PR #5737, opened 2017-09-14
fix issue 6895 - isCovariantWith doesn't work for function pointer and delegate

BBasile:
This is mostly a copy and paste of a patch sent to the bug tracker and I'm not 100% convinced by the validity of the issue.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -4869,7 +4869,9 @@ Determines whether the function type $(D F) is covariant with $(D G), i.e.,
 functions of the type $(D F) can override ones of the type $(D G).
  */
 template isCovariantWith(F, G)
-    if (is(F == function) && is(G == function))
+    if ((is(F == function) && is(G == function)) ||
+        (is(F == delegate) && is(G == delegate)) ||
+        (isFunctionPointer!F && isFunctionPointer!G))
 {
     static if (is(F : G))
         enum isCovariantWith = true;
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -5057,6 +5059,29 @@ template isCovariantWith(F, G)
...
+
+    static assert(isCovariantWith!(typeof(derived_function), typeof(derived_function)));
+    static assert(isCovariantWith!(typeof(base_delegate), typeof(base_delegate)));
+
+    static assert(isCovariantWith!(typeof(derived_function), typeof(base_function)));
+    static assert(isCovariantWith!(typeof(*derived_function), typeof(*base_function)));
+    static assert(isCovariantWith!(typeof(derived_delegate), typeof(base_delegate)));
+
+    static assert(!isCovariantWith!(typeof(invalid_function), typeof(base_function)));
+    static assert(!isCovariantWith!(typeof(invalid_delegate), typeof(base_delegate)));
```
wilzbach: One could check the function - delegate combinations too

std/traits.d
```diff
@@ -4869,7 +4869,9 @@ Determines whether the function type $(D F) is covariant with $(D G), i.e.,
 functions of the type $(D F) can override ones of the type $(D G).
  */
 template isCovariantWith(F, G)
-    if (is(F == function) && is(G == function))
+    if ((is(F == function) && is(G == function)) ||
```
MetaLang: Is there a reason you didn't use [isSomeFunction](https://dlang.org/library/std/traits/is_some_function.html) here?

BBasile: I can't say. It 's an old patch found on bugzilla.

BBasile: That being said it seems obvious that the author wanted `F` and `G` to have the same `function membership type` (how to name that ?), so that covariance couldn't be tested with a pointer to static member function and a delegate.

BBasile:
It seems that nobody's really confident enough w/ the concept of covariance. After a quik look at the dedicated WP page it seems that it would be better not to blend everything: covariance in OO and covariance of functions.

PR #5738, opened 2017-09-15
Fix issue 9591 - Add ApplyWith template

schveiguy:
Bikeshedding for `ApplyWith` name welcome. If you look at the example, it reads well, but maybe not when specified on its own.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -1328,7 +1328,7 @@ private template SmartAlias(T...)
   *    Alias to instantiated template passed as last parameter, with other
   *    parameters used for instantiation.
   */
-template ApplyWith(Args...) if(Args.length >= 1)
+template ApplyWith(Args...) if (Args.length >= 1)
 {
     alias ApplyWith = Instantiate!(Args[$-1], Args[0 .. $-1]);
 }
@@ -1340,7 +1340,7 @@ template ApplyWith(Args...) if(Args.length >= 1)
     alias functions = staticMap!(ApplyLeft!(ApplyWith, string),
                                  leftJustify, center, rightJustify);
     string result = "";
-    static foreach(f; functions)
+    static foreach (f; functions)
     {
         {
             auto x = &f; // not a template, but a function instantiation
--- a/std/meta.d
+++ b/std/meta.d
@@ -1313,31 +1313,45 @@ private template SmartAlias(T...)
     })));
 }
 
+// used to actually instantiate the ApplyWith template. The last parameter is
+// the template to instantiation, while the other parameters are the ones to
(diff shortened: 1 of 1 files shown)


std/meta.d
```diff
@@ -1313,6 +1314,45 @@ private template SmartAlias(T...)
...
+}
+
+///
+@safe unittest
+{
+    import std.string: leftJustify, center, rightJustify;
+    alias functions = staticMap!(ApplyLeft!(ApplyWith, string),
+                                 leftJustify, center, rightJustify);
+    string result = "";
+    static foreach(f; functions)
```
wilzbach: Spacing after `foreach` applies to static foreach too (that's why CircleCi is failing)

schveiguy:
> Spacing

I will never be trained at this point to do that spacing properly... :)

Updated.

std/meta.d
```diff
@@ -1313,6 +1314,45 @@ private template SmartAlias(T...)
 }
 
 /**
+  * This template instantiates a template by taking its parameters first, and
+  * the template last. This can be useful when using staticMap to apply various
+  * templates to the same set of parameters, since staticMap always appends the
+  * list of arguments to one template.
+  *
+  * `ApplyWith!(A0, A1, ..., An-1, An)` is equivalent to
+  * `An!(A0, A1, ..., An-1)`
```
edi33416: I kept looking at this and I might be wrong, but I can't stop wondering if it wouldn't be more natural for the call to be `A0!(A1, ..., An)` ?
If so, maybe we could have the template declaration as `ApplyWith(Func, Args...)` ?
What do you think?

schveiguy: I get what you are saying. By itself, `ApplyWith` (or whatever we should call it) looks weird when you add the template at the end. The whole point of this template is to work with `ApplyLeft`. It's like specifying all the parameters, and leaving the actual template to be used up to a later time.

For example:

```D
template T1(T) {}
template T2(T) {}

alias X = ApplyLeft!(ApplyWith!string);

alias a = X!T1; // equivalent to T1!string
alias b = X!T2; // equivalent to T2!string
```

I'm thinking now, since pretty much `ApplyLeft` is required to make any sense, `ApplyWith` should use it internally instead of `Instantiate`. The fact that it's used can be hidden.

schveiguy:
OK, I redesigned `ApplyWith` so it is much less confusing. Thanks @edi33416 for the review, please take a look at the new design.

schveiguy:
@klickverbot Yes, I didn't think of that mechanism. However, `Instantiate` is private. Maybe I can just close this and make a PR to make it public?

schveiguy:
Opened said PR: #5739

PR #5739, opened 2017-09-17
Fix issue 9591 - Allow Instantiate template to be public

schveiguy:
This solves the problem in the referenced issue. I've added the solution as a documented unit test for `Instantiate`.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -54,6 +54,7 @@
  * $(TR $(TD Template instantiation) $(TD
  *           $(LREF ApplyLeft)
  *           $(LREF ApplyRight)
+ *           $(LREF Instantiate)
  * ))
  * ))
  *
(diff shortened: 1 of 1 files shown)


schveiguy:
ping @wilzbach I simply reused my branch and opened a new PR, but dlang-bot didn't notice?

wilzbach:
> ping @wilzbach I simply reused my branch and opened a new PR, but dlang-bot didn't notice?

Don't worry about it. The essential things are run in a loop and run every five minutes. It's probably an issue with the GitHub API, but I am currently on the way for a week long vacation, so I sadly can't check the logs to be sure. Thanks for the ping!

schveiguy:
Ah, @wilzbach it's because I squashed the original branch, so must be dlang-bot didn't think there were changes (or github didn't notify?).

Anyway, seems good now, I had to fix the docs a bit and the bot showed up.

wilzbach:
Shouldn't this have been formally approved by @andralex? (after all it's a "new" public symbol)

atilaneves:
@wilzbach Good question. I'm new around these parts.

wilzbach:
> @wilzbach Good question. I'm new around these parts.

No worries - it was a rhetorical one ;-)
Every new public symbol needs to be approved by @andralex
It's probably not a big deal for this one, maybe quickly pinging him on slack is enough?

schveiguy:
It's newly public, but not new. I don't speak for Andrei, but I think the naming of the symbol is correct -- I can't imagine a better name is possible. But the question to answer is, does this solve a big enough problem such that it's worthy of a new symbol? Given the problem it solves in the context of `staticMap`, I think it is. It's also needed any time you are finding templates inside an `AliasSeq` that you need to instantiate. Clearly it was needed inside `std.meta` as well.

andralex:
I'm not thrilled about this. A compelling example should be found, if not I think we shouldn't add this kind of subtle artifacts of very limited usefulness.

schveiguy:
This is as essential as `AliasSeq`. Before it existed (as `TypeTuple` before), one had to always make a simple declaration in their file where they use it. So much easier when everyone uses the same name and meaning.

`Instantiate` is the same thing -- it's obvious, easy, but something you just have to declare somewhere, so it might as well go in a central place.

I can't think of a more compelling example besides trying to apply a bunch of similar templates to the same parameters.

All you have to do is look at all the simple templates in this file to see things that are easy to declare, but are needed many places, so this module collects them. `AliasSeq`, `Alias`, `templateNot`, etc. Many of the things in here are one-liners.

Also note that `Instantiate` was already used internally, so I didn't add anything. I just moved it to a public area and documented it.

andralex:
@schveiguy what would a good standalone example not necessitating much additional information?

andralex:
Thanks for the due diligence @klickverbot!

jmdavis:
Template lambdas would be awesome. The more you do template metaprogramming, the more painful it seems that they aren't there. I have no idea what a reasonable syntax for them would be though, which is part of why I've never really tried to push the idea.

ntrel:
This is missing a changelog entry.

PR #5741, opened 2017-09-19
Fix Issue 17832 - std.random.choice cannot be used with other random generators

dunkyp:
Random choice default argument was deciding the type of the
template. I'm not sure if this is a deeper bug, but it can easily be
fixed with this patch which simply overloads the function.
Also added a unittest which would have failed under the old code.

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -2018,13 +2018,19 @@ Returns:
     a copy.
  */
 auto ref choice(Range, RandomGen = Random)(auto ref Range range,
-                                           ref RandomGen urng = rndGen)
+                                           ref RandomGen urng)
 if (isRandomAccessRange!Range && hasLength!Range && isUniformRNG!RandomGen)
 {
     assert(range.length > 0,
            __PRETTY_FUNCTION__ ~ ": invalid Range supplied. Range cannot be empty");
 
     return range[uniform(size_t(0), $, urng)];
+}
+
+/// ditto
+auto ref choice(Range)(auto ref Range range)
+{
+    return choice(range, rndGen);
 }
 
 ///
@@ -2041,6 +2047,10 @@ if (isRandomAccessRange!Range && hasLength!Range && isUniformRNG!RandomGen)
     auto urng = Random(unpredictableSeed);
     elem = choice(array, urng);
 
+    assert(canFind(array, elem),
+           "Choice did not return a valid element from the given Range");
+    auto rng2 = Xorshift();
+    elem = choice(array, rng2);
     assert(canFind(array, elem),
            "Choice did not return a valid element from the given Range");
 }


quickfur:
Found underlying cause: https://issues.dlang.org/show_bug.cgi?id=16467

Basically, the problem is that a template function like `void func(T)(T t = 123)` is just shorthand for:
````
template func(T)
{
    void func(T t = 123) { ... }
}
````

The problem is that during IFTI of a call like `func("abc")`, `T` is found to be `string`, which makes the declaration equivalent to `void func(string t = 123)`, which is invalid.  I'm not 100% sure what happens after that in the compiler, but I suspect something along the lines of a compilation error in a template being gagged and the compiler moving on to the next possible overload / instantiation or whatever, which eventually ends with it trying to shoehorn a `string` into a `func!int` overload, causing it to abort with a compile error.

One possible solution is to change the declaration to:

````
template func(T)
{
    static if (is(T == typeof(123)))
        void func(T t = 123) { ... }
    else
        void func(T t) { ... }
}
````

Since this is a lot of ugly boilerplate, I'm tempted to say the compiler should perform this lowering automatically. However, that unfortunately breaks the valid case where you *do* want the default parameter to apply across different types, for example:

````
void func(T)(T t = 123) if (isIntegral!T) { ... }
````

In this case, the default argument could arguably(!) be construed to impose a constraint that `T` must be a type for which `123` is a valid value.

So, unfortunately, it seems that the workaround in this PR is probably a necessary one.

**Comment by MetaLang** [comment]
Thanks for the explanation. When you put it that way, this is definitely a fairly subtle bug in the code and needs to be fixed.

**Comment by quickfur** [comment]
ping @wilzbach 

Does the above address your concerns? Should we merge this? It's been rotting here for far too long already.

**Comment by quickfur** [comment]
ping @wilzbach  I'm in favor of merging this, because the bug is real, and needs to be addressed somehow. The fact that Phobos looks a little uglier to me seems secondary, and something that we can work on afterwards.  I've thought of an idea that might work, but it will probably involve a DIP, so I'd much rather get this fix in now rather than later.  What do you think?

File: std/random.d [hunk 1]
```diff
@@ -2008,6 +2008,11 @@ if (isRandomAccessRange!Range && hasLength!Range && isUniformRNG!RandomGen)
     return range[uniform(size_t(0), $, urng)];
 }
 
+auto ref choice(Range)(auto ref Range range)
```
wilzbach: Should be `/// ditto` - this tells the Ddoc engine to visually group the overloads.

wilzbach:
Sorry getting too many GH lately. Feel free to ping me on Slack for something that requires immediate attention.

1) I agree that the bug is real and needs to be fixed. _All_ other methods in `std.random` are probably affected by this too :scream: 

2) I'm still not a huge friend of uglifying Phobos to workaround compiler bugs / implementation problems, but that sadly seems to be a common motive this month :/

3) Let's look closer at the compiler bug / limitation

Here's a reduction:

```d
struct Foo {}
struct Bar {}
Foo foo;
void myFun(D)(D d = foo){}

void bar()
{
    myFun(); // works
    myFun(Bar()); // error
}
```

```
> dmd -c foo.d
foo.d(6): Error: cannot implicitly convert expression foo of type Foo to Bar
foo.d(11): Error: template instance foo.myFun!(Bar) error instantiating
```

As you might have guessed, it's a known bug:
https://issues.dlang.org/show_bug.cgi?id=17036

There's also the pre-approved https://issues.dlang.org/show_bug.cgi?id=17186 which even had a PR (https://github.com/dlang/dmd/pull/2270).
It could fix this problem by allowing `auto d = foo`.

4) > I've thought of an idea that might work, but it will probably involve a DIP

Without knowing your idea, I can say the following: "you might be luckily and it could be considered as a bug fix". Remember that 17186 is pre-approved.

5) >  I'm in favor of merging this, because the bug is real

Yeah, fair enough, but let's fix entire `std.random` then (doesn't have to be this PR)?
Thanks for pushing this!

wilzbach:
> Should be `/// ditto` - this tells the Ddoc engine to visually group the overloads.

I quickly fixed this myself.

quickfur:
Is the CircleCI failure relevant? All I can glean from the output is that some built-in timeout was exceeded while running tests. Doesn't seem related to this PR directly?

wilzbach:
It has to do with something else, but so far I haven't had time to dig deeper into it.

PR #5742, opened 2017-09-20
Fix Issue 17844 - std.process.execute should allow not capturing stderr

CyberShadow:


CyberShadow:
Picking a good name for this flag is a bit tricky, because there is a `retainStderr` flag with a different meaning. Not too happy with `passStderr` because of the consecutive `s` letters.

CyberShadow:
> In hindsight, stderr should not have been automatically redirected without being told to do so.

Agreed.

> When I first read passStderr without ingesting the whole change and bug report, I thought it was a way to pass a specialized stderr to the child. Really what it should be is passThroughStderr, or maybe ignoreStderr. What you are doing is leaving the stderr setup alone.

Interesting you should mention those name suggestions. My initial idea was `ignoreStderr`, however I thought that one obvious interpretation of that name implies that anything written to `stderr` would be completely discarded. My second idea was `passThruStderr`, similar to [the PHP `passthru` function](http://php.net/manual/en/function.passthru.php), but I thought it was too long, and the function's effect in PHP when processing a HTTP request (PHP's most common execution context) is rather different from how stderr is typically encountered by D programs.

We can go with `passThroughStderr` I guess, unless there are better ideas?

schveiguy:
`passThruStderr` is shorter, go with that.

CyberShadow:
> `thru` is a "chiefly North American informal spelling"

Well, there's industry precedent with `passthru`.

> in an infrequently used option

Did you read the above discussion and bug report? If we could make it the default, we would.

schveiguy:
It's OK to use `passThroughStderr` instead, but I disagree with the objections. The question to answer is: can `passThru` mean something other than "pass through"? If not, it's not surprising or confusing. It's just a standard quirky abbreviation that programming libraries and languages are full of.

schveiguy:
> The situation where it is confusing is not when reading code, but when writing it. You are looking for a parameter you probably remembered as "pass through stderr".

I don't think I would remember either name, even if I'm writing it often.

Either way, it's a poor way to name it. It should really be the opposite (opt-in to receiving stderr output, not opt-out).

I think we've spent too much time already debating it, just pick one and merge.

CyberShadow:
> That's not frequent enough to try and be cute

Nobody is trying to be cute.

> Just because you got talked into changing it too readily doesn't somehow make it the right thing to do.

Nice attack on my character drawn out of an assumption. 

I have limited time and patience to go back-and-forth on the same nitpicking matters, as should everyone.

> The spelling "pass-thru" isn't in Merriam-Webster, while "pass-through" is.

Neither is `writeln` or many other names in Phobos. We use them because they are common in programming.

Meanwhile, 'passthrough` is not a function in PHP, while `passthru` is.

> That implication is fallacious.

So was much of your post. It was very frustrating to read.

Since you feel so strongly about the current name, I am going to change the name to `stderrPassThrough` (to also disambiguate from passing something through `stderr`). I suggest that you use your time to contribute more constructively in the future, though.

schveiguy:
I'm assuming the name is good now.

kyllingstad:
I realise that I am somewhat late in reviewing this now (sorry!), but I would like to add my comments anyway. And the first thing that comes to mind is that this actually changes something fundamental about `Config`: It is no longer just a set of flags that control the behaviour of `spawnProcess()` and `spawnShell()`; now we have a flag which uniquely affects `execute()` and `executeShell()`. Is this OK? (I'm not saying it isn't, just that it is worth some thought.)

Assuming it is, my second comment is that this change should be reflected in the main `Config` documentation, which currently specifies that the flags apply to `spawnProcess()` and `spawnShell()`.

Thirdly, the documentation for the new flag should then specify that the flag actually has *no* effect on `spawnProcess()` and `spawnShell()`.

Again, sorry to be Johnny-come-lately here.

schveiguy:
All excellent points. I think the reason for using `Config` is because it's already being passed to `execute` and `executeShell`. Adding a new set of `Config` flags just for those functions would be detrimental and confusing. Adding a simple bool flag at the end (or `Flag`) would be ugly too.

In reality, the best way to have fixed this (in hindsight) is to not have the automatic redirection of stderr to stdout, and let the user specify it in the config in the first place. But it's hard to undo that at this point.

We should definitely update the docs to reflect the special nature of the new flag.

CyberShadow:
Thanks Lars, good suggestions. I'll send a follow-up PR.

CyberShadow:
-> https://github.com/dlang/phobos/pull/5744

PR #5743, opened 2017-09-20
Fix Issue 17847 - Properly sanitize seeds for Park–Miller engines

n8sh:
As in XorshiftEngine, sanitizing the seed is more useful than throwing
an exception. The original check was broken because it checked the seed
before taking the modulus. Making this change allows unpredictableSeed
to be marked nothrow and @nogc.

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -362,7 +362,7 @@ The parameters of this distribution. The random number is $(D_PARAM x
 Constructs a $(D_PARAM LinearCongruentialEngine) generator seeded with
 $(D x0).
  */
-    this(UIntType x0) @safe pure
+    this(UIntType x0) @safe pure nothrow @nogc
     {
         seed(x0);
     }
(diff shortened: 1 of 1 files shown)


DmitryOlshansky:
@n8sh For the benefit of relese note please file a short enhancement/bug report in

issue.dlang.org

And add Fix issue xxx in your commit message

Sorry for this busy-work.

std/random.d
```diff
@@ -1291,7 +1299,7 @@ random number sequences every run.
 Returns:
 A single unsigned integer seed value, different on each successive call
 */
-@property uint unpredictableSeed() @trusted
+@property uint unpredictableSeed() @trusted nothrow @nogc
```
wilzbach: FYI: there is a pending PR and interest to improve the seeding in Phobos. This is usually done by querying the OS APIs which can throw errors, but for me it makes sense to have this function as `nothrow` - especially because we can always fallback to this "naive"  seeding variant in case the improvement ever lands in Phobos.

PR #5745, opened 2017-09-24
Fix Issue 6447: iota(BigInt) too

Biotronic:
This adds the missing overload for three-parameter iota as pointed out in [Issue 6447](https://issues.dlang.org/show_bug.cgi?id=6447), and should be enough to finally close the issue.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
     ---
 */
 auto iota(B, E, S)(B begin, E end, S step)
-if ((isIntegral!(CommonType!(B, E)) || isPointer!(CommonType!(B, E)))
-        && isIntegral!S)
 {
-    import std.conv : unsigned;
-
-    alias Value = CommonType!(Unqual!B, Unqual!E);
-    alias StepType = Unqual!S;
-
-    assert(step != 0 || begin == end);
-
-    static struct Result
+    static if (!is(B == Unqual!B) || !is(E == Unqual!E) || !is(S == Unqual!S))
     {
-        private Value current, last;
-        private StepType step; // by convention, 0 if range is empty
-
-        this(Value current, Value pastLast, StepType step)
-        {
-            if (current < pastLast && step > 0)
-            {
-                // Iterating upward
-                assert(unsigned((pastLast - current) / step) <= size_t.max);
-                // Cast below can't fail because current < pastLast
-                this.last = cast(Value) (pastLast - 1);
-                this.last -= unsigned(this.last - current) % step;
-            }
-            else if (current > pastLast && step < 0)
-            {
-                // Iterating downward
-                assert(unsigned((current - pastLast) / (0 - step)) <= size_t.max);
-                // Cast below can't fail because current > pastLast
-                this.last = cast(Value) (pastLast + 1);
-                this.last += unsigned(current - this.last) % (0 - step);
-            }
-            else
-            {
-                // Initialize an empty range
-                this.step = 0;
-                return;
-            }
-            this.step = step;
-            this.current = current;
-        }
-
-        @property bool empty() const { return step == 0; }
-        @property inout(Value) front() inout { assert(!empty); return current; }
-        void popFront()
-        {
-            assert(!empty);
-            if (current == last) step = 0;
-            else current += step;
-        }
-
-        @property inout(Value) back() inout
-        {
-            assert(!empty);
-            return last;
-        }
-        void popBack()
-        {
-            assert(!empty);
-            if (current == last) step = 0;
-            else last -= step;
-        }
-
-        @property auto save() { return this; }
-
-        inout(Value) opIndex(ulong n) inout
-        {
-            assert(n < this.length);
-
-            // Just cast to Value here because doing so gives overflow behavior
-            // consistent with calling popFront() n times.
-            return cast(inout Value) (current + step * n);
-        }
-        auto opBinaryRight(string op)(Value val) const
-        if (op == "in")
-        {
-            if (empty) return false;
-            //cast to avoid becoming unsigned
-            auto supposedIndex = cast(StepType)(val - current) / step;
-            return supposedIndex < length && supposedIndex * step + current == val;
-        }
-        auto contains(Value x){return x in this;}
-        inout(Result) opSlice() inout { return this; }
-        inout(Result) opSlice(ulong lower, ulong upper) inout
-        {
-            assert(upper >= lower && upper <= this.length);
-
-            return cast(inout Result) Result(
-                cast(Value)(current + lower * step),
-                cast(Value)(current + upper * step),
-                step);
-        }
-        @property size_t length() const
-        {
-            if (step > 0)
-                return 1 + cast(size_t) (unsigned(last - current) / step);
-            if (step < 0)
-                return 1 + cast(size_t) (unsigned(current - last) / (0 - step));
-            return 0;
-        }
-
-        alias opDollar = length;
+        return iota(Unqual!B(begin), Unqual!E(end), Unqual!S(step));
     }
-
-    return Result(begin, end, step);
+    else
+    {
+        return IotaResult!(B, E, S)(begin, end, step);
+    }
 }
 
 /// Ditto
 auto iota(B, E)(B begin, E end)
-if (isFloatingPoint!(CommonType!(B, E)))
 {
-    return iota(begin, end, CommonType!(B, E)(1));
-}
-
-/// Ditto
-auto iota(B, E)(B begin, E end)
-if (isIntegral!(CommonType!(B, E)) || isPointer!(CommonType!(B, E)))
-{
-    import std.conv : unsigned;
-
-    alias Value = CommonType!(Unqual!B, Unqual!E);
-
-    static struct Result
+    static if (!is(B == Unqual!B) || !is(E == Unqual!E))
     {
-        private Value current, pastLast;
-
-        this(Value current, Value pastLast)
-        {
-            if (current < pastLast)
-            {
-                assert(unsigned(pastLast - current) <= size_t.max);
-
-                this.current = current;
-                this.pastLast = pastLast;
-            }
-            else
-            {
-                // Initialize an empty range
-                this.current = this.pastLast = current;
-            }
-        }
-
-        @property bool empty() const { return current == pastLast; }
-        @property inout(Value) front() inout { assert(!empty); return current; }
-        void popFront() { assert(!empty); ++current; }
-
-        @property inout(Value) back() inout { assert(!empty); return cast(inout(Value))(pastLast - 1); }
-        void popBack() { assert(!empty); --pastLast; }
-
-        @property auto save() { return this; }
-
-        inout(Value) opIndex(size_t n) inout
-        {
-            assert(n < this.length);
-
-            // Just cast to Value here because doing so gives overflow behavior
-            // consistent with calling popFront() n times.
-            return cast(inout Value) (current + n);
-        }
-        auto opBinaryRight(string op)(Value val) const
-        if (op == "in")
-        {
-            return current <= val && val < pastLast;
-        }
-        auto contains(Value x){return x in this;}
-        inout(Result) opSlice() inout { return this; }
-        inout(Result) opSlice(ulong lower, ulong upper) inout
-        {
-            assert(upper >= lower && upper <= this.length);
-
-            return cast(inout Result) Result(cast(Value)(current + lower),
-                                            cast(Value)(pastLast - (length - upper)));
-        }
-        @property size_t length() const
-        {
-            return cast(size_t)(pastLast - current);
-        }
-
-        alias opDollar = length;
+        return iota(Unqual!B(begin), Unqual!E(end));
     }
-
-    return Result(begin, end);
+    else static if (is(typeof(begin + B(1))))
+    {
+        return iota(begin, end, B(1));
+    }
+    else static if (is(typeof(++begin)) && !isPointer!B)
+    {
+        return IotaResult!(B, E)(begin, end);
+    }
+    else static if (is(typeof(begin + 1)))
+    {
+        return iota(begin, end, 1);
+    }
+    else
+    {
+        static assert(false, "iota does not support the types ("~B.stringof~", "~E.stringof~").");
+    }
 }
 
 /// Ditto
 auto iota(E)(E end)
-if (is(typeof(iota(E(0), end))))
+if (is(typeof(E(0))))
 {
-    E begin = E(0);
-    return iota(begin, end);
+    static if (is(E == Unqual!E))
+        return iota(E(0), end);
+    else
+        return iota(Unqual!E(end));
 }
 
-/// Ditto
-// Specialization for floating-point types
-auto iota(B, E, S)(B begin, E end, S step)
-if (isFloatingPoint!(CommonType!(B, E, S)))
-in
+/*
+ * Overload for anything supporting regular arithmetic.
+ */
+struct IotaResult(B, E, S)
+if (!anySatisfy!(isFloatingPoint, B, E, S) && is(typeof(B.init + S.init)))
 {
-    assert(step != 0, "iota: step must not be 0");
-    assert((end - begin) / step >= 0, "iota: incorrect startup parameters");
-}
-do
-{
-    alias Value = Unqual!(CommonType!(B, E, S));
-    static struct Result
+    static if (is(CommonType!(B, E, S)) && !is(CommonType!(B, E, S) == void))
+        alias ValueType = CommonType!(B, E, S);
+    else static if (is(typeof(B.init + E.init + S.init)))
+        alias ValueType = typeof(B.init + E.init + S.init);
+    else static if (is(typeof(B.init + S.init)))
+        alias ValueType = typeof(B.init + S.init);
+    else
+        static assert(false, "iota does not support types that cannot be added or incremented.");
+
+    ValueType begin;
+    ValueType end;
+    S step;
+
+    private auto unsigned(T)(T value)
     {
-        private Value start, step;
-        private size_t index, count;
-
-        this(Value start, Value end, Value step)
-        {
-            import std.conv : to;
-
-            this.start = start;
-            this.step = step;
-            immutable fcount = (end - start) / step;
-            count = to!size_t(fcount);
-            auto pastEnd = start + count * step;
-            if (step > 0)
-            {
-                if (pastEnd < end) ++count;
-                assert(start + count * step >= end);
-            }
-            else
-            {
-                if (pastEnd > end) ++count;
-                assert(start + count * step <= end);
-            }
-        }
-
-        @property bool empty() const { return index == count; }
-        @property Value front() const { assert(!empty); return start + step * index; }
-        void popFront()
-        {
-            assert(!empty);
-            ++index;
-        }
-        @property Value back() const
-        {
-            assert(!empty);
-            return start + step * (count - 1);
-        }
-        void popBack()
-        {
-            assert(!empty);
-            --count;
-        }
-
-        @property auto save() { return this; }
-
-        Value opIndex(size_t n) const
-        {
-            assert(n < count);
-            return start + step * (n + index);
-        }
-        inout(Result) opSlice() inout
-        {
-            return this;
-        }
-        inout(Result) opSlice(size_t lower, size_t upper) inout
-        {
-            assert(upper >= lower && upper <= count);
-
-            Result ret = this;
-            ret.index += lower;
-            ret.count = upper - lower + ret.index;
-            return cast(inout Result) ret;
-        }
-        @property size_t length() const
-        {
-            return count - index;
-        }
-
-        alias opDollar = length;
+        static if (is(Unsigned!T)) return cast(Unsigned!T) value;
+        else return value;
     }
 
-    return Result(begin, end, step);
+    this(B b, E e, S s)
+    {
+        step = 0;
+        if (s == 0) return;
+        if (b < e && s < 0) return;
+        if (b > e && s > 0) return;
+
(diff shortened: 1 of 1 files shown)


std/range/package.d
```diff
@@ -5898,6 +5898,63 @@ if (!isIntegral!(CommonType!(B, E)) &&
...
 
+/// ditto
+auto iota(B, E, S)(B begin, E end, S step)
+if (!isIntegral!(CommonType!(B, E)) &&
+    !isFloatingPoint!(CommonType!(B, E)) &&
+    !isPointer!(CommonType!(B, E)) &&
+    is(typeof((ref B b, S s) { b += s; })) &&
+    is(typeof(S.init == 0)) &&
+    is(typeof(S.init < 0)) &&
+    is(typeof(B.init < E.init)))
```
wilzbach: Could we unite the overloads, have only one function and use static if internally so that a user isn't hit by multiple giant blocks?

std/range/package.d
```diff
@@ -5898,6 +5898,63 @@ if (!isIntegral!(CommonType!(B, E)) &&
...
+            assert(!empty);
+            current += step;
+        }
+    }
+    return Result(begin, end, step);
+}
+
+@system unittest
+{
+    import std.bigint;
```
wilzbach: Phobos style: use selective imports whenever possible

std/range/package.d
```diff
@@ -5898,6 +5898,63 @@ if (!isIntegral!(CommonType!(B, E)) &&
...
+    auto a = iota(BigInt(1), BigInt(100), BigInt(5));
+    auto b = iota(BigInt(1), BigInt(100), 5);
+    auto c = iota(BigInt(100), BigInt(1), -5);
+    assert(equal(a, b));
+    assert(equal(a, iota(1, 100, 5)));
+    assert(equal(c, iota(100, 1, -5)));
+}
```
wilzbach: Shouldn't we test what happens for very big numbers (aka over ulong)?

std/range/package.d
```diff
@@ -5898,6 +5898,63 @@ if (!isIntegral!(CommonType!(B, E)) &&
...
+    return Result(begin, end, step);
+}
+
+@system unittest
+{
+    import std.bigint;
+    auto a = iota(BigInt(1), BigInt(100), BigInt(5));
+    auto b = iota(BigInt(1), BigInt(100), 5);
+    auto c = iota(BigInt(100), BigInt(1), -5);
+    assert(equal(a, b));
```
wilzbach: You need to import `equal` from `std.algorithm.comparison`

std/range/package.d
```diff
@@ -5898,6 +5898,63 @@ if (!isIntegral!(CommonType!(B, E)) &&
...
+        B current;
+        E end;
+        S step;
+
+        this(B current, E end, S step)
+        {
+            this.current = current;
+            this.end = end;
+            this.step = step;
+            if (step == 0 || (current < end) != (step > 0))
```
wilzbach: Your tests don't cover all cases

Biotronic: Should be fixed now.

Biotronic:
As @wilzbach suggested, I united all iota overloads. This has the benefit of uniting functionality - where previously different overloads of iota supported different operations, now these operations are available on any iota overload whose parameter types support the necessary operations.

DmitryOlshansky:
Autotester is not happy(

Biotronic:
Believe me, I know. I've failed to get some of the unittests to run on my local machine, and the specific one that fails now works when I copy the code to a separate file.

MetaLang:
I'd suggest we leave uniting the overloads to a separate PR and get the fix merged.

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
     }
     ---
 */
+auto iota(E)(E end)
+{
+    static if (is(typeof(end += 1)))
+        return IotaResult!(Unqual!E, Unqual!E, int)(E(0), end, 1);
+    else
+        return IotaResult!(Unqual!E, Unqual!E, void)(E(0), end);
```
andralex: this code is duplicated, should just forward to `iota(E(0), end)`

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
     }
     ---
 */
+auto iota(E)(E end)
+{
+    static if (is(typeof(end += 1)))
```
andralex: This is not good. We should not keep unnecessarily 1 as state and lose the benefit of simpler and faster `++` as opposed to `+= variable`.

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
-if ((isIntegral!(CommonType!(B, E)) || isPointer!(CommonType!(B, E)))
-        && isIntegral!S)
 {
-    import std.conv : unsigned;
+    return IotaResult!(Unqual!B, Unqual!E, Unqual!S)(begin, end, step);
+}
+
+struct IotaResult(B, E, S)
+{
+    alias ValueType = CommonType!(B,E);
```
andralex: space after comma

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
 
-    assert(step != 0 || begin == end);
+    static if (isFloatingPoint!ValueType)
+        size_t count, index;
+    static if (hasStep)
+        S step;
 
-    static struct Result
+    enum hasStep = !is(S == void);
+    enum hasLength = is(typeof((end - current) / step) : size_t);
```
andralex: Such properties should migrate to the top of the type so as to make code easier to follow.

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
-                assert(unsigned((current - pastLast) / (0 - step)) <= size_t.max);
-                // Cast below can't fail because current > pastLast
-                this.last = cast(Value) (pastLast + 1);
-                this.last += unsigned(current - this.last) % (0 - step);
+                this.step = step;
+                this.end = end;
             }
             else
             {
-                // Initialize an empty range
```
andralex: Why was this comment lost?

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
+        else static if (!is(typeof(current < end)))
+            return current == end;
+        else
         {
-            assert(!empty);
-            if (current == last) step = 0;
-            else current += step;
+            if (step > 0)
+                return !(current < end);
+            else if (step < 0)
```
andralex: redundant else

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
+                return !(current > end);
+            else
```
andralex: redundant else

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
-        {
-            assert(!empty);
-            return last;
-        }
-        void popBack()
+    @property
+    bool empty()() const
+    if (is(typeof((const B b, const E e) => b < e)) ||
+        is(typeof((const B b, const E e) => b == e)))
+    {
```
andralex: Why is this overload necessary? A comment would be in order.

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
+        else static if (!is(typeof(current < end)))
+            return current == end;
+        else
         {
-            assert(!empty);
-            if (current == last) step = 0;
-            else last -= step;
+            if (step > 0)
+                return !(current < end);
+            else if (step < 0)
```
andralex: redundant else

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
+                return !(current > end);
+            else
```
andralex: redundant else

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
+            ++index;
+        else static if (hasStep)
         {
-            if (step > 0)
-                return 1 + cast(size_t) (unsigned(last - current) / step);
-            if (step < 0)
-                return 1 + cast(size_t) (unsigned(current - last) / (0 - step));
-            return 0;
+            if ((step > 0 && current + step > end) ||
+                (step < 0 && current + step < end))
```
andralex: one more indent

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
+    {
+        return this;
+    }
 
-/// Ditto
-auto iota(B, E)(B begin, E end)
-if (isIntegral!(CommonType!(B, E)) || isPointer!(CommonType!(B, E)))
-{
-    import std.conv : unsigned;
+    auto opSlice() const
```
andralex: does inout work here?

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
+    {
+        return this;
+    }
 
-    alias Value = CommonType!(Unqual!B, Unqual!E);
+    enum canMultiply(T) = is(typeof((B b, E e, S s, T t){ return b + t * s ; }));
+    enum canDivide(T) = is(typeof((B b, E e, S s, T t){ return (e - b)/t ; }));
```
andralex: whitespace

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
+        size_t length() const
         {
-            return current <= val && val < pastLast;
+            static if (isFloatingPoint!ValueType)
+                return count - index;
+            else
+            {
+                if (step > 0)
+                    return 1 + cast(size_t) (cast(size_t)(end - current -1) / step);
+                else if (step < 0)
```
andralex: redundant else

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
+                    return 1 + cast(size_t) (cast(size_t)(current - end -1) / (0 - step));
+                else
```
andralex: redundant else

std/range/package.d
```diff
@@ -5297,285 +5297,279 @@ auto sequence(alias fun, State...)(State args)
...
-        inout(Result) opSlice() inout { return this; }
-        inout(Result) opSlice(ulong lower, ulong upper) inout
-        {
-            assert(upper >= lower && upper <= this.length);
 
-            return cast(inout Result) Result(cast(Value)(current + lower),
-                                            cast(Value)(pastLast - (length - upper)));
-        }
-        @property size_t length() const
+        auto opSlice(I1, I2)(I1 idx1, I2 idx2)
```
andralex: This should only take `size_t` in both positions.

Biotronic:
After repeated failed attempts at getting the unit tests to run, I decided to revert to the simple change that fixes 6447, and a greater rewrite will have to wait.

std/range/package.d
```diff
@@ -5898,6 +5898,76 @@ if (!isIntegral!(CommonType!(B, E)) &&
...
+        }
+
+        @property bool empty()
+        {
+            if (step > 0)
+                return !(current < end);
+            else if (step < 0)
+                return !(current > end);
+            else
+                return true;
```
JackStouffer:
```d
if (step > 0)
    return !(current < end);
if (step < 0)
    return !(current > end);

return true;
```

std/range/package.d
```diff
@@ -5898,6 +5898,76 @@ if (!isIntegral!(CommonType!(B, E)) &&
...
+        }
+        @property auto front() { return current; }
+        void popFront()
+        {
+            assert(!empty);
```
JackStouffer: Needs an assert message

std/range/package.d
```diff
@@ -5898,6 +5898,76 @@ if (!isIntegral!(CommonType!(B, E)) &&
...
+        @property bool empty()
+        {
+            if (step > 0)
+                return !(current < end);
+            else if (step < 0)
+                return !(current > end);
+            else
+                return true;
+        }
+        @property auto front() { return current; }
```
JackStouffer: add assert not empty with a message

JackStouffer:
Ping @andralex

std/range/package.d
andralex: Yes, that would be great. Just one function that accepts the union of all cases, and distinguish inside as needed using `static if` please.

quickfur: Yes, please reduce unnecessary user-facing overloads, unify into an API that makes sense from the user's POV.  Internally, either use static ifs inside the function body, or module-private overloads (e.g., `iotaImpl`) for dispatching to different specializations.

std/range/package.d
```diff
@@ -5665,16 +5768,6 @@ pure nothrow @nogc @safe unittest
     assert(iota(ssize_t.max, ssize_t.min, -3).length == size_t.max / 3);
 }
 
-debug @system unittest
-{//check the contracts
-    import core.exception : AssertError;
-    import std.exception : assertThrown;
-    assertThrown!AssertError(iota(1,2,0));
```
Biotronic: To unify the interface, I opted to remove the assertions that are tested here. The other overloads did not assert, and removing assertions break less code than adding them.

std/range/package.d
```diff
@@ -5353,284 +5353,387 @@ pure @safe nothrow unittest
...
-            else
-            {
-                // Initialize an empty range
-                this.step = 0;
-                return;
-            }
-            this.step = step;
-            this.current = current;
-        }
+    this()(B b, E e, S s)
```
wilzbach: What's the reason for templating the constructor? `IotaResult` is already a template.

quickfur:
ping @andralex  Do the changes look OK now?  Apparently this has stalled. Let's get it moving again!

std/range/package.d
```diff
@@ -5369,284 +5369,387 @@ pure @safe nothrow unittest
...
+    }
+    else static if (is(typeof(begin + B(1))))
+    {
+        return iota(begin, end, B(1));
+    }
+    else static if (is(typeof(begin + 1)))
+    {
+        return iota(begin, end, 1);
+    }
+    else static if (is(typeof(++begin)))
```
andralex: this should go before the test `begin + 1` i.e. prefer increment to adding 1

Biotronic: Wait... I'm not sure you're right about that. A more correct test would put `begin + 1` first, and check in `popFront()` if it can use `begin++`. That way we can have efficient slicing on types that support both.

thewilsonator:
Ping @andralex this needs a rebase but is this otherwise goo to go?

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
     ---
 */
 auto iota(B, E, S)(B begin, E end, S step)
-if ((isIntegral!(CommonType!(B, E)) || isPointer!(CommonType!(B, E)))
```
andralex: There should still be a constraint, right?

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
 {
-    import std.conv : unsigned;
+    static if (!is(B == Unqual!B) || !is(E == Unqual!E) || !is(S == Unqual!S))
+    {
+        return iota(Unqual!B(begin), Unqual!E(end), Unqual!S(step));
+    }
+    else
+    {
+        return IotaResult!(B, E, S)(begin, end, step);
+    }
```
andralex: I think this was meant:

```d
return IotaResult!(Unqual!B, Unqual!E, Unqual!S)(begin, end, step);
```

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
+}
 
-    alias Value = CommonType!(Unqual!B, Unqual!E);
-    alias StepType = Unqual!S;
+/// Ditto
+auto iota(B, E)(B begin, E end)
```
andralex: please add constraint

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
+        return iota(E(0), end);
+    else
+        return iota(Unqual!E(end));
+}
 
-    static struct Result
+/*
+ * Overload for anything supporting regular arithmetic.
+ */
+struct IotaResult(B, E, S)
```
andralex: Is this meant to be public?

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
+        if (s == 0) return;
+        if (b < e && s < 0) return;
+        if (b > e && s > 0) return;
 
-        @property bool empty() const { return step == 0; }
-        @property inout(Value) front() inout { assert(!empty); return current; }
-        void popFront()
+        begin = b;
+        end = e;
+        step = s;
```
andralex: Why the structure whereby `step` is written to three times (once in .init, then assigned from 0, then assigned from `s`? Would it work to make the member `= 0` by default? If not:

```d
this(B b, E e, S s)
{
    if (s == 0 || b < e && s < 0 || b > e && s > 0)
    {
        // degenerate cases
        step = 0;
        return;
    }
    step = s;
    ...
}
```

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
+
+        if (begin < end && step > 0)
```
andralex: This is redundant because this case has already been eliminated by the early returns above. `if (step > 0)` should suffice.

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
-            else current += step;
+            // Iterating upward
+            assert(unsigned((end - begin) / step) <= size_t.max);
+            // Cast below can't fail because begin < end
+            end = cast(E) (end - 1);
+            end -= unsigned(end - begin) % step;
         }
-
-        @property inout(Value) back() inout
+        else if (begin > end && step < 0)
```
andralex: There seems to be redundancy here as well.

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
+            end += unsigned(begin - end) % (0 - step);
         }
-        void popBack()
+        else
         {
-            assert(!empty);
-            if (current == last) step = 0;
-            else last -= step;
+            // Initialize an empty range
+            step = 0;
```
andralex: This seems overdone. There have been several tests above that eliminated the empty range cases. There would be the case where `begin == end` and `step != 0` but those should be handled properly as limits of the normal cases.

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
         }
-        auto contains(Value x){return x in this;}
-        inout(Result) opSlice() inout { return this; }
-        inout(Result) opSlice(ulong lower, ulong upper) inout
+        else static if (is(typeof(begin == end)))
         {
-            assert(upper >= lower && upper <= this.length);
+            return begin == end;
+        }
+        else static assert(false, "Arguments to iota must be comparable using > or ==");
```
andralex: Shouldn't this case be eliminated by the template constraints?

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
+    }
```
andralex: `empty` seems off, too. A simple invariant that is established by the constructor and popXxx is "a range is empty if and only if the step is zero". Then `empty` would be a one-liner using that invariant.

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
-                cast(Value)(current + lower * step),
-                cast(Value)(current + upper * step),
-                step);
+    @property
+    size_t length() const
+    {
+        if (step > 0)
+        {
+            auto delta = cast(size_t)(end - begin);
+            return 1+cast(size_t)(delta / step);
```
andralex: spaces around operator

andralex: It's a one-liner, why is a named temporary needed?

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
-            if (step > 0)
-                return 1 + cast(size_t) (unsigned(last - current) / step);
-            if (step < 0)
-                return 1 + cast(size_t) (unsigned(current - last) / (0 - step));
-            return 0;
+            auto delta = cast(size_t)(begin - end);
+            auto absStep = 0-step;
+            return 1+cast(size_t)(delta / absStep);
         }
+        else return 0;
```
andralex: redundant `else`

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
+    @property
+    size_t length() const
+    {
+        if (step > 0)
+        {
+            auto delta = cast(size_t)(end - begin);
+            return 1+cast(size_t)(delta / step);
         }
-        @property size_t length() const
+        else if (step < 0)
```
andralex: redundant `else`

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
 
-    static struct Result
+    this(B b, E e, S s)
     {
-        private Value current, pastLast;
-
-        this(Value current, Value pastLast)
+        begin = b;
+        step = s;
+        count = cast(size_t)((e - begin)/step+0.5);
```
andralex: ...spacing...

std/range/package.d
```diff
@@ -5887,284 +5887,387 @@ pure @safe nothrow unittest
...
+        auto end = begin + (count+1) * step;
```
andralex: ...spacing...

LightBender:

PR #5746, opened 2017-09-27
Fix Issue 17862 - std.random.XorshiftEngine.min is wrong when bits == 32

n8sh:
XorshiftEngine.min is defined as 0 regardless of template parameters
but an XorshiftEngine cannot produce a value of zero if its internal
state has the same number of bits as the output element type.

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -1027,7 +1027,7 @@ if (isUnsigned!UIntType)
     /// Always $(D false) (random generators are infinite ranges).
     enum empty = false;
     /// Smallest generated value.
-    enum UIntType min = 0;
+    enum UIntType min = seeds_.length == 1 ? 1 : 0;
     /// Largest generated value.
     enum UIntType max = UIntType.max;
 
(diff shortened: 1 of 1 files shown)


JackStouffer:
Pinging our std.random experts @wilzbach @WebDrake @andralex

quickfur:
Code looks OK to me.  But I'll defer to others who know better about PRNGs.

Also, needs rebase.

n8sh:
>I'll defer to others who know better about PRNGs.

Here's the code for `front`:
```d
    static if (bits == 192)
        return value_;
    else
        return seeds_[size - 1];
```

Here's the `popFront` code for the `seeds_.length == 1` case:
```d
    UIntType temp;

    static if (bits == 32)
    {
        temp      = seeds_[0] ^ (seeds_[0] << a);
        temp      = temp ^ (temp >> b);
        seeds_[0] = temp ^ (temp << c);
    }
    else
        /* all other branches are irrelevant */
```
For unsigned types you can see `x ^ (x << k)` and `x ^ (x >> k)` are zero only when `x` is zero, so when `seeds_.length == 1` the only way for `front` to be zero after `popFront` is if it were also zero _before_ `popFront`.

The constructor and the `seed` method call `sanitizeSeeds` which disallows 0 from appearing in `seeds_`, and `seeds_` has a non-zero initializer, so when `seeds_.length == 1` there is no way to get `front == 0` without memory corruption.

>Also, needs rebase.

Rebased.

WebDrake:
Oh cripes, sorry for forgetting about this one.  It looks reasonable, and I'd expect the impact to be minimal.

PR #5749, opened 2017-10-01
std.windows.registry: Keep Key object alive while iterating over it

kinke:
The `Key` and `{Key,Value}NameSequence` objects weren't referenced during and after iteration in `...NameSequence.opApply()`, so there was no guarantee the `Key` object owning the HKEY handle would stay alive and keep the handle open - its destructor closes the handle.

This lead to a failing `std.datetime.timezone` unittest for optimized LDC builds.

Pinging @klickverbot @ZombineDev @rainers as involved in the discussion starting [here](https://github.com/ldc-developers/ldc/pull/2252#issuecomment-333325654).

### DIFF
--- a/std/windows/registry.d
+++ b/std/windows/registry.d
@@ -735,20 +735,21 @@ body
         "Value cannot be set: \"" ~ subKey ~ "\"");
 }
 
-private void regProcessNthKey(HKEY hkey, scope void delegate(scope LONG delegate(DWORD, out string)) dg)
+private void regProcessNthKey(Key key, scope void delegate(scope LONG delegate(DWORD, out string)) dg)
 {
     DWORD cSubKeys;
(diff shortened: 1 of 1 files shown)


std/windows/registry.d
```diff
@@ -735,20 +735,21 @@ body
         "Value cannot be set: \"" ~ subKey ~ "\"");
 }
 
-private void regProcessNthKey(HKEY hkey, scope void delegate(scope LONG delegate(DWORD, out string)) dg)
+private void regProcessNthKey(Key key, scope void delegate(scope LONG delegate(DWORD, out string)) dg)
```
kinke: Anything blocking? ;) - I'm not sure whether decorating the `key` param with `scope` makes a difference.

kinke:
Thanks. :)

PR #5603, opened 2017-07-12
Issue 17127 - bad example code for std.concurrency.Generator

wilzbach:
The only way we can ensure that an example works is by actually executing it!

PR #5752, opened 2017-10-01
Issue 17385: Update findSplit doc so example compiles.

jondegenhardt:
As described in [17835](https://issues.dlang.org/show_bug.cgi?id=17835), the code example does not compile. Investigation by Duncan Patterson indicated changing the code to support the use of 'const' shown in the example will be problematic. Instead changing the documentation to a form that works.

### DIFF
--- a/std/algorithm/searching.d
+++ b/std/algorithm/searching.d
@@ -2791,9 +2791,9 @@ shown in the following example.
 
 Example:
 ---
-if (const split = haystack.findSplit(needle))
+if (auto split = haystack.findSplit(needle))
 {
-     doSomethingWithSplit(split);
+    doSomethingWithSplit(split);
 }
 ---
  */


MetaLang:
You should go all the way and make this a documented unit test.

jondegenhardt:
@MetaLang I looked through the existing unit tests. They do test that opcast operator involved is adequately tested. Assignment to 'auto' as in the example is tested as well. There doesn't seem any additional value in adding the example as written as another unit test.

MetaLang:
There's no reason _not_ to make it a documented unit test, since it's already there in the documentation and making it a runnable unit test will ensure that it never breaks again.

jondegenhardt:
@MetaLang It's simply prioritization of my time. The example as written is part of the flow of existing written text. It is illustrating an idiom, and is written with that intent. It is very concise. Four lines. However, it won't compile/run, even with the change I made, as there is no `doSomethingWithSplit` function.

To make an actual unit test read as well means I'll have to spend some time thinking about. My quick thoughts about how to do this meant adding an else clause, which would increase the length. And it's debatable whether asserts will read as well.

So yes, it's certainly doable, but I when I saw the existing doc structure I decided to limit the scope to the most pressing issue, which is that the declaration form used (`const`) does not compile.  I did check to ensure that with the unit tests in place the `auto` declaration and use of the opcast operator used in the example will continue to work.

wilzbach:
> @MetaLang It's simply prioritization of my time. The example as written is part of the flow of existing written text. It is illustrating an idiom, and is written with that intent. It is very concise. Four lines. However, it won't compile/run, even with the change I made, as there is no doSomethingWithSplit function.
> To make an actual unit test read as well means I'll have to spend some time thinking about it.

I don't follow - what's wrong with sth. simple like this?

```d
// findSplit returns a triplet
if (auto split = "dlang-rocks".findSplit("-"))
	assert(split[2] == "rocks");
```

> And it's debatable whether asserts will read as well.

That's why you don't see them on dlang.org - we automagically convert them ;-)

jondegenhardt:
> That's why you don't see them on dlang.org - we automagically convert them ;-)

I'm not following. I see asserts in the unit tests exposed in the documentation on dlang.org everywhere I look. Here is are the initial examples from `findSplit` as displayed on web page:

    import std.range.primitives : empty;
    auto a = "Carl Sagan Memorial Station"; 
    auto r = findSplit(a, "Velikovsky");
    import std.typecons : isTuple;
    static assert(isTuple!(typeof(r.asTuple)));
    static assert(isTuple!(typeof(r)));
    assert(!r);

There doesn't seem to be conversion going on. So there must be something else going on here I'm not aware of.

To me, this conversation is misguided. The PR was a quick and easy fix to what I see as an meaningful problem: The form in the documentation won't work, and users who try to copy it will get an annoying compilation error. This focus on re-writing the documentation so that the example is auto-generated from unit tests rather than being in-line text is desirable, but not nearly as big a deal. If someone want to show exactly how to do this, or show an existing example that does this, that would be helpful. But so far no one has.

MetaLang:
>To me, this conversation is misguided. The PR was a quick and easy fix
I'm not sure what the issue is here. Making it a unit test is almost as quick and easy:

```
unittest
{
    import std.range;
    // findSplit returns a triplet
    if (auto split = "dlang-rocks".findSplit("-"))
	    assert(split[2] == "rocks");
}
```

That said, thank you for the PR; I'll merge as-is and we can make it a unit test in a follow-up.

jondegenhardt:
Thank you.

FWIW, I assume the unit test would have to take the form:

    unittest
    {
        import std.range;
        if (auto split = "dlang-rocks".findSplit("-"))
        {
    	    assert(split[2] == "rocks");
        }
        else
        {
            assert(false);
        }
    }

This is what is behind my comments about needing to consider whether it would constitute a loss of readability when replacing the current text.

MetaLang:
No, I don't think so. Why would the test have to be changed at all?

jondegenhardt:
If I was reviewing the unit test above without the `else assert(false)` clause, that would my comment. That as a unit test, it should also ensure that findSplit actually found what it was looking for. (By the way, for me this is a more useful line of conversation. I'm guessing this topic, how to balance unit test correctness vs documentation, has come up before and people are settling on approaches.)

wilzbach:
> I'm not following. I see asserts in the unit tests exposed in the documentation on dlang.org everywhere I look. Here is are the initial examples from findSplit as displayed on web page:

It only works for `assert` with `==` at the moment:

-> https://dlang.org/blog/2017/03/08/editable-and-runnable-doc-examples-on-dlang-org/

> This focus on re-writing the documentation so that the example is auto-generated from unit tests rather than being in-line text is desirable, but not nearly as big a deal.

Well the point here was that `unittest` blocks get automatically tested for every PR at dmd, druntime, phobos etc., so once you have a `unittest` block in Phobos, it is almost guaranteed that the code will continue to work the documented way. There are dozens of plaintext examples in Phobos which don't work anymore - just one recent example: https://github.com/dlang/phobos/pull/5603

> If I was reviewing the unit test above without the else assert(false) clause, that would my comment. That as a unit test, it should also ensure that findSplit actually found what it was looking for. (By the way, for me this is a more useful line of conversation. I'm guessing this topic, how to balance unit test correctness vs documentation, has come up before and people are settling on approaches.)

Public examples (the ddoced ones with `///`) are for a user-facing documentation where the focus is to have simple examples which are easy to understand. Usually one tries to have two or three 4-5 line examples to explain the API. However, of course, a good Phobus function is then accompanied by hundreds of lines of internal tests which try to test the functionality and possible usages almost completely.

jondegenhardt:
Thanks for the follow-up. This is very helpful information.

I would like to suggest that reviews of PRs like this would benefit from a slightly different tone. Note that in this case I was not introducing new documentation, I was correcting an existing piece of documentation. It was for a bug I filed, it had been sitting there for a while, and was likely to be in the queue for a while. I had a free half hour and felt I could make a useful change in that time. Importantly, I did check the existing unit tests to ensure that the code in text would continue to work in the future, as long as the unit tests were not changed.

Instead of being thanked, my contribution was initially rejected because it didn't also change the existing documentation to conform to the new guidelines. Now, I realize there was no ill will intended by this. But, the review comments could have recognized that correcting the writing structure was a separate task, and asked if I would be willing to also do this as part of the PR. This would have been a much more satisfying interaction. Instead I was left wondering why I bothered. 

Now, in retrospect, I can see that what the reviewers were trying to say is that making the structure change is easy with the documentation generation system. And indeed, it's clearly not hard. There has been a lot of great work done in this area. Still, I haven't worked with documentation system, did not know how unit test examples get formatted, about the conversion of assert to writeln (very cool by the way), etc. I would have had to spend at least a half hour more learning the documentation system before doing this, and more time reviewing my changes to see if they read well. I guess what I'm suggesting is to recognize that less frequent contributors will need to make this type of time investment.

MetaLang:
Sorry Jon, I think there's been a mis-understanding. Your contribution was not initially rejected, it's just that these are old-style examples and our policy is to always use documented unit tests now. It'a hard to keep track of who knows what, and I thought I recognized your name as someone else who occasionally contributes and thus thought you would already know how to change this example into a documented unit test. My mistake, sorry.

jondegenhardt:
NP. It was my 12th contribution, which puts me 81st on the phobos contributors list, but it was the first that modified documentation in any form. And really, I do appreciate being informed of how to do things better and in a preferred way.

PR #5755, opened 2017-10-03
Make statTimeToStdTime public

FreeSlave:
That might be useful for someone who's working with stat structs and wants to avoid a call of std.file functions like getTimes which accept only filenames and have to call stat function internally again.

Note: I could not figure out why statTimeToStdTime does not appear in generated docs though I added doc comment and added the function to Timestamp category.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -1826,7 +1826,7 @@ else version(Posix)
         return statTimeToStdTime!'c'(statbuf);
     }
 
-    unittest
+    @safe unittest
     {
         stat_t statbuf;
         // check that both lvalues and rvalues work
--- a/std/file.d
+++ b/std/file.d
@@ -1180,7 +1180,7 @@ if (isConvertibleToString!R)
 
 // Reads a time field from a stat_t with full precision.
 version(Posix)
-private SysTime statTimeToStdTime(char which)(ref stat_t statbuf)
+private SysTime statTimeToStdTime(char which)(ref const stat_t statbuf)
 {
     auto unixTime = mixin(`statbuf.st_` ~ which ~ `time`);
     long stdTime = unixTimeToStdTime(unixTime);
@@ -1793,7 +1793,7 @@ version(StdDdoc)
      Params:
         statbuf = stat_t retrieved from file.
      +/
-    SysTime timeLastModified(ref stat_t statbuf) pure nothrow;
+    SysTime timeLastModified()(auto ref stat_t statbuf) pure nothrow {assert(false);}
     /++
      $(BLUE This function is Posix-Only.)
 
@@ -1801,7 +1801,7 @@ version(StdDdoc)
      Params:
         statbuf = stat_t retrieved from file.
      +/
-    SysTime timeLastAccessed(ref stat_t statbuf) pure nothrow;
+    SysTime timeLastAccessed()(auto ref stat_t statbuf) pure nothrow {assert(false);}
     /++
      $(BLUE This function is Posix-Only.)
 
@@ -1809,22 +1809,29 @@ version(StdDdoc)
      Params:
         statbuf = stat_t retrieved from file.
      +/
-    SysTime timeStatusChanged(ref stat_t statbuf) pure nothrow;
+    SysTime timeStatusChanged()(auto ref stat_t statbuf) pure nothrow {assert(false);}
 }
-else
-version(Posix)
+else version(Posix)
 {
-    SysTime timeLastModified(ref stat_t statbuf) pure nothrow
+    SysTime timeLastModified()(auto ref stat_t statbuf) pure nothrow
     {
         return statTimeToStdTime!'m'(statbuf);
     }
-    SysTime timeLastAccessed(ref stat_t statbuf) pure nothrow
+    SysTime timeLastAccessed()(auto ref stat_t statbuf) pure nothrow
     {
         return statTimeToStdTime!'a'(statbuf);
     }
-    SysTime timeStatusChanged(ref stat_t statbuf) pure nothrow
+    SysTime timeStatusChanged()(auto ref stat_t statbuf) pure nothrow
     {
         return statTimeToStdTime!'c'(statbuf);
+    }
+
+    unittest
+    {
+        stat_t statbuf;
+        // check that both lvalues and rvalues work
+        timeLastAccessed(statbuf);
+        timeLastAccessed(stat_t.init);
     }
 }
 
--- a/std/file.d
+++ b/std/file.d
@@ -54,6 +54,8 @@ $(TR $(TD Timestamp) $(TD
           $(LREF getTimesWin)
           $(LREF setTimes)
           $(LREF timeLastModified)
+          $(LREF timeLastAccessed)
+          $(LREF timeStatusChanged)
 ))
 $(TR $(TD Other) $(TD
           $(LREF DirEntry)
@@ -1780,6 +1782,50 @@ if (isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R))
     assert(target.timeLastModified(SysTime.min) < source.timeLastModified);
     target.write(".");
     assert(target.timeLastModified(SysTime.min) >= source.timeLastModified);
+}
+
+version(StdDdoc)
+{
+    /++
+     $(BLUE This function is Posix-Only.)
+
+     Returns the time that the given file was last modified.
+     Params:
+        statbuf = stat_t retrieved from file.
+     +/
+    SysTime timeLastModified(ref stat_t statbuf) pure nothrow;
+    /++
+     $(BLUE This function is Posix-Only.)
+
+     Returns the time that the given file was last accessed.
+     Params:
+        statbuf = stat_t retrieved from file.
+     +/
+    SysTime timeLastAccessed(ref stat_t statbuf) pure nothrow;
+    /++
+     $(BLUE This function is Posix-Only.)
+
+     Returns the time that the given file was last changed.
+     Params:
+        statbuf = stat_t retrieved from file.
+     +/
+    SysTime timeStatusChanged(ref stat_t statbuf) pure nothrow;
+}
+else
+version(Posix)
+{
+    SysTime timeLastModified(ref stat_t statbuf) pure nothrow
+    {
+        return statTimeToStdTime!'m'(statbuf);
+    }
+    SysTime timeLastAccessed(ref stat_t statbuf) pure nothrow
+    {
+        return statTimeToStdTime!'a'(statbuf);
+    }
+    SysTime timeStatusChanged(ref stat_t statbuf) pure nothrow
+    {
+        return statTimeToStdTime!'c'(statbuf);
+    }
 }
 
 @safe unittest


CyberShadow:
Not sure about exposing the `which` parameter as-is to Phobos users. Perhaps some `timeLastAccessed`/etc. aliases/wrapper functions, or a `getTimes` overload taking a `stat_t`, would be better.

CyberShadow:
Pinging @andralex as he asked to be notified for any symbol additions to Phobos, and this effectively adds a new symbol accessible to Phobos users.

FreeSlave:
So which name should be used? Overload of getTimes? Maybe getTimesStat?
Note that original getTime has 3 arguments. But the new function must have 4 (stat buf, access time, modification time, change time).
Out of this PR, we also could have getTimesPosix for file names (similarly to getTimesWin) to retrieve a change time.

jmdavis:
I don't know. std.file is really supposed to be abstracting away stuff like `stat_t`, and `DirEntry` is essentially the abstraction for `stat_t` specifically. So, I'd be inclined to argue for just constructing a `DirEntry` from the `stat_t` and getting at the times that way except that `DirEntry` also includes the file's name, which `stat_t` doesn't have. So, for that to work, the programmer would either have to supply the file name, or we'd have to be okay with having the filename be null in that case, which seems questionable.

`statTimeToStdTime` seems a bit low level to expose (and actually it really should be `statTimeToSysTime` anyway; it returns a `SysTime`, not a "std time"), but I also really don't like the idea of adding a bunch of free functions which take a `stat_t` and return the various times. It's cleaner to use `getTimes`, because then we only end up with one function taking a `stat_t`, but that's less efficient if you only care about one of the times - which is usually the case. It makes sense in the case where you pass a file name, because then you're grabbing everything with one `stat` call when you do want more than one of the times, but it doesn't make a lot of sense for `stat_t`.

I'd be in favor of just arguing that if someone is doing something as low level as using `stat_t`, they should just do the conversion themselves, but with the addition of the timensec fields on some systems, that gets complicated enough that it arguably should be wrapped. When it was just some `time_t` fields, it was as simple as calling `SysTime.fromUnixTime`, but with the timensec fields present, you have to add in those values if you want the full time - and those fields only exist on some systems and aren't named the same on all systems. So, doing that yourself in a platform-independent way pretty much forces you to implement `statTimeToStdTime`.

If we think that we have to expose this functionality, then it's probably best to just make `statTimeToStdTime` public, but it would need to have its name fixed, it would need a template constraint, and it would need unit tests. It should probably also be changed to use `core.time.convert` rather than hardcoding the unit conversions, since using `convert` would be more idiomatic.

FreeSlave:
@jmdavis got it right. The one who does not want to use std.file must re-implement (or copy) the whole `statTimeToStdTime` to match the results returned by phobos. This is the main reason to expose it. The function may reside in some other module than std.file (some core.*?) so user would not need to import std.file at all.

jmdavis:
core is primarily for bindings. There really isn't any place else that we'd put a helper function like this. Normally, folks are either using the higher level API that std.file provides and not using system-specific stuff like `stat_t`, or they're doing the lower level stuff themselves, in which case, they're not using std.file (or probably Phobos in general). There are a few places that system-specific stuff like `stat_t` are exposed in std.file for convenience (e.g. on a `DirEntry`), but we've generally avoided providing any code that operates on lower level stuff without encapsulating it. It might make sense in this case to just expose the helper function, but it's not really what we're generally trying to do with std.file.

And the more I think about it, the less this makes sense. What use case is there for exposing this? The vast majority of D programs are going to use `DirEntry` or `getTimes` for this kind of information. Relatively few will be operating on a `stat_t`. And out of those that are, this functionality really only matters if they care about having file times be more precise than a second. If all you need is second-level precision, then you can just grab the corresponding field from the `stat_t` and feed it to `SysTime.fromUnixTime`, and you're good to go. This functionality only matters if you're getting a `stat_t` from somewhere other than D code, and you need file times to be more precise than a second. And how many programs are in that boat? _Very_ few, I expect.

So, while I'm not entirely opposed to exposing this functionality given that it _does_ nicely deal with the differences across systems for getting the sub-second portion of the file time, I definitely question that it makes sense to be exposing it, as it doesn't really fit in with what std.file is trying to do, and very few programs are likely to need it. So, I don't know.

But if we do expose it, it needs to be fixed up first as I commented before. It can't simply be exposed as-is.

CyberShadow:
> and actually it really should be statTimeToSysTime anyway; it returns a SysTime, not a "std time"

Oops, it returned a std time (long) in the initial implementation (like the FILETIME one) but I changed the return type without changing the name.

> What use case is there for exposing this?

There are libraries which do use `stat_t` as first-class API types, such as FUSE.

There may be also some rather platform- or system-specific syscalls that use `stat_t`, such as getting a file's time by its file descriptor or inode.

> And out of those that are, this functionality really only matters if they care about having file times be more precise than a second.

Using full precision is actually more important than might be obvious. For example, for a while (still? don't remember) std.file.copy would copy files' timestamps with just second precision, discarding sub-second time data. However, as a result, when you used the function to copy files around, the discarded timestamp information would throw off tools that relied on it, such as build systems. Essentially it only copied the data half-way, which is sometimes worse than failing.

> If all you need is second-level precision, then you can just grab the corresponding field from the stat_t and feed it to SysTime.fromUnixTime, and you're good to go.

We should support full timestamp precision in all Phobos code that deals with timestamps.

CyberShadow:
BTW, if we are exposing getters, we probably also want to have setters as well, so we should probably discuss/design the interface with having both.

jmdavis:
> We should support full timestamp precision in all Phobos code that deals with timestamps.

I don't disagree. The issue here is whether it makes sense to expose functions for operating on `stat_t` in std.file. We've mostly avoided putting system-specific stuff in std.file's API or providing anything to operate on lower level constructs without encapsulating them. Normally, we've left stuff like that to the user to deal with, because it's a lower level than we've wanted to expose in Phobos' API. So, this isn't really about exposing full timestamp precision in Phobos API so much as whether we should provide helper function for dealing with full timestamp precision with lower level constructs rather than with the constructs that std.file provides and normally operates on in its API.

And maybe in this case, it really does make sense to expose one or more functions which operate on a `stat_t`, but I don't particularly like the idea given that it's lower level, system-specific stuff, and that really isn't what std.file is aimed at. So, while I'm not completely opposed to the idea, I don't really like it either.

CyberShadow:
Isn't this pretty much exactly like `FILETIMEToSysTime`, though (which is public and documented)?

Perhaps we should move this helper function to `std.datetime.systime`, near `FILETIMEToSysTime`?

FreeSlave:
`The vast majority of D programs are going to use DirEntry or getTimes for this kind of information. Relatively few will be operating on a stat_t`

Well, anyone who wants to keep their code nothrow nogc will not use DirEntry and getTimes.

jmdavis:
> Isn't this pretty much exactly like FILETIMEToSysTime, though (which is public and documented)?
> Perhaps we should move this helper function to std.datetime.systime, near FILETIMEToSysTime?

Maybe. This is somewhat of a weird case in that we're not converting a type to `SysTime`; we're converting specific pieces of a type to a `SysTime`. The biggest problem with putting it on std.datetime.systime would be that that would then force std.datetime.systime to import the core module containing `stat_t`, and arguably too much gets pulled in when importing std.datetime.systime already. Also, if anyone is looking for this functionality, I bet that they'd look in std.file before they looked in std.datetime, but I don't know. `FILETIMEToSysTime` arguably shouldn't be in std.datetime, but that decision was made ages ago, so it's a bit late now. It's not entirely a straightforward decision though.

> Well, anyone who wants to keep their code nothrow nogc will not use DirEntry and getTimes.

Anyone who wants their code to be `nothrow` isn't going to use std.file in general and is basically saying that they're not interested in using D's standard library to deal with files, since both std.file and std.stdio take the approach of using exceptions to handle errors with files. And honestly, I would argue that it really doesn't make sense in most cases to be operating on files and not use exceptions unless you're stuck in a language that doesn't have exceptions. That gets ugly fast, and it's certainly never been one of the goals of std.file.

But since we already have to have code to extract the times from a `stat_t`, and it's non-trivial thanks to the fractional seconds of a stat_t being non-standard, there's certainly an argument to be made for exposing that functionality even if we're not providing functions for explicitly operating on `stat_t` in general.

jmdavis:
If we're going to expose `statTimeToStdTime`, I'd say that it probably makes the most sense to fix it up and expose it rather than splitting it into multiple functions or adding an overload to `getTimes`. If we want to add a setter too, then I guess that that makes sense, but I don't know what its use case would be. From the looks of what std.file is currently doing, the time would normally be set on a file using either a `timeval` or a `timespec`, not a `stat_t`. AFAIK, normally, a `stat_t` is only used for conveying information about an existing file not for setting it.

CyberShadow:
> AFAIK, normally, a stat_t is only used for conveying information about an existing file not for setting it.

I was thinking of FUSE again - the `getattr` interface function from Facebook's DFuse requires populating a `stat_t`.

jmdavis:
Well, then we can provide a corresponding setter. Either way, the name will need to be fixed, and unit tests will need to be added. I'd also suggest making the getter `auto ref` rather than `const ref`.

andralex:
I'm a bit confused. Isn't there a fair amount of engineering going on in here? If somebody has a `stat_t` object, it contains the three times already in `time_t` format, right? So all that's needed is a primitive `time_tToSysTime` that converts an individual `time_t` value to a `SysTime` value. Then use `buf.st_mtime.time_tToSysTime` etc. What am I missing?

std/file.d
```diff
@@ -930,10 +931,13 @@ if (isConvertibleToString!R)
...
-// Reads a time field from a stat_t with full precision.
-version(Posix)
-private SysTime statTimeToStdTime(char which)(ref stat_t statbuf)
+version(StdDdoc)
+{
+    /// Reads a time field from a stat_t with full precision.
+    SysTime statTimeToStdTime(char which)(const ref stat_t statbuf) if (which == 'a' || which == 'c' || which == 'm');
+}
+else version(Posix)
+SysTime statTimeToStdTime(char which)(const ref stat_t statbuf) if (which == 'a' || which == 'c' || which == 'm')
```
andralex: As a rule of thumb we should use template parameters only if there's a good chance the user will have the arguments in the same format. In this case, it's unlikely someone has a compile-time char that is `a`, `m`, or `c` received from elsewhere so a template is not warranted here.

CyberShadow:
> What am I missing?

Sub-second precision, which is not representable in a `time_t`, and instead stored in `stat_t` as additional non-standard platform-specific fields.

andralex:
@CyberShadow thanks. Then three functions sharing implementation would be the ticket.

quickfur:
ping @FreeSlave @CyberShadow @jmdavis 

Seems Andrei has given the green light on this one pending proposed changes.  Let's unclog this one from the PR queue, shall we?

FreeSlave:
I'm not sure how you want to name these potential three functions.

CyberShadow:
How about `timeCreated`, `timeLastAccessed` and `timeLastModified` overloads taking a `stat_t` ?

FreeSlave:
@CyberShadow there are member functions of DirEntry, so not really overload.
And ctime is not timeCreated. It's time of change (not only content modifiction, but also permission change), so it should be timeLastChanged I guess? Btw DirEntry lacks such kind of member function.

CyberShadow:
> @CyberShadow these are member functions of DirEntry, so not really overload.

Ah. `timeLastModified` is also a top-level function, so an overload for that one.

> And ctime is not timeCreated.

Right.

FreeSlave:
@CyberShadow strangely only timeLastModified exists as free function. It's inconsistent to add others just for stat_t.

CyberShadow:
Right, the others won't be overloads, just free functions. We'd probably want to encourage using them via UFCS so that they appear like methods/properties of `stat_t`.

FreeSlave:
@CyberShadow what I meant is that it would weird to have timeLastAccessed only for stat, while not for filename. Should we add counterparts that accept files by name too?

CyberShadow:
> @CyberShadow what I meant is that it would be weird to have timeLastAccessed only for stat, while not for filename.

I guess so, but do you suggest we do something else instead?

> Should we add counterparts that accept files by name too?

I think that's a separate issue that should be discussed elsewhere.

std/file.d
```diff
@@ -1779,6 +1781,50 @@ if (isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R))
...
 }
 
+version(StdDdoc)
+{
+    /++
+     $(BLUE This function is Posix-Only.)
+
+     Returns the time that the given file was last modified.
+     Params:
+     statbuf = statbuf retrieved from file.
```
CyberShadow: Why `statbuf retrieved from file`? Is `statbuf` a pre-established term? Maybe "`stat_t` retrieved from file"?

FreeSlave:
> I guess so, but do you suggest we do something else instead?

There's also an option to add getTimes overload that takes stat_t (btw we probably need also getTimesPosix returning change time in addition to getTimesWin).
I don't know why timeLastModified exists as a separate function without other counterparts in the first place.

CyberShadow:
> There's also an option to add getTimes overload that takes stat_t

Well, the thing with `getTimes` is that its API is somewhat unwieldy (you need to declare the three variables whether you're going to use their values or not), because if you need more than one timestamp you'd otherwise need more than one `stat` call. But this problem doesn't exist when we query an existing `stat_t` structure, so we can avoid having the unwieldy API for this case.

> btw we probably need also getTimesPosix, returning change time, in addition to getTimesWin

Sure.

> I don't know why timeLastModified exists as a separate function without other counterparts in the first place.

If I had to guess, probably because it's the file timestamp needed by far the most often.

Having just one shorthand for `getTimes` to return just one file timestamp is fine, but once you have more than one, it becomes easy to accidentally write inefficient programs that invoke `stat` more times than necessary.

FreeSlave:
@CyberShadow I added getTimesPosix https://github.com/dlang/phobos/pull/6539
I'm still not sure how to implement this one without introducing too many new functions.

FreeSlave:
Updated documentation comments, add nothrow pure attributes.

CyberShadow:
I don't have a strong opinion about it, we can make it not an overload. Did you have a better function name in mind?

FreeSlave:
Maybe statTimeLastModified, etc.

CyberShadow:
Well, I still don't have a strong opinion about it, but `statTimeLastModified` seems redundant in that the subject is already disambiguated as the type of the only argument, and since D allows overloading throughout and across modules, there is generally no need to disambiguate the subject by putting its type in the function name.

@jmdavis Any thoughts?

jmdavis:
> @jmdavis Any thoughts?

We should take advantage of overloading. There's no reason to give these different names.

std/file.d
```diff
@@ -1782,6 +1784,50 @@ if (isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R))
...
+version(StdDdoc)
+{
+    /++
+     $(BLUE This function is Posix-Only.)
+
+     Returns the time that the given file was last modified.
+     Params:
+        statbuf = stat_t retrieved from file.
+     +/
+    SysTime timeLastModified(ref stat_t statbuf) pure nothrow;
```
jmdavis: Why are any of these `ref`? That's just going to be really annoying any time someone has an rvalue or if the argument is `const`. The function is not modifying the argument, so it really shouldn't be `ref`. If you're really that worried about copying the `stat_t`, then just templatize these functions and make them take `auto ref` - e.g.

    SysTime timeLastModified()(auto ref stat_t statbuf);

or

    SysTime timeLastModified(T)(auto ref T statbuf)
    if (is(Unqual!T == stat_t))
    {...}

std/file.d
```diff
@@ -1782,6 +1784,50 @@ if (isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R))
...
+     $(BLUE This function is Posix-Only.)
+
+     Returns the time that the given file was last changed.
+     Params:
+        statbuf = stat_t retrieved from file.
+     +/
+    SysTime timeStatusChanged(ref stat_t statbuf) pure nothrow;
+}
+else
+version(Posix)
```
jmdavis: I know this is nitpicky, but please put these two lines on one line. This is like doing

    else
    if(cond)

which is a bit weird and just wastes a line.

std/file.d
CyberShadow: I think the compiler has been automatically templatizing `auto ref` functions, so `()(` is not necessary.

jmdavis: If so, that's a new thing, but I just tried it locally, and the compiler still gives an error if you use `auto ref` and the function isn't templated.

jmdavis: I should probably point out though that if these are templated instead of just accepting `stat_t` without `ref`, then their bodies in the `version(StdDoc)` section will probably need `assert(0);` in them to make the compiler shut up about the lack of function body for a templated function.

PR #5757, opened 2017-10-04
Fix Issue 14794 - Add optional position parameter to std.file.read

RazvanN7:


FreeSlave:
What's the point to worry about offsets? std.file.read does not allow reading from some previously opened file descriptor so you won't get any conflicts.

CyberShadow:
Do we really need this? This smells of pointless feature creep... `std.file.read` and `write` are meant as convenience functions for things you can do anyway with `std.stdio.File`. How often do you need to open a file just to read something at a specific position once, then close it and never do anything with it?

I also don't like that the parameter is added in the middle, i.e. it doesn't act like an optional parameter, and `read(fn, a)` and `read(fn, a, 0)` do different things (but I see why having the position come after the length is awkward).

RazvanN7:
I also think that using std.stdio is better for this kind of stuff, but having a PR is the best way to bring the issue to attention. In my opinion, if read is useful, read with offset may also be useful sometimes.

Anyway, if you think this is unnecessary, we may close this PR and the bug report as WONTFIX.

CyberShadow:
@andralex Since you filed the enhancement request, any thoughts?

That `pread` is faster than `seek`+`read` doesn't really matter considering we're still opening the file just for that one read.

std/file.d
```diff
@@ -353,22 +377,36 @@ version (Posix) private void[] readImpl(const(char)[] name, const(FSChar)* namez
     stat_t statbuf = void;
     cenforce(fstat(fd, &statbuf) == 0, name, namez);
 
+    if (offset >= statbuf.st_size)
+        offset = to!size_t(statbuf.st_size + 1);
     immutable initialAlloc = min(upTo, to!size_t(statbuf.st_size
-        ? min(statbuf.st_size + 1, maxInitialAlloc)
+        ? min(statbuf.st_size + 1 - offset, maxInitialAlloc)
```
andralex: That won't work. One reason for the seemingly awkward way we currently implemented read (with the loop below) is that in Unix there are files that report size 0 yet when you start reading from them, reading works. So the only way to read from those files is to consider `statbuf.st_size` of only advisory value, not ground truth.

The new code assumes `statbuf.st_size` as ground truth, which won't work.

jmdavis: Do you specifically mean special files like /dev/zero or /dev/random? I've certainly never heard of normal files reporting an incorrect size, but that could certainly happen with special files. It had never occurred to me to use functions like /dev/zero or /dev/random with std.file, but I just tried it (with a value for `upTo`, of course), and it does work.

andralex: Yah, it's the special files. People need to read those as well... there was a bug report about it involving a popular special file, forgot which that was.

CyberShadow:
> Having the ability to skip a fixed-size header is good to have with read.

How often does this need occur in practice? I don't think I've ever encountered needing to do this to the point where I wished there was a Phobos function, and a lot of my D programs deal with files of various kinds...

andralex:
(Trying to reply via email because I have no access to gihub.com right 
now...)

When I submitted the issue I did have a use case that made the addition 
an obvious improvement. Sadly I forgot what it was, I now wish I'd have 
written it down. It wasn't the need to skip a header.

Anyhow, by means of a different use case: consider we want to 
occasionally update the contents of a file meant for append only (like a 
log). Then:

// first
auto buf = read(filename);
...
// later
buf ~= read(filename, buf.length, size_t.max);

If this is done frequently, keeping the file opened and aiming for a 
"tail -f" style strategy makes more sense. Otherwise, it's a convenient 
method to keep the buffer updated when the file grows.

Implementing positioned read well is surprisingly subtle, so it's fit 
for encapsulating in the stdlib. BTW we need to make sure the 
implementation works on special files.

CyberShadow:
> Otherwise, it's a convenient method to keep the buffer updated when the file grows.

Sounds good!

> Implementing positioned read well is surprisingly subtle,

Opening the file is still probably be going to be more overhead than whatever difference the read implementation could gain...

RazvanN7:
@andralex The linux verion now works with special files, but I'm having problems on windows. I will test it later today on a windows machine, but at a first glance I can't figure why the code with SetFilePointerEx doesn't work.

CyberShadow:
> , but at a first glance I can't figure why the code with SetFilePointerEx doesn't work.

Looks like it's because `numRead != chunkSize`?

wilzbach:
> The linux version now works with special files, but I'm having problems on windows. I will test it later today on a windows machine,

Ping @RazvanN7

PR #5758, opened 2017-10-05
Make DirEntry and DirIterator safe

jercaianu:
I have put trusted for some small functions which are basically wrappers for C calls, and made sure that the strings used in the implementation are null terminated.

Since RefCounted is not yet safe, I have also put trusted for the constructor of DirIterator.
The only thing preventing DirIterator to be completely safe, is this issue here [1].

[1] - https://issues.dlang.org/show_bug.cgi?id=17867

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -2808,6 +2808,7 @@ version(StdDdoc)
       +/
     struct DirEntry
     {
+        @safe:
         /++
             Constructs a $(D DirEntry) for the given file (or directory).
 
--- a/std/file.d
+++ b/std/file.d
@@ -3076,7 +3076,7 @@ else version(Posix)
     public:
         alias name this;
 
-        this(string path)
+        this(string path) @safe
         {
             if (!path.exists)
                 throw new FileException(path, "File does not exist");
@@ -3088,11 +3088,16 @@ else version(Posix)
             _dTypeSet = false;
         }
 
-        private this(string path, core.sys.posix.dirent.dirent* fd)
+        private this(string path, core.sys.posix.dirent.dirent* fd) @safe
         {
             import std.path : buildPath;
 
-            immutable len = core.stdc.string.strlen(fd.d_name.ptr);
+            static auto trustedStrlen(char* str) @trusted
+            {
+                return core.stdc.string.strlen(str);
+            }
+
+            immutable len = trustedStrlen(&fd.d_name[0]);
             _name = buildPath(path, fd.d_name[0 .. len]);
 
             _didLStat = false;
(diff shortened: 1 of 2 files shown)


jercaianu:
Thanks for the review @andralex !

@CyberShadow I saw that your build test failed, but I am not sure why.
Can you please point out how I can fix it?

Thanks,
Alex

CyberShadow:
> @CyberShadow I saw that your build test failed, but I am not sure why.
> Can you please point out how I can fix it?

You can look at the log to see the problem. Here is the relevant part:

```
std/file.d(3856): Error: @safe function 'std.file.DirIteratorImpl.mayStepIn' cannot call @system function 'std.file.DirEntry.isDir'
std/file.d(3856): Error: @safe function 'std.file.DirIteratorImpl.mayStepIn' cannot call @system function 'std.file.DirEntry.linkAttributes'
std/file.d(3909): Error: @safe function 'std.file.DirIteratorImpl.popFront' cannot call @system function 'std.file.DirEntry.name'
std/file.d(3923): Error: @safe function 'std.file.DirIteratorImpl.popFront' cannot call @system function 'std.file.DirEntry.name'
std/conv.d(4372): Error: static assert  "DirIteratorImpl cannot be emplaced from (string, SpanMode, bool)."
std/conv.d(4453):        instantiated from here: emplaceRef!(DirIteratorImpl, DirIteratorImpl, string, SpanMode, bool)
std/typecons.d(5418):        instantiated from here: emplace!(DirIteratorImpl, string, SpanMode, bool)
std/typecons.d(5509):        instantiated from here: initialize!(string, SpanMode, bool)
std/file.d(3947):        instantiated from here: __ctor!(string, SpanMode, bool)
```

jercaianu:
@CyberShadow 
Thanks, however if you check the PR, I added @safe at the top level of the DirEntry struct, but the checks complain that the functions are still @system. 

Should I put @safe somewhere else?
Thanks,
Alex

CyberShadow:
@jercaianu You need to update the `version(StdDdoc)` block.

andralex:
Placement of safe seems legit, this is odd. @CyberShadow I'll be on the road, could you please follow up and see if a bug report is warranted? @jercaianu as a workaround please return to your previous approach :/.

andralex:
Oh wait it's fixed. Nevermind.

jercaianu:
@andralex @CyberShadow 
The problem is that in the stdDdoc version for RefCounted there are 2 overloads for "refCountedPayload". One of them is safe and the other one isn't. 
In my code I am calling the safe one.

If I comment out the one which is not safe, the build is successful. 
How should I approach this?

Thanks,
Alex

andralex:
@jercaianu feel free to mark both overloads as trusted in the StdDdoc version

std/file.d
```diff
@@ -3088,11 +3093,12 @@ else version(Posix)
...
         }
 
-        private this(string path, core.sys.posix.dirent.dirent* fd)
+        private this(string path, core.sys.posix.dirent.dirent* fd) @trusted
         {
             import std.path : buildPath;
+            import std.algorithm.searching : countUntil;
 
-            immutable len = core.stdc.string.strlen(fd.d_name.ptr);
+            immutable len = fd.d_name[].countUntil(0);
```
jmdavis: I'll create a bug report and and a PR to fix this shortly, but using `countUntil` instead of `strlen` is completely incorrect here. I don't know why that was done, but the decoding that `countUntil` does is completely unnecessary, and counting code points and then slicing on code units is fundamentally wrong, and it means that this code no longer correctly handles file names with non-ASCII characters. Their names get truncated.

PR #4894, opened 2016-11-10
Disallow std.variant visiting with handler type parameters not held by variant

ntrel:
See commits.

PR #5759, opened 2017-10-07
Ensure that Variant.visit only accepts handlers that handle allowed types

MetaLang:
This is a simplified reworking of https://github.com/dlang/phobos/pull/4894

### DIFF
--- a/std/variant.d
+++ b/std/variant.d
@@ -2337,6 +2337,11 @@ if (isAlgebraic!VariantType && Handler.length > 0)
                 static if (isSomeFunction!dg)
                 {
                     alias Params = Parameters!dg;
+                    //Make sure the type handled by dg is in AllowedTypes
+                    static if (Params.length > 0)
+                        static assert(VariantType.allowed!(Params[0]),
+                            "Type " ~ Params[0].stringof ~ " does not match any of " ~ AllowedTypes.stringof);
+
                     static if (Params.length == 0)
                     {
                         // Just check exception functions in the first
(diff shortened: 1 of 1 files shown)


MetaLang:
And this is why you test a PR on your local system before committing.

BBasile:
Yeah and just `dmd theModule.d -unittest -main && ./theModule` is usually enough when only 1 module is involved.

andralex:
lgtm, but what's with the autotester failure?

MetaLang:
Unfortunately I think I'll have to close this PR as it breaks some of the `Algebraic` unit tests.

ntrel:
The failing code is:
```
 static assert( __traits(compiles, number.visit!((string x) => x ~ "a", x => x + 1)));
```
This is just after the declaration of `maybenumber`, which is unused. I think these tests are erroneously using `number` instead of `maybenumber`. Fixing that then causes an error that `x => x + 1` cannot be instantiated with `string` (one of `maybenumber.AllowedTypes`). This bug was not hit before as each of `number.AllowedTypes` can instantiate the generic lambda.

MetaLang:
Yes, I've already gone down this rabbit hole and came to the conclusion that this functionality can't be implemented without breaking `Algebraic`. I can't remember what the exact issue was, though, so I will go over it again to double check. However, in general, we're going from more permissive to less permissive, so it's more or less a given that some code will break. It all depends on whether we _want_ to break code that uses `Algebraic` or not.

ntrel:
There seems to be a problem with my setup, so I can't run tests ATM. But, the `maybenumber` test needs to be fixed, and is fairly straightforward using `to!string`: https://github.com/ntrel/phobos/commit/d5c5dbb69948c0d6c5ec7e2725ade703ea076849

PR #5761, opened 2017-10-07
Fix Issue 6244 - Add powmod / modpow function to std.math

jercaianu:
This implementation for powmod correctly handles overflow for unsigned types.
For at most 32-bit inputs, to avoid overflow during modulo sum or multiplication, we can cast the values to 64-bit types and do the operations.

For unsigned 64-bit types I implemented a special mulmod and addmod functionality to avoid overflows.

Tested with boost multiprecision powm function.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -6832,6 +6832,143 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
     assert(approxEqual(pow(twoI, three), 8.0));
 }
 
+/** Computes the value of a positive integer x, raised to the power n, modulo m.
+ *
+ * All values have to be unsigned types, otherwise it will raise a compilation error.
+ */
+Unqual!(Largest!(F, H)) powmod(F, G, H)(F x, G n, H m)
+if (isIntegral!F && isIntegral!G && isIntegral!H &&
+    isUnsigned!F && isUnsigned!G && isUnsigned!H)
+{
+    import std.traits : AliasSeq;
+
+    alias T = Unqual!(Largest!(F, H));
+    static if (T.sizeof <= 4)
+    {
+        alias DoubleT = AliasSeq!(void, ushort, uint, void, ulong)[T.sizeof];
+    }
+
+    static T mulmod(T a, T b, T c)
+    {
+        static if (T.sizeof == 8)
+        {
+            import core.checkedint : addu;
+
+            static T addmod(T a, T b, T c)
+            {
+                b = c - b;
+                if (a >= b)
+                    return a - b;
+                else
+                    return c - b + a;
+            }
+
+            T result = 0, tmp;
+
+            b %= c;
+            while (a > 0)
+            {
+                if (a & 1)
+                    result = addmod(result, b, c);
+
+                a >>= 1;
+                b = addmod(b, b, c);
+            }
+
+            return result;
+        }
+        else
+        {
+            DoubleT result = cast(DoubleT) (cast(DoubleT) a * cast(DoubleT) b);
+            return result % c;
+        }
+    }
+
+    T base = x, result = 1, modulus = m;
+    Unqual!G exponent = n;
+
+    while (exponent > 0)
+    {
+        if (exponent & 1)
+            result = mulmod(result, base, modulus);
+
+        base = mulmod(base, base, modulus);
+        exponent >>= 1;
+    }
+
+    return result;
+}
+
+@safe pure nothrow @nogc unittest
+{
+    ulong a = 18446744073709551615u, b = 20u, c = 18446744073709551610u;
+    assert(powmod(a, b, c) == 95367431640625u);
+    a = 100; b = 7919; c = 18446744073709551557u;
+    assert(powmod(a, b, c) == 18223853583554725198u);
+    a = 117; b = 7919; c = 18446744073709551557u;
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -6832,6 +6832,149 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
...
 }
 
+/** Computes the value of a positive integer x, raised to the power n, modulo m.
+ *
+ * All values have to be unsigned types, otherwise it will raise a compilation error.
+ *
+ */
+Unqual!(Largest!(F, H)) powmod(F, G, H)(F x, G n, H m) @nogc @safe pure nothrow
+if (isIntegral!(F) && isIntegral!(G) && isIntegral!(H) &&
+    is (Unsigned!F == F) && is (Unsigned!G == G) && is (Unsigned!H == H))
```
ZombineDev: The template constraint can be simplified if you use [`isUnsigned`](https://dlang.org/phobos-prerelease/std_traits#isUnsigned) directly. Also when there's a single template argument, the convention is to drop the parentheses.

```D
if (isUnsigned!F && isUnsigned!G && isUnsigned!H)
```

std/math.d
```diff
@@ -6832,6 +6832,149 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
     assert(approxEqual(pow(twoI, three), 8.0));
 }
 
+/** Computes the value of a positive integer x, raised to the power n, modulo m.
+ *
+ * All values have to be unsigned types, otherwise it will raise a compilation error.
+ *
+ */
+Unqual!(Largest!(F, H)) powmod(F, G, H)(F x, G n, H m) @nogc @safe pure nothrow
```
ZombineDev: Template functions should not rely on attributes inference and have unittest that verifies that the correct set of attributes is inferred -> drop `@nogc @safe pure nothrow` here and add the missing attributes to the unittest below.

std/math.d
```diff
@@ -6832,6 +6832,149 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
...
+        }
+        else
+        {
+            DoubleT result = cast(DoubleT) (cast(DoubleT) a * cast(DoubleT) b);
+            return result % c;
+        }
+    }
+
+    T base = x, result = 1, modulus = m;
+    Unqual!(G) exponent = n;
```
ZombineDev: `Unqual!G`

std/math.d
```diff
@@ -6832,6 +6832,149 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
...
+if (isIntegral!(F) && isIntegral!(G) && isIntegral!(H) &&
+    is (Unsigned!F == F) && is (Unsigned!G == G) && is (Unsigned!H == H))
+{
+    import std.stdint : uint8_t, uint16_t, uint32_t, uint64_t;
+
+    alias RetT = Unqual!(Largest!(F, H));
+
+    static if (RetT.sizeof == 1)
+    {
+        alias T = uint8_t;
```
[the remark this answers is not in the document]
ZombineDev: The whole thing can be simplified as:
```D
alias DoubleT = AliasSeq!(void, ushort, uint, void, ulong)[RetT.sizeof];
```

std/math.d
```diff
@@ -6832,6 +6832,141 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
     assert(approxEqual(pow(twoI, three), 8.0));
 }
 
+/** Computes the value of a positive integer x, raised to the power n, modulo m.
+ *
+ * All values have to be unsigned types, otherwise it will raise a compilation error.
+ */
+Unqual!(Largest!(F, H)) powmod(F, G, H)(F x, G n, H m)
+if (isIntegral!F && isIntegral!G && isIntegral!H &&
```
ZombineDev: You don't need `isIntegral`, as `isUnsigned` already allow only a subset of the types for which `isIntegral` is true.

std/math.d
```diff
@@ -6832,6 +6832,141 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
...
+    isUnsigned!F && isUnsigned!G && isUnsigned!H)
+{
+    import std.traits : AliasSeq;
```
ZombineDev: `AliasSeq` is defined in `std.meta` - change the import to: `import std.meta : AliasSeq;`

jercaianu:
@ZombineDev 
Thanks a lot for the comments!

Alex

ZombineDev:
Glad to help, @jercaianu. For more general advise on contributing to phobos, be sure to check https://wiki.dlang.org/Contributing_to_Phobos and of course ask around ;)

std/math.d
```diff
@@ -6832,6 +6832,149 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
...
+ *  Params:
+ *      x = base
+ *      n = exponent
+ *      m = modulus
+ *
+ *  Returns:
+ *      `x` to the power `n`, modulo `m`.
+ *      The return type is the largest between `x`'s and `m`'s type.
+ *
+ * All values have to be unsigned types, otherwise it will raise a compilation error.
```
andralex: Replace with:

The function is only defined for unsigned types and statically rejects all others.

PR #5762, opened 2017-10-07
std.random: No equality checks for doc example unittests

kinke:


ZombineDev:
Perhaps we should make the assertion conditional on the size of `real`?

kinke:
Even leaving the `alias` thing aside, I think checking for exact FP equality here is a bad idea (and no need for reals - just use double for portability; the rhs is a double too btw).

WebDrake:
It does indeed seem dodgy to assume that `Random` is `Mt19937`, although right now the assumption holds true for all supported architectures.  Note that this affects _all_ the equality checks for variates in this unittest example, not just the `real`-valued one.

TBH I think the problems of this unittest come out of a confused purpose: it's really a module-level documentation example, so there's no _need_ for it to check these result values at all.  Checks on the results of deterministically seeded RNGs can be (and are) made in the more in-depth checks of those RNGs themselves.

I'd suggest just cutting all the equality checks on results in this unittest block.  If not I guess `approxEqual` will suffice: `feqrel` is much better for comparing floating-point results, but is harder for the reader to understand, which in this context is probably more important.

std/random.d
```diff
@@ -68,7 +68,8 @@ import std.traits;
 
     // Generate a uniformly-distributed real in the range [0, 100)
     auto r = uniform(0.0L, 100.0L, rnd);
-    assert(r == 79.65429843861011285);
+    import std.math : approxEqual;
+    assert(approxEqual(r, 79.65429843861011285));
```
ZombineDev: Give @WebDrake advice, I think it's better 5o simply remove the equality comparisons and instead assert that the  numbers are in the expected intervals.

WebDrake: @ZombineDev I take it you're talking here only about the floating-point number?  The integer variates generated elsewhere in the unittest should be exactly equal, assuming that `Random` is reliably `Mt19937`.

ZombineDev: Yes, I mean the integer cases too, since I don't think knowledge of the Random's implementation should be assumed for the readers of the first example in the documentation. Testing the exact values should be done only when the exact random engine is explicitly specified, in addition to its seed, IMO.

WebDrake: Oh, so you mean in this case you'd check that the resulting number is `>= 0` and `< 100`, and for the `uniform(0, 15, rnd)` call you'd check that it's `>= 0` and `< 15`?

That sounds reasonable, yes.  Note that for the `uniform!uint` case this won't really work since the range of possible values covers the entire range of possible values for that type; but you don't really have to validate that result at all anyway.

kinke: Done.

PR #5765, opened 2017-10-08
Move some deprecations along.

jmdavis:


ZombineDev:
@dlang/team-phobos In future let's make sure that every PR that removes a deprecated function from phobos, has a corresponding PR to add it to https://github.com/dlang/undeaD. I was too hasty to merge this one and I found the hard way that we need to be more careful (and of course I already submitted a PR for undead: https://github.com/dlang/undeaD/pull/28).

jmdavis:
Historically, we've only put major stuff in undead. If you really want to put stuff in undead, then okay, but in general, I think that it's just creating another place for code to go for folks to complain about being broken, since either undead isn't being maintained, or its existence is defeating part of the purpose of deprecating stuff. We certainly haven't been putting every function that gets deprecated in undead, and I have no interest in making that normal policy.

ZombineDev:
> If you really want to put stuff in undead

It's not that I want, we have no other choice as it doesn't compile otherwise. I can imagine other projects that haven't upgraded their compiler in a long time easily ending up in the same situation.

> I think that it's just creating another place for code to go for folks to complain about being broken, since either undead isn't being maintained

With the ubiquitous access to free continuous integration services, I don't think we have an excuse for not maintaining stuff.

I think it is a question of priorities. I want us to be able aggressively deprecate functionality that is not up to standards of Phobos. For this to happen we want to have an easy migration path for our users.

> We certainly haven't been putting every function that gets deprecated in undead, and I have no interest in making that normal policy.

If you don't care about user experience or the stability of the ecosystem, that's ok. But please don't make it harder for everyone else.

jmdavis:
> I think it is a question of priorities. I want us to be able to aggressively deprecate functionality that is not up to standards of Phobos. For this to happen we want to have an easy migration path for our users.

Users have _two years_ to fix their code after something is deprecated before it's gone from Phobos. That should be more than plenty of time for folks to migrate their code to something else, and if they don't want to, they can always just copy the deprecated stuff into their own projects. Undead was created so that folks could more easily find the old code if they were really slow about upgrading, not so that folks could just use it from the undead repo. Undead is not supposed to be part of a normal migration path, and it's definitely not supposed to be as a means of continuing to use old, deprecated code that was remove from Phobos in part so that the Phobos team wouldn't have to spend time maintaining it.

> If you don't care about user experience or the stability of the ecosystem, that's ok. But please don't make it harder for everyone else.

Honestly, I think that the fact than Undead even exists is a problem. This is stuff that projects shouldn't even be using anymore. So, instead of it going away into oblivion like it's supposed to, it's sitting there for folks to continue to use and where they can complain that it's broken and ask the Phobos team to waste time fixing it. The stuff that's in undead shouldn't even be part of the normal user experience or general ecosystem. IMHO, if there is anything on code.dlang.org which depends on Undead outside of copying it and maintaining it as part of those projects, that's a serious problem.

And it's only going to get worse if/when stuff like DIP 1000 becomes the normal behavior and a bunch of code has to be updated accordingly.

CyberShadow:
> Users have two years to fix their code after something is deprecated before it's gone from Phobos.

This assumes that all D code that is going to be used has maintainers that are constantly updating it and have an arbitrary amount of time at their disposal in order to do so.

The reality can be very different. Sometimes you have a program written 5 years ago that has worked just fine in its compiled form, and now needs a few updates in some places here and there. Sometimes these updates must be done under strong time pressure (i.e. things are on fire and they need to be fixed urgently). 

There **absolutely** needs to be an easy migration path for all deprecated things that can reasonably have one. This is not up for discussion - by not providing one we'd be doing a great disservice to our users and erode trust in D's longevity at the same time. Even very small functions are thus worth moving to Undead for this purpose, because adding/updating imports is still much easier than finding the functions' definitions and pasting them in the right place, especially when it needs to be done for multiple definitions.

jmdavis:
If a project that you're trying to build hasn't been maintained, then use whatever version of the compiler it was developed with. IMHO, it's utterly ridiculous to have to maintain everything that was ever deprecated from Phobos for all time just because someone might try and compile an old, unmaintained project with a newer compiler.

CyberShadow:
> If a project that you're trying to build hasn't been maintained, then use whatever version of the compiler it was developed with.

That's not always an option. It's not always obvious or even possible to install an older D compiler version (see e.g. the recent issues with libbfd and fPIC). Sometimes the changes require integrating with a newer library which does require a newer compiler version.

> IMHO, it's utterly ridiculous to have to maintain everything that was ever deprecated from Phobos for all time just because someone might try and compile an old, unmaintained project with a newer compiler.

If the amount of work such maintenance requires is not large, then it's not an issue. But if some language change would require considerable effort to update Undead. then it's a good indication that the change is likely to be a bad idea. Undead consists of mostly simple, D1-like code that really should not be broken very often at all.

PR #5766, opened 2017-10-08
Fix Issue 4763 - std.stdio.File.open() : more efficient implementation

jercaianu:
Everytime open or popen are called, we reuse the File object instead of creating a new one.
Also, in some cases, we avoid extra calls to malloc for the Impl struct.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -534,6 +534,7 @@ Throws: $(D ErrnoException) in case of error.
 
     private void closeHandles() @trusted
     {
+        assert(_p);
         import std.exception : errnoEnforce;
 
         version (Posix)
@@ -547,6 +548,7 @@ Throws: $(D ErrnoException) in case of error.
                 errnoEnforce(res != -1,
                         "Could not close pipe `"~_name~"'");
                 _p.handle = null;
+                return;
             }
         }
         if (_p.handle)
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -370,6 +370,11 @@ struct File
 
         assert(!_p);
         _p = cast(Impl*) enforce(malloc(Impl.sizeof), "Out of memory");
+        initImpl(handle, name, refs, isPopened);
+    }
+
+    private void initImpl(FILE* handle, string name, uint refs = 1, bool isPopened = false)
+    {
         _p.handle = handle;
         _p.refs = refs;
         _p.isPopened = isPopened;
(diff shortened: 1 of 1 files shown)


jercaianu:
Worth noting that I tested opening 50000 files and I got a speed improvement from 17 seconds to 2 seconds.
Although this may not be a real life situation, it shows that there is indeed a performance gain if avoiding destructing and creating File objects each time we call open.

CyberShadow:
> Worth noting that I tested opening 50000 files and I got a speed improvement from 17 seconds to 2 seconds.

Wow, that's huge! Definitely something worth fixing. We should not be spending more time in D code than the OS for I/O operations.

However, I'm worried about the amount of code this adds, and the entire idea of carefully mutating existing state just to simulate it being initialized from scratch. Generally such approaches have been fragile, both in terms of initial implementation (there seems to be a non-trivial amount of logic involved) and in keeping up with future changes (any changes to the state model need to be reflected in the logic across the board).

If you haven't done so, do you think it's possible to profile the current implementation and see if there are any simpler approaches to reduce unnecessary overhead?

> Your PR doesn't reference any Bugzilla issue.

The PR title is good, but the commit message isn't. Please copy the PR title to the commit message (using rebase/amend and force-push).

jercaianu:
Thanks for your comments!
I will check and see how can improvements be done in a way that is beneficial to everyone.

 However, the amount of new logic that I added in this PR is almost nonexistent. That is because I "duplicated" the logic in the destructor/constructor but without redundant allocations.

Thanks,
Alex

andralex:
@CyberShadow made a careful pass through the code with @jercaianu and it seems the complexity of the code (after a few more simplifications) is commensurate with the work done.

jercaianu:
@andralex @CyberShadow
Since this was not merged yet, I also added the fix for issue 10395.
In case of failure popen returns -1, therefore the enforce that I removed in my latest commit was not correct.

Can you please check again ?
Thanks,
Alex

std/stdio.d
```diff
@@ -472,17 +470,107 @@ file.
...
+        version (Posix)
+        {
+            import core.sys.posix.stdio : pclose;
+            import std.format : format;
+
+            if (_p.isPopened)
+            {
+                auto res = pclose(_p.handle);
+                errnoEnforce(res != -1,
+                        "Could not close pipe `"~_name~"'");
```
andralex: No need for a temporary here, suffices to say:

```d
errnoEnforce(pclose(_p.handle) != -1,
                        "Could not close pipe `"~_name~"'");
```

But it's a minor point.

std/stdio.d
```diff
@@ -472,17 +470,107 @@ file.
...
+        version (MICROSOFT_STDIO)
+        {
+            setAppendWin(stdioOpenmode);
+        }
+    }
+
+    private void closeHandles() @trusted
+    {
+        import std.exception : errnoEnforce;
+
```
andralex: I guess `assert(_p);` is in order here.

PR #4272, opened 2016-04-30
std.math: Use core.stdc.fenv for exception mask values

ibuclaw:
See also: https://github.com/dlang/druntime/pull/1554

`FloatingPointControl` rounding control and interrupt masks should be given similar treatment.  Where their values all correspond to macros from `fpu_control.h` in libc (there does exist any binding in druntime).

PR #5769, opened 2017-10-08
Small clean-ups of architecture-specific code in FloatingPointControl

ibuclaw:
Slight continuation of #4272.

Changes:

- `FloatingPointControl.roundingMask`: New public RoundingMode value
- `FloatingPointControl.ExceptionMask`: New public base type for exception mask enum.
- `EXCEPTION_MASK`, `ROUNDING_MODE`: Remove internal values.

Now, the only parts that need changing for adding ports are FPU exception masks, and the asm code which gets/set the control word.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -4946,34 +4946,25 @@ struct FloatingPointControl
             roundToZero /// ditto
         }
     }
-    else version(ARM)
+    else version (CRuntime_Microsoft)
     {
-        enum : RoundingMode
-        {
-            roundToNearest = 0x000000,
-            roundDown      = 0x800000,
-            roundUp        = 0x400000,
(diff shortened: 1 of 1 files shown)


ibuclaw:
Ping?

std/math.d
```diff
@@ -4939,60 +4939,60 @@ struct FloatingPointControl
...
              * The default mode is roundToNearest.
+             *
+             *  roundingMask = A mask of all rounding modes.
              */
             roundToNearest,
             roundDown, /// ditto
             roundUp, /// ditto
-            roundToZero /// ditto
+            roundToZero, /// ditto
+            roundingMask, /// ditto
```
andralex: nice touch with that comma

PR #5771, opened 2017-10-09
Explore additions to phobos with DIP 1011 extern(delegate)

marler8997:


wilzbach:
@marler8997 I am going to close this proof of concept as it fills up the PR queue and won't be merged in this form anyhow. However, please don't take this as discouragement for your DIP - after all the diff will still be as available.

PR #5773, opened 2017-10-10
std.math: Fix names of PI/4 constants for RealFormat.ieeeQuadruple

ibuclaw:
Introduced by #5459.

FYI @klickverbot as the author of the referenced PR.


ibuclaw:
Ping?  This fixes a build failure on 128-bit real platforms.

PR #5774, opened 2017-10-10
Fix Issue 10131 - To remove duplicates and keep order

RazvanN7:
I'm not sure if this function should be added as an overload to uniq.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -189,7 +189,7 @@ PACKAGE_std = array ascii base64 bigint bitmanip compiler complex concurrency \
   outbuffer parallelism path process random signals socket stdint \
   stdio string system traits typecons uni \
   uri utf uuid variant xml zip zlib
-PACKAGE_std_experimental = checkedint typecons
+PACKAGE_std_experimental = checkedint typecons algorithm
 PACKAGE_std_algorithm = comparison iteration mutation package searching setops \
   sorting
 PACKAGE_std_container = array binaryheap dlist package rbtree slist util
--- a/std/experimental/algorithm.d
+++ b/std/experimental/algorithm.d
@@ -1,0 +1,49 @@
+// Written in the D programming language.
+
+/**
+This module implements experimental additions/modifications to $(MREF std, _algorithm).
+
+Use this module to test out new functionality for $(REF wrap, std, _algorithm)
+
+Source:    $(PHOBOSSRC std/experimental/_algorithm.d)
+
+Copyright: Copyright the respective authors, 2008-
+License:   $(HTTP boost.org/LICENSE_1_0.txt, Boost License 1.0).
+Authors:   Andrea Fontana
+           Razvan Nitu
+ */
+module std.experimental.algorithm;
+
+import std.range : isInputRange, ElementType;
+
+/**
+Lazily produces a range with no duplicated elements from the input
+range $(D r). The order of elements is the original order.
+
+Params:
+    Range r - the input range to eliminate duplicates from
+
+Returns:
+    A range containing all the unique elements
+*/
+auto distinct(Range)(Range r)
+if (isInputRange!Range)
+{
+    import std.algorithm.iteration : filter;
+    bool[ElementType!Range] justSeen;
+    return r.filter!((k)
+        {
+            if (k in justSeen) return false;
+            return justSeen[k] = true;
+        });
+}
+
+///
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+
+    assert("AAAA".distinct.equal("A"));
+    assert("ABAC".distinct.equal("ABC"));
+    assert([1, 2, 3, 1, 2, 1, 1].distinct.equal([1, 2, 3]));
+}
--- a/win32.mak
+++ b/win32.mak
@@ -293,7 +293,7 @@ SRC_STD_INTERNAL_WINDOWS= \
 	std\internal\windows\advapi32.d
 
 SRC_STD_EXP= \
-	std\experimental\checkedint.d std\experimental\typecons.d
+	std\experimental\checkedint.d std\experimental\typecons.d std\experimental\algorithm.d
 
 SRC_STD_EXP_ALLOC_BB= \
 	std\experimental\allocator\building_blocks\affix_allocator.d \


BBasile:
`deduplicate()` ?

RazvanN7:
@BBasile maybe it's just me, but deduplicate sounds very confusing.

RazvanN7:
@wilzbach why is dscanner failing? The code is valid user code which should be unittested.

std/algorithm/iteration.d
```diff
@@ -5185,3 +5185,100 @@ if (isRandomAccessRange!Range && hasLength!Range)
...
+if the $(D opEquals) operator is not overriden for classes, $(D noDupes) considers the class
+pointer for duplication removal.
+*/
+T[] noDupes(T)(T[] s)
+{
+    static if (is(T == class) && !__traits(isOverrideFunction, T.toHash))
+    {
+        import std.algorithm : canFind;
+        T[] result;
+        foreach (T elem; s)
```
BBasile: first element of `s` can be inserted, always, unconditionally. (in the other else part of the static branch too)

RazvanN7: this would imply checking manually for an empty array, instead the foreach does that for you, plus the code is more readable this way

BBasile:
@RazvanN7: about `deduplicate()`, no it's not confusing it's kinda of a valid english word that tells well what the function does and that's used in programming ([1](http://www.programming-idioms.org/idiom/119/deduplicate-list), [2](http://www.oracle.com/webfolder/technetwork/data-quality/edqhelp/Content/processor_library/matching/match_deduplication.htm) and probably much more)

std/algorithm/iteration.d
```diff
@@ -5185,3 +5185,100 @@ if (isRandomAccessRange!Range && hasLength!Range)
...
+            {
+                set[elem] = true;
+                result ~= elem;
+            }
+        return result;
+    }
+}
+
+///
+@system unittest
```
BBasile: it's not safe to deduplicate an array of int ?

std/algorithm/iteration.d
```diff
@@ -5185,3 +5185,100 @@ if (isRandomAccessRange!Range && hasLength!Range)
...
 }
+/*
+Returns the array without any duplicates, keeping the initial order of appearance.
+If $(D T) is a basic type, a struct or a class which overrides the $(D toHash) operator
+the resulting array is computed in $(D O(n)) steps. If $(D T) is a class which does not
+override the $(D toHash) operator, the result is computed in $(D O(n^2)) steps. Aditionally,
+if the $(D opEquals) operator is not overriden for classes, $(D noDupes) considers the class
+pointer for duplication removal.
+*/
+T[] noDupes(T)(T[] s)
```
BBasile: why not `auto ref ` ?

std/algorithm/iteration.d
```diff
@@ -5185,3 +5185,100 @@ if (isRandomAccessRange!Range && hasLength!Range)
          [1, 2, 0],
          [2, 1, 0]]));
 }
+/*
+Returns the array without any duplicates, keeping the initial order of appearance.
```
BBasile: `Params` and `Returns` standard DDOC sections ?

std/algorithm/iteration.d
```diff
@@ -5185,3 +5185,100 @@ if (isRandomAccessRange!Range && hasLength!Range)
...
+        T[] result;
+        foreach (T elem; s)
+            if (!result.canFind(elem))
+                result ~= elem;
+        return result;
+    }
+    else
+    {
+        bool[T] set;
+        T[] result;
```
BBasile: can't an `Appender` be possible here ? we know how many to reserve (top).

std/algorithm/iteration.d
BBasile: By the way the content here is quite horrid for a documentation unittest

std/algorithm/iteration.d
```diff
@@ -5185,3 +5185,100 @@ if (isRandomAccessRange!Range && hasLength!Range)
...
+        foreach (T elem; s)
+            if (elem !in set)
+            {
+                set[elem] = true;
```
BBasile: I'm not sure if this strategy is something really clever with struct that are postblitable. The postblit will run once in the AA and then a second time in the result.

std/algorithm/iteration.d
BBasile: Aw and `inout` ? Since deduplication doesn't happend in place ?

trikko:
I have a custom function for this:

```D
auto distinct(Range)(Range r) if (isInputRange!Range)
{
   bool[ForeachType!Range] justSeen;
   return r.filter!(
      (k) {
         if (k in justSeen) return false;
         return justSeen[k] = true;
      }
   );
}
```
It works in a lazy way, so it works for infinite ranges too like: 
```d
roundRobin(sequence!"n"(0), sequence!"n"(0)).distinct.take(10).writeln;
```

and doesn't read the whole array if not needed.

wilzbach:
> @wilzbach why is dscanner failing? The code in the unittest is valid user code.

Have a look at the error messages:

```
std/algorithm/iteration.d(5198:5)[warn]: Public declaration 'noDupes' is undocumented.
std/algorithm/iteration.d(5233:25)[warn]: Methods 'opCmp', 'toHash', 'opEquals', 'opCast', and/or 'toString' are non-const.
std/algorithm/iteration.d(5234:23)[warn]: Methods 'opCmp', 'toHash', 'opEquals', 'opCast', and/or 'toString' are non-const.
std/algorithm/iteration.d(5244:11)[warn]: 'B' has method 'toHash', but not 'opEquals'.
std/algorithm/iteration.d(5249:25)[warn]: Methods 'opCmp', 'toHash', 'opEquals', 'opCast', and/or 'toString' are non-const.
std/algorithm/iteration.d(5257:11)[warn]: 'C' has method 'opEquals', but not 'toHash'.
std/algorithm/iteration.d(5262:23)[warn]: Methods 'opCmp', 'toHash', 'opEquals', 'opCast', and/or 'toString' are non-const.
```

Regarding `distinct`
---------------------------

Btw I have something similar to @trikko's solution as well. However, as of now the paradigm is that `std.algorithm` should be `@nogc` in theory (not all functions are this in practice, but they could), so I have never considered suggesting it.
I would be in favor of considering to add such a utility function like `distinct`, but I think we should go for `std.experimental.algorithm` (or similar), so that we are flexible and can experiment with adding an allocator API to it.

Regarding this PR
------------------------

FWIW there are at least these valid approaches:
A) `sort.uniq` ("Phobos out of the box")
B) filter + AA (e.g. by @trikko)

(A) is good if you want to be `@nogc` or don't have many duplicates, but (B) is the lazy and usually faster approach. Imho this PR combines both disadvantages of (A) and (B)

wilzbach:
@RazvanN7 please don't forget to add a link to your PR to the regarding Bugzilla issue, s.t. people in the future know about all work done and duplicate the work (I have added it for this one).

See also: https://github.com/dlang/dmd/pull/7205

RazvanN7:
@wilzbach 

Regarding circleCi : since the compiler allows having to hash without opEquals why does dscanner complain about that? I admit that having a non-const toHash, opCmp etc. is most likely a mistake. 

Other then that, I haven't thought about the @nogc aspect. I will put @trikko 's code into std/algorithm/experimental and hopefully that will close the bug.

Thanks for the extensive explanation. I'm pretty noobish when it comes to function annotations and your explanations are extremely helpful.

PR #5775, opened 2017-10-10
Move new top-level selective imports in std.uni to local scopes

joakim-noah:
Dmitry said he had problems with locally scoping these, [see issue 17813](https://issues.dlang.org/show_bug.cgi?id=17813), but I haven't been able to reproduce on linux/x64 with the latest dmd 2.076.1.  Let's see what the auto-tester says.


joakim-noah:
The single `std.socket` failure on the Mac auto-tester seems unrelated and spurious.

joakim-noah:
Thanks, the associated issue can also be closed.  I didn't say this pull fixed it in the commit, because there was nothing to fix.  I simply moved the imports and there was no bug.

PR #5776, opened 2017-10-10
Fix Issue 16487 - Add function to obtain the available disk space

jercaianu:
removed redundant line

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -5264,7 +5264,7 @@ ulong getAvailableDiskSpace(scope const(char)[] path) @safe
     }
     else version (Posix)
     {
-        import std.string : toStringz;
+        import std.internal.cstring : tempCString;
 
         version (FreeBSD)
         {
@@ -5272,7 +5272,7 @@ ulong getAvailableDiskSpace(scope const(char)[] path) @safe
 
             statfs_t stats;
             auto err = () @trusted {
-                return statfs(path.toStringz(), &stats);
+                return statfs(path.tempCString(), &stats);
             } ();
             cenforce(err == 0, "Cannot get available disk space");
 
@@ -5284,7 +5284,7 @@ ulong getAvailableDiskSpace(scope const(char)[] path) @safe
 
             statvfs_t stats;
             auto err = () @trusted {
-                return statvfs(path.toStringz(), &stats);
+                return statvfs(path.tempCString(), &stats);
             } ();
             cenforce(err == 0, "Cannot get available disk space");
 
--- a/std/file.d
+++ b/std/file.d
@@ -5246,7 +5246,7 @@ Returns:
 Throws:
     $(LREF FileException) in case of failure
 */
-ulong getAvailableDiskSpace(string path) @safe
+ulong getAvailableDiskSpace(scope const(char)[] path) @safe
 {
     version (Windows)
     {
--- a/std/file.d
+++ b/std/file.d
@@ -62,6 +62,7 @@ $(TR $(TD Other) $(TD
           $(LREF FileException)
           $(LREF PreserveAttributes)
           $(LREF SpanMode)
+          $(LREF getAvailableDiskSpace)
 ))
 )
 
@@ -5231,4 +5232,75 @@ string tempDir() @trusted
 
     myFile.write("hello");
     assert(myFile.readText == "hello");
+}
+
+/**
+Returns the available disk space based on a given path.
+On Windows, `path` must be a directory; on Posix systems, it can be a file or directory.
+
+Params:
+    path = on Windows, it must be a directory; on Posix it can be a file or directory
+Returns:
+    Available space in bytes
+
+Throws:
+    $(LREF FileException) in case of failure
+*/
+ulong getAvailableDiskSpace(string path) @safe
+{
+    version (Windows)
+    {
+        import core.sys.windows.winbase : GetDiskFreeSpaceExW;
+        import core.sys.windows.winnt : ULARGE_INTEGER;
+        import std.internal.cstring : tempCStringW;
+
+        ULARGE_INTEGER freeBytesAvailable;
+        auto err = () @trusted {
+            return GetDiskFreeSpaceExW(path.tempCStringW(), &freeBytesAvailable, null, null);
+        } ();
+        cenforce(err != 0, "Cannot get available disk space");
+
+        return freeBytesAvailable.QuadPart;
+    }
+    else version (Posix)
+    {
+        import std.string : toStringz;
+
+        version (FreeBSD)
+        {
+            import core.sys.posix.sys.statvfs : statfs, statfs_t;
+
+            statfs_t stats;
+            auto err = () @trusted {
+                return statfs(path.toStringz(), &stats);
+            } ();
+            cenforce(err == 0, "Cannot get available disk space");
+
+            return stats.f_bavail * stats.f_bsize;
+        }
+        else
+        {
+            import core.sys.posix.sys.statvfs : statvfs, statvfs_t;
+
+            statvfs_t stats;
+            auto err = () @trusted {
+                return statvfs(path.toStringz(), &stats);
+            } ();
+            cenforce(err == 0, "Cannot get available disk space");
+
+            return stats.f_bavail * stats.f_frsize;
+        }
+    }
+    else static assert(0, "Unsupported platform");
+}
+
+///
+@safe unittest
+{
+    import std.exception : assertThrown;
+
+    auto space = getAvailableDiskSpace(".");
+    assert(space > 0);
+
+    assertThrown!FileException(getAvailableDiskSpace("ThisFileDoesNotExist123123"));
 }
--- a/std/file.d
+++ b/std/file.d
@@ -5268,7 +5268,7 @@ ulong getAvailableDiskSpace(scope const(char)[] path) @safe
 
         version (FreeBSD)
         {
-            import core.sys.posix.sys.statvfs : statfs, statfs_t;
+            import core.sys.freebsd.sys.mount : statfs, statfs_t;
 
             statfs_t stats;
             auto err = () @trusted {
--- a/changelog/std-file-getAvailableDiskSpace.dd
+++ b/changelog/std-file-getAvailableDiskSpace.dd
@@ -1,0 +1,17 @@
+Added the `std.file.getAvailableDiskSpace` functionality.
+
+$(REF getAvailableDiskSpace, std,file) receives as a parameter the path of a file or
+directory in the file system, and returns the available disk space on the mounted filesystem.
+If the given path is nonexistent, an exception is thrown.
+
+---
+import std.file;
+ulong size = getAvailableDiskSpace(".");
+assert(size > 0);
+---
+
+---
+import std.file;
+assertThrown(getAvailableDiskSpace("NonExistentFile"));
+---
+


std/file.d
```diff
@@ -4323,3 +4323,65 @@ string tempDir() @trusted
...
+        return stats.f_bavail * stats.f_frsize;
+    }
+}
+
+@safe unittest
+{
+    auto space = getAvailableDiskSpace(".");
+    assert(space >= 100000);
+
+    try {
```
BBasile: curly brace on its own line PLZ (D style)

std/file.d
```diff
@@ -4323,3 +4323,65 @@ string tempDir() @trusted
...
+        space = getAvailableDiskSpace("ThisFileDoesNotExist123123");
+        assert(0);
+    } catch (Exception e){
```
BBasile: brace style

std/file.d
```diff
@@ -4323,3 +4323,65 @@ string tempDir() @trusted
...
+        assert(1);
```
BBasile: huh ? it's `assert(0)` for non reachable-ility !!!!

jercaianu: Thanks for the comments!
I wanted to test here that it throws exception in case of an invalid path

BBasile: there's `assertThrown` for this, in std.exception.

std/file.d
```diff
@@ -4323,3 +4323,62 @@ string tempDir() @trusted
...
+        import core.sys.windows.winnt;
+        import std.internal.cstring;
+
+        ULARGE_INTEGER freeBytesAvailable;
+        auto err = GetDiskFreeSpaceExW(path.tempCStringW(), &freeBytesAvailable, null, null);
+        enforce(err != 0, text("Cannot get available disk space: ", GetLastError()));
+
+        return freeBytesAvailable.QuadPart;
+    }
+    else
```
BBasile: else `version(POSIX)`

std/file.d
```diff
@@ -4323,3 +4323,62 @@ string tempDir() @trusted
...
+    {
+        import core.sys.posix.sys.statvfs;
+        import std.string: toStringz;
+
+        statvfs_t stats = void;
+        auto err = statvfs(path.toStringz(), &stats);
+        enforce(err == 0, text("Cannot get available disk space: ", err));
+
+        return stats.f_bavail * stats.f_frsize;
+    }
```
BBasile: `else static assert(0, "unsupported platform")`

though the message is not 100 % required.

std/file.d
```diff
@@ -4323,3 +4323,62 @@ string tempDir() @trusted
...
+}
+
+@safe unittest
+{
+    import std.exception : assertThrown;
+
+    auto space = getAvailableDiskSpace(".");
+    assert(space >= 100000);
```
nemanja-boric-sociomantic: Better go with the `assert(space != 0), although that could fail as well.

std/file.d
BBasile: sorry it's `Posix`.

BBasile: I mean the identifier...not fully Upper case.

std/file.d
wilzbach: It's doesn't hurt and would be helpful to the user...

std/file.d
```diff
@@ -4323,3 +4323,62 @@ string tempDir() @trusted
...
+
+        statvfs_t stats = void;
+        auto err = statvfs(path.toStringz(), &stats);
+        enforce(err == 0, text("Cannot get available disk space: ", err));
+
+        return stats.f_bavail * stats.f_frsize;
+    }
+}
+
+@safe unittest
```
wilzbach: Please use `///` to mark it as a documented unittest. It then will appear on dlang.org in the documentation (you can use DAutoTest to preview the generated documentation or of course build dlang.org yourself).

std/file.d
```diff
@@ -4323,3 +4323,62 @@ string tempDir() @trusted
...
     return cache;
 }
+
+/**
+ * Returns the available space in a file system.
+ *
+ * On Windows, this function uses the result of the Windows API function
+ * $(LINK2 https://msdn.microsoft.com/en-us/library/windows/desktop/aa364937(v=vs.85).aspx, $(D GetDiskFreeSpaceExW)).
+ *
+ * On POSIX platforms, it uses the $(LINK2 https://msdn.microsoft.com/en-us/library/windows/desktop/aa364937(v=vs.85).aspx,
```
joakim-noah: Link is the same as above, presumably you have another in mind.

jercaianu:
Thanks for the comments, I'll shortly look into them as soon as I figure out what's going on with the FreeBSD failure.

Also, since we're here, do you think it would be a good idea to add functionality for getFreeDiskSpace or getDiskSpaceCapacity (similar to boost::filesystem) ?

Thanks,
Alex

jercaianu:
I have figured out what's wrong with the FreeBSD test.
It seems that statvfs does not work on FreeBSD as expected, but statfs works.

The question is now, should I add statfs in the runtime, or just assume return 0 for FreeBSD?

Thanks,
Alex

wilzbach:
> The question is now, should I add statfs in the runtime, or just assume return 0 for FreeBSD?

Extending the bindings in druntime is almost always welcome as *BSD is vastly lacking behind in terms of maintenance manpower (iirc only one or two of the core people around here use it regularly).

n8sh:
>The question is now, should I add statfs in the runtime, or just assume return 0 for FreeBSD?

You throw an exception on failure, so isn't the correct behavior to throw an exception?

std/file.d
```diff
@@ -4323,3 +4323,62 @@ string tempDir() @trusted
...
+ * Params:
+ *     path = a valid path to any file in the file system
+ *
+ * Returns:
+ *    ulong which contains the available space in bytes
+ *
+ * Throws:
+ *     `Exception` in case statvfs (POSIX) or GetDiskFreeSpaceExW (Windows) fails.
+ */
+ulong getAvailableDiskSpace(string path) @trusted
```
quickfur: I'm automatically scared of putting `@trusted` on a non-trivial function.  Wouldn't it be better to put `@trusted` around smaller blocks of code, i.e., around the relevant OS API calls?

std/file.d
quickfur: There needs to be an `else static assert` otherwise the function will refuse to compile on an unsupported platform (the else path will not return anything, but the function must return `ulong`).

std/file.d
quickfur: Yeah, there is no way to predict available space in advance.  Certainly, there is no guarantee at all that it will be `>= 100_000`. At least `> 0` has only a vanishingly small chance of failing.

quickfur:
Tagging @andralex since this adds a new public symbol.

quickfur:
ping @jercaianu   This has stalled for far too long.  Any chance we can get things moving again?

jercaianu:
@quickfur 
Hi, the reason I can't move forward with this, is because I need to add statvfs on FreeBSD [1].
However, this in turn, is blocked by this one [2].

If statvfs would be added, then I can easily finish this.

[1] - https://github.com/dlang/druntime/pull/1937
[2] - https://issues.dlang.org/show_bug.cgi?id=17596#c8

n8sh:
@jercaianu I think it would be reasonable to move forward and add a note to the documentation that this function doesn't work on FreeBSD, since the function has a way to signal failure.

n8sh:
FreeBSD could be added later.

JackStouffer:
>FreeBSD could be added later.

Yes, this is acceptable for file based functions, as we already have posix and windows only functions.

jercaianu:
@JackStouffer @n8sh 
Thanks for the advice, I'll update this PR as soon as possible, with the mention that it will not work correctly on FreeBSD

jercaianu:
Should I static assert(0) for FreeBSD, or just mention that it may not work correctly?

std/file.d
```diff
@@ -5119,3 +5120,73 @@ string tempDir() @trusted
...
+        return freeBytesAvailable.QuadPart;
+    }
+    else version (Posix)
+    {
+        import core.sys.posix.sys : statvfs;
+        import std.string: toStringz;
+
+        statvfs_t stats = void;
+        auto err = statvfs(path.toStringz(), &stats);
+        enforce(err == 0, text("Cannot get available disk space: ", err));
```
JackStouffer: `enforce!FileException`

std/file.d
```diff
@@ -5119,3 +5120,73 @@ string tempDir() @trusted
...
+
+    version (Windows)
+    {
+        import core.sys.windows.winbase : GetDiskFreeSpaceExW, GetLastError;
+        import core.sys.windows.winnt : ULARGE_INTEGER;
+        import std.internal.cstring : tempCStringW;
+
+        ULARGE_INTEGER freeBytesAvailable;
+        auto err = GetDiskFreeSpaceExW(path.tempCStringW(), &freeBytesAvailable, null, null);
+        enforce(err != 0, text("Cannot get available disk space: ", GetLastError()));
```
JackStouffer: `enforce!FileException`

std/file.d
```diff
@@ -5119,3 +5120,73 @@ string tempDir() @trusted
...
+ * `getAvailableDiskSpace` may not work correctly on `FreeBSD`.
+ *
+ * Params:
+ *     path = a valid path to any file in the file system
+ *
+ * Returns:
+ *    ulong which contains the available space in bytes
+ *
+ * Throws:
+ *     `Exception` in case statvfs (`POSIX`) or GetDiskFreeSpaceExW (`Windows`) fails.
```
JackStouffer: \`Exception\` => $(LREF FileException)

JackStouffer:
I'd static assert on it.

joakim-noah:
That'll break the auto-tester, just mention it.

std/file.d
```diff
@@ -5119,3 +5120,74 @@ string tempDir() @trusted
     myFile.write("hello");
     assert(myFile.readText == "hello");
 }
+
+
+version(FreeBSD) {}
```
JackStouffer: This should be inside of the function body and part of the already existing version branches so documentation generates on all platforms.

And there should be a static assert with a message inside this branch.

JackStouffer:
@joakim-noah Just version out the tests. We do that all over the place already.

std/file.d
jercaianu: And this is guaranteed not to trigger if we're not using the function?

JackStouffer: Yes, if it's never called on FreeBSD the assert will not trigger.

jercaianu: Hi, It seems the tester triggers the assert.
Can you please help me figure out why?

Thanks

std/file.d
```diff
@@ -5119,3 +5120,77 @@ string tempDir() @trusted
...
+* Params:
+*     path = a valid path to any file in the file system
+*
+* Returns:
+*    ulong which contains the available space in bytes
+*
+* Throws:
+*     $(LREF FileException) in case statvfs (`POSIX`) or GetDiskFreeSpaceExW (`Windows`) fails.
+*/
+ulong getAvailableDiskSpace(string path) @trusted
```
wilzbach: Needs to be a template to avoid instantiation by default.
Also note that the other functions accept `const(char)[]` which is different to `immutable(char)[]` which is `string`.

std/file.d
```diff
@@ -5119,3 +5120,77 @@ string tempDir() @trusted
...
+* `FreeBSD` does not support `getAvailableDiskSpace`.
+*
+* Params:
+*     path = a valid path to any file in the file system
+*
+* Returns:
+*    ulong which contains the available space in bytes
+*
+* Throws:
+*     $(LREF FileException) in case statvfs (`POSIX`) or GetDiskFreeSpaceExW (`Windows`) fails.
```
wilzbach: No leading stars please: https://dlang.org/dstyle.html#phobos_documentation

std/file.d
```diff
@@ -5119,3 +5120,77 @@ string tempDir() @trusted
...
 }
+
+/**
+* Returns the available space in a file system.
+*
+* On `Windows`, this function uses the result of the Windows API function
+* $(LINK2 https://msdn.microsoft.com/en-us/library/windows/desktop/aa364937(v=vs.85).aspx, $(D GetDiskFreeSpaceExW)).
+*
+* On `POSIX` platforms, it uses the $(LINK2 http://man7.org/linux/man-pages/man3/statvfs.3.html,
+* $(D statvfs)) function.
```
wilzbach: A user doesn't care about this. Describe the differences / implication if you want to, but do not just link to other documentation.

std/file.d
```diff
@@ -5119,3 +5120,77 @@ string tempDir() @trusted
...
+*
+* Due to inconsistent behaviour with the `statvfs` `Posix` function,
+* `FreeBSD` does not support `getAvailableDiskSpace`.
```
wilzbach: That's not true, we were essentially just to lazy to do this.

> At the moment, `getAvailableDiskSpace` is not supported on FreeBSD.

std/file.d
```diff
@@ -5119,3 +5120,77 @@ string tempDir() @trusted
...
+*
+* Params:
+*     path = a valid path to any file in the file system
+*
+* Returns:
+*    ulong which contains the available space in bytes
```
wilzbach: Available space in `bytes` (the user already sees the return type and we typically don't mention it in the return type)

std/file.d
```diff
@@ -5119,3 +5120,77 @@ string tempDir() @trusted
...
+        return stats.f_bavail * stats.f_frsize;
+    }
+    else static assert(0, "Unsupported platform");
+}
+
+version (FreeBSD) {}
+else
+{
+    ///
+    @safe unittest
```
wilzbach:
```
version (FreeBSD) {} else
@safe unittest
```

is typically used for this. However, the documentation checker enforces that the examples on the documentation work on all platforms, so this needs to be moved in the test for now (or the documentation example checker needs to be adapted to support `version` blocks).

std/file.d
```diff
@@ -5119,3 +5120,77 @@ string tempDir() @trusted
...
+    {
+        import std.exception : assertThrown;
+
+        auto space = getAvailableDiskSpace(".");
+        assert(space > 0);
+
+        assertThrown(getAvailableDiskSpace("ThisFileDoesNotExist123123"));
```
wilzbach: `assertThrown!FileException`

std/file.d
wilzbach: Oh and ideally the `@trusted` goes inline to only to the needed lines

std/file.d
```diff
@@ -5119,3 +5120,77 @@ string tempDir() @trusted
...
+*     $(LREF FileException) in case statvfs (`POSIX`) or GetDiskFreeSpaceExW (`Windows`) fails.
+*/
+ulong getAvailableDiskSpace(string path) @trusted
+{
+    import std.exception : enforce;
+    import std.conv : text;
+
+    version (FreeBSD)
+    {
+        static assert(0, "Unsupported platform");
```
wilzbach: FYI: You could put the `extern(C)` declarations from your druntime PR here inline.
It seems like getting D to work with FreeBSD12 is a lot of work anyways (and not supported at the moment) and in this way `getAvailableDiskSpace` at least supports "all currently supported versions" of FreeBSD and this can be tweaked later once a solution to the druntime PR is found.

wilzbach:
From the discussion. Here's Walter's summary:

> We don't seem to have a good solution for backwards compatibility, or anyone particularly interested in supporting it. So we should just update the structs, abandon the old FreeBSDs, and move on with the newer ones.

So I suggest to merge this and put the newer version in a `FreeBSD12` version block, s.t. once we switch to FreeBSD12 and abandon FreeBSD11 it's a simple removal.

CC @jmdavis @dkgroot

std/file.d
jmdavis: Please no. We have sometimes put `extern(C)` declarations in Phobos, but in general, they should go in druntime, and every one of them that ends up in Phobos, it's something that has to be fixed later. And in the case of FreeBSD with regards to 11 vs 12, having `extern(C)` declarations in Phobos is definitely going to make life harder. If there is a missing C declaration in druntime, just add it in druntime first, and if there's a PR for that, please wait on it rather than adding the declaration here to push this PR through. Doing that is just going to create more problems later.

jmdavis:
> So I suggest to merge this and put the newer version in a FreeBSD12 version block, s.t. once we switch to FreeBSD12 and abandon FreeBSD11 it's a simple removal.

It's not actually possible to do that. We have no version identifiers for specific versions of any OS, and I'd rather not go into the details of FreeBSD 11 vs 12 here, since it's complicated. Either way, we shouldn't be putting `extern(C)` declarations in Phobos (and doing so would make dealing with FreeBSD 12 harder). We should be using the ones in druntime. So, FreeBSD 11 vs 12 should be irrelevant here. What needs to work here with regards to FreeBSD is the version on the auto-tester, which is FreeBSD 11.x. As long as that works, FreeBSD 12 should not be a concern here. And when druntime is updated for FreeBSD 12, this code should then just work.

std/file.d
```diff
@@ -5119,3 +5120,87 @@ string tempDir() @trusted
...
+$(D statvfs)) function. On `FreeBSD` however, it uses the $(LINK2 https://www.freebsd.org/cgi/man.cgi?query=statfs&sektion=2,
+$(D statfs)) function.
+
+Params:
+    path = a valid path to any file in the file system
+Returns:
+    Available space in bytes
+
+Throws:
+$(LREF FileException) in case of failure
```
wilzbach: Indent after new sections

std/file.d
```diff
@@ -5119,3 +5120,87 @@ string tempDir() @trusted
...
+
+/**
+Returns the available space in a file system.
+
+On `Windows`, this function uses the result of the Windows API function
+$(LINK2 https://msdn.microsoft.com/en-us/library/windows/desktop/aa364937(v=vs.85).aspx, $(D GetDiskFreeSpaceExW)).
+
+On `POSIX` platforms, it uses the $(LINK2 http://man7.org/linux/man-pages/man3/statvfs.3.html,
+$(D statvfs)) function. On `FreeBSD` however, it uses the $(LINK2 https://www.freebsd.org/cgi/man.cgi?query=statfs&sektion=2,
+$(D statfs)) function.
```
wilzbach: I still think that the user doesn't care about this. He cares about the _implications_ of using these APIs.
If there's something different on Windows, note this, but otherwise the users expects that we have done our job properly.

std/file.d
```diff
@@ -5119,3 +5120,87 @@ string tempDir() @trusted
...
+        import core.sys.windows.winbase : GetDiskFreeSpaceExW, GetLastError;
+        import core.sys.windows.winnt : ULARGE_INTEGER;
+        import std.internal.cstring : tempCStringW;
+
+        ULARGE_INTEGER freeBytesAvailable;
+        auto err = () @trusted {
+            return GetDiskFreeSpaceExW(path.tempCStringW(), &freeBytesAvailable, null, null);
+        } ();
+        enforce(err != 0,
+            new FileException(text("Cannot get available disk space: ", GetLastError())));
```
wilzbach: Have a look at the default constructor of FileException:

```
    version(Windows) this(in char[] name,
                          uint errno = .GetLastError(),
                          string file = __FILE__,
                          size_t line = __LINE__) @safe
    {
```

tl;dr: use `cenforce`

std/file.d
```diff
@@ -5119,3 +5120,87 @@ string tempDir() @trusted
...
+        version (FreeBSD)
+        {
+            import core.sys.posix.sys.statvfs : statfs, statfs_t;
+
+            statfs_t stats = void;
+            auto err = () @trusted {
+                return statfs(path.toStringz(), &stats);
+            } ();
+            enforce(err == 0,
+                new FileException(text("Cannot get available disk space: ", err)));
```
wilzbach: `cenforce`

std/file.d
```diff
@@ -5119,3 +5120,87 @@ string tempDir() @trusted
...
+        else
+        {
+            import core.sys.posix.sys.statvfs : statvfs, statvfs_t;
+
+            statvfs_t stats = void;
+            auto err = () @trusted {
+                return statvfs(path.toStringz(), &stats);
+            } ();
+            enforce(err == 0,
+                new FileException(text("Cannot get available disk space: ", err)));
```
wilzbach: `cenforce`

std/file.d
```diff
@@ -5119,3 +5119,86 @@ string tempDir() @trusted
...
+    import std.exception : enforce;
+    import std.conv : text;
+
+    version (Windows)
+    {
+        import core.sys.windows.winbase : GetDiskFreeSpaceExW, GetLastError;
+        import core.sys.windows.winnt : ULARGE_INTEGER;
+        import std.internal.cstring : tempCStringW;
+
+        ULARGE_INTEGER freeBytesAvailable;
```
wilzbach: `= void`?

wilzbach:
As this is a new function, it might be worthwhile to check [this PR](https://github.com/dlang/phobos/pull/6389). 
The idea is to mark it as unstable (in some way) for a few releases.

std/file.d
jercaianu: https://www.boost.org/doc/libs/1_48_0/libs/filesystem/v3/doc/reference.html#space
Boost explains very clearly what happens in their filesystem library. I think it's a good idea to let the user know how we are getting the sizes.

std/file.d
```diff
@@ -5119,3 +5119,86 @@ string tempDir() @trusted
...
+///
+@safe unittest
+{
+    import std.exception : assertThrown;
+
+    auto space = getAvailableDiskSpace(".");
+    assert(space > 0);
+
+    assertThrown!FileException(getAvailableDiskSpace("ThisFileDoesNotExist123123"));
+}
```
wilzbach: I think we can do more tests here. Simple create a directory with `deleteme` and write files to it and check the resulting file size (if this varies on Windows, `version(Posix)` it).

Example:

https://dlang.org/phobos-prerelease/std_file.html#mkdir

std/file.d
n8sh: As long as this extra information comes "below the fold" I think it's fine to have any amount of extra technical information.

wilzbach: Nope it's not. It's our good to provide a "high-level" abstraction (as in the user shouldn't need to care about the low-level).
He definitely doesn't want to read 3 different documentations just to know "hey they return the disk size". If he really wants to, he can always look this up in the sources.

Also have a look at the existing std.file - notice how clean it is?
One of the few function where OS details are mentioned is [tempDir](https://dlang.org/phobos-prerelease/std_file.html#.tempDir) and it matters there.

n8sh: Taking a second look , the "however" makes it seem like there's an important distinction the user needs to be aware of. Is there one?
>On `FreeBSD` however, it uses the statfs function.

wilzbach: Here's a good example of an OS abstraction from Python's standard library:

https://docs.python.org/3/library/shutil.html#shutil.disk_usage

std/file.d
andralex: That'd be a bit excessive - we'd be testing the OS primitives.

andralex:
@jercaianu what's with the linux32 failure?

andralex:
@JackStouffer good to go?

JackStouffer:
I won’t be able to review this until late tonight. Please merge without my review if you think this is good to go.

tibi77:
Why is this stil open?

thewilsonator:
Because it fails ons FreeBSD.

thewilsonator:
@n8sh this probably needs a rebase to pass on circle.

std/file.d
JohanEngelen: You should test that the function works for `null` and `""` (empty strings) input. See e.g. https://issues.dlang.org/show_bug.cgi?id=14980

std/file.d
```diff
@@ -5232,3 +5233,74 @@ string tempDir() @trusted
...
+        auto err = () @trusted {
+            return GetDiskFreeSpaceExW(path.tempCStringW(), &freeBytesAvailable, null, null);
+        } ();
+        cenforce(err != 0, "Cannot get available disk space");
+
+        return freeBytesAvailable.QuadPart;
+    }
+    else version (Posix)
+    {
+        import std.string : toStringz;
```
n8sh: Since this function's argument type was changed from `string` to `const char[]` it should probably use `tempCString` instead of `toStringz` in the Posix path. `toStringz` always allocates if it is given a `const char[]`.

JohanEngelen:
https://github.com/dlang/phobos/pull/5776#discussion_r295124957  ?

PR #5726, opened 2017-09-07
Fix Issue 6409 - std.array.empty for associative arrays too

RazvanN7:
Should we implement other range primitives for AA's?

PR #5777, opened 2017-10-11
Fix issue 17894 - std.regex.internal.ir.Bytecode.length is defined as uint

ZombineDev:
See https://github.com/dlang/phobos/pull/5726#issuecomment-335768162 for more context.

### DIFF
--- a/std/regex/internal/generator.d
+++ b/std/regex/internal/generator.d
@@ -148,7 +148,7 @@ module std.regex.internal.generator;
                     if (app.data.length < limit && rand(3) > 0)
                         pc = pc - len;
                     else
-                        pc = pc + re.ir[pc].length;
+                        pc = pc + cast(uint) re.ir[pc].length;
                     break;
                 case IR.GroupStart, IR.GroupEnd:
                     pc += IRL!(IR.GroupStart);
--- a/std/regex/internal/ir.d
+++ b/std/regex/internal/ir.d
@@ -301,7 +301,7 @@ struct Bytecode
     }
 
     //full length of instruction
-    @property uint length() const
+    @property size_t length() const
     {
         return lengthOfIR(code);
     }
@@ -323,7 +323,7 @@ struct Bytecode
     uint indexOfPair(uint pc) const
     {
         assert(isStart || isEnd);
-        return isStart ? pc + data + length  : pc - data - lengthOfPairedIR(code);
+        return isStart ? pc + data + cast(uint) length  : pc - data - lengthOfPairedIR(code);
     }
 }
 
--- a/std/regex/internal/parser.d
+++ b/std/regex/internal/parser.d
@@ -88,7 +88,7 @@ if (isSomeString!S)
     {
         for (uint pc = start; pc < end; )
         {
-            immutable len = code[pc].length;
+            immutable len = cast(uint) code[pc].length;
             if (code[pc].code == IR.GotoEndOr)
                 break; //pick next alternation branch
             if (code[pc].isAtom)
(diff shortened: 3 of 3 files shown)


ZombineDev:
Nevermind @dlang-bot - we need this fix only master.

DmitryOlshansky:
@ZombineDev Please don't Bytecode isn't a range and never ment to be one. Length is uint is perfectly fine.

ZombineDev:
> Bytecode isn't a range and never meant to be one

I agree 100%. My intention is **not** to make `Bytecode` pass `isInputRange`, but instead to allow it fail *gracefully* without emitting warning that break the build.

DmitryOlshansky:
> I agree 100%. My intention is not to make Bytecode pass isInputRange, but instead to allow it fail gracefully without emitting any warnings that break the build.

This is patently wrong. If *unrelated field* called length breaks some build check *that build check is wrong*.

ZombineDev:
So which course of action do you propose? Removing the check for warnings or to removing the warning?

DmitryOlshansky:
Remove the check if there are false positives.
Do no mutilate code to fit some poorly designed check obviously.

ZombineDev:
Without this check input ranges with whose `.length` field is not defined as `size_t` will unconditionally fail the `isInputRange` check. So you propose breaking the code of all Phobos users? :D

DmitryOlshansky:
> Without this check input ranges with whose .length field is not defined as size_t will unconditionally fail the isInputRange check. So you propose breaking the code of all Phobos users, whose input ranges are not using size_t? :D

Something is seriously damaged if that is the case. A single unrelated struct needs to be reworked because some check for Ranges somewhere else won't pass otherwise for user-defined structs. Am I mad thinking this is sheer idiocy?

DmitryOlshansky:
> So you propose breaking the code of all Phobos users, whose input ranges are not using size_t? :D

So don't. Why std.regex.Bytecode is the victim of poorly devised checks in std.range?

ZombineDev:
> Am I mad thinking this is sheer idiocy?

Hah. Initially the commit message for the PR was named "Workaround stupid bug", however the more I investigated the less certain I was that there is a clear answer here. If you examine carefully all the code in that stack, you'll find that every single one is doing the right thing in isolation, it's just when we integrate everything it all falls apart.

ZombineDev:
> A single unrelated struct needs to be reworked because some check for Ranges somewhere else won't pass otherwise for user-defined structs.

If there's someone to put to the blame, it's the probably the template constraint in `std.array.insertInPlace(T, U...)(ref T[] array, size_t pos, U stuff)`, which `std.regex.internal.parser.optimize` internally uses.

DmitryOlshansky:
> But still, I don't see a clear way forward, save for the workaround proposed by this pull request.

You do understand that some innocent user-defined object may have defined length to be something other then size_t and it may have called insertInPlace

DmitryOlshansky:
> If there's someone to put to the blame, it's the probably the template constraint in std.array.insertInPlace(T, U...)(ref T[] array, size_t pos, U stuff), which std.regex.internal.parser.optimize internally calls. But still, I don't see a clear way forward, save for the workaround proposed by this pull request.

Obviously we need somehow to first check is smth is a range and then issue a warning or whatever if the length is not size_t. Bytecode is not a range hence no warning.

ZombineDev:
> Obviously we need somehow to first check is smth is a range and then issue a warning or whatever if the length is not size_t. Bytecode is not a range hence no warning.

The code goes the other way around. In order to check if `R` is an input range, it checks if `R` defines `empty`. If `empty` is not defined, the compiler implicitly uses via UFCS the version of `empty` defined in `std.range.primitives`, which calls `hasLength`.

ZombineDev:
> You do understand that some innocent user-defined object may have defined length to be something other then size_t and it may have called insertInPlace

This is not problem though. Their build system is unlikely to fail if the compiler emits a `pragma(msg, ...)`, while the phobos build system will.

DmitryOlshansky:
> while the phobos build system will.

Why do we need this?

ZombineDev:
To ensure that users of Phobos won't get unrelated warnings. That way it forces authors of PRs to fix warnings, before they reach the end-users.

ZombineDev:
Fine, let's see how this goes.

PR #5778, opened 2017-10-11
Fix Issue 10131 - To remove duplicates and keep order

RazvanN7:
This is the same as #5774 . I closed the PR and made a new one to some github branch mix-ups I made.

### DIFF
--- a/std/experimental/algorithm.d
+++ b/std/experimental/algorithm.d
@@ -1,0 +1,49 @@
+// Written in the D programming language.
+
+/**
+This module implements experimental additions/modifications to $(MREF std, _algorithm).
+
+Use this module to test out new functionality for $(REF wrap, std, _algorithm)
+
+Source:    $(PHOBOSSRC std/experimental/_algorithm.d)
+
+Copyright: Copyright the respective authors, 2008-
+License:   $(HTTP boost.org/LICENSE_1_0.txt, Boost License 1.0).
+Authors:   Andrea Fontana
+           Razvan Nitu
+ */
+module std.experimental.algorithm;
+
+import std.range : isInputRange, ElementType;
+
+/**
+Lazily produces a range with no duplicated elements from the input
+range $(D r). The order of elements is the original order.
+
+Params:
+    Range r - the input range to eliminate duplicates from
+
+Returns:
+    A range containing all the unique elements
+*/
+auto distinct(Range)(Range r)
+if (isInputRange!Range)
+{
+    import std.algorithm.iteration : filter;
+    bool[ElementType!Range] justSeen;
+    return r.filter!((k)
+        {
+            if (k in justSeen) return false;
+            return justSeen[k] = true;
+        });
+}
+
+///
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+
+    assert("AAAA".distinct.equal("A"));
+    assert("ABAC".distinct.equal("ABC"));
+    assert([1, 2, 3, 1, 2, 1, 1].distinct.equal([1, 2, 3]));
+}
--- a/posix.mak
+++ b/posix.mak
@@ -189,7 +189,7 @@ PACKAGE_std = array ascii base64 bigint bitmanip compiler complex concurrency \
   outbuffer parallelism path process random signals socket stdint \
   stdio string system traits typecons uni \
   uri utf uuid variant xml zip zlib
-PACKAGE_std_experimental = checkedint typecons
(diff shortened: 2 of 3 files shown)


std/experimental/algorithm.d
```diff
@@ -0,0 +1,49 @@
...
+}
+
+///
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+
+    assert("AAAA".distinct.equal("A"));
+    assert("ABAC".distinct.equal("ABC"));
+    assert([1, 2, 3, 1, 2, 1, 1].distinct.equal([1, 2, 3]));
```
trikko: I think an example with infinite ranges could be useful. Something like:
```d
	auto ascInt = recurrence!((x,n) => n)(0);    				// 0,1,2,3,4,5,6...
	auto descInt = recurrence!((x,n) => n)(0).map!(x => 10-x);  // 10,9,8,7,6,5,4 ..
	
	roundRobin(ascInt, descInt).distinct.take(20).writeln;

```

std/experimental/algorithm.d
```diff
@@ -0,0 +1,49 @@
+// Written in the D programming language.
+
+/**
+This module implements experimental additions/modifications to $(MREF std, _algorithm).
+
+Use this module to test out new functionality for $(REF wrap, std, _algorithm)
```
wilzbach: `wrap`?

std/experimental/algorithm.d
```diff
@@ -0,0 +1,49 @@
...
+module std.experimental.algorithm;
+
+import std.range : isInputRange, ElementType;
+
+/**
+Lazily produces a range with no duplicated elements from the input
+range $(D r). The order of elements is the original order.
+
+Params:
+    Range r - the input range to eliminate duplicates from
```
wilzbach: Use `r = input range from which duplicates should be eliminated`

std/experimental/algorithm.d
```diff
@@ -0,0 +1,49 @@
...
+
+Returns:
+    A range containing all the unique elements
```
wilzbach: Nit: misses a final dot ;-)

std/experimental/algorithm.d
```diff
@@ -0,0 +1,49 @@
...
+Authors:   Andrea Fontana
+           Razvan Nitu
+ */
+module std.experimental.algorithm;
+
+import std.range : isInputRange, ElementType;
+
+/**
+Lazily produces a range with no duplicated elements from the input
+range $(D r). The order of elements is the original order.
```
wilzbach: Should note that this uses the GC if we really want to go down this road.

std/experimental/algorithm.d
```diff
@@ -0,0 +1,49 @@
...
+
+///
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+
+    assert("AAAA".distinct.equal("A"));
+    assert("ABAC".distinct.equal("ABC"));
+    assert([1, 2, 3, 1, 2, 1, 1].distinct.equal([1, 2, 3]));
+}
```
wilzbach: This should have a lot more internal test. Please take a lot at other Phobos function for an impression - basic things, for example:
- `const`
- DummyRanges (e.g. RefRanges)
- infinite ranges
- strings (especially unicode ones)
- `nothrow`

std/experimental/algorithm.d
```diff
@@ -0,0 +1,49 @@
...
+    Range r - the input range to eliminate duplicates from
+
+Returns:
+    A range containing all the unique elements
+*/
+auto distinct(Range)(Range r)
+if (isInputRange!Range)
+{
+    import std.algorithm.iteration : filter;
+    bool[ElementType!Range] justSeen;
```
wilzbach: When I suggested moving this to `stdx.algorithm`, I was suggesting that we can use this to experiment with APIs that allow the user to control how the allocation happens.
There are a couple of ways to do this. Some quick ideas (there are more!):
- Do the deallocation automatically, e.g. like RefCounted and provide an optional Allocator
- Let the user do the deallocation (e.g. by returning the range and the allocated array as tuple)

Please note that it's very important that this function can be used in `@nogc`.

andralex:
I'll add that this (and other related functions) is an important function for things like relational joins etc. Martin Odersky showed me some pretty amazing short example that do powerful things because of such.

andralex:
After planning things with @RazvanN7 will close this in wait for scoped hashtables. cc @edi33416 :)

PR #5331, opened 2017-04-11
Make emplace infer class ctor safety

edi33416:

PR #5779, opened 2017-10-12
Split unittest6.obj to get past the 32767 max symbols limitation

edi33416:
On #5331 I've ran into the following issue:
On Windows 32bit, dmd fails to generate `unittest6.obj` because this object has more than 32767 symbols. This PR fixes this issue.

### DIFF
--- a/win32.mak
+++ b/win32.mak
@@ -561,6 +561,7 @@ UNITTEST_OBJS= \
 		unittest4.obj \
 		unittest5.obj \
 		unittest6.obj \
+		unittest6b.obj \
 		unittest7.obj \
 		unittest8a.obj \
 		unittest8b.obj \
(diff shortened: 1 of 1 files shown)


edi33416:
@ZombineDev @wilzbach Could you please have a look?

ZombineDev:
> used the opportunity to split it up fully ;-)

And we agreed that it's better to invest the time in migrating the Windows build-system to use posix.mak ;)

wilzbach:
> And we agreed that it's better to invest the time in migrating the Windows build-system to use posix.mak and leave this one as it is ;)

Yes, but sadly that's a long-term goal :/
Btw I was just referring to the line modified (i.e. `$(SRC_STD_CONTAINER) $(SRC_STD_EXP_ALLOC)` into `unittest6a` `unittest6c`)

ZombineDev:
I see. I though you meant split everything in the make file, which is why I thought that it's better to leave such a huge task for another time.

edi33416:
@wilzbach Sorry about that.. I misunderstood. As @ZombineDev, I should do the same change as I did to prove the this fixed the issue and leave the rest for the migration phase.

PR #5780, opened 2017-10-13
Fix Issue 7957 - std.functional untuple/untupleReversed too

RazvanN7:
I didn't implement the reversed part since I am not sure if this addition is going to make it to phobos and anyway there is the function reverse for tuples which can be used.

### DIFF
--- a/std/functional.d
+++ b/std/functional.d
@@ -15,6 +15,9 @@ $(TR $(TH Function Name) $(TH Description)
         $(TD Joins a couple of functions into one that executes the original
         functions independently and returns a tuple with all the results.
     ))
+    $(TR $(TD $(LREF apply))
+        $(TD Applies the items of a tuple as arguments to a function.
+    ))
     $(TR $(TD $(LREF compose), $(LREF pipe))
         $(TD Join a couple of functions into one that executes the original
         functions one after the other, using one function's result for the next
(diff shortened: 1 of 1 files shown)


std/functional.d
```diff
@@ -1562,3 +1565,45 @@ template forward(args...)
...
+Returns:
+    The return value of the function `callable` for the arguments
+    represented by the items of `t`.
+
+*/
+template apply(alias callable)
+{
+    import std.typecons : isTuple;
+    auto apply(Tup)(Tup t)
+    if (isTuple!Tup && Parameters!callable.length == Tup.length)
```
Biotronic: Instead of checking if the number of arguments matches (which might cause trouble with overload sets), why not check `__traits(compiles, callable(t.expand))`?

std/functional.d
```diff
@@ -1562,3 +1565,45 @@ template forward(args...)
...
+    {
+        return callable(t.tupleof);
```
Biotronic: The recommended property here is t.expand. I'm not sure that actually matters, though.

std/functional.d
```diff
@@ -1562,3 +1565,45 @@ template forward(args...)
...
+    auto a2 = [10, 20, 30, 40];
+    auto a3 = [100, 200, 300, 400];
+
+    int foo(dchar a, int b) { return a + b; }
+    assert(tup.apply!foo() == 107);
+    assert(zip(a1, a2).map!(apply!foo).equal([107, 118, 129, 140]));
+
+    int foo2(dchar a, int b, int c) { return a + b + c; }
+    assert(zip(a1, a2, a3).map!(apply!foo2).equal([207, 318, 429, 540]));
+}
```
Biotronic: Please also test overload sets - two or more functions with the same name and different parameters.

RazvanN7:
@Biotronic thanks for the review. I added a test for overload sets, although I if the test for overload sets wouldn't have worked that would have been a compiler issue, not a phobos one. Anyway, I hope this is better.

Biotronic:
I suggested the overload test to highlight a limitation of your previous implementation. By having the test there, we guard against implementations with the same limitations in the future, should someone feel the need to rewrite this.

MetaLang:
Is there a good reason why a user could not just do `foo(t.expand)`? Of the top of my head, I can't think of a situation where `t.apply(foo)` works but the former does not.

andralex:
@MetaLang the issue as described in https://issues.dlang.org/show_bug.cgi?id=7957 was that people can write `map!(apply!fun)` in a chain instead of `map!(t => fun(t.expand))`. So a convenience more than a primitive facility.

The issue is old, which gives us perspective. There seem to be little proliferation of the idiom using `.expand`, which makes for a weak case for adding this facility.

RazvanN7:
@MetaLang My initial position was to close the bug as WONTFIX due to the exact thought process of @andralex , but after a discussion with @ZombineDev on the bug-triage channel on slack we agreed that making a PR will bring the issue to the attention of the community. I am in favor of closing the PR and the bug if no strong case for this facility is presented.

Either way, the good thing is that we clean the bugzilla list.

andralex:
Cool @RazvanN7 thanks for the work and it's not for naught - the closure of the bug is good progress. Will close this and the bug. Thanks!

PR #5330, opened 2017-04-10
Make the allocators' unittests and allocate, owns and resolveInternal…

edi33416:
…Pointer methods safe

Allocating memory should be a safe operation and so should functions like owns and resolveInternalPointer.

PR #5783, opened 2017-10-16
Make resolveInternalPointer @safe

edi33416:
For stand-alone allocators, such as `GCAllocator`, `resolveInternalPointer` is a `@safe` function.
Allocators that are building on top of such allocators should infer the safety from their parents.

### DIFF
--- a/std/experimental/allocator/gc_allocator.d
+++ b/std/experimental/allocator/gc_allocator.d
@@ -70,7 +70,7 @@ struct GCAllocator
     }
 
     /// Ditto
-    pure nothrow
+    pure nothrow @trusted @nogc
     Ternary resolveInternalPointer(const void* p, ref void[] result) shared
     {
         auto r = GC.addrOf(cast(void*) p);
@@ -165,4 +165,18 @@ struct GCAllocator
 
     // anything above a page is simply rounded up to next page
     assert((() nothrow @safe @nogc => GCAllocator.instance.goodAllocSize(4096 * 4 + 1))() == 4096 * 5);
+}
+
+@safe unittest
+{
+    import std.typecons : Ternary;
+
+    void[] buffer = GCAllocator.instance.allocate(42);
+    void[] result;
+    Ternary found = GCAllocator.instance.resolveInternalPointer(&buffer[0], result);
+
+    assert(found == Ternary.yes && &result[0] == &buffer[0] && result.length >= buffer.length);
+    assert(GCAllocator.instance.resolveInternalPointer(null, result) == Ternary.no);
+    void *badPtr = (() @trusted => cast(void*)(0xdeadbeef))();
+    assert(GCAllocator.instance.resolveInternalPointer(badPtr, result) == Ternary.no);
 }
--- a/std/experimental/allocator/gc_allocator.d
+++ b/std/experimental/allocator/gc_allocator.d
@@ -151,8 +151,8 @@ struct GCAllocator
         scope(exit) GCAllocator.instance.deallocate(buffer);
 
         void[] p;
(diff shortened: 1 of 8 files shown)


std/experimental/allocator/package.d
```diff
@@ -2844,7 +2857,7 @@ private struct InternalPointersTree(Allocator)
     /// Ditto
     bool deallocate(void[] b)
     {
-        if (!b.ptr) return;
+        if (!b.ptr) return true;
```
andralex: how in the world did this pass until now?

ZombineDev: The only reasonable explanation is that it was never instantiated.

edi33416: This code doesn't get hit because of the `__EOF__` [here](https://github.com/edi33416/phobos/blob/d621a465202844548338c126b2ec362a18f0b48f/std/experimental/allocator/package.d#L2545).

edi33416:
@andralex I missed that. I think we should make this pure as well

std/experimental/allocator/gc_allocator.d
```diff
@@ -165,3 +165,17 @@ struct GCAllocator
...
+    import std.typecons : Ternary;
+
+    void[] buffer = GCAllocator.instance.allocate(42);
+    void[] p;
+    Ternary r = GCAllocator.instance.resolveInternalPointer(&buffer[0], p);
+
+    assert(r == Ternary.yes && &p[0] == &buffer[0] && p.length >= buffer.length);
+    assert(GCAllocator.instance.resolveInternalPointer(null, p) == Ternary.no);
+    void *badPtr = (() @trusted => cast(void*)(0xdeadbeef))();
+    assert(GCAllocator.instance.resolveInternalPointer(badPtr, p) == Ternary.no);
```
ZombineDev: Circle CI fails with:
> std/experimental/allocator/gc_allocator.d(180:56)[warn]: Argument 2 is named 'p', but this is the name of parameter 1

So the course of action is to use a slightly more informative name for the local variables. I would suggest renaming:
* `void[] p` -> `void[] result`
* `Ternary r` -> `Ternary found`

ZombineDev: Or alternatively, `void[] p` -> `void[] res` and `Ternary r` -> `Ternary ok` if you prefer to be more terse.

edi33416:
Should I remove [this](https://github.com/dlang/phobos/pull/5783/files#diff-0002a05f29875f98fa52cfba8996ac49R2545) `__EOF__` ?

Please note that by doing so, we will make [`struct ThreadLocal`](https://github.com/dlang/phobos/pull/5783/files#diff-0002a05f29875f98fa52cfba8996ac49R2559) public.

andralex:
Yes, ThreadLocal is good but let's make that a separate PR with a changelog entry associated. Thanks!

andralex:
@edi33416 rebase

andralex:
cc @Cybershadow, DAutoTest fails with:

```
Root package assert_writeln_magic references unknown package libdparse
```

Will force pull this.

CyberShadow:
> references unknown package

This usually means that code.dlang.org is down.

andralex:
@CyberShadow thx, can you translate to a more informative error message on the scripting side? Thx!

CyberShadow:
I feel really bad about doing that. @s-ludwig Could you please fix Dub to provide a better errod message instead?

wilzbach:
> I feel really bad about doing that. @s-ludwig Could you please fix Dub to provide a better error message instead?

The 2.077.0 release contains DUB 1.6.0 which has fallback mirrors and better error messages, so the easiest path would be to bump the D compiler version

PR #5784, opened 2017-10-16
Fix Issue 9362 - Add a method to remove one item to std.container.SList

Darredevil:
https://issues.dlang.org/show_bug.cgi?id=9362

If everyone agrees with this I will go ahead and do the same for Dlist.

### DIFF
--- a/std/container/slist.d
+++ b/std/container/slist.d
@@ -123,6 +123,18 @@ struct SList(T)
         return n;
     }
 
+    private static Node* findNodeByValue(Node* n, T value)
+    {
+        if (!n) return null;
+        auto ahead = n._next;
+        while (ahead && ahead._payload != value)
+        {
+            n = ahead;
+            ahead = n._next;
+        }
+        return n;
+    }
+
 /**
 Constructor taking a number of nodes
      */
@@ -589,6 +601,61 @@ Complexity: $(BIGOH n)
 
 /// ditto
     alias stableLinearRemove = linearRemove;
+
+/**
+Removes the first occurence of an element from the list in linear time.
+
+Returns: True if the element existed and was successfully removed, false otherwise.
+
+Params:
+    value = value of the node to be removed
+
+Complexity: $(BIGOH n)
+     */
+    bool linearRemoveElement(T value)
+    {
+        auto n1 = findNodeByValue(_root, value);
+
+        if (n1 && n1._next)
+        {
+            n1._next = n1._next._next;
+            return true;
+        }
+
+        return false;
+    }
+}
+
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+
+    auto e = SList!int();
+    auto b = e.linearRemoveElement(2);
+    assert(b == false);
+    assert(e.empty());
+    auto a = SList!int(-1, 1, 2, 1, 3, 4);
+    b = a.linearRemoveElement(1);
+    assert(equal(a[], [-1, 2, 1, 3, 4]));
+    assert(b == true);
+    b = a.linearRemoveElement(-1);
+    assert(b == true);
+    assert(equal(a[], [2, 1, 3, 4]));
+    b = a.linearRemoveElement(1);
+    assert(b == true);
+    assert(equal(a[], [2, 3, 4]));
+    b = a.linearRemoveElement(2);
+    assert(b == true);
+    b = a.linearRemoveElement(20);
+    assert(b == false);
+    assert(equal(a[], [3, 4]));
+    b = a.linearRemoveElement(4);
+    assert(b == true);
+    assert(equal(a[], [3]));
+    b = a.linearRemoveElement(3);
+    assert(b == true);
+    assert(a.empty());
+    a.linearRemoveElement(3);
 }
 
 @safe unittest


std/container/slist.d
```diff
@@ -123,6 +123,19 @@ struct SList(T)
         return n;
     }
 
+    private static Node * findNodeByValue(Node * n, T value)
```
wilzbach: Nit: style: `Node*`

std/container/slist.d
```diff
@@ -123,6 +123,19 @@ struct SList(T)
...
+    {
+        assert(n);
```
wilzbach: Needs an error message, e.g. "The node can't be null"

(I know that there are other methods in this module don't do this either, but that's no reason to continue using the old style)

std/container/slist.d
```diff
@@ -589,6 +602,39 @@ Complexity: $(BIGOH n)
 
 /// ditto
     alias stableLinearRemove = linearRemove;
+
+/**
+Removes an element from the list in linear time.
+
```
wilzbach: Parameter block is missing.

std/container/slist.d
```diff
@@ -589,6 +602,39 @@ Complexity: $(BIGOH n)
...
+Complexity: $(BIGOH n)
+     */
+    void linearRemoveElement(T value)
+    {
+        enforce(_first);
```
wilzbach: Why is removing `0` invalid?

std/container/slist.d
```diff
@@ -589,6 +602,39 @@ Complexity: $(BIGOH n)
...
+    a.linearRemoveElement(-1);
+    assert(equal(a[], [2, 3, 4]));
+    a.linearRemoveElement(2);
+    a.linearRemoveElement(20);
+    assert(equal(a[], [3, 4]));
+    a.linearRemoveElement(4);
+    assert(equal(a[], [3]));
+    a.linearRemoveElement(3);
+    assert(a.empty());
 }
```
wilzbach: Would need a test for the behavior when one value occurs multiple times.

std/container/slist.d
```diff
@@ -123,6 +123,19 @@ struct SList(T)
...
     }
 
+    private static Node * findNodeByValue(Node * n, T value)
+    {
+        assert(n);
+        auto ahead = n._next;
+        while (ahead && ahead._payload != value)
+        {
+            n = ahead;
+            enforce(n);
```
wilzbach: Why do you test `n` / `ahead` here? The loop body is only executed if `ahead` is evaluated to `true` (i.e. not `null`).

std/container/slist.d
```diff
@@ -589,6 +602,39 @@ Complexity: $(BIGOH n)
 
 /// ditto
     alias stableLinearRemove = linearRemove;
+
+/**
+Removes an element from the list in linear time.
```
wilzbach: Define behavior when the value occurs multiple times (e.g. the list is iterated from front to back and the first match is removed).

std/container/slist.d
Darredevil: Was trying to maintain the general style of the Slist implementation. I will update it.

std/container/slist.d
Darredevil: Added test for this.

std/container/slist.d
ZombineDev: As mentioned in https://wiki.dlang.org/Contributing_to_Phobos:

> Be critical when looking at Phobos code (some parts of Phobos are a bit older and have some relics...)

edi33416:
LGTM

std/container/slist.d
ZombineDev: For example:

```
Removes the first an element from the list that equals `value` in $(BIGOH n) time.
```

std/container/slist.d
ZombineDev: Also I think it would be better to return a boolean value - `true` iff the element existed and was successfully removed (`false` otherwise).

std/container/slist.d
```diff
@@ -589,6 +601,43 @@ Complexity: $(BIGOH n)
...
+
+/**
+Removes the first occurence of an element from the list in linear time.
+
+Params:
+    value = value of the node to be removed
+
+Complexity: $(BIGOH n)
+     */
+    void linearRemoveElement(T value)
```
ZombineDev: I think it would be better to return a boolean value - `true` iff the element existed and was successfully removed (`false` otherwise).

ZombineDev: In which case, `findNodeByValue` should probably return `null` (without asserting) if `_root` is also `null`.

Darredevil:
@ZombineDev @wilzbach this PR should be ready for merge. Jenkins seems to be unhappy with something unrelated with this PR.

wilzbach:
Being a public symbol this still needs an approval from @andralex

std/container/slist.d
```diff
@@ -123,6 +123,18 @@ struct SList(T)
         return n;
     }
 
+    private static Node* findNodeByValue(Node* n, T value)
+    {
+        assert(n, "Error: un-initialized List");
```
andralex: should return null for empty list

std/container/slist.d
```diff
@@ -589,6 +601,58 @@ Complexity: $(BIGOH n)
...
+
+        return false;
+    }
+}
+
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+
+    auto a = SList!int(-1, 1, 2, 1, 3, 4);
```
andralex: should test against an empty list

Darredevil:
@andralex @ZombineDev should be ready for merge

std/container/slist.d
```diff
@@ -589,6 +601,62 @@ Complexity: $(BIGOH n)
...
+Params:
+    value = value of the node to be removed
+
+Complexity: $(BIGOH n)
+     */
+    bool linearRemoveElement(T value)
+    {
+        auto n1 = findNodeByValue(_root, value);
+
+        // if node exists
```
andralex: delete

andralex:
@wilzbach this fails the style checker, which segfaults. I'll pull this, could you please investigate the matter?

PR #5785, opened 2017-10-17
Issue 17905: Add source property to byCodeUnit

jmdavis:
This makes it possible to extract the underlying string from the result of `byCodeUnit` so that it's possible to temporarily wrap a string to pass it to a function that would otherwise auto-decode it but still be able to have a string afterwards (assuming that the function that's called doesn't wrap the `byCodeUnit` range in its own range of course).

Without this, you're frequently stuck having to allocate a new string from the `byCodeUnit` range just so that you can have a `string` again, which is just plain wasteful.


andralex:
OK, I see it's undocummented for now, probably a good idea to experiment with.

PR #5786, opened 2017-10-17
Add liniarRemoveElement for Dlist

Darredevil:
https://issues.dlang.org/show_bug.cgi?id=9362

Followed the idea from https://github.com/dlang/phobos/pull/5784 and implemented the same for Dlist.

### DIFF
--- a/std/container/dlist.d
+++ b/std/container/dlist.d
@@ -702,7 +702,44 @@ Complexity: $(BIGOH r.walkLength)
     /// ditto
     alias stableLinearRemove = linearRemove;
 
+/**
+Removes the first occurence of an element from the list in linear time.
+
+Returns: True if the element existed and was successfully removed, false otherwise.
+
+Params:
+    value = value of the node to be removed
+
+Complexity: $(BIGOH n)
+     */
+    bool linearRemoveElement(T value)
+    {
+        auto n1 = findNodeByValue(_root, value);
+        if (n1)
+        {
+            auto n2 = n1._next._next;
+            BaseNode.connect(n1, n2);
+            return true;
+        }
+
+        return false;
+    }
+
+
 private:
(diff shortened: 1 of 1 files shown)


Darredevil:
@andralex awaiting approval

std/container/dlist.d
```diff
@@ -702,8 +702,46 @@ Complexity: $(BIGOH r.walkLength)
...
+
+        return false;
+    }
+
+
 private:
 
+    BaseNode* findNodeByValue(BaseNode* n, T value)
+    {
+        assert(n, "Error: un-initialized List");
```
andralex: shouldn't you just return null if the list is not initialized?

std/container/dlist.d
```diff
@@ -766,6 +804,34 @@ private:
 {
     import std.algorithm.comparison : equal;
 
+    auto a = DList!int(-1, 1, 2, 1, 3, 4);
```
andralex: should include a test involving an empty list

std/container/dlist.d
ZombineDev: I would also prefer to return null instead of asserting, though all the other `findNode*` functions do the same.

Darredevil:
@andralex @ZombineDev should be ready for merge

std/container/dlist.d
```diff
@@ -702,8 +702,46 @@ Complexity: $(BIGOH r.walkLength)
...
+
+Params:
+    value = value of the node to be removed
+
+Complexity: $(BIGOH n)
+     */
+    bool linearRemoveElement(T value)
+    {
+        auto n1 = findNodeByValue(_root, value);
+        // if node exists
```
andralex: delete, just repeats what the code says

PR #5787, opened 2017-10-17
Make goodAllocSize @safe

edi33416:
For stand-alone allocators, such as `GCAllocator`, `goodAllocSize` is a `pure nothrow @safe @nogc` function.

### DIFF
--- a/std/experimental/allocator/building_blocks/affix_allocator.d
+++ b/std/experimental/allocator/building_blocks/affix_allocator.d
@@ -439,3 +439,12 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
     Ternary r = MyAllocator.instance.resolveInternalPointer(d.ptr, p);
     assert(p.ptr is d.ptr && p.length >= d.length);
 }
+
+// Check that goodAllocSize inherits from parent, i.e. GCAllocator
+@system unittest
+{
+    import std.experimental.allocator.gc_allocator;
+    alias a = AffixAllocator!(GCAllocator, uint).instance;
+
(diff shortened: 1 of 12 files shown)


edi33416:
I believe that `goodAllocSize` should also be `static`.
What do you think?

std/experimental/allocator/gc_allocator.d
```diff
@@ -140,11 +141,15 @@ struct GCAllocator
     import std.typecons : Ternary;
 
     // test allocation sizes
-    assert(GCAllocator.instance.goodAllocSize(1) == 16);
+    () nothrow @safe @nogc {
+        assert(GCAllocator.instance.goodAllocSize(1) == 16);
+    }();
```
andralex: same here and everywhere, i.e. we want assert to be top-level to make sure all code goes away in release mode

ZombineDev: Why? This is inside a unittest after all. Unittests should be able to verify that code built with optimizations is also correct.

std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -439,3 +439,12 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
     Ternary r = MyAllocator.instance.resolveInternalPointer(d.ptr, p);
     assert(p.ptr is d.ptr && p.length >= d.length);
 }
+
+// Check that goodAllocSize inherits from parent, i.e. GCAllocator
+@system unittest
+{
+    import std.experimental.allocator.gc_allocator;
+    alias a = AffixAllocator!(GCAllocator, uint).instance;
```
andralex: no need for empty lines in short functions, it's excessive

PR #5230, opened 2017-03-03
RFC: Improve std.random.unpredictableSeed

yshui:
Implemented according to suggestions from Cédric Picard (see [1](https://cym13.github.io/article/unpredictableSeed.html) [2](https://forum.dlang.org/post/atpkaykjsfiolghtivfq@forum.dlang.org))

I only have linux machine to test with, so I opened this pull request to abuse the build bot.

Shoot if you spot any problems.

PR #5788, opened 2017-10-17
Issue 17912: Add function for creating a temporary file.

jmdavis:
This adds std.file.createTempFile, which creates a temporary file with a
random name (optionally with a specified prefix and/or suffix). By
default, the file is empty, but if data is passed to createTempFile,
then the file is populated with that data just like it would have been
with std.file.write. The name of the file that was created is returned.
So, essentially, createTempFile is like write except that it generates
the file name for you and returns it.

Previously, we added scratchFile to std.stdio, which was like
createTempFile except that it gave you an open File object, but it was
scrapped, because the dependencies that it added to std.stdio made
"hello world" bigger. See:
https://issues.dlang.org/show_bug.cgi?id=14599

Just like scratchFile did, createTempFile ensures that a filename is
selected which did not exist previously so that the file can't be
hijacked with different permissions by someone creating a file with the
same name beforehand (the randomness of the name should make that
effectively impossible, but the way that createTempFile opens the file
guarantees it).

These changes also consolidate some of write's implementation across
OSes, since the writing portion of writeImpl was esentially identical
across OSes but needlessly duplicated. That consolidated functionality
is then also used by createTempFile.

Unfortunately, I don't currently have a Windows setup that I can test this with, so depending on what happens with the autotester, I may need to make a revision or two to get it to pass on Windows. I might get lucky though.

### DIFF
--- a/changelog/std-file-createTempFile.dd
+++ b/changelog/std-file-createTempFile.dd
@@ -1,0 +1,15 @@
+`createTempFile` added to std.file
+
+$(REF std, file, createTempFile) creates a file in $(REF tempDir, std, file)
+with a random name (with an optionally provided prefix and suffix) and returns
+the file name. It also optionally takes data to write to the file when creating
+it (if no data is provided, then the newly created file will be empty).
+
+This is different from $(REF File.tmpfile, std, stdio), which creates a
+temporary file with no name, returns a $(REF File, std, stdio) to that file,
+and then deletes the file when the $(REF File, std, stdio) is closed. The file
+created by $(REF std, file, createTempFile) is a normal file which can be
+opened and closed as many times as desired and will only be deleted by the
+program if the program explicitly does so (though it is sitting in a temp
+directory and is subject to the normal things tha happen to such files - e.g.
+some OSes delete those files on boot).
--- a/std/file.d
+++ b/std/file.d
@@ -35,6 +35,7 @@ $(TR $(TD Files) $(TD
           $(LREF remove)
           $(LREF slurp)
           $(LREF write)
+          $(LREF createTempFile)
 ))
 $(TR $(TD Symlinks) $(TD
           $(LREF symlink)
@@ -809,8 +810,6 @@ if (isConvertibleToString!R)
     static assert(__traits(compiles, append(TestAliasedString("foo"), [0, 1, 2, 3])));
 }
 
-// Posix implementation helper for write and append
-
 version(Posix) private void writeImpl(const(char)[] name, const(FSChar)* namez,
         in void[] buffer, bool append) @trusted
 {
@@ -821,26 +820,8 @@ version(Posix) private void writeImpl(const(char)[] name, const(FSChar)* namez,
                        : O_CREAT | O_WRONLY | O_TRUNC;
 
     immutable fd = core.sys.posix.fcntl.open(namez, mode, octal!666);
-    cenforce(fd != -1, name, namez);
-    {
-        scope(failure) core.sys.posix.unistd.close(fd);
-
-        immutable size = buffer.length;
-        size_t sum, cnt = void;
-        while (sum != size)
-        {
-            cnt = (size - sum < 2^^30) ? (size - sum) : 2^^30;
-            const numwritten = core.sys.posix.unistd.write(fd, buffer.ptr + sum, cnt);
-            if (numwritten != cnt)
-                break;
-            sum += numwritten;
-        }
-        cenforce(sum == size, name, namez);
-    }
-    cenforce(core.sys.posix.unistd.close(fd) == 0, name, namez);
+    writeToOpenFile(fd, buffer, name, namez);
 }
-
-// Windows implementation helper for write and append
 
 version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez,
         in void[] buffer, bool append) @trusted
@@ -868,19 +849,201 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
         h = CreateFileW(namez, defaults);
         cenforce(h != INVALID_HANDLE_VALUE, name, namez);
     }
+    writeToOpenFile(h, buffer, name, namez);
+}
+
+private void writeToOpenFile(FD)(FD fd, const void[] buffer, const(char)[] name,
+                                 const(FSChar)* namez) @trusted
+{
     immutable size = buffer.length;
     size_t sum, cnt = void;
-    DWORD numwritten = void;
+
     while (sum != size)
     {
-        cnt = (size - sum < 2^^30) ? (size - sum) : 2^^30;
-        WriteFile(h, buffer.ptr + sum, cast(uint) cnt, &numwritten, null);
-        if (numwritten != cnt)
+        cnt = size - sum < 2^^30 ? size - sum : 2^^30;
+
+        version(Posix)
+            immutable numWritten = core.sys.posix.unistd.write(fd, buffer.ptr + sum, cnt);
+        else version(Windows)
+        {
+            DWORD numWritten = void;
+            WriteFile(fd, buffer.ptr + sum, cast(uint) cnt, &numWritten, null);
+        }
+        else
+            static assert(0, "Unsupported OS");
+
+        if (numWritten != cnt)
             break;
-        sum += numwritten;
+        sum += numWritten;
     }
-    cenforce(sum == size && CloseHandle(h), name, namez);
+
+    version(Posix)
+        cenforce(sum == size && core.sys.posix.unistd.close(fd) == 0, name, namez);
+    else version(Windows)
+        cenforce(sum == size && CloseHandle(fd), name, namez);
+    else
+        static assert(0, "Unsupported OS");
 }
+
+
+/++
+    Creates a file with a random name in $(LREF tempDir) and returns the name
+    of the file. If data is passed in, then the file is written with that data;
+    otherwise, it's empty.
+
+    The idea is that this creates a temporary file for testing or whatever
+    other purpose a temporary file might be needed for. However, it will only
+    be deleted if it's explicitly deleted or if the OS decides to delete it as
+    some OSes do with files in temp directories on startup or shutdown. So,
+    unlike $(REF File.tmpfile, std, stdio), the file has a name, and it can have
+    all of the things done to it that one might typically do with a file
+    (including reopening it after closing it).
+
+    The file is created with R/W permissions. On POSIX systems, the permissions
+    are restricted to the current user, though the effective permissions are
+    modified by the process' umask in the usual way.
+
+    $(REF rndGen, std, random) is used as the random number generator.
+
+    Params:
+        prefix = Prefix to the random portion of the file name.
+        suffix = Suffix to the random portion of the file name (e.g. for a file
+                 extension).
+        buffer = Data to populate the file with.
+
+    Returns:
+        The name of the generated file.
+
+    Throws:
+        $(LREF FileException) if it fails to create the file.
+
+    See_Also:
+        $(REF File.tmpfile, std, stdio)
+  +/
+string createTempFile(const void[] buffer = null) @safe
+{
+    return createTempFile(null, null, buffer);
+}
+
+/// Ditto
+string createTempFile(const(char)[] prefix, const(char)[] suffix, const void[] buffer = null) @trusted
+{
+    import std.path : absolutePath, baseName, buildPath, dirName;
+
+    static string genTempName(const(char)[] prefix, const(char)[] suffix)
+    {
+        import std.ascii : digits, letters;
+        import std.random : choice, rndGen;
+        import std.range : chain;
+        import std.string : representation;
+
+        auto name = new char[](prefix.length + 15 + suffix.length);
+        name[0 .. prefix.length] = prefix;
+        name[$ - suffix.length .. $] = suffix;
+
+        auto random = &rndGen();
+        rndGen.popFront();
+
+        auto chars = chain(letters.representation, digits.representation);
+        foreach (ref c; name[prefix.length .. $ - suffix.length])
+        {
+            c = choice(chars);
+            random.popFront();
+        }
+
+        return buildPath(tempDir, name);
+    }
+
+    while (1)
+    {
+        auto filename = genTempName(prefix, suffix);
+
+        version(Posix)
+        {
+            auto fd = open(tempCString!FSChar(filename), O_RDWR | O_CREAT | O_EXCL, S_IRUSR | S_IWUSR);
+            if (fd == -1)
+            {
+                immutable errno = .errno;
+                if (errno == EEXIST)
+                    continue;
+                else
+                    throw new FileException("Failed to create a temporary file", errno);
+            }
+        }
+        else version(Windows)
+        {
+            auto fd = CreateFileW(tempCString!FSChar(filename), GENERIC_WRITE, 0, null, CREATE_NEW,
+                                  FILE_ATTRIBUTE_NORMAL | FILE_FLAG_SEQUENTIAL_SCAN, HANDLE.init);
+            if (fd == INVALID_HANDLE_VALUE)
+            {
+                immutable errno = .GetLastError();
+                if (errno == ERROR_FILE_EXISTS)
+                    continue;
+                else
+                    throw new FileException("Failed to create a temporary file", errno);
+            }
+        }
+        else
+            static assert(0, "Unsupported OS");
+
+        writeToOpenFile(fd, buffer, filename, null);
+
+        return filename;
+    }
+}
+
+///
+@safe unittest
+{
+    import std.algorithm.searching : startsWith;
+    import std.range.primitives;
+    import std.path : baseName, buildNormalizedPath, dirName, extension;
+
+    {
+        auto path = createTempFile();
+        scope(exit) if (path.exists) remove(path);
+        assert(!path.empty);
+        assert(buildNormalizedPath(path.dirName) == buildNormalizedPath(tempDir));
+        assert(!path.baseName.empty);
+        assert(path.exists);
+        assert(path.isFile);
+        assert(readText(path).empty);
+    }
+    {
+        auto path = createTempFile("hello world");
+        scope(exit) if (path.exists) remove(path);
+        assert(!path.empty);
+        assert(buildNormalizedPath(path.dirName) == buildNormalizedPath(tempDir));
+        assert(!path.baseName.empty);
+        assert(path.exists);
+        assert(path.isFile);
+        assert(readText(path) == "hello world");
+    }
+    {
+        auto path = createTempFile("prefix_", ".txt");
+        scope(exit) if (path.exists) remove(path);
+        auto name = path.baseName;
+        assert(name.startsWith("prefix_"));
+        assert(name.extension == ".txt");
+        assert(name.length > "prefix_.txt".length);
+        assert(buildNormalizedPath(path.dirName) == buildNormalizedPath(tempDir));
+        assert(path.exists);
+        assert(path.isFile);
+        assert(readText(path).empty);
+    }
+    {
+        auto path = createTempFile(null, ".txt", "D rocks");
+        scope(exit) if (path.exists) remove(path);
+        auto name = path.baseName;
+        assert(name.extension == ".txt");
+        assert(name.length > ".txt".length);
+        assert(buildNormalizedPath(path.dirName) == buildNormalizedPath(tempDir));
+        assert(path.exists);
+        assert(path.isFile);
+        assert(readText(path) == "D rocks");
+    }
+}
+
 
 /***************************************************
  * Rename file $(D from) _to $(D to).


jmdavis:
Great. By some miracle, Windows went perfectly on the first go, but then Mac OS X failed for some reason. :|

jmdavis:
It passed Mac OS X this time though, so I have no clue what happened. In theory, it should act the same on Mac OS X as the other POSIX platforms.

std/file.d
```diff
@@ -670,20 +650,191 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+        }
+        else
+            static assert(0, "Unsupported OS");
+
+        writeToOpenFile(fd, buffer, filename, null);
+
+        return filename;
+    }
+
+    throw new FileException("Failed to create a temporary file");
```
BBasile: The `Throws` ddoc section should be added for this exception.

CyberShadow:
Why not `mkstemp` on POSIX?

Why not `GetTempFileName ` on Windows?

jmdavis:
> Why not mkstemp on POSIX?

Because the last time I looked into it, there were some POSIX systems that could only generate something like 26 file names with mkstemp. I don't think that the ones dmd supports at the moment have that problem, but others did, which then potentially becomes a problem for LDC or GDC, and I see no advantage to using mkstemp over just doing it ourselves when some systems do a terrible job with mkstemp. The code used here was in Phobos previously with `scratchFile` in std.stdio. It just got ripped out because it made "hello world" bigger.

> Why not GetTempFileName on Windows?

It doesn't support suffixes (which IMHO is more important than a prefix, since it can be used for file extensions, which sometimes matter a great deal; prefixes matter a lot less frequently). It also doesn't let you write to the file when it's created. You have to reopen it and then write to it. Also, the docs themselves seem to think that you shouldn't use it, saying

"To avoid problems resulting when converting an ANSI string, an application should call the CreateFile function to create a temporary file."

So, GetTempFile can't be used unless we don't have a suffix or if we call rename on the function afterwards (which is likely a bad idea), and Microsoft's docs are basically saying to do what I did.

So, I'd just as soon avoid mkstemp regardless of what Windows is up to, but it looks like we need to implement this ourselves on Windows anyway, and this way, what it's doing should be consistent across systems.

std/file.d
```diff
@@ -670,20 +650,210 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+            c = chars[rndGen.front % chars.length];
+            rndGen.popFront();
+        }
+
+        return buildPath(tempDir, name);
+    }
+
+    // Once should be enough given the randomness, but we'll give it three
+    // times to be safe.
+    foreach (i; 0 .. 3)
```
CyberShadow: Doing things "just to be safe" smells - this should be an infinite loop.

std/file.d
```diff
@@ -670,20 +650,210 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+                writeln(filename);
+                writeln("-------------------------");
+            }
+        }
+        else version(Windows)
+        {
+            auto fd = CreateFileW(tempCString!FSChar(filename), GENERIC_WRITE, 0, null, CREATE_NEW,
+                                  FILE_ATTRIBUTE_NORMAL | FILE_FLAG_SEQUENTIAL_SCAN, HANDLE.init);
+            if (fd == INVALID_HANDLE_VALUE)
+                continue;
```
CyberShadow: The loop should be unconditionally retried on EEXIST or equivalent. Other errors, like ENOPERM or ENOENT (temporary directory inaccessible on nonexisting) should throw immediately.

CyberShadow:
I don't think we should be using the global `rndGen` for this. Some part of the program may initialize it to some seed, which could result in a denial of service as this function retries the same file names over and over. Another source of randomness would be better.

CyberShadow:
> Because the last time I looked into it, there were some POSIX systems that could only generate something like 26 file names with mkstemp.

What are they?

> and I see no advantage to using mkstemp over just doing it ourselves when some systems do a terrible job with mkstemp.

There's a very good chance that we get things wrong too - definitely some things that might be worth worrying about in the initial implementation here. Generally I don't think D is expected to fix or work around bugs in ancient/unsupported libcs.

> So, GetTempFile can't be used unless we don't have a suffix or if we call rename on the function afterwards (which is likely a bad idea), and Microsoft's docs are basically saying to do what I did.

Yeah, on closer look, GetTempFileName looks not that great. You have to specify the random value anyway, and it's limited to 65535 unique file names (with the same prefix).

std/file.d
jmdavis: I'm not a big fan of infinite loops when you can't absolutely guarantee that it will exit, which we can't here, because it depends on the file system - but in theory, it should be essentially impossible for the loop to be infinite if we're picky about the error code like you suggested.

std/file.d
jmdavis: Fixed.

jmdavis:
Okay, per the review comments, it now is pickier about which errors it retries for, and it's not using `rndGen`, so it doesn't run the risk of being screwed up by `rndGen` being badly reseeded by the program or the random numbers being reused within the program due to the issues caused by the random number generators being value types instead of reference types.

> >    Because the last time I looked into it, there were some POSIX systems that could only generate something like 26 file names with mkstemp.

> What are they?

I _think_ that it was Solaris that was the problem, but it's been quite a while since I implemented `scratchFile`. Digging around now, SmartOS' man page (which is a "distro" of Illumos and thus the version of Solaris that matters), and all it says on the matter is that ["It is possible to run out of letters."](https://smartos.org/man/3C/mkstemp) It does make it sound like it only supports 6 X's though, whereas FreeBSD's seems to support an arbitrary number. It might be that Illumos' implementation would be fine, but that man page does exactly give me warm fuzzies about it (whereas FreeBSD is quite explicit about how many possible values it has).

But given that we need to implement this ourselves on Windows anyway, I'd just as soon use our implementation on all systems, and then if there's a problem, it will be caught more easily. The key thing to getting it right is how the file is opened (it needs to make sure that the file is created when it's opened and can't do that by checking for existence first, or there's a race condition); the rest is mostly just a question of having a suitably random name, and unless we have serious problems with our random number generator, that shouldn't be a problem (and at the moment, this picks 15 random alphanumeric characters, so the number of possible values is huge).

quickfur:
This is very nice, in that it frees us from having to rely on potentially poorly-implemented `mkstemp` functions in the OS / standard C library.

However, it warrants *very* close scrutiny, because when it comes to security, one can never be too paranoid.  As far as POSIX is concerned, I *think* calling `open()` with O_EXCL | O_CREAT ought to rule out race conditions with file creation; and making the file read/write only upon creation by the user should rule out an outside attacker overwriting the file with potential attack vectors.  It doesn't exclude the possibility of someone who has access to the current user account from manipulating the file in some way, but if they can do that, we're already screwed anyway, so presumably that should be OK as well.  So it seems OK to me.

But I think reviewing it against various `mkstemp` implementations would be wise before we merge this.  The last thing we want is for users to start using this and then discovering a security loophole, and D will get very bad rep for having insecure functions in its standard library.

quickfur:
Oh, one potential thing: this function ought to be thread-safe. I believe it is (the static RNG seed should be in TLS, right?), but someone should review it carefully just in case.

jmdavis:
> But I think reviewing it against various mkstemp implementations would be wise before we merge this. The last thing we want is for users to start using this and then discovering a security loophole, and D will get very bad rep for having insecure functions in its standard library.

If you can think of a potential security hole, then great, but all this does is generate a random file name and then ensure that when it creates the file, that file doesn't already exist. Ensuring that it doesn't overwrite any files solves the key security problem. I don't see how it would be possible to do anything else to protect against security problems, because there's so little involved here. Certainly, if there's a potential security problem with creating a file, closing it, and reopening it, then there could be a problem, but there's nothing that we can do about that. If that's your problem, then go use `std.stdio.tmpfile`, but if you need to create a file and then reopen it, that doesn't cut it, and in my experience, that's always what I've needed from a temp file; as a result, I've found `tmpfile` to be worse than useless.

> Oh, one potential thing: this function ought to be thread-safe. I believe it is (the static RNG seed should be in TLS, right?), but someone should review it carefully just in case.

Of course, it's thread-safe. Nothing in it is `shared`, and std.random doesn't use `shared` anywhere. `__gshared` isn't used either. The only risk of thread-safety issues would be the C functions involved, and if `open` isn't thread-safe, well, we're screwed anyway. D's type system makes it wonderfully easy to see that everything involved is in TLS.

jmdavis:
@quickfur If it makes you feel any better, I just looked over FreeBSD's implementation of `mkstemp`, and it's doing basically the same thing that we're doing here. The main difference is that instead of randomly generating file names after the first one already exists, it increments each of the characters in turn. But the file is opened the same way.

std/file.d
```diff
@@ -670,20 +650,224 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+        prefix = Prefix to the random portion of the file name.
+        suffix = Suffix to the random portion of the file name (e.g. for a file
+                 extension).
+        buffer = Data to populate the file with.
+
+    Returns:
+        The name of the generated file.
+
+    Throws:
+        $(LREF FileException) if it fails to create the file reason.
```
BBasile: "for some" ?

BBasile: One last thing about the doc is that you must add `createFile` to the table in the top level module comment.

nemanja-boric-sociomantic:
Instead of random number generator, could you use the approach used by other implementations: process' id and the current time (incrementing every loop)? That should be "random" enough for this purpose.

std/file.d
jmdavis: done.

jmdavis:
Okay. The random failure on Mac OS X was because sometimes `tempDir` returns /tmp, which `readLink` claims is a symlink to `private/tmp`, but which doesn't exist (I think that it's supposed to be `/private/tmp`). I'm not sure what the deal with that is, but upon rereading the Linux and Mac OS X man pages for `open`, I think that I misunderstood the bit about `O_EXCL` not working with symlinks. I thought that it meant that directory it was in would fail if it was a symlink, but I think that it means that it fails if the target is a symlink (which makes sense, since it would be stupid to overwrite it), and Mac OS X does seem to pass even when `tempDir` is `/tmp` if I don't try to call `readLink` on it. So, I've removed the call to `readLink`, since it doesn't seem to be necessary. I don't know if the bit about `private/tmp` is a bug in `readLink` or what, but if `readLink` isn't called here, then it doesn't matter for this PR.

> Instead of random number generator, could you use the approach used by other implementations: process' id and the current time (incrementing every loop)? That should be "random" enough for this purpose.

And what would be the benefit of that over using a random number generator? If anything, I would think that a random number generator would be better, because the value would be less predictable, so if there are any security concerns about a malicious program with the correct permissions guessing what the file is and mucking with it, then the risk would be lower (albeit not zero). I don't think that that's a huge concern (and using `O_EXCL` with `open` solves the problem of programs that wouldn't have permission being able to hijack the file), but I don't see any benefit in avoiding the random number generator either. FWIW, FreeBSD uses a random number generator with `mkstemp` just like we're doing here. It's just that it's using its own random number generator, and it increments the result if there's a conflict rather than generating another random name, whereas here, we just generate another random name. As it is, it should be extremely rare that the file would already exist given how much randomness we're using.

nemanja-boric-sociomantic:
> And what would be the benefit of that over using a random number generator?

I've read other (now outdated comments), but I missed that you're not using the global RNG instance, but you create your own here. That looks fair to me.

CyberShadow:
> I think that it's supposed to be /private/tmp

Symlink destinations are relative to the directory the symlink itself is in, so that would make no difference, except when the filesystem is mounted elsewhere than on /, in which case `private/tmp` would be better. Operating systems / distributions which utilize symlinks in their standard filesystem layout will often use relative symlinks for this purpose.

n8sh:
`Random` is an alias to a 32-bit MersenneTwisterEngine which has 2.5 kilobytes of state. This is excessive for this task, which does not benefit from the properties the Mersenne Twister provides in exchange for taking up so much space. I recommend one of the following:

1) Use a much smaller PRNG, such as `std.random.Xorshift`, which has 16 bytes of state and is completely adequate for this task; or

2) Going against @CyberShadow's recommendation at https://github.com/dlang/phobos/pull/5788#issuecomment-337727725 and using a reference to the thread-local `rndGen`.

std/file.d
```diff
@@ -670,20 +651,214 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+    // would be).
+    static Random random;
+    static bool initialized;
+    if (!initialized)
+    {
+        import std.algorithm.iteration : map;
+        import std.range : repeat;
+        static if (isSeedable!(Random, typeof(map!((a) => unpredictableSeed)(repeat(0)))))
+            random.seed(map!((a) => unpredictableSeed)(repeat(0)));
+        else
```
n8sh: This method of seeding calls `unpredictableSeed` 624 times to initialize the MersenneTwisterEngine's internal array. Currently this just means multiplying a number 624 times modulo 2^^31-1 and XOR-ing the result against the clock each time. This is pointless but fairly harmless, although it means that initially each number will have the same high bit. It would be faster and give a better result to just seed MersenneTwisterEngine with a single number. Also, if a proposal like https://github.com/dlang/phobos/pull/5230 to make `unpedictableSeed` more "unpredictable" is implemented, this mode of seeding could become pathological.

I hope this comment will be unnecessary due to the code having been changed either to use `rndGen` or to use a more lightweight PRNG than the Mersenne Twister.

jmdavis: I just copied what std.random was doing for initializing `Random`. I do think that Cybershadow raised a legitimate concern with `rndGen`, though honestly I think that the bigger concern is the fact that `Random` is a value type, making it stupidly easy to reuse random numbers across the program, and the risk of it being reseeded isn't all that high and causing problems - particularly when this is all TLS, and worst case, it has to go through every random number used to generate a temporary file previously where that file hasn't been deleted (which isn't great, but it's not terribly likely either). So, while I do think that using `rndGen` is a concern, I'm not sure that it's so big a concern that we shouldn't do it - especially if it's so large (which certainly wasn't something that I thought of).

jmdavis:
Okay. It's using `rndGen` again. That does have the downside that someone could reseed it, and cause `createTempFile` to have to make more attempts to find a valid file, but that's only an issue if someone is stupid enough to keep reseeding with the same seed, and I don't think that taking up extra memory for a separate random number generator to avoid that is worth it. Someone being that dumb can just shoot themselves in the foot. The fact that random numbers that were used prior to calling `createTempFile` could be reused if `rndGen` was used incorrectly is arguably problematic, but that's a problem with `rndGen` and std.random in general, and if we want to fix that, we really need to fix std.random.

CyberShadow:
Why do that instead of use XorShift? That seemed less compromising.

CyberShadow:
> Someone being that dumb can just shoot themselves in the foot.

This is a really unhelpful way of thinking when designing standard libraries.

jmdavis:
> Why do that instead of use XorShift? That seemed less compromising.

Because the more I think about it, the less reasonable it seems to create a static variable just for this function, and if `rndGen` isn't acceptable, we have far bigger problems. The default random number generator should be fine. The only downsides of which I am aware are

1. If the user keeps reseeding `rndGen` with the same seed.
2. If the user incorrectly uses `rndGen` before calling `createTemporaryFile` (e.g. by copying it when getting random numbers), and the random numbers that `createTemporaryFile` uses were some of the ones that were just used.

I don't see it as particularly realistic that someone is going to keep reseeding `rndGen` with the same value. Yes, it could happen, but it's a pretty stupid thing to do, and there's only so much that we can do to protect the user. And even if they do keep reseeding it with the same value, unless they're creating a lot of temporary files, it still won't matter, because worst case, all of the previously created temporary files still exist, and `createTemporaryFile` has to try each of them before getting to one that works, and there would have to be a lot of temporary files that were being created and not deleted for that to be a problem. If someone actually is foolish enough to keep reseeding `rndGen` with the same seed while using `createTemporaryFile`, and they use enough temporary files that it's a problem, then their program may be slow, but then they can debug it and fix the fact that they keep reseeding `rndGen`. It shouldn't be hard to figure out what's slow, even if it wouldn't be immediately obvious as to why.

The fact that `rndGen` is annoyingly easy to use incorrectly by accidentally copying it and reusing random numbers is definitely a design flaw of std.random, and it definitely does risk random numbers being reused, but the numbers that were used for a previous call to `createTemporaryFile` won't be reused, since it's not using it incorrectly. So, it's not going to result in a name conflict that will require trying a new name. Worst case, it makes the random names more predictable, but because of how the files are created and opened (specifically, the file must not exist previously), there isn't much of a security risk there. The only risk is that a program with the same permissions as the one using `createTemporaryFile` could access the file, and such a program could just watch `/tmp` for new files if that's what it's trying to do.

So, while I do agree that there are some problems using `rndGen`, I don't think that they're serious, and I don't like the idea of wasting memory for a static variable that shouldn't be necessary. I also don't think that it's good policy to be avoiding our own standard random number generator.

>> Someone being that dumb can just shoot themselves in the foot.

> This is a really unhelpful way of thinking when designing standard libraries.

There's a limit to how much we can reasonably protect a user from themselves, and I don't see why reseeding `rndGen` is something that many programs would be doing, let alone reseeding it with the same value continuously. To do that would show a complete lack of understanding of how pseudo-random number generators work, and the programmer doing it would likely run into far worse problems than having their program be slow when they create a lot of temporary files. And the fact that their program would be slow might actually help clue them into what they're doing wrong.

CyberShadow:
I think the implications of the possibility of security risks far outweigh the 4-byte cost of a counter variable.

jmdavis:
> I think the implications of the possibility of security risks far outweigh the 4-byte cost of a counter variable.

What risks? The fact that it would be easier to guess the file name? As I said, anything trying to do that could just watch /tmp. The fact that the file name is harder to guess just makes it harder. It doesn't actually stop anything. It's the file permissions that actually protect the file, and `createTempFile` deals with that problem correctly.

jmdavis:
Any time you open a file, you run the risk of another program with the appropriate permissions messing with it - and many of those files have known names and are in known locations. I don't see why temporary files would be special in that regard or why they would need would need more protections than a normal file. As long as we don't overwrite an existing file (which could therefore have different permissions and allow a program access which shouldn't have access), we should be fine. Or is there some other issue here that I'm not seeing?

CyberShadow:
Sorry if I was too terse or ambiguous.

Here's the situation I'm thinking of:

A program is generating random tokens (nonces or such) as part of a security protocol. It uses the global `rndGen` to do so.

At the same time, another part of the program (perhaps, in a library three Dub dependencies away) needs to create a temporary file. The function which does this also uses the global `rndGen`.

Now, because temporary directories are often global per machine, and any user is allowed to enumerate their contents, this leaks the state of `rndGen`. With enough temporary files, it should be possible to guess the initial random seed, and predict what random tokens the program has generated or will generate.

Now, by itself the description of the above doesn't seem likely that it could lead to an exploitable vulnerability. However, most security breaches these days occur due to a sequence of security weaknesses, which when exploited together, allow the attackers to reach their goal.

This is why I think that with any matter where the choice is a possible security weakness and a slight performance cost, the more secure solution should be heavily preferred.

DmitryOlshansky:
> 29 окт. 2017 г., в 20:07, Vladimir Panteleev <notifications@github.com> написал(а):
> 
> Sorry if I was too terse or ambiguous.
> 
> Here's the situation I'm thinking of:
> 
> A program is generating random tokens (nonces or such) as part of a security protocol. It uses the global rndGen to do so.
> 

WAT? That is precisely what it shouldn’t do.
There are strong PRNGs for cryptographically secure random numbers (CPRNGs). 

Fir instance XorShift can be cracked by observing a handful of numbers. Same for MerseneTwister.

CyberShadow:
Yes. The protocol might not require that the token be generated from a cryptographically secure random source. As I already said, it might very well be just part of the complete attack. For example, knowing the token might greatly reduce the amount of computation needed to brute-force the plaintext from the cyphertext.

DmitryOlshansky:
> 29 окт. 2017 г., в 21:16, Vladimir Panteleev <notifications@github.com> написал(а):
> 
> Yes. The protocol might not require that the token be generated from a cryptographically secure random source. As I already said, it might very well be just part of the complete attack. For example, knowing the token might greatly reduce the amount of computation needed to brute-force the plaintext from the cyphertext.
> 

Then said protocol is as secure as this attack is feasible. If you can break it knowing the token then the protocol is broken.

Protocol that allow easily guessable token are generally secured in some other way e.g. preshared secret.

I think the level of paranoia should be lowered a notch. People who do security for living are usually skeptical of std implementations of anything even remotely close to the sensitive data - nonces, keys, signatures, OTPs, temp files, caches, etc. Those who don’t ain’t worth their money and we can’t devise every mistake they might make by blindly using std library.

CyberShadow:
> Then said protocol is as secure as this attack is feasible. If you can break it knowing the token then the protocol is broken.

Computer security doesn't work that way. If you can design a flawless protocol or a write bug-free program, then none of this would be necessary, but the reality is the extreme opposite of that. Consider how many revisions of SSL and TLS it took to get to where we are now, and then how many bugs in their implementations have been found.

> People who do security for living

This is also a very unhelpful way to think about software today. Practically everything is Internet-connected, and a lot of things can become an attack vector.

Rust is winning mind-share because they are serious and proactive about security, and I don't mean just memory safety. For example, their hashmaps use a random value as part of the hash, to prevent denial of services occurring from an attacker sending input that generates tons of hash (bucket) collisions. We need to do the same.

DmitryOlshansky:
> 30 окт. 2017 г., в 0:06, Vladimir Panteleev <notifications@github.com> написал(а):
> 
> Then said protocol is as secure as this attack is feasible. If you can break it knowing the token then the protocol is broken.
> 
> Computer security doesn't work that way.
> 
It does work exactly like that. Making something less insecure is how it doesn’t work.
> If you can design a flawless protocol or a write bug-free program, then none of this would be necessary, but the reality is the extreme opposite of that. Consider how many revisions of SSL and TLS it took to get to where we are now, and then how many bugs in their implementations have been found.
> 

Flawless protocol is strange remark. Protocol  have to be flawless to the point that nobody can find a sufficient vulnerability (yet), implementations typically are the problem.

All in all, protocols are designed such that
1. Something is a secret and must be protected (including from guessing).
2. Something is not a secret and may be shared freely.

Anything in “gray” area is just asking for problems. You don’t secure things by making them less insecure.
> People who do security for living
> 
> This is also a very unhelpful way to think about software today. Practically everything is Internet-connected, and a lot of things can become an attack vector.
> 
Then whoever does that software will have to care about security as well. We have to hire people if don’t grasp the subject.
> Rust is winning mind-share because they are serious and proactive about security, and I don't mean just memory safety. For example, their hashmaps use a random value as part of the hash, to prevent denial of services occurring from an attacker sending input that generates tons of hash (bucket) collisions.
> 
100% to be precise. Which is importantly a proper well known attack vector that even Perl has patched.
> We need to do the same.
> 

But if you were to follow the lead fool proofing random generator you’d need CPRNG e.g. Fortuna as default rng. Picking XorShift is exactly what you don’t do. 

I wouldn’t be opposed to introduce secure rngs by default but it’s none of this PR business. If somebody can get it right that is.

All in all fool proofing things with an eye to potential security problem is fine, but:
1. You don’t pick an arbitrary PR to do this, it is a full-scale project itself that may entail dozens of PRs.
2. Even then some caution is required to not go overboard as we should not impose “less insecure” primitives on people that just run simulations and mostly care about speed. An extreme example would be imposing constant time string compare on everybody because the might compare secure hashes that way. 
3. False sense of security. So imagine we done wonders to fool-proof at library level trying to make sure a user can’t use it insecurely. It says nothing about security of actual applications, sadly.
4. A great deal of exploits are trivial buffer overruns still, so focus is important as well.
5. We need at least a couple of security audits. Rust has a trove of experience with Firefox, we have nothing.

Anyhow before this goes OT.

Please stop mutilating this PR motivating it by security. Your suggestion still doesn’t help it in any capacity.

CyberShadow:
> 1. You don’t pick an arbitrary PR to do this, it is a full-scale project itself that may entail dozens of PRs.

Wrong. This is something we need to be vigilant about everywhere. "Major cleanup efforts" almost never happen, incremental improvements and improvements in the process that lead to higher quality additions do.

> Even then some caution is required to not go overboard as we should not impose “less insecure” primitives on people that just run simulations and mostly care about speed.

There are no measurable performance issues here.

> So imagine we done wonders to fool-proof at library level trying to make sure a user can’t use it insecurely. It says nothing about security of actual applications, sadly.

It would still be a huge improvement! This is something I've failed to get across - security is NOT black-and-white, it comes in layers and needs to be done in-depth as well as on the surface (the protocols being "secure in theory"...). If you think otherwise, you *should not be writing standard library code*!

> We have to hire people if don’t grasp the subject.

As far as I know I'm the closest we have to a "security person" in this discussion, but apparently my opinion isn't worth anything.

> Anyhow before this goes OT.

There is no better place to discuss policy than regarding a concrete example affected by it.

Anyway, here's what I'm upset about:

This pull request had the more secure version at some point, and despite consensus, it was changed to a less secure version. All you had to do was change the RNG type to `XorShift`!

Anyhow, none of this is needed. You can have an implementation which neither relies on leaking global state nor needs its own private static variables - just use a sufficiently precise clock, possibly with the PID/TID, and keep retrying on EEXIST. @jmdavis let me know if you'd like me to have a go at implementing that.

jmdavis:
I'm going to leave the decision about whether `rndGen` is enough of a security risk up to @andralex . He has to approve new symbols to Phobos anyway.

Personally, I don't think that it's a big enough deal to be concerned. Anyone doing anything that needs to care about security with random numbers shouldn't be using `rndGen`. They should be using whatever algorithms are appropriate to the task and probably need a cryptographically secure random number generator. And I would expect it to be an extreme case that a program would be creating enough temporary files that someone could reasonably use it to figure out the state of `rndGen`. In this case, I think that avoiding our own standard number generator is just being paranoid, particularly since it would only be an issue if a program is both misusing `rndGen` and creating a ton of temporary files.

FWIW, the FreeBSD implementation of `mkstemp` does essentially what we're doing here to generate random names (the main difference being that it iterates characters on a conflict rather than generating a new name), and it uses the system's `arc4random` call. It doesn't do anything special with it in an attempt to have its own random number state or anything like that. Granted, (IIUC) it's a stronger number generator than Mersenne Twister, but they don't appear to be concerned about avoiding the system's random number generator.

std/file.d
```diff
@@ -670,20 +651,200 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+    See_Also:
+        $(REF tmpfile,std,stdio)
+  +/
+string createTempFile(const void[] buffer = null) @safe
+{
+    return createTempFile(null, null, buffer);
+}
+
+/// Ditto
+string createTempFile(const(char)[] prefix, const(char)[] suffix, const void[] buffer = null) @trusted
```
wilzbach: Why not `string` or using templates?

std/file.d
```diff
@@ -670,20 +651,200 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+        name[$ - suffix.length .. $] = suffix;
+
+        auto random = &rndGen();
+        rndGen.popFront();
+
+        auto chars = chain(letters.representation, digits.representation);
+        foreach (ref c; name[prefix.length .. $ - suffix.length])
+        {
+            c = chars[random.front % chars.length];
+            random.popFront();
```
wilzbach: Why not simply using [`std.random.choice`](https://dlang.org/phobos/std_random.html#.choice)? It's made for this ;-)

std/file.d
n8sh: If it were `string` and the prefix/suffix weren't already immutable then it would be necessary to create copies of them to pass to the function. What benefit would immutability provide in exchange?

If the prefix/suffix were template parameters then it would be impossible for them to be specified at runtime, so I think the question is whether using templates gives some benefit that justifies that limitation.

quickfur:
Wow. I'm almost sorry that my comments about security led to a big sidetracked discussion.

I think as far as moving this PR forward is concerned, there are two points:

1. The current implementation doesn't necessarily need to be perfect from a security POV, *as long as the API does not preclude a later, fully-secure solution*.  I think this is a classic case of letting the perfect be the enemy of the good.  Creating temporary files with unique names is a fairly common operation, and not every use case needs to be as completely paranoid as the comments on this PR seem to have been, so it's something we *should* provide.

2. The current hesitation about this PR seems to be centered around the security of `std.random`, which is used for generating the temporary filename.  So what about making it possible for the *user* to supply an RNG of his choice? Make it a default parameter that defaults to `std.random.rng` or whatever it is we wish to use by default.  If the user is paranoid, he can supply his own CSPRNG to this function for use.

So I'd say let's do this:

1. Review the API of the proposed function to see if it is sufficient to allow a possibly-future implementation to be fully secure. If not, let's devise an API that does allow this. While it would be nice if the current implementation can also be fully secure, IMO we shouldn't wait around until that happens. Don't let the perfect become the enemy of the good.

2. Add an optional parameter for the user to supply a (CS)PRNG of his own, if he's paranoid enough. AFAICT, this is probably the only thing that needs to be done to make the proposed API secure.

3. As long as the API is secure in the above sense (let's call it a "securable API"), I think that's good enough to merge this PR.  If security holes are discovered afterwards, we can fix it without changing the API and breaking user code.

4. Get @andralex's attention on approving this new symbol, then merge ahoy!

quickfur:
ping @jmdavis

jmdavis:
Okay. I rebased on master and updated it to use `choice`, but as far as having the random number generator as a parameter, I think that that's overkill. `createTempFile` already has a default parameter. I think that adding another is just making it unnecessarily complicated and that the concerns about using `rndGen` are overblown - not to mention, I think that we're in serious trouble if we're avoiding our own random number generator.

Think about this for a moment. The only case where there's even potentially a security problem with using `rndGen` in `createTempFile` is if someone uses `rndGen` for something in their program where they shouldn't (e.g. cryptography) and then creates so many temporary files with `createTempFile` in a short enough span of time that someone else with access to the computer is able to figure out in what state the default random number generator is in and exploit the program based on that. Not only is such a case highly unlikely (due to how many temp files they'd have to generate in a relatively short period of time), but it requires that someone misuse `rndGen` in their code. Having `createTempFile` use a different random number generator might reduce their risk, but they're already at risk from having used `rndGen` when they shouldn't, and having an optional argument to `createTempFile` to specify an alternate random number generator isn't going to help them any, because that requires that they then have the presence of mind to realize that they need to pass a different random number generator to `createTempFile`, because they're incorrectly using `rndGen` elsewhere. And if they have the presence of mind to do that, then they just won't misuse `rndGen`. The optional argument for the random number generator doesn't really help anyone.

The fact is that if someone is going to be doing something that requires secure random numbers, then they need to know enough to use a proper random number generator for the task, and avoiding `rndGen` everywhere just because someone might be an idiot with it ultimately isn't going to save them, and IMHO, it actively makes Phobos worse. If we're avoiding using our own random number generator in cases which don't need stuff like cryptographic security, we have a serious problem. Phobos functions should be able to use `rndGen` - especially for something as simple and stupid as creating a name for a temporary file.

The only way that I'm doing anything different with the random number generator here is if @andralex  requires it for `createTempFile` to be included in Phobos.

`createTempFile` solves the real security problem, which is ensuring that the file is created when it's written so that the file permissions don't get screwed up by someone having created the file with other permissions first.

quickfur:
When it comes to security, there's always a loophole if you look hard enough. (And unfortunately, in this day and age, there are too many who would want to do so.)

It doesn't have to be *another* part of the code that (ab)uses `rndGen` for cryptographic purposes; `createTempFile` itself can be exploited once the state of the PRNG is known. Imagine if the program is a web server that needs to create a temporary file in order to perform some computation that generates a dynamic webpage in response to an incoming request. If an attacker knows the state of the PRNG, he can predict what the next "random" filename would be, and in fact, the next *n* filenames for an arbitrarily large *n*. So he creates them in advance, and forces `createTempFile` to be unable to find a suitable filename. The result is an effective denial-of-service attack against the web server.

How likely is this to happen? Not very likely. And how relevant is this to this PR? IMO, not too much. But let's not claim wrongly that it's impossible. In this day and age, a slight flaw like this can easily be part of a larger exploit scheme, which, while it may seem harmless on its own, can add up to something not so harmless once coaxed into the wrong context.

My point, though, was not to get into these sorts of what-if scenarios, which will trigger a discussion that may literally never end. Rather, the point is to design an API that can *in theory* be secured, even if the current implementation isn't secure. That way, we leave ourselves the door to future improvements as necessary.  Otherwise the perfect will forever be the enemy of the good, and this PR will languish in the queue forever.

jmdavis:
IMHO, it simply is not worth complicating the API yet further by adding another optional argument - especially over concern about something that is realistically a non-issue.

jmdavis:
Regardless, at this point, I'm leaving the decision up to @andralex. If it were my decision, it would go in as-is, but @andralex needs to approve the symbol regardless.

CyberShadow:
@jmdavis Why not use the clock as suggested above?

jmdavis:
> @jmdavis Why not use the clock as suggested above?

The random number generator is for generating random values. If we can't use our own random number generator when we want to generate random numbers, then we have a serious problem. This code is using `rndGen` for exactly what it's for.

CyberShadow:
That sounds like an ideological argument in the face of a practical solution. The argument is also not as simple, since there are several kinds of random numbers, and `rndGen` is not even "our own random number generator" - it's also the user's, and the user can e.g. seed it to the same seed that might cause lots of collisions in this function.

Any practical objections to using the clock, esp. compared to the downsides of using a RNG as discussed above?

std/file.d
```diff
@@ -672,20 +653,200 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+        static assert(0, "Unsupported OS");
 }
 
+
+/++
+    Creates a file with a random name in $(LREF tempDir) and returns the name
+    of the file. If data is passed in, then the file is written with that data;
+    otherwise, it's empty.
+
+    The idea is that this creates a temporary file for testing or whatever
```
wilzbach: "The idea is" sounds colloquial. How about simply "This creates ..."

std/file.d
```diff
@@ -672,20 +653,200 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+    other purpose a temporary file might be needed for. However, it will only
+    be deleted if it's explicitly deleted or if the OS decides to delete it as
+    some OSes do with files in temp directories on startup or shutdown. So,
+    unlike $(REF tmpfile,std,stdio), the file has a name, and it can have all
```
wilzbach: `$(REF tmpfile, std,stdio.File`) - https://dlang.org/phobos/std_stdio.html#.File.tmpfile

std/file.d
```diff
@@ -672,20 +653,200 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+        suffix = Suffix to the random portion of the file name (e.g. for a file
+                 extension).
+        buffer = Data to populate the file with.
+
+    Returns:
+        The name of the generated file.
+
+    Throws:
+        $(LREF FileException) if it fails to create the file.
+
```
wilzbach: If we are really paranoid a note could be added that this function:

> This function isn't designed to be cryptographically secure, and are therefore unsuitable for cryptographic or security-related purposes

But then again Captain Obvious tells you that using temporary files in a shared directory is a pretty bad idea for anything that claims to be secure.

std/file.d
```diff
@@ -672,20 +653,200 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+    See_Also:
+        $(REF tmpfile,std,stdio)
```
wilzbach: `std,stdio,File`

std/file.d
aG0aep6G: `$(REF File.tmpfile, std,stdio)`

CyberShadow:
> Clock and PID/TID are global state. Try accessing PID/TID in pure.

Same with `rndGen`, so I don't see your point.

> Sure but the same can be said about clock + tid.

No, that misses the point. The current time is not a secret!

> Summary

Here is the summary from where I'm standing:

The current implementation:
- Requires that `rndGen` is in a sane state, which it might not be, because it is exposed to the user, and reseeding it is a supported operation
- Requires initializing `rndGen` for each new thread, which (surprise!) uses [the same data as suggested above](https://github.com/dlang/phobos/blob/5d37ac123f6a812246eaa5f3fdd36aa31d576526/std/random.d#L1309-L1310) (time + PID + TID)
- Leaks the state of `rndGen`, whether you consider it a problem or not

We've now had several proposals that improve on the above list, and you reject all of them with "stop mutilating this PR motivating it by security"?

Anyway, I'm not blocking this PR as it is. For better or for worse, a lot of languages do no better than the current implementation. The public interface is fine as it is, and the implementation can be improved afterwards. But I can't understand the joint opposition to some (IMHO) sensible ideas for improving it now.

jmdavis:
> But I can't understand the joint opposition to some (IMHO) sensible ideas for improving it now.

Because you're trying to get the code changed to not use the normal random number generator to generate random numbers and instead do something less obvious because you're afraid that someone is going to do something stupid with the default random number generator. If we go with that argument, then we shouldn't ever be using `rndGen`, which makes it kind of pointless. This code is using `rndGen` for what it's designed for, and I don't think that any of the concerns posed merit avoiding using a feature for exactly what it's designed for. If someone wants to shoot themselves in the foot by doing something with `rndGen`, then I'm inclined to let them.

CyberShadow:
> If we go with that argument, then we shouldn't ever be using `rndGen`, which makes it kind of pointless.

Yes, I think we probably shouldn't be using `rndGen` in Phobos at all. There ought to be an internal instance for use in Phobos, and one public for the user (only!), which we never touch. The state of the RNG is still state that ought to be properly encapsulated.

std/file.d
```diff
@@ -672,20 +653,200 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+    be deleted if it's explicitly deleted or if the OS decides to delete it as
+    some OSes do with files in temp directories on startup or shutdown. So,
+    unlike $(REF tmpfile,std,stdio), the file has a name, and it can have all
+    of the things done to it that one might typically do with a file (including
+    reopening it after closing it).
+
+    The file is created with R/W permissions. On POSIX systems, the permissions
+    are restricted to the current user, though the effective permissions are
+    modified by the process' umask in the usual way.
+
```
CyberShadow: Could you add that it uses the global `rndGen` to generate the file names? This should allow us to close the `rndGen` discussion.

jmdavis:
I added a not to the documentation about `rndGen`, and I added a changelog entry.

std/file.d
quickfur: Sometimes it's good to state the obvious, since what's obvious to us may not be obvious to others, e.g., a new learner of D.

n8sh:
AFAICT the 'controversy' was put to rest on January 26. @CyberShadow [was satisfied with this resolution](https://github.com/dlang/phobos/pull/5788/files#r164132896):
>Could you add that it uses the global rndGen to generate the file names? This should allow us to close the rndGen discussion.

std/file.d
```diff
@@ -672,20 +653,202 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+///
+@safe unittest
+{
+    import std.algorithm.searching : startsWith;
+    import std.range.primitives;
+    import std.path : baseName, buildNormalizedPath, dirName, extension;
+
+    {
+        auto path = createTempFile();
+        scope(exit) if (path.exists) remove(path);
```
wilzbach: Why `if (path.exists)`. Should this either be `scope(exit) assert(path.exists), remove(path);` or `scope(exit) remove(path);`?

jmdavis: The `path.exists` portion could be removed, but I pretty much always use it before removing a file so that I don't have to worry about `remove` throwing due to something that happened when the code is changed later or something else unforeseen goes wrong.

jmdavis: If `tempFile` _does_ fail to create the file for some reason, then one of the tests after the scope statement will fail, and the `AssertError` will bubble up properly, but if the test isn't there, then in that case, we'd then get a `FileException` being thrown instead when `remove` failed. So, IMHO, it is better to have the check even if it's useless if everything is working correctly.

jmdavis:
Ah. Great. It looks like someone committed a change to std.random which breaks this PR.

quickfur:
@n8sh I agree, I think we should just merge this once the autotester failure has been fixed.  A large part of the discussion in this PR is just Yet Another Case of the perfect being the enemy of the good.  Surely we can merge this first and then fix it later if it's really *that* necessary.  My suspicion is that nobody will actually care (and those who are paranoid enough to care will probably just write their own version of the function anyway, instead of relying on Phobos).

jmdavis:
The breakage was caused by [this PR](https://github.com/dlang/phobos/pull/6302), which is going to need to be reverted.

jmdavis:
Okay. It looks like this is passing again now that the breaking change to std.random.choice has been reverted, and in theory, the controversy with the random number generator was solved by adding a note to tho documentation about it. But someone still has to actually give an approving review for the PR.

std/file.d
```diff
@@ -821,27 +820,9 @@ version(Posix) private void writeImpl(const(char)[] name, const(FSChar)* namez,
...
-        {
-            cnt = (size - sum < 2^^30) ? (size - sum) : 2^^30;
-            const numwritten = core.sys.posix.unistd.write(fd, buffer.ptr + sum, cnt);
-            if (numwritten != cnt)
-                break;
-            sum += numwritten;
-        }
-        cenforce(sum == size, name, namez);
-    }
-    cenforce(core.sys.posix.unistd.close(fd) == 0, name, namez);
```
CyberShadow: General note, separating related refactorings into their own commit makes reviews easier and faster

std/file.d
```diff
@@ -809,8 +810,6 @@ if (isConvertibleToString!R)
     static assert(__traits(compiles, append(TestAliasedString("foo"), [0, 1, 2, 3])));
 }
 
-// Posix implementation helper for write and append
-
```
CyberShadow: Why?

std/file.d
```diff
@@ -868,20 +849,202 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
     immutable size = buffer.length;
     size_t sum, cnt = void;
-    DWORD numwritten = void;
+
     while (sum != size)
     {
-        cnt = (size - sum < 2^^30) ? (size - sum) : 2^^30;
-        WriteFile(h, buffer.ptr + sum, cast(uint) cnt, &numwritten, null);
-        if (numwritten != cnt)
+        cnt = size - sum < 2^^30 ? size - sum : 2^^30;
```
CyberShadow: Why? The parens were redundant but they were clearly there for a reason (readability).

CyberShadow: Also, mixing stylistic changes, refactorings, and adding new code, all in one commit, is pretty bad. Please avoid doing this in the future.

std/file.d
CyberShadow: Bump, this wasn't addressed.

std/file.d
```diff
@@ -868,20 +849,202 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+    (including reopening it after closing it).
+
+    The file is created with R/W permissions. On POSIX systems, the permissions
+    are restricted to the current user, though the effective permissions are
+    modified by the process' umask in the usual way.
+
+    $(REF rndGen, std, random) is used as the random number generator.
+
+    Params:
+        prefix = Prefix to the random portion of the file name.
```
CyberShadow: Can this be a directory path? In some terminology (e.g. autotools) "prefix" is implied to mean a path prefix. Please clarify.

std/file.d
```diff
@@ -868,20 +849,202 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+        foreach (ref c; name[prefix.length .. $ - suffix.length])
+        {
+            c = choice(chars);
+            random.popFront();
+        }
+
+        return buildPath(tempDir, name);
+    }
+
+    while (1)
```
CyberShadow: Using `1` as `true` is a C-ism (which does not have a `bool` type). (Since D has a `bool` type, numbers should be used to indicate quantity, and 1 here does not indicate the quantity of anything.) Please use `while (true)` in D.

std/file.d
```diff
@@ -868,20 +849,202 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+        else
+            static assert(0, "Unsupported OS");
+
+        writeToOpenFile(fd, buffer, filename, null);
+
+        return filename;
+    }
+}
+
+///
```
CyberShadow: This is a fine unit test, but not a good example. Would be nice to have a human-friendly example alongside this unit test.

std/file.d
```diff
@@ -672,20 +653,200 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+        import std.random : choice, rndGen;
+        import std.range : chain;
+        import std.string : representation;
+
+        auto name = new char[](prefix.length + 15 + suffix.length);
+        name[0 .. prefix.length] = prefix;
+        name[$ - suffix.length .. $] = suffix;
+
+        auto random = &rndGen();
+        rndGen.popFront();
```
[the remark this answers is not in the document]
n8sh: I think the idea here is that the value currently in `rndGen.front` could have already been used by someone else who didn't then pop it off. `choice` uses `uniform` which uses `front` then pops afterwards.

std/file.d
```diff
@@ -868,20 +849,202 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+                if (errno == ERROR_FILE_EXISTS)
+                    continue;
+                else
+                    throw new FileException("Failed to create a temporary file", errno);
+            }
+        }
+        else
+            static assert(0, "Unsupported OS");
+
+        writeToOpenFile(fd, buffer, filename, null);
```
n8sh: Is returning without closing the file descriptor okay?

quickfur: Good catch! This is a raw OS handle, not a `std.stdio.File`, so there is no refcounting or GC or anything of that sort.  So unless I'm missing something obvious, yes we will need to close the file descriptor.

wilzbach:
Bump. Maybe move this to `std.experimental` for one release or so, s.t. the nitpicks and final issues can be addressed without this PR lingering in the queue forever?

jmdavis:
> Bump. Maybe move this to std.experimental for one release or so, s.t. the nitpicks and final issues can be addressed without this PR lingering in the queue forever?

We've never done that, and I see no reason to start now. It would just result in quickly breaking the code of anyone who actually used it. I just need to address the last few comments, and it should be good to go. Hopefully, I can get to it this weekend.

n8sh:
BTW, why does this close the file and return only its name rather than returning a `std.stdio.File` that has both an fd and a name?

jmdavis:
> BTW, why does this close the file and return only its name rather than returning a std.stdio.File that has both an fd and a name?

std.file does not use `std.stdio.File`. std.file always operates on an entire file at a time. Anything which uses `File` goes in std.stdio. Previously, we briefly had a function like that in std.stdio, but it got removed, because of issues with the compiler, it increased the size of hello world, and that was deemed unacceptable. This is an alternate way to go about the problem that does not run into that issue.

And personally, for basically every use case that I would have for a function like this, returning a `File` would just be annoying. The typical use case would be writing a file for a unit test (or maybe to pass off to some script) where something else would be reading the file afterwards. So, the only reason that having a `File` would make sense would be if you were looking to write to it in pieces rather than just write it all at once (and in general, if it's not the plan to close and reopen the file, then `std.stdio.File.tmpfile` would work just fine, and having a name for the file is pretty pointless). Anyone who wants a `File` can always use `createTempFile` to create the file and then open it with `File`, which is a bit annoying if you really want a `File`, but it works, and it would not be appropriate to return a `std.stdio.File` in std.file.

n8sh:
@jmdavis got it

changelog/std-file-createTempFile.dd
```diff
@@ -0,0 +1,15 @@
+`createTempFile` added to std.file
+
+$(REF std, file, createTempFile) creates a file in $(REF tempDir, std, file)
+with a random name (with an optionally provided prefix and suffix) and returns
```
andralex: repeated "with", use e.g. s/with an/including an/

changelog/std-file-createTempFile.dd
```diff
@@ -0,0 +1,15 @@
...
+the file name. It also optionally takes data to write to the file when creating
+it (if no data is provided, then the newly created file will be empty).
```
andralex: It's awkward that the file is created, closed, to presumably be opened by the user again right after. Shouldn't this function return e.g. an opened `File`? Then people can call `name` to get its name.

changelog/std-file-createTempFile.dd
```diff
@@ -0,0 +1,15 @@
...
+
+This is different from $(REF File.tmpfile, std, stdio), which creates a
+temporary file with no name, returns a $(REF File, std, stdio) to that file,
+and then deletes the file when the $(REF File, std, stdio) is closed. The file
+created by $(REF std, file, createTempFile) is a normal file which can be
+opened and closed as many times as desired and will only be deleted by the
+program if the program explicitly does so (though it is sitting in a temp
```
andralex: s/is sitting/is created/

changelog/std-file-createTempFile.dd
```diff
@@ -0,0 +1,15 @@
...
+directory and is subject to the normal things tha happen to such files - e.g.
```
andralex: "normal things" is too colloquial, please rephrase

andralex: typo: "tha"

std/file.d
```diff
@@ -821,27 +820,9 @@ version(Posix) private void writeImpl(const(char)[] name, const(FSChar)* namez,
...
-            cnt = (size - sum < 2^^30) ? (size - sum) : 2^^30;
-            const numwritten = core.sys.posix.unistd.write(fd, buffer.ptr + sum, cnt);
-            if (numwritten != cnt)
-                break;
-            sum += numwritten;
-        }
-        cenforce(sum == size, name, namez);
-    }
-    cenforce(core.sys.posix.unistd.close(fd) == 0, name, namez);
+    writeToOpenFile(fd, buffer, name, namez);
```
andralex: who closes fd?

std/file.d
```diff
@@ -868,20 +849,202 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+    else version(Windows)
+        cenforce(sum == size && CloseHandle(fd), name, namez);
+    else
+        static assert(0, "Unsupported OS");
 }
 
+
+/++
+    Creates a file with a random name in $(LREF tempDir) and returns the name
+    of the file. If data is passed in, then the file is written with that data;
```
andralex: s/returns the name of the file/returns its name/

Generally the prose could use some editing.

std/file.d
andralex: "whatever" is also odd

std/file.d
```diff
@@ -868,20 +849,202 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+    otherwise, it's empty.
+
+    The idea is that this creates a temporary file for testing or whatever
+    other purpose a temporary file might be needed for. However, it will only
+    be deleted if it's explicitly deleted or if the OS decides to delete it as
+    some OSes do with files in temp directories on startup or shutdown. So,
+    unlike $(REF File.tmpfile, std, stdio), the file has a name, and it can have
+    all of the things done to it that one might typically do with a file
```
andralex: again with "things"...

std/file.d
```diff
@@ -868,20 +849,202 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+    (including reopening it after closing it).
+
+    The file is created with R/W permissions. On POSIX systems, the permissions
+    are restricted to the current user, though the effective permissions are
+    modified by the process' umask in the usual way.
```
andralex: Would it be important to allow configurable permissions?

std/file.d
```diff
@@ -868,20 +849,202 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
...
+    Returns:
+        The name of the generated file.
+
+    Throws:
+        $(LREF FileException) if it fails to create the file.
+
+    See_Also:
+        $(REF File.tmpfile, std, stdio)
+  +/
+string createTempFile(const void[] buffer = null) @safe
```
andralex: Thsi shouldn't be an overload. Consider: `createTempFile("my_prefix_")`. The user will get the wrong idea. Let them type `null`, `null` if they don't care.

jmdavis:
> I'm trying to figure what this has over

>     auto tmpName = tempDir.buildPath("my_" ~ uniformAlphaNumerics(15) ~ ".txt");

The biggest thing that a function like this does is guarantee that the file is created when it's opened, since if it's not actually created when it's opened, you have a security flaw, since someone could theoretically create the file beforehand with different permissions. And it's not the sort of thing that someone is usually going to think of unless it's been explained to them before. So, a large percentage of folks who would try something like this would try a line like you just suggested and simply open the file with that and have a security flaw in their program. All other details aside, that is the big reason to have a function like this.

changelog/std-file-createTempFile.dd
jmdavis: Returning an open `File` would fit in std.stdio, not std.file, because std.file specifically only ever deals with entire files at at time. Also, in most cases, I'd personally find it very annoying to have a `File` to deal with. The main purpose is to write data to a file for testing or to pass to another program and then pass the file name to something else to open it. Having an open `File` rather than just a filename just makes that use case more annoying. We did have a version of this in std.stdio briefly that did that, but it got removed from std.stdio on the grounds that it increased the size of "hello world" due to issues with the compiler.

The way I would expect to use this normally would be to just call this function like `write` and write the whole thing at once, and in the rarer case where they want a `File`, they can just not pass any data to it and then reopen it with `File` (in which case, since the file was already created, there is no security risk like there is the first time).

std/file.d
jmdavis: Important? I don't think so. We don't provide that ability with `write`, and that does start getting a bit system-specific to implement. It might be a reasonable enhancement, but if we did something like that, then we'd need to do the same with other functions like `write`. So, if we want to do that, I think that it should be left to a separate PR.

std/file.d
jmdavis: Fine with me. I'd expect the normal use case to involve passing data anyway. I suspect that I assumed that folks would find it annoying to have to pass `null` in the case where they didn't want to write data, but it's been a while since I wrote this, so I don't remember for sure.

std/file.d
jmdavis: Looking at the code, `writeToOpenFile` does. I suppose that I could rename it to `writeToOpenFileAndClose` or something to make it clearer. It's just a private function that was created to refactor some code that `writeImpl` uses rather than duplicating it.

std/file.d
jmdavis: Actually, looking at the code `writeToOpenFile` closes the file descriptor so that the code that involves closing the file descriptor is shared with `writeImpl` just like the code to actually write to the open file is. I guess that I should have named it in a way that made it clear.

Flying-Toast:
How is this different than `std.stdio.File.tmpfile`?

FraMecca:
What is this missing (apart from the small conflicts in std/file.d) for merging?
I can't see the CI failures.

It seems like most of the work was done by @jmdavis so I am bumping this

jmdavis:
@FraMecca  IIRC, it's basically good but needs to be cleaned up a bit (e.g. one of the function names was named badly and was confusing reviewers about whether the file descriptor was being closed properly or not), though I'd have to look over it all again to see what exactly would need to be tweaked. I've just never gotten around to cleaning it up.

Imperatorn:
Status?

LightBender:

PR #5662, opened 2017-07-29
Fix Issue 16744: Add Typeof

MetaLang:
This an alternative implementation to https://github.com/dlang/phobos/pull/4920

PR #4920, opened 2016-11-24
Implement issue# 16744: TypeOf template helper.

jmdavis:
typeof cannot be used directly with templates such as staticMap, because
we don't have the template equivalent of lambda functions. TypeOf is
therefore a template wrapper around typeof so that it can be used in
contexts that require a template.

PR #5789, opened 2017-10-18
Fix Issue 16744: Add Typeof (Redux)

MetaLang:
Previous PR: https://github.com/dlang/phobos/pull/5662. I'm resubmitting this after it was reverted in https://github.com/dlang/phobos/pull/5664.

This an alternative implementation to #4920.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -5551,24 +5551,24 @@ template BuiltinTypeOf(T)
 
 /**
     A template designed to function in the same manner as the
-    built-in `typeof` utility. As it is a template, Typeof can be
-    passed around like any other symbol, unlike `typeof`.
-    Also, whereas `typeof` will not accept symbols that are already
-    types (e.g., `typeof(int)` will not compile), Typeof will accept
-    both values and types.
+    built-in `typeof` utility. As it is a template, `Typeof` can be
+    passed around like any other symbol, unlike `typeof`. Also,
+    whereas `typeof` will not accept symbols or expressions that
+    evaluate to a type (e.g., `typeof(int)` will not compile),
+    `Typeof` will accept both values and types.
 
     Params:
-        T =  The symbol to take the type of. Can be either a type
+        T =  The symbol or expression to take the type of. Can be either a type
              or a value. If it is something that does not have a valid
              type (such as a module name or an expression that
-             evaluates to the special `__error__` symbol), an error
+             evaluates to the special `__error__` type), an error
              will be raised at compile time.
 
     Returns:
-        The type of the given symbol.
+        The type of the given symbol or expression.
 
     Note:
-        If the given symbol is a template, the result type will be `void`.
+        If a template is passed to `Typeof`, the result type will be `void`.
 */
 template Typeof(T...)
 if (T.length == 1)
--- a/std/traits.d
+++ b/std/traits.d
@@ -75,6 +75,7 @@
  *           $(LREF SharedConstOf)
  *           $(LREF ImmutableOf)
  *           $(LREF QualifierOf)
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -5548,6 +5549,132 @@ template BuiltinTypeOf(T)
...
+    else static if (is(T[0] TypeOfT0))
+        alias Typeof = TypeOfT0;
+    else static if (is(typeof(T[0]) TypeOfT0))
+        alias Typeof = TypeOfT0;
+    else
+        static assert(false, "Cannot take the type of " ~ T[0].stringof);
+}
+
+///
+@safe unittest
```
atilaneves: I'd add `pure` and convert the `assert`s into `static assert`s.

MetaLang: I don't see a point in adding pure as `Typeof` is a template, not a function.

Do we have a policy on using `static assert` vs. `assert` in unit tests for stuff that can be checked at compile time? I don't know if it's necessarily a good idea for _compilation_ to fail when building with tests enabled, but that's just a vague sense of unease on my part.

andralex: static assert is better than assert but keep in in the unittest so it doesn't slow down regular compilation

andralex:
Wished there were more examples where `Typeof` saves the day. Are there examples without `staticMap`? If not, perhaps `mapTypes` would be a better way to go.

std/traits.d
MetaLang: I've changed all asserts to static asserts.

MetaLang:
@andralex I've run into this a lot in template-heavy generic code. I'll try and find a few examples.

This could easily be done in user code as it's a very small template to define, but IMO the value in having this in the standard library is that it handles non-obvious edge cases, such as trying to pass a module or `__error__`.

Now that I think about it, however, this also reifies the `typeof(template) -> void` behaviour of DMD, which may(?) change in the future. Should I remove that and just return whatever regular `typeof` returns, or should this behaviour be kept?

andralex:
I'd want to focus on something higher-level and useful instead of something that is just like typeof but a bit different. Those fare poorly because people are never sure which one they're supposed to use etc. In contrast, a higher-level primitive will shine by its own utility.

That's why I'm saying: if most of the time the use of `Typeof` is in conjunction with `staticMap`, we're better off with a template `Types`. When given a list of types, it yields an `AliasSeq` containing the type of each. No type? The empty `AliasSeq`. One type? AliasSeq with one element. Perhaps that would have more use cases.

MetaLang:
@andralex you're confusing me now, because you seem to be talking about something different from what's requested in the bug report. `Typeof` does the exact same thing as `typeof`, except when passed a type it just aliases itself to that type instead of causing a compile error (if you try to do `typeof(int)`, it won't compile), and it can be passed as a normal symbol (i.e., `staticMap!(Typeof, <list of values and/or types>)` will work whereas `staticMap!(typeof, <list of values and/or types>)` will not).

To my knowledge, `Typeof` has the same behaviour as `typeof` in every other respect. If you pass a symbol that's a module name to `Typeof`, or the special `__error__` symbol, it will generate a compile-time error just like `typeof` would. Likewise, if you pass a template to `Typeof`, it'll return `void` just like `typeof` currently does (the only reason I brought it up is because I'm pretty sure this is an implementation detail and not in the spec).

andralex:
> @andralex you're confusing me now, because you seem to be talking about something different from what's requested in the bug report.

Yes. More specifically: I am trying to resolutely steer away from "just about the same, only a bit different".

MetaLang:
Okay, then the enhancement request should be closed as WONTFIX. I'll close this PR and close the bug.

PR #5790, opened 2017-10-18
allow n=0 for std.meta.Repeat

9il:


std/meta.d
```diff
@@ -1317,9 +1317,12 @@ private template SmartAlias(T...)
  * Creates an `AliasSeq` which repeats a type or an `AliasSeq` exactly `n` times.
  */
 template Repeat(size_t n, TList...)
-if (n > 0)
```
MetaLang: Please add a `n >= 0` template constraint.

9il: no need, size_t is unsigned

MetaLang: Whoops, missed that.

PR #5791, opened 2017-10-19
Make owns pure nothrow @safe @nogc

edi33416:
For allocators that implement their own `owns` method, such as `Region`, `owns` is a `pure nothrow @safe @nogc` function.

### DIFF
--- a/std/experimental/allocator/building_blocks/bitmapped_block.d
+++ b/std/experimental/allocator/building_blocks/bitmapped_block.d
@@ -350,12 +350,12 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
     `Ternary.no` otherwise. Never returns `Ternary.unkown`. (This
     method is somewhat tolerant in that accepts an interior slice.)
     */
-    Ternary owns(void[] b) const
+    pure nothrow @trusted @nogc
+    Ternary owns(const void[] b) const
     {
-        //if (!b.ptr) return Ternary.no;
-        assert(b.ptr !is null || b.length == 0, "Corrupt block.");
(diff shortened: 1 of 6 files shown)


edi33416:

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -350,12 +350,12 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
...
-    Ternary owns(void[] b) const
+    pure nothrow @trusted @nogc
+    Ternary owns(const void[] b) const
     {
-        //if (!b.ptr) return Ternary.no;
-        assert(b.ptr !is null || b.length == 0, "Corrupt block.");
-        return Ternary(b.ptr >= _payload.ptr
-            && b.ptr + b.length <= _payload.ptr + _payload.length);
+        assert(b || b.length == 0, "Corrupt block.");
+        return Ternary(b && _payload && (&b[0] >= &_payload[0])
```
andralex: If you made it trusted you can keep the previous implementation, which is marginally better on the eyes.

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -843,6 +843,19 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
     assert(h3.totalAllocation(64 * 4096 + 1) >= 65 * 4096);
 }
 
+// Test owns
+@system unittest
```
andralex: I assume this is not safe because of some other PRs that need to complete the feature. Please don't forget to make a pass later to migrate all attributes into the unittest.

PR #5792, opened 2017-10-19
Fix Issue 9661 - LockingTextWriter should increment file handle reference count

jercaianu:
I also did the same for LockingBinaryWriter and added a test case for each.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2717,7 +2717,10 @@ $(D Range) that locks the file and allows fast writing to it.
     {
     private:
         import std.range.primitives : ElementType, isInfinite, isInputRange;
-        // the shared file handle
+
+        // We need the 'file_' member to keep the object alive by refcounting.
+        // The 'fps_' member is also kept to avoid constant checking if 'file_'
+        // members are correctly initialized, when trying to get the 'FILE*' from the 'File' object
         FILE* fps_;
         File file_;
 
@@ -2898,6 +2901,10 @@ See $(LREF byChunk) for an example.
     {
         import std.traits : hasIndirections;
     private:
+
+        // We need the 'file_' member to keep the object alive by refcounting.
+        // The 'fps_' member is also kept to avoid constant checking if 'file_'
+        // members are correctly initialized, when trying to get the 'FILE*' from the 'File' object
         FILE* fps;
         File file;
         string name;
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2719,6 +2719,7 @@ $(D Range) that locks the file and allows fast writing to it.
         import std.range.primitives : ElementType, isInfinite, isInputRange;
         // the shared file handle
         FILE* fps_;
+        File file_;
 
         // the unshared version of fps
         @property _iobuf* handle_() @trusted { return cast(_iobuf*) fps_; }
@@ -2733,6 +2734,7 @@ $(D Range) that locks the file and allows fast writing to it.
             import std.exception : enforce;
 
             enforce(f._p && f._p.handle, "Attempting to write to closed File");
+            file_ = f;
             fps_ = f._p.handle;
             orientation_ = fwide(fps_, 0);
             FLOCK(fps_);
@@ -2897,6 +2899,7 @@ See $(LREF byChunk) for an example.
         import std.traits : hasIndirections;
     private:
         FILE* fps;
+        File file;
         string name;
 
         version (Windows)
@@ -2910,7 +2913,7 @@ See $(LREF byChunk) for an example.
         this(ref File f)
         {
             import std.exception : enforce;
-
+            file = f;
             enforce(f._p && f._p.handle);
             name = f._name;
             fps = f._p.handle;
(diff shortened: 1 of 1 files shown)


CyberShadow:
Any reason not to turn the `FILE *fps` fields into properties, now that they can be reached through the added `File` field?

std/stdio.d
```diff
@@ -2897,6 +2899,7 @@ See $(LREF byChunk) for an example.
         import std.traits : hasIndirections;
     private:
         FILE* fps;
```
DmitryOlshansky: Do we still need fps when we have file?

jercaianu:
@CyberShadow 
I don't know if it makes sense, but due to extra dereferences caused by the Impl* inside the File struct, I think it could be slower.
Maybe if we change the Impl to actually be embedded inside the File struct, we could do this.

jercaianu:
@CyberShadow @DmitryOlshansky 
Hi, is it alright now?

Thanks,
Alex

jercaianu:
@CyberShadow @DmitryOlshansky 
Trying to remove the FILE* member I have encountered several issues.
When accessing the FILE* through the File object, we have to add boilerplate code to check if File is correctly initialized (_p member is not null). And this has to be done every time we are trying to access the FILE*.

I think it's better to keep it this way, with the FILE* member also.
Thanks,
Alex

CyberShadow:
> I think it's better to keep it this way, with the FILE* member also.

Sounds reasonable; feel free to leave this for another day. Maybe add this as a comment.

jercaianu:
@CyberShadow 
Where should I add the comment, in code or commit message?

schveiguy:
> And this has to be done every time we are trying to access the FILE*

First, the extra indirection is a bit more, but there's no reason we need to check to see if the File's implementation is `null`, as that is checked on entry, and it won't be null as long as the ref count is not 0. We know it's not 0 because we still have a reference to it! Or am I mistaken?

jercaianu:
@schveiguy
In case the object is created using the constructor provided, then there is no need for checks.
The LockingTextWriter can however be created using .init, just as in the test case at line 3729, which would lead to an invalid File object.

schveiguy:
Sure, but what are the consequences of that? I mean, it just dereferences a null pointer and crashes. I don't know what someone would expect to happen if you created a `LockingTextReader` that you can never properly use (the `_file` member and `_fps` member are private so you can't set them).

Note, the test you reference still would pass, because it doesn't actually use the LTR.

jercaianu:
@schveiguy 
The test would fail if no checks are added in the destructor, for example, when accessing the FILE* through the File object.

The way I see it, we have two options.
Either we assume that someone may want to use .init (like the test i mentioned) and keep it the way it is, or we assume that .init for LTW is invalid, remove that test and keep only the File object and access the FILE* though the 'file' member.

schveiguy:
> The test would fail if no checks are added in the destructor

Checking in the destructor is OK. That doesn't mean we have to check it everywhere. Note, we are already checking it in the destructor, so it won't be any worse.

> we assume that .init for LTW is invalid

the .init value of a struct by default, must be destructable. Otherwise, you can't compose  It doesn't have to be *usable*.

I don't think the following use case makes any sense to do anything but crash or throw:

```D
LockingTextWriter ltw;
ltw.put('c');
```

We can safely assume that this isn't going to happen. What we can do is `assert` the pointer in all uses *except* the destructor, which should remove the checks in release mode, but throw an `Error` in debug mode.

jercaianu:
@schveiguy 
What do you think about my latest commit?
Is it alright?

Thanks!

std/stdio.d
```diff
@@ -2896,7 +2911,9 @@ See $(LREF byChunk) for an example.
     {
         import std.traits : hasIndirections;
     private:
-        FILE* fps;
+        // Access the FILE* handle through the 'file_' member
+        // to keep the object alive through refcounting
+        File file;
```
schveiguy: I'd say you should be consistent with LockingTextWriter and use `file_` here instead of `file`, but either way, the comment doesn't match the member.

std/stdio.d
```diff
@@ -3271,8 +3314,7 @@ void main()
     scope(exit) std.file.remove(deleteme);
 
     {
-        File f = File(deleteme, "w");
-        auto writer = f.lockingTextWriter();
+        auto writer = File(deleteme, "w").lockingTextWriter();
```
schveiguy: nice!

std/stdio.d
jercaianu: Ok, thanks for the comments!
Fixing :)

std/stdio.d
```diff
@@ -2733,25 +2734,39 @@ $(D Range) that locks the file and allows fast writing to it.
...
+                fps = file_._p.handle;
+            }
+
+            if (fps)
             {
-                FUNLOCK(fps_);
-                fps_ = null;
+                FUNLOCK(fps);
             }
         }
```
andralex:
```d
if (auto p = file_._p)
{
    if (p && p.handle) FUNLOCK(p.handle);
}
```

Please simplify similar snippets accodringly

schveiguy:
Restarted Circle CI, seems to have segfaulted.

PR #5793, opened 2017-10-20
fix math.d for 16997

WalterBright:
Blocking https://github.com/dlang/dmd/pull/7013

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -4518,7 +4518,7 @@ real round(real x) @trusted nothrow @nogc
     {
         auto old = FloatingPointControl.getControlState();
         FloatingPointControl.setControlState(
-            (old & ~FloatingPointControl.ROUNDING_MASK) | FloatingPointControl.roundToZero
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -4518,7 +4518,7 @@ real round(real x) @trusted nothrow @nogc
     {
         auto old = FloatingPointControl.getControlState();
         FloatingPointControl.setControlState(
-            (old & ~FloatingPointControl.ROUNDING_MASK) | FloatingPointControl.roundToZero
+            (old & (-1-FloatingPointControl.ROUNDING_MASK)) | FloatingPointControl.roundToZero
```
ibuclaw: Why not a cast instead?

WalterBright: I like this better, as it will invoke the compiler's normal integral promotion. I try to avoid casts if I can.

andralex: spaces around binary `-`...

PR #5794, opened 2017-10-20
Fix Issue 17919: Make std.container.Array use pureMalloc

n8sh:
Using pureMalloc like std.typecons.RefCounted does would make it
possible to use std.container.Array in  `pure` code.


andralex:
good idea, thx

PR #5796, opened 2017-10-21
Documentation improvements for std.algorithm.splitter.

jondegenhardt:
This is an attempt to improve documentation for `std.algorithm.splitter`. This was done mostly by changes to the examples. Partly by starting with a more obvious example and adding examples of different results cases. Also by putting the example with adjacent spaces in a context where it was clearer that there are adjacent spaces rather than a single space when reading on the web.

It turns out there is a bug in the documentation as well, as the overload taking a range as a separator indicates a bidirectional range will be returned. This is incorrect. See this [learn forum answer](https://forum.dlang.org/post/mailman.234.1506810353.3410.digitalmars-d-learn@puremagic.com) by @jmdavis. This was corrected as well.

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -3855,14 +3855,14 @@ if (is(typeof(binaryFun!pred(r.front, s)) : bool)
     import std.range : empty, retro;
 
     // Basic splitting with characters and numbers.
-    assert(equal(splitter("a.bc.def", '.'), [ "a", "bc", "def" ]));
+    assert(equal(splitter("a|bc|def", '|'), [ "a", "bc", "def" ]));
 
     int[] a = [1, 0, 2, 3, 0, 4, 5, 6];
     int[][] w = [ [1], [2, 3], [4, 5, 6] ];
     assert(equal(splitter(a, 0), w));
 
     // Adjacent separators.
-    assert(equal(splitter("a.b..c", '.'), [ "a", "b", "", "c" ]));
+    assert(equal(splitter("a|b||c", '|'), [ "a", "b", "", "c" ]));
     assert(equal(splitter("hello  world", ' '), [ "hello", "", "world" ]));
 
     a = [ 1, 2, 0, 0, 3, 0, 4, 5, 0 ];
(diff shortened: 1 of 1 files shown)


andralex:
please babysit the tests a little

jondegenhardt:
Thanks for the review, especially given the large backlog currently. 

> please babysit the tests a little

You mean the failed circleci test? I apologize, I'm not sure what the appropriate action was for me to take here. I saw it was failing due to the dscanner seg fault, same as on pretty much all PRs. What was best action for me to take?

andralex:
cc @MartinNowak

jmdavis:
> You mean the failed circleci test? I apologize, I'm not sure what the appropriate action was for me to take here. I saw it was failing due to the dscanner seg fault, same as on pretty much all PRs. What was best action for me to take?

That seems to be happening randomly from what I've seen, so I think that dscanner has a bug that we're hitting some of the time but not consistently.

> Update: Maybe I could have rebased the PR, once newer PRs started passing the circleci test.

Doing `git commit --amend` without changing the commit message should allow you to then force push without changing anything (since that changes the commit ID), and all of the tests will run again, in which case, dscanner might work this time, since the failure does not appear to be consistent.

jondegenhardt:
> Doing git commit --amend without changing the commit message should allow you to then force push without changing anything (since that changes the commit ID), and all of the tests will run again, in which case, dscanner might work this time, since the failure does not appear to be consistent.

Thanks. I assume I need to do this now to trigger the tests and allow the merge to occur. Is that is correct?

jondegenhardt:
fyi: On the CircleCI page for the test there is a "Rebuild" button, which looks like it might re-run the test. However, it indicates write access to the repo is required to use it. I don't have write access, but it might be an option for people who do.

jmdavis:
> Thanks. I assume I need to do this now to trigger the tests and allow the merge to occur. Is that is correct?

If you do that, it will automatically retrigger the tests, since it's effectively a new commit. The auto-merge label will then have to be re-added (since it gets removed whenever changes are pushed to a PR so that nothing sneaks in under the radar), but if the tests pass on the next run, and the auto-merge label is re-added, then the merge should happen.

jondegenhardt:
Well... CircleCI finally ran without dscanner seg faulting. Then it found what it identified as a style error. But it's really a bug in the style checker. It ran the check on the literal strings in the examples, as if the literal string was code. Here is the failed test:
```
Enforce space between a .. b
grep -nrE '[[:alnum:]][.][.][[:alnum:]]|[[:alnum:]] [.][.][[:alnum:]]|[[:alnum:]][.][.] [[:alnum:]]' $(find etc std -name '*.d' | grep -vE 'std/string.d|std/uni.d') ; test $? -eq 1
std/algorithm/iteration.d:3865:    assert(equal(splitter("a.b..c", '.'), [ "a", "b", "", "c" ]));
std/algorithm/iteration.d:4122:    assert(equal(splitter("a.b..c", ".."), [ "a.b", "c" ]));
make: *** [style_lint] Error 1
```
Can't have something in the code-base that triggers this failure, so I'll change the examples to use a different character. It'll be few days before I can get back to this.

jondegenhardt:
FWIW - When submitting this originally I tried to run the style checkers locally. The [Starting as a Contributer](https://wiki.dlang.org/Starting_as_a_Contributor#Run_CI_checks) wiki lists two steps:
```
make -j16 -f posix.mak checkwhitespace
make -f posix.mak style
```
The first succeeded. The second ran a number of checks successfully, then tried to build `dscanner`. Building `dscanner` failed. Looked like the link paths on the build command were wrong, but I didn't take the time to figure out what the correct setup was and moved on.

jondegenhardt:
Switched from periods to vertical bars to avoid the false positive `Enforce space between a .. b` style check. Took five submissions, but finally all tests passed on the same run. Should be good now.

PR #5797, opened 2017-10-22
Issue 17269: formattedWrite of struct with Nullable value fails

CromFr:
The bug was caused by formatElement [implicitly converting](https://github.com/dlang/phobos/blob/dc07f1ae3fe1a5b1fbf8a95f39861ce87964ec58/std/typecons.d#L3136) Nullable!string to string, that resulted in [an error](https://github.com/dlang/phobos/blob/dc07f1ae3fe1a5b1fbf8a95f39861ce87964ec58/std/typecons.d#L3114) when the Nullable was null.

New behavior is to use the toString method when available, instead of implicitly converting the value.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -3830,6 +3830,20 @@ if (is(AssocArrayTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     assert(w.data == "TestContainer(helloworld)", w.data);
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=17269
+@safe unittest
+{
+    import std.typecons : Nullable;
+
+    struct Foo
+    {
+        Nullable!string bar;
+    }
+
+    Foo f;
+    formatTest(f, "Foo(Nullable.null)");
+}
+
 @safe unittest
 {
     assert(collectExceptionMsg!FormatException(format("%d", [0:1])).back == 'd');
--- a/std/format.d
+++ b/std/format.d
@@ -3638,7 +3638,7 @@ private void formatChar(Writer)(ref Writer w, in dchar c, in char quote)
 // undocumented because of deprecation
 // string elements are formatted like UTF-8 string literals.
 void formatElement(Writer, T, Char)(auto ref Writer w, T val, scope const ref FormatSpec!Char f)
-if (is(StringTypeOf!T) && !is(T == enum))
+if (is(StringTypeOf!T) && !hasToString!(T, Char) && !is(T == enum))
 {
     import std.array : appender;
     import std.utf : decode, UTFException;
@@ -3749,7 +3749,7 @@ if (is(CharTypeOf!T) && !is(T == enum))
 // undocumented
 // Maybe T is noncopyable struct, so receive it by 'auto ref'.
 void formatElement(Writer, T, Char)(auto ref Writer w, auto ref T val, scope const ref FormatSpec!Char f)
-if (!is(StringTypeOf!T) && !is(CharTypeOf!T) || is(T == enum))
+if ((!is(StringTypeOf!T) || hasToString!(T, Char)) && !is(CharTypeOf!T) || is(T == enum))
 {
     formatValue(w, val, f);
 }
@@ -3805,6 +3805,29 @@ if (is(AssocArrayTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     }
     if (spec == 's')
         put(w, f.seqAfter);
+}
+
+@safe unittest
+{
+    // Bug #17269. Behavior similar to `struct A { Nullable!string B; }`
+    struct StringAliasThis
+    {
+        @property string value() const { assert(0); }
+        alias value this;
+        string toString() { return "helloworld"; }
+        private string _value;
+    }
+    struct TestContainer
+    {
+        StringAliasThis testVar;
+    }
+
+    import std.array : appender;
+    auto w = appender!string();
+    auto spec = singleSpec("%s");
+    formatElement(w, TestContainer(), spec);
+
+    assert(w.data == "TestContainer(helloworld)", w.data);
 }
 
 @safe unittest


std/format.d
```diff
@@ -3251,6 +3251,29 @@ if (is(AssocArrayTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
 
 @safe unittest
 {
+    // Bug #17269. Behavior similar to `struct A { Nullable!string B; }`
```
atilaneves: Why not use a struct with `Nullable!string`?

MetaLang:
Urgh, it's all `StringTypeOf`'s fault. That template is way too permissive.

Looks like `formatValue` uses this same method so we might as well keep it consistent. Sorry for the noise.

std/format.d
CromFr: If I use a Nullable!string, it causes an [assert failure on another unittest block](https://github.com/dlang/phobos/blob/master/std/format.d#L4035)
```
expected = `void delegate() @system`, result = `void delegate()`
```

I don't really understand why, so I wrote a custom struct to:
- Show the `alias this` and the implicit conversion problem
- Do not depend on Nullable if it is changed in the future

CromFr:
fyi the first fix I thought of was to modify `StringTypeOf` to return false if the struct is implicitly convertible to string using an `alias this` to a function that can throw.
But since Nullable does an assert, it wouldn't work (and modifying `StringTypeOf` has too many consequences)

std/format.d
wilzbach: Huh? `Nullable!string` works without your PR: https://run.dlang.io/is/Jjzpvo

wilzbach: Could you provide a test case where it fails and maybe open another bug report?

CromFr: you need to put the Nullable!string inside a struct to see it fail:
```d
struct NullableStr {
   Nullable!string value;
}
NullableStr f;
format("%s", f).writeln;
```

wilzbach:
Reworded the commit message, s.t. the dlang bot recognizes it and it can be auto-closed and part of the changelog (and also rebased to upstream/master).

wilzbach:
I added the initial example of as an additional `unittest` example - this is considered a good practice.
Anyhow now we are failing due to:

```
core.exception.AssertError@std/format.d(4097): expected = `void delegate() @system`, result = `void delegate()`
```

CromFr:
I encountered the same unittest error when initially trying to fix formattedWrite

Since it didn't make much sense to me, I changed the test to use a custom struct that has the same problem as Nullable!string and the unittest error disappeared, so I concluded that it was because of a weird import conflict or something like that, without much more investigations

quickfur:
Rebased. Hopefully it will go through this time.

quickfur:
Hmph. Seems that avoiding `Nullable` doesn't fix this problem. :-(  Gotta get to the bottom of that `@system delegate` issue.

quickfur:
I think there's a compiler bug associated with this. Moving the offending unittest above the unittest that imports `Nullable` fixes the problem. Unfortunately, it's a rather elusive one, as adding `version(none)` to preceding unittests makes the problem go away. Very strange.

CromFr:
I think dlang-bot should add a message when the PR becomes un-mergeable. I didn't receive any notification on this :/

wilzbach:
That would be quite spammy: https://github.com/dlang-bots/dlang-bot/issues/123

Note that if the PR gets traction or approval, we as maintainers can easily rebase this.

quickfur:
@CromFr BTW, I already rebased your pull, you may want to fetch from dlang/phobos to update your local branch if you're planning to adding more commits.

quickfur:
Managed to reduce the bug to a minimal(-ish) case: https://issues.dlang.org/show_bug.cgi?id=18269

A truly bizarre compiler bug where unrelated declarations can affect the type of a delegate!

MetaLang:
That's a nasty one. Thanks for the reduction @quickfur.

Anyone have a suggestion for how to proceed? I don't want this to fall by the wayside because of an obscure compiler bug.

quickfur:
One way to deal with this, if it's deemed urgent to merge this PR sooner rather than later, is to `version` out the unittest with a comment indicating that it's being blocked by issue 18269. That way, once 18269 is fixed we can turn on the unittest again.

wilzbach:
How about moving the unittest to a different module?

quickfur:
On Tue, Jan 23, 2018 at 07:11:46PM +0000, Sebastian Wilzbach wrote:
>    How about moving the unittest to a different module?
[...]

I suppose we could try that too. No idea whether it will work, though.
The compiler bug seems highly sensitive and fragile to small changes.
I'd like to dig to the bottom of it sometime, but it's deep in the
complicated and messy IFTI code, and I'm not sure I'll be able to figure
it out!


--T

quickfur:
In the meantime, perhaps we could change the order of the offending unittests so as the bypass the compiler bug?  I'm not sure when the compiler bug will be fixed; it's a pretty tricky one involving an ambiguity in the global stringtable.

wilzbach:
Yeah that's why I thought that putting it in a separate module will help as it reduces the chances of someone accidentally running into this in the near future again.

quickfur:
Unfortunately, the problem is cross-module. The original problem was caused by the import of `std.variant`, which in turn contains a template argument containing `void delegate()`. Basically, there's some kind of ambiguity in the compiler's stringtable, where both `void delegate()` and `void delegate() @system` have the same deco in their mangling, and whichever gets seen by the compiler first gets into the stringtable, and from that point on becomes the "official" string representation of that type. So no, putting it in a different module won't help as long as the order of unittests (as seen by the compiler) is wrong. See: https://issues.dlang.org/show_bug.cgi?id=18269

berni44:
@RazvanN7 With #7771 fixing the problem, after fixing the merge conflict (formatspec is meanwhile scope), IMHO this can be merged now.

RazvanN7 (approved):

RazvanN7:
@CromFr Please rebase this and we will get it in.

CromFr:
Should be good now :)

RazvanN7 (approved):

berni44:
@RazvanN7 circleci seems not to respond... Is there anything, that can be done?

RazvanN7:
Superweird. I can't even access the details page.

RazvanN7:
@berni44 I just restarted it.

PR #5798, opened 2017-10-23
fix Issue 17730 - move escapes scope variable in @safe code

MartinNowak:
- needs to be annotated with return scope, so that the return value
  lifetime depends on the argument's lifetime
- cannot be tested because phobos doesn't yet work with DIP1000 and
  also because of Issue 17932


MartinNowak:
Apparently dscanner dies because of return scope :/.

PR #5799, opened 2017-10-23
Make empty @safe

edi33416:
For allocators that implement their own `empty` method, such as `KRRegion`, `empty` is a `pure nothrow @safe @nogc` function.

In this PR, there are also a couple of bug-fixes:
- `FallbackAllocator` should attempt to test only if it has supplied at least one stateless allocator
- `Region` didn't take into consideration the `growDownwards` flag
- `Segregator` should use the `Ternary` op `&` instead of `&&`

### DIFF
--- a/std/experimental/allocator/building_blocks/affix_allocator.d
+++ b/std/experimental/allocator/building_blocks/affix_allocator.d
@@ -402,6 +402,20 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
     });
 }
 
+// Test empty
+@system unittest
+{
+    import std.experimental.allocator.building_blocks.bitmapped_block : BitmappedBlock;
+    import std.typecons : Ternary;
+
+    auto a = AffixAllocator!(BitmappedBlock!128, ulong, ulong)
+                (BitmappedBlock!128(new ubyte[128 * 4096]));
+    assert((() pure nothrow @safe @nogc => a.empty)() == Ternary.yes);
+    auto b = a.allocate(42);
+    assert(b.length == 42);
+    assert((() pure nothrow @safe @nogc => a.empty)() == Ternary.no);
+}
+
 @system unittest
 {
     import std.experimental.allocator.mallocator : Mallocator;
--- a/std/experimental/allocator/building_blocks/allocator_list.d
+++ b/std/experimental/allocator/building_blocks/allocator_list.d
@@ -503,6 +503,7 @@ struct AllocatorList(Factory, BookkeepingAllocator = GCAllocator)
      Returns `Ternary.yes` if no allocators are currently active,
     `Ternary.no` otherwise. This methods never returns `Ternary.unknown`.
     */
+    pure nothrow @safe @nogc
     Ternary empty() const
     {
         return Ternary(!allocators.length);
@@ -615,9 +616,10 @@ version(Posix) @system unittest
     b1 = a.allocate(1024 * 10);
     assert(b1.length == 1024 * 10);
     a.allocate(1024 * 4095);
+    assert((() pure nothrow @safe @nogc => a.empty)() == Ternary.no);
     // Ensure deallocateAll infers from parent
     assert((() nothrow @nogc => a.deallocateAll())());
-    assert(a.empty == Ternary.yes);
+    assert((() pure nothrow @safe @nogc => a.empty)() == Ternary.yes);
 }
 
 @system unittest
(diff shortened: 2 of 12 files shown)


edi33416:
@andralex @wilzbach @ZombineDev Could you have a look?

std/experimental/allocator/building_blocks/fallback_allocator.d
```diff
@@ -247,7 +249,10 @@ struct FallbackAllocator(Primary, Fallback)
 
     Returns: $(D primary.empty & fallback.empty)
     */
-    static if (hasMember!(Primary, "empty") && hasMember!(Fallback, "empty"))
+    static if (hasMember!(Primary, "empty")
+               && is(typeof(Ternary(primary.empty)))
```
andralex: Why this? We assume that if someone defines "empty" it must return `bool`. True, that assumption is not verified but we will insert all verifications later in a single place. So no need to attach tests to each constraint.

std/experimental/allocator/building_blocks/fallback_allocator.d
```diff
@@ -247,7 +249,10 @@ struct FallbackAllocator(Primary, Fallback)
...
+               && hasMember!(Fallback, "empty")
+               && is(typeof(Ternary(fallback.empty))))
```
andralex: remove

std/experimental/allocator/building_blocks/segregator.d
```diff
@@ -239,10 +239,12 @@ struct Segregator(size_t threshold, SmallAllocator, LargeAllocator)
         }
 
         static if (hasMember!(SmallAllocator, "empty")
-                && hasMember!(LargeAllocator, "empty"))
+                && is(typeof(Ternary(_small.empty)))
```
andralex: remove

std/experimental/allocator/building_blocks/segregator.d
```diff
@@ -239,10 +239,12 @@ struct Segregator(size_t threshold, SmallAllocator, LargeAllocator)
...
+                && hasMember!(LargeAllocator, "empty")
+                && is(typeof(Ternary(_large.empty))))
```
andralex: remove

andralex:
@edi33416 rebase

andralex:
Tried to resolve conflicts by using the web editor...

std/experimental/allocator/building_blocks/region.d
```diff
@@ -356,9 +363,12 @@ struct Region(ParentAllocator = NullAllocator,
     // Create a 64 KB region allocated with malloc
     auto reg = Region!(Mallocator, Mallocator.alignment,
         Yes.growDownwards)(1024 * 64);
+
+    assert((() pure nothrow @safe @nogc => reg.empty)() ==  Ternary.yes);
```
MetaLang: Since the unit test is explicitly marked as `@system`, it would probably have been better to put these in their own unit test block and mark it `pure nothrow @safe @nogc`.

std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -402,6 +402,20 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
...
 
+// Test empty
+@system unittest
+{
+    import std.experimental.allocator.building_blocks.bitmapped_block : BitmappedBlock;
+    import std.typecons : Ternary;
+
+    auto a = AffixAllocator!(BitmappedBlock!128, ulong, ulong)
+                (BitmappedBlock!128(new ubyte[128 * 4096]));
+    assert((() pure nothrow @safe @nogc => a.empty)() == Ternary.yes);
```
MetaLang: Since the unit test is explicitly marked as @system, it would probably have been better to put these in their own unit test block and mark it pure nothrow @safe @nogc.

std/experimental/allocator/building_blocks/allocator_list.d
```diff
@@ -614,8 +615,9 @@ version(Posix) @system unittest
     b1 = a.allocate(1024 * 10);
     assert(b1.length == 1024 * 10);
     a.allocate(1024 * 4095);
+    assert((() pure nothrow @safe @nogc => a.empty)() == Ternary.no);
```
MetaLang: ditto

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -707,6 +708,17 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
 
 @system unittest
 {
+    import std.typecons : Ternary;
+
+    auto a = BitmappedBlock!(64, 64)(new ubyte[10_240]);
+    assert((() nothrow @safe @nogc => a.empty)() == Ternary.yes);
+    const b = a.allocate(100);
+    assert(b.length == 100);
+    assert((() nothrow @safe @nogc => a.empty)() == Ternary.no);
```
MetaLang: ditto (you get the idea)

andralex:
@MetaLang we'll make one pass to clean this up once we add attributes to all primitives.

PR #5183, opened 2017-02-23
Markdownify code syntax from the old $(D code) macros to Markdown (i.e. `code`) 

wilzbach:
Gets Phobos one step closer to pretty code ;-)

While the detection is easy with a stupid RegExp, I needed to [write a small lexer](https://github.com/wilzbach/dscanner-fix/blob/master/d_code_markdown.d) as it's context-sensitive:

```
$(D foo) -> `foo`
$(D writeln("foo")) -> `writeln("foo` /// <- with a naive RegExp
```

Ideally DAutoTest shouldn't show any changes.

PR #2877, opened 2015-01-16
$(D code) -> `code`

CyberShadow:
http://forum.dlang.org/post/m9btie$pcl$1@digitalmars.com

PR #5801, opened 2017-10-24
Change single-word uses of the D macro in documentation with backticks

andralex:
This replaces uses of `$(D ...)` with backticks for single-word instances. This makes the doc easier on the eyes and easier to edit. To effect this I ran:

```bash
sed -i '' -e 's/\$(D \([^ )]*\))/`\1`/g' std/**/*.d
```


andralex:
cc @wilzbach can we also make this a standard check for future code?

std/random.d
```diff
@@ -1367,22 +1367,22 @@ A singleton instance of the default random number generator
 }
 
 /**
-Generates a number between $(D a) and $(D b). The $(D boundaries)
+Generates a number between `a` and `b`. The `boundaries`
 parameter controls the shape of the interval (open vs. closed on
-either side). Valid values for $(D boundaries) are $(D "[]"), $(D
-"$(LPAREN)]"), $(D "[$(RPAREN)"), and $(D "()"). The default interval
+either side). Valid values for `boundaries` are `"[]"`, $(D
+"$(LPAREN)]"), `"[$(RPAREN`"), and `"(`"). The default interval
```
aG0aep6G: Bad substitutions here.

andralex:
Changed the sed expression to:

```bash
sed -i '' -e 's/\$(D \([^ ()]*\))/`\1`/g' std/**/*.d
```

andralex:
What the heck did I do? Ugh...

andralex:
OK here we go...

andralex:
This should be merged soon, it's a bonanza of rebase conflicts...

andralex:
cool thx

ZombineDev:
@andralex the ddoc build passed, so you can verify if there's anything unexpected: http://dtest.dlang.io/results/d970593fffa023018370dde1852998cd6bd83ff2/f8d3d9b2ade3031dc671481be922425e8577263a/. (I'll be able to look at it in 30-40 mins)

andralex:
Seeing artifacts that shouldn't be there, removed auto-merge for now to investigate.

CyberShadow:
Here is the previous attempt, may contain useful discussion: https://github.com/dlang/phobos/pull/2877

andralex:
@CyberShadow thanks. This particular PR is more conservative so it's mainly exposed to the problem of double highlighting of focal words (those that people get rid of by using a leading underscore).

That problem is benign - it applies an anchor and a span style twice, so it increases file size but does not affect functionality. I suggest we pull this and create an issue for the backticks feature.

std/base64.d
```diff
@@ -53,7 +53,7 @@
  * Authors:   Masahiro Nakagawa, Daniel Murphy (Single value Encoder and Decoder)
  * Source:    $(PHOBOSSRC std/_base64.d)
  * Macros:
- *      LREF2=<a href="#$1">$(D $2)</a>
+ *      LREF2=<a href="#$1">`$2`</a>
```
ZombineDev: I think this is one of the culprits - see [here](http://dtest.dlang.io/diff/website-7a608bf8457130bb33dae89311a1dcdebf01bc3b-3bcdcebf6c39ba84e8fd3f5faa3064a7/website-7a608bf8457130bb33dae89311a1dcdebf01bc3b-74648b88f0af419c3de1b2b2d2d6e5c6/web/phobos-prerelease/std_base64.html):
```diff
-  <a href="#.Base64Impl.Encoder.empty"><span class="d_inlinecode donthyphenate notranslate">empty</span></a> returns <span class="d_inlinecode donthyphenate notranslate"><code class="ddoc_keyword">true</code></span>.</div></div>
+  <a href="#.Base64Impl.Encoder.empty">`empty`</a> returns <span class="d_inlinecode donthyphenate notranslate">true</span>.</div></div>
```

andralex:
OK @CyberShadow warned that there are additional `_` inserted in the generated HTML, which makes this PR not viable without fixing the ddoc bug.

andralex:
@ZombineDev good insight, I changed the macro to see how things go. BTW we already have that macro, right? No need for it.

wilzbach:
> Here is the previous attempt, may contain useful discussion: #2877

FWIW I also tried this before and failed: :cry: 
https://github.com/dlang/phobos/pull/5183

> which makes this PR not viable without fixing the ddoc bug.

Well you do see all places where a `_` needs to be inserted on DAutoTest.

@andralex let me know if I should adopt this PR and bring it to the finish line. After all, this is now at least the third attempt...

andralex:
@wilzbach yah, that would be nice but do it only if technically easy - it's not high impact. Thx!

quickfur:
We really *really* need to kill that autoexpansion habit of ddoc that leads to `_` proliferation. It's clearly causing far more trouble than it's worth, and for far too long already.

wilzbach:
> @wilzbach yah, that would be nice but do it only if technically easy - it's not high impact. Thx!

Yeah, it's pretty trivial to do, but I just did it for `std.algorithm` as a test:
https://github.com/dlang/phobos/pull/5970

PR #5803, opened 2017-10-24
Ternary opBinary should be able to take bool

edi33416:
I believe that logical operations between `Ternary`s and `bool`s should be valid.


andralex:
please also add a changelog entry

PR #5804, opened 2017-10-24
prevent a few implicit truncating FP conversions

BBasile:
Prevent warnings emitted after a possible compiler change.

### DIFF
--- a/std/numeric.d
+++ b/std/numeric.d
@@ -3034,7 +3034,7 @@ private:
             else if (i == size * 3 / 4)
                 lastRow[i] = 1;  // -sin(pi * 3 / 2) == 1
             else
-                lastRow[i] = -sin(i * 2.0L * PI / size);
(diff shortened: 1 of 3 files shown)


std/numeric.d
```diff
@@ -3034,7 +3034,7 @@ private:
             else if (i == size * 3 / 4)
                 lastRow[i] = 1;  // -sin(pi * 3 / 2) == 1
             else
-                lastRow[i] = -sin(i * 2.0L * PI / size);
+                lastRow[i] = -sin(i * cast(float)(2.0 * PI) / size);
```
andralex: use `float(2.0 * PI)` here instead

BBasile:
Improvements are not the point of the PR which is rather to remove the warnings emitted in case https://github.com/dlang/dmd/pull/7240#partial-pull-merging would be accepted, but the test failures show that the change cause problems.

andralex:
I see, thanks. cc @WalterBright

PR #5805, opened 2017-10-24
Fix Issue 17742 - std.range.transposed does not have opIndex

Darredevil:
https://issues.dlang.org/show_bug.cgi?id=17742


std/range/package.d
```diff
@@ -6629,10 +6630,10 @@ if (isForwardRange!RangeOfRanges &&
         }
     }
 
-    // ElementType opIndex(size_t n)
-    // {
-    //     return _input[n].front;
-    // }
+    ElementType opIndex(size_t n)
```
andralex: Cool! Let's do this simpler implementation:

```d
auto ref opIndex(size_t n) { return _input[n]; }
```

Then you can remove the import and the alias.

andralex:
@wilzbach the style checker fails again with a segfault, could you please look into it? Thanks!

std/range/package.d
RazvanN7: I would throw const into the mix:
```
auto ref opIndex(size_t n) const { return _input[n]; }

```

andralex: const would unnecessarily limit what can be done and the ranges this could work with - it's more restrictive than C++'s.

ZombineDev: `inout` then?

schveiguy: `inout` has a problem here, because the `_input` type's `opIndex` must support `inout` in order for it to work properly. Otherwise, it would not work with a range that supplies only mutable `opIndex`.

Regardless, you need a static check around `opIndex` to make sure the underlying range supports it.

andralex:
Actually could you also please add slicing primitives to this diff? For examples git grep opSlice. Thanks!

std/range/package.d
```diff
@@ -6629,10 +6627,10 @@ if (isForwardRange!RangeOfRanges &&
         }
     }
 
-    // ElementType opIndex(size_t n)
-    // {
-    //     return _input[n].front;
-    // }
+    static if (isRandomAccessRange!RangeOfRanges)
+    {
+        auto ref opIndex(size_t n) { return _input[n]; }
```
schveiguy: I don't think this is right at all. For instance, I would expect this:

```D
auto x = [[1,2,3],[4,5,6]].transposed;
assert(x.front.equal([1,4])); // passes today
assert(x[0].equal([1,4])); // I don't think this would pass with your change.
```

std/range/package.d
```diff
@@ -6674,6 +6672,16 @@ private:
...
 }
 
+// Issue 17742
+@safe unittest
+{
+    import std.algorithm.iteration : map;
+    auto ror = 5.iota.map!(y => 5.iota.map!(x => x * y).array).array;
+    assert(ror[3][2] == 6);
+    auto result = ror.transposed;
+    assert(result[3][2] == 6);
```
schveiguy: This is actually a really bad test case. It's a matrix that is equal to itself transposed.

I'd rather see an array of array literal, and make sure you don't repeat elements so it's less likely you get a false positive:
```D
[
   [1, 2, 3, 4, 5],
   [6, 7, 8, 9, 10],
   ...
]
```

schveiguy: Also, your test should be of `result[2][3]`, as the ror is transposed, so the indexes should be reversed.

andralex:
ping @Darredevil

Darredevil:
@schveiguy thanks for pointing this out, however this leads to another issue. I can make opIndex work lazy in O(n) but I assume that is unacceptable. The other option would be to make Transposed eager all together. @andralex what do you think?

schveiguy:
The result of the opIndex could be something entirely lazy that obtains the opIndex value from each of the ranges.

For example:

```D
struct OpIndexedTransposed
{
   RangeOfRanges ror;
   immutable size_t idx;
   auto front() { return ror.front[idx]; }
   bool empty() { return ror.empty; }
   void popFront() { ror.popFront; }
}
```

You just have to check that `typeof(ror.front())` supports `opIndex`.

Darredevil:
@schveiguy I still don't see how your suggestion solved the O(n) issue. Edited my PR to make opIndex work correctly for Transpose although it's still in O(n).

std/range/package.d
```diff
@@ -6629,10 +6627,20 @@ if (isForwardRange!RangeOfRanges &&
...
-    // ElementType opIndex(size_t n)
-    // {
-    //     return _input[n].front;
-    // }
+    static if (isRandomAccessRange!RangeOfRanges)
+    {
+        auto ref opIndex(size_t n)
+        {
+            auto r = this.save;
+            r._input = _input.dup;
```
schveiguy: This isn't necessary, nor is it supported by forward ranges in general.

std/range/package.d
```diff
@@ -6629,10 +6627,20 @@ if (isForwardRange!RangeOfRanges &&
...
+            while(n)
+            {
+                r.popFront();
+                n--;
+            }
+            return r.front.array;
```
schveiguy: No, this is definitely not what we want. You only want to support indexing if the RangeOfRange's element type supports indexing. Then you "index on demand". All the accesses should be O(1).

std/range/package.d
Darredevil: If you eliminate that line and change the result of this.save it will affect the original instance as well, due to how `@property Transposed save()` is currently implemented.

std/range/package.d
Darredevil: Forgot the .array there, was testing something.

std/range/package.d
schveiguy: Not sure what you mean? if save doesn't make it so you can iterate independently of the original, then it's not implemented correctly. But you definitely should be able to change the underlying values via `opIndex`.

For example, this should work:
```D
int[][] x = [[1,2,3],[4,5,6]];
auto t = x.transposed;
x[0][1] = 10;
assert(t == [[1,4],[10,5],[3,6]]);
t[2][1] = 15;
assert(x == [[1,10,3],[4,5,15]]);
```

std/range/package.d
schveiguy: To be clear, that wasn't the reason for my critique, yes the array call was unnecessary, but I don't think the `popFront` loop is necessary either.

schveiguy:
Hm... just realized something. opIndex on the subsequent transposed range cannot work, since the front call uses `filter` to tell if each range is empty. Honestly, I'm not sure this is the expectation when the ranges are each of the same length.

So we may not be able to do everything that has been discussed here. I think that's why the auto tester is failing.

std/range/package.d
Darredevil: What I meant, without that line we have the following behaviour:
`
auto x = [[1,2,3],[4,5,6]].transposed;
writeln(x[1]); // prints [2,5] OK
writeln(x[1]); // prints [3,6] NOT OK
`

Which I assume is not ok?

schveiguy: Oh, definitely that's not OK. Let me think about this a bit, as the root cause isn't apparent to me. I don't think this is just a problem with `opIndex`.

schveiguy: Yeah, this is basically a problem of `save`:

```D
auto x = [[1,2,3],[4,5,6]];
auto y = x.save;
y.popFront;
writeln(x.front); // prints [2,5]
```
Clearly, save doesn't work here the way it's implemented, this is a bug in the original code.

schveiguy: https://issues.dlang.org/show_bug.cgi?id=17952

Darredevil:
@schveiguy @andralex implemented opIndex using Transversal
Also the style check is acting out again with a segfault.

std/range/package.d
```diff
@@ -6592,13 +6592,12 @@ Transversal!(RangeOfRanges, opt) transversal
     assert(mat1[1] == 10);
 }
 
-struct Transposed(RangeOfRanges)
+struct Transposed(RangeOfRanges,
+    TransverseOptions opt = TransverseOptions.assumeJagged)
```
schveiguy: Now you've done it :) We need to implement the assumption for all of transposed if we are going this route. Which gives us a lot of good options, nice idea!

std/range/package.d
```diff
@@ -6629,10 +6628,13 @@ if (isForwardRange!RangeOfRanges &&
         }
     }
 
-    // ElementType opIndex(size_t n)
-    // {
-    //     return _input[n].front;
-    // }
+    static if (isRandomAccessRange!RangeOfRanges)
```
schveiguy: This needs to test random access on the front of the ROR, not the actual ROR.

std/range/package.d
```diff
@@ -6674,16 +6676,37 @@ private:
...
+    auto result = transposed!(TransverseOptions.assumeNotJagged)(ror);
+    assert(result[2][3] == 6);
+
+    auto x = [[1,2,3],[4,5,6]];
+    auto y = transposed!(TransverseOptions.assumeNotJagged)(x);
+    assert(y.front.equal([1,4]));
+    assert(y[0].equal([1,4]));
+    assert(y[0][0] == 1);
+    assert(y[1].equal([2,5]));
+    assert(y[1][1] == 5);
```
schveiguy: Some more tests to ensure we get this right:

```D
import std.range.interfaces;
// limit input range to only a forward range
ForwardRange fr = inputRangeObject(x);

auto z = transposed!(TransverseOptions.assumeNotJagged)(fr);
assert(z.front.equal([1, 4]));
assert(z[0].equal([1,4]));
assert(z[1].equal([2, 5]));
assert(!is(typeof(z[0][0]))); // cannot index a forward range
```

std/range/package.d
schveiguy: BTW, I didn't even know about `transverse`, this is a perfect solution for it!

std/range/package.d
schveiguy: And once you implement the full gamut of `TransverseOptions`, those should be tested.

Darredevil: After spending more time with this I realized there is a problem, this fails:
```
static assert(hasAssignableElements!(typeof(fr)));
// aka
auto z = transposed!.... // will fail
```

Darredevil:
@andralex @schveiguy one final review please

std/range/package.d
schveiguy: Ugh ok. We will have to make a custom harness to test. Nothing’s ever easy ;)

schveiguy: Oops never mind I see you found an easy way!

schveiguy:
This looks pretty good. I think we can overhaul `Transposed` by using `transversal` for front as well, because it has a rich, already-implemented set of forwarding methods that is exactly what we need, and it already works with the `TransverseOptions`. But for now, this does solve the original bug. Feel free to experiment with adding the other options.

One thing that I think needs adding, but if you don't I can trivially add it later, is explaining what the `opt` compile-time option does in the ddocs.

std/range/package.d
```diff
@@ -6629,10 +6628,13 @@ if (isForwardRange!RangeOfRanges &&
...
 
-    // ElementType opIndex(size_t n)
-    // {
-    //     return _input[n].front;
-    // }
+    static if (isRandomAccessRange!(ElementType!RangeOfRanges))
+    {
+        auto ref opIndex(size_t n)
+        {
+            return transversal!(opt)(_input, n);
```
andralex: `transversal!opt`

Darredevil: Fixed.

PR #5806, opened 2017-10-24
Fix issue 17901 - FreeBSD SbrkRegion alignment

edi33416:
@andralex

### DIFF
--- a/std/experimental/allocator/building_blocks/region.d
+++ b/std/experimental/allocator/building_blocks/region.d
@@ -777,7 +777,7 @@ version(Posix) @system unittest
 version(Posix) @system unittest
 {
     import std.typecons : Ternary;
-    alias alloc = SbrkRegion!(8).instance;
+    alias alloc = SbrkRegion!().instance;
     auto a = alloc.alignedAllocate(2001, 4096);
     assert(a.length == 2001);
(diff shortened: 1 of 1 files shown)


jmdavis:
My local FreeBSD system (which passed all of the Phobos unit tests until recently) is now failing with

    core.exception.AssertError@std/experimental/allocator/building_blocks/region.d(660): Too large alignment chosen for shared(SbrkRegion!16u)

and the 64-bit FreeBSD autotester that Brad set up to test a newer version of FreeBSD is failing as well (though I'm not sure if that one is running FreeBSD 10.3 or 11, since IIRC, one of the two is running 10.3 and the other is running 11, but they just list their architectures, not the version of the OS):

https://auto-tester.puremagic.com/show-run.ghtml?projectid=15&runid=129654&dataid=898553

The 32-bit autotester for the newer FreeBSD is currently failing for other reasons, whereas it was previously failing with the message in the bug report. So, my guess is that 32-bit was fixed by this and 64-bit was broken in the process.

edi33416:
This is weird.. I'm going to set up 4 VMs (FBSD 10.3 and 11, 32 and 64 bit) and I'm going to look into this. I'm going to come with updates as soon as I have them.

PR #5807, opened 2017-10-24
Solaris: add import clock_gettime to currStdTime

nykytenko:
fix skipped "import clock_gettime"

### DIFF
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -220,7 +220,7 @@ public:
                     return unixTimeToStdTime(core.stdc.time.time(null));
                 else
                 {
-                    import core.sys.solaris.time : CLOCK_REALTIME;
+                    import core.sys.solaris.time : clock_gettime, CLOCK_REALTIME;
                     static if (clockType == ClockType.coarse)       alias clockArg = CLOCK_REALTIME;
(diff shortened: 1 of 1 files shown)


nykytenko:
Yes, but with LDC only.

DmitryOlshansky:
> std/typecons.d(2018:36)[warn]: Left side of logical or is identical to right side.

WAT?

I'm really impressed by poor quality of our mandatory lint check.
https://github.com/dlang/phobos/blob/master/std/typecons.d#L2018

jmdavis:
> I'm really impressed by poor quality of our mandatory lint check.
> https://github.com/dlang/phobos/blob/master/std/typecons.d#L2018

Yeah, not only does that have nothing to do with this PR, it's not even true.

@wilzbach Any idea what's going on here? Was dscanner updated to be more stupid?

ZombineDev:
The problem with this error is that I can't reproduce it locally (`make -f posix.mak style_lint`), even with the exact same git commit hash on several pull-requests. I'll try to investigate more later today.

jmdavis:
@nykytenko Can you just do a quick `git commit --amend` (you don't have to change anything, just run the command and save the message as-is so that the commit has a new ID), and then do `git push -f` to push the changes to your branch again so that all of the checks will be run again, and hopefully this time, the bizarre circleci failure won't happen, and the changes can be merged.

PR #5809, opened 2017-10-26
Remove gcstub.

skl131313:


ZombineDev:
@rainers this is in your area of expertise. Are you ok with removing gcstub from Druntime?

rainers:
I'm fine with removing it, but it's more @WalterBright s baby. See also https://issues.dlang.org/show_bug.cgi?id=17326

Jebbs:
Probably worth mentioning that the "manual" gc (`src/impl/manual/gc.d`) is the same implementation as gcstub, or it was when I added it.

gcstub probably should have been removed when it was added.

skl131313:
@ZombineDev 
Well if rainers is alright with it and @WalterBright doesn't have anything to say against it, can this be merged in along with the druntime pull request?

skl131313:
I think it's fine, it's been long enough and it's not like this is removing a working feature ; it's been broken for a year or more.

rainers:
Maybe a bit of information how to implement something similar to the gcstub (with a big red warning about its limitations) in the wiki would be nice, i.e. not linking in the regular GC, but allowing to proxy it elsewhere. I believe linking in something hacky like this should work (not tested but I used something similar recently):
```
import core.thread;
import core.demangle;
import gc.gcinterface;
import gc.impl.manual.gc;

// access to private symbol
pragma(mangle, mangle!GC("gc.proxy.instance")) extern __gshared GC instance;

extern (C)
{
   // replace COMDATs in druntime lib
    void gc_init()
    {
        config.gc = "manual";
        ManualGC.initialize(instance);
        thread_init();
    }
    void gc_term()
    {
        thread_term();
        ManualGC.finalize(instance);
    }
}
```

skl131313:
@andralex Any thoughts? Walter seems to be busy.

wilzbach:
Okay as no one is responding and this has been broken for quite a while and existing as a copy in `src/impl/manual/gc.d` too, I will set it to `auto-merge`. 
Thanks @skl131313 for pushing this!

wilzbach:
> Maybe a bit of information how to implement something similar to the gcstub (with a big red warning about its limitations) in the wiki would be nice, i.e. not linking in the regular GC, but allowing to proxy it elsewhere. 

I didn't find a good place in the wiki for this (maybe someone else did?). However, I believe this would be more than interesting for a new article in the `@nogc` series...

skl131313:
Thanks for merging it!

PR #5810, opened 2017-10-26
Fix Issue 17741 - std.range.transposed does not work with Map

Darredevil:
I had to change a test for issue 8764 however I checked and it does not introduce regression.
https://issues.dlang.org/show_bug.cgi?id=8764

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -6594,8 +6594,7 @@ Transversal!(RangeOfRanges, opt) transversal
 
 struct Transposed(RangeOfRanges)
 if (isForwardRange!RangeOfRanges &&
-    isInputRange!(ElementType!RangeOfRanges) &&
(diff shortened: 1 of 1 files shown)


std/range/package.d
```diff
@@ -6678,12 +6677,18 @@ private:
...
+    isInputRange!(ElementType!RangeOfRanges))
 {
-    return Transposed!RangeOfRanges(rr);
+    static if (hasAssignableElements!RangeOfRanges)
+    {
+        return Transposed!RangeOfRanges(rr);
+    }
+    else
+    {
+        return Transposed!(typeof(rr.array))(rr.array);
```
andralex: We can't accept this. It eagerly copies the range, which is inefficiently and unexpected - entirely out of character for std.range. If a user wants this, they can use `r.array.transposed` themselves.

PR #5472, opened 2017-06-11
Eliminate shared this from std/socket.d

andralex:
This ends up simplifying Posix a lot. It could be simplified even further.

PR #5813, opened 2017-10-26
Eliminate static shared this in std.socket

andralex:
A fresh approach based loosely on https://github.com/dlang/phobos/pull/5472.

### DIFF
--- a/std/socket.d
+++ b/std/socket.d
@@ -335,9 +335,6 @@ version(Windows) private
 
     void initialize() @trusted
     {
-        static bool tlsInitialized;
-        if (tlsInitialized) return;
-
         static shared bool processInitialized;
         import std.concurrency : initOnce;
 
@@ -368,7 +365,6 @@ version(Windows) private
         }
 
         initOnce!processInitialized(perform());
-        tlsInitialized = true;
     }
 }
 else version(Posix)
--- a/std/socket.d
+++ b/std/socket.d
@@ -302,7 +302,8 @@ version(Posix) private enum bool getnameinfoSupported = true;
 version(Windows) private @property bool getaddrinfoSupported()
 {
     initialize;
-    return getaddrInfoPointer !is null;
+    assert((getaddrinfoPointer is null) == (freeaddrinfoPointer is null));
+    return getaddrinfoPointer !is null;
 }
 // Ditto (always supported on Posix).
 version(Posix) private enum bool getaddrinfoSupported = true;
@@ -318,24 +319,26 @@ version(Windows) private
     auto our_getnameinfo(A...)(A a) @system
     {
         initialize;
-        return atomicLoad(getnameinfoPointer)(a);
+        return getnameinfoPointer(a);
     }
 
     auto our_getaddrinfo(A...)(A a) @system
     {
         initialize;
-        return atomicLoad(getaddrinfoPointer)(a);
+        return getaddrinfoPointer(a);
     }
 
     auto our_freeaddrinfo(A...)(A a) @system
     {
         initialize;
-        return atomicLoad(freeaddrinfoPointer)(a);
+        return freeaddrinfoPointer(a);
     }
 
     void initialize() @trusted
     {
         static shared bool processInitialized;
+        if (atomicLoad(processInitialized)) return;
+
         import std.concurrency : initOnce;
 
         static bool perform()
@@ -354,12 +357,12 @@ version(Windows) private
             auto ws2Lib = GetModuleHandleA("ws2_32.dll");
             if (ws2Lib)
             {
-                atomicStore(getnameinfoPointer, cast(typeof(getnameinfoPointer))
-                            GetProcAddress(ws2Lib, "getnameinfo"));
-                atomicStore(getaddrinfoPointer, cast(typeof(getaddrinfoPointer))
-                            GetProcAddress(ws2Lib, "getaddrinfo"));
-                atomicStore(freeaddrinfoPointer, cast(typeof(freeaddrinfoPointer))
-                            GetProcAddress(ws2Lib, "freeaddrinfo"));
+                getnameinfoPointer = cast(typeof(getnameinfoPointer))
+                            GetProcAddress(ws2Lib, "getnameinfo");
+                getaddrinfoPointer = cast(typeof(getaddrinfoPointer))
+                            GetProcAddress(ws2Lib, "getaddrinfo");
+                freeaddrinfoPointer = cast(typeof(freeaddrinfoPointer))
+                            GetProcAddress(ws2Lib, "freeaddrinfo");
             }
             return true;
         }
--- a/std/socket.d
+++ b/std/socket.d
@@ -289,55 +289,93 @@ bool wouldHaveBlocked() nothrow @nogc
     assert(rec == -1 && wouldHaveBlocked());
 }
 
-
-private immutable
+// Returns true iff getnameinfo is supported on this platform.
+version(Windows) private @property bool getnameinfoSupported()
 {
-    typeof(&getnameinfo) getnameinfoPointer;
-    typeof(&getaddrinfo) getaddrinfoPointer;
-    typeof(&freeaddrinfo) freeaddrinfoPointer;
+    initialize;
+    return getnameinfoPointer !is null;
 }
+// Ditto (always supported on Posix).
+version(Posix) private enum bool getnameinfoSupported = true;
 
-shared static this() @system
+// Returns true iff getaddrinfo is supported on this platform.
+version(Windows) private @property bool getaddrinfoSupported()
 {
-    version (Windows)
+    initialize;
+    return getaddrInfoPointer !is null;
+}
+// Ditto (always supported on Posix).
+version(Posix) private enum bool getaddrinfoSupported = true;
+
+version(Windows) private
+{
+    import core.atomic;
+
+    shared typeof(&getnameinfo) getnameinfoPointer;
+    shared typeof(&getaddrinfo) getaddrinfoPointer;
+    shared typeof(&freeaddrinfo) freeaddrinfoPointer;
+
+    auto our_getnameinfo(A...)(A a) @system
     {
-        WSADATA wd;
+        initialize;
+        return atomicLoad(getnameinfoPointer)(a);
+    }
 
-        // Winsock will still load if an older version is present.
-        // The version is just a request.
-        int val;
-        val = WSAStartup(0x2020, &wd);
-        if (val)         // Request Winsock 2.2 for IPv6.
-            throw new SocketOSException("Unable to initialize socket library", val);
+    auto our_getaddrinfo(A...)(A a) @system
+    {
+        initialize;
+        return atomicLoad(getaddrinfoPointer)(a);
+    }
 
-        // These functions may not be present on older Windows versions.
-        // See the comment in InternetAddress.toHostNameString() for details.
-        auto ws2Lib = GetModuleHandleA("ws2_32.dll");
-        if (ws2Lib)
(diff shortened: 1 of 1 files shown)


std/socket.d
```diff
@@ -307,56 +307,94 @@ bool wouldHaveBlocked() nothrow @nogc
...
+                throw new SocketOSException("Unable to initialize socket library", val);
+            static extern(C) void cleanup() { WSACleanup(); }
+            atexit(&cleanup);
+
+            // These functions may not be present on older Windows versions.
+            // See the comment in InternetAddress.toHostNameString() for details.
+            auto ws2Lib = GetModuleHandleA("ws2_32.dll");
+            if (ws2Lib)
+            {
+                atomicStore(getnameinfoPointer, cast(typeof(getnameinfoPointer))
```
MartinNowak: If I read this correctly, it's supported since WinXP.
https://msdn.microsoft.com/en-us/library/windows/desktop/ms738520(v=vs.85).aspx#SUPPORT_FOR_GETADDRINFO_ON_WINDOWS_2000_AND_OLDER_VERSIONS_

std/socket.d
```diff
@@ -307,56 +307,94 @@ bool wouldHaveBlocked() nothrow @nogc
...
+    }
 
-shared static ~this() @system nothrow @nogc
-{
-    version(Windows)
+    private void initialize() @trusted
     {
-        WSACleanup();
+        static bool tlsInitialized;
+        if (tlsInitialized) return;
```
MartinNowak: ???
Why do you put a *TLS* flag around the double-checked locking in initOnce?
https://github.com/dlang/phobos/blob/27072f8282fd60ef2af025280d217729d2b5770f/std/concurrency.d#L2487
Considering how slow name lookups are, performance shouldn't be a hindrance. And even then you'd better spent some time to optimize initOnce for dmd than to add more and more workarounds.

std/socket.d
andralex: Yah, that's what @CyberShadow said too. Then we had linking problems so I backed off of that.

std/socket.d
andralex: Just wanted to make `initialize` cheap to call in all instances.

andralex:
Fixed typo on Windows.

std/socket.d
MartinNowak: It's already cheap to call initOnce, that's the purpose of it.
And besides a TLS access isn't actually free either.

MartinNowak: I'm a bit worried about cargo cult here.
That TLS locking pattern came up as alternative to "double-checked locking", only that double-checked locking works fine with atomic acquire&release primitives (which are free on X86).
So in total you have the overhead of a TLS access against a normal X86 load instruction.
Now in this PR you added both to be even faster.?

std/socket.d
MartinNowak: Just seeing those linking issues myself, it's an outed ws2_32.lib (OMF import lib for optlink) shipping with dmd, so much to that rotten format.

std/socket.d
andralex: No problem, glad to remove.

std/socket.d
```diff
@@ -307,56 +307,94 @@ bool wouldHaveBlocked() nothrow @nogc
...
+
+        static bool perform()
+        {
+            // Winsock will still load if an older version is present.
+            // The version is just a request.
+            WSADATA wd;
+            const val = WSAStartup(0x2020, &wd);
+            if (val)         // Request Winsock 2.2 for IPv6.
+                throw new SocketOSException("Unable to initialize socket library", val);
+            static extern(C) void cleanup() { WSACleanup(); }
```
MartinNowak: Actually quite nice that dmd mangles nested C functions, no?

std/socket.d
```diff
@@ -307,56 +307,90 @@ bool wouldHaveBlocked() nothrow @nogc
...
+            atexit(&cleanup);
+
+            // These functions may not be present on older Windows versions.
+            // See the comment in InternetAddress.toHostNameString() for details.
+            auto ws2Lib = GetModuleHandleA("ws2_32.dll");
+            if (ws2Lib)
+            {
+                atomicStore(getnameinfoPointer, cast(typeof(getnameinfoPointer))
+                            GetProcAddress(ws2Lib, "getnameinfo"));
+                atomicStore(getaddrinfoPointer, cast(typeof(getaddrinfoPointer))
```
MartinNowak: Those don't need to be atomic, since initOnce is holding a lock for you.

std/socket.d
```diff
@@ -307,56 +307,90 @@ bool wouldHaveBlocked() nothrow @nogc
...
     }
-    else version(Posix)
+
+    auto our_getaddrinfo(A...)(A a) @system
     {
-        getnameinfoPointer = &getnameinfo;
-        getaddrinfoPointer = &getaddrinfo;
-        freeaddrinfoPointer = &freeaddrinfo;
+        initialize;
+        return atomicLoad(getaddrinfoPointer)(a);
```
MartinNowak: Those don't need to be atomic either since in case it really did initialize, the mutex has the necessary memory barriers.

std/socket.d
MartinNowak: > Yah, that's what @CyberShadow said too. Then we had linking problems so I backed off of that.

I just fixed those linking issues, https://github.com/dlang/installer/pull/264, after 6 years :o, but it'll take a while till those libs arrive at the testers.

MartinNowak:
Looks like you still need to wrap `gethostbyname` and such functions not loaded from the DLL.

andralex:
@MartinNowak thanks. I've made the updates prompted by your review save for calls to `initialize`. That'd be difficult to do without a Windows host. If you can get to this great, otherwise I'll return to this later.

quickfur:
ping @andralex 

Are we moving forward with this?  Is it just a matter of a rebase?

andralex:
@quickfur yes I'll get to this soon thanks for the prod

dejlek:
So... where are we with this PR?

andralex:
OK, I rebased. Let's see how it goes now.

std/socket.d
```diff
@@ -289,56 +289,93 @@ bool wouldHaveBlocked() nothrow @nogc
...
+        import std.concurrency : initOnce;
+
+        static bool perform()
+        {
+            // Winsock will still load if an older version is present.
+            // The version is just a request.
+            WSADATA wd;
+            const val = WSAStartup(0x2020, &wd);
+            if (val)         // Request Winsock 2.2 for IPv6.
+                throw new SocketOSException("Unable to initialize socket library", val);
```
andralex, line 351: I'm considering aborting the application here. It pretty much means the Windows installation is messed up. Thoughts?

MoonlightSentinel: Sounds like an horrible idea for GUI applications. Why?

std/socket.d
```diff
@@ -307,56 +307,93 @@ bool wouldHaveBlocked() nothrow @nogc
...
+    auto our_freeaddrinfo(A...)(A a) @system
+    {
+        initialize;
+        return freeaddrinfoPointer(a);
+    }
 
-shared static ~this() @system nothrow @nogc
-{
-    version(Windows)
+    private void initialize() @trusted
```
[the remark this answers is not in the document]
andralex: done

std/socket.d
andralex: There's just no realistic scenario that warrants throwing an exception (and forcing all transitive clients of this function to mind it). Last time `WSAStartup()` failed was on Windows 7 or whatever: https://www.youtube.com/watch?v=MWc6fGipU28.

Normally I wouldn't mind much, but it's time to acknowledge that throwing exceptions is painful for programmatic clients and we should only throw them when it makes sense. We need to tighten the screws on this stuff.

andralex: Oh, there is an alternative: simply return and leave the pointers to functions to be null. Then nothing will work.

andralex: OK, I think this is nice: if the lib doesn't load, no networking functionality is supported. There was code before that accounted for that anyway.

std/socket.d
```diff
@@ -289,56 +289,96 @@ bool wouldHaveBlocked() nothrow @nogc
     assert(rec == -1 && wouldHaveBlocked());
 }
 
+// Returns true iff getnameinfo is supported on this platform.
+version(Windows) private @property bool getnameinfoSupported()
```
CyberShadow, line 293: Why add `getaddrinfoSupported` and `getnameinfoSupported` and not just make `getnameInfoPointer` / `getaddrinfoPointer` lazily-initialized property functions?

std/socket.d
```diff
@@ -289,56 +289,96 @@ bool wouldHaveBlocked() nothrow @nogc
...
+        if (freeaddrinfoPointer)
+            // May be null if WSAStartup() failed, just do nothing
+            return;
+        return freeaddrinfoPointer(a);
+    }
 
-shared static ~this() @system nothrow @nogc
-{
-    version (Windows)
+    void initialize() @trusted nothrow
```
CyberShadow, line 340: If I'm reading this correctly, this function needs to be called in a lot more places now (every single bit of public code that eventually calls WSA functions, so that it calls `WSAStartup`), and I don't see this patch doing it.

std/socket.d
andralex: That's not new to this PR - old code handled the cases when the pointers were null and used workaround solutions.

So, old code:

* try to fetch the pointers at startup
* compare the pointers against null in code

New code:

* define xxxSupported and xxx, both doing initialization on first call
* use xxxSupported instead of comparing pointers to null

std/socket.d
andralex: Got it, will fix. Thanks.

std/socket.d
CyberShadow: Right, but, why not just keep comparing pointers to null?

andralex: Because you need to call initialize before comparing them to null.

CyberShadow: Yes, so, why not turn the function pointer *variables* into function pointer returning *property functions*, in the same way that `getnameinfoSupported` are in the current version, but returning the pointer directly instead of a `bool`?

andralex: That would be an alternative, but it seems more roundabout. To make sure I understand:

```D
@property auto our_getnameinfoPtr() {
    initialize;
    return &getNameInfoPtr;
}
... using code ...
if (our_getnameinfoPtr)
{
    ...
    our_getnameinfoPtr(... arguments ...);
}
```

Is this the idea? Mine was to not use pointers at all, but instead have the simple "if xxxSupported returns true, you may call xxx" convention. It seems yours could be used to avoid two calls to initialize by using local variables:

```d
... using code ...
auto gni = our_getnameinfoPtr;
if (gni)
{
    ...
    gni(... arguments ...);
}
```

CyberShadow: > That would be an alternative, but it seems more roundabout.

I don't see how that would be more roundabout. If anything, adding an additional pair of functions would better fit the description of "roundabout".

> It seems yours could be used to avoid two calls to initialize by using local variables:

I don't see the need to avoid performing two calls. The compiler ought to inline the function, at which point it can trivially conclude that under the `if` branch, the pointer cannot possibly be null, and eliminate the second comparison.

andralex: Cool, yeah, two functions are better than four.

andralex:
Folks, I'll need to give up on this. There's too much work for Windows (albeit fun) but I don't have the time to set up Windows right now.

Eliminating static cdtors from druntime and phobos is key to lean and mean executables. Any function called from a static cdtors ends up linked in the program, transitively with all other functions it calls. And you know Phobos... you link one, you link'em all :)

PR #5470, opened 2017-06-11
Eliminate static shared this from std/parallelism.d

andralex:
Wondering how to get rid of the `shared static ~this`. cc @dsimcha @wilzbach @quickfur @JackStouffer @MartinNowak

PR #5814, opened 2017-10-27
Revert "Eliminate static shared this from std/parallelism.d"

ZombineDev:
Reverts dlang/phobos#5470, since it breaks several projects.


ZombineDev:
Jenkins didn't like this PR - I'm giving it another try in https://github.com/dlang/phobos/pull/5815.

PR #5815, opened 2017-10-27
Revert "Merge pull request #5470

ZombineDev:
This reverts commit 87bdeaee41f7bedbc3a5f974eedc013377ef3822, reversing changes made to b5e6365a9bb9e0ca3af194d7add212c092ad1b13, since it breaks the Dustmite, among probably other projects.

### DIFF
--- a/std/parallelism.d
+++ b/std/parallelism.d
@@ -99,80 +99,86 @@ else version(NetBSD)
     version = useSysctlbyname;
 }
 
-/*
-A lazily initialized global constant. The underlying value is a shared global
-statically initialized to `outOfBandValue` which must not be a legit value of
-the constant. Upon the first call the situation is detected and the global is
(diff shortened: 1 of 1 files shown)


andralex:
So what happened here? I see two unrelated pull requests mentioned.

ZombineDev:
@andralex The first part of my pull request description was generated automatically by git. It means that your pull-request (#5470) will be reverted and after that the master branch will be in the same state as before (#5470) was merged - the last commit effective commit will be b5e6365 (which remains intact).

andralex:
I suggest https://github.com/dlang/phobos/pull/5816 instead. I plan to expose the functionality publicly but would like to work the kinks out for a release cycle.

ZombineDev:
Yes, I would prefer #5816 too.

PR #5816, opened 2017-10-27
Fix visibility issue for https://github.com/dlang/phobos/pull/5470

andralex:


andralex:
thx @ZombineDev @jmdavis for a quick turnaround

PR #5817, opened 2017-10-27
Fix FreeBSD SbrkRegion alignment issue

edi33416:
Please see discussion at #5806 

@andralex @jmdavis


jmdavis:
Not being particularly familiar with std.experimental.allocator, I can't really properly review these changes, but I can confirm that with these changes, the unit tests now pass again on my local FreeBSD system (which is TrueOS stable, so it's a 64-bit version of FreeBSD 12-to-be). And if you tested them on the 4 VMs like you said you were going to in the other PR, then presumably, it works for 10.3 and 11 for both 32-bit and 64-bit, but the autotesters with the newer versions of FreeBSD don't run for PRs.

PR #5818, opened 2017-10-27
Added templates for working with overload sets

Biotronic:
std.traits currently has no tools for working with overload sets. This PR adds two templates:

- `getOverloads`, which is a wrapper around `__traits(getOverloads)`. It takes its argument as an alias, instead of `__traits`'s more complicated `aggregate, name` pattern, and matches more closely how other templates in std.traits works.

- `overloadFor`, which finds the overload for a function that matches the given argument list.


Biotronic:
An argument could be presented for the addition also of `exactOverloadFor`, which would work like `overloadFor`, but do precise matching - `overloadFor` checks if the function can be called with the given arguments, while `exactOverloadFor` would check that the argument types match exactly. It's my impression that `exactOverloadFor` would generally not be what users want, and that having two templates would increase the cognitive load for little benefit.

ZombineDev:
> An argument could be presented for the addition also of `exactOverloadFor` [..]

Yes, for example one may need to obtain a function pointer to the function, in which case the ABI would be different.
How about:
```D
template overloadFor(alias func, bool exactMatch, Args...)
```

Obviously we can't use an optional parameter, but anyone could add a wrapper in their code if they always prefer exact behavior or not.

Biotronic:
I don't quite follow - there's nothing about `overloadFor` as currently implemented that prevents taking the address of the result.

As for the bool argument, it's a lot worse than having two templates - seeing `overloadFor!(foo, false, 1, 2)` does not in any way convey that the `false` means it's a fuzzy match - it could be one of the function arguments, and even when you know it isn't, does false mean fuzzy or exact?

As I started implementing `exactOverloadFor`, I found that it's more complex than I first imagined - function parameters can have qualifiers that aliases cannot. In other words, `(ref int) => {}` is impossible to match exactly with an `AliasSeq`, since `Parameters!((ref int) => {})` is `(ref int)`, and `AliasSeq!(ref int)` fails to compile. Not only that, but `alias a = Parameters!((ref int) => {})[0];` makes `a` = `int`, not `ref int`.

This indicates the only possible signature for `exactOverloadFor` is one that takes a callable as its pattern, and calling it as `exactOverloadFor!(fun, (ref int) => 1)`.

std/traits.d
```diff
@@ -8087,3 +8089,97 @@ enum isCopyable(S) = is(typeof(
...
+{
+    alias p = AliasSeq!(__traits(parent, func))[0];
+    static if (is(p) || p.stringof[0 .. 7] == "module ")
+        alias getOverloads = AliasSeq!(__traits(getOverloads, __traits(parent, func), __traits(identifier, func)));
+    else
+        alias getOverloads = func;
+}
+
+version(unittest)
+{
```
wilzbach: `private:`

std/traits.d
```diff
@@ -8087,3 +8089,97 @@ enum isCopyable(S) = is(typeof(
...
 }
+
+/**
+ * Determines the type of the argument.
+ *
+ * Returns:
+ *   T[0] if it's a type, typeof(T[0]) if it's a value, void if it's a template.
+ *
+ */
+template TypeOf(T...)
```
wilzbach: I would set this to `private` and don't expose via ddoc. At least until you have convinced us that it's useful on its own.

std/traits.d
JackStouffer: These should instead be defined as static functions in the unit test block. Currently, the reader of the doc has no idea what's being tested.

std/traits.d
```diff
@@ -8087,3 +8089,97 @@ enum isCopyable(S) = is(typeof(
...
+        else
+            alias match = match!(fns[1 .. $]);
+    }
+    alias overloadFor = match!(getOverloads!func);
+}
+
+///
+@safe unittest
+{
+    assert(is(typeof(overloadFor!(getOverloadsTest)) == typeof(getOverloads!getOverloadsTest[0])));
```
JackStouffer: Re-define the functions here for clarity

std/traits.d
```diff
@@ -8087,3 +8089,97 @@ enum isCopyable(S) = is(typeof(
...
+{
+    assert(is(TypeOf!3 == int));
+    assert(is(TypeOf!int == int));
+    assert(is(TypeOf!([]) == void[]));
+    assert(!is(TypeOf!(mixin(__MODULE__))));
+    assert(is(TypeOf!TypeOf));
+}
+
+/**
+ * Get a list of the overloads for the given callable.
```
JackStouffer: `Params:` and `Returns:` sections should be defined here.

std/traits.d
```diff
@@ -8087,3 +8089,97 @@ enum isCopyable(S) = is(typeof(
...
+    alias fn = getOverloadsTest;
+    assert(is(typeof(&fn) == typeof(&getOverloads!getOverloadsTest[0])));
+    assert(!is(typeof(&fn) == typeof(&getOverloads!getOverloadsTest[1])));
+    assert(getOverloads!getOverloadsTest2.length == 1);
+    assert(getOverloads!(() => 3).length == 1);
+    assert(getOverloads!(() => 3).length == 1);
+}
+
+/**
+ * Gets the overload for `func` that can be called with `Args` as arguments, if any.
```
JackStouffer: `Params:` and `Returns:` sections should be defined here.

std/traits.d
```diff
@@ -8087,3 +8089,97 @@ enum isCopyable(S) = is(typeof(
...
+ */
+template overloadFor(alias func, Args...)
+    if (isCallable!func)
+{
+    alias ArgTypes = staticMap!(TypeOf, Args);
+    template match(fns...)
+    {
+        static if (fns.length == 0)
+            static assert(false, "No overload found for parameter types "~Args.stringof);
```
JackStouffer: `static assert`-ing here seems incorrect on first glance. Why not return an empty `AliasSeq`?

std/traits.d
```diff
@@ -8087,3 +8089,97 @@ enum isCopyable(S) = is(typeof(
...
+ *
+ */
+template TypeOf(T...)
+    if (T.length == 1 && (is(T[0]) || is(typeof(T[0]))))
+{
+    static if (is(T[0]))
+        alias TypeOf = T[0];
+    else static if (is(typeof(T[0])))
+        alias TypeOf = typeof(T[0]);
+}
```
MetaLang: Needs an else path for things without a type like modules and `__error__`.

std/traits.d
```diff
@@ -8087,3 +8089,97 @@ enum isCopyable(S) = is(typeof(
...
+
+/**
+ * Get a list of the overloads for the given callable.
+ */
+template getOverloads(func...)
+    if (func.length == 1 && isCallable!func)
+{
+    alias p = AliasSeq!(__traits(parent, func))[0];
+    static if (is(p) || p.stringof[0 .. 7] == "module ")
```
MetaLang: Instead of this ugly hack you could use this _slightly_ less ugly hack:
`enum isModuleOrPackage(alias s) = !__traits(compiles, { alias _ = typeof(s); })`

It works because the only two things that don't have a type (that I'm aware of, anyway...) that can still be passed as a template alias parameter are a module name and a package name. This has the advantage that it also works for packages.

std/traits.d
Biotronic: No can do, since this is testing overloads, and function-local functions can't have those.

std/traits.d
Biotronic: Done.

std/traits.d
Biotronic: Done.

std/traits.d
```diff
@@ -8087,3 +8089,105 @@ enum isCopyable(S) = is(typeof(
...
+    else
+        alias getOverloads = func;
+}
+
+version (unittest)
+{
+    // Local functions can't have overloads, so the functions used by
+    // getOverloads and overloadFor need to be defined here.
+    private:
+    static int getOverloadsTest(int n = 3) { return n; }
```
JackStouffer: Since you can't add these in the unit test, just add them as three commented lines so the doc reader knows what's being tested.

Biotronic: Sorry, I don't quite get what you're asking me to do. You mean each line has to have a separate comment? Doc comment (`///`)?
```
/// For getOverloads and overloadFor
private static int getOverloadsTest(int n = 3) { return n; }
/// Ditto
static int getOverloadsTest(int, int)  { return 3; }
/// Ditto
static int getOverloadsTest2(int, int) { return 3; }
```

JackStouffer: like this

```
@safe unittest
 {
     // int getOverloadsTest(int n = 3) { return n; }
     // int getOverloadsTest(int, int)  { return 3; }
     // int getOverloadsTest2(int, int) { return 3; }

     assert(getOverloads!getOverloadsTest.length == 2);
     assert(is(typeof(&getOverloadsTest) == typeof(&getOverloads!getOverloadsTest[0])));
     assert(!is(typeof(&getOverloadsTest) == typeof(&getOverloads!getOverloadsTest[1])));
     alias fn = getOverloadsTest;
     assert(is(typeof(&fn) == typeof(&getOverloads!getOverloadsTest[0])));
     assert(!is(typeof(&fn) == typeof(&getOverloads!getOverloadsTest[1])));
     assert(getOverloads!getOverloadsTest2.length == 1);
     assert(getOverloads!(() => 3).length == 1);
     assert(getOverloads!(() => 3).length == 1);
 }
```

quickfur:
ping @andralex

std/traits.d
```diff
@@ -8087,3 +8089,124 @@ enum isCopyable(S) = is(typeof(
...
+ * Params:
+ *    func = the overload set to search.
+ *    Args = the set of argument types to look for.
+ * Returns:
+ *    The overload that matches the argument types in Args
+ */
+template overloadFor(alias func, Args...)
+    if (isCallable!func)
+{
+    alias ArgTypes = staticMap!(TypeOf, Args);
```
ntrel: The docs say that `Args` is a set of types - this should be *sequence* of types. The docs don't mention also accepting values - is that something that `std.traits` has a precedent for? If so the docs for `Args` must mention values too.

If we want to support both types and values, maybe we only need to support one or the other, and not allow interleaving. In that case, we could remove the module symbol `TypeOf` and instead have:
```d
alias ArgTypes = is(Args) ? Args : typeof(Args);
```
If we do want interleaving, the `staticMap` expression could be rolled into an overload of `TypeOf` taking more than one argument.

std/traits.d
```diff
@@ -8087,3 +8089,124 @@ enum isCopyable(S) = is(typeof(
...
+template getOverloads(func...)
+    if (func.length == 1 && isCallable!func)
+{
+    enum isModuleOrPackage(alias s) = !__traits(compiles, { alias _ = typeof(s); });
+    alias Parent = AliasSeq!(__traits(parent, func))[0];
+
+    static if (is(Parent) || isModuleOrPackage!Parent)
+        alias getOverloads = AliasSeq!(__traits(getOverloads, Parent, __traits(identifier, func)));
+    else
+        alias getOverloads = func;
```
ntrel: Perhaps add a comment with an example of when the `else` branch is triggered?

thewilsonator:
ping @andralex

std/traits.d
```diff
@@ -8087,3 +8089,124 @@ enum isCopyable(S) = is(typeof(
...
 }
+
+private template TypeOf(T...)
+    if (T.length == 1 && (is(T[0]) || is(typeof(T[0]))))
+{
+    static if (is(T[0]))
+        alias TypeOf = T[0];
+    else static if (is(typeof(T[0])))
+        alias TypeOf = typeof(T[0]);
+}
```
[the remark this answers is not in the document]
andralex: Oh, one more thing. The constraint is not needed on this private template; any error would be eliminated upstream in the public interface. So you can drop `TypeOf`'s constraint.

std/traits.d
Biotronic: Done.

std/traits.d
```diff
@@ -8087,3 +8089,152 @@ enum isCopyable(S) = is(typeof(
...
+ *     A tuple of the overloads for the given callable.
+ */
+template getOverloads(func...)
+if (func.length == 1 && isCallable!func)
+{
+    enum isModuleOrPackage(alias s) = !is(typeof(s));
+    alias Parent = __traits(parent, func);
+
+    static if (is(Parent) || isModuleOrPackage!Parent)
+        alias getOverloads = AliasSeq!(__traits(getOverloads, Parent, __traits(identifier, func)));
```
BBasile: `AliasSeq` is not required anymore to alias `__traits` results.

BBasile: Also don't forget that `getOverloads` has an optional 3rd parameter since a few releases. It's a bool and it allows to include templates (esp. function templates).

thewilsonator:
@Biotronic please address @BBasile's comments and then this looks good to go, modulo whatever the CI is doing.

thewilsonator:
```
Running DScanner
generated/dscanner-383fcb84d892e5169c134e282878ee2c51e4265f/dsc --config .dscanner.ini --styleCheck etc std -I.
std/traits.d(8749:121)[warn]: Line is longer than 120 characters
std/traits.d(8810:121)[warn]: Line is longer than 120 characters
std/traits.d(8815:121)[warn]: Line is longer than 120 characters
std/traits.d(8816:109)[warn]: Line is longer than 120 characters
```

Also under buildkite/style
```
generated/linux/release/64/publictests/std_traits.d(8726): Error: static assert:  `is(int function(int n = 3) == int function(int = 3) pure nothrow @nogc @safe)` is false
--
```
Not sure if this is dscanner doing something weird here?

std/traits.d
```diff
@@ -8664,3 +8666,152 @@ enum isCopyable(S) = is(typeof(
...
+/**
+ * Gets the overload for `func` that can be called with `Args` as arguments, if any.
+ *
+ * Params:
+ *    func = the overload set to search.
+ *    Args = the set of argument types to look for, specified as types or as values.
+ * Returns:
+ *    The overload that matches the argument types in `Args`.
+ */
+template overloadFor(alias func, Args...)
```
thaven: Can you please have a look at PR #6885, which I created, unaware of this PR. It basically does the same thing, but also allows to specify function attributes (which may be the only thing distinguishing overloads).

May be a good idea to integrate that into this PR.

Biotronic: I've had a look-see, and my opinion is everything #6885 does can be done in a library without big syntax sacrifices. Function attributes can also be handled (but aren't at the moment), as pointed out above by letting the template take a lambda: `exactOverloadFor!(fun, (ref int) => 1)`

thaven: #6885 does not propose new syntax (how would it, being a Phobos PR?). There's only a comment  there suggesting an alternative solution using ˋ__traitsˋ.

Anyway, both this PR and #6885 seem to have stalled and I'd like to see some solution for getting an alias to a specific overload.

By the way, I like the idea of taking a function (possibly lambda) as a template parameter, but there should probably be a way to just specify types as well, because parameter attributes might be unknown, making it impossible to create a correct lambda.

PR #5820, opened 2017-10-28
std.math: Implement more optimal floorImpl for ieeeSingle

ibuclaw:
Less branching, and doesn't care about endian-ness.  Tested against PPC64/LE.

For comparison between old (source view 1) and new (source view 2) https://explore.dgnu.org/g/fTvttz

Probably better to review with https://github.com/dlang/phobos/pull/5820/files?w=1

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -475,7 +475,7 @@ T floorImpl(T)(const T x) @trusted pure nothrow @nogc
             // Clear all bits representing the fraction part.
             const uint fraction_mask = F.MANTISSAMASK_INT >> exp;
 
-            if ((y.vi & i) != 0)
+            if ((y.vi & fraction_mask) != 0)
             {
                 // If 'x' is negative, then first substract 1.0 from the value.
                 if (y.vi < 0)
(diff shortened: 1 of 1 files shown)


ibuclaw:
Ping?

JackStouffer:
Pinging people who are qualified to review this from git blame @yebblies @wilzbach @don-clugston-sociomantic @klickverbot

ibuclaw:
Or you know, you could just request a review. :)

std/math.d
```diff
@@ -464,24 +463,42 @@ T floorImpl(T)(const T x) @trusted pure nothrow @nogc
...
-        exp -= 16;
+        if (exp < (T.mant_dig - 1))
+        {
+            const uint i = F.MANTISSAMASK_INT >> exp;
+
+            // Clear all bits representing the fraction part.
+            if ((y.vi & i) != 0)
+            {
+                if (y.vi < 0)
+                    y.vi += 0x00800000 >> exp;
```
don-clugston-sociomantic: This REALLY needs a comment. My guess is that this is the float value 0.5, added so that it rounds up towards zero? Or something like that.

ibuclaw:
@redstar - And this you might be interested in also.

std/math.d
ibuclaw: This is subtracting one, adding comment.

PR #5822, opened 2017-10-29
std.math: Implement floor for ibmExtended reals.

ibuclaw:
This is just a slight extension of #5820.  Tested against PPC64/LE.

https://github.com/dlang/phobos/pull/5822/files?w=1

PR #4336, opened 2016-05-18
Begin replacing floatTraits pointer math with RealRep union

tsbockman:
Pros:
- Prepare the way for CTFE floating-point math by switching from explicit pointer reinterpretation to a `union`.
- `union` code is easier to read and understand.
- `union` code is also easier to write, thus freeing mental energy for performance optimizations.
- Consolidate tons of near-duplicate code-paths into (usually) two: one for `ibmExtended`, and one for everything else.
- Adding support for new IEEE-style floating-point types (such as `half` or `quad`) gets easier as more of `std.math` is converted to use `RealRep`. (`ucent` would help here, though.)
- By the time all of `std.math` has been converted, I expect that (in net) `RealRep` will have reduced the total line count substantially.
- It may be possible to replace some of the inline assembler with `RealRep` code; the codegen seems to be very clean, for the most part.

Cons:
- Many large diffs (starting with this one) will be required to complete the conversion.
- `std.math` will become substantially more dependant upon inlining for good performance.

PR #5823, opened 2017-10-29
std.math: Use DOUBLEPAIR_MSB/LSB for accessing high/low parts of ibmExtended format

ibuclaw:
A continuation of #5822 (or a prelude, works both ways).

When running the unittests for PPC64/LE, found many failures due to it checking the wrong part (`MANTISSA_MSB` value is the least significant index on `LittleEndian`).

Fixed places where Big/Little endian would otherwise differ, but actually should do the same thing.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -375,15 +375,22 @@ template floatTraits(T)
         enum ushort EXPMASK = 0x7FF0;
         enum ushort EXPSHIFT = 4;
         enum realFormat = RealFormat.ibmExtended;
-        // the exponent byte is not unique
+
+        // For IBM doubledouble the larger magnitude double comes first.
+        // It's really a double[2] and arrays don't index differently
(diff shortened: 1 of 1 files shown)


ibuclaw:
Maybe its time to resurrect  #4336

redstar:
Very nice - I found this, too. I will have a closer look this evening. One point: you are enabling doubledouble support on little endian. This should also be reflected in the version check at the beginning of the file (https://github.com/dlang/phobos/blob/master/std/math.d#L233).

ibuclaw:
@redstar, oh and there's also #5825 which fixes this using an entirely different approach.

redstar:
Ok, this is the simple approach. :-) Looks good to me. There are no overlooked instance, so it should work out of the box.

ibuclaw:
Rebased.

PR #5824, opened 2017-10-29
Add std.getopt.config.hidden

zorael:
This configuration flag hides the option following it from being printed in defaultGetoptPrinter's --help screen.

The use-cases for this include:
* hidden and undocumented options
* maintaining old and deprecated options without encouraging their use by listing them
* stealth convenience aliases for alternative spellings; --colour and --color, --summarise and --summarize

Alternative workarounds include:
* faith-in-fellow-man approaches such as marking options "[deprecated please do not use]" in their description texts
* using "colour|color" tags and living with both being displayed (currently not possible due to [bug 17945](https://issues.dlang.org/show_bug.cgi?id=17945))
* possible to reinvent the wheel and roll your own printer that manually filters

A different approach could be to allow for tagging individual flags with some token to hide them, much like appending a '+' to an option changes its behaviour to be an incremental option.

    // example from the docs
    uint paranoid;
    getopt(args, "paranoid+", &paranoid);

    bool secret;
    getopt(args, "secret@", &secret);  // @^_'#%...

No unit tests were added; I'm unsure how to test the output of the printer.

### DIFF
--- a/std/getopt.d
+++ b/std/getopt.d
@@ -408,11 +408,18 @@ $(D args) after $(D getopt) returns.
 
 Help_Information_Generation:
 If an option string is followed by another string, this string serves as a
-description for this option. The $(D getopt) function returns a struct of type
+description for this option. This field is optional, and omitting it will yield
+the same result as when supplying an empty string.
+
+The $(D getopt) function returns a struct of type
 $(D GetoptResult). This return value contains information about all passed options
 as well a $(D bool GetoptResult.helpWanted) flag indicating whether information
 about these options was requested. The $(D getopt) function always adds an option for
 `--help|-h` to set the flag if the option is seen on the command line.
+
+An option can be prevented from being displayed in this information list by
+giving it a description value of `null`. The option will still be available for
+use, but it will be hidden.
 
 Options_Terminator:
 A lone double-dash terminates $(D getopt) gathering. It is used to
@@ -500,7 +507,7 @@ struct GetoptResult {
 struct Option {
     string optShort; /// The short symbol for this option
     string optLong; /// The long symbol for this option
-    string help; /// The description of this option
+    string help = ""; /// The description of this option
     bool required; /// If a option is required, not passing it will result in an error
 }
 
@@ -555,7 +562,7 @@ follow this pattern:
 
  - config override: a config value, optional
  - option:          a string or a char
- - description:     a string, optional
+ - description:     a string, optional, may be explicitly null
  - receiver:        a pointer or a callable
 */
 private template optionValidator(A...)
@@ -583,12 +590,13 @@ private template optionValidator(A...)
             else foreach (i; staticIota!(1, A.length))
             {
                 static if (!isReceiver!(A[i]) && !isOptionStr!(A[i]) &&
-                    !(is(A[i] == config)))
+                    !is(A[i] == config) && !is(A[i] == typeof(null)))
                 {
                     msg = format(fmt, "invalid argument type: " ~ A[i].stringof, i);
                     break;
                 }
-                else static if (isReceiver!(A[i]) && !isOptionStr!(A[i-1]))
+                else static if (isReceiver!(A[i]) && !isOptionStr!(A[i-1]) &&
+                    !is(A[i-1] == typeof(null)))
                 {
                     msg = format(fmt, "a receiver can not be preceeded by a receiver", i);
                     break;
@@ -1655,6 +1663,8 @@ void defaultGetoptFormatter(Output)(Output output, string text, Option[] opt)
 
     foreach (it; opt)
     {
+        if (it.help is null) continue;
+
         output.formattedWrite("%*s %*s%*s%s\n", ls, it.optShort, ll, it.optLong,
             hasRequired ? re.length : 1, it.required ? re : " ", it.help);
     }
@@ -1854,4 +1864,48 @@ void defaultGetoptFormatter(Output)(Output output, string text, Option[] opt)
     assert(n == -50);
     assert(c == '-');
     assert(f == "-");
+}
+
+@system unittest // Hidden option when description is explicitly null
+{
+    import std.array : appender;
+    import std.string : indexOf;
+
+    auto args = ["program", "--normal", "--empty", "--missing", "--null"];
+    bool normal, empty, missing, null_;
+
+    auto result = getopt(args,
+        "normal", "description", &normal,
+        "empty", "", &empty,
+        "missing", &missing,
+        "null", null, &null_); // hidden
+
+    assert(normal);
+    assert(empty);
+    assert(missing);
+    assert(null_);
+
+    auto app = appender!string();
+    defaultGetoptFormatter(app, "Help", result.options);
+    string helpMsg = app.data;
+
+    assert(helpMsg.indexOf("--normal") != -1);
+    assert(helpMsg.indexOf("--empty") != -1);
+    assert(helpMsg.indexOf("--missing") != -1);
+    assert(helpMsg.indexOf("--null") == -1);
+
+    // Added support for typeof(null) parameters, so test that only the
+    // description may be null
+    bool foo;
+
+    // Normal getopt with config.required flag
+    static assert(__traits(compiles, getopt(args, config.required, "option", "description", &foo)));
+    // null description
+    static assert(__traits(compiles, getopt(args, "option", null, &foo)));
+    // null flag
+    static assert(!__traits(compiles, getopt(args, null, "option", "description", &foo)));
+    // null option string
+    static assert(!__traits(compiles, getopt(args, null, "description", &foo)));
+    // null receiver
+    static assert(!__traits(compiles, getopt(args, "option", "description", null)));
 }
--- a/std/getopt.d
+++ b/std/getopt.d
@@ -390,22 +390,6 @@ getopt(args,
 Only the option directly following $(D std.getopt.config.required) is
 required.
 
-Hidden:
-An option can likewise be marked as hidden. Flagged thus it will still be a
-valid option, but it will not show up the options list produced by
-$(D defaultGetoptPrinter).
-
----------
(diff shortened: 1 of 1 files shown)


jondegenhardt:
One thought about functionality - It would seem useful to defaultGetoptPrinter provide an option include the hidden options in the output. That would provide a backdoor for listing all the options.

For unit tests - Should be able to add unit tests by testing the underlying function DefaultGetoptFormatter. See the unit tests starting at line 1689 in PR.

andralex:
What would be some use cases? And can't we implement that by e.g. passing a `null` string for documentation?

zorael:
My concrete use-case was wanting to silently accept both American and British versions of words for options (colour/color), without having both explicitly listed.

Passing `null` as the description string is clever, conveys meaning too. Better yet, omitting the parameter entirely seems to have the same effect. Both reduce the changes to a one-liner.

    // getopt.d:1656
    foreach (it; opt)
    {
        if (it.help is null) continue;  // <--

        output.formattedWrite("%*s %*s%*s%s\n", ls, it.optShort, ll, it.optLong,
            hasRequired ? re.length : 1, it.required ? re : " ", it.help);
    }

Picking `it.help is null` over `!it.help.length` allows for empty-string descriptions (`getopt(args, "flag", "", &flag)`).

This is my second PR ever. Should I close it and prepare a new one with that change instead? Or revert the commit and keep this one?

andralex:
We'd rather avoid distinguishing across `null` and `""` as subtly different cases. You may want to go with `length == 0`. Feel free to close this or work on it. Thanks!

zorael:
There are two behaviours we would want to support; having options with no descriptions, and having hidden options. No-description options are possible today and are in fact showcased in the very first code example in the `getopt` docs.

    auto helpInformation = getopt(
        args,
        "length",  &length,    // numeric
        "file",    &data,      // string
        "verbose", &verbose,   // flag
        "color", "Information about this color", &color);    // enum

...gives the help screen...

    Some information about the program.
        --length 
          --file 
       --verbose 
         --color Information about this color
    -h    --help This help information.

The description argument is missing in the original `getopt` call, and the `help` string was never initialised.
* If we go by `length == 0` to tell when something should be hidden, options like those in that example will get trapped in the middle and be hidden as well. Current behaviour changes.
* If we go by `is null` and change nothing else, they will still be hidden but can be forced to show by supplying an empty string (since `""` is not `null`). Current behaviour changes.
* If we go by `is null` and modify `optionValidator` to accept `null` arguments (it doesn't currently), then modify the `Option.help` field to be default-initialised to `""`, we preserve the current behaviour and add our ability to hide options by explicitly specifying `null` as description argument ("overwriting" the default `""`). Omitting the argument or specifying `""` will still display them, and nothing breaks.

Like so:

    struct Option {
        string optShort; /// The short symbol for this option
        string optLong; /// The long symbol for this option
        string help = ""; /// The description of this option, initialised to not be null
        bool required; /// If a option is required, not passing it will result in an error
    }

Is the third approach okay? It's a bit hackish but I don't see a way to do this without some interplay between `null` and merely empty strings.

wilzbach:
> Is the third approach okay? It's a bit hackish but I don't see a way to do this without some interplay between null and merely empty strings.

Why not sth. like the current option style?

```d
getopt(args,
    std.getopt.config.caseSensitive,
    "foo", &foo,
    std.getopt.config.hideFromOverview,
    "bar", &bar);
```

Note that the `caseSensitive` option already works...

zorael:
Indeed, the original PR was to add such a flag (std.getopt.config.hidden). The no-string solution was @andralex's idea to work around having to add a new flag by using an empty string as description instead. I don't dislike it if that's the way we want to go, but it needs some dancing around null if we don't want to clobber current behaviour.

quickfur:
Here's a better idea: declare an empty struct in `std.getopt`, say something like this:
````
struct Hidden { }
static immutable hidden = Hidden();
````

and then recognize instances of this struct as a flag to mark a particular option hidden, e.g.:
````
getopt(args,
    "someOption", &someOption,
    "hiddenOption", std.getopt.hidden, &hiddenOption,
    "someOtherOption", &someOtherOption
);
````

In general, we should try our best to avoid relying on the difference between `null` and an empty but non-null string.

quickfur:
Ah nevermind, that was the original implementation and Andrei wanted something else. Fair enough.

quickfur:
ping @andralex 

How should we move this forward?  Should we go back to the original `Hidden` attribute, or should we do something with null / `length==0`?  Let's not let this rot in the queue.

jondegenhardt:
I was reluctant to contradict Andrei, but since this has been stalled for a bit - To me this has the feel of feature creep that is likely to grow further. I'm thinking specifically of LDC's `--help-hidden` option, which exposes the full set of options, including the hidden ones (not shown by `--help`). It's a logical next addition after implementing this. Other selection mechanisms might follow.

This leaves me wondering how much further to make incremental extensions to `std.getopt`, and when it's better to do a bit more of a redesign. This is very subjective thought. It's partly colored by my use of `getopt`. I think `getopt` serves it purposes really well, but I also think it is nearing the limits of what it does well with the current design.

andralex:
I think we shouldn't address the matter of hidden/undocumented flags as a feature of getopt. If an applications needs that they can simply call `getopt` twice (first time with passThrough) or do custom preprocessing to the arg list. It's an exceptional case.

>This leaves me wondering how much further to make incremental extensions to std.getopt, and when it's better to do a bit more of a redesign. 

I think getopt is in a sweet spot between simplicity, ease of use, and flexibility. I'd be curious to see how a redesign would make it significantly better. I encourage anyone considering working on such a project to focus instead on the many high-impact items to choose from.

quickfur:
I agree that the current design is nearing the limit of its usefulness. There's only so much you can do with a long variadic argument list before things just get completely out-of-hand.

Personally, I'm leaning towards a more declarative approach to parsing command-line options, via UDAs and compile-time introspection.  I've implemented something similar in the past (though not 100% the same), where usage is along these lines:

````
struct ProgramOptions
{
    @Desc("X resolution of output image") // detailed description
    @Alt("xres") // optional command-line shorthand
    int xResolution;

    @Desc("Y resolution of output image")
    @Alt("yres")
    int yResolution;

    @Nested // tells option parser to recursively inspect SubsystemOptions
    SubsystemOptions subsystem;

    @Direct // means value should come from args[i] directly
    string inputFile;
}
struct SubsystemOptions
{
    @Desc("Output format")
    @Alt("fmt")
    string outputFmt = "png"; // default arguments are displayed in help
}
void main(string[] args)
{
    ProgramOptions cfg;
    parseOptions(args, cfg); // updates cfg in-place with command-line values
    auto f = File(cfg.inputFile);
    auto xres = cfg.xResolution;
    auto yres = cfg.yResolution;

    // Nice compartmentalization of subsystem options so that subsystem modules
    // won't have unwanted dependencies on global program options.
    configureSubsystem(cfg.subsystem);
}
````

An example command-line would be something like:
````
program -xres 50 -yres 50 -fmt=jpg inputdata.txt
````

The nice thing about using UDAs and declarative structs is that everything is in one place; no fragile editing of overly long variadic arguments and unreadable clusters of meta-arguments, completely automated help text generation, no need to declare variables to hold the result of options -- the option value is directly assigned into struct fields. The `@Nested` UDA lets you organize options for specific subsystems separately without clutter or leaking options across subsystem boundaries -- just pass `cfg.subsystem` to the subsystem containing only the settings relevant to it, instead of polluting the program with dependencies on global options everywhere.

You can even substitute different option parsers that populate the same structs using the same interpretations of the UDAs, e.g., from JSON retrieved over the network, or from a database backend, or even a GUI configuration page. The help text can be used to automatically generate tooltips / GUI help pages, for example.  Or a network JSON API can be automatically reply to an info request with a response describing option names, description text, etc..  The program becomes free from an explicit dependency on command-line arguments.

Furthermore, it is much more amenable to extensions, e.g., you could have a UDA that describes the range of allowed values for numerical parameters, etc., and the `parseOptions` could automatically enforce that for you, without needing you to explicitly write range-checking code. The possibilities are endless.

andralex:
For my money, the more I see such heavy-handed approaches to command line parsing, the more I like getopt.

jondegenhardt:
> I think getopt is in a sweet spot between simplicity, ease of use, and flexibility. I'd be curious to see how a redesign would make it significantly better.

I agree with this. `getopt` hits a very nice sweet spot. At this point I've used it quite a bit, I do have suggestions about useful changes. I have been wondering about a productive venue for describing them.

I wouldn't want to infer that my ideas should be given high priority. More that should someone start looking at updates to `getopt`, they might benefit from reading some of the specifics I've identified. Neither the forums nor the bugzilla enhancement feel like a productive place to put this, but I could be wrong. Thoughts?

quickfur:
I agree that `getopt` suffices for simple one-off programs that have a simple set of options.

As soon as you want to provide a nicer interface like printing a list of option descriptions, or managing a very large set of possible options, sooner or later you begin to approximate the declarative approach.  The ad hoc approach of `getopt` is OK for small programs with small option sets, but it simply doesn't scale.

quickfur:
@jondegenhardt On the contrary, I think the forum is the perfect place for this sort of discussion.

quickfur:
@andralex Anyway, coming back on topic, should we just close this PR then?

andralex:
I'd recommend anyone considering improving getopt the following list in decreasing order of importance:

* Work on things related to shared, concurrency, parallelism
* Work on safety, e.g. make more functions in phobos safe
* Work on gc, e.g. make more functions in phobos `@nogc`
* Work on closing literally any bug we have
* ... a million other things ...
* add a dub package with an alternative to getopt

jondegenhardt:
> I'd recommend anyone considering improving getopt the following list in decreasing order of importance: ...

Fully agree. That's why I'm concerned about venue for recording what I have found. It's easy for people to draw the conclusion that an immediate change is being recommended.

PR #5825, opened 2017-10-29
std.math: Add RealRep union for accessing floating point bits via a union.

ibuclaw:
As I'd like to see this move forwards, but am in the middle of finishing up work on new ports (making sure that 128-bit is working properly for both IEEE and IBM reals) so can't really commit at the moment.

This is a slightly more conservative version of #4336.  As far as I can see, GDC at least produces identical assembly in the conversion from old to new.

Should allow the bigger refactorings in the other PR to be added gracefully.  @tsbockman - would you agree?

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -256,6 +256,163 @@ enum RealFormat
     ieeeQuadruple,
 }
 
+// A union which permits converting between a floating point
+// and an integer, or array of integers for low level access.
+union RealRep(N)
+if (isFloatingPoint!N)
+{
+pure: @safe: nothrow: @nogc:
+pragma(inline, true):
+    // Field accessors:
+    //      value = the floating point value to manipulate.
+    //      bytes = view as an array of bytes.
+    //      shorts = view as an array of shorts.
+
+    // 32-bit, 64-bit and 80-bit reals only:
+    //      word = view as an unsigned integer.
+    //      sword = view as a signed integer.
+
+    // 64-bit, 80-bit and 128-bit reals only:
+    //      ints = view as an array of ints.
+    //      msw = view the most significant part.
+    //      lsw = view the least significant part.
+
+    // 80-bit real only:
+    //      exponent = view the exponent part.
+
+    // 128-bit reals only:
+    //      longs = view as an array of longs.
+
+    // 128-bit ibm real only:
+    //      high = view the underlying higher double format.
+    //      low = view the underlying lower double format.
+    Unqual!N value;
+    alias value this;
+
+    this (const N x) inout
+    {
+        this.value = x;
+    }
+
+    alias traits = floatTraits!N;
+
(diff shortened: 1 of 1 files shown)


tsbockman:
Unless someone has added floating-point `union` support to DMD since I last checked, this PR has the same problem which stalled mine: [it breaks some parts of `std.math` which currently work in CTFE](https://github.com/dlang/phobos/pull/4336#issuecomment-302196046).

It could be re-written to emulate a `union` with a `struct` and properties encapsulating some pointer code, but given that the response to [my efforts at extending the CTFE pointer repainting capabilities to support all basic operations](https://github.com/dlang/dmd/pull/6811) was fairly hostile, it didn't seem worth my time.

I think the compiler team needs to pick a long-term solution to the floating-point CTFE issue before it's worth refactoring `std.math`.

ibuclaw:
> it breaks some parts of std.math which currently work in CTFE.

I don't particularly buy that, as many functions that do pointer casting are not CTFE-able anyway.

Typical getter/setters look like this.
```
ushort *vu = cast(ushort *)&f;
vu[EXPPOS_SHORT] = 0;
```
The value of `vu[0]` might be inferable but any other index certainly isn't.

tsbockman:
Specifically, this will break `isInfinity` and `isNaN` at CTFE for `double` and `float`:
```D
import std.math, std.stdio;

void main() {
	enum float x = float.nan;
	enum a = isInfinity(x);
	enum b = isNaN(x);
	writeln(a, ", ", b);

	enum double y = double.infinity;
	enum c = isInfinity(y);
	enum d = isNaN(y);
	writeln(c, ", ", d);
}
```

Admittedly, these can both be reimplemented in a CTFE-compatible way by the user easily enough:
```D
bool isInfinity(X)(X x) { return x > X.max || x < -X.max; }
bool isNaN(X)(X x) { return x != x; }
```

But, it is still a breaking change. If that's considered acceptable, then why not just go with my original pull request?

ibuclaw:
I'm in the middle of testing ieeeQuadruple and ibmExtended, and finding bugs in both.  I'd want to make sure that is complete before introducing yours.  To make sure we start from a working implementation before the redesign.

tsbockman:
That makes sense.

quickfur:
ping @ibuclaw 

Please rebase?

ibuclaw:
Rebased.

wilzbach:
@ibuclaw maybe it's better if you rebase yourself this time?

ibuclaw:
Rebased, but I'll have to double-check that ibmExtended is still fine (did some educated guesswork on reimplementing the floorImpl I added in the new format).

wilzbach:
@ibuclaw would need another rebase.

LightBender:
PR closed as stalled. @thewilsonator would like to split this into two PR's. One for the new feature and another for the refactoring.

PR #5826, opened 2017-10-30
Fix Issue 11644 - EvictingStrategy.LRU for std.functional.memoize

jercaianu:
I implemented the O(1) algorithm for "get" and "set" from the cache similar to what is described here[1].

Thanks,
Alex

### DIFF
--- a/std/functional.d
+++ b/std/functional.d
@@ -1068,6 +1068,147 @@ template memoize(alias fun, uint maxSize)
     }
 }
 
+/// ditto
+template memoize(alias fun, uint maxSize, bool lru = true)
+{
+    import std.traits : ReturnType;
+    import std.typecons : Tuple;
+
+    class Node
+    {
+    public:
+        Node next;
+        Node prev;
+        Tuple!(Parameters!fun) args;
+        ReturnType!fun res;
+
+        this(Tuple!(Parameters!fun) args = Tuple!(Parameters!fun).init,
+             ReturnType!fun res = ReturnType!fun.init)
+        {
+            this.args = args;
+            this.res = res;
+        }
+
+        void removeNode()
+        {
+            Node prev = this.prev;
+            Node next = this.next;
+
+            prev.next = next;
+            next.prev = prev;
+        }
+
+        void addNode(Node head)
+        {
+            Node nextHead = head.next;
+
+            nextHead.prev = this;
+            head.next = this;
+            this.prev = head;
+            this.next = nextHead;
+        }
+
+        void moveNode(Node head)
+        {
+            removeNode();
+            addNode(head);
+        }
+    }
+
+    ReturnType!fun memoize(Parameters!fun args)
+    {
+        import std.typecons : Tuple;
+        import std.conv;
+
+        alias Args = Parameters!fun;
+
+        static Node head;
+        static Node tail;
+        static Node[Tuple!Args] cache;
+        static int capacity = maxSize;
+
+        if (capacity == maxSize)
+        {
+            head = new Node();
+            tail = new Node();
+
+            head.prev = tail;
+            head.next = tail;
+
+            tail.next = head;
+            tail.prev = head;
+        }
+
+        Node node = null;
+        auto t = Tuple!Args(args);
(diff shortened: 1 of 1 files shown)


MartinNowak:
Thanks for your PR.

The usage of a doubly linked list and classes bothers me a little from a performance perspective.
Could you use indices and a lazily pre-allocated array instead?

Also most of the issue has been addressed by the random replacement from the cuckoo hashing. Do you have an example that has severe performance issues with that approach?

andralex:
Where is the link to [1]?

std/functional.d
```diff
@@ -1068,6 +1068,146 @@ template memoize(alias fun, uint maxSize)
     }
 }
 
+/// ditto
+template memoize(alias fun, uint maxSize, bool lru = true)
```
andralex: I searched this page and found no use of `lru`.

std/functional.d
```diff
@@ -1068,6 +1068,146 @@ template memoize(alias fun, uint maxSize)
...
+{
+    import std.traits : ReturnType;
+    import std.typecons : Tuple;
+
+    class Node
```
andralex: A class with dynamic allocation and linked in a doubly linked list is textbook-correct but certainly out of character for phobos and for high-performance applications. Per @MartinNowak we'd be looking at a properly managed contiguous array,

std/functional.d
```diff
@@ -1068,6 +1068,146 @@ template memoize(alias fun, uint maxSize)
...
+    {
+    public:
+        Node next;
+        Node prev;
+        Tuple!(Parameters!fun) args;
+        ReturnType!fun res;
+
+        this(Tuple!(Parameters!fun) args = Tuple!(Parameters!fun).init,
+             ReturnType!fun res = ReturnType!fun.init)
+        {
```
andralex: Hmmm, the defaults here are suprising - most functions will NOT return the default result for the default parameter values. Where are the defaults used?

std/functional.d
```diff
@@ -1068,6 +1068,146 @@ template memoize(alias fun, uint maxSize)
...
+    ReturnType!fun memoize(Parameters!fun args)
+    {
+        import std.typecons : Tuple;
+
+        alias Args = Parameters!fun;
+
+        static Node head;
+        static Node tail;
+        static Node[Tuple!Args] cache;
+        static int capacity = maxSize;
```
andralex: `size_t` probably

andralex:
By the way LRU sux we should use LIRS: https://en.wikipedia.org/wiki/LIRS_caching_algorithm

MartinNowak:
It's very welcome to review and properly benchmark the current approach.
https://github.com/dlang/phobos/pull/2591
If we find this to be inadequate for certain usages, we could look at different replacement policies.

Not too certain we really need a fancy eviction policy here, since we have a uniformly distributed hash (MD5), random replacement should be fairly OKish, and just using a slightly bigger cache will fix performance problems.
Of course a better eviction policy can further optimize for cache size vs. hit ratio.

If we go that route we should replace the current cuckoo hash implementation.

Hopefully we can find a general purpose replacement policy. Making it selectable via a template parameter does increase the complexity of `memoize` even further.

ben-manes:
LIRS is very complicated to implement correctly. Most implementations miss key details that degrades the hit rates, have memory leaks due to the paper neglecting to bound the ghost entries (as done in their C code), and may have jitter due to slow stack pruning if bounded inefficiently. However, it does have an excellent hit rate across many workloads and can be speedy.

After doing a deep dive into approaches for my caching library, I chose [TinyLFU](https://github.com/ben-manes/caffeine/wiki/Efficiency) with some minor adaptions. It is very easy to implement, time/space efficient, and beats LIRS in many real-world workloads. As an admission policy, it can be paired with any eviction policy (such as an array-based approach rather than linked lists - e.g. sampled LRU).

jercaianu:
@ben-manes @MartinNowak
Thanks for the comments!
I have left this issue on the backburner, since I've been working more on allocators.
I will definitely take a look at TinyLFU.

Alex

LightBender:

PR #5827, opened 2017-10-30
std/process.test fails if ran as root

jercaianu:
This test checks that we are unable to enter a directory if we don't have the right permissions.
However root can chdir to a directory, despite it not having execute permissions.

The solution would be to skip this test if ran as root.
Thanks,
Alex

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -1188,13 +1188,16 @@ version (Posix) @system unittest
     // can't run in directory if user does not have search permission on this directory
     version(Posix)
     {
-        import core.sys.posix.sys.stat : S_IRUSR;
-        auto directoryNoSearch = uniqueTempPath();
-        mkdir(directoryNoSearch);
-        scope(exit) rmdirRecurse(directoryNoSearch);
-        setAttributes(directoryNoSearch, S_IRUSR);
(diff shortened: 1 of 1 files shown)


CyberShadow:
Looks good, but please copy the pull request title and description (sans signature) to the commit message. "check root" is anything but descriptive.

jercaianu:
@CyberShadow 
Thanks for the comment!
Is it alright now?

Alex

CyberShadow:
@jercaianu Please have a look at https://chris.beams.io/posts/git-commit/. The first line should not be longer than 70 characters (preferably 50).

CyberShadow:
Here is the commit message I would use:

```
std.process: Skip permission test when running as root

A spawnProcess unittest checks that we are unable to enter a directory
if we don't have the right permissions.  However, programs running
under the root user can chdir to a directory, despite it not having
execute permissions.

Thus, skip this test when running as root.
```

jercaianu:
@CyberShadow 
Third time does the trick?

RazvanN7:
Looks good. LGTM

edi33416:
Great catch! This is subtle and gave me a lot of pain while testing on FreeBSD.

PR #5829, opened 2017-10-31
Fix Quantizer alignedAllocate typo

edi33416:
It looks like `alignedAllocate` had some residues, probably from a copy-paste.

Fixed the typos and added an unittest.

### DIFF
--- a/std/experimental/allocator/building_blocks/quantizer.d
+++ b/std/experimental/allocator/building_blocks/quantizer.d
@@ -85,9 +85,9 @@ struct Quantizer(ParentAllocator, alias roundingFunction)
     $(D parent.alignedAllocate(goodAllocSize(n), a)).
     */
     static if (hasMember!(ParentAllocator, "alignedAllocate"))
-    void[] alignedAllocate(size_t n, uint)
+    void[] alignedAllocate(size_t n, uint a)
     {
-        auto result = parent.alignedAllocate(goodAllocSize(n));
+        auto result = parent.alignedAllocate(goodAllocSize(n), a);
         return result.ptr ? result.ptr[0 .. n] : null;
     }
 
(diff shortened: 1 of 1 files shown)


MetaLang:
Please change the title of the PR to be more descriptive of what is being fixed, and if possible provide a short summary.

std/experimental/allocator/building_blocks/quantizer.d
```diff
@@ -75,6 +75,7 @@ struct Quantizer(ParentAllocator, alias roundingFunction)
     */
     void[] allocate(size_t n)
     {
+        if (n == 0) return null;
```
MetaLang: Is this necessary? If parent.allocate returns `null`, then that should automatically be propagated by the ternary expression on line 80.

edi33416: This is might be necessary because `parent.allocate` might not return `null`, which is exactly what is happening with `Region.allocate`. Please see #5830.
Another reason would be to avoid calling `parent.allocate`, since this should return `null`.

What do you think?

edi33416:
How can the code coverage decrease when I actually add a unittest? 😄

std/experimental/allocator/building_blocks/quantizer.d
MetaLang: As far as I can tell from [the documentation](https://dlang.org/phobos/std_experimental_allocator_building_blocks.html), passing 0 to allocate should always return `null`, and if it doesn't, that's a bug and should be correct "upstream".

std/experimental/allocator/building_blocks/quantizer.d
```diff
@@ -171,21 +171,14 @@ struct Quantizer(ParentAllocator, alias roundingFunction)
     {
         if (!b.ptr)
         {
-            b = alignedAllocate(s);
+            b = alignedAllocate(s, a);
```
RazvanN7: This line is not covered. Test with a null **b**

RazvanN7:
@edi33416 

> How can the code coverage decrease when I actually add a unittest? :smile:
easy, your PR is +15| -12. You add 3 lines in previously not tested blocks and voila!

Sorry if it was a rhetorical question :))

PR #5830, opened 2017-10-31
Fix region allocate and expand 0 sz

edi33416:
Allocating `0` bytes should return `null`.
Expanding with `0` bytes should return `true`.


edi33416:
@andralex @MetaLang

std/experimental/allocator/building_blocks/region.d
```diff
@@ -125,7 +125,7 @@ struct Region(ParentAllocator = NullAllocator,
     {
         static if (growDownwards)
```
andralex: hoist the test and put it here right at the beginning: `if (n == 0) return null`

PR #5831, opened 2017-11-01

edi33416:


edi33416:
I left a bunch of comments in the PR asking this: should owns actually be defined as
`Ternary owns(const void[] b) const` ?
My opinion is that it should be.

PR #5832, opened 2017-11-01
Fix Issue 17952 - std.range.transposed save is invalid

Darredevil:
After discussing with @andralex we decided to deprecate this.
CC @schveiguy

### DIFF
--- a/changelog/std-range-package-Transposed-deprecate-save.dd
+++ b/changelog/std-range-package-Transposed-deprecate-save.dd
@@ -1,7 +1,7 @@
-`std.range.package.Transposed` deprecate `save`
+Deprecate `savce` for `std.range.package.Transposed`
 
-Transposed incorrectly is marked as a forward range. It cannot be used
-without affecting any other copies made with `save`. `save` will be
+Transposed is incorrectly marked as a forward range. Its `popFront` primitive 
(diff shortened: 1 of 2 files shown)


MetaLang:
The changelog explanation could be a bit longer.

schveiguy:
Massaged the docs a bit.

BTW, I forgot to make a commit message on the first update (did it in github). Not sure how I can fix that ☹️, but at least it's on a temporary file!

schveiguy:
Hm. The fact that there’s no doc changes worries me. Probably the docs about save (the warning in particular) should be moved to the transpose function doc.

schveiguy:
Or... maybe just ditto the Transposed struct similar to how other functions are in the same file.

andralex:
cc @MartinNowak

changelog/std-range-package-Transposed-deprecate-save.dd
```diff
@@ -0,0 +1,14 @@
+Deprecate `savce` for `std.range.package.Transposed`
```
wilzbach: Typo: `save`

wilzbach: Ah I didn't see the follow-up PR. Thanks @schveiguy!

PR #5833, opened 2017-11-01
Mark deallocate as nothrow @nogc

edi33416:
For allocators that define their own `deallocate`, this does not produce any garbage. All the other allocators must infer this from their parents.

### DIFF
--- a/std/experimental/allocator/mallocator.d
+++ b/std/experimental/allocator/mallocator.d
@@ -84,7 +84,7 @@ struct Mallocator
     {
         int* p = null;
         p = cast(int*) A.instance.allocate(int.sizeof);
-        scope(exit) A.instance.deallocate(p[0 .. int.sizeof]);
+        scope(exit) () nothrow @nogc { A.instance.deallocate(p[0 .. int.sizeof]); }();
         *p = 42;
         assert(*p == 42);
     }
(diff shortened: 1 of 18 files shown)


andralex:
rebase

MetaLang:
@CyberShadow how can the doc tester be re-triggered?

braddr:
This needs to be re-based for it to be buildable.  Removing the auto-merge label to avoid wasting builder time.

std/experimental/allocator/mallocator.d
```diff
@@ -84,7 +84,7 @@ struct Mallocator
     {
         int* p = null;
         p = cast(int*) A.instance.allocate(int.sizeof);
-        scope(exit) A.instance.deallocate(p[0 .. int.sizeof]);
+        scope(exit) () nothrow @nogc { A.instance.deallocate(p[0 .. int.sizeof]); }();
```
ZombineDev: This unittest was already marked with `@nogc nothrow`. Why are you wrapping that expression in a lambda?

edi33416: Ops.. sorry about that. Should I do a quick PR for this, or I'll clean it along with a new one?

andralex: @edi33416 that'd be great thx

PR #5834, opened 2017-11-01
Followup to #5832

schveiguy:
Original PR #5832


schveiguy:
ping @Darredevil @andralex Just a couple of minor fixups that didn't make it into the original.

andralex:
thx!

PR #5835, opened 2017-11-02
Add std.concurrency.Generator overload to accept Fiber's stack guard'…

nemanja-boric-sociomantic:
…s size

As of 2.075 Fiber's stack protection pages are on by default, preventing
OS to merge mmaped ranges, so the total number of mmaped ranges is much
higher than before. In case there's no need for the stack overflow
protection and the high number of fibers, it may be wise to change this
setting, so the Generator's constructor overload is added which supports
this.

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -1533,6 +1533,27 @@ class Generator(T) :
     }
 
     /**
+     * Initializes a generator object which is associated with a static
+     * D function.  The function will be called once to prepare the range
+     * for iteration.
+     *
+     * Params:
+     *  fn = The fiber function.
+     *  sz = The stack size for this fiber.
+     *  guardPageSize = size of the guard page to trap fiber's stack
+     *                  overflows. Refer to $(REF Fiber, core,thread)'s
+     *                  documentation for more details.
+     *
+     * In:
+     *  fn must not be null.
+     */
+    this(void function() fn, size_t sz, size_t guardPageSize)
+    {
+        super(fn, sz, guardPageSize);
+        call();
+    }
+
+    /**
      * Initializes a generator object which is associated with a dynamic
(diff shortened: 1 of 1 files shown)


nemanja-boric-sociomantic:
cc @MartinNowak

schveiguy:
Need delegate version too.

nemanja-boric-sociomantic:
Ah, good catch, thanks! Updated.

schveiguy:
If a unit test won't always pass, we shouldn't add it. Otherwise, it causes spurious failures for other PRs.

Hackerpilot:
This should be enough to work around my use case's problem.

Burgos:
Ping @MartinNowak :-)

schveiguy:
So I have a question here. This targets stable, but is introducing new symbols. Is that OK?

nemanja-boric-sociomantic:
That's a good question. As we already missed the train to fix it in the last release, I have no problem targeting master instead.

wilzbach:
>> This targets stable, but is introducing new symbols. Is that OK?
> As we already missed the train to fix it in the last release, I have no problem targeting master instead.

Yeah, I think this would be the "safer" way ;-)

nemanja-boric-sociomantic:
Changed the base.

MartinNowak:
Quite an ugly way to deal with default values in the base class, but anyhow.

MartinNowak:
This addition is API/ABI compatible with existing code, so it would be fine for a minor release, but currently we only do major and patch releases.
https://forum.dlang.org/post/vsepgetbiovfiboklsbk@forum.dlang.org

nemanja-boric-sociomantic:
> Quite an ugly way to deal with default values in the base class, but anyhow.

Do you have any suggestion for an alternative approach?

PR #5838, opened 2017-11-03
Fix issue 17962: dirEntries is truncating Unicode file names.

jmdavis:
This fixes a regression that was introduced in https://github.com/dlang/phobos/pull/5758. Fortunately, looking at the history in git, I'm pretty sure that it has not yet made it into a release.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -3093,12 +3093,13 @@ else version(Posix)
             _dTypeSet = false;
         }
 
-        private this(string path, core.sys.posix.dirent.dirent* fd) @trusted
+        private this(string path, core.sys.posix.dirent.dirent* fd) @safe
         {
-            import std.path : buildPath;
             import std.algorithm.searching : countUntil;
+            import std.path : buildPath;
+            import std.string : representation;
 
-            immutable len = fd.d_name[].countUntil(0);
+            immutable len = fd.d_name[].representation.countUntil(0);
             _name = buildPath(path, fd.d_name[0 .. len]);
 
             _didLStat = false;
@@ -4207,6 +4208,33 @@ auto dirEntries(string path, string pattern, SpanMode mode,
             assert(de.attributes == de.statBuf.st_mode);
         }
     }
+}
+
+// Bugzilla 17962 - Make sure that dirEntries does not butcher Unicode file names.
+@system unittest
+{
+    import std.algorithm.comparison : equal;
+    import std.algorithm.iteration : map;
+    import std.algorithm.sorting : sort;
+    import std.array : array;
+    import std.path : buildPath;
+    import std.uni : normalize;
+
+    // The Unicode normalization is required to make the tests pass on Mac OS X.
+    auto dir = deleteme ~ normalize("𐐷");
+    scope(exit) if (dir.exists) rmdirRecurse(dir);
+    mkdir(dir);
+    auto files = ["Hello World",
+                  "Ma Chérie.jpeg",
+                  "さいごの果実.txt"].map!(a => buildPath(dir, normalize(a)))().array();
+    sort(files);
+    foreach (file; files)
+        write(file, "nothing");
+
+    auto result = dirEntries(dir, SpanMode.shallow).map!(a => a.name.normalize()).array();
+    sort(result);
+
+    assert(equal(files, result));
 }
 
 


d-random-contributor:
The difference is that countUntil doesn't go out of bounds, but strlen does.

d-random-contributor:
Then you slice `d_name` with len bigger than it is and use the resulting string.

std/file.d
```diff
@@ -3096,9 +3096,8 @@ else version(Posix)
         private this(string path, core.sys.posix.dirent.dirent* fd) @trusted
         {
             import std.path : buildPath;
-            import std.algorithm.searching : countUntil;
 
-            immutable len = fd.d_name[].countUntil(0);
```
MetaLang: I'm not clear on why calling `countUntil` truncates the file name. Is it somehow due to autodecoding?

jmdavis: Yes. `countUntil` uses ranges, so it auto-decodes and counts code points, whereas `strlen` just counts the number of chars until it hits the null character.

jmdavis:
> The difference is that countUntil doesn't go out of bounds, but strlen does.

I don't follow. Why would `strlen` go out-of-bounds? It counts until it encounters the null character, which the char* is going to have following the name unless the OS screwed up. Are you suggesting that that the OS doesn't guarantee that the null character is at the end of the name? As far as I can tell from the man pages, it does make that guarantee, and it would be very weird if it didn't.

aG0aep6G:
> Are you suggesting that that the OS doesn't guarantee that the null character is at the end of the name?

If you want the constructor to be properly safe, you can't assume that the `dirent` is set up that way. You can only rely on the definition of the struct.

jmdavis:
> If you want the constructor to be properly safe, you can't assume that the dirent is set up that way. You can only rely on the definition of the struct.

It's a private constructor. There's no reason that we can't guarantee where the `dirent` comes from.

aG0aep6G:
> It's a private constructor. There's no reason that we can't guarantee where the dirent comes from.

That doesn't make the constructor itself any more safe. Either make it properly safe, or mark the calls `@trusted` instead of the constructor.

Or, if you guys decide that misapplying `@trusted` is somehow ok here, at least explain in a comment that it's actually only safe when `fd.d_name` contains a null.

jmdavis:
Sometimes, `@trusted` does depend on the arguments being valid, but the only functions that call this constructor are already marked with `@trusted`. I really don't care about how `@trusted` is dealt with here so long as it's actually `@safe`, what is being done here with `dirent` is perfectly safe. The only place that it comes from is the system, and its API requires that the name be null terminated.

aG0aep6G:
> Sometimes, @trusted does depend on the arguments being valid,

That's always breaking the rules, and should at least be commented.

> but the only functions that call this constructor are already marked with @trusted

Then there's no point in the constructor being `@trusted`, and it shouldn't be.

quickfur:
Walter has stated very clearly before that the only time a function should be marked `@trusted` is if *there is no combination of arguments* that will lead to unsafe memory access, i.e., the function API should be inherently safe.

This doesn't seem to be the case here, since one could conceivably create a `dirent` struct, or modify one returned by the OS, that has a non-null-terminated filename.  So the ctor here ought not to be marked `@trusted`. 

In perspective, however, I checked all calls to the `DirEntry` ctor in this file and found that the only two places where it gets called is from functions that are already `@trusted`.  Therefore, the ctor does not need to be marked `@trusted` at all, and probably should instead be marked `@system` because of its inherently-unsafe API.  This will reduce the number of needless usages of `@trusted` in Phobos.

jmdavis:
> Walter has stated very clearly before that the only time a function should be marked @trusted is if there is no combination of arguments that will lead to unsafe memory access, i.e., the function API should be inherently safe.

Except that we consistently depend on stuff like the pointer pointing to valid memory, so it's not clearcut, and there have been plenty of arguments on the matter. But as I said, I really don't care here so long as the public API is correctly `@safe` and the code is doing the right thing, and when someone started marking up the code as `@trusted`, for some reason, they switched to using `countUntil` - and while still marking the constructor as `@trusted`, so it really makes no sense to me. `strlen` on the other hand does the right thing.

> In perspective, however, I checked all calls to the DirEntry ctor in this file and found that the only two places where it gets called is from functions that are already @trusted.

Yes, as I said, the callers are all `@trusted`. I'm fine with removing `@trusted` from the constructor. I don't care. Since it's only called from `@trusted` functions, the `@trusted` on the constructor doesn't really matter anyway. So, this code has no memory safety issues. What I care about is that the code is doing the right thing, and right now, it isn't, and it broke my stuff in the process.

jmdavis:
Okay. The constructor is not marked `@system` instead of `@trusted`.

CyberShadow:
No reason for either `@system` or `strlen`, just use `.representation.countUntil`.

schveiguy:
> just use `.representation.countUntil`

Was just about to come here and say that. Do we even need `@trusted` here if we change to this? I think it can be `@safe`

jmdavis:
I've got to say that it just seems bizarre to avoid using `strlen` for exactly what `strlen` is for, but whatever. It's now using `representation.countUntil`.

schveiguy:
> it just seems bizarre to avoid using `strlen` for exactly what `strlen` is for

Using `strnlen` would be safe as well, but the idea here is to not assume that all parameters have been properly set up by outside callers. IMO, we should never use `strlen`.

jmdavis:
> Using strnlen would be safe as well, but the idea here is to not assume that all parameters have been properly set up by outside callers. IMO, we should never use strlen.

If we were talking about a public function, I might agree, but this is a private function where it's well-known where the arguments come from. So, honestly, I think worrying about it here is a bit paranoid, but what I really care about is that the code works, and `representation.countUntil` works even if I'm inclined to think that it's a bit silly.

CyberShadow:
> If we were talking about a public function, I might agree, but this is a private function where it's well-known where the arguments come from.

Good point, but we've had requests to make some private symbols public in the past (e.g. the `DirEntry` constructor or `statTimeToStdTime`), and such subtleties may slip through the review of such a change.

d-random-contributor:
As to the philosophy: 
* `@trusted` depends on the arguments being constructable in `@safe` code and not worse than that; unexpected argument should be constructed manually, but it's not an unsafe operation.
* safety is checked on function granularity (so was my analysis), private access sort of alleviates the problem, but broadens the context to the module granularity.
* `strnlen` would be fine indeed.

schveiguy:
Seems like everyone is OK with this, so...

PR #5839, opened 2017-11-03
win64: build C files with /GS- to avoid "security" calls

rainers:
This avoids some symbol references into the VC compiler runtime and makes it easier to replace it with another runtime.
I suspect there are no similar security checks on other platforms.

### DIFF
--- a/etc/c/zlib/win64.mak
+++ b/etc/c/zlib/win64.mak
@@ -13,7 +13,7 @@ LDFLAGS=/nologo
 O=.obj
 
 # do not preselect a C runtime (extracted from the line above to make the auto tester happy)
-CFLAGS=$(CFLAGS) /Zl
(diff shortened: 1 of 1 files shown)


MartinNowak:
I'm somewhat uneasy about this @rainers, any chance we could stub the necessary methods in toolchains that lack them?
Disabling protections in a library that process user input (zlib) and might not get updated regularly sounds like a bad idea (thx to Iain for the recent updates, https://github.com/dlang/phobos/commits/master/etc/c/zlib/README).
https://www.zlib.net/ChangeLog.txt
[-GS (Buffer Security Check)](https://msdn.microsoft.com/en-us/library/8dbf701c.aspx)

rainers:
@MartinNowak I don't think we need to be very pessimistic here. zlib is a pretty well tested library. I suspect there are no such tests on any other platform.

PR #5842, opened 2017-11-03
Fix Issue 17467 - BitArray are broken with <<= 64

Darredevil:


Darredevil:
PS. Fixed rollRight as well.

RazvanN7:
LGTM.

std/bitmanip.d
```diff
@@ -1979,7 +1979,10 @@ public:
     }
     body
     {
-        return (upper << (bitsPerSizeT - nbits)) | (lower >> nbits);
+        if (nbits == 0)
+            return lower;
+        else
```
andralex: remove

std/bitmanip.d
```diff
@@ -2014,7 +2017,10 @@ public:
     }
     body
     {
-        return (upper << nbits) | (lower >> (bitsPerSizeT - nbits));
+        if (nbits == 0)
+            return upper;
+        else
```
andralex: remove

Darredevil: Fixed.

PR #5844, opened 2017-11-03
Do not expose unmangled ZLIB_VERSION, Z_NULL and ZLIB_VERNUM needlessly

mathias-lang-sociomantic:
Before this change, ZLIB_VERSION and ZLIB_VERNUM were unmangled TLS variables.
That exposure was unneeded, and could lead to some clash with other zlib binding which would prevent linking.

### DIFF
--- a/etc/c/zlib.d
+++ b/etc/c/zlib.d
@@ -228,7 +228,8 @@ enum
 }
 /* The deflate compression method (the only one supported in this version) */
 
-const int Z_NULL = 0;  /* for initializing zalloc, zfree, opaque */
+/// for initializing zalloc, zfree, opaque (extern(D) for mangling)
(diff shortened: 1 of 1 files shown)


CyberShadow:
Looks good, but I would prefer to either use `enum` or `extern(D)`, as moving the definitions might make updating the file a bit more difficult (I assume the definitions here are in the same order as in the original C header file).

mathias-lang-sociomantic:
I guess we can go with `extern (D)` then. I'm not comfortable with `enum` because it's turning an lvalue into an rvalue. While I doubt anyone rely on that, you never know :tm:
Will update tomorrow.

mathias-lang-sociomantic:
Updated

PR #5845, opened 2017-11-04
Remove trailing whitespace in transposed changelog entry

wilzbach:
https://github.com/dlang/dlang.org/pull/1921 is failing due to the trailing whitespace in the
changelog entry.

Ideally we should build the preview changelog again, s.t. new changelog files will be tested during the stable/master cutoff.


wilzbach:
Force-pushing this as it's as trivial as possible (just trailing whitespace) & it unblocks the dlang.org PR for 2.077.0 which should be online ASAP.

PR #5846, opened 2017-11-04
Bump stable DMD and DUB version

wilzbach:
With DUB 1.6.0 (part of the DMD 2.077.0 release), DUB will retry
downloads from the DUB registry and alternatively use fallback mirrors.
../dmd/src/posix.mak uses 2.072.0 atm, hence the bump of the host
version.

### DIFF
--- a/circleci.sh
+++ b/circleci.sh
@@ -2,8 +2,8 @@
 
 set -uexo pipefail
 
-HOST_DMD_VER=2.068.2 # same as in dmd/src/posix.mak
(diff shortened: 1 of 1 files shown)


CyberShadow:
From https://circleci.com/gh/dlang/phobos/4633:

> Package libdparse not found in registry at http://code.dlang.org/: HTTP request returned status code 502 (Bad Gateway)

Well, things are still broken, but at least we're getting sensible error messages, eh?

PR #5847, opened 2017-11-04
Fix bucketizer reallocate bug

edi33416:
`Bucketizer.reallocate` should check if `expand` was successful.


edi33416:
codecov was complaining because I didn't test for a successful expand. I had just updated the unittests, but not fast enough to prevent the merge 😄 
Should we leave it as it is?

MetaLang:
Submit a PR with a quick test case and I'll merge it.

edi33416:
@MetaLang cool: it's PR #5848

PR #5849, opened 2017-11-04
Fix bucketizer alignedAllocate bug

edi33416:
@andralex @MetaLang

### DIFF
--- a/std/experimental/allocator/building_blocks/bucketizer.d
+++ b/std/experimental/allocator/building_blocks/bucketizer.d
@@ -82,7 +82,7 @@ struct Bucketizer(Allocator, size_t min, size_t max, size_t step)
         {
             const actual = goodAllocSize(bytes);
             auto result = a.alignedAllocate(actual, alignment);
-            return result !is null ? (() @trusted => (&result)[0 .. bytes])() : null;
+            return result !is null ? (() @trusted => (&result[0])[0 .. bytes])() : null;
         }
         return null;
     }
@@ -155,7 +155,7 @@ struct Bucketizer(Allocator, size_t min, size_t max, size_t step)
             return true;
         }
         // Move cross buckets
-        return .alignedReallocate(this, b, size, a);
+        return common.alignedReallocate(this, b, size, a);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/building_blocks/bucketizer.d
```diff
@@ -82,7 +82,7 @@ struct Bucketizer(Allocator, size_t min, size_t max, size_t step)
         {
             const actual = goodAllocSize(bytes);
             auto result = a.alignedAllocate(actual, alignment);
-            return result !is null ? (() @trusted => (&result)[0 .. bytes])() : null;
+            return result !is null ? (() @trusted => (&result[0])[0 .. bytes])() : null;
```
edi33416: Took me a while to discover this bug, as, imo, it's easy to miss 😄

MetaLang: Can you explain for my benefit?

edi33416: The intent here is to return a slice of the underlying memory chunk (`result.ptr`). In other words, we want `result.ptr[0 .. bytes]`.

The bug is that `&result` is different from `result.ptr`; `&result` takes the address of the struct that holds the actual raw pointer and the length information.

The equivalent of `result.ptr` is `&result[0]`.

MetaLang: Okay, thought that was it but wasn't sure. I'm surprised that the struct's layout doesn't have the pointer first so this will Just Work.

std/experimental/allocator/building_blocks/bucketizer.d
```diff
@@ -271,3 +271,18 @@ struct Bucketizer(Allocator, size_t min, size_t max, size_t step)
...
 }
+
+// Test alignedAllocate
+@system unittest
+{
+    import std.experimental.allocator.building_blocks.bitmapped_block : BitmappedBlock;
+    import std.experimental.allocator.gc_allocator : GCAllocator;
+
+    Bucketizer!(BitmappedBlock!(64, 8, GCAllocator), 1, 512, 64) a;
+    for (auto i = 0; i < a.buckets.length; ++i) {
```
MetaLang: Braces on their own line

MetaLang:
```
foreach (ref bucket; a.buckets)
{
    //etc.
}
```

andralex:
cc @MetaLang will override in the interest of speed as I see your concerns have been addressed - holler if not thx!

andralex:
Codecov will be improved once the related prs go in, I'll override after the autotester passes.

andralex:
Auto-merge toggled on

PR #5851, opened 2017-11-09
alignedReallocate returns false on failure and FallbackAllocator needs stateless allocators for test

jercaianu:
alignedReallocate should return false if alignedAllocate fails.

This change however caused the FallbackAllocator tests to fail.
This is because testAllocator uses a FallbackAllocator with implicit constructor, which initializes its members, the primary and fallback allocators, to default values. 
Therefore stateful allocators used by the FallbackAllocator would be not properly initialized, causing tests to fail.

The change is to run tests for FallbackAllocator only for stateless allocators.

Thanks,
Alex

### DIFF
--- a/std/experimental/allocator/building_blocks/fallback_allocator.d
+++ b/std/experimental/allocator/building_blocks/fallback_allocator.d
@@ -401,6 +401,13 @@ fallbackAllocator(Primary, Fallback)(auto ref Primary p, auto ref Fallback f)
     assert(a.primary.owns(b2) == Ternary.no);
 }
 
+@system unittest
+{
+    import std.experimental.allocator.building_blocks.region : Region;
+    import std.experimental.allocator.gc_allocator : GCAllocator;
+    testAllocator!(() => fallbackAllocator(Region!GCAllocator(1024), GCAllocator.instance));
+}
+
 // Ensure `owns` inherits function attributes
 @system unittest
 {
--- a/std/experimental/allocator/building_blocks/fallback_allocator.d
+++ b/std/experimental/allocator/building_blocks/fallback_allocator.d
@@ -24,7 +24,8 @@ struct FallbackAllocator(Primary, Fallback)
     import std.traits : hasMember;
     import std.typecons : Ternary;
 
-    // Need at least one stateless allocator for this to work
+    // Need both allocators to be stateless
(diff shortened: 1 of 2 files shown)


jercaianu:
I also need to add some tests

std/experimental/allocator/building_blocks/fallback_allocator.d
```diff
@@ -24,8 +24,9 @@ struct FallbackAllocator(Primary, Fallback)
     import std.traits : hasMember;
     import std.typecons : Ternary;
 
-    // Need at least one stateless allocator for this to work
-    static if (!stateSize!Primary || !stateSize!Fallback)
+    // Need both allocators to be stateless
+    // This is to avoid using default initialized stateful allocators
+    static if (!stateSize!Primary && !stateSize!Fallback)
```
andralex: Clever idea. I suspect it won't run tests on some stateful allocators that would work properly, such as `InSituRegion`.

Can you try being a bit more specific? For example:

```
@system unittest
{
    Primary testing;
    if (testing.allocate(1) is null)
        return; // cannot test with degenerate primary allocator
    testAllocator!(() => FallbackAllocator());
}
```

jercaianu: Thanks for the comment, fixing

jercaianu: However, would this be a strong enough guarantee that the allocator is or isn't degenerate?
``testing.allocate`` in the case mentioned above could return not null and the allocator still be ill formed.

I think we should manually add tests for for stateful allocators used as parameters for the FallbackAllocator.

andralex:
Cool. Can you please look into the coverage? Looks like it went down a bit.

jercaianu:
@andralex 
Since the testAllocator is now run for statless allocators only, I manually added a test for a Region allocator, which is stateful, increasing the coverage.

std/experimental/allocator/common.d
```diff
@@ -391,6 +392,7 @@ bool alignedReallocate(Allocator)(ref Allocator alloc,
         if (b.length == s) return true;
```
edi33416: This should also check that `b` has the desired `alignment`.
Could you please add this test?

Thanks,
Edi

PR #5852, opened 2017-11-09
Fix Issue 15637 - Region allocator assert failure when expanding the last allocation

jercaianu:


jercaianu:

PR #5853, opened 2017-11-09
[Documentation] fold takes a seed first.

thewilsonator:


atilaneves:
`fold` takes a range first, that's the whole point of it existing and replacing `reduce`, which took the seed first and therefore couldn't be used with UFCS when a seed was needed.

PR #5854, opened 2017-11-09
Fixed Issue 16253 - BitmappedBlock allocator not working with chooseAtRuntime

jercaianu:
When passed chooseAtRuntime as a template parameter, BitmappedBlock receives an extra constructor parameter for the block size.

Thanks,
Alex


jercaianu:

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -186,6 +194,14 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
     }
 
     /// Ditto
+    static if (chooseAtRuntime == theBlockSize)
```
andralex: prefer `var == constant` in new code

PR #5855, opened 2017-11-10
Handle structs with @disable this in Nullable

FeepingCreature:
Currently, `Nullable` always constructs a default value for the type stored in it. This breaks the use of Nullable with types that `disable @this()`, for no good reason. If such types are detected, and the type in question does not use elaborate assignment (which would allow it to read uninitialized values via = void), then switch to `T _value = void;`. Otherwise, error.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2356,7 +2356,7 @@ struct Nullable(T)
     }
     else
     {
-        static assert(false, "Cannot construct Nullable("~T.stringof~"): type has no default constructor and overloaded assignment.");
+        static assert(false, "Cannot construct " ~ typeof(this).stringof ~ ": type has no default constructor and overloaded assignment.");
     }
 
     private bool _isNull = true;
@@ -2957,6 +2957,19 @@ auto nullable(T)(T t)
     }
     Nullable!TestToString ntts = new TestToString(2.5);
     assert(ntts.to!string() == "2.5");
+}
+
+// Bugzilla 14477
+@safe unittest
+{
+    static struct DisabledDefaultConstructor
+    {
+        @disable this();
+        this(int i) { }
+    }
+    Nullable!DisabledDefaultConstructor var;
+    var = DisabledDefaultConstructor(5);
+    var.nullify;
 }
 
 /**
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2357,7 +2357,7 @@ struct Nullable(T)
     else
     {
         static assert(false,
-                      "Cannot construct "~typeof(this).stringof~
+                      "Cannot construct " ~ typeof(this).stringof ~
                       ": type has no default constructor and overloaded assignment."
         );
     }
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2356,7 +2356,10 @@ struct Nullable(T)
     }
     else
     {
-        static assert(false, "Cannot construct " ~ typeof(this).stringof ~ ": type has no default constructor and overloaded assignment.");
+        static assert(false,
+                      "Cannot construct "~typeof(this).stringof~
+                      ": type has no default constructor and overloaded assignment."
+        );
     }
 
     private bool _isNull = true;
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2352,7 +2352,7 @@ struct Nullable(T)
     // that we're assigning to an uninitialized variable.
     else static if (!hasElaborateAssign!T)
     {
-        private T _value = void;
+        private T _value = T.init;
     }
     else
     {
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2343,7 +2343,22 @@ Practically $(D Nullable!T) stores a $(D T) and a $(D bool).
  */
 struct Nullable(T)
 {
-    private T _value;
+    // simple case: type is freely constructable
+    static if (__traits(compiles, { T _value; }))
+    {
+        private T _value;
+    }
+    // type is not constructable, but also has no way to notice
+    // that we're assigning to an uninitialized variable.
+    else static if (!hasElaborateAssign!T)
+    {
+        private T _value = void;
+    }
+    else
+    {
+        static assert(false, "Cannot construct Nullable("~T.stringof~"): type has no default constructor and overloaded assignment.");
+    }
+
     private bool _isNull = true;
 
 /**


std/typecons.d
```diff
@@ -2343,7 +2343,22 @@ Practically $(D Nullable!T) stores a $(D T) and a $(D bool).
...
+    }
+    // type is not constructable, but also has no way to notice
+    // that we're assigning to an uninitialized variable.
+    else static if (!hasElaborateAssign!T)
+    {
+        private T _value = void;
+    }
+    else
+    {
+        static assert(false, "Cannot construct Nullable("~T.stringof~"): type has no default constructor and overloaded assignment.");
```
BBasile: Prefer `typeof(this).stringof` since it will yield the whole thing including with the exclamation mark.

FeepingCreature:
I am deliberately *not* using emplace/destroy, as I believe they're an overcomplication that sank the previous pull request. Instead, the code checks for an elaborate assignment operator, for the simple reason that while the struct may not be initialized properly, without an assignment overload there is logically no way for this fact to become visible to user code. This does not cover every case, but it covers an additional case compared to the original setup, one that _really should_ be supported, and it does so in a safe fashion. You're right that I probably should be using `= T.init`, but I'm not convinced that this approach will be honored by the compiler in the future; `= void` may be unsafe, but that seems easier to work around with `@trusted`.

std/typecons.d
```diff
@@ -2944,6 +2959,17 @@ auto nullable(T)(T t)
     assert(ntts.to!string() == "2.5");
 }
 
+// Bugzilla 14477
+@safe unittest {
+    static struct DisabledDefaultConstructor {
```
BBasile: brace style !

FeepingCreature: oops, will fix tomorrow

std/typecons.d
```diff
@@ -2944,6 +2959,19 @@ auto nullable(T)(T t)
...
+// Bugzilla 14477
+@safe unittest
+{
+    static struct DisabledDefaultConstructor
+    {
+        @disable this();
+        this(int i) { }
+    }
+    Nullable!DisabledDefaultConstructor var;
+    var = DisabledDefaultConstructor(5);
```
wilzbach: Save the value and do an assert here.

std/typecons.d
```diff
@@ -2944,6 +2959,19 @@ auto nullable(T)(T t)
...
+    var.nullify;
+}
```
wilzbach: I think it would be a good idea to add a couple of more cases here, e.g. the one with an overloaded assignment, s.t. someone in the future doesn't accidentally "fix" the test.

std/typecons.d
FeepingCreature: What do you mean?

std/typecons.d
FeepingCreature: Wouldn't that be testing internal properties? Nullable with overloaded assignment *could* work, in theory, if e.g. emplace supported const; the fact that it doesn't work now is merely because I didn't want to put in the (significant) effort to support it correctly.

std/typecons.d
wilzbach: Sorry, the idea was to assert that the  custom constructor has been called

FeepingCreature: Seems a bit redundant - the custom constructor is just so I can make a value of that type at all, and the point of the unittest is to test Nullable, not @disable this or struct initialization.

FeepingCreature:
I'd really like to be able to trust = T.init to avoid the unsafetiness of =void, except for two things:

first, I'm half-worried = T.init will be declared unsafe itself,

second, alternately I'm half-worried that = T.init will go away at some point, since it's basically doing an end-run around the type system.

is .get unsafe? If so, I should test that and annotate it with @trusted.

JackStouffer:
Yes, I would use a `@trusted` lambda in `get` in the `!hasElaborateAssign!T` case

FeepingCreature:
In local testing, it seems to compile even without that? I am confused. Is that a bug? Are you sure @trusted ing the get is necessary?

JackStouffer:
For some reason I was absolutely positive that using void initialization was considered unsafe. Turns out it's not.

Otherwise, @MetaLang, do you see any other issues with this?

schveiguy:
> I was absolutely positive that using void initialization was considered unsafe

It is unsafe, but only for pointers or pointer-containing types.

FeepingCreature:
Can confirm. However, I don't see that there's _any_ way to construct a void initializer in that case. `= T.init` still seems to work, but that seems more like an oversight.

schveiguy:
` = T.init` is supposed to work, even for `@disable this()`. It's not an oversight.

https://dlang.org/spec/property.html#init

Note that even though it says it's "bad", it doesn't mean that it doesn't compile. However, you can have problems if there is an invariant (not sure that means anything here though).

MetaLang:
I still don't like the idea of this as it seems like a half measure, but I suppose it _does_ enable code that wasn't working before.

JackStouffer:
Uh, dscanner seg faulted. Is there anyway to restart the tests?

schveiguy:
@JackStouffer log into circle ci via github (as long as you are a dlang member I think), then you can restart. I did it for you.

std/typecons.d
```diff
@@ -2343,7 +2343,22 @@ Practically $(D Nullable!T) stores a $(D T) and a $(D bool).
...
+    }
+    // type is not constructable, but also has no way to notice
+    // that we're assigning to an uninitialized variable.
+    else static if (!hasElaborateAssign!T)
+    {
+        private T _value = void;
+    }
+    else
+    {
+        static assert(false, "Cannot construct " ~ typeof(this).stringof ~ ": type has no default constructor and overloaded assignment.");
```
JackStouffer: This line is too long. Needs to be < 120 characters.

FeepingCreature:
Maybe the `= T.init` case should be the default, but then Nullable would become incorrect for types with overloaded assignment but enabled default construction, unless we used moveEmplace which does not seem to work with inout. I agree that this is suboptimal, but the perfect solution does not seem to work yet.

std/typecons.d
```diff
@@ -2352,11 +2352,14 @@ struct Nullable(T)
...
     else static if (!hasElaborateAssign!T)
     {
-        private T _value = void;
+        private T _value = T.init;
     }
     else
     {
-        static assert(false, "Cannot construct " ~ typeof(this).stringof ~ ": type has no default constructor and overloaded assignment.");
+        static assert(false,
+                      "Cannot construct "~typeof(this).stringof~
```
MetaLang: This needs to be properly spaced, e.g.:

`"Cannot construct " ~ typeof(this).stringof ~`

FeepingCreature:
Ping?

MetaLang:
Did you use a shell command to add whitespace around all the concatenation operators? There's a ton of unrelated changes in your last commit that made the fix I requested, but also a lot more throughout the code. Could you remove these? We don't want to mess up git blame.

FeepingCreature:
Sorry, I was feeling a little snippish because more than half of std.typecons does not obey operator spacing. Fixed now. Well, unfixed.

MetaLang:
I understand, but most of `std.typecons` is old code that was written before we had the Phobos style guide. Any new code that we add should follow the Phobos guidelines.

FeepingCreature:
Ping?

BBasile:
approval needed i think otherwise auto-merge has no effect.

schveiguy:
I looked earlier, but figured others still needed to review. I didn't realize the auto-merge label was added already. LGTM.

PR #5856, opened 2017-11-11
Fix Issue 15140 - std.experimental.allocator.building_blocks.free_list.FreeList leaks memory

jercaianu:
I also slightly updated the documentation.

Thanks,
Alex


std/experimental/allocator/building_blocks/free_list.d
```diff
@@ -11,7 +11,9 @@ another allocator. Allocation requests between $(D min) and $(D max) bytes are
 rounded up to $(D max) and served from a singly-linked list of buffers
 deallocated in the past. All other allocations are directed to $(D
 ParentAllocator). Due to the simplicity of free list management, allocations
-from the free list are fast.
+from the free list are fast. If $(D adaptive) is set to $(D Yes.adaptive),
```
andralex: Use backquotes for single word

PR #5857, opened 2017-11-13
fix issue 16398 - experimental allocators, add aligned reallocation for Posix

BBasile:
The rationale for adding the primitive is that, like explained on bugzilla, the Windows version made for the _Digital Mars_ standard C library already included a "pseudo" alignedReallocate that actually just allocates a new block, copies the old one and frees it.

### DIFF
--- a/std/experimental/allocator/mallocator.d
+++ b/std/experimental/allocator/mallocator.d
@@ -68,16 +68,14 @@ struct Mallocator
 }
 
 ///
-@nogc nothrow
-@system unittest
+@nogc @system nothrow unittest
 {
     auto buffer = Mallocator.instance.allocate(1024 * 1024 * 4);
     scope(exit) Mallocator.instance.deallocate(buffer);
     //...
 }
 
-@nogc nothrow
-@system unittest
+@nogc @system nothrow unittest
 {
     @nogc nothrow
     static void test(A)()
@@ -91,8 +89,7 @@ struct Mallocator
     test!Mallocator();
 }
 
-@nogc nothrow
-@system unittest
+@nogc @system nothrow unittest
 {
     static void test(A)()
     {
@@ -280,16 +277,10 @@ struct AlignedMallocator
     else static assert(0);
 
     /**
-    On Posix, forwards to $(D realloc). On Windows, forwards to
-    $(D alignedReallocate(b, newSize, platformAlignment)).
+    Forwards to $(D alignedReallocate(b, newSize, platformAlignment)).
+    Should be used with bocks obtained with `allocate` otherwise the custom
+    alignment passed with `alignedAllocate` can be lost.
     */
-    version (Posix)
-    @system @nogc nothrow
-    bool reallocate(ref void[] b, size_t newSize) shared
-    {
-        return Mallocator.instance.reallocate(b, newSize);
-    }
-    version (Windows)
     @system @nogc nothrow
     bool reallocate(ref void[] b, size_t newSize) shared
     {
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/mallocator.d
```diff
@@ -318,6 +307,38 @@ struct AlignedMallocator
...
+        if (!s)
+        {
+            deallocate(b);
+            b = null;
+            return true;
+        }
+        auto p = alignedAllocate(s, a);
+        if (!p.ptr)
+        {
+            deallocate(b);
```
andralex: We don't want to deallocate the original memory here - just signal failure and let the client decide.

std/experimental/allocator/mallocator.d
```diff
@@ -318,6 +307,38 @@ struct AlignedMallocator
...
+            b = null;
+            return false;
+        }
+        else
+        {
+            if (b.ptr)
```
andralex: no need for this test - it works itself out

std/experimental/allocator/mallocator.d
```diff
@@ -318,6 +307,38 @@ struct AlignedMallocator
...
+            {
+                import std.algorithm.comparison : min;
+                const upTo = min(s, b.length);
+                p[0..upTo] = b[0..upTo];
```
andralex: space around `..`

std/experimental/allocator/mallocator.d
```diff
@@ -68,16 +68,14 @@ struct Mallocator
 }
 
 ///
-@nogc nothrow
-@system unittest
+@nogc @system nothrow unittest
 {
     auto buffer = Mallocator.instance.allocate(1024 * 1024 * 4);
```
andralex: I wonder why these lines appear green? Is it because they weren't covered and now they are?

std/experimental/allocator/mallocator.d
```diff
@@ -280,26 +277,18 @@ struct AlignedMallocator
...
     bool reallocate(ref void[] b, size_t newSize) shared
     {
         return alignedReallocate(b, newSize, alignment);
     }
 
     /**
-    On Posix, uses $(D alignedAllocate) and copies data around because there is
-    no realloc for aligned memory. On Windows, calls
-    $(HTTP msdn.microsoft.com/en-US/library/y69db7sx(v=vs.80).aspx,
+    On Posix this function is a helper that allocates and copies data around
```
andralex: The previous Posix-related explanation was more informative, even though it was incorrect :o).

std/experimental/allocator/mallocator.d
```diff
@@ -280,26 +277,18 @@ struct AlignedMallocator
...
+    because there is no realloc for aligned memory.
```
andralex: please put "realloc" in backquotes

std/experimental/allocator/mallocator.d
```diff
@@ -280,26 +277,18 @@ struct AlignedMallocator
...
+    On Windows, calls $(HTTP msdn.microsoft.com/en-US/library/y69db7sx(v=vs.80).aspx,
```
andralex: Use HTTPS so we're with the cool kids. Also the link has changed, now it's https://msdn.microsoft.com/en-us/library/y69db7sx.aspx

You may want to even try something like `$(LUCKY __aligned_realloc site:microsoft.com, text text text)`

std/experimental/allocator/mallocator.d
BBasile: I have not the CodeCov plugin enabled and i see it green too, so it's the standard diff color.

std/experimental/allocator/mallocator.d
```diff
@@ -280,26 +277,20 @@ struct AlignedMallocator
...
     bool reallocate(ref void[] b, size_t newSize) shared
     {
         return alignedReallocate(b, newSize, alignment);
     }
 
     /**
-    On Posix, uses $(D alignedAllocate) and copies data around because there is
-    no realloc for aligned memory. On Windows, calls
-    $(HTTP msdn.microsoft.com/en-US/library/y69db7sx(v=vs.80).aspx,
+    On Posix this function is non system call that allocates and copies old data
```
andralex: "this function is non system call" does not parse, I'll fix the text myself

std/experimental/allocator/mallocator.d
```diff
@@ -318,6 +309,33 @@ struct AlignedMallocator
...
+            deallocate(b);
+            b = null;
+            return true;
+        }
+        auto p = alignedAllocate(s, a);
+        if (!p.ptr)
+        {
+            return false;
+        }
+        else
```
andralex: drop

BBasile:
hmmm what did you do Andrei ? It was green.

andralex:
there was a whitespace

BBasile:
thanks, green again.

PR #5858, opened 2017-11-13
fix issue 12064 - std.typecons.wrap doesn't handle NVI

BBasile:


std/typecons.d
```diff
@@ -4684,16 +4684,24 @@ if (Targets.length >= 1 && allSatisfy!(isMutable, Targets))
             alias type = F;
         }
 
+        // Remove NVI members
+        template RemNVI(members...)
```
MetaLang: I don't think there's a need for such an abbreviation here. Please change this to something more descriptive like `removeNonVirtual`.

BBasile: Really ? can someone confirm, I'm afraid that `RemoveNonVirtualMembers` is a bit too much for a local helper template, also NVI is quite known. If you look at the comments inside the parent template there's one that contains NVI too.

BBasile:
Adressed the review comment but it seems that now there's a random error due to win32 sockets, see auto-tester.

PR #5859, opened 2017-11-13
hasLength should check unqualified length type

John-Colvin:
This silences the `pragma(msg,...)` for types like `const(size_t)` which are fine as range lengths.

### DIFF
--- a/std/range/primitives.d
+++ b/std/range/primitives.d
@@ -1397,7 +1397,8 @@ Deprecation: Historically `hasLength!R` yielded `true` for types whereby
 */
 template hasLength(R)
 {
-    static if (is(typeof(((R* r) => r.length)(null)) Length))
+    import std.traits : Unqual;
+    static if (is(Unqual!(typeof(((R* r) => r.length)(null))) Length))
     {
         static if (is(Length == size_t))
         {
--- a/std/range/primitives.d
+++ b/std/range/primitives.d
@@ -1435,10 +1435,12 @@ template hasLength(R)
 
     struct A { ulong length; }
     struct B { size_t length() { return 0; } }
-    struct C { @property size_t length() { return 0; } }
+    struct C { immutable(size_t) length() { return 0; } }
+    struct D { @property size_t length() { return 0; } }
     static assert( hasLength!(A));
     static assert( hasLength!(B));
     static assert( hasLength!(C));
+    static assert( hasLength!(D));
 }
 
 /**


andralex:
I'm not a fan of this, and I think we should become increasingly biased against such changes. It's "distinction without a difference" or adapted to this context "flexibility without usefulness".

People shouldn't return `immutable(size_t)` from a function. They could. They just shouldn't because it makes no practical sense. We shouldn't complicate library code to support this kind of useless flexibility.

ZombineDev:
(Removing the `auto-merge` to allow for discussion)

@andralex while I agree that defining `immutable(size_t) length()` doesn't make much sense, I think that defining `length` as field is a valid design choice. By extension `hasLength` should return `true` for such types regardless if they are qualified as `const` or `immutable`. Or alternatively ranges should be free to define the primitive as `immutable size_t length` if appropriate (i.e. the length doesn't change after construction).

John-Colvin:
@ZombineDev I'm pretty sure having a `const`/`immutable` `length` field makes no sense for a range.

@andralex I would have to think about it w.r.t. explicitly returning `immutable(size_t)`, but this change was motivated by this case:

```D
struct S
{
    private size_t _length;
    auto length() const
    {
        return _length;
    }
}
```

or more complex versions of that where the return type is being inferred from accessing a member from a const function.

jmdavis:
@John-Colvin The obvious counter to that is that if it's `length`, it should just return `size_t`, and there's really no reason to use `auto`.

Another issue is that having `length` return a non-mutable type can cause problems when `length` is called and `auto` is used on the caller's side. e.g.

    auto len = range.length;
    --len;

isn't going compile if `length` returns a non-mutable `size_t`, whereas it normally would. Now, given that `length` should always be `size_t`, the caller could use `size_t` explicitly too, but fewer places are likely to have to explicitly use `size_t` if it's forced on the function itself.

andralex:
Just have `length` return `size_t` instead of `auto`. I love offering flexibility, but we seem prone to err on the side of allowing just odd client code.

John-Colvin:
`auto` also means attribute inference. Of course, you can make it a no-arg template instead, but then that has other consequences.

Having thought about it, I'm ok with it either way. I can see the value in more tightly constraining what the range API actually is.

ZombineDev:
@John-Colvin Sorry for not being more concrete. Yes mutable ranges with immutable length do not make sense. Here's what use cases I had in mind:
1) Immutable or fixed-length containers, where the length is known only at run-time, but remains constant for the lifetime of the object.
2) Normal ranges where the length is defined as a field, e.g.:
```D
void main()
{
    auto r = countDown(10);
    use(r);
}

// Returns the range [start, start - 1, ..., 2, 1]
auto countDown(size_t start)
{
    struct Result
    {
        size_t length;
        void popFront() { --length; }
        bool empty() { return !length; }
        size_t front() { return length; }
    }

    return Result(start);
}

void use(T)(const ref T obj)
{
    print(obj);
}

void print(T)(ref T obj)
{
    import std.range.primitives : hasLength;
    import std.stdio : writefln;
    
    static if (hasLength!T)
    	writefln("Received a sequence of %s elements.", obj.length);
	else
        writeln("Received %s", obj);
}
```

John-Colvin:
@ZombineDev 

1) If the length of a range can't be changed, then it's not a range at all, because ranges require `popFront`. If `popFront` doesn't change `length` then that's an infinite range, which means it shouldn't define `length` anyway. The only edge case is for ranges that are permanently empty by construction, which seems special enough to require the user to do a bit of work.

2) In the case you've shown there,  `hasLength` isn't being called on a `const` type anyway, so I'm not sure I understand you.


The more interesting case is with `shared` ranges. I think this is a valuable concept that's been underexplored.

ZombineDev:
@John-Colvin 
> 1) If the length of a range can't be changed, then it's not a range at all, because ranges require popFront. If popFront doesn't change length then that's an infinite range, which means it shouldn't define length anyway*. The only edge case is for ranges that are permanently empty by construction, which seems special enough to require the user to do a bit of work.

This is all obvious to me, which is why I never used the word "range" for this point. I'm talking about containers.

> 2) In the case you've shown there, hasLength isn't being called on a const type anyway, so I'm not sure I understand you.

On the contrary. If you run the program you get [this](https://github.com/dlang/phobos/blob/v2.077.0/std/range/primitives.d#L1411) deprecation message and if you add `pragma (msg, T)` statement to the `print` function, the compiler will print `const(Result)` - see https://run.dlang.io/is/4r7Gk0.

John-Colvin:
@ZombineDev 

`std.range.primitives.hasLength` is a test of the range API. I'm not following why it matters here how a container that isn't a range defines its length.

w.r.t. the example, sorry, my mistake. I guess the answer would be that `hasLength` doesn't mean "does this type have a member `.length`", it means "does this type have a validly defined `.length` according to the range API". If the type you've passed in isn't a range (e.g. it doesn't work because it's `const`) then `hasLength` is fine to evaluate to `false`.

Somewhat relevant aside: Const objects can be ranges:
```D
int[][string] q;
class R
{
    string key;
    this(string key) const
    {
        this.key = key;
        q[key] = [3, 2, 1];
    }

    auto front() const { return q[key][0]; }
    bool empty() const { return q[key].length == 0; }
    void popFront() const { q[key] = q[key][1 .. $]; }
}

void main()
{
    import std.algorithm : equal;
    auto s = new const R("fdsa");
    assert(s.equal([3, 2, 1]));
}
```

jmdavis:
> The more interesting case is with shared ranges. I think this is a valuable concept that's been underexplored.

Actually, I don't think that it really matters. You really don't operate on `shared` objects normally, because the operations need to be atomic for that to work. If you had a `shared` range, you would need to lock a mutex to protect it and then cast away `shared` to actually operate on it (be that an explicit cast or implicit by the type system via something like `synchronized` classes if we ever get them). So, you wouldn't ever really be operating on a `shared` object, and I don't see why `shared` and ranges would then need to really coexist beyond being able to label a range as `shared` so that the data is shared across threads. The operations themselves wouldn't need to work as `shared`.

> Somewhat relevant aside: Const objects can be ranges:

Only because none of the actual state is `const`. That's just an example of why `const` is somewhat meaningless when `pure` is not involved.

quickfur:
Why can't we just `return cast() length;` to remove the immutable qualifier from the length?  It does seem pointless, as @andralex said, to return `immutable(size_t)`, since, being a value type, it shouldn't matter.

quickfur:
ping @ZombineDev @andralex @jmdavis @wilzbach 

This has been rotting in the queue. How should we proceed with this?  Should we close this, since it's deem superfluous?

andralex:
Time to close. Feel free to reopen if strong new motivation comes up.

PR #5862, opened 2017-11-16
Fix Issue 16564 - KRRegion.empty sometimes returns Ternary.no

jercaianu:
For region mode, the root node is not necessarily the one with the smallest address, therefore the empty condition would only be valid for free list mode.

I have added a variable which tracks the number of allocated bytes in region mode, to check if it is empty.

Also, I figured that on dellocateAll, since we reclaim all memory, we might as well switch back to region mode, since it's faster.

Thanks,
Alex

### DIFF
--- a/std/experimental/allocator/building_blocks/kernighan_ritchie.d
+++ b/std/experimental/allocator/building_blocks/kernighan_ritchie.d
@@ -172,6 +172,7 @@ struct KRRegion(ParentAllocator = NullAllocator)
     private void[] payload;
     private Node* root;
     private bool regionMode = true;
+    private size_t bytesUsedRegionMode = 0;
 
     auto byNodePtr()
     {
@@ -394,6 +395,7 @@ struct KRRegion(ParentAllocator = NullAllocator)
             if (root.size >= actualBytes)
             {
                 // Enough room for allocation
+                bytesUsedRegionMode += actualBytes;
                 void* result = root;
                 immutable balance = root.size - actualBytes;
                 if (balance >= Node.sizeof)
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/building_blocks/kernighan_ritchie.d
```diff
@@ -172,6 +172,7 @@ struct KRRegion(ParentAllocator = NullAllocator)
     private void[] payload;
     private Node* root;
     private bool regionMode = true;
+    private size_t bytesUsedRegionMode = 0;
```
andralex: Interesting approach. Could you please eliminate `regionMode`? Make it distinguished by `bytesUsedRegionMode == size_t.max`. So then:

```d
private bool regionMode() const { return bytesUsedRegionMode != size_t.max; }
private void cancelRegionMode() { bytesUsedRegionMode = size_t.max; }
```

std/experimental/allocator/building_blocks/kernighan_ritchie.d
```diff
@@ -490,9 +493,10 @@ struct KRRegion(ParentAllocator = NullAllocator)
             assert(pnode && pnode.next);
             assert(pnode != n);
             assert(pnode.next != n);
+
             if (pnode < pnode.next)
             {
-                if (pnode >= n || n >= pnode.next) continue;
+                if (pnode > n || n > pnode.next) continue;
```
andralex: what the heck was that???

jercaianu: That if was confusing, because memory that we want to deallocate should not be in the list (check the asserts above).

PR #5863, opened 2017-11-17
Fix Issue 17986 - Erratic failure with std/experimental/allocator/common.d(445): unittest failure

jercaianu:
alignedReallocate test assumed that if a pointer would be aligned to 'alignment', it wouldn't be aligned to 'alignment + 1'.
This is clearly wrong and it caused the test to fail at times.

It should be fine now and sorry for causing this bug,
Alex


WalterBright:
Thanks for the quick fix!

ZombineDev:
Thanks @jercaianu!

PR #5867, opened 2017-11-20
Fix issue 17995

Biotronic:


std/meta.d
```diff
@@ -487,7 +487,11 @@ template NoDuplicates(TList...)
     {
         enum steps = 16;
         alias first = NoDuplicates!(TList[0 .. steps]);
-        alias NoDuplicates = NoDuplicates!(EraseAllN!(first.length, first, TList[steps..$]));
+        alias NoDuplicates = NoDuplicates!(
```
MartinNowak: In case there are no duplicates, it looks like this would expand to the original arguments. The remaining work to perform is eliminating duplicates between the first and the second half below.

MartinNowak:
Bugfixes should always target stable btw.

MartinNowak:
That was still somewhat hard to understand and failed for e.g.
```d
static assert(NoDuplicates!(aliasSeqOf!(iota(350)), aliasSeqOf!(iota(350))) == aliasSeqOf!(iota(350)));
```
.
Changed this to a proper partition-and-conquer for both, NoDuplicates and the internal EraseAllN.
Now only uses `O(N * log(N))` template instances with a max recursion of `log(N)`.

MetaLang:
@MartinNowak while I'm sure the changes you made are good ones, you've essentially created a new PR and merged it yourself. I think it would've been better had you submitted an alternate PR.

MartinNowak:
> @MartinNowak while I'm sure the changes you made are good ones, you've essentially created a new PR and merged it yourself.

Yes the changes were quite big, but just adding them seems more appropriate than a "please change things like so ping-pong".

> I think it would've been better had you submitted an alternate PR.

Closing and superseding a PR can be somewhat demotivating for contributors, in particular as the tests and initial idea were close. Also we should have enough test coverage in phobos to not worry too much about pragmatic moves.

Biotronic:
> Closing and superseding a PR can be somewhat demotivating for contributors

I'd argue the same effect could be had by basically throwing away the entirety of the PR and replacing it. :p No worries though, I very much agree with the changes.

PR #5868, opened 2017-11-20
fix Issue 17996 - don't build libphobos2.a with PIC for i386

MartinNowak:
- use MODEL=64/32 to decide on default
- allow overriding default with explicit PIC=0/1 make args

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -54,12 +54,7 @@ ifneq ($(BUILD),release)
     endif
 endif
 
-# -fPIC is enabled by default and can be disabled with DISABLE_PIC=1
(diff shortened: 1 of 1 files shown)


MartinNowak:
The dscanner style_lint test segfaults? Is this somehow related to PIC?

MartinNowak:
Saw some mentionings of DScanner in the original PIC PRs @wilzbach.

wilzbach:
> The dscanner style_lint test segfaults? Is this somehow related to PIC?

It's a random failure that happens on every second build, but no ones has found time to look into it yet.

PR #5869, opened 2017-11-20
Replace `body` keyword with its replacement: `do`

wilzbach:
Automatic replacement with

    sed -i "s/^\([ ]*\)body/\1do/" -i **/*.d

Prompted by https://github.com/dlang/dlang.org/pull/1929:

> Even in phobos you still use body.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -69,7 +69,7 @@ ROOT_OF_THEM_ALL = generated
 ROOT = $(ROOT_OF_THEM_ALL)/$(OS)/$(BUILD)/$(MODEL)
 DUB=dub
 TOOLS_DIR=../tools
-DSCANNER_HASH=eb16fe4e8610e7477e09710fee1137d657a56f72
+DSCANNER_HASH=bb32e9f1e3e5206deb11b3dbc43ad44e23fabf96
 DSCANNER_DIR=../dscanner-$(DSCANNER_HASH)
 
 # Documentation-related stuff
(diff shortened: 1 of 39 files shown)


ZombineDev:
@wilzbach we need to upgrade DScanner first.

wilzbach:
>  we need to upgrade DScanner first.

Ok -> https://github.com/dlang/phobos/pull/5874

ZombineDev:
@wilzbach I'm not sure, but I don't think DScanner has support for the `do` keyword (in place of `body`) yet.

BBasile:
There's no emergency for the tools based on dparse. Once again, it's only a problem of communication. Specs have been update a too eagerly. Just updating the website should be okay, even if of course dparse needs an update too.

wilzbach:
> @wilzbach I'm not sure, but I don't think DScanner has support for the do keyword (in place of body) yet.

Ah. I wrongly remembered seeing such a PR, but adding support for the do keyword shouldn't be too difficult and we have to do it anyways, so I gave it a quick shot:

-> https://github.com/dlang-community/libdparse/pull/180

wilzbach:
As dlang-community/libdparse#180 was merged in light speed (thanks @BBasile !), I updated libdparse for the D-Scanner version used at Phobos (https://github.com/dlang-community/D-Scanner/pull/538) and now CircleCI is passing :tada: and this PR should be good to go :)

PR #5870, opened 2017-11-20

wilzbach:

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -619,28 +619,6 @@ private void _send(T...)(MsgType type, Tid tid, T vals)
  * matched by an earlier delegate.  If more than one argument is sent,
  * the $(D Variant) will contain a $(REF Tuple, std,typecons) of all values
  * sent.
- *
- * Example:
- * ---
- * import std.stdio;
- * import std.variant;
- * import std.concurrency;
- *
- * void spawnedFunction()
- * {
- *     receive(
- *         (int i) { writeln("Received an int."); },
- *         (float f) { writeln("Received a float."); },
- *         (Variant v) { writeln("Received some other type."); }
- *     );
- * }
- *
- * void main()
- * {
- *      auto tid = spawn(&spawnedFunction);
- *      send(tid, 42);
- * }
- * ---
  */
 void receive(T...)( T ops )
 in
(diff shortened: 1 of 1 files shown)


wilzbach:
Seems like the Jenkins instance is running out of space: 

```
/var/lib/jenkins/dlang_projects/rejectedsoftware/vibe.d/web/.dub/build/library-debug-linux.posix-x86_64-dmd_2077-1C2854E892D43B7F2EDB1600D2705677/: No space left on device

script returned exit code 2
```

std/concurrency.d
```diff
@@ -656,6 +634,36 @@ body
     thisInfo.ident.mbox.get( ops );
 }
 
+///
+@system unittest
+{
+    import std.variant;
+    __gshared string received;
```
ZombineDev: The documentation examples are supposed to show the best practices of using the provided functionality, in the specific case - means to avoid memory sharing and low-level synchronization by using the message passing paradigm. Please refrain from using global variables in the examples of this module, especially `__gshared` ones.

The examples should ideally look as if you could easily scale from threads, to process and to network nodes, i.e avoid sharing of state.

std/concurrency.d
```diff
@@ -656,6 +634,36 @@ body
...
+
+    static void spawnedFunction()
+    {
+        receive(
+            (int i) {
+                received = "Received an int.";
+                ownerTid.send(true); // signal success
```
ZombineDev: If you want to check that the functions are called, I suggest something along the lines of:
```D
auto process = ()
{
    receive(
        (int i) { ownerTid.send(1); },
        (double f) { ownerTid.send(2); },
        (Variant v) { ownerTid.send(3); }
    );
};

{
    auto tid = spawn(process);
    send(tid, 42);
    assert(receiveOnly!int == 1);
}

{
    auto tid = spawn(process);
    send(tid, 3.14);
    assert(receiveOnly!int == 2);
}

{
    auto tid = spawn(process);
    send(tid, "something else");
    assert(receiveOnly!int == 3);
}
```

Otherwise, it would be be better to keep the old examples.

wilzbach:
> The documentation examples are supposed to show the best practices of using the provided functionality,

Fair enough - rebased.

PR #5871, opened 2017-11-20

wilzbach:


ZombineDev:
On second thoughts, I decided to go a step further: https://github.com/dlang/phobos/pull/5875 - let's continue the discussion there.

PR #5872, opened 2017-11-20

wilzbach:

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -429,40 +429,60 @@ private template isSpawnable(F, T...)
  *  to $(D fn) must either be $(D shared) or $(D immutable) or have no
  *  pointer indirection.  This is necessary for enforcing isolation among
  *  threads.
- *
- * Example:
- * ---
- * import std.stdio, std.concurrency;
- *
- * void f1(string str)
- * {
- *     writeln(str);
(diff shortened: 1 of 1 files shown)


wilzbach:
Rebased to make the examples look nicer

std/concurrency.d
```diff
@@ -429,42 +429,55 @@ private template isSpawnable(F, T...)
...
+    }
+
+    auto tid = spawn(&f, "Hello World");
+}
+
+/// Fails: char[] has mutable aliasing.
+@system unittest
+{
+    static void f(char[] str) {}
+    static assert(!__traits(compiles, spawn(&f, msg.dup)));
```
JackStouffer: This isn't failing because of mutable aliasing; it's because `msg` doesn't exist.

Perhaps this would clearer

```d
char[] msg = "Hello, World!";

static void f(char[] str) {}

static assert(!__traits(compiles, spawn(&f, msg.dup)));
static assert(__traits(compiles, spawn(&f, msg.idup)));
```

wilzbach:
> This isn't failing because of mutable aliasing; it's because msg doesn't exist.

Ah good point, but then we need to show all possibilities to make it clear?
(rebased)

PR #5873, opened 2017-11-20

wilzbach:

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -1515,35 +1515,6 @@ private interface IsGenerator {}
 /**
  * A Generator is a Fiber that periodically returns values of type T to the
  * caller via yield.  This is represented as an InputRange.
- *
- * Example:
- * ---
- * import std.concurrency;
- * import std.stdio;
- *
- *
- * void main()
- * {
- *     auto tid = spawn(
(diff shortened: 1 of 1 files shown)


edi33416:
Any updates on this? I strongly agree that we should have working examples in the docs.

wilzbach:
Oh this example fails on 32-bit and my distro doesn't even support 32-bit anymore, so building a debug version of dmd, druntime and phobos is a bit of effort for which I hadn't had the time so far (and probably won't in the near future).
@edi33416 interested to dive into this?

edi33416:
@wilzbach I'm going to spin-up a 32bit machine and give it a go

wilzbach:
Cool! On a typical Linux distro, `gcc-multilib` is all you need, but be careful to use `AUTO_BOOTSTRAP=1` when building DMD.

jercaianu:
@edi33416  @wilzbach 
This is failing because the test contains a subtle bug.
`receiveOnly` expects a `size_t`, but the `Generator` is generating `ints`, causing `receiveOnly` to silently throw a `MessageMismatch` exception.

Either catch the exception, or change the `Generator` type to `uint`.

PR #5875, opened 2017-11-20
Improve std.concurrency.receiveOnly example

ZombineDev:
* Make the example a DDoc-ed unittest
* Show receiving scalars, as well as tuples

Alternative to #5871.

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -712,24 +712,6 @@ private template receiveOnlyRet(T...)
  *
  * Returns: The received message.  If $(D T.length) is greater than one,
  *          the message will be packed into a $(REF Tuple, std,typecons).
- *
- * Example:
- * ---
- * import std.concurrency;
- *
- * void spawnedFunc()
- * {
- *     auto msg = receiveOnly!(int, string)();
- *     assert(msg[0] == 42);
- *     assert(msg[1] == "42");
- * }
- *
- * void main()
- * {
- *     auto tid = spawn(&spawnedFunc);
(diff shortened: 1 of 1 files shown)


std/concurrency.d
```diff
@@ -765,6 +747,42 @@ do
         return ret;
 }
 
+///
+@system unittest
```
MetaLang: Why the `@system`?

ZombineDev: I tried `@safe` but `spawn` turned out to be system, though I haven't investigated further.

std/concurrency.d
```diff
@@ -765,6 +747,42 @@ do
...
+{
+    auto tid = spawn(
+    {
+        assert(receiveOnly!int == 42);
```
wilzbach: I agree that this way is cleaner, but how do you ensure that the `assert` was actually called?

ZombineDev: That's an example, the focus is on good looking code. We can add a non-documented real test if necessary.

std/concurrency.d
MetaLang: Yes but why the _explicit_ @system?

wilzbach: So that they can be found easily via grep and there is an active awareness for both the submitter and reviewers. That's why there is a rule that checks that a unittest is annotated with either `@safe` or `@system`

MetaLang: I did not know about that rule. I don't think I've ever ran into it before.

PR #5876, opened 2017-11-20
AscendingPageAllocator

jercaianu:
This is my work so far on a new allocator which is useful for large objects.
The allocator maps on construction a range of virtual adresses, without allocating any physical memory.

All allocations are rounded to a multiple of the page size and take place at increasing virtual adresses. Each allocation is mapped onto a number of virtual pages, for which we change the protection to read/write access. Physical memory is only allocated on demand, when we access the buffer.

On deallocation, we set the protection to none and use various syscalls to tell the OS to reclaim its memory.

The main advantage of this allocator is that it is safe, because we do not reuse memory and fast for large objects.
After some more tweaks I will also come back with a benchmark.

Future work: 
- integrate it with AllocatorList
- come up with a similar solution for smaller objects

Any advice is much appreciated!
Thanks,
Alex

### DIFF
--- a/std/experimental/allocator/building_blocks/ascending_page_allocator.d
+++ b/std/experimental/allocator/building_blocks/ascending_page_allocator.d
@@ -91,7 +91,7 @@ public:
 
     Params:
     n = Bytes to allocate
-    
+
     Returns:
     `null` on failure or if the requested size exceeds the remaining capacity.
     */
--- a/std/experimental/allocator/building_blocks/ascending_page_allocator.d
+++ b/std/experimental/allocator/building_blocks/ascending_page_allocator.d
@@ -121,7 +121,7 @@ public:
                 {
                     import core.sys.windows.windows : VirtualAlloc, MEM_COMMIT, PAGE_READWRITE;
 
-                    auto ret = VirtualAlloc(readWriteLimit, newReadWriteLimit - readWriteLimit, 
+                    auto ret = VirtualAlloc(readWriteLimit, newReadWriteLimit - readWriteLimit,
                         MEM_COMMIT, PAGE_READWRITE);
                     if (!ret)
                         assert(0, "Failed to allocate memory, VirtualAlloc failure");
--- a/std/experimental/allocator/building_blocks/ascending_page_allocator.d
+++ b/std/experimental/allocator/building_blocks/ascending_page_allocator.d
@@ -1,14 +1,15 @@
 module std.experimental.allocator.building_blocks.ascending_page_allocator;
 import std.experimental.allocator.common;
+
 /**
-`AscendingPageAllocator` is a fast and safe allocator which rounds all allocations
-to page size multiples. It reserves a range of virtual addresses (`mmap` for Posix
-and `VirtualAlloc` for Windows) and allocates memory at consecutive virtual
+`AscendingPageAllocator` is a fast and safe allocator that rounds all allocations
+to multiples of the system's page size. It reserves a range of virtual addresses 
+(using `mmap` on Posix and `VirtualAlloc` on Windows) and allocates memory at consecutive virtual
 addresses.
 
 When a chunk of memory is requested, the allocator finds a range of
-virtual pages which satisfy the requested size, changing their protection to
-read and write using OS primitives (`mprotect` and `VirtualProtect`).
+virtual pages that satisfy the requested size, changing their protection to
+read/write using OS primitives (`mprotect` and `VirtualProtect`, respectively).
 The physical memory is allocated on demand, when the pages are accessed.
 
 Deallocation removes any read/write permissions from the target pages
@@ -17,6 +18,9 @@ memory.
 
 Because the allocator does not reuse memory, any dangling references to
 deallocated memory will always result in deterministically crashing the process.
+
+See_Also:
+$(HTTPS microsoft.com/en-us/research/wp-content/uploads/2017/07/snowflake-extended.pdf, Project Snoflake) for the general approach.
 */
 struct AscendingPageAllocator
 {
@@ -73,7 +77,7 @@ public:
         }
         else
         {
-            assert(0, "Unsupported OS version");
+            static assert(0, "Unsupported OS version");
         }
 
         offset = data;
@@ -81,10 +85,15 @@ public:
     }
 
     /**
-    Round the allocation to the next multiple of the page size.
+    Rounds the allocation size to the next multiple of the page size.
     The allocation only reserves a range of virtual pages but the actual
-    physical memory is allocated on demand, when accessing the memory
-    Returns `null` on failure or if the requested size exceeds the remaining capacity.
+    physical memory is allocated on demand, when accessing the memory.
+    
+    Params:
+    n = Bytes to allocate
+    
+    Returns:
+    `null` on failure or if the requested size exceeds the remaining capacity.
     */
     void[] allocate(size_t n)
     {
@@ -133,7 +142,7 @@ public:
     }
 
     /**
-    Rounds the requested size to the next multiple of the page size
+    Rounds the requested size to the next multiple of the page size.
     */
     size_t goodAllocSize(size_t n)
     {
@@ -144,8 +153,8 @@ public:
     Decommit all physical memory associated with the buffer given as parameter,
     but keep the range of virtual addresses.
 
-    On POSIX systems we call `mmap` with `MAP_FIXED' a second time to decommit the memory.
-    On Windows we use `VirtualFree` with `MEM_DECOMMIT`.
+    On POSIX systems `deallocate` calls `mmap` with `MAP_FIXED' a second time to decommit the memory.
+    On Windows, it uses `VirtualFree` with `MEM_DECOMMIT`.
     */
     version(Posix)
     {
@@ -198,14 +207,11 @@ public:
     }
     else
     {
-        bool deallocate(void[] buf)
-        {
-            assert(0, "Unsupported OS version");
-        }
+        static assert(0, "Unsupported OS");
     }
 
     /**
-    Return `true` if the passed buffer is inside the range of virtual adresses.
+    Returns `true` if the passed buffer is inside the range of virtual adresses.
     Does not guarantee that the passed buffer is still valid.
     */
     bool owns(void[] buf)
--- a/std/experimental/allocator/building_blocks/ascending_page_allocator.d
+++ b/std/experimental/allocator/building_blocks/ascending_page_allocator.d
@@ -121,13 +121,14 @@ public:
                 {
                     import core.sys.windows.windows : VirtualAlloc, MEM_COMMIT, PAGE_READWRITE;
 
-                    auto ret = VirtualAlloc(readWriteLimit, newReadWriteLimit - readWriteLimit, MEM_COMMIT, PAGE_READWRITE);
+                    auto ret = VirtualAlloc(readWriteLimit, newReadWriteLimit - readWriteLimit, 
+                        MEM_COMMIT, PAGE_READWRITE);
                     if (!ret)
                         assert(0, "Failed to allocate memory, VirtualAlloc failure");
                 }
                 else
                 {
-                    assert(0, "Unsupported OS version");
+                    static assert(0, "Unsupported OS");
                 }
 
                 readWriteLimit = newReadWriteLimit;
@@ -342,7 +343,8 @@ public:
                 else version(Windows)
                 {
                     import core.sys.windows.windows : VirtualAlloc, PAGE_READWRITE, MEM_COMMIT;
-                    auto ret = VirtualAlloc(readWriteLimit, newReadWriteLimit - readWriteLimit, MEM_COMMIT, PAGE_READWRITE);
+                    auto ret = VirtualAlloc(readWriteLimit, newReadWriteLimit - readWriteLimit,
+                        MEM_COMMIT, PAGE_READWRITE);
                     if (!ret)
                         assert(0, "Failed to expand, VirtualAlloc failure");
                 }
--- a/std/experimental/allocator/building_blocks/ascending_page_allocator.d
+++ b/std/experimental/allocator/building_blocks/ascending_page_allocator.d
@@ -56,7 +56,7 @@ public:
         {
             import core.sys.posix.sys.mman : mmap, MAP_ANON, PROT_NONE,
                 MAP_PRIVATE, MAP_FAILED;
-            import core.sys.posix.unistd;
+            import core.sys.posix.unistd : sysconf;
 
             pageSize = cast(size_t) sysconf(_SC_PAGESIZE);
             data = mmap(null, pageSize * pages, PROT_NONE, MAP_PRIVATE | MAP_ANON, -1, 0);
--- a/std/experimental/allocator/building_blocks/ascending_page_allocator.d
+++ b/std/experimental/allocator/building_blocks/ascending_page_allocator.d
@@ -56,7 +56,7 @@ public:
         {
             import core.sys.posix.sys.mman : mmap, MAP_ANON, PROT_NONE,
                 MAP_PRIVATE, MAP_FAILED;
-            import core.sys.posix.unistd : sysconf;
+            import core.sys.posix.unistd : sysconf, _SC_PAGESIZE;
 
             pageSize = cast(size_t) sysconf(_SC_PAGESIZE);
             data = mmap(null, pageSize * pages, PROT_NONE, MAP_PRIVATE | MAP_ANON, -1, 0);
--- a/std/experimental/allocator/building_blocks/ascending_page_allocator.d
+++ b/std/experimental/allocator/building_blocks/ascending_page_allocator.d
@@ -17,7 +17,7 @@ memory.
 */
 struct AscendingPageAllocator
 {
-    enum size_t pageSize = 4096;
+    size_t pageSize;
     size_t numPages;
     bool valid;
 
@@ -29,6 +29,9 @@ struct AscendingPageAllocator
 
     // Number of pages which contain alive objects
     size_t pagesUsed;
+
+    void* readWriteLimit;
+    enum extraAllocPages = 1000;
 
     /**
     The allocator receives as a parameter the size in pages of the virtual
@@ -44,17 +47,23 @@ struct AscendingPageAllocator
         {
             import core.sys.posix.sys.mman : mmap, MAP_ANON, PROT_READ,
                    PROT_WRITE, PROT_NONE, MAP_PRIVATE, MAP_FAILED;
+            import core.sys.posix.unistd;
+            pageSize = cast(size_t) sysconf(_SC_PAGESIZE);
             data = mmap(null, pageSize * pages, PROT_NONE, MAP_PRIVATE | MAP_ANON, -1, 0);
             enforce(data != MAP_FAILED, "Failed to mmap memory");
         }
         else version(Windows)
         {
             import core.sys.windows.windows : VirtualAlloc, PAGE_NOACCESS,
-                   MEM_COMMIT, MEM_RESERVE;
-            data = VirtualAlloc(null, pageSize * pages, MEM_COMMIT | MEM_RESERVE, PAGE_NOACCESS);
+                   MEM_COMMIT, MEM_RESERVE, GetSystemInfo, SYSTEM_INFO;
+            SYSTEM_INFO si;
+            GetSystemInfo(&si);
+            pageSize = cast(size_t) si.dwPageSize;
+            data = VirtualAlloc(null, pageSize * pages, MEM_RESERVE, PAGE_NOACCESS);
             enforce(data != null, "Failed to VirtualAlloc memory");
         }
         offset = data;
+        readWriteLimit = data;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+read and write using OS primitives (`mprotect` and `VirtualProtect`).
+The physical memory is allocated on demand, when the pages are accessed.
+
+Deallocation removes any read/write permissions from the target pages
+and notifies the OS to reclaim the physical memory, while keeping the virtual
+memory.
+*/
+struct AscendingPageAllocator
+{
+    enum size_t pageSize = 4096;
```
Burgos: This should be queried from the environment - it's possible that page size is different than 4096. On Linux that would be `sysconf(_SC_PAGE_SIZE).` - see `core/thread.d` in druntime.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+        version(Posix)
+        {
+            import core.sys.posix.sys.mman : mprotect, PROT_READ, PROT_WRITE;
+            auto ret = mprotect(offset, extraPages * pageSize, PROT_READ | PROT_WRITE);
+            enforce(ret == 0, "Failed to expand, mprotect failure");
+        }
+        else version(Windows)
+        {
+            import core.sys.windows.windows : VirtualProtect, PAGE_READWRITE;
+            uint oldProtect;
```
Burgos: This is a good place to `assert` on the fact `oldProtect` should be no access.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+
+            // madvise to let the OS reclaim the resources
+            ret = posix_madvise(buf.ptr, goodSize, POSIX_MADV_DONTNEED);
+            enforce(ret == 0, "Failed to deallocate, posix_madvise failure");
+            pagesUsed -= goodSize / pageSize;
+
+            if (!valid && pagesUsed == 0)
+            {
+                munmap(data, numPages * pageSize);
+                data = null;
```
Burgos: I have a feeling you should reset `offset` to null as well. Also it doesn't seem that `allocate` would gracefully fail, except on mprotect failure. Maybe there should be `assert(valid)` there.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+            enforce(ret != 0, "Failed to deallocate memory, VirtualProtect failure");
+
+            // MSDN states for VirtualAlloc: Calling VirtualUnlock on a range
+            // of memory that is not locked releases the pages from the process's working set.
+            VirtualUnlock(buf.ptr, goodSize);
+            pagesUsed -= goodSize / pageSize;
+
+            if (!valid && pagesUsed == 0)
+            {
+                VirtualFree(data, 0, MEM_RELEASE);
```
Burgos: missing check for success/failure.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+            import std.exception : enforce;
+
+            uint oldProtect;
+            size_t goodSize = goodAllocSize(buf.length);
+            auto ret = VirtualProtect(buf.ptr, goodSize, PAGE_NOACCESS, &oldProtect);
+            enforce(ret != 0, "Failed to deallocate memory, VirtualProtect failure");
+
+            // MSDN states for VirtualAlloc: Calling VirtualUnlock on a range
+            // of memory that is not locked releases the pages from the process's working set.
+            VirtualUnlock(buf.ptr, goodSize);
```
Burgos: missing check for success/failure.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+            enforce(ret == 0, "Failed to deallocate memory, mprotect failure");
+
+            // madvise to let the OS reclaim the resources
+            ret = posix_madvise(buf.ptr, goodSize, POSIX_MADV_DONTNEED);
+            enforce(ret == 0, "Failed to deallocate, posix_madvise failure");
+            pagesUsed -= goodSize / pageSize;
+
+            if (!valid && pagesUsed == 0)
+            {
+                munmap(data, numPages * pageSize);
```
Burgos: missing check for success/failure.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+        if (!newSize) return deallocate(b);
+        if (!b) return true;
+
+        if (newSize >= b.length && expand(b, newSize - b.length))
+            return true;
+
+        void[] newB = allocate(newSize);
+        if (newB.length <= b.length) newB[] = b[0 .. newB.length];
+        else newB[0 .. b.length] = b[];
+        deallocate(b);
```
Burgos: Could we somehow make use of holes left by this?

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+addresses.
+
+When a chunk of memory is requested, the allocator finds a range of
+virtual pages which satisfy the requested size, changing their protection to
+read and write using OS primitives (`mprotect` and `VirtualProtect`).
+The physical memory is allocated on demand, when the pages are accessed.
+
+Deallocation removes any read/write permissions from the target pages
+and notifies the OS to reclaim the physical memory, while keeping the virtual
+memory.
```
Burgos: The thought that I just had is that you don't need to mess with the protection at all. If you just allocate from a pool and give back the arrays, you can rely on the bounds check to protect you from going out of bounds. Because, if you have two buffers adjacently allocated in your address range, there's really no protection at all (you don't leave range of memory with no access between them), and writing over the end of the first one will corrupt memory. So, protecting the memory by the OS only works in some cases, and not all of them.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+        {
+            import core.sys.posix.sys.mman : posix_madvise, POSIX_MADV_DONTNEED, mprotect, PROT_NONE, munmap;
+            import std.exception : enforce;
+
+            size_t goodSize = goodAllocSize(buf.length);
+            auto ret = mprotect(buf.ptr, goodSize, PROT_NONE);
+            enforce(ret == 0, "Failed to deallocate memory, mprotect failure");
+
+            // madvise to let the OS reclaim the resources
+            ret = posix_madvise(buf.ptr, goodSize, POSIX_MADV_DONTNEED);
```
DmitryOlshansky: posix_madvise doesn't guarantee a damn thing. PROT_NONE might help, but you are definetly after MADV_FREE found in FreeBSD/OSX and on recent Linux. It is a non-portable interface of madvise.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
DmitryOlshansky: Also see MEM_RESET in VirtualAlloc, might be a better match for your needs. Anyhow - benchmark/test is appropriate.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+            import core.sys.posix.sys.mman : mmap, MAP_ANON, PROT_READ,
+                   PROT_WRITE, PROT_NONE, MAP_PRIVATE, MAP_FAILED;
+            data = mmap(null, pageSize * pages, PROT_NONE, MAP_PRIVATE | MAP_ANON, -1, 0);
+            enforce(data != MAP_FAILED, "Failed to mmap memory");
+        }
+        else version(Windows)
+        {
+            import core.sys.windows.windows : VirtualAlloc, PAGE_NOACCESS,
+                   MEM_COMMIT, MEM_RESERVE;
+            data = VirtualAlloc(null, pageSize * pages, MEM_COMMIT | MEM_RESERVE, PAGE_NOACCESS);
```
DmitryOlshansky: MEM_COMMIT will actually reserve swap space on Windows and crash loudly if you reserve more the than you have.

DmitryOlshansky: So on Windows you may want to try MEM_RESERVE in ctor and MEM_COMMIT in allocate + MEM_RESET in deallocate.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
DmitryOlshansky: Yeah, just look for pageSize in DRuntime/Phobos there are plenty of places where it is used.

DmitryOlshansky:
> The main advantage of this allocator is that it is safe, because we do not reuse memory and fast for large objects.
> After some more tweaks I will also come back with a benchmark.

Yes a separate strategy for large object is great. In fact large arrays / hash tables would benefit the most.

Speaking of does not reuse memory - it may be fine, but even on 64-bit you might run out of virtual address space if you keep allocating/deallocating. Especially in long-running process.

Some sane strategy of "wrap-around" on end of virtual memory range might be a good idea. Making it tunable as a template parameter would cover both cases.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
DmitryOlshansky: Interestingly if we were to e.g. store a bit per page, the memory overhead would be smallish (1byte per 32 kbytes) but we can find adjucent free pages when calling expand easily.

Other strategies are possible as well.

jercaianu:
@DmitryOlshansky @Burgos 
Thanks for the comments, will shortly look into them
Alex

std/experimental/allocator/building_blocks/ascending_page_allocator.d
jercaianu: Yes, changing protection to none does not solve this alone. Maybe I can add guard pages between allocations for the ultimate defense :)

std/experimental/allocator/building_blocks/ascending_page_allocator.d
jercaianu: Yes, MADV_FREE would be better but just as you said, it's available since Linux 4.5.
I could add some version checks to see if we're using a Linux version which has MADV_FREE.
I'm not sure yet, but maybe MADV_REMOVE could be of help.

jercaianu:
@DmitryOlshansky 
I was recently talking with @andralex  about this, and we thought that an ideal use case for this would be together with AllocatorList.

Let's say that the allocator would cover a 16GB range of virtual addresses. The next step is to chain a bunch of these in an AllocatorList. Whenever an allocation is requested, either it is satisfied by the last arena in this list, or we create a new range of 16GB virtual memory.

Whenever an allocator in this list has all memory deallocated, we remove it from the list and unmap the pages.

I think that by using this strategy, is it highly unlikely that we would run out of virtual memory.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
jercaianu: VirtualUnlock here is expected to fail, but have the intended behavior [1].
I'll also experiment in more detail with all params for VirtualAlloc.

[1] - https://msdn.microsoft.com/en-us/library/windows/desktop/aa366910(v=vs.85).aspx

andralex: @jercaianu you may want to put this as a code comment just before the call

nemanja-boric-sociomantic: This blog post from MSDN might come in handy: https://blogs.msdn.microsoft.com/oldnewthing/20170113-00/?p=95185

jercaianu: @nemanja-boric-sociomantic 
Thanks for the post :)

After reading this I think that only VirtualUnlock is not enough, because as it states here, the pages would be evicted from the working set, but memory would still be accessible. From my understanding this means that the pages will be swapped out on disk, right?

Maybe we can use this together with VirtualAlloc/MEM_RESET, which states that it will make the contents unavailable after eviction.

I'll test this, but can someone please confirm/infirm, just in theory, if my judgement is right?
Thanks

nemanja-boric-sociomantic: Yes, `VirtualUnlock` would do similar thing to the `munlock` in Posix - just allow system to swap pages out, but they would still be transferred to pagefile (which you don't need). I indeed think you want `VirtualAlloc/MEM_RESET`, I don't think you need to call `VirtualUnlock` at all.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
andralex: thx @DmitryOlshansky

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+            enforce(data != null, "Failed to VirtualAlloc memory");
```
andralex: same here use assert(0) instead

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+
+    /**
+    Round the allocation to the next multiple of the page size.
+    The allocation only reserves a range of virtual pages but the actual
+    physical memory is allocated on demand, when accessing the memory
+    Returns `null` on failure or if the requested size exceeds the remaining capacity.
+    */
+    void[] allocate(size_t n)
+    {
+        import std.exception : enforce;
```
andralex: basically this should never be used - no exceptions in allocators at all

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+
+        size_t goodSize = goodAllocSize(n);
+        if (offset - data > numPages * pageSize - goodSize)
```
andralex: pedantically nicer because it has fewer chance of overflowing: `if (offset + goodSize > data + numPages * pageSize)`

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+    ``VirtualUnlock`` for Windows).
+    If the allocator is no longer valid, and all objects have been deallocated,
+    the range of virtual addresses is unmapped.
+    */
+    version(Posix)
+    {
+        bool deallocate(void[] buf)
+        {
+            import core.sys.posix.sys.mman : posix_madvise, POSIX_MADV_DONTNEED, mprotect, PROT_NONE, munmap;
+            import std.exception : enforce;
```
andralex: should go away

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    /**
+    Marks the allocator unavailable for further allocations and sets the ``valid``
+    flag to ``false``, which unmaps the virtual address range when all memory is deallocated.
+    */
+    void invalidate()
+    {
+        valid = false;
+        if (pagesUsed == 0) {
```
andralex: fix brace style

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+    Return ``true`` if the passed buffer is inside the range of virtual adresses.
+    Does not guarantee that the passed buffer is still valid.
+    */
+    bool owns(void[] buf)
+    {
+        return buf.ptr >= data && buf.ptr < buf.ptr + numPages * pageSize;
+    }
+
+    /**
+    Marks the allocator unavailable for further allocations and sets the ``valid``
```
andralex: why two backticks? just one should be the case

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -0,0 +1,418 @@
...
+        b = cast(void[]) b.ptr[0 .. b.length + delta];
+        return true;
+    }
+
+    /**
+    First it tries to ``expand`` to satisfy ``newSize``.
+    If not possible, it allocates a new buffer of length ``newSize``,
+    copies the contents and deallocates the previous buffer.
+    */
+    bool reallocate(ref void[] b, size_t newSize)
```
andralex: If this is not different from the stock reallocate, you could remove and just use that.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
andralex: We don't want to reuse holes. Recall that one important purpose of this allocator is to help with use-after-free.

andralex:
According to @jercaianu 's measurements it seems this is quite slow, and expectedly so. One way to accelerate things is to reduce the syscalls. When an allocation comes, we could commit more memory than needed (and then use that memory for future allocations). When freeing occurs, we free pages eagerly. So we have the same frequency of page deallocations but we allocate pages less frequently and in smaller batches.

That would be a combination of this allocator and BitmappedBlock.

jercaianu:
Thanks for all the comments!
One benchmark that I tried does a series of allocations and deallocations of 10 pages, covering about 100GB - [1].

The problem is that because we do not reuse memory, the allocator spends a lot of time in I/O with page faults. 

While malloc has no problems finishing this in a couple of seconds, the AscendingPageAllocator after a couple of minutes, causes memory pressure issues, because of a lot of page fault activity.

std/experimental/allocator/building_blocks/ascending_page_allocator.d
Burgos: @andralex @jercaianu all right, thanks, with that note in mind both this and the no-holes-reuse makes sense. With re-reading the documentation I can see the idea, but maybe it needs to be articulated more clearer to the casual observer.

DmitryOlshansky:
> The problem is that because we do not reuse memory, the allocator spends a lot of time in I/O with page faults.

This. Please try MADV_FREE on free-d pages, especially on Mac. It should make more soft-faults as opposed to hard faults.

> While malloc has no problems finishing this in a couple of seconds, the AscendingPageAllocator after a couple of minutes, causes memory pressure issues, because of a lot of page fault activity.

Indeed. But unlike malloc you actually provide pages that are zeroed-out.
More apple-to-apple would be using calloc.

DmitryOlshansky:
> I think that by using this strategy, is it highly unlikely that we would run out of virtual memory.
> I know there is a lot more to it and I haven't figured out everything, but I think it's a good direction.

Indeed, however I should warn that memory fragmentation is a thing and a single pesky long-lived object can pin down an entire 16Gb span. Now if you are using this as a basis for region-style allocation such cases are impossible, otherwise they might not be.

jercaianu:
@DmitryOlshansky @andralex
Thanks for advice/comments!

I have tested a bit more and here is how it performs so far.
I have 2 tests:
1. Do 1000 small allocations(10 pages each),dirty the pages and then deallocate everything.
Do all the above for a large number of times.

2. Do 1000 random allocations(each allocation ranging from 1 to 1000 pages), dirty the pages and then deallocate everything in random order.
Do all the above for a large number of times. [2]

Here is how the allocator did compared to malloc.

Linux 4.4.0-101-generic
Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
8GB RAM 2133 MHz

Test1:
malloc: 22s
page_allocator: 22s

Test2:
malloc: 1min 13s
page_allocator: 56s
-----------------------------

Darwin 15.6.0
Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
16GB RAM 1600MHz

Test1:
malloc: 13s
page_allocator: 42s

Test2:
malloc: 3min 7s
page_allocator: 3min
-----------------------------

I have tried MADV_FREE, but it did not totally decommit memory in my case.
What solved my problem, is calling mmap(MAP_FIXED) a second time over already mmaped memory.
I haven't played so far with any other parameters of madvise, but there is plenty of room for experiments.

It seems that for the first test, malloc significantly outperforms the allocator on my mac, because of no page faults.
On randomized tests however, we slightly outperform malloc on both Linux and Mac.
[1] - https://pastebin.com/sFgGnbyJ
[2] - https://pastebin.com/ZaCrJgDU

jercaianu:
Also, if reading the code, please try to ignore for the moment the documentation.
I started too early writing documentation and now I've changed some stuff.
After the code reaches a stable form, I will update the documentation.
Thanks!

DmitryOlshansky:
> I have tried MADV_FREE, but it did not totally decommit memory in my case.

The whole point is that teh memory is released to OS lazily. Yes it will count as commited for a short while or if the OS is fine without it for a long time.

andralex:
@DmitryOlshansky @jercaianu let's get this in as it's fundamentally sound, so we can improve it!

jercaianu:
@DmitryOlshansky 
I have not been explicit enough about my problem.
What happened during my benchmarks, is that even though RAM usage remained fairly low, after a couple of mins, I would get from my Mac an error that my system has run out of memory. 
This happened both with MADV_FREE and with MADV_DONTNEED. It seemed that even MADV_FREE did not totally decommit the memory (I think that I ran out of disk space).

The only workaround I found so I don't get the 'run out of memory thing' was either to 'munmap' on deallocation, or this approach in my most recent commit, to call mmap(MAP_FIXED).

As I also found here, It seems that mmap(MAP_FIXED) does the trick on Darwin [1].
Thanks!

[1] - https://github.com/jemalloc/jemalloc/issues/255

jercaianu:
I have also tested on a 64bit Windows7 machine.
Running tests multiple times, I have noticed that the allocator performs just as good as malloc does, with oscilating time differences of less than a second for very large tests.

For deallocation on Windows, VirtualFree with MEM_DECOMMIT does the trick. Not only does it decommit the physical memory, but further attempts to access that memory result in an access violation exception.

Just as MSDN[1] states: 
'After memory is released or decommited, you can never refer to the memory again. Any information that may have been in that memory is gone forever. Attempting to read from or write to a free page results in an access violation exception. If you need to keep information, do not decommit or free memory that contains the information'.

[1] - https://msdn.microsoft.com/ru-ru/library/windows/desktop/aa366892(v=vs.85).aspx

I am open to any further comments/review.

Thanks,
Alex

jercaianu:
Looking again at this, I don't think it makes sense to assert(0) in allocate/deallocate.
In case of memory pressure, these calls may fail, but this kind of failure is not irreparable.
I think it's better to return false

andralex:
@jercaianu makes sense

andralex:
@jercaianu I'm working on getting this in, feel free to follow up in a different PR

DmitryOlshansky:
> As I also found here, It seems that mmap(MAP_FIXED) does the trick on Darwin [1].
> Thanks!

That's about right, cool trick. And yes, jemalloc is an excellent resource on things like that :)
Also could be useful for my druntime PR for VMM module to make portable code that manages virtual memory easier.

PR #5880, opened 2017-11-21

edi33416:

This can now be safely merged as the vibe-d [patch](https://github.com/vibe-d/vibe.d/pull/1969) was merged. For more references see #5837 .


wilzbach:
FYI: you don't need to create a new pull request to change the branch ...

Just click on "edit" :

image

andralex:
@edi33416 anyhow it seems we now are safe to return to the master branch...

PR #5881, opened 2017-11-21

edi33416:

### DIFF
--- a/std/experimental/allocator/building_blocks/affix_allocator.d
+++ b/std/experimental/allocator/building_blocks/affix_allocator.d
@@ -112,7 +112,7 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
             }
         }
 
-        private void[] actualAllocation(void[] b) const
+        private inout(void)[] actualAllocation(inout(void)[] b) const
         {
             assert(b !is null);
             return (b.ptr - stateSize!Prefix)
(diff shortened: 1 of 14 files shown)


std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -168,10 +168,11 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
         }
 
         static if (hasMember!(Allocator, "owns"))
-        Ternary owns(void[] b)
+        //Ternary owns(const void[] b) const ?
```
wilzbach: Remove

std/experimental/allocator/building_blocks/allocator_list.d
```diff
@@ -349,7 +349,8 @@ struct AllocatorList(Factory, BookkeepingAllocator = GCAllocator)
     returned  `Ternary.unknown`.
     */
     static if (hasMember!(Allocator, "owns"))
-    Ternary owns(void[] b)
+    //Ternary owns(const void[] b) const ?
```
wilzbach: ?

std/experimental/allocator/building_blocks/bucketizer.d
```diff
@@ -166,11 +166,11 @@ struct Bucketizer(Allocator, size_t min, size_t max, size_t step)
     static if (hasMember!(Allocator, "owns"))
     Ternary owns(void[] b)
```
wilzbach: Shouldn't this be `const`?

edi33416: Yes it should. Apparently the cherry pick didn't go as smooth as I had hoped 😄

std/experimental/allocator/building_blocks/affix_allocator.d
edi33416: The question is, should the function header be
- `Ternary owns(const void[] b);` or
- `Ternary owns(const void[] b) const;`

andralex:
Per the discussion with @edi33416 let's close this as it has low impact.

andralex:
We can return to it later.

PR #5403, opened 2017-05-19
Use OpaquePointer as a safe support for region based allocators

edi33416:

PR #5882, opened 2017-11-21
Use OpaquePointer as a safe support for region based allocators

edi33416:


andralex:
@edi33416 I recall there was a related bug, was there?

andralex:
What's the deal with "stderr test, please ignore" in the output followed by unittest failure?

n8sh:
>Using `OpaquePointer` will ensure that
>the garbage collector will mark the memory for scanning.
>Note: the responsibility of allocating the buffer properly falls on the caller.

What is the benefit of OpaquePointer over void[]? Does the current garbage collector not scan memory allocated with `new void[1024]`?

n8sh:
Or I suppose `void*[]`

andralex:
@n8sh thanks for asking!

* `void[]` is not good because an array of anything converts to `void[]`, thus allowing the allocator to trample over memory of other objects.
* `void*[]` is better, but it allows the nonsensical operation of assigning individual pointers

An `OpaquePointer` canot be assigned meaningfully so it's a bit more descriptive. But it does seem that `void*[]` is a solid solution worth looking twice at. @edi33416 do you think that would work out easier?

edi33416:
> @edi33416 I recall there was a related bug, was there?

I'm still not convinced that those changes were buggy. For reference, please see #5403

edi33416:
> What's the deal with "stderr test, please ignore" in the output followed by unittest failure?

I don't know what I should make of this output

edi33416:
> An `OpaquePointer` canot be assigned meaningfully so it's a bit more descriptive. But it does seem that `void*[]` is a solid solution worth looking twice at. @edi33416 do you think that would work out easier?

An `OpaquePointer` is just a struct that wraps a `void*`, so an `OpaquePointer[]` is equivalent to `void*[]`. From a users perspective, I think that using `OpaquePointer[]` instead of `void*[]` is less error prone, as it would be easy to miss the `*`, but it would be a hard to find bug.

@andralex @n8sh  what do you think?

n8sh:
That's reasonable. A tradeoff between verbosity and clarity is a matter of style/taste.

My kneejerk reaction to `OpaquePointer` was that introducing a type used here and nowhere else increased the chance of having to write `(() @trusted => cast(OpaquePointer[])allocated_memory)()`, but that may not be a real concern.

std/experimental/allocator/building_blocks/region.d
```diff
@@ -53,21 +54,23 @@ struct Region(ParentAllocator = NullAllocator,
...
+    The underlying `store` must be of $(LREF OpaquePointer) type for a safe
+    allocation of objects that may contain pointers, and thus for a safe allocator.
+    n = Bytes to allocate using `ParentAllocator`. This constructor is only
+    defined if `ParentAllocator` is different from `NullAllocator`. If
+    $(D parent.allocate(n)) returns `null`, the region will be initialized
     as empty (correctly initialized but unable to allocate).
     */
-    this(ubyte[] store)
+    @system this(ubyte[] store)
     {
```
n8sh: Why is it necessary to explicitly mark this constructor as `@system`?

wilzbach:
Removing "auto-merge" until this passes on Win32_64

andralex:
>I think that using OpaquePointer[] instead of void*[] is less error prone, as it would be easy to miss the *, but it would be a hard to find bug.

At least in this case we have a `ubyte[]` vs a `void*[]` overload, so any confusion between the two would be caught by the compiler.

I'd say let's go qwith `void*[]`, we don't want to complicate ourselves without necessity. `OpaquePointer` was my (apparently not so good) idea, sorry @edi33416 for the wasted work.

PR #5883, opened 2017-11-22
Modified a unittest to work independent of IEEE rounding mode 

bvoq:
The unittest converts a double to a string, which depends on the IEEE rounding mode.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -635,16 +635,11 @@ if (distinctFieldNames!(Specs))
         mixin(injectNamedFields());
 
         ///
-        static if (Specs.length == 0) @system unittest
-        {
-            //The following conversion from a double to a string doesn't work for all rounding modes, therefore set the rounding mode.
-            import std.math;
-            FloatingPointControl fpctrl;
-            uint previousRoundingMode = fpctrl.rounding;
-            fpctrl.rounding = FloatingPointControl.roundToNearest;
-            auto t1 = tuple(1, " hello ", 2.3);
-            assert(t1.toString() == `Tuple!(int, string, double)(1, " hello ", 2.3)`);
-            fpctrl.rounding = previousRoundingMode;
+        
+        static if (Specs.length == 0) @safe unittest
+        {    
+            auto t1 = tuple(1, " hello ", 'a');
+            assert(t1.toString() == `Tuple!(int, string, char)(1, " hello ", 'a')`);
                     
             void takeSeveralTypes(int n, string s, bool b)
             {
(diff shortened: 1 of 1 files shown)


MetaLang:
@wilzbach ever seen this one? It failed over to a local dub mirror and eventually found the package, but this step of the build looks like it _still_ failed.

```
[ERROR] /var/lib/jenkins/dlang_projects@2/dlang/dub/test/feat663-search.sh:11 `dub search dub` failed
        Using dub registry url 'https://code-mirror.dlang.io/'
        Refreshing local packages (refresh existing: true)...
        Looking for local package map at /var/lib/dub/packages/local-packages.json
        Looking for local package map at /var/lib/jenkins/dlang_projects@2/.dub/packages/local-packages.json
        Try to load local package map at /var/lib/jenkins/dlang_projects@2/.dub/packages/local-packages.json
        Determined package version using GIT: dub 1.6.0
        ==== registry at https://code-mirror.dlang.io/ (fallback registry at https://code-mirror.dlang.io/) ====
        dub (1.6.0)                     Package manager for D packages
        gitcompatibledubpackage (1.0.4) Example of a DUB package also usable as git submodule. For DUB test suite.
        arsd-official (1.2.2)           A container of various subpackages that do lots of different things. You should use one of the subpackages instead of the main package in most cases, but you can try the complete package if you get duplicated dependency issues.
        dub-registry (2.2.0)            Online registry for dub packages
        dfl2 (1.0.7)                    dfl2 is a GUI library for Windows,now it can be built by dco, DUB or Visual Studio .NET(such as VS 2010).
        dco (0.1.2)                     dco is a build tool, it can be built by DUB and Visual Studio .NET(such as VS 2010).
        hmod-dub (0.1.1)                Generates DDoc documentation for any package in the DUB registry
        dwtlib (3.1.1)                  The dub package for dwt
        rdub (3.0.1)                    RDUB is a front end for DUB, a D language build tool.
        dubimport (1.1.2)               A command line tool to extract import paths on dub project.
        dubbio (0.5.0)                  Dub wrapper
        dub-data-mod (0.9.23-mod.1)     Temporary fork of dub that includes 'dub describe --data=' feature (until it's merged and released in mainline dub).
        gdub (1.1.0)                    GDUB is a DWT GUI front end for DUB, a D language build tool
[ERROR] Script failure.
```

ZombineDev:
@wilzbach 
> FWIW as we already look at this code: this is a unittest within a template and as long as DIP82 isn't reality, such unittest blocks should be versioned with something like unittest_phobos to avoid compilation for every template instance - even in user code.

No need for that, as the test in question is already guarded by `static if (Specs.length == 0)`.

ZombineDev:
@MetaLang looks like the error was intermittent and went away.

MetaLang:
It went away because I restarted it. At least we know that it can be worked around.

PR #5884, opened 2017-11-22
[RFC] std.algorithm.stats

JackStouffer:
This PR is a proposition for two separate things, which should be evaluated in turn, but are inevitably tied together:

1. Adding the basic statistic functions to Phobos
2. Moving those functions to their own module

This PR also adds a `variance` function, the reason for which is explained below.

# Stat Functions

Because the API is relatively simple and unchanging, and they have such wide range of use, I believe that the following stat functions should make their way into Phobos in some fashion:

- mean (done)
- mode
- variance
- standard deviation
- skewness
- quartile
- correlation
- covariance

More complex functions such as CDFs, PDFs etc. can be added later if so desired. But I think they are a lower priority as they're used less.

# Module Organization

If it's decided that the above stat functions should be added to Phobos, we have several options on how to do so

1. Add the functions to `std.numeric`
2. Add the functions to a new module, `std.algorithm.stats`/`statistics` (implemented in this PR)
3. Add the functions to a new module, `std.numeric.stats`/`statistics`
4. Add the functions to a new module, `std.experimental.stats`/`statistics`

## Pros and Cons

Adding (or moving) all of these functions to std.numeric is the most simple option. Plus, it allows these functions to be added one at a time and get into user's hands eariliest. There is also precedent, as there's already an entropy function in `std.numeric` The downside is it reduces the visibility in comparison to the separate module option. In a perfect world, googling "dlang stats" would bring up these functions.

Options two and three are relatively similar and are mostly bikesheding. I added `variance` here incase we go with this option, as I didn't think we would want a module with just one function.

As for the experimental option, the main upside of making a new module in std.experimental would allow for more deliberation on where we want the direction of the module to go. Do we just want the module to have simple stat functions? Do we want some form of summary `struct` like https://github.com/DlangScience/dstats? etc. The downside is that approach takes much longer to get into user's hands for a stable API (if it's decided to just go with the above list of functions). 

Please comment with your thoughts.

I'm partial to options 1 or 2.


std/algorithm/stats.d
```diff
@@ -0,0 +1,267 @@
...
+    !isInfinite!R)
+{
+    Unqual!T mean = 0;
+    Unqual!T var = 0;
+
+    return varianceImpl(r, mean, var, population);
+}
+
+/// ditto
+T variance(R, T)(R r, T seed, bool population = false)
```
wilzbach: Why not something like `Flag!WholePopulation`?

std/algorithm/stats.d
```diff
@@ -0,0 +1,267 @@
...
+if (isInputRange!R &&
+    !isInfinite!R &&
+    is(typeof(r.front + seed)) &&
+    is(typeof(r.front / size_t(1))))
+{
+    Unqual!T mean = seed;
+    Unqual!T var = seed;
```
wilzbach: Why don't you move this to the Impl too?

std/algorithm/stats.d
```diff
@@ -0,0 +1,267 @@
...
+
+    return varianceImpl(r, mean, var, population);
+}
+
+private T varianceImpl(R, T)(R r, ref T mean, ref T var, bool population)
```
wilzbach: `R` could be `auto ref` to save an additional copy.

std/algorithm/stats.d
```diff
@@ -0,0 +1,267 @@
...
+    // Welford’s single pass variance method
+    foreach (e; r)
+    {
+        Unqual!T oldMean = mean;
+        mean = mean + (e - mean) / i++;
+        var = var + (e - mean) * (e - oldMean);
+    }
+
+    static if (!hasLength!R)
+        if (i == 3)
```
wilzbach: <=

andralex:
We're at the end of a protracted discussion about how `std.experimental.allocator` needs to be stable (!), should have its own repo (!), etc. Now literally the _opening_ of this new module is making a breaking change by moving a just-added function elsewhere.

Please do all the work in a dub repository, then propose it for Phobos. Thanks.

wilzbach:
> just-added function

FYI: There hasn't been a release with the newly added function yet, hence the discussion.

andralex:
@wilzbach understood; my point was this kind of stuff is foreshadowing future API changes and instabilities

PR #5885, opened 2017-11-23
[WIP] Make allocators expand safe

edi33416:


std/experimental/allocator/building_blocks/allocator_list.d
```diff
@@ -381,7 +381,7 @@ struct AllocatorList(Factory, BookkeepingAllocator = GCAllocator)
         && hasMember!(Allocator, "owns"))
     bool expand(ref void[] b, size_t delta)
     {
-        if (!b.ptr) return delta == 0;
+        if (!b) return delta == 0;
```
andralex: I think `!b.length` would be a better check here.

std/experimental/allocator/building_blocks/bucketizer.d
```diff
@@ -96,12 +96,12 @@ struct Bucketizer(Allocator, size_t min, size_t max, size_t step)
     */
     bool expand(ref void[] b, size_t delta)
     {
-        if (!b.ptr) return delta == 0;
+        if (!b || delta == 0) return delta == 0;
```
andralex: here too etc

PR #5886, opened 2017-11-23
[WIP] Make reallocate nothrow nogc

edi33416:


MetaLang:
I just dropped a review without comment but I should say that while the code looks good, that "approve" is really conditional on the code actually passing all tests (of course).

edi33416:
I want to double check something before I'm ok with merging this. Thanks for the patience

PR #5888, opened 2017-11-23
win32: split unittest6.obj, it's becoming too large for OMF

rainers:
The current object file has 31929 symbols, pushing some PRs over the OMF limit of 32767, for example https://github.com/dlang/druntime/pull/1603

### DIFF
--- a/win32.mak
+++ b/win32.mak
@@ -561,6 +561,7 @@ UNITTEST_OBJS= \
 		unittest4.obj \
 		unittest5.obj \
 		unittest6.obj \
+		unittest6a.obj \
 		unittest6b.obj \
 		unittest7.obj \
 		unittest8a.obj \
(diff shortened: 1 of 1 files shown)


nordlow:
Does this mean that https://github.com/dlang/druntime/pull/1603 is ready for merge?

rainers:
> Does this mean that dlang/druntime#1603 is ready for merge?

It should be back to passing the tests now. But @MartinNowak doesn't want to commit to the changes to the GC interface unless it has proven to be faster than the existing GC (disregarding other benefits).

nordlow:
So, how do we make it faster? 😉

ZombineDev:
@rainers what do you think about breaking this into multiple PRs each laying down part of the infrastructure bits? Hopefully we could move faster with unobjectionable parts and focus the discussion on the trickier ones?

rainers:
Discussion should probably happen in https://github.com/dlang/druntime/pull/1603.

PR #5889, opened 2017-11-26
Fix SBrkRegion expand bug

edi33416:
`SBrkRegion.expand` must take into account the rounding due to alignment.


edi33416:
@andralex @wilzbach @ZombineDev @MetaLang

andralex:
@edi33416 looks like DAutoTest timed out, can you push again?

PR #5890, opened 2017-11-27
Make std.range.slide public

dhasenan:
It was scheduled to become public as of 2.076, and we're a bit past that. Plus several people have requested it on IRC.


wilzbach:
There is still an API issue (see the other PRs for details) with it which has to be worked out. I am sorry for being busy lately, but please don't set it simply to public. Hopefully I will get to it this week.

PR #5891, opened 2017-11-27
Correct wrongful example for chunkBy

Timoses:
chunkBy short description displayed a wrong example. (As far as I understood chunkBy).


wilzbach:
> displayed a wrong example. (As far as I understood chunkBy).

Could you elaborate on this?
Maybe you were simply confused:
https://run.dlang.io/is/v4lWuz

JackStouffer:
Ah, I was incorrect as well. Forgot that it goes in order like uniq or groupby

Timoses:
> Could you elaborate on this?
Maybe you were simply confused:
https://run.dlang.io/is/v4lWuz

Ah *facepalm*...

PR #5894, opened 2017-11-29
Integrate AscendingPageAllocator with AllocatorList

jercaianu:
- Added a couple of tests for AllocatorList together with AscendingPageAllocator.
- AscendingPageAllocator returns null/false for allocate/deallocate in case of failure
- AscendingPageAllocator receives as a constructor parameter the size in bytes, instead of the number of pages
- Removed some unused functions from AscendingPageAllocator
- When in ouroboros mode, the allocator which holds all the other allocators does not simply call deallocate, since it may have state. In case it has state, it creates a copy and then calls deallocate from that copy.

### DIFF
--- a/std/experimental/allocator/building_blocks/allocator_list.d
+++ b/std/experimental/allocator/building_blocks/allocator_list.d
@@ -476,7 +476,18 @@ struct AllocatorList(Factory, BookkeepingAllocator = GCAllocator)
         assert(special || !allocators.ptr);
         if (special)
         {
-            special.deallocate(allocators);
+            static if (stateSize!Allocator)
+            {
+                import core.stdc.string : memcpy;
+                Allocator specialCopy;
+                memcpy(&specialCopy, &special.a, Allocator.sizeof);
+                emplace(&special.a);
+                specialCopy.deallocate(allocators);
+            }
+            else
+            {
+                special.deallocate(allocators);
+            }
         }
         allocators = null;
         root = null;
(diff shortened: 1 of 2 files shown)


jercaianu:
Also, this fixes https://issues.dlang.org/show_bug.cgi?id=17393

std/experimental/allocator/building_blocks/allocator_list.d
```diff
@@ -476,7 +476,19 @@ struct AllocatorList(Factory, BookkeepingAllocator = GCAllocator)
...
         {
-            special.deallocate(allocators);
+            static if (stateSize!SAllocator)
+            {
+                import core.stdc.string : memcpy;
+                SAllocator specialCopy;
+                assert(special.a.sizeof == specialCopy.sizeof);
+                memcpy(&specialCopy, &special.a, specialCopy.sizeof);
+                emplace(&special.a);
+                specialCopy.deallocateAll();
```
andralex: nice!

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -100,7 +102,7 @@ public:
         import std.algorithm.comparison : min;
 
         size_t goodSize = goodAllocSize(n);
-        if (offset - data > numPages * pageSize - goodSize)
+        if (goodSize > numPages * pageSize || offset - data > numPages * pageSize - goodSize)
```
andralex: You may want to introduce `immutable pagedBytes = numPages * pageSize;` because it's used a number of times throughout the function,

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -212,44 +196,14 @@ public:
     }
 
     /**
-    Returns `true` if the passed buffer is inside the range of virtual adresses.
+    Returns `Ternary.yes` if the passed buffer is inside the range of virtual adresses.
     Does not guarantee that the passed buffer is still valid.
     */
-    bool owns(void[] buf)
+    Ternary owns(void[] buf)
```
wilzbach: What's the advantage of using `Ternary` here if there are only two states?

jercaianu: I used Ternary for uniformity just as the Region or KRR allocator implement for owns.

PR #5896, opened 2017-11-30
Issue 18021 - FileLogger Member not accessible to subclasses

burner:
fix Issue 18021

https://issues.dlang.org/show_bug.cgi?id=18021

### DIFF
--- a/std/experimental/logger/filelogger.d
+++ b/std/experimental/logger/filelogger.d
@@ -170,8 +170,11 @@ class FileLogger : Logger
         return this.filename;
     }
 
-    private File file_;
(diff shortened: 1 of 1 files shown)


BBasile:
There is already a property returning the file: https://dlang.org/phobos/std_experimental_logger_filelogger.html#.FileLogger.file

burner:
that property gets you another handle to it in a ref counted manner. A user stated that this given him the capabilities he needed when subclassing filelogger. Also you can't change the filename after construction.

burner:
@wilzbach thank you

PR #5897, opened 2017-11-30
fix Issue 18023 - remove std.net.isemail until license issues are resolved

WalterBright:
We cannot copy other peoples' code, translate it to D, and change the license on it. This module must be removed ASAP until we get permission from the copyright holder to change the license.

Note that the BSD license is incompatible with Boost because it requires binary attributeion, which is why we do not use it.

https://github.com/dominicsayers/isemail/blob/master/license.md


andralex:
No need for this. The email from Dominic Sayers on Tue, 22 Feb 2011 12:36:59 +0000 has sent you a zip file containing the code licensed under Boost. That was the basis of the current code.

WalterBright:
Phew! Dodged a bullet there!

jmh530:
It probably couldn't hurt to put a notice to this effect in the relevant files.

quickfur:
Walter already did this: https://github.com/dlang/phobos/pull/5898

jmh530:
@quickfur Ah, good.

PR #5898, opened 2017-11-30
std.net.isemail: document relicense

WalterBright:
Leave a trail to make it clear we have permission.


quickfur:
Should a link to the email be provided, or is the email private only?

WalterBright:
It was private and is in my email archives. I can produce it if needed, I just need a note in the source that it exists. I've run into these sorts of issues before, and having a breadcrumb trail of permissions is invaluable in resolving them quickly.

PR #5899, opened 2017-12-01
Add a cross-platform Makefile

wilzbach:
This will include the respective Makefile depending on the users's OS.

This is for all the people out there who are too lazy to type `make -f posix.mak` all the time and of course all the people who are new to the D build chain and who can now simply type the familiar `make` without needing to know anything about `posix.mak`.

See also: https://github.com/dlang/dmd/pull/7382

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,33 @@
+# Includes the respective Makefile depending on the OS
+# Have a look at either posix.mak, win32.mak or win64.mak for a list of available commands.
+
+DMD_DIR=../dmd
+include $(DMD_DIR)/src/osmodel.mak
+
+ifeq (Windows_NT,$(OS))
+    ifeq ($(findstring WOW64, $(shell uname)),WOW64)
+	OS:=win64
+	MODEL:=64
+    else
+	OS:=win32
+	MODEL:=32
+    endif
+endif
+ifeq (Win_32,$(OS))
+    OS:=win32
+    MODEL:=32
+endif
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,33 @@
+# Includes the respective Makefile depending on the OS
+# Have a look at either posix.mak, win32.mak or win64.mak for a list of available commands.
+
+DMD_DIR=../dmd
+include $(DMD_DIR)/src/osmodel.mak
```
wilzbach: The Phobos Makefiles assume the existence of a DMD directory (see e.g. `posix.mak`).

quickfur: Hmm. I looked at `dmd/src/osmodel.mak` but it only seems to detect Posix OS flavors, there doesn't seem to be anything that detects windows?

edi33416: I believe @quickfur is right. I had a look at `dmd/test/Makefile` and it seems to have the Windows checks written aside from including `dmd/src/osmodel.mak`. Maybe `osmodel.mak` needs an update?

edi33416:
Any updates on the status of this or plans of going forward?

wilzbach:
See https://github.com/dlang/dmd/pull/7382

edi33416:
Does it mean that the faith of this PR has been decided?

JackStouffer:
Seems like that's a "no" from Walter. Closing.

PR #5901, opened 2017-12-01
Update interfaces.d

Dashster:
$(TR $(TD $(LREF OutputRangeObject))
        $(TD Class that implements the $(D OutputRange) interface and wraps the
        $(D put) methods in virtual functions.
*************************************************************
    ))   <-   add the missing ending double parentheses
***************************************************************

### DIFF
--- a/win32.mak
+++ b/win32.mak
@@ -59,9 +59,10 @@ MAKE=make
 
 ## D compiler
 
-DMD=$(DIR)\bin\dmd
-#DMD=..\dmd
-DMD=dmd
+DMD_DIR=..\dmd
+BUILD=release
+OS=windows
+DMD=$(DMD_DIR)\generated\$(OS)\$(BUILD)\$(MODEL)\dmd
 
 ## Location of where to write the html documentation files
 
--- a/win64.mak
+++ b/win64.mak
@@ -63,9 +63,10 @@ MAKE=make
 
 ## D compiler
 
-DMD=$(DIR)\bin\dmd
-#DMD=..\dmd
-DMD=dmd
+DMD_DIR=..\dmd
+BUILD=release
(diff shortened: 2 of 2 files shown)


Dashster:
"you might want to target stable. If you don't know how to do this, let us know."

I don't know how to do that. I  initiated this action by the "Improve this page" link on 
https://dlang.org/phobos/std_range_interfaces.html

However, I closed my browser tab unintentionally after starting and did not know how to get back to where I was, so I again clicked on "Improve this page" and began again. But I noticed that there was now a 2 instead of a 1 after patch- . So I don't know if that created a problem or  not.

schveiguy:
> But I noticed that there was now a 2 instead of a 1 after patch-

so how this works is that github automatically forks the phobos project into your user's account, and then creates a "patch-X" branch so you can save your edits. Your patch-1 is still here: https://github.com/Dashster/phobos/tree/patch-1

> I don't know how to do that.

You can do it (from the start) by changing the base target before submitting the PR. You can also still do it, just click on the `Edit` button at the top of this page, and change the base.

schveiguy:
Wow, that didn't work. @wilzbach how to do this the correct way?

wilzbach:
> how to do this the correct way?

Usually `git rebase --onto upstream/stable` works nicely, but it seems that in this case it didn't and as I accidentally force-pushed it directly, it seems that I can't reopen nor push again :/
-> I opened a new PR (https://github.com/dlang/phobos/pull/5903) and I'm terribly sorry for the noise.

@Dashster thanks again for the PR and I hope you don't mind too much about how difficult it can be to get a one-line documentation change into Phobos. Hopefully we do a better job on your next submission ;-)

PR #5358, opened 2017-05-03
ARM and Android fixes: version out some HardFloat tests, use TMPDIR logic and declaration from druntime, and correct comment

joakim-noah:
x86/x64 has no problem with unaligned reads, but the llvm optimizer gets confused by the cast in the `foreach` and assumes the `Element` is 32-bit aligned, generating ARM instructions that require word alignment and causing problems with the last unaligned test.

As for `posix_memalign`, turns out its declared in druntime but wasn't for `OSX`, so added it in dlang/druntime#1819 and used the import instead.  That druntime PR will need to be merged first.

PR #5902, opened 2017-12-01
Upstream some LDC patches

kinke:
Pinging @klickverbot and @ibuclaw.

### DIFF
--- a/std/datetime/timezone.d
+++ b/std/datetime/timezone.d
@@ -334,6 +334,7 @@ public:
             else version(DragonFlyBSD)  enum utcZone = "UTC";
             else version(linux)         enum utcZone = "UTC";
             else version(OSX)           enum utcZone = "UTC";
+            else version(Solaris)       enum utcZone = "UTC";
             else static assert(0, "The location of the UTC timezone file on this Posix platform must be set.");
 
             auto tzs = [testTZ("America/Los_Angeles", "PST", "PDT", dur!"hours"(-8), dur!"hours"(1)),
@@ -2006,6 +2007,14 @@ public:
     {
         // Android concatenates all time zone data into a single file and stores it here.
         enum defaultTZDatabaseDir = "/system/usr/share/zoneinfo/";
+    }
+    else version(Solaris)
+    {
+        /++
+            The default directory where the TZ Database files are. It's empty
+            for Windows, since Windows doesn't have them.
+          +/
+        enum defaultTZDatabaseDir = "/usr/share/lib/zoneinfo/";
     }
     else version(Posix)
     {
--- a/std/digest/murmurhash.d
+++ b/std/digest/murmurhash.d
@@ -38,6 +38,11 @@ $(BR) $(LINK2 https://en.wikipedia.org/wiki/MurmurHash, Wikipedia)
  */
 module std.digest.murmurhash;
 
+version (X86)
+    version = HaveUnalignedLoads;
+else version (X86_64)
+    version = HaveUnalignedLoads;
+
 ///
 @safe unittest
 {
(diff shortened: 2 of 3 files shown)


std/digest/murmurhash.d
```diff
@@ -500,28 +509,75 @@ struct MurmurHash3(uint size /* 32 or 128 */ , uint opt = size_t.sizeof == 8 ? 6
         // Buffer should never be full while entering this function.
         assert(bufferSize < Element.sizeof);
 
-        // Check if we have some leftover data in the buffer. Then fill the first block buffer.
+        // Check if we don't even fill up a single block buffer.
```
MetaLang: This is very awkward wording. What about "check if we can fill up a single block buffer"?

kinke: The true-branch with early return is for the opposite case, i.e., where the current bytes in the block buffer and the new bytes to be hashed don't add up to a full `Element` size.

andralex: Not sure whether applicable here, but don't forget that for simple documentation matters patching someone else's PR is almost as easy as submitting a comment.

MetaLang: I don't like the idea of editing someone's pull request unless it's necessary.

jpf91: @kinke "check if we can **'t** fill up a single block buffer?" ;-)

std/digest/murmurhash.d
```diff
@@ -500,28 +509,75 @@ struct MurmurHash3(uint size /* 32 or 128 */ , uint opt = size_t.sizeof == 8 ? 6
...
+            putElement(buffer.block);
+            element_count += Element.sizeof;
+            data = data[bufferLeeway .. $];
+        }
 
         // Do main work: process chunks of `Element.sizeof` bytes.
         const numElements = data.length / Element.sizeof;
         const remainderStart = numElements * Element.sizeof;
-        foreach (ref const Element block; cast(const(Element[]))(data[0 .. remainderStart]))
+        version (haveUnalignedLoads)
```
ibuclaw: Haven't we already talked about this before?  Or am I thinking of another patch.

std/math.d
```diff
@@ -949,7 +952,8 @@ Lret: {}
         z = ((x - y * P1) - y * P2) - y * P3;
         const real zz = z * z;
 
-        if (zz > 1.0e-20L)
+        enum zzThreshold = (floatTraits!real.realFormat == RealFormat.ieeeDouble ? 1.0e-14L : 1.0e-20L);
```
ibuclaw: Store `floatTraits!real` in an alias instead of instantiating it multiple times in the same function?

std/digest/murmurhash.d
kinke: Possibly, not with me though; a first incarnation of this murmurhash patch is in https://github.com/dlang/phobos/pull/5358 (which I only realized after creating this PR). This one here still isn't fully satisfactory as casting a static array to another static array type may not be well-defined in the spec (haven't checked, but it's reinterpret_cast vs. conversion to new rvalue). The latest (LDC-specific) version without ambiguities is https://github.com/ldc-developers/phobos/pull/44.

std/math.d
kinke: Twice to be precise, and an `enum isRealDouble` would suffice. If you insist on adding a new line for shortening these 2 lines, I won't refuse. ;)

joakim-noah:
@kinke, you can remove the second and third commits relating to Android and `std.allocator`, as they were just merged in another pull.

kinke:
Rebased and slightly reworked.

std/digest/murmurhash.d
kinke: Now: `Check if we don't fill up a full block buffer.`

std/digest/murmurhash.d
kinke: Now in PascalCase.

std/math.d
kinke: There was a third usage I previously overlooked => `enum realFormat`.

std/datetime/timezone.d
```diff
@@ -2007,6 +2008,14 @@ public:
         // Android concatenates all time zone data into a single file and stores it here.
         enum defaultTZDatabaseDir = "/system/usr/share/zoneinfo/";
     }
+    else version(Solaris)
+    {
+        /++
+            The default directory where the TZ Database files are. It's empty
+            for Windows, since Windows doesn't have them.
+          +/
```
joakim-noah: Why does this comment need to be copy-pasted for every platform?

kinke: AFAICT just because the `version(TZDatabaseDir)` branch (which seems obsolete) doesn't start with the enum, but needs a prior import; otherwise, the comment could be placed once before the `version` branches. [Outside the scope of this PR though.]

wilzbach: It was added a week ago: https://github.com/dlang/phobos/pull/5966

wilzbach: FWIW we only support building the docs on `Posix` since 1.5 years, so adding comments in other branches isn't needed, but doesn't hurt either.
Anyhow @kinke said, it doesn't block this PR.

jmdavis: The ddoc comment really should be put in a StdDdoc version block and removed from the other versions.

joakim-noah: I see no reason to put this off for later, I'd rather it were at the top with a StdDdoc version or just as a comment.  Just change it to say that Windows doesn't have one and TZDatabaseDir lets you set one at compile-time in an external file.

andralex: Who can take care of this in a follow up?

andralex:
@kinke autotester plz

wilzbach:
The failure is unrelated - see https://github.com/dlang/phobos/pull/6061

joakim-noah:
I'll fix the comments in another pull, if that's what you're asking.

std/digest/murmurhash.d
kinke: Just noticed `fill up a full ...` =>`Check if we don't fill up a whole block buffer.`

quickfur:
Are @MetaLang's concerns already addressed? Should we merge?

joakim-noah:
I think so.  It was just about a comment, which has been changed since.  Can probably dismiss his review if he doesn't.

MetaLang:
I still don't like the wording (_check if we don't_ is very awkward to read and try to understand), but if this is ready to go it's not worth holding up for a comment.

PR #5903, opened 2017-12-02
Update interfaces.d [rebased to stable]

wilzbach:
See https://github.com/dlang/phobos/pull/5901

CC @Dashster

### DIFF
--- a/std/range/interfaces.d
+++ b/std/range/interfaces.d
@@ -42,6 +42,7 @@ $(BOOKTABLE ,
     $(TR $(TD $(LREF OutputRangeObject))
         $(TD Class that implements the $(D OutputRange) interface and wraps the
         $(D put) methods in virtual functions.
+    ))
     $(TR $(TD $(LREF outputRangeObject))
         Convenience function for creating an $(D OutputRangeObject) with a base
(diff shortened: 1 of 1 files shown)


JackStouffer:
On my phone atm but the new version looks incorrect 
216b56af-4197-457d-8f38-f289a8299663

Dashster:
Sorry, it looks like my requested change was incomplete. 

Line 46 in the original file, 47 in the changed file is:

"Convenience function for creating an $(D OutputRangeObject) with a base"

I believe it should have a "$(TD" prefixed:

"$(TD Convenience function for creating an $(D OutputRangeObject) with a base"

Also, line 58/59 

"$(TD Returns the interface type that best matches the range.)"

has a closing parentheses at the end that I think should be removed as there is one for that purpose as part of the next line.

schveiguy:
@Dashster thx, added your suggestion. We'll see how it looks.

schveiguy:
I see a stray closing parenthesis, will investigate...

schveiguy:
> Also, line 58/59

oops, didn't see this at first, so your suggestions were complete after all :) Thanks @Dashster

PR #5904, opened 2017-12-03
[Trivial] improve documentation of popBackN

JohanEngelen:
note that the popBackN doc is shared with popFrontN. This change moves the description of popBackN up, such that it is closer to where the reader will want to find it. Also adds more explicit mention of popFrontN, such that it is clear which one of the two is being described.

offtopic: About the `O(.)` part, I think it should be `O(O(length) + O(slice)` and `O(n * O(popFront))`?

### DIFF
--- a/std/range/primitives.d
+++ b/std/range/primitives.d
@@ -1636,18 +1636,18 @@ if (isInputRange!Range)
 }
 
 /**
-    Eagerly advances $(D r) itself (not a copy) up to $(D n) times (by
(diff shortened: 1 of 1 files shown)


ZombineDev:
> offtopic: About the O(.) part, I think it should be O(O(length) + O(slice) and O(n * O(popFront))?

The contract states that implementations of the range interface should implement all range primitives in `O(1)` time, AFAIK.

PR #5905, opened 2017-12-04
Limitation note to lockstep and zip.

dukc:
The added lines explain it all.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -3946,8 +3946,8 @@ private alias lengthType(R) = typeof(R.init.length.init);
         An `Exception` if all of the _ranges are not the same length and
         `sp` is set to `StoppingPolicy.requireSameLength`.
 
-    Limitations: The compiler cannot infer @nogc or nothrow attributes for this range
-    because $(LREF StoppingPolicy) is not known at compile time.
+    Limitations: The compiler cannot infer the `@nogc` or `nothrow` attributes for this range
+    because $(LREF StoppingPolicy) is not known during compilation.
 */
 struct Zip(Ranges...)
 if (Ranges.length && allSatisfy!(isInputRange, Ranges))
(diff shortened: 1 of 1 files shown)


quickfur:
LGTM

std/range/package.d
```diff
@@ -4663,6 +4667,10 @@ private string lockstepMixin(Ranges...)(bool withIndex, bool reverse)
 
    By default $(D StoppingPolicy) is set to $(D StoppingPolicy.shortest).
 
+   Limitations: Compiler cannot currently infer purity, safety, use of garbage
+   collection nor possibility of an exception from iteration of a lockstep.
+   $(LREF zip) can infer the former two due to a different implementation.
+
    See_Also: $(LREF zip)
```
n8sh: Consider changing "nor" to "or" and "former two" to "first two".

std/range/package.d
```diff
@@ -3945,6 +3945,10 @@ private alias lengthType(R) = typeof(R.init.length.init);
     Throws:
         An `Exception` if all of the _ranges are not the same length and
         `sp` is set to `StoppingPolicy.requireSameLength`.
+
+    Limitations: Compiler always has to assume use of garbage collection
```
MetaLang: I think we should word this in the opposite way for clarity. Ex:

```
Limitations: the compiler cannot guarantee @nogc or nothrow for this function because StoppingPolicy is not known at compile time.
```

Thoughts?

std/range/package.d
```diff
@@ -4663,6 +4667,10 @@ private string lockstepMixin(Ranges...)(bool withIndex, bool reverse)
 
    By default $(D StoppingPolicy) is set to $(D StoppingPolicy.shortest).
 
+   Limitations: Compiler cannot currently infer purity, safety, use of garbage
```
MetaLang: Ditto.

std/range/package.d
dukc: I'm bad at being clear so I better heed this.

std/range/package.d
n8sh: Word order side, I like the phrasing "cannot infer" more than "cannot guarantee".

dukc: Lol, you do all the hard work for me making the phrases. Perhaps that means I'm doing a bit too small PRs.

std/range/package.d
dukc: How so? Unlike the previous one, this one is already in the negative can-not form instead of has-to.

MetaLang: I mean, use the attributes directly. Purity -> `pure`, safety -> `@safe`, etc.

dukc:
Autotest failure, seems incidental. Trying to restart...

PR #5906, opened 2017-12-04
Added more attributes to unittests so they can catch more regressions.

dukc:
More is more here, no?

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -5007,20 +5007,20 @@ struct Recurrence(alias fun, StateType, size_t stateSize)
 }
 
 ///
-pure @safe nothrow @nogc unittest
+pure @safe nothrow unittest
 {
     import std.algorithm.comparison : equal;
 
     // The Fibonacci numbers, using function in string form:
     // a[0] = 1, a[1] = 1, and compute a[n+1] = a[n-1] + a[n]
     auto fib = recurrence!("a[n-1] + a[n-2]")(1, 1);
-    assert(fib.take(10).equal(only(1, 1, 2, 3, 5, 8, 13, 21, 34, 55)));
+    assert(fib.take(10).equal([1, 1, 2, 3, 5, 8, 13, 21, 34, 55]));
 
     // The factorials, using function in lambda form:
     auto fac = recurrence!((a,n) => a[n-1] * n)(1);
-    assert(take(fac, 10).equal(only(
+    assert(take(fac, 10).equal([
         1, 1, 2, 6, 24, 120, 720, 5040, 40320, 362880
-    )));
+    ]));
 
     // The triangular numbers, using function in explicit form:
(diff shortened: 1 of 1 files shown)


std/range/package.d
```diff
@@ -6269,15 +6269,22 @@ FrontTransversal!(RangeOfRanges, opt) frontTransversal(
...
+pure @safe nothrow unittest
 {
     int[][] arr = [[], []];
 
     auto ft1 = frontTransversal!(TransverseOptions.assumeNotJagged)(arr);
     assert(ft1.empty);
 
-    auto ft2 = frontTransversal!(TransverseOptions.enforceNotJagged)(arr);
-    assert(ft2.empty);
+    try
```
MetaLang: It's not worth having `nothrow` on the unittest if it makes the example this ugly.

std/range/package.d
```diff
@@ -5006,28 +5007,28 @@ struct Recurrence(alias fun, StateType, size_t stateSize)
...
-@safe unittest
+pure @safe nothrow @nogc unittest
 {
     import std.algorithm.comparison : equal;
 
     // The Fibonacci numbers, using function in string form:
     // a[0] = 1, a[1] = 1, and compute a[n+1] = a[n-1] + a[n]
     auto fib = recurrence!("a[n-1] + a[n-2]")(1, 1);
-    assert(fib.take(10).equal([1, 1, 2, 3, 5, 8, 13, 21, 34, 55]));
+    assert(fib.take(10).equal(only(1, 1, 2, 3, 5, 8, 13, 21, 34, 55)));
```
JackStouffer: I don't think we should complicate ddoc-ed unit tests

std/range/package.d
MetaLang: I didn't realize this was not a DDOC'd unit test, so it's not as big of a deal, but I still don't like this and don't really see much of a point.

jmdavis: The test can be split into a separate `unittest` block if it can't be `nothrow`.

std/range/package.d
dukc: I can revert that, but this means I also have to take off the nogc attribute I added.

dukc:
Even that style article itself does not obey that in the other guidelines, lol!

std/range/package.d
JackStouffer: Yes, I believe so. Making the ddoc-ed unit tests simpler is always better IMO. If you want to, you can always add a separate unit test block for the express purpose of testing `@nogc`, e.g. https://github.com/dlang/phobos/blob/9503a1e11ed9c474464b1ea6e5b07391407b18d3/std/range/package.d#L11862 (I would merge this even without that addition though).

dukc:
Autotester stuck. Restarting.

PR #5907, opened 2017-12-05
Support std.typecons.Tuple expansion in 'any' and 'all'

FeepingCreature:
It would be convenient if ranges that iterated over typecons Tuples could be used with predicates that took multiple parameters. This PR adds tests and implementation for 'all' and 'any' to start, by changing 'find' to try Tuple expansion and 'any' and 'all' to check more relaxedly.

### DIFF
--- a/std/algorithm/searching.d
+++ b/std/algorithm/searching.d
@@ -117,16 +117,15 @@ Checks if $(I _all) of the elements verify $(D pred).
  +/
 template all(alias pred = "a")
 {
+    import std.functional : not;
     /++
     Returns $(D true) if and only if $(I _all) values $(D v) found in the
     input _range $(D range) satisfy the predicate $(D pred).
     Performs (at most) $(BIGOH range.length) evaluations of $(D pred).
      +/
     bool all(Range)(Range range)
-    if (isInputRange!Range && is(typeof(unaryFun!pred(range.front))))
+    if (isInputRange!Range && is(typeof(find!(not!(unaryFun!pred))(range))))
     {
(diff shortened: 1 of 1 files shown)


MetaLang:
I dunno about this... auto-expansion for tuples in `foreach` loops is considered to have been a mistake; I don't know if we want to codify it in Phobos as well.

FeepingCreature:
Personally I feel it's something that seems like it _should_ work if viewed intuitively. If I'm looking at a range of tuples, I'm rarely interested in the tuples as values, but usually in their individual members.

wilzbach:
> Personally I feel it's something that seems like it should work if viewed intuitively. 

Well, yes `10.iota.enumerate.find!(e => e.value == 7)` isn't as powerful as `10.iota.enumerate.find!((idx, val) => val == 7)`, but note that with the extra braces needed you actually don't save keystrokes - especially if you consider that you could write `10.iota.enumerate.find!(e => e[1] == 7)` too.

Anyhow, the biggest problem with this is that once we start doing this for a few ranges, a user would expect it to work for all ranges. Even if we say we do this for Phobos, there are functions that accept a `pred` that could be unary or binary as well, so we might run into problems there.

FeepingCreature:
I'd been trying to see if there was some way to support this within unaryPred, but I couldn't figure it out. There doesn't seem to be a way to, depending on arguments, either define a function or an alias, in a way that still works with ifti.

andralex:
This seems to add a special case that needs more explanation than just following the simpler rules. I suggsest closing.

PR #5908, opened 2017-12-05
Added alignedAllocate for AscendingPageAllocator

jercaianu:
I also added and refactored some tests for the new functionality.

### DIFF
--- a/std/experimental/allocator/building_blocks/ascending_page_allocator.d
+++ b/std/experimental/allocator/building_blocks/ascending_page_allocator.d
@@ -101,13 +101,14 @@ public:
     {
         import std.algorithm.comparison : min;
 
+        immutable pagedBytes = numPages * pageSize;
         size_t goodSize = goodAllocSize(n);
-        if (goodSize > numPages * pageSize || offset - data > numPages * pageSize - goodSize)
+        if (goodSize > pagedBytes || offset - data > pagedBytes - goodSize)
             return null;
 
         if (offset + goodSize > readWriteLimit)
         {
-            void* newReadWriteLimit = min(data + numPages * pageSize, offset + goodSize + extraAllocPages * pageSize);
+            void* newReadWriteLimit = min(data + pagedBytes, offset + goodSize + extraAllocPages * pageSize);
             if (newReadWriteLimit != readWriteLimit)
             {
                 assert(newReadWriteLimit > readWriteLimit);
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -101,13 +101,14 @@ public:
     {
         import std.algorithm.comparison : min;
 
+        immutable pagedBytes = numPages * pageSize;
         size_t goodSize = goodAllocSize(n);
-        if (goodSize > numPages * pageSize || offset - data > numPages * pageSize - goodSize)
+        if (goodSize > pagedBytes || offset - data > pagedBytes - goodSize)
```
andralex: `if (goodSize > pagedBytes || offset + goodSize > data + pagedBytes)` would be better from a defensive viewpoint (no subtraction of unsigned values) but you decide whether it's at all sensible :)

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -101,13 +101,14 @@ public:
...
             return null;
 
         if (offset + goodSize > readWriteLimit)
```
andralex: ah, so here we have again `offset + goodSize` as a meaningful quantity

std/experimental/allocator/building_blocks/ascending_page_allocator.d
```diff
@@ -144,6 +145,38 @@ public:
...
+    `null` on failure or if the requested size exceeds the remaining capacity.
+    */
+    void[] alignedAllocate(size_t n, uint a)
+    {
+        void* alignedStart = cast(void*) roundUpToMultipleOf(cast(size_t) offset, a);
+        assert(alignedStart.alignedAt(a));
+        immutable pagedBytes = numPages * pageSize;
+        size_t goodSize = goodAllocSize(n);
+        if (goodSize > pagedBytes ||
+            alignedStart - data > pagedBytes - goodSize)
```
andralex: one more indent here or full block

std/experimental/allocator/building_blocks/ascending_page_allocator.d
jercaianu: When I did this, I was afraid that some sort of an overflow could happen.
By doing substraction and knowing that ``offset > data`` and ``pagedBytes > goodSize`` I didnt have to do any overflow checks.

PR #5911, opened 2017-12-08
Fix for two problems with format separators

burner:
fix issue 17459
fix issue 18047

This is in one commit as the two problem are located at the place

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -1994,7 +1994,7 @@ private void formatUnsigned(Writer, T, Char)
             + (prefix2 != 0)
             + zerofill
             + digits.length
-            + ((fs.flSeparator != 0) * (digits.length / fs.separators))
+            + ((fs.flSeparator != 0) * ((digits.length - 1)/ fs.separators))
         );
     if (spacesToPrint > 0) // need to do some padding
     {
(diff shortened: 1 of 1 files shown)


std/format.d
```diff
@@ -1995,6 +1995,7 @@ private void formatUnsigned(Writer, T, Char)
             + zerofill
             + digits.length
             + ((fs.flSeparator != 0) * (digits.length / fs.separators))
+            - (fs.flSeparator != 0 && digits.length % fs.separators == 0)
```
andralex: `((fs.flSeparator != 0) * ((digits.length - 1) / fs.separators))` replaces the two lines and is quite a bit simpler

std/format.d
```diff
@@ -2013,8 +2014,34 @@ private void formatUnsigned(Writer, T, Char)
...
+            if (j != fs.width && j % fs.separators == 0)
+            {
+                put(w, fs.separatorChar);
+                ++i;
+            }
+            if (i < zerofill)
+            {
+                   put(w, '0');
+            }
+            --j;
```
andralex: move to loop increment

std/format.d
burner: fxied

std/format.d
burner: and fixed

std/format.d
```diff
@@ -2013,8 +2013,33 @@ private void formatUnsigned(Writer, T, Char)
...
-        put(w, '0');
+    if (fs.flSeparator)
+    {
+        if (zerofill > 0)
+        {
+            put(w, '0');
+            --zerofill;
+        }
+
+        int j = fs.width;
```
andralex: should this be uint?

burner: FormatSpec.width is an int. So I think int is good

PR #5912, opened 2017-12-08
Added allocateFresh for BitmappedBlock

jercaianu:
allocateFresh returns unused memory from a BitmappedBlocks.
This is just a puzzle piece, together with AscendingPageAllocator, to create a fast and safe allocator which does not reuse memory.

### DIFF
--- a/std/experimental/allocator/building_blocks/bitmapped_block.d
+++ b/std/experimental/allocator/building_blocks/bitmapped_block.d
@@ -129,6 +129,11 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
     private BitVector _control;
     private void[] _payload;
     private size_t _startIdx;
+
+    // Keeps track of first block which has never been used in an allocation.
+    // All blocks which are located right to the '_freshBit', should have never been
+    // allocated
+    private ulong _freshBit;
     // }
 
     pure nothrow @safe @nogc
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -246,6 +251,15 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
         }
     }
 
+    /*
+    Based on the latest allocated bit, 'newBit', it adjusts '_freshBit'
+    */
+    void adjustFreshBit(const ulong newBit)
+    {
+        import std.algorithm.comparison : max;
+        _freshBit = max(newBit, _freshBit);
```
andralex: `if (newBit > _freshBit) _freshBit = newBit;` seems to do better for performance, see https://goo.gl/WeQ1KS and https://godbolt.org/g/YUnzNw

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -255,6 +269,7 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
         assert(msbIdx <= 63);
         const start = (wordIdx * 64 + msbIdx) * blockSize;
         const end = start + blockSize * howManyBlocks;
+        if (start == end) return null;
```
andralex: This seems to fix an unrelated bug?

PR #5913, opened 2017-12-08
Fixed 18048 std.bigint.toDecimalString is impure

Biotronic:


Biotronic:
CircleCI is segfaulting. I don't see anything here that should do that, but I'm just a hairy northerner, so I could be wrong.

wilzbach:
> CircleCI is segfaulting. I don't see anything here that should do that, but I'm just a hairy northerner, so I could be wrong.

You won't believe this, but you aren't wrong. We had some issues with random segfaults of DScanner lately. I just restarted your build that usually makes the ugly red cross go away.

std/bigint.d
```diff
@@ -1082,16 +1082,16 @@ Returns:
     A $(D string) that represents the $(D BigInt) as a decimal number.
 
 */
-string toDecimalString(const(BigInt) x)
+string toDecimalString(const(BigInt) x) pure
```
wilzbach: This function should be `deprecated`. `myBigInt.to!string` is so much more idiomatic D.

JackStouffer: Problem is, that `BigInt.toString` uses the `delegate sink` method of `toString`, which makes it impossible to be `pure`.

I've been meaning to write up a bug report on how std.format should be reworked so that a possible `toString` overload of

```d
void toString(Writer)(auto ref Writer w) if (isOutputRange!(Writer, char))
```

is given precedence over every other `toString` overload.  The above signature allows all attributes to be inferred, along with fitting much better into the range paradigm.

JackStouffer: Is this also `nothrow`?

schveiguy: There is another way to do it. And that is, to pass the char buffer directly to toString, and let it work with that buffer. The only problem is, you need some way to communicate back to the caller if you don't have enough space.

I've thought of this since implementing iopipe, where I try to do everything directly in the buffer. The output range method works as well, but now the toString function is responsible for buffer management. For some things it works well, for others it may not.

The reason we opted (a long time ago) to use a char[] delegate instead of an output range is to allow virtual toString functions to work. Not sure it's as relevant today.

std/bigint.d
```diff
@@ -1082,16 +1082,16 @@ Returns:
...
 {
-    string outbuff="";
-    void sink(const(char)[] s) { outbuff ~= s; }
-    x.toString(&sink, "%d");
-    return outbuff;
+    auto buff = x.data.toDecimalString(0);
+
+    if (x.isNegative()) return '-' ~ buff;
+    return buff;
```
schveiguy: `BigUint.toDecimalString` takes a parameter telling it how much "extra" padding to put at the front. I think this is specifically to allow adding a '-' sign. We should take advantage of that here:

```D
auto buff = x.data.toDecimalString(x.isNegative() ? 1 : 0);
if(x.isNegative())
    buff[0] = '-';
```

std/bigint.d
wilzbach: There's even another: instead of returning a string, you return a range. If combined with a yielding Fiber, it's equally attractive syntaxwise and deals with the problem of passing the writer object around.

schveiguy: > instead of returning a string, you return a range.

I think the toString version that receives an output range is more optimal, simply because then you can avoid dealing with buffering (if you don't need it), and the function could be implemented to take advantage of any optimizations based on the output range type. For example, if the output range given is an array, write directly to the array instead of using it as an output range.

Simple thought exercise, how would you return a toString range from an integer? Typically you construct the string from the right-most digit first! This does not work well in range form.

JackStouffer: Also, the main attraction of the output range overload (other than attributes) is that it allows the user to control the memory used.

schveiguy:
I didn’t know and find it interesting that you can just return a char[] as a string. Cool!

PR #5914, opened 2017-12-08
Run DScanner with gdb to catch the stack on a SEGFAULT

wilzbach:
As neither @MartinNowak nor I can reproduce the spurious DScanner failures locally, getting a stack trace on a SEFAULT is probably the best thing we can do.

CC @CyberShadow @ZombineDev @JackStouffer

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -545,7 +545,7 @@ style: publictests style_lint
 # runs static code analysis with Dscanner
 dscanner: | $(DSCANNER_DIR)/dsc
 	@echo "Running DScanner"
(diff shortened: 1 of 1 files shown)


JackStouffer:
Is the dscanner executable you're using on Circle compiled with debug symbols?

wilzbach:
> Is the dscanner executable you're using on Circle compiled with debug symbols?

Yeah, check the target (it's a couple of lines above):

```
DC=$(DMD) DFLAGS="$(DFLAGS) -defaultlib=$(LIB)" make -C $(DSCANNER_DIR) githash debug
```

PR #5915, opened 2017-12-09
Fix std.utf for -dip1000

wilzbach:
AFAICT we have finally reached the point.

### DIFF
--- a/std/utf.d
+++ b/std/utf.d
@@ -4091,7 +4091,19 @@ if (isSomeChar!C)
 
                 static if (isForwardRange!R)
                 {
+                    static if (hasIndirections!R)
                     @property auto save() return scope
+                    /* `return scope` cannot be inferred because compiler does not
+                     * track it backwards from assignment to local `ret`
+                     */
+                    {
+                        auto ret = this;
+                        ret.r = r.save;
+                        return ret;
+                    }
+
+                    static if (!hasIndirections!R)
+                    @property auto save() scope
                     /* `return scope` cannot be inferred because compiler does not
                      * track it backwards from assignment to local `ret`
                      */


wilzbach:
I should have written more description about this PR. This only enables building Phobos with `-dip1000`, the testsuite still throws errors (one step at a time).
However, it looks like building with `-dip1000` changes the mangling (?) and thus leads to an error while building `d_do_test` during the dmd build stage:

```
generated/d_do_test.o: In function `_D3std3uni__T8CowArrayTSQwQu8GcPolicyZQz__T6__ctorTAkZQlMFNaNbNcNfQpZSQCqQCp__TQCoTQCiZQCw':
d_do_test.d:(.text._D3std3uni__T8CowArrayTSQwQu8GcPolicyZQz__T6__ctorTAkZQlMFNaNbNcNfQpZSQCqQCp__TQCoTQCiZQCw[_D3std3uni__T8CowArrayTSQwQu8GcPolicyZQz__T6__ctorTAkZQlMFNaNbNcNfQpZSQCqQCp__TQCoTQCiZQCw]+0x60): undefined reference to `_D3std9algorithm8mutation__T4copyTAkTQdZQmFNaNbNiNfQrQtZQw'
generated/d_do_test.o: In function `_D3std5regex8internal6parser__T6ParserTAyaTSQBqQBpQBmQBg7CodeGenZQBi11parseEscapeMFNeZv':
d_do_test.d:(.text._D3std5regex8internal6parser__T6ParserTAyaTSQBqQBpQBmQBg7CodeGenZQBi11parseEscapeMFNeZv[_D3std5regex8internal6parser__T6ParserTAyaTSQBqQBpQBmQBg7CodeGenZQBi11parseEscapeMFNeZv]+0x7d5): undefined reference to `_D3std3uni__T5StackTkZQj3topMFNaNbNcNdNiNfZk'
```

So I guess we can't go one step at a time?

CyberShadow:
Huh, isn't that bad? Doesn't this mean that if we start building Phobos with -dip1000, all D users would then be forced to start using -dip1000 too?

nordlow:
What about making -dip25 default aswell? Does -dip1000 include -dip25?

schveiguy:
The mangling has to be identical, or we can't do a dip1000 incremental change. Is it possible to change the mangling, but not enforce the semantics for dip1000?

schveiguy:
> It is especially fun once attribute inference gets involved

ugh. This is going to cause a lot of problems. We have 3 options I think at this point:

1. Leave the name mangling changes out, and re-introduce them once -dip1000 becomes permanent
2. Somehow detect the attributes for mangling purposes only, and only enforce the semantics of dip1000 when the switch is given.
3. Ship multiple versions of libraries.

Option 3 seems the easiest, but also the worst user experience.

ZombineDev:
@klickverbot what problems do you anticipate? I don't see problem with e.g. compiling phobos with `-dip1000` when running the tests, while shipping the final library build without `-dip1000`.

ZombineDev:
@nordlow `-dip1000` includes `-dip25` - see https://github.com/dlang/dmd/blob/64424027452125693483b36022827e6c401e2307/src/dmd/mars.d#L1956.

JackStouffer:
Whatever approach we decide to take towards DIP1000, simply throwing the switch seems like the wrong option. Closing to free up the PR queue; please reopen if you want to continue the discussion.

wilzbach:
We could have at least kept the existing `-dip1000` fixes. Rebased to that. Though I would like to wait until https://github.com/dlang/phobos/pull/6278 has been merged, s.t. we can ensure that we actually make progress.

std/utf.d
```diff
@@ -4100,6 +4101,17 @@ if (isSomeChar!C)
                         ret.r = r.save;
                         return ret;
                     }
+
+                    static if (!hasIndirections!R)
+                    @property auto save() scope
+                    /* `return scope` cannot be inferred because compiler does not
```
JackStouffer: change comment to remove mention of return

wilzbach:
Closing this and submitting this as a new approach with `-dip1000.mak`

PR #5917, opened 2017-12-11
Changed zip and lockstep of std.range to assert same length

dukc:
…instead of enforcing. This is of course controversal since it's breaking but I decided it's worth a discussion. Isn't it after all clearly a bug if zipped ranges with `StoppingPolicy.RequireSameLength` have different lengths?

I made the same change for enforcement of `!stoppingPolicy != StoppingPolicy.Longest` for `LockStep` and enforcement of default-initializable elemets.

The advantages are that functions can be faster in production code and, in case of `zip`, `nothrow` can be inferred.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -3993,7 +3993,7 @@ private alias lengthType(R) = typeof(R.init.length.init);
         An `Exception` if all of the _ranges are not the same length and
         `sp` is set to `StoppingPolicy.requireSameLength`.
 
-    Limitations: The `@nogc` or `nothrow` attributes for this range cannot be inferred
+    Limitations: The `@nogc` attribute for this range cannot be inferred
     because $(LREF StoppingPolicy) is not known during compilation.
 */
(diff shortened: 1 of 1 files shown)


dukc:
Such code is doubtlessly bug-prone since the length property does not check for all ranges, resulting in overflow sooner or later. But that makes the code no less existent...

JackStouffer:
I can tell you from experience that Andrei will reject this due to its breaking nature, and I'd have to agree with him.

I would refactor this to add a new `StoppingPolicy` which gives this behavior.

dukc:
Good idea, I think. Does not let us infer `nothrow`, though.

Already 3 to be considered against. Enough to convince me off.

wilzbach:
FWIW [DIP1008](https://github.com/dlang/dmd/pull/6681) has been merged two days ago. So while you might not get `nothrow`, inference of `@nogc` due to Exceptions being `@nogc` might happen very soon ;-)

PR #5918, opened 2017-12-12
[DDoc] Avoid auto-escaping of StopWatch in std.datetime

wilzbach:
Ddoc's annoying auto-escaping strikes again!

Before:

image

https://dlang.org/phobos/std_datetime.html#.StopWatch

After:

image


jmdavis:
Killing the auto-escaping would be _way_ more valuable than doing anything with adding any kind of markdown to ddoc. :|

PR #5919, opened 2017-12-12
Improve wording in the Transposed changelog entry

wilzbach:
The current changelog can be previewed at https://dlang.org/changelog/pending.html

@ reviewers DAutotest builds the pending changelog too, s.t. changes to it can be conveniently previewed.

In this case, just dropping the Bugzilla link as plain text looks a bit weird.

Before:

image

After:

image


JackStouffer:
Looking at the pending changelog, the entry on `Ternary` needs some work as well. I'll open a quick PR

PR #5922, opened 2017-12-13
Move deprecations along: std.range.hasLength

wilzbach:


jmdavis:
The tests need to be fixed.

        struct A { ulong length; }
        static assert( hasLength!(A));

will not pass on 32-bit systems, so the auto-tester is failing. Also, tests should probably be added to ensure that `long` and `int` never work and so that `long` doesn't work on 32-bit systems, and `int` doesn't work on 32-bit systems. As it is, they really don't test that types other than `size_t` don't work, just that `size_t` works.

PR #5191, opened 2017-02-24
Simply the docs of std.format.unformatValue

JackStouffer:
This PR brings `unformatValue` docs to the same style as the docs for `std.conv.to`. The user doesn't need to know the exact overloads that they can use, it's covered in the docs. And if they get it wrong, the compiler still prints out the overloads of `unformatValueImpl`.

PR #4208, opened 2016-04-17
Make all toImpl overloads private

JackStouffer:
http://forum.dlang.org/post/hapxiqrdmjdiavqlpzlv@forum.dlang.org

PR #5923, opened 2017-12-13
Unify overloads and documentation for formatValue

JackStouffer:
In the vein of https://github.com/dlang/phobos/pull/5191 and https://github.com/dlang/phobos/pull/4208, this hides all of the individual overloads of `formatValue` and brings the documentation and examples into one place.

This is especially useful in the ddox version of the docs, as every overload will no longer be shown on a different page.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -1715,19 +1715,286 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
     assertThrown(singleSpec("%2.3eTest"));
 }
 
+/**
+ * Formats any value into `Char` accepting `OutputRange`, using the given `FormatSpec`.
+ *
+ * Aggregates:
+ * `struct`, `union`, `class`, and `interface` are formatted by calling `toString`.
+ *
+ * `toString` should have one of the following signatures:
+ *
+ * ---
+ * const void toString(scope void delegate(const(char)[]) sink, FormatSpec fmt);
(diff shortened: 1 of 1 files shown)


std/format.d
```diff
@@ -1716,15 +1716,286 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
...
+    {
+        int* p = new int;
+        return &foo;
+    }
+
+    assert(to!string(&bar) == "int delegate(short) @nogc delegate() pure nothrow @system");
+}
+
+/*
+    Bools
```
wilzbach: I would leave the full comment line here, e. g. 

> bool`s are formatted as `"true"` or `"false"` with `%s` and as `1` or `0` with integral-specific format specs.

In this case

JackStouffer:
@wilzbach Done

std/format.d
```diff
@@ -1716,15 +1716,287 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
...
+
+    int foo(short k) @nogc
+    {
+        return i + k;
+    }
+
+    @system int delegate(short) @nogc bar() nothrow pure
+    {
+        int* p = new int;
+        return &foo;
```
wilzbach: The function body seems unused. It could be dropped - this will then also increase the coverage and remove the red bits.

PR #5546, opened 2017-07-05
Trigger a hard error on deprecation messages

wilzbach:
https://github.com/dlang/phobos/pull/5083#issuecomment-313140290

> Yeah but it doesn't make sense otherwise, does it? Seems antithetical to deprecate something for everyone else but not ourselves.
> Any reason not to enable `-de` on the test suite then?

CC @CyberShadow

PR #5925, opened 2017-12-14
Fix intentional discarding of return value in topN

JohanEngelen:
This fixes a warning one gets when compiling with warnings enabled:
"phobos/std/algorithm/sorting.d(3122): Warning: calling testcase.foo.__lambda1 without side effects discards return value of type bool, prepend a cast(void) if intentional"

Testcase that fails compilation with `-w`:
```
import std.algorithm.sorting;

auto foo() {
    int[] v = [ 25, 7, 9, 2, 0, 5, 21 ];
    return topN!((a, b) => a > b)(v, 100);
}
```

Phobos unittests are compiled with `-w` right? In that case, the expanded testcase takes care of testing this fix.

### DIFF
--- a/std/algorithm/sorting.d
+++ b/std/algorithm/sorting.d
@@ -3119,7 +3119,7 @@ if (isRandomAccessRange!(Range) && hasLength!Range && hasSlicing!Range)
         // Workaround for https://issues.dlang.org/show_bug.cgi?id=16528
         // Safety checks: enumerate all potentially unsafe generic primitives
         // then use a @trusted implementation.
-        binaryFun!less(r[0], r[r.length - 1]);
+        cast(void) binaryFun!less(r[0], r[r.length - 1]);
         import std.algorithm.mutation : swapAt;
         r.swapAt(size_t(0), size_t(0));
         static assert(is(typeof(r.length) == size_t));
--- a/std/algorithm/sorting.d
+++ b/std/algorithm/sorting.d
@@ -3137,7 +3137,7 @@ if (isRandomAccessRange!(Range) && hasLength!Range && hasSlicing!Range)
     topN!"a < b"(v, 100);
     assert(v == [ 25, 7, 9, 2, 0, 5, 21 ]);
     auto n = 4;
-    topN!"a < b"(v, n);
+    topN!((a, b) => a < b)(v, n);
     assert(v[n] == 9);
 }
 


wilzbach:
> Phobos unittests are compiled with -w right? In that case, the expanded testcase takes care of testing this fix.

Yes - you can easily check this yourself:

```
>  make -f posix.mak std/algorithm/sorting.test
...
../dmd/generated/linux/release/64/dmd -conf= -I../druntime/import  -w -de -dip25 -m64 ...
...
std/algorithm/sorting.d(3122): Warning: calling std.algorithm.sorting.__unittest_std_algorithm_sorting_d_3134_45.__lambda1 without side effects discards return value of type bool, prepend a cast(void) if intentional
make: *** [posix.mak:385: std/algorithm/sorting.test] Error 1
```

std/algorithm/sorting.d
```diff
@@ -3119,7 +3119,7 @@ if (isRandomAccessRange!(Range) && hasLength!Range && hasSlicing!Range)
         // Workaround for https://issues.dlang.org/show_bug.cgi?id=16528
         // Safety checks: enumerate all potentially unsafe generic primitives
         // then use a @trusted implementation.
-        binaryFun!less(r[0], r[r.length - 1]);
+        cast(void) binaryFun!less(r[0], r[r.length - 1]);
```
wilzbach: FWIW this looks like unfinished work. This is in an `if (false)` block ...

JohanEngelen:
Can one of you retrigger the failed testers so this is merged? Thanks.

ZombineDev:
Sure, done. Ping again if you need help.

JohanEngelen:
Thanks.

Dscanner crashes on this code??

ZombineDev:
Dscanner has random chance of crashing on any PR. We're running it under DScanner in order to try to narrow down the problem, as it doesn't reproduce locally...

wilzbach:
Sadly you aren't that special, it's DScanner has been segfaulting spuriously for quite a while.

See e.g. https://github.com/dlang/phobos/pull/5929

As I have actually restarted your PR more then three times yesterday, let's try to temporarily ignore the failing DScanner (-> https://github.com/dlang/phobos/pull/5933), s.t. we aren't blocked here.

JohanEngelen:
Thanks.
FYI, this issue may seem trivial, but actually it is very bad in practice: it prevents upgrading the compiler if one is building with `-w`, warnings==errors; and there is no viable workaround other than to not use that particular Phobos function with a lambda.
(it is similar for the deprecation messages that I see in the build log when building phobos/druntime !)

wilzbach:
> (it is similar for the deprecation messages that I see in the build log when building phobos/druntime !)

Hehe, I'm aware and you shouldn't see any for Phobos since this summer: https://github.com/dlang/phobos/pull/5546 (we now build with `-de` by default). 

There's a similar [PR for druntime](https://github.com/dlang/druntime/pull/1982), but I haven't figured out yet how to fix the deprecations for Martin's new `arrayop` package.

wilzbach:
Force-merging this due to our CircleCi/DScanner SEGFAULT issue. We should probably go with https://github.com/dlang/phobos/pull/5546 for now :/

PR #5926, opened 2017-12-15
Make imports in std.net.curl local & lazy

wilzbach:
A first step for https://github.com/dlang/phobos/pull/5916 - there's still much work to do.

### DIFF
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -157,6 +157,8 @@ Distributed under the Boost Software License, Version 1.0.
 module std.net.curl;
 
 public import etc.c.curl : CurlOption;
+import core.time : dur;
+import etc.c.curl : CURLcode;
 import std.concurrency : Tid;
 import std.range.primitives;
 import std.encoding : EncodingScheme;
@@ -165,6 +167,8 @@ import std.typecons : Flag, Yes, No, Tuple;
 
 version(unittest)
 {
+    import std.socket : Socket;
+
     // Run unit test with the PHOBOS_TEST_ALLOW_NET=1 set in order to
     // allow net traffic
     private struct TestServer
@@ -232,11 +236,8 @@ version(unittest)
         immutable(T)[] bdy;
     }
 
-    private Request!T recvReq(T=char, S)(S s)
+    private Request!T recvReq(T=char)(Socket s)
     {
-        import std.socket : Socket;
-        static assert (is(S == Socket), "Invalid type used. Use std.socket.Socket.");
-
         import std.algorithm.comparison : min;
         import std.algorithm.searching : find, canFind;
         import std.conv : to;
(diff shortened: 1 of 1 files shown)


std/net/curl.d
```diff
@@ -156,36 +156,26 @@ Distributed under the Boost Software License, Version 1.0.
...
-import std.typecons;
-
-import std.internal.cstring;
-
 public import etc.c.curl : CurlOption;
+import std.concurrency : Tid;
+import std.range.primitives;
+import std.encoding : EncodingScheme;
+import std.traits : isSomeChar;
+import std.typecons : Flag, Yes, No, Tuple;
```
wilzbach: So when do we get [DIP1005](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1005.md)?

std/net/curl.d
```diff
@@ -235,8 +232,11 @@ version(unittest)
         immutable(T)[] bdy;
     }
 
-    private Request!T recvReq(T=char)(Socket s)
+    private Request!T recvReq(T=char, S)(S s)
     {
+        import std.socket : Socket;
+        static assert (is(S == Socket), "Invalid type used. Use std.socket.Socket.");
```
wilzbach: I really wanted to avoid that there are different imports available during the unittest mode, hence this workaround.

std/net/curl.d
```diff
@@ -4300,7 +4352,7 @@ struct Curl
                 interleavefunction, chunk_data, chunk_bgn_function,
                 chunk_end_function, fnmatch_data, fnmatch_function, cookiejar, postfields);
 
-            foreach (option; tt)
+            static foreach (option; tt)
```
CyberShadow: I would generally advise against mixing different kinds of changes in one commit.

std/net/curl.d
```diff
@@ -313,7 +313,11 @@ version(unittest)
 version(StdDdoc) import std.stdio;
 
 // Default data timeout for Protocols
-private enum _defaultDataTimeout = dur!"minutes"(2);
+private auto _defaultDataTimeout() {
+    import core.time : dur;
+    static res = dur!"minutes"(2);
```
CyberShadow: `static` is now going to make it take up space in TLS (and be initialized for each new thread), for something that should be a manifest constant...

std/net/curl.d
CyberShadow: Ew.

I think this cure is worse than the disease.

std/net/curl.d
```diff
@@ -4099,15 +4144,16 @@ class HTTPStatusException : CurlException
     immutable int status; /// The HTTP status code
 }
 
+import etc.c.curl : CURLcode;
```
JackStouffer: should be at the top of the file

PR #5927, opened 2017-12-15
Add a StdUnittest version

wilzbach:
There's still quite a bit of stuff in Phobos, that shouldn't be parsed & executed when Phobos is imported by a user who runs his tests.

For the motivation of excluding `unittest` in templates, please see https://wiki.dlang.org/DIP82

(I don't have any preferences on the name of this `version`.)

### DIFF
--- a/std/experimental/allocator/building_blocks/bitmapped_block.d
+++ b/std/experimental/allocator/building_blocks/bitmapped_block.d
@@ -50,6 +50,7 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
     import std.typecons : Ternary;
     import std.typecons : tuple, Tuple;
 
+    version(StdUnittest)
     @system unittest
     {
         import std.algorithm.comparison : max;
--- a/posix.mak
+++ b/posix.mak
@@ -138,6 +138,7 @@ endif
 ifdef ENABLE_COVERAGE
 DFLAGS  += -cov
 endif
+UDFLAGS=-unittest -version=StdUnittest
 
 # Set DOTOBJ and DOTEXE
 ifeq (,$(findstring win,$(OS)))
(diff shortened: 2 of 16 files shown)


CyberShadow:
Did we not have this or something very similar to this before? Am I thinking of another component?

CyberShadow:
Oh, I was probably thinking of `StdDdoc`.

In which case maybe make the two consistent.

JackStouffer:
bikesheading: any `version` you use should probably have two underscores prepended to it to avoid naming collisions with any existing code.

wilzbach:
> Oh, I was probably thinking of StdDdoc.
> In which case maybe make the two consistent.

Renamed to `StdUnittest`, which @JackStouffer I think is unique enough and considering that this sadly will be used often in Phobos I would prefer to avoid anything with two ugly `__`.
We could add a changelog entry?

wilzbach:
> I'd say let's just "reserve" all StdXyz identifiers by mentioning that in the Phobos docs, and then stick to that naming scheme.

Great idea!
-> https://github.com/dlang/dlang.org/pull/1999

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -50,6 +50,7 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
     import std.typecons : Ternary;
     import std.typecons : tuple, Tuple;
 
+    version(StdUnittest)
```
andralex: why do we need this here? I'd assume if no unittest is conducted, unittests are ignored anyway.

wilzbach: From DIP82:

> At present, unittest blocks inside of a templated type are compiled into every instantation, which only works with tests that are generic, which most tests aren't, and it almost never works with examples and thus doesn't work for ddoc-ed unittest blocks.

1) these unittest blocks might not even be run (I have a couple of these issues during the runnable examples story)
2) these unittest blocks will be executed as part of a user's testsuite

wilzbach:
> I think we need here a version(unittest)

This already exists, but it's defined during a user's test run too.

wilzbach:
FYI there have been two recent pull requests for which this identifier would have been useful:
- https://github.com/dlang/phobos/pull/5932 (already merged, I added an additional commit for it)
- https://github.com/dlang/phobos/pull/5951

So maybe we can simply move on with this? There are many other places in Phobos which should have used `StdUnittest` ...

PR #5930, opened 2017-12-15
Upgrade to CircleCi 2.0

wilzbach:
Looks fairly straight-forward so far.

Note that CircleCi 2.0 allows to use custom Docker images.
So for example, we could built the DMD compiler into it and avoid the network issues during it's download.
Also allows this other customization, e.g. sth. like @ZombineDev's PIE-hardened docker images (https://github.com/ZombineDev/dmd-test-suite-docker) to run all testsuites on a PIE-hardened system.

PR #5929, opened 2017-12-15
Run CircleCi in BUILD=debug

wilzbach:
To investigate the repeated, spurious failures on CircleCi.
This should allow us to get a better stack trace.
It also tweaks the gdb command.

PR #5920, opened 2017-12-12
[Trivial] Fix Link and Grammar in Ternary Changelog

JackStouffer:

PR #5933, opened 2017-12-16
Temporarily allow DScanner failures

wilzbach:
I really don't want to do this as we then most likely will forget about it, but the failures are getting more persistent :/

This works by replacing the Makefile target with `dscanner_disabled` and then executing it separately with the good old `|| true`.

See also:
- https://github.com/dlang/phobos/pull/5929
- https://github.com/dlang/phobos/pull/5930
- https://github.com/dlang/phobos/pull/5920#issuecomment-351355514


wilzbach:
Closing in favor of https://github.com/dlang/phobos/pull/5930 (Upgrade to CircleCi 2.0)

PR #5934, opened 2017-12-16
[DO NOT MERGE] Test Circleci with debug GC

wilzbach:
Trying @rainers's advice from https://github.com/dlang/phobos/pull/5929#issuecomment-352184482 to debug the DScanner failure.

### DIFF
--- a/circleci.sh
+++ b/circleci.sh
@@ -87,6 +87,9 @@ setup_repos()
         fi
     done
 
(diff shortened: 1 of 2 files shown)


wilzbach:
We now build with `BUILD=debug` so one
- one already sees the entire stacktrace nicely
- one can ssh into the machine easily and grab the file

-> closing.

PR #5937, opened 2017-12-17
Fix Issue 18086 - BigInt DivMod

Biotronic:
Made `divMod` accessible for `BigInt`. The function already exists in biguintcore, but is not publicly accessible.

### DIFF
--- a/std/bigint.d
+++ b/std/bigint.d
@@ -1836,3 +1836,65 @@ unittest
     enum BigInt test1 = BigInt(123);
     enum BigInt test2 = plusTwo(test1);
 }
+
+/**
+ * Finds the quotient and remainder for the given dividend and divisor in one operation.
+ *
+ * Params:
+ *     dividend = the $(LREF BigInt) to divide
+ *     divisor = the $(LREF BigInt) to divide the dividend by
+ *     quotient = is set to the result of the division
+ *     remainder = is set to the remainder of the division
+ */
+void divMod(const BigInt dividend, const BigInt divisor, out BigInt quotient, out BigInt remainder) pure nothrow
+{
+    BigUint q, r;
+    BigUint.divMod(dividend.data, divisor.data, q, r);
+    quotient.sign = dividend.sign != divisor.sign;
+    quotient.data = q;
+    remainder.sign = dividend.sign;
+    remainder.data = r;
+}
+
+///
+@system pure nothrow unittest
+{
+    auto a = BigInt(123);
+    auto b = BigInt(25);
+    BigInt q, r;
+
+    divMod(a, b, q, r);
+
+    assert(q == 4);
+    assert(r == 23);
+    assert(q * b + r == a);
+}
+
+// Issue 18086
+@system pure nothrow unittest
+{
+    BigInt q = 1;
+    BigInt r = 1;
+    BigInt c = 1024;
+    BigInt d = 100;
+
+    divMod(c, d, q, r);
+    assert(q ==  10);
+    assert(r ==  24);
+    assert((q * d + r) == c);
+
+    divMod(c, -d, q, r);
+    assert(q == -10);
+    assert(r ==  24);
+    assert(q * -d + r == c);
+
+    divMod(-c, -d, q, r);
+    assert(q ==  10);
+    assert(r == -24);
+    assert(q * -d + r == -c);
+
+    divMod(-c, d, q, r);
+    assert(q == -10);
+    assert(r == -24);
+    assert(q * d + r == -c);
+}
--- a/std/internal/math/biguintcore.d
+++ b/std/internal/math/biguintcore.d
@@ -847,6 +847,29 @@ public:
         return BigUint(removeLeadingZeros(trustedAssumeUnique(rem)));
     }
 
+    // Return x / y in quotient, x % y in remainder
+    static void divMod(BigUint x, BigUint y, out BigUint quotient, out BigUint remainder) pure nothrow
+    {
+        if (y.data.length > x.data.length)
+        {
+            quotient = 0uL;
+            remainder = x;
+        }
+        else if (y.data.length == 1)
+        {
+            quotient = divInt(x, y.data[0]);
+            remainder = BigUint([modInt(x, y.data[0])]);
+        }
+        else
+        {
+            BigDigit[] quot = new BigDigit[x.data.length - y.data.length + 1];
+            BigDigit[] rem = new BigDigit[y.data.length];
+            divModInternal(quot, rem, x.data, y.data);
+            quotient = BigUint(removeLeadingZeros(trustedAssumeUnique(quot)));
+            remainder = BigUint(removeLeadingZeros(trustedAssumeUnique(rem)));
+        }
+    }
+
     // return x op y
     static BigUint bitwiseOp(string op)(BigUint x, BigUint y, bool xSign, bool ySign, ref bool resultSign)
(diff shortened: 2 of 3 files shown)


std/bigint.d
```diff
@@ -1761,3 +1761,90 @@ unittest
...
+    quotient.sign = x.sign != y.sign;
+    quotient.data = q;
+    remainder.sign = x.sign;
+    remainder.data = r;
+}
+
+void divMod(T1, T2)(T1 x, T2 y, out BigInt quotient, out BigInt remainder)
+    if (is(T1 : const BigInt) || is(T2 : const BigInt) && is(typeof(BigInt(x))) && is(typeof(BigInt(y))))
+{
+    return divMod(BigInt(x), BigInt(y), quotient, remainder);
```
JackStouffer: static if this to only create a new big int when it’s not already one

std/bigint.d
```diff
@@ -1761,3 +1761,90 @@ unittest
...
+{
+    BigUint q, r;
+    BigUint.divMod(x.data, y.data, q, r);
+    quotient.sign = x.sign != y.sign;
+    quotient.data = q;
+    remainder.sign = x.sign;
+    remainder.data = r;
+}
+
+void divMod(T1, T2)(T1 x, T2 y, out BigInt quotient, out BigInt remainder)
```
JackStouffer: /// ditto

std/bigint.d
```diff
@@ -1761,3 +1761,90 @@ unittest
     auto b = immutable BigInt("123");
     assert(b == 123);
 }
+
+// Issue 18086
```
JackStouffer: Documentation?

std/bigint.d
```diff
@@ -1761,3 +1761,90 @@ unittest
...
+    remainder.data = r;
+}
+
+void divMod(T1, T2)(T1 x, T2 y, out BigInt quotient, out BigInt remainder)
+    if (is(T1 : const BigInt) || is(T2 : const BigInt) && is(typeof(BigInt(x))) && is(typeof(BigInt(y))))
+{
+    return divMod(BigInt(x), BigInt(y), quotient, remainder);
+}
+
+@system unittest
```
JackStouffer: This is pure and no throw right?

std/bigint.d
```diff
@@ -1761,3 +1761,90 @@ unittest
...
+{
+    import std.conv : to;
```
JackStouffer: This doesn’t seem used

std/bigint.d
Biotronic: Since BigInt is COW, the constructor call is basically free (it's an array assignment and a bool assignment - no copying takes place). Using static if to conditionally call the constructor complicates the code to no benefit.

Biotronic: However, creating two overloads on the form `divMod(T)(T x, const BigInt y, ...) if (is(typeof(BigInt(x))))` is more readable than the current template constraint.

std/bigint.d
```diff
@@ -1761,3 +1761,111 @@ unittest
     auto b = immutable BigInt("123");
     assert(b == 123);
 }
+
+/// Extracts the quotient and remainder for the given dividend and divisor.
```
JackStouffer:
```d
Finds the quotient and remainder for the given dividend and divisor in one operation.

Params:
    dividend = the number or $(LREF BigInt) to divide
    divisor = the number or $(LREF BigInt) to divide the dividend by
    quotient = is set to the result of the division
    remainder = is set to the remainder of the division
```

JackStouffer:
Ping @andralex, this is a new symbol

wilzbach:
> this is a new symbol

Also new symbols should come with a changelog entry (see the `/changelog` directory for details).

std/internal/math/biguintcore.d
```diff
@@ -706,6 +706,29 @@ public:
...
+            remainder = x;
+        }
+        else if (y.data.length == 1)
+        {
+            quotient = divInt(x, y.data[0]);
+            remainder = BigUint([modInt(x, y.data[0])]);
+        }
+        else
+        {
+            BigDigit [] quot = new BigDigit[x.data.length - y.data.length + 1];
```
wilzbach: Nit: `BigDigit[]`

std/bigint.d
```diff
@@ -1761,3 +1761,111 @@ unittest
...
+    BigUint q, r;
+    BigUint.divMod(dividend.data, divisor.data, q, r);
+    quotient.sign = dividend.sign != divisor.sign;
+    quotient.data = q;
+    remainder.sign = dividend.sign;
+    remainder.data = r;
+}
+
+/// ditto
+void divMod(T)(const BigInt dividend, T divisor, out BigInt quotient, out BigInt remainder) pure nothrow
```
wilzbach: Remove the attributes here, they are automatically inferred for templated functions and by manually specifying them, future improvements which e.g. would make this `@nogc` are overwritten.

std/bigint.d
```diff
@@ -1761,3 +1761,111 @@ unittest
     auto b = immutable BigInt("123");
     assert(b == 123);
 }
+
+/// Extracts the quotient and remainder for the given dividend and divisor.
+void divMod(const BigInt dividend, const BigInt divisor, out BigInt quotient, out BigInt remainder) pure nothrow
```
wilzbach: The common Phobos style is to avoid using overloads if not necessary as it overwhelms the user with documentation and it also allows to provide good error messages.
In your case as an additional benefit, it also allows you to get rid of hard-coding the attributes.
Use sth. like this instead when unifying the overloads:

```d
void divMod...
is(Dividend == BigInt) || is(Divisor == BigInt)
```

std/internal/math/biguintcore.d
```diff
@@ -706,6 +706,29 @@ public:
         return BigUint(removeLeadingZeros(trustedAssumeUnique(rem)));
     }
 
+    // Return x / y in quotient, x % y in remainder
```
wilzbach: Should add a warning comment that it doesn't handle signs...

wilzbach: Or maybe it's better to simply move this logic into this function, so that future callers don't get surprised?

std/bigint.d
```diff
@@ -1761,3 +1761,111 @@ unittest
...
+{
+    divMod(dividend, BigInt(divisor), quotient, remainder);
+}
+
+/// ditto
+void divMod(T)(T dividend, const BigInt divisor, out BigInt quotient, out BigInt remainder) pure nothrow
+    if (is(typeof(BigInt(dividend))))
+{
+    divMod(BigInt(dividend), divisor, quotient, remainder);
+}
```
andralex: There's not a lot of value in these last two convenience functions - after all, client code may simply construct the BigInt themselves. Worse, their existence suggest the caller some good optimizations are at work when in fact none is. Let's drop the overloads and only keep the first.

std/internal/math/biguintcore.d
Biotronic: I don't think anyone expects a function that works on `BigUint`s to handle signs.

std/bigint.d
Biotronic: Beautiful. Thanks. :)

JackStouffer:
Ping @Biotronic

Biotronic:
Pong.

Biotronic:
Added missing changelog entry, and addressed all comments in reviews. Pong?

PR #5938, opened 2017-12-18
Remove docs build from the Windows Makefiles

wilzbach:
Building the documentation under Windows has been abandoned (see e.g. https://github.com/dlang/dlang.org/pull/1976).

Hence the Windows doc targets here aren't run by any CI. This will leave a bad impression when used and being broken. Also these abandoned Makefile targets just add up to the huge chunk of redundancy that needs to be maintained between `posix.mak` and `win{32,64}.mak` (try to grep for e.g. `checkedint` in the diff).
Also the documentation listing are 2/3 of the Windows Makefiles,  this should make them quite a bit leaner.


ntrel:
@wilzbach So now there's no way to build even a single module's docs from Windows? Can't we keep it but print a message about it being out of date? I was literally using it today and was puzzled when it stopped working.

wilzbach:
The doc build via Phobos was legacy for a long time - the way to build the docs locally is via dlang.org: https://github.com/dlang/dlang.org/blob/master/CONTRIBUTING.md

See also: https://github.com/dlang/dlang.org/pull/1976?

> Support for Windows has been abandoned for more than a year. It's dangerous to leave unmaintained scripts in the official repository as it can easily lead to a frustrating experience.

Apart the enormous PITA to maintain the DigitalMars Makefiles (and the bad experience of people when things don't work), the other reasons for removing it is that DAutoTest does an excellent job at building the docs for the contributor on the PR and if the Ddoc in a Phobos module is really so complicated that it needs to be compiled locally, the doc string should be fixed.
Also run.dlang.io supports `-D`, so you can use it for quick previews, e.g. https://run.dlang.io/is/uteSju

What's your motivation for building the docs locally?
Note that if you really want to, you can still do the following - even from the Phobos dir:

```d
 dmd -conf= -m64 -w -c -o- -version=StdDdoc -I../druntime/import project.ddoc ../dlang.org/macros.ddoc ../dlang.org/html.ddoc ../dlang.org/dlang.org.ddoc ../dlang.org/std.ddoc -Df../dlang.org/std/test.html std/path.d
```

Also note that the DigitalMars files will hopefully be replaced soon (the decision as already been made) and GNUmake will be required to build DMD etc. on Windows (it is already required to build DMD's testsuite).

ntrel:
OK, thanks for explaining, I didn't think of using `posix.mak`. I build a module's docs locally so I can review the changes in the browser, helping me to see the whole page instead of just the ddoc lines on screen. This often makes me think of extra information to add or formatting changes. Anyway, I can use that command if necessary.

> GNUmake will be required to build DMD etc. on Windows

Sounds good. I did find it pretty confusing when I first tried to run the dmd test suite with DM make.

PR #5939, opened 2017-12-18
win64.mak: Try to unify source files

wilzbach:
According to Bugzilla [3372](https://issues.dlang.org/show_bug.cgi?id=3372) has been fixed six years ago.

In lack of a machine to test it and as it needs to work on auto-tester anyhow. Here's an attempt to remove the split into different unittest object files. If this works, the next stage would be to put the source files in a separate file which can be imported by `posix.mak` and `win{32,64}.mak` like e.g. it's done at DRuntime.

### DIFF
--- a/win64.mak
+++ b/win64.mak
@@ -94,46 +94,26 @@ test.obj : test.d
 test.exe : test.obj $(LIB)
 	$(DMD) -conf= test.obj -m$(MODEL) -g -L/map
 
-#	ti_bit.obj ti_Abit.obj
-
-SRC= \
+SRC_STD= \
 	unittest.d \
 	index.d
-
-# The separation is a workaround for bug 4904 (optlink bug 3372).
-SRC_STD_1= \
 	std\stdio.d \
 	std\string.d \
 	std\format.d \
 	std\file.d
-
-SRC_STD_2a= \
-	std\array.d \
 	std\functional.d \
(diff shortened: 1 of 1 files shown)


ZombineDev:
We're now hitting the 32k limit ;)

IIRC, this year we had to split the object files on Windows 1-2 times, you can look at commit history for details.

OptLink needs support at least 128k symbols, IMO in order for us to be certain this won't break later down the road.

wilzbach:
> OptLink needs support at least 128k symbols, IMO in order for us to be certain this won't break later down the road.

Hmm is this as simple as increasing the constants in the DMD source code or are the 32k a hard limit defined by OPTLINK?
Is there any chance that we can switch to mscoff?

ZombineDev:
We need changes to optlink and we can't switch away from using it in the short-term.

rainers:
The problem here is not a limit on the number of symbols (win64 uses COFF, not OMF), but a limit on available memory. I've not yet seen phobos unittest compile to a single object file. Last time I tried it run out of memory/swap space at around 15GB.
BTW: index.d is not a D source file to compile together with the other files, it's a ddoc file. Also, this PR is missing a lot of line continuation `\`s.

wilzbach:
Doesn't look like this is ever going to pass :/

PR #5940, opened 2017-12-18
Use a common Makefile for win{32,64}.mak

wilzbach:
Gets rid off of a huge chunk of redundancy by simply moving common variables to a shared file.

### DIFF
--- a/mak/import_win.mak
+++ b/mak/import_win.mak
@@ -1,0 +1,394 @@
+# Common parts shared between win32.mak and win64.mak
+
+# Copy command
+
+CP=cp
+
+## Directory where dmd has been installed
+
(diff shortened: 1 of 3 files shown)


wilzbach:
I'm a bit worried about auto-tester's patching here:
https://github.com/braddr/at-client/blob/master/src/diff-phobos-win64.diff

andralex:
This is cool, cuts a lot of fat.

wilzbach:
Well, it's blocked on the auto-tester still doing it's magic patching which would have been addressed by https://github.com/braddr/at-client/pull/8 ...

wilzbach:
Doesn't look like this is ever going to pass :/

PR #5941, opened 2017-12-18
Add DragonFlyBSD support for phobos

dkgroot:
Related: 
- [DMD Pull Request](https://github.com/dlang/dmd/pull/7463)
- [DRuntime Pull Request](https://github.com/dlang/druntime/pull/1999)

Bootstrapped via 'dmd-cxx' branch: [See](https://github.com/dkgroot/dragonflybsd_dmd_port)

Notes:
- Exclude assertion for dragonfly in std/experimental/allocator/mallocator.d (Needs further investigation, not sure why aligned_realloc is causing trouble).

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -13,7 +13,8 @@
 #
 # make BUILD=debug unittest => builds all unittests (for debug) and runs them
 #
-# make DEBUGGER=ddd std/XXXXX.debug => builds the module XXXXX and executes it in the debugger ddd
+# make DEBUGGER=ddd std/XXXXX.debug => builds the module XXXXX and executes it
+#                                      in the debugger ddd
 #
 # make html => makes html documentation
 #
@@ -25,10 +26,10 @@
 ################################################################################
 # Configurable stuff, usually from the command line
 #
-# OS can be linux, win32, win32wine, osx, or freebsd. The system will be
-# determined by using uname
+# OS can be linux, win32, win32wine, osx, freebsd, netbsd or dragonflybsd.
+# The system will be determined by using uname
 
-QUIET:=
+QUIET:=@
 
 DEBUGGER=gdb
 GIT_HOME=https://github.com/dlang
--- a/std/experimental/allocator/mallocator.d
+++ b/std/experimental/allocator/mallocator.d
@@ -278,7 +278,7 @@ struct AlignedMallocator
 
     /**
     Forwards to $(D alignedReallocate(b, newSize, platformAlignment)).
-    Should be used with bocks obtained with `allocate` otherwise the custom
+    Should be used with blocks obtained with `allocate` otherwise the custom
     alignment passed with `alignedAllocate` can be lost.
     */
     @system @nogc nothrow
@@ -375,6 +375,9 @@ version(Posix)
     AlignedMallocator.instance.alignedReallocate(c, 32, 32);
     assert(c.ptr);
 
+    version (DragonFlyBSD) {} else    /* FIXME: Malloc on DragonFly does not return NULL when allocating more than UINTPTR_MAX
+                                       * $(LINK: https://bugs.dragonflybsd.org/issues/3114, dragonfly bug report)
+                                       * $(LINK: https://github.com/dlang/druntime/pull/1999#discussion_r157536030, PR Discussion) */
     assert(!AlignedMallocator.instance.alignedReallocate(c, size_t.max, 4096));
     AlignedMallocator.instance.deallocate(c);
 }
--- a/std/math.d
+++ b/std/math.d
@@ -7444,6 +7444,34 @@ private real polyImpl(real x, in real[] A) @trusted pure nothrow @nogc
                 ;
             }
         }
+        else version (DragonFlyBSD)
+        {
+            asm pure nothrow @nogc // assembler by W. Bright
+            {
+                // EDX = (A.length - 1) * real.sizeof
+                mov     ECX,A[EBP]              ; // ECX = A.length
+                dec     ECX                     ;
+                lea     EDX,[ECX*8]             ;
+                lea     EDX,[EDX][ECX*4]        ;
+                add     EDX,A+4[EBP]            ;
+                fld     real ptr [EDX]          ; // ST0 = coeff[ECX]
+                jecxz   return_ST               ;
+                fld     x[EBP]                  ; // ST0 = x
+                fxch    ST(1)                   ; // ST1 = x, ST0 = r
+                align   4                       ;
+        L2:     fmul    ST,ST(1)                ; // r *= x
+                fld     real ptr -12[EDX]       ;
+                sub     EDX,12                  ; // deg--
+                faddp   ST(1),ST                ;
+                dec     ECX                     ;
+                jne     L2                      ;
+                fxch    ST(1)                   ; // ST1 = r, ST0 = x
+                fstp    ST(0)                   ; // dump x
+                align   4                       ;
+        return_ST:                              ;
+                ;
+            }
+        }
         else
         {
             static assert(0);
--- a/std/parallelism.d
+++ b/std/parallelism.d
@@ -94,6 +94,10 @@ else version(FreeBSD)
 {
     version = useSysctlbyname;
 }
+else version(DragonFlyBSD)
+{
+    version = useSysctlbyname;
+}
 else version(NetBSD)
 {
     version = useSysctlbyname;
@@ -981,6 +985,10 @@ uint totalCPUsImpl() @nogc nothrow @trusted
             auto nameStr = "machdep.cpu.core_count\0".ptr;
         }
         else version(FreeBSD)
+        {
+            auto nameStr = "hw.ncpu\0".ptr;
+        }
+        else version(DragonFlyBSD)
         {
             auto nameStr = "hw.ncpu\0".ptr;
         }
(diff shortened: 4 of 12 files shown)


posix.mak
```diff
@@ -25,9 +25,10 @@
 ################################################################################
 # Configurable stuff, usually from the command line
 #
-# OS can be linux, win32, win32wine, osx, or freebsd. The system will be
+# OS can be linux, win32, win32wine, osx, freebsd, netbsd or dragonflybsd. The system will be
 # determined by using uname
 
+#QUIET:=@
```
wilzbach: Any reason for this to be added?

posix.mak
```diff
@@ -25,9 +25,10 @@
 ################################################################################
 # Configurable stuff, usually from the command line
 #
-# OS can be linux, win32, win32wine, osx, or freebsd. The system will be
+# OS can be linux, win32, win32wine, osx, freebsd, netbsd or dragonflybsd. The system will be
```
wilzbach: Nit: break it at 80 chars here (i.e. after dragonflybsd).

std/conv.d
```diff
@@ -5519,7 +5519,7 @@ pure nothrow @safe @nogc unittest
 }
 
 // Test attribute propagation for UDTs
-pure nothrow @safe /* @nogc */ unittest
+pure nothrow @safe /+@nogc+/ unittest
```
wilzbach: Unrelated change - remove?

std/math.d
```diff
@@ -7444,6 +7444,34 @@ private real polyImpl(real x, in real[] A) @trusted pure nothrow @nogc
                 ;
             }
         }
+        else version (DragonFlyBSD)
+        {
+            asm pure nothrow @nogc // assembler by W. Bright
```
wilzbach: For the record: I'm not a huge fan of copying blocks as this means a bug fix will very likely miss one of these bits.
Anyhow as far as I know is Walter a huge proponent of using `version(A)` and not of `static if( A || B)`

std/parallelism.d
```diff
@@ -984,6 +988,10 @@ uint totalCPUsImpl() @nogc nothrow @trusted
         {
             auto nameStr = "hw.ncpu\0".ptr;
         }
+        else version(DragonFlyBSD)
+        {
+            auto nameStr = "hw.ncpu\0".ptr;
```
wilzbach: Off topic: this looks like a very old style. IIRC are strings literals always zero-terminated?

std/math.d
dkgroot: I completely agree. I did see this 'argument' pop up several time in different PR's and Forum discussions. I even wanted to suggest a generic 'BSD' version flag, as they all have common origins, similar to the 'Posix' version. But i did not want to reopen this discussion.

std/conv.d
dkgroot: Made it easier to exclude ranges of unittests, while trying to track down a particular issue.
But reverting the change.

std/parallelism.d
dkgroot: Don't know, maybe someone else can answer this one.

posix.mak
dkgroot: It was needed while debugging, accidentally committed. Retracted now.

std/parallelism.d
nemanja-boric-sociomantic: Yes, that's true, the string literals are always zero-terminated. I was wondering if something is also doing `nameStr.length` in which case it makes sense putting this `\0` but it doesn't look like it.

dkgroot: @nemanja-boric-sociomantic 
So should i rewrite this to: ```auto nameStr = "hw.ncpu".ptr;```, or just leave it as is.

The surrounding lines are doing the same, though. and i do not want to correct them (Only DragonFlyBSD related lines should be touched in this PR).

nemanja-boric-sociomantic: I would leave them as is IMHO and do this separately.

std/math.d
ibuclaw: Or maybe we do away with the asm implementation and implement a better poly algorithm.

https://issues.dlang.org/show_bug.cgi?id=12084

BBasile: It's possible too to put the implementation in a token string  `enum polyAsm = q{ asm pure nothrow @nogc { ... } };` and to mix it

dkgroot: I think this discussion should be had outside this PR. Sounds more like a structural (ie ASM) discussion, that pertaining to this particular PR (correct).

std/experimental/allocator/mallocator.d
```diff
@@ -375,6 +375,7 @@ version(Posix)
     AlignedMallocator.instance.alignedReallocate(c, 32, 32);
     assert(c.ptr);
 
+    version (DragonFlyBSD) {} else                    /* FIXME: assertion below fails, have not been able to figure out why, yet */
     assert(!AlignedMallocator.instance.alignedReallocate(c, size_t.max, 4096));
```
ZombineDev: If this assert fails it follows that `alignedReallocate` successfully called `posix_memalign`, allocating more virtual memory than what could be reasonably available on the system. Can you try to debug this test and see what exactly breaks down?

dkgroot: Reported this error to dragonfly ([dragonfly bug report](https://bugs.dragonflybsd.org/issues/3114))
See [PR Discussion](https://github.com/dlang/druntime/pull/1999#discussion_r157536030)
Will update this entry to add both these links, so the issue can be tracked.

ZombineDev: Thanks!

std/regex/internal/parser.d
```diff
@@ -1033,6 +1033,7 @@ if (isForwardRange!R && is(ElementType!R : dchar))
 {//@@@BUG@@@ write is @system
     with(zis)
     {
+        import std.regex.internal.kickstart;
```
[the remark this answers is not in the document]
dkgroot: I did not know i added this.  I think this might be wrongfully merged in. Thanks for noticing.
I will remove it and re-run my tests.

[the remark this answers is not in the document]
dkgroot: I had some trouble in line 1095 when running without this added import. I will recheck.

dkgroot:
@ZombineDev Thanks for Reviewing BTW
@jmdavis Thanks !

std/regex/internal/parser.d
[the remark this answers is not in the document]
dkgroot: Looks like the issue was resolved (possible/maybe by the git rebase i did before committing the PR).
Will remove the import

dkgroot:
@ZombineDev Thanks for the review. I guess i will still have to wait a little, to allow other to review, before squashing, right ? How do you guys normally like to see this squash done, using 'rebase -i HEAD~5'?

ibuclaw:
@dkgroot. Yes, git rebase and then fix up all commits into one.  Amending the original commit message if necessary.

ZombineDev:
I think the PR is good to go, so we don't need to wait more. I prefer doing squashing with `git rebase -i HEAD~<number of commits on top of master>` (in your case `git rebase -i HEAD~8`).

dkgroot:
@ZombineDev I did the `git rebase -i HEAD~8` as you suggested, but now this PR includes other/unrelated commits (ie: #a11f63ff,#4a1065e1,#1ad40c60,#cae4a305), [See](https://github.com/dlang/phobos/pull/5941/commits). I hope this is not going to cause issues.

nemanja-boric-sociomantic:
Just do `git fetch upstream` followed by `git rebase -i upstream/master`, given that the `upstream` is the name of the blessed remote.

ZombineDev:
@dkgroot  Yes, as @nemanja-boric-sociomantic said, first rebase your branch on top of master and then squash it.

nemanja-boric-sociomantic:
I see you lost all of your commits :-). `git reflog` will save you in case you don't know it already.

andralex:
@dkgroot what I do when I have unrelated commits in my PR is I remove them during `git rebase -i`

dkgroot:
Damn i completely messed it up. Will need a little time to figure this out.

andralex:
@dkgroot you may want to send me email (see http://erdani.com/index.php/contact) and I'll invite you to https://dlang.slack.com. Then you can get community help in real time via text chat in the browser.

dkgroot:
Did a 'git reset --hard ca27cdfec' which was my last known good state.
Will have to go a little slower this time.

dkgroot:
I hope I managed to do better this time.

ibuclaw:
@andralex - He is already on slack. :-)

dkgroot:
Sorry for the 'git mess' I created. Accoding to @geod24 it is merge able as is, so leaving it alone for now.
Thanks for all your help guys

andralex:
Thanks for the work and many thanks to the reviewers! I'll do the honors unless someone beats me to it.

andralex:
Oh, was too hasty - many commits. Will undo this.

dkgroot:
Thanks a lot everyone for the help / review / git stuff and the final merge !

MartinNowak:
Thanks for your contribution.

> Oh, was too hasty - many commits. Will undo this.

Well a missing squash isn't the end of the world, it's only a single merge commit on master after all.

PR #5945, opened 2017-12-19
Make import of std.internal.unicode_tables lazy.

wilzbach:
There's no need to open and open and parse the gigantic
`std.internal.unicode_tables` if it's never needed.

My journey:

`std.net.curl` -> `etc.c.curl` -> `std.socket` -> `std.stdio` -> `std.uni` -> `std.internal.unicode_tables`.

```shell
> cd std && echo "import std.uni;" > foo.d
> time dmd -c -o- foo.d
dmd -c -o- foo.d  0.10s user 0.00s system 99% cpu 0.101 total

> time dmd -c -o- foo.d -I..
dmd -c -o- foo.d -I..  0.04s user 0.01s system 99% cpu 0.050 total
```

Follow-up to https://github.com/dlang/phobos/pull/5916

Alternatively a benchmark with `avgtime`:

```
> avgtime -r100 dmd -c -o- foo.d
Total time (ms): 9942.26
Repetitions    : 100
Sample mode    : 96 (15 occurrences)
Median time    : 96.908
Avg time       : 99.4226
Std dev.       : 6.38707
Minimum        : 92.079
Maximum        : 120.269
95% conf.int.  : [86.9042, 111.941]  e = 12.5184
99% conf.int.  : [82.9706, 115.875]  e = 16.452
EstimatedAvg95%: [98.1707, 100.674]  e = 1.25184
EstimatedAvg99%: [97.7774, 101.068]  e = 1.6452

> avgtime -r100 dmd -c -o- foo.d -I..
Total time (ms): 4678.42
Repetitions    : 100
Sample mode    : 46 (40 occurrences)
Median time    : 46.537
Avg time       : 46.7842
Std dev.       : 1.39462
Minimum        : 44.338
Maximum        : 53.491
95% conf.int.  : [44.0508, 49.5176]  e = 2.73341
99% conf.int.  : [43.1919, 50.3765]  e = 3.59231
EstimatedAvg95%: [46.5108, 47.0575]  e = 0.273341
EstimatedAvg99%: [46.425, 47.1434]  e = 0.359231
```

### DIFF
--- a/std/uni.d
+++ b/std/uni.d
@@ -1588,7 +1588,7 @@ private auto packedArrayView(T)(inout(size_t)* ptr, size_t items) @trusted pure 
 // Partially unrolled binary search using Shar's method
 //============================================================================
 
-string genUnrolledSwitchSearch(size_t size) @safe pure nothrow
+auto genUnrolledSwitchSearch(size_t size) @safe pure nothrow
 {
     import core.bitop : bsr;
     import std.array : replace;
@@ -2760,7 +2760,7 @@ public:
         }
         ---
     */
-    string toSourceCode(string funcName="")
+    string toSourceCode()(string funcName="")
     {
         import std.array : array;
         auto range = byInterval.array();
@@ -6135,7 +6135,7 @@ package @trusted auto memoizeExpr(string expr)()
 }
 
 //property for \w character class
-package @property @safe CodepointSet wordCharacter()
+package @property @safe auto wordCharacter()
 {
     return memoizeExpr!("unicode.Alphabetic | unicode.Mn | unicode.Mc
         | unicode.Me | unicode.Nd | unicode.Pc")();
@@ -6212,7 +6212,7 @@ package dchar parseUniHex(Range)(ref Range str, size_t maxDigit)
       .canFind("invalid codepoint"));
 }
 
-auto caseEnclose(CodepointSet set)
+auto caseEnclose()(CodepointSet set)
 {
     auto cased = set & unicode.LC;
     foreach (dchar ch; cased.byCodepoint)
@@ -6226,7 +6226,7 @@ auto caseEnclose(CodepointSet set)
 /+
     fetch codepoint set corresponding to a name (InBlock or binary property)
 +/
-@trusted CodepointSet getUnicodeSet(in char[] name, bool negated,  bool casefold)
+@trusted CodepointSet getUnicodeSet()(in char[] name, bool negated,  bool casefold)
 {
     CodepointSet s = unicode(name);
     //FIXME: caseEnclose for new uni as Set | CaseEnclose(SET && LC)
--- a/std/uni.d
+++ b/std/uni.d
@@ -8900,11 +8900,25 @@ private dchar toTitlecase(dchar c)
     return c;
 }
 
-enum MAX_SIMPLE_UPPER = 1051;
-alias UpperTriple = AliasSeq!(toUpperIndex, MAX_SIMPLE_UPPER, toUpperTab);
-
-enum MAX_SIMPLE_LOWER = 1043;
-alias LowerTriple = AliasSeq!(toLowerIndex, MAX_SIMPLE_LOWER, toLowerTab);
+/*
+Needs to be a function to make the import of std.internal.unicode_tables lazy
+As of 2.078.0 DMD
+- doesn't allow templated structs with zero arguments nor
+- doesn't allow tuples (AliasSeq) as return value
+- doesn't perform imports in structs lazily
+*/
+private auto Triples()
+{
+    static struct TriplesImpl
+    {
+        import std.internal.unicode_tables : MAX_SIMPLE_UPPER, MAX_SIMPLE_LOWER;
+        //enum MAX_SIMPLE_UPPER = 1041;
+        alias UpperTriple = AliasSeq!(toUpperIndex, MAX_SIMPLE_UPPER, toUpperTab);
+        //enum MAX_SIMPLE_LOWER = 1043;
+        alias LowerTriple = AliasSeq!(toLowerIndex, MAX_SIMPLE_LOWER, toLowerTab);
+    }
+    return TriplesImpl();
+}
 
 // generic toUpper/toLower on whole string, creates new or returns as is
 private S toCase(alias indexFn, uint maxIdx, alias tableFn, alias asciiConvert, S)(S s) @trusted pure
@@ -9080,7 +9094,7 @@ if (isInputRange!Range && isSomeChar!(ElementEncodingType!Range) &&
     else
     {
         static import std.ascii;
-        return toCaser!(LowerTriple, std.ascii.toLower)(str);
+        return toCaser!(Triples.LowerTriple, std.ascii.toLower)(str);
     }
 }
 
@@ -9099,7 +9113,7 @@ if (isInputRange!Range && isSomeChar!(ElementEncodingType!Range) &&
     else
     {
         static import std.ascii;
-        return toCaser!(UpperTriple, std.ascii.toUpper)(str);
+        return toCaser!(Triples.UpperTriple, std.ascii.toUpper)(str);
     }
 }
 
@@ -9292,11 +9306,11 @@ if (isInputRange!Range && isSomeChar!(ElementEncodingType!Range) &&
         import std.utf : byDchar;
 
         // Decode first
-        return toCapitalizer!UpperTriple(str.byDchar);
+        return toCapitalizer!(Triples.UpperTriple)(str.byDchar);
     }
     else
     {
-        return toCapitalizer!UpperTriple(str);
+        return toCapitalizer!(Triples.UpperTriple)(str);
     }
 }
 
@@ -9557,7 +9571,7 @@ private template toCaseLength(alias indexFn, uint maxIdx, alias tableFn)
 
 @safe unittest
 {
-    alias toLowerLength = toCaseLength!(LowerTriple);
+    alias toLowerLength = toCaseLength!(Triples.LowerTriple);
     assert(toLowerLength("abcd") == 4);
     assert(toLowerLength("аБВгд456") == 10+3);
 }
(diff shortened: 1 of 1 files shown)


std/uni.d
```diff
@@ -8888,8 +8900,11 @@ private dchar toTitlecase(dchar c)
     return c;
 }
 
-private alias UpperTriple = AliasSeq!(toUpperIndex, MAX_SIMPLE_UPPER, toUpperTab);
-private alias LowerTriple = AliasSeq!(toLowerIndex, MAX_SIMPLE_LOWER, toLowerTab);
+enum MAX_SIMPLE_UPPER = 1051;
+alias UpperTriple = AliasSeq!(toUpperIndex, MAX_SIMPLE_UPPER, toUpperTab);
+
+enum MAX_SIMPLE_LOWER = 1043;
+alias LowerTriple = AliasSeq!(toLowerIndex, MAX_SIMPLE_LOWER, toLowerTab);
```
wilzbach: This is the only bit I couldn't figure out how to do lazily. Better ideas?

wilzbach:
So this will make literally every D program 0.05s faster to parse as `std.uni` is omnipresent in almost all Phobos modules, e.g.:

```
> cd std && echo "import std.stdio;" > foo.d
> time dmd -c -o- foo.d
dmd -c -o- foo.d  0.03s user 0.01s system 99% cpu 0.038 total

> time dmd -c -o- foo.d -I..
dmd -c -o- foo.d -I..  0.02s user 0.01s system 99% cpu 0.031 total
```

std/uni.d
quickfur: Where are these aliases used?  Since they were originally private, you may be able to just rewrite the places where they're used to lazily instantiate the definition. Or maybe something like:
````
private alias UpperTriple() = ...
````
might do the trick? Not 100% this syntax is supported, or whether it will do what I think it will do.

**Comment by wilzbach** [comment]
I managed to make a couple of other imports lazy and now `std.uni` is quite slim:

Before:

```d
parse     foo
importall foo
import    object	(/usr/include/dlang/dmd/object.d)
import    std.uni	(/usr/include/dlang/dmd/std/uni.d)
import    std.meta	(/usr/include/dlang/dmd/std/meta.d)
import    std.traits	(/usr/include/dlang/dmd/std/traits.d)
import    std.functional	(/usr/include/dlang/dmd/std/functional.d)
import    std.range.primitives	(/usr/include/dlang/dmd/std/range/primitives.d)
import    std.internal.unicode_tables	(/usr/include/dlang/dmd/std/internal/unicode_tables.d)
semantic  foo
import    std.range	(/usr/include/dlang/dmd/std/range/package.d)
import    std.typecons	(/usr/include/dlang/dmd/std/typecons.d)
import    core.stdc.stdint	(/usr/include/dlang/dmd/core/stdc/stdint.d)
import    core.stdc.stddef	(/usr/include/dlang/dmd/core/stdc/stddef.d)
import    core.stdc.signal	(/usr/include/dlang/dmd/core/stdc/signal.d)
import    core.stdc.wchar_	(/usr/include/dlang/dmd/core/stdc/wchar_.d)
import    core.stdc.config	(/usr/include/dlang/dmd/core/stdc/config.d)
import    core.stdc.stdarg	(/usr/include/dlang/dmd/core/stdc/stdarg.d)
import    core.stdc.stdlib	(/usr/include/dlang/dmd/core/stdc/stdlib.d)
import    core.stdc.stdio	(/usr/include/dlang/dmd/core/stdc/stdio.d)
import    core.stdc.time	(/usr/include/dlang/dmd/core/stdc/time.d)
import    core.sys.posix.sys.types	(/usr/include/dlang/dmd/core/sys/posix/sys/types.d)
import    core.sys.posix.config	(/usr/include/dlang/dmd/core/sys/posix/config.d)
import    std.array	(/usr/include/dlang/dmd/std/array.d)
import    std.algorithm.iteration	(/usr/include/dlang/dmd/std/algorithm/iteration.d)
import    std.range.interfaces	(/usr/include/dlang/dmd/std/range/interfaces.d)
import    std.format	(/usr/include/dlang/dmd/std/format.d)
import    core.vararg	(/usr/include/dlang/dmd/core/vararg.d)
import    std.exception	(/usr/include/dlang/dmd/std/exception.d)
semantic2 foo
semantic3 foo
code      foo
```

--------

After

```d
parse     foo
importall foo
import    object	(/home/seb/dlang/druntime/import/object.d)
import    std.uni	(/home/seb/dlang/phobos/std/uni.d)
import    std.meta	(/home/seb/dlang/phobos/std/meta.d)
import    std.traits	(/home/seb/dlang/phobos/std/traits.d)
import    std.functional	(/home/seb/dlang/phobos/std/functional.d)
import    std.range.primitives	(/home/seb/dlang/phobos/std/range/primitives.d)
semantic  foo
semantic2 foo
semantic3 foo
code      foo
```

std/uni.d
```diff
@@ -1588,7 +1588,7 @@ private auto packedArrayView(T)(inout(size_t)* ptr, size_t items) @trusted pure
 // Partially unrolled binary search using Shar's method
 //============================================================================
 
-string genUnrolledSwitchSearch(size_t size) @safe pure nothrow
+auto genUnrolledSwitchSearch(size_t size) @safe pure nothrow
```
quickfur: What's the reason for this change?

std/uni.d
```diff
@@ -10320,30 +10350,98 @@ private:
...
 }
 
 @safe pure nothrow @nogc @property
 {
-    import std.internal.unicode_tables; // generated file
-
     // It's important to use auto return here, so that the compiler
     // only runs semantic on the return type if the function gets
     // used. Also these are functions rather than templates to not
     // increase the object size of the caller.
```
quickfur: I'm not sure I understand the logic of the comment here. Why would templates increase the object size of the caller if they never get used? And does returning `auto` really suppress the import inside the function body until the function is called?

std/uni.d
```diff
@@ -6079,12 +6079,16 @@ else
...
 
-template SetSearcher(alias table, string kind)
+template SetSearcher(string tableName, string kind)
 {
     /// Run-time checked search.
     static auto opCall(C)(in C[] name)
         if (is(C : dchar))
     {
+        // lazily import the big unicode_table
+        mixin(q{
```
DmitryOlshansky: OMG, this going beyond acceptable mutilation imho.

std/uni.d
```diff
@@ -6095,6 +6099,10 @@ template SetSearcher(alias table, string kind)
     /// Compile-time checked search.
     static @property auto opDispatch(string name)()
     {
+        // lazily import the big unicode_table
+        mixin(q{
+            import std.internal.unicode_tables;
```
DmitryOlshansky: Same here

quickfur:
Maybe another approach that will speed up imports without massive mutilation of `std.uni` might be to generate `.di` file from `unicode_tables.d` and import that instead. That way, no actual code change will be required, and we avoid template bloat.

quickfur:
ping @wilzbach @DmitryOlshansky 

Any ideas how to move this forward?

CyberShadow:
Can we use `.di` files here to just provide an interface to the tables, so that the actual tables are parsed only when building Phobos and never when building user code?

wilzbach:
> Can we use .di files here to just provide an interface to the tables, so that the actual tables are parsed only when building Phobos and never when building user code?

Yeah that should work in theory, but then we would probably have to change the layout of Phobos a bit (e.g. like Druntime with it's `import` folder). However, I have an idea and will give it a try.

wilzbach:
> Can we use .di files here to just provide an interface to the tables, so that the actual tables are parsed only when building Phobos and never when building user code?

Tried it here: https://github.com/dlang/phobos/pull/6123.
tl;dr: DMD's header generation has many bugs, I bumped into two. They can be fixed though (submitted PRs). The big problem though is that the `unicode_tables` file uses templates too, so they won't be removed by the header file generation.

std/uni.d
```diff
@@ -8771,20 +8777,27 @@ else
...
-    dchar toTitleTab(size_t idx) { return toTitleTable[idx]; }
+    dchar toTitleTab()(size_t idx) {
+        import std.internal.unicode_tables : toTitleTable; // generated file
+        return toTitleTable[idx];
+    }
 
     ushort toUpperIndex(dchar c) { return toUpperIndexTrie[c]; }
     ushort toUpperSimpleIndex(dchar c) { return toUpperSimpleIndexTrie[c]; }
-    dchar toUpperTab(size_t idx) { return toUpperTable[idx]; }
+    dchar toUpperTab()(size_t idx) {
```
MartinNowak: Why not a function?

Actually making things a function above was a decision to also save on codegen time when sth. gets used, because the binary is already in phobos. Imports in function bodies do not leak to the call site, so you also save those when merely calling.
The compiler can still inline things if found appropriate at which point you pay the same price as for instantiating the template.

std/uni.d
```diff
@@ -6079,12 +6079,16 @@ else
...
+template SetSearcher(string tableName, string kind)
 {
     /// Run-time checked search.
     static auto opCall(C)(in C[] name)
         if (is(C : dchar))
     {
+        // lazily import the big unicode_table
+        mixin(q{
+            import std.internal.unicode_tables;
+            alias table = } ~ tableName ~ ";");
```
MartinNowak: How about
```d
mixin("std.internal.unicode_tables : table="~tableName~";")
```
instead?

std/uni.d
```diff
@@ -6218,7 +6226,7 @@ auto caseEnclose(CodepointSet set)
 /+
     fetch codepoint set corresponding to a name (InBlock or binary property)
 +/
-@trusted CodepointSet getUnicodeSet(in char[] name, bool negated,  bool casefold)
```
MartinNowak: Why make it a template?

MartinNowak:
> Yeah that should work in theory, but then we would probably have to change the layout of Phobos a bit (e.g. like Druntime with it's import folder). However, I have an idea and will give it a try.

Dmd prefers importing .di over .d files, so you could have them side by side. But our experience with .di files in druntime was fairly bad, try to avoid them.

std/uni.d
MartinNowak: > And does returning auto really suppress the import inside the function body until the function is called?

Yes it does, the compiler needs to have the full function body and run semantic3 on it to determine the return type. Auto return functions behave a lot like templates in that sense, they don't get compiled into the call-site though (AFAIR).

LightBender:

PR #5946, opened 2017-12-19
Fix Issue 18105 - std.conv.parse!wchar cannot take a string

Dechcaudron:
I also edited the associated test to make it slightly stricter.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -3342,24 +3342,17 @@ Throws:
     A $(LREF ConvException) if the range is empty.
  */
 Target parse(Target, Source)(ref Source s)
-if (isSomeString!Source && !is(Source == enum) &&
-    staticIndexOf!(Unqual!Target, dchar, Unqual!(ElementEncodingType!Source)) >= 0)
+if (isSomeString!Source && 
+    isSomeChar!Target &&
+    !is(Source == enum) &&
+    !is(Target == enum) &&
+    Target.sizeof >= (ElementEncodingType!Source).sizeof)
 {
     if (s.empty)
         throw convError!(Source, Target)(s);
-    static if (is(Unqual!Target == dchar))
-    {
-        Target result = s.front;
-        s.popFront();
-        return result;
-    }
-    else
-    {
-        // Special case: okay so parse a Char off a Char[]
-        Target result = s[0];
-        s = s[1 .. $];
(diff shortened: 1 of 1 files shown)


Dechcaudron:
You are right, my bad. I did not realize what the special case was doing. @schveiguy said in the bugzilla thread it would be an error to add this particular feature. I want to properly understand what he's on to before pushing through with this. Will update as soon as I hear from him.

schveiguy:
Yeah, this stems from a fundamental misunderstanding of how UTF encoding works.

The biggest issue with this proposed feature I see is that `parse` has no way of advancing the range past the first surrogate pair in a `wchar` sequence. In other words, if your UTF-8 sequence contains a surrogate pair, you must return a single `wchar` out of `parse`. you can't advance the range, since you need the entire code point to know what the other part of the pair is, and there's no way to store somewhere "next time, get me the second part". So `parse` cannot proceed.

The only logical step at that point would be to throw an exception. But that isn't great either, because someone might expect `parse` to simply keep going given that it works most of the time.

I think you should just use `std.utf.byUTF` to convert to `wchars`.

Dechcaudron:
Okay, I think I get it. If you tried to parse the "á" `string` as a `wchar`, it cannot break the utf-8 representation of the `á` Unicode code unit in a way that it returns the 2 parts of the equivalent utf-16 surrogate pair in different calls. Is that what you mean? If so, I now realize and apologize for the noise. This particular issue cannot be fixed.

However, I set out to fix this issue because to!wchar(string) does not compile due to it. Does @schveiguy agree with me in that it should be supported by Phobos? I can try to implement it in a different way.

schveiguy:
I get what you mean, yes, since `to!char(string)` works for strings with a single code unit (i.e. single `char`), and simply throws when given a string with more than one code-unit, it stands to reason that `to!wchar(string)` should just throw when it can't convert a string with a single code point into a single wchar.

My recommendation is to first parse out a `dchar`. If the string isn't empty, error.

Then, decode the `dchar` into a `wchar[2]` using `std.utf`. If the length decoded isn't 1, error.

Otherwise return the first `wchar`.

Using this mechanism, you should be able to support all conversions between the UTF types.

FYI, a code-point with a surrogate pair to test with is 𐐷

PR #4047, opened 2016-03-03
gamma function: add constants for 128-bit reals.

redstar:

PR #5947, opened 2017-12-19
Porting: gammafunction: Add constants for ieeeQuadruple reals

jpf91:
Setting MAXGAMMA to the same as for ieeeExtended reals means the range of the gamma function will be limited to the same range as ieeeExtended. However, until someone reviews the algorithm to work fine for ieeeQuadruple ranges, this workaround seems fine. We just lose some 128bit real features, but we're still as good as x86.

This is a revival of https://github.com/dlang/phobos/pull/4047


jpf91:
Thanks!

PR #5584, opened 2017-07-09
Use more selective imports at module imports

wilzbach:
AFAICT selective imports have been fixed in [2.071.0](http://dlang.org/changelog/2.071.0.html#imports-314) and now enough time has passed down the road.

FWIW I still hope that we get [Dependency-Carrying Declarations (DIP1005)](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1005.md) at some point and can completely get rid of these global imports and make Phobos entirely pay for what you need.

PR #5948, opened 2017-12-19
Dramatically faster to import std.datetime (and anything under it)

andralex:
This has been an iterative process of pushing imports from the top level into the places where imports are being used. That helped not only clarifying dependencies, but also accelerates import time by about 3x. Further improvements are possible.

### DIFF
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -7,9 +7,13 @@
 +/
 module std.datetime.systime;
 
-import core.time;
-import std.datetime.date;
-import std.datetime.timezone;
+// Note: reconsider using specific imports below after
+// https://issues.dlang.org/show_bug.cgi?id=17630 has been fixed
+import core.time;// : ClockType, convert, dur, Duration, seconds, TickDuration, TimeException;
+import std.datetime.date;// : _monthNames, AllowDayOverflow, CmpTimeUnits, Date,
+    //DateTime, DateTimeException, DayOfWeek, enforceValid, getDayOfWeek, maxDay,
+    //Month, splitUnitsFromHNSecs, TimeOfDay, validTimeUnits, yearIsLeapYear;
+import std.datetime.timezone;// : LocalTime, SimpleTimeZone, TimeZone, UTC;
 import std.exception : enforce;
 import std.format : format;
 import std.range.primitives;
@@ -70,6 +74,7 @@ public:
     {
         import std.format : format;
         import std.stdio : writefln;
+        import core.time;
         assert(currTime().timezone is LocalTime());
         assert(currTime(UTC()).timezone is UTC());
 
@@ -426,6 +431,7 @@ public:
 
     @safe unittest
     {
+        import core.time;
         static void test(DateTime dt, Duration fracSecs, immutable TimeZone tz, long expected)
         {
             auto sysTime = SysTime(dt, fracSecs, tz);
@@ -1729,6 +1735,7 @@ public:
     @safe unittest
     {
         import std.range : chain;
+        import core.time;
 
         assert(SysTime(0, UTC()).fracSecs == Duration.zero);
         assert(SysTime(1, UTC()).fracSecs == hnsecs(1));
@@ -1822,6 +1829,7 @@ public:
     @safe unittest
     {
         import std.range : chain;
+        import core.time;
 
         foreach (fracSec; testFracSecs)
         {
@@ -1856,6 +1864,7 @@ public:
 
     @safe unittest
     {
+        import core.time;
         assert(SysTime(0).stdTime == 0);
         assert(SysTime(1).stdTime == 1);
         assert(SysTime(-1).stdTime == -1);
@@ -1883,6 +1892,7 @@ public:
 
     @safe unittest
     {
+        import core.time;
         static void test(long stdTime, in SysTime expected, size_t line = __LINE__)
         {
             auto st = SysTime(0, UTC());
@@ -1967,6 +1977,7 @@ public:
 
     @safe unittest
     {
+        import core.time;
         {
             auto sysTime = SysTime(DateTime(1982, 1, 4, 8, 59, 7), hnsecs(27));
             assert(sysTime == sysTime.toLocalTime());
@@ -1999,6 +2010,7 @@ public:
 
     @safe unittest
     {
+        import core.time;
         auto sysTime = SysTime(DateTime(1982, 1, 4, 8, 59, 7), hnsecs(27));
         assert(sysTime == sysTime.toUTC());
         assert(sysTime._stdTime == sysTime.toUTC()._stdTime);
@@ -2022,6 +2034,7 @@ public:
 
     @safe unittest
     {
+        import core.time;
         auto stz = new immutable SimpleTimeZone(dur!"minutes"(11 * 60));
         auto sysTime = SysTime(DateTime(1982, 1, 4, 8, 59, 7), hnsecs(27));
         assert(sysTime == sysTime.toOtherTZ(stz));
@@ -2089,6 +2102,7 @@ public:
     @safe unittest
     {
         import std.meta : AliasSeq;
+        import core.time;
         assert(SysTime(DateTime(1970, 1, 1), UTC()).toUnixTime() == 0);
         foreach (units; AliasSeq!("hnsecs", "usecs", "msecs"))
             assert(SysTime(DateTime(1970, 1, 1, 0, 0, 0), dur!units(1), UTC()).toUnixTime() == 0);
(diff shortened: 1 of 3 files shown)


quickfur:
Very nice!  Hooray for local imports. They are an overall win.

You could also reference https://issues.dlang.org/show_bug.cgi?id=13253 in one of your commits; it's a useful way to keep track of improvements that we've made in turning more imports local.

wilzbach:
BTW during my work on this, I noticed a couple of things:

- We shouldn't be doing this by hand. At least the conversion from full imports to selective imports could be automated
- If we would have lazy selective imports, then all this work isn't needed and from a naive point of view, this shouldn't be too hard? The compiler knows where he can find a symbol and only needs to so if he really needs that symbol resolved...
- There's currently no way to make imports in aggregated types lazy (except for using Voldemort types, but that would be a breaking change for public symbols). Example:

```d
struct Foo
{
    import std.internal.unicode_tables; // not lazy, but ok
    import std.internal.unicode_tables : SimpleCaseEntry; // not lazy either :/
}
```

- Technically converting global imports into local or selective ones, is a breaking change due to the broken import system in the compiler see e.g. https://github.com/dlang/phobos/pull/5584

wilzbach:
> continuous-integration/jenkins/pr-merge — This commit cannot be built

Due to a vibe.d test that hangs constantly (not related to this PR) -> https://github.com/dlang/ci/pull/96

jmdavis:
I don't see anything obviously wrong, though it's not surprising if the Windows build broke in the process given that std.datetime.timezone contains Windows-specific code. That's either going to have to be dealt with on a Windows system or risk a fair bit of back and forth with the auto-tester.

I confess thought that this PR helps show why I hate local and selective imports. They help with performance at least some of the time (apparently quite a bit here), so they can be necessary (especially for Phobos, which is imported by everything), and being more explicit about what uses which symbols might help some of the time, but man, the code just gets littered with import statements, and whenever you make changes, you frequently have to muck with the various selective imports instead of just writing code. It would be _so_ much nicer if we could just have the imports at the top without bothering with selective or local imports, but at least right now, the compiler isn't able to do that as efficiently.

andralex:
>I confess thought that this PR helps show why I hate local and selective imports. They help with performance at least some of the time (apparently quite a bit here), so they can be necessary (especially for Phobos, which is imported by everything), and being more explicit about what uses which symbols might help some of the time, but man, the code just gets littered with import statements, and whenever you make changes, you frequently have to muck with the various selective imports instead of just writing code. It would be _so_ much nicer if we could just have the imports at the top without bothering with selective or local imports, but at least right now, the compiler isn't able to do that as efficiently.

I'm actually in two minds about this. I am old enough to have seen all of these arguments made earnestly (roughly in chronological order):

* "These local variables are so tedious. You mean I need to define `i` with a type again and again whenever I need to loop over something? I wish I wouldn't need to bother with locals, globals are already there when you need them."
* "You mean now some random statement can introduce a new variable? Wish I wouldn't need to bother with that, I prefer it when all variables in a function are nicely declared at the top of the function."
* "With virtuals there's no way to tell what code gets executed. Such code is impossible to test!"
* "You mean every function call can throw an exception so no more single exit? That's the end of civilization!"

... you get my drift. Local imports are a new technology. I know of no language that does them similarly to the way D does them. It is expected that new technology comes with new advantages, new liabilities, and  generally changes in style that are not universally liked.

This PR has been cathartic. Yeah, imports at the top are there for the whole module, but... they're there for the whole module. And I mean all 11,000 lines of it. Initially I thought I'd never figure it out - cutting these mutually recursive imports is surprisingly difficult because compile-time errors don't always give the right information, e.g. when CTFE and constraints are involved. (I will look into filing some bugs.) With this PR it's way more clear what depends on what, where the large dependency knots are, etc.

It may be the case that global imports will become seen as okay for small projects, an antipattern for large programs.

jmdavis:
Yeah, there are pros and cons, and in larger projects, they matter more. I even use them quite a bit even though I hate them. It's just that they get so tedious and repetitive, and when templates aren't involved, there really isn't going to be a performance boost from local imports, since the imports will happen anyway. It's just that it then helps show the dependencies in a more fine-grained manner, which can help avoid importing stuff when you don't need to after refactoring. But if there are a lot of duplicate imports, I'm quickly inclined to put the imports at the top.

In some ways though, the selective imports are far worse than the local imports, because they require more maintenance - and it's particularly bad if you go so far as to use selective imports for the range primitives given how many you end up using for even simple code that uses arrays as ranges. std.range.primitives is one that I pretty much never use selective imports for. In the end, these seem like a prime target for an automated tool, which isn't exactly a good sign. In theory, we're trying to avoid needing extra tools to make code manageable and avoid boilerplate.

If the compiler were smarter and did more imports in a lazy fashion, we could at least mitigate the need to do this for performance, but that still leaves all of the arguments about how much it improves things in terms of how it clarifies dependencies and whether that's worth all of the code duplication.

In any case, as with function attributes, this is one area where good, idiomatic D code seems to be really verbose with the stuff that isn't actual code. And I just can't look at a module with tons of import statements with many of them followed by whole lists of symbols being selectively imported and not get annoyed at how verbose and repetitive it is - especially when the verbosity isn't actually required by the language.

quickfur:
I've been slowly shifting away from global imports in my own code and moving towards local imports.  Just like the old C++ innovation of declaring variables close to where you actually need them vs. at the top of the function like older forms of C, I've been finding that local imports are much more well-encapsulated, esp. if put close to or even next to the line where it's actually needed, and blocks of code written this way can be moved around easily without having to constantly fiddle with updating import statements at the top of the scope / module.

Like manually typing long type names in (older forms of?) C++, though, there is definitely a noticeable maintenance cost to it.  Which makes me wonder if some import analogue of type inference might be the next step forward.  Like some kind of `lazy import` or `auto import`.  But presently I can't think of any concrete realization of this idea that doesn't either just send us straight back to square one (i.e., declaring modules at the top of the file / start of scope, etc.), or has significant implementational difficulties.

But maybe somebody can take this idea and run with it, and get somewhere new. :-)

andralex:
>when templates aren't involved, there really isn't going to be a performance boost from local imports, since the imports will happen anyway.

Nonono. Careful here, I also was confused by that. When you import module `xyz`, only top-level imports and imports inside aggregates (struct/class/union) are opened. All others are NOT opened. That includes imports inside non-template functions.

MartinNowak:
> ... you get my drift. Local imports are a new technology. I know of no language that does them similarly to the way D does them. It is expected that new technology comes with new advantages, new liabilities, and generally changes in style that are not universally liked.

Yes, that's how things go. Also new things are often introduced with a more verbose syntax and get simplified later on (e.g `delegate (arg) { return expr; }` => `arg => expr`).

In that sentiment, now that things get more adopted, maybe it'd be worth to think about some shorter syntax for local imports. https://github.com/dlang/dmd/pull/6589

Other candidates:
`() @trusted { stmt; }();` => `@trusted { stmt; }`
`immutable` => `imm` or `let`

Not that urgent and a great bike-shedding topic though.

std/datetime/systime.d
```diff
@@ -8560,6 +8630,7 @@ public:
     // bug# 17801
     @safe unittest
     {
+        import core.time;
```
MartinNowak: How about a top-level `version (unittest) import core.time;` instead? Maybe even just a selective one?
It's indeed annoying that privately imported modules are resolved eagerly, albeit none of their symbols might ever be needed.

andralex: I considered it but am weary of the bugs that could happen. Consider:

```d
version(unittest) import std.stdio;
void fun(T)(T value)
{
    writeln(value);
}
unittest
{
    fun(5);
}
```

This module compiles on its own, passes unittests, is importable when not used, but issues an error whenever the user attempts to use `fun`. I think we should make it a guideline that `version(unittest)` top-level imports should be used with caution.

wilzbach: FYI: At least if it's a publicly documented test, this can't happen as we execute them separately exactly for the reason to protect against these kinds of bugs.

jmdavis: > I think we should make it a guideline that version(unittest) top-level imports should be used with caution.

It's more or less been a guideline for a while now, precisely because it caused problems in the past. I don't know that we've actually written it down anywhere though, just told people not to do it when it comes up in PRs.

andralex:
@MartinNowak good points. I think we should in the short term move forward with dlang/dmd#6589. No need for a DIP as long as we're removing a limitation.

andralex:
ping?

std/datetime/timezone.d
```diff
@@ -7,11 +7,9 @@
...
 module std.datetime.timezone;
 
-import core.time;
-import std.datetime.date;
-import std.datetime.systime;
-import std.exception : enforce;
-import std.range.primitives;
+import core.time : abs, convert, dur, Duration, hours, minutes;
+import std.datetime.systime : Clock, stdTimeToUnixTime, SysTime;
+import std.range.primitives : back, front, empty, popFront;
```
joakim-noah: As Seb pointed out, this could break code because of [the selective import symbol leak laid out in Issue 17630](https://issues.dlang.org/show_bug.cgi?id=17630) and demonstrated in #5584, which is unfixed according to bugzilla.  We cannot push all D code to selective imports until all known leaks are fixed without breaking code.

andralex: I've backed off this part of the PR and left a note for the future.

andralex:
squashed, will put up for auto-merge

andralex:
fixed a couple more imports

PR #5950, opened 2017-12-20
Revert "Add DragonFlyBSD support for phobos"

andralex:
Bit more work on consolidating the commits needed


wilzbach:
The git history is supposed to be immutable, so except you want to force-push over it, we have to live with this ;-)

andralex:
coll thanks guys

PR #5952, opened 2017-12-21
Make std.typecons.Tuple betterC compatible

wilzbach:
I'm not sure how people want to start testing `-betterC` compatibility with Phobos. I assume the best way would be to do extract all unittests marked with `@betterC`?

```d
unittest @betterC
{
....
}
```

For now I went with a more simple and straight-forward way of simply executing all files in `test/betterC`.

~~~However, I still get an error when using the `alias BOMSeq = Tuple!(int, "schema");`  line:~~~

Also I need to define `main` manually as the default main with `void main` returns with `44` :/

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -512,7 +512,7 @@ if (distinctFieldNames!(Specs))
     //      :
     // NOTE: field[k] is an expression (which yields a symbol of a
     //       variable) and can't be aliased directly.
-    string injectNamedFields()
+    enum injectNamedFields = ()
     {
         string decl = "";
         static foreach (i, val; fieldSpecs)
@@ -525,7 +525,7 @@ if (distinctFieldNames!(Specs))
             }
         }}
         return decl;
-    }
+    };
 
     // Returns Specs for a subtuple this[from .. to] preserving field
     // names if any.
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -1498,6 +1498,26 @@ if (distinctFieldNames!(Specs))
     assert(t == AliasSeq!(1, Bad(1), "asdf"));
 }
 
+// Ensure Tuple.toHash works
+@safe unittest
+{
+    Tuple!(int, int) point;
+    assert(point.toHash == typeof(point).init.toHash);
+    assert(tuple(1, 2) != point);
+    assert(tuple(1, 2) == tuple(1, 2));
+    point[0] = 1;
+    assert(tuple(1, 2) != point);
+    point[1] = 2;
+    assert(tuple(1, 2) == point);
+}
+
+@safe @betterC unittest
+{
+    auto t = tuple(1, 2);
+    assert(t == tuple(1, 2));
+    auto t3 = tuple(1, 'd');
+}
+
 /**
     Creates a copy of a $(LREF Tuple) with its fields in _reverse order.
 


posix.mak
```diff
@@ -411,6 +411,18 @@ unittest/%.run : $(ROOT)/unittest/test_runner
...
 	 BUILD=debug $(MAKE) -f $(MAKEFILE) $(basename $<).debug_with_debugger
 
+################################################################################
+# Run separate -betterC tests
+################################################################################
+
+test/betterC/%.run: test/betterC/%.d $(DMD) $(LIB)
+	mkdir -p $(ROOT)/unittest/betterC
+	$(DMD) $(DFLAGS) -of$(ROOT)/unittest/betterC/$(notdir $(basename $<)) -betterC $(UDFLAGS) $(LIB) \
+		-defaultlib= -debuglib= $(LINKDL) $<
```
ZombineDev: `-betterC` implies that we're not linking druntime and phobos -> drop `$(LIB)`. This should hopefully get rid of the current linker error, though we get a new one.

wilzbach: Hehe I thought so too and I tried this in the beginning, but then I get more linker errors:

```
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiVAyaa6_736368656d61Z17injectNamedFieldsFZQBo: error: undefined reference to '_D12TypeInfo_Aya6__initZ'
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiVAyaa6_736368656d61Z17injectNamedFieldsFZQBo: error: undefined reference to '_d_arraycatnTX'
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiVAyaa6_736368656d61Z17injectNamedFieldsFZQBo: error: undefined reference to '_D12TypeInfo_Aya6__initZ'
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiVAyaa6_736368656d61Z17injectNamedFieldsFZQBo: error: undefined reference to '_d_arrayappendT'
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiVAyaa6_736368656d61Z17injectNamedFieldsFZQBo: error: undefined reference to '_D12TypeInfo_Aya6__initZ'
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiVAyaa6_736368656d61Z17injectNamedFieldsFZQBo: error: undefined reference to '_d_arraycatnTX'
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiVAyaa6_736368656d61Z17injectNamedFieldsFZQBo: error: undefined reference to '_D12TypeInfo_Aya6__initZ'
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiVAyaa6_736368656d61Z17injectNamedFieldsFZQBo: error: undefined reference to '_d_arrayappendT'
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiZ17injectNamedFieldsFZAya: error: undefined reference to '_d_arraycatnTX'
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiZ17injectNamedFieldsFZAya: error: undefined reference to '_d_arrayappendT'
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiVAyaa6_736368656d61ZQBc6toHashMxFNbNeZm: error: undefined reference to '_D10TypeInfo_i6__initZ'
generated/linux/release/64/unittest/betterC/test18101.o:test/betterC/test18101.d:function _D3std8typecons__T5TupleTiZQj6toHashMxFNbNeZm: error: undefined reference to '_D10TypeInfo_i6__initZ'
collect2: error: ld returned 1 exit status
```

ZombineDev: There are two culprits here:
1) `injectNamedFields` gets codegened -> we need to make sure that it exist only in CTFE context. One way to do this is to implement it as `enum injectNamedFields = { /* code */ }();
2. `toHash` uses `typeid(T).getHash` -> try replacing this with some template function which will get codegen-ed on demand (so that we don't need to linking druntime). You can try with `core.internal.hash : hashOf`, or if that doesn't work, we can workaround it temporary with:
```D
version (D_BetterC)
    static assert (0, "Not implemented");
else:
    /* same as before ... */
```

and making `Tuple.toHash` a parameter-less template.

wilzbach: Worked like a charm. Thanks!

std/typecons.d
```diff
@@ -1135,12 +1135,20 @@ if (distinctFieldNames!(Specs))
             Returns:
                 A `size_t` representing the hash of this `Tuple`.
          */
-        size_t toHash() const nothrow @trusted
+        size_t toHash()() const nothrow @trusted
```
wilzbach: Looks like why can't change this to a template:

```
std/functional.d(1003): Error: AA key type Tuple should have 'size_t toHash() const nothrow @safe' if opEquals defined
generated/linux/debug/64/publictests/std_functional.d(193): Error: template instance std_functional.__unittest_generated_linux_debug_64_publictests_std_functional_d_187_15.memoize!(fib) error instantiating
std/functional.d(1003): Error: AA key type Tuple should have 'size_t toHash() const nothrow @safe' if opEquals defined
generated/linux/debug/64/publictests/std_functional.d(205): Error: template instance std_functional.__unittest_generated_linux_debug_64_publictests_std_functional_d_199_16.memoize!(fact) error instantiating
std/functional.d(1003): Error: AA key type Tuple should have 'size_t toHash() const nothrow @safe' if opEquals defined
generated/linux/debug/64/publictests/std_functional.d(219): Error: template instance std_functional.__unittest_generated_linux_debug_64_publictests_std_functional_d_211_17.memoize!(factImpl) error instantiating
```

How about simply removing the method for `betterC` for now?

ZombineDev: How about this:
```D
size_t toHash() const nothrow @trusted
{
    version(D_BetterC)
    {
        // Disabled in -betterC for now as TypeInfo isn't present
        assert(0, "Not implemented");
    }
    else
    {
        size_t h = 0;
        foreach (i, T; Types)
            h += typeid(T).getHash(cast(const void*)&field[i]);
        return h;
    }
}
```

DmitryOlshansky: That would fail at runtime which is strictly worse then compile-time failure.

ZombineDev: Of course, that's why my first suggestion was to make `toHash` a template function and use `static assert`. Thinking about it more I don't see why we define our own `opEquals` and `toHash`, instead of relying on the compiler generated ones.

wilzbach:
Finally got around rebasing this. It looks like we can get around the problems of the `static assert` errors if we use the user-facing `hashOf` implementation.

wilzbach:
Opened a separate PR that just adds the testsuite (https://github.com/dlang/phobos/pull/6640), s.t. we can already add the betterC testsuite.

wilzbach:
Well, we can't use `Tuple!(int, Json)` yet, but thanks to a few recent changes in druntime this seems to be finally passing.
I added a very simplistic `-betterC` test in the `test/betterc` testsuite, but once we finally get `@betterC` I plan on annotating a few unittests, s.t. we can ensure `Tuple` will continue to work in `-betterC`.

test/betterC/typecons.d
```diff
@@ -0,0 +1,7 @@
+extern(C) void main()
```
ZombineDev: Can you now move it inside `std.typecons` as a `@BetterC unittest`?

wilzbach: I can/will do this once https://github.com/dlang/phobos/pull/6645 has been merged.

std/typecons.d
```diff
@@ -1169,12 +1169,12 @@ if (distinctFieldNames!(Specs))
...
                 A `size_t` representing the hash of this `Tuple`.
          */
-        size_t toHash() const nothrow @safe
+        size_t toHash() const nothrow @trusted
         {
             size_t h = 0;
             static foreach (i, T; Types)
             {{
-                const k = typeid(T).getHash((() @trusted => cast(const void*) &field[i])());
+                const k = hashOf(field[i]);
```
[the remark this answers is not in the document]
n8sh: @wilzbach `hashOf` void arrays is currently working in druntime `master`.

ZombineDev:
@n8sh the enum function trick used in this PR may help for `hashOf`'s use of CTFE. (Though of course the compiler should also be fixed.)

std/typecons.d
```diff
@@ -1168,12 +1168,12 @@ if (distinctFieldNames!(Specs))
             Returns:
                 A `size_t` representing the hash of this `Tuple`.
          */
-        size_t toHash() const nothrow @safe
+        size_t toHash() const nothrow @trusted
```
n8sh: This can't be legitimately marked `@trusted` because `hashOf(x)` can call `x.toHash()` which may do any number of unsafe things. It also can't be `@trusted` because of the "Object.toHash() is not const" fiasco.

n8sh:
@ZombineDev I don't think it can be used there: you can't treat function arguments as compile-time constants even in an `if (__ctfe)` block. Or did you mean something else?

wilzbach:
Rebased and removed the `@trusted`. Looks like it's no longer required (: 

CC @thewilsonator

PR #5956, opened 2017-12-22
Update CodeCov configuration to be always green/successful

wilzbach:
Propagating the config from https://github.com/dlang/dmd/pull/7470.


quickfur:
LGTM

PR #5957, opened 2017-12-22
Reduced copying in BigInt.opCmp

JackStouffer:


quickfur:
Hmm. I was under the impression that `BigInt` was just a lightweight wrapper around a reference to the actual big int data that sits on the heap, with COW semantics. So passing the struct by ref or not shouldn't matter?  Or does it implement by-value semantics, as in copies the actual bigint data every time you assign a BigInt?

JackStouffer:
BigInt holds a struct which holds the reference to the heap, so you're not copying the data when passing a bigint by value but you are (as I understand it) copying the reference and the sign data when you don't need to. So while there won't be a huge difference, there is unnecessary passing by value which can be reduced.

quickfur:
How big is the struct itself, though? If it fits in a couple of registers or stack words, it might not be worth the extra indirection since it would only be slightly bigger than the pointer underneath the ref.

JackStouffer:
> How big is the struct itself, though?

A `BigInt` is 24 bytes.

quickfur:
As far as I can tell from the source code, it's basically just an array slice (ptr + size) and a bool indicating sign.  I suppose the compiler pads it to 24 bytes for alignment or whatever reasons, but it seems hardly necessary to use `auto ref` just for passing a slice + flag.

Or am I missing some other fields that are also in there somewhere?

Either way, I don't feel strongly enough about this to say yes or no to this PR, but it's just an observation that we may not be gaining much (if anything at all) just by taking an extra indirection to a BigInt struct, which is already itself pretty small. We do have to pay for an extra level of indirection, which may be more costly in terms of CPU cache misses, than using a few extra bytes to pass it by value. But it's hard to tell without some actual benchmarks to show the difference.

JackStouffer:
Ok, I knew the differences are too small to measure, so just quickly threw this together thinking that avoiding copying the data would automatically mean less work but I didn't think of the cache misses.

ZombineDev:
A better optimization would be to drop the bool field and to store its information in of the free bits inside the slice (and add accessor properties to hide the bit twiddling).

quickfur:
Unfortunately, that would mean essentially merging `BigInt` with `BigUint`, which is a pretty big job because you'd have to update every function in `BigUint` to work with the sign bit.

Although... actually, perhaps one possibility would be to "waste" the first `BigDigit` in the data array to store just the sign, then slice the data with `data[1 .. $]` before handing it over to the `BigUint` functions.  Won't be a trivial change, but easier and (much) less error-prone than touching the `BigUint` functions themselves.

PR #5958, opened 2017-12-23
prevent deprecation message caused by std.math.abs() on short and byte.

BBasile:
I target stable since you don't allow deprecation in the standard library anymore

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -527,14 +527,20 @@ enum real SQRT1_2 =    SQRT2/2;                               /** $(SQRT)$(HALF)
  * = hypot(z.re, z.im).
  */
 Num abs(Num)(Num x) @safe pure nothrow
-if (is(typeof(Num.init >= 0)) && is(typeof(-Num.init)) &&
+if ((is(typeof(Num.init >= 0)) && is(typeof(-Num.init)) ||
+    (is(Num == short) || is(Num == byte))) &&
     !(is(Num* : const(ifloat*)) || is(Num* : const(idouble*))
     || is(Num* : const(ireal*))))
 {
     static if (isFloatingPoint!(Num))
         return fabs(x);
     else
-        return x >= 0 ? x : -x;
+    {
+        static if (is(Num == short) || is(Num == byte))
+            return x >= 0 ? x : cast(Num) -int(x);
+        else
+            return x >= 0 ? x : -x;
+    }
 }
 
 /// ditto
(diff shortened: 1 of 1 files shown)


BBasile:
The tests pass on my linux x86_64 box with both DMD 2.078-beta.1 and DMD ~master so i don't get what's happening with the tests. The change should have been rather trivial.

std/math.d
```diff
@@ -527,14 +527,21 @@ enum real SQRT1_2 =    SQRT2/2;                               /** $(SQRT)$(HALF)
...
     || is(Num* : const(ireal*))))
 {
     static if (isFloatingPoint!(Num))
         return fabs(x);
     else
-        return x >= 0 ? x : -x;
+    {
+        // NOTE: The operation is not safe for BigInt on 32 bits systems
+        static if (is(Num == short) || is(Num == byte))
+            return x >= 0 ? x : cast(Num) (x * -1);
```
BBasile: note about performances: the optimizer [turns the multiplication into an asm NEG](https://godbolt.org/g/m8MHzE) so there are in practice no change at all.

BBasile:
ping this is trivial, but it's for 2.078 so don't forget it (6 days if i trust the planning).

ibuclaw:
Why `* - 1`? Isn't casting explicitly enough?

First you came for our blessed tilde, now you want to make our negation obfuscated too. 😛

BBasile:
Okay, let's go for a basic type constructor then.

MetaLang:
I hate how complex this signature has become. Couldn't we add overloads for `abs(byte)` and `abs(short)` instead?

BBasile:
@MetaLang, the constraint of the current function will not be more simple if we put the two cases away.
It will be even less clear because an explanation will be required. For now the strange processing is hidden inside.

BBasile:
@ibuclaw, now that it's merged, you said

> First you came for our blessed tilde

There must me a misunderstanding, I've never been in touch with anyone about the tidle. My own code has been affected but it was about the minus prefix, and I've only talked about this on IRC, where after no help i did [this](https://github.com/BBasile/iz/commit/14ddcc727e81957f76982a1f456c9d233d48c8a4#diff-4cb8438a3d0a8d2c18f7f483131b1978R721).

So...i don't know what you are talking about with this tidle story.

ibuclaw:
https://github.com/dlang/phobos/commit/1882758ab1e47e943e9b96f4b16a92f833ddcaa4#diff-8b9f1870415268a4c40b628d596dc405

BBasile:
okay...but you know...i'm not in the organization.

BBasile:
anyway, end of story ~

BBasile:
Should have tested the Unqual type... https://issues.dlang.org/show_bug.cgi?id=18349

PR #5959, opened 2017-12-23
Fixes Issue 18116 - to!wchar([string, dstring]), and to!char([wstring, dstring]) don't compile

Dechcaudron:
Fix for issue [18116](https://issues.dlang.org/show_bug.cgi?id=18116). Mentioning @schveiguy for review, due to discussion [here](https://github.com/dlang/phobos/pull/5946)

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -1884,6 +1884,49 @@ if (isInputRange!S && !isInfinite!S && isSomeChar!(ElementEncodingType!S) &&
 }
 
 /**
+String, or string-like input range, to char type not directly
+supported by parse parses the first dchar of the source.
+
+Returns: the first code point of the input range, converted
+         to type T.
+
+Throws: ConvException if the input range contains more than
+        a single code point, or if the code point does not
+        fit into a code unit of type T.
+*/
+private T toImpl(T, S)(S value)
+if (isSomeChar!T && !is(typeof(parse!T(value))) &&
+    is(typeof(parse!dchar(value))))
+{
+    import std.utf : encode;
+    
+    immutable dchar codepoint = parse!dchar(value);
+    if (!value.empty)
+        throw new ConvException(convFormat("Cannot convert \"%s\" to %s because it " ~
+                                           "contains more than a single code point.",
+                                           value, T.stringof));
+    T[dchar.sizeof / T.sizeof] decodedCodepoint;
+    if (encode(decodedCodepoint, codepoint) != 1)
+        throw new ConvException(convFormat("First code point '%s' of \"%s\" does not fit into a " ~
+                                           "single %s code unit", codepoint, value, T.stringof));
+    return decodedCodepoint[0];
+}
+
+@safe pure unittest
+{
+    import std.exception : assertThrown;
+
+    assert(toImpl!wchar("a") == 'a');
+
+    assert(toImpl!char("a"d) == 'a');
+    assert(toImpl!char("a"w) == 'a');
+    assert(toImpl!wchar("a"d) == 'a');
+
+    assertThrown!ConvException(toImpl!wchar("ab"));
+    assertThrown!ConvException(toImpl!char("😃"d));
+}
+
+/**
 Convert a value that is implicitly convertible to the enum base type
 into an Enum value. If the value does not match any enum member values
 a ConvException is thrown.
--- a/std/conv.d
+++ b/std/conv.d
@@ -1899,7 +1899,7 @@ if (isSomeChar!T && !is(typeof(parse!T(value))) &&
     is(typeof(parse!dchar(value))))
 {
     import std.utf : encode;
-    
+
     immutable dchar codepoint = parse!dchar(value);
     if (!value.empty)
         throw new ConvException(convFormat("Cannot convert \"%s\" to %s because it " ~


std/conv.d
```diff
@@ -835,8 +844,8 @@ if (!isImplicitlyConvertible!(S, T) &&
     class C : B, I, J {}
     class D : I {}
 
-    static foreach (m1; 0 .. 5) // enumerate modifiers
-    static foreach (m2; 0 .. 5) // ditto
+    foreach (m1; AliasSeq!(0,1,2,3,4)) // enumerate modifiers
+    foreach (m2; AliasSeq!(0,1,2,3,4)) // ditto
```
wilzbach: Huh?

std/conv.d
```diff
@@ -1324,7 +1333,7 @@ in
 {
     assert(radix >= 2 && radix <= 36);
 }
-do
+body
```
wilzbach: Huh?

std/conv.d
```diff
@@ -2454,7 +2495,7 @@ in
...
```
wilzbach: Huh?

std/conv.d
```diff
@@ -1619,16 +1628,6 @@ if (!isImplicitlyConvertible!(S, T) && isAssociativeArray!S &&
...
 
-@safe unittest
-{
-    import std.algorithm.comparison : equal;
-    import std.array : byPair;
-
-    int[int] a;
-    assert(a.to!(int[int]) == a);
-    assert(a.to!(const(int)[int]).byPair.equal(a.byPair));
-}
```
wilzbach: Why did you delete this test?

std/conv.d
```diff
@@ -1883,6 +1882,48 @@ if (isInputRange!S && !isInfinite!S && isSomeChar!(ElementEncodingType!S) &&
...
+         to type T.
+
+Throws: ConvException if the input range contains more than
+        a single code point, or if the code point does not
+        fit into a code unit of type T.
+*/
+private T toImpl(T, S)(S value)
+if (isInputRange!S && isSomeChar!(ElementEncodingType!S) &&
+     isSomeChar!T && !is(typeof(parse!T(value))) &&
+     is(typeof(parse!double(value))))
```
wilzbach: This looks weird. There shouldn't be a need to check for `parse!double`

std/conv.d
```diff
@@ -1883,6 +1882,48 @@ if (isInputRange!S && !isInfinite!S && isSomeChar!(ElementEncodingType!S) &&
...
+{
+    immutable dchar codepoint = parse!dchar(value);
+    if (!value.empty)
+        throw new ConvException(convFormat("Cannot convert \"%s\" to %s because it " ~
+                                           "contains more than a single code point.",
+                                           value, T.stringof));
```
wilzbach: Nit: two empty lines -> remove one.

std/conv.d
schveiguy: It looks like potentially @Dechcaudron used a stale version of the repository to base his changes from?

wilzbach: Hmm git is smart enough for this (it thinks in hunks/changes, not files). It looks more like a rebase with mistakes or copy/pasting an old version of this file into a up-to-date branch.

schveiguy:
@Dechcaudron in addition to the changes already requested, you should change the log message in your commit to reference the issue fixed. i.e. say "Fixes issue 18116" somewhere in the log message. That triggers the automatic changelog editing.

std/conv.d
Dechcaudron: I use parse!double(value) within the method, and given the argument is of type S, it may not compile. I'll check and see if instantiation conditions for parse overlap the ones on S for this method,

std/conv.d
Dechcaudron: @wilzbach's assesment is accurate. I fucked up real bad with this one. I'll fix it and let you guys know.

std/conv.d
Dechcaudron: I don't see the empty lines. Do you mind pointing them out?

wilzbach: Ah the GitHub mobile view didn't display the "contains more" lines. Sorry. Please ignore

Dechcaudron:
Okay, everything up to code now (pun intended).

std/conv.d
```diff
@@ -1883,6 +1883,47 @@ if (isInputRange!S && !isInfinite!S && isSomeChar!(ElementEncodingType!S) &&
...
+private T toImpl(T, S)(S value)
+if (isSomeChar!T && !is(typeof(parse!T(value))) &&
+    is(typeof(parse!double(value))))
+{
+    immutable dchar codepoint = parse!dchar(value);
+    if (!value.empty)
+        throw new ConvException(convFormat("Cannot convert \"%s\" to %s because it " ~
+                                           "contains more than a single code point.",
+                                           value, T.stringof));
+    auto equivalentTArray = to!(T[])(codepoint);
```
schveiguy: Use `std.utf.encode`, as `to!(T[])` will allocate.

```D
T[dchar.sizeof / T.sizeof] outbuf; // hacky sizing, but it's correct.
auto resultSize = encode(outbuf, codepoint);
if(resultSize != 1) { ... } // throw conv error.
return outbuf[0];
```

std/conv.d
```diff
@@ -1883,6 +1883,47 @@ if (isInputRange!S && !isInfinite!S && isSomeChar!(ElementEncodingType!S) &&
...
+    if (equivalentTArray.length > 1)
+        throw new ConvException(convFormat("Code point '%s' does not fit into a single " ~
+                                           "%s code unit", codepoint, T.stringof));
```
schveiguy: Slightly adjust the message here, remember the input was a string, not a code point.

std/conv.d
```diff
@@ -1883,6 +1883,47 @@ if (isInputRange!S && !isInfinite!S && isSomeChar!(ElementEncodingType!S) &&
...
+    import std.exception : assertThrown;
+
+    assert(toImpl!wchar("a") == 'a');
+
+    assert(toImpl!char("a"d) == 'a');
+    assert(toImpl!char("a"w) == 'a');
+    assert(toImpl!wchar("a"d) == 'a');
+
+    assertThrown!ConvException(toImpl!wchar("ab"));
+    assertThrown!ConvException(toImpl!char("😃"d));
```
schveiguy: looks good!

std/conv.d
```diff
@@ -1883,6 +1883,47 @@ if (isInputRange!S && !isInfinite!S && isSomeChar!(ElementEncodingType!S) &&
...
+Returns: the first code point of the input range, converted
+         to type T.
+
+Throws: ConvException if the input range contains more than
+        a single code point, or if the code point does not
+        fit into a code unit of type T.
+*/
+private T toImpl(T, S)(S value)
+if (isSomeChar!T && !is(typeof(parse!T(value))) &&
+    is(typeof(parse!double(value))))
```
schveiguy: What is this `parse!double` for? did you mean `parse!dchar`?

Dechcaudron: Yep, my bad. Fixing.

std/conv.d
Dechcaudron: [As per the documentation](https://dlang.org/phobos/std_utf.html#encode) encode will throw UTFException if the input is not a valid UTF code point. Should I catch said exception and rethrow a ConvException?

std/conv.d
Dechcaudron: Updated, let's see if it's better now.

wilzbach:
This isn't a regression. Please target master. The 2.078 release is about to happen soon and we should only merge regression fixes to it. You still need to rebase your PR locally as I just changed the target branch. If you need help, please let me/us know.

std/conv.d
schveiguy: Let's follow precedent -- whatever `to!wstring(string)` does with an invalid code point. I suspect it lets the exception through.

Dechcaudron: Okay, no further changes then

schveiguy: Yep, just tested, throws a `UTFException`.

Dechcaudron:
My pleasure 😃. But why are the checks failing?

schveiguy:
> But why are the checks failing?

See my commit, trailing whitespace 😄

Dechcaudron:
Oh, so that's why. Thanks!

std/conv.d
```diff
@@ -1883,6 +1883,49 @@ if (isInputRange!S && !isInfinite!S && isSomeChar!(ElementEncodingType!S) &&
     assert(to!int(byDchar("10"), 10) == 10);
 }
 
+/**
+String, or string-like input range, to char type not directly
```
JackStouffer: Sorry to jump in after this is already been merged,  but this documentation will not be seen because this function is private. This should have gone as the docs to an example underneath to.

schveiguy: Hm... the docs don't talk about `to!char(string)` which worked before this change (and was the impetus to finding this limitation). I suppose it wouldn't hurt to rearrange the unit test to be part of the public docs and add some more comments.

Dechcaudron: The thing is, since all `to` does is call `toImpl`, all documentation of the latter does not show up in the generated docs. As a matter of fact, the unit tests I added do not call `to` directly, but rather `toImpl`, so there is not much sense in having them document `to`. But I do agree `to` could benefit from better documentation. Should that be a separate issue?

ZombineDev: Sounds good

schveiguy: @Dechcaudron would you mind refactoring your unit tests to call `to` instead of `toImpl` (I should have caught that), moving the unit test to be under the `to` function, making it documented and putting a little blurb about it? I think you could even add calls for `to!char(string)` to show that works also.

Dechcaudron: I think it makes sense to duplicate the tests rather than to plain move them. After all, we still want this particular override of `toImpl` tested. If I'm going to do something that affects the public documentation of `to`, I want to put some more thought into it. I'll get back to you guys about this in some days. I need to disconnect from Phobos a little. 

Should I create a separate pull request once I've gone about the above? Shall we create a new issue then?

Dechcaudron: Guys, ready to do this over here. New issue, or just a standalone PR?

JackStouffer: just make a new PR, no need for a bugzilla issue

PR #5960, opened 2017-12-24
Don't expect root home dir in unittest

ThomasMader:
Running Phobos unittests in the NixOS build environment while building DMD 2.067.1 or newer, doesn't work because the user root has home directory /build instead of /root.
For now I patch it in the Nix build script but it would be better to not depend on the existence of the /root directory.

See also https://github.com/NixOS/nixpkgs/issues/29443

Here is the error of the build machine with DMD 2.067.1:

****** FAIL release64 std.path
core.exception.AssertError@std/path.d(3083): /build

### DIFF
--- a/std/path.d
+++ b/std/path.d
@@ -4005,11 +4005,13 @@ string expandTilde(string inputPath) nothrow
 }
 
 
-version(unittest) import std.process : environment;
 @system unittest
 {
     version (Posix)
     {
+        import std.process : executeShell, environment;
+        import std.string : strip;
+
         // Retrieve the current home variable.
         auto oldHome = environment.get("HOME");
 
@@ -4032,20 +4034,13 @@ version(unittest) import std.process : environment;
         if (oldHome !is null) environment["HOME"] = oldHome;
         else environment.remove("HOME");
 
-        // Test user expansion for root, no /root on Android
-        version (OSX)
-        {
-            assert(expandTilde("~root") == "/var/root", expandTilde("~root"));
-            assert(expandTilde("~root/") == "/var/root/", expandTilde("~root/"));
-        }
-        else version (Android)
-        {
-        }
-        else
-        {
-            assert(expandTilde("~root") == "/root", expandTilde("~root"));
-            assert(expandTilde("~root/") == "/root/", expandTilde("~root/"));
-        }
+        immutable tildeUser = "~" ~ environment.get("USER");
+        immutable path = executeShell("echo " ~ tildeUser).output.strip();
+        immutable expTildeUser = expandTilde(tildeUser);
+        assert(expTildeUser == path, expTildeUser);
+        immutable expTildeUserSlash = expandTilde(tildeUser ~ "/");
+        assert(expTildeUserSlash == path ~ "/", expTildeUserSlash);
+
         assert(expandTilde("~Idontexist/hey") == "~Idontexist/hey");
     }
 }


std/path.d
```diff
@@ -4032,20 +4035,15 @@ version(unittest) import std.process : environment;
...
-        else version (Android)
-        {
-        }
-        else
-        {
-            assert(expandTilde("~root") == "/root", expandTilde("~root"));
-            assert(expandTilde("~root/") == "/root/", expandTilde("~root/"));
-        }
+        immutable user = environment.get("USER");
+        immutable tildeUser = "~" ~ user;
```
wilzbach: I just tested this on my Android phone and this doesn't seem to work there.
`echo ~$USER` evaluates to `~shell`

Maybe simply use `version(Android) {} else` in front of this block?

CC @joakim-noah

ThomasMader: Yes I thought about Android but didn't know how to test it locally in a fast way so I hoped the auto tester will do it.
Thanks for testing it manually.
Since it is not working I will add the Android block as it was before.

CyberShadow: > Yes I thought about Android but didn't know how to test it locally in a fast way so I hoped the auto tester will do it.

If you have an Android phone, you can reproduce this by enabling USB debugging, plugging in your phone, running `adb shell` on your PC, and typing `echo ~$USER`.

wilzbach: As DMD can't compile for arm, we don't have a CI testing for it. Most of the Android work is actually based on @joakim-noah's hard work at LDC. With LDC being so up to date, we should be able to incorporate their diff and run the Phobos testsuite directly with LDC too. Well that's a goal for 2018...

std/path.d
```diff
@@ -4005,11 +4005,14 @@ string expandTilde(string inputPath) nothrow
 }
 
 
-version(unittest) import std.process : environment;
+version(unittest)
```
wilzbach: You can remove this now ...

std/path.d
ThomasMader: Ok little misunderstanding. I thought you ran the actual test.
Even if the user on Android is shell and there is no home directory configured the test should not fail.
I already tested that case by creating a user without home directory. (useradd -r testuser)

std/path.d
ThomasMader: I was wondering about that construct. :-)
Didn't understand why the import was placed there.

joakim-noah:
This pull will remove the special case testing for root that was there before, pinging the author, what do you think, @kyllingstad?

CyberShadow:
>  Even if the user on Android is shell and there is no home directory configured the test should not fail.

No, that's not the problem. It looks like the default shell on Android can't do tilde expansion with a username.

joakim-noah:
> It looks like the default shell on Android can't do tilde expansion with a username.

That'd make sense since [Android assigns a different UID per app](https://android.stackexchange.com/questions/8452/how-can-i-find-app-name-by-uid), but I just tried running this patch in the Termux terminal app and the main issue here is that the Termux app doesn't set a `USER` env variable.  However, that means these tests should work in Termux, as it just expands the `~` alone and gets the Termux `HOME`.  I haven't tried it in an apk though, which is a different environment.

Don't worry about Android, just run your tests there too and I'll fix it later if needed.

I'm more wary of taking out tests that work fine most everywhere else: maybe just put your new tests under `version(linux)`, which will cover Android too, and let the old tests be on `OSX` and the BSDs?

ThomasMader:
> I'm more wary of taking out tests that work fine most everywhere else: maybe just put your new tests under version(linux), which will cover Android too, and let the old tests be on OSX and the BSDs?

Nix is available for OSX too and I think one can also setup OSX without root or without home dir.
Using the current user ensures that that exists at least.

std/path.d
```diff
@@ -4032,20 +4034,21 @@ version(unittest) import std.process : environment;
...
         {
         }
         else
         {
-            assert(expandTilde("~root") == "/root", expandTilde("~root"));
-            assert(expandTilde("~root/") == "/root/", expandTilde("~root/"));
+            immutable user = environment.get("USER");
+            immutable tildeUser = "~" ~ user;
+            immutable echoCmd = executeShell("echo " ~ tildeUser);
+            immutable pathFromEcho = echoCmd.output.strip();
```
wilzbach: Could have been one line

std/path.d
```diff
@@ -4032,20 +4034,21 @@ version(unittest) import std.process : environment;
...
+            immutable expTildeUser = expandTilde(tildeUser);
+            immutable expTildeUserSlash = expandTilde(tildeUser ~ "/");
```
wilzbach: We usually put the expression to validate directly before its respective `assert`, i.e. we would have moved this `asser` one line below.

std/path.d
ThomasMader: I changed it slightly so I saved two lines of code but still don't have too many columns.
Putting everything into one line is way too much for my taste. It's also easier to debug with multiple lines.

std/path.d
ThomasMader: Moved one assert up, I guess that's what you meant.

std/path.d
```diff
@@ -4032,20 +4034,19 @@ version(unittest) import std.process : environment;
...
 
-        // Test user expansion for root, no /root on Android
-        version (OSX)
-        {
-            assert(expandTilde("~root") == "/var/root", expandTilde("~root"));
-            assert(expandTilde("~root/") == "/var/root/", expandTilde("~root/"));
-        }
-        else version (Android)
+        version (Android)
         {
```
joakim-noah: As I said, take out the Android exception.

ThomasMader: Done

PR #5961, opened 2017-12-24
std.range.nullSink - allow convenient use of the NullSink

wilzbach:
`nullSink` returns the `NullSink` by reference and thus allows using it
directly in code which expects an OutputRange by reference.

The pattern is providing convenience wrappers for struct construction is
very common in Phobos, e.g. `rebindable`, `nullable`, `refCounted`, ...

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -117,7 +117,7 @@ $(BOOKTABLE ,
         loop. Similar to $(D zip), except that $(D lockstep) is designed
         especially for $(D foreach) loops.
     ))
-    $(TR $(TD $(LREF NullSink))
+    $(TR $(TD $(LREF nullSink))
         $(TD An output _range that discards the data it receives.
     ))
     $(TR $(TD $(LREF only))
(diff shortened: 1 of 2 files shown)


std/range/package.d
```diff
@@ -11547,14 +11547,45 @@ struct NullSink
     void put(E)(E){}
 }
 
+/// ditto
+auto ref nullSink() {
```
JackStouffer: brace style

JackStouffer:
Ping @andralex

JackStouffer:
Ping @andralex

quickfur:
ping @andralex 

This one is simple to review.  Should we merge?

quickfur:
P.S. The Phobos PR queue is back below 100 PRs!  Hooray, Team Phobos!  Let's keep pushing to merge / close the rest of the rotting PRs to get the queue size back under control!

andralex:
Not a fan guys. The null sink is a degenerate output range with limited usefulness. Dedicating a convenience function in addition to a type to it - isn't that a bit much?

Is NullSink public?

quickfur:
`NullSink` has been public for at least the past several releases, if not longer. It was added by @WalterBright (see https://github.com/dlang/phobos/pull/2030).

wilzbach:
> Is NullSink public?

Yes. I would be happy to remove it from the docs as part of this PR.

> Dedicating a convenience function in addition to a type to it

There are many precedents for this, e.g. `refCounted`, `nullable`, `rebindable` and most of the functions in `std.range` still have their public range struct exposed ...

BTW it's not only about this - `NullSink()` is basically useless as it's ironically not an output range - it ain't a lvalue. From the attached examples:

```
line.csvNextToken(NullSink(), ',', '"'); // Error: cannot pass type NullSink as a function argument
```

https://run.dlang.io/is/Hc61LZ

quickfur:
Well, it was added back in 2014, when it was still "cool" to write code along the lines of:
````
Nullsink ns;
myRange.blah.blah.blah.copy(ns);
````

andralex:
fine grumble

andralex:
oh at least can the docs be merged together? NullSink and nullSink...

quickfur:
`nullSink` already has a `/// ditto`, so ddoc should be able to merge it automatically.

andralex:
oh ok thx

PR #5963, opened 2017-12-25
Add explicit return types in std.regex

dhasenan:
https://issues.dlang.org/show_bug.cgi?id=18124 : explicit return types are good for documentation. Mildly annoying for editing when you are breaking the public API, but that shouldn't happen, right?

### DIFF
--- a/std/regex/package.d
+++ b/std/regex/package.d
@@ -356,7 +356,7 @@ public alias StaticRegex = Regex;
 
     Throws: $(D RegexException) if there were any errors during compilation.
 +/
-@trusted public auto regex(S)(S[] patterns, const(char)[] flags="")
+@trusted public Regex!(BasicElementOf!S) regex(S)(S[] patterns, const(char)[] flags="")
 if (isSomeString!(S))
 {
     import std.array : appender;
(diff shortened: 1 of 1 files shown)


JackStouffer:
>but that shouldn't happen, right?

Quite the contrary, **because** it's `auto` we can "break" the API as long as the interface remains the same, e.g. https://github.com/dlang/phobos/pull/5436/files

Thanks for the PR but I'm going to close this because this is churn which has little benefit and restricts us in the future. If you find the documentation lacking, please bug/PR it.

dhasenan:
The types are owned by std.regex; they aren't shared with other modules. They have no other purpose. So we can alter them however we wish without having to rename them.

If we needed to switch to a different return type that is shared by another module, it is easy to add an alias.

Finally, the argument that we can alter return types at any time if they're listed as `auto` applies to **literally every function in phobos**. What's so special about Capture and RegexMatch that we need to be able to swap them out?

Temtaime:
Even worse, such an explicit return type makes function signature looks ugly and any changes to it force us to modify it.
I see no benefit in such prs. Improve the documentation if you want.

wilzbach:
@dhasenan thanks a lot for your pull request and wanting to make Phobos a better library. However, I think you should put your energy in something new or real bugs instead.

> this is churn which has little benefit and restricts us in the future

+1 (and I also think that `auto` reads itself a lot better in the documentation than `RegexMatch!(Unqual!R)`).

> If we needed to switch to a different return type that is shared by another module, it is easy to add an alias.

No we wouldn't be able to:
- we can't change the meaning of public types
- just look at the `Unqual`'s you used

If people really need the return type, it's as simple as `typeof(regex("foo")`.

> What's so special about Capture and RegexMatch that we need to be able to swap them out?

Nothing, but development is happening and why should we restrict ourselves without need?

PR #5942, opened 2017-12-18
Faster to import std.regex, by lazy eval of wordMatcher

DmitryOlshansky:
After:
▶ time dmd test_case.d
dmd test_case.d  0.24s user 0.03s system 98% cpu 0.269 total

▶ gco master
Switched to branch 'master'

Before:
▶ time dmd test_case.d
dmd test_case.d  1.10s user 0.12s system 99% cpu 1.228 total

PR #5964, opened 2017-12-25
Deprecate public aliases for Regex, StaticRegex, etc

dhasenan:
As per the discussion at https://github.com/dlang/phobos/pull/5963, we want the option to swap out the types for Regex, Capture, etc. This is impossible if people are able to refer to Regex by name. Furthermore, according to @wilzbach, aliases are not an option. Therefore, this PR deprecates public uses of regex types (Regex, StaticRegex, Captures, and RegexMatch) to ensure we can later swap out the implementation types..

### DIFF
--- a/std/regex/package.d
+++ b/std/regex/package.d
@@ -332,6 +332,8 @@ import std.typecons; // : Flag, Yes, No;
     assert(nc[1..$].equal(["var"]));
     ----
 +/
+// See commentary at https://github.com/dlang/phobos/pull/5963
+deprecated(`use typeof(regex("foo"))`)
 public alias Regex(Char) = std.regex.internal.ir.Regex!(Char);
 
 /++
@@ -340,7 +342,12 @@ public alias Regex(Char) = std.regex.internal.ir.Regex!(Char);
 
     No longer used, kept as alias to Regex for backwards compatibility.
 +/
+// See commentary at https://github.com/dlang/phobos/pull/5963
+deprecated(`use typeof(ctRegex!("foo"))`)
 public alias StaticRegex = Regex;
+
+private alias _Regex(Char) = std.regex.internal.ir.Regex!(Char);
+private alias _StaticRegex = _Regex;
 
 /++
     Compile regular expression pattern for the later execution.
@@ -459,9 +466,13 @@ template ctRegexImpl(alias pattern, string flags=[])
 +/
 public enum ctRegex(alias pattern, alias flags=[]) = ctRegexImpl!(pattern, flags).wrapper;
 
-enum isRegexFor(RegEx, R) = is(Unqual!RegEx == Regex!(BasicElementOf!R)) || is(RegEx : const(Regex!(BasicElementOf!R)))
-     || is(Unqual!RegEx == StaticRegex!(BasicElementOf!R));
+enum isRegexFor(RegEx, R) = is(Unqual!RegEx == _Regex!(BasicElementOf!R)) || is(RegEx : const(_Regex!(BasicElementOf!R)))
+     || is(Unqual!RegEx == _StaticRegex!(BasicElementOf!R));
 
+/// Compatibility alias for _Captures
+// See commentary at https://github.com/dlang/phobos/pull/5963
+deprecated(`use typeof(matchOnce("", "").front)`)
+public alias Captures(R) = _Captures!R;
 
 /++
     $(D Captures) object contains submatches captured during a call
@@ -469,7 +480,7 @@ enum isRegexFor(RegEx, R) = is(Unqual!RegEx == Regex!(BasicElementOf!R)) || is(R
 
     First element of range is the whole match.
 +/
-@trusted public struct Captures(R)
+@trusted private struct _Captures(R)
 if (isSomeString!R)
 {//@trusted because of union inside
     alias DataIndex = size_t;
@@ -498,7 +509,7 @@ private:
         _f = 0;
     }
 
-    this(ref RegexMatch!R rmatch)
+    this(ref _RegexMatch!R rmatch)
     {
         _input = rmatch._input;
         _names = rmatch._engine.pattern.dict;
@@ -695,13 +706,18 @@ public:
     assert(!matchFirst("nothing", "something"));
 }
 
+/// Compatibility alias for _Captures
+// See commentary at https://github.com/dlang/phobos/pull/5963
+deprecated(`use typeof(match("", ""))`)
+public alias RegexMatch(R) = RegexMatch!R;
+
 /++
     A regex engine state, as returned by $(D match) family of functions.
 
     Effectively it's a forward range of Captures!R, produced
     by lazily searching for matches in a given input.
 +/
-@trusted public struct RegexMatch(R)
+@trusted private struct _RegexMatch(R)
 if (isSomeString!R)
 {
 private:
@@ -709,7 +725,7 @@ private:
     Matcher!Char _engine;
     const MatcherFactory!Char _factory;
     R _input;
-    Captures!R _captures;
+    _Captures!R _captures;
 
     this(RegEx)(R input, RegEx prog)
     {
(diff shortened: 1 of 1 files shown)


std/regex/package.d
```diff
@@ -340,8 +342,13 @@ public alias Regex(Char) = std.regex.internal.ir.Regex!(Char);
 
     No longer used, kept as alias to Regex for backwards compatibility.
 +/
+// See commentary at https://github.com/dlang/phobos/pull/5963
+deprecated(`use typeof(ctRegex!("foo"))`)
 public alias StaticRegex = Regex;
 
+private alias _Regex(Char) = std.regex.internal.ir.Regex!(Char);
```
DmitryOlshansky: Ugh!

DmitryOlshansky:
I think this quickly got out of hand. I think it's fine to keep Regex!Char and in fact I'd drop isRegexFor instead.

StaticRegex!X is indeed on its way out.

Captures!R is a non-standard concept (unlike e.g. forward range) that we are pretty much stuck with.

RegexMatch is indeed just a forward range of Captures so can deprecate that .

DmitryOlshansky:
The biggest problem is that we should move docs to package.d so that Regex alias contains DDoc from internal/ir.d

dhasenan:
> I think this quickly got out of hand. 

Thank you! I strive to be efficient.

> I think it's fine to keep Regex!Char

But then we can't change `regex()` to return a value of a different type without a deprecation cycle. Since the consensus seems to be that this is an important thing for us to be able to do, we should get the necessary deprecation cycle started ASAP.

> RegexMatch is indeed just a forward range of Captures so can deprecate that .

Not quite. It's a forward range of Captures with extra convenience methods `pre`, `post`, and `hit`. While we could add free functions to be used with UFCS to emulate this if we change to a more widely used type, this would break code that uses selective imports.

So we can't switch from RegexMatch to something that's currently more widely used. But we could come up with a type similar to RegexMatch that can be more widely used (eg a more advanced type for searching through random-access ranges). If we did that, I would be able to add an alias for compatibility, but that apparently isn't a universal trait of phobos devs and I can't commit to being available to help out. So that's not an option.

std/regex/package.d
wilzbach: Btw that's another advantage of using `auto`:
We don't have to pay any initialization cost until the function is called once :)

dhasenan: You mean import cost, no? Because "initializers" would be constructors, usually. You never pay for an instance initializer until an instance is created, and you always pay for a static constructor.

Anyway, that's easily addressed:

```D
template _Regex(Char)
{
  private import std.regex.internal.ir : IRRegex = Regex;
  alias _Regex = IRRegex!Char;
}
```

Pretty sure this technique has been discussed on the newsgroup several times.

wilzbach:
> I strive to be efficient.

https://issues.dlang.org/buglist.cgi?component=phobos&list_id=218415&product=D&resolution=---

Any help would be highly appreciated!

> Furthermore, according to @wilzbach, aliases are not an option. 
> ...
> I would be able to add an alias for compatibility, but that apparently isn't a universal trait of phobos devs

Oh I think you have misunderstood me a bit. I was trying to say that we can't use aliases to change the return type of a function if it's not `auto`, e.g.:

```d
Bar myFancyPhobosFunction()
{
...
}
```

and in a later version do the following:

```
alias Bar = Bar2;
```

`auto` essentially means to a developer "hey the return type of this function might change, use `typeof(..)` if you really need to store my type".

std/regex/package.d
wilzbach: I meant CTFE "initialization" cost - evaluation would have been a better word.

> Pretty sure this technique has been discussed on the newsgroup several times.

Well NG and actual PRs is a huge difference. Here's an overview on the status quo on making Phobos slicker:
- https://github.com/dlang/phobos/pull/5942
- https://github.com/dlang/phobos/pull/5931
- https://github.com/dlang/phobos/pull/5945
- https://github.com/dlang/phobos/pull/5948
- https://github.com/dlang/phobos/pull/5916
- http://forum.dlang.org/post/zezhvfdwutpdqkbkjymp@forum.dlang.org (latest NG thread)

tl;dr: there are still a lot of easy fruits to pick :)

dhasenan:
>> I strive to be efficient.

> https://issues.dlang.org/buglist.cgi?component=phobos&list_id=218415&product=D&resolution=---

> Any help would be highly appreciated!

Ah, I see the problem. When I said I strive to be efficient, you *thought* I said that I strive to do whatever you think is important whenever you want. I really have no clue how people keep confusing the two, but since they do it a lot, apparently it's an easy mistake to make.

> I was trying to say that we can't use aliases to change the return type of a function if it's not `auto`

Some discussion of why this is the case would be salient. Removing one type in favor of another and adding a compatibility alias seems like the obvious way to make a change with a reasonable deprecation cycle.

schveiguy:
FYI, here is a PR that fixes just the immediate issue for `RegexMatch.front`: https://github.com/dlang/phobos/pull/5976

> auto essentially means to a developer "hey the return type of this function might change, use typeof(..) if you really need to store my type".

This is not correct IMO. Only Voldemort returns make this guaranteed. And even there, it doesn't give you liberty to change the API of the return type.

For instance, if you have:

```D
auto foo(T)(T t)
{
   static if(is(T == int)) return cast(long)t * 2;
   else return t * 2;
}
```
`auto` doesn't say we can change the type without notice. It says "I don't want to have to make a template `IntOrLong` just to write a return type for this"

std/regex/package.d
```diff
@@ -695,21 +657,19 @@ public:
...
-/++
-    A regex engine state, as returned by $(D match) family of functions.
+// See commentary at https://github.com/dlang/phobos/pull/5963
+deprecated(`use typeof(match("", ""))`)
+public alias RegexMatch(R) = RegexMatch!R;
 
-    Effectively it's a forward range of Captures!R, produced
-    by lazily searching for matches in a given input.
-+/
-@trusted public struct RegexMatch(R)
```
DmitryOlshansky: Basically of this PR, 
only this can be kept

std/regex/package.d
```diff
@@ -734,42 +694,26 @@ public:
...
     }
 
-    ///ditto
     @property R post()
     {
         return _captures.post;
     }
 
-    ///ditto
     @property R hit()
```
DmitryOlshansky: The 2-nd thing to keep from this PR

std/regex/package.d
```diff
@@ -911,20 +851,48 @@ if (isSomeString!R && isRegexFor!(RegEx, R))
...
+    following additional members:
+
+    $(UL
+      $(LI pre: shortcut for $(D .front.pre).)
+      $(LI post: shortcut for $(D .front.post).)
+      $(LI hit: shortcut for $(D .front.hit).)
+      $(LI captures: compatibility alias for previous versions of std.regex.)
+    )
+    
+    Each capture struct offers the following methods:
```
DmitryOlshansky: Instead just document the Captures struct. This description is at best a skeleton

std/regex/package.d
```diff
@@ -789,16 +733,12 @@ public:
...
-    ///ditto
     auto save(){ return this; }
 
-    ///Test if this match object is empty.
     @property bool empty() const { return _captures._nMatch == 0; }
 
-    ///Same as !(x.empty), provided for its convenience  in conditional statements.
     T opCast(T:bool)(){ return !empty; }
 
-    /// Same as .front, provided for compatibility with original std.regex.
```
DmitryOlshansky: The 3-rd thing to keep from this PR

dhasenan:
> Instead just document the Captures struct.

We can't use real return types because we need to be able to change out the types returned without a deprecation cycle. But in other places where that's done, we either use a private type or a Voldemort type to make it actually work. Nobody can actually refer to `MapResult`, so we can switch it out trivially.

That's not the case with `Captures`, so we need to make it private as soon as possible, since that will require a deprecation cycle and we don't want to have to wait in order to change the return types here.

However, once `Captures` is private, it won't appear in ddoc. Therefore we need to document it elsewhere.

(Or we could just be honest about our inclination to change return types and stop using `auto`, but that idea was shot down.)

Anyway, I don't have the patience to deal with open source anymore, so you all can decide what you want. I'm out.

PR #5965, opened 2017-12-25
std.conv: ieeeQuadruple (AArch64) support for parse!real

jpf91:
I've added some code to support ieeeQuadruple reals in `parse!real`. I've also documented some of the parser. As I do not fully understand the internals of the parser, I didn't want to extend the parser to allow for higher precision. We'd have to use cent (or, as there is no cent type, two longs for a manual 128 bit in implementation) which probably makes the code slower. So I leave that to somebody who fully understands the parser code. (It would also make more sense to limit precision in the parser for double precision reals. Currently we parse at higher precision and remove precision when converting to the final value. This just adds unnecessary overhead).

I've added code to stuff the parse values into an ieeeQuadruple real though. This just means parse!real on ieeeQuadruple targets is limited to the same precision as on X86_64. Note that I had to implement bit-shifts for 128bit integers using two ulongs, which doesn't exactly make the code more readable...

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -2872,7 +2872,79 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
         enforce(ndigits, new ConvException("Error converting input"~
                         " to floating point"));
 
-        static if (real.mant_dig == 64)
+        import std.math : floatTraits, RealFormat;
+
+        static if (floatTraits!real.realFormat == RealFormat.ieeeQuadruple)
+        {
+            if (msdec)
+            {
+                /*
+                 * For quad precision, we currently allow max mantissa precision
+                 * of 64 bits, simply so we don't have to change the mantissa parser
+                 * in the code above. Feel free to adapt the parser to support full
+                 * 113 bit precision.
+                 */
+
+                // Exponent bias + 112:
+                // After shifting 112 times left, exp must be 1
+                int e2 = 0x3FFF + 112;
+
+                /*
+                 * left justify mantissa: The implicit bit (bit 112) must be 1
+                 * after this, (it is implicit and always defined as 1, so making
+                 * sure we end up with 1 at 112 means we adjust mantissa and eponent
+                 * to fit the ieee format)
+                 * For quadruple, this is especially fun as we have to use 2 longs
+                 * to store the mantissa and care about endianess...
+                 * quad_mant[0]               | quad_mant[1]
+                 * S.EEEEEEEEEEEEEEE.MMMMM .. | MMMMM .. 00000
+                 *                48       0  |
+                 */
+                ulong[2] quad_mant;
+                quad_mant[1] = msdec;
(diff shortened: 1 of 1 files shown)


std/conv.d
```diff
@@ -2849,7 +2872,79 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
...
+                        msw = &(cast(ulong*)&ldval)[0];
+                        lsw = &(cast(ulong*)&ldval)[1];
+                    }
+
+                    // Stuff mantissa directly into double
+                    // (first including implicit bit)
+                    *msw = quad_mant[0];
+                    *lsw = quad_mant[1];
+
+                    //Store exponent, now overwriting implicit bit
```
ibuclaw: nit, space.

std/conv.d
```diff
@@ -2743,8 +2743,21 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
 
     if (isHex)
     {
+        /*
+         * The following algorithm consists of mainly 3 steps (and maybe should
+         * be refactored into functions accordingly)
+         * 1) Parse the textual input into msdec and exp variables:
+         *    input is 0xaaaaa...p+000... where aaaa is the mantissa in hex and
+         *    000 is the exponent in decimal format.
+         * 2) Rounding / sticky bit, whatever?
```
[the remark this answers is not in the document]
jpf91: That was a note that I do not totally understand what the algorithm is doing with these guard bits... Anyway, forgot to remove that note :-)

std/conv.d
```diff
@@ -2849,7 +2872,79 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
...
+                    else
+                    {
+                        msw = &(cast(ulong*)&ldval)[0];
+                        lsw = &(cast(ulong*)&ldval)[1];
+                    }
+
+                    // Stuff mantissa directly into double
+                    // (first including implicit bit)
+                    *msw = quad_mant[0];
+                    *lsw = quad_mant[1];
```
ibuclaw: Would using a union here be better? I'm a bit weary of the optimizer here, incase it just sees this as dead code (msw/lsw are set but not read).

Would probably be more clearer as well that you are changing the value of `ldval` directly.

ibuclaw: Probably not a problem however. I just have a slight distaste for using pointers for bit manipulation in std.math that carries over here.

jpf91: The IEEE extended and double branches used pointers as well and I'm not sure if this is relevant for CTFE-ability? I think the optimizer shouldn't be much of a problem, as the pointers are used to write to the value?

ibuclaw: I've seen it go wrong before when I implemented `floorImpl`. From what I recall, it couldn't trace back to the original type, and so thought that the program was modifying bits outside the range of the type it was pointing to.

PR #4924, opened 2016-11-24
std.random enhancement: overload uniform to return array of elements

edi33416:
Overload uniform to return an array of elements of the given type.

For example, uniform!(int[])(10) will return an array of ten uniformly generated integers between int.min and int.max.

PR #5967, opened 2017-12-28
Add Array of Random Numbers Example

JackStouffer:
Supersedes https://github.com/dlang/phobos/pull/4924

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -1402,6 +1402,17 @@ if (!is(CommonType!(T1, T2) == void))
     auto b = uniform(0.0f, 1.0f, gen);
 }
 
+/// Create an array of random numbers using range functions and UFCS
+@safe unittest
+{
+    import std.array : array;
+    import std.range : generate, takeExactly;
+
+    int[] arr = generate!(() => uniform(0, 100)).takeExactly(10).array;
+    assert(arr.length == 10);
+    assert(arr[0] >= 0 && arr[0] < 100);
+}
+
 @safe unittest
 {
     MinstdRand0 gen;


quickfur:
Nice.  But checking just the first element for range is somewhat underwhelming.  Anything more interesting we could do with an array of random numbers?

std/random.d
```diff
@@ -1402,6 +1402,17 @@ if (!is(CommonType!(T1, T2) == void))
...
+/// Create an array of random numbers using range functions and UFCS
+@safe unittest
+{
+    import std.array : array;
+    import std.algorithm.iteration : iota, map;
+
+    int[] arr = iota(100).map!(() => uniform(0, 100)).array;
+    assert(arr.length == 100);
+    assert(arr[0] >= 0 && arr[0] < 100);
+}
```
wilzbach: Ideas:
- `arr.sum > 0`
- use a seed
- seed the global RNG before

I would go with (b). It's common practice to use explicit seeding with RNGs.

JackStouffer:
I'm not really looking for the tests to actually test anything of substance. The point of the example is to be illustrative of different range functions, not to test functionality of `uniform`.

std/random.d
```diff
@@ -1402,6 +1402,18 @@ if (!is(CommonType!(T1, T2) == void))
...
 }
 
+/// Create an array of random numbers using range functions and UFCS
+@safe unittest
+{
+    import std.array : array;
+    import std.algorithm.iteration : map;
+    import std.range : iota;
+
+    int[] arr = iota(100).map!((_) => uniform(0, 100)).array;
```
wilzbach: You could have omitted the parentheses around the underscore.

std/random.d
```diff
@@ -1402,6 +1402,18 @@ if (!is(CommonType!(T1, T2) == void))
...
+    assert(arr.length == 100);
+    assert(arr[0] >= 0 && arr[0] < 100);
+}
+
```
wilzbach: BTW here's another way:

```d
import std.stdio, std.range, std.random;
void main(string[] args)
{
    generate!(() => uniform(0, 100)).take(10).writeln;
}
```

https://run.dlang.io/is/5hjbtK

klickverbot: Yep, generate is probably the way to go rather than the "old-school" iota/map. Not using the same arbitrary choice (100) for both length and range also helps comprehensibility.

PR #5968, opened 2017-12-28
findSplit: return named tuples instead of an anonymous tuple

wilzbach:
Split-up from https://github.com/dlang/phobos/pull/5962

Motivation: Named tuples are a lot better to read.

There has been a precedent for this, see #5436

### DIFF
--- a/changelog/std-algorithm-searching-findSplit.dd
+++ b/changelog/std-algorithm-searching-findSplit.dd
@@ -1,0 +1,23 @@
+`std.algorithm.searching.findSplit` returns a named tuple
+
+$(REF findSplit, std,algorithm,searching),
+$(REF findSplitBefore, std,algorithm,searching), and
+$(REF findSplitAfter, std,algorithm,searching)
+will return a named tuple instead of an anonymous tuple:
+
+---
+import std.algorithm.searching;
+auto a = "Carl Sagan Memorial Station";
+auto r = findSplit(a, " ");
+assert(r.pre == "Carl");
+assert(r.match == " ");
+assert(r.post == "Sagan Memorial Station");
+
+auto r1 = findSplitBefore(a, "Sagan");
+assert(r1.pre == "Carl ");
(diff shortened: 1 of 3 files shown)


wilzbach:
Note that `findSplit` return `auto`, so it's less problematic to change the return type here.

std/algorithm/searching.d
```diff
@@ -2353,6 +2353,8 @@ if (Ranges.length > 1 && is(typeof(startsWith!pred(haystack, needles))))
 
 @safe unittest
 {
+    import std.typecons : tuple;
```
JackStouffer: not necessary, it's imported at the top

changelog/std-algorithm-searching-findSplit.dd
```diff
@@ -0,0 +1,23 @@
+`std.algorithm.searching.findSplit` returns a named tuple
+
+$(REF findSplit, std,algorithm,searching),
+$(REF findSplitBefore, std,algorithm,searching), and
+$(REF findSplitAfter, std,algorithm,searching)
+will return a named tuple tuple instead of an anonymous tuple:
```
quickfur: Typo: "tuple tuple" > "tuple"

JackStouffer:
Let's move forward with this.

jacob-carlborg:
This broke DStep [1]. It's the same error as the second commit https://github.com/dlang/phobos/pull/5968/commits/8becc70da790826bccec734d40021dd7ff22f485 in this PR fixes. But according to the documentation of `Tuple` [2] it should be possible to use both indexing and names to access the elements of a tuple.

[1] https://travis-ci.org/jacob-carlborg/dstep/jobs/333759750
[2] https://dlang.org/phobos/std_typecons.html#Tuple

jacob-carlborg:
I can we please revert this. It broke Phobos and DStep, then it's likely that it will break other projects as well.

PR #5970, opened 2017-12-29
Markdownify std.algorithm

wilzbach:
Done with: https://github.com/wilzbach/dscanner-fix/blob/master/d_code_markdown.d

Revival of https://github.com/dlang/phobos/pull/5801 for `std.algorithm`

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -8,41 +8,41 @@ $(BOOKTABLE Cheat Sheet,
 $(TR $(TH Function Name) $(TH Description))
 $(T2 among,
         Checks if a value is among a set of values, e.g.
-        $(D if (v.among(1, 2, 3)) // `v` is 1, 2 or 3))
+        `if (v.among(1, 2, 3)) // `v` is 1, 2 or 3`)
 $(T2 castSwitch,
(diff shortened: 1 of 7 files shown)


wilzbach:
We can also do this file by file if that's easier.

wilzbach:
> Because this was done via program and not by hand you can infer the correct-ness by looking at a few examples of the output (

Not that there are still some cases in which DDoc auto-highlights a token, compare e.g.

Old: http://dtest.dlang.io/artifact/website-d4a3e415e1933b1bf95375258832afc0fa2d38d9-cf0b6d829835d57b259c20e7d81ed874/web/library-prerelease/std/algorithm/sorting/top_n.html
New: http://dtest.dlang.io/artifact/website-d4a3e415e1933b1bf95375258832afc0fa2d38d9-dd7fe723dce5f3ab6d81d0f7a8671d92/web/library-prerelease/std/algorithm/sorting/top_n.html

and look at "r.length" after "satisfy".

Though AFAICT these changes look benign as the auto-highlighting actually does make sense in these spots.

JackStouffer:
Let's get on with this.

d-random-contributor:
First change: backticks don't nest.

PR #5971, opened 2017-12-29
Make std.format.formattedRead usable with rvalues

Kozzi11:
http://forum.dlang.org/post/pmvhclhwxebldzffdwyg@forum.dlang.org

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -631,7 +631,7 @@ matching failure happens.
 Throws:
     An `Exception` if `S.length == 0` and `fmt` has format specifiers.
  */
-uint formattedRead(alias fmt, R, S...)(ref R r, auto ref S args)
+uint formattedRead(alias fmt, R, S...)(auto ref R r, auto ref S args)
 if (isSomeString!(typeof(fmt)))
 {
     alias e = checkFormatException!(fmt, S);
@@ -640,7 +640,7 @@ if (isSomeString!(typeof(fmt)))
 }
 
 /// ditto
-uint formattedRead(R, Char, S...)(ref R r, const(Char)[] fmt, auto ref S args)
+uint formattedRead(R, Char, S...)(auto ref R r, const(Char)[] fmt, auto ref S args)
 {
     import std.typecons : isTuple;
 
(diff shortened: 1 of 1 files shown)


std/format.d
```diff
@@ -978,6 +978,18 @@ uint formattedRead(R, Char, S...)(ref R r, const(Char)[] fmt, auto ref S args)
     assert(aa3 == ["hello":1, "world":2]);
 }
 
+// test rvalue using
+@system pure unittest
+{
+    string[int] aa1;
+    formattedRead!("%s")(`[1:"hello", 2:"world"]`, &aa1);
```
wilzbach: You don't need to take `aa1` by reference here. That's the old style, which we should look into deprecating. The compiler and `formattedRead` are smart enough to automatically do it for you.

std/format.d
```diff
@@ -978,6 +978,18 @@ uint formattedRead(R, Char, S...)(ref R r, const(Char)[] fmt, auto ref S args)
...
+    assert(aa1 == [1:"hello", 2:"world"]);
+
+    int[string] aa2;
+    formattedRead(`{"hello"=1; "world"=2}`, "{%(%s=%s; %)}", &aa2);
```
wilzbach: I see that we only have one public example here. What do you think about adding another public example which shows fancy stuff like reading an array or associative array?
(This is independent from this PR, of course)

quickfur:
OK, I filed a bugzilla bug for you: https://issues.dlang.org/show_bug.cgi?id=18152

schveiguy:
@Kozzi11 if you could amend your commit message to say "Fixes issue 18152", we are all set to merge.

wilzbach:
> @Kozzi11 if you could amend your commit message to say "Fixes issue 18152", we are all set to merge.

@Kozzi11 I quickly did it for you to reduce the latency ;-)

quickfur:
@wilzbach Ah you beat me to it.  I was about to do the exact same thing. :-D

wilzbach:
> @wilzbach Ah you beat me to it. I was about to do the exact same thing. :-D

Hehe, I'm fully automated: https://wiki.dlang.org/Guidelines_for_maintainers#Write_access_to_PRs :innocent:

schveiguy:
I didn't know that meant we could edit commits :) I thought we could just add more.

quickfur:
In retrospect, though, I would've added an empty commit on top of the OP's, so that the git history would still have the proper attribution to his work.

wilzbach:
> In retrospect, though, I would've added an empty commit on top of the OP's, so that the git history would still have the proper attribution to his work.

Huh? I'm not sure I follow. GitHub is smart enough to differentiate between the author and committer:

https://github.com/dlang/phobos/commit/146306a27b80c407d619a4aa8aa44f8fc562e343

In you local `git log`, there's only `author`.

quickfur:
Actually, I just checked `git log`: the author is still set to the OP. Nevermind what I said then. :D

PR #5975, opened 2017-12-30
Fix 'among' doc link in std.algorithm.searching

dcarp:


wilzbach:
I will rebase this to `stable`, s.t. it will be released on dlang.org soon.

PR #5976, opened 2017-12-31
Fix issue 18124 -- RegexMatch.front isn't clear as to what it returns

schveiguy:
Easiest thing to do was explicitly type the result. It's probably less effort than documenting what it returns in ddoc, and is clearer to the user.

In all other cases in this file, the auto returns have explicit docs as to the return type. This really was the only anomaly.

### DIFF
--- a/std/regex/package.d
+++ b/std/regex/package.d
@@ -764,7 +764,7 @@ public:
         assert(m.empty);
         ---
     +/
-    @property auto front()
+    @property inout(Captures!R) front() inout
     {
         return _captures;
     }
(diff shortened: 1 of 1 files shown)


quickfur:
I'm not sure this is a good idea.  It binds us to an explicit return type that we may not be able to change in the future, e.g., if for example a future implementation wants to return distinct types for `matchFirst` and `matchAll` (since the former only needs to encapsulate a single match whereas the latter a range of matches).

schveiguy:
> It binds us to an explicit return type that we may not be able to change in the future

That ship has sailed, the return type is not a voldemort type, so code may already exists that uses the real type name. It's what I would do if I were stuffing the result into a struct.

schveiguy:
In fact, the docs of `RegexMatch` already state it's a forward range of `Captures` elements. This just clarifies it close to where you would be looking for the element type.

JackStouffer:
Wait, I apologize. I didn't see your other comments before I submitted the review.

PR #5977, opened 2017-12-31
[WIP] Add std.meta.apply

PetarKirov:
TODO:

- [ ] add more tests which properly motivate this addition
- [ ] add changelog entry

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -1733,8 +1733,16 @@ if (stepSize != 0)
  * Params:
  *    Template = The template to instantiate.
  *    Params = The parameters with which to instantiate the template.
+ *
  * Returns:
  *    The instantiated template.
+ *
+ * See_Also: $(LREF apply)
+ *
+ * $(LREF apply) offers a superset of `Instantiate`'s functionality, but
+ * its use may not always be appropriate because it blurs the difference
+ * between template instantiation and compile-time or run-time function
+ * calling in generic code.
  */
 alias Instantiate(alias Template, Params...) = Template!Params;
 
--- a/std/meta.d
+++ b/std/meta.d
@@ -56,6 +56,7 @@
  *           $(LREF staticIsSorted)
  * ))
  * $(TR $(TD Template instantiation) $(TD
+ *           $(LREF apply)
  *           $(LREF ApplyLeft)
  *           $(LREF ApplyRight)
  *           $(LREF Instantiate)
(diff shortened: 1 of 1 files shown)


schveiguy:
Doesn't `Instantiate` already do something similar to this? Would it not be better to simply extend `Instantiate` to do things you need?

JackStouffer:
bike-sheding: this should be named `Apply` to conform with the naming conventions of the rest of std.meta

ZombineDev:
@schveiguy yes, unfortunately `Instantiate` is similar, but not quite as useful, I need to add more tests which properly motivate this addition.

@JackStouffer I considered this, but it wasn't a clear call since it can be used with values, where `apply` would be more appropriate and with templates, where `Apply` would be better. However the most important point is that `apply` / `Apply` combines both behaviours, which brings a tremendous simplification in generic code.

std/meta.d
```diff
@@ -1210,6 +1211,147 @@ template aliasSeqOf(alias range)
     }
 }
 
+/**
+ * Applies the sequence of arguments `args` to the function or template `func`
+ * either as function or template arguments, at compile-time or at run-time
+ * (if CTFE is not possible).
```
andralex: The introduction paragraph should answer the obvious question "What's wrong with `func(args)`?"

std/meta.d
```diff
@@ -1210,6 +1211,147 @@ template aliasSeqOf(alias range)
...
+ *
+ * `apply` is most likely to be found used as a building block in higher-order
+ * templates and $(LREF AliasSeq) algorithms like $(LREF staticMap) which need
+ * to call or instantiate (as appropriate) functions or templates, passed as
+ * parameters.
+ *
+ * Params:
+ *      func  = function or template to call or instantiate
+ *      args = template / function arguments to `apply` to `fun`
```
andralex: no spaces around "/" in text

std/meta.d
```diff
@@ -1210,6 +1211,147 @@ template aliasSeqOf(alias range)
...
+    int add(int a, int b) { return a + b; }
+    enum int a = 2, b = 3;
+    enum int sum = apply!(add, a, b);
+    static assert(sum == 5);
+
+    int x = 2, y = 5;
+    int z = apply!((a, b) => a + b, x, y);
+    assert(z == 7);
+
+    // TODO: Show higher-order template example
```
andralex: This can't be part of the published documentation.

LightBender:
PR closed as stalled. If you wish to resurrect it you are welcome to do so. @jmdavis If this has any relevance to your Phobos 3 work let us know.

PR #5982, opened 2018-01-02
Fix Issue 18153 - deprecate public symbols ByLine, ByRecord, ByChunk

wilzbach:
Note: these symbols were documented until 2013/08 in https://github.com/dlang/phobos/pull/1448 (rationale back then: https://github.com/dlang/phobos/pull/1408#discussion_r5178781).

Please see the issue for the motivation.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -252,7 +252,7 @@ version(HAS_GETDELIM) extern(C) nothrow @nogc
 }
 
 //------------------------------------------------------------------------------
-struct ByRecord(Fields...)
+private struct ByRecordImpl(Fields...)
 {
 private:
     import std.typecons : Tuple;
@@ -307,9 +307,17 @@ public:
     }
 }
 
+// @@@DEPRECATED_2019-01@@@
+deprecated("Use .byRecord")
+struct ByRecord(Fields...)
+{
+    ByRecordImpl!Fields payload;
+    alias payload this;
+}
+
 template byRecord(Fields...)
 {
-    ByRecord!(Fields) byRecord(File f, string format)
+    auto byRecord(File f, string format)
     {
         return typeof(return)(f, format);
     }
@@ -2067,7 +2075,7 @@ Range that reads one line at a time.  Returned by $(LREF byLine).
 
 Allows to directly use range operations on lines of a file.
 */
-    struct ByLine(Char, Terminator)
+    private struct ByLineImpl(Char, Terminator)
     {
     private:
         import std.typecons : RefCounted, RefCountedAutoInitialize;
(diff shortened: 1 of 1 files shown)


std/stdio.d
```diff
@@ -307,9 +307,13 @@ public:
     }
 }
 
+/// @@@DEPRECATED_2019-01@@@
+deprecated("Use .byRecord")
+alias ByRecord = ByRecordImpl;
```
schveiguy: Stupid question: Does alias make `ByRecord` public even though it's private? I seem to remember even if you public aliased a private symbol, it still didn't work.

quickfur: I just tested this:
````
// mod.d
module mod;
private struct S {}
alias PubS = S;

// test.d
import mod;
void main() {
        PubS s; // OK
        //S t;  // NG
}
````
Compiler accepts the first declaration `s`, but rejects the declaration `t`. So we're good.

Comment (jmdavis) [reply std/stdio.d]: I thought that that was the case, but some investigation shows that that's not quite right - e.g. [this comment](https://issues.dlang.org/show_bug.cgi?id=4533#c7) by Marten on an old enhancement request indicates that it works except for functions, and testing it, that seems to be the case. I get an error if I try to use a public alias of a private function, but it works with a variable or type.

However, I would point out that the comment shouldn't be a ddoc comment. `ByRecord` isn't documented, and this then documents it. The same goes for the other comments. They should be regular comments, not documentation comments.

Comment (jmdavis) [reply std/stdio.d]: And H.S.Teoh replied while I was in the middle of my own reply. :)

Comment (wilzbach) [reply std/stdio.d]: I just tested this as well by simply inserting `ByLine!(char, char)` into a random Phobos module:

```
std/file.d(4384): Deprecation: std.stdio.File.ByLine!(char, char).ByLineImpl is not visible from module file
```

Hence, I used the `alias this` trick ->


```
std/file.d(4384): Deprecation: std.stdio.File.ByLine!(char, char).ByLineImpl is not visible from module file
```

**Comment by schveiguy** [comment]
OK, so are we settled on using alias this'd structs or should it be put back to simple aliases? Sorry, didn't mean to throw such a wrench in here with my ignorance 😆

File: std/stdio.d (hunk 1 above)
Comment (schveiguy) [reply std/stdio.d]: Is there supposed to be a difference between those two?

**Comment by wilzbach** [comment]
> OK, so are we settled on using alias this'd structs or should it be put back to simple aliases? Sorry, didn't mean to throw such a wrench in here with my ignorance :laughing:

On the contrary, you realized that the simple `alias` doesn't work :/
Copy/pasting from the now hidden discussion:

----

I just tested this as well by simply inserting `ByLine!(char, char)` into a random Phobos module:

```
std/file.d(4384): Deprecation: std.stdio.File.ByLine!(char, char).ByLineImpl is not visible from module file
```

With `alias this`:

```
std/file.d(4384): Deprecation: struct std.stdio.File.ByLine!(char, char).ByLine is deprecated - Use .byLine
```

std/stdio.d
schveiguy: ok, I see it now.

wilzbach: Copy/pasting mistake - sorry.

schveiguy: Hm... maybe it's because it's a template. In any case, the alias this trick works.

schveiguy:
Thanks, looks like it's all working now.

PR #5985, opened 2018-01-03
@safeify more unittests II

wilzbach:
Slowly, but steadily.

### DIFF
--- a/std/base64.d
+++ b/std/base64.d
@@ -1766,12 +1766,11 @@ class Base64Exception : Exception
 }
 
 ///
-@system unittest
+@safe unittest
 {
     import std.exception : assertThrown;
     assertThrown!Base64Exception(Base64.decode("ab|c"));
 }
-
 
 @system unittest
 {
@@ -2039,7 +2038,7 @@ class Base64Exception : Exception
 }
 
 // Regression control for the output range ref bug in encode.
-@system unittest
+@safe unittest
 {
     struct InputRange
     {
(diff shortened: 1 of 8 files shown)


std/base64.d
```diff
@@ -2094,6 +2095,8 @@ class Base64Exception : Exception
     // Verify that any existing workaround that uses & still works.
     InputRange ir2;
     OutputRange or2;
-    assert(Base64.decode(ir2, &or2) == 6);
+    () @trusted {
```
JackStouffer: This seems kind of besides the point. decode is the function really being tested here, so it makes sense to leave this unit test system if it's system. Same for the above test

std/format.d
```diff
@@ -1909,13 +1909,13 @@ void formatValue(Writer, T, Char)(auto ref Writer w, auto ref T val, const ref F
...
  */
-@system unittest
+@safe unittest
 {
    struct Point
    {
        int x, y;
 
-       void toString(scope void delegate(const(char)[]) sink,
+       void toString(scope void delegate(const(char)[]) @safe sink,
```
JackStouffer: Technically, this wouldn't actually work with a lot of real world code (and I think std.format). This is one of my big gripes with the delegate sink to string is that attributes don't really work (and I still have to formally propose the output range overload).

std/base64.d
wilzbach: Did you look the the full context?
This is just a legacy line - the real test is five lines above.

```d
assert(Base64.decode(ir, or) == 6);
assert(or.result == [0x1a, 0x2b, 0x3c, 0x4d, 0x5d, 0x6e]);
 	 
// Verify that any existing workaround that uses & still works
```

PR #5986, opened 2018-01-03
Fix Issue 18186 - std.array.replaceSlice should be usable in @safe

wilzbach:


std/array.d
```diff
@@ -2799,25 +2799,27 @@ in
...
     auto result = new T[s.length - slice.length + replacement.length];
-    immutable so = slice.ptr - s.ptr;
+    immutable so = &slice[0] - &s[0];
     result[0 .. so] = s[0 .. so];
     result[so .. so + replacement.length] = replacement[];
     result[so + replacement.length .. result.length] =
         s[so + slice.length .. s.length];
 
-    return cast(inout(T)[]) result;
+    return () @trusted inout {
```
JackStouffer: Can you check if this gets inlined by DMD? If not I don't think this needs to be stalled, but I've run into cases where trusted escapes aren't inlined and I want to make sure incase a bug report should be made.

wilzbach: LDC is smart enough: https://godbolt.org/g/oeq5xh, DMD isn't. 
There are already existing issues for this, it's a well-known problem, e.g.

https://issues.dlang.org/show_bug.cgi?id=10848
https://issues.dlang.org/show_bug.cgi?id=9785

ZombineDev: Does the lambda really need the `inout` attribute? I believe that only member functions and free functions where there's a relationship between the const-ness of a parameter and the return type can have this qualifier.

wilzbach: Yes otherwise DMD complains:

```
std/array.d(2808): Error: inout on return means inout must be on a parameter as well for pure nothrow @nogc @trusted inout(char)[]()
```

PR #5987, opened 2018-01-03
Fix Issue 17249: Added BigInt.getDigit

JackStouffer:
Get the elements of the underlying representation. This would also be useful in my `stdxdecimal` library.

### DIFF
--- a/std/bigint.d
+++ b/std/bigint.d
@@ -1019,6 +1019,53 @@ public:
         assert(aa[BigInt(456)] == "def");
     }
 
+    /**
+     * Gets the nth number in the underlying representation that makes up the whole
+     * `BigInt`.
+     *
+     * Params:
+     *     T = the type to view the underlying representation as
+     *     n = The nth number to retrieve. Must be less than $(LREF ulongLength) or
+     *     $(LREF uintLength) with respect to `T`.
+     * Returns:
+     *     The nth `ulong` in the representation of this `BigInt`.
+     */
+    T getDigit(T = ulong)(size_t n) const
+    if (is(T == ulong) || is(T == uint))
+    {
+        static if (is(T == ulong))
+        {
+            assert(n < ulongLength(), "getDigit index out of bounds");
+            return data.peekUlong(n);
+        }
+        else
+        {
+            assert(n < uintLength(), "getDigit index out of bounds");
+            return data.peekUint(n);
+        }
+    }
+
+    ///
+    @system pure unittest
+    {
+        auto a = BigInt("1000");
+        assert(a.ulongLength() == 1);
+        assert(a.getDigit(0) == 1000);
+
+        assert(a.uintLength() == 1);
+        assert(a.getDigit!uint(0) == 1000);
+
+        auto b = BigInt("2_000_000_000_000_000_000_000_000_000");
+        assert(b.ulongLength() == 2);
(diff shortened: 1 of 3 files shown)


std/bigint.d
```diff
@@ -1019,6 +1019,65 @@ public:
         assert(aa[BigInt(456)] == "def");
     }
 
+    /**
+     * Params:
+     *     n = The nth number to retrieve. Must be less than $(LREF ulongLength).
+     * Returns:
+     *     The nth `ulong` in the representation of this `BigInt`.
+     */
+    ulong ulongDigit(int n) pure nothrow const @safe @nogc
```
wilzbach: How about using only one functions?
Something lilke:

```d
Type digit(Type)(size_t n)
```

Usage:

```d
BigInt(20).digit!ulong(0).
```

JackStouffer:
@wilzbach done

std/bigint.d
```diff
@@ -1019,6 +1019,53 @@ public:
...
+     *
+     * Params:
+     *     T = the type to view the underlying representation as
+     *     n = The nth number to retrieve. Must be less than $(LREF ulongLength) or
+     *     $(LREF uintLength) with respect to `T`.
+     * Returns:
+     *     The nth `ulong` in the representation of this `BigInt`.
+     */
+    T getDigit(T = ulong)(size_t n) const
+        if (is(T == ulong) || is(T == uint))
```
wilzbach: Constraint indentation should be on the same level... I thought there was a test for this :/

wilzbach: https://github.com/dlang-community/D-Scanner/pull/450 - it just never made it into DScanner.

std/internal/math/biguintcore.d
```diff
@@ -168,7 +169,7 @@ public:
             return (n & 1) ? cast(uint)(x >> 32) : cast(uint) x;
         }
     }
-public:
+
     ///
```
quickfur: Was this removed because the following method wasn't intended to be public?  Also, does this change somehow relate to this PR?

wilzbach: Public is already set a couple of lines above. It's pretty redundant and I don't think it's a problem to include such small things if they are trivial.

quickfur: Fair enough.

Though I'm starting to agree with @andralex that `private` should be attached per declaration rather than wholesale, for better context when reading the code. But that's for another day.

quickfur:
ping @andralex 

This one has been rotting here for a while. Let's move it forward!

std/bigint.d
```diff
@@ -1019,6 +1019,53 @@ public:
...
+     *
+     * Params:
+     *     T = the type to view the underlying representation as
+     *     n = The nth number to retrieve. Must be less than $(LREF ulongLength) or
+     *     $(LREF uintLength) with respect to `T`.
+     * Returns:
+     *     The nth `ulong` in the representation of this `BigInt`.
+     */
+    T getDigit(T = ulong)(size_t n) const
+    if (is(T == ulong) || is(T == uint))
```
andralex: I'd think ushort and ubyte would also be useful, but can be added later.

JackStouffer:
A note could be added but the docs say it's about the representation and the representation is the same for both negative an positive numbers.

PR #5989, opened 2018-01-03
Use static foreach in Phobos

wilzbach:
All the cool kids in town use `static foreach`. Let's join the club.
Motivation: `static foreach` makes it clear to the reader that the loop is unrolled
during compile-time.
The replacements were done semi-automatically.

I also removed the workarounds against BUG 2396 - `static foreach` doesn't seem
to be affected by this as overall execution time of the entire Phobos testsuite
doesn't with this PR.

I know that this got quite large, but the changes itself are trivial.

### DIFF
--- a/std/bitmanip.d
+++ b/std/bitmanip.d
@@ -2428,8 +2428,8 @@ private ulong swapEndianImpl(ulong val) @trusted pure nothrow @nogc
 @safe unittest
 {
     import std.meta;
-    foreach (T; AliasSeq!(bool, byte, ubyte, short, ushort, int, uint, long, ulong, char, wchar, dchar))
-    {
+    static foreach (T; AliasSeq!(bool, byte, ubyte, short, ushort, int, uint, long, ulong, char, wchar, dchar))
+    {{
         scope(failure) writefln("Failed type: %s", T.stringof);
         T val;
         const T cval;
@@ -2470,7 +2470,7 @@ private ulong swapEndianImpl(ulong val) @trusted pure nothrow @nogc
                 right <<= 8;
             }
         }
-    }
+    }}
 }
 
 
@@ -2544,7 +2544,7 @@ if (isFloatOrDouble!T)
 @safe unittest
 {
     import std.meta;
-    foreach (T; AliasSeq!(bool, byte, ubyte, short, ushort, int, uint, long, ulong,
+    static foreach (T; AliasSeq!(bool, byte, ubyte, short, ushort, int, uint, long, ulong,
                          char, wchar, dchar
         /* The trouble here is with floats and doubles being compared against nan
          * using a bit compare. There are two kinds of nans, quiet and signaling.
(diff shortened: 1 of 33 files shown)


std/bitmanip.d
```diff
@@ -3994,17 +3994,17 @@ if (canSwapEndianness!T && isOutputRange!(R, ubyte))
 {
     import std.array;
     import std.format : format;
-    import std.meta;
-    foreach (endianness; AliasSeq!(Endian.bigEndian, Endian.littleEndian))
-    {
+    import std.meta : AliasSeq;
+    static foreach (endianness; [Endian.bigEndian, Endian.littleEndian])
```
MetaLang: Why the change from `AliasSeq` to an array?

wilzbach: Because `static foreach` can unroll arrays and thus there's no need for `AliasSeq` anymore.

MetaLang: Yes but why only this one specifically? I know most of these loops are over lists of types, but there's at least one other one that loops over a list of values.

wilzbach: Then I overlooked the other case. This PR catches only 95% of the `static foreach`'s.
Anyways, we can easily make more improvements to this. Where did you see the other list of values?

wilzbach:
BTW seeing how easy it is to replace `foreach` with `static foreach`, should we deprecate compile-time loop unrolling in the normal `foreach`?
If that's not wanted, we could at least think about introducing a DScanner check.

jmdavis:
Is there any point to using `static` on the `foreach`es which use `AliasSeq` other than style?

std/bitmanip.d
MetaLang: It's not that important; I was just trying to understand why you did it for one and not the others. Looking at the other cases, however, an array can't or shouldn't be used because they're separate types.

MetaLang:
Does anyone know if `dfix` is still under active development? This type of code rewriting is exactly what it was made for.

jmdavis:
> BTW seeing how easy it is to replace foreach with static foreach, should we deprecate compile-time loop unrolling in the normal foreach?

There's a _lot_ of code out there doing stuff like using `AliasSeq` with `foreach`. Unless we have a really good reason for deprecating the behavior, I suggest that we leave it alone. Also, I don't particularly want to restrict the compiler's ability to optimize `foreach`, and if trying to shift some of its current semantics over to `static foreach` does that, then that seems like a negative (I don't know if it would or not).

wilzbach:
> Does anyone know if dfix is still under active development? This type of code rewriting is exactly what it was made for.

AFAICT only @UplinkCoder and Sociomantic have done a bit of work on it recently.
I started https://github.com/wilzbach/dscanner-fix with the idea of automatically fixing DSCanner warnings a while ago, but it's a more or less a collection of crude hacks (though some of them have been used to upgrade Phobos).

PR #4712, opened 2016-08-04
file.d: add overflow checks

WalterBright:

PR #5990, opened 2018-01-03
file.d: add overflow checks

JackStouffer:
Taken over from https://github.com/dlang/phobos/pull/4712

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -336,6 +336,7 @@ version (Posix) private void[] readImpl(const(char)[] name, const(FSChar)* namez
     import std.algorithm.comparison : min;
     import std.array : uninitializedArray;
     import std.conv : to;
+    import std.experimental.checkedint : checked;
 
     // A few internal configuration parameters {
     enum size_t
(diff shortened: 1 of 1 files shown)


std/file.d
```diff
@@ -2645,14 +2647,19 @@ version(Posix) @system unittest // input range of dchars
...
+    import std.experimental.checkedint : checked;
     /* GetCurrentDirectory's return value:
         1. function succeeds: the number of characters that are written to
     the buffer, not including the terminating null character.
         2. function fails: zero
         3. the buffer (lpBuffer) is not large enough: the required size of
     the buffer, in characters, including the null-terminating character.
     */
-    wchar[4096] buffW = void; //enough for most common case
+    version (unittest)
```
wilzbach: `StdUnittest`

wilzbach: See https://github.com/dlang/phobos/pull/5932

JackStouffer:
@wilzbach done

adamdruppe:
Is it sane to depend on a std.experimental module in the main production branch of the library?

wilzbach:
We already discussed this in https://github.com/dlang/phobos/pull/4712 where the consensus between everyone (including @WalterBright and @andralex) was that using the `Checked` wrapper is a lot better than creeping in ugly manual overflow checks.
std.experimental is experimental because of its API not because it's not tested. Using it internally within Phobos is no problem because the user won't even know it and we can still do breaking changes with its API. It's essentially the same as if we would have added our own SafeInt wrapper internally.

PR #5991, opened 2018-01-03
[RFC] Make Output Ranges Using Methods The Preferred toString Option

JackStouffer:
This proposal

* Modifies `std.format` to recognize any `toString` that accepts a character accepting output range as having a valid `toString` overload
* Modifies `std.format` to prioritize and use the output range method when formatting aggregate types if it exists
* De-emphasizes the `scope delegate` `toString` methods in the docs

## Rationale

Firstly, speed. [According to Walter](https://issues.dlang.org/show_bug.cgi?id=9785#c3), the delegate passed as a parameter can never be in-lined because of how DMD does data flow analysis. Using instead the defined `put` method of an output range would solve that problem.

Secondly, the vision of Phobos.  Ranges are the present and the future. Output ranges really need more support in Phobos (std.conv).

Thirdly, safety. The vast majority of the code I've seen doesn't mark the `delegate` as `@safe` to allow non-safe sinks to work with it. This means that the `toString` method itself can never be marked `@safe` either. Making `toString` a template who's safety relies on the `put` method of the output range side-steps this problem entirely.

### DIFF
--- a/changelog/toString.dd
+++ b/changelog/toString.dd
@@ -20,11 +20,11 @@ auto t = MyType();
 writeln(t); // writes "Custom toString"
 -------
 
-This has several benefits for the user. Firstly, this design is much friendlier
+This has several benefits for the user. First, this design is much friendlier
 to inlining than the `toString(scope void delegate(const(char)[]) sink)` method of
-`toString`. Secondly, this cuts down on memory usage, as characters are placed right
-into the output buffers of functions like $(REF format, std, format). Thirdly,
-because `toString` is now a template, can be made `@safe` via inference much more
+`toString`. Second, this cuts down on memory usage, as characters are placed right
+into the output buffers of functions like $(REF format, std, format). Third,
+because `toString` is now a template, can be marked `@safe` via inference much more
 often.
 
 All previous forms of `toString` will continue to work.
--- a/changelog/toString.dd
+++ b/changelog/toString.dd
@@ -1,0 +1,30 @@
+toString Can Now Use Output Ranges
+
+The standard library has been modified to recognize and use `toString` overloads
+that accept $(REF_ALTTEXT output ranges, isOutputRange, std, range, primitives)
+when such overloads exist.
+
+-------
+import std.range.primitives;
+import std.stdio;
+
+struct MyType
+{
+    void toString(W)(ref W writer) if (isOutputRange!(W, char))
+    {
+        put(writer, "Custom toString");
+    }
+}
+
+auto t = MyType();
+writeln(t); // writes "Custom toString"
+-------
+
+This has several benefits for the user. Firstly, this design is much friendlier
+to inlining than the `toString(scope void delegate(const(char)[]) sink)` method of
+`toString`. Secondly, this cuts down on memory usage, as characters are placed right
+into the output buffers of functions like $(REF format, std, format). Thirdly,
+because `toString` is now a template, can be made `@safe` via inference much more
+often.
+
+All previous forms of `toString` will continue to work.
--- a/changelog/toString.dd
+++ b/changelog/toString.dd
@@ -1,4 +1,4 @@
-toString Can Now Use Output Ranges
+`toString` Can Now Use Output Ranges
 
 The standard library has been modified to recognize and use `toString` overloads
 that accept $(REF_ALTTEXT output ranges, isOutputRange, std, range, primitives)
--- a/std/format.d
+++ b/std/format.d
@@ -1735,10 +1735,22 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
  * `toString` should have one of the following signatures:
  *
  * ---
- * const void toString(scope void delegate(const(char)[]) sink, FormatSpec fmt);
- * const void toString(scope void delegate(const(char)[]) sink, string fmt);
- * const void toString(scope void delegate(const(char)[]) sink);
- * const string toString();
+ * void toString(W)(ref W w, const ref FormatSpec fmt)
+ * void toString(W)(ref W w)
+ * string toString();
+ * ---
+ *
+ * Where `W` is an $(REF_ALTTEXT output range, isOutputRange, std,range,primitives)
+ * which accepts characters. The template type does not have to be called `W`.
+ *
+ * The following overloads are also accepted for legacy reasons or for use in virtual
+ * functions. It's recommended that any new code forgo these overloads if possible for
+ * speed and attribute acceptance reasons.
+ *
+ * ---
+ * void toString(scope void delegate(const(char)[]) sink, const ref FormatSpec fmt);
+ * void toString(scope void delegate(const(char)[]) sink, string fmt);
+ * void toString(scope void delegate(const(char)[]) sink);
  * ---
  *
  * For the class objects which have input range interface,
@@ -1756,7 +1768,7 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
  * Otherwise, are formatted just as their type name.
  *
  * Params:
- *     w = The $(REF_ALTTEXT output _range, isOutputRange, std,_range,primitives) to write to.
+ *     w = The $(REF_ALTTEXT output range, isOutputRange, std,_range,primitives) to write to.
  *     val = The value to write.
  *     f = The $(REF FormatSpec, std, format) defining how to write the value.
  */
@@ -1917,13 +1929,52 @@ void formatValue(Writer, T, Char)(auto ref Writer w, auto ref T val, const ref F
 }
 
 /**
- *  Formatting of a `struct` with a defined `toString`.
+ * Formatting a struct by defining a method `toString`, which takes an output
+ * range.
  *
- *  `formatValue` also allows to reuse existing format specifiers:
+ * It's recommended that any `toString` using $(REF_ALTTEXT output ranges, isOutputRange, std,range,primitives)
+ * use $(REF put, std,range,primitives) rather than use the `put` method of the range
+ * directly.
  */
 @safe unittest
 {
-   struct Point
+    import std.array : appender;
+    import std.range.primitives;
+
+    static struct Point
+    {
+        int x, y;
+
+        void toString(W)(ref W writer, const ref FormatSpec!char f)
+        if (isOutputRange!(W, char))
+        {
+            // std.range.primitives.put
+            put(writer, "(");
+            formatValue(writer, x, f);
+            put(writer, ",");
+            formatValue(writer, y, f);
+            put(writer, ")");
+        }
+    }
+
+    auto w = appender!string();
+    auto spec = singleSpec("%s");
+    auto p = Point(16, 11);
+
+    formatValue(w, p, spec);
+    assert(w.data == "(16,11)");
+}
+
+/**
+ * Another example of formatting a `struct` with a defined `toString`,
+ * this time using the `scope delegate` method.
+ *
+ * $(RED This method is now discouraged for non-virtual functions).
+ * If possible, please use the output range method instead.
+ */
+@safe unittest
+{
+   static struct Point
    {
        int x, y;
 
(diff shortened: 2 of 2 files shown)


JackStouffer:
Ho man, this is a fun bug. Turns out, if you don't define your `FormatSpec` as `const ref` in the parameters of your `toString` (and why would you when the docs don't tell you to do that), it wasn't recognized as a valid overload by `formatObject` before this change.

Case in point, BigInt's `toString(scope void delegate(const(char)[]) sink, ref FormatSpec!char f)` overload wasn't ever called by std.format. It called the string format option every time.

Fix incoming.

JackStouffer:
https://github.com/dlang/phobos/pull/5992

std/format.d
```diff
@@ -1735,10 +1735,22 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
...
- * const void toString(scope void delegate(const(char)[]) sink, string fmt);
- * const void toString(scope void delegate(const(char)[]) sink);
- * const string toString();
+ * void toString(W)(W w, FormatSpec fmt)
+ * void toString(W)(W w)
+ * string toString();
+ * ---
+ *
+ * Where `W` allows a character accepting
+ * $(REF_ALTTEXT ouput range, isOutputRange, std,_range,primitives) with its template constraints.
```
quickfur: Nitpick: ouput -> output

std/format.d
```diff
@@ -1735,10 +1735,22 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
...
- * const void toString(scope void delegate(const(char)[]) sink, FormatSpec fmt);
- * const void toString(scope void delegate(const(char)[]) sink, string fmt);
- * const void toString(scope void delegate(const(char)[]) sink);
- * const string toString();
+ * void toString(W)(W w, FormatSpec fmt)
+ * void toString(W)(W w)
+ * string toString();
+ * ---
+ *
+ * Where `W` allows a character accepting
```
quickfur: What about "Where W is an output range that accepts ...".

std/format.d
```diff
@@ -1735,10 +1735,22 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
...
+ * $(REF_ALTTEXT ouput range, isOutputRange, std,_range,primitives) with its template constraints.
+ * The template type does not have to be called `W`.
+ *
+ * The following overloads are also accepted for legacy reasons. It's recommended that
+ * any new code forgo these overloads for speed and attribute acceptance reasons.
```
quickfur: What about also accepting a `toString` overload that takes an output range that can accept any range of characters?  If efficiency is a concern here, it would make sense for `toString` to be able to take a, output range of char/wchar/dchar instead of needing a function call roundtrip per character.

quickfur:
In general I like the direction this is going.  The main concern I have is that we should support `toString` taking an output range that accepts ranges of characters, in order to minimize roundtrips when outputting strings.

std/format.d
jmdavis: Also, it really should be `ref W`. Accepting an output range by value is begging for trouble.

std/format.d
jmdavis: Is it really just for legacy reasons though? What about classes? As soon as you templatize `toString`, it can't be virtual.

quickfur: Hmm, that's true. Virtual `toString` methods could be problematic because `format` may not know how to construct an output range of a type that will be accepted by `toString`.

Perhaps one way out is use `std.range.interfaces.OutputRangeObject`. Since such an object is an output range, it can be recognized by `format`, and since it gives an opaque interface to arbitrary underlying output range implementations, it will also give us the flexibility to choose the best implementation used by `format`.

OTOH, since `scope void delegate(const(char)[])` is already established in current code, and already allows arbitrary implementations, perhaps we should just retain support for it.  Code that can afford to have `toString` templatized can still take advantage of a more flexible implementation of `format`.

quickfur: `@safe`ty is still an issue for virtual `toString` functions, though. Well, as well as other attributes.  I've no idea how to work around that.

JackStouffer:
Changed the requirements to look for a `string` accepting output range instead of a `char` accepting one. Also added checks to make sure the output range is being taken by ref.

As for virtualizing `toString`, IMO the best option would be to use the `string toString()` overload in that case, as it's probably going to be faster with the inlining issue I mentioned. Plus, it can also be `@safe` in that case.

JackStouffer:
https://github.com/dlang/phobos/pull/5995

JackStouffer:
Thought about the round-trip per character issue some more, and I remembered that `put` solves this problem automatically for the user. Emphasized it in the docs.

quickfur:
@JackStouffer It took me a few minutes to understand why and how exactly `std.range.put` solves the problem of round-trip per character.  You might want to put in a slightly more detailed explanation than just pointing the reader to `put`. :-) Also, it would be nice if there's a hyperlink to the documentation for `put` somewhere.

JackStouffer:
@quickfur I'm not planning on deprecating the other overloads anytime soon, if ever. There's just too much code out there that uses it, and this is such a common function, that I'm not comfortable removing the delegate overloads from `std.format`.

std/format.d
```diff
@@ -1735,10 +1735,21 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
...
+ * void toString(W)(ref W w, const ref FormatSpec fmt)
+ * void toString(W)(ref W w)
+ * string toString();
+ * ---
+ *
+ * Where `W` is an $(REF_ALTTEXT output range, isOutputRange, std,_range,primitives)
+ * which accepts characters. The template type does not have to be called `W`.
+ *
+ * The following overloads are also accepted for legacy reasons. It's recommended that
+ * any new code forgo these overloads for speed and attribute acceptance reasons.
```
quickfur: You might want to reword this, if the older overloads are still recommended for class hierarchies where `toString` needs to be virtual.

JackStouffer:
@quickfur Addressed comments

std/format.d
```diff
@@ -3484,11 +3553,59 @@ template hasToString(T, Char)
...
+    {
+        val.toString(w, f);
+    }
+    else static if (overload == 5)
+    {
+        val.toString(w);
+    }
+    // not using the overload enum to not break badly defined toString overloads
+    // e.g. defining the FormatSpec as ref and not const ref led this function
+    // to ignore that toString overload
```
quickfur: Just a thought: to prevent similar breakages in the future, what about if the new unittest added above is changed to have non-trivial return values for the `toString` overloads, and the unittest actually checks for these values?  The idea being that if a particular `toString` overload is ignored or not called correctly for whatever reason, the unittest would catch it.

std/format.d
```diff
@@ -1917,13 +1929,53 @@ void formatValue(Writer, T, Char)(auto ref Writer w, auto ref T val, const ref F
...
+{
+    import std.array : appender;
+
+    static struct Point
+    {
+        int x, y;
+
+        void toString(W)(ref W writer, const ref FormatSpec!char f) if (isOutputRange!(W, char))
+        {
+            // make sure to use std.range.primitives.put and not writer.put directly
```
wilzbach: Mentioning this in the summary page.

std/format.d
```diff
@@ -1917,13 +1929,53 @@ void formatValue(Writer, T, Char)(auto ref Writer w, auto ref T val, const ref F
...
+    formatValue(w, p, spec);
+    assert(w.data == "(16,11)");
+}
+
+/**
+ * Another example of formatting a `struct` with a defined `toString`,
+ * this time using the `scope delegate` method.
  *
- *  `formatValue` also allows to reuse existing format specifiers:
+ * This method is now discouraged. Please use the output range method
```
wilzbach: `$(RED ..)` the first sentence. Or maybe just bold?

std/format.d
```diff
@@ -3455,13 +3507,30 @@ if (is(AssocArrayTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
...
     static if (isPointer!T && !isAggregateType!T)
     {
         // X* does not have toString, even if X is aggregate type has toString.
         enum hasToString = 0;
     }
+    else static if (is(typeof(
+        {T val = void; FormatSpec!Char f; static struct S {void put(Char s){}}
+        S s; val.toString(s, f);
+        // force toString to take output range by ref
+        static assert(ParameterStorageClassTuple!(T.toString!(S))[0] == ParameterStorageClass.ref_);}
```
wilzbach: How about enforcing that the second parameter (FormatSpec) is const and ref?

std/format.d
```diff
@@ -1735,10 +1735,22 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
  * `toString` should have one of the following signatures:
  *
  * ---
- * const void toString(scope void delegate(const(char)[]) sink, FormatSpec fmt);
- * const void toString(scope void delegate(const(char)[]) sink, string fmt);
- * const void toString(scope void delegate(const(char)[]) sink);
- * const string toString();
+ * void toString(W)(ref W w, const ref FormatSpec fmt)
+ * void toString(W)(ref W w)
```
wilzbach: You don't have a public unittest showing this to the user. I know it's trivial, but I still would add one to avoid guaranteed questions on the NG.

std/format.d
JackStouffer: If it's ref compilation will already fail if it's not const

std/format.d
```diff
@@ -1917,13 +1929,50 @@ void formatValue(Writer, T, Char)(auto ref Writer w, auto ref T val, const ref F
...
+        }
+    }
+
+    auto w = appender!string();
+    auto spec = singleSpec("%s");
+    auto p = Point(16, 11);
+
+    formatValue(w, p, spec);
+    assert(w.data == "(16,11)");
+}
```
wilzbach: This public tests is failing, run with:

```
make -f posix.mak std/format.publictests
```

wilzbach: The line number on the error message is the one from the generated file.
(it probably makes sense to insert empty lines in it, s.t. the line numbers match up)

JackStouffer: I'm not really sure how to debug this. This is green on the autotester

wilzbach: 1) Compile Phobos
2) Run the extracted test with Phobos-dev:

```
unittest
{
    import std.array : appender;
    import std.format;

    static struct Point
    {
        int x, y;

        void toString(W)(ref W writer, const ref FormatSpec!char f) if (isOutputRange!(W, char))
        {
            // std.range.primitives.put
            put(writer, "(");
            formatValue(writer, x, f);
            put(writer, ",");
            formatValue(writer, y, f);
            put(writer, ")");
        }
    }

    auto w = appender!string();
    auto spec = singleSpec("%s");
    auto p = Point(16, 11);

    formatValue(w, p, spec);
    assert(w.data == "(16,11)");
}
```

BTW seeing that we just catched this issue, you might want to add more public tests (or tests in a different module).

wilzbach: The issue is that `w.data` is `Point(16, 11)` here.

std/format.d
```diff
@@ -1917,13 +1929,50 @@ void formatValue(Writer, T, Char)(auto ref Writer w, auto ref T val, const ref F
...
+ */
+@safe unittest
+{
+    import std.array : appender;
+
+    static struct Point
+    {
+        int x, y;
+
+        void toString(W)(ref W writer, const ref FormatSpec!char f) if (isOutputRange!(W, char))
```
wilzbach: If constraints should be in the line of their declaration:

```
void toString(W)(ref W writer, const ref FormatSpec!char f)
if (isOutputRange!(W, char))
```

JackStouffer:
@wilzbach Fixed. The public tests didn't have access to `std.range.primitives`.

JackStouffer:
Ping @andralex

changelog/toString.dd
```diff
@@ -0,0 +1,30 @@
...
+    {
+        put(writer, "Custom toString");
+    }
+}
+
+auto t = MyType();
+writeln(t); // writes "Custom toString"
+-------
+
+This has several benefits for the user. Firstly, this design is much friendlier
```
schveiguy: Firstly -> First, Secondly -> Second...

schveiguy:
This is cool, I like it.

CyberShadow:
This pull request seems to have exposed an `-allinst` DMD bug, thus causing a regression:
https://issues.dlang.org/show_bug.cgi?id=19406

PR #5995, opened 2018-01-04
Changed all instances of FormatSpec in toString to const ref

JackStouffer:
See https://github.com/dlang/phobos/pull/5991#issuecomment-355122818

While in this instance, the code compiled just fine, the problem is that a `const ref` parameter in `formatObject` ended up being copied in said function if `FormatSpec` wasn't marked as `ref`. This poses a problem because `FormatSpec` is quite a large struct, not to mention the fact that we've already gone through the trouble of making a pointer to one, so making a copy is just extra work.

### DIFF
--- a/std/bitmanip.d
+++ b/std/bitmanip.d
@@ -2231,8 +2231,7 @@ public:
      * $(LI $(B %b) which prints the bits as 8-bit byte packets)
      * separated with an underscore.
      */
-    void toString(scope void delegate(const(char)[]) sink,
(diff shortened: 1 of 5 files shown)


JackStouffer:
This is technically a breaking change mind you. However, it's very rare that a user would call these methods directly and not just use `to` or `format`.

andralex:
The size of FormatSpec is 72 bytes making the argument difficult. Might be better to just compress it to 64 bytes...

I'll accept this anyway.

PR #5996, opened 2018-01-04
Removed debug printfs from std.bitmanip

JackStouffer:


wilzbach:
CircleCI doesn't like this:

```
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2382): Error: undefined identifier `writefln`
std/bitmanip.d(2509): Error: undefined identifier `writefln`
std/bitmanip.d(2509): Error: undefined identifier `writefln`
std/bitmanip.d(2509): Error: undefined identifier `writefln`
std/bitmanip.d(2509): Error: undefined identifier `writefln`
std/bitmanip.d(2509): Error: undefined identifier `writefln`
std/bitmanip.d(2509): Error: undefined identifier `writefln`
std/bitmanip.d(2509): Error: undefined identifier `writefln`
std/bitmanip.d(2509): Error: undefined identifier `writefln`
```

There are still a couple of `scope(failure) writefln("Failed type: %s", T.stringof);` blocks in the codebase.

JackStouffer:
@wilzbach done

std/bitmanip.d
```diff
@@ -2722,7 +2669,6 @@ if (isFloatOrDouble!T)
                          char, wchar, dchar/*,
                          float, double*/))
     {{
-        scope(failure) writefln("Failed type: %s", T.stringof);
```
wilzbach: On a second thought I'm not so sure about removing these `scope(failure)` blocks. This looks like it might be useful to debug failures on DAutoTest given that D's `assert` is quite dumb.
What do you or other people think? Should we leave this in and do the following at the top / within the test?

```d
version(StdUnittest) import std.stdio : writefln;
```

JackStouffer: I'm not a fan of code bases like DMD that leave in lots of residual `print`s that add clutter but might one day be useful to debug the code. Better to either use a debugger, add in the `writeln`s yourself when you need them, or fix `assert` to give better diagnostics.

andralex: Feel free to leave those in on a case basis. First thing I'd do if I had to debug that would be to insert such stuff. I don't think that versioned import is a good idea.

andralex:
@JackStouffer this stuff is at best low-impact. We don't use this tracing style often but actively removing it seems like time spent best elsewhere.

JackStouffer:
I know, it's just a trend that I've been removing them whenever I see them on code I'm working on. Gotten rid of them in about five other modules already.

andralex:
>I know, it's just a trend that I've been removing them whenever I see them on code I'm working on. Gotten rid of them in about five other modules already.

So that means they were somewhat popular and you're pushing the way you prefer. It did happen that other contributors spent time marking their territory with their favorite way of doing things. Not necessarily a bad thing but hardly something that moves the needle, and it's just churn. I suggest you leave this be and focus on something else.

andralex:
I'll let you close as I noticed if I close something it causes a riot.

JackStouffer:
Ok, I'm not particularly attached to this

>I'll let you close as I noticed if I close something it causes a riot.

lol

wilzbach:
Oh wow that went quickly of the road. I though that we all agree that versioned global imports are _bad_?

> I don't think that versioned import is a good idea.

`std.bitmanip` does the following on top:

```d
version(unittest)
{	
    import std.stdio;	
}
```

For my lazy import story, I have to get rid of as many of these ugly global imports anyways (ideally all of them!). Note that this is why I suggested `version(StdUnittest)` as the import then won't be executed in user code.

@JackStouffer do you mind restoring your branch and reopen this (with the `writeln` imports moved into the `unittest`?

PR #5999, opened 2018-01-05
[trivial] fix ddoc typo

BBasile:


JackStouffer:
It's not a typo https://en.wikipedia.org/wiki/If_and_only_if

But, most people don't know about this usage, so it should be changed to directly say "if and only if".

BBasile:
yes, i don't know if it's a good idea to use this in the documentation. In the meantime let's release the pressure on the autotester resources.

PR #6000, opened 2018-01-05
Slightly simplify std.internal.cstring realloc logic

kinke:


std/internal/cstring.d
```diff
@@ -111,26 +111,25 @@ if (isSomeChar!To && (isInputRange!From || isSomeString!From) &&
         auto r = cast(const(CF)[])str;  // because inout(CF) causes problems with byUTF
         if (r is null)  // Bugzilla 14980
         {
+            res._length = 0;
```
kinke: This is a fix; _length was previously undefined due to void-initialization.

std/internal/cstring.d
```diff
@@ -111,26 +111,25 @@ if (isSomeChar!To && (isInputRange!From || isSomeString!From) &&
...
-            q = p;
         }
-        q[i++] = c;
+        p[i++] = c;
     }
-    q[i] = 0;
+    p[i] = 0;
     res._length = i;
-    res._ptr = p_is_onstack ? res.useStack : &p[0];
+    res._ptr = p_is_onstack ? res.useStack : p.ptr;
```
wilzbach: You can't do that:

```
std/file.d(2338): Error: @safe function 'std.file.mkdirRecurse' cannot call @system function 'std.file.exists!(const(char)[]).exists'
std/file.d(2344): Error: @safe function 'std.file.mkdirRecurse' cannot call @system function 'std.file.ensureDirExists!().ensureDirExists'
std/file.d(3099): Error: @safe constructor 'std.file.DirEntry.this' cannot call @system function 'std.file.exists!string.exists'
std/file.d(2338): Error: @safe function 'std.file.mkdirRecurse' cannot call @system function 'std.file.exists!(const(char)[]).exists'
std/file.d(2344): Error: @safe function 'std.file.mkdirRecurse' cannot call @system function 'std.file.ensureDirExists!().ensureDirExists'
std/file.d(3099): Error: @safe constructor 'std.file.DirEntry.this' cannot call @system function 'std.file.exists!string.exists'
std/stdio.d(423): Error: @safe constructor 'std.stdio.File.this' cannot call @system function 'std.stdio.fopen!(string, const(char)[]).fopen'
std/stdio.d(423): Error: @safe constructor 'std.stdio.File.this' cannot call @system function 'std.stdio.fopen!(string, const(char)[]).fopen'
std/experimental/logger/filelogger.d(76): Error: @safe constructor 'std.experimental.logger.filelogger.FileLogger.this' cannot call @system function 'std.file.exists!string.exists'
std/experimental/logger/filelogger.d(76): Error: @safe constructor 'std.experimental.logger.filelogger.FileLogger.this' cannot call @system function 'std.file.exists!string.exists'
```

std/internal/cstring.d
```diff
@@ -234,31 +233,30 @@ private:
...
+    if (buf_is_onstack)
     {
-        size_t newlen = res.length * 3 / 2;
         if (newlen <= strLength)
             newlen = strLength + 1; // +1 for terminating 0
         auto ptr = cast(To*) pureMalloc(newlen * To.sizeof);
         if (!ptr)
             onOutOfMemoryError();
-        memcpy(ptr, res.ptr, i * To.sizeof);
+        ptr[0 .. buf.length] = buf[];
```
kinke: [This will copy 1 more element than the previous code.]

std/internal/cstring.d
kinke: Yep, plus a second `To[]` variable (now `heapBuffer` instead of `q`) is required for safe-ty.

std/internal/cstring.d
```diff
@@ -234,31 +233,30 @@ private:
     static TempCStringBuffer trustedVoidInit() { TempCStringBuffer res = void; return res; }
 }
 
-private To[] trustedRealloc(To)(To[] buf, size_t i, To[] res, size_t strLength, bool res_is_onstack)
+private To[] trustedRealloc(To)(To[] buf, size_t strLength, bool buf_is_onstack)
```
wilzbach: If you already change this, I think we should stick to the [DStyle](https://dlang.org/dstyle.html), e.g. sth. like `bufIsOnStack`

std/internal/cstring.d
wilzbach: Don't you 1 element _less_? `i` is `buf.length + 1`
See: https://run.dlang.io/is/DJicoO

std/internal/cstring.d
wilzbach: Are there even cases where `res` won't be `null` from `trustedVoidInit`?
See also: https://issues.dlang.org/show_bug.cgi?id=14980

But we can try this in another PR ;-)

std/internal/cstring.d
kinke: Nope, that'd be a serious problem ;). `i` was `buf.length - 1`; the realloc already happens when about to fill the last element in the old buffer (due to the additionally required null-terminator).

std/internal/cstring.d
kinke: `res` is the void-initialized struct, so all of its members are undefined. Just setting the ptr to null but leaving the length undefined doesn't sound too good.

std/internal/cstring.d
kinke: Will do when at home.

kinke:
6000, nice indeed. :) - Nit addressed.

PR #4692, opened 2016-07-29
std.typecons: make RefCounted @safe

wilzbach:
follow-up to #4685

PR #6001, opened 2018-01-05
Fix Issue 18163 - std.container.array should be usable in @safe

n8sh:
I think that standard library Array should be just as `@safe` as built-in dynamic arrays. Thanks to @wilzbach for opening this issue. This patch requires marking some code `@trusted` but in each case I believe it is defensible. Some changes were required in RefCounted because it is used in Array.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -5495,10 +5495,16 @@ if (!is(T == class) && !(is(T == interface)))
             import core.exception : onOutOfMemoryError;
             import std.conv : emplace;
 
+            // @trusted for the cast from void* to Impl*. Could only be unsafe
+            // if _store were dereferenced before being initialized, which
+            // cannot happen since we next call emplace.
             _store = (() @trusted => cast(Impl*) pureMalloc(Impl.sizeof))();
             if (_store is null)
                 onOutOfMemoryError();
             static if (hasIndirections!T)
+                // @trusted for the pureGcAddRange. Okay because this cannot
+                // cannot conceivably cause memory corruption: pointer is
+                // non-null and size is correct.
                 (() @trusted => pureGcAddRange(&_store._payload, T.sizeof))();
             emplace(&_store._payload, args);
             _store._count = 1;
@@ -5509,10 +5515,17 @@ if (!is(T == class) && !(is(T == interface)))
             import core.exception : onOutOfMemoryError;
             import core.stdc.string : memcpy, memset;
 
+            // @trusted for the cast from void* to Impl*. Could only be unsafe
+            // if _store were dereferenced before being initialized, which
+            // cannot happen since we next either initialize with memcpy or
+            // with a blit.
             _store = (() @trusted => cast(Impl*) pureMalloc(Impl.sizeof))();
             if (_store is null)
                 onOutOfMemoryError();
             static if (hasIndirections!T)
+                // @trusted for the pureGcAddRange. Okay because this cannot
+                // cannot conceivably cause memory corruption: pointer is
+                // non-null and size is correct.
                 (() @trusted => pureGcAddRange(&_store._payload, T.sizeof))();
 
             // Can't use std.algorithm.move(source, _store._payload)
@@ -5521,6 +5534,11 @@ if (!is(T == class) && !(is(T == interface)))
 
             // Can avoid destructing result.
             static if (hasElaborateAssign!T || !isAssignable!T)
+                // @trusted for the memcpy. Okay because this cannot cannot
+                // cause memory corruption barring unsafe previous operations.
+                // By specification it is not safe for any structure to contain
+                // a pointer to itself, so the change in address is allowed;
+                // _store is non-null; and _store has the correct size.
                 (() @trusted => memcpy(&_store._payload, &source, T.sizeof))();
             else
                 _store._payload = source;
@@ -5537,8 +5555,12 @@ if (!is(T == class) && !(is(T == interface)))
 
                 auto init = typeid(T).initializer();
                 if (init.ptr is null) // null ptr means initialize to 0s
+                    // @trusted for memset. Address is valid and size is valid.
                     (() @trusted => memset(&source, 0, sz))();
                 else
+                    // @trusted for memcpy. Destination is non-null, source is
+                    // valid if there has not already been memory corruption,
+                    // and size does not overrun.
                     (() @trusted => memcpy(&source, init.ptr, sz))();
             }
 
@@ -5626,6 +5648,11 @@ to deallocate the corresponding resource.
         .destroy(_refCounted._store._payload);
         static if (hasIndirections!T)
         {
+            // @trusted for pureGcRemoveRange. Cannot possibly cause memory
+            // corruption unless the payload has pointers to garbage-collected
+            // memory which is referred to by pointers elsewhere not in a
+            // location known to the garbage collector, and there is no safe
+            // way to obtain such pointers.
             (() @trusted => pureGcRemoveRange(&_refCounted._store._payload))();
         }
 
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -5495,11 +5495,11 @@ if (!is(T == class) && !(is(T == interface)))
             import core.exception : onOutOfMemoryError;
             import std.conv : emplace;
 
-            _store = cast(Impl*) pureMalloc(Impl.sizeof);
+            _store = (() @trusted => cast(Impl*) pureMalloc(Impl.sizeof))();
             if (_store is null)
                 onOutOfMemoryError();
             static if (hasIndirections!T)
-                pureGcAddRange(&_store._payload, T.sizeof);
+                (() @trusted => pureGcAddRange(&_store._payload, T.sizeof))();
             emplace(&_store._payload, args);
             _store._count = 1;
         }
@@ -5509,11 +5509,11 @@ if (!is(T == class) && !(is(T == interface)))
             import core.exception : onOutOfMemoryError;
             import core.stdc.string : memcpy, memset;
 
-            _store = cast(Impl*) pureMalloc(Impl.sizeof);
+            _store = (() @trusted => cast(Impl*) pureMalloc(Impl.sizeof))();
             if (_store is null)
                 onOutOfMemoryError();
             static if (hasIndirections!T)
-                pureGcAddRange(&_store._payload, T.sizeof);
+                (() @trusted => pureGcAddRange(&_store._payload, T.sizeof))();
 
             // Can't use std.algorithm.move(source, _store._payload)
             // here because it requires the target to be initialized.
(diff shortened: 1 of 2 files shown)


JackStouffer:
Marking RefCounted as trusted is very unsafe. See http://forum.dlang.org/post/n0nnu0$1tth$1@digitalmars.com

n8sh:
> Marking RefCounted as trusted is very unsafe.
Thanks for the link @JackStouffer. I'm not making the entirety of RefCounted trusted, just certain things in `initialize`, `move`,  `~this`, and `opAssign`. I think these changes are sound, but I'll go through them one-by-one.

std/typecons.d
```diff
@@ -5495,11 +5495,11 @@ if (!is(T == class) && !(is(T == interface)))
             import core.exception : onOutOfMemoryError;
             import std.conv : emplace;
 
-            _store = cast(Impl*) pureMalloc(Impl.sizeof);
+            _store = (() @trusted => cast(Impl*) pureMalloc(Impl.sizeof))();
```
n8sh: `@trusted` for the cast from void* to Impl*. Could only be unsafe if _store were dereferenced before being initialized, which cannot happen since we next call emplace.

std/typecons.d
```diff
@@ -5495,11 +5495,11 @@ if (!is(T == class) && !(is(T == interface)))
...
             if (_store is null)
                 onOutOfMemoryError();
             static if (hasIndirections!T)
-                pureGcAddRange(&_store._payload, T.sizeof);
+                (() @trusted => pureGcAddRange(&_store._payload, T.sizeof))();
```
n8sh: `@trusted` for the pureGcAddRange. Okay because this cannot cannot conceivably cause memory corruption.

std/typecons.d
```diff
@@ -5509,19 +5509,19 @@ if (!is(T == class) && !(is(T == interface)))
             import core.exception : onOutOfMemoryError;
             import core.stdc.string : memcpy, memset;
 
-            _store = cast(Impl*) pureMalloc(Impl.sizeof);
+            _store = (() @trusted => cast(Impl*) pureMalloc(Impl.sizeof))();
```
n8sh: `@trusted` for the cast from void* to Impl*. Could only be unsafe if _store were dereferenced before being initialized, which cannot happen since we next either initialize with memcpy or with a blit.

std/typecons.d
```diff
@@ -5509,19 +5509,19 @@ if (!is(T == class) && !(is(T == interface)))
...
             if (_store is null)
                 onOutOfMemoryError();
             static if (hasIndirections!T)
-                pureGcAddRange(&_store._payload, T.sizeof);
+                (() @trusted => pureGcAddRange(&_store._payload, T.sizeof))();
```
n8sh: `@trusted` for the pureGcAddRange. Okay because this cannot cannot conceivably cause memory corruption (pointer is non-null and size is correct).

std/typecons.d
```diff
@@ -5509,19 +5509,19 @@ if (!is(T == class) && !(is(T == interface)))
...
 
             // Can't use std.algorithm.move(source, _store._payload)
             // here because it requires the target to be initialized.
             // Might be worth to add this as `moveEmplace`
 
             // Can avoid destructing result.
             static if (hasElaborateAssign!T || !isAssignable!T)
-                memcpy(&_store._payload, &source, T.sizeof);
+                (() @trusted => memcpy(&_store._payload, &source, T.sizeof))();
```
n8sh: `@trusted` for the memcpy. Okay because this cannot cannot cause memory corruption barring unsafe previous operations. By specification it is not `safe` for any structure to contain a pointer to itself, so the change in address is allowed; _store is non-null; and _store has the correct size.

std/typecons.d
```diff
@@ -5537,9 +5537,9 @@ if (!is(T == class) && !(is(T == interface)))
 
                 auto init = typeid(T).initializer();
                 if (init.ptr is null) // null ptr means initialize to 0s
-                    memset(&source, 0, sz);
+                    (() @trusted => memset(&source, 0, sz))();
```
n8sh: `@trusted` for memset. Address is valid and size is valid.

std/typecons.d
```diff
@@ -5537,9 +5537,9 @@ if (!is(T == class) && !(is(T == interface)))
...
                 else
-                    memcpy(&source, init.ptr, sz);
+                    (() @trusted => memcpy(&source, init.ptr, sz))();
```
n8sh: `@trusted` for memcpy. Destination is non-null, source is valid if there has not already been memory corruption, and size does not overrun.

std/typecons.d
```diff
@@ -5626,17 +5626,17 @@ to deallocate the corresponding resource.
         .destroy(_refCounted._store._payload);
         static if (hasIndirections!T)
         {
-            pureGcRemoveRange(&_refCounted._store._payload);
+            (() @trusted => pureGcRemoveRange(&_refCounted._store._payload))();
```
n8sh: `@trusted` for pureGcRemoveRange. Cannot possibly cause memory corruption unless the payload has pointers to garbage-collected memory which is referred to by pointers elsewhere not in a location known to the garbage collector, and there is no `safe` way to obtain such pointers.

quickfur:
I would feel slightly less uncomfortable at the number of `@trusted` escapes introduced, if there was at least one unittest that ensures that instantiating `RefCounted` with an unsafe payload actually causes the compiler to infer `@system` for the resulting type / method calls. I.e., something along the lines of:
````
@system unittest
{
    struct UnsafePayload { ... /* do something unsafe here */ }
    auto x = RefCounted!UnsafePayload(...);
    void canary(T)(T t) @safe
    {
        ... /* call various RefCounted methods here */
    }
    static assert(!__traits(compiles, canary(x));
}
````
The point being, ensure that the new `@trusted` additions aren't introducing unintended safety loopholes for payloads that are obviously unsafe.

File: std/typecons.d [hunk 9]
```diff
@@ -5626,17 +5626,17 @@ to deallocate the corresponding resource.
...
         }
 
-        pureFree(_refCounted._store);
+        (() @trusted => pureFree(_refCounted._store))();
```
n8sh: `@trusted` for pureFree. The pointer is known not to be null due to isInitialized. Is this a bit that could cause problems if an "alias" to the stored object is captured elsewhere as in http://forum.dlang.org/post/n0nnu0$1tth$1@digitalmars.com?

std/typecons.d
```diff
@@ -5626,17 +5626,17 @@ to deallocate the corresponding resource.
...
         _refCounted._store = null;
     }
 
 /**
 Assignment operators
  */
-    void opAssign(typeof(this) rhs)
+    void opAssign(typeof(this) rhs) @trusted
     {
```
n8sh: `@trusted` for `std.algorithm.swap(_refCounted._store, rhs._refCounted._store)`. I believe this is `safe` due to the prohibition against structs holding pointers to themselves and due to the lack of any `safe` way for anyone else to have the address of either _store (meaning there is no need to account for this).

std/typecons.d
```diff
@@ -5626,17 +5626,17 @@ to deallocate the corresponding resource.
...
-        pureFree(_refCounted._store);
+        (() @trusted => pureFree(_refCounted._store))();
         _refCounted._store = null;
     }
 
 /**
 Assignment operators
  */
-    void opAssign(typeof(this) rhs)
+    void opAssign(typeof(this) rhs) @trusted
```
n8sh: `@trusted` for `std.algorithm.swap(_refCounted._store, rhs._refCounted._store)`. I believe this is `safe` due to the prohibition against structs holding pointers to themselves and due to the lack of any `safe` way for anyone else to have the address of either _store (meaning there is no need to account for this).

n8sh:
>The point being, ensure that the new @trusted additions aren't introducing unintended safety loopholes for payloads that are obviously unsafe.

Okay, will add a test in a few hours.

n8sh:
BTW, I acknowledge that there are a lot of `@trusted` escapes introduced, but I think the standard library is the correct place to bear this pain. Not putting them here where they can be checked once for correctness means that instead every single user of `std.container.Array` must clutter his own code with similar escapes -- possibly making a mistake -- or eschew safety altogether.

schveiguy:
> ensure that the new `@trusted` additions aren't introducing unintended safety loopholes for payloads that are obviously unsafe

And, even if they aren't (and I don't think they are), make sure future improvements don't!

wilzbach:
@n8sh thanks a lot for working on this. Just an FYI @edi33416 is working on a new containers library for Phobos (though it still might take some time ...).

> Marking RefCounted as trusted is very unsafe. See 

See also: https://github.com/dlang/phobos/pull/4692

std/typecons.d
schveiguy: Yeah, this is where it breaks down. Because you can return a raw reference to the payload, you could potentially have `RefCounted` free the memory from underneath you.

std/typecons.d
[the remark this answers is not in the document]
n8sh: Oddly it does. Just verified.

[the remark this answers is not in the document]
schveiguy: Weird. I tested locally, and swap with 2 pointers works in `@safe`. I would have expected this to infer that.

In any case, this should be marked `@safe`, not `@trusted` I would think.

std/typecons.d
n8sh: If the change were made in Array instead, the RefCounted payload is private to Array so a reference to the payload itself couldn't escape, but we're still left with the problem of an escaping reference to one of the items contained in the Array whose space will be freed when the reference count hits zero. Is there a way forward that doesn't rely on [DIP-1000](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md)?

n8sh: One approach that comes to mind is making Array's `opIndex` return by value instead of by reference. To avoid breaking existing code or giving unpleasant surprises we'd have to limit this to types that are reasonably small and don't have a complex destructor and don't have a complex copy constructor. (Of course for those types there would need to be an `opIndexAssign`.) Then we know that no references to the to-be-freed memory could have escaped without unsafe shenanigans.

schveiguy: If there was a way to say "in safe mode, only return values, not references", that would be ideal. I'm unaware if there is such a thing.

quickfur: One thought that occurs to me is to use `std.traits.isSafe` with a suitably-crafted lambda / function in a template sig constraint, and thereby alter the function's refness depending on where it's called from.  Unfortunately, since we can't overload on `@safe`/`@system`, I'm not sure if this is achievable.

n8sh: What I am thinking of now is adding a `Flag!"returnByValue"` parameter to the Array template that defaults to `!hasElaborateDestructor!T && !hasElaborateCopyConstructor!T && T.sizeof <= (uint[]).sizeof` but can be manually specified if desired.

std/typecons.d
wilzbach: These comments should be put in the source code, s.t. they are easy to catch by future readers ;-)

n8sh:
I am closing this pull request because I currently don't see a way to deal with the lifetime problems associated with return by reference (see above comments).

PR #6003, opened 2018-01-05
add a helper in allocator to prevent the usage of the `new` operator

BBasile:
This could work for `delete` too but actually there's a bug: https://issues.dlang.org/show_bug.cgi?id=18198.
Since `delete` is likely to be deprecated it's acceptable not to handle it here.

### DIFF
--- a/std/experimental/allocator/package.d
+++ b/std/experimental/allocator/package.d
@@ -31,6 +31,9 @@ $(TR $(TD Class interface) $(TD
     $(LREF CAllocatorImpl)
     $(LREF IAllocator)
 ))
+$(TR $(TD Other) $(TD
+    $(LREF disableNew)
+))
 )
 
 Synopsis:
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/package.d
```diff
@@ -3113,3 +3139,4 @@ unittest
     b = alloc2.allocate(101);
     assert(alloc2.deallocate(b));
 }
+
```
wilzbach: Unrelated?

wilzbach:
(rebased to `upstream/master` to fix the CI failures)

std/experimental/allocator/package.d
BBasile: Yes, things were initially at the end, until i realize that because of the `__EOF__` the test were not executed and the ddoc not rendered.

wilzbach:
After talking with @andralex shortly about it, here's his feedback:

- it can be reduced to `@disable new (size_t);` (no implementation needed)
- Phobos isn't about including one-lines `mixin(disableNew);` is exactly as long and still requires the import
- `@disable new (size_t);` is part of the language, `disableNew` will require people to lock it up in the docs.
- this lacks good use cases

-> I will close this now as our PR queue is exploding. If you or someone else comes up with a better motivation, please resubmit (or ping us for reopening).

PR #6004, opened 2018-01-07
Fixed formatValue for string with multibyte chars

biozic:
Strings containing multibyte chars wouldn't be formatted with the correct width. Using walkLength instead of length fixes it.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -3043,16 +3043,16 @@ if (isInputRange!T)
                 if (!f.flDash)
                 {
                     // right align
-                    if (f.width > s.length)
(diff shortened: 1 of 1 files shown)


JackStouffer:
This needs a unit test illustrating the issue as well as a corresponding issue in bugzilla. Once you have the issue number, please use it in the commit message like so

```
Fix Issue xxxx: Fixed formatValue for string with multibyte chars
```

biozic:
Yes, and it obviously needs further inversigation ! I'll file a simple bug issud.

PR #6008, opened 2018-01-07
fix issue 7054 - format() aligns using code units instead of graphemes

BBasile:
I didn't see it in bugzilla before cooking the fix but apparently it's controversial for some people.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -3040,19 +3040,45 @@ if (isInputRange!T)
             static if (is(StringTypeOf!T))
             {
                 auto s = val[0 .. f.precision < $ ? f.precision : $];
+
+                size_t width;
+                if (f.width > 0)
+                {
+                    // strings that are fully made of ASCII characters
+                    // can be aligned w/o graphemeStride
+                    bool onlyAscii = true;
+                    for (size_t i; i < s.length; i++)
+                    {
+                        if (s[i] > 0x7F)
+                        {
+                            onlyAscii = false;
+                            break;
+                        }
+                    }
+                    if (!onlyAscii)
+                    {
+                        //TODO: optimize this
+                        import std.uni : graphemeStride;
+                        for (size_t i; i < s.length; i += graphemeStride(s, i))
+                            width++;
+                    }
+                    else width = s.length;
+                }
+                else width = s.length;
+
                 if (!f.flDash)
                 {
                     // right align
-                    if (f.width > s.length)
-                        foreach (i ; 0 .. f.width - s.length) put(w, ' ');
+                    if (f.width > width)
+                        foreach (i ; 0 .. f.width - width) put(w, ' ');
                     put(w, s);
                 }
                 else
                 {
                     // left align
                     put(w, s);
-                    if (f.width > s.length)
-                        foreach (i ; 0 .. f.width - s.length) put(w, ' ');
+                    if (f.width > width)
+                        foreach (i ; 0 .. f.width - width) put(w, ' ');
                 }
             }
             else
@@ -5739,7 +5765,19 @@ private bool needToSwapEndianess(Char)(const ref FormatSpec!Char f)
     assert(r == ">            15<, [1, 2, 3]");
 
     assert(format("%8s", "bar") == "     bar");
-    assert(format("%8s", "b\u00e9ll\u00f4") == " b\u00e9ll\u00f4");
+    assert(format("%8s", "b\u00e9ll\u00f4") == "   b\u00e9ll\u00f4");
+}
+
+@safe pure unittest // bugzilla 18205
+{
+    assert("|%8s|".format("abc")        == "|     abc|");
+    assert("|%8s|".format("αβγ")        == "|     αβγ|");
+    assert("|%8s|".format("   ")        == "|        |");
+    assert("|%8s|".format("été"d)       == "|     été|");
+    assert("|%8s|".format("été 2018"w)  == "|été 2018|");
+
+    assert("%2s".format("e\u0301"w) == " e\u0301");
+    assert("%2s".format("a\u0310\u0337"d) == " a\u0310\u0337");
 }
 
 @safe unittest


biozic:
Yes, having a formatted string that is longer than what the width specification specifies for some inputs is probably controversial. The workaround is to transcode the input to `dstring`. That's why I closed my bugzilla issue.

BBasile:
To me it's more logic to get a well aligned output, whatever is the encoding. What do the foreign standard libraries do ?

biozic:
Python 3:
```python
print('|{:>8}|'.format('abc'))
print('|{:>8}|'.format('αβγ'))
```
prints
```
|     abc|
|     αβγ|
```

BBasile:
[Go too](https://play.golang.org/p/xnvXrubWsCl) so your report was valid after all. I don't say that we should align on the others but `format()` is one quite a universal function.

JackStouffer:
This doesn't really fix the problem:

```d
auto a = "\u0065\u0301";
writeln(a); // é
assert(a.walklength == 2);
```

Operating on the code point level rather than the code unit level just opens up different different issues.

If you want the same behavior as Python, you need to use `byGrapheme.walkLength`, but it's much slower. I think there is a case to be made here and would be much more sensible to the user in a string formatting function.

There's also the issue of backwards incompatibility.

JackStouffer:
I would changes this to use `byGrapheme` and solicit opinions from other Phobos maintainers.

biozic:
Yes `byGrapheme.walkLength` is more to the point, since the problem was that the number of graphemes was inconsistent.

If performance is an issue, we could add a specific flag, say `!` in the format-specification grammar, to tell that the width specification is not a number of code units, but a number of graphemes, e.g. `%8s` for the former, `%!8s` for the latter.

BBasile:
`byGrapheme.walkLength` is not CTFE-able but this seems to be a requirement for `format()`.

BBasile:
> Operating on the code point level rather than the code unit level just opens up different issues.

_different_ issues ? It seems that what would happen is that in the current state, the fix would be incomplete, e.g only working when `walkLength == byGrapheme.walkLength`.

> There's also the issue of backwards incompatibility.

Likely only some ugly workarounds, which then would not be necessary anymore.
The most simple workaround that people could used (calling  `spec.format(stuff.to!dstring)`) would still work.

JackStouffer:
It brings up different issues because it's slower and it breaks from an easy to understand but simplistic rule of string width == length, while also not changing the output if the string was made with or normalized with decomposing rules. With either code points or code units, the behavior is not intuitive but this method gives the *illusion* of correctness.

I don't have the time atm to throughly investigate this but I don't off the bat see any reason why `byGrapheme` wound't work in CTFE, especially since encoding/decoding work at CTFE. Fixing this would be the right way to go.

BBasile:
@JackStouffer, just compiles this

```d
import std.range, std.uni;
enum len = "abc".byGrapheme.walkLength;
```

To see what i'm talking about. If you think that `byGrapheme` is CTFE-able maybe that's because something is not **yet** tested.

BBasile:
blocked by https://issues.dlang.org/show_bug.cgi?id=18209

quickfur:
This is a long-standing issue. See: https://issues.dlang.org/show_bug.cgi?id=7054

It's not so easy to fix because of safety, purity, and performance issues with byGrapheme. But if you can make it work, it will be awesome.

quickfur:
@JackStouffer  The last time I checked, byGrapheme didn't work in CTFE because for runtime performance reasons it uses some hacks to store the first few characters in situ, so that memory allocation won't be needed unless it's an unusually-long grapheme cluster.  But there's probably a way to make it work by special-casing it for CTFE.

JackStouffer:
@quickfur Are you aware of a method of using static if to check if code is being executed in CTFE? If that's possible, then this is an easy fix.

BBasile:
I'm a partisan of _"this is still better than nothing"_, meaning:
- merge this partial fix
- open a proper issue on bugzilla: _byGrapheme not usable with CTFE_
- document the issue that's still there in the code. If someone wants to fix it he'll likely reach the comment saying _see http..._

About the if `(__ctfe)`. It appears that invalid branches are eliminated by the optimizer but in reality the two branches have to be valid before. So, i guess i teach you nothing, this is not usable for now, it's not like a `static if`.

Finally i'd like to say that it's a shame that Phobos cannot be right for this format thing "just" because of a core feature that is CTFE. Paradoxically.

JackStouffer:
Another possibility is to use this functionality for CTFE and byGrapheme for runtime.

quickfur:
The problem is that ever since `std.format` because CTFE-able, lots of code have come to rely on it, for better or for worse.  So we can't afford to break that now.

So here's a possible approach that allows us to at least make progress on *one* front, without holding everything up until somebody figures out how to make `byGrapheme` CTFE-able:

1. Use `if (__ctfe)` to isolate the current wrong width formatting to CTFE only, and use the `!__ctfe` branch to implement walking by grapheme cluster length.

2. File a bug to the effect that width formatting is wrong in CTFE.

3. Merge this fix, so that we get the correct behaviour at least at runtime.

4. Work on making byGrapheme CTFE-able, then fix the rest of the bug.

JackStouffer:
So for this PR, I'd approve it if it used walklength in CTFE, used byGrapheme in runtime, added a note to the docs about this discrepancy, and added a changelog entry.

quickfur:
@JackStouffer Unfortunately, we can't mix `static if` with CTFE. See: https://wiki.dlang.org/User:Quickfur/Compile-time_vs._compile-time

BBasile:
That's the idea assuming the difference between CT and RT can be made.

quickfur:
I just checked, it *is* possible to make this partial fix work. The trick here is that `byGrapheme` is compilable, and code that depends on it can be used in CTFE as long as the CTFE engine never actually evaluates it. So here's a proof of concept:

````
import std.range.primitives;
import std.stdio;
import std.uni;

size_t width(string s) {
        if (__ctfe) {
                return s.length;
        } else {
                return s.byGrapheme.walkLength;
        }
}

enum str = "Это досадно!";
pragma(msg, width(str));
void main() {
        writeln(width(str));
}
````

Here's the compiler output:
````
22LU
````

And here's the runtime output:
````
12
````

This means that a simple `if (__ctfe)` branch can be used, for now, to implement the partial fix.  Let's do that first, then we can think about how to make `byGrapheme` itself work in CTFE.

**Comment by BBasile** [comment]
Good. I thought i wouldn't work. Although it's still strange to have different result at CT and RT, so just `walkLength` for now ?

**Comment by quickfur** [comment]
Actually, I just started looking into the `byGrapheme` implementation, and I noticed that somebody has made it so that `graphemeStride` is actually usable from CTFE, you just can't extract the `Grapheme` struct itself because of the aforementioned runtime optimizations.  But for our purposes, we don't *need* to access the actual `Grapheme` units; all we care about is the *number* of grapheme clusters!  For that, we can just loop over the string using `graphemeStride` and count the number of clusters, and this does work in CTFE:

````
import std.range.primitives;
import std.stdio;
import std.uni;

size_t width(string s) {
        if (__ctfe) {
                size_t w = 0;
                size_t stride;
                for (size_t i = 0; i < s.length; i += graphemeStride(s, i)) {
                        w++;
                }
                return w;
        } else {
                return s.byGrapheme.walkLength;
        }
}

enum str = "Это доса\u0301дно!";
pragma(msg, width(str));
void main() {
        writeln(width(str));
        writeln(str);
}
````

Compiler output:
````
12LU
````

Runtime output:
````
12
````

I left the `if (__ctfe)` branch intact here just so you can see that the loop is returning the same value as `byGrapheme.walkLength`.  Since we don't actually need the `Grapheme` structs returned by `byGrapheme`, we can eliminate the 2nd branch altogether and use the loop for both runtime and CTFE.  So we can actually fix this bug in full right now!

**Comment by JackStouffer** [comment]
@quickfur Nice find.

**Comment by quickfur** [comment]
The only remaining concern is that grapheme clustering is, in general, rather expensive.  So we should ensure that `graphemeStride` is only called when necessary.  But that should not block this PR; it's better to be correct first, and optimize later if needed.

**Comment by BBasile** [comment]
@quickfur, thanks. I 'll update the PR tomorrow unless you want to do it.

File: std/format.d [hunk 1]
```diff
@@ -5777,7 +5777,16 @@ private bool needToSwapEndianess(Char)(const ref FormatSpec!Char f)
...
+    assert(format("%8s", "b\u00e9ll\u00f4") == "   b\u00e9ll\u00f4");
+}
+
+@safe pure unittest // bugzilla 18205
+{
+    assert("|%8s|".format("abc")        == "|     abc|");
+    assert("|%8s|".format("αβγ")        == "|     αβγ|");
+    assert("|%8s|".format("   ")        == "|        |");
+    assert("|%8s|".format("été")        == "|     été|");
+    assert("|%8s|".format("été 2018")   == "|été 2018|");
```
quickfur: Just a thought: since Unicode contains precomposed characters for `é`, etc., it might be better to spell out the grapheme clustering explicitly, e.g., `"e\u0301"`, so that it's clear what the unittest is checking.

std/format.d
```diff
@@ -3040,19 +3040,49 @@ if (isInputRange!T)
...
+                    // FIXME: byGrapheme.walkLength cannot be used in CTFE
+                    // see https://issues.dlang.org/show_bug.cgi?id=18209
+                    import std.uni : byGrapheme, graphemeStride;
+                    if (__ctfe)
+                    {
+                        size_t stride;
+                        for (size_t i = 0; i < s.length; i += graphemeStride(s, i))
+                            width++;
+                    }
+                    else width = val.byGrapheme.walkLength;
```
quickfur: We don't really need to use `byGrapheme` at all, since we don't care about the individual `Grapheme` structs. The loop in the `__ctfe` branch works equally well in CTFE and at runtime. Just use the `graphemeStride` loop always, no need to distinguish between CTFE and runtime.

std/format.d
```diff
@@ -3040,19 +3040,49 @@ if (isInputRange!T)
...
+                static if (is(CharTypeOf!E == char))
+                {
+                    onlyAscii = true;
+                    for (size_t i; i < s.length; i++)
+                        if (s[i] > 0x7F)
+                    {
+                        onlyAscii = false;
+                        break;
+                    }
+                }
```
quickfur: I'm not 100% sure this is necessary.  Well, I *did* say yesterday that grapheme clustering in general is expensive, but without profiling the code to be sure that that's actually a problem, I would lean towards not adding more complications to the code than necessary to fix the bug. Besides, the fix proposed here just involves walking the entire string one more time anyway, so it may not actually buy us any extra performance.

std/utf.d
```diff
@@ -3161,7 +3161,7 @@ version(unittest)
         @property C front() { return _str[0]; }
         void popFront() { _str = _str[1 .. $]; }
 
-        this(inout(C)[] str)
+        this(const(C)[] str)
```
quickfur: What's the reason for this change?

std/utf.d
```diff
@@ -3180,7 +3180,7 @@ version(unittest)
         @property auto save() { return BidirCU(_str); }
         @property size_t length() { return _str.length; }
 
-        this(inout(C)[] str)
+        this(const(C)[] str)
```
quickfur: ditto

std/utf.d
```diff
@@ -3201,7 +3201,7 @@ version(unittest)
         C opIndex(size_t i) { return _str[i]; }
         auto opSlice(size_t i, size_t j) { return RandomCU(_str[i .. j]); }
 
-        this(inout(C)[] str)
+        this(const(C)[] str)
```
quickfur: ditto

std/utf.d
```diff
@@ -3220,7 +3220,7 @@ version(unittest)
         @property auto save() { return new RefBidirCU(_str); }
         @property size_t length() { return _str.length; }
 
-        this(inout(C)[] str)
+        this(const(C)[] str)
```
quickfur: ditto

std/utf.d
```diff
@@ -3241,7 +3241,7 @@ version(unittest)
         C opIndex(size_t i) { return _str[i]; }
         auto opSlice(size_t i, size_t j) { return new RefRandomCU(_str[i .. j]); }
 
-        this(inout(C)[] str)
+        this(const(C)[] str)
```
quickfur: ditto

std/format.d
wilzbach: I would be interested in whether this improves the performance. `std.format` is very heavily used, so imho it's important to avoid introducing a huge performance regression.

std/utf.d
BBasile: it leads to instantiate [`byGrapheme!(inout(char)[])`](https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=2986788&isPull=true) which is an impossible type ctor, but since i'll remove byGrapheme, there are string probabilities that this will not required.

std/format.d
BBasile:
```d
benchmark!((){auto s = format("%100s","SSßŸé©é©é§SSÃŸß");})(1_000_000)[0].writeln;
```

before : 2 sec 13 ms
after: 3 sec 96 ms

When no width is specified there's no change (after a change just pushed)

std/utf.d
BBasile: removed

std/format.d
quickfur: What about a benchmark to compare the performance of an ASCII string vs. a UTF-8 string without grapheme clusters vs. UTF-8 with grapheme clusters?  Also, compare the performance with and without the `onlyAscii` optimization.

From an initial glance, it looks like the `onlyAscii` optimization might be necessary after all.  2 sec 13 ms to 3 sec 96 ms is a pretty big change.

std/format.d
```diff
@@ -3040,19 +3040,30 @@ if (isInputRange!T)
             static if (is(StringTypeOf!T))
             {
                 auto s = val[0 .. f.precision < $ ? f.precision : $];
+
+                size_t width;
+                if (f.width > 0)
+                {
+                    import std.uni : graphemeStride;
+                    size_t stride;
```
quickfur: Looks like an unused variable.

BBasile:
The onlyAscii optimization allows to maintain current performances but new benchmarks are worst than announced previously: 4  to 7  times slower. You'll have to choose between speed or correctness.

```d
#!dmd -O -release -inline -boundscheck=off
import std.stdio, std.datetime.stopwatch, std.format;

string test_ascii = "abcdefghijkl";
string test_utf_n = "ÃŸÃ©Ã Ã§ÃŸÃ©Ã Ã§ÃŸÃ©Ã Ã§";
string test_utf_g = "e\u0301e\u0301e\u0301e\u0301e\u0301e\u0301";

/*
before:
46 ms, 904 Î¼s, and 2 hnsecs
23 ms, 466 Î¼s, and 9 hnsecs
36 ms, 789 Î¼s, and 2 hnsecs

after, w/o ascii opt:
136 ms, 802 Î¼s, and 7 hnsecs
150 ms, 675 Î¼s, and 5 hnsecs
121 ms, 825 Î¼s, and 9 hnsecs

after, w/ ascii opt:
49 ms and 769 Î¼s
148 ms, 805 Î¼s, and 5 hnsecs
120 ms, 711 Î¼s, and 5 hnsecs
*/

void main()
{
    benchmark!((){auto s = format("%20s", test_ascii);})(100_000)[0].writeln;
    benchmark!((){auto s = format("%20s", test_utf_n);})(100_000)[0].writeln;
    benchmark!((){auto s = format("%20s", test_utf_g);})(100_000)[0].writeln;
}
```

quickfur:
OK, so looks like we definitely need the `onlyAscii` optimization. Without that, we're looking at a pretty significant performance deterioration.

Having said that, let's think about how to improve the `onlyAscii` loop. I'm not 100% certain, but I think having that if statement may be one cause of the serious performance degradation, because of the branch hazard it introduces. With modern CPUs, generally we can run a lot faster if there are no branches in the inner loop. So one idea is to use bitwise OR on the MSB of each char, which allows us to do something like this:

````
ubyte hasUpperBit = 0;
foreach (ch; s)
{
    hasUpperBit |= (ch & 0x80);
}
onlyAscii = (hasUpperBit == 0);
````

The trouble with this, of course, is if the string is very long and non-ASCII bytes appear early, then we may end up doing worse, because we don't know to stop the loop early.

One possible compromise is to scan the string by ulongs instead of individual bytes, something like this:
````
... // fill-in-the-blank here: take care of boundary cases where s.ptr is not ulong-aligned or s.length ends on a non-aligned boundary
foreach (word; cast(ulong[]) s)
{
    if ((word & 0x8080808080808080) != 0)
    {
        onlyAscii = false;
        break;
    }
}
````

This entails some pretty tricky code, though, so we might want to factor this out into a separate function so that we don't clutter the current already very-long function.

Another possibility is, we note that we actually don't need to call `graphemeStride` on every iteration when computing the number of clusters: if the string doesn't have any bytes above `0x7f`, for example.  So another way of fixing this bug that doesn't need to iterate over the string twice, is to dynamically switch to `graphemeStride` only if we encounter a byte that's > 127:

````
auto width = 0;
for (i=0; i < s.length;)
{
    if ((s[i] & 0x80) == 0)
        width++;
    else
        width += graphemeStride(s, i);
}
````

This does require a branch inside the loop, so I'm not 100% sure it will be an improvement.  OTOH, it eliminates the need to separate scan the string for non-ASCII characters, so depending on how bad the branch hazard is, it may actually save us some performance. It also makes the code a little shorter and easier to read, so I'm tempted to suggest to do it this way and merge the PR, then we can think about how to implement a more performant way of computing grapheme count.

What I'm thinking is, first let's merge a fix to this bug that isn't too horrible, then we should write a new function in std.uni, let's tentatively call it `graphemeCount`, that will implement a highly-optimized way of counting graphemes. Then we can refactor this function to just call `graphemeCount`.  Perhaps there's a way to squeeze more performance out of it, but let's not delay the correctness fix in this PR just because of performance.

**Comment by quickfur** [comment]
This performance thing really piqued my interest, so I did a quick benchmark of three approaches to counting graphemes: (1) the wrong approach (i.e., just count the number of code points via auto-decoding), (2) call `graphemeStride` to loop over the string, and (3) call `graphemeStride` only if the character value is > 127.  I got some pretty surprising results (see below).

Here's the code:
````
import std.stdio;
import std.uni;

size_t width0(string s) @safe pure
{
    size_t count;
    foreach (ch; s)
        count++;
    return count;
}

size_t width1(string s) @safe pure
{
    size_t count;
    for (size_t i = 0; i < s.length; i += graphemeStride(s, i))
        count++;
    return count;
}

size_t width2(string s) @safe pure
{
    size_t count;
    for (size_t i = 0; i < s.length;
         i += ((s[i] & 0x80) == 0) ? 1 : graphemeStride(s, i))
    {
        count++;
    }
    return count;
}

version(none)
size_t width3(string s) @safe pure
{
    size_t count;
}

alias Seq(A...) = A;

void main()
{
    auto str1 = "This is an ASCII-only string. Should be a performance baseline.";
    auto str2 = "Это строковый, который имеет не от таблицы ASCII символы.";
    auto str3 = "Этот stríng cõntaîns nón-ASCII\0301 и combi\0302ning diacri\u0301tics\u0301\u0302\u0303.";

    foreach (func; Seq!(width0, width1, width2))
    {
        auto result = benchmark!({
            auto w = func(str1);
        })(10_000);
        writefln("%s: %s", __traits(identifier, func), result[0]);

        result = benchmark!({
            auto w = func(str2);
        })(10_000);
        writefln("%s: %s", __traits(identifier, func), result[0]);
    
        result = benchmark!({
            auto w = func(str3);
        })(10_000);
        writefln("%s: %s", __traits(identifier, func), result[0]);

        writeln;
    }
}
````

Here's the output:
````
width0: 11 ms, 3 μs, and 7 hnsecs
width0: 6 ms and 942 μs
width0: 3 ms, 58 μs, and 5 hnsecs

width1: 183 ms, 163 μs, and 3 hnsecs
width1: 174 ms and 610 μs
width1: 172 ms, 9 μs, and 7 hnsecs

width2: 2 ms, 163 μs, and 6 hnsecs
width2: 132 ms, 300 μs, and 6 hnsecs
width2: 37 ms, 717 μs, and 1 hnsec
````

Unsurprisingly, just counting code points is generally speaking the fastest.  However, it is also wrong. Also unsurprisingly, calling `graphemeStride` per iteration is pretty slow, like over an order of magnitude slower.

Calling `graphemeStride` only when non-ASCII characters are found, though, seems to perform much faster. Surprisingly, it even beats counting code points when there are no non-ASCII characters!  Well, perhaps this is not *that* surprising, it is just proof that autodecoding is slow.  But anyway, it was able to take advantage of the ASCII characters in `str2` and `str3`, so that its performance more-or-less gracefully deteriorates depending on how many non-ASCII characters were encountered.  In this way it could beat calling `graphemeStride` per iteration even with `str2`, which is very heavy on non-ASCII characters. In `str3` we have some longer grapheme clusters sprinkled amidst a medium-to-high ratio of ASCII characters, and the loop was able to take advantage of this by avoiding a good number of calls to `graphemeStride`, which helped boost the performance significantly.

So it looks like this is the best way to go forward.  It's a relatively simple code change, has OK performance characteristics, and we can improve on it as a standalone unit later if we think of more ways of optimizing it.

**Comment by quickfur** [comment]
Actually, come to think of it, I wonder if we could just modify `graphemeStride` to just return 1 immediately if the first char in the string is < 128. Then we don't even need to change calling code to implement this optimization.

**Comment by BBasile** [comment]
I tried the "dynamic switch" and it is actually slightly slower. The problem is that a lookup (so a branch) is necessary since the cluster may start with an ASCII:

```d
for (size_t i; i < s.length;)
{
    import std.uni : graphemeStride;
    if ((s[i] & 0x80) == 0)
    {
        // next not ASCII: a cluster may start here
        if (i + 1 < s.length && (s[i+1] & 0x80))
            goto LStride;

        width++;
        i++;
        continue;
    }
LStride:
    i += graphemeStride(s, i);
    width ++;
}
```

quickfur:
Ohhh, you're right. :-(  So the reason I got a faster result was because my code was wrong. Haha.  Oh well, I guess that means we're back to optimizing non-ASCII detection.

DmitryOlshansky:
Otherwise I agree that using codepoints in __ctfe and graphemes at run-time seems a sensible choice

quickfur:
@DmitryOlshansky I think there's some misunderstanding here, we found out that `graphemeStride` is usable in CTFE, so the current consensus is to just use that for both CTFE and runtime.

As far as performance is concerned, I've been mulling over speeding up grapheme counting when the string is ASCII.  The latest idea is to search the string for non-ASCII bytes, and begin grapheme counting 1 character before that (to account for the case where the grapheme cluster begins with an ASCII base character). If the entire string is ASCII, or most of the string is ASCII, we can skip the slow grapheme decoding enough that we should be able to squeeze some performance out of it. For strings that are mostly non-ASCII, we have no choice but to decode every cluster anyway, so it's just a price we have to pay for correctness.

Also, basing string width on grapheme count is not the end of the story, though it is an important start.  There's still the additional pathology of zero-width characters and double-width characters (don't you just love Unicode?!). But let's not get bogged down in these details. At least get a reasonably efficient fix in first (well, reasonably efficient if the string is ASCII), then we can incrementally fix the rest.

std/format.d
```diff
@@ -3040,19 +3040,46 @@ if (isInputRange!T)
...
+                if (f.width > 0)
+                {
+                    // ASCII dstring & string can be aligned w/o graphemeStride
+                    bool onlyAscii;
+                    static if (!is(Unqual!(ElementEncodingType!T) == wchar))
+                    {
+                        onlyAscii = true;
+                        for (size_t i; i < s.length; i++)
+                            if (s[i] > 0x7F)
+                        {
```
quickfur: Style nitpick: please add braces around the loop body and an extra indent for the if-statement body. I know it's annoying to have another level of nesting for just a single statement, but I'd rather we stick to the Phobos style guidelines for consistency's sake.

BBasile: Okay done, but i'm sure i've already been merged in Phobos even if the code included an usage of this particular brace style.

std/format.d
```diff
@@ -3040,19 +3040,46 @@ if (isInputRange!T)
             static if (is(StringTypeOf!T))
             {
                 auto s = val[0 .. f.precision < $ ? f.precision : $];
+
+                size_t width;
+                if (f.width > 0)
+                {
+                    // ASCII dstring & string can be aligned w/o graphemeStride
```
quickfur: Actually, that's wrong. `dstring` also needs grapheme cluster decoding, because a grapheme like `a\u0310\u0337` occupies multiple dchars but only takes up one space when displayed.

BBasile: No this is right, this works, but i can add a test for that `assert("%2s".format("a\u0310\u0337"d) == " a\u0310\u0337");`
i meant  _"dstring only made of ASCII and string only made of ASCII..."_. In this case we have 3 `dchars`. only 1 is ASCII so `onlyAscii` is false and then the execution branches to the stride code.

BBasile: you seem to forget that `dstring`s are decoded... ;)

std/format.d
```diff
@@ -3040,19 +3040,47 @@ if (isInputRange!T)
             static if (is(StringTypeOf!T))
             {
                 auto s = val[0 .. f.precision < $ ? f.precision : $];
+
+                size_t width;
+                if (f.width > 0)
+                {
+                    // dstring & string that are fully made of ASCII characters
+                    // can be aligned w/o graphemeStride
```
quickfur: Why can't wstring containing only ASCII be aligned w/o graphemeStride?

std/format.d
quickfur: Yes, but why is the optimization not applied to `wstring`? If I have a wstring containing only ASCII characters, shouldn't it skip the grapheme stride code too?

std/format.d
BBasile: fixed

DmitryOlshansky:
> we have no choice but to decode every cluster anyway, so it's just a price we have to pay for correctness.

Actually there probably are ways to do that if you were to look at genericDecodeGrapheme:
https://github.com/dlang/phobos/blob/v2.078.0/std/uni.d#L6958

You'll see that it only applies to certain character classes and if your purpose is only to count width not to actually delimt graphemes.

You don't need state machine, instead just build a Trie of dchar --> bool, that indicates if a character is zero-width or not. We can prebuild the trie  but for benchmark just do:
```d
auto sizeTrie = codepointSetTrie!(8, 5, 8)(unicode.Mc | unicode.Grapheme_Extend);
```

Then the quick and dirty algo look like this:

for(size_t i = 0; i < s.length; )
{
  dchar ch = decode(s, i); // has 0x80 special case baked in
  if (ch < X)
     size += 1;
  else
     size += cast(int)sizeTrie[ch];
}

Where X is the biggest value until we hit the first codepoint of graphemeExtend or spacingMark.

It will work for anything but Korean algorithmic Hangul.

That can be fixed with separate branch.
Basically hangul is a sequence of codepoints that can have any of L, V, T "bits". Just look at state-machine and I'm certain even hangul logic path can be greatly simplified by "just count the size" vs  "find the end of current cluster".

For the record, hangul codepoints are:
```d
bool isHangL(dchar ch) @safe pure nothrow
{
    if (ch < 4352) return false;
    if (ch < 4448) return true;
    if (ch < 43_360) return false;
    if (ch < 43_389) return true;
    return false;
}

bool isHangV(dchar ch) @safe pure nothrow
{
    if (ch < 4448) return false;
    if (ch < 4520) return true;
    if (ch < 55_216) return false;
    if (ch < 55_239) return true;
    return false;
}

bool isHangT(dchar ch) @safe pure nothrow
{
    if (ch < 4520) return false;
    if (ch < 4608) return true;
    if (ch < 55_243) return false;
    if (ch < 55_292) return true;
    return false;
}
```

std/format.d
```diff
@@ -3040,19 +3040,45 @@ if (isInputRange!T)
             static if (is(StringTypeOf!T))
             {
                 auto s = val[0 .. f.precision < $ ? f.precision : $];
+
+                size_t width;
+                if (f.width > 0)
+                {
+                    // strings that are fully made of ASCII characters
+                    // can be aligned w/o graphemeStride
+                    bool onlyAscii = true;
```
JackStouffer:
```d
bool onlyAscii = s.byCodeUnit.all!isASCII;
```

If only. I understand why you broke it out into a custom loop; it's such a shame we have to work around DMD's inliner all the time.

BBasile: Yes, indeed. I thought to use a higher order function but since testing for "ascii-ness" is trivial i choose to avoid imports to std.algo + std.scii.

quickfur: IMO one of these days we should spearhead a dmd optimizer improvement project, where we identify all such instances of deficiency in the dmd inliner and push for them to be fixed.  I'm thinking that all range-based idioms really need stronger support from D's reference compiler, since they're supposed to be one of D's selling points. More aggressive inlining is part of the necessary optimization in order to make range-based code more performant.

The other big area that will probably make a big impact is loop unrolling, which IIRC Walter says isn't done by dmd except in special cases. But that is probably far more complex than improving the inliner.

quickfur:
@DmitryOlshansky Sounds like a promising way forward. Let's merge this PR as-is first, then we can implement what you suggest in a followup PR, since it's likely to improve performance as well as correctness: if we build the size trie to account for double-width characters like the CJK block, we can address that part of the bug as well.

quickfur:
Unless anybody has objections, I'm going to merge this and submit further improvements as followup PRs.

quickfur:
I spent the better part of today looking into computing the display width of a string in monospace font.  Which led to the following cases that haven't yet been considered (don't worry, I'm not asking for all of this to be implemented in this PR, let's just do that in a followup PR/PRs):

* Zero-width characters, such as U+200B, that are segmented as individual graphemes by `graphemeStride` but should not take up any space in the output.

* East Asian Widths: TR11 specifies a (rather large!) number of blocks as being "wide" or "full width". On most monospace text terminals, these are rendered in square cells that take up twice the width of a traditional character in European script.  Ostensibly, characters from these blocks ought to count for 2 spaces rather than 1.

* Hangul / Jamo syllables: while these are correctly segmented as single graphemes by `graphemeStride`, Jamo syllables happen to be in the block designated as "Wide", meaning that a 3-code-point LVT sequence like U+1100 U+1161 U+11A8 ought to be counted as taking up *two* spaces, not 1 or 3. Haha.

I wrote up a bunch of test cases and a (slow) reference implementation that computes string width, as groundwork for exploring how to implement an efficient solution that addresses all of the above.   Some of the test cases are sufficiently deep into the dark corners of Unicode that they exposed bugs in my current terminal(!).

Anyway, here's my current idea on how to implement efficient width computation:

* As @DmitryOlshansky pointed out, if we don't care about actual grapheme segmentation, we can cut some corners and get better performance than `graphemeStride` (note that we have to do more than `graphemeStride` to address the above points, so a naïve implementation will be even worse than what we have in this PR).  Namely, all combining marks and grapheme extenders can be ignored.

* For Jamo, my understanding is that "proper" Jamo text should always consist of an L code point, followed by V and optionally T, i.e., proper combinations are LV and LVT.  For our purposes, all we care about are L code points, since an LV or LVT sequence is supposed to be interpreted as a single grapheme (single glyph).  So we can just ignore V and T code points altogether, from what I understand.

* Code points with East Asian Width value W (wide) or F (full) need to count as 2 spaces rather than 1. Unfortunately, Unicode being Unicode, there are Ambiguous cases. But thankfully, according to TR11 section 5, we can treat ambiguous as narrow by default unless we have more information, so my inclination is just to ignore them and default to narrow width. Similarly, neutral (N) characters can probably be counted the same as narrow.  The codepoint set for wide characters W + F, unfortunately, is quite complex and has no simple pattern, so the best way is probably to use a `CodepointTrie` for O(1) lookup at runtime.

Given all of the above, I'm thinking to construct a single `CodepointTrie` that encapsulates all of the above by mapping every codepoint to either 0, 1, or 2, and the width computation is simply a matter of summing these values. All combining diacritics, zero width characters, grapheme extenders, and Jamo V and T codepoints will be assigned width 0. Of the remaining codepoints, those designated to be W (wide) or Full-width (F) by EastAsianWidth.txt will be assigned width 2, and everything else is width 1.

Of course, looking up the trie per code point is still a pretty expensive operation, even if it's O(1), so we'll want to short-circuit the loop by skipping codepoints < N, where N is the first codepoint that has width != 1 (the ASCII block has all width 1, AFAICT). The number of skipped codepoints is added to the total width since they all have width 1.  Better yet, don't even decode if the char is < 128, since we know it's ASCII with 1 byte per char with width 1.

JackStouffer:
>the ASCII block has all width 1, AFAICT

ASCII 1-30 have no width, but I have never seen these characters in the wild. IIRC they were all old terminal control characters.

BBasile:
It's still time to remove the auto-merge label and handle these values.

wilzbach:
> It's still time to remove the auto-merge label and handle these values.

As you are the author of this PR, I just did so (FYI: pushing to a PR will also remove it).

BBasile:
Thanks, you can re-label for merging.

BBasile:
@wilzbach: ci/circleci never start. I think that a manual intervention is required or something.

std/format.d
```diff
@@ -3040,19 +3040,46 @@ if (isInputRange!T)
...
+                    for (size_t i; i < s.length; i++)
+                    {
+                        width += s[i] >= 30;
```
quickfur: I disagree with this change.  The characters under ASCII 0x20 (32) are control characters, and can have effects on formatting. For example, `\n` is ASCII 0x0A (13) (on Posix anyway, on Windows you also get 0x0D with it), and has the side-effect of breaking the line completely. So what "width" should it have?  Similarly, `\t` is ASCII 0x09, and has the side-effect of changing the current horizontal position of the output differently depending on how the target terminal interprets it.  There are a few others that have similar side-effects, not to mention escape sequences prefixed by ASCII 0x1B, that changes the interpretation of an arbitrary number of following characters, again depending on how the terminal interprets it.

IMO it's outside the charter of `std.format` to interpret these control characters, so asking it to format these characters into a particular width is basically UB.  As such, the implementation should not sacrifice speed/complexity to treat them specially.  It's not *wrong* per se to treat them as if they have zero width, but I don't think we should go out of our way just to do that.

quickfur: Note furthermore that we cannot just ignore them, because the user *may* have inserted them specifically because they wanted the side-effects associated with them. E.g., we regularly write `\n` to introduce line breaks into the output, and conceivably, someone could call `format("%s", obj)` and `obj.toString` might return a multi-line string.   But what should be done in the case `format("%.2s", "\t")`?  Or indeed, `format("%.1s", "\nX")`?  Since these cases have no sane answer, my inclination is to call them UB and just not handle them at all in the code.

quickfur: (Also, the code as-is is wrong, because ASCII 30 and 31 are also control characters, but the current if-condition would treat them as having width 1. The correct condition should be `>= 32`.)

BBasile: Ok removed. Since you're there you can re-label for merging.

quickfur:
@DmitryOlshansky @BBasile  I've put up my latest experiments in computing string width in a public repo: https://github.com/quickfur/strwidth

This isn't intended to supplant this PR, but the plan is to eventually submit it as a followup PR. Basically it implements what I described a few comments up.

Things to be done:

* Benchmark it with a proper data set in addition to the correctness test cases I already have, so that we can optimize the code further.

* Implement basic ASCII string optimization by skipping substrings that are purely ASCII.

* Any other optimizations that benchmarking / profiling may suggest.

* Turn the Trie into a static table ala std/internal/unicode_tables.d, rather than constructing this in a static ctor.  The current construction solely relies on the public API of std.uni, so it's terribly inefficient (loops over individual codepoints rather than adding ranges directly, since we already have range data handy).   But it's Good Enough for a proof-of-concept.  Will need @DmitryOlshansky's help on how to generate the static Trie nodes.

BBasile:
If you think it could be ready in the next 6 weeks maybe it's better to close this PR without merging since actually you piloted most of the things after the naive initial version that used `.walkLength`.

quickfur:
I think it's okay to merge this PR first, in case unforeseen problems prevent my version from being merged before next release, then at least we'll have a more correct implementation.

JackStouffer:
>If you think it could be ready in the next 6 weeks maybe it's better to close this PR without merging 

I have to disagree; let's not let perfect be the enemy of the good. This is a clear improvement over the status quo that's ready *now*, so let's get this merged.

JackStouffer:
I'd just like to point out that @quickfur and I commented on the exact same second.

wilzbach:
> @wilzbach: ci/circleci never starts. I think that a manual intervention is required or something.

Ehm I also have now idea what that happens.
Here's what I can see:

image

Usually the URL looks like https://circleci.com/gh/dlang/phobos/tree/pull%2F6005 but https://circleci.com/gh/dlang/phobos/tree/pull%2F6008 is empty.

I will wait until Jenkins has built this PR and then manually over-write auto-merge.
FYI: auto-tester is already happy:

image

I just did the `git commit --amend` trick to restart DAutoTest.

std/format.d
```diff
@@ -3040,19 +3040,45 @@ if (isInputRange!T)
...
             {
                 auto s = val[0 .. f.precision < $ ? f.precision : $];
+
+                size_t width;
+                if (f.width > 0)
+                {
+                    // strings that are fully made of ASCII characters
+                    // can be aligned w/o graphemeStride
+                    bool onlyAscii = true;
+                    for (size_t i; i < s.length; i++)
```
wilzbach: You could have used at least `foreach`

std/format.d
wilzbach: @BBasile do you remember how big the performance penalty of using `s.byCodeUnit.all!isASCII` was?

jondegenhardt:
> I spent the better part of today looking into computing the display width of a string in monospace font. Which led to the following cases that haven't yet been considered (don't worry, I'm not asking for all of this to be implemented in this PR, let's just do that in a followup PR/PRs):

@quickfur I think [Issue 17810 - Add wcwidth/wcswidth equivalents to std.uni](https://issues.dlang.org/show_bug.cgi?id=17810) is related to this.

quickfur:
@jondegenhardt Thanks for the pointer! This definitely looks like what my new string width implementation is doing.

std/format.d
andralex: for obvious adjustments feel free to just patch the code

jondegenhardt:
@quickfur As an fyi, I filed [17810](https://issues.dlang.org/show_bug.cgi?id=17810) when I was writing a tool pretty printing for tabular data files, with the columns nicely aligned. For that it was necessary to estimate the expected print width.

std/format.d
```diff
@@ -3040,19 +3040,45 @@ if (isInputRange!T)
...
+                    {
+                        if (s[i] > 0x7F)
+                        {
+                            onlyAscii = false;
+                            break;
+                        }
+                    }
+                    if (!onlyAscii)
+                    {
```
DmitryOlshansky: Obviously we can start from the last place where ASCII-only portion ends.

BBasile:
@MartinNowak : i don't see it in next change log: https://dlang.org/changelog/pending.html#bugfix-list.
Is there any reason why ?

wilzbach:
> @MartinNowak : i don't see it in next change log: https://dlang.org/changelog/pending.html#bugfix-list.
Is there any reason why ?

image

BBasile:
Other issues should have been opened. The situation was already greatly enhanced.
By the way since you're there @wilzbach, i see you still don't have updated D-Scanner for the layout PR of Alexandrescu.

wilzbach:
> Other issues should have been opened. The situation was already greatly enhanced.

Done. I opened separate issues for the individual items left:

https://issues.dlang.org/show_bug.cgi?id=18465
https://issues.dlang.org/show_bug.cgi?id=18466
https://issues.dlang.org/show_bug.cgi?id=18467

Once the changelog is regenerated (i.e. the next PR merged, `pending.html` will be updated).
[The first beta for 2.079 has already been generated](https://dlang.org/changelog/2.079.0.html), so it currently doesn't appear on this list, but on the next manual generation the list will be updated too.

> By the way since you're there @wilzbach, i see you still don't have updated D-Scanner for the layout PR of Alexandrescu.

Thanks for the ping. I was away over the weekend and got flooded as usual with emails.

-> https://github.com/dlang/phobos/pull/6197

BBasile:
Thanks you.

PR #6006, opened 2018-01-07
Add PHOBOSSRC urls to std.experimental

wilzbach:

PR #6009, opened 2018-01-08
Expose AscendingPageAllocator on dlang.org

wilzbach:
Follow-up to https://github.com/dlang/phobos/pull/6006/files#r160067505

Note that the module title is too long for the menu atm.

image

image

CC @Darredevil

BTW why don't we put everything into a `building_blocks.d` file like it's
done for all the other Phobos modules?

### DIFF
--- a/std/experimental/allocator/building_blocks/ascending_page_allocator.d
+++ b/std/experimental/allocator/building_blocks/ascending_page_allocator.d
@@ -1,3 +1,7 @@
+// Written in the D programming language.
+/**
+Source: $(PHOBOSSRC std/experimental/allocator/building_blocks/_ascending_page_allocator.d)
+*/
 module std.experimental.allocator.building_blocks.ascending_page_allocator;
 import std.experimental.allocator.common;
 


wilzbach:
I pinged the wrong Alex. Ping @jercaianu 

@jercaianu you added this module in https://github.com/dlang/phobos/pull/6009 - it's not displayed on the docs yet. Can you please have a look the description above and let me know your thoughts? Thanks!

jercaianu:
Yes, it's pretty long, however it's not intended for common use.
An alternative would be to make the name shorter, but to change it just because the name does not fit the CSS, I don't think it's right.

Is there any upper limit on module names?

Thanks,
Alex

jercaianu:
@CyberShadow @wilzbach 
What I could do is change the CSS such that for longer module names,  I might add trailing '...'
For example, instead of 'ascending_page_allocator', it would be 'ascending_page_...'.

Would that be reasonable?
Alex

wilzbach:
> What I could do is change the CSS such that for longer module names, I might add trailing '...'

Sounds like a good idea!

> Yes, it's pretty long, however it's not intended for common use.

Then why is it in the standard library ;-)

jercaianu:
@wilzbach I think I fixed it here [1].
Thanks,
Alex

[1] - https://github.com/dlang/dlang.org/pull/2123

JackStouffer:
dlang/dlang.org#2123 has been merged

PR #4628, opened 2016-07-21
fix Issue 16307 - [ndslice] should be safe

9il:

PR #6010, opened 2018-01-08
Remove unsafe, indiscriminate '@trusted:' from std.uni

JackStouffer:
I'm leaning towards removing all `attribute:` usage outside of structs from Phobos...

### DIFF
--- a/std/uni.d
+++ b/std/uni.d
@@ -7369,8 +7369,6 @@ if (isInputRange!Range && is(Unqual!(ElementType!Range) == dchar))
     assert("абвгд"d.byCodePoint.length == 5);
 }
 
-@trusted:
-
 /++
     $(P A structure designed to effectively pack $(CHARACTERS)
     of a $(CLUSTER).
@@ -7442,13 +7440,13 @@ public:
         Warning: Invalidates when this Grapheme leaves the scope,
         attempts to use it then would lead to memory corruption.
     +/
-    @system SliceOverIndexed!Grapheme opSlice(size_t a, size_t b) pure nothrow @nogc
+    SliceOverIndexed!Grapheme opSlice(size_t a, size_t b) pure nothrow @nogc
     {
         return sliceOverIndexed(a, b, &this);
     }
 
     /// ditto
-    @system SliceOverIndexed!Grapheme opSlice() pure nothrow @nogc
+    SliceOverIndexed!Grapheme opSlice() pure nothrow @nogc
     {
         return sliceOverIndexed(0, length, &this);
     }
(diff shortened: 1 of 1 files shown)


std/uni.d
```diff
@@ -7442,13 +7440,13 @@ public:
         Warning: Invalidates when this Grapheme leaves the scope,
         attempts to use it then would lead to memory corruption.
     +/
-    @system SliceOverIndexed!Grapheme opSlice(size_t a, size_t b) pure nothrow @nogc
+    SliceOverIndexed!Grapheme opSlice(size_t a, size_t b) pure nothrow @nogc
```
JackStouffer: So this function is supposedly supposed to be `@system`, but the `@trusted:` over-rid (overrided?) that, and now there's a ton of functionality which relies on it being `@trusted`.

wilzbach:
> I'm leaning towards removing all attribute: usage from outside of structs on Phobos...

It shouldn't be too tricky to add a DScanner check for it ;-)

BBasile:
[I remember this issue now](https://github.com/dlang-community/D-Scanner/issues/425). Maybe it wasn't well formulated enough but it's basically the same problem.

quickfur:
IIRC, somebody added the `@trusted` annotations so that std.uni is usable from `@safe`. There really needs to be a whole bunch of unittest to test usability from `@safe` and so forth, so that changes like these are less likely to break existing code without getting caught by the autotester.

JackStouffer:
@quickfur All of the unittests are marked as either safe or system already which is how I knew to fix normalize

jmdavis:
> I'm leaning towards removing all attribute: usage outside of structs from Phobos...

There are plenty of cases where it makes sense to use attributes directly on functions. Even when a function is templated, it can make sense. The bigger problem is when they're used at the module-level, though honestly, I think that using them within a `struct` or `class` is almost as bad. In general, I think that mass-applying attributes is a mistake and that they should be applied to each function individually, since it's so easy to screw them up otherwise. `@trusted` is particularly bad for that, particularly since there is not compiler check for screwing that one up (unlike `@safe`, `pure`, or `nothrow` which will complain at least some of the time, though using them incorrectly on templated code can go uncaught depending on the unit tests), and in general, I think that the use of `@trusted` should be restricted to the most local level that makes sense in order to reduce the risk of misuse. `@trusted` is basically like a cast to force the compiler to shut up, and applying that at a higher level is just asking for it.

wilzbach:
> @trusted is particularly bad for that, particularly since there is not compiler check for screwing that one up

Btw there has been one precedent I can think of: https://github.com/dlang/phobos/pull/4628 (it got rejected)

std/uni.d
stefan-koch-sociomantic: overwrote

PR #6011, opened 2018-01-08
Added note to isInstanceOf docs on template usage

JackStouffer:


std/traits.d
```diff
@@ -6501,6 +6501,30 @@ template isInstanceOf(alias S, alias T)
     static assert(isInstanceOf!(templ, templ!int));
 }
 
+/**
+ * To use `isInstanceOf` to check the type of a template while inside of said
```
BBasile: `identity of...` rather then `type of...`

JackStouffer:
fixed

JackStouffer:
Wait, screwed up.

```
static assert(!__traits(compiles, a1.func(a1)));
```

should be

```
static assert(!__traits(compiles, a1.func(a3)));
```

PR #6012, opened 2018-01-08

wilzbach:


wilzbach:
For those interested. This is done with: https://github.com/wilzbach/git-tools/blob/master/merge_stable

wilzbach:
> CyberShadow/DAutoTest — Waiting for status to be reported

Seems like DAutoTest lost the hook. Tried the goold old `git commit --amend` trick to rekick it.

PR #5698, opened 2017-08-19
Add std.complex.coshisinh and remove uses of creal math.

ibuclaw:
Small addition for https://github.com/dlang/dmd/pull/7081

Everything else looks like just deleting support code from phobos.

PR #4978, opened 2016-12-20
Remove deprecated doFormat

wilzbach:
Apparently we missed the deprecation cleanup on this one.

It's part of undeaD:

https://github.com/dlang/undeaD/blob/master/src/undead/doformat.d

PR #6014, opened 2018-01-08
Start to move away from complex and imaginary types

wilzbach:
This is first PR which works in the direction of finally going through with the
complex and imaginary type deprecation.

See also: https://github.com/dlang/dmd/pull/7640

### DIFF
--- a/std/complex.d
+++ b/std/complex.d
@@ -821,7 +821,6 @@ Complex!(CommonType!(T, U)) fromPolar(T, U)(T modulus, U argument)
 */
 Complex!T sin(T)(Complex!T z)  @safe pure nothrow @nogc
 {
-    import std.math : expi, coshisinh;
     auto cs = expi(z.re);
     auto csh = coshisinh(z.im);
     return typeof(return)(cs.im * csh.re, cs.re * csh.im);
@@ -839,7 +838,6 @@ Complex!T sin(T)(Complex!T z)  @safe pure nothrow @nogc
 /// ditto
 Complex!T cos(T)(Complex!T z)  @safe pure nothrow @nogc
 {
-    import std.math : expi, coshisinh;
     auto cs = expi(z.re);
     auto csh = coshisinh(z.im);
     return typeof(return)(cs.re * csh.re, - cs.im * csh.im);
@@ -849,12 +847,16 @@ Complex!T cos(T)(Complex!T z)  @safe pure nothrow @nogc
 @safe pure nothrow unittest
 {
     import std.complex;
-    import std.math;
     assert(cos(complex(0.0)) == 1.0);
-    assert(cos(complex(1.3L)) == std.math.cos(1.3L));
-    assert(cos(complex(0, 5.2L)) == cosh(5.2L));
 }
 
+deprecated
+@safe pure nothrow unittest
+{
+    import std.math;
+    assert(cos(complex(0, 5.2L)) == cosh(5.2L));
+    assert(cos(complex(1.3L)) == std.math.cos(1.3L));
+}
 
 /**
     Params: y = A real number.
@@ -875,15 +877,60 @@ Complex!real expi(real y)  @trusted pure nothrow @nogc
 ///
 @safe pure nothrow unittest
 {
+    import std.math : cos, sin;
+    assert(expi(0.0L) == 1.0L);
+    assert(expi(1.3e5L) == complex(cos(1.3e5L), sin(1.3e5L)));
+}
+
+deprecated
+@safe pure nothrow unittest
+{
     static import std.math;
 
     assert(expi(1.3e5L) == complex(std.math.cos(1.3e5L), std.math.sin(1.3e5L)));
-    assert(expi(0.0L) == 1.0L);
     auto z1 = expi(1.234);
     auto z2 = std.math.expi(1.234);
     assert(z1.re == z2.re && z1.im == z2.im);
 }
 
+/**
+    Params: y = A real number.
+    Returns: The value of cosh(y) + i sinh(y)
+
+    Note:
+    $(D coshisinh) is included here for convenience and for easy migration of code
+    that uses $(REF _coshisinh, std,math).
+*/
+Complex!real coshisinh(real y) @safe pure nothrow @nogc
+{
+    static import std.math;
+    if (std.math.fabs(y) <= 0.5)
+        return Complex!real(std.math.cosh(y), std.math.sinh(y));
+    else
+    {
+        auto z = std.math.exp(y);
+        auto zi = 0.5 / z;
+        z = 0.5 * z;
+        return Complex!real(z + zi, z - zi);
+    }
+}
+
+///
+@safe pure nothrow @nogc unittest
+{
+    import std.math : cosh, sinh;
+    assert(coshisinh(3.0L) == complex(cosh(3.0L), sinh(3.0L)));
+}
+
+deprecated
+@safe pure nothrow @nogc unittest
+{
+    static import std.math;
+    assert(coshisinh(3.0L) == complex(std.math.cosh(3.0L), std.math.sinh(3.0L)));
+    auto z1 = coshisinh(1.234);
+    auto z2 = std.math.coshisinh(1.234);
+    assert(z1.re == z2.re && z1.im == z2.im);
+}
 
 /**
     Params: z = A complex number.
--- a/std/format.d
+++ b/std/format.d
@@ -2590,6 +2590,7 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
 /*
     Formatting a $(D creal) is deprecated but still kept around for a while.
  */
+deprecated("Use of complex types is deprecated. Use std.complex")
 private void formatValueImpl(Writer, T, Char)(auto ref Writer w, T obj, const ref FormatSpec!Char f)
 if (is(Unqual!T : creal) && !is(T == enum) && !hasToString!(T, Char))
 {
@@ -2604,6 +2605,8 @@ if (is(Unqual!T : creal) && !is(T == enum) && !hasToString!(T, Char))
     put(w, 'i');
 }
 
+version(TestComplex)
+deprecated
 @safe /*pure*/ unittest     // formatting floating point values is now impure
 {
     import std.conv : to;
@@ -2621,6 +2624,8 @@ if (is(Unqual!T : creal) && !is(T == enum) && !hasToString!(T, Char))
     }
 }
 
+version(TestComplex)
+deprecated
 @system unittest
 {
     formatTest( 3+2.25i, "3+2.25i" );
@@ -2641,6 +2646,7 @@ if (is(Unqual!T : creal) && !is(T == enum) && !hasToString!(T, Char))
 /*
     Formatting an $(D ireal) is deprecated but still kept around for a while.
  */
+deprecated("Use of imaginary types is deprecated. Use std.complex")
 private void formatValueImpl(Writer, T, Char)(auto ref Writer w, T obj, const ref FormatSpec!Char f)
 if (is(Unqual!T : ireal) && !is(T == enum) && !hasToString!(T, Char))
 {
@@ -2650,6 +2656,8 @@ if (is(Unqual!T : ireal) && !is(T == enum) && !hasToString!(T, Char))
     put(w, 'i');
 }
 
+version(TestComplex)
+deprecated
 @safe /*pure*/ unittest     // formatting floating point values is now impure
 {
     import std.conv : to;
@@ -2661,6 +2669,8 @@ if (is(Unqual!T : ireal) && !is(T == enum) && !hasToString!(T, Char))
     }
 }
 
+version(TestComplex)
+deprecated
 @system unittest
 {
     formatTest( 2.25i, "2.25i" );
@@ -5494,20 +5504,22 @@ private bool needToSwapEndianess(Char)(const ref FormatSpec!Char f)
 
     s = format("%d %s", 0x1234AF, 0xAFAFAFAF);
     assert(s == "1193135 2947526575");
+}
 
-    //version(X86_64)
-    //{
-    //    pragma(msg, "several format tests disabled on x86_64 due to bug 5625");
-    //}
-    //else
-    //{
-        s = format("%s", 1.2 + 3.4i);
+version(TestComplex)
+deprecated
+@system unittest
+{
+        string s = format("%s", 1.2 + 3.4i);
         assert(s == "1.2+3.4i", s);
+}
 
-        //s = format("%x %X", 1.32, 6.78f);
-        //assert(s == "3ff51eb851eb851f 40D8F5C3");
+@system unittest
+{
+    import std.conv : octal;
 
-    //}
+    string s;
+    int i;
 
     s = format("%#06.*f",2,12.345);
     assert(s == "012.35");
--- a/std/math.d
+++ b/std/math.d
@@ -49,9 +49,6 @@ $(TR $(TDNW Introspection) $(TD
     $(MYREF isNormal) $(MYREF isSubnormal) $(MYREF signbit) $(MYREF sgn)
     $(MYREF copysign) $(MYREF isPowerOf2)
 ))
-$(TR $(TDNW Complex Numbers) $(TD
-  $(MYREF abs) $(MYREF conj) $(MYREF sin) $(MYREF cos) $(MYREF expi)
-))
 $(TR $(TDNW Hardware Control) $(TD
     $(MYREF IeeeFlags) $(MYREF FloatingPointControl)
 ))
@@ -507,6 +504,36 @@ enum real SQRT2 =      0x1.6a09e667f3bcc908b2fb1366ea958p+0L; /** $(SQRT)2 = 1.4
 enum real SQRT1_2 =    SQRT2/2;                               /** $(SQRT)$(HALF) = 0.707106... */
 // Note: Make sure the magic numbers in compiler backend for x87 match these.
 
+// it's quite tricky check for a type who will trigger a deprecation when accessed
+template isDeprecatedComplex(T)
+{
+    static if (__traits(isFloating, T) && __traits(isDeprecated, T))
+    {
+        enum isDeprecatedComplex = true;
+    }
+    else
+    {
+        enum m = T.mangleof;
+        // cfloat, cdouble, creal
+        // ifloat, idouble, ireal
+        enum isDeprecatedComplex = m == "q" || m == "r" || m == "c" ||
+                                   m == "o" || m == "p" || m == "j";
+    }
+}
+
+deprecated unittest
+{
+    static assert(isDeprecatedComplex!cfloat);
+    static assert(isDeprecatedComplex!cdouble);
+    static assert(isDeprecatedComplex!creal);
+    static assert(isDeprecatedComplex!ifloat);
+    static assert(isDeprecatedComplex!idouble);
+    static assert(isDeprecatedComplex!ireal);
+
+    static assert(!isDeprecatedComplex!float);
+    static assert(!isDeprecatedComplex!double);
+    static assert(!isDeprecatedComplex!real);
+}
 
 /***********************************
  * Calculates the absolute value of a number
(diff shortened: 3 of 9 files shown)


std/format.d
```diff
@@ -5425,37 +5425,23 @@ enum Mangle : char
...
-    }
-    auto p = m in dic;
-    return p ? *p : null;
+    enum types = [
+        "void", "bool", "byte", "ubyte",
+        "short", "ushort", "int", "uint", "long", "ulong",
+        "float", "double", "real",
+        "char", "wchar", "dchar",
+        "ifloat", "idouble", "ireal",
+        "cfloat", "cdouble", "creal",
```
wilzbach: How can we call `typeid` without a deprecation warning?

std/math.d
```diff
@@ -526,50 +555,48 @@ enum real SQRT1_2 =    SQRT2/2;                               /** $(SQRT)$(HALF)
...
-}
-
 /// ditto
 @safe pure nothrow @nogc unittest
 {
+    import std.complex : complex;
     assert(isIdentical(abs(-0.0L), 0.0L));
     assert(isNaN(abs(real.nan)));
     assert(abs(-real.infinity) == real.infinity);
-    assert(abs(-3.2Li) == 3.2L);
```
wilzbach: The complex deprecation currently triggers error in deprecated code. See https://issues.dlang.org/show_bug.cgi?id=18212

std/format.d
ibuclaw: I think the right steps are to first deprecate in library (user-code).  Once enough time has passed, remove support for complex and imaginary in the library.

Then we kill it in the compiler.

wilzbach:
> I think the right steps are to first deprecate in library (user-code). Once enough time has passed, remove support for complex and imaginary in the library.
> Then we kill it in the compiler starting with deprecation by default.

We can't do that. Phobos needs to build without deprecation messages. This was enabled about 9 months ago, because (1) deprecation messages were shown to users and (2) people always introduced deprecations without considering the fallout and viability. By `-de` we can enforce that if a symbol gets deprecated in Phobos, all usages of it have been removed (or are in `deprecated` blocks). Since it was introduced, it has proven as quite helfpul.

andralex:
>Phobos needs to build without deprecation messages.

Slight amendment: it's okay to have deprecation messages in phobos' unittest builds.

wilzbach:
> Slight amendment: it's okay to have deprecation messages in phobos' unittest builds.

Yep, but only if they are marked as `deprecated`. However, the complex deprecation messages doesn't respect this at the moment: https://issues.dlang.org/show_bug.cgi?id=18212

std/math.d
```diff
@@ -2545,42 +2667,45 @@ private real exp2Impl(real x) @nogc @trusted pure nothrow
...
-                fsincos;
-                fxch ST(1), ST(0);
-            }
-        }
-    }
-    else
-    {
-        return cos(y) + sin(y)*1i;
-    }
+auto expi(real y) @trusted pure nothrow @nogc
```
wilzbach: Not sure how to deal with this in another way.

std/math.d
```diff
@@ -732,50 +818,72 @@ float sin(float x) @safe pure nothrow @nogc { return sin(cast(real) x); }
...
+    assert(sin(complex(2.0, 0.0)) == sin(2.0L));
 }
 
 /***********************************
  *  cosine, complex and imaginary
  *
  *  cos(z) = cos(z.re)*cosh(z.im) - sin(z.re)*sinh(z.im)i
  */
-creal cos(creal z) @safe pure nothrow @nogc
+auto cos(C)(C z) @safe pure nothrow @nogc
```
wilzbach: Not sure how to deal with this in another way.

std/format.d
```diff
@@ -5419,31 +5420,6 @@ enum Mangle : char
     Timmutable = 'y',
 }
 
-// return the TypeInfo for a primitive type and null otherwise.  This
-// is required since for arrays of ints we only have the mangled char
-// to work from. If arrays always subclassed TypeInfo_Array this
-// routine could go away.
-private TypeInfo primitiveTypeInfo(Mangle m)
```
wilzbach: So `primitiveTypeInfo` is a relict from the old `doFormat` days. It's not used anywhere anymore.
See e.g. https://github.com/dlang/phobos/pull/4978

std/math.d
```diff
@@ -507,30 +504,38 @@ enum real SQRT2 =      0x1.6a09e667f3bcc908b2fb1366ea958p+0L; /** $(SQRT)2 = 1.4
 enum real SQRT1_2 =    SQRT2/2;                               /** $(SQRT)$(HALF) = 0.707106... */
 // Note: Make sure the magic numbers in compiler backend for x87 match these.
 
+// it's quite tricky check for a type who will trigger a deprecation when accessed
+template isDeprecatedComplex(T)
+{
+    static if (__traits(isDeprecated, T))
```
wilzbach: Requires https://github.com/dlang/dmd/pull/7665

std/complex.d
```diff
@@ -874,16 +876,61 @@ Complex!real expi(real y)  @trusted pure nothrow @nogc
...
 
+/**
+    Params: y = A real number.
+    Returns: The value of cosh(y) + i sinh(y)
+
+    Note:
+    $(D coshisinh) is included here for convenience and for easy migration of code
+    that uses $(REF _coshisinh, std,math).
+*/
+Complex!real coshisinh(real y) @safe pure nothrow @nogc
```
wilzbach: This is part of https://github.com/dlang/phobos/pull/5698

wilzbach:
So it was quite to tricky to get Phobos to build, but thanks to https://github.com/dlang/dmd/pull/7647
we can now "poison" deprecate functions and the diff got a lot cleaner.

-------

BTW it was quite tricky to get `abs` working as every usage of the complex type will rightfully trigger a deprecation warning, even hacks like this:

```d
pragma(msg, creal.mangleof);
```
https://run.dlang.io/is/iNPtWM

However, there's the new `isDeprecated` traits which is exactly intended for such purposes. It just needs to be made aware of the complex types: https://github.com/dlang/dmd/pull/7665

wilzbach:
So while this is slowly falling into the abyss of the universe, I completed the quest of finding the unittests using complex or imaginary numbers. It turns out, there aren't too many, but as some are in `std.format` it still lead to very long deprecation messages.

Anyhow we can now use `-transition=complex` on Phobos for everything (build, unittest) and once this is merged should be able to finally turn the switch at DMD :)

However, while doing this I hit two more bugs:
- https://issues.dlang.org/show_bug.cgi?id=18250
- https://issues.dlang.org/show_bug.cgi?id=18249

The problem is that deprecated works poorly with templates, so it doesn't mark a template as deprecated when instantiated:

```d
auto foo(T)(T t) // <- deprecation message
{
    return T.init;
}

deprecated unittest
{
    auto b = foo(2 + 2i);
}
``` 
// 18250

and more annoyingly even does this if constraint evaluates to false.

```
auto foo(T)(T t) // <- deprecation message
if (!__traits(isDeprecated, T))
{
    return T.init;
}

deprecated
auto foo(T)(T t)
if (__traits(isDeprecated, T))
{
    return T.init;
}

deprecated unittest
{
    auto b = foo(2 + 2i);
}
```

The underlying problem is that the `DSymbolSemanticVisitor` just checks its in the `VarDeclaration` overload whether the type of the currently visited `Var` is complex/imaginary:

https://github.com/dlang/dmd/blob/11eda055563457560ff2510048805880ed8d8c28/src/dmd/dsymbolsem.d#L385

std/format.d
```diff
@@ -2604,6 +2605,8 @@ if (is(Unqual!T : creal) && !is(T == enum) && !hasToString!(T, Char))
     put(w, 'i');
 }
 
+version(TestComplex)
```
wilzbach: https://issues.dlang.org/show_bug.cgi?id=18250 - I didn't know a better way to workaround it.

std/format.d
```diff
@@ -2621,6 +2624,8 @@ if (is(Unqual!T : creal) && !is(T == enum) && !hasToString!(T, Char))
     }
 }
 
+version(TestComplex)
```
wilzbach: https://issues.dlang.org/show_bug.cgi?id=18250 - I didn't know a better way to workaround it.

std/math.d
```diff
@@ -543,20 +563,22 @@ if ((is(typeof(Num.init >= 0)) && is(typeof(-Num.init)) ||
     }
 }
 
-/// ditto
-auto abs(Num)(Num z) @safe pure nothrow @nogc
-if (is(Num* : const(cfloat*)) || is(Num* : const(cdouble*))
-    || is(Num* : const(creal*)))
+import std.meta : AliasSeq;
+deprecated("Please use std.complex")
+static foreach (Num; AliasSeq!(cfloat, cdouble, creal, ifloat, idouble, ireal))
```
wilzbach: This is pretty ugly, but my attempt at working around https://issues.dlang.org/show_bug.cgi?id=18251

wilzbach:
Meh. https://github.com/dlang/druntime/pull/2047 is needed first :/

9il:
Please do not deprecate native complex numbers!

std.complex does not work with LDC optimisation flags. I would need to copy std.complex to mir, mark with appropriate LDC flags, add my own conditional compilation flags, and add this informations to few readmes.

Native complex numbers are easy to use, they do not require Phobos and they works as _native_ . This means that `a + b` is not a function. And this is very important for LDC optimization flags. I do not want to copy phobos module again :-(

Complex numbers can be implemented as library but LDC should inline this function in front-end, not it LLVM back-end. Otherwise they would be much slooooower. Proved in GLAS&

andralex:
@9il I'm interested in learning more about this. Can't compilers inline `a + b` as well as a builtin would do?

ibuclaw:
@9il - slow down. Say one thing at a time, and be concise. Your arguments either make no sense or are conflicting.

thewilsonator:
@ibuclaw I get what @9il is saying

He means that `std.complex` is not annotated with LLVM optimisation hints (inlining, fast math). This leads to either duplication, or performance penalties and a dependance on phobos. Neither of which is desirable.

@andralex it should be possible to hook up the phobos symbols to the current code generating machinery, but this doesn't solve the dependance on phobos.

The simplest option would be to annotate phobos (with `version(LDC)` or in LDC's copy of phobos) with optimisation hints and since everything except `expi` is a template that shouldn't introduce a dependance on phobos if it is made to be a template.

The only remaining issue is the use of `real` (e.g. in `std.complex.sqrt`), which is very undesirable when speed is of greatest concern.

andralex:
First order of business (assuming we can implement complex as a lib) seems to move fundamental complex support into druntime. Code using complex numbers should only depend on core support.

>The only remaining issue is the use of real (e.g. in std.complex.sqrt), which is very undesirable when speed is of greatest concern.

This should be addressed separately, anyone working on this?

thewilsonator:
>First order of business (assuming we can implement complex as a lib) seems to move fundamental complex support into druntime. Code using complex numbers should only depend on core support.

It will need to work with `-betterC` (i.e. not depend on `libdruntime.a`)to have any hope of making Ilya happy. But assuming that all of the basic operations are `pragma(inline, true)`, I don't think this will be a problem. 
> This should be addressed separately, anyone working on this?

This shouldn't be hard, `s/real/T/g` should do the trick. The question is how to transition the behaviour  properly.

9il:
> @9il I'm interested in learning more about this. Can't compilers inline a + b as well as a builtin would do?
> 
>  Revert View details

Plz, take a look
https://godbolt.org/g/xvEpNH

9il:
> dependance on phobos

The Phobos dependency is not a big problem. The problem that I can not modify phobos module for a target compiler like i can do with a dub package.

9il:
> It will need to work with -betterC (i.e. not depend on libdruntime.a)to have any hope of making Ilya happy. But assuming that all of the basic operations are pragma(inline, true), I don't think this will be a problem.

`std.math` that is used in `std.complex` is not betterC

thewilsonator:
>Plz, take a look
https://godbolt.org/g/xvEpNH

I see only two calls to std.math.fabs in the entire dump. I was kinda expecting more.

9il:
> Plz, take a look
> https://godbolt.org/g/xvEpNH
> 
> I see only two calls to std.math.fabs in the entire dump. I was kinda expecting more.

Yes, I used only basic arithmetic ops. The problem is huge performance degradation.

thewilsonator:
>The Phobos dependency is not a big problem. The problem that I can not modify phobos module for a target compiler like i can do with a dub package.

If this is to become `core.complex` as Andrei suggested  then I guarantee that LDC will allow you to optimise the LDC druntime implementation for as may arch's as you need to.

wilzbach:
> Also druntime is not allowed to use phobos so its betterC issues will not apply.

Do we need to move it to `core`? We can also test that certain modules in Phobos are `-betterC`, see e.g.

https://github.com/dlang/phobos/pull/5952

9il:
> If this is to become core.complex as Andrei suggested then I guarantee that LDC will allow you to optimise the LDC druntime implementation for as may arch's as you need to.
> 
> Also druntime is not allowed to use phobos so its betterC issues will not apply.

Hmm, as you know `fastmath` is a set of some llvm flags. In practice we may need to mark function with one of other flags. But what flags should be choose `core./std.complex`? Should it be always `fastmath`? I should it be always not `@fastmath`? With native complex number we do not need to care about what kind of flags has library implementation. Native complex number inherit flags from the current function

thewilsonator:
@andralex We should be able to have almost full backwards compatibility if we cannibalise the builtin names, the only thing we lose is the ability to have e.g. 0.0 + 0.0i as a valid initialiser (would become `cdouble(0.0, 0.0)`, I'm sure we can hack the lever/parser in the meantime.) and require people to `import core.complex;` everything else should be able to be kept the same.

9il:
> Do we need to move it to core? We can also test that certain modules in Phobos are -betterC, see e.g.
> 
> #5952

I do not think this is an issue. see https://github.com/dlang/phobos/pull/6014#issuecomment-358856178

thewilsonator:
> In practice we may need to mark function with one of other flags. But what flags should be choose core./std.complex? Should it be always fastmath? I should it be always not @fastmath? With native complex number we do not need to care about what kind of flags has library implementation. Native complex number inherit flags from the current function

Perhaps we should add an `@inherit` attribute for inline functions?

9il:
> Perhaps we should add an @inherit attribute for inline functions?

Yes!!! But plus `__traits(llvmAttributes)` to chose algorithm depend on flags!
This would be awesome !

9il:
> Perhaps we should add an @inherit attribute for inline functions?
> 
> Yes!!! But plus __traits(llvmAttributes) to chose algorithm depend on flags!
> This would be awesome !

This will allow solve all problems I have with std.math and std.complex. And I will be happy to remove `mir.math.common` then

thewilsonator:
>But plus `__traits(llvmAttributes)` to chose algorithm depend on flags!

Please elaborate, I'm not sure what this would do.

9il:
> But plus __traits(llvmAttributes) to chose algorithm depend on flags!
> 
> Please elaborate, I'm not sure what this would do.

`__traits(llvmAttributes)` returns an alias sequence of LLVM specific attributes that function has.

```d
@fastmath auto foo(double a); { return bar(a); }
@inherit auto bar(double a) {
   // this trivial case can be done by LLVM optimizer,
   // but helpful for other tricky math hacks
    static if (AllowsReciprocal!(__traits(llvmAttributes)))
        return PI * 1 / a; // one algo
    else
        return PI / a; // another algo
}

template AllowsReciprocal(Attrs...) { ... }
```

thewilsonator:
That won't work because the body of `bar` must be independent of it's inlining into `foo` i.e. that static if has no knowledge of the inlining. You could template bar on a given set of attributes.

9il:
@thewilsonator can it be just a template without args? E. g. the attributes are special hiden template Park.

wilzbach:
> We should be able to have almost full backwards compatibility

How would `core.complex` (i.e. `-betterC` `complex`) work?
`std.complex`
- has nice formatting (requires `std.format`)
- uses a lot of things from `std.math` which aren't `-betterC`

Wouldn't it then be just a copy of `std.complex` or in your words:

> I would need to copy std.complex to mir, mark with appropriate LDC flags

thewilsonator:
Formating not so sure, any reason this can't be done one `std.format` as is done now?
Excluding the trig and `^^` and friends the only use of `std.math` is `fabs`.

thewilsonator:
@9il I don't think so.

ibuclaw:
> He means that std.complex is not annotated with LLVM optimisation hints (inlining, fast math). This leads to either duplication, or performance penalties and a dependance on phobos. Neither of which is desirable.

Neither are std.math or any other functions that use complex annotated with compiler specific attributes -> not a valid argument.

ibuclaw:
> Perhaps we should add an @inherit attribute for inline functions? Edit: or make the ldc.attributes template inheritable like the builtin attributes.

If it's not part of the language, then I can't see it going in. Likewise fastmath, which means that code generated is no longer IEEE compliant.

9il:
> Neither are std.math or any other functions that use complex annotated with compiler specific attributes -> not a valid argument

There are no D libs for math and algorithms, but Phobos?

Entire mir-algorithm is annotated. I do not  whant to argue why it is fine for IEEE, you can do experiments with LDC to ensure that IEEE is preserved if a lambda is not annotated.

9il:
> If it's not part of the language, then I can't see it going in. Likewise fastmath, which means that code generated is no longer IEEE compliant.

It is part for the best D compiler for math and finance.

stefan-koch-sociomantic:
@9il `AllowsReciprocal` ? 

when would that ever not be allowed ?
b/c of precision issues ?

9il:
@stefan-koch-sociomantic div operation can be optimised in fastmath to mul and reciprocal. This optimization is not allowed by default because of IEEE

ibuclaw:
> Entire mir-algorithm is annotated. I do not whant to argue why it is fine for IEEE

You're free to do that if you value speed over correctness.

Talking about annotating functions with non-standard attributes is not up for debate though.

ibuclaw:
> It is part of the best D compiler for math and finance.

The best compiler for math and finance is gdc (I say that with a grin, but the point stands - there is no best).

(incidentally one of the Berlin meet up talks did a live demo talk benchmarking math performance between dmd, ldc, gdc - the winner was gdc by a landslide, the loser was ldc but the author was surprised and didn't know why).

stefan-koch-sociomantic:
gdc's frontend does a better job then ldc does.
also the gcc optimizers are capable of performing rewrites in more cases.
llvm's development is mostly driven by clang which emits rather specific patterns.
and in the optimizers only those patterns get recognized.

9il:
@ibuclaw this is your opinion. Mir-algorithm does not break IEEE, probably you do not understand how this attributes affects lambdas.

I do not debate. I just showed how the prs kill perfomance. I will be happy with my own library

ibuclaw:
@9il I'll start by removing fastmath from your examples and seeing what the real difference is.

thewilsonator:
>>Perhaps we should add an @inherit attribute for inline functions? Edit: or make the ldc.attributes template inheritable like the builtin attributes.
>If it's not part of the language, then I can't see it going in. Likewise fastmath, which means that code generated is no longer IEEE compliant.

@ibuclaw Quite the opposite.`@inherit` has nothing to do directly with IEEE conformance except propagating attributes of a function to functions inlined into it e.g.
 ```
@infer void bar(...) { ... } // <-not @fastmath
@fastmath void foo(...)
{
    bar(...); // inlined into foo. foo's inline copy of bar now @fastmath
}
```
 If there becomes a `core.complex` (and probably even if it stays as `std.complex`) LDC's implementation may add that attribute.

The question is if there is benefit in standardising it across compilers for propagating optimisation directives.

ibuclaw:
> The question is if there is benefit in standardising it across compilers for propagating optimisation directives.

You could start with a DIP. ☺️

thewilsonator:
Well would GDC benefit from this? DMD doesn't have much in the way of perf tweaking attributes.

ibuclaw:
Try playing around with gdc yourself, and see if there might be: explore.dgnu.org

The module you're looking for is `gcc.attribute`, and the uda you should test is `@attribute("optimize", "-fsome-optimizer-flag")`. 

From what you describe though, I'm not sure.  I would expect all inlined functions to be applied the same optimization flags as the function they are inlined into without the need to be explicit about it.

thewilsonator:
Well in LDC the IEEE compliance (or lack thereof) is encoded into the function such that inlining will result in some instructions being `@fastmath` and some not. Maybe it is done differently in GCC?

ibuclaw:
In any case. Any solution or discussion needs to revolve around compiler-agnostic tests for performance and attributes. Anything else is out of scope.

We could introduce `core.complex`, but for what purpose? There is no complex in the language. So it will end up being a copy of std.complex.

ibuclaw:
> Well in LDC the IEEE compliance (or lack thereof) is encoded into the function such that inlining will result in some instructions being @fastmath and some not. Maybe it is done differently in GCC?

I couldn't say off the top of my head. If anything it probably depends entirely on what order optimization passes are done in. I.e: whether or not inlining is done before or after reordering.

I'm pretty certain that optimisation attributes are applied only at a function level and not a block level. So I don't think ldc's problem really applies for gdc. But as I said please test this! (I'm on my phone and can't for the moment)

PR #6016, opened 2018-01-08
Fix Issue 18214 - TemplateOf should return void for non-templated symbols

wilzbach:
See also: http://forum.dlang.org/post/bpnfmtqafvasbehzhnom@forum.dlang.org
I also bumped into this a while ago.
While the `is` operator should be preferred anyways, there's no reason we can't
be nice people and return `void` if a basic type was given.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -4367,6 +4367,7 @@ template MemberFunctionsTuple(C, string name)
 
 /**
 Returns an alias to the template that $(D T) is an instance of.
+It will return `void` if a symbol without a template is given.
  */
 template TemplateOf(alias T : Base!Args, alias Base, Args...)
 {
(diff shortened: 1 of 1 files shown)


JackStouffer:
Out of curiosity, what's the current behavior?

wilzbach:
```d
import std.traits;
pragma(msg, TemplateOf!(int[]));
```

```
onlineapp.d(5): Error: template std.traits.TemplateOf does not match any template declaration
onlineapp.d(5):        while evaluating pragma(msg, TemplateOf!(int[]))
```

https://run.dlang.io/is/2hubRP

MetaLang:
Thanks @wilzbach. This needs to be done for a lot of the stuff in `std.traits`.

PR #5596, opened 2017-07-11
Add substitute for lazy replacements

wilzbach:
As I realized in https://github.com/dlang/phobos/pull/5565 that Phobos has no way to have a lazy range of replacements, I ported @nordlow's `substitute` from [phobos-next](https://github.com/nordlow/phobos-next/blob/1992896e945111aa3f47eb94a42dcc57be4cc61b/src/substitution.d).

`substitute` allows lazy replacements:
--------------------------------------------------

```d
import std.algorithm.comparison : equal;
import std.algorithm.searching : substitute;

// substitute single elements
assert("do_it".substitute('_', ' ').equal("do it"));

// substitute multiple, single elements
assert("do_it".substitute('_', ' ',
                           'd', 'g',
                           'i', 't',
                           't', 'o')
              .equal("go to"));

// substitute subranges
assert("do_it".substitute("_", " ",
                          "do", "done")
              .equal("done it"));

// substitution works for any ElementType
int[] x = [1, 2, 3];
auto y = x.substitute(1, 0.1);
assert(y.equal([0.1, 2, 3]));
static assert(is(typeof(y.front) == double));
```

If the substitution parameters are known at Compile-Time, the faster
template overload can be used:

```d
import std.algorithm.comparison : equal;
import std.algorithm.searching : substitute;

// substitute subranges of a range
assert("apple_tree".substitute!("apple", "banana",
                                "tree", "shrub").equal("banana_shrub"));

// substitute elements in a range
assert("apple_tree".substitute!('a', 'b',
                                't', 'f').equal("bpple_free"));

// substitute values
assert('a'.substitute!('a', 'b', 't', 'f') == 'b');
```

PR #6017, opened 2018-01-09
Fix Issue 18215 - std.array.replace throws a range violation if the from range is longer than the array

wilzbach:
Maybe I'm just too tired, but how is it possible that there wasn't a test for replacing on an array of arrays?

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -2124,6 +2124,7 @@ if (isDynamicArray!(E[]) && isForwardRange!R1 && isForwardRange!R2
         && (hasLength!R2 || isSomeString!R2))
 {
     import std.algorithm.searching : find;
+    import std.range : dropOne;
 
     if (from.empty) return subject;
 
@@ -2134,7 +2135,11 @@ if (isDynamicArray!(E[]) && isForwardRange!R1 && isForwardRange!R2
     auto app = appender!(E[])();
     app.put(subject[0 .. subject.length - balance.length]);
     app.put(to.save);
-    replaceInto(app, balance[from.length .. $], from, to);
+    // replacing an element in an array is different to a range replacement
+    static if (is(Unqual!E : Unqual!R1))
+        replaceInto(app, balance.dropOne, from, to);
+    else
+        replaceInto(app, balance[from.length .. $], from, to);
 
     return app.data;
 }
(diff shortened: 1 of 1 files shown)


std/array.d
```diff
@@ -2146,13 +2151,26 @@ if (isDynamicArray!(E[]) && isForwardRange!R1 && isForwardRange!R2
...
+@safe unittest
+{
+    auto arr = ["aaa.dd", "b"];
+    arr = arr.replace("aaa.dd", ".");
+    assert(arr == [".", "b"]);
+
+    arr = ["_", "_", "aaa.dd", "b", "c", "aaa.dd", "e"];
+    arr = arr.replace("aaa.dd", ".");
+    assert(arr == ["_", "_", ".", "b", "c", ".", "e"]);
+}
```
quickfur: Is there already a unittest that covers the case `"abababc".replace("abcdef","g")`? I.e., the second branch of the static if.

wilzbach: Yeah that's the original case and covered by the existing tests. Expand the diff above.

image

BTW you might want to install the CodeCov browser extension: https://github.com/codecov/browser-extension

wilzbach: I still added more tests ;-)

quickfur: Yeah I was confused by your change to use `min(from.length, $)`, which you just reverted. I suppose that was not necessary? :-P

std/array.d
```diff
@@ -2146,13 +2151,42 @@ if (isDynamicArray!(E[]) && isForwardRange!R1 && isForwardRange!R2
...
+
+// https://issues.dlang.org/show_bug.cgi?id=18215
+@safe unittest
+{
+    assert([[0], [1, 2], [0], [3]].replace([0], [4]) == [[4], [1, 2], [4], [3]]);
+    assert([[0], [1, 2], [0], [3], [1, 2]].replace([1, 2], [0]) == [[0], [0], [0], [3], [0]]);
```
quickfur: Nitpick: This line is a little too long.

wilzbach:
I just found https://issues.dlang.org/show_bug.cgi?id=10930 in Bugzilla:
It's about:

```d
[1, 2, 3].replace(2, 4);
```

With this fix, it will be easy to fix the issue as follows:

```d
diff --git a/std/array.d b/std/array.d
index 3a44612cb..65e161504 100644
--- a/std/array.d
+++ b/std/array.d
@@ -2120,21 +2120,28 @@ if (isInputRange!RoR &&
         $(REF map, std,algorithm,iteration) which can act as a lazy replace
  +/
 E[] replace(E, R1, R2)(E[] subject, R1 from, R2 to)
-if (isDynamicArray!(E[]) && isForwardRange!R1 && isForwardRange!R2
-        && (hasLength!R2 || isSomeString!R2))
+if (isDynamicArray!(E[]))
 {
     import std.range : dropOne;
 
-    if (from.empty) return subject;
+    static if (isInputRange!R1)
+    {
+        if (from.empty) return subject;
+        alias rSave = a => a.save;
+    }
+    else
+    {
+        alias rSave = a => a;
+    }
 
-    auto balance = find(subject, from.save);
+    auto balance = find(subject, rSave(from));
     if (balance.empty)
         return subject;
 
     auto app = appender!(E[])();
     app.put(subject[0 .. subject.length - balance.length]);
-    app.put(to.save);
+    app.put(rSave(to));
     // replacing an element in an array is different to a range replacement
     static if (is(Unqual!E : Unqual!R1))
         replaceInto(app, balance.dropOne, from, to);
@@ -2158,6 +2165,11 @@ if (isDynamicArray!(E[]) && isForwardRange!R1 && isForwardRange!R2
     assert([3, 3, 4, 3].replace([3, 3], [1, 1, 1]) == [1, 1, 1, 4, 3]);
 }
 
+@safe unittest
+{
+    assert([0, 1, 2].replace(1, 4) == [0, 4, 2]);
+}
+
 // https://issues.dlang.org/show_bug.cgi?id=18215
 @safe unittest
 {
@@ -2181,28 +2193,34 @@ if (isDynamicArray!(E[]) && isForwardRange!R1 && isForwardRange!R2
 
 /// ditto
 void replaceInto(E, Sink, R1, R2)(Sink sink, E[] subject, R1 from, R2 to)
-if (isOutputRange!(Sink, E) && isDynamicArray!(E[])
-    && isForwardRange!R1 && isForwardRange!R2
-    && (hasLength!R2 || isSomeString!R2))
+if (isOutputRange!(Sink, E) && isDynamicArray!(E[]))
 {
     import std.range : dropOne;
 
-    if (from.empty)
+    static if (isInputRange!R1)
+    {
+        if (from.empty)
+        {
+            sink.put(subject);
+            return;
+        }
+        alias rSave = a => a.save;
+    }
+    else
     {
-        sink.put(subject);
-        return;
+        alias rSave = a => a;
     }
     for (;;)
     {
-        auto balance = find(subject, from.save);
+        auto balance = find(subject, rSave(from));
         if (balance.empty)
         {
             sink.put(subject);
             break;
         }
         sink.put(subject[0 .. subject.length - balance.length]);
-        sink.put(to.save);
+        sink.put(rSave(to));
         // replacing an element in an array is different to a range replacement
         static if (is(Unqual!E : Unqual!R1))
             subject = balance.dropOne;
```

Of course constraints need be set, but to be fair currently there aren't any for this (and probably `save` should be called if `R1` is a range
I would do this in another PR for `master` as this is enhancement request (the `replace` currently rejects this), whereas `replace` currently accepts RoR (ranges of ranges), but returns invalid output.
I'm aware that `[1, 2, 3].replace(2.only, 4.only)` or `[1, 2, 3].replace([2], [4])`, but this is such a common function that there's no reason why we shouldn't allow it.
Also there will be the lazy `substitute` soon: https://github.com/dlang/phobos/pull/5596

std/array.d
wilzbach: I split it into two lines.

quickfur:
Nice. Well, let's merge this first, then we can fix 10930 next.

PR #6020, opened 2018-01-09
Remove unused std.format.primitiveTypeInfo and deprecate unused std.format.Mangle

wilzbach:
`primitiveTypeInfo` is a relict from the old doFormat days.
It's not used anywhere anymore.

See also: https://github.com/dlang/phobos/pull/4978
  
Separated from: https://github.com/dlang/phobos/pull/6014

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -5375,7 +5375,8 @@ if (isInputRange!Range)
 
 
 // Legacy implementation
-
+// @@@DEPRECATED_2019-01@@@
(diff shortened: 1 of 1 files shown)


wilzbach:
> It's doubtful whether any user code would depend on this, but the deprecation should serve as fair warning.

I don't think anyone is using this either, but technically it's an undocumented, _public_ symbol, so :man_shrugging:

PR #6022, opened 2018-01-09
Fix Issue 10930 - std.array.replace cannot simple replace an element in array

wilzbach:
Follow-up to https://github.com/dlang/phobos/pull/6017

Currently there's no constraints that the ranges return similar types.
Should I add such a constraint too?
It would start to get a monster though:

```d
if (isOutputRange!(Sink, E) && isDynamicArray!(E[]) &&
    ((isForwardRange!R1 && isForwardRange!R2
    && (hasLength!R2 || isSomeString!R2))
    && is(typeof(R1.init.front) : typeof(R2.init.front)) ||
    is(Unqual!E : Unqual!R1)) && is(Unqual!R1 : Unqual!R2)))
```

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -2336,21 +2336,30 @@ if (isInputRange!RoR &&
         $(REF substitute, std,algorithm,iteration) for a lazy replace.
  +/
 E[] replace(E, R1, R2)(E[] subject, R1 from, R2 to)
-if (isDynamicArray!(E[]) && isForwardRange!R1 && isForwardRange!R2
-        && (hasLength!R2 || isSomeString!R2))
+if (isDynamicArray!(E[]) &&
+    ((isForwardRange!R1 && isForwardRange!R2 && (hasLength!R2 || isSomeString!R2)) ||
+    is(Unqual!E : Unqual!R1)))
 {
     import std.algorithm.searching : find;
     import std.range : dropOne;
 
-    if (from.empty) return subject;
+    static if (isInputRange!R1)
+    {
+        if (from.empty) return subject;
+        alias rSave = a => a.save;
+    }
+    else
+    {
+        alias rSave = a => a;
+    }
 
-    auto balance = find(subject, from.save);
+    auto balance = find(subject, rSave(from));
     if (balance.empty)
         return subject;
 
     auto app = appender!(E[])();
     app.put(subject[0 .. subject.length - balance.length]);
-    app.put(to.save);
+    app.put(rSave(to));
     // replacing an element in an array is different to a range replacement
     static if (is(Unqual!E : Unqual!R1))
         replaceInto(app, balance.dropOne, from, to);
@@ -2396,6 +2405,20 @@ if (isDynamicArray!(E[]) && isForwardRange!R1 && isForwardRange!R2
             .replace([[0], [1, 2]], [[4]]) == [[4], [0], [3], [1, 2], [4]]);
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=10930
+@safe unittest
+{
+    assert([0, 1, 2].replace(1, 4) == [0, 4, 2]);
+    assert("äbö".replace('ä', 'a') == "abö");
+}
+
+// empty array
+@safe unittest
+{
+    int[] arr;
+    assert(replace(arr, 1, 2) == []);
+}
+
 /++
     Replace occurrences of `from` with `to` in `subject` and output the result into
     `sink`.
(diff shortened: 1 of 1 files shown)


quickfur:
My opinion is that whenever a sig constraint grows out of hand and becomes monstrous, that's a sign that implementation details are leaking into the API.

Barring the rare case where whatever the function does is *actually* that complicated and requires a monstrous sig constraint to describe its parameters, I'd say user-facing sig constraints should *always* be readable, and short.  It boils down to asking yourself the question: "If somebody handed to me a black box that does what this function is supposed to do, what do I expect would be the argument types it should take?"

Usually the answer is simple: parameter 1 should be a range of some sort, for example, parameter 2 should be a range with compatible element type, etc.. Anything else that is currently in the sig constraint IMO doesn't belong there, but instead should go either into a `static if` inside the function body, or else be moved to a module-private overload set on a private implementation symbol (like `funcImpl`) that the user-facing function dispatches to.

wilzbach:
@JackStouffer why did you add "needs work" here? This is passing on all CIs and @quickfur's comment was of more general nature and in response to the huge constraint I proposed in the description.
FYI: @andralex recently pointed out that he isn't a huge friend of mixing `static assert` and template constraints when there's a good reason to do - see: https://github.com/dlang/phobos/pull/5596#discussion_r161896302

That being said when looking at the coverage annotation I found a red spot for which I just added a test.

quickfur:
I don't have a strong enough opinion about this to insist on changes, especially if @andralex prefers putting stuff in sig constraints instead of static ifs in the function body, but, looking at the new sig constraints again, I have to say I'm not a fan of it.

If I read it correctly (and that's a sign that it's too big, btw), `replace` now can take *either* a range in its `from` argument or a single element of the array being operated on.  To me, that just screams "new overload". IMO, it would be much better off if we explicitly created a separate overload that indicates clearly that the `from` argument can be a single element, e.g.:
````
E[] replace(E,R1,R2)(E[] arr, R1 from, R2 to) if (...) { ... }
E[] replace(E,E1,R)(E[] arr, E1 from, R to) if (...) { ... }
````
This makes it clear that we can either pass a range in the 2nd argument, or a single element.

The way we currently have it, `R1` can be either a range or a single element, which is very confusing to read, in addition to bloating the sig constraint (there was an effort some time ago to refactor Phobos sig constraints into conjunctive normal form, and this breaks that). Splitting it into two overloads that makes it clear exactly what is allowed is much clearer documentation-wise.

Now of course the implementation is mostly the same, so *internally*, we can still use the currently proposed code, with the above two new overloads simply redirecting to it. So, something like this:
````
E[] replace(E,R1,R2)(E[] arr, R1 from, R2 to)
if (isDynamicArray!(E[]) && isForwardRange!R1 && isForwardRange!R2)
{
    return replaceImpl(arr, from, to);
}

E1[] replace(E1,E2,R)(E1[] arr, E2 from, R to)
if (isDynamicArray!(E1[]) && is(Unqual!E1 : Unqual!E2) && isForwardRange!R)
{
    return replaceImpl(arr, from, to);
}

E[] replaceImpl(E,R1,R2)(E[] arr, R1 from, R2 to)
if (... /* current monstrous sig, optionally */)
{
    /* currently proposed code */
}
````

File: std/array.d [hunk 1]
```diff
@@ -2119,21 +2119,30 @@ if (isInputRange!RoR &&
         $(REF map, std,algorithm,iteration) which can act as a lazy replace
  +/
 E[] replace(E, R1, R2)(E[] subject, R1 from, R2 to)
-if (isDynamicArray!(E[]) && isForwardRange!R1 && isForwardRange!R2
-        && (hasLength!R2 || isSomeString!R2))
+if (isDynamicArray!(E[]) &&
```
andralex: Wait, `isDynamicArray!(E[])` is always true innit?

wilzbach:
I found another bug that this issue fixes: https://issues.dlang.org/show_bug.cgi?id=9702

wilzbach:
So the overload got shrunken down and is now:

```d
+if ((isForwardRange!R1 && isForwardRange!R2 && (hasLength!R2 || isSomeString!R2)) ||
       is(Unqual!E : Unqual!R1))
```

We have dozens of functions that use bigger constraints. Still required to split it into two functions? Or can we move forward here?

std/array.d
```diff
@@ -2119,21 +2119,29 @@ if (isInputRange!RoR &&
...
+    is(Unqual!E : Unqual!R1))
 {
     import std.algorithm.searching : find;
     import std.range : dropOne;
 
-    if (from.empty) return subject;
+    static if (isInputRange!R1)
+    {
+        if (from.empty) return subject;
+        alias rSave = a => a.save;
```
n8sh: What is the purpose of this when R1 and R2 are required to be forward ranges?

n8sh: My mistake, I misread the constraint!

wilzbach:
Rebased to master.

thewilsonator:
Dunno. Approved.

thewilsonator:
(Temporarily removing automerge due to OSX builder shortage)

PR #6023, opened 2018-01-10
Fix issue 13632: Enhancement to std.string.strip

aravindavk:
Added second argument similar to Python `str.strip`

Second argument accepts list of chars to strip and strips only
those chars.

Examples:

    "xyzhello".stripLeft("xyz") == ""hello"
    "helloxy ".stripRight("xy ") == "hello"
    "xhellox".strip("x") == "hello"

Signed-off-by: Aravinda VK <mail@aravindavk.in>

### DIFF
--- a/changelog/std-string-strip.dd
+++ b/changelog/std-string-strip.dd
@@ -1,0 +1,14 @@
+`std.string.strip` now accepts a string of characters to be stripped
+
+New overload functions of $(REF strip, std, string),
+$(REF stripLeft, std, string) and
+$(REF stripRight, std, string) accepts a string of characters to be
+stripped instead of stripping only whitespace.
+
+---
+import std.string: stripLeft, stripRight, strip;
+
+assert(stripLeft("www.dlang.org", "w.") == "dlang.org");
+assert(stripRight("dlang.org/", "/") == "dlang.org");
+assert(strip("www.dlang.org/", "w./") == "dlang.org");
+---
--- a/std/string.d
+++ b/std/string.d
@@ -2846,13 +2846,16 @@ if (isSomeChar!C)
 }
 
 /++
-    Strips leading whitespace (as defined by $(REF isWhite, std,uni)).
+    Strips leading whitespace (as defined by $(REF isWhite, std,uni)) or
+    as specified in the second argument.
 
     Params:
         input = string or $(REF_ALTTEXT forward range, isForwardRange, std,range,primitives)
         of characters
+        chars = string of characters to be stripped
 
-    Returns: $(D input) stripped of leading whitespace.
+    Returns: $(D input) stripped of leading whitespace or characters
+    specified in the second argument.
 
     Postconditions: $(D input) and the returned value
     will share the same tail (see $(REF sameTail, std,array)).
@@ -2920,14 +2923,72 @@ if (isConvertibleToString!Range)
     assert(testAliasedString!stripLeft("  hello"));
 }
 
+/// Ditto
+auto stripLeft(Range, Char)(Range input, const(Char)[] chars)
+if (((isForwardRange!Range && isSomeChar!(ElementEncodingType!Range)) ||
+     isConvertibleToString!Range) && isSomeChar!Char)
+{
+    static if (isConvertibleToString!Range)
+        return stripLeft!(StringTypeOf!Range)(input, chars);
+    else
+    {
+        for (; !input.empty; input.popFront)
+        {
+            if (chars.indexOf(input.front) == -1)
+                break;
+        }
+        return input;
+    }
+}
+
+///
+@safe pure unittest
+{
+    assert(stripLeft("     hello world     ", " ") ==
+           "hello world     ");
+    assert(stripLeft("xxxxxhello world     ", "x") ==
+           "hello world     ");
+    assert(stripLeft("xxxyy    hello world     ", "xy ") ==
+           "hello world     ");
+}
+
+///
+@safe pure unittest
+{
+    import std.array : array;
+    import std.utf : byChar, byWchar, byDchar;
+
+    assert(stripLeft("  xxxyy hello world     "w.byChar, "xy ").array ==
+           "hello world     ");
+
+    assert(stripLeft("\u2028\u2020hello world\u2028"w.byWchar,
+                     "\u2028").array == "\u2020hello world\u2028");
+    assert(stripLeft("\U00010001hello world"w.byWchar, " ").array ==
+           "\U00010001hello world"w);
+    assert(stripLeft("\U00010001 xyhello world"d.byDchar,
+                     "\U00010001 xy").array == "hello world"d);
+
+    assert(stripLeft("\u2020hello"w, "\u2020"w) == "hello"w);
+    assert(stripLeft("\U00010001hello"d, "\U00010001"d) == "hello"d);
+    assert(stripLeft(" hello ", "") == " hello ");
+}
+
+@safe pure unittest
+{
+    assert(testAliasedString!stripLeft(" xyz  hello", "xyz "));
+}
+
 /++
-    Strips trailing whitespace (as defined by $(REF isWhite, std,uni)).
+    Strips trailing whitespace (as defined by $(REF isWhite, std,uni)) or
+    as specified in the second argument.
 
     Params:
         str = string or random access range of characters
+        chars = string of characters to be stripped
 
     Returns:
-        slice of $(D str) stripped of trailing whitespace.
+        slice of $(D str) stripped of trailing whitespace or characters
+        specified in the second argument.
 
     See_Also:
         Generic stripping on ranges: $(REF _stripRight, std, algorithm, mutation)
@@ -3080,16 +3141,73 @@ if (isConvertibleToString!Range)
     cast(void) stripRight(ws.byUTF!wchar).array;
 }
 
+/// Ditto
+auto stripRight(Range, Char)(Range str, const(Char)[] chars)
+if (((isBidirectionalRange!Range && isSomeChar!(ElementEncodingType!Range)) ||
+     isConvertibleToString!Range) && isSomeChar!Char)
+{
+    static if (isConvertibleToString!Range)
+        return stripRight!(StringTypeOf!Range)(str, chars);
+    else
+    {
+        for (; !str.empty; str.popBack)
+        {
+            if (chars.indexOf(str.back) == -1)
+                break;
+        }
+        return str;
+    }
+}
+
+///
+@safe pure
+unittest
+{
+    assert(stripRight("     hello world     ", "x") ==
+           "     hello world     ");
+    assert(stripRight("     hello world     ", " ") ==
+           "     hello world");
+    assert(stripRight("     hello worldxy     ", "xy ") ==
+           "     hello world");
+}
+
+@safe pure unittest
+{
+    assert(testAliasedString!stripRight("hello xyz  ", "xyz "));
+}
+
+@safe pure unittest
+{
+    import std.array : array;
+    import std.utf : byChar, byDchar, byUTF, byWchar;
+
+    assert(stripRight("     hello world  xyz   ".byChar,
+                      "xyz ").array == "     hello world");
+    assert(stripRight("\u2028hello world\u2020\u2028"w.byWchar,
+                      "\u2028").array == "\u2028hello world\u2020");
+    assert(stripRight("hello world\U00010001"w.byWchar,
+                      " ").array == "hello world\U00010001"w);
+    assert(stripRight("hello world\U00010001 xy"d.byDchar,
+                      "\U00010001 xy").array == "hello world"d);
+    assert(stripRight("hello\u2020"w, "\u2020"w) == "hello"w);
+    assert(stripRight("hello\U00010001"d, "\U00010001"d) == "hello"d);
+    assert(stripRight(" hello ", "") == " hello ");
+}
+
 
 /++
     Strips both leading and trailing whitespace (as defined by
-    $(REF isWhite, std,uni)).
+    $(REF isWhite, std,uni)) or as specified in the second argument.
 
     Params:
         str = string or random access range of characters
+        chars = string of characters to be stripped
+        leftChars = string of leading characters to be stripped
+        rightChars = string of trailing characters to be stripped
 
     Returns:
-        slice of $(D str) stripped of leading and trailing whitespace.
+        slice of $(D str) stripped of leading and trailing whitespace
+        or characters as specified in the second argument.
 
     See_Also:
         Generic stripping on ranges: $(REF _strip, std, algorithm, mutation)
@@ -3174,6 +3292,125 @@ if (isConvertibleToString!Range)
     wstring s = " ";
     assert(s.sameTail(s.stripLeft()));
     assert(s.sameHead(s.stripRight()));
+    });
+}
+
+/// Ditto
+auto strip(Range, Char)(Range str, const(Char)[] chars)
+if (((isBidirectionalRange!Range && isSomeChar!(ElementEncodingType!Range)) ||
+     isConvertibleToString!Range) && isSomeChar!Char)
+{
+    static if (isConvertibleToString!Range)
+        return strip!(StringTypeOf!Range)(str, chars);
+    else
+        return stripRight(stripLeft(str, chars), chars);
+}
+
+///
+@safe pure unittest
+{
+    assert(strip("     hello world     ", "x") ==
+           "     hello world     ");
+    assert(strip("     hello world     ", " ") ==
+           "hello world");
+    assert(strip("   xyxyhello worldxyxy     ", "xy ") ==
+           "hello world");
+    assert(strip("\u2020hello\u2020"w, "\u2020"w) == "hello"w);
+    assert(strip("\U00010001hello\U00010001"d, "\U00010001"d) == "hello"d);
+    assert(strip(" hello ", "") == " hello ");
+}
+
+@safe pure unittest
+{
+    assert(testAliasedString!strip("  xyz   hello world  xyz   ", "xyz "));
+}
+
+/// Ditto
+auto strip(Range, Char)(Range str, const(Char)[] leftChars, const(Char)[] rightChars)
+if (((isBidirectionalRange!Range && isSomeChar!(ElementEncodingType!Range)) ||
+     isConvertibleToString!Range) && isSomeChar!Char)
+{
+    static if (isConvertibleToString!Range)
+        return strip!(StringTypeOf!Range)(str, leftChars, rightChars);
+    else
+        return stripRight(stripLeft(str, leftChars), rightChars);
+}
+
+///
+@safe pure unittest
+{
+    assert(strip("xxhelloyy", "x", "y") == "hello");
+    assert(strip("   xyxyhello worldxyxyzz    ", "xy ", "xyz ") ==
+           "hello world");
+    assert(strip("\u2020hello\u2028"w, "\u2020"w, "\u2028"w) == "hello"w);
+    assert(strip("\U00010001hello\U00010002"d, "\U00010001"d, "\U00010002"d) ==
+           "hello"d);
+    assert(strip(" hello ", "", "") == " hello ");
+}
+
+@safe pure unittest
+{
+    assert(testAliasedString!strip("  xy   hello world  pq   ", "xy ", "pq "));
+}
+
+@safe pure unittest
+{
+    import std.algorithm.comparison : equal;
+    import std.conv : to;
+    import std.exception : assertCTFEable;
+
+    assertCTFEable!(
+    {
+    static foreach (S; AliasSeq!( char[], const  char[],  string,
+                          wchar[], const wchar[], wstring,
+                          dchar[], const dchar[], dstring))
+    {
+        assert(equal(stripLeft(to!S("  \tfoo\t "), "\t "), "foo\t "));
+        assert(equal(stripLeft(to!S("\u2008  foo\t \u2007"), "\u2008 "),
+                     "foo\t \u2007"));
+        assert(equal(stripLeft(to!S("\u0085 μ \u0085 \u00BB \r"), "\u0085 "),
+                     "μ \u0085 \u00BB \r"));
+        assert(equal(stripLeft(to!S("1"), " "), "1"));
+        assert(equal(stripLeft(to!S("\U0010FFFE"), " "), "\U0010FFFE"));
+        assert(equal(stripLeft(to!S(""), " "), ""));
+
+        assert(equal(stripRight(to!S("  foo\t "), "\t "), "  foo"));
+        assert(equal(stripRight(to!S("\u2008  foo\t \u2007"), "\u2007\t "),
+                     "\u2008  foo"));
+        assert(equal(stripRight(to!S("\u0085 μ \u0085 \u00BB \r"), "\r "),
+                     "\u0085 μ \u0085 \u00BB"));
+        assert(equal(stripRight(to!S("1"), " "), "1"));
+        assert(equal(stripRight(to!S("\U0010FFFE"), " "), "\U0010FFFE"));
+        assert(equal(stripRight(to!S(""), " "), ""));
+
+        assert(equal(strip(to!S("  foo\t "), "\t "), "foo"));
+        assert(equal(strip(to!S("\u2008  foo\t \u2007"), "\u2008\u2007\t "),
+                     "foo"));
+        assert(equal(strip(to!S("\u0085 μ \u0085 \u00BB \r"), "\u0085\r "),
+                     "μ \u0085 \u00BB"));
+        assert(equal(strip(to!S("\U0010FFFE"), " "), "\U0010FFFE"));
+        assert(equal(strip(to!S(""), " "), ""));
+
+        assert(equal(strip(to!S("  \nfoo\t "), "\n ", "\t "), "foo"));
+        assert(equal(strip(to!S("\u2008\n  foo\t \u2007"),
+                           "\u2008\n ", "\u2007\t "), "foo"));
+        assert(equal(strip(to!S("\u0085 μ \u0085 \u00BB μ \u00BB\r"),
+                           "\u0085 ", "\u00BB\r "), "μ \u0085 \u00BB μ"));
+        assert(equal(strip(to!S("\U0010FFFE"), " ", " "), "\U0010FFFE"));
+        assert(equal(strip(to!S(""), " ", " "), ""));
+    }
+    });
+}
+
+@safe pure unittest
+{
+    import std.array : sameHead, sameTail;
+    import std.exception : assertCTFEable;
+    assertCTFEable!(
+    {
+    wstring s = " xyz ";
+    assert(s.sameTail(s.stripLeft(" ")));
+    assert(s.sameHead(s.stripRight(" ")));
     });
 }
 


std/string.d
```diff
@@ -2840,15 +2841,17 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
         auto c = input.front;
         if (std.ascii.isASCII(c))
         {
-            if (!std.ascii.isWhite(c))
+            if ((chars == "" && !std.ascii.isWhite(c)) ||
```
wilzbach: `is null` is the more idiomatic D way to check for empty strings.

aravindavk: Thanks for review, I will update the patch to use null and refresh the patch.

jmdavis: `is null` is most definitely _not_ the way to check for empty strings. That's how you check that a string is actually `null`. A `null` string is empty, but an empty string isn't necessarily null. The idiomatic way to check for empty is the `empty` function.

std/string.d
```diff
@@ -2827,7 +2828,7 @@ if (isConvertibleToString!Range)
     See_Also:
         Generic stripping on ranges: $(REF _stripLeft, std, algorithm, mutation)
   +/
-auto stripLeft(Range)(Range input)
+auto stripLeft(Range)(Range input, string chars = "")
```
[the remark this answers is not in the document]
aravindavk: Makes sense.  I will work on changing this as overloaded function as you suggested. Thanks for the review

std/string.d
quickfur: Yes, please don't use null to check for empty string, as it will introduce subtle bugs where empty strings with a non-null `.ptr` will not get recognized as empty.  Either use `.empty` or do an explicit length check `chars.length == 0`.

burner:
I'm pretty sure that is not @jmdavis had in mind when he said overload.
This api is way to easy to use wrong.
Please, split it into two functions.

jmdavis:
Yeah, this is way too error-prone. The fact that you're providing a second function argument should be what determines whether the normal version of strip is used or not, not whether you passed a template argument _and_ passed a second function argument. The simplest way to deal with this is to just create two separate functions:

    auto stripLeft(Range)(Range input)
    if(...)
    {...}

    auto stripLeft(Range)(Range input, string chars)
    if(...)
    {...}

That way, the function signatures are clear, and the efficiency of the existing `stripLeft` is not affected by the new one.

aravindavk:
> Yeah, this is way too error-prone. The fact that you're providing a second function argument should be what determines whether the normal version of strip is used or not, not whether you passed a template argument and passed a second function argument. The simplest way to deal with this is to just create two separate functions:

Using second argument as compile time switch didn't worked since value should be known during compile time.

Without duplicating the function content I couldn't achieve two overloaded functions. Template argument is used in the intermediate function, both `strip(Range)` and `strip(Range, chars)` calls that function.

    auto stripLeft(Range, bool useCustomChars)(Range input, string chars)

    auto stripLeft(Range)(Range input)
    if(...)
    {
        return stripLeft(Range, false)(input)
    }

    auto stripLeft(Range)(Range input, string chars)
    if(...)
    {
         return stripLeft(Range, true)(input, chars)
    }

First function is internal function, I have to explore how it can be made private. Other two functions are exposed which doesn't use template parameter.

Thanks.

jmdavis:
> Without duplicating the function content I couldn't achieve two overloaded functions.

Just create two completely separate functions. It's simpler. e.g.

    auto stripLeft(R)(R range)
    if(...)
    {
        ...
    }

    auto stripLeft(R)(R range, string chars)
    if(...)
    {
        ...
    }

wilzbach:
> Without duplicating the function content I couldn't achieve two overloaded functions.

To avoid duplication, you could use a templated `stripLeftImpl`:

```d
auto stripLeft(R)(R range)
if(...)
{
    stripLeftImpl!(Yes.useChars)(range, chars);
}

auto stripLeft(R)(R range, string chars)
if(...)
{
    stripLeftImpl!(No.useChars)(range, chars);
}

private stripLeftImpl(Flag!"UseChars" useChars, R, Args...)(R range, Args args)
{
   ... code ...
   static if (useChars)
   {
       ...
   }
}
```

jmdavis:
> To avoid duplication, you could use a templated stripLeftImpl:

If you go that route, the flag is pointless. The number of arguments gives you the information you need. There is zero need to pass any kind of template argument indicating which version of the function to use.

jmdavis:
Looking over these again, I definitely vote for separate functions. Each overload should be doing fundamentally different things, and honestly, the fact that you've just swapped out `isWhite` for a call to `indexOf` implies that you don't really understand what the implementations of `stripLeft` and `stripRight` are doing.

With the new overload, for `stripLeft`, it's going to need to decode each character and compare it against the characters in `chars`. There's no point in the checks for `isASCII` and all of that, not without getting into checking that _all_ of the characters in `chars` are ASCII and then having different branches depending on that information.

And with `stripRight`, that big long set of if-elses has to do with looking for Unicode whitespace efficiently and has _nothing_ to do with `chars`

Just create completely separate overloads for these versions of `stripLeft` and `stripRight`, and don't just copy the existing ones and tweak them. How would you write these two functions if you weren't concerned about the existing overloads? Write their implementations as if the current ones didn't exist, and we can see what that would look like. The way that you've tried to tweak the existing ones really doesn't make sense. All of the fancy stuff in there is to make dealing with checking for whitespace more efficient and doesn't belong in overloads that aren't stripping whitespace.

aravindavk:
Thanks @jmdavis for detailed review, will work on the suggested changes and refresh the patch by today.

aravindavk:
Updated the code as suggested. Please review.

aravindavk:
Build is failing in CI, looks like unable to find `libdparse`,

    + make -f posix.mak -j2 publictests DUB=dub BUILD=debug
    make -C ../druntime -f posix.mak MODEL=64 DMD=../dmd/generated/linux/debug/64/dmd OS=linux BUILD=debug
    mkdir -p generated/linux/debug/64/publictests || exists generated/linux/debug/64/publictests
    make[1]: Entering directory '/home/circleci/druntime'
    touch generated/linux/debug/64/publictests/.directory
    make[1]: Nothing to be done for 'target'.
    make[1]: Leaving directory '/home/circleci/druntime'
    DFLAGS="-conf= -I../druntime/import  -w -de -dip25 -m64 -fPIC -g -debug generated/linux/debug/64/libphobos2.a -defaultlib= -debuglib= -L-ldl" dub build --force --compiler=${PWD}/../dmd/generated/linux/debug/64/dmd --single ../tools/tests_extractor.d
    Root package tests_extractor references unknown package libdparse
    posix.mak:610: recipe for target 'generated/linux/debug/64/tests_extractor' failed
    make: *** [generated/linux/debug/64/tests_extractor] Error 2
    rm generated/linux/debug/64/publictests/.directory
    Exited with code 2

std/string.d
```diff
@@ -2870,10 +2900,24 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
     assert(stripLeft([paraSep] ~ "hello world" ~ paraSep) ==
            "hello world" ~ [paraSep]);
 
+    assert(stripLeft("     hello world     ", " ") ==
```
quickfur: Please put the new unittests in a separate unittest block, and also, please move the new `stripLeft` overload *after* the original unittest block. This is so that the generated documentation will make sense.

In other words, what you currently have is this:
````
/** ... */
auto stripLeft(...) ... // original stripLeft()

/** ... */
auto stripLeft(...) // new stripLeft

///
unittest
{
   ... // original test cases
   ... // new test cases
}
````

What it *should* be is this:

````
/** ... */
auto stripLeft(...) ... // original stripLeft

///
unittest { ... } // original test cases

/** ... */
auto stripLeft(...) ... // new stripLeft

///
unittest { ... } // new test cases
````

File: std/string.d [hunk 4]
```diff
@@ -2882,6 +2926,12 @@ if (isConvertibleToString!Range)
     return stripLeft!(StringTypeOf!Range)(str);
 }
 
+auto stripLeft(Range)(auto ref Range str, string chars)
```
quickfur: These two overloads also need to have the right order.  Here's what the order should be:
````
/** ... */
auto stripLeft(...) ... // original stripLeft

///
unittest { ... } // original test cases

/// ditto   <--- add this comment
auto stripLeft(Range)(auto ref Range str)
if (isConvertibleToString!Range) ...

///
unittest { ... } // more original test cases

/** ... */
auto stripLeft(...) ... // new stripLeft

///
unittest { ... } // new test cases

/// ditto   <--- add this comment
auto stripLeft(Range)(auto ref Range str, string chars)
if (isConvertibleToString!Range) ...

///
unittest { ... } // more new test cases (if there are any specific to the 2nd overload)
````

File: std/string.d [hunk 5]
```diff
@@ -3010,6 +3107,13 @@ unittest
            [lineSep] ~ "hello world");
     assert(stripRight([paraSep] ~ "hello world" ~ paraSep) ==
            [paraSep] ~ "hello world");
+
+    assert(stripRight("     hello world     ", "x") ==
+           "     hello world     ");
+    assert(stripRight("     hello world     ", " ") ==
+           "     hello world");
+    assert(stripRight("     hello worldxy     ", "xy ") ==
+           "     hello world");
```
quickfur: Same goes for `stripRight`.  The original overload + unittests should be left as-is in their original order, and the new `stripRight` should be added *after* the original unittest, and the new tests should follow, in their own unittest block.

std/string.d
```diff
@@ -3092,6 +3236,12 @@ if (isConvertibleToString!Range)
     return strip!(StringTypeOf!Range)(str);
 }
 
+auto strip(Range)(auto ref Range str, string chars)
+if (isConvertibleToString!Range)
+{
+    return strip!(StringTypeOf!Range)(str, chars);
+}
+
```
quickfur: Same goes for these overloads.  Keep the original single-argument overloads grouped together with their respective unittest blocks, followed by the new 2-argument overloads and the new unittests.

aravindavk:
@quickfur thanks for the review, I will make the suggested changes and refresh the patch.

aravindavk:
@quickfur updated the patch as suggested. Please review

aravindavk:
Hi Reviewers, Let me know if any more changes required to the patch. Thanks.

std/string.d
```diff
@@ -3047,6 +3109,91 @@ if (isConvertibleToString!Range)
...
+    Returns:
+        slice of $(D str) stripped of trailing extra characters.
+
+    See_Also:
+        Generic stripping on ranges: $(REF _stripRight, std, algorithm, mutation)
+  +/
+auto stripRight(Range)(Range str, string chars)
+if (isSomeString!Range ||
+    isRandomAccessRange!Range && hasLength!Range && hasSlicing!Range &&
```
quickfur: Technically, we don't need to require a random access range here.  It is enough to require a bidirectional range and use `popBack` to strip from the end, in a similar way to `stripLeft`. In that case, we also don't need to require `hasLength` and `hasSlicing`, though of course in the implementation we can take advantage of those attributes where available.

std/string.d
```diff
@@ -3144,6 +3291,104 @@ if (isConvertibleToString!Range)
...
+
+    Returns:
+        slice of $(D str) stripped of leading and trailing extra characters.
+
+    See_Also:
+        Generic stripping on ranges: $(REF _strip, std, algorithm, mutation)
+  +/
+auto strip(Range)(Range str, string chars)
+if (isSomeString!Range ||
+    isRandomAccessRange!Range && hasLength!Range && hasSlicing!Range &&
```
quickfur: ditto, we can relax the constraints here to just `isBidirectionalRange!Range`, and that should be sufficient to implement this function.

aravindavk:
@quickfur updated the code to use `isBidirectionalRange`, please review.

std/string.d
```diff
@@ -2887,6 +2887,68 @@ if (isConvertibleToString!Range)
...
+
+    Params:
+        input = string or $(REF_ALTTEXT forward range, isForwardRange, std,range,primitives)
+        of characters
+        chars = list of chars to strip
+
+    Returns: $(D input) stripped of extra characters.
+
+    Postconditions: $(D input) and the returned value
+    will share the same tail (see $(REF sameTail, std,array)).
```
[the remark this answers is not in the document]
wilzbach: Also you should use 'ditto' here and add the parameters to Ddoc header of the stripRight above 
This will allow Ddoc to group both functions.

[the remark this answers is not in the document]
aravindavk: > Not sure whether this is really necessary. We typically just state that it doesn't allocate.

I followed this similar to stringLeft above.

Is there a way to generate the documentation locally to verify before sending patch? (Sorry if it is documented elsewhere)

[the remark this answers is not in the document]
aravindavk: got it, testing documentation using `dmd -D std/string.d`, will refresh the patch with doc fixes. Thanks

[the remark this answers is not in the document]
wilzbach: You can also preview the documentation at DAutoTest here or of course clone dlang.org and build it yourself

aravindavk:
@wilzbach tested doc generation locally(`dmd -D std/string.d` and `firefox string.html`) and modified doc text as you suggested. Please review.

aravindavk:
Let me know if any changes required.

std/string.d
```diff
@@ -3144,6 +3252,89 @@ if (isConvertibleToString!Range)
...
+
+@safe pure unittest
+{
+    import std.array : sameHead, sameTail;
+    import std.exception : assertCTFEable;
+    assertCTFEable!(
+    {
+    wstring s = " xyz ";
+    assert(s.sameTail(s.stripLeft(" ")));
+    assert(s.sameHead(s.stripRight(" ")));
```
quickfur: Nitpick: please indent the body of the function literal here.

std/string.d
```diff
@@ -3144,6 +3252,89 @@ if (isConvertibleToString!Range)
...
+
+@safe pure unittest
+{
+    import std.algorithm.comparison : equal;
+    import std.conv : to;
+    import std.exception : assertCTFEable;
+
+    assertCTFEable!(
+    {
+    static foreach (S; AliasSeq!( char[], const  char[],  string,
```
quickfur: Come to think of it, this should also be indented too.

quickfur:
Actually, nevermind that, I see that you're just following the style in other parts of the same file. Can't say I'm a fan of this style, but since it's consistent, let's just roll with it.

JackStouffer:
Needs a changelog entry

aravindavk:
> Needs a changelog entry

added.

quickfur:
LGTM. Thanks!

quickfur:
ping @JackStouffer @wilzbach @burner @jmdavis 

Any more comments before I pull the trigger? ;-)

jmdavis:
Technically, I think that it requires @andralex's approval, because it's adding new public symbols.

quickfur:
ping @andralex

andralex:
Hm, I fear people might get confused. Consider: `strip("www.dlang.org", "www.")` whereby the user expects the `www.` to be stripped if present.

changelog/std-string-strip.dd
```diff
@@ -0,0 +1,14 @@
+`std.string.strip` now accepts a list of characters to be stripped
```
andralex: "list" is not the right word

changelog/std-string-strip.dd
```diff
@@ -0,0 +1,14 @@
...
+
+New overload functions of $(REF strip, std, string),
+$(REF stripLeft, std, string) and
+$(REF stripRight, std, string) accepts a list of characters to be
```
andralex: no list

std/string.d
```diff
@@ -2887,14 +2890,64 @@ if (isConvertibleToString!Range)
     assert(testAliasedString!stripLeft("  hello"));
 }
 
+/// Ditto
+auto stripLeft(Range)(Range input, string chars)
+if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
+    !isInfinite!Range && !isConvertibleToString!Range)
```
andralex: range can be infinite

std/string.d
```diff
@@ -3047,16 +3100,71 @@ if (isConvertibleToString!Range)
     cast(void) stripRight(ws.byUTF!wchar).array;
 }
 
+/// Ditto
+auto stripRight(Range)(Range str, string chars)
+if (isBidirectionalRange!Range && isSomeChar!(ElementEncodingType!Range) &&
+    !isInfinite!Range && !isConvertibleToString!Range)
```
andralex: bidirectional and infinite not very likely

std/string.d
```diff
@@ -3047,16 +3100,71 @@ if (isConvertibleToString!Range)
...
+{
+    while (!str.empty)
+    {
+        auto c = str.back;
+        if (indexOf(chars, c) == -1)
+            break;
+        str.popBack();
+    }
```
andralex: There's too much pomp and circumstance in this code. It's a simple for loop with no temporaries:

```d
for (; !str.empty; str.popBack)
{
    if (chars.indexOf(str.back) == -1)
        break;
}
```

std/string.d
```diff
@@ -3144,6 +3252,89 @@ if (isConvertibleToString!Range)
     });
 }
 
+/// Ditto
+auto strip(Range)(Range str, string chars)
+if (isBidirectionalRange!Range && isSomeChar!(ElementEncodingType!Range) &&
+    !isInfinite!Range && !isConvertibleToString!Range)
```
andralex: drop isInfinite

std/string.d
quickfur: Hold on. How do you strip characters to the *right* of an infinite range?!

E.g., if the infinite range consists of `aaabaaabaaa...` repeated forever, and you call `stripRight` to remove `a`s from the right, how would the code know when to start dropping `a`s? And if `stripRight` was asked to remove *both* `a`s and `b`s, how would the code know if there's a `z` 500 elements into the range so that it shouldn't just return an empty range?

std/string.d
quickfur: Again, how do you strip characters from the right of an infinite range?

std/string.d
andralex: The range has `back` and `popBack` and that seems to be all that's needed. It's difficult to define a range that is bidirectional and infinite.

std/string.d
andralex: With the code as is :)

std/string.d
quickfur: Ahh, I misunderstood that you wanted to remove *both* bidirectional and infinite from the constraints.  Removing isInfinite and leaving bidirectional makes sense. :-)

std/string.d
quickfur: Gotcha.

std/string.d
quickfur: P.S. It's not *that* hard to make an infinite range that's also bidirectional; the hard part is coming up with a sensible use case for it. :-P  The traditional infinite range is open to the right, i.e., it has no end. But it's also possible to have a range that's open in the middle, i.e., the front and back parts never meet. Such a range actually *would* work with this code, as is, if we drop `isInfinite`.  Though of course, I've a hard time imagining a real-world use case for this kind of bizarre infinite range!

aravindavk:
Thanks @andralex for the review, reworked on the patch as suggested.

Also changed the second argument from string to `const(Char)[]`(similar to other functions in the same file) and added unit tests for the same.

aravindavk:
> Hm, I fear people might get confused. Consider: strip("www.dlang.org", "www.") whereby the user expects the www. to be stripped if present.

Many languages provides the same API(Python, PHP, Golang). 

In Go, `TrimPrefix` and `TrimSuffix` is added along with `TrimLeft` and `TrimRight` to avoid this confusion. Equivalant in D is `std.string.chompPrefix` (I couldn't find `std.string.chompSuffix`)

aravindavk:
jenkins/pr-merge test failed with following error. Is there a way I can retrigger this test?

    [diet-ng] Running shell script
    + dub build -b release
    Performing "release" build using /var/lib/jenkins/dlang_projects/distribution/bin/dmd for x86_64.
    diet-ng 1.4.4: building configuration "library"...
    Error: Error writing file '.dub/build/library-release-linux.posix-x86_64-dmd_2078-47FD7784A9F78ED910EB095685955196/libdiet-ng.a'
    /var/lib/jenkins/dlang_projects/distribution/bin/dmd failed with exit code 1.
    script returned exit code 2

jmdavis:
> Is there a way I can retrigger this test?

It triggers on new commits. So, just do a `git commit --amend` with no changes to the message, and then `git push -f` to your github repo, and it'll rerun everything.

wilzbach:
> Is there a way I can retrigger this test?

Only people with write permissions can login at Jenkins -> I just restarted Jenkins for you.
You could do it yourself by changing your commit sha:

```bash
git commit --amend
git push -f
```

Note that we currently have a couple of issues with Jenkins. The one you saw is https://github.com/dlang/ci/issues/135
tl;dr: we are aware of the Jenkins failures and they won't block your PR.

aravindavk:
Thanks a lot @wilzbach @jmdavis

quickfur:
LGTM

aravindavk:
Thanks for all the review comments. Let me know if any change required.

@andralex please review the changes. (jenkins/pr-merge failure is not related to the patch https://github.com/dlang/phobos/pull/6023#issuecomment-360804197)

std/string.d
```diff
@@ -2887,14 +2890,69 @@ if (isConvertibleToString!Range)
...
+           "\U00010001hello world"w);
+    assert(stripLeft("\U00010001 xyhello world"d.byDchar,
+                     "\U00010001 xy").array == "hello world"d);
+
+    assert(stripLeft("\u2020hello"w, "\u2020"w) == "hello"w);
+    assert(stripLeft("\U00010001hello"d, "\U00010001"d) == "hello"d);
+    assert(stripLeft(" hello ", "") == " hello ");
+}
+
+auto stripLeft(Range, Char)(auto ref Range str, const(Char)[] chars)
```
andralex: Public and undocumented not good. Please merge constraints and bodies with the overload above.

std/string.d
```diff
@@ -3047,16 +3105,72 @@ if (isConvertibleToString!Range)
...
+{
+    assert(stripRight("     hello world     ", "x") ==
+           "     hello world     ");
+    assert(stripRight("     hello world     ", " ") ==
+           "     hello world");
+    assert(stripRight("     hello worldxy     ", "xy ") ==
+           "     hello world");
+}
+
+auto stripRight(Range, Char)(auto ref Range str, const(Char)[] chars)
```
andralex: please merge with the other overload

std/string.d
```diff
@@ -3144,6 +3258,92 @@ if (isConvertibleToString!Range)
...
+    assert(strip("     hello world     ", " ") ==
+           "hello world");
+    assert(strip("   xyxyhello worldxyxy     ", "xy ") ==
+           "hello world");
+    assert(strip("\u2020hello\u2020"w, "\u2020"w) == "hello"w);
+    assert(strip("\U00010001hello\U00010001"d, "\U00010001"d) == "hello"d);
+    assert(strip(" hello ", "") == " hello ");
+}
+
+auto strip(Range, Char)(auto ref Range str, const(Char)[] chars)
```
andralex: please merge with the other oveload

std/string.d
aravindavk: I followed similar to existing `stripLeft(Range)(auto ref Range str)`. Confused because none of the functions with `auto ref` input are documented in this file. Please clarify.

andralex: We pursue lifting the quality of code overall, and in the process of doing that it's possible that new code is subject to a higher bar than existing code. Two standards arose from experience:

* All public symbols must be documented
* Prefer fewer overloads

I understand this can be frustrating - please bear with us. Thanks!

aravindavk: No problem, happy to work on the changes, just needed a clarification. Thanks

aravindavk:
@andralex merged the overloads as suggested. Please review.

aravindavk:
ci/circleci is failing due to following error.

    posix.mak:599: recipe for target 'std/traits.publictests' failed
    make: *** [std/traits.publictests] Hangup
    make: *** Deleting intermediate file 'generated/linux/debug/64/publictests/.directory'
    Too long with no output (exceeded 10m0s)

Is this a known failure or do I need to do anything to fix this?

wilzbach:
Known

wilzbach:
> Known

And fixed: https://github.com/dlang/phobos/pull/6099

std/string.d
```diff
@@ -3080,16 +3137,71 @@ if (isConvertibleToString!Range)
...
+if (((isBidirectionalRange!Range && isSomeChar!(ElementEncodingType!Range)) ||
+     isConvertibleToString!Range) && isSomeChar!Char)
+{
+    static if (isConvertibleToString!Range)
+        return stripRight!(StringTypeOf!Range)(str, chars);
+    else
+    {
+        for (; !str.empty; str.popBack)
+        {
+            if (chars.indexOf(str.back) == -1)
```
wilzbach: Why don't you use `foreach` here?
In this case there's _any_ performance penalty of this abstraction.
https://github.com/dlang/phobos/pull/6092#issuecomment-361397749

std/string.d
```diff
@@ -2920,14 +2923,68 @@ if (isConvertibleToString!Range)
...
+/// Ditto
+auto stripLeft(Range, Char)(Range input, const(Char)[] chars)
+if (((isForwardRange!Range && isSomeChar!(ElementEncodingType!Range)) ||
+     isConvertibleToString!Range) && isSomeChar!Char)
+{
+    static if (isConvertibleToString!Range)
+        return stripLeft!(StringTypeOf!Range)(input, chars);
+    else
+    {
+        for (; !input.empty; input.popFront)
```
wilzbach: Why don't you use `foreach` here?
In this case there's _any_ performance penalty of this abstraction.
https://github.com/dlang/phobos/pull/6092#issuecomment-361397749

quickfur: I think the catch here is that we need to iterate *from the back*, not from the front. So we'll either have to use `foreach_reverse` (*ick*), or `std.range.retro`, which then would require passing the range in by reference, since we need to guarantee that the range is actually modified in order to return it to the user. So it'll end up being something ugly like this:
````
import std.range : retro;
foreach (ch; (&input).retro) { ... }
````
which is equally *ick*.

File: std/string.d (hunk 21 above)
Comment (quickfur) [reply std/string.d]: ditto

**Comment by aravindavk** [comment]
> And fixed: #6099 (I restarted your build on CircleCi)

Thanks @quickfur

File: std/string.d (hunk 22 above)
Comment (quickfur) [reply std/string.d]: Oh wait, what I said only applies to the `stripRight` variant.  Here we're moving forwards, so there's no `foreach_reverse` or `retro` needed.

I'd say the `for` loop here is justified (using foreach would require passing the range by ref, just in case, so it still entails the ugly `&range` syntax).  Perhaps a more self-documenting way to write this would be:
````
while (!input.empty && chars.indexOf(input.front) != -1)
    input.popFront;
````
which, at least to me, reads more intuitively as "while the range is not empty and the front element is a character to be stripped, pop the front element".

File: std/string.d (hunk 21 above)
Comment (quickfur) [reply std/string.d]: Here we're iterating from the back, so we'd need `foreach_reverse` or `retro` if you wanted to use an abstraction.

Or we could write this as:
````
while (!str.empty && chars.indexOf(str.back) != -1)
    str.popBack;
````
which to me reads more intuitively as "while the range is not empty and the back is a character to the stripped, pop the back".

File: std/string.d [hunk 23]
```diff
@@ -3177,6 +3289,89 @@ if (isConvertibleToString!Range)
     });
 }
 
+/// Ditto
+auto strip(Range, Char)(Range str, const(Char)[] chars)
```
andralex: An additional overload would take a third argument:

```d
auto strip(Range, Char)(Range str, const(Char)[] leftChars, const(Char)[] rightChars)
```

and would strip `leftChars` on the left and `rightChars` on the right.

aravindavk: Nice. I will add it

std/string.d
aravindavk: Shall I change this to use `while` as suggested?

andralex: Not important, but I like the `for` form. Kurz und gut.

std/string.d
aravindavk: Added the new overload as suggested and added the respective tests.

std/string.d
wilzbach: > Not important, but I like the for form. Kurz und gut.

Yes, I was referring to `foreach_reverse`:

```d
foreach_reverse (s; str)
    if (chars.indexOf(s) == -1)
        break;
```

FWIW LDC optimizes `foreach_reverse` a lot better:
 
https://godbolt.org/g/k9tr4S

```asm
_D7example3fooFAyaQdZv:
  sub rsp, 72
  mov qword ptr [rsp + 64], rcx
  mov qword ptr [rsp + 56], rdx
  mov qword ptr [rsp + 40], rdi
  mov qword ptr [rsp + 48], rsi
  mov rcx, qword ptr [rsp + 64]
  mov rdx, qword ptr [rsp + 56]
  mov qword ptr [rsp + 24], rdx
  mov qword ptr [rsp + 32], rcx
  mov rcx, qword ptr [rsp + 24]
  mov qword ptr [rsp + 16], rcx
.LBB0_1:
  mov rax, qword ptr [rsp + 16]
  mov rcx, rax
  sub rcx, 1
  mov qword ptr [rsp + 16], rcx
  cmp rax, 0
  je .LBB0_6
  mov eax, 1
  mov rcx, qword ptr [rsp + 16]
  mov rdx, qword ptr [rsp + 32]
  mov sil, byte ptr [rdx + rcx]
  mov byte ptr [rsp + 15], sil
  movzx esi, byte ptr [rsp + 15]
  mov rdx, qword ptr [rsp + 40]
  mov rcx, qword ptr [rsp + 48]
  mov edi, 1
  mov dword ptr [rsp + 8], eax
  call _D3std6string__T7indexOfTAyaZQnFNaNbNiNfQpxwxEQBs8typecons__T4FlagVQBqa13_6361736553656e736974697665ZQBoZl@PLT
  cmp rax, -1
  jne .LBB0_4
  jmp .LBB0_6
.LBB0_4:
  jmp .LBB0_5
.LBB0_5:
  jmp .LBB0_1
.LBB0_6:
  add rsp, 72
  ret
```

I hope you notice that it could inline `back` completely.

For comparison, the `for`:

```asm
_D7example3fooFAyaQdZv:
  push r15
  push r14
  push r12
  push rbx
  sub rsp, 24
  mov r14, rsi
  mov r12, rdi
  mov qword ptr [rsp + 8], rdx
  mov qword ptr [rsp + 16], rcx
  test rdx, rdx
  je .LBB0_9
  mov rdi, rdx
  mov rsi, rcx
  call _D3std5range10primitives__T4backTyaZQjFNaNdNfAyaZw@PLT
  mov edi, 1
  mov esi, eax
  mov rdx, r12
  mov rcx, r14
  call _D3std6string__T7indexOfTAyaZQnFNaNbNiNfQpxwxEQBs8typecons__T4FlagVQBqa13_6361736553656e736974697665ZQBoZl@PLT
  cmp rax, -1
  je .LBB0_9
  lea r15, [rsp + 8]
.LBB0_3:
  mov rbx, qword ptr [rsp + 8]
  test rbx, rbx
  je .LBB0_6
  mov rdi, rbx
  mov rsi, r15
  call _D3std3utf__T10strideBackTAyaZQrFNaNfKQmmZk@PLT
  mov eax, eax
  sub rbx, rax
  cmp rbx, qword ptr [rsp + 8]
  ja .LBB0_5
  mov qword ptr [rsp + 8], rbx
  test rbx, rbx
  je .LBB0_9
  mov rsi, qword ptr [rsp + 16]
  mov rdi, rbx
  call _D3std5range10primitives__T4backTyaZQjFNaNdNfAyaZw@PLT
  mov edi, 1
  mov esi, eax
  mov rdx, r12
  mov rcx, r14
  call _D3std6string__T7indexOfTAyaZQnFNaNbNiNfQpxwxEQBs8typecons__T4FlagVQBqa13_6361736553656e736974697665ZQBoZl@PLT
  cmp rax, -1
  jne .LBB0_3
.LBB0_9:
  add rsp, 24
  pop rbx
  pop r12
  pop r14
  pop r15
  ret
.LBB0_6:
  lea rsi, [rip + .L.str]
  lea rcx, [rip + .L.str.1]
  mov edi, 69
  mov edx, 92
  mov r8d, 2206
  call _d_assert_msg@PLT
.LBB0_5:
  lea rsi, [rip + .L.str.1]
  mov edi, 92
  mov edx, 2207
  call _d_arraybounds@PLT
```

 If we would have perfect abstraction, I would have suggested `str.retro.until!(s => chars.indexOf(s) > 0);`, but unfortunately that's - even for the LLVM optimizer - too much.

std/string.d
```diff
@@ -2920,14 +2923,68 @@ if (isConvertibleToString!Range)
...
+                break;
+        }
+        return input;
+    }
+}
+
+///
+@safe pure unittest
+{
+     assert(stripLeft("     hello world     ", " ") ==
```
wilzbach: Indentation

std/string.d
```diff
@@ -2920,14 +2923,68 @@ if (isConvertibleToString!Range)
...
+           "hello world     ");
+    assert(stripLeft("xxxxxhello world     ", "x") ==
+           "hello world     ");
+    assert(stripLeft("xxxyy    hello world     ", "xy ") ==
+           "hello world     ");
+
+    import std.array : array;
```
wilzbach: This is a good sign that it should be a separate unittest.

std/string.d
wilzbach: Well, it boils down to whether we want to do auto-decoding or not ;-)

std/string.d
aravindavk: fixed

std/string.d
aravindavk: done

aravindavk:
@andralex @quickfur @wilzbach is this patch looking good now. 

Thanks

quickfur:
Alright, I think this is good to go.  Any further fixes can be done in followup PRs. Merge ahoy!

aravindavk:
Feeling very happy to see my first contribution to D language :)

Contribution experience was very nice, review comments were very valid and helped me to understand the language better. Happy to contribute more to D language in future. Thanks a lot

ZombineDev:
Nice work @aravindavk, you made a solid contribution to the standard library!

wilzbach:
> Feeling very happy to see my first contribution to D language :)

BTW it's online on the changelog already:

https://dlang.org/changelog/pending.html#std-string-strip

> Happy to contribute more to D language in future.

:+1: 
I have sent out an invite to our Slack channel too (it's a good way to quickly ask questions) - you deserve d it!

Also, https://issues.dlang.org is open 24/7 :wink:

PR #6024, opened 2018-01-11
Fix Issue 18223: use memset in uninitializedFillDefault(T[])

n8sh:
Current function in std.experimental.allocator.package:
```d
private T[] uninitializedFillDefault(T)(T[] array) nothrow
{
    T t = T.init;
    fillWithMemcpy(array, t);
    return array;
}
```

When we can statically determine that the representation of T.init consists of nothing but zeroes we can instead use memset. char and wchar can also be special-cased.

### DIFF
--- a/std/experimental/allocator/common.d
+++ b/std/experimental/allocator/common.d
@@ -743,3 +743,327 @@ version(StdUnittest)
         }
     }
 }
+
+/+
+Can the representation be determined at compile time to consist of nothing but
+zero bits? Padding between a struct's fields is not considered.
++/
+private template isAllZeroBits(T, T value)
+{
+    static if (isDynamicArray!(typeof(value)))
+        enum isAllZeroBits = value is null && value.length == 0;
+    else static if (is(typeof(value is null)))
+        enum isAllZeroBits = value is null;
+    else static if (is(typeof(value is 0)))
+        enum isAllZeroBits = value is 0;
+    else static if (isStaticArray!(typeof(value)))
+        enum isAllZeroBits = ()
+        {
+            bool b = true;
+            // Use index so this works when T.length is 0.
+            static foreach (i; 0 .. T.length)
+            {
+                b &= isAllZeroBits!(typeof(value[i]), value[i]);
+                if (b == false) return b;
+            }
+
+            return b;
+        }();
+    else static if (is(typeof(value) == struct) || is(typeof(value) == union))
+        enum isAllZeroBits = ()
+        {
+            bool b = true;
+            static foreach (e; value.tupleof)
+            {
+                b &= isAllZeroBits!(typeof(e), e);
+                if (b == false) return b;
+            }
+
+            return b;
+        }();
+    else
+        enum isAllZeroBits = false;
+}
+
+@nogc nothrow pure @safe unittest
+{
+    import std.meta : AliasSeq;
+    static foreach (Int; AliasSeq!(bool, char, wchar, dchar, byte, ubyte,
+        short, ushort, int, uint, long, ulong))
+    {
+        static assert(isAllZeroBits!(Int, Int(0)));
+        static assert(!isAllZeroBits!(Int, Int(1)));
+    }
+
+    foreach (Float; AliasSeq!(float, double, real))
+    {
+        assert(isAllZeroBits!(Float, 0.0));
+        assert(!isAllZeroBits!(Float, -0.0));
+        assert(!isAllZeroBits!(Float, Float.nan));
+    }
+
+    static assert(isAllZeroBits!(void*, null));
+    static assert(isAllZeroBits!(int*, null));
+    static assert(isAllZeroBits!(Object, null));
+}
+
+/+
+Is the representation of T.init known at compile time to consist of nothing but
+zero bits? Padding between a struct's fields is not considered.
++/
+package template isInitAllZeroBits(T)
+{
+    static if (isStaticArray!T && __traits(compiles, T.init[0]))
+        enum isInitAllZeroBits = __traits(compiles, {
+            static assert(isAllZeroBits!(typeof(T.init[0]), T.init[0]));
+        });
+    else
+        enum isInitAllZeroBits = __traits(compiles, {
+            static assert(isAllZeroBits!(T, T.init));
+        });
+}
+
+@nogc nothrow pure @safe unittest
+{
+    static assert(isInitAllZeroBits!(Object));
+    static assert(isInitAllZeroBits!(void*));
+    static assert(isInitAllZeroBits!uint);
+    static assert(isInitAllZeroBits!(uint[2]));
+
+    static assert(!isInitAllZeroBits!float);
+    static assert(isInitAllZeroBits!(float[0]));
+    static assert(!isInitAllZeroBits!(float[2]));
+
+    static struct S1
+    {
+        int a;
+    }
+    static assert(isInitAllZeroBits!S1);
+
+    static struct S2
+    {
+        int a = 1;
+    }
+    static assert(!isInitAllZeroBits!S2);
+
+    static struct S3
+    {
+        S1 a;
+        int b;
+    }
+    static assert(isInitAllZeroBits!S3);
+    static assert(isInitAllZeroBits!(S3[2]));
+
+    static struct S4
+    {
+        S1 a;
+        S2 b;
+    }
+    static assert(!isInitAllZeroBits!S4);
+
+    static struct S5
+    {
+        real r = 0;
+    }
+    static assert(isInitAllZeroBits!S5);
+
+    static struct S6
+    {
+
+    }
+    static assert(isInitAllZeroBits!S6);
+
+    static struct S7
+    {
+        float[0] a;
+    }
+    static assert(isInitAllZeroBits!S7);
+
+    static class C1
+    {
+        int a = 1;
+    }
+    static assert(isInitAllZeroBits!C1);
+
+    // Ensure Tuple can be read.
+    import std.typecons : Tuple;
+    static assert(isInitAllZeroBits!(Tuple!(int, int)));
+    static assert(!isInitAllZeroBits!(Tuple!(float, float)));
+
+    // Ensure private fields of structs from other modules
+    // are taken into account.
+    import std.random : Mt19937;
+    static assert(!isInitAllZeroBits!Mt19937);
+    // Check that it works with const.
+    static assert(isInitAllZeroBits!(const(Mt19937)) == isInitAllZeroBits!Mt19937);
+    static assert(isInitAllZeroBits!(const(S5)) == isInitAllZeroBits!S5);
+}
+
+/+
+Can the representation be determined at compile time to consist of nothing but
+1 bits? This is reported as $(B false) for structs with padding between
+their fields because `opEquals` and hashing may rely on those bits being
+zero.
+
+Note:
+A bool occupies 8 bits so `isAllOneBits!(bool, true) == false`
+
+See_Also:
+https://forum.dlang.org/post/hn11oh$1usk$1@digitalmars.com
++/
+private template isAllOneBits(T, T value)
+{
+    static if (isIntegral!T || isSomeChar!T)
+    {
+        import core.bitop : popcnt;
+        static if (T.min < T(0))
+            enum isAllOneBits = popcnt(cast(Unsigned!T) value) == T.sizeof * 8;
+        else
+            enum isAllOneBits = popcnt(value) == T.sizeof * 8;
+    }
+    else static if (isStaticArray!(typeof(value)))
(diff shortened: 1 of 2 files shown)


std/experimental/allocator/common.d
```diff
@@ -743,3 +743,135 @@ version(unittest)
         }
     }
 }
+
+/+
+Can the representation be determined at compile time to consist of nothing but
+zero bits? Padding between a struct's fields is not considered.
++/
+private template isAllZeroBits(T, T value)
+{
```
n8sh: Is there a more elegant way than this to check if T.init is entirely zero bits? `typeid(T).initializer() is null` isn't available.

std/experimental/allocator/common.d
```diff
@@ -743,3 +743,135 @@ version(unittest)
...
+            {
+                enum N = typeof(value).length;
+                char[] buffer = new char[32 * value.length]; // Pre-size for value.length <= 999.
+                buffer.length = 0;
+                buffer ~= "isAllZeroBits!(E,value[0])";
+                import std.conv: to;
+                foreach (i; 1 .. N)
+                    buffer ~= " && isAllZeroBits!(E,value["~to!string(i)~"])";
+                return buffer;
+            }());
```
n8sh: Is there a more elegant way to check all the elements of a static array?

std/experimental/allocator/common.d
```diff
@@ -743,3 +743,135 @@ version(unittest)
...
+            {
+                if (buffer.length != 0)
+                    buffer ~= " && isAllZeroBits!(typeof(value."~fieldName
+                        ~"), value."~fieldName~")";
+                else
+                    buffer ~= "isAllZeroBits!(typeof(value."~fieldName
+                        ~"), value."~fieldName~")";
+            }
+            return buffer;
+        }());
```
n8sh: Is there a more elegant way to check all the fields of a struct?

std/experimental/allocator/common.d
```diff
@@ -743,3 +743,135 @@ version(unittest)
...
+                enum N = typeof(value).length;
+                char[] buffer = new char[32 * value.length]; // Pre-size for value.length <= 999.
+                buffer.length = 0;
+                buffer ~= "isAllZeroBits!(E,value[0])";
+                import std.conv : to;
+                foreach (i; 1 .. N)
+                    buffer ~= " && isAllZeroBits!(E,value["~to!string(i)~"])";
+                return buffer;
+            }());
+        }
```
n8sh: Is there a more elegant way to check all the elements of a static array?

std/experimental/allocator/common.d
```diff
@@ -743,3 +743,135 @@ version(unittest)
...
+    {
+        static if (value.length == 0)
+            enum isAllZeroBits = true;
+        else
+        {
+            alias E = typeof(value[0]);
+            enum isAllZeroBits = mixin(()
+            {
+                enum N = typeof(value).length;
+                char[] buffer = new char[32 * value.length]; // Pre-size for value.length <= 999.
```
DmitryOlshansky: This doesn't reserve the size. On append the slice is not the tail so it wil reallocate.

n8sh: Thanks for the correction! `reserve` isn't available at compile time so I'll just remove this.

n8sh:
Is there an escape hatch to bypass read-access protection during CTFE? Being unable to inspect private members of structs greatly reduces the applicability. Alternatively is there a better way to check at compile time if a struct's representation is entirely zero?

std/experimental/allocator/common.d
```diff
@@ -743,3 +743,147 @@ version(unittest)
...
+            {
+                string n = to!string(i);
+                if (buffer.length != 0)
+                    buffer ~= " && ";
+                buffer ~= "isAllZeroBits!(typeof(value.tupleof["~n~"]),value.tupleof["~n~"])";
+            }
+            if (buffer.length == 0)
+                buffer ~= "true";
+            return buffer;
+        }());
```
wilzbach: How about:

```d
static if (value.tupleof.length == 0)
{
    enum isAllZeroBits = true;
}
else
{
    enum isAllZeroBits = ()
    {
        int b;
        static foreach (e; value.tupleof)
            b += !isAllZeroBits!(typeof(e), e);

        return b == 0;
    }();
}
```

std/experimental/allocator/common.d
wilzbach: And I think we can do even better here:

```d
enum isAllZeroBits = isAllZeroBits!(typeof(value[0]), value[0]);
```

Static arrays need to have the same type, don't they?

std/experimental/allocator/common.d
```diff
@@ -743,3 +743,147 @@ version(unittest)
...
+        else
+        {
+            alias E = typeof(value[0]);
+            enum isAllZeroBits = mixin(()
+            {
+                enum N = typeof(value).length;
+                char[] buffer;
+                buffer ~= "isAllZeroBits!(E,value[0])";
+                import std.conv : to;
+                foreach (i; 1 .. N)
```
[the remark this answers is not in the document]
n8sh: Because it could have been either I avoided it due to [seeing this advice on the forums](http://forum.dlang.org/post/mailman.1994.1515288062.9493.digitalmars-d-learn@puremagic.com):
>When you can use the old style do so. Since it puts less stress on the compiler in the general case.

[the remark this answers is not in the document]
wilzbach: That's bad advice. There isn't any measurable performance penality - I tried a large-scale test: https://github.com/dlang/phobos/pull/5989

std/experimental/allocator/common.d
n8sh: That works only when the static array is the outermost type and not a field of another struct . That optimization could go in `isInitAllZeroBits` though.

std/experimental/allocator/package.d
```diff
@@ -922,9 +922,26 @@ private void fillWithMemcpy(T)(void[] array, auto ref T filler) nothrow
...
+        if (array !is null)
+            memset(array.ptr, 0, T.sizeof * array.length);
+        return array;
+    }
+    else static if (is(T : char) || (is(T : wchar) && T.init == 0xffff))
+    {
+        import core.stdc.string : memset;
+        if (array !is null)
+            memset(array.ptr, cast(ubyte) T.init, T.sizeof * array.length);
+        return array;
```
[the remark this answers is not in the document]
n8sh: Added a unittest.

std/experimental/allocator/package.d
```diff
@@ -922,9 +922,26 @@ private void fillWithMemcpy(T)(void[] array, auto ref T filler) nothrow
...
-    fillWithMemcpy(array, t);
-    return array;
+    static if (isInitAllZeroBits!T)
+    {
+        import core.stdc.string : memset;
+        if (array !is null)
+            memset(array.ptr, 0, T.sizeof * array.length);
+        return array;
+    }
+    else static if (is(T : char) || (is(T : wchar) && T.init == 0xffff))
```
wilzbach: 1) What's the reason for excluding `dchar` here?

2) Shouldn't the type classifiers be removed? E.g.
- [`Unqual`]((https://dlang.org/phobos/std_traits.html#Unqual)
- [isSomeChar](https://dlang.org/phobos/std_traits.html#isSomeChar) - doesn't allow `alias this`

3) Also the `T.init == 0xffff` check is to support user-extended versions of (char) that don't overwrite the `init` value, right? This should probably get a test too.

std/experimental/allocator/common.d
```diff
@@ -743,3 +743,138 @@ version(unittest)
...
+
+    // Ensure Tuple can be read.
+    import std.typecons : Tuple;
+    static assert(isInitAllZeroBits!(Tuple!(int, int)));
+    static assert(!isInitAllZeroBits!(Tuple!(float, float)));
+
+    // Has non-zero private fields.
+    import std.random : Mt19937;
+    static assert(!isInitAllZeroBits!Mt19937);
+}
```
wilzbach: It might be worthwhile to check whether this works with `const`.
[`ConstOf`](https://dlang.org/phobos/std_traits.html#ConstOf) might be helpful.

std/experimental/allocator/package.d
```diff
@@ -943,6 +960,24 @@ pure nothrow @nogc
...
+}
+
+@system unittest
+{
+    static struct P { float x = 0; float y = 0; }
+
+    static assert(isInitAllZeroBits!P);
+    P[] a = [P(10, 11), P(20, 21), P(40, 41)];
+    uninitializedFillDefault(a);
+    assert(a == [P.init, P.init, P.init]);
```
wilzbach: Needs tests for `const`, e.g

```d
static struct S { float x = 0; float y = 0; }
static foreach (e; ["mutable", "non-mutable"])
{{

    static if (e == "mutable")
        alias P = S;
    else
        alias P = ConstOf!S;

    static assert(isInitAllZeroBits!P);
    P[] a = [P(10, 11), P(20, 21), P(40, 41)];
    uninitializedFillDefault(a);
    assert(a == [P.init, P.init, P.init]);
}}
```

fails with:

```
std/experimental/allocator/package.d(929): Error: function core.stdc.string.memset(return scope void* s, int c, ulong n) is not callable using argument types (const(S)*, int, ulong)
std/experimental/allocator/package.d(994): Error: template instance std.experimental.allocator.uninitializedFillDefault!(const(S)) error instantiating
```

std/experimental/allocator/common.d
```diff
@@ -743,3 +743,138 @@ version(unittest)
...
+
+/+
+Is the representation of T.init known at compile time to consist of nothing but
+zero bits? Padding between a struct's fields is not considered.
++/
+package template isInitAllZeroBits(T)
+{
+    import std.traits : isStaticArray;
+    static if (isStaticArray!T)
```
ZombineDev: I would use a more generic trait: `static if (T.sizeof == 0)`.

std/experimental/allocator/package.d
n8sh: 1. `dchar.init == 0x0000_ffff`. Can't use memset.

2. `static assert(is(immutable(char) : char))` compiles. I don't know that Unqual is necessary. I hadn't given much though to `alias this`. What do you suggest?

3. Since `char.sizeof == 1` memset will work regardless of what char.init is. Since `wchar.sizeof == 2` memset will work if the high 8 bits are the same as the low 8 bits; the main purpose of the `T.init == 0xffff` check was reminding the casual reader that is is true for wchar. Maybe`0 == T.init ^ (T.init >>> 8)` would be better?

std/experimental/allocator/package.d
n8sh: Shouldn't trying to write to `const` without casting const-ness away result in compilation failure?

n8sh:
@wilzbach I've changed the code around to avoid issues with `alias this` and the awkward special casing of char. Now instead of special casing specific types I check `isInitAllOneBits!T` so the code is symmetrical with the all zero bits case.

n8sh:
@DmitryOlshansky I've removed the erroneous code you mentioned in https://github.com/dlang/phobos/pull/6024#discussion_r160918451. Does the pull request look okay to you now?

JackStouffer:
Ping @DmitryOlshansky

n8sh:
Bump

apz28:
Improvement codes, loop should not be continue if isAllZeroBits return false.

+            static foreach (i; 0 .. T.length)
+                b &= isAllZeroBits!(typeof(value[i]), value[i]);

n8sh:
@apz28 The problem is because I am using `static foreach` if I return before the end the compiler will complain about unreachable code.

n8sh:
@apz28 Actually I've found a way that compiles.

n8sh:
Ping @wilzbach @JackStouffer @ZombineDev @DmitryOlshansky.

wilzbach:
Putting this on the merge queue (if no objections are raised in the next three days, this can be merged).

wilzbach:
BTW @n8sh we have this semi-official Slack channel that we often for some quick discussion. I couldn't find your email address, so if you want to join, just send me an email ;-)

std/experimental/allocator/common.d
```diff
@@ -743,3 +743,307 @@ version(StdUnittest)
...
+        enum isAllOneBits = false;
+    }
+}
+
+@nogc nothrow pure @safe unittest
+{
+    static assert(isAllOneBits!(char, 0xff));
+    static assert(isAllOneBits!(wchar, 0xffff));
+    static assert(isAllOneBits!(byte, cast(byte) 0xff));
+    static assert(isAllOneBits!(int, 0xffff_ffff));
```
nordlow: I would prefer to see tests for all integer types here just to be sure.

PR #6025, opened 2018-01-11
Fix issue 18224 - BigInt modulo uint must return long.

markus-oberhumer:


std/bigint.d
```diff
@@ -479,6 +479,7 @@ public:
 
         // BigInt % long => long
         // BigInt % ulong => BigInt
+        // BigInt % uint => long
```
wilzbach: What about the documentation table above?

std/bigint.d
```diff
@@ -494,6 +495,14 @@ public:
                 return r;
             }
         }
+        else static if (is(Unqual!T == uint))
+        {
+            immutable uint u = absUnsign(y);
+            long rem = BigUint.modInt(data, u);
```
wilzbach: Why can't you join this with the else clause below?
AFAICT the only thing that is different is the type of `rem`:

```d
static if (static if (is(Unqual!T == uint))
    alias T = long;
else
    alias T = int;
```

markus-oberhumer:
I've created a new version of this pull request to address the reviews above.

markus-oberhumer:
Thanks!

PR #6026, opened 2018-01-11
Optimized format with compile time format strings

JackStouffer:
Added function to std.format to make an educated guess at the length of the resulting string when it's given at compile time, and use that data to reserve capacity in the appender before anything is written to it.

Performance wise, I only saw differences outside of random-ness in longer format strings.

```
old 1			1 sec, 223 ms, 469 μs, and 8 hnsecs
old 2			2 secs, 427 ms, 447 μs, and 1 hnsec
old 3			2 secs, 448 ms, 949 μs, and 8 hnsecs
old 4			2 secs, 655 ms, 333 μs, and 4 hnsecs

new 1			1 sec, 191 ms, 30 μs, and 2 hnsecs
new 2			2 secs, 432 ms, 636 μs, and 8 hnsecs
new 3			2 secs, 489 ms, 238 μs, and 6 hnsecs
new 4			2 secs, 155 ms, 551 μs, and 3 hnsecs
```

```d
import std.stdio;
import std.algorithm;
import std.conv;
import std.format;
import std.range;
import std.traits;
import std.string;
import std.utf;

enum testCount = 6_000_000;

typeof(fmt) format1(alias fmt, Args...)(Args args)
if (isSomeString!(typeof(fmt)))
{
    alias e = checkFormatException!(fmt, Args);
    static assert(!e, e.msg);
    return format(fmt, args);
}

typeof(fmt) format2(alias fmt, Args...)(Args args)
if (isSomeString!(typeof(fmt)))
{
    import std.array : appender;

    alias e = checkFormatException!(fmt, Args);
    alias Char = Unqual!(ElementEncodingType!(typeof(fmt)));

    static assert(!e, e.msg);
    auto w = appender!(immutable(Char)[]);

    // no need to traverse the string twice during compile time
    if (!__ctfe)
    {
        enum len = guessLength!(fmt, Char)();
        w.reserve(len);
    }

    formattedWrite(w, fmt, args);
    return w.data;
}

private size_t guessLength(alias fmtString, Char)()
{
    import std.array : appender;
    pragma(msg, fmtString);

    size_t len;
    auto output = appender!(immutable(Char)[])();
    auto spec = FormatSpec!Char(fmtString);
    while (spec.writeUpToNextSpec(output))
    {
        // take a guess
        if (spec.width == 0 && (spec.precision == spec.UNSPECIFIED || spec.precision == spec.DYNAMIC))
        {
            if (spec.spec == 's')
                len += 6;
            else if (spec.spec == 'd' || spec.spec == 'x' || spec.spec == 'x')
                len += 3;
            else if (spec.spec == 'b')
                len += 8;
            else if (spec.spec == 'e' || spec.spec == 'E')
                len += 12;
            else if (spec.spec == 'f' || spec.spec == 'F')
                len += 10;
            else if (spec.spec == 'c')
                ++len;

            continue;
        }

        if ((spec.spec == 'e' || spec.spec == 'E') && spec.precision != spec.UNSPECIFIED && spec.precision == spec.DYNAMIC)
        {
            len += spec.precision + 5;
            continue;
        }

        if (spec.width == spec.precision)
            len += spec.width;
        else if (spec.width > 0 && (spec.precision == spec.UNSPECIFIED || spec.width > spec.precision))
            len += spec.width;
        else if (spec.precision != spec.UNSPECIFIED && spec.precision > spec.width)
            len += spec.precision;
    }
    len += output.data.length;
    return len;
}

void main()
{
    string res;

    auto result1 = to!Duration(benchmark!(() => res = format1!"%d"(42))(testCount)[0]);
    auto result2 = to!Duration(benchmark!(() => res = format1!"%e"(4.2))(testCount)[0]);
    auto result3 = to!Duration(benchmark!(() => res = format1!"%02d%02d%02d"(10, 30, 50))(testCount)[0]);
    auto result4 = to!Duration(benchmark!(() => res = format1!"Error code %d: longer format string to represent real world work (%s)"(10, 3000))(testCount)[0]);

    writeln("old 1", "\t\t\t", result1);
    writeln("old 2", "\t\t\t", result2);
    writeln("old 3", "\t\t\t", result3);
    writeln("old 4", "\t\t\t", result4);
    writeln;

    result1 = to!Duration(benchmark!(() => res = format2!"%d"(42))(testCount)[0]);
    result2 = to!Duration(benchmark!(() => res = format2!"%e"(4.2))(testCount)[0]);
    result3 = to!Duration(benchmark!(() => res = format2!"%02d%02d%02d"(10, 30, 50))(testCount)[0]);
    result4 = to!Duration(benchmark!(() => res = format2!"Error code %d: longer format string to represent real world work (%s)"(10, 3000))(testCount)[0]);

    writeln("new 1", "\t\t\t", result1);
    writeln("new 2", "\t\t\t", result2);
    writeln("new 3", "\t\t\t", result3);
    writeln("new 4", "\t\t\t", result4);
    writeln;
}
```

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -5771,18 +5771,44 @@ package static const checkFormatException(alias fmt, Args...) =
     return null;
 }();
 
-/*****************************************************
+/**
  * Format arguments into a string.
+ *
+ * If the format string is fixed, passing it as a template parameter checks the
+ * type correctness of the parameters at compile-time. This also can result in
+ * better performance.
  *
  * Params: fmt  = Format string. For detailed specification, see $(LREF formattedWrite).
  *         args = Variadic list of arguments to _format into returned string.
+ *
+ * Throws:
+ *     $(LREF, FormatException) if the number of arguments doesn't match the number
+ *     of format parameters and vice-versa.
  */
 typeof(fmt) format(alias fmt, Args...)(Args args)
 if (isSomeString!(typeof(fmt)))
 {
+    import std.array : appender;
+
     alias e = checkFormatException!(fmt, Args);
+    alias Char = Unqual!(ElementEncodingType!(typeof(fmt)));
+
     static assert(!e, e.msg);
-    return .format(fmt, args);
+    auto w = appender!(immutable(Char)[]);
+
+    // no need to traverse the string twice during compile time
+    if (!__ctfe)
+    {
+        enum len = guessLength!Char(fmt);
+        w.reserve(len);
+    }
+    else
+    {
+        w.reserve(fmt.length);
+    }
+
+    formattedWrite(w, fmt, args);
+    return w.data;
 }
 
 /// Type checking can be done when fmt is known at compile-time:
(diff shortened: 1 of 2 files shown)


quickfur:
Very interesting.

Just out of curiosity, what heuristic did you use for the various guessed length values for each %-spec?

JackStouffer:
For strings I just choose 12 because it seemed reasonable.

For numbers, basically what I did is saw how many chars the number 999 took and used that.

std/format.d
```diff
@@ -5796,6 +5823,54 @@ if (isSomeString!(typeof(fmt)))
...
+
+    size_t len;
+    auto output = appender!(immutable(Char)[])();
+    auto spec = FormatSpec!Char(fmtString);
+    while (spec.writeUpToNextSpec(output))
+    {
+        // take a guess
+        if (spec.width == 0 && (spec.precision == spec.UNSPECIFIED || spec.precision == spec.DYNAMIC))
+        {
+            if (spec.spec == 'd' || spec.spec == 'x' || spec.spec == 'x')
```
n8sh: I believe thee second `'x'` should be `'X'`.

JackStouffer:
Added changelog entry

std/format.d
```diff
@@ -5796,6 +5823,79 @@ if (isSomeString!(typeof(fmt)))
     static assert(!__traits(compiles, {s = format!"%d"(4.03);})); // incompatible arg
 }
 
+// called during compilation to guess the length of the
+// result of format
+private size_t guessLength(alias fmtString, Char)()
```
n8sh: In guessLength is there a reason to prefer `alias fmtString ` over making fmtString a function parameter? It seems to me that `guessLength(FmtStringType, Char)(FmtStringType fmtString)` would result in fewer template instantiations.

quickfur: That's a good point. We don't need `guessLength` to be re-instantiated per format string. Since it's going to be run in CTFE anyway, making the format string a "runtime" parameter should work just as well, and likely faster / less memory hogging, given the way dmd currently handles templates.

JackStouffer:
Done

std/format.d
```diff
@@ -5771,18 +5771,45 @@ package static const checkFormatException(alias fmt, Args...) =
...
+    // no need to traverse the string twice during compile time
+    if (!__ctfe)
+    {
+        enum len = guessLength!Char(fmt);
+        w.reserve(len);
+    }
+    else
+    {
+        enum len = fmt.length;
+        w.reserve(len);
```
quickfur: Hmm, just out of curiosity. Why not use `guessLength` at runtime too?  I think runtime performance could benefit from not having to reallocate as frequently.

JackStouffer:
Going ahead and merging this.

std/format.d
```diff
@@ -5796,6 +5823,79 @@ if (isSomeString!(typeof(fmt)))
...
+    len += output.data.length;
+    return len;
+}
+
+@safe pure nothrow unittest
+{
+    static assert(guessLength!char("%02d") == 2);
+    static assert(guessLength!char("%02d:%02d:%02d") == 8);
+    static assert(guessLength!char("%s") == 12);
+    static assert(guessLength!char("%0.2f") == 7);
```
wilzbach: I thought that PRs which decrease the coverage need a good justification to do so?

image

-> Evaluate the tests at runtime, s.t. the coverage can be counted and we can see whether its fully covered.

wilzbach:
I don't know why CodeCov doesn't display its status updates here (I have contacted their support), but this diff isn't well covered:

image

I am sorry for being so late to the party, but it shouldn't be too hard to evaluate `guessLength` at runtime, s.t. it's properly annotated with code coverage.

JackStouffer:
Ping @wilzbach

JackStouffer:
Going forward with this again. Any other issues can be resolved in follow ups.

PR #6027, opened 2018-01-12
Fix issue 18227 - Assertion failure in iota for floating point

Hackerpilot:


Hackerpilot:
What's going on with the auto-tester? It looks like this has both passed and failed on some of the FreeBSD 32-bit machines.

std/range/package.d
```diff
@@ -5590,12 +5590,20 @@ do
             auto pastEnd = start + count * step;
             if (step > 0)
             {
-                if (pastEnd < end) ++count;
+                while (pastEnd < end)
+                {
+                    ++count;
+                    pastEnd = start + count * step;
+                }
```
quickfur: Wow, took me a few minutes of staring hard at this code to understand what it's trying to do.

Can't help wondering if it's worth a comment or two explaining why the loop is necessary.  Also, would be nice to explain why a loop is necessary, and whether there's a better way of converging upon the correct count besides incrementing the multiplier to `step`.

Hackerpilot: What's really great is that my first attempt involved adding `step` to `pastEnd` each time through the loop, but that didn't work because floating point is "magical".

quickfur: Yeah, floats fail the Archimedean property, yet they're supposed to be a model of real numbers... :-P

BBasile: If you test with -m32 you should see the failure too. I do this with my IDE but this command should work: `dmd  package.d -m32 -unittest -main -run`.

One way to fix the issue is to extend the precision of the value as hold out of the FPU

```diff
    alias Value = Unqual!(CommonType!(B, E, S));
+    version(X86)
+    {
+        static if (isFloatingPoint!Value)
+            alias InternalValue = real;
+        else
+            alias InternalValue = Value;
+    }
+    else alias InternalValue = Value;

    static struct Result
    {
-        private Value start, step;
+        private InternalValue start, step;
        private size_t index, count;

        this(Value start, Value end, Value step)
        {
            import std.conv : to;

            this.start = start;
            this.step = step;
            immutable fcount = (end - start) / step;
            count = to!size_t(fcount);
-            Value pastEnd = start + count * step;
+            InternalValue pastEnd = start + count * step;
```

this patch + the new unittest is enough, i.e no need for the `while` loop. (at least everything is okay here).

BBasile: Actually the part of the fix with while is needed for x86_64

BBasile: Proper patch https://github.com/economicmodeling/phobos/pull/1

LightBender:
PR is ancient. Closing as stalled. @thewilsonator wants to look into this more.

PR #6029, opened 2018-01-14
fix issue 14767 - Support CTFE of BigInt under x86

BBasile:
This allows to declare very big BigInt constants but has the same limitation as x86_64 BigInt CTFE.

### DIFF
--- a/std/internal/math/biguintcore.d
+++ b/std/internal/math/biguintcore.d
@@ -56,21 +56,23 @@ private:
 // dipatchers to the right low-level primitives. Added to allow BigInt CTFE for
 // 32 bit systems (issue 14767) although it's used by the other architectures too.
 // See comments below in case it has to be refactored.
-pragma(inline, true)
+version(X86)
 uint multibyteAddSub(char op)(uint[] dest, const(uint)[] src1, const (uint)[] src2, uint carry)
 {
-    // CTFE, no arch distinction
+    // must be checked before, otherwise D_InlineAsm_X86 is true.
     if (__ctfe)
         return std.internal.math.biguintnoasm.multibyteAddSub!op(dest, src1, src2, carry);
-    // Run-time, X86 with asm
+    // Runtime.
     else version(D_InlineAsm_X86)
         return std.internal.math.biguintx86.multibyteAddSub!op(dest, src1, src2, carry);
-    // Run-time, not X86
+    // Runtime if no asm available.
     else
         return std.internal.math.biguintnoasm.multibyteAddSub!op(dest, src1, src2, carry);
 }
+// Any other architecture
+else alias multibyteAddSub = std.internal.math.biguintnoasm.multibyteAddSub;
 
-pragma(inline, true)
+version(X86)
 uint multibyteIncrementAssign(char op)(uint[] dest, uint carry)
 {
     if (__ctfe)
@@ -80,8 +82,9 @@ uint multibyteIncrementAssign(char op)(uint[] dest, uint carry)
     else
         return std.internal.math.biguintnoasm.multibyteIncrementAssign!op(dest, carry);
 }
+else alias multibyteIncrementAssign = std.internal.math.biguintnoasm.multibyteIncrementAssign;
 
-pragma(inline, true)
+version(X86)
 uint multibyteShl()(uint[] dest, const(uint)[] src, uint numbits)
 {
     if (__ctfe)
@@ -91,8 +94,9 @@ uint multibyteShl()(uint[] dest, const(uint)[] src, uint numbits)
     else
         return std.internal.math.biguintnoasm.multibyteShl(dest, src, numbits);
 }
+else alias multibyteShl = std.internal.math.biguintnoasm.multibyteShl;
 
-pragma(inline, true)
+version(X86)
 void multibyteShr()(uint[] dest, const(uint)[] src, uint numbits)
 {
     if (__ctfe)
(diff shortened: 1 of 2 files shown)


std/internal/math/biguintcore.d
```diff
@@ -52,6 +52,130 @@ import std.range.primitives;
 import std.traits;
 
 private:
+
+// compile-time & run-time dipatchers to allows BigInt CTFE for 32 bit OS.
```
wilzbach: `/for 32bit OS/d` - it should work for all OS at CTFE?

std/internal/math/biguintcore.d
```diff
@@ -35,12 +35,12 @@ module std.internal.math.biguintcore;
...
 version(D_InlineAsm_X86)
 {
-    import std.internal.math.biguintx86;
-}
-else
-{
-    import std.internal.math.biguintnoasm;
+    static import std.internal.math.biguintx86;
 }
+static import std.internal.math.biguintnoasm;
```
wilzbach: This now will lead to a bigger object size on `x86` as this file will always be imported, but `dmd` isn't smart enough to lazy load local imports anyways. Besides `if (__ctfe)` isn't even `static if`, so imports within will always be loaded.

std/internal/math/biguintcore.d
```diff
@@ -52,6 +52,130 @@ import std.range.primitives;
 import std.traits;
 
 private:
+
+// compile-time & run-time dipatchers to allows BigInt CTFE for 32 bit OS.
+
+pragma(inline, true)
+uint multibyteAddSub(char op)(uint[] dest, const(uint)[] src1, const (uint)[] src2, uint carry) pure
+{
+    if (__ctfe)
```
wilzbach: Does are the times where CTFE + `static if` would be really cool:

```
static if(__ctfe || isNotX86)
```

std/internal/math/biguintcore.d
```diff
@@ -52,6 +52,130 @@ import std.range.primitives;
 import std.traits;
 
 private:
+
+// compile-time & run-time dipatchers to allows BigInt CTFE for 32 bit OS.
+
+pragma(inline, true)
+uint multibyteAddSub(char op)(uint[] dest, const(uint)[] src1, const (uint)[] src2, uint carry) pure
```
wilzbach: Why did you specify the attributes manually? It's a template and DMD should infer the attributes automatically.
The best practice is to avoid specifying the attributes manually, because then future improvements will propagate.

std/internal/math/biguintcore.d
BBasile: No but now it will likely be the case. Even if CTFE isn't a requirement i think that it's a problem when it works for an arch and not the other

std/internal/math/biguintcore.d
BBasile: Yes. I remember that Stefan said that it could be possible, so it may be true in the future.

BBasile:
It seems that the autotester, for some reason, stops after FreeBSD32 test.
In addition this PR has never been reported to CircleCI (same problem with other PRs of mine last week)

BBasile:
I authorized CircleCi for my own stuff again (revoked last week because i didn't use them) and after reopening the status is reported. They have a bug at CircleCi, it seems that they don't make the difference between people own stuff and their PRs in foreign repo.

wilzbach:
> It seems that the autotester, for some reason, stops after FreeBSD32 test.

Not for some reason. There are a couple of PRs with "auto-merge" in the queue. They receive a higher priority:
https://auto-tester.puremagic.com/pulls.ghtml?projectid=1

Windows + MacOS build hosts are typically quite busy:
https://auto-tester.puremagic.com/hosts/

> In addition this PR has never been reported to CircleCI (same problem with other PRs of mine last week)

I still don't know why that's just happening with your account, but it seems like it works now.

BBasile:
It works now because [see my other comment.](https://github.com/dlang/phobos/pull/6029#issuecomment-357584346), we've posted more or less at the same time.

std/internal/math/biguintcore.d
BBasile: Yeah i guess this is the counterpart of the trick used to handle symbol conflicts.

BBasile:
@wilzbach, can you accept as you stated before that if i 

> Remove pure from the templates and we should be good to go.

The pragma for inline guarantee that there's no performance regressions and they don't fail locally.

std/internal/math/biguintcore.d
quickfur: I was trying to think of various ways to work around this, but could not.

I wonder if there's a way to introduce a CTFE-only template, basically one that is not instantiated even if it's referred to, until the CTFE engine tries to execute it.

std/internal/math/biguintcore.d
```diff
@@ -52,6 +52,130 @@ import std.range.primitives;
 import std.traits;
 
 private:
+
+// compile-time & run-time dipatchers to allows BigInt CTFE for 32 bit systems.
```
wilzbach: typo: "to allow:

Also as I mentioned earlier, the dispatchers are more general than 32 bit, so either X86 or remove the last part of the sentence.

BBasile: Should be much better now, in no way someone will ever be confused about the role of the new functions.

BBasile:
Actually i can make a direct path for X86_64, saving the inlining step.

BBasile:
And bad news, i have to remove the pragma for inlining. Actually not all the primitives can be inlined.

quickfur:
For performance-related issues, I'm tempted to say check with ldc/gdc where possible. Dmd's optimizer leaves much to be desired.

BBasile:
Okay so this is still accepted.

BBasile:
The latest version is better. It doesn't affect X86_64, which is widely used, it only slightly affects X86, less and less used.

wilzbach:
DAutoTest fails due to https://github.com/dlang/dub/issues/1336

BBasile:
Thanks for the information buddy.

PR #6032, opened 2018-01-15
Fix Issue 18230 - multiwayUnion sets wrong pred lambdas

RazvanN7:
cc @9il @wilzbach

### DIFF
--- a/std/algorithm/setops.d
+++ b/std/algorithm/setops.d
@@ -950,7 +950,7 @@ auto multiwayUnion(alias less = "a < b", RangeOfRanges)(RangeOfRanges ror)
 {
     import std.algorithm.iteration : uniq;
     import std.functional : not;
-    return multiwayMerge!less(ror).uniq!(not!less);
+    return ror.multiwayMerge!(less).uniq!(not!less);
 }
 
 ///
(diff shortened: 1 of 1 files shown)


std/algorithm/setops.d
```diff
@@ -949,7 +949,8 @@ See also: $(LREF multiwayMerge)
 auto multiwayUnion(alias less = "a < b", RangeOfRanges)(RangeOfRanges ror)
 {
     import std.algorithm.iteration : uniq;
-    return ror.multiwayMerge.uniq;
+    import std.functional : not;
+    return multiwayMerge!less(ror).uniq!(not!less);
```
JackStouffer: nitpick: please don't mix non-UFCS and UFCS.

Either `ror.multiwayMerge!less.uniq!(not!less)` or `uniq!(not!less)(multiwayMerge!less(ror))`.

JackStouffer: The pred use for `uniq` doesn't seem correct. `uniq`'s pred is supposed to be for equality

std/algorithm/setops.d
```diff
@@ -980,6 +981,15 @@ auto multiwayUnion(alias less = "a < b", RangeOfRanges)(RangeOfRanges ror)
         [ 7 ],
     ];
     assert(equal(multiwayUnion(b), witness));
+
+    double[][] c =
+    [
+        [9, 8, 7, 6],
```
wilzbach: Add: `8, 8, 8` here

std/algorithm/setops.d
9il: It is correct here for current uniq implementation. And it is a big task to invent a uniq implementation that will not work here. The `not!less` solutions seems to be the best here

9il:
LGTM

PR #6034, opened 2018-01-15
Fix Issue 18239: fillWithMemcpy use memset when T.sizeof == 1

n8sh:
Current function in std.experimental.allocator.package:
```d
private void fillWithMemcpy(T)(void[] array, auto ref T filler) nothrow
{
    import core.stdc.string : memcpy;
    if (!array.length) return;
    memcpy(array.ptr, &filler, T.sizeof);
    // Fill the array from the initialized portion of itself exponentially.
    for (size_t offset = T.sizeof; offset < array.length; )
    {
        size_t extent = min(offset, array.length - offset);
        memcpy(array.ptr + offset, array.ptr, extent);
        offset += extent;
    }
}
```

When T.sizeof==1 we could use memset instead.

### DIFF
--- a/std/experimental/allocator/package.d
+++ b/std/experimental/allocator/package.d
@@ -895,7 +895,17 @@ nothrow @safe @nogc unittest
     assertThrown(make!InvalidImpureStruct(Mallocator.instance, 42));
 }
 
-private void fillWithMemcpy(T)(void[] array, auto ref T filler) nothrow
+private void fillWithMemcpy(T)(scope void[] array, auto ref T filler) nothrow
+if (T.sizeof == 1)
+{
+    import core.stdc.string : memset;
+    import std.traits : CopyConstness;
+    if (!array.length) return;
+    memset(array.ptr, *cast(CopyConstness!(T*, ubyte*)) &filler, array.length);
+}
+
+private void fillWithMemcpy(T)(scope void[] array, auto ref T filler) nothrow
+if (T.sizeof != 1)
 {
     import core.stdc.string : memcpy;
     import std.algorithm.comparison : min;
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/package.d
```diff
@@ -895,7 +895,17 @@ nothrow @safe @nogc unittest
...
 }
 
-private void fillWithMemcpy(T)(void[] array, auto ref T filler) nothrow
+private void fillWithMemcpy(T)(scope void[] array, auto ref T filler) nothrow
+if (T.sizeof == 1)
+{
+    import core.stdc.string : memset;
+    import std.traits : CopyConstness;
+    if (!array.length) return;
+    memset(array.ptr, *cast(CopyConstness!(T*, ubyte*)) &filler, array.length);
```
wilzbach: Why not `cast(int) filler` as the API expects?

```
pure nothrow @nogc @system void* memset(return void* s, int c, size_t n); 
```

https://dlang.org/phobos/core_stdc_string.html#.memset

For other reviewers are wondering too why `int` is required by the C header:

> The value is passed as an int, but the function fills the block of memory using the unsigned char conversion of this value.

http://www.cplusplus.com/reference/cstring/memset/

n8sh: > Why not `cast(int) filler`

My rationale is that it's possible for to T override opCast so `cast(int) filler` could return something whose low byte is not bitwise identical to T.init.

>as the API expects

ubyte will be promoted to int

quickfur: What about using a union to make the intent clearer, instead of that line noise complicated cast?
````
union U { T t; ubyte asByte; }
U u;
u.t = filler;
memset(array.ptr, u.filler, array.length);
````

Comment (n8sh) [reply std/experimental/allocator/package.d]: > `u.t = filler;`

I need to avoid `opAssign` in order for this function to always have the same behavior as the existing `fillWithMemcpy` which never calls `opAssign`. I could do:

```d
union U { Unqual!T t; ubyte asByte; }
U u = void;
memcpy(&u.t, &filler, 1);
memset(array.ptr, u.asByte, array.length);
```
if you think that would be clearer.

quickfur: Hmm. If you wish to avoid opAssign, I suppose there's really no other way than to dereference a pointer cast to `ubyte`. Fair enough.

PR #6039, opened 2018-01-16
Fix Issue 15798 - std.algorithm.mutation.copy takes target by value

JackStouffer:
`copy` is designed in such a way that passing in a generic output range as the target value makes no sense, as `copy` is supposed to return the "remainder" of target. As such, its inputs should be restricted to inputs that are assignable input ranges or slices.

Consider the following from the stackoverflow example

```d
import std.stdio;
import std.digest.digest;
import std.digest.md;
import std.algorithm;

void main() {
    string s = "Hello!\n";
    auto d1 = makeDigest!MD5;
    auto d2 = makeDigest!MD5;
    foreach (ubyte b; s) {
        d1.put(b);
    }
    s.copy(d2);
    writeln(digest!MD5(s).toHexString);
    writeln(d1.finish().toHexString);
    writeln(d2.finish().toHexString);
}
```

This outputs

```
E134CED312B3511D88943D57CCD70C83
E134CED312B3511D88943D57CCD70C83
D41D8CD98F00B204E9800998ECF8427E
```

Because copy takes the target by value. What the user should have done is use `std.range.primitives.put` to copy an input range to an output range which is taken by reference.

This change adds a deprecation message that alerts the user. Any usage of `copy` in this way is very likely deprecating broken code.

### DIFF
--- a/changelog/copy-auto-ref.dd
+++ b/changelog/copy-auto-ref.dd
@@ -1,0 +1,34 @@
+`std.algorithm.mutation.copy` Has Been Modified For `put`-style Output Ranges
+
+With this release, when the target output range in a call to
+$(REF copy, std, algorithm, mutation) is any `put` defining output range (and it
+doesn't have $(REF_ALTTEXT assignable elements, hasAssignableElements, std, range, primitives)),
+the target is taken as `auto ref`. This new overload is also `void` returning
+vs the other overloads which return the unfilled portion of the given range
+or array.
+
+Before this release, $(REF copy, std, algorithm, mutation) took all
+$(REF_ALTTEXT output ranges, isOutputRange, std, range, primitives) by value.
+This leads to some issues when output ranges aren't designed around a reference
+to heap memory, e.g. the hash ranges in $(MREF std,digest):
+
+-------
+// v2.078 and earlier
+import std.algorithm.mutation : copy;
+import std.digest.digest : toHexString;
+import std.digest.md : MD5;
+import std.range : put;
+
+auto s = "Hello!\n";
+auto h1 = makeDigest!MD5;
+auto h2 = makeDigest!MD5;
+
+put(h1, s); // takes h1 by ref
+copy(s, h2); // takes h2 by value
+
+assert(h1.finish().toHexString == "E134CED312B3511D88943D57CCD70C83");
+// Different result because copy didn't properly insert the contents of s
+assert(h2.finish().toHexString == "D41D8CD98F00B204E9800998ECF8427E");
+-------
+
+This issue no longer occurs in 2.079 and later as long as `target` is an lvalue.
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -78,7 +78,7 @@ T2=$(TR $(TDNW $(LREF $1)) $(TD $+))
 module std.algorithm.mutation;
 
 import std.range.primitives;
-import std.traits : isArray, isBlitAssignable, isNarrowString, Unqual, isSomeChar, isDynamicArray;
+import std.traits : isArray, isBlitAssignable, isNarrowString, Unqual, isSomeChar;
 // FIXME
 import std.typecons; // : tuple, Tuple;
 
@@ -369,7 +369,7 @@ Params:
 
 Returns:
     The unfilled part of `target` if `target` is not a `put` defining output
-    range.
+    range. `void` otherwise.
 
 See_Also:
     $(HTTP sgi.com/tech/stl/_copy.html, STL's _copy)
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -78,7 +78,7 @@ T2=$(TR $(TDNW $(LREF $1)) $(TD $+))
 module std.algorithm.mutation;
 
 import std.range.primitives;
-import std.traits : isArray, isBlitAssignable, isNarrowString, Unqual, isSomeChar;
+import std.traits : isArray, isBlitAssignable, isNarrowString, Unqual, isSomeChar, isDynamicArray;
 // FIXME
 import std.typecons; // : tuple, Tuple;
 
@@ -351,18 +351,25 @@ private enum bool areCopyCompatibleArrays(T1, T2) =
 
 // copy
 /**
-Copies the content of `source` into `target` and returns the
-remaining (unfilled) part of `target`.
+Copies the content of `source` into `target`.
 
-Preconditions: `target` shall have enough room to accommodate
+If `target` is an $(REF_ALTTEXT output range, isOutputRange, std,range,primitives)
+which is a range with assignable elements, then `copy` returns the remaining
+(unfilled) part of `target`. In this case, `target` must have enough room to accommodate
 the entirety of `source`.
+
+If `target` is an $(REF_ALTTEXT output range, isOutputRange, std,range,primitives)
+which defines a `put` method, than nothing is returned. In this case, it's generally
+recommended to pass `target` as an lvalue, to avoid any issues with copying the output
+data by value.
 
 Params:
     source = an $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
-    target = an output range
+    target = an $(REF_ALTTEXT output range, isOutputRange, std,range,primitives)
 
 Returns:
-    The unfilled part of target
+    The unfilled part of `target` if `target` is not a `put` defining output
+    range.
 
 See_Also:
     $(HTTP sgi.com/tech/stl/_copy.html, STL's _copy)
@@ -399,7 +406,8 @@ if (areCopyCompatibleArrays!(SourceRange, TargetRange))
 TargetRange copy(SourceRange, TargetRange)(SourceRange source, TargetRange target)
 if (!areCopyCompatibleArrays!(SourceRange, TargetRange) &&
     isInputRange!SourceRange &&
-    isOutputRange!(TargetRange, ElementType!SourceRange))
+    isOutputRange!(TargetRange, ElementType!SourceRange) &&
+    (isArray!TargetRange || hasAssignableElements!TargetRange))
 {
     // Specialize for 2 random access ranges.
     // Typically 2 random access ranges are faster iterated by common
@@ -420,6 +428,17 @@ if (!areCopyCompatibleArrays!(SourceRange, TargetRange) &&
         put(target, source);
         return target;
     }
+}
+
+/// ditto
+void copy(SourceRange, TargetRange)(SourceRange source, auto ref TargetRange target)
+if (!areCopyCompatibleArrays!(SourceRange, TargetRange) &&
+    isInputRange!SourceRange &&
+    isOutputRange!(TargetRange, ElementType!SourceRange) &&
+    !isArray!TargetRange &&
+    !hasAssignableElements!TargetRange)
+{
+    put(target, source);
 }
 
 ///
--- a/std/digest/package.d
+++ b/std/digest/package.d
@@ -265,7 +265,7 @@ version(ExampleDigest)
 
     auto oneMillionRange = repeat!ubyte(cast(ubyte)'a', 1000000);
     auto ctx = makeDigest!MD5();
-    copy(oneMillionRange, &ctx); //Note: You must pass a pointer to copy!
+    oneMillionRange.copy(ctx);
     assert(ctx.finish().toHexString() == "7707D6AE4E027C70EEA2A935C2296F21");
 }
 
@@ -433,10 +433,10 @@ DigestType!Hash digest(Hash, Range)(auto ref Range range)
 if (!isArray!Range
     && isDigestibleRange!Range)
 {
-    import std.algorithm.mutation : copy;
+    import std.range.primitives : put;
     Hash hash;
     hash.start();
-    copy(range, &hash);
+    put(hash, range);
     return hash.finish();
 }
 
@@ -634,7 +634,7 @@ interface Digest
 
     auto oneMillionRange = repeat!ubyte(cast(ubyte)'a', 1000000);
     auto ctx = new MD5Digest();
-    copy(oneMillionRange, ctx);
+    oneMillionRange.copy(ctx);
     assert(ctx.finish().toHexString() == "7707D6AE4E027C70EEA2A935C2296F21");
 }
 
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -15,7 +15,6 @@ module std.stdio;
 
 import core.stdc.stddef; // wchar_t
 public import core.stdc.stdio;
-import std.algorithm.mutation; // copy
 import std.meta; // allSatisfy
 import std.range.primitives; // ElementEncodingType, empty, front,
     // isBidirectionalRange, isInputRange, put
@@ -3143,7 +3142,7 @@ void main()
 
     @system unittest
     {
-        import std.algorithm.mutation : reverse;
+        import std.algorithm.mutation : copy, reverse;
         import std.exception : collectException;
         static import std.file;
         import std.range : only, retro;
@@ -4572,16 +4571,17 @@ private struct ChunksImpl
 
 /**
 Writes an array or range to a file.
-Shorthand for $(D data.copy(File(fileName, "wb").lockingBinaryWriter)).
+Shorthand for `put(data, File(fileName, "wb").lockingBinaryWriter)`.
 Similar to $(REF write, std,file), strings are written as-is,
 rather than encoded according to the $(D File)'s $(HTTP
 en.cppreference.com/w/c/io#Narrow_and_wide_orientation,
 orientation).
 */
 void toFile(T)(T data, string fileName)
-if (is(typeof(copy(data, stdout.lockingBinaryWriter))))
+if (is(typeof({ auto w = stdout.lockingBinaryWriter; put(w, data); })))
 {
-    copy(data, File(fileName, "wb").lockingBinaryWriter);
+    auto w = File(fileName, "wb").lockingBinaryWriter;
+    put(w, data);
 }
 
 @system unittest


std/stdio.d
```diff
@@ -4683,18 +4683,19 @@ alias stdin = makeGlobal!(StdFileHandle.stdin);
...
     import std.typecons : Yes;
 
     void main() {
-        stdin                       // read from stdin
-        .byLineCopy(Yes.keepTerminator) // copying each line
-        .array()                    // convert to array of lines
-        .sort()                     // sort the lines
-        .copy(                      // copy output of .sort to an OutputRange
-            stdout.lockingTextWriter()); // the OutputRange
+        put(
```
JackStouffer: Case in-point: the only reason this worked is because `stdout.lockingTextWriter` represents an output range which has only one global output destination.

CyberShadow: Boo, this breaks the UFCS chain. Is there as UFCS-able `put` in Phobos?

std/stdio.d
```diff
@@ -4683,18 +4683,20 @@ alias stdin = makeGlobal!(StdFileHandle.stdin);
...
-        .sort()                     // sort the lines
-        .copy(                      // copy output of .sort to an OutputRange
-            stdout.lockingTextWriter()); // the OutputRange
+        auto output = stdout.lockingTextWriter();
+        put(
+            output,                             // writing directly to stdout
+            stdin                               // read from stdin
+                .byLineCopy(Yes.keepTerminator) // copying each line
+                .array()                        // convert to array of lines
+                .sort()                         // sort the lines
```
wilzbach: I'm going to repeat @CyberShadow's [comment](https://github.com/dlang/phobos/pull/6039#discussion_r161854016) as it got hidden with the last rebase:

> Boo, this breaks the UFCS chain. Is there as UFCS-able `put` in Phobos?

JackStouffer:
@CyberShadow In this specific instance you can call `lockingTextWriter.put`. But it's better to encourage people to use the `global` put as it takes care of the type of use cases where output ranges only define a `char` accepting `put`, and a string is passed.

JackStouffer:
I get that using `std.range.put` isn't pretty, but it is more correct. This code worked by chance and not by design.

CyberShadow:
But it's not UFCS-able :(

wilzbach:
> I get that using std.range.put isn't pretty,
> But it's not UFCS-able :(

Wasn't this example one of the few prime examples of how elegant D code can look like? :disappointed:

CyberShadow:
UFCS is one of the more beautiful parts of D. Even though the code worked by chance, it saddens me that code like this now has to become uglier.

I often try to structure my code to improve its UFCS-ability, e.g. [[1]](https://github.com/CyberShadow/ae/blob/master/sys/install/vs.d#L160-L167) [[2]](https://github.com/CyberShadow/choose-product/blob/master/lenovo/common.d#L33-L40). I remember using this very pattern some time recently as well.

std/stdio.d
andralex: Yah, that's a problem.

JackStouffer:
There is a solution here: define some function `X` (we can bikeshed this) that just forwards to put, but because it's not called `put` we don't have to worry about name collision with the method of the type. That way it can be used in UFCS chains.

CyberShadow:
Yep, + invert args. I guess `.reverseArgs!put` will work now, but it's not really idiomatic. Anyway not a blocker I guess.

std/stdio.d
schveiguy: This is an existential problem with `put`, because the member hook is also called `put`.

schveiguy:
I'm not 100% convinced this is the right path. "it should only work for arrays" seems short-sighted. It should only work for output ranges that don't copy their state by value. However, we have no mechanism to test for this. Even if we made it a rule that output ranges must be reference types, there's no static way to check for this.

Is there no other path forward here?

JackStouffer:
@schveiguy it also works for ranges with assignable values a-la the `retro` example with `copy`.

schveiguy:
> it also works for ranges with assignable values

That's not necessarily a requirement:

```D
struct ValidOutputRange
{
   int front;
   void popFront() {front = 0;}
   enum empty = false;
}
```

JackStouffer:
@schveiguy I don't understand what that's supposed to illustrate. This would be treated as an output range by `copy` before this change and would discard all the values it's given except the last.

schveiguy:
What I'm trying to say is, I don't think there's a static test that accepts all ranges that would be valid for `copy` and rejects all ranges that would be invalid.

For an example of something that should be fine with `copy` but rejected after your change (I think):

```D
struct List(T)
{
    T value;
    List *next;
}

struct ListRange(T) // should be OK to use with copy
{
    List!T *cur;
    ref T front() { return cur.value; }
    void popFront() { cur = cur.next; }
    bool empty() { return cur == null; }
}
```

schveiguy:
Wait, I'm misunderstanding your tests. You are looking at ranges that *don't* have assignable elements, are input ranges, but are also output ranges. This... seems very odd.

JackStouffer:
@schveiguy Specifically I'm disallowing all `TargetRange`s which are output ranges but are also input ranges, because according to `put` any input range with assignable elements is an output range.

I'm trying to limit the deprecation to those output ranges which are `put` based.

schveiguy:
OK, I am confusing the `TargetRange` and the `SourceRange` checks. Regrouping...

schveiguy:
I'm focusing on the wrong thing. It's not valid input ranges that can be output ranges that cause problems, it's the output-only ranges that should also work. `LockingTextWriter` should work with `copy`, even if passed by value, because you aren't discarding the data. This is going to break lots of code that uses i/o via output ranges.

The real issue here is that `copy` needs to know whether the target is putting the data outside itself (and therefore can be passed by value). A crude test may be whether it has any pointers in it. But that still may not be enough, as you don't know if the pointers relate to the data itself, or something else (an unrelated string or something). As I said earlier, I don't think we can create such a test, unless we have a convention (a-la `save`, but that hasn't been stellar either).

The problem with this fix is that it purports to have such a test, but the test isn't good enough. This means you will be rejecting valid calls (such as with `LockingTextWriter`) and causing a different kind of confusion (why can't I call copy with XYZ?).

schveiguy:
I wonder if checking for purity could help in some way...

JackStouffer:
@schveiguy I agree, it's not a perfect solution. There's no way to know a-priori if a output range is properly copy-able by value. But, the way we've figured out how to get around the problem is to always use `ref` when passing them around. It sidesteps the whole issue.

Like I noted above, `LockingTextWriter` just happens to work. There are many other forms of output ranges which fail silently with copy as it exists today. Further, `copy`'s semantics are designed around the idea of array-like inputs (all of its tests using arrays at the bottom isn't an accident IMO). Most notably, the concept of `copy` returning the "unfilled" portion of the output range is not applicable to `put` based output ranges.

JackStouffer:
>What about taking the second overload by `auto ref`?

My problem with that is you still have the issue that `auto ref` will allow users to do the wrong thing and many output ranges will fail silently. I'd rather not give the user the chance to make a mistake, if possible.

JackStouffer:
Ping @jmdavis. You had a comment https://github.com/dlang/phobos/pull/5991#discussion_r159545584 on my output range `toString` PR which I think is relevant here, so I'd like to get your opinion.

schveiguy:
> many output ranges will fail silently

I don't think there are many output ranges that have this problem. Most reference the data anyway, or are lvalues.

I am on the side of not breaking working, valid, reasonable code like the example `stdin.byLineCopy(Yes.keepTerminator).array.sort.copy(stdout.lockingTextWriter)`.

std/stdio.d
```diff
@@ -4683,18 +4684,20 @@ alias stdin = makeGlobal!(StdFileHandle.stdin);
...
-        .array()                    // convert to array of lines
-        .sort()                     // sort the lines
-        .copy(                      // copy output of .sort to an OutputRange
-            stdout.lockingTextWriter()); // the OutputRange
+        auto output = stdout.lockingTextWriter();
+        put(
+            output,                             // writing directly to stdout
+            stdin                               // read from stdin
+                .byLineCopy(Yes.keepTerminator) // copying each line
+                .array()                        // convert to array of lines
```
andralex: as an aside, the parens here must go

schveiguy:
> have you considered changing copy such that it takes its target by auto ref?

We can't do that completely, because that changes the semantics of `copy` with forward ranges that are input ranges with assignable elements:

```D
auto buffer = new int[100];
iota(15).copy(buffer);
writeln(buffer[0 .. 15]);
```
Currently, this will print the numbers from 0 to 14. If `copy` is auto-ref, it will print 15 zeros.

I think we can use @JackStouffer's identification of the types of ranges that copy has problems with in this PR and just make THAT overload auto-ref. See https://github.com/dlang/phobos/pull/6039#issuecomment-358125018

wilzbach:
As no one pointed this out before - Jenkins failure is at tools:

```
changed.d(122): Deprecation: function std.algorithm.mutation.copy!(MapResult!(to, FilterResult!(__lambda5, Splitter!(cast(Flag)false, char[], Wrapper))), Appender!(int[])).copy is deprecated - std.algorithm.mutation.copy to non-arrays should be replaced with std.range.primitives.put
```

https://github.com/dlang/tools/blob/master/changed.d#L122

JackStouffer:
@schveiguy @andralex @CyberShadow Removed the deprecation and made the new overload take the output range by `auto ref`. For the new overload, I also made it return `void` to avoid (heh) confusion with returning "unfilled" portions of output ranges when that doesn't apply.

schveiguy:
I like this direction better.

I have considered further that it's possible an output range accepted by reference may break some code. Namely an output range that does put elements into an external location, but isn't expected to affect the lvalue that was used.

It might be possible to deprecate this functionality, but it might be too clever. If you pass in an lvalue by reference, you can `pragma(msg, "Warning, in version 2.0XX, this call will accept the target by reference, and affect the original")`, and then make a copy before using it.

std/digest/package.d
```diff
@@ -123,7 +123,7 @@ import std.traits;
     {
         File file = File(filename);
 
-        //As digests imlement OutputRange, we could use std.algorithm.copy
+        //As digests imlement OutputRange, we could use std.range.put
```
schveiguy: Can we revert this? `copy` should work for this now. Alternatively, the comment looks odd since you are calling `hash.put`, which I first thought was a UFCS call of `std.range.put`. If we leave the comment as calling `std.range.put`, maybe include the full call to show how it's different.

std/digest/package.d
```diff
@@ -162,7 +162,7 @@ import std.traits;
     {
         File file = File(filename);
 
-        //As digests implement OutputRange, we could use std.algorithm.copy
+        //As digests implement OutputRange, we could use std.range.put
```
schveiguy: ditto

std/digest/package.d
```diff
@@ -259,13 +259,12 @@ version(ExampleDigest)
...
     //Using the OutputRange feature
-    import std.algorithm.mutation : copy;
     import std.digest.md;
-    import std.range : repeat;
+    import std.range : repeat, put;
 
     auto oneMillionRange = repeat!ubyte(cast(ubyte)'a', 1000000);
     auto ctx = makeDigest!MD5();
-    copy(oneMillionRange, &ctx); //Note: You must pass a pointer to copy!
+    put(ctx, oneMillionRange);
```
schveiguy: Can revert this, and remove the `&`. Leave the note, but just show how `ctx` must be an lvalue for this to work.

std/digest/package.d
```diff
@@ -433,10 +432,10 @@ DigestType!Hash digest(Hash, Range)(auto ref Range range)
 if (!isArray!Range
     && isDigestibleRange!Range)
 {
-    import std.algorithm.mutation : copy;
+    import std.range.primitives : put;
     Hash hash;
     hash.start();
-    copy(range, &hash);
+    put(hash, range);
```
schveiguy: Can revert, and also switch to UFCS if you want.

std/digest/package.d
```diff
@@ -628,13 +627,12 @@ interface Digest
...
     //Using the OutputRange feature
-    import std.algorithm.mutation : copy;
     import std.digest.md;
-    import std.range : repeat;
+    import std.range : repeat, put;
 
     auto oneMillionRange = repeat!ubyte(cast(ubyte)'a', 1000000);
     auto ctx = new MD5Digest();
-    copy(oneMillionRange, ctx);
+    put(ctx, oneMillionRange);
```
schveiguy: ditto.

std/stdio.d
```diff
@@ -4572,16 +4571,17 @@ private struct ChunksImpl
...
 en.cppreference.com/w/c/io#Narrow_and_wide_orientation,
 orientation).
 */
 void toFile(T)(T data, string fileName)
-if (is(typeof(copy(data, stdout.lockingBinaryWriter))))
+if (is(typeof({ auto w = stdout.lockingBinaryWriter; put(w, data); })))
 {
-    copy(data, File(fileName, "wb").lockingBinaryWriter);
+    auto w = File(fileName, "wb").lockingBinaryWriter;
+    put(w, data);
```
schveiguy: I'd prefer reverting this, but I'm fine with the new version as well.

JackStouffer: I left this one to remove the global std.algorithm.mutation import.

std/digest/package.d
JackStouffer: Since this one isn't a public example, and std.range is more likely to have already been imported, I'll leave this

changelog/copy-auto-ref.dd
```diff
@@ -0,0 +1,32 @@
...
+With this release, when the target output range in a call to 
+$(REF copy, std, algorithm, mutation) is any `put` defining output range, the
+target is taken as `auto ref`. This new overload is also `void` returning
+vs the other overloads which return the unfilled portion of the given range
+or array.
+
+Before this release, $(REF copy, std, algorithm, mutation) took all 
+$(REF_ALTTEXT output ranges, isOutputRange, std, range, primitives) by value.
+Leads to some issues when output ranges aren't designed around a reference to
+heap memory, e.g. the hash ranges in $(MREF std,digest,package):
```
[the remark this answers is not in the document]
schveiguy: Is there a place to see what this actually produces?

std/algorithm/mutation.d
```diff
@@ -399,7 +406,8 @@ if (areCopyCompatibleArrays!(SourceRange, TargetRange))
 TargetRange copy(SourceRange, TargetRange)(SourceRange source, TargetRange target)
 if (!areCopyCompatibleArrays!(SourceRange, TargetRange) &&
     isInputRange!SourceRange &&
-    isOutputRange!(TargetRange, ElementType!SourceRange))
+    isOutputRange!(TargetRange, ElementType!SourceRange) &&
+    (isArray!TargetRange || hasAssignableElements!TargetRange))
```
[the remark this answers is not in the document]
schveiguy: Hm.. I think `hasAssignableElements` would fail on a `char[]`, but not sure if that's a valid output range anyway.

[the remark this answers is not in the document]
schveiguy: tested, `char[]` is not an output range for `char`, which is... strange. It looks like there's special handling for characters in `put`, so I would have expected this to work.

std/algorithm/mutation.d
```diff
@@ -422,6 +430,17 @@ if (!areCopyCompatibleArrays!(SourceRange, TargetRange) &&
     }
 }
 
+/// ditto
+void copy(SourceRange, TargetRange)(SourceRange source, auto ref TargetRange target)
+if (!areCopyCompatibleArrays!(SourceRange, TargetRange) &&
+    isInputRange!SourceRange &&
+    isOutputRange!(TargetRange, ElementType!SourceRange) &&
+    !isArray!TargetRange &&
+    !hasAssignableElements!TargetRange)
```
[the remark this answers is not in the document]
schveiguy: Not sure what you are meaning here, but I don't know how you do `auto ref` correctly without an overload.

changelog/copy-auto-ref.dd
```diff
@@ -0,0 +1,32 @@
+`std.algorithm.mutation.copy` Has Been Modified For `put`-style Output Ranges
+
+With this release, when the target output range in a call to 
+$(REF copy, std, algorithm, mutation) is any `put` defining output range, the
+target is taken as `auto ref`. This new overload is also `void` returning
+vs the other overloads which return the unfilled portion of the given range
+or array.
```
schveiguy: This isn't exact enough. It's really if the output range is not an input range with assignable elements. By deduction, this means it defines `put`, which is a necessary condition, but not a sufficient condition to trigger the auto-refness.

schveiguy: e.g. one could define an input range with assignable elements that ALSO defines `put` for some optimization. In this case, it won't be auto-ref.

changelog/copy-auto-ref.dd
```diff
@@ -0,0 +1,32 @@
...
+auto h2 = makeDigest!MD5;
+
+put(h1, s); // takes h1 by ref
+copy(s, h2); // takes h2 by value
+
+assert(h1.finish().toHexString == "E134CED312B3511D88943D57CCD70C83");
+assert(h2.finish().toHexString == "D41D8CD98F00B204E9800998ECF8427E");
+-------
+
+This issue no longer occurs as long as `target` is an lvalue.
```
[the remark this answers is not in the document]
schveiguy: You can just note that in 2.079 going forward, `h1` and `h2`'s results will be identical.

std/algorithm/mutation.d
```diff
@@ -78,7 +78,7 @@ T2=$(TR $(TDNW $(LREF $1)) $(TD $+))
 module std.algorithm.mutation;
 
 import std.range.primitives;
-import std.traits : isArray, isBlitAssignable, isNarrowString, Unqual, isSomeChar;
+import std.traits : isArray, isBlitAssignable, isNarrowString, Unqual, isSomeChar, isDynamicArray;
```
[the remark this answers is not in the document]
schveiguy: I think this is what the test originally was (not `isArray`). It may be leftover.

std/digest/package.d
schveiguy: OK, I didn't notice that it wasn't an example.

schveiguy:
I would like to hear from @andralex about the implications of using auto ref here, particularly surrounding this scenario: https://github.com/dlang/phobos/pull/6039#issuecomment-358347576

But other than that, LGTM.

changelog/copy-auto-ref.dd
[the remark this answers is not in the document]
wilzbach: Yes, DAutoTest will show a preview, but it's failing atm due to trailing whitespace:

```
Searching for trailing whitespace
./changelog/pending.dd:180:With this release, when the target output range in a call to 
./changelog/pending.dd:188:Before this release, $(REF copy, std, algorithm, mutation) took all 
posix.mak:875: recipe for target 'test' failed
```

std/algorithm/mutation.d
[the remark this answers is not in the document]
wilzbach: Ah fair enough :/

changelog/copy-auto-ref.dd
JackStouffer: Gah! If someone defines an output range which defines `put` and also has assignable elements, logically one would assume that any library function would prioritize `put` because that's supposedly the fundamental operation of output ranges. But `copy` prioritizes the `r.front` assignment, which according to the principal of least surprise is wrong. But "fixing" it would be a breaking change! 

Thankfully, an output range which is also an input range with assignable elements is really rare.

JackStouffer:
Fixed

changelog/copy-auto-ref.dd
schveiguy: Well, I would expect that if it does define assignable elements, either interface will properly work in this scenario (copy by value). So I think it's OK.

The real problem would be if you somehow had an assignable element input range of `T`, but an output range of `U` ;) But I won't worry about that case...

changelog/copy-auto-ref.dd
```diff
@@ -0,0 +1,35 @@
...
+auto h1 = makeDigest!MD5;
+auto h2 = makeDigest!MD5;
+
+put(h1, s); // takes h1 by ref
+copy(s, h2); // takes h2 by value
+
+assert(h1.finish().toHexString == "E134CED312B3511D88943D57CCD70C83");
+assert(h2.finish().toHexString == "D41D8CD98F00B204E9800998ECF8427E");
+// this fails in 2.079 and later
+assert(h1.finish().toHexString != h2.finish().toHexString)
```
schveiguy: Are you allowed to call `finish` more than once? I'm not familiar with the API of digests. Also the assert is trivially obvious (the two strings are different per the above asserts).

I think in the suggestion I was saying, you could just put a note after the code saying the two digests will be the same in 2.079 and later.

Alternatively, you could leave out the 3rd assert and put as a comment: "In 2.079 and later, both digests will have the same value".

JackStouffer:
Ping @andralex

schveiguy:
Just wanted to say I am good with the rewording, thanks.

schveiguy:
@andralex sorry, I'm not sure if you saw my previous comment. Switching to auto-ref *does* matter for arrays -- now you will advance the array along.

I guarantee there is code out there that does something like this:

```D
auto buffer = new int[100];
iota(15).copy(buffer);
writeln(buffer[0 .. 15]);
```

Which will be semantically different if copy takes it's target via ref.

andralex:
@schveiguy just create a local copy of the array inside copy

JackStouffer:
Closing this as I'm no longer interested. If someone else wants to shepherd these changes, feel free.

PR #6041, opened 2018-01-17
Fix Issue 17961 - std.uni does not compile with -unittest -dip1000

carblue:
see also my comment #7 in issue 17961

### DIFF
--- a/std/uni.d
+++ b/std/uni.d
@@ -1578,7 +1578,7 @@ if (is(Unqual!T == T))
     assert(nullSlice.empty);
 }
 
-private auto packedArrayView(T)(inout(size_t)* ptr, size_t items) @trusted pure nothrow
+private inout(PackedArrayView!T) packedArrayView(T)(inout(size_t)* ptr, size_t items)
 {
     return inout(PackedArrayView!T)(ptr, 0, items);
 }
@@ -1754,7 +1754,7 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
 
 
 // Simple storage manipulation policy
-@trusted private struct GcPolicy
+@safe private struct GcPolicy
 {
     import std.traits : isDynamicArray;
 
@@ -1791,7 +1791,7 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
         insertInPlace(arr, arr.length, value);
     }
 
-    static void destroy(T)(ref T arr)
+    static void destroy(T)(ref T arr) pure // pure required for -dip25, inferred for -dip1000
         if (isDynamicArray!T && is(Unqual!T == T))
     {
         debug
@@ -1801,7 +1801,7 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
         arr = null;
     }
 
-    static void destroy(T)(ref T arr)
+    static void destroy(T)(ref T arr) pure // pure required for -dip25, inferred for -dip1000
         if (isDynamicArray!T && !is(Unqual!T == T))
     {
         arr = null;
@@ -1809,7 +1809,7 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
 }
 
 // ditto
-@trusted struct ReallocPolicy
+@safe struct ReallocPolicy
 {
     import std.range.primitives : hasLength;
 
@@ -1820,9 +1820,9 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
         return result;
     }
 
-    static T[] alloc(T)(size_t size)
+    static T[] alloc(T)(size_t size) @trusted
     {
-        import core.stdc.stdlib : malloc;
+        import core.memory : pureMalloc;
         import std.exception : enforce;
 
         import core.checkedint : mulu;
@@ -1830,13 +1830,13 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
         size_t nbytes = mulu(size, T.sizeof, overflow);
         if (overflow) assert(0);
 
-        auto ptr = cast(T*) enforce(malloc(nbytes), "out of memory on C heap");
+        auto ptr = cast(T*) enforce(pureMalloc(nbytes), "out of memory on C heap");
         return ptr[0 .. size];
     }
 
-    static T[] realloc(T)(T[] arr, size_t size)
+    static T[] realloc(T)(scope T[] arr, size_t size) @trusted
     {
-        import core.stdc.stdlib : realloc;
+        import core.memory : pureRealloc;
         import std.exception : enforce;
         if (!size)
         {
@@ -1849,7 +1849,7 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
         size_t nbytes = mulu(size, T.sizeof, overflow);
         if (overflow) assert(0);
 
-        auto ptr = cast(T*) enforce(realloc(arr.ptr, nbytes), "out of memory on C heap");
+        auto ptr = cast(T*) enforce(pureRealloc(arr.ptr, nbytes), "out of memory on C heap");
         return ptr[0 .. size];
     }
 
@@ -1866,7 +1866,7 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
         arr[$-1] = force!T(value);
     }
 
-    @safe unittest
+    pure @safe unittest
     {
         int[] arr;
         ReallocPolicy.append(arr, 3);
@@ -1889,7 +1889,7 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
         copy(value, arr[$-value.length..$]);
     }
 
-    @safe unittest
+    pure @safe unittest
     {
         int[] arr;
         ReallocPolicy.append(arr, [1,2,3]);
@@ -1898,11 +1898,11 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
         assert(equal(arr, [1,2,3]));
     }
 
-    static void destroy(T)(ref T[] arr)
+    static void destroy(T)(scope ref T[] arr) @trusted
     {
-        import core.stdc.stdlib : free;
+        import core.memory : pureFree;
         if (arr.ptr)
-            free(arr.ptr);
+            pureFree(arr.ptr);
         arr = null;
     }
 }
@@ -1910,7 +1910,7 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
 //build hack
 alias _RealArray = CowArray!ReallocPolicy;
 
-@safe unittest
+pure @safe unittest
 {
     import std.algorithm.comparison : equal;
 
@@ -2064,7 +2064,7 @@ pure:
     alias $(LREF CodepointSet) throughout the whole code base.
     )
 */
-@trusted public struct InversionList(SP=GcPolicy)
+public struct InversionList(SP=GcPolicy)
 {
     import std.range : assumeSorted;
 
@@ -2150,7 +2150,7 @@ pure:
     }
 
     ///
-    @safe unittest
+    pure @safe unittest
     {
         import std.algorithm.comparison : equal;
 
@@ -2164,6 +2164,12 @@ pure:
         auto set2 = CodepointSet('а', 'я'+1, 'a', 'd', 'b', 'z'+1);
         //the same end result
         assert(set2.byInterval.equal(set.byInterval));
+        // test constructor this(Range)(Range intervals)
+        auto chessPiecesWhite = CodepointInterval(9812, 9818);
+        auto chessPiecesBlack = CodepointInterval(9818, 9824);
+        auto set3 = CodepointSet([chessPiecesWhite, chessPiecesBlack]);
+        foreach (v; '♔'..'♟'+1)
+            assert(set3[v]);
     }
 
     /**
@@ -2179,7 +2185,7 @@ pure:
         assert(set.byInterval.equal([tuple('A','E'), tuple('a','e')]));
         -----------
     */
-    @property auto byInterval()
+    @property auto byInterval() scope
     {
         return Intervals!(typeof(data))(data);
     }
@@ -2201,7 +2207,7 @@ pure:
     }
 
     ///
-    @safe unittest
+    pure @safe unittest
     {
         auto gothic = unicode.Gothic;
         // Gothic letter ahsa
@@ -2287,7 +2293,7 @@ public:
     }
 
     ///
-    @safe unittest
+    pure @safe unittest
     {
         import std.algorithm.comparison : equal;
         import std.range : iota;
@@ -2353,7 +2359,7 @@ public:
     }
 
     ///
-    @safe unittest
+    pure @safe unittest
     {
         assert('я' in unicode.Cyrillic);
         assert(!('z' in unicode.Cyrillic));
@@ -2376,7 +2382,7 @@ public:
     */
     @property auto byCodepoint()
     {
-        @trusted static struct CodepointRange
+        static struct CodepointRange
         {
             this(This set)
             {
@@ -2415,7 +2421,7 @@ public:
     }
 
     ///
-    @safe unittest
+    pure @safe unittest
     {
         import std.algorithm.comparison : equal;
         import std.range : iota;
@@ -2473,7 +2479,7 @@ public:
     }
 
     ///
-    @safe unittest
+    pure @safe unittest
     {
         import std.conv : to;
         import std.format : format;
@@ -2494,7 +2500,7 @@ public:
             ~"[0XA640..0XA698) [0XA69F..0XA6A0)");
     }
 
-    @safe unittest
+    pure @safe unittest
     {
         import std.exception : assertThrown;
         import std.format : format, FormatException;
@@ -2512,7 +2518,7 @@ public:
     }
 
     ///
-    @safe unittest
+    pure @safe unittest
     {
         CodepointSet someSet;
         someSet.add('0', '5').add('A','Z'+1);
@@ -2548,7 +2554,7 @@ private:
         return this;
     }
 
-    @safe unittest
+    pure @safe unittest
     {
         assert(unicode.Cyrillic.intersect('-').byInterval.empty);
     }
@@ -2615,7 +2621,7 @@ public:
     }
 
     ///
-    @safe unittest
+    pure @safe unittest
     {
         auto set = unicode.ASCII;
         // union with the inverse gets all of the code points in the Unicode
@@ -2657,7 +2663,7 @@ public:
             return result;
         }
 
-        static string binaryScope(R)(R ivals, string indent)
+        static string binaryScope(R)(R ivals, string indent) @safe
         {
             // time to do unrolled comparisons?
             if (ivals.length < maxBinary)
@@ -2775,7 +2781,7 @@ public:
     }
 
     ///
-    @safe unittest
+    pure @safe unittest
     {
         CodepointSet emptySet;
         assert(emptySet.length == 0);
@@ -2789,14 +2795,14 @@ private:
     // a random-access range of integral pairs
     static struct Intervals(Range)
     {
-        this(Range sp)
+        this(Range sp) scope
         {
             slice = sp;
             start = 0;
             end = sp.length;
         }
 
-        this(Range sp, size_t s, size_t e)
+        this(Range sp, size_t s, size_t e) scope
         {
             slice = sp;
             start = s;
@@ -3120,7 +3126,7 @@ private:
     CowArray!SP data;
 }
 
-@system unittest
+pure @system unittest
 {
     import std.conv : to;
     assert(unicode.ASCII.to!string() == "[0..128)");
@@ -3405,7 +3411,7 @@ private:
     uint[] data;
 }
 
-@safe unittest// Uint24 tests
+pure @safe unittest// Uint24 tests
 {
     import std.algorithm.comparison : equal;
     import std.algorithm.mutation : copy;
@@ -3497,7 +3503,7 @@ version(unittest)
     private alias AllSets = AliasSeq!(InversionList!GcPolicy, InversionList!ReallocPolicy);
 }
 
-@safe unittest// core set primitives test
+pure @safe unittest// core set primitives test
 {
     import std.conv : text;
     foreach (CodeList; AllSets)
@@ -3574,7 +3580,7 @@ version(unittest)
 
 
 //test constructor to work with any order of intervals
-@safe unittest
+pure @safe unittest
 {
     import std.algorithm.comparison : equal;
     import std.conv : text, to;
@@ -3618,7 +3624,7 @@ version(unittest)
 }
 
 
-@safe unittest
+pure @safe unittest
 {   // full set operations
     import std.conv : text;
     foreach (CodeList; AllSets)
@@ -3723,7 +3729,7 @@ version(unittest)
 
 }
 
-@safe unittest// vs single dchar
+pure @safe unittest// vs single dchar
 {
     import std.conv : text;
     CodepointSet a = CodepointSet(10, 100, 120, 200);
@@ -3731,7 +3737,7 @@ version(unittest)
     assert((a & 'B') == CodepointSet(66, 67));
 }
 
-@safe unittest// iteration & opIndex
+pure @safe unittest// iteration & opIndex
 {
     import std.algorithm.comparison : equal;
     import std.conv : text;
@@ -4620,7 +4626,7 @@ public struct MatcherConcept
         assert(false);
     }
     ///
-    @safe unittest
+    pure @safe unittest
     {
         string truth = "2² = 4";
         auto m = utfMatcher!char(unicode.Number);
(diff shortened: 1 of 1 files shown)


std/uni.d
```diff
@@ -2541,10 +2541,14 @@ private:
...
     ref intersect()(dchar ch)
     {
-        foreach (i; byInterval)
+        scope arr = data;
+        foreach (i; Intervals!(typeof(data))(arr))
             if (i.a <= ch && ch < i.b)
-                return this = This.init.add(ch, ch+1);
-        this = This.init;
+            {
+                data.length(0);
```
quickfur: Nitpick: I think the intended usage is:
````
data.length = 0;
````

File: std/uni.d [hunk 2]
```diff
@@ -2541,10 +2541,14 @@ private:
...
+                return this.add(ch, ch+1);
+            }
+        data.length(0);
```
quickfur: ditto

std/uni.d
carblue: I didn't go through all details of CowArray and Your comment was my first idea too, but as it is reference counted, I opted for the method
@property void length(size_t len)
which cares for refCount

nemanja-boric-sociomantic: @quickfur 's suggestion still calls that method:

```
void length (int val)
{
    import std.stdio;
    writeln(val);
}

void main()
{
    length = 1;
}
```

(prints 1).

DmitryOlshansky: I think just using return scope should be the ticket.

No idea what this muckering with scope here acomplishes but I think it just fools compiler

quickfur: @carblue I think you misunderstood. The syntax `data.length = 0;` is lowered to `data.length(0)`. It's not a big deal, but usually when you write `@property void length(size_t) { ... }`, the intention is to use it as a property setter method, i.e., you would invoke it with assignment syntax rather than function call syntax.

carblue: Agreed about data.length = 0; and thanks, I forgot about that lowering.

@DmitryOlshansky The scope topic is tricky; I watched Walter Bright's Pointers Gone Wild: Memory Safety and D several times, introduced that into some code and know from tests, that the implementation still needs some refinement: I wouldn't call it "fool", but just "help" the compiler with it's current, hypersensitive dip1000 code and/or help me save time: Whether legitimately or not  (I didn't go to the bottom of that in static struct Intervals(Range) or wherever, which may be worthwhile), the compiler - in the first place - rejects byInterval in foreach with the error mentioned in issue 17961. The simple, yet maybe temporary solution I opted for, is, to give the compiler what it seemingly wants to see here, a scope variable (arr) -based iteration.
I hope, when dmd is dip1000-ready, the compiler will be that 'clever' to detect the helping hand case here, making arr dispensable.

If I understood Your ticket right, 'return scope this;' doesn't compile for syntax reason.

quickfur:
LGTM.  Though I'll defer to @DmitryOlshansky to take another look at the new proposed fix.

DmitryOlshansky:
Errors on circle-ci and ddoc-tester seem spurious, could you please rebase?

carblue:
Yes I will rebase, but first learn how to do that.
I tried to dig into the ci/circleci pure/impure errors and injected behind function byInterval():
pragma(msg, "std.uni.d:", __LINE__, "  ",   typeof(&byInterval));
resulting in:
std.uni.d:2186  Intervals!(CowArray!(GcPolicy)) function() pure nothrow @nogc @property scope @trusted
I have no clue what's going on there: The only function I touched didn't change it's pure adornment

wilzbach:
CircleCi runs `publictests`, which in turn runs e.g. `std/uni.publictests`

The target ensures that the public unittest in Phobos don't contain any private dependency and are really runnable. It does nothing more than extracting all public unittest and runs them against the freshly compiled Phobos library.

wilzbach:
BTW here's a simple trick that allows to gradually enable `-dip1000`:

```diff
diff --git a/posix.mak b/posix.mak
index d0576574a..e30b687c7 100644
--- a/posix.mak
+++ b/posix.mak
@@ -357,7 +357,10 @@ UT_D_OBJS:=$(addprefix $(ROOT)/unittest/,$(addsuffix .o,$(D_MODULES)))
 $(UT_D_OBJS): $(ALL_D_FILES)
 $(UT_D_OBJS): $(ROOT)/unittest/%.o: %.d
 	@mkdir -p $(dir $@)
-	$(DMD) $(DFLAGS) $(UDFLAGS) -c -of$@ $<
+	declare -A aa=( \
+		["std/uni.d"]=-dip1000 \
+	) \
+	$(DMD) $(DFLAGS) $(UDFLAGS) $${aa["$<"]}  -c -of$@ $<
 
 ifneq (1,$(SHARED))
```

carblue:
@wilzbach Thank You very much for Your last 2 comments. They are precisely what I currently required and enable me to go on pushing forward -dip1000 for phobos.

Btw, I did the rebase asked for and it's exactly summarizing the 3 outdated commits, which did pass  jenkins and IIRC auto-tester as well (now they seem to have stalled). Since 2.078.1 is released, the rebase to stable seems superfluous now, but I'll pay attention next time.
From my side, everything seems to have been done for this PR.

wilzbach:
> @wilzbach Thank You very much for Your last 2 comments. They are precisely what I currently required and enable me to go on pushing forward -dip1000 for phobos.

You are welcome. Let me know if you have further questions / problems. I also sent you an invite to our Slack channel - it's often the quickest way to get answers.

> auto-tester as well (now they seem to have stalled)

Auto-tester will invalidate all results after every PR merged to dmd,druntime and Phobos. Don't worry about it being pending, that's the case for most PR. Once you new commits are added to a PR or you force-push, you will land on top of the queue.
Also once this gets the auto-merge label, it will have priority on the auto-tester.

>  Since 2.078.1 is released, the rebase to stable seems superfluous now, but I'll pay attention next time.

This is just a general warning which we display for issues labelled as regression, but technically this isn't a regression, because `-dip1000` isn't supported by Phobos officially yet. So don't worry about it ;-)

> From my side, everything seems to have been done for this PR.

We still need to find out why `make -f posix.mak std/datetime/interval.publictests` fails (i.e. why CircleCi fails).
In other words this fails to compile with your PR:

```d
import std.datetime.interval;
import std.datetime.date : Date, DayOfWeek;
auto interval = Interval!Date(Date(2010, 9, 2), Date(2010, 9, 27));
```

carblue:
Thanks for Your detailed explanations and the invitation to join the dlang team on slack. I registered and will explore slack tomorrow.
Okay I'll dig deeper into the CircleCi failure and envision, there will be a cascade of other std.uni.d fixes being required before this PR can be merged then: I've seen some mention of std.uni in bugzilla ref. @safe and -dip1000. Thus I'll now start from the end of the queue.

wilzbach:
@quickfur FYI the CircleCi failure here isn't random or spurious, it's a real one. See the discussions above.

quickfur:
Ah, I missed that. Sorry about that.

carblue:
~/gitclones/phobos$  make -j6 -f posix.mak unittest
doesn't show any errors when applying both this PR #6041 and PR #6074.

@wilzbach
BTW Your trick didn't work for me exactly that way, the aa keys didn't get recognized, thus I changed aa to be a global associative array; (maybe specific for bash shell? and still not the final version; allows fine-grained -dipXYZ setting; as far as I got, there are already a lot -dip1000 compatible):
```
--- phobos/posix.mak    2018-01-26 19:22:27.655289238 +0100
+++ posix.mak   2018-01-27 20:57:04.340708054 +0100
@@ -340,12 +340,97 @@
 $(addprefix $(ROOT)/unittest/,$(DISABLED_TESTS)) :
        @echo Testing $@ - disabled
 
+aa[array]=-dip1000      # Can add multiple values one by one
+aa[ascii]=-dip1000      # Can add multiple values one by one
+aa[base64]=-dip1000
+aa[bigint]=-dip1000
+aa[bitmanip]=-dip1000
+aa[compiler]=-dip1000
+aa[complex]=-dip1000
+aa[concurrency]=-dip1000
+aa[conv]=-dip25
+aa[bitmanip]=-dip1000
+aa[csv]=-dip1000
+aa[demangle]=-dip1000
+aa[encoding]=-dip1000
+aa[exception]=-dip1000
+aa[file]=-dip1000
+aa[format]=-dip25
+aa[functional]=-dip1000
+aa[getopt]=-dip1000
+aa[json]=-dip1000
+aa[math]=-dip1000
+aa[mathspecial]=-dip1000
+aa[meta]=-dip1000
+aa[mmfile]=-dip1000
+aa[numeric]=-dip1000
+aa[outbuffer]=-dip1000
+aa[parallelism]=-dip1000
+aa[path]=-dip25
+aa[process]=-dip1000
+aa[random]=-dip1000
+aa[signals]=-dip1000
+aa[socket]=-dip25
+aa[stdint]=-dip1000
+aa[stdio]=-dip1000
+aa[string]=-dip1000
+aa[system]=-dip1000
+aa[traits]=-dip1000
+aa[typecons]=-dip25
+aa[typetuple]=-dip1000
+aa[uni]=-dip25
+aa[uri]=-dip1000
+aa[utf]=-dip1000
+aa[uuid]=-dip1000
+aa[variant]=-dip1000
+aa[xml]=-dip1000
+aa[zip]=-dip1000
+aa[zlib]=-dip1000
+
+#std/algorithm
+aa[comparison]=-dip25
+aa[internal]=-dip1000
+aa[iteration]=-dip25
+aa[mutation]=-dip1000
+aa[package]=-dip25
+aa[searching]=-dip25
+aa[setops]=-dip1000
+aa[sorting]=-dip25
+
+#std/container
+#aa[array]=-dip1000
+aa[binaryheap]=-dip1000
+aa[dlist]=-dip1000
+#aa[package]=-dip1000
+aa[rbtree]=-dip1000
+aa[slist]=-dip25
+aa[util]=-dip25
+
+#std/datetime
+aa[date]=-dip1000
+aa[interval]=-dip1000
+#aa[package]=-dip1000
+aa[stopwatch]=-dip1000
+aa[systime]=-dip25
+aa[timezone]=-dip1000
+
+#std/digest
+aa[crc]=-dip1000
+aa[digest]=-dip1000
+aa[hmac]=-dip1000
+aa[md]=-dip1000
+aa[murmurhash]=-dip1000
+#aa[package]=-dip1000
+aa[ripemd]=-dip1000
+aa[sha]=-dip1000
+#...
+
 UT_D_OBJS:=$(addprefix $(ROOT)/unittest/,$(addsuffix .o,$(D_MODULES)))
 # need to recompile all unittest objects whenever sth. changes
 $(UT_D_OBJS): $(ALL_D_FILES)
 $(UT_D_OBJS): $(ROOT)/unittest/%.o: %.d
        @mkdir -p $(dir $@)
-       $(DMD) $(DFLAGS) $(UDFLAGS) -c -of$@ $<
+       $(DMD) $(DFLAGS) $(UDFLAGS) ${aa[$(basename $(notdir $<))]}  -c -of$@ $<
 
 ifneq (1,$(SHARED))
```

carblue:
Arrrggghhh, PR #6074 is fine, but PR #6041  on top is still erroring with: make -f posix.mak publictests

wilzbach:
Great job with creating the list!
I started a GH project to give this better visibility and nicely track the progress: https://github.com/dlang/phobos/projects/6

We are currently having some troubles with this CircleCi check, so it might be that the failure here was just the first indication of a wider problem. I will have a more detailed look into this later.

wilzbach:
> We are currently having some troubles with this CircleCi check, so it might be that the failure here was just the first indication of a wider problem. I will have a more detailed look into this later.

Managed to find the cause of the permanent CircleCi issues: https://github.com/dlang/phobos/pull/6099, but I think your failure is different (see [my comment above](https://github.com/dlang/phobos/pull/6041#issuecomment-360308904))

wilzbach:
> Merge remote-tracking branch 'upstream/master' into issue17961

Please don't use merge, but rebase. This helps to keep the history sane.
All CIs will merge your PR with latest master anyhow.

See also: https://wiki.dlang.org/Starting_as_a_Contributor#Rebasing

carblue:
The solution will be far more complicated than/different from simply adding scope to byInterval() (which seems to work for InversionList!GcPolicy only). The impure errors stem from InversionList!ReallocPolicy.
It's currently like "herding cats": When I (assume to) solve one specific error, others come up instead: there's so much interrelation in this large file. I'm trying to isolate as much as possible issues to easily reviewable PRs like https://github.com/dlang/phobos/pull/6104.
I'm -dip1000-testing targets std/uni.test, publictests and unittest (enabling -dip1000 for std/uni.d only) and they come up with diverging errors, and I'm still struggling with DIP1000.md to reside in my memory: It really deserves an overhauled representation and some more examples: We have for parameter(s): scope, return scope, scope ref and even return scope ref; for the this implicit parameter: scope; for the return value: scope and scope ref.
auto-generated functions don't infer scope correctly in all cases #18000 (especially auto-generated opAssign is not scope aware which causes an error in current intersect function), there are still other unsolved dip1000-implementation issues in bugzilla, and whether scope was inferred or not and where is invisible to me (pragma(msg...)) except referring to the this implicit parameter, as they are not type attributes (maybe -vcg-ast can help). Thus it's currently a maze to find a way (I hope there is one) around bugs and complexity, but currently my learning experience stilll outweighs the hassle.

carblue:
@wilzbach @DmitryOlshansky @quickfur 
The issue should be fixed by latest commit.
In order to test against **released** libphobos2.so etc. (which come with
pure nothrow @nogc **@trusted** void std.uni.GcPolicy.destroy!(uint[]).destroy(ref uint[])  )
temporarily change in Your local copy of std.uni.d : **@safe** private struct GcPolicy 
back to **@trusted** private struct GcPolicy.
I tried hard not to change semantics and almost/virtually succeeded, except there are 2 replacements of ref return-type to auto return-type (functions 'tab' in template Utf8Matcher and 'ascii' in template Utf16Matcher). I didn't find any combination of attributes/storage class to keep them as ref and be compatible with -dip1000. But that's currently no problem, as the ref return of tab/ascii is nowhere used as such and I doubt it makes sense to be ref. As a reminder, I kept /*ref*/ in there.
As far as I can interpret the failing tests jenkins/codecov, there is no remaining issue ?!

std/uni.d
```diff
@@ -2629,7 +2629,7 @@ public:
         import std.algorithm.searching : countUntil;
         import std.format : format;
         enum maxBinary = 3;
-        static string linearScope(R)(R ivals, string indent)
+        @safe static string linearScope(R)(R ivals, string indent)
```
wilzbach: Be style consistent ;-)

std/uni.d
```diff
@@ -1898,7 +1898,7 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
         assert(equal(arr, [1,2,3]));
     }
 
-    static void destroy(T)(ref T[] arr)
+    @trusted static void destroy(T)(scope ref T[] arr)
```
wilzbach: Freeing memory is not safe in general. Move the `@trusted` to the call site where track of the reference is kept.

std/uni.d
```diff
@@ -1834,7 +1834,7 @@ alias sharSwitchLowerBound = sharMethod!switchUniformLowerBound;
         return ptr[0 .. size];
     }
 
-    static T[] realloc(T)(T[] arr, size_t size)
+    @trusted static T[] realloc(T)(scope T[] arr, size_t size)
```
wilzbach: `destroy` isn't safe in general, move this to the call site.

wilzbach:
> I didn't find any combination of attributes/storage class to keep them as ref and be compatible with -dip1000. But that's currently no problem, as the ref return of tab/ascii is nowhere used as such and I doubt it makes sense to be ref. As a reminder, I kept /ref/ in there.

Ehm I would even go one step further and remove the `ref`. Otherwise people in the future looking at the code will be confused.

> As far as I can interpret the failing tests jenkins/codecov, there is no remaining issue ?!

Yes those are unrelated. I will look into those. A start for the Jenkins failures: https://github.com/dlang/ci/pull/163

std/uni.d
carblue: Yes, I agree about alloc, realloc and destroy, but in the end: I didn't change their attribute @trusted, they already were declared that way by original @trusted struct GcPolicy and @trusted struct ReallocPolicy. As I outlined in #6104, I currently won't touch function bodies (as far as possible) and be focussed on function/struct signatures for -dip1000 compliance. Frankly speaking, looking deeper into phobos, I was surprised how much @trusted still is in there, part of which possibly undermining memory safety, thus I will leave Your issue for Dmitry or whoever.
Reviewing my changes, there is only 1 place, where I added @trusted where no one was before: mixin template ForwardStrings function fwdStr. I forgot to mention that.

carblue:
@wilzbach 
A recapitulation of what changed, all limited to signatures of functions/structs, except the added lines of unittest:
Changes acc. to DIP25 and DIP1000 to be compilable with -dip1000 (roughly in the context of InversionList, it's alloc-policies, CowArray and Grapheme).
-Attributes/storage class removed where they get inferred (all removed @trusted are inferred @safe now etc.; checked for all removed with pragma(msg typeof...))
-@trusted was added in a single place, where no one was before: mixin template ForwardStrings, function fwdStr. If @trusted is not justified here, the whole fix won't work !
-@trusted was moved from some @trusted struct to some of their member functions where currently required for @safe struct (e.g. ReallocPolicy needs more work done on it where @trusted can't be guaranteed) !
-Changed return type from ref to auto for funtions ascii and tab
-Some D style changes like: Auto return type replaced in some locations (stated explicitly), alphabetical ordering of attibutes (in the end for functions, in the beginning for structs)

-added unittesting of InversionList.this(Range)(Range intervals) to calm codecov (required anyway but should ideally go in isolated PR)
There are still CI errors, but seem unrelated to this PR

std/uni.d
```diff
@@ -4711,7 +4717,7 @@ enum Mode {
 
 mixin template ForwardStrings()
 {
-    private bool fwdStr(string fn, C)(ref C[] str) const pure
+    private bool fwdStr(string fn, C)(ref C[] str) const @trusted
```
wilzbach: Do we have a test for the purity here?

wilzbach:
> There are still CI errors, but seem unrelated to this PR

Yes, restarted them.

> -added unittesting of InversionList.this(Range)(Range intervals) to calm codecov (required anyway but should ideally go in isolated PR)

Nice! BTW since https://github.com/dlang/phobos/pull/6124 CodeCov will always be green / "calm".

std/uni.d
carblue: Yes, with latest commit: Any utf\d*Matcher calling match/test/skip has it in it's call chain, e.g. pure @safe unittest beginning at line 4634 or 4670 or 5319 or 5374 or 5424. Added pure to many @safe unittest.
BTW, I failed introducing Mallocator in ReallocPolicy.alloc/realloc/destroy due to @system calls, like this:

@@ -1812,6 +1812,7 @@
 @safe struct ReallocPolicy
 {
     import std.range.primitives : hasLength;
+    import std.experimental.allocator.mallocator : Mallocator;
 
     static T[] dup(T)(const T[] arr)
     {
@@ -1820,9 +1821,10 @@
         return result;
     }
 
-    static T[] alloc(T)(size_t size) @trusted
+    static T[] alloc(T)(size_t size) //@trusted
     {
-        import core.stdc.stdlib : malloc;
+//        import core.memory : pureMalloc; //import core.stdc.stdlib : malloc;
+        import std.experimental.allocator : makeArray;
 
         import core.checkedint : mulu;
@@ -1830,13 +1832,14 @@
         size_t nbytes = mulu(size, T.sizeof, overflow);
         if (overflow) assert(0);
 
-        auto ptr = cast(T*) enforce(malloc(nbytes), "out of memory on C heap");
-        return ptr[0 .. size];
+        return enforce(Mallocator.instance.makeArray!ubyte(nbytes), "out of memory on C heap"); //auto ptr = cast(T*) enforce(pureMalloc(nbytes), "out of memory on C heap");
+//        return ptr[0 .. size];
     }

wilzbach:
> incompatible with onlineapp

Out of interest: what do you mean by this?
If you are referring to a problem with run.dlang.io, see https://github.com/dlang-tour/core/issues/672

carblue:
After commit 2f590da I checked the cyberShadow/DAutoTest generated 
http://dtest.dlang.io/artifact/website-9f2164728b4bbb036aec0fc431428ba026513b42-a3ecbbd38ec1e34f7756deb6bdfa3123/web/phobos-prerelease/std_uni.html
and run the unitest touched (that one with the chessPieces: It failed with:
onlineapp.d(8): Error: undefined identifier `CodepointSetRP`, did you mean alias `CodepointSet`?

wilzbach:
Ah don't worry about this, but the backend (run.dlang.io) can't build Docker containers for every PR. We simply use master (aka dmd-nightly) for the preview documentation.
There's a CI (CircleCI) in-place that guarantees that all examples are always runnable. 

If you want to learn more: https://dlang.org/blog/2017/03/08/editable-and-runnable-doc-examples-on-dlang-org

carblue:
I'm aware, that there's still a lot to be done for std/uni.d, e.g. looking at it's unittest-coverage, but I try to not get bogged down, focus on -dip1000 first and DIP1000.md is 'fresh' in my brain: Thus the reverted unittest will come back later again at another (non-public unittest) location.

carblue:
This PR expired my "languishing acceptance", just closing it now

WalterBright:
Please don't close PRs based on arbitrary time limits. I simply do not have time to work on it at the moment. That doesn't make it a bad PR.

carblue:
This is on my TODO-List for a new PR, as proposed by @wilzbach :
```
std.uni: boost coverage

1. Required anyway
2. https://github.com/dlang/phobos/pull/6041 - according to D style - removes a lot of attributes/storage class from functions, where they get inferred (including @safe and pure).
Thus this PR ensures, that (ref. #6041) every function with @safe/@trusted removed has at least 1 usage in a @safe unittest, same as every function with pure removed has at least 1 usage in a pure unittest.
3. (re-)establish a unittest, that once was reverted from #6041

I don't know any guideline where to place unittest, except recommended directly after a function and not within template,
but current std.uni practice doesn't/can't obey this strictly, thus my decision is arbitrary to
place all new unittest together in the files end. May then later be reorganized and decided which shall serve as public example


#6104 line                                                                 | tested in New PR #? unittest line
----------------------------------------------------------------------------------------------------------
1581: removed nothrow pure @trusted (from private auto packedArrayView(T)) | slice, the (only) function calling packedArrayView, has attributes @nogc nothrow pure;
                                                                           | by-passing -dip1000 safety checks seems to be wrong here (also, no invocation from @safe code happens)

1794: added   pure (to static void destroy(T)(ref T)                       |
1804: added   pure (to static void destroy(T)(ref T)                       |
2067: removed @trusted (from public struct InversionList(SP=GcPolicy))     |
2385: removed @trusted (from byCodepoint(). static struct CodepointRange)  |
2666: added   @safe (to static string binaryScope(R)(R, string)            |
4714: removed pure                                                         | yes, see comment in #6104
4829/4835: removed pure (from public bool match(Range)(ref Range)          |
4853/4859: removed pure @trusted (from public bool skip(Range)(ref Range)) |
4874/4880: removed pure @trusted (from public bool test(Range)(ref Range)) |
4891/4897: removed pure @trusted (from bool match(C)(ref C[]))
4897/4903: removed pure @trusted (from bool skip(C)(ref C[]))
4903/4909: removed pure @trusted (from bool test(C)(ref C[]))
4925/4931: removed @trusted (from auto subMatcher(SizesToPick...)()
4930/4936: removed pure @trusted (from bool lookup(int, Mode, Range)(ref Range))

4995/5001: removed pure (from @property auto tab(int i)())
4996/5002: removed pure (from bool lookup(int, Mode, Range)(ref Range))

5008/5014: added @safe (to void badEncoding())

5058/5064: removed pure @trusted (from public bool match(Range)(ref Range)) |
5084/5090: removed pure @trusted (from public bool skip(Range)(ref Range)   |
5105/5111: removed pure @trusted (from public bool test(Range)(ref Range))  |

5117/5123: removed pure @trusted (from bool match(C)(ref C[] str))
5123/5129: removed pure @trusted (from bool skip(C)(ref C[] str))
5129/5135: removed pure @trusted (from bool test(C)(ref C[] str))

5161/5167: removed @trusted (from auto subMatcher(SizesToPick...)())
5166/5172: removed pure (from bool lookupUni(Mode, Range)(ref Range))

5237/5243: removed pure (from @property auto ascii()())

5240/5246: removed pure (from bool lookupUni(Mode, Range)(ref Range)

5250/5256: removed @trusted (from private auto utf8Matcher(Set)(Set set))
5255/5261: removed @trusted (from private auto utf16Matcher(Set)(Set set))
5267/5273: removed @trusted (from public auto utfMatcher(Char, Set)(Set set))

5283/5289: @nogc nothrow pure @safe (from auto decoder(C)(C[], size_t))

5935/5941: removed pure @trusted (from ptrdiff_t findUnicodeSet(alias table, C)(const scope C[]))
5948/5954: removed pure @trusted (from bool loadUnicodeSet(alias table, Set, C)(const scope C[], ref Set)

5959/5965: removed @trusted (from bool loadProperty(Set=CodepointSet, C)(const scope C[], ref Set))

6116/6122: removed @trusted (from package CodepointSet memoizeExpr(string expr)())

6234/6240: removed @safe (from struct UnicodeSetParser(Range))
```

PR #6044, opened 2018-01-17
Fix Issue 18259 - allocatorObject's CAllocatorImpl should store the passed allocator within

edi33416:


std/experimental/allocator/package.d
```diff
@@ -2030,12 +2031,15 @@ if (!isPointer!A)
...
+        import std.algorithm.mutation : move;
         import std.traits : hasMember;
         static if (hasMember!(A, "deallocate"))
         {
             scope(failure) a.deallocate(state);
         }
-        return cast(CAllocatorImpl!A) emplace!(CAllocatorImpl!A)(state);
+        auto tmp = cast(CAllocatorImpl!A) emplace!(CAllocatorImpl!A)(state);
+        move(a, tmp.impl);
+        return tmp;
```
andralex: Looking below (the else branch) it looks like this implementation would work for that as well. So no need for the two branches - just unite them to use this code. Thanks!

PR #6046, opened 2018-01-19
std.traits: move isDesiredUDA out of getUDAs, since it depends only on attribute and uda.

FeepingCreature:
This conserves memory, since isDesiredUDA instances can be reused between symbols. Usually we want to check a large number of symbols against a small number of possible attributes. With this change, isDesiredUDA only depends on the number of attributes, not spuriously on the symbol they appear on.

In a large project, this brought compiler memory usage from 8.2GB to 7.3GB.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -7668,28 +7668,7 @@ template getUDAs(alias symbol, alias attribute)
 {
     import std.meta : Filter;
 
-    template isDesiredUDA(alias toCheck)
-    {
-        static if (is(typeof(attribute)) && !__traits(isTemplate, attribute))
-        {
-            static if (__traits(compiles, toCheck == attribute))
-                enum isDesiredUDA = toCheck == attribute;
-            else
-                enum isDesiredUDA = false;
-        }
-        else static if (is(typeof(toCheck)))
-        {
-            static if (__traits(isTemplate, attribute))
-                enum isDesiredUDA =  isInstanceOf!(attribute, typeof(toCheck));
-            else
-                enum isDesiredUDA = is(typeof(toCheck) == attribute);
-        }
-        else static if (__traits(isTemplate, attribute))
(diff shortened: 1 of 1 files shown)


quickfur:
IMO we should have more refactorings along these lines, to reduce the amount of template bloat generated by Phobos.  After trying (unsuccessfully, sad to say) to build several D projects on a low-memory system, I have to say that reducing compiler memory usage in any way possible is a big plus in my book.

std/traits.d
```diff
@@ -7668,6 +7668,11 @@ template getUDAs(alias symbol, alias attribute)
 {
     import std.meta : Filter;
 
+    alias getUDAs = Filter!(isDesiredUDA!attribute, __traits(getAttributes, symbol));
+}
+
```
wilzbach: Hmm Ddoc doesn't seem to like this: http://dtest.dlang.io/diff/website-5d84a318ec802dc2fdc00d2f5a8b7565167d489c-f505cb0447f0cc98330d0ee0ce1c53b4/website-5d84a318ec802dc2fdc00d2f5a8b7565167d489c-06d63a0531f58a1b7a505cb68dc55acc/web/phobos-prerelease/std_traits.html

wilzbach: As fixing Ddoc is probably out of reach for the PR, other ideas:
- try whether `/// ditto` works
- move the `private` impl below the public test

wilzbach:
> IMO we should have more refactorings along these lines, to reduce the amount of template bloat generated by Phobos.

Yes, we can do that for the standard library, but the end goal should be to fix the compiler as there is a lot of "template-bloat generating" code already out there.

> After trying (unsuccessfully, sad to say) to build several D projects on a low-memory system, I have to say that reducing compiler memory usage in any way possible is a big plus in my book.

FWIW it also affects the auto-tester:

https://github.com/braddr/d-tester/issues/66

PR #6047, opened 2018-01-22
fix issue 11747 - Better error message with @disabled toString

BBasile:
This is not an important fix. It allows to stop compiling before the linker failure happening when an aggregate has its `to string` marked with `@disable`, so very unlikely to happen, ever. However since i managed to get the new trait accepted in the language, here is an example of usage.

Funnily enough the fix is impossible to test, see the note in the unittest.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -3609,7 +3609,7 @@ if (is(T == class) && !is(T == enum))
     // TODO: Change this once toString() works for shared objects.
     static assert(!is(T == shared), "unable to format shared objects");
 
-    // TODO: remove this check once `@dsiable override` deprecation cycle is finished
+    // TODO: remove this check once `@disable override` deprecation cycle is finished
     static if (__traits(hasMember, T, "toString") && isSomeFunction!(val.toString))
         static assert(!__traits(isDisabled, T.toString), T.stringof ~
             " cannot be formatted because its `toString` is marked with `@disable`");
--- a/std/format.d
+++ b/std/format.d
@@ -3609,6 +3609,11 @@ if (is(T == class) && !is(T == enum))
     // TODO: Change this once toString() works for shared objects.
     static assert(!is(T == shared), "unable to format shared objects");
 
+    // TODO: remove this check once `@dsiable override` deprecation cycle is finished
+    static if (__traits(hasMember, T, "toString") && isSomeFunction!(val.toString))
+        static assert(!__traits(isDisabled, T.toString), T.stringof ~
+            " cannot be formatted because its `toString` is marked with `@disable`");
+
     if (val is null)
         put(w, "null");
     else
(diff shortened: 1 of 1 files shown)


std/format.d
```diff
@@ -3609,6 +3609,11 @@ if (is(T == class) && !is(T == enum))
     // TODO: Change this once toString() works for shared objects.
     static assert(!is(T == shared), "unable to format shared objects");
 
+    // TODO: remove this check once `@dsiable override` deprecation cycle is finished
```
quickfur: Nitpick: typo `@dsiable` -> `@disable`

std/format.d
```diff
@@ -3609,6 +3609,11 @@ if (is(T == class) && !is(T == enum))
     // TODO: Change this once toString() works for shared objects.
     static assert(!is(T == shared), "unable to format shared objects");
 
+    // TODO: remove this check once `@disable override` deprecation cycle is finished
+    static if (__traits(hasMember, T, "toString") && isSomeFunction!(val.toString))
```
quickfur: There's a flaw with this test: if a struct defines `@disable toString` but it isn't a function, `isSomeFunction!(val.toString)` will not compile (this appears to be a bug in `std.traits` or the compiler, will investigate further).  Furthermore, if `toString` is not disabled but also isn't a function, `__traits(isDisabled)` will still work.  So `isSomeFunction` really isn't necessary here.

quickfur: Actually, correction: `__traits(isDisabled)` will *not* return the correct result if `toString` is not a function but annotated with `@disable`.  This is a compiler bug.

BBasile: the trait return false for any declaration that's not a function.

BBasile:
Actually feel free to close the PR and close the issue as invalid or wontfix. The author used to spam bugzilla a lot with small reports like this.

quickfur:
No, this PR is useful, even if it's only a minor issue.  Plus, it did expose a compiler bug: https://issues.dlang.org/show_bug.cgi?id=18310

std/format.d
quickfur: There's a chicken-and-egg problem here.  A struct member that's declared `@disable` cannot even be referenced without producing a compile error, even if passed directly to a `is(T.stringof == function)` test. So there's no way to check whether the symbol is even a function to begin with, and yet `__traits(isDisabled)` returns false for it.  So basically there's no way to avoid a compile error.

quickfur: On second thoughts, the proposed change isn't making things any worse than they are now (compile error with `@disabled`). We can handle the remaining case of declaring `@disable toString` as a non-function later, since it is so rare that it's probably not worth dealing with. The proposed change is bringing *some* improvement, so let's not let the perfect be the enemy of the good.

BBasile:
Sorry about 18130. I dont know how did i miss this problem. I'm almost certain that i tested a non function somewhere.

PR #6048, opened 2018-01-22
Fix erroneous imports

RazvanN7:
This currently blocks : https://github.com/dlang/dmd/pull/7760

### DIFF
--- a/std/experimental/typecons.d
+++ b/std/experimental/typecons.d
@@ -23,8 +23,7 @@ module std.experimental.typecons;
 import std.meta; // : AliasSeq, allSatisfy;
 import std.traits;
 
-import std.typecons : Tuple, tuple, Bind, DerivedFunctionType,
-       isImplicitlyConvertible, mixinAll, staticIota,
+import std.typecons : Tuple, tuple, Bind, DerivedFunctionType, mixinAll, staticIota,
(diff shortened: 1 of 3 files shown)


std/experimental/typecons.d
```diff
@@ -23,8 +23,7 @@ module std.experimental.typecons;
 import std.meta; // : AliasSeq, allSatisfy;
 import std.traits;
 
-import std.typecons : Tuple, tuple, Bind, DerivedFunctionType,
-       isImplicitlyConvertible, mixinAll, staticIota,
+import std.typecons : Tuple, tuple, Bind, DerivedFunctionType, mixinAll, staticIota,
```
wilzbach: We can even remove `mixinAll` + `staticIota` entirely -> https://github.com/dlang/phobos/pull/6049

andralex:
Will force merge in the interest of time.

wilzbach:
> Will force merge in the interest of time.

While I like the enthusiasm and action, I think it would be best of force merge is __only__  used if __absolutely__ necessary, a criteria not fulfilled by this PR:

1) It would have been merged in ~20 mins anyways:

image

2)  dlang/dmd#7760 needs to be changed to emit deprecations anyways

PR #6052, opened 2018-01-22
Remove staticIota: std_regex_internal_backtracking

wilzbach:


wilzbach:
Probably the tests were moved or changed and removing the global alias was forgotten.

PR #4265, opened 2016-04-29
std.algorithm.searching: no mapping-specialization for extremum

wilzbach:
As #4257 (exposing the extremum function) depends on andralex, I thought I separate the optimization for the noop-function case in a separate PR. Here it goes. The idea is to reduce the unnecessary element and function call if no mapping is specifiecd, e.g. from:

```
MapType mapElement = mapFun(r[i]);
if (selectorFun(mapElement, extremeElementMapped))
{
   extremeElement = r[i];
   extremeElementMapped = mapElement;
}
```

this PR reduces it down to:

```
if (selectorFun(r[i], extremeElement))
{
   extremeElement = r[i];
}
```

Moreover a small goodie was added that checks whether the first template function can be used as an mapping function (unary function) and otherwise it falls back to the selector (binary function). Useful in terms of #4257

PR #6056, opened 2018-01-22
Fix Issue 18280 - std.algorithm.comparison.cmp for non-strings should call opCmp only once per item pair

n8sh:
Right now when comparing non-string ranges std.algorithm.comparison.cmp calls the comparison predicate twice for every pair of elements to determine their order. When the elements have overloaded `opCmp` and the predicate is "a < b" we only need to call `opCmp` once for each pair.

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -580,43 +580,64 @@ do
 
 // cmp
 /**********************************
-Performs three-way lexicographical comparison on two
-$(REF_ALTTEXT input ranges, isInputRange, std,range,primitives)
-according to predicate `pred`. Iterating `r1` and `r2` in
-lockstep, `cmp` compares each element `e1` of `r1` with the
-corresponding element `e2` in `r2`. If one of the ranges has been
-finished, `cmp` returns a negative value if `r1` has fewer
-elements than `r2`, a positive value if `r1` has more elements
-than `r2`, and `0` if the ranges have the same number of
-elements.
+Performs a lexicographical comparison on two
+$(REF_ALTTEXT input ranges, isInputRange, std,range,primitives).
+Iterating `r1` and `r2` in lockstep, `cmp` compares each element
+`e1` of `r1` with the corresponding element `e2` in `r2`. If one
+of the ranges has been finished, `cmp` returns a negative value
+if `r1` has fewer elements than `r2`, a positive value if `r1`
+has more elements than `r2`, and `0` if the ranges have the same
+number of elements.
 
 If the ranges are strings, `cmp` performs UTF decoding
 appropriately and compares the ranges one code point at a time.
 
+A custom predicate may be specified, in which case `cmp` performs
+a three-way lexicographical comparison using `pred`. Otherwise
+the elements are compared using `opCmp`.
+
 Params:
-    pred = The predicate used for comparison.
+    pred = Predicate used for comparison. Without a predicate
+        specified the ordering implied by `opCmp` is used.
     r1 = The first range.
     r2 = The second range.
 
 Returns:
-    0 if both ranges compare equal. -1 if the first differing element of $(D
-    r1) is less than the corresponding element of `r2` according to $(D
-    pred). 1 if the first differing element of `r2` is less than the
-    corresponding element of `r1` according to `pred`.
+    `0` if the ranges compare equal. A negative value if `r1` is a prefix of `r2` or
+    the first differing element of `r1` is less than the corresponding element of `r2`
+    according to `pred`. A positive value if `r2` is a prefix of `r1` or the first
+    differing element of `r2` is less than the corresponding element of `r1`
+    according to `pred`.
 
+Note:
+    An earlier version of the documentation incorrectly stated that `-1` is the
+    only negative value returned and `1` is the only positive value returned.
+    Whether that is true depends on the types being compared.
 */
-int cmp(alias pred = "a < b", R1, R2)(R1 r1, R2 r2)
+auto cmp(R1, R2)(R1 r1, R2 r2)
 if (isInputRange!R1 && isInputRange!R2)
 {
     static if (!(isSomeString!R1 && isSomeString!R2))
     {
         for (;; r1.popFront(), r2.popFront())
         {
-            if (r1.empty) return -cast(int)!r2.empty;
-            if (r2.empty) return !r1.empty;
-            auto a = r1.front, b = r2.front;
-            if (binaryFun!pred(a, b)) return -1;
-            if (binaryFun!pred(b, a)) return 1;
+            static if (is(typeof(r1.front.opCmp(r2.front)) R))
+                alias Result = R;
+            else
+                alias Result = int;
+            if (r2.empty) return Result(!r1.empty);
+            if (r1.empty) return Result(-1);
+            static if (is(typeof(r1.front.opCmp(r2.front))))
+            {
+                auto c = r1.front.opCmp(r2.front);
+                if (c != 0) return c;
+            }
+            else
+            {
+                auto a = r1.front, b = r2.front;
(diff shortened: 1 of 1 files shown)


std/algorithm/comparison.d
```diff
@@ -610,13 +610,26 @@ if (isInputRange!R1 && isInputRange!R2)
 {
     static if (!(isSomeString!R1 && isSomeString!R2))
     {
+        static if (is(typeof(pred) : string))
+            enum isLessThan = pred == "a < b";
```
wilzbach: FYI: @andralex isn't a huge fan of string comparisons, see https://github.com/dlang/phobos/pull/4265#issuecomment-266234691

std/algorithm/comparison.d
```diff
@@ -610,13 +610,26 @@ if (isInputRange!R1 && isInputRange!R2)
...
         for (;; r1.popFront(), r2.popFront())
         {
             if (r1.empty) return -cast(int)!r2.empty;
             if (r2.empty) return !r1.empty;
-            auto a = r1.front, b = r2.front;
-            if (binaryFun!pred(a, b)) return -1;
-            if (binaryFun!pred(b, a)) return 1;
+            static if (isLessThan && is(typeof(r1.front.opCmp(r2.front)) : int))
+            {
+                immutable int c = r1.front.opCmp(r2.front);
```
wilzbach: Why not `r1.front < r2.front`?

std/algorithm/comparison.d
wilzbach: As I did a bit of digging the string specialization of `cmp` was added in 2011: https://github.com/dlang/phobos/commit/a0ecf2a10e319e387e269fd6310ad2acd244cc92

andralex: Good news, https://github.com/dlang/dmd/pull/7484/ is on. With it we should be able to switch to lambdas soon. cc @RazvanN7

std/algorithm/comparison.d
```diff
@@ -689,6 +698,54 @@ if (isInputRange!R1 && isInputRange!R2)
...
 }
 
+/// ditto
+int cmp(alias pred, R1, R2)(R1 r1, R2 r2)
+if (isInputRange!R1 && isInputRange!R2)
+{
+    static if (is(typeof(pred) : string)
+        && (__traits(compiles, { static assert("a < b" == pred); })))
+    {
+        // In case someone was explicitly using the default predicate.
```
andralex: please remove, no need to accommodate this case

n8sh: Removed

n8sh:
>I committed a few changes.

Sorry, I clobbered your changes when I force pushed to add a unittest.

andralex:
OK, I submitted them again.

I think since we got here we should fix the `float` return. So cmp should return auto and let opCmp drive.

n8sh:
For float `opCmp` what do we do with a NaN result? "Not positive or negative, so we keep going" or "not zero, so we return it"?

andralex:
@n8sh Stop the comparison and forward the NaN result. So the core of the loop would be:

```d
auto c = r1.front.opCmp(r2.front);
if (c != 0) return c;
```

std/algorithm/comparison.d
```diff
@@ -614,40 +623,41 @@ if (isInputRange!R1 && isInputRange!R2)
         {
             if (r1.empty) return -cast(int)!r2.empty;
             if (r2.empty) return !r1.empty;
```
andralex: For these two returns you need a bit of massaging. I'd say something like this:

```d
static if (is(typeof(r1.front.opCmp(r2.front))) R)
    alias Result = R;
else
    alias Result = int;
if (r1.empty) return -R(!r2.empty);
if (r2.empty) return R(!r1.empty);
...
```

std/algorithm/comparison.d
```diff
@@ -724,6 +776,60 @@ if (isInputRange!R1 && isInputRange!R2)
...
+@safe unittest
+{
+    int result;
+
+    result = cmp!"a > b"("abc", "abc");
+    assert(result == 0);
+    result = cmp!"a > b"("", "");
+    assert(result == 0);
+    result = cmp!"a > b"("abc", "abcd");
+    assert(result > 0);
```
n8sh: This behavior is a bug. r1 is a prefix of r2, so the result should be negative even if we are comparing in reverse alphabetical order, but `cmp` is using `pred` to compared the lengths! I have no sense of whether anyone might have been relying on this for reverse order comparison. It is a violation of the documentation and it only works for strings. Should it be fixed?

std/algorithm/comparison.d
n8sh: Fixed

std/algorithm/comparison.d
andralex: That's terrible. I'd say open a bug with this repro then mark it as fixed by this PR. That way we'll have it in the changelog.

andralex:
@n8sh you'll need to add a few more unittests because now you are generalizing cmp. Sorry this fell on you and thanks for the quality code!

n8sh:
Found another bad bug. This code:
```if (c1 != c2) return threeWayInt(cast(int) c1, cast(int) c2);```
meant that no custom predicate that treats different values the same (for example, ignoring case) will work correctly. They'll incorrectly return 0 prematurely.

Example https://run.dlang.io/is/feq4KK
```d
void main()
{
    import std.algorithm.comparison : cmp;
    static bool lessThanCaseInsensitive(size_t a, size_t b)
    {
        import std.ascii : toUpper;
        return toUpper(cast(dchar) a) < toUpper(cast(dchar) b);
    }
    static assert(cmp!lessThanCaseInsensitive("apple2", "APPLE1") != 0,
                 "These are clearly not the same!");
}
```

andralex:
@n8sh it's a shame these haven't been exposed already. Thanks for your work! Please file that bug, too (or massage it with the other) and then mark them as fixed with this PR.

andralex:
@n8sh I made a push to your code that removes `cast` in promotions.

andralex:
Approved, please squash and then we'll merge. Thanks!!

andralex:
Aaand made one more minor optimization - `empty` was sometimes checked twice at the end of the ranges.

ZombineDev:
@n8sh perhaps the way forward is to make it an option but also keep the existing behavior by querying the predicate type via static if.

andralex:
There's always another name such as `cmp3`... but I'd say distinguishing a predicate that returns `bool` from one that returns anything else is a fine option too.

aG0aep6G:
The bot picked up 4 issues here, but it only closed one. Bug?
https://github.com/dlang-bots/dlang-bot/issues/175

n8sh:
After @aG0aep6G's message I closed the issues manually, but I don't know if that might result in them not getting added to the next release changelog.

wilzbach:
> After @aG0aep6G's message I closed the issues manually, but I don't know if that might result in them not getting added to the next release changelog.

A bit of explanation:

The bugs are closed by the Bugzilla integration - not the bot.
The changelog is generated by https://github.com/dlang/tools/blob/master/changed.d - it parses the git log between two revisions with the same regex as the bot and the bugzilla integration does. However, it then queries Bugzilla and filters for only referenced bugs which are closed.
So the bug overview by the bot is merely a help for reviewers as (a) it provides a handy link to the references issues and (b) allows them to verify that the issues have been referenced correctly.
See also: https://github.com/dlang-bots/dlang-bot#automated-references

Now to the bug - it looks like the Bugzilla integration is a bit dumber and only picks up one bug regex per commit. There's much I can do about (except for writing our own Bugzilla integration).
In any case, a "workaround" for the future if you have to use one commit is to include multiple issues in one regex, e.g. "Fixes issues 17494, 17505, 17506".

PR #6057, opened 2018-01-23
Fix erroneous imports in unittests

RazvanN7:
This is part 2 of #6048 and is currently blocking : dlang/dmd#7760 because deprecations are treated as errors.

### DIFF
--- a/std/socket.d
+++ b/std/socket.d
@@ -85,10 +85,10 @@ else version(Posix)
         }
     }
 
+    public import core.sys.posix.netinet.in_;
     import core.sys.posix.arpa.inet;
     import core.sys.posix.fcntl;
     import core.sys.posix.netdb;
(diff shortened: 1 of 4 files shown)


std/socket.d
```diff
@@ -85,10 +85,10 @@ else version(Posix)
         }
     }
 
+    public import core.sys.posix.netinet.in_;
```
wilzbach: Argh - this is needed because the Windows branch is `public import core.sys.windows.winsock2;`

RazvanN7:
The fail in jenkins seems unrelated. @atilaneves does this affect your project?

atilaneves:
@RazvanN7 I think that's a spurious error that has to do with an stdio bug. I'm confident that if rerun it'll be green.

RazvanN7:
@atilaneves ok. thanks!

PR #6059, opened 2018-01-23
Add policy hooks to std.json to support custom extensions

marler8997:
This is an exploration on enhancing `std.json`.  The idea is to add support to preserve order of object members.  An example of where this is useful is if you'd like to parse some JSON and then later write it somewhere, but you'd like to maintain the original order of the object members.

Note that the techniques used in this PR could also be reused for adding other optional features.  It allows the application to pass in a Policy structure that can enable/disable features.  If you wanted to add another feature, say, "enableComments", this could be added as a Policy setting as well.  This technique could also be ported to the upcoming `std.data.json`.

Preserving field order is beneficial for any kind of tool that modifies JSON files.  You could imagine tools that modify `dub.json` files, DMD JSON output files, sublime/vscode config files.  If a tool like this couldn't guarantee preserving the order of object fields, then this would make the tools much less usable.

NOTE: This PR is not fully documented.  The work to fully document is pending approval that this is a desirable enhancement.

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -96,11 +96,34 @@ enum JSON_TYPE : byte
     FALSE    /// ditto
 }
 
+private struct DefaultPolicy { }
+
 /**
 JSON value node
 */
-struct JSONValue
+alias JSONValue = JSONValueTemplate!DefaultPolicy;
+
+/**
+A JSONValue template. Features can be enabled/disabled via setting members in $(D Policy).
+Example:
+------------
+struct MyPolicy
 {
+    enum sortObjectMembers = false;
+    alias MyObjectType(K,V) = OrderedAA!(K,V);
+}
+alias MyJSONValue = JSONValueTemplate!MyPolicy;
+------------
+**/
+struct JSONValueTemplate(alias Policy)
+{
+    alias Value = typeof(this);
+
+    static if (hasMember!(Policy, "ObjectType"))
+        alias JSONObject = Policy.ObjectType!(string, Value);
+    else
+        alias JSONObject = Value[string];
+
     import std.exception : enforceEx, enforce;
 
     union Store
@@ -109,8 +132,8 @@ struct JSONValue
         long                            integer;
         ulong                           uinteger;
         double                          floating;
-        JSONValue[string]               object;
-        JSONValue[]                     array;
+        JSONObject                      object;
+        Value[]                         array;
     }
     private Store store;
     private JSON_TYPE type_tag;
@@ -126,7 +149,7 @@ struct JSONValue
     @safe unittest
     {
           string s = "{ \"language\": \"D\" }";
-          JSONValue j = parseJSON(s);
+          Value j = parseJSONTemplate!Policy(s);
           assert(j.type == JSON_TYPE.OBJECT);
           assert(j["language"].type == JSON_TYPE.STRING);
     }
@@ -151,7 +174,7 @@ struct JSONValue
     ///
     @safe unittest
     {
-        JSONValue j = [ "language": "D" ];
+        Value j = [ "language": "D" ];
 
         // get value
         assert(j["language"].str == "D");
@@ -227,14 +250,14 @@ struct JSONValue
        (*a)["hello"] = "world";  // segmentation fault
        ---
      */
-    @property ref inout(JSONValue[string]) object() inout pure @system
+    @property ref inout(JSONObject) object() inout pure @system
     {
         enforce!JSONException(type == JSON_TYPE.OBJECT,
                                 "JSONValue is not an object");
         return store.object;
     }
     /// ditto
-    @property JSONValue[string] object(JSONValue[string] v) pure nothrow @nogc @safe
+    @property JSONObject object(JSONObject v) pure nothrow @nogc @safe
     {
         assign(v);
         return v;
@@ -255,7 +278,7 @@ struct JSONValue
      * Throws: $(D JSONException) for read access if $(D type) is not
      * $(D JSON_TYPE.OBJECT).
      */
-    @property inout(JSONValue[string]) objectNoRef() inout pure @trusted
+    @property inout(JSONObject) objectNoRef() inout pure @trusted
     {
         enforce!JSONException(type == JSON_TYPE.OBJECT,
                                 "JSONValue is not an object");
@@ -273,14 +296,14 @@ struct JSONValue
        (*a)[0] = "world";  // segmentation fault
        ---
      */
-    @property ref inout(JSONValue[]) array() inout pure @system
+    @property ref inout(Value[]) array() inout pure @system
     {
         enforce!JSONException(type == JSON_TYPE.ARRAY,
                                 "JSONValue is not an array");
         return store.array;
     }
     /// ditto
-    @property JSONValue[] array(JSONValue[] v) pure nothrow @nogc @safe
+    @property Value[] array(Value[] v) pure nothrow @nogc @safe
     {
         assign(v);
         return v;
@@ -302,7 +325,7 @@ struct JSONValue
      * Throws: $(D JSONException) for read access if $(D type) is not
      * $(D JSON_TYPE.ARRAY).
      */
-    @property inout(JSONValue[]) arrayNoRef() inout pure @trusted
+    @property inout(Value[]) arrayNoRef() inout pure @trusted
     {
         enforce!JSONException(type == JSON_TYPE.ARRAY,
                                 "JSONValue is not an array");
@@ -355,40 +378,38 @@ struct JSONValue
             type_tag = JSON_TYPE.FLOAT;
             store.floating = arg;
         }
-        else static if (is(T : Value[Key], Key, Value))
+        else static if (is(T : JSONObject))
+        {
+            type_tag = JSON_TYPE.OBJECT;
+            JSONObject t = arg;
+            () @trusted { store.object = t; }();
+        }
+        else static if (is(T : AAValue[Key], Key, AAValue))
         {
             static assert(is(Key : string), "AA key must be string");
             type_tag = JSON_TYPE.OBJECT;
-            static if (is(Value : JSONValue))
-            {
-                JSONValue[string] t = arg;
-                () @trusted { store.object = t; }();
-            }
-            else
-            {
-                JSONValue[string] aa;
-                foreach (key, value; arg)
-                    aa[key] = JSONValue(value);
-                () @trusted { store.object = aa; }();
-            }
+            JSONObject aa;
+            foreach (key, value; arg)
+                aa[key] = Value(value);
+            () @trusted { store.object = aa; }();
         }
         else static if (isArray!T)
         {
             type_tag = JSON_TYPE.ARRAY;
-            static if (is(ElementEncodingType!T : JSONValue))
+            static if (is(ElementEncodingType!T : Value))
             {
-                JSONValue[] t = arg;
+                Value[] t = arg;
                 () @trusted { store.array = t; }();
             }
             else
             {
-                JSONValue[] new_arg = new JSONValue[arg.length];
+                Value[] new_arg = new Value[arg.length];
                 foreach (i, e; arg)
-                    new_arg[i] = JSONValue(e);
+                    new_arg[i] = Value(e);
                 () @trusted { store.array = new_arg; }();
             }
         }
-        else static if (is(T : JSONValue))
+        else static if (is(T : Value))
         {
             type_tag = arg.type;
             store = arg.store;
@@ -402,15 +423,15 @@ struct JSONValue
     private void assignRef(T)(ref T arg) if (isStaticArray!T)
     {
         type_tag = JSON_TYPE.ARRAY;
-        static if (is(ElementEncodingType!T : JSONValue))
+        static if (is(ElementEncodingType!T : Value))
         {
             store.array = arg;
         }
         else
         {
-            JSONValue[] new_arg = new JSONValue[arg.length];
+            Value[] new_arg = new Value[arg.length];
             foreach (i, e; arg)
-                new_arg[i] = JSONValue(e);
+                new_arg[i] = Value(e);
             store.array = new_arg;
         }
     }
@@ -437,7 +458,7 @@ struct JSONValue
         assignRef(arg);
     }
     /// Ditto
-    this(T : JSONValue)(inout T arg) inout
+    this(T : Value)(inout T arg) inout
     {
         store = arg.store;
         type_tag = arg.type;
@@ -445,17 +466,17 @@ struct JSONValue
     ///
     @safe unittest
     {
-        JSONValue j = JSONValue( "a string" );
-        j = JSONValue(42);
+        Value j = Value( "a string" );
+        j = Value(42);
 
-        j = JSONValue( [1, 2, 3] );
+        j = Value( [1, 2, 3] );
         assert(j.type == JSON_TYPE.ARRAY);
 
-        j = JSONValue( ["language": "D"] );
+        j = Value( ["language": "D"] );
         assert(j.type == JSON_TYPE.OBJECT);
     }
 
-    void opAssign(T)(T arg) if (!isStaticArray!T && !is(T : JSONValue))
+    void opAssign(T)(T arg) if (!isStaticArray!T && !is(T : Value))
     {
         assign(arg);
     }
@@ -469,7 +490,7 @@ struct JSONValue
      * Array syntax for json arrays.
      * Throws: $(D JSONException) if $(D type) is not $(D JSON_TYPE.ARRAY).
      */
-    ref inout(JSONValue) opIndex(size_t i) inout pure @safe
+    ref inout(Value) opIndex(size_t i) inout pure @safe
     {
         auto a = this.arrayNoRef;
         enforceEx!JSONException(i < a.length,
@@ -479,7 +500,7 @@ struct JSONValue
     ///
     @safe unittest
     {
-        JSONValue j = JSONValue( [42, 43, 44] );
+        Value j = Value( [42, 43, 44] );
         assert( j[0].integer == 42 );
         assert( j[1].integer == 43 );
     }
@@ -488,16 +509,16 @@ struct JSONValue
      * Hash syntax for json objects.
      * Throws: $(D JSONException) if $(D type) is not $(D JSON_TYPE.OBJECT).
      */
-    ref inout(JSONValue) opIndex(string k) inout pure @safe
+    ref inout(Value) opIndex(string k) inout pure @safe
     {
         auto o = this.objectNoRef;
         return *enforce!JSONException(k in o,
-                                        "Key not found: " ~ k);
+                                      "Key not found: " ~ k);
     }
     ///
     @safe unittest
     {
-        JSONValue j = JSONValue( ["language": "D"] );
+        Value j = Value( ["language": "D"] );
         assert( j["language"].str == "D" );
     }
 
@@ -514,7 +535,7 @@ struct JSONValue
     {
         enforceEx!JSONException(type == JSON_TYPE.OBJECT || type == JSON_TYPE.NULL,
                                 "JSONValue must be object or null");
-        JSONValue[string] aa = null;
+        JSONObject aa = null;
         if (type == JSON_TYPE.OBJECT)
         {
             aa = this.objectNoRef;
@@ -526,9 +547,9 @@ struct JSONValue
     ///
     @safe unittest
     {
-            JSONValue j = JSONValue( ["language": "D"] );
-            j["language"].str = "Perl";
-            assert( j["language"].str == "Perl" );
+        Value j = Value( ["language": "D"] );
+        j["language"].str = "Perl";
+        assert( j["language"].str == "Perl" );
     }
 
     void opIndexAssign(T)(T arg, size_t i) pure
@@ -542,21 +563,21 @@ struct JSONValue
     ///
     @safe unittest
     {
-            JSONValue j = JSONValue( ["Perl", "C"] );
+            Value j = Value( ["Perl", "C"] );
             j[1].str = "D";
             assert( j[1].str == "D" );
     }
 
-    JSONValue opBinary(string op : "~", T)(T arg) @safe
+    Value opBinary(string op : "~", T)(T arg) @safe
     {
         auto a = this.arrayNoRef;
         static if (isArray!T)
         {
-            return JSONValue(a ~ JSONValue(arg).arrayNoRef);
+            return Value(a ~ Value(arg).arrayNoRef);
         }
-        else static if (is(T : JSONValue))
+        else static if (is(T : Value))
         {
-            return JSONValue(a ~ arg.arrayNoRef);
+            return Value(a ~ arg.arrayNoRef);
         }
         else
         {
@@ -569,9 +590,9 @@ struct JSONValue
         auto a = this.arrayNoRef;
         static if (isArray!T)
         {
-            a ~= JSONValue(arg).arrayNoRef;
+            a ~= Value(arg).arrayNoRef;
         }
-        else static if (is(T : JSONValue))
+        else static if (is(T : Value))
         {
             a ~= arg.arrayNoRef;
         }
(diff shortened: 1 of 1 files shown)


std/json.d
```diff
@@ -96,11 +96,72 @@ enum JSON_TYPE : byte
...
+                if (name == member.name)
+                    return &member.value;
+                return null;
+            }
+            ref auto opIndex(in char[] name) const
+            {
+                foreach (ref member; members)
+                if (name == member.name)
+                    return member.value;
+                assert(0, "JSON object does not contain member named: " ~ name);
```
timotheecour: enforce for external / environmental errors

std/json.d
```diff
@@ -96,11 +96,72 @@ enum JSON_TYPE : byte
...
+    enum preserveMemberOrder = true;
+}
+
 /**
 JSON value node
 */
-struct JSONValue
+alias JSONValue = JSONValueTemplate!DefaultPolicy;
+
+private struct JSONValueTemplate(alias Policy)
```
timotheecour: not private

std/json.d
```diff
@@ -96,11 +96,72 @@ enum JSON_TYPE : byte
...
+            {
+                foreach (ref member; members)
+                if (name == member.name)
+                    return &member.value;
+                return null;
+            }
+            ref auto opIndex(in char[] name) const
+            {
+                foreach (ref member; members)
+                if (name == member.name)
```
timotheecour: missing indent

std/json.d
```diff
@@ -96,11 +96,72 @@ enum JSON_TYPE : byte
...
+                    return member.value;
+                assert(0, "JSON object does not contain member named: " ~ name);
+            }
+            ref auto opIndexAssign(Value value, string name)
```
timotheecour: ref auto => u probably wanna return *this ?

std/json.d
```diff
@@ -96,11 +96,72 @@ enum JSON_TYPE : byte
...
 */
-struct JSONValue
+alias JSONValue = JSONValueTemplate!DefaultPolicy;
+
+private struct JSONValueTemplate(alias Policy)
 {
+    alias Value = typeof(this);
+    static if (Policy.preserveMemberOrder)
+    {
+        struct JSONObject
```
timotheecour: optional: this is basically an AA with ordered fields; it's re-usable; why not expose it as a generic:
```
struct AAOrdered(K,V){
...
}
```

std/json.d
```diff
@@ -96,11 +96,86 @@ enum JSON_TYPE : byte
...
+private struct DefaultPolicy
+{
+    enum preserveMemberOrder = false;
+}
+private struct PreserveMemberOrderPolicy
+{
+    enum preserveMemberOrder = true;
+}
+
+struct OrderedAA(K,V)
```
timotheecour: move to std.array? (eg has assocArray etc)

std/json.d
```diff
@@ -96,11 +96,86 @@ enum JSON_TYPE : byte
...
+{
+    struct Member
+    {
+        K name;
+        V value;
+    }
+    private Member[] members;
+    size_t length() const { return members.length; }
+    inout(V)* opIn(in char[] name) inout
```
timotheecour: char[] instead of K ? (do we need specialization for case K==string to allow char[] in that case?)
=> a unitest with orderedAA!(int, double) would've caught this bug

std/json.d
```diff
@@ -96,11 +96,86 @@ enum JSON_TYPE : byte
...
+    {
+        foreach (ref member; members)
+        if (name == member.name)
+            return &member.value;
+        return null;
+    }
+    ref auto opIndex(in char[] name) const
```
timotheecour: ditto

std/json.d
```diff
@@ -96,11 +96,86 @@ enum JSON_TYPE : byte
...
+    {
+        foreach (ref member; members)
+        if (name == member.name)
+            return member.value;
+        assert(0, "OrderedAA does not contain member named: " ~ name);
+    }
+    void opIndexAssign(V value, K name)
```
timotheecour: s/name/key/ everywhere in this struct

std/json.d
```diff
@@ -96,11 +96,86 @@ enum JSON_TYPE : byte
...
+        if (name == member.name)
+            return &member.value;
+        return null;
+    }
+    ref auto opIndex(in char[] name) const
+    {
+        foreach (ref member; members)
+        if (name == member.name)
+            return member.value;
+        assert(0, "OrderedAA does not contain member named: " ~ name);
```
timotheecour: enforce!JSONException ?

std/json.d
```diff
@@ -1239,35 +1353,53 @@ string toJSON(const ref JSONValue root, in bool pretty = false, in JSONOptions o
                     putCharAndEOL('{');
                     bool first = true;
 
-                    void emit(R)(R names)
+                    static if (Policy.preserveMemberOrder)
                     {
-                        foreach (name; names)
+                        foreach (member; obj.members)
```
timotheecour: foreach (key, value; obj.members) {...}

marler8997: Actually, in the DefaultPolicy case, `names` is just an array of strings, not an associative array.

std/json.d
```diff
@@ -355,40 +433,46 @@ struct JSONValue
...
             {
-                JSONValue[string] aa;
+                JSONObject aa;
                 foreach (key, value; arg)
-                    aa[key] = JSONValue(value);
+                    aa[key] = Value(value);
                 () @trusted { store.object = aa; }();
             }
         }
+        else static if (Policy.preserveMemberOrder && is(T == JSONObject))
```
timotheecour: that seems restrictive compared to other policy which didn't only accept JSONObject but also things that converted to it via `Key : string` and `Value(value)`

marler8997: Maybe.  It seems a bit odd to take something that is unordered and then copy it to something that is ordered.  Not sure.

marler8997: An application could always do the conversion themselves to.  This is something that doesn't have to be implemented in the library (AFAICT).

timotheecour: no, it could be AAordered!(Key, Value) with Key and Value convertible to destination types

std/array.d
```diff
@@ -3889,3 +3889,50 @@ unittest
...
+            if (key == member.key)
+                return &member.value;
+        return null;
+    }
+    ref auto opIndex(in K key) const
+    {
+        foreach (ref member; array)
+        if (key == member.key)
+            return member.value;
+        assert(0, "OrderedAA does not contain value with key: " ~ key);
```
timotheecour: * `~ key` will fail with non-string types
* unittest to catch that eg OrderedAA!(int, double)
* not sure if assert is appropriate; it probably should be same type of error as corresponding error in an  associative array

std/array.d
```diff
@@ -3889,3 +3889,50 @@ unittest
...
 }
+
+struct OrderedAA(K,V)
+{
+    struct KeyValue
+    {
+        K key;
+        V value;
+    }
+    private KeyValue[] array;
```
timotheecour: is that more efficient with `Appender!KeyValue`?

std/array.d
```diff
@@ -3889,3 +3889,50 @@ unittest
     assert(appS.data == "hellow");
     assert(appA.data == "hellow");
 }
+
+struct OrderedAA(K,V)
+{
+    struct KeyValue
```
timotheecour: static struct? to avoid hidden __this pointer

JackStouffer:
This seems like a misapplication of JSON. JSON objects are designed to be unordered; from the spec:

>An object is an **unordered** set of name/value pairs. 

My inclination is that if what you're doing with JSON requires you to keep the same order of keys, then there's something wrong with your design.

quickfur:
Yeah, let's not implement features that are not in the JSON spec. We don't want user code to start relying on non-standard behaviour.

If a fixed order is required in JSON, an array can always be used instead:
````
[
    { "name": "field1", "value": value1 },
    { "name": "field2", "value": value2 },
    ...
]
````
instead of
````
{
    "field1": value1,
    "field2": value2,
    ...
}
````

**Comment by marler8997** [comment]
> This seems like a misapplication of JSON. JSON objects are designed to be unordered; from the spec:
> > An object is an unordered set of name/value pairs.

> My inclination is that if what you're doing with JSON requires you to keep the same order of keys, then there's something wrong with your design.

I don't blame you since that was also my initial inclination :)

In the description I mentioned an application where this is useful, namely, if you'd like to parse some JSON and at a later time write it back out somewhere while preserving the order of object members.  The specific application we would use this for is in the JSON generated by DMD.  We'd like to "sanitize" the JSON by normalizing/removing environment specific values so it can be compared with an expected output.  In this case it's nice to keep the original order of the object members since they are already ordered nicely when they are generated and changing that order makes it harder to compare it with an expected JSON file and makes the JSON less readable.

I questioned this use case initially but after thinking about it I decided it makes sense.  Note that when the JSON spec says an object is an "unordered set of name/value pairs" it is saying that the order doesn't matter when it comes to the data structures, however, that doesn't mean an application or user doesn't care about the order.  For example, when you look at a `dub.json` file, the order of each property has some importance.  Most `dub.json` files start with the projection information such as name/description/author.  It would be quite odd to suddenly put the platform specific settings/compiler flags at the beginning.  If you wanted to write a tool that modified the JSON file, you would want a JSON parser that maintained the original order of these fields would you not?

**Comment by timotheecour** [comment]
(copied from private slack discussion with @marler8997 )

other use cases of your json_ordered (which could be mentioned in your PR): 
* clang-format equivalent for json based config files (eg dub.json, sublime text config files etc), that doesn’t change field order (if so ppl would not use it); likewise with tooling to process json files (eg json aware string replace)
* json based serialization (eg orange library): D structs (and tuples) are ordered but order is lost if using json object to represent them; this could be undesirable
* supports more efficient comparison based on memcmp instead of json parsing; this is especially relevant for fast lookups; eg this explains why bson (binary json extension) DOES support field order preserving in mongodb (cf rationale here:  https://softwareengineering.stackexchange.com/questions/319468/why-does-mongodb-check-for-order-of-keys-when-matching-embedded-documents)

**Comment by CyberShadow** [comment]
I did end up implementing optional preservation of object field order in my JSON library, because there are occasions where this is genuinely useful (e.g. you need readable diffs of produced files, which is pretty much always the case when you want them in version control).

It looks like this implementation is no longer using a hashmap for lookup, which could result in a performance regression. FWIW, here's my (pretty basic) `OrderedMap` implementation: https://github.com/CyberShadow/ae/blob/ab685d4ff0994f7a6656a86b981ce2d106e4be95/utils/aa.d#L141

**Comment by marler8997** [comment]
benchmark would be good here.  I think the performance will depend on the key type.  For strings the hash will likely be better for larger data sets (prevents accessing loading cache for a bunch of strings that don't matter), however, for "immediate value" keys (like integers) the array would likely be faster (https://www.youtube.com/watch?v=YQs6IC-vgmo).  Since JSON uses keys for strings, I'm sure your orderedAA would end up being faster given large enough data sets.  I'll see if I can get a benchmark.

**Comment by marler8997** [comment]
Wow I don't know why, but the new `AssocRange` data structure included in this PR is WAAAAY slower than `OrderedMap` from (https://github.com/CyberShadow/ae/blob/ab685d4ff0994f7a6656a86b981ce2d106e4be95/utils/aa.d#L141).

Inserting 100,000 elements using `size_t` as an index took 6 seconds for `AssocRange` and only 28 milliseconds for `OrderedMap`.  That just goes to show that you should always measure (as Andrei says).  I'm not sure exactly why it's so much slower, might dig a little deeper to figure out why but that's the verdict for now.

**Comment by marler8997** [comment]
Adding some performance logging, it looks like the bulk of the wasted time is in the loop to lookup whether an existing key exists.  Looks like using a hash is the way to go here.

**Comment by quickfur** [comment]
@timotheecour  You have a point that tools that operate on JSON would prefer to preserve input ordering, in order to avoid unnecessary reorderings that are unrelated to the actual change being applied, which could result in noisy diffs, for example.

I don't agree about using JSON objects to represent structs or tuples, though. Usually, in such cases, the receiving end already has the struct/tuple definition with the right ordering, and the code would just transcribe from JSON in an order-agnostic way. In fact, the lack of ordering here is actually an advantage, because you could be communicating between two programs that define the struct/tuple fields in a different ordering, but transcribing the fields by name means you never have to worry about the ordering used on the sending end.

OTOH, if you want to *derive* struct/tuple definitions from JSON, then IMO you should be using a representation that *explicitly* encodes the field ordering, instead of assuming order in something that the JSON spec explicitly states is unordered. Relying on this seems to me to be just an unwise way to code, and opens the door to implementation-dependent behaviour and general flakiness.  If you want ordering, use an array or an object with an explicit ordering index:
````
[
    { "fieldname": "name", "fieldvalue": "John Doe" },
    { "fieldname": "phone", "fieldvalue": "123-456-7890" },
    ...
]
// or:
{
    "name": { "fieldIndex": 1, "fieldvalue": "John Doe" },
    "phone": { "fieldIndex": 2, "fieldvalue": "123-456-7890" },
    ...
}
````

And on that note, @marler8997, the fact that in the first example above `fieldname` and `fieldvalue` may be ordered differently is not a problem, because the code should be looking them up by name rather than by position anyway.

**Comment by marler8997** [comment]
@quickfur Are saying that tools that modify JSON files while preserving the order don't need to exist?

Say I want to write a tool that simply goes through a JSON file and capitalizes all the key names.  And I don't want it to mix up the fields from the order they were read in.  How would you do that with `std.json`?

**Comment by quickfur** [comment]
@marler8997 Of course not, my first paragraph explicitly conceded that it's a valid use case. :-)

Though to be honest, I feel `std.json` isn't suitable for implementing that sort of JSON editing. If the goal is to edit JSON rather than consume JSON data (involving parsing, building up data structures that map to the JSON, etc.), I would use a drastically different approach, like the JSON analogue of a SAX parser as opposed to a DOM parser, which is the analogue of what's currently being done (build a "DOM" or data tree that represents the JSON data). Parsing the entire JSON input and building data structures, then operating on it afterwards, seems to me to be a rather inefficient approach to implement JSON editing. A SAX-style parser would be much more suitable, since it can just pass over the data once, and memory usage is minimal because it doesn't have to allocate data structures for data that has already been processed and transformed.

**Comment by quickfur** [comment]
P.S. A SAX-style parser would be like the analogue of a range-based lazy algorithm, whereas the current DOM-style parser is like an array.  It could be beneficial to implement a SAX-style parser in `std.json` as an optional alternative, which might even address some of the performance problems that people often complain about.

**Comment by wilzbach** [comment]
> It could be beneficial to implement a SAX-style parser in std.json as an optional alternative, which might even address some of the performance problems that people often complain about.

FYI: I thought `std.json` will be replaced soon by `std.data.json`?
https://github.com/s-ludwig/std_data_json

**Comment by quickfur** [comment]
It would be awesome if it did, because I see that `std.data.json` has a stream-based ("SAX-style") parsing option, which would address the issue in this PR in a nice way that doesn't depend on questionable interpretations of the JSON spec.

**Comment by marler8997** [comment]
The "lazy" vs "aggressive" loading style of the parser is independent of whether or not a tool would want to preserve the object field order.  Whether or not another parser (like a SAX style parser) should support it is a different topic.  I have presented use cases where tools could use this feature with `std.json`.  I'm not saying they couldn't use another parser, but just because another parser could support such a feature doesn't mean that `std.json` shouldn't also support it.  Do you have any reasons why  `std.json` should not support this feature?

**Comment by JackStouffer** [comment]
The std.data.json module was given a bunch of feedback in the NG and it kind-of went nowhere from there.

**Comment by quickfur** [comment]
My main hesitation with this feature is that it doesn't conform to the JSON spec.  It's one thing to make it *possible* for user code to do whatever it wants, which is why I gravitate towards a SAX-style / stream-based parser, because that *enables* user code to implement order-dependence without actually explicitly doing that in Phobos. But it's another thing for Phobos to explicitly support something that, strictly speaking, doesn't follow the JSON spec.  I'm all for *enabling* the user, but I'm not a fan of vendor-specific extensions (which is basically what this PR is) that eventually leads to fragmentation.

And just to be clear, I'm not complaining merely because I like being pedantic about spec conformance; the underlying issue is that we're trying to impose order upon a structure that has been *explicitly* defined to be unordered. This, to me, is a code smell.  In my experience, this style of coding often has or leads to fragility, unhandled corner cases, exceptional behaviours, and lots of little crufty things like that, that are symptomatic of an inherent impedance mismatch between what the code does and what the data is supposed to represent.   I don't feel strongly enough about `std.json` to actively blockade this PR, the others can merge if they think it's a good idea, but I'll just say that I won't actively support this direction.

**Comment by marler8997** [comment]
@quickfur thanks for explaining yourself, I see where you are coming from now.  I'm thinking a better solution might be to provide a mechanism to allow the user to override the type used for `JSONObject` with the default being `JSONValue[string]`.  If you're ok with this I'll go ahead and modify this PR to do that.

**Comment by marler8997** [comment]
Pinging @quickfur and @JackStouffer 

I've refactored the change.  Now it allows you to override the default type for JSON objects, but does not provide any of these types.  This allows an application to customize the parser to their needs without the parser providing any non-spec functionality out of the box.  Note that I've included the `OrderedAA` as a unittest, but not as a publicly accessible structure.  Please have a look and let us know if you are OK with this design.

File: std/json.d [hunk 16]
```diff
@@ -488,17 +522,24 @@ struct JSONValue
      * Hash syntax for json objects.
      * Throws: $(D JSONException) if $(D type) is not $(D JSON_TYPE.OBJECT).
      */
-    ref inout(JSONValue) opIndex(string k) inout pure @safe
+    ref inout(Value) opIndex(string k) inout pure @safe
     {
         auto o = this.objectNoRef;
-        return *enforce!JSONException(k in o,
-                                        "Key not found: " ~ k);
+        static if (hasMember!(Policy, "ObjectType"))
```
timotheecour: DRY:
static if (hasMember!(Policy, "ObjectType"))
  auto ok=o.opIn(k);
else
  auto ok=k in o;
return *enforce!JSONException(ok, "Key not found: " ~ k);

std/json.d
```diff
@@ -96,11 +96,36 @@ enum JSON_TYPE : byte
     FALSE    /// ditto
 }
 
+private struct DefaultPolicy
+{
+    enum sortObjectMembers = true;
+}
```
timotheecour: you can simplify a lot the rest of the code by always requiring a template ObjectType to Policy:
```
private struct DefaultPolicy
 { 
   enum sortObjectMembers = true;
   alias ObjectType(Value)=Value[string];
 }
```

std/json.d
```diff
@@ -1110,6 +1165,91 @@ if (isInputRange!T && !isInfinite!T && isSomeChar!(ElementEncodingType!T))
     assert(json["key1"]["key2"].integer == 1);
 }
 
+version(unittest)
+{
+    struct OrderedAA(Key, Value)
```
timotheecour: I think others (not me...) were concerned about exposing json with ordered_fields, but I didn't see objections to exposing OrderedAA in std.array; otherwise this'll require someone who wants to use json with ordered_fields to reimplement `OrderedAA` below, which is not DRY and much worse than exposing json with ordered_fields. In short: using json with ordered_fields should be as simple as:
```
import std.array:OrderedAA;
static struct OrderedAAPolicy
{
alias ObjectType = OrderedAA;
enum sortObjectMembers = false;
}

auto json1 = parseJSONTemplate!OrderedAAPolicy(...);

```

std/json.d
marler8997: Just discovered that `opIn_r` was a thing...now that I know about it, this code only has the one case now :)

std/json.d
marler8997: Which code can be simplified?

std/json.d
marler8997: Agreed.  However, now that `std.json` does not depend on having `OrderedAA` defined, these can be 2 separate changes.  When I make a separate PR to add `OrderedAA` to `std.array`, we can remove this instance I've included in the unittest and swap it out for that.

Note that by separating these changes into distinct PRs, it makes this one easier to accept on it's own.  Once this one gets accepted, we can start using this feature in `dmd` by simply including our own definition of `OrderedAA` until we can add it to `std.array`.  I'm guessing that adding `OrderedAA` is going to be a fair amount of work with lots of revision and documentation so I would rather not have this change depend on that one.

marler8997:
Per @quickfur's comments, I've created a PR to remove the object field sorting that is currently present in `std.json`: https://github.com/dlang/phobos/pull/6084

wilzbach:
BTW here's what the [ECMA15 spec](http://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-enumerate) says - it requires insertion order to be preserved (it didn't before).

> order of enumerating the properties is not specified but must conform to the rules specified below.

```
Let keys be a new empty List.
For each own property key P of O that is an integer index, in ascending numeric index order
Add P as the last element of keys.
For each own property key P of O that is a String but is not an integer index, in property creation order
Add P as the last element of keys.
For each own property key P of O that is a Symbol, in property creation order
Add P as the last element of keys.
Return keys.
```

Granted, JS and JSON are different things, but people mix them up often enough, s.t. preserving insertion order is an interesting use case.

marler8997:
> BTW here's what the ECMA15 spec says - it requires insertion order to be preserved (it didn't before).

Can't say I'm surprised by this change to ECMA15 :)  Based on @quickfur 's comments I refactored this change to remove the option to preserve order, maybe I should bring it back in?  I've currently been working on modifying this PR to appease @quickfur but after I refactored the change based on his comments he has been silent.  This seems to be a pattern among some of dlang's developers and definitely not a good one.

quickfur:
@marler8997  Please understand that (1) I'm not paid for doing this; (2) I'm doing this entirely out of my free time; (3) my free time is very scant (my weekends are usually completely booked); (4) Phobos has tons of other PRs and things that I can look at; and (5) I'm not under contract to respond to stuff by some deadline. Basically, the only reason I even respond at all is only out of goodwill.  So I don't understand why a pause of 4 days is unacceptable.  If you paid me to work on this full-time, then I would respond within a couple of hours, but IMO the accusatory tone in your comment is uncalled for. If you feel the response rate of Phobos devs is not up to your standard, then perhaps you should consider donating to the D foundation to hire full-time programmers to work on this.

marler8997:
@quickfur 

Sorry for offending you.  I don't mean to say that you NEED to be more responsive in general.  My point was that if you decide to engage in a PR with criticism that you should follow through when the person takes your criticism and spends their time refactoring their work based on your comments.  Of course you're not under any obligation to respond quickly, but I do think it's inconsiderate to criticize someone else's work and then fail to respond in a timely manner after they've gone out of their way to redo all their work based on that criticism.

quickfur:
Well, your last change was 3 days ago, and as I said, my weekend are booked, so if you subtract that, that's really only 1 day.  I don't think that's unreasonable at all, given that there's a whole bunch of other stuff I was reviewing, and my 1 day is really a lot less since I have a full-time job and can't be spending all day working on Phobos -- my employer would fire me.  If I didn't respond for a week, then you may have a point, but if your expectations are that I have to respond instantly, then perhaps I should refrain from reviewing your PRs in the future since I can't keep up with your expectations.

marler8997:
@quickfur I'll know in the future to expect it to take up to a week for you to respond.  Glad we've got that cleared up.  Let's continue the more useful discussion at hand, namely, what to do with this PR.

std/json.d
```diff
@@ -96,11 +96,34 @@ enum JSON_TYPE : byte
...
+A JSONValue template. Features can be enabled/disabled via setting members in $(D Policy).
+Example:
+------------
+struct MyPolicy
+{
+    enum sortObjectMembers = false;
+    alias MyObjectType(K,V) = OrderedAA!(K,V);
+}
+alias MyJSONValue = JSONValueTemplate!MyPolicy;
+------------
```
quickfur: The example here really should be in a ddoc'd unittest.  And it should show how to actually use `JSONValueTemplate`, not just in how to instantiate it, but how to use it with the rest of the module, e.g., to configure it with a different `JSONObject` type, etc..

marler8997: Agreed.  Was going to wait to fully document  until a general "approved" consensus was reached.

marler8997:
I'm going to have to disagree on the size of this change.  Most of the lines the actual change are unit tests/documentation and renaming `JSONValue` to `Value` (internally).  The only functional change is redifining the implementation of JSONValue, toJSON and parseJSON as templates with hooks to override functionality.

Also, saying that this only adds support for "one use case' is just not true.  The change is providing a couple hooks to override to parts of the implementation, along with a framework for providing more hooks in the future.

quickfur:
OK, then let's let @andralex decide whether or not to go ahead with this.

andralex:
On a cursory look I see two issues with this:

* Preserving order of fields is mentioned repeatedly as an important application, but relying on order of fields is specifically listed as a pattern in JSON canon - see e.g. https://stackoverflow.com/questions/3948206/json-order-mixed-up
* std.json is generally disliked. This feature addition does not improve on any of the dislikes. 

It don't see how this can be brought to a solid addition.

PR #6060, opened 2018-01-24
Partially Fix Issue 10828 - Add output range version of toString functions to std.datetime.Date

JackStouffer:
Using the `toString` format from https://github.com/dlang/phobos/pull/5991, update `Date`'s string methods to accept output ranges.

### DIFF
--- a/std/datetime/date.d
+++ b/std/datetime/date.d
@@ -12,6 +12,7 @@ module std.datetime.date;
 import core.time;// : TimeException;
 import std.traits : isSomeString, Unqual;
 import std.typecons : Flag;
+import std.range.primitives : isOutputRange;
 
 version(unittest) import std.exception : assertThrown;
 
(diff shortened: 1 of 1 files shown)


std/datetime/date.d
```diff
@@ -12,6 +12,7 @@ module std.datetime.date;
 import core.time;// : TimeException;
 import std.traits : isSomeString, Unqual;
 import std.typecons : Flag;
+import std.range.primitives;// : isOutputRange;
```
wilzbach: Please use selective imports for new imports - otherwise you risk leaking the entire module.

JackStouffer: I thought it was the opposite?

wilzbach: We can't convert the existing imports to selective imports because that would close the symbol leakage: https://github.com/dlang/phobos/pull/5584

JackStouffer:
These random vibed errors are getting pretty annoying

wilzbach:
Yeah, I am opening a _new_ Jenkins issue almost every second day -  https://github.com/dlang/ci/issues (not mentioning hitting the existing ones), but I don't have a concrete proposal except for dropping Jenkins, creating one common bash script and running that one all repos on a reliable CI (e.g. CircleCi or SemaphoreCI). As we already have the build scripts for Jenkins it probably wouldn't take that long to do...

PR #6063, opened 2018-01-24
Removed unnecessary atomicLoad inside initOnce

marler8997:
I believe the second `atomicLoad` in this double-checked lock is unnecessary.

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -2578,7 +2578,7 @@ auto ref initOnce(alias var)(lazy typeof(var) init, shared Mutex mutex)
     {
         synchronized (mutex)
         {
-            if (!atomicLoad!(MemoryOrder.acq)(flag))
+            if (!atomicLoad!(MemoryOrder.raw)(flag))
             {
                 var = init;
                 atomicStore!(MemoryOrder.rel)(flag, true);


n8sh:
Could you explain why, and why this doesn't also apply to the atomicStore inside that `synchronized` block?

n8sh:
For that matter, why is `atomicLoad` used the first time? https://wiki.dlang.org/Low-Lock_Singleton_Pattern

marler8997:
Because of the mutex, there are 3 operations a thread can perform on the variable:
1. Load Outside the Mutex
2. Load Inside the Mutex
3. Store Inside the Mutex

> Note: the case of storing outside the mutex is forbidden

Let's imagine 2 threads performing any one of these actions simultaneously.  That's 6 Cases to consider without loss of generality.  In each case, both operations need to be atomic if and only if one of the threads is performing a store.  So that leaves us with 3 cases to check:
1. Thread 1 "stores inside mutex" while thread 2 "loads outside mutex"
2. Thread 1 "stores inside mutex" while thread 2 "loads inside mutex"
3. Thread 1 "stores inside mutex" while thread 2 "stores inside mutex"

Because of how a mutex works, cases 2 and 3 are impossible.  That leaves us with one case where the operations MUST be atomic:

Thread 1 "stores inside mutex" while thread 2 "loads outside mutex".  In this case both operations need to be atomic.  Using this logic, you can infer that the 2nd operation (Load Inside Mutex) has no cases where atomicity is required.  The following table summarizes the final result:

|                           | Load                 | Store |
|------------------|------------------------|---------------|
| Outside Mutex |  Must Use Atomic     | Forbidden |
| Inside Mutex    | Atomic Unnecessary | Must Use Atomic |

n8sh:
Thanks for the explanation!

marler8997:
Sorry you're using a couple terms I'm unfamiliar with.  What do you mean by "please explicitly use `raw`"?  And what does "make them implicitly SC at various points" mean?

marler8997:
Ok, changed to `MemoryOrder.raw`.

wilzbach:
(rebased to restart Jenkins)

PR #6068, opened 2018-01-25
Fix Issue 18299: std.datetime.date.cmpTimeUnits does not throw a Date…

JackStouffer:
…TimeException

### DIFF
--- a/std/datetime/date.d
+++ b/std/datetime/date.d
@@ -9986,12 +9986,11 @@ int cmpTimeUnits(string lhs, string rhs) @safe pure
     import std.exception : enforce;
     import std.format : format;
 
-    auto tstrings = timeStrings;
-    immutable indexOfLHS = countUntil(tstrings, lhs);
-    immutable indexOfRHS = countUntil(tstrings, rhs);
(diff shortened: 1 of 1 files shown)


std/datetime/date.d
```diff
@@ -9986,12 +9986,11 @@ int cmpTimeUnits(string lhs, string rhs) @safe pure
     import std.exception : enforce;
     import std.format : format;
 
-    auto tstrings = timeStrings;
-    immutable indexOfLHS = countUntil(tstrings, lhs);
-    immutable indexOfRHS = countUntil(tstrings, rhs);
+    immutable indexOfLHS = countUntil(timeStrings, lhs);
+    immutable indexOfRHS = countUntil(timeStrings, rhs);
```
quickfur: Just curious, what was the reason for this change?

JackStouffer: No reason to make a second copy of the reference when countUntil takes it by value

quickfur: Fair enough.

wilzbach:
Once https://github.com/dlang/phobos/pull/6070 is merged, CircleCi can be restarted here.

PR #6069, opened 2018-01-25
Publicly document all TimeOfDay.roll overloads

JackStouffer:


wilzbach:
Once https://github.com/dlang/phobos/pull/6070 is merged, CircleCi can be restarted here.

PR #6071, opened 2018-01-26
Work On Issue 10828 - Add output range versions of toString functions to std.datetime.TimeOfDay

JackStouffer:
See #5991 and https://github.com/dlang/phobos/pull/6060 for more details

### DIFF
--- a/std/datetime/date.d
+++ b/std/datetime/date.d
@@ -8966,15 +8966,32 @@ public:
 
 
     /++
-        Converts this $(LREF TimeOfDay) to a string with the format HHMMSS.
+        Converts this $(LREF TimeOfDay) to a string with the format `HHMMSS`.
+        If `writer` is set, the resulting string will be written directly to it.
+
+        Params:
+            writer = A `char` accepting $(REF_ALTTEXT output range, isOutputRange, std, range, primitives)
+        Returns:
+            A `string` when not using an output range; `void` otherwise.
       +/
     string toISOString() const @safe pure nothrow
     {
-        import std.format : format;
+        import std.array : appender;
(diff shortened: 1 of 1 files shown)


std/datetime/date.d
```diff
@@ -8966,15 +8966,32 @@ public:
...
+    void toISOString(W)(ref W writer) const
+    if (isOutputRange!(W, char))
+    {
+        import std.format : formattedWrite;
         try
-            return format("%02d%02d%02d", _hour, _minute, _second);
+            formattedWrite(writer, "%02d%02d%02d", _hour, _minute, _second);
         catch (Exception e)
-            assert(0, "format() threw.");
+            assert(0, "formattedWrite() threw.");
```
wilzbach: ### Off-topic

FYI there's a trick to avoid the red coverage here - from https://github.com/dlang/druntime/pull/2057#discussion_r163652061:

```d
scope (failure) assert(0, "formattedWrite() threw.");
formattedWrite(writer, "%02d%02d%02d", _hour, _minute, _second);
```

Also I wonder whether `formattedWrite` can be made `nothrow` when the input string is passed in at CTFE and thus checked at compile-time.

JackStouffer: https://issues.dlang.org/show_bug.cgi?id=18304

wilzbach:
BTW as you seem to be working on adding Writer support to `std.datetime`, there's `core.time` too and IIRC they all only have `string toString`.
For example, `Duration` - https://github.com/dlang/druntime/blob/594e3c66e0c0a38a4e8fa19cbfdf05601d7c0b3c/src/core/time.d#L1529

PR #6077, opened 2018-01-27
Add warning to isConvertibleToString.

jmdavis:
I should have added this warning months ago, but here it is. Hopefully, it won't take all that long to remove uses of `isConvertibleToString` from Phobos so that we can deprecate it, but it is unfortunately used in a number of places, so it will take several PRs to remove them - especially if they sit and rot like my previous PRs for removing uses of `isConvertibleToString` sat and rotted.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -6130,8 +6130,27 @@ deprecated
 }
 
 /**
- * Detect whether $(D T) is a struct, static array, or enum that is implicitly
- * convertible to a string.
+  $(RED Warning: This trait will be deprecated as soon as it is no longer used
+                 in Phobos. For a function parameter to safely accept a type
+                 that implicitly converts to string as a string, the conversion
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -6130,8 +6130,27 @@ deprecated
 }
 
 /**
- * Detect whether $(D T) is a struct, static array, or enum that is implicitly
- * convertible to a string.
+  $(RED Warning: This trait will be deprecated as soon as it is no longer used
+                 in Phobos. For a function parameter to safely accept a type
+                 that implicitly converts to string as a string, the conversion
+                 needs to happen at the callsite; otherwise, the conversion is
+                 done inside the function, and in many (most?) cases, that
```
wilzbach: I would do `s/many (most?)/most/` - the official documentation shouldn't speculate.

PR #6079, opened 2018-01-27
Fix Issue 18316 - std.net.curl.SMTP.mailTo fails to compile

shoo:
https://issues.dlang.org/show_bug.cgi?id=18316


wilzbach:
Thanks!

PR #5234, opened 2017-03-04
std.exception: undocument enforceEx

wilzbach:
Just saw this while reviewing https://github.com/dlang/phobos/pull/5233.
The docs for `enforceEx` explicitly say:

> This is legacy name, it is recommended to use enforce!E instead.

While there has been some resistance against `deprecation`s, a good step to help the reader and avoid confusion would be to undocument the symbol.

PR #6080, opened 2018-01-28
Fix Issue 18319 - std.exception: enforce example does not compile

wilzbach:
I used the opportunity to improve the docs and unify all three overloads under one documentation block.


wilzbach:
Note that [`enforcEx`](https://dlang.org/phobos/std_exception.html#errnoEnforce) has a similar problem, but that's just historical baggage that should be deprecated or at least undocumented (see https://github.com/dlang/phobos/pull/5234).

PR #6081, opened 2018-01-28
Rewrite synopsis example of std.exception

wilzbach:
It didn't compile (see https://github.com/dlang/phobos/pull/6080).

- converted to a module unittest
- fixed the example, s.t. it compiles
- made it to a full synposis by adding more cool functions from `std.exception`

### DIFF
--- a/std/exception.d
+++ b/std/exception.d
@@ -33,35 +33,6 @@ $(TR $(TD Other) $(TD
 ))
 )
 
-    Synopsis of some of std.exception's functions:
-    --------------------
-    string synopsis()
-    {
-        FILE* f = enforce(fopen("some/file"));
-        // f is not null from here on
-        FILE* g = enforce!WriteException(fopen("some/other/file", "w"));
-        // g is not null from here on
-
-        Exception e = collectException(write(g, readln(f)));
-        if (e)
-        {
-            ... an exception occurred...
-            ... We have the exception to play around with...
-        }
-
-        string msg = collectExceptionMsg(write(g, readln(f)));
-        if (msg)
-        {
-            ... an exception occurred...
-            ... We have the message from the exception but not the exception...
-        }
-
-        char[] line;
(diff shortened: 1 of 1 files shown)


std/exception.d
```diff
@@ -71,6 +42,78 @@ $(TR $(TD Other) $(TD
  +/
 module std.exception;
 
+/// Synopis
+unittest
```
JackStouffer: `@system`

std/exception.d
```diff
@@ -71,6 +42,78 @@ $(TR $(TD Other) $(TD
...
+    static class MeaCulpa : Exception
+    {
+        mixin basicExceptionCtors;
+    }
+    e = collectException((){throw new MeaCulpa("diagnostic message");}());
+    assert(e.msg == "diagnostic message");
+    assert(e.file == __FILE__);
+    assert(e.line == __LINE__ - 3);
+
+    // doesPointTo can be used to check whether the source contains pointers / references
```
klickverbot: "The source"? Also, it doesn't just check for pointers/references, but ones pointing into the target. Perhaps this is best removed from the synopsis? The function doesn't really fit into std.exception anyway.

wilzbach: > The function doesn't really fit into std.exception anyway.

Excellent point (I was wondering about the same). Suggestions?
Probably `stdx.allocator` would probably be the best fit...

wilzbach: Anyhow, I removed it for now and opened a Bugzilla issue: https://issues.dlang.org/show_bug.cgi?id=18387

wilzbach:
Rebased to `stable`, s.t. this gets deployed soon to the documentation.
Looks like nothing is blocking this and it has two approvals. Anyone around to do the honor?

(I moved the discussion of the ill-fit `doesPointTo` to https://issues.dlang.org/show_bug.cgi?id=18387)

PR #6083, opened 2018-01-28
Document std.conv.to for string-to-char conversions

Dechcaudron:
As requested in [this thread](https://github.com/dlang/phobos/pull/5959) by @JackStouffer and @schveiguy.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -290,6 +290,29 @@ template to(T)
 }
 
 /**
+   Conversion from string types to char types enforces the input
+   to consist of a single code point, and said code point must
+   fit in the target type. Otherwise, $(LREF ConvException) is thrown.
+ */
+@safe pure unittest
+{
+    import std.exception : assertThrown;
+
+    assert(to!char("a") == 'a');
+    assertThrown(to!char("ñ")); // 'ñ' does not fit into a char
+    assert(to!wchar("ñ") == 'ñ');
+    assertThrown(to!wchar("😃")); // '😃' does not fit into a wchar
+    assert(to!dchar("😃") == '😃');
+
+    // Using wstring or dstring as source type does not affect the result
+    assert(to!char("a"w) == 'a');
+    assert(to!char("a"d) == 'a');
+
+    // Two code points cannot be converted to a single one
+    assertThrown(to!char("ab"));
+}
+
+/**
  * Converting an array _to another array type works by converting each
  * element in turn. Associative arrays can be converted _to associative
  * arrays as long as keys and values can in turn be converted.


std/conv.d
```diff
@@ -289,6 +289,31 @@ template to(T)
     assert(to!int(to!float(-a)) == -a);
 }
 
+/**
+ * Conversion from string types to char types enforces the input
+ * to have a single code point, and then it converts said code point
+ * to its representation in a single code unit of the target type.
+ * If the code point cannot be represented in a single code unit of
+ * the target type, ConvException is thrown.
+ */
```
wilzbach: FYI: we recently added to the [DStyle](https://dlang.org/dstyle.html) that new code [shouldn't not have leading stars](https://dlang.org/dstyle.html#phobos_documentation) - though you are the first one to submit a PR with new documentation since this has been added.

std/conv.d
```diff
@@ -289,6 +289,31 @@ template to(T)
     assert(to!int(to!float(-a)) == -a);
 }
 
+/**
+ * Conversion from string types to char types enforces the input
+ * to have a single code point, and then it converts said code point
+ * to its representation in a single code unit of the target type.
+ * If the code point cannot be represented in a single code unit of
+ * the target type, ConvException is thrown.
```
wilzbach: `$(LREF ConvException)`

std/conv.d
```diff
@@ -289,6 +289,31 @@ template to(T)
     assert(to!int(to!float(-a)) == -a);
 }
 
+/**
+ * Conversion from string types to char types enforces the input
+ * to have a single code point, and then it converts said code point
+ * to its representation in a single code unit of the target type.
```
wilzbach: That sounds needlessly complicated. I would simplify it to sth. like:

>  Conversion from string types to char types requires the input to consist of a single code point of the target type only. Otherwise a $(LREF ConvException) is thrown).

std/conv.d
```diff
@@ -289,6 +289,31 @@ template to(T)
...
+ * If the code point cannot be represented in a single code unit of
```
wilzbach: s/the/a/

std/conv.d
schveiguy: No this is inaccurate. It must fit in a single code unit, not a code point.

It could be shortened. Just say “... code point, and that said code point fits into the target type”

std/conv.d
Dechcaudron: I'm sorry, what?

std/conv.d
Dechcaudron: Noted and modified, thanks.

Dechcaudron:
Suggestions implemented 😄

std/conv.d
```diff
@@ -289,6 +289,29 @@ template to(T)
     assert(to!int(to!float(-a)) == -a);
 }
 
+/**
+   Conversion from string types to char types enforces the input
+   to consist of a single code point, and said code point must
```
JackStouffer: trailing whitespace here.

wilzbach:
Removed the trailing whitespace and squashed to one commit.

schveiguy:
@Dechcaudron thanks for the extra effort!

Dechcaudron:
Hey, my pleasure. I committed to do it and so I did. Thanks to all of you for making Phobos possible.

PR #6085, opened 2018-01-28
Issue 18314 - std.traits.getSymbolsByUDA only considers the first sym…

Biotronic:
…bol of an overload set

This change expands overload sets such that each overload is considered separately.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -7824,36 +7824,56 @@ private template isDesiredUDA(alias attribute)
  */
 template getSymbolsByUDA(alias symbol, alias attribute)
 {
-    import std.format : format;
-    import std.meta : AliasSeq, Filter;
-
-    // translate a list of strings into symbols. mixing in the entire alias
-    // avoids trying to access the symbol, which could cause a privacy violation
-    template toSymbols(names...)
-    {
-        static if (names.length == 0)
-            alias toSymbols = AliasSeq!();
-        else
-            mixin("alias toSymbols = AliasSeq!(symbol.%s, toSymbols!(names[1..$]));"
-                  .format(names[0]));
-    }
-
-    // filtering inaccessible members
-    enum isAccessibleMember(string name) = __traits(compiles, __traits(getMember, symbol, name));
-    alias accessibleMembers = Filter!(isAccessibleMember, __traits(allMembers, symbol));
-
-    // filtering not compiled members such as alias of basic types
-    enum hasSpecificUDA(string name) = mixin("hasUDA!(symbol." ~ name ~ ", attribute)");
-    enum isCorrectMember(string name) = __traits(compiles, hasSpecificUDA!(name));
-
-    alias correctMembers = Filter!(isCorrectMember, accessibleMembers);
-    alias membersWithUDA = toSymbols!(Filter!(hasSpecificUDA, correctMembers));
+    alias membersWithUDA = getSymbolsByUDAImpl!(symbol, attribute, __traits(allMembers, symbol));
 
     // if the symbol itself has the UDA, tack it on to the front of the list
     static if (hasUDA!(symbol, attribute))
         alias getSymbolsByUDA = AliasSeq!(symbol, membersWithUDA);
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -7856,6 +7833,49 @@ template getSymbolsByUDA(alias symbol, alias attribute)
         alias getSymbolsByUDA = membersWithUDA;
 }
 
+template getSymbolsByUDAImpl(alias symbol, alias attribute, names...)
```
wilzbach: `private`?

wilzbach:
(The Jenkins + CircleCI failures are unrelated)

Biotronic:
The thing I dislike about my implementation is `getSymbolsByUDAImpl` is recursive, and explicitly and verbosely does what the previous version did with `Filter`. Sadly, [5710](https://issues.dlang.org/show_bug.cgi?id=5710) is still around, and I'm becoming more and tired of it.

Dechcaudron:
Thanks for the PR @Biotronic. I was planning on fixing it myself, but you were way too fast 😄

std/traits.d
```diff
@@ -7921,6 +7941,26 @@ template getSymbolsByUDA(alias symbol, alias attribute)
...
+    {
+        @lal
+        int n;
+        @lal
+        void foo();
+        @lol
+        void foo(int a);
+    }
+
+    assert(getSymbolsByUDA!(A, lal).length == 2);
```
BBasile: `static assert(...` twice.

std/traits.d
```diff
@@ -7921,6 +7941,26 @@ template getSymbolsByUDA(alias symbol, alias attribute)
...
 
+// Issue 18314
+@safe unittest
+{
+    enum lol;
+    enum lal;
+
+    struct A
+    {
+        @lal
```
BBasile: insert this declaration **first**

```d
@lal
void foo()(string){}
```

update the assertions accordingly, and have fun.

std/traits.d
Dechcaudron: Right. This should be done at CT.

std/traits.d
Dechcaudron: For the sake of style, could we use `@Attr1` and `@Attr2` instead of `@lol` and `@lal`? I know I originally used those in my issue report, but it definitely was not my intention that such funny names ended up in the source code.

Biotronic: This bug has already been filed in BugZilla. Looks to be by you: 
https://issues.dlang.org/show_bug.cgi?id=16206

It's not the mandate of this PR to fix bugs in DMD.

BBasile: Indeed. I planned to put the link to the issue after a possible "wtf" reaction from your part. But things didn't happen as expected.

Biotronic: WTF?!
Here. Feeling better now?

wilzbach: > I originally used those in my issue report, but it definitely was not my intention that such funny names ended up in the source code.

If you already come with the style, it should be `attr1` and `attr2` - https://dlang.org/dstyle.html#naming_udas

But meh, I don't have a preference between these two. It's quite typical for contributors to put the code from the issue 1:1 into Phobos. Also it doesn't end up in the user documentation.
And if you disagree, this can always be improved. It shouldn't block this PR.

wilzbach: > Indeed. I planned to put the link to the issue after a possible "wtf" reaction from your part. But things didn't happen as expected.

Please refrain from wasting valuable time of other contributors in the future.
A simple comment:

> I just bumped into a similar issue. See XYZ

would have saved all parties time...

Biotronic: Sure, but we'd have less fun that way.

Biotronic:
And apparently I broke the auto-merge when I addressed the stuff that was mentioned above.

wilzbach:
> And apparently I broke the auto-merge when I addressed the stuff that was mentioned above.

Yes, it's not a bug it's a security feature.

image

PR #3395, opened 2015-06-08
Add new algorithm: either()

nordlow:
Thanks to D's lazy parameters this function simplify pattern matching:

`either()` similar to Python/Lisp's `or`. Same as D's `any` but returns first parameter value which implicitly convert to true, or last argument otherwise. I guess `any` could be modified but that would cause breakage on its return type.

I've found real use of `either()`in various parsing contexts.

I'm very open to changing the namings if you feel they are misguiding.

Note that `either` currently cannot be `nothrow` because of issue    https://issues.dlang.org/show_bug.cgi?id=12647

PR #6086, opened 2018-01-28
Turn enforce into an eponymous template + undocument enforceEx

wilzbach:
Follow-up to https://github.com/dlang/phobos/pull/5234 and https://github.com/dlang/phobos/pull/6080.

tl;dr summary

> However, if enforce can't fully replace enforceEx, then it needs to stick around for at least one more release so that we can have a release where enforce can fully replace enforceEx. That way, someone can build their code with both the latest release and master without getting a bunch of deprecation messages. And clearly, from how enforceEx is being used in Phobos, enforce needs to be improved first. So, please create a PR to improve enforce, and then we can deprecate enforceEx after the improved enforce has been put out in a release.

I already added the warning in the documentation and removed `enforceEx` from the bookmark summary at the top.

### DIFF
--- a/std/exception.d
+++ b/std/exception.d
@@ -17,7 +17,6 @@ $(TR $(TD Assumptions) $(TD
 $(TR $(TD Enforce) $(TD
         $(LREF doesPointTo)
         $(LREF enforce)
-        $(LREF enforceEx)
         $(LREF errnoEnforce)
 ))
 $(TR $(TD Handlers) $(TD
@@ -386,12 +385,17 @@ void assertThrown(T : Throwable = Exception, E)
         If a delegate is passed, the safety and purity of this function are inferred
         from `Dg`'s safety and purity.
  +/
-T enforce(E : Throwable = Exception, T)(T value, lazy const(char)[] msg = null,
-string file = __FILE__, size_t line = __LINE__)
-if (is(typeof({ if (!value) {} })))
+template enforce(E : Throwable = Exception)
+if (is(typeof(new E("", __FILE__, __LINE__)) : Throwable) || is(typeof(new E(__FILE__, __LINE__)) : Throwable))
 {
-    if (!value) bailOut!E(file, line, msg);
-    return value;
+
+    T enforce(T)(T value, lazy const(char)[] msg = null,
+    string file = __FILE__, size_t line = __LINE__)
+    if (is(typeof({ if (!value) {} })))
+    {
+        if (!value) bailOut!E(file, line, msg);
+        return value;
+    }
 }
 
 /// ditto
@@ -453,6 +457,15 @@ unittest
         assert(e.file == __FILE__);
         assert(e.line == __LINE__-7);
     }
+}
+
+/// Alias your own enforce function
+@safe unittest
+{
+    import std.conv : ConvException;
+    alias convEnforce = enforce!ConvException;
+    assertNotThrown(convEnforce(true));
+    assertThrown!ConvException(convEnforce(false, "blah"));
 }
 
 private void bailOut(E : Throwable = Exception)(string file, size_t line, in char[] msg)
@@ -585,7 +598,7 @@ T errnoEnforce(T, string file = __FILE__, size_t line = __LINE__)
 }
 
 
-/++
+/+
     If $(D !value) is $(D false), $(D value) is returned. Otherwise,
     $(D new E(msg, file, line)) is thrown. Or if $(D E) doesn't take a message
     and can be constructed with $(D new E(file, line)), then
@@ -611,7 +624,7 @@ if (is(typeof(new E("", __FILE__, __LINE__))))
     }
 }
 
-/++ Ditto +/
+/+ Ditto +/
 template enforceEx(E : Throwable)
 if (is(typeof(new E(__FILE__, __LINE__))) && !is(typeof(new E("", __FILE__, __LINE__))))
 {
--- a/std/exception.d
+++ b/std/exception.d
@@ -597,8 +597,10 @@ T errnoEnforce(T, string file = __FILE__, size_t line = __LINE__)
     return value;
 }
 
+// @@@DEPRECATED_2.084@@@
+/++
+    $(RED Deprecated. Please use $(LREF enforce) instead. This function will be removed 2.084.)
 
-/+
     If $(D !value) is $(D false), $(D value) is returned. Otherwise,
     $(D new E(msg, file, line)) is thrown. Or if $(D E) doesn't take a message
     and can be constructed with $(D new E(file, line)), then
@@ -613,6 +615,7 @@ T errnoEnforce(T, string file = __FILE__, size_t line = __LINE__)
     enforceEx!DataCorruptionException(line.length);
     --------------------
  +/
+//deprecated("Please use enforce instead")
 template enforceEx(E : Throwable)
 if (is(typeof(new E("", __FILE__, __LINE__))))
 {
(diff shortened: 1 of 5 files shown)


std/json.d
```diff
@@ -101,7 +101,7 @@ JSON value node
 */
 struct JSONValue
 {
-    import std.exception : enforceEx, enforce;
+    import std.exception : enforce, enforce;
```
CyberShadow: Just one should be enough.

CyberShadow:
Existing usage: https://github.com/search?q=enforceEx+language%3Ad&type=Code&utf8=%E2%9C%93

Related older change: https://github.com/dlang/phobos/pull/2462

std/exception.d
```diff
@@ -455,6 +459,15 @@ unittest
     }
 }
 
+/// Alias your own enforce function
+@safe unittest
+{
+    import std.conv : ConvException;
+    alias enf = enforce!ConvException;
+    assertNotThrown(enf(true));
+    assertThrown(enf(false, "blah"));
```
CyberShadow: How about `assertThrown!ConvException` ?

std/exception.d
```diff
@@ -455,6 +459,15 @@ unittest
     }
 }
 
+/// Alias your own enforce function
+@safe unittest
+{
+    import std.conv : ConvException;
+    alias enf = enforce!ConvException;
```
CyberShadow: `alias convEnforce = enforce!ConvException;`

std/exception.d
```diff
@@ -584,8 +597,11 @@ T errnoEnforce(T, string file = __FILE__, size_t line = __LINE__)
     return value;
 }
 
-
+// @@@DEPRECATED_2019-07@@@
 /++
+    $(RED Deprecated. Please use $(LREF enforce) instead. This function will be removed
+               in July 2019.)
```
CyberShadow: Hmm, are we still doing deprecations by dates? I thought there was consensus that deprecating by D versions is overall better.

wilzbach:
> Existing usage: https://github.com/search?q=enforceEx+language%3Ad&type=Code&utf8=%E2%9C%93

Thanks! Though this PR doesn't depend on a decision on whether is worthwhile to deprecate this as it's just the preparation to make `enforce` equal to `enforceEx`.
Also we have multiple options here that improve the status quo without any cost of breakage:
- just mark `enforceEx` as deprecated in the docs
- just remove `enforceEx` from the docs
- trigger a deprecation warning, but never follow through with the removal (preferred)

std/exception.d
wilzbach: Ehm our deprecation process definitely needs some guidelines because I wasn't aware of this.
I would prefer D versions too.

CyberShadow:
This case doesn't seem to different from the others in which we've deprecated then removed Phobos symbols, so I see no reason to not stick to the current policy. It's also a super-easy fix, that's going to be backwards-compatible through a few releases.

std/exception.d
```diff
@@ -386,12 +385,17 @@ void assertThrown(T : Throwable = Exception, E)
...
-if (is(typeof({ if (!value) {} })))
+template enforce(E : Throwable = Exception)
+if (is(typeof(new E("", __FILE__, __LINE__)) : Throwable) || is(typeof(new E(__FILE__, __LINE__)) : Throwable))
 {
-    if (!value) bailOut!E(file, line, msg);
-    return value;
+
+    T enforce(T)(T value, lazy const(char)[] msg = null,
+    string file = __FILE__, size_t line = __LINE__)
+    if (is(typeof({ if (!value) {} })))
```
andralex: We have `ifTestable`, though I have to say this is nicer.

std/exception.d
andralex: By version would be nice. @MartinNowak what do you think?

jmdavis: I've been using month/year, because historically, our releases haven't been consistent enough to use any kind of release numbers when deprecating things. I can easily pick a month a year from now and put that in the message. Historically, that hasn't worked at all with versions. Martin has improved that situation considerably, but they're still not completely consistent. And as a user, it's far easier to look at a date and see about when the symbol is going away than it is to try and figure out when on earth a specific version is going to be released. If the date is a year from now, then it's clear that I have a year. If it's a version number, then it's not at all clear how long I have. I'd have to examine what the release schedule has been and estimate about when the symbol is going away.

jmdavis: However, the date in this case should be more like 2019-02, since it's supposed to be about a year from now (after which, it will be undocumented and have a non-ddoc comment indicating a date about a year from then for when it will be fully removed).

CyberShadow: > And as a user, it's far easier to look at a date and see about when the symbol is going away than it is to try and figure out when on earth a specific version is going to be released. 

Err, sorry, I think that's a bit removed from reality. Essentially that's only true (and useful) if you're always using the latest version of DMD, and use nothing but the latest version of DMD. While this is mostly true for us who contribute to DMD and core D components frequently, it likely isn't for most D users.

Using dates is less useful for answering these questions:
- Upgrading to which version of DMD will break my code?
- Downgrading to which version of DMD will get my code to build again?
- In which version of LDC or GDC will this code stop working?
- Will upgrading to the next release of my Linux distribution mean I'll have fix my code?

Essentially, dates become a lot less useful in any situation where the user is not using the latest version of Phobos, whether that's thanks to a rolling release Linux distro, using D-apt, or religious frequent manual updates. In all other cases, I believe version numbers would be more useful and easier to look up.

jmdavis: Well, technically, it keeps working for about a year after it's removed from the docs, so I'd be inclined to argue that the imprecision with regards to how it matches up with ldc releases or whatnot doesn't matter much, and the site always has the latest docs only, but if we get to the point that we have the docs for previous releases like we've discussed before, then dates definitely become worse.

So, fine, if you guys really want version numbers, then we go with version numbers, but it's going to make the actual time frame of deprecations more inconsistent, since it'll be harder to target a year ahead, but while they've usually been in the ballpark of where they should be, they don't always get moved along quite when they should anyway.

In that case, let's change it to say

    // @@@DEPRECATED_2.084@@@

and

    This function will be removed in 2.084.

and I'll have to create some PRs to fix the existing deprecations.

CyberShadow: > since it'll be harder to target a year ahead,

I don't think we need to try to target deprecations using calendar time, just use "N major versions ahead" instead of "N months ahead".

> but while they've usually been in the ballpark of where they should be, they don't always get moved along quite when they should anyway.

Yeah, one advantage of using dates is that when we don't honor our own deprecation deadlines, it's easier to excuse that for dates than version numbers.

std/exception.d
wilzbach: Yes, `ifTestable` is one of those one-liners that cost a lot more than they are worth it:

https://github.com/dlang/phobos/blob/3f30d5c02bae9c5b2af3d3e5e2602f0f5403a547/std/traits.d#L7996

Interestingly it crept into Phobos with other one-liners: https://github.com/dlang/phobos/pull/3395

std/exception.d
CyberShadow: > And as a user, it's far easier to look at a date and see about when the symbol is going away than it is to try and figure out when on earth a specific version is going to be released.

Thinking more on this, I think I understand better what you meant here.

If you are maintaining an open-source project (especially a library), a deprecation date (on a symbol used in your code) will tell you what's the latest point in time that the deprecation is guaranteed to still work for everyone, everywhere, so you don't need to really worry about it until then. Anything past that date, though, and there appears the chance that someone, somewhere is going to start having trouble building your software.

So... I'd say version numbers are more useful if you're writing code for yourself or your company, and deprecation dates are more useful if you're writing open-source software.

wilzbach: > In that case, let's change it to say

OK. Changed. We should probably add a short documentation in the wiki or as a markdown document this repo about the deprecation process.

wilzbach:
(added a changelog entry)

Anything else needed for this PR? It uses a version for the deprecation now and once @JackStouffer's deprecation DIP has been approved we will have to update all the existing deprecations anyways.

It would be great to get this in before 2.079, s.t. `enforce` is fully capable of replacing `enforceEx` and we can start the deprecation cycle with `2.080`.

PR #6089, opened 2018-01-29
Fix issue 15949: Make readText handle mismatched BOMs.

jmdavis:
This adds a BOMException to std.encoding and makes readText check the
BOM. If there is one, and it doesn't match the requested type, then it
throws a BOMException. If the BOM matches and UTF validation fails or if
there is no BOM, and the size of the file mismatches with the requested
type (e.g. it's not divible by 2, so it can't be valid UTF-16), then a
UTFException is thrown.

readBOM was also added to read a BOM from a file. readText uses it, but
it's public so that anyone can use it.

I'm also looking into adding a function (probably called `readFileAsRange`) which returns a range of the requested character type and converts the file to that encoding regardless of whether it was UTF-8, UTF-16 BE, UTF-16 LE, UTF-32 BE, or UTF-32 LE, which would make it possible not to care about the file's encoding so long as it's at least UTF-8, UTF-16, or UTF-32. It would also make it possible to use the replacement character instead of throwing a `UTFException` when invalid UTF is encountered. But that function is clearly going to be a bit more involved to implement than this, so I'm doing it separately. These changes at least make it possible to cleanly deal with UTF-8, UTF-16 of the native endianness, and UTF-32 of the native endianness, whereas before they just resulted in a `UTFException` or possibly an `Error` (if the attempted cast to the requested string type would result in a misaligned array).

### DIFF
--- a/changelog/std-file-readText.dd
+++ b/changelog/std-file-readText.dd
@@ -1,0 +1,15 @@
+$(REF readBOM, std, file) added and $(REF readText, std, file) now reports mismatched BOMs.
+
+$(REF readBOM, std, file) attempts to read the
+$(HTTP https://en.wikipedia.org/wiki/Byte_order_mark, BOM) from the front of the
+file and returns the corresponding $(REF BOMSeq, std, encoding).
+
+$(REF readText, std, file) now checks the BOM (if present) and throws a
+$(REF BOMException, std, encoding) if the the BOM is for an encoding that does
+not match the requested return type. A $(REF UTFException, std, utf) will still
+be thrown if the file contains invalid UTF, but now, if there isn't a BOM, it
+also checks whether the length of the file indicates that it can't have the
+requested encoding, in which case an $(REF UTFException, std, utf) is thrown.
+Previously, in that case, when $(REF readText, std, file) attempt to cast to
+the requested type, an Error was thrown because the resulting array was
+misaligned.
--- a/std/file.d
+++ b/std/file.d
@@ -32,6 +32,7 @@ $(TR $(TD Files) $(TD
           $(LREF copy)
           $(LREF read)
           $(LREF readText)
+          $(LREF readBOM)
           $(LREF remove)
           $(LREF slurp)
           $(LREF write)
@@ -83,6 +84,7 @@ import core.time : abs, dur, hnsecs, seconds;
 
 import std.datetime.date : DateTime;
 import std.datetime.systime : Clock, SysTime, unixTimeToStdTime;
+import std.encoding : BOMSeq;
 import std.internal.cstring;
 import std.meta;
 import std.range.primitives;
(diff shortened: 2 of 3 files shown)


std/file.d
```diff
@@ -465,30 +467,98 @@ version (linux) @safe unittest
...
+    import std.exception : enforce;
+    import std.format : format;
+    import std.utf : UTFException, validate;
+
+    static if(is(StringTypeOf!R))
+        StringTypeOf!R filename = name;
+    else
+        auto filename = name;
+
+    immutable bomSeq = readBOM(filename);
```
CyberShadow: Won't opening the file twice break reading from streams? Better read the file first, then use `getBOM`. `readBOM` seems unnecessary.

std/file.d
```diff
@@ -465,30 +467,98 @@ version (linux) @safe unittest
...
+    immutable bom = bomSeq.schema;
+
+    static if(is(Unqual!(ElementEncodingType!S) == char))
+    {
+        with(BOM) switch(bom)
+        {
+            case none:
+            case utf8: break;
+            default: throw new BOMException(format!"UTF-8 requested, but BOM is %s"(bom), bomSeq);
```
CyberShadow: This doesn't look right; e.g. it will do the wrong thing if the file starts with the UTF-7 BOM (`+/v8`). This is still valid UTF-8. Instead we probably want to detect cases where the file is not valid in the specified encoding, and throw a better error message than "invalid UTF". IIRC the UTF8/16/32 BOMs are designed to be invalid in all of those encodings except their own, so we can take advantage of that.

std/file.d
```diff
@@ -465,30 +467,98 @@ version (linux) @safe unittest
...
+        }
+    }
+
+    auto data = read(filename);
+    if (data.length % ElementEncodingType!S.sizeof != 0)
+        throw new UTFException(format!"The content of %s is not UTF-%s" (filename, ElementEncodingType!S.sizeof * 8));
+
     static auto trustedCast(void[] buf) @trusted { return cast(S) buf; }
-    auto result = trustedCast(read(name));
+    auto result = trustedCast(data[bomSeq.sequence.length .. $]);
```
CyberShadow: Stripping the BOM may be a breaking change in some situations, e.g. if the code already expects the BOM and will do `s = fn.readText[3..$];`, so I'm not sure about this change.

changelog/std-file-readText.dd
```diff
@@ -0,0 +1,15 @@
...
+
+$(REF readText, std, file) now checks the BOM (if present) and throws a
+$(REF BOMException, std, encoding) if the the BOM is for an encoding that does
+not match the requested return type. A $(REF UTFException, std, utf) will still
+be thrown if the file contains invalid UTF, but now, if there isn't a BOM, it
+also checks whether the length of the file indicates that it can't have the
+requested encoding, in which case an $(REF UTFException, std, utf) is thrown.
+Previously, in that case, when $(REF readText, std, file) attempt to cast to
+the requested type, an Error was thrown because the resulting array was
+misaligned.
```
CyberShadow: I have to say that this changelog entry was confusing to me, and wasn't enough to understand the change until I read the source diff.

Some questions that might be better to answer more directly:
- What kind of files, if any, that were previously valid, are now considered invalid?
- What kind of files, if any, that were previously invalid, are now considered valid?
- Of the files that are now valid, how will the in-memory representation (returned array) differ from the file on disk? (The answer being that a BOM is stripped if present, in the current implementation)
- Do I need to add more `catch` clauses to my exception handling after this change?

std/file.d
jmdavis: I forgot about streams. I did that with the idea that it's quite inefficient to read in the entire file just to then throw an exception due to the fact that the actual encoding doesn't match the requested encoding. But I guess that if someone is really worried about that, they can call `readBOM` themselves.

std/file.d
jmdavis: Are you sure that UTF-7 is valid UTF-8? I didn't think that it was, and `char` is supposed to be UTF-8 by definition.

std/file.d
jmdavis: It's my understanding that it's never actually guaranteed that there will be a BOM (though I think that it's usually there for UTF-16). So, as long as existing code is checking that there is a BOM, and removing it if present, then it will be fine. It's only code that simply assumes that a BOM is present that would be in trouble, and that code would be wrong anyway. And it's far more user friendly to strip it off. Otherwise, technically, everyone needs to always be calling `getBOM` to see if there's a BOM and then strip it off, otherwise they have junk at the beginning of the text some percentage of the time. The main reason that that doesn't happen more is that UTF-8 doesn't usually have a BOM.

So, given that anyone handling the BOM now would need to be checking for it and then stripping it if present, the fact that it's removed shouldn't cause problems. It just means that the existing checks would become unnecessary.

changelog/std-file-readText.dd
jmdavis: Clearly, I'll have to rework it then.

std/file.d
CyberShadow: I think we shouldn't add `readBOM` 1) because we're not going to use it ourselves anyway 2) because it doesn't look like good practice due to being susceptible to race conditions.

std/file.d
CyberShadow: > Are you sure that UTF-7 is valid UTF-8? 

Yes, I'm pretty sure; but that's beside the point. Just the UTF-7 BOM is valid UTF-8, and that's a problem by itself.

> I didn't think that it was, and char is supposed to be UTF-8 by definition.

I don't see how that's related. The problem is that, after this change, you can no longer use `readText` to read text files (which are otherwise valid UTF-8) that start with `+/v8`, which creates an opportunity for a denial of service.

It's not just UTF-7 - look at the `BOM` enum, there's lots of other stuff in there that we don't care about (like EBCDIC).

std/file.d
CyberShadow: > It's my understanding that it's never actually guaranteed that there will be a BOM (though I think that it's usually there for UTF-16). 

Err, that completely depends on the use case.

> It's only code that simply assumes that a BOM is present that would be in trouble, and that code would be wrong anyway. 

No, that lies on the assumption that `readText` will only ever be used to read arbitrary files with no foreknowledge of what they may contain. In contrast, consider a program that creates and then reads its own text files: it will put a BOM at the beginning when writing the file, so it will want to remove it when reading it.

> And it's far more user friendly to strip it off.

Well, I don't disagree. I was actually a little surprised the initial `readText` implementation didn't do this, years ago.

On the other hand, that would break the following contract, which might be surprising:

- `fileName.readText.toFile(fileName)` should not change the contents of `fileName`.

Hypothetically, there might be a program that needs to process a large number of text files, do some processing that would affect 1% of them, and then commit them to git or something; a change in `readText` as done here would mean that 100% files will change instead of that 1%.

> Otherwise, technically, everyone needs to always be calling getBOM to see if there's a BOM and then strip it off, otherwise they have junk at the beginning of the text some percentage of the time.

I think a `stripBOM` helper would be much better than `getBOM`, i.e.:

```d
string fileContent = fileName.readText.stripBOM;
```

> The main reason that that doesn't happen more is that UTF-8 doesn't usually have a BOM.

Yeah, I noticed the `dos2unix` utility will remove the BOM on the files it processes by default. HTML and PHP also both don't recommend using a BOM.

Considering the BOM is not used a lot, maybe it's better to err on the side of caution than a breaking change without a significant benefit.

std/file.d
jmdavis: My point is that because `readText` returns a string, and by definition, strings in D are supposed to be UTF-8, UTF-16, or UTF-32 depending on the character type, any other encoding is invalid and should be processed using `read` and turned into `ubyte`s are `uint`s or whatever from there. As it stands, such encodings are likely to result in a `UTFException` being thrown, but if the BOM clearly indicates that the file is not UTF-8, UTF-16, or UTF-32, then there should be no need to call `validate`.

std/file.d
jmdavis: Hmm. The main advantage to `getBOM` would be that you can determine the encoding ahead of time. Then if it's not UTF-8, UTF-16, or UTF-32, you can do whatever you need to do with `read` to process it, and if it's UTF-8, UTF-16, or UTF-32, you can call the `readText` with the correct string type. Having `stripBOM` would allow us to avoid having `readText` change the file contents for those cases where it matters, though that's less user-friendly for the common case. Stripping it mostly just matters for UTF-16 though, since UTF-8 almost never has a BOM, and UTF-32 isn't used much. So, I guess that it would be better to have `stripBOM` than to have `readText` strip it off, though that is a bit annoying for any use case I've had. I think that `readBOM` still has value though in and of itself.

std/file.d
CyberShadow: Sorry, I think we may be talking past each other.

Here is a D program:

```d
import std.file;

void main()
{
	foreach (string s; ["+/v6", "+/v7", "+/v8"])
	{
		write("test.txt", s);
		readText("test.txt");
	}
}
```

The program is not doing anything invalid. The strings it reads and writes are all perfectly valid UTF-8 (just plain ol' 7-bit ASCII, in fact, which is a subset of UTF-8).

Before this pull request, the program works fine.

After this pull request, the program stops working with:
```
std.encoding.BOMException@.../std/file.d(518): UTF-8 requested, but BOM is utf7
```

This is not a good thing :)

std/file.d
CyberShadow: > I think that readBOM still has value though in and of itself.

Not saying it's not useful, but I think we should avoid encouraging patterns in the standard library that are susceptible to filesystem race conditions (i.e. TOCTOU).

jmdavis: I would not expect there to be much risk of a file suddenly being rewritten with a different encoding, and how on earth else are you going to deal with this? Call `readText` and have it fail with the requested encoding after reading in the entire file and then read in the entire file again with its proper encoding? You have the same race condition in that case, but you ended up reading the entire file twice.

std/file.d
jmdavis: Hmm. Well, I guess that we can not throw on UTF-7 then and just let it try and validate it as UTF-8, but really, `readText` is intended for reading UTF-8, UTF-16, and UTF-32 only given that it returns a string.

std/file.d
CyberShadow: I think that if you want your program to read UTF-8, UTF-16 and UTF-32 and then process each of those as `string`, `wstring` and `dstring` respectively, you should read the file as an `ubyte` array first, look at the BOM, and then cast the `ubyte[]` to the corresponding string type.

I don't know if there's much demand for it, but the above sounds like something that ought to be handled by the standard library, not user code...

jmdavis: Unless we're going to just convert the entire thing to the desired string type regardless of the original encoding, I don't see how a library function could wrap that, because the return type would differ - though unless you've templatized all of your string processing, you're going to have to convert it to a specific string type anyway, or your code won't be able to handle it. I don't think that there's truly a clean way to handle it though. :|

I have started writing a function which is supposed to return a range rather than a string, and the range will take whatever the original encoding was (at least, out of UTF-8, UTF-16, and UTF-32) and convert it to the requested encoding, but it's a bit tricky, especially if the file doesn't have a BOM. It's too bad that we can't just assume UTF-8 for everything...

CyberShadow: > Unless we're going to just convert the entire thing to the desired string type regardless of the original encoding, I don't see how a library function could wrap that, because the return type would differ - though unless you've templatized all of your string processing, you're going to have to convert it to a specific string type anyway, or your code won't be able to handle it. I don't think that there's truly a clean way to handle it though. :|

Err, yes, I meant eagerly convert it to the return type, which would probably still be a negligible cost for most use cases. But we can keep the type too, by allowing the user to pass a lambda, i.e.: `fileName.readText!(s => s.retro.toFile("reversed.txt"));`

jmdavis:
Okay. I created a different PR which doesn't introduce any new symbols: https://github.com/dlang/phobos/pull/6113

PR #6090, opened 2018-01-29
Group documentation in std.algorithm

wilzbach:
From https://forum.dlang.org/post/spytixhkojgrsbdlgtmy@forum.dlang.org

> Note to self: learn to scroll down in the docs to find other definitions. I was convinced that remove only acted on indices :p

I also took the opportunity to improve the docs in std.algorithm.searching slightly.

Documentation of std.algorithm
------------------------------

BTW in general while I think std.algorithm is one of the best modules in Phobos,
I don't like it's churned with so many redundant symbols - a quick list:

- min, minElement, minIndex, minPos
- find, findAdjacent, findAmong, findSkip, findSplit, findSplitBefore, findSplitAfter
- levenshteinDistance, levenshteinDistanceAndPath
- move, moveAll, moveEmplaceAll, moveSome, moveEmplaceSome
- largestPartialIntersection, largestPartialIntersectionWeighted
- isSorted, isStrictlyMonotonic
- ordered, strictlyOrdered
- nextPermutation, nextEvenPermutation
- sort, schwartzSort (`sort` already accepts a `SwapStrategy`), `multiSort`, `completeSort`
- topN, topNCopy, topNIndex
- partition, partition3, pivotPartition
- fill, uninitializedFill
- strip, stripLeft, stripRight
- cache, cacheBidirectional
- filter, filterBidirectional

It took me a long time to learn to navigate in this chaos and I observed newcomers
struggling at this too. It also takes a toll from experienced developers.

Of course, some methods in this list are actually useful, but here's what I suggest
to do on case by case evaluation:

If there's a simple unification e.g. `largestPartialIntersection!Weightes.yes`,
add it to Phobos and undocument the legacy baggage and make it an `alias`.
That way we risk zero breakage and can do a major cleanup of std.algorithm
and people will benefit immediately from it as , after all, the rendered docs
are important - not the exposed symbols. People looking for the old symbols
will still find them on docarchives.dlang.io and we can add a nice changelog
entry that these symbols have been removed from the documentation if really necessary.
Imho, std.algorithm should be one of the flagship modules and have superb documentation.

### DIFF
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -541,12 +541,22 @@ $(HTTP sgi.com/tech/stl/copy_backward.html, STL's copy_backward'):
 /**
 Assigns `value` to each element of input _range `range`.
 
+Alternatively, instead of using a single `value` to fill the `range`,
+a `filter` $(REF_ALTTEXT forward _range, isForwardRange, std,_range,primitives)
+can be provided. The length of `filler` and `range` do not need to match, but
+`filler` must not be empty.
+
 Params:
         range = An
                 $(REF_ALTTEXT input _range, isInputRange, std,_range,primitives)
                 that exposes references to its elements and has assignable
                 elements
         value = Assigned to each element of range
+        filler = A
+                $(REF_ALTTEXT forward _range, isForwardRange, std,_range,primitives)
+                representing the _fill pattern.
+
+Throws: If `filler` is empty.
 
 See_Also:
         $(LREF uninitializedFill)
@@ -712,18 +722,7 @@ if ((isInputRange!Range && is(typeof(range.front = value)) ||
     }
 }
 
-/**
-Fills `range` with a pattern copied from `filler`. The length of
-`range` does not have to be a multiple of the length of $(D
-filler). If `filler` is empty, an exception is thrown.
-
-Params:
-    range = An $(REF_ALTTEXT input _range, isInputRange, std,_range,primitives)
-            that exposes references to its elements and has assignable elements.
-    filler = The
-             $(REF_ALTTEXT forward _range, isForwardRange, std,_range,primitives)
-             representing the _fill pattern.
- */
+/// ditto
 void fill(InputRange, ForwardRange)(InputRange range, ForwardRange filler)
 if (isInputRange!InputRange
     && (isForwardRange!ForwardRange
@@ -2181,14 +2180,19 @@ if (isBidirectionalRange!Range
 
 // reverse
 /**
-Reverses `r` in-place.  Performs `r.length / 2` evaluations of $(D
-swap).
+Reverses `r` in-place.  Performs `r.length / 2` evaluations of `swap`.
+UTF sequences consisting of multiple code units are preserved properly.
+
 Params:
     r = a $(REF_ALTTEXT bidirectional range, isBidirectionalRange, std,range,primitives)
-    with swappable elements or a random access range with a length member
+        with either swappable elements, a random access range with a length member,
+        or a narrow string
 
-See_Also:
-    $(HTTP sgi.com/tech/stl/_reverse.html, STL's _reverse), $(REF retro, std,range) for a lazy reversed range view
+Note:
+    When passing a string with unicode modifiers on characters, such as `\u0301`,
+    this function will not properly keep the position of the modifier. For example,
+    reversing `ba\u0301d` ("bád") will result in d\u0301ab ("d́ab") instead of
+    `da\u0301b` ("dáb").
 */
 void reverse(Range)(Range r)
 if (isBidirectionalRange!Range && !isRandomAccessRange!Range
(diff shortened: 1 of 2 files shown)


std/algorithm/mutation.d
```diff
@@ -541,12 +541,22 @@ $(HTTP sgi.com/tech/stl/copy_backward.html, STL's copy_backward'):
...
+can be provided. The length of `filler` and `range` do not need to match, but
+`filler` must not be empty.
+
 Params:
         range = An
                 $(REF_ALTTEXT input _range, isInputRange, std,_range,primitives)
                 that exposes references to its elements and has assignable
                 elements
         value = Assigned to each element of range
+        filler = The
```
wilzbach: The was copied from the previous overload, but I think "An" would be better here.

std/algorithm/mutation.d
```diff
@@ -2181,14 +2180,22 @@ if (isBidirectionalRange!Range
...
+    with swappable elements, a random access range with a length member or a narrow string
 
 See_Also:
     $(HTTP sgi.com/tech/stl/_reverse.html, STL's _reverse), $(REF retro, std,range) for a lazy reversed range view
+
+Note:
+    When passing a string with unicode modifiers on characters, such as `\u0301`,
+    this function will not properly keep the position of the modifier. For example,
+    reversing `ba\u0301d` ("bád") will result in d\u0301ab ("d́ab") instead of
+    `da\u0301b` ("dáb").
```
wilzbach: Imho it's bad to have "Bug" written to loud if it's not really a bug, but rather an important "Gotcha".

std/algorithm/searching.d
```diff
@@ -3738,15 +3699,15 @@ Params:
     range = The $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
     to search.
 
-Complexity: O(n)
-    Exactly `n - 1` comparisons are needed.
+Complexity: $(BIGOH range.length)
+    Exactly `range.length - 1` comparisons are needed.
```
wilzbach: Even though it's obvious, `n` isn't defined.

std/algorithm/searching.d
```diff
@@ -1457,35 +1447,51 @@ private auto extremum(alias selector = "a < b", Range,
...
 
-`haystack` advanced such that the front element is the one searched for;
-that is, until `binaryFun!pred(haystack.front, needle)` is `true`. If no
-such position exists, returns an empty `haystack`.
+    `haystack` advanced such that the front element is the one searched for;
+    that is, until `binaryFun!pred(haystack.front, needle)` is `true`. If no
+    such position exists, returns an empty `haystack`.
 
-See_Also:
-     $(HTTP sgi.com/tech/stl/_find.html, STL's _find)
```
wilzbach: There's no reason to link to C++ here.
See also: https://github.com/dlang/dlang.org/pull/2103

std/algorithm/searching.d
```diff
@@ -1457,35 +1447,51 @@ private auto extremum(alias selector = "a < b", Range,
...
+
+Note:
+    `find` behaves similar to `dropWhile` in other languages.
+
+Complexity:
+    `find` performs $(BIGOH walkLength(haystack)) evaluations of `pred`.
+    There are specializations that improve performance by taking
+    advantage of $(REF_ALTTEXT bidirectional range, isBidirectionalRange, std,range,primitives)
+    or random access in the given ranges (where possible), depending on the statistics
+    of the two ranges' content.
```
aliak00: ...advantage of bidirectional or random access ranges, depending...

Note: as a reader, I'm not sure I'd understand what these statistics are.. by statistics do you mea type qualifiers i.e. const/immutable/etc?

std/algorithm/mutation.d
```diff
@@ -541,12 +541,22 @@ $(HTTP sgi.com/tech/stl/copy_backward.html, STL's copy_backward'):
 /**
```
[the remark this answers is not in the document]
wilzbach: This is all very valuable input!
BTW most of the bits were actually just copy/pasted from the existing Ddoc overloads which I removed, so if you ever have a bit time, you should definitely try to do a pass over std.algorithm. I doubt that many programmers here used to edit articles for a living ;-)

std/algorithm/mutation.d
```diff
@@ -2181,14 +2180,21 @@ if (isBidirectionalRange!Range
...
-Reverses `r` in-place.  Performs `r.length / 2` evaluations of $(D
-swap).
+Reverses `r` in-place.  Performs `r.length / 2` evaluations of `swap`.
+UTF sequences consisting of multiple code units are preserved properly.
+
 Params:
     r = a $(REF_ALTTEXT bidirectional range, isBidirectionalRange, std,range,primitives)
-    with swappable elements or a random access range with a length member
+    with swappable elements, a random access range with a length member or a narrow string
```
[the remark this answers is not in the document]
wilzbach: There are no super nits - everything is valuable input. Thanks!

std/algorithm/searching.d
wilzbach: That was moved 1:1 from the existing overload. It's hard for me to understand as well and I guess whoever wrote this wanted to refer to taking advantage of a range e.g. being random access.

I simply removed this confusing bit.

std/algorithm/mutation.d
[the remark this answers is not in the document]
aliak00: Ah right, I didn't notice the red parts of the diff so failed to realize that actually. But thanks! I'll try and have a go at it some day :)

andralex:
Thanks for this work!

andralex:
pls take a look at the autotester

JackStouffer:
That find example needs some work.

PR #6093, opened 2018-01-30
Fix Issue 18328 - algorithm.startsWith can compare narrow string leng…

n8sh:
…ths in more circumstances


JackStouffer:
Merging manually because of Circle

PR #6094, opened 2018-01-30
Fix Issue 18329 - std.algorithm.startsWith & endsWith can sometimes statically determine decoding unnecessary

n8sh:
Uses an ugly string comparison to check for the default predicate.


JackStouffer:
Merging since Circle is broken

PR #6095, opened 2018-01-30
Fix some comments from #5902

joakim-noah:


std/datetime/timezone.d
```diff
@@ -1994,41 +1994,37 @@ public:
...
-            The default directory where the TZ Database files are. It's empty
-            for Windows, since Windows doesn't have them.
+            The default directory where the TZ Database files are stored. It's
+            empty for Windows, since Windows doesn't have them. You can also use
+            the TZDatabaseDir version to pass an arbitrary path at compile-time,
+            rather than hard-coding it here. Android concatenates all time zone
+            data into a single file called tzdata and stores it in the directory
+            below.
           +/
+        enum defaultTZDatabaseDir = "";
```
joakim-noah: First `StdDdoc` comment I've written, not sure if I need to put this blank enum here.

jmdavis: You do. You can't document a symbol that isn't there.

std/digest/murmurhash.d
```diff
@@ -505,15 +505,16 @@ struct MurmurHash3(uint size /* 32 or 128 */ , uint opt = size_t.sizeof == 8 ? 6
         // Buffer should never be full while entering this function.
         assert(bufferSize < Element.sizeof);
 
-        // Check if we don't fill up a whole block buffer.
+        // Check if the incoming data doesn't fill up a whole block buffer.
```
joakim-noah: @MetaLang, this is for you.

wilzbach:
Used the `git commit --amend` trick to restart DAutoTest (it was failing due to https://github.com/dlang/dub/issues/1336)

PR #6097, opened 2018-01-30
Remove references to STL

wilzbach:
There's no reason and benefit to link to C++ in our standard library exception for a few
special cases (e.g. saying that this function behaves like XYZ in another language
for better searchability).

See also: dlang/dlang.org#2103

### DIFF
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -363,9 +363,6 @@ Params:
 
 Returns:
     The unfilled part of target
-
-See_Also:
-    $(HTTP sgi.com/tech/stl/_copy.html, STL's _copy)
  */
 TargetRange copy(SourceRange, TargetRange)(SourceRange source, TargetRange target)
 if (areCopyCompatibleArrays!(SourceRange, TargetRange))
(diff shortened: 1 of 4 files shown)


std/algorithm/mutation.d
```diff
@@ -116,9 +116,6 @@ Params:
 
 Returns:
     The number of elements brought to the front, i.e., the length of `back`.
-
-See_Also:
-    $(HTTP sgi.com/tech/stl/_rotate.html, STL's rotate)
```
JackStouffer: I would keep this one for search-ability since the names are different.

std/algorithm/searching.d
```diff
@@ -2587,9 +2587,6 @@ Returns:
 `r` advanced to the first occurrence of two adjacent elements that satisfy
 the given predicate. If there are no such two elements, returns `r` advanced
 until empty.
-
-See_Also:
-     $(HTTP sgi.com/tech/stl/adjacent_find.html, STL's adjacent_find)
```
JackStouffer: ditto

std/algorithm/mutation.d
wilzbach: Ok. Updated.

PR #6098, opened 2018-01-30
Removed auto-decoding from std.algorithm.searching.balancedParens

JackStouffer:
I'm not sold on this function's usefulness, but fixing auto-decoding is simple so we might as well.

```
0
old		4 secs, 968 ms, 536 μs, and 6 hnsecs
new		1 sec, 988 ms, and 715 μs
$ dmd -O -inline -release test.d && ./test
0
old		16 secs, 874 ms, 710 μs, and 5 hnsecs
new		6 secs, 49 ms, 780 μs, and 8 hnsecs
```

```d
import std.stdio;
import std.algorithm;
import std.conv;
import std.ascii;
import std.range;
import std.traits;
import std.string;
import std.datetime.timezone;
import std.utf;
import std.random;

enum testCount = 100_000_000;

__gshared a = "1 + (2 * (3 + 1 / 2)";

bool balancedParens1(Range, E)(Range r, E lPar, E rPar,
        size_t maxNestingLevel = size_t.max)
if (isInputRange!(Range) && is(typeof(r.front == lPar)))
{
    size_t count;

    for (; !r.empty; r.popFront())
    {
        if (r.front == lPar)
        {
            if (count > maxNestingLevel) return false;
            ++count;
        }
        else if (r.front == rPar)
        {
            if (!count) return false;
            --count;
        }
    }
    return count == 0;
}

bool balancedParens2(Range, E)(Range r, E lPar, E rPar,
        size_t maxNestingLevel = size_t.max)
if (isInputRange!(Range) && is(typeof(r.front == lPar)))
{
    size_t count;

    static if (is(Unqual!(ElementEncodingType!Range) == Unqual!E) && isNarrowString!Range)
        auto rn = r.byCodeUnit;
    else
        alias rn = r;

    for (; !rn.empty; rn.popFront())
    {
        if (rn.front == lPar)
        {
            if (count > maxNestingLevel) return false;
            ++count;
        }
        else if (rn.front == rPar)
        {
            if (!count) return false;
            --count;
        }
    }
    return count == 0;
}

void main()
{
    long res;

    auto result = to!Duration(benchmark!(() => res += a.balancedParens1('(', ')'))(testCount)[0]);
    auto result2 = to!Duration(benchmark!(() => res += a.balancedParens2('(', ')'))(testCount)[0]);

    writeln(res); // force execution

    writeln("old", "\t\t", result);
    writeln("new", "\t\t", result2);
}
```

### DIFF
--- a/std/algorithm/searching.d
+++ b/std/algorithm/searching.d
@@ -233,14 +233,25 @@ bool balancedParens(Range, E)(Range r, E lPar, E rPar,
 if (isInputRange!(Range) && is(typeof(r.front == lPar)))
 {
     size_t count;
-    for (; !r.empty; r.popFront())
+
+    static if (is(Unqual!(ElementEncodingType!Range) == Unqual!E) && isNarrowString!Range)
     {
-        if (r.front == lPar)
+        import std.utf : byCodeUnit;
+        auto rn = r.byCodeUnit;
+    }
+    else
+    {
+        alias rn = r;
+    }
+
+    for (; !rn.empty; rn.popFront())
+    {
+        if (rn.front == lPar)
         {
             if (count > maxNestingLevel) return false;
             ++count;
         }
-        else if (r.front == rPar)
+        else if (rn.front == rPar)
         {
             if (!count) return false;
             --count;
@@ -250,7 +261,7 @@ if (isInputRange!(Range) && is(typeof(r.front == lPar)))
 }
 
 ///
-@safe unittest
+@safe pure unittest
 {
     auto s = "1 + (2 * (3 + 1 / 2)";
     assert(!balancedParens(s, '(', ')'));
@@ -260,6 +271,8 @@ if (isInputRange!(Range) && is(typeof(r.front == lPar)))
     assert(!balancedParens(s, '(', ')', 0));
     s = "1 + (2 * 3 + 1) / (2 - 5)";
     assert(balancedParens(s, '(', ')', 0));
+    s = "f(x) = ⌈x⌉";
+    assert(balancedParens(s, '⌈', '⌉'));
 }
 
 /**


wilzbach:
> I'm not sold on this function's usefulness

Me neither, imho it's pretty much useless (or only in very rare occasion useful and shouldn't have been added to Phobos)
BTW a generalization of this - a function that maintains its own stack and returns a range like `untilClosingParentheses` in the ddoc parser on dlang.org would have been useful - see also:
https://github.com/dlang/dlang.org/pull/2060#discussion_r163830331

std/algorithm/searching.d
```diff
@@ -233,14 +233,25 @@ bool balancedParens(Range, E)(Range r, E lPar, E rPar,
...
+        import std.utf : byCodeUnit;
+        auto rn = r.byCodeUnit;
+    }
+    else
     {
-        if (r.front == lPar)
+        alias rn = r;
+    }
+
+    for (; !rn.empty; rn.popFront())
```
wilzbach: You could have used the opportunity to use a `foreach` here ;-)

std/algorithm/searching.d
```diff
@@ -233,14 +233,25 @@ bool balancedParens(Range, E)(Range r, E lPar, E rPar,
 if (isInputRange!(Range) && is(typeof(r.front == lPar)))
 {
     size_t count;
-    for (; !r.empty; r.popFront())
+
+    static if (is(Unqual!(ElementEncodingType!Range) == Unqual!E) && isNarrowString!Range)
+    {
+        import std.utf : byCodeUnit;
+        auto rn = r.byCodeUnit;
```
wilzbach: I think this requires a test for a `lPar` character - maybe someone used one of the many variations: http://xahlee.info/comp/unicode_matching_brackets.html

std/algorithm/searching.d
JackStouffer: I disagreed with the rejection of `for` for DMD, so I'm not going to go through the trouble of removing it when it's already there.

JackStouffer:
@wilzbach added test

andralex:
Yah we should have an algorithm `findMatchingParen` that assumes `r.front` is the opening paren and advances the range until it's positioned on the corresponding closing paren. An optional `rParen` can be passed for unusual closing parens.

quickfur:
@andralex But to be fully useful, the function will need to count nested parens, so it will need to know what the start parens is. Or are you saying it will just take `r.front` as the opening parens?

andralex:
Yes, the range is positioned initially on the opening paren, and is advanced through to the corresponding closing paren. The balance of the range is returned (empty if no closing paren).

wilzbach:
> Yes, the range is positioned initially on the opening paren, and is advanced through to the corresponding closing paren. The balance of the range is returned (empty if no closing paren).

How would it know what's matching to `[`, `(`, `<`, ... without hard-coding this?

JackStouffer:
@andralex @quickfur @wilzbach Let's move this discussion to https://issues.dlang.org/show_bug.cgi?id=18336

PR #6099, opened 2018-01-30
Bump CircleCi to 2.078.1

wilzbach:
I can't reproduce the failures locally, so I assume bumping is the easiest
course of action.

### DIFF
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -2,8 +2,7 @@
 
 set -uexo pipefail
 
-HOST_DMD_VER=2.072.2 # same as in dmd/src/posix.mak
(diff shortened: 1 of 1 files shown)


wilzbach:
So I can reproduce the failure locally with:

```
. $(~/dlang/install.sh dmd-2.072.1 -a)
make -f posix.mak clean
make -f posix.mak std/traits.publictests
```

So DMD 2.072.1 is definitely running into an infinite loop here. Luckily that has been fixed...

wilzbach:
Great - now we still have disabled DScanner as while running it, the GC is segfaulting (see https://github.com/dlang/phobos/pull/5929), but maybe that problem is now gone too...

PR #6101, opened 2018-01-30
Issue 10828 - Add Output Range Overloads to DateTime toString Methods

JackStouffer:


std/datetime/date.d
```diff
@@ -2861,23 +2880,42 @@ public:
...
+                writer,
                 _tod._hour,
                 _tod._minute,
                 _tod._second
             );
         }
         catch (Exception e)
         {
-            assert(0, "format() threw.");
+            assert(0, "formattedWrite() threw.");
```
n8sh: Does it make sense to have this try/catch assert when `writer` can be any arbitrary output range?

JackStouffer:
@n8sh @wilzbach Fixed. Also fixed it for all of the others while I was at it.

JackStouffer:
Also, please view the diff using this link https://github.com/dlang/phobos/pull/6101/files?w=1

It removes the white space changes and makes reviewing a lot easier.

PR #6102, opened 2018-01-31
fix Issue 18318 - std.net.curl.download silently ignores non-2xx...

MartinNowak:
- document, handle, and throw HTTPStatusException on non-2xx responses in other high-level API functions (download, upload, byLineAsync, byChunkAsync)
- similar to how get, post, et.al. work
- still downloads/fetches the error content (e.g. 404 page)

Not fun to work on this at all, as both std.net.curl and std.concurrency are full of issues and pitfalls.
Maybe we should deprecate the async functions, and point people to use std.parallelism instead, e.g. with a document example in std.net.curl.

### DIFF
--- a/changelog/curl-http-status-exception.dd
+++ b/changelog/curl-http-status-exception.dd
@@ -1,0 +1,6 @@
+std.net.curl methods now throw on non-2xx status codes
+
+The methods $(REF get,std,net,curl), $(REF post,std,net,curl), $(REF put,std,net,curl), $(REF del,std,net,curl), $(REF trace,std,net,curl), $(REF connect,std,net,curl), $(REF patch,std,net,curl), $(REF byLine,std,net,curl), $(REF byChunk,std,net,curl), $(REF byLineAsync,std,net,curl), and $(REF byChunkAsync,std,net,curl) were changed to throw an $(REF HTTPStatusException,std,net,curl) when the server responds with a non-success status code.
+
+For download, byLineAsync, and byChunkAsync the exception is thrown
+after receiving any error response body.
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -432,8 +432,7 @@ if (isCurlConn!Conn)
         }
         conn.perform();
         static if (is(Conn : HTTP))
-            enforce(statusLine.code / 100 == 2, new HTTPStatusException(statusLine.code,
-                    format("HTTP request returned status code %d (%s)", statusLine.code, statusLine.reason)));
+            enforce(statusLine.code / 100 == 2, new HTTPStatusException(statusLine));
     }
     else
     {
@@ -542,8 +541,7 @@ if (isCurlConn!Conn)
         }
         conn.perform();
         static if (is(Conn : HTTP))
-            enforce(statusLine.code / 100 == 2, new HTTPStatusException(statusLine.code,
-                    format("HTTP request returned status code %d (%s)", statusLine.code, statusLine.reason)));
+            enforce(statusLine.code / 100 == 2, new HTTPStatusException(statusLine));
     }
 }
 
@@ -1179,8 +1177,7 @@ private auto _basicHTTP(T)(const(char)[] url, const(void)[] sendData, HTTP clien
     };
     client.onReceiveStatusLine = (HTTP.StatusLine l) { statusLine = l; };
     client.perform();
-    enforce(statusLine.code / 100 == 2, new HTTPStatusException(statusLine.code,
-            format("HTTP request returned status code %d (%s)", statusLine.code, statusLine.reason)));
+    enforce(statusLine.code / 100 == 2, new HTTPStatusException(statusLine));
 
     return _decodeContent!T(content.data, client.p.charset);
 }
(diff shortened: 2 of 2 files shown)


MartinNowak:
FYI @ikod in case this affects [ikod/dlang-requests](https://github.com/ikod/dlang-requests).

wilzbach:
> FYI @ikod in case this affects ikod/dlang-requests.

dlang-requests is written in native D, so I don't see how it would be affected.

> Maybe we should deprecate the async functions, and point people to use std.parallelism instead, 

Imho we should deprecate the entire module and point people to the `requests` package.
It's so much easier to use, native D and supports async with Vibe.d eventcore out-of-the-box.

CyberShadow:
Thanks!

> Imho we should deprecate the entire module and point people to the requests package.

I quite like std.net.curl for small programs.

- Very simple API, doesn't get simpler than `get` / `download`.
- Using it does not slow down the build of the D program.
- Verbose mode is a simple, single switch that enables output useful for diagnosing incorrect requests.
- Supports cookie jars, SSL client (!) certificates, FTP, and other weird things that come with using cURL.

ikod:
> FYI @ikod in case this affects ikod/dlang-requests.

Hello Martin,

No, dlang-requests do not use curl, so it can't be affected.

Anyway, thanks for the info!

MartinNowak:
IMO, simple D out-of-the box examples are enough of an argument to keep std.net.curl.
But it's indeed very difficult to use for anything non-trivial, and the async APIs should better be deprecated.

MartinNowak:
Unittest on Win32 seems to hang, closing PR to save tester resources until that's resolved.

wilzbach:
@MartinNowak there are no resources wasted with keeping a PR open for a while.
Auto-tester doesn't rerun old PRs anyhow.

wilzbach:
(rebased)

wilzbach:
The failure is `fetchzip` seems to be permanent:

```
[ERROR] :DUB should have retried download on server error multiple times, but only tried 0 times. command failed
```

CyberShadow:
I think this needs a changelog entry, as programs that "worked" previously will now throw. (Programs written before this change might have elected to search for an error message string in the output, or otherwise validate the downloaded data, to detect HTTP errors.)

andre2007:
As far as I understand dub issue https://github.com/dlang/dub/issues/1530 is related to this pr.
Dub has a retry logic if http download is failing. This retry logic only works if HTTPStatusException is thrown. But at the moment instead (e.g. in case of a timeout) CurlException is thrown.

Is my understanding correct, with this pr an HTTPStatusException is thrown also for timeout issues?
What is missing in this PR?

MartinNowak:
> Is my understanding correct, with this pr an HTTPStatusException is thrown also for timeout issues?

No, as it's name and docs say it's thrown when an exceptional HTTP status is returned.

changelog/curl-http-status-exception.dd
```diff
@@ -0,0 +1,6 @@
+std.net.curl methods now throw on non-2xx status codes
+
+The methods $(REF get,std,net,curl), $(REF post,std,net,curl), $(REF put,std,net,curl), $(REF del,std,net,curl), $(REF trace,std,net,curl), $(REF connect,std,net,curl), $(REF patch,std,net,curl), $(REF byLine,std,net,curl), $(REF byChunk,std,net,curl), $(REF byLineAsync,std,net,curl), and $(REF byChunkAsync,std,net,curl) were changed to throw an $(REF HTTPStatusException,std,net,curl) when the server responds with a non-success status code.
+
+For download, byLineAsync, and byChunkAsync the exception is thrown
```
CyberShadow: Why no `REF` here? Or at least backticks for D symbols

CyberShadow:
@MartinNowak Do you think the custom logic could be replaced with `CURLOPT_FAILONERROR`, so that we could match the behavior with `curl --fail`?

wilzbach:
With this change, `test/fetchzip.sh` fails in DUB.

```
Full exception: object.Exception@source/dub/packagesuppliers/registry.d(71): Failed to download package gitcompatibledubpackage from http://localhost:10255/packages/gitcompatibledubpackage/1.0.2.zip
```

```
[ERROR] :DUB should have retried download on server error multiple times, but only tried 0 times. command failed
```

So it looks like DUB retry logic doesn't handle catching these CurlExceptions that well.

changelog/curl-http-status-exception.dd
```diff
@@ -0,0 +1,6 @@
+std.net.curl methods now throw on non-2xx status codes
+
+The methods $(REF get,std,net,curl), $(REF post,std,net,curl), $(REF put,std,net,curl), $(REF del,std,net,curl), $(REF trace,std,net,curl), $(REF connect,std,net,curl), $(REF patch,std,net,curl), $(REF byLine,std,net,curl), $(REF byChunk,std,net,curl), $(REF byLineAsync,std,net,curl), and $(REF byChunkAsync,std,net,curl) were changed to throw an $(REF HTTPStatusException,std,net,curl) when the server responds with a non-success status code.
```
berni44: Shouldn't this line be broken into several lines?

LightBender:

PR #6104, opened 2018-02-01
std.uni struct Grapheme mem fun signatures: Replace some @trusted by @safe;...

carblue:
…preparation for -dip1000, issues #17961, #18110.

Changes apply to function signatures only and are trivial except referring to return attribute: Seems to be undocumented in DIP25/DIP1000, similar to https://dlang.org/spec/function.html#return-ref-parameters;
It's what template inference does for opSlice: "return attribute ensures the returned object will not outlive the Grapheme instance".
Without return, there are "-dip1000 errors" like this one: returning sliceOverIndexed(a, b, &this) escapes a reference to parameter this, perhaps annotate with return.

### DIFF
--- a/std/uni.d
+++ b/std/uni.d
@@ -7395,7 +7395,7 @@ if (isInputRange!Range && is(Unqual!(ElementType!Range) == dchar))
 
     See_Also: $(LREF decodeGrapheme), $(LREF graphemeStride)
 +/
-@trusted struct Grapheme
+@safe struct Grapheme
 {
     import std.exception : enforce;
     import std.traits : isDynamicArray;
@@ -7417,7 +7417,7 @@ public:
     }
 
     /// Gets a $(CODEPOINT) at the given index in this cluster.
-    dchar opIndex(size_t index) const pure nothrow @nogc
+    dchar opIndex(size_t index) const pure nothrow @nogc @trusted
     {
         assert(index < length);
         return read24(isBig ? ptr_ : small_.ptr, index);
@@ -7430,7 +7430,7 @@ public:
         Use of this facility may invalidate grapheme cluster,
         see also $(LREF Grapheme.valid).
     +/
-    void opIndexAssign(dchar ch, size_t index) pure nothrow @nogc
+    void opIndexAssign(dchar ch, size_t index) pure nothrow @nogc @trusted
     {
         assert(index < length);
         write24(isBig ? ptr_ : small_.ptr, ch, index);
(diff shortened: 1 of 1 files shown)


std/uni.d
```diff
@@ -7478,7 +7478,7 @@ public:
 
         See_Also: $(LREF Grapheme.valid)
     +/
-    ref opOpAssign(string op)(dchar ch)
+    ref opOpAssign(string op)(dchar ch) @trusted
     {
         static if (op == "~")
         {
```
wilzbach: It we can use stdx.checked in general - it's just not @safe by default atm, but there's https://github.com/dlang/phobos/pull/5928

std/uni.d
```diff
@@ -7560,7 +7560,7 @@ public:
         return r.length == 0;
     }
 
-    this(this) pure @nogc nothrow
+    this(this) pure @nogc nothrow @trusted
```
wilzbach: This should use stdx.allocator.makeArray - it's `@safe`

```d
import std.experimental.allocator.mallocator : Mallocator;
import std.experimental.allocator : makeArray;

auto p = Mallocator.instance.makeArray!ubyte(20);
```

https://run.dlang.io/is/iZGDzT

std/uni.d
```diff
@@ -7578,7 +7578,7 @@ public:
         }
     }
 
-    ~this() pure @nogc nothrow
+    ~this() pure @nogc nothrow @trusted
     {
         import core.memory : pureFree;
```
wilzbach: Go with `dispose`, instead:

```d
import std.experimental.allocator.mallocator : Mallocator;
import std.experimental.allocator : makeArray, dispose;
alias alloc = Mallocator.instance;
auto p = alloc.makeArray!ubyte(20);
alloc.dispose(p);
```
https://run.dlang.io/is/t2IMq2

but it needs to be fixed first (https://issues.dlang.org/show_bug.cgi?id=18347)

std/uni.d
```diff
@@ -7612,7 +7612,7 @@ private:
         }
     }
 
-    void convertToBig() pure @nogc nothrow
+    void convertToBig() pure @nogc nothrow @trusted
```
wilzbach: stdx.allocator to the help!

std/uni.d
```diff
@@ -7417,7 +7417,7 @@ public:
     }
 
     /// Gets a $(CODEPOINT) at the given index in this cluster.
-    dchar opIndex(size_t index) const pure nothrow @nogc
+    dchar opIndex(size_t index) const pure nothrow @nogc @trusted
     {
         assert(index < length);
         return read24(isBig ? ptr_ : small_.ptr, index);
```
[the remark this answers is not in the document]
carblue: Yep!

std/uni.d
carblue: Thanks for Your review and helpful hints!
Currently I won't touch function bodies and be focussed on function/struct signatures for -dip1000 compliance.
@wilzbach Please close this PR, it will be integrated in an update for PR #6041 as there are overlaps.

std/uni.d
n8sh: Can't use it here because it's not `pure`.

carblue:
```
pragma(msg, "std.uni.d:Grapheme.this(this):", __LINE__, " ", typeof(&Grapheme.__postblit));
```

PR #6106, opened 2018-02-01
Fix Issue 15461 - std.stdio.File ctor that accepts FILE* should not be hidden

JackStouffer:
From my cursory investigation there's no cross-platform way to tell from a `FILE*` what its name is, or to tell if it's still open.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -420,10 +420,14 @@ Throws: $(D ErrnoException) if the file could not be opened.
         import std.conv : text;
         import std.exception : errnoEnforce;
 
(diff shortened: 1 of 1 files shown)


JackStouffer:
Wait, I just saw `wrapFile` exists. \*sigh\*

wilzbach:
Then improve the documentation?

JackStouffer:
It's more a problem of I missed it initally because Sublime Text mis-parses the comments in std.stdio

screen shot 2018-02-01 at 1 22 12 pm

PR #6107, opened 2018-02-01
Replaced version(unittest) blocks with version(StdUnittest)

JackStouffer:
Should result in faster compiles for unittest builds in some cases.


JackStouffer:
Fixed

PR #6110, opened 2018-02-02
Fix issue 18134 - BitArray right shift broken if length is multiple of 8*size_t.sizeof

byebye:
There was a bug caused by misunderstanding `endMask` - which is a mask for bits after the full word. Therefore, when length is multiple of `size_t.sizeof`, i.e. `dim == fullWords`, then `endMask == 0`.

I fixed also a similar problem present in `flip()`, which would reset all the bits in the word after the array.

### DIFF
--- a/std/bitmanip.d
+++ b/std/bitmanip.d
@@ -1006,7 +1006,8 @@ public:
         foreach (i; 0 .. fullWords)
             _ptr[i] = ~_ptr[i];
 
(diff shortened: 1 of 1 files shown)


byebye:
Tests weren't passing because I made a mistake in unittests - I specified length as `size_t.sizeof` instead of `8*size_t.sizeof`. Hopefully they will pass now.

PR #6111, opened 2018-02-02
Add warning about the old std.json module from D1

wilzbach:
We shouldn't advertise code like this:

```
JSONValue json = parseJSON(`{"foo": 1}`);
auto a = &(json.object());
json.uinteger = 0;
(*a)["hello"] = 1;
```
https://run.dlang.io/is/KW86zH

https://dlang.org/phobos/std_json.html

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -3,6 +3,12 @@
 /**
 JavaScript Object Notation
 
+$(MESSAGE_BOX red, This module is considered out-dated and not up to Phobos'
+      current standards. It will remain until we have a suitable replacement,
+      but be aware that it will not remain long term.
+      A potential replacement is $(LINK2 https://github.com/s-ludwig/std_data_json, `std.experimnetal.json`).
+)
+
 Copyright: Copyright Jeremie Pelletier 2008 - 2009.
 License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
 Authors:   Jeremie Pelletier, David Herberth
--- a/std/xml.d
+++ b/std/xml.d
@@ -1,9 +1,11 @@
 // Written in the D programming language.
 
 /**
-$(RED Warning: This module is considered out-dated and not up to Phobos'
+$(MESSAGE_BOX red, This module is considered out-dated and not up to Phobos'
       current standards. It will remain until we have a suitable replacement,
-      but be aware that it will not remain long term.)
+      but be aware that it will not remain long term.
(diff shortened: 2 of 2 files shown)


JackStouffer:
I don't know about this. I know that people come into the Learn group and and are confused about the status of std.xml and want to know what to do. While these modules aren't fast or pretty, they do work while the linked modules do not. I don't think we should worry users about the stability of these modules when we don't have a road-map to replacing them. What we currently have are modules which *might* replace the Phobos versions if someone picks up the torch.

std/json.d
```diff
@@ -3,6 +3,12 @@
 /**
 JavaScript Object Notation
 
+$(RED Warning: This module is considered out-dated and not up to Phobos'
+      current standards. It will remain until we have a suitable replacement,
+      but be aware that it will not remain long term.
+      You can preview the future $(LINK2 https://github.com/s-ludwig/std_data_json, `std.data.json`).
+)
```
andralex: A mention of the upcoming package is worthwhile, but we do not lay judgment.

JackStouffer:
AFAIK Sönke has stopped working on `std.data.json` to focus on vibe.d (last commit was almost a year ago). That's why I said we need someone to pick up the torch before we start endorsing it.

quickfur:
Yeah, if nobody is actively working on `std.data.json`, then we shouldn't be advertising it in the Phobos docs.

CyberShadow:
> auto a = &(json.object());

BTW, I think we can fix this example by making JSONValue not use unions. It will use more memory, but it will be safe. I think that's more important for Phobos.

wilzbach:
> Yeah, if nobody is actively working on std.data.json, then we shouldn't be advertising it in the Phobos docs.

Hmm, how should somebody pick up the torch if they don't know about it?
I reworded that sentence and also used the nicer message boxes:

image

image

andralex:
Folks, this is really simple. If we have a definite existing or upcoming replacement in the standard library, we can advertise it. If we invite others to work on one, that's nice to mention too. A "see also" section somewhere? Fine, too. Otherwise the standard library isn't in the business of passing judgment about itself. This is not open for negotiation.

JackStouffer:
@wilzbach Adding a simple see also section to the top with a link to Sönke's library that states

> For an alternative JSON library, see ...

seems like the best option here.

rikkimax:
This can be closed, std.xml is gone and std.json has a warning that it may not be what you want.

PR #6112, opened 2018-02-02
Fix issue 18133 - improve docs and add code examples for BitArray constructors

byebye:


std/bitmanip.d
```diff
@@ -799,6 +799,140 @@ private:
...
+        import std.algorithm.comparison : equal;
+
+        auto bc = BitArray(source, si + 1);
+        assert(bc.bitsSet.equal([0, si]));
+        // The unmapped bits from the final word have been cleared.
+        assert(source[1] == 1);
+    }
+
+    // Deliberately undocumented: raw initialization of bit array.
+    this(size_t len, size_t* ptr) nothrow @nogc pure
```
byebye: I'm wondering if this constructor should remain undocumented? I mean, it works almost in the same way as the one above taking `void[]`.

std/bitmanip.d
```diff
@@ -799,18 +801,157 @@ private:
...
+        assert(a == b);
+    }
+
+    ///
+    @system unittest
+    {
+        import std.algorithm.comparison : equal;
+
+        size_t[] source = [0b1100, 0b0011];
+        auto sbits = size_t.sizeof * 8;
```
JackStouffer: `enum`

std/bitmanip.d
```diff
@@ -799,18 +801,157 @@ private:
...
+        auto ba = BitArray(source, source.length * sbits);
+        // The least significant bit in each unit is this unit's starting bit.
+        assert(ba.bitsSet.equal([2, 3, sbits, sbits + 1]));
+    }
+
+    ///
+    @system unittest
+    {
+        size_t[] source = [1, 0b101, 3, 3424234, 724398, 230947, 389492];
+        auto sbits = size_t.sizeof * 8;
```
JackStouffer: `enum`

std/bitmanip.d
```diff
@@ -799,18 +801,157 @@ private:
...
     {
         return lenToDim(_len);
     }
 
-    /**********************************************
-     * Gets the amount of bits in the $(D BitArray).
-     */
-    @property size_t length() const @nogc pure nothrow @safe
+    /**
+    Returns the number of bits in the `BitArray`.
```
JackStouffer: "Returns:"

byebye: I have one issue with that, because when I specify just `Returns:` without any other description, ddox is going to generate an empty description:
ddox_missing_desc
Is this something I should care about? If yes, what is the best way to avoid writing the same thing in both the description and the `Returns` section?

std/bitmanip.d
```diff
@@ -799,18 +801,157 @@ private:
...
+            this[i] = b;
+        }
+    }
+
+    ///
+    @system unittest
+    {
+        import std.algorithm.comparison : equal;
+
+        auto a = BitArray([1, 0, 0, 1, 1]);
```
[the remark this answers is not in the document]
wilzbach: It's a bit array, so using `0` and `1` seems fine to me.

std/bitmanip.d
```diff
@@ -799,18 +801,157 @@ private:
...
+    ---
+    size_t[] source = [1, 2, 3, 3424234, 724398, 230947, 389492];
+    auto sbits = size_t.sizeof * 8;
+    auto ba = BitArray(source, source.length * sbits);
+    foreach (n; 0 .. source.length * sbits)
+    {
+        auto nth_bit = cast(bool) (source[n / sbits] & (1L << (n % sbits)));
+        assert(ba[n] == nth_bit);
+    }
+    ---
```
wilzbach: This example should be added 1:1 as a unittest,s.t. it can't get broken by future released.

std/bitmanip.d
wilzbach: Opening an issue at Ddox: https://github.com/rejectedsoftware/ddox/issues
Our main documentation engine is still Ddoc though, there are still many visual glitches like this one with Ddox.

std/bitmanip.d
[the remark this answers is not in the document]
JackStouffer: Then do an explicit cast, as it's not clear to everyone that the implicit cast is happening.

std/bitmanip.d
byebye: Seems good, I'll change it. I found out this issue is already reported: https://github.com/rejectedsoftware/ddox/issues/172.

std/bitmanip.d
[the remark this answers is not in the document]
byebye: I think I'll make this example to contain both methods and write a comment that an implicit cast occurs.

std/bitmanip.d
byebye: Yeah and actually it's already there (3rd example for this constructor).

JackStouffer:
Please squash the commits.

byebye:
Squashed and rebased onto master. CircleCI is failing, but it doesn't seem like an error caused by this PR.

PR #6113, opened 2018-02-02
Fix issue 15949: Make readText check BOMs.

jmdavis:
This makes it so that readText checks for a BOM. If there is a BOM, it
is for UTF-8, UTF-16, or UTF-32, and it doesn't match the requested
string type, then a UTFException is thrown. Other encodings are let
through in case they happen to work with the requested string type and
pass UTF validation.

Also, this makes it so that readText checks the alignment of the buffer
against the requested string type and throws a UTFException instead of
letting the cast throw an Error.

### DIFF
--- a/changelog/std-file-readText.dd
+++ b/changelog/std-file-readText.dd
@@ -1,0 +1,20 @@
+readText now checks BOMs
+
+$(REF readText, std, file) now checks for a
+$(HTTP https://en.wikipedia.org/wiki/Byte_order_mark, BOM). If a BOM is present
+and it is for UTF-8, UTF-16, or UTF-32, $(REF readText, std, file) verifies
+that it matches the requested string type and the endianness of the machine,
+and if there is a mismatch, a $(REF UTFException, std, utf) is thrown without
+bothering to validate the string.
+
+If there is no BOM, or if the BOM is not for UTF-8, UTF-16, or UTF-32, then the
+behavior is what it's always been, and UTF validation continues as per normal,
+so if the text isn't valid for the requested string type, a
+$(REF UTFException, std, utf) will be thrown.
+
+In addition, before the buffer is cast to the requested string type, the
+alignment is checked (e.g. 5 bytes don't fit cleanly in an array of $(D wchar)
+or $(D dchar)), and a $(REF UTFException, std, utf) is now throw if the
+number of bytes does not align with the requested string type. Previously, the
+alignment was not checked before casting, so if there was an alignment mismatch,
+the cast would throw an Error, killing the program.
--- a/std/file.d
+++ b/std/file.d
@@ -465,54 +465,242 @@ version (linux) @safe unittest
     assert(read(deleteme) == "abcd");
 }
 
-/********************************************
-Read and validates (using $(REF validate, std,utf)) a text file. $(D S)
-can be a type of array of characters of any width and constancy. No
-width conversion is performed; if the width of the characters in file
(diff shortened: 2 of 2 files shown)


changelog/std-file-readText.dd
```diff
@@ -0,0 +1,20 @@
...
+behavior is what it's always been, and UTF validation continues as per normal,
+so if the text isn't valid for the requested string type, a
+$(REF UTFException, std, utf) will be thrown.
+
+In addition, before the buffer is cast to the requested string type, the
+alignment is checked (e.g. 5 bytes don't fit cleanly in an array of $(D wchar)
+or $(D dchar)), and a $(REF UTFException, std, utf) is now throw if the
+number of bytes does not align with the requested string type. Previously, the
+alignment was not checked before casting, so if there was an alignment mismatch,
+the cast would throw an Error, killing the program.
```
CyberShadow: This is much better, thanks!

jmdavis:
I just added [this PR](https://github.com/dlang/phobos/pull/6114), which is somewhat related.

std/file.d
```diff
@@ -465,54 +465,248 @@ version (linux) @safe unittest
...
+        // Strips BOM if present.
+        content.skipOver('\uFEFF');
+        assert(content == "abc"w);
+    }
+    {
+        write(deleteme, "abc");
+        scope(exit) remove(deleteme);
+        // Throws because the file is not valid UTF-16.
+        assertThrown!UTFException(readText!wstring(deleteme));
+    }
```
CyberShadow: This example is a bit of a big, dense ball of code. Some whitespace and comments describing what is being tested / presented would be nice. I don't think we need two `skipOver` examples, too.

jmdavis: Does this look better then?

CyberShadow: Yes; I'd still prefix each code block with a one-line comment of what's being tested. E.g.:

```
// Read UTF-8 text file:
{ ... }
// Attempt to read invalid UTF-16 file:
{ ... }
// Read UTF-8 file with BOM:
{ ... }
```

I still don't think we need two `skipOver` examples.

jmdavis: One shows it with a BOM and one without, which makes it clearer that it does nothing when there is no BOM while still showing that it does something when there is a BOM.

CyberShadow: Yeah, but I don't think that's enough value per LOC. People really shouldn't need to play "spot the difference" with our examples.

jmdavis: I don't agree, but I don't care enough to keep it if you really don't want it there.

wilzbach: How about splitting them up into individual unittests? It will look better on the docs and you can use the comment directly on the unittest:

```d
/// Read UTF-8 ...
unittest
...
```

CyberShadow: > I don't agree, but I don't care enough to keep it if you really don't want it there.

No, I don't feel strongly about it.

> How about splitting them up into individual unittests?

I think that would be best. I was thinking that duplicating imports for each block would be ugly, but that's no longer the case.

changelog/std-file-readText.dd
n8sh: "is now thrown"

std/file.d
jmdavis: I don't really agree that splitting them up is better, but I just did it. So, hopefully this is ready to go now.

jmdavis:
Any more objections, or can we merge this now?

PR #6114, opened 2018-02-03
Add std.encoding.stripBOM.

jmdavis:
stripBOM examines the given range of characters or ubytes for a BOM at
the front of the range and strips it if present (or just returns the
original range if there is no BOM). This should make it more pleasant to
deal with files that start with a BOM - particularly if the program
doesn't want to care if the file starts with a BOM.


jmdavis:
@andralex since this is a new symbol.

std/encoding.d
```diff
@@ -3907,3 +3907,121 @@ if (isForwardRange!Range && is(Unqual!(ElementType!Range) == ubyte))
...
+
+        static if (is(E == char))
+        {
+            import std.range : only;
+            if (bcu.save.startsWith(only(cast(char)0xEF, cast(char)0xBB, cast(char)0xBF)))
+                bcu.popFrontN(3);
+        }
+        else static if (is(E == wchar))
+        {
+            if (bcu.save.byCodeUnit().startsWith(cast(wchar)0xFEFF))
```
JackStouffer: calling byCodeUnit twice

std/encoding.d
```diff
@@ -3907,3 +3907,121 @@ if (isForwardRange!Range && is(Unqual!(ElementType!Range) == ubyte))
...
+                bcu.popFront();
+        }
+        else
+        {
+            if (range.save.byCodeUnit().startsWith(cast(dchar)0x0000FEFF))
```
JackStouffer: ditto

changelog/std-encoding-stripbom.dd
```diff
@@ -0,0 +1,27 @@
...
+(if any) from the front of the range.
+
+If there is no BOM, then the original range is returned.
+
+If the range is a range of $(D ubyte)s, then if $(REF getBOM, std, encoding)
+returns a BOM for the range, that BOM is stripped.
+
+If the range is a range of characters, then the BOM that matches that encoding
+of the string and endianness of the machine is checked for and stripped if
+present, but other types are BOMs are not checked. Strings are treated as
```
CyberShadow: other types of* BOMs

std/encoding.d
jmdavis: *sigh* I guess that I refactored that bit of code a few too many times and missed that. It's fixed now.

std/encoding.d
```diff
@@ -3907,3 +3907,121 @@ if (isForwardRange!Range && is(Unqual!(ElementType!Range) == ubyte))
...
+                bcu.popFront();
+        }
```
CyberShadow: Any reason this entire `static if` couldn't be replaced with `if (range.front == '\uFEFF') range.popFront();` ?

std/encoding.d
```diff
@@ -3907,3 +3907,121 @@ if (isForwardRange!Range && is(Unqual!(ElementType!Range) == ubyte))
...
 /** Constant defining a fully decoded BOM */
 enum dchar utfBOM = 0xfeff;
+
+
+/++
+    Strips the BOM (if present) from the front of the given range and returns
+    the range.
+
+    If the range is a range of $(D ubyte), then $(LREF getBOM) will be used to
+    determined if there is a BOM present, and the matching BOM will be stripped.
```
CyberShadow: determine*

std/encoding.d
```diff
@@ -3907,3 +3907,121 @@ if (isForwardRange!Range && is(Unqual!(ElementType!Range) == ubyte))
...
+    assert(stripBOM("hello world") == "hello world");
+
+    assert(stripBOM((cast(string)[0xEF, 0xBB, 0xBF]) ~ "hello world") ==
+           "hello world");
+
+    assert(stripBOM("\uFEFFhello world"w) == "hello world"w);
+
+    // utf-7
+    ubyte[] utf7 = (cast(ubyte[])[0x2B, 0x2F, 0x76, 0x38, 0x2D]) ~
+                   "hello world".representation;
```
CyberShadow: I'm not sure about this... this is going back to the previous problem of accidentally stripping a piece of UTF-8 text just because it happens to be UTF-7.

I think that the `ubyte` overload should take a `BOM` parameter, which tells it which `BOM` to strip. It should then only strip only that `BOM` (if it's present).

If the program isn't sure which BOM is in the content, it should use `getBOM` first. Stripping arbitrary BOMs doesn't make sense anyway, because if you don't know or care what BOM is in the data, then you won't know the data's encoding and how to process it correctly!

std/encoding.d
```diff
@@ -3907,3 +3907,121 @@ if (isForwardRange!Range && is(Unqual!(ElementType!Range) == ubyte))
...
+        else static if (is(E == wchar))
+        {
+            if (bcu.save.startsWith(cast(wchar)0xFEFF))
+                bcu.popFront();
+        }
+        else
+        {
+            if (range.save.startsWith(cast(dchar)0x0000FEFF))
+                bcu.popFront();
+        }
```
CyberShadow: Reposting obsolete comment:

Any reason this entire `static if` couldn't be replaced with `if (range.front == '\uFEFF') range.popFront();` ?

jmdavis: Do you mean all of the branches or just the `dchar` one? If you mean the `dchar` one, using `startsWith` makes it work with empty ranges, though it could check for empty and then check front. If you mean all of them, I don't know if that works. Does `\uFEFF` encode to the BOM for each encoding? I was going off of what wikipedia and std.encoding said the bytes were.

std/encoding.d
jmdavis: I don't know. I honestly don't care much about encodings other than UTF-8, UTF-16, and UTF-32. Stripping off whichever BOM was there made sense to me for ubytes, since if you want the BOM gone, at that point you have to check them all, though yeah, if you want to do anything meaningful with the range at that point, you already have to know what the encoding is at that point, and you probably had to call `getBOM` anyway. I don't see much point to `stripBOM` if you have to pass it the BOM, especially if you already had to call `getBOM`. At that point, it's trivial to just pop the elements based on the return value of `getBOM`. So, I guess that I can just remove ubyte support from `stripBOM`.

std/encoding.d
CyberShadow: > Does \uFEFF encode to the BOM for each encoding? 

Yes, that's how it's defined. From Wikipedia: "The byte order mark (BOM) is a Unicode character, U+FEFF byte order mark (BOM), [...]"

CyberShadow: > Do you mean all of the branches

Yes, all branches.

> though it could check for empty

Right.

CyberShadow: It will if you let auto-decoding do its thing... *shudder*

jmdavis: No, it won't, because that only works for strings. I mean actual ranges of `char` or `wchar`. I can use `decodeFront` explicitly though.

std/encoding.d
jmdavis: If we don't support ranges of `ubyte` though, maybe this belongs in std.utf, since it's UTF-specific.

jmdavis:
Okay. I made it so that `stripBOM` just works on ranges of `char`, `wchar`, and `dchar` and greatly simplified its implementation. I also moved it to std.utf, since it now only works on UTF.

CyberShadow:
Thanks! Looks great, just needs Andrei's approval.

jmdavis:
> Thanks! Looks great, just needs Andrei's approval.

You'll need to approve too, since you marked it as needing changes.

CyberShadow:
Maybe the `readText` documentation could mention / link to this function.

CyberShadow:
> You'll need to approve too, since you marked it as needing changes.

Thanks for the reminder, done.

jmdavis:
> Maybe the readText documentation could mention / link to this function.

It probably should, but since I have a PR changing `readText` right now, that would create a conflict. So, I figured that either one would be merged, and then I'd add the link to the other, or I'd create a trivial PR after both were merged.

jondegenhardt:
I gave it a quick try in the context of `byLine`, it worked fine.

CyberShadow:
Not UFCSable, for one. Can't do `string text = fn.readText.stripBOM;`. Also won't work with ranges of chars. Also the actual BOM character (`\FEFF`) is kind of an implementation detail that ought to be encapsulated.

andralex:
@CyberShadow Understood. I stay with my decision this is not worth adding a function. Thanks.

CyberShadow:
Fine by me. Jonathan, anything to add?

jmdavis:
> Fine by me. Jonathan, anything to add?

Well, I really don't agree with the decision. It's so easy to screw up checking for the BOM (e.g. I got it backwards quite a bit when writing tests), and it's arguably an implementation detail for most folks (certainly, most folks are just going to want the BOM out of their way so that it doesn't screw up the text), so having to explicitly type the BOM to use it with `skipOver` is a lot less user-friendly IMHO - and it's even worse with ranges of `char` or `wchar` such that it's pretty ugly for any code avoiding auto-decoding. But if Andrei doesn't agree, I don't know what I could do to convince him.

CyberShadow:
If it's of any consolation, and as you said, BOMs aren't very common any more... and UTF-16 / UCS-2 popularity is waning as UTF-8 is taking over, so it's likely that if there was no demand for this function until now, it probably won't be sorely missed in the future.

andralex:
The implementation was overly complex because it did not use `skipOver`.

```d
R stripBOM(R)(R range)
if (isForwardRange!R && isSomeChar!(ElementType!R))
{
    range.skipOver('\uFEFF');
    return range;
}
```

It passes your own unittests out of the box!

Once implemented like that, it becomes painfully obvious this is not really its own function. We can't go around telling people "if you want to skip over an optional prefix, use `skipOver`. If said prefix is the BOM, we have another function for that".

Converting this PR to adding an example to `skipOver` on how to skip the BOM is an awesome thing to do. It will make searches for "skip the BOM" land there. Also it would make excellent use of your and reviewers' time for this PR.

andralex:
@jmdavis I compel you to please do the example thing - thanks!

CyberShadow:
> It passes your own unittests out of the box!

It doesn't. The 2nd and 3rd-to-last tests fail because `skipOver` won't do the right thing for `char`/`wchar` ranges.

andralex:
@CyberShadow good, point taken. I looked into the matter a bit more. The example would be https://gist.github.com/anonymous/51ace27cf3186550aa716df49a93236c where a couple of asserts had to be commented out. A few things that would be very useful here:

* skipOver should be made nogc nothrow (technically a breaking change, but a worthwhile one)
* apparently BOM is unrecommended in UTF8: https://stackoverflow.com/questions/2223882/whats-different-between-utf-8-and-utf-8-without-bom
* the small subtleties involved make a BOM skipping example a great addition to `skipOver`

wilzbach:
Even though it's mostly a graveyard, I submitted these points as issues ...

https://issues.dlang.org/show_bug.cgi?id=18369
https://issues.dlang.org/show_bug.cgi?id=18370

PR #6115, opened 2018-02-03
Issue 13121 - std.algorithm.joiner should return a bidirectional range if possible

wilzbach:
The `joiner` overload with separators is more complex and will require a RandomAccessRange, so I'm leaving that one to another PR.


wilzbach:
Have a look at the changelog entry for a motivation.
This PR enables this fully lazy code:

```d
import std.range : chain, cycle, iota, only, retro, take, zip;
import std.format : format;

static immutable number = "12345678";
static immutable delimiter = ",";
auto formatted = number.retro
    .zip(3.iota.cycle.take(number.length))
    .map!(z => chain(z[0].only, z[1] == 2 ? delimiter : null))
    .joiner
    .retro; // "12,345,678"
```

std/algorithm/iteration.d
```diff
@@ -2495,16 +2475,39 @@ if (isInputRange!RoR && isInputRange!(ElementType!RoR))
...
+            }
+            // We cannot export .save method unless we ensure subranges are not
+            // consumed when a .save'd copy of ourselves is iterated over. So
+            // we need to .save each subrange we traverse.
+            static if (isForwardRange!RoR && isForwardRange!(ElementType!RoR))
+                _current = _items.front.save;
+            else
+                _current = _items.front;
+        end:
+            assert(1);
```
n8sh: What is the purpose of this line?

wilzbach: DMD complains with "found EOF instead of statement" otherwise.

wilzbach:
Ping. Two months passed. Anything blocking this?

n8sh:
You only call `checkFinalElement` to update `reachedFinalElement` in `popBack()`. Isn't that a problem if you pop from both ends? All of the tests you added consume the range without changing direction.

wilzbach:
> You only call checkFinalElement to update reachedFinalElement in popBack(). Isn't that a problem if you pop from both ends? All of the tests you added consume the range without changing direction.

Added two tests, addressed the other pending comments and rebased.
checkFinalElement is only needed in popBack because it's necessary for the case of reaching the final element from both sides. checkFinalElement makes sure that when this happens popBack/back uses `_current` (i.e. the one that is uses by front/popFront) and not its _currentBack element cache.
BTW this is done in this way, s.t. if the bidirectional access is never used, you essentially don't pay for it and only pay for it when you actually use it.

So this should be ready to go.

n8sh:
@wilzbach Please look at the failing build of BlackEdder/ggplotd.

>[...]/std/algorithm/iteration.d(2493,9): Error: field `_currentBack` must be initialized in constructor, because it is nested struct
>[...]/std/algorithm/iteration.d-mixin-2507(2524,25): Error:  field `_currentBack` initialization is not allowed in loops or after labels
>[...]/std/algorithm/iteration.d-mixin-2507(2521,21): Error: one path skips field `_currentBack`
>[...]/ggplotd/stat.d(583,11): Error: template instance `std.algorithm.iteration.joiner!(MapResult!(__lambda5, ZipShortest!(cast(Flag)true, double[], double[])))` error instantiating

The error comes from [line 583 of ggplotd.stat](https://github.com/BlackEdder/ggplotd/blob/d4428c08db5ffdc05dfd29690bf7da9073ea1dc5/source/ggplotd/stat.d#L562-L583):

```d
        // Use statFunction with the kernel to produce a line
        // Also add points to close the path down to zero
        auto stepX = (minmax[1] - minmax[0])/25.0;
        auto stepY = (minmax[3] - minmax[2])/25.0;
        auto xCoords = iota( minmax[0], minmax[1], stepX ).array ~ minmax[1];
        auto yCoords = iota( minmax[2], minmax[3], stepY ).array ~ minmax[3];
        auto coords = zip(xCoords, xCoords[1..$]).map!( (xr) {
                return zip(yCoords, yCoords[1..$]).map!( (yr) {
                        // Two polygons
                        return [
                        g.front.mergeRange(Aes!( double[], "x", double[], "y", double[], "colour" )
                        ([xr[0], xr[0], xr[1]],
                         [yr[0], yr[1], yr[1]],
                         [density(xr[0],yr[0]),density(xr[0],yr[1]),density(xr[1],yr[1])])),
                        g.front.mergeRange(
                                Aes!( double[], "x", double[], "y", double[], "colour" )
                        ([xr[0], xr[1], xr[1]],
                         [yr[0], yr[0], yr[1]],
                         [density(xr[0],yr[0]),density(xr[1],yr[0]),density(xr[1],yr[1])]))
                         ];
                }).joiner;
        }).joiner;
```

PR #5637, opened 2017-07-20
Show warnings about outdated Phobos modules

wilzbach:
As moving old & ugly Phobos modules to undeaD doesn't seem to be [an option](https://github.com/dlang/phobos/pull/5488) for now, let's at least try to warn the user.
I used the same message as we already display on std.xml.

Are there other modules that you would like to be included on this list?

Previous discussion: https://github.com/dlang/dlang.org/pull/1778

PR #6116, opened 2018-02-03
Issue 18298 - add note about frozen curl package

wilzbach:
I think this is the third time I have the __same__ issue on the issue tracker and I don't know how many times people have been having asking for a better http client on the forums.
This PR makes the standard answer: "No one is or will be working on that. However, try requests it's great, written entirely in D, compatible with Vibe.d eventloop for async requests and as fast as curl"

See also:
- https://forum.dlang.org/post/vjfrbipjdtyuesndlvcj@forum.dlang.org

The author of D's requests package shares a similar opinion as the Python requests developer have:

> Essentially, the standard library is where a library goes to die.
It is appropriate for a module to be included when active
development is no longer necessary.

https://github.com/requests/requests/issues/2424
http://docs.python-requests.org/en/latest/dev/philosophy/

With the web consistently moving (HTTP/2, newer SSL releases, etc.),
it's unlikely that the development will stop.

Hence, my proposal is to state the facts and to prevent people from bumping into the same issue over and over again.

Preview:

image

### DIFF
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -22,6 +22,10 @@ SMTP) )
 )
 )
 
+$(MESSAGE_BOX gray, The API of `std.net.curl` has been $(B frozen) for stability.
+For a high-level, pure D networking client, see $(LINK2 https://github.com/ikod/dlang-requests, requests).
+)
+
 Note:
 You may need to link to the $(B curl) library, e.g. by adding $(D "libs": ["curl"])
 to your $(B dub.json) file if you are using $(LINK2 http://code.dlang.org, DUB).


CyberShadow:
I don't understand. What makes std.net.curl different than any other Phobos module (except for std.experimental) in this regard?

I don't see the point in this change.

CyberShadow:
Instead of passing judgement, one thing we could do is add some links to some established Dub repositories, e.g. "Similar packages on code.dlang.org: (list packages)" or just a link to "Browse similar packages on code.dlang.org".

Ideally such lists would be automatically generated - filtered by topic and sorted by popularity, but doing it manually could be fine for now.

In any case, this is up to @andralex. I suspect he won't like the change in this PR either, but maybe we could discuss something along the lines of the above.

wilzbach:
> What makes std.net.curl different than any other Phobos module (except for std.experimental) in this regard?

No active development, but people expect a lot of functionality from it.

> I don't see the point in this change.

That I'm tired of explaining to people that std.net.curl isn't under development and that requests is to the go to alternative.

CyberShadow:
> No active development, but people expect a lot of functionality from it.

cURL itself is actively developed.

What's missing from std.net.curl? Shouldn't any new cURL functionality wouldn't be hard to wrap?

CyberShadow:
> No active development, but people expect a lot of functionality from it.

I don't understand how it's not any less under active development than the rest of Phobos. It receives periodic pull requests, like all Phobos modules, does it not?

JackStouffer:
I'm going to have to agree with @CyberShadow here and echo the comments made by Andrei and Johnathan here: https://github.com/dlang/phobos/pull/5637

If you feel strongly about this, please reopen.

PR #6117, opened 2018-02-03
Fix Issue 18229 - Misleading documentation of std.process.environment.get

wilzbach:
Imho it's an API flow if the behavior is different depending on the OS.
A standard library is supposed to abstract this for the user.

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -3341,7 +3341,7 @@ static:
     Retrieves the value of the environment variable with the given $(D name),
     or a default value if the variable doesn't exist.
 
-    Unlike $(LREF environment.opIndex), this function never throws.
+    Unlike $(LREF environment.opIndex), this function never throws on Posix.
     ---
     auto sh = environment.get("SHELL", "/bin/sh");
     ---
@@ -3357,6 +3357,13 @@ static:
         // empty.
     }
     ---
+    Params:
+        name = name of the environment variable to retrieve
+        defaultValue = default value to return if the environment variable doesn't exist.
+
+    Returns:
+        the value of the environment variable if found, otherwise
+        `null` if the environment doesn't exist.
 
     Throws:
     $(REF UTFException, std,utf) if the variable contains invalid UTF-16


wilzbach:
I added another commit to fix this issue once and for all and make `get` truly `nothrow` once and for all.
Use https://github.com/dlang/phobos/pull/6117/files?w=1 for a more pleasant diff.

std/process.d
```diff
@@ -3364,15 +3370,25 @@ static:
...
-        auto found = getImpl(name, value);
+        bool found;
+        version(Windows)
+        {
+            try
+            {
+                found = getImplWindows(name, value);
+            }
+            catch (WindowsException) {}
+            catch (UTFException) {}
```
CyberShadow: This isn't right. We shouldn't be treating UTF-malformed variables or errors from the Windows API as if the variable doesn't exist. Silently ignoring errors in general is almost never a good idea.

wilzbach: But having a function which is `nothrow` on Posix and throws on Windows is pretty ugly too.

schveiguy: Well, you could throw an Error instead.

schveiguy: Wait, it wasn't `nothrow` on posix, was it?

std/process.d
```diff
@@ -3357,15 +3363,29 @@ static:
...
+    string get(in char[] name, string defaultValue = null) @safe nothrow
     {
         string value;
-        auto found = getImpl(name, value);
+        bool found;
+        version(Windows)
+        {
+            try
+                found = getImplWindows(name, value);
+            catch (WindowsException) {}
```
schveiguy: Is this really enough to shut up the compiler? I mean, you're not handling all exceptions...

schveiguy: answer: it's not :)
https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3020178&isPull=true

std/process.d
schveiguy: Can we fix the documentation issue in a PR, and then work on the `nothrow` issue as a separate bugzilla enhancement/PR?

CyberShadow: I think throwing an `Error` would probably be fine. One could argue that this opens the window for a DoS attack by someone somehow sneaking bad UTF-16 into the environment (see Shellshock et al), but... this being Windows and all, I don't know if that's worth worrying about in this particular case.

CyberShadow: Also we probably want to throw Errors instead of UTFExceptions when auto-decoding bad UTF-8 anyway...

CyberShadow: Simplest and best solution, IMO: just don't make it `nothrow`. Why the heck does it need to be `nothrow` anyway? That the POSIX primitive never returns an error is really kind of a leaking implementation detail of how the environment is stored and encoded on whatever UNIX implementations POSIX was based off.

wilzbach:
OK, I dropped the second commit and now only the documentation commit is in.

> With the proposed change, a variable that contains invalid UTF16 will appear as nonexistent, which is wrong IMO.

Hmm, imho this isn't too unexpected, because a variable with invalid UTF16 chars can never exist.

kyllingstad:
> a variable with invalid UTF16 chars can never exist.

I think it can, e.g. if a garbage string is passed to [`SetEnvironmentVariable()`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686206(v=vs.85).aspx).

In hindsight, it seems to me that `std.process.environment` should have dealt in `ubyte[]` rather than `string`, to avoid making assumptions about the encoding. I am not 100% sure about Windows, but POSIX, at least, does not specify any particular encoding. In other words, an environment variable on POSIX could contain ASCII, UTF-8, UTF-16, binary data – anything other than a null character.

PR #6121, opened 2018-02-03
Update win32 libcurl link

0xEAB:
The real download can be found in the directory `other` of the *Release Archive*. It wasn't really straight forward to just link the general download page. A link to the archive can be found on the dl page, though.

(Usually people don't like to play hide and seek while coding 🙈)

> @0xNiklas on Discord:
>> well there is no download ^^
>> or i am blind

### DIFF
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -28,7 +28,7 @@ to your $(B dub.json) file if you are using $(LINK2 http://code.dlang.org, DUB).
 
 Windows x86 note:
 A DMD compatible libcurl static library can be downloaded from the dlang.org
-$(LINK2 http://dlang.org/download.html, download page).
+$(LINK2 http://downloads.dlang.org/other/index.html, download archive page).
 
 Compared to using libcurl directly this module allows simpler client code for
 common uses, requires no unsafe operations, and integrates better with the rest


wilzbach:
FYI: We are about to switch the curl build to AppVeyor:

https://github.com/dlang/installer/pull/288
https://ci.appveyor.com/project/greenify/installer/build/1.0.86/artifacts

---------------------

Also why is this note still there - I thought that curl is shipped with dmd?
I don't use Windows, but here's what I see:

```
als dmd.2.078.1.windows.7z | grep curl
2018-01-21 14:45:54 ....A         9216               dmd2/windows/lib/curl.lib
2018-01-21 14:45:54 ....A        14074               dmd2/windows/lib64/curl.lib
2018-01-21 14:45:54 ....A       451072               dmd2/windows/bin/libcurl.dll
2018-01-21 14:45:54 ....A       424960               dmd2/windows/bin64/libcurl.dll
```

CyberShadow:
The description of the download link says it's for a static library. Is that right? I think we've been using a dynamic library for a while with the cURL shipped with DMD, so a static alternative wouldn't be bad.

0xNiklas:
It's a static library

CyberShadow:
Thanks.

PR #6122, opened 2018-02-04
std.net.curl: Move asynchronous stuff to lazily-instantiated namespace

CyberShadow:
Allows avoiding importing std.concurrency unless asynchronous functionality actually needed.
Improves impact on import time by about 15%.

https://github.com/dlang/phobos/pull/5916#issuecomment-362900099

### DIFF
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -159,7 +159,6 @@ module std.net.curl;
 public import etc.c.curl : CurlOption;
 import core.time : dur;
 import etc.c.curl : CURLcode;
-import std.concurrency : Tid;
 import std.range.primitives;
 import std.encoding : EncodingScheme;
 import std.traits : isSomeChar;
@@ -171,6 +170,8 @@ version(StdUnittest)
 
     private struct TestServer
     {
+        import std.concurrency : Tid;
+
         import std.socket : Socket, TcpSocket;
 
         string addr() { return _addr; }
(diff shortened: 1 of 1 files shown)


std/net/curl.d
```diff
@@ -1612,7 +1584,7 @@ private static struct AsyncLineInputRange(Char)
  *
  * If no data is available and the main thread accesses the range it will block
  * until data becomes available. An exception to this is the $(D wait(Duration)) method on
- * the $(LREF AsyncLineInputRange). This method will wait at maximum for the
+ * the $(LREF LineInputRange). This method will wait at maximum for the
```
wilzbach: :+1:

std/net/curl.d
```diff
@@ -4883,285 +4778,397 @@ private struct Pool(Data)
...
-                                          ref bool aborted)
+// Lazily-instantiated namespace to avoid importing std.concurrency until needed.
+private struct _async()
 {
-    import std.concurrency : receive, send, thisTid;
+static:
+    // @@@@BUG 15831@@@@
+    // this should be inside byLineAsync
+    // Range that reads one chunk at a time asynchronously.
+    private static struct ChunkInputRange
```
wilzbach: `static` is redundant here.

CyberShadow: Err actually there were two more non-templated functions.

PR #6125, opened 2018-02-04
Fix #9584 - Using dirEntries and chdir() can have unwanted results.

### LINKED ISSUE #9584: Using dirEntries and chdir() can have unwanted results
### CC List

* andrei (@andralex)
* francesco.galla3
* greensunny12
* hsteoh
* razvan.nitu1305
* timothee.cour2

### Description

````markdown
I can't officially classify this as a bug. Jonathan said it might be a good idea to put it here just so it doesn't get lost, and then we can later decide if it's a bug or if we should just be careful what we do. 

Here's some buggy code which might throw an exception (if you're lucky!):
   foreach (string entry; dirEntries(curdir, SpanMode.shallow))
   {
       if (entry.isdir)
       {
           foreach (string subentry; dirEntries(entry, SpanMode.shallow))
           {
               if (subentry.isdir)
               {
                   chdir(rel2abs(subentry));
               }
           }
       }
   }

This might throw something like this:
std.file.FileException@std\file.d(1124): .\foo\bar.d: The system cannot find the path specified.

The problem is chdir will modify curdir, and dirEntries doesn't store 'curdir' internally as an absolute address, so calling chdir() screws up the behavior of dirEntries.

A workaround is to call rel2abs as soon as possible when using 'curdir' (the first line changed here):

   foreach (string entry; dirEntries(rel2abs(curdir), SpanMode.shallow))
   {
       if (entry.isdir)
       {
           foreach (string subentry; dirEntries(entry, SpanMode.shallow))
           {
               if (subentry.isdir)
               {
                   chdir(rel2abs(subentry));
               }
           }
       }
   }

I've had a use for code like this where I was invoking a script file which was always relative to some subdirectory.
````

### PR DESCRIPTION
**Pull request by GallaFrancesco**
After following issue 6138 on the bug tracker [*edit:* #9584 on GitHub], I am proposing this fix for the `std.file` module.

All it does is calling `absolutePath` from `std.path` before passing the argument to `dirEntries`.
The overload of dirEntries which also accepts string patterns doesn't seem to be affected by this, therefore no changes were proposed.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -4138,11 +4138,13 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
     dirEntries("", SpanMode.shallow).walkLength();
 
     // issue 6138
-    foreach (string entry; dirEntries(testdir, SpanMode.shallow) {
+    string cwd = getcwd();
+    foreach (string entry; dirEntries(testdir, SpanMode.shallow)) {
         if (entry.isDir) {
              chdir(entry);
         }
     }
+    chdir (cwd); // needed for the directories to be removed
 }
 
 /// Ditto
--- a/std/file.d
+++ b/std/file.d
@@ -4038,7 +4038,7 @@ foreach (d; dFiles)
  +/
 auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
 {
-	import std.path: absolutePath;
+    import std.path: absolutePath;
     return DirIterator(absolutePath(path), mode, followSymlink);
 }
 
@@ -4058,7 +4058,7 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
             .array;
     }
 
-	void main(string[] args)
+    void main(string[] args)
     {
         import std.stdio;
 
--- a/std/file.d
+++ b/std/file.d
@@ -4038,7 +4038,8 @@ foreach (d; dFiles)
  +/
 auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
 {
-    return DirIterator(path, mode, followSymlink);
+	import std.path: absolutePath;
+    return DirIterator(absolutePath(path), mode, followSymlink);
 }
 
 /// Duplicate functionality of D1's $(D std.file.listdir()):
@@ -4057,7 +4058,7 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
             .array;
     }
 
-    void main(string[] args)
+	void main(string[] args)
     {
         import std.stdio;
 
@@ -4109,7 +4110,7 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
     foreach (string name; dirEntries(testdir, SpanMode.breadth))
     {
         //writeln(name);
-        assert(name.startsWith(testdir));
+        assert(name.startsWith(absolutePath(testdir)));
     }
     foreach (DirEntry e; dirEntries(absolutePath(testdir), SpanMode.breadth))
     {
--- a/std/file.d
+++ b/std/file.d
@@ -4139,8 +4139,10 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
 
     // issue 6138
     string cwd = getcwd();
-    foreach (string entry; dirEntries(testdir, SpanMode.shallow)) {
-        if (entry.isDir) {
+    foreach (string entry; dirEntries(testdir, SpanMode.shallow))
+    {
+        if (entry.isDir)
+        {
              chdir(entry);
         }
     }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: std/file.d [hunk 1]
```diff
@@ -4038,7 +4038,8 @@ foreach (d; dFiles)
  +/
 auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
 {
-    return DirIterator(path, mode, followSymlink);
+	import std.path: absolutePath;
+    return DirIterator(absolutePath(path), mode, followSymlink);
```
wilzbach: No tabs please (see https://dlang.org/dstyle.html)

std/file.d
```diff
@@ -4057,7 +4058,7 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
             .array;
     }
 
-    void main(string[] args)
+	void main(string[] args)
```
wilzbach: tab.

GallaFrancesco:
I've updated the source code removing tabs.

std/file.d
```diff
@@ -4109,7 +4110,7 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
     foreach (string name; dirEntries(testdir, SpanMode.breadth))
     {
         //writeln(name);
-        assert(name.startsWith(testdir));
+        assert(name.startsWith(absolutePath(testdir)));
```
CyberShadow: Do not break existing tests!

Changing existing unit tests to conform to changes in the implementation is almost always a mistake.

CyberShadow:
@GallaFrancesco Have a look at the project tester (Jenkins) - this change actually breaks existing D code out there.

GallaFrancesco:
@CyberShadow thank you for your reply.

I will look into the approach you suggested (`fdopendir` + `openat`). I wonder how this could apply to non-POSIX environments though, given that I am mainly familiar with POSIX ones. Any suggestions? Could this be solved on POSIX systems only?

CyberShadow:
Windows is the only non-POSIX system we care about.

It doesn't look like Windows has native equivalents for `fdopendir` or `openat`. Here is the approach that Gnulib uses:

- When beginning iteration (i.e. when `dirEntries` is called), save the current working directory.
- Restore the working directory before every `FindFileFirst` call, but only for the duration of the `FindFileFirst` call.

However, I can't recommend the above because on Windows, the current directory is a process-wide value, so it could interfere with other threads (which are presumably aware that the user changed the current directory after calling `dirEntries`).

Another approach would be similar to your current implementation here, but would additionally remove the absolute path prefix in returned DirEntries, so as to not break existing code.

I think we should do the `fdopendir` thing on POSIX anyway, since it's more pragmatic and likely to improve performance, I'm not sure if doing the above is worthwhile on Windows... because, returning any sort of relative paths when the current directory changes is just not useful. If the path is relative to the initial working directory, then it will point to a non-existing file (or an existing file but not the one you want). If the path is relative to the current working directory, then recording it anywhere is likely to be meaningless.

[Your example on Bugzilla](https://issues.dlang.org/show_bug.cgi?id=6138#c5) also doesn't make sense - you're calling `absolutePath` twice in a row with a `chdir` in between, but that is not going to work right (because `absolutePath` uses the current directory as the base) unless the path in question was already absolute, in which case the `absolutePath` call is also meaningless.

So... maybe improve the POSIX implementation, because that won't hurt and could bring other benefits, but as for fixing the Bugzilla issue as it was stated... that strikes me more of a kind of "shooting yourself in the foot" situation. If you expect to change directories during directory iteration, just pass an absolute path to `dirEntries`.

I should add that as far as I can see, most D programs have no business at all in setting the current directory. For process creation, `std.process` functions can take a `workDir` parameter that sets directories' initial working directories. For most other things, just use `std.path.buildPath` & co. If your program has a "home directory" where it keeps all its work/resource files, you might be tempted to change into it when it starts, but you will later find out that this is not such a great idea when you attempt to process relative paths passed by users on your program's command line.

GallaFrancesco:
I agree with you, the situations in which one needs to `chdir` inside a program are very few and errors such as this one or the one recently reported in the [bug tracker](https://issues.dlang.org/show_bug.cgi?id=6138) can be solved by passing `absolutePath` to `dirEntries`. 

Nevertheless, it seems to me that this behavior is poorly documented, which is the main reason I opened this PR in the first place.

The POSIX solution looks feasible though, the only issue I encountered is the absence of the `fdopendir` and `openat` declarations from [DRuntime](https://github.com/dlang/druntime/), where `fdopendir` is supposed to be in module [dirent](https://github.com/dlang/druntime/blob/master/src/core/sys/posix/dirent.d) and `openat` in module [fcntl](https://github.com/dlang/druntime/blob/master/src/core/sys/posix/fcntl.d).

Would this require a PR to DRuntime? I'm not sure it is worth the minimal benefits brought by these changes.

CyberShadow:
> Would this require a PR to DRuntime?

Improving our coverage of OS APIs is a worthwhile goal in itself.

LightBender:
@thewilsonator wants to take a look at this one.

0xEAB:
This looks like a dead end:
image

Revival → #10666

PR #6128, opened 2018-02-05
Fix Issue 18381 - std.random.uniform01!(double,Mt19937) should produce at least 2^^53 distinct values

n8sh:
Do this by popping off the front two uints from the PRNG and combining them. Also make same change to std.random.uniform.

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -1473,10 +1473,42 @@ if (isFloatingPoint!(CommonType!(T1, T2)) && isUniformRNG!UniformRandomNumberGen
     enforce(_a <= _b,
             text("std.random.uniform(): invalid bounding interval ",
                     boundaries[0], a, ", ", b, boundaries[1]));
-    NumberType result =
-        _a + (_b - _a) * cast(NumberType) (urng.front - urng.min)
-        / (urng.max - urng.min);
-    urng.popFront();
+
+    alias R = typeof(urng.front);
+    // Special case: pop off the front two items to get enough
+    // bits for a double or a real when typeof(rng.front) is uint.
+    // (Generalized but in practice it's only uint.)
+    enum useFrontTwoItems = isIntegral!R
+        && NumberType.mant_dig > R.sizeof * 8
+        && R.sizeof <= uint.sizeof
+        && (hasIndirections!(typeof(urng)) || typeof(urng).sizeof >= R.sizeof * 2
+            || !__traits(compiles, () pure { return urng.front; }()))
+        && (urng.min == R.min || urng.min == R.min + 1)
+        && urng.max == R.max;
+
+    static if (useFrontTwoItems)
+    {
+        import std.algorithm.comparison : min;
+        import std.meta : AliasSeq;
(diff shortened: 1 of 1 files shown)


n8sh:
The eBay/tsv-utils-dlang test is [failing spuriously](https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-6128/2/pipeline/220) because it uses `uniform01` with a known seed for weighted sampling and checks if it matches a predefined result. That test would also fail if `std.random.Random` were changed to another PRNG.

wilzbach:
Yes, no test suite should rely on the default seed of std.random and always set its seed and PRNG explicitly.
If they want to do additional random fuzzing, they can of course do so, but not on the project tester.

Do you have time to submit a quick PR to the tsv-utils? Our policy is that if that one isn't merged within three days, the project will/can be temporarily disabled from dlang/ci to avoid blocking a PR.

n8sh:
>Do you have time to submit a quick PR to the tsv-utils?

What should it entail? Changing the expected results to match the new `uniform01` wouldn't work since the test would fail with the current standard library, and I can't imagine the maintainers would want to remove the tests outright.

n8sh:
>The unit tests assume that the result of `uniform01` is stable across releases. It's also a feature provided to users by `tsv-sample`. The caller gets consistent results run-to-run if the choose a stable seed (the built-in 2438424139), or provide their own seed.

That's an important consideration. Would copying and pasting a legacy `uniform01` into the project be an acceptable fix?

jondegenhardt:
> That's an important consideration. Would copying and pasting a legacy uniform01 into the project be an acceptable fix?

Do you mean copying the code for `uniform01` into tsv-utils?

n8sh:
>Do you mean copying the code for `uniform01` into tsv-utils?

That's what I meant, but continuing to think about this, even if that were acceptable there could be any number of projects in similar situations. If there is already-deployed code that is counting on the algorithms in `std.random` never changing (which is not an unreasonable thing to assume) maybe we'll just have to swallow the bitter pill that things we didn't get 100% right the first time can't be changed after being released into the wild.

jondegenhardt:
Okay, thanks for considering this case. For tsv-utils, I'd prefer to create a breaking change rather than copy a legacy piece of code. But, that's because the utilities are relatively young, not that many users, there's a good chance the number of people counting on this feature is zero.

But, I think your larger point is the more important one. Allowing apps/users to specify a static seed and get a consistent set of "random" values is relatively common, there could easily be tools other than mine expecting this behavior.

jondegenhardt:
I haven't looked at the code in question, but would a way forward be to version the random generator in some way?

yebblies:
I've used `std.random` for procedural generation assuming that since `Mt19937` is deterministic for a given seed the results from `uniform*` would be stable.  This needs to be weighed against the distribution quality, since reading twice from the rng will completely change any results.  `uniform` probably needs to document if the number of reads and generation method are stable or not.

n8sh:
Closing this pull request due to the problems that it would cause.

PR #6129, opened 2018-02-06
Fix Issue 18384 - std.net.isemail is slow to import due to regex

n8sh:
Solution is to remove regex from std.net.isemail. May be worth revisiting if std.regex compile times improve. On my machine making the change decreased the time measured by [@wilzbach's import benchmark script](https://gist.github.com/wilzbach/afb9f2edec8ddfabfdd6becec55e62a7) from 0.15 seconds to 0.03 seconds (consistently over repeated trials).

[Link to mention of this issue on the forums.](https://forum.dlang.org/post/qltouuhsxsqltpspuhye@forum.dlang.org)

### DIFF
--- a/std/net/isemail.d
+++ b/std/net/isemail.d
@@ -27,7 +27,6 @@ module std.net.isemail;
 
 // FIXME
 import std.range.primitives; // : ElementType;
-import std.regex;
 import std.traits;
 import std.typecons : Flag, Yes, No;
 
@@ -73,10 +72,6 @@ if (isSomeChar!(Char))
 
     alias tstring = const(Char)[];
     alias Token = TokenImpl!(Char);
-
-    static ipRegex = ctRegex!(`\b(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}`~
-                        `(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$`.to!(const(Char)[]));
-    static fourChars = ctRegex!(`^[0-9A-Fa-f]{0,4}$`.to!(const(Char)[]));
 
     enum defaultThreshold = 16;
     int threshold;
@@ -399,12 +394,11 @@ if (isSomeChar!(Char))
                             auto maxGroups = 8;
                             size_t index = -1;
                             auto addressLiteral = parseData[EmailPart.componentLiteral];
-                            auto matchesIp = addressLiteral.matchAll(ipRegex).map!(a => a.hit).array;
+                            const(Char)[] ipSuffix = matchIPSuffix(addressLiteral);
 
-                            if (!matchesIp.empty)
+                            if (ipSuffix.length)
                             {
-                                index = addressLiteral.lastIndexOf(matchesIp.front);
-
+                                index = addressLiteral.length - ipSuffix.length;
                                 if (index != 0)
                                     addressLiteral = addressLiteral[0 .. index] ~ "0:0";
                             }
@@ -418,7 +412,7 @@ if (isSomeChar!(Char))
                             else
                             {
                                 auto ipV6 = addressLiteral[5 .. $];
-                                matchesIp = ipV6.split(Token.colon);
+                                auto matchesIp = ipV6.split(Token.colon);
                                 immutable groupCount = matchesIp.length;
                                 index = ipV6.indexOf(Token.doubleColon);
 
@@ -453,7 +447,7 @@ if (isSomeChar!(Char))
                                     returnStatus ~= EmailStatusCode.rfc5322IpV6ColonEnd;
 
                                 else if (!matchesIp
-                                        .filter!(a => a.matchFirst(fourChars).empty)
+                                        .filter!(a => !isUpToFourHexChars(a))
                                         .empty)
                                     returnStatus ~= EmailStatusCode.rfc5322IpV6BadChar;
 
@@ -1862,4 +1856,97 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
 {
     assert("abc".get(1, 'b') == "b");
     assert("löv".get(1, 'ö') == "ö");
+}
+
+/+
+Replacement for:
+---
+static fourChars = ctRegex!(`^[0-9A-Fa-f]{0,4}$`.to!(const(Char)[]));
+...
+a => a.matchFirst(fourChars).empty
+---
++/
+private bool isUpToFourHexChars(Char)(scope const(Char)[] s)
+{
+    import std.ascii : isHexDigit;
+    if (s.length > 4) return false;
+    foreach (c; s)
+        if (!isHexDigit(c)) return false;
+    return true;
+}
+
+version(StdUnittest) @nogc nothrow pure @safe unittest
+{
+    assert(!isUpToFourHexChars("12345"));
+    assert(!isUpToFourHexChars("defg"));
+    assert(isUpToFourHexChars("1A0a"));
+}
+
+/+
+Replacement for:
+---
+static ipRegex = ctRegex!(`\b(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}`~
+                    `(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$`.to!(const(Char)[]));
+...
+auto matchesIp = addressLiteral.matchAll(ipRegex).map!(a => a.hit).array;
+----
+Note that only the first item of "matchAll" was ever used in practice
+so we can return `const(Char)[]` instead of `const(Char)[][]` using a
+zero-length string to indicate no match.
++/
+private const(Char)[] matchIPSuffix(Char)(return const(Char)[] s) @nogc nothrow pure @safe
+{
+    size_t end = s.length;
+    if (end < 7) return null;
+    // Check the first three `[.]\d{1,3}`
+    foreach (_; 0 .. 3)
+    {
+        size_t start = void;
+        if (end >= 2 && s[end-2] == '.')
+            start = end - 2;
+        else if (end >= 3 && s[end-3] == '.')
+            start = end - 3;
+        else if (end >= 4 && s[end-4] == '.')
+            start = end - 4;
+        else
+            return null;
+        uint x = 0;
+        foreach (i; start + 1 .. end)
+        {
+            uint c = cast(uint) s[i] - '0';
+            if (c > 9) return null;
+            x = x * 10 + c;
+        }
+        if (x > 255) return null;
+        end = start;
+    }
+    // Check the final `\d{1,3}`.
+    if (end < 1) return null;
+    size_t start = end - 1;
+    uint x = cast(uint) s[start] - '0';
+    if (x > 9) return null;
+    if (start > 0 && cast(uint) s[start-1] - '0' <= 9)
+    {
+        --start;
+        x += 10 * (cast(uint) s[start] - '0');
+        if (start > 0 && cast(uint) s[start-1] - '0' <= 9)
+        {
+            --start;
+            x += 100 * (cast(uint) s[start] - '0');
+        }
+    }
+    if (x > 255) return null;
+    // Must either be at start of string or preceded by a non-word character.
+    // (TO DETERMINE: is the definition of "word character" ASCII only?)
+    if (start == 0) return s;
+    const b = s[start - 1];
+    import std.ascii : isAlphaNum;
+    if (isAlphaNum(b) || b == '_') return null;
+    return s[start .. $];
+}
+
+version(StdUnittest) @nogc nothrow pure @safe unittest
+{
+    assert(matchIPSuffix("255.255.255.255") == "255.255.255.255");
+    assert(matchIPSuffix("babaev 176.16.0.1") == "176.16.0.1");
 }


std/net/isemail.d
```diff
@@ -1863,3 +1857,95 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
...
+    if (s.length < "0.0.0.0".length) return null;
+    size_t end = s.length;
+    // Check the first three `[.]\d{1,3}`
+    foreach (_; 0 .. 3)
+    {
+        size_t start = void;
+        if (end >= 2 && s[end-2] == '.')
+            start = end - 2;
+        else if (end >= 3 && s[end-3] == '.')
+            start = end - 3;
```
schveiguy: This line is untested, try an IP address with 2 digits in an octet.

std/net/isemail.d
```diff
@@ -1863,3 +1857,95 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
...
+...
+a => a.matchFirst(fourChars).empty)
+---
++/
+private static bool isUpToFourHexChars(Char)(scope const(Char)[] s)
+{
+    import std.ascii : isHexDigit;
+    if (s.length > 4) return false;
+    foreach (i; 0 .. s.length)
+        if (!isHexDigit(s[i])) return false;
```
schveiguy:
```D
foreach (c; s)
    if (!isHexDigit(c)) return false;
```

std/net/isemail.d
n8sh: Fixed

std/net/isemail.d
n8sh: Although that is nicer syntax, it would auto-decode which is not necessary (all characters I'm searching for are <= 0x7F) and would be a performance hit.

aG0aep6G: `foreach (c; s)` doesn't do auto-decoding.

quickfur: It does if `c` is `dchar` and `s` is a narrow string.

n8sh: My mistake: I thought `c` would default to `dchar` unless given a specific type. Changed to use `foreach`.

wilzbach: Are you sure? https://run.dlang.io/is/UGY8jP

schveiguy: > I thought c would default to dchar unless given a specific type.

This is a common misconception -- because Phobos treats arrays of `char` weirdly. The language doesn't, it's still an array to the language.

std/net/isemail.d
```diff
@@ -1863,3 +1857,95 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
...
+    {
+        size_t start = void;
+        if (end >= 2 && s[end-2] == '.')
+            start = end - 2;
+        else if (end >= 3 && s[end-3] == '.')
+            start = end - 3;
+        else if (end >= 4 && s[end-4] == '.')
+            start = end - 4;
+        else
+            return null;
```
wilzbach:
```d
start = end - s.byCodeUnit.retro.take(4).filter!(a == '.').count;
```

std/net/isemail.d
```diff
@@ -1863,3 +1857,95 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
...
+        uint x = 0;
+        foreach (i; start + 1 .. end)
+        {
+            uint c = cast(uint) s[i] - '0';
+            if (c > 9) return null;
+            x = x * 10 + c;
+        }
```
wilzbach:
```d
uint x = s[start + 1 .. end].byCodeUnit.take(2).to!uint.ifThrown(256);
```

std/net/isemail.d
```diff
@@ -1863,3 +1857,95 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
...
+    if (start > 0 && cast(uint) s[start-1] - '0' <= 9)
+    {
+        --start;
+        x += 10 * (cast(uint) s[start] - '0');
+        if (start > 0 && cast(uint) s[start-1] - '0' <= 9)
+        {
+            --start;
+            x += 100 * (cast(uint) s[start] - '0');
+        }
+    }
```
wilzbach:
```d
s[max(0, end - 3) .. $].byCodeUnit.to!uint.ifThrown(256);
```

std/net/isemail.d
```diff
@@ -1863,3 +1857,95 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
...
+...
+auto matchesIp = addressLiteral.matchAll(ipRegex).map!(a => a.hit).array;
+----
+Note that only the first item of "matchAll" was ever used in practice
+so we can return `const(Char)[]` instead of `const(Char)[][]` using a
+zero-length string to indicate no match.
++/
+private static const(Char)[] matchIPSuffix(Char)(return const Char[] s)
+{
+    if (s.length < "0.0.0.0".length) return null;
```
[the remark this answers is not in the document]
n8sh: Fixed

std/net/isemail.d
```diff
@@ -1863,3 +1857,95 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
...
+            --start;
+            x += 100 * (cast(uint) s[start] - '0');
+        }
+    }
+    if (x > 255) return null;
+    // Must either be at start of string or preceded by a non-word character.
+    // (TO DETERMINE: is the definition of "word character" ASCII only?)
+    if (start == 0) return s;
+    const b = cast(uint) s[start - 1];
+    if (b - 'A' < 26 || b - 'a' < 26 || b - '0' < 10 || b == '_') return null;
```
wilzbach: Use `isAlphaNum` from [`std.ascii`](https://dlang.org/phobos/std_ascii.html)

std/net/isemail.d
```diff
@@ -1863,3 +1857,95 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
...
+static ipRegex = ctRegex!(`\b(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}`~
+                    `(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$`.to!(const(Char)[]));
+...
+auto matchesIp = addressLiteral.matchAll(ipRegex).map!(a => a.hit).array;
+----
+Note that only the first item of "matchAll" was ever used in practice
+so we can return `const(Char)[]` instead of `const(Char)[][]` using a
+zero-length string to indicate no match.
++/
+private static const(Char)[] matchIPSuffix(Char)(return const Char[] s)
```
wilzbach: :+1: for the use of `return`

std/net/isemail.d
```diff
@@ -1863,3 +1857,95 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
...
+{
+    if (s.length < 7) return null;
+    size_t end = s.length;
```
wilzbach: Nit: you could now move `end` one line above and do `if (end < 7) ...`

std/net/isemail.d
n8sh: not `@nogc` and not `nothrow`

std/net/isemail.d
wilzbach: You can do even one better:

```d
int blocks = s.byCodeUnit
.splitter(".")
.filter!checkValid // check an individual block
.take(4) // avoids problems with infinite strings
.walkLength;
if (blocks < 4) return null;
```

std/net/isemail.d
n8sh: Fixed

std/net/isemail.d
n8sh: sure

std/net/isemail.d
wilzbach: FYI: `@nogc` isn't a requirement for Phobos (we prefer nice, readable code instead of mir). In this case it will be fixed once the already merged `-dip1008` gets activated by default.
Also the previous `matchAll` isn't neither `@nogc` nor `nothrow` either (and neither is the code which uses it).

Regarding `nothrow`, how about `scope(failure) return null;`?
That would save even more lines.

std/net/isemail.d
n8sh: Don't think that does the right thing. The blocks need to be consecutive and they need to be the tail of the string, and we need the total length of the tail.

std/net/isemail.d
JackStouffer: While making things `@nogc nothrow` is nice when ever possible, in this specific instance `isEmail` is not `@nogc nothrow`, so this change will have no effect on the user and there's no benefit.

schveiguy: What about converting to `ubyte`? then let `to` do the work of making sure the octet is in range.

FWIW, I like the idea of tagging internal functions `@nogc` and `nothrow`, even if the users aren't, because it helps keep GC usage down. Otherwise, little fixes creep in here and there.

std/net/isemail.d
n8sh: Can't use this here because unlike above I don't know where the encoded octet starts.

schveiguy:
I know it's tempting here to find the "best way" to parse an ip address. When I first saw the function, I spent about 15 minutes writing an alternative one as a suggestion that looked more at the text pattern rather than creating a uint out of the numbers. After a while, I just deleted it, since it's really not insanely complex either way, and the end result isn't different enough to warrant changes to this PR.

I think probably we can pull this as-is, and improve the internal details later. I'd like to get the regex problem solved.

std/net/isemail.d
klickverbot: `static` is pointless, though.

WalterBright: add `pure nothrow @safe @nogc`

luismarques:
Can't you just manually include the D code generated and mixin()ed by the ctRegex? Leave the ctRegex version(none)ed out to indicate where the code comes from, and to reenable it once the performance situation changes.

std/net/isemail.d
```diff
@@ -1863,3 +1857,96 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
...
+        else if (end >= 4 && s[end-4] == '.')
+            start = end - 4;
+        else
+            return null;
+        uint x = 0;
+        foreach (i; start + 1 .. end)
+        {
+            uint c = cast(uint) s[i] - '0';
+            if (c > 9) return null;
+            x = x * 10 + c;
```
WalterBright: need to check for integer overflow

n8sh: Can't overflow, we're summing a maximum of 3 digits.

std/net/isemail.d
n8sh: Fixed

JackStouffer:
>Can't you just manually include the D code generated and mixin()ed by the ctRegex? Leave the ctRegex version(none)ed out to indicate where the code comes from, and to reenable it once the performance situation changes.

Yes, this makes sense. @n8sh There's a debug flag in std.regex which will print out the code. Can you see which one is faster?

n8sh:
@luismarques @JackStouffer I've run a comparison and the new code is dramatically faster.

Time to extract either an empty string or an IP address suffix from "addressLiteral" in a list of 32 strings, repeated 10_000 times:

|                   | dmd -O     | ldc2 -O2  |
|-------------------|------------|-----------|
| **ctRegex**       | 1667 msecs | 451 msecs |
| **matchIPSuffix** |    3 msecs |   2 msecs |

The "addressLiteral" strings used for the benchmark were captured from the current std.netisemail unittests plus the unittest added by this pull request. I can post the benchmark code if you suggest a convenient place for it (it's a bit long for this comment).

n8sh:
Note that the way the ctRegex is used in the current code -- which is how I benchmarked it -- includes dynamic memory allocation. It is effectively:

```d
auto matchesIp = addressLiteral.matchAll(ipRegex).map!(a => a.hit).array;
string ipSuffix = matchesIp.empty ? null : matchesIp.front;
```

JackStouffer:
@n8sh You can post it here via the following

    <details>
    ```d
    code here
    ```
    </details>

n8sh:
@JackStouffer Thanks.
<details>
<summary><b>Benchmark source code</b></summary>

```d

// Strings taken from current unittests.
immutable string[32] address_literal_dataset =
[
    `IPv6:1111:2222:3333:4444:5555:6666::8888`,
    `255.255.255.255`,
    `255.255.255`,
    `255.255.255.255.255`,
    `255.255.255.256`,
    `1111:2222:3333:4444:5555:6666:7777:8888`,
    `IPv6:1111:2222:3333:4444:5555:6666:7777`,
    `IPv6:1111:2222:3333:4444:5555:6666:7777:8888`,
    `IPv6:1111:2222:3333:4444:5555:6666:7777:8888:9999`,
    `IPv6:1111:2222:3333:4444:5555:6666:7777:888G`,
    `IPv6:1111:2222:3333:4444:5555:6666::8888`,
    `IPv6:1111:2222:3333:4444:5555::8888`,
    `IPv6:1111:2222:3333:4444:5555:6666::7777:8888`,
    `IPv6::3333:4444:5555:6666:7777:8888`,
    `IPv6:::3333:4444:5555:6666:7777:8888`,
    `IPv6:1111::4444:5555::8888`,
    `IPv6:::`,
    `IPv6:1111:2222:3333:4444:5555:255.255.255.255`,
    `IPv6:1111:2222:3333:4444:5555:6666:255.255.255.255`,
    `IPv6:1111:2222:3333:4444:5555:6666:7777:255.255.255.255`,
    `IPv6:1111:2222:3333:4444::255.255.255.255`,
    `IPv6:1111:2222:3333:4444:5555:6666::255.255.255.255`,
    `IPv6:1111:2222:3333:4444:::255.255.255.255`,
    `IPv6::255.255.255.255`,
    `255.255.255.255`,
    `RFC-5322-domain-literal`,
    `RFC-5322`,
    `RFC5322domainliteral`,
    `RFC-5322-domain-literal`,
    `IPv6:1::2:`,
    `255.255.255.255`,
    `babaev 176.16.0.1`
];

void main(string[] args)
{
    import std.stdio : writeln;
    import std.datetime.stopwatch : AutoStart, StopWatch;
    enum iterations = 10_000;
    auto test_data = address_literal_dataset[];
    size_t output;
    StopWatch sw = StopWatch(AutoStart.no);
    
    // Method 1

    sw.reset();
    output = 0;
    sw.start();
    foreach_reverse (_; 0.. iterations)
    {
        foreach (address_literal; test_data)
            output += method1(address_literal).length;
    }
    sw.stop();
    writeln("method1 x", iterations, " = ", sw.peek().total!"msecs", " msecs [checksum ", output, "]");

    sw.reset();
    output = 0;
    sw.start();
    foreach_reverse (_; 0.. iterations)
    {
        foreach (address_literal; test_data)
            output += method1(address_literal).length;
    }
    sw.stop();
    writeln("method1 x", iterations, " = ", sw.peek().total!"msecs", " msecs [checksum ", output, "]");

    // Method 2

    sw.reset();
    output = 0;
    sw.start();
    foreach_reverse (_; 0.. iterations)
    {
        foreach (address_literal; test_data)
            output += method2(address_literal).length;
    }
    sw.stop();
    writeln("method2 x", iterations, " = ", sw.peek().total!"msecs", " msecs [checksum ", output, "]");

    sw.reset();
    output = 0;
    sw.start();
    foreach_reverse (_; 0.. iterations)
    {
        foreach (address_literal; test_data)
            output += method2(address_literal).length;
    }
    sw.stop();
    writeln("method2 x", iterations, " = ", sw.peek().total!"msecs", " msecs [checksum ", output, "]");

    // Method 1 again

    sw.reset();
    output = 0;
    sw.start();
    foreach_reverse (_; 0.. iterations)
    {
        foreach (address_literal; test_data)
            output += method1(address_literal).length;
    }
    sw.stop();
    writeln("method1 x", iterations, " = ", sw.peek().total!"msecs", " msecs [checksum ", output, "]");

    sw.reset();
    output = 0;
    sw.start();
    foreach_reverse (_; 0.. iterations)
    {
        foreach (address_literal; test_data)
            output += method1(address_literal).length;
    }
    sw.stop();
    writeln("method1 x", iterations, " = ", sw.peek().total!"msecs", " msecs [checksum ", output, "]");
}

pragma(inline, false)
const(Char)[] method1(Char)(const(Char)[] s)
{
    import std.algorithm.iteration : map;
    import std.array : array, split;
    import std.range.primitives : empty, front;
    import std.regex : ctRegex, matchAll;
    static ipRegex = ctRegex!(`\b(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}`~
                            `(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$`);
    auto matchesIp = s.matchAll(ipRegex).map!(a => a.hit).array;
    return matchesIp.empty ? null : matchesIp.front;
}

pragma(inline, false)
private const(Char)[] method2(Char)(return const(Char)[] s)
{
    size_t end = s.length;
    if (end < 7) return null;
    // Check the first three `[.]\d{1,3}`
    foreach (_; 0 .. 3)
    {
        size_t start = void;
        if (end >= 2 && s[end-2] == '.')
            start = end - 2;
        else if (end >= 3 && s[end-3] == '.')
            start = end - 3;
        else if (end >= 4 && s[end-4] == '.')
            start = end - 4;
        else
            return null;
        uint x = 0;
        foreach (i; start + 1 .. end)
        {
            uint c = cast(uint) s[i] - '0';
            if (c > 9) return null;
            x = x * 10 + c;
        }
        if (x > 255) return null;
        end = start;
    }
    // Check the final `\d{1,3}`.
    if (end < 1) return null;
    size_t start = end - 1;
    uint x = cast(uint) s[start] - '0';
    if (x > 9) return null;
    if (start > 0 && cast(uint) s[start-1] - '0' <= 9)
    {
        --start;
        x += 10 * (cast(uint) s[start] - '0');
        if (start > 0 && cast(uint) s[start-1] - '0' <= 9)
        {
            --start;
            x += 100 * (cast(uint) s[start] - '0');
        }
    }
    if (x > 255) return null;
    // Must either be at start of string or preceded by a non-word character.
    // (TO DETERMINE: is the definition of "word character" ASCII only?)
    if (start == 0) return s;
    const b = s[start - 1];
    import std.ascii : isAlphaNum;
    if (isAlphaNum(b) || b == '_') return null;
    return s[start .. $];
}
```
</details>

std/net/isemail.d
n8sh: >in this specific instance `isEmail` is not `@nogc nothrow`

It probably should be but I'll leave that for another pull request. It looks like `isEmail` is incorrectly using Exceptions rather than assert() to debug logic errors in the function itself ([see forum discussion](https://forum.dlang.org/thread/m07gf1$18jl$1@digitalmars.com)).

std/net/isemail.d
```diff
@@ -1863,3 +1857,96 @@ const(T)[] get (T) (const(T)[] str, size_t index, dchar c)
...
+static fourChars = ctRegex!(`^[0-9A-Fa-f]{0,4}$`.to!(const(Char)[]));
+...
+a => a.matchFirst(fourChars).empty
+---
++/
+private bool isUpToFourHexChars(Char)(scope const(Char)[] s)
+{
+    import std.ascii : isHexDigit;
+    if (s.length > 4) return false;
+    foreach (c; s)
```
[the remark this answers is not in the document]
schveiguy: Also, this would involve auto-decoding I think.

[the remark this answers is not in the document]
wilzbach: `.byCodeUnit` - no autodecoding and in `@nogc nothrow pure @safe`

https://run.dlang.io/is/PvbTfz

schveiguy:
I think this is in a good enough state, we can have further PRs to include uses of Phobos and nifty range pipelines as improvements. Given the two approvals, I'll pull.

PR #6132, opened 2018-02-06
Remove auto-decoding from startsWith when Needles are all narrow chars

JackStouffer:


std/algorithm/searching.d
```diff
@@ -4104,7 +4104,22 @@ if (isInputRange!Range && Needles.length > 1 &&
     is(typeof(.startsWith!pred(doesThisStart, withOneOfThese[0])) : bool ) &&
     is(typeof(.startsWith!pred(doesThisStart, withOneOfThese[1 .. $])) : uint))
 {
-    alias haystack = doesThisStart;
+    import std.meta : allSatisfy;
+
+    template checkType(T)
+    {
+        enum checkType = is(Unqual!(ElementEncodingType!Range) == Unqual!T);
```
BBasile: Any particular reason why you use a template and not a templated enum ?

n8sh:
You need to check that that `startsWith` is using the default predicate `"a == b"` for this to be sound.

wilzbach:
> You need to check that that startsWith is using the default predicate "a == b" for this to be sound.

Lambda comparison has been merged a few days ago:

https://dlang.org/changelog/pending.html#lambdacomp

schveiguy:
> Lambda comparison has been merged a few days ago:

There's a list issue with that, but I don't know where the source is, so not sure how to change it.

It looks like item 12 really should be a sub-bullet of item 11.

wilzbach:
>  but I don't know where the source is, so not sure how to change it.

https://github.com/dlang/dmd/tree/master/changelog

I will add a few notes and links in the header.

schveiguy:
Hm... the docs look correct to me (The parentheses match as expected), so maybe it's an issue with the thing that assembles it together? Where is that code?

Sorry for hijacking this thread, don't know of a better place to ask these questions...

wilzbach:
> Hm... the docs look correct to me (The parentheses match as expected), so maybe it's an issue with the thing that assembles it together? Where is that code?

https://github.com/dlang/tools/blob/master/changed.d
https://github.com/dlang/dlang.org/blob/master/posix.mak#L900

> Sorry for hijacking this thread, don't know of a better place to ask these questions...

Slack?

wilzbach:
@schveiguy fix for the changelog entry: https://github.com/dlang/dmd/pull/7847

JackStouffer:
What does `isSame` return if it can't compare the lambdas? Does anyone know?

wilzbach:
`false`

std/algorithm/searching.d
```diff
@@ -4104,7 +4104,24 @@ if (isInputRange!Range && Needles.length > 1 &&
...
+    }
+
+    // auto-decoding special case
+    static if (__traits(isSame, binaryFun!pred, binaryFun!"a == b") &&
```
wilzbach: Use `alias pred = "a == b"` to `(a, b) => a == b` and then you can use here `(a, b) => a == b`
No need to stress the compiler more than necessary with mixins.

andralex:
Cool! I think things are fine as they are to avoid too much change of interface.

andralex:
Does this help with using `skipOver` to skip over the BOM per discussion in https://github.com/dlang/phobos/pull/6114? If so, adding some examples with this PR would be awesome.

wilzbach:
> Cool! I think things are fine as they are to avoid too much change of interface.

FWIW my comment was just about reducing the compile-time by avoiding an unnecessary mixin (and the longer-term goal to fade out string functions). I don't think that the interface of a template would be affected by changing by changing the default value of a parameter.

JackStouffer:
>Does this help with using skipOver to skip over the BOM per discussion in #6114?

probably not, as `\uFEFF` is always going to be a `dchar` right?

PR #6133, opened 2018-02-06
fix doc inaccuracy in dirEntries

timotheecour:
cf FORUM: dirEntries returns relative, not absolute paths


CyberShadow:
> The name of each iterated directory entry contains the absolute or relative path (depending on pathname).

Well, what else could it contain? What's the point of this sentence? Why keep it at all?

PR #6135, opened 2018-02-07
fix issue 18349 - std/math.d(543,33): Deprecation: integral promotion not done for -x

BBasile:
last fix was incomplete. abs didn't handle type constructors yet.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -553,13 +553,13 @@ auto abs(Num)(Num x)
 // workaround for https://issues.dlang.org/show_bug.cgi?id=18251
 //if (!isDeprecatedComplex!Num &&
     //(is(typeof(Num.init >= 0)) && is(typeof(-Num.init)) ||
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -553,13 +553,13 @@ auto abs(Num)(Num x)
...
     //(is(typeof(Num.init >= 0)) && is(typeof(-Num.init)) ||
-    //(is(Num == short) || is(Num == byte))))
+    //(is(Num : const(short)) || is(Num : const(byte)))))
 {
     static if (isFloatingPoint!(Num))
         return fabs(x);
     else
     {
-        static if (is(Num == short) || is(Num == byte))
+        static if (is(Num : const(short)) || is(Num : const(byte)))
```
schveiguy: Note, this isn't the same. `ubyte`, `char`, `wchar`, etc. are going to pass these.

I suggest `is(Unqual!Num == short)`

BBasile: okay, done.

PR #6137, opened 2018-02-07
std.process: Get rid of goofy warning in unit test

CyberShadow:
Refactor out the need of the warning by placing the part of the test containing "return" statements into its own function.


CyberShadow:
Diff without whitespace: https://github.com/dlang/phobos/pull/6137/files?w=1

PR #6138, opened 2018-02-08
fix Issue 18397 - Poor implementation of std.conv.hexString results i…

WalterBright:
…n unintended bloat

The trick is to make it an enum, use a mixin string literal rather than an array, and instantiate the template in std.conv rather than in user code.

For:
```
import std.conv;
void test() {
    auto data = cast(ubyte[]) hexString!"deadbeef";
}
```
now compiles to:
```
___a4_deadbeef	comdat
	db	0deh,0adh,0beh,0efh,000h	;.....
___a4_deadbeef	ends
_D5test24testFZv	comdat
	assume	CS:_D5test24testFZv
		mov	ECX,offset FLAT:___a4_deadbeef
		mov	EAX,4
		ret
_D5test24testFZv	ends
```
instead of the hundreds of bytes of code it used to.

### DIFF
--- a/changelog/fix18397.dd
+++ b/changelog/fix18397.dd
@@ -1,18 +1,18 @@
-Changed std.conv.hexString() to return an immutable string literal
-rather than an array of ubytes. This is more in keeping with
-the documentation that says it is a replacement for the deprecated
-x"deadbeef" string literal syntax.
+Changed `std.conv.hexString` to return an immutable string literal
 
-https://dlang.org/phobos/std_conv.html#hexString
+$(REF hexString, std,conv) now returns an immutable string literal rather than an array of ubytes.
+This is more in keeping with
+the documentation that says it is a replacement for the deprecated
+`x"deadbeef"` string literal syntax.
 
 The benefits of this change are:
 
-* behavior consistency with the documentation
-* the hexString template instantiations no longer appear in the
-generated object file
-* the generated object file no longer contains references to TypeInfo
-and various druntime functions
-* it is now compatible with -betterC mode
+$(UL
+  $(LI behavior consistency with the documentation)
+  $(LI the `hexString` template instantiations no longer appear in the generated object file)
+  $(LI the generated object file no longer contains references to `TypeInfo` and various druntime functions)
+  $(LI it is now compatible with `-betterC` mode)
+)
 
 In some cases, code did rely on it being an array, and a cast will
 fix the issue:
@@ -26,4 +26,3 @@ enum ubyte[8] input = hexString!"c3 fc 3d 7e fb ea dd aa";
 // add cast to fix:
 enum ubyte[8] input = cast(ubyte[8]) hexString!"c3 fc 3d 7e fb ea dd aa";
 ---
-
--- a/changelog/fix18397.dd
+++ b/changelog/fix18397.dd
@@ -1,0 +1,29 @@
+Changed std.conv.hexString() to return an immutable string literal
+rather than an array of ubytes. This is more in keeping with
+the documentation that says it is a replacement for the deprecated
+x"deadbeef" string literal syntax.
+
+https://dlang.org/phobos/std_conv.html#hexString
+
+The benefits of this change are:
+
+* behavior consistency with the documentation
+* the hexString template instantiations no longer appear in the
+generated object file
+* the generated object file no longer contains references to TypeInfo
+and various druntime functions
+* it is now compatible with -betterC mode
+
+In some cases, code did rely on it being an array, and a cast will
+fix the issue:
+
+---
+// now an error:
+enum ubyte[8] input = hexString!"c3 fc 3d 7e fb ea dd aa";
+---
+
+---
+// add cast to fix:
+enum ubyte[8] input = cast(ubyte[8]) hexString!"c3 fc 3d 7e fb ea dd aa";
+---
+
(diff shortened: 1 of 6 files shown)


WalterBright:
```
std/digest/md.d(483): Error: cannot implicitly convert expression `"\xc3\xfc\xd3\xd7a\x92\xe4\x00}\xfbIl\xcag\xe1;"` of type `string` to `ubyte[16]`
```
I knew somebody was likely misusing hexString :-(

std/digest/crc.d
```diff
@@ -381,7 +381,7 @@ struct CRC(uint N, ulong P) if (N == 32 || N == 64)
                     "1234567890123456789012345678901234567890");
     assert(digest == cast(ubyte[]) hexString!"724aa97c");
 
-    enum ubyte[4] input = hexString!"c3fcd3d7";
+    enum ubyte[4] input = cast(ubyte[4]) hexString!"c3fcd3d7";
```
wilzbach: If we decide the breakage is worth it, it should at least be mentioned in the changelog.

std/digest/crc.d
```diff
@@ -421,7 +421,7 @@ struct CRC(uint N, ulong P) if (N == 32 || N == 64)
                          "1234567890123456789012345678901234567890");
     assert(digest == cast(ubyte[]) hexString!"bd3eb7765d0a22ae");
 
-    enum ubyte[8] input = hexString!"c3fcd3d7efbeadde";
+    enum ubyte[8] input = cast(ubyte[8]) hexString!"c3fcd3d7efbeadde";
```
carun: Just wondering if it is mandatory to use `enum ubyte[8] input`? `auto input` would look idiomatic.

std/digest/crc.d
carun: Just wondering if it is mandatory to use `enum ubyte[8] input`? `auto input` would look idiomatic.

std/digest/crc.d
```diff
@@ -461,7 +461,7 @@ struct CRC(uint N, ulong P) if (N == 32 || N == 64)
                         "1234567890123456789012345678901234567890");
     assert(digest == cast(ubyte[]) hexString!"760cd2d3588bf809");
 
-    enum ubyte[8] input = hexString!"c3fcd3d7efbeadde";
+    enum ubyte[8] input = cast(ubyte[8]) hexString!"c3fcd3d7efbeadde";
```
carun: Just wondering if it is mandatory to use `enum ubyte[8] input`? `auto input` would look idiomatic.

std/digest/ripemd.d
```diff
@@ -649,7 +649,7 @@ struct RIPEMD160
                     "1234567890123456789012345678901234567890");
     assert(digest == cast(ubyte[]) hexString!"9b752e45573d4b39f4dbd3323cab82bf63326bfb");
 
-    enum ubyte[20] input = hexString!"f71c27109c692c1b56bbdceb5b9d2865b3708dbc";
+    enum ubyte[20] input = cast(ubyte[20]) hexString!"f71c27109c692c1b56bbdceb5b9d2865b3708dbc";
```
carun: Just wondering if it is mandatory to use `enum ubyte[16] input`? `auto input` would look idiomatic.

std/digest/sha.d
```diff
@@ -1117,7 +1117,7 @@ alias SHA512_256 = SHA!(1024, 256); /// SHA alias for SHA-512/256, hash is ubyte
     assert(digest512_224 == cast(ubyte[]) hexString!"37ab331d76f0d36de422bd0edeb22a28accd487b7a8453ae965dd287");
     assert(digest512_256 == cast(ubyte[]) hexString!"9a59a052930187a97038cae692f30708aa6491923ef5194394dc68d56c74fb21");
 
-    enum ubyte[20] input = hexString!"a9993e364706816aba3e25717850c26c9cd0d89d";
+    enum ubyte[20] input = cast(ubyte[20]) hexString!"a9993e364706816aba3e25717850c26c9cd0d89d";
```
carun: Just wondering if it is mandatory to use `enum ubyte[20] input`? `auto input` would look idiomatic.

JackStouffer:
@carun When you have a suggestion that applies to multiple pieces of code, simply stating something like "same throughout" will suffice.

And it must be at least `enum input`, as `auto` turns it into a variable.

WalterBright:
> What are those non-hex chars doing in there?

The string-to-ascii conversion represents non-printable characters as \xnn

carun:
@JackStouffer I see, thanks!

schveiguy:
> The string-to-ascii conversion represents non-printable characters as \xnn

OK, so it produces the correct string literal, but then when displaying the error message shows the printable ones as the actual chars.

I was looking at the code, and couldn't figure out of something weird was sneaking in!

WalterBright:
added changelog entry

WalterBright:
oops, pushed wrong button

WalterBright:
For reference, the documentation: https://dlang.org/phobos/std_conv.html#hexString

changelog/fix18397.dd
```diff
@@ -1,18 +1,18 @@
...
-generated object file
-* the generated object file no longer contains references to TypeInfo
-and various druntime functions
-* it is now compatible with -betterC mode
+$(UL
+  $(LI behavior consistency with the documentation)
+  $(LI the `hexString` template instantiations no longer appear in the generated object file)
+  $(LI the generated object file no longer contains references to `TypeInfo` and various druntime functions)
+  $(LI it is now compatible with `-betterC` mode)
+)
```
wilzbach: FYI @WalterBright even if `changed.d` overlays some goodies over Ddoc (no need for the ugly `$(P)`, Ddoc is still used in the background :/

wilzbach:
@WalterBright FYI: I "DDocified" your changelog entry. The changelog generator still uses old Ddoc behind the scenes.

PR #6139, opened 2018-02-08
Issue 18398 - std.datetime.stopwatch documented examples could be better

n8sh:
Someone using `std.datetime.stopwatch` probably wants to use it to obtain a time in milliseconds or some other unit of time but none of the example code shows how to do this.

Proposed addition:

```d
    /// Measure a time in milliseconds, microseconds, or nanoseconds
    @safe nothrow @nogc unittest
    {
        auto sw = StopWatch(AutoStart.no);
        sw.start();
        // ... Insert operations to be timed here ...
        sw.stop();

        long msecs = sw.peek.total!"msecs";
        long usecs = sw.peek.total!"usecs";
        long nsecs = sw.peek.total!"nsecs";

        assert(usecs >= msecs * 1000);
        assert(nsecs >= usecs * 1000);
    }
```

### DIFF
--- a/std/datetime/stopwatch.d
+++ b/std/datetime/stopwatch.d
@@ -82,6 +82,22 @@ struct StopWatch
 {
 public:
 
+    /// Measure a time in milliseconds, microseconds, or nanoseconds
+    @safe nothrow @nogc unittest
+    {
+        auto sw = StopWatch(AutoStart.no);
+        sw.start();
+        // ... Insert operations to be timed here ...
+        sw.stop();
+
+        long msecs = sw.peek.total!"msecs";
+        long usecs = sw.peek.total!"usecs";
+        long nsecs = sw.peek.total!"nsecs";
+
+        assert(usecs >= msecs * 1000);
+        assert(nsecs >= usecs * 1000);
+    }
+
     ///
     @system nothrow @nogc unittest
     {


std/datetime/stopwatch.d
```diff
@@ -82,6 +82,22 @@ struct StopWatch
 {
 public:
 
+    /// Measure a time in milliseconds, microseconds, or nanoseconds
+    @safe nothrow @nogc unittest
+    {
+        auto sw = StopWatch(AutoStart.no);
+        sw.start();
+        // ... Insert operations to be timed here ...
+        sw.stop();
```
wilzbach: The super-idiomatic use is:

```d
{
    sw.start;
    scope(exit) sw.stop;

    // ... Insert operations to be timed here ...
}
```

but even I use this hardly ;-)
We could add a method that stops when the dtor is called:

```d
with (sw.startScope) {
  // benchmark here
}
```

But I'm not sure whether it's a good idea and worth bothering. https://github.com/dlang/phobos/pull/5416 (dontOptimizeAway) is a lot more important.

std/datetime/stopwatch.d
```diff
@@ -82,6 +82,22 @@ struct StopWatch
...
+
+        long msecs = sw.peek().total!"msecs";
+        long usecs = sw.peek().total!"usecs";
+        long nsecs = sw.peek().total!"nsecs";
```
wilzbach: Idiomatic D doesn't use parenthesis in UFCS calls -> `sw.peek.total!"msecs"`.
(Users will most likely use exactly what is written here, so it might make sense to be picky about the style of the first example)

n8sh: Makes sense

std/datetime/stopwatch.d
jmdavis: I actually rather doubt that using `scope` with `StopWatch` is all that idiomatic, because stopping it is not the last thing you do with it. It's also designed with the idea that you can start and stop it multiple times (if you don't want to do that, arguably just using `MonoTime` directly makes more sense). So, while I'm sure that there are cases where using `scope` makes sense, I question that that's the typical use case. Regardless, the point of the docs is to show the basics of how to use it. I would hope that your typical D programmer could figure out when it's appropriate to use `scope` with `StopWatch` in their own code.

n8sh:
The example is not showing up in the ddoc for the master branch. Did I put it in the wrong place?
https://dlang.org/phobos-prerelease/std_datetime_stopwatch.html

jmdavis:
I don't see anything obviously wrong. I suspect that the documentation simply hasn't been rebuilt yet after the update.

jmdavis:
Actually. I take that back. You added a ddoc-ed unittest block that's suppose to go with the type, but there was already one there, and it's not showing up either. So, there is something wrong.

wilzbach:
It's showing up on the ddox pages: https://dlang.org/library-prerelease/std/datetime/stopwatch/stop_watch.html

wilzbach:
Related issue: https://issues.dlang.org/show_bug.cgi?id=16992

wilzbach:
A workaround would be to set the example as module-level, documented unittest.

PR #6140, opened 2018-02-08
Use REF1 macro instead of OBJECTREF

wilzbach:
The REF1 macro should have been used here instead.

See also: https://github.com/dlang/dlang.org/pull/2191

edit: also fixed a few typos while I was at it.

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -80,8 +80,7 @@ License:
 Source:
     $(PHOBOSSRC std/_process.d)
 Macros:
-    OBJECTREF=$(D $(LINK2 object.html#$0,$0))
-    LREF=$(D $(LINK2 #.$0,$0))
+    OBJECTREF=$(REF1 $0, object)
 */
 module std.process;
 


CyberShadow:
> child process's standard streams 

Not sure about this. There's 4 `s` letters in a row here.

Some commentary on Wikipedia:
https://en.wikipedia.org/wiki/Apostrophe#Singular_nouns_ending_with_an_%22s%22_or_%22z%22_sound

wilzbach:
> The Supreme Court of the United States is split on whether a possessive singular noun that ends with s should always have an additional s after the apostrophe, sometimes have an additional s after the apostrophe (for instance, based on whether the final sound of the original word is pronounced /s/ or /z/), or never have an additional s after the apostrophe. The informal majority view (5–4, based on past writings of the justices) has favoured the additional s, but a strong minority disagrees.

:scream: - maybe we need a typography styleguide too :confused: 
Okay will drop the second commit for now.

n8sh:
@wilzbach Don't be silly. Please add it back. Documentation should not without good reason use spellings that a majority of readers will perceive as an error.

wilzbach:
> Don't be silly. 

In my experience, _anything_ that is potentially controversial in a PR can (or will) block it.
Hence, it's better to move such things into separate PRs, s.t. the main PR can be merged quickly.
I submitted this because of the `REF1` macro (sorry for screwing up the initial title) - that is/was my main concern.

> Please add it back. 

Sure. Let's have the discussion here: https://github.com/dlang/phobos/pull/6141

PR #6141, opened 2018-02-08
Add missing 's

wilzbach:
(This seems to be a controversial topic)

https://github.com/dlang/phobos/pull/6140#issuecomment-364065744

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -1389,7 +1389,7 @@ enum Config
     By default, the child process inherits the parent's environment,
     and any environment variables passed to $(LREF spawnProcess) will
     be added to it.  If this flag is set, the only variables in the
-    child process' environment will be those given to spawnProcess.
+    child process's environment will be those given to spawnProcess.
     */
     newEnv = 1,
 
@@ -2023,7 +2023,7 @@ input, output and/or error streams.
 
 $(D pipeProcess) and $(D pipeShell) are convenient wrappers around
 $(LREF spawnProcess) and $(LREF spawnShell), respectively, and
-automate the task of redirecting one or more of the child process'
+automate the task of redirecting one or more of the child process's
 standard streams through pipes.  Like the functions they wrap,
 these functions return immediately, leaving the child process to
 execute in parallel with the invoking process.  It is recommended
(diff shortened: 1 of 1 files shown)


CyberShadow:
Let's ask the experts:
https://english.stackexchange.com/questions/430160/singular-posessive-ending-with-ss-before-another-word-starting-with-s

n8sh:
>Let's have the discussion here

I think this pull request should go ahead. Technical documentation should not without good reason use spellings that a majority of readers will perceive as errors. Perceived spelling mistakes are distracting and give a document a slipshod appearance regardless of whether they are truly mistakes or are accepted alternative spellings.

n8sh:
>Since, 1810, forms like James’s (which I will call type A) have generally been more commonly used than forms like James’ (type B), according to my research using the Corpus of Historical American English (COHA).

Is the current year after 1810?

schveiguy:
Found this: http://www.thepunctuationguide.com/apostrophe.html

> Nouns that end in an s sound take only an apostrophe when they are followed by sake.

quickfur:
Technically, words that end with s take only an apostrophe without another s.

But seriously, this is a ridiculous amount of spat over a minor grammar nitpick.  *I* say we should just reword the entire lousy sentence to read something like "X of Y" instead of "Y's X". E.g., "the environment of the child process" instead of "the child process('s|'|s|whatever) environment".  The former is much clearer to read anyway, and sidesteps this entire lousy argument.

MetaLang:
Let's not start grammar fights and introduce useless churn into Phobos. I'm not even supportive of changing everything to one consistent style; just leave it as it is.

wilzbach:
From the resource @schveiguy posted:

> The possessive of a plural noun is formed by adding only an apostrophe when the noun ends in s, and by adding both an apostrophe and s when it ends in a letter other than s.

Maybe we should add this excellent Punctuation Guide to the Phobos part of the DStyle? It would help to avoid such discussions in the future as there is a goto resource we can link to.

> Let's not start grammar fights and introduce useless churn into Phobos.

Fair enough.

> I'm not even supportive of changing everything to one consistent style; just leave it as it is.

I disagree - a consistent style is important. It leaves an unprofessional touch.

PR #6143, opened 2018-02-08
Fix Issue 12335 - std.algorithm.skipOver should support multiple args like startsWith

wilzbach:
Now you can do the following:

```d
auto s = "DLang.rocks";
assert(!s.skipOver("dlang", "DLF", "DLang "));
assert(s == "DLang.rocks");

assert(s.skipOver("dlang", "DLANG", "DLF", "D", "DL", "DLanp"));
assert(s == "ang.rocks");
s = "DLang.rocks";
```

### DIFF
--- a/std/algorithm/searching.d
+++ b/std/algorithm/searching.d
@@ -3915,62 +3915,125 @@ if (isInputRange!Range && !isInfinite!Range &&
 }
 
 /**
-Skip over the initial portion of the first given range that matches the second
-range, or if no second range is given skip over the elements that fullfil pred.
+Skip over the initial portion of the first given range (`haystack`) that matches
+any of the additionally given ranges (`needles`) fully, or
+if no second range is given skip over the elements that fulfill pred.
 Do nothing if there is no match.
 
 Params:
     pred = The predicate that determines whether elements from each respective
         range match. Defaults to equality `"a == b"`.
 */
-template skipOver(alias pred = "a == b")
+template skipOver(alias pred = (a, b) => a == b)
 {
+    import std.meta : allSatisfy;
+
+    enum bool isPredComparable(T) = ifTestable!(T, binaryFun!pred);
+
     /**
-    r1 = The $(REF_ALTTEXT forward range, isForwardRange, std,range,primitives) to
-        move forward.
-    r2 = The $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
-        representing the initial segment of `r1` to skip over.
-    e = The element to match.
+    Params:
+        haystack = The $(REF_ALTTEXT forward range, isForwardRange, std,range,primitives) to
+                   move forward.
+        needles = The $(REF_ALTTEXT input ranges, isInputRange, std,range,primitives)
+                  representing the prefix of `r1` to skip over.
+        es = The element to match.
 
     Returns:
-    true if the initial segment of `r1` matches `r2` or `pred` evaluates to true,
-    and `r1` has been advanced to the point past this segment; otherwise false, and
-    `r1` is left in its original position.
+        `true` if the prefix of `haystack` matches any range of `needles` fully
+        or `pred` evaluates to true, and `haystack` has been advanced to the point past this segment;
+        otherwise false, and `haystack` is left in its original position.
+
+    Note:
+        By definition, empty ranges are matched fully and if `needles` contains an empty range,
+        `skipOver` will return `true`.
     */
-    bool skipOver(R1, R2)(ref R1 r1, R2 r2)
-    if (is(typeof(binaryFun!pred(r1.front, r2.front))) &&
-        isForwardRange!R1 &&
-        isInputRange!R2)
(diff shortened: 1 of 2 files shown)


std/algorithm/searching.d
```diff
@@ -3915,62 +3915,124 @@ if (isInputRange!Range && !isInfinite!Range &&
...
+    Note:
+        By definition, empty ranges are matched fully and if `needles` contains an empty range,
+        `skipOver` will return `true`.
     */
-    bool skipOver(R1, R2)(ref R1 r1, R2 r2)
-    if (is(typeof(binaryFun!pred(r1.front, r2.front))) &&
-        isForwardRange!R1 &&
-        isInputRange!R2)
+    bool skipOver(Haystack, Needles...)(ref Haystack haystack, Needles needles)
+    if (is(typeof(binaryFun!pred(haystack.front, needles[0].front))) &&
```
dukc: allSatisfy!(isInputRange, Needles) is double-checked

std/algorithm/searching.d
```diff
@@ -4027,6 +4101,18 @@ template skipOver(alias pred = "a == b")
     assert(s4.skipOver!isWhite && s3.empty);
 }
 
+/// Variadic skipOver
```
dukc: These can probably be nothrow pure too?

wilzbach: Due to UTF auto-decoding strings aren't `nothrow` by default. It be prevented by adding a few `byCodeUnit` in there, but typically one test is added to ensure that all attributes work (i.e. `@safe @nogc pure nothrow` in this case), s.t. the other tests don't need to bother with `static immutable`, `byCodeUnit` or other unrelated "tricks". This is especially important because this test is the publicly documented one.

The test with all attributes is a few lines below.

dukc: But can they still be marked pure without changing the internals? I see nothing that should prevent it.

If so IMO it will add to the documentation, as the user can see it works in pure code.

wilzbach: Unittest attributes aren't displayed on the documentation, but yes `pure` could be added here.

wilzbach:
Ping - I believe this is a quite useful generalization of `skipOver` and this has been sitting in the queue for two months now.

std/algorithm/searching.d
```diff
@@ -3915,62 +3915,124 @@ if (isInputRange!Range && !isInfinite!Range &&
...
+        allSatisfy!(isInputRange, Needles) &&
+        !is(CommonType!(staticMap!(ElementType, staticMap!(Unqual, Needles))) == void))
     {
-        static if (is(typeof(pred) : string) && pred == "a == b"
-                && is(typeof(r1[0 .. $] == r2) : bool)
-                && is(typeof(r2.length > r1.length) : bool)
-                && is(typeof(r1 = r1[r2.length .. $])))
+        static if (__traits(isSame, pred, (a, b) => a == b)
+                && is(typeof(haystack[0 .. $] == needles[0]) : bool)
+                && is(typeof(haystack = haystack[size_t.init .. $]))
```
n8sh: Not that it matters in this instance, but writing `size_t.init` here is no better than writing 0. If for some reason the first argument of `opIndex` had to be `uint` this would still compile.

JackStouffer: >is(typeof(haystack = haystack[size_t.init .. $])

this unnecessarily limits those ranges who've not defined opDollar

std/algorithm/searching.d
```diff
@@ -3915,62 +3915,124 @@ if (isInputRange!Range && !isInfinite!Range &&
...
+                && hasLength!Haystack && allSatisfy!(hasLength, Needles))
         {
-            if (r2.length > r1.length || r1[0 .. r2.length] != r2)
+            ptrdiff_t longestMatch = -1;
+            static foreach (r2; needles)
```
JackStouffer: where's this `r2` definition coming from? You changed the name of the parameters and I can't see it redefined anywhere.

std/algorithm/searching.d
wilzbach: I use opDollar below and I thought we agreed a few months ago that we don't cater for those ranges and either deprecate this or let the compiler automatically emit `alias opDollar = length` if it's not defined.

std/algorithm/searching.d
wilzbach: r2 used to be name of the needle to search for. 
As with this PR multiple needles are supported, it's no longer a parameter (needles is), but I kept using r2 in this static foreach loop over all needles to make reviewing easier as some point stayed the same.

wilzbach:
Ping  - this has been open now for quite a while and I don't see anything blocking it.

wilzbach:
The DAutoTest failure is unrelated and I would like to get this in for 2.082. Any final objections?

wilzbach:
Urgh so it looks like Buildkite never ran here (and the old Jenkins result didn't include dxml).
Thus, we managed to introduce a regression that has been noticed in dxml - see e.g. https://github.com/dlang/ci/pull/286

PR #5001, opened 2016-12-27
std.algorithm.searching: no mapping-specialization for extremum

wilzbach:
In #4265 I showed that `extremum` (aka `minElement` and `maxElement`) can be made a lot faster if we provide a special path for the identity case, because the compiler can then make use of SSE instructions.
In theory the compiler should be smart enough to figure it out himself, but well neither dmd nor ldc are that smart :/

For reference here are the benchmarks from #4265:

```
dmd -release -O foo.d && ./foo
0 4 secs, 610 ms, 609 μs, and 5 hnsecs
1 3 secs, 769 ms, 89 μs, and 4 hnsecs
```

```
ldc -release -O3 foo.d && ./foo
0 1 sec, 345 ms, and 914 μs
1 216 ms, 347 μs, and 1 hnsec
```

Source for benchmarks: http://sprunge.us/QSWP

This PR does the specialization via overloads whereas #4265 does them via comparison of strings.

PR #6144, opened 2018-02-08
Use __traits(isSame) for min,maxElement

wilzbach:
`__traits(isSame, a, b)` rocks :)

The specialization was discussed in https://github.com/dlang/phobos/pull/4265 and merged in https://github.com/dlang/phobos/pull/5001.

Use https://github.com/dlang/phobos/pull/6144/files?w=1 for a nicer diff.

### DIFF
--- a/std/algorithm/searching.d
+++ b/std/algorithm/searching.d
@@ -1311,33 +1311,64 @@ if (isInputRange!Range && !isInfinite!Range &&
     alias CommonElement = CommonType!(Element, RangeElementType);
     Unqual!CommonElement extremeElement = seedElement;
 
-    alias MapType = Unqual!(typeof(mapFun(CommonElement.init)));
-    MapType extremeElementMapped = mapFun(extremeElement);
 
-    // direct access via a random access range is faster
-    static if (isRandomAccessRange!Range)
+    // if we only have one statement in the loop, it can be optimized a lot better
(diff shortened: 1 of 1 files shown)


std/algorithm/searching.d
```diff
@@ -1311,33 +1311,64 @@ if (isInputRange!Range && !isInfinite!Range &&
...
-    MapType extremeElementMapped = mapFun(extremeElement);
 
-    // direct access via a random access range is faster
-    static if (isRandomAccessRange!Range)
+    // if we only have one statement in the loop, it can be optimized a lot better
+    static if (__traits(isSame, map, a => a))
     {
-        foreach (const i; 0 .. r.length)
+
+        // direct access via a random access range is faster
```
n8sh: Is this really true? Looping over an index is faster than repeated `empty`/`front`/`popFront()`?

wilzbach: Yes, D ranges have an overhead of ~20-50% for arrays. It's pretty bad for e.g. joiner (see https://gist.github.com/wilzbach/ffd5d20639766a831fd4b1962572897a), but it's a problem here too.

Check also https://github.com/dlang/phobos/pull/4265

The reason is that the compiler can inline the commands a lot better for native arrays and even do vectorization better than for the generic loop.

PR #6147, opened 2018-02-09
Add isSliceOf to check the origin of a slice

wilzbach:
This is has come up before:
- https://github.com/dlang/phobos/pull/2416
- https://forum.dlang.org/post/lsaktk$1ob5$1@digitalmars.com

tl;dr: of the previous discussion:
- `overlap` exists, but is confusing to use
- there were no "real world" use cases / huge interest
  - @dcarp provided some and I have run into similar uses when working with allocators (it's analog to `owns`)
  - `sameHead` and `sameTail` already exists
  - `@trusted` doesn't work for one-liners

CC @nordlow @dcarp


JackStouffer:
`std.exception.doesPointTo`?

wilzbach:
> std.exception.doesPointTo?

Sorry for not mentioning that one. It's pretty useless (and in the wrong module):

>If target is a pointer, a dynamic array or a class, then these functions will only check if source points to target, not what target references.

Convince yourself: https://run.dlang.io/is/NoZVWq

wilzbach:
Argh, the spuriously failing std.regex due to its high memory consumption is getting more and more annoying.

See also: https://issues.dlang.org/show_bug.cgi?id=18411

std/array.d
```diff
@@ -1360,6 +1363,84 @@ pure nothrow bool sameTail(T)(in T[] lhs, in T[] rhs)
...
+Checks whether a `slice` is part of a `whole` array.
+
+Params:
+
+    slice = The array to check for its origin
+    whole = The origin array to be queried
+
+Returns: `true` if `slice` is a part of `whole`.
+*/
+bool isSliceOf(T)(const scope T[] slice, const scope T[] whole) @trusted
```
nordlow: Use `in` instead of `const scope`. They are equivalent.

std/array.d
```diff
@@ -1360,6 +1363,84 @@ pure nothrow bool sameTail(T)(in T[] lhs, in T[] rhs)
     }}
 }
 
+/**
+Checks whether a `slice` is part of a `whole` array.
```
nordlow: I would have used `part` instead of `slice`.

std/array.d
wilzbach: Hehe, that used to be in the spec, but the compiler never did this:

- https://issues.dlang.org/show_bug.cgi?id=17928
- https://github.com/dlang/dlang.org/pull/2126
- https://github.com/dlang/phobos/pull/6074

std/array.d
wilzbach: Ok renamed.

std/array.d
jmdavis: I've actually argued for years that `in` should never be used, because `scope` wasn't actually implemented for anything other than delegates, so using `in` was just risking having your code break as soon as `scope` was more fully implemented, but precisely because of that problem, Walter change `in` to just be `const`, meaning that `in` is now completely useless.

It's meaning may or may not change later, but either way, I think that it should be actively avoided unless/until it's actually implemented to mean something rather than just `const` and is clearly going to stay that way, at which point, the code can actually be designed with its proper meaning in mind instead of using `in` because of what it might mean in the future or because it seems like the opposite of `out`.

nordlow: Hmm, that's unfortunate. I've gotten used to using `in`.

JackStouffer:
Any reason to not use the `&whole[0]` style rather than `whole.ptr`?

wilzbach:
> Any reason to not use the &whole[0] style rather than whole.ptr?

Yes - `isSliceOf` should work for empty arrays too, e.g.

```
assert(null.isSliceOf(null));
```

JackStouffer:
`null` isn't a slice of itself because `null` isn't anything. Nothing can't be a part of nothing.

wilzbach:
> Nothing can't be a part of nothing.

So you would treat `null` like FP's `NaN` where the default value is poison and  yields `false`? Interesting thought!
The problem here is that [`null is null`](https://run.dlang.io/is/m6UCoZ), so I thought it makes sense that `isSliceOf` matches this behavior.

JackStouffer:
If I had to be pedantic, I would say the nothing is the same as nothing but nothing is not a memory view of nothing.

changelog/std-array-sliceOf.dd
```diff
@@ -0,0 +1,18 @@
+std.array.sliceOf has been added
+
+$(REF sliceOf, std,array) provides a `@safe` way to query
```
JackStouffer: a safe way to query what?

n8sh:
Tangential to this pull request:

>overlap exists, but is confusing to use

Maybe the documentation needs improvement then, because `if (a.overlap(b))` should be crystal clear!

andralex:
This API could be better. As it is, it returns very little information i.e. "is this range overlapping this other range in any position?" with no further information on position.

The same information can be trivially fetched using `overlap` in one of two ways:

```d
small.overlap(large) is small
```

or

```d
small.overlap(large).length == small.length
```

I like the first one more for generality, and the second more for efficiency. Given the rarity of such a query, I think we don't need a new function for it.

BUT! All is not lost. I propose to convert this PR into a documentation PR. The definition and use of `isSliceOf` can be put in a documentation unittest. That way we match google queries like "dlang isSliceOf" and similar. Win-win-win!

andralex (dismissed):
To clarify I am asking for changes.

thewilsonator:
IMO your suggestion fall foul of 

> `overlap` exists, but is confusing to use

If I saw `small.overlap(large) is small` it would take me a good while to figure out what that meant.

> As it is, it returns very little information i.e. "is this range overlapping this other range in any position?" with no further information on position.

c.f. `canFind`(/`find`), so what? The index is trivially `whole.ptr - part.ptr` but you'd just use `part.ptr`.

>Given the rarity of such a query, I think we don't need a new function for it.

>@dcarp provided some and I have run into similar uses when working with allocators (it's analog to owns)

The existence of one-liners on phobos exist precisely to a) stop people reinventing the wheel, and b) use a common name: this fit both of those.

Please reconsider.

andralex:
>The existence of one-liners on phobos exist precisely to a) stop people reinventing the wheel, and b) use a common name: this fit both of those.

(c) must be in frequent use, this doesn't pass that test

andralex:
>If I saw small.overlap(large) is small it would take me a good while to figure out what that meant.

"The overlap of small and large is the same as small itself, i.e. large subsumes small." For somebody in a place in life where they need overlap, this is not quite the leap.

thewilsonator:
>For somebody in a place in life where they need overlap, this is not quite the leap.

I agree, but for those not in need of or familiar with `overlap`, `small.overlap(large) is small` is obtuse.

andralex:
@thewilsonator wouldn't disagree. I still think this should be relegated to a documentation example. Thanks.

RazvanN7:
@andralex I have put `isSliceOf` into a documented unittest and have used `overlap` instead of pointer arithmetics. However, I had to delete a few tests regardind the behavior in the presence of `null`. However, I think this is fine since the point is to provide a use case for `overlap`.

LightBender:
@thewilsonator you wanted this one.

LightBender:
@thewilsonator you wanted this one.

std/array.d
```diff
@@ -1297,6 +1297,63 @@ if (is(typeof(a.ptr < b.ptr) == bool))
...
+    assert(isSliceOf(x[3..$], x));
+    assert(isSliceOf(x[], x));
+    assert(!isSliceOf(x, x[3..$]));
+    assert(!isSliceOf([7, 8], x));
+    assert(isSliceOf(null, x));
+
+    // null is a slice of itself
+    assert(isSliceOf(null, null));
+
+    foreach (T; AliasSeq!(int[], const(int)[], immutable(int)[], const int[], immutable int[]))
```
thewilsonator, line 1322:
```
generated/linux/debug/64/publictests/std_array.d(1322): Error: template instance `AliasSeq!(int[], const(int)[], immutable(int)[], const(int[]), immutable(int[]))` template `AliasSeq` is not defined
```

0xEAB: Thanks :)

0xEAB:
I assume, we still want to keep this merely as a documentation enhancement – correct?

thewilsonator:
For the moment yes. We can reconsider later, e.g. phobos 3

PR #6149, opened 2018-02-09
Remove use of undefined ddoc macros from Phobos

wilzbach:
This is required for https://github.com/dlang/dmd/pull/7834
The uses of `_` has been wrong in both cases.

CC @quickfur

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -497,7 +497,7 @@ enum real LOG2 =       0x1.34413509f79fef311f12b35816f92p-2L; /** $(SUB log, 10)
 enum real LOG10E =     0x1.bcb7b1526e50e32a6ab7555f5a67cp-2L; /** $(SUB log, 10)e = 0.434294... */
 enum real LN2 =        0x1.62e42fefa39ef35793c7673007e5fp-1L; /** ln 2  = 0.693147... */
 enum real LN10 =       0x1.26bb1bbb5551582dd4adac5705a61p+1L; /** ln 10 = 2.302585... */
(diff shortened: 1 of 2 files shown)


quickfur:
Whoa, what's with the ddoc test failure?

quickfur:
I don't think you want to use `&pi;` here; it will break LaTeX generation, won't it?

wilzbach:
> Whoa, what's with the ddoc test failure?

Argh damn. We still use the old Ddoc behavior here. I expanded the `PI` macro - an alternative would have been to rename it.

quickfur:
But what about the other lines that reference `$(PI)`? How come they compile, but this line doesn't?

wilzbach:
> it will break LaTeX generation, won't it?

We only generate LaTeX from the spec, not the documentation.
I just did inserted, what Ddoc would have done manually:

https://github.com/dlang/phobos/blob/b3ebd07bbe1f7737350d5397ada63720fb6a0536/std/math.d#L106

But as mentioned I can also rename the macro if you prefer this.

quickfur:
Ohhh I know why... because of the stupid auto-highlight, since `PI` is being defined, it needs to be escaped otherwise it will mangle the macro. :-(

wilzbach:
> But what about the other lines that reference $(PI)? How come they compile, but this line doesn't?

Auto-escaping - the thing you want to change in your PR ;-)

quickfur:
Haha, yeah.  *This* is why we need to kill that misfeature with fire. :-(  I guess we should just go with `&pi;` for now.

wilzbach:
Yes I think so too.
Once you PR is in, we can easily go back to `PI` (though honestly it doesn't matter).

PR #6150, opened 2018-02-09
Issue 8166 - remove dead _backLength code in std.algorithm.splitter

wilzbach:
The implementation already was partially there.
Has anyone a memory why it wasn't enabled?

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -4016,8 +4016,6 @@ if (is(typeof(binaryFun!pred(r.front, s.front)) : bool)
         Separator _separator;
         // _frontLength == size_t.max means empty
         size_t _frontLength = size_t.max;
-        static if (isBidirectionalRange!Range)
-            size_t _backLength = size_t.max;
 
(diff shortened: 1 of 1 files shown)


jmdavis:
Making `splitter` a bidirectional range fundamentally broken - at least if the delimiter has more than one element. Think about something like

   auto result = "*******".splitter("**"); // odd number in the string being split, even number in the delimeter

The range is not going to have the same elements if it's iterated from the back. It will end up being split differently.

wilzbach:
> Making splitter a bidirectional range fundamentally broken - at least if the delimiter has more than one element. Think about something like

Argh. I almost expected something like this, but couldn't think of it. Thanks!
I used this opportunity to remove the old, dead code then. No need to set and store `_backLength` if we never use it.

JackStouffer:
You can still do it if `is(typeof(separator) == ElementType!Range)`

wilzbach:
> You can still do it if is(typeof(separator) == ElementType!Range)

That's a separate overload, which already does so.

PR #6151, opened 2018-02-10
Fix all DScanner warnings + use -return-exit-code re-enable it

wilzbach:
tl;dr: DScanner started to segfaulted in December due to the introduction of the AscendingPageAllocator, which used `~this(){}`. The DScanner bug has been fixed and now we can fix the DScanner warnings that sneaked in and re-enable in.

See also: https://github.com/dlang/phobos/pull/6148

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -535,7 +535,7 @@ style: publictests style_lint
 # runs static code analysis with Dscanner
 dscanner: | $(DSCANNER_DIR)/dsc
 	@echo "Running DScanner"
-	$(DEBUGGER) -q -ex run -ex bt -batch --args $(DSCANNER_DIR)/dsc --config .dscanner.ini --styleCheck etc std -I.
+	$(DEBUGGER) -return-child-result -q -ex run -ex bt -batch --args $(DSCANNER_DIR)/dsc --config .dscanner.ini --styleCheck etc std -I.
 
 style_lint: dscanner $(LIB)
 	@echo "Check for trailing whitespace"
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -4689,6 +4689,47 @@ if (isSomeChar!C)
     }
 }
 
+// In same combinations substitute needs to calculate the auto-decoded length
+// of its needles
+private template hasDifferentAutodecoding(Range, Needles...)
+{
+    import std.meta : anySatisfy;
+    /* iff
+       - the needles needs auto-decoding, but the incoming range doesn't (or vice versa)
+       - both (range, needle) need auto-decoding and don't share the same common type
+    */
+    enum needlesAreNarrow = anySatisfy!(isNarrowString, Needles);
+    enum sourceIsNarrow = isNarrowString!Range;
+    enum hasDifferentAutodecoding = sourceIsNarrow != needlesAreNarrow ||
+                                    (sourceIsNarrow && needlesAreNarrow &&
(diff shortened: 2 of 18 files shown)


std/net/curl.d
```diff
@@ -313,7 +313,7 @@ version(StdUnittest)
 version(StdDdoc) import std.stdio;
 
 // Default data timeout for Protocols
-enum _defaultDataTimeout = dur!"minutes"(2);
+private enum _defaultDataTimeout = dur!"minutes"(2);
```
wilzbach: Undocumented public symbol, it's not too late for this one. It has been only added recently: https://github.com/dlang/phobos/commit/91eaa508c4

std/traits.d
```diff
@@ -5238,7 +5238,8 @@ template IntegralTypeOf(T)
             static assert( is(Q!T == IntegralTypeOf!( SubTypeOf!(Q!T) )));
         }
 
-    static foreach (T; AliasSeq!(void, bool, FloatingPointTypeList, /*ImaginaryTypeList, ComplexTypeList,*/ CharTypeList))
+    static foreach (T; AliasSeq!(void, bool, FloatingPointTypeList,
+                /*ImaginaryTypeList, ComplexTypeList,*/ CharTypeList))
```
wilzbach: Long line

posix.mak
```diff
@@ -535,7 +535,7 @@ style: publictests style_lint
 # runs static code analysis with Dscanner
 dscanner: | $(DSCANNER_DIR)/dsc
 	@echo "Running DScanner"
-	$(DEBUGGER) -q -ex run -ex bt -batch --args $(DSCANNER_DIR)/dsc --config .dscanner.ini --styleCheck etc std -I.
+	$(DEBUGGER) -return-child-result -q -ex run -ex bt -batch --args $(DSCANNER_DIR)/dsc --config .dscanner.ini --styleCheck etc std -I.
```
wilzbach: FYI: DScanner doesn't seem to be super stable, so imho it's better to keep running it within GDB for a while. The overhead isn't too big and we have done so for the last two months.

wilzbach:
image

Yeah, this is finally ready. Sorry for the commit noise.

std/algorithm/iteration.d
```diff
@@ -4689,6 +4689,47 @@ if (isSomeChar!C)
...
+    static assert(hasDifferentAutodecoding!(dstring, AliasSeq!(wstring, wstring)));
+
+    // both (range, needle) need auto-decoding and don't share the same common type
+    static foreach (T; AliasSeq!(string, wstring, dstring))
+    {
+        static assert(hasDifferentAutodecoding!(T, AliasSeq!(wstring, string)));
+        static assert(hasDifferentAutodecoding!(T, AliasSeq!(dstring, string)));
+        static assert(hasDifferentAutodecoding!(T, AliasSeq!(wstring, dstring)));
+    }
+}
```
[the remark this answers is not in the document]
JackStouffer: I don't understand. This symbol is not documented and the unit test is not an example.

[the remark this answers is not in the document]
wilzbach: The warning was for `template substitute`. I don't remember which one, but one of the documentation engines used to dislike different symbols (even if they are private) between declaration and unittest. It looks like this has been fixed, so I can look into updating DScanner... though maybe we can live with this workaround until DScanner has been upgraded?

PR #6152, opened 2018-02-10
Issue 18398 - std.datetime.stopwatch documented examples could be better (x2)

n8sh:
Revision of https://github.com/dlang/phobos/pull/6139 which isn't showing up due to a ddoc bug documented in [Issue 16992](https://issues.dlang.org/show_bug.cgi?id=16992):

>A workaround would be to set the example as module-level, documented unittest.

### DIFF
--- a/std/datetime/stopwatch.d
+++ b/std/datetime/stopwatch.d
@@ -82,55 +82,6 @@ struct StopWatch
 {
 public:
 
-    /// Measure a time in milliseconds, microseconds, or nanoseconds
-    @safe nothrow @nogc unittest
-    {
-        auto sw = StopWatch(AutoStart.no);
-        sw.start();
-        // ... Insert operations to be timed here ...
-        sw.stop();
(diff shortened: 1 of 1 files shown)


wilzbach:
Thanks. I just verified and there are two other potential workaround (sorry for not checking before).

1) move the unittests below the `this` constructor

image

2) move the unittests after the struct

image

Which one do you like the best? And while you are at it, could you make the other unittest visible too? Thanks!

jmdavis:
Putting them after the struct would be better, since they really go with the type, not the module.

n8sh:
@jmdavis OK moved

n8sh:
>And while you are at it, could you make the other unittest visible too? Thanks!

Oops, missed this. Fixed.

>Which one do you like the best?

Both in the source code and in the docs I'd like the example to appear as early as possible. Because I can't have both I favored the docs.

PR #6154, opened 2018-02-10
Fix Issue 12086 - std.algorithm.remove + range of indices produces wrong results

wilzbach:
See the Bugzilla message or the changelog entry for a motivation.

----

Previously, `std.algorithm.remove` used to accept any values as `offset`.
This behavior was very error-prone:

```d
import std.algorithm, std.stdio;
[0, 1, 2, 3, 4].remove(1, 3).writeln; // 0, 2, 4  -- correct
[0, 1, 2, 3, 4].remove([1, 3]).writeln; // 0, 3, 4  -- incorrect
```

With this release, using arrays as individual element is deprecated.
`tuple` can be used to explicitly indicate that a range
from `start` to `stop` (non-enclosing) should be removed:

```d
import std.algorithm, std.stdio, std.typecons;
[0, 1, 2, 3, 4].remove(tuple(1, 3)).writeln; // 0, 3, 4
```

However, only 2-tuples are allowed to avoid this un-intuitive scenario:

```d
import std.algorithm, std.stdio, std.typecons;
[0, 1, 2, 3, 4].remove(tuple(1, 3, 4)).writeln; // 0, 3, 4
```

### DIFF
--- a/changelog/std-algorithm-mutation-remove.dd
+++ b/changelog/std-algorithm-mutation-remove.dd
@@ -1,0 +1,27 @@
+`std.algorithm.mutation.remove` now only accepts integral values or pair of integral values as offset
+
+Previously, without being stated in the documentation,
+$(REF remove, std,algorithm) used to accept any values as `offset`.
+This behavior was very error-prone:
+
+---
+import std.algorithm, std.stdio;
+[0, 1, 2, 3, 4].remove(1, 3).writeln; // 0, 2, 4  -- correct
+[0, 1, 2, 3, 4].remove([1, 3]).writeln; // 0, 3, 4  -- incorrect
+---
+
+With this release, using arrays as individual elements is no longer valid.
+$(REF tuple, std,typecons) can be used to explicitly indicate that a range
+from `start` to `stop` (non-enclosing) should be removed:
+
+---
+import std.algorithm, std.stdio, std.typecons;
+[0, 1, 2, 3, 4].remove(tuple(1, 3)).writeln; // 0, 3, 4
+---
+
+However, only 2-tuples are allowed to avoid this un-intuitive scenario:
+
+---
+import std.algorithm, std.stdio, std.typecons;
+[0, 1, 2, 3, 4].remove(tuple(1, 3, 4)).writeln; // 0, 4?
+---
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -80,6 +80,7 @@ module std.algorithm.mutation;
 import std.range.primitives;
 import std.traits : isArray, isAssignable, isBlitAssignable, isNarrowString,
        Unqual, isSomeChar, isMutable;
+import std.meta : allSatisfy;
 // FIXME
 import std.typecons; // : tuple, Tuple;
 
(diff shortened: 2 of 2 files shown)


wilzbach:
There's also a similar discussion about renaming `remove` here: https://issues.dlang.org/show_bug.cgi?id=10959

Quotes:

> Currently std.algorithm.remove is a landmine
..
> It's a landmine, and it's bit me a few times in the past too.

Note that this PR just deprecates the "landmine" input. Ideally renaming to `removeIndex` would be best, but that's not something this PR proposes. Let's fix the landmines first!

std/algorithm/mutation.d
```diff
@@ -1695,6 +1696,59 @@ enum SwapStrategy
...
+        else
+        {
+            enum isIntegralPair = false;
+        }
+    }
+    else
+    {
+        enum isIntegralPair = false;
+    }
+}
```
dukc: On the bottom of this statement you use single-statement curly brace block for static else. But at line 1713 you newline after it without curly braces, a style considered error-prone by some. Should you change these to be consistent with each other? On the other hand, this might be an over-pedantic suggestion.

wilzbach: The current convention is that curly braces only need to be used for more than one line, i.e. when they are required.
I don't mind changing but to my knowledge always requiring curly braces isn't part of the DStyle (as of now).

jmdavis: And please don't make it part of the style. That would be incredibly annoying. Personally, I always use them when the condition is longer than one line, since it gets hard to read otherwise, but other than that, I never use braces when their contents are just one line long, and many other contributors to Phobos don't use the pointless braces either. They just add unnecessary vertical space to the file.

dukc: I did not mean that lines like that should have the curly braces always, but that they should (maybe) be consistent with each other in the same function. Meaning, either use or don't use curly braces for if or else clause at one line and one statement at another line, but don't do both in the same function.

wilzbach:
Looks like no one wants to approve a deprecation without @andralex's approval :/

changelog/std-algorithm-mutation-remove.dd
```diff
@@ -0,0 +1,26 @@
...
+Previously, $(REF remove, std,algorithm) used to accept any values as `offset`.
+This behavior was very error-prone:
+
+---
+import std.algorithm, std.stdio;
+[0, 1, 2, 3, 4].remove(1, 3).writeln; // 0, 2, 4  -- correct
+[0, 1, 2, 3, 4].remove([1, 3]).writeln; // 0, 3, 4  -- incorrect
+---
+
+With this release, using arrays as individual element is deprecated.
```
andralex: s/element/elements/

std/algorithm/mutation.d
```diff
@@ -1695,6 +1696,59 @@ enum SwapStrategy
     assert(arr == [10, 9, 8, 4, 5, 6, 7, 3, 2, 1]);
 }
 
+/**
+Checks whether a type is either an integral static array (e.g. int[2])
+or a tuple of two integral types
+*/
+private template isIntegralPair(T)
```
andralex: That's a lot of work! Just do this inline:

```d
    static if (is(T == Tuple!(U1, U2), U1, U2))
        enum isIntegralPair = isIntegral!U1 && isIntegral!U2;
    else
        enum isIntegralPair = false;
}
```

We don't want to support static arrays.

No need even to define a separate template, just do this in the body:

`static assert(is(T == Tuple!(U1, U2), U1, U2) && isIntegral!(T[0]) && isIntegral!(T[1]),
    "Each offset must be an integral or a tuple of two integrals");`

std/algorithm/mutation.d
```diff
@@ -1788,15 +1842,30 @@ Params:
...
 (SwapStrategy s = SwapStrategy.stable, Range, Offset...)
 (Range range, Offset offset)
-if (s != SwapStrategy.stable
-    && isBidirectionalRange!Range
+if (isBidirectionalRange!Range
     && hasLvalueElements!Range
     && hasLength!Range
-    && Offset.length >= 1)
+    && Offset.length >= 1
+    && allSatisfy!(templateOr!(isIntegral, isIntegralPair),  Offset))
```
andralex: No private symbols in a public constraint - how would the user ever figure what `isIntegralPair` does?

wilzbach:
Okay I added the deprecation warnings back in & rebased this to master. Should be good to go now.
Maybe it can still be part of 2.081?

PR #6155, opened 2018-02-10
Fix Issue 5502 - More handy ways to create associative arrays

wilzbach:
I'm not entirely sure about this one.
One one side, it's trivial to add this and it falls in the category of: "We don't add one-liners to Phobos".
OTOH the symbol already exists and it feels more natural for it, to accept two arguments too.

What do you think?

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -410,16 +410,25 @@ if (isNarrowString!String)
 }
 
 /**
-Returns a newly allocated associative array from a range of key/value tuples.
+Returns a newly allocated associative array from a range of key/value tuples
+or from a range of keys and a range of values.
 
 Params:
     r = An $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
     of tuples of keys and values.
-Returns: A newly allocated associative array out of elements of the input
-range, which must be a range of tuples (Key, Value). Returns a null associative
-array reference when given an empty range.
-Duplicates: Associative arrays have unique keys. If r contains duplicate keys,
-then the result will contain the value of the last pair for that key in r.
+    keys = An $(REF_ALTTEXT input range, isInputRange, std,range,primitives) of keys
+    values = An $(REF_ALTTEXT input range, isInputRange, std,range,primitives) of values
+
+Returns:
+
+    A newly allocated associative array out of elements of the input
+    range, which must be a range of tuples (Key, Value) or
+    a range of keys and a range of values. If given two ranges of unequal
+    lengths after the elements of the shorter are exhausted the remaining
+    elements of the longer will not be considered.
+    Returns a null associative array reference when given an empty range.
+    Duplicates: Associative arrays have unique keys. If r contains duplicate keys,
+    then the result will contain the value of the last pair for that key in r.
 
 See_Also: $(REF Tuple, std,typecons), $(REF zip, std,range)
  */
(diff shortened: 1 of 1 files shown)


n8sh:
Is there no way to presize an associative array?

wilzbach:
> Is there no way to presize an associative array?

See: https://github.com/dlang/druntime/pull/1929

> Could this use _d_assocarrayliteralTX when both inputs are dynamic arrays?

Yeah, `zip` is very naive, but the issue was about the API. The implementation can always be tweaked later.

n8sh:
How about this? At minimum it means that this function no longer falls afoul of "we don't add one-liners to Phobos."

```d
auto assocArray(Keys, Values)(Keys keys, Values values)
if (isInputRange!Values && isInputRange!Keys)
{
    static if (isDynamicArray!Keys && isDynamicArray!Values)
    {
        void* aa = aaLiteral(keys, values);
        alias Key = typeof(keys[0]);
        alias Value = typeof(values[0]);
        return (() @trusted => cast(Value[Key]) aa)();
    }
    else
    {
        import std.range : zip;
        return zip(keys, values).assocArray;
    }
}
```

wilzbach:
> How about this? 

Nice! I thought this would be more complex and delayed the investigation until this gets approval.
Thanks a lot!

JackStouffer:
Even with the dynamic array specialization, I think there's not enough usefulness gained from this. In most cases all you're doing for the user is hiding a call to `zip`.

I think the right path forward here is to add an example showing how to combine two ranges with `zip` and then pass it to `assocArray`.

n8sh:
>In most cases all you're doing for the user is hiding a call to `zip`.

Because `zip` isn't yet `@safe` (https://github.com/dlang/phobos/pull/6254), there could be value in hiding a verbose substitute for `zip`:

```d
    else
    {
        alias Key = ElementType!Keys;
        alias Value = ElementType!Values;
        static assert(isMutable!Value, "assocArray: value type must be mutable");
        Value[Key] aa;
        foreach (key; keys)
        {
            if (values.empty) break;
            aa[key] = values.front;
            values.popFront();
        }
        return aa;
    }
```

wilzbach:
> Even with the dynamic array specialization, I think there's not enough usefulness gained from this. In most cases all you're doing for the user is hiding a call to zip.

Well, as it turns out, it's a bit more complicated than just `zip` if you want it to be `@safe` and `nothrow`.
Have a look at the current implementation - it's more than 60 lines.

Also note that the symbol already exists and we just add an overload to it.

std/array.d
```diff
@@ -386,11 +386,17 @@ Returns a newly allocated associative _array from a range of key/value tuples.
...
-Duplicates: Associative arrays have unique keys. If r contains duplicate keys,
-then the result will contain the value of the last pair for that key in r.
+    keys = An $(REF_ALTTEXT input range, isInputRange, std,range,primitives) of keys
+    values = An $(REF_ALTTEXT input range, isInputRange, std,range,primitives) of values
+
+Returns:
+
+    A newly allocated associative array out of elements of the input
+    range, which must be a range of tuples (Key, Value) or
+    a range of keys and a range of tuples.
```
n8sh: Do you mean "a range of keys and a range of values"?

std/array.d
```diff
@@ -413,18 +419,93 @@ if (isInputRange!Range)
...
+    else
+    {
+        // zip can't infer nothrow
+        alias Key = ElementType!Keys;
+        alias Value = ElementType!Values;
+        static assert(isMutable!Value, "assocArray: value type must be mutable");
+        Value[Key] aa;
+        foreach (key; keys)
+        {
+            if (values.empty) break;
```
JackStouffer: This behavior has to be documented. You might want to consider using something like `StoppingPolicy` to deal with different length ranges.

std/array.d
```diff
@@ -413,18 +419,93 @@ if (isInputRange!Range)
...
+                    KeyElement.init.__xdtor();
+
+                alias ValueElement = ElementType!Values;
+                static if (hasElaborateDestructor!ValueElement)
+                    ValueElement.init.__xdtor();
+            })))
+            {
+                scope(failure) assert(0);
+            }
+            aa = aaLiteral(keys, values);
```
n8sh: If allowing ranges of unequal length, above this should go code along the lines of:
```d
if (keys.length > values.length)
    keys = keys[0 .. values.length];
else if (values.length > keys.length)
    values = values[0 .. keys.length;
```
to avoid an assertion error from `aaLiteral`.

DmitryOlshansky:
> 'm not entirely sure about this one.
> One one side, it's trivial to add this and it falls in the category of: "We don't add one-liners to Phobos".
> OTOH the symbol already exists and it feels more natural for it, to accept two arguments too.
> 
> What do you think?

I guess adding overloads to existing symbol that hide some boilerplate to use the original is perfectly fine. New symbol for this thing - certainly no.

std/array.d
```diff
@@ -413,18 +419,93 @@ if (isInputRange!Range)
...
+        }
+        alias Key = typeof(keys[0]);
+        alias Value = typeof(values[0]);
+        return (() @trusted => cast(Value[Key]) aa)();
+    }
+    else
+    {
+        // zip can't infer nothrow
```
n8sh: Noting here that `zip` can infer `nothrow` as of https://github.com/dlang/phobos/pull/6307.

std/array.d
n8sh: Documentation added.

std/array.d
```diff
@@ -442,18 +451,97 @@ if (isInputRange!Range)
     return aa;
 }
 
+/// ditto
+auto assocArray(Keys, Values)(Keys keys, Values values)
+if (isInputRange!Values && isInputRange!Keys)
+{
+    static if (isDynamicArray!Keys && isDynamicArray!Values
+        && !isNarrowString!Keys && !isNarrowString!Values)
```
n8sh: Added `&& !isNarrowString!Keys && !isNarrowString!Values` for consistency with other Phobos functions that treat strings as ranges.

PR #6156, opened 2018-02-10
Fix Issue 8341 - topN(zip()) doesn't work

wilzbach:
`topN` requires an assignable range. While fixing the constraint of `topN`, I found that almost the entire module had missing constraints.
This should improve the error messages slightly, i.e. now they aren't in `std.algorithm` anymore, but the user at least gets "no matching function found" and sees the constraints.

### DIFF
--- a/std/algorithm/sorting.d
+++ b/std/algorithm/sorting.d
@@ -112,8 +112,9 @@ Params:
         sorted.
 */
 void completeSort(alias less = "a < b", SwapStrategy ss = SwapStrategy.unstable,
-        RandomAccessRange1, RandomAccessRange2)(SortedRange!(RandomAccessRange1, less) lhs, RandomAccessRange2 rhs)
-if (hasLength!(RandomAccessRange2) && hasSlicing!(RandomAccessRange2))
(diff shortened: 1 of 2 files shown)


std/algorithm/sorting.d
```diff
@@ -3100,7 +3102,8 @@ Stable topN has not been implemented yet.
 auto topN(alias less = "a < b",
         SwapStrategy ss = SwapStrategy.unstable,
         Range)(Range r, size_t nth)
-if (isRandomAccessRange!(Range) && hasLength!Range && hasSlicing!Range)
+if (isRandomAccessRange!(Range) && hasLength!Range &&
+    hasSlicing!Range && hasAssignableElements!Range)
```
JackStouffer: Shouldn't this be hasSwappableElements?

wilzbach: No it calls `pivotPartition` which calls `moveAt` and assigns to the element:

```d
auto save = r.moveAt(hi);
...
r[lo] = save;
```

`hasSwappableElements` just requires that `swap(a, b)` works.

PR #6157, opened 2018-02-11
Fix Issue 17982 - Support for const(Class) in algorithm.searching.extremum

wilzbach:


std/algorithm/searching.d
```diff
@@ -1466,6 +1476,25 @@ private auto extremum(alias selector = "a < b", Range,
...
+    }
+
+    const(B) doStuff(const(B)[] v)
+    {
+        return v.extremum!"a.val";
+    }
+
+    const(B)[] arr = [new B(0), new B(1)];
+    // can't compare directly - https://issues.dlang.org/show_bug.cgi?id=1824
+    assert(arr.extremum!"a.val".val == 0);
```
wilzbach: Imho this is the real bummer - `arr.extremum!"a.val" == new B(0)` doesn't work :/

std/algorithm/searching.d
```diff
@@ -3456,6 +3485,38 @@ if (isInputRange!Range && !isInfinite!Range &&
...
 
+// https://issues.dlang.org/show_bug.cgi?id=17982
+@safe unittest
+{
+    struct A
+    {
+      int val;
+    }
+
+    import std.stdio;
```
JackStouffer: remove

std/algorithm/searching.d
```diff
@@ -1470,6 +1480,25 @@ private auto extremum(alias selector = "a < b", Range,
...
+// https://issues.dlang.org/show_bug.cgi?id=17982
+@safe unittest
+{
+    class B
+    {
+        int val;
+        this(int val){ this.val = val; }
+    }
+
+    const(B) doStuff(const(B)[] v)
```
JackStouffer: This is never called

wilzbach: It doesn't need to be called (it's a copy from the bug report).
Anyhow, I added an call of `doStuff` for "more coverage".

PR #6159, opened 2018-02-11
Fix Issue 18419 - make all Phobos unittests version(StdUnittest)

n8sh:
This is the purpose for which `version(StdUnittest)` was added. The pain is one-time, the benefit is forever. It is also a good practice for other library authors to copy.

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -116,7 +116,7 @@ if (isExpressionTuple!values)
 }
 
 ///
-@safe unittest
+version(StdUnittest) @safe unittest
 {
     assert(3.among(1, 42, 24, 3, 2));
 
@@ -133,13 +133,13 @@ if (isExpressionTuple!values)
 Alternatively, `values` can be passed at compile-time, allowing for a more
 efficient search, but one that only supports matching on equality:
 */
-@safe unittest
+version(StdUnittest) @safe unittest
 {
     assert(3.among!(2, 3, 4));
     assert("bar".among!("foo", "bar", "baz") == 2);
 }
 
-@safe unittest
+version(StdUnittest) @safe unittest
 {
     import std.meta : AliasSeq;
 
@@ -373,7 +373,7 @@ auto castSwitch(choices...)(Object switchObject)
 }
 
 ///
-@system unittest
+version(StdUnittest) @system unittest
 {
     import std.algorithm.iteration : map;
     import std.format : format;
@@ -404,7 +404,7 @@ auto castSwitch(choices...)(Object switchObject)
 }
 
 /// Using with void handlers:
-@system unittest
+version(StdUnittest) @system unittest
 {
     import std.exception : assertThrown;
 
@@ -425,7 +425,7 @@ auto castSwitch(choices...)(Object switchObject)
     )();
 }
 
-@system unittest
+version(StdUnittest) @system unittest
 {
     import core.exception : SwitchError;
     import std.exception : assertThrown;
@@ -498,7 +498,7 @@ auto castSwitch(choices...)(Object switchObject)
                            )());
 }
 
-@system unittest
+version(StdUnittest) @system unittest
 {
     interface I { }
     class B : I { }
@@ -541,7 +541,7 @@ do
 }
 
 ///
-@safe unittest
+version(StdUnittest) @safe unittest
 {
     assert(clamp(2, 1, 3) == 2);
     assert(clamp(0, 1, 3) == 1);
@@ -552,7 +552,7 @@ do
     assert(clamp(5, -1, 2u) == 2);
 }
 
-@safe unittest
+version(StdUnittest) @safe unittest
 {
     int a = 1;
     short b = 6;
@@ -752,7 +752,7 @@ if (isInputRange!R1 && isInputRange!R2)
 }
 
 ///
-pure @safe unittest
+version(StdUnittest) pure @safe unittest
 {
     int result;
 
@@ -789,7 +789,7 @@ pure @safe unittest
 }
 
 /// Example predicate that compares individual elements in reverse lexical order
-pure @safe unittest
+version(StdUnittest) pure @safe unittest
 {
     int result;
 
@@ -825,7 +825,7 @@ pure @safe unittest
     assert(result > 0);
 }
 
-@nogc nothrow pure @safe unittest
+version(StdUnittest) @nogc nothrow pure @safe unittest
 {
     // Issue 18286: cmp for string with custom predicate fails if distinct chars can compare equal
     static bool ltCi(dchar a, dchar b)// less than, case insensitive
@@ -840,7 +840,7 @@ pure @safe unittest
     static assert(cmp!ltCi("apple", "APPLE") == 0);
 }
 
-@nogc nothrow @safe unittest
+version(StdUnittest) @nogc nothrow @safe unittest
 {
     // Issue 18280: for non-string ranges check that opCmp is evaluated only once per pair.
     static int ctr = 0;
@@ -859,7 +859,7 @@ pure @safe unittest
     assert(ctr == a.length, "opCmp should be called exactly once per pair of items!");
 }
 
-nothrow pure @safe unittest
+version(StdUnittest) nothrow pure @safe unittest
 {
     // Test cmp when opCmp returns float.
     struct F
@@ -881,7 +881,7 @@ nothrow pure @safe unittest
     assert(result > 0);
 }
 
-nothrow pure @safe unittest
+version(StdUnittest) nothrow pure @safe unittest
 {
     // Parallelism (was broken by inferred return type "immutable int")
     import std.parallelism : task;
@@ -984,7 +984,7 @@ template equal(alias pred = "a == b")
 }
 
 ///
-@safe unittest
+version(StdUnittest) @safe unittest
 {
     import std.algorithm.comparison : equal;
     import std.math : approxEqual;
@@ -1010,7 +1010,7 @@ This can be very useful when the element type of a range is itself a
 range. In particular, `equal` can be its own predicate, allowing
 range of range (of range...) comparisons.
  +/
-@safe unittest
+version(StdUnittest) @safe unittest
 {
     import std.algorithm.comparison : equal;
     import std.range : iota, chunks;
@@ -1020,7 +1020,7 @@ range of range (of range...) comparisons.
     ));
 }
 
-@safe unittest
+version(StdUnittest) @safe unittest
 {
     import std.algorithm.iteration : map;
     import std.internal.test.dummyrange : ReferenceForwardRange,
@@ -1084,7 +1084,7 @@ range of range (of range...) comparisons.
     assert(!equal(cir, ifr));
 }
 
-@safe pure unittest
+version(StdUnittest) @safe pure unittest
 {
     import std.utf : byChar, byWchar, byDchar;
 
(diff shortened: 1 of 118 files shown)


jmdavis:
IMHO, if we need to do this, there is something very wrong about how D handles unit tests.

CyberShadow:
Indeed. This change would fix a very real problem, but ... perhaps this is better fixed elsewhere.

n8sh:
This won't interfere with someone making unittest-related improvements to other parts of the D toolchain. It may also take some of the pressure off of whoever is working on that, so they can take their time to get the design right instead of feeling like they have to rush something out the door.

wilzbach:
For reference, the discussion is here: https://github.com/dlang/phobos/pull/6142

> This change would fix a very real problem, but ... perhaps this is better fixed elsewhere.

Ok, alternatives:

- revert `StdUnittest` for the next release
- avoid emitting `unittest` blocks with `-deps` (or potentially other flags)

Better ideas?

---

@FeepingCreature - any chance we can add a unittest for your `-deps` issue? Or maybe you can simple add your project to the Project Tester?

adamdruppe:
so now when you build the docs with dmd, you need -version=StdUnittest yes?

but i say don't let the perfect solution be the enemy of a workable stopgap. we can always fix root cause later and revert this at that time

FeepingCreature:
> @FeepingCreature - any chance we can add a unittest for your -deps issue? Or maybe you can simple add your project to the Project Tester?

When you say "your project", I think you still underestimate the scale of the problem.

This is what it takes on master, to cause a failing build:

    echo "import std.stdio; void main() { }" > test.d
    dmd -unittest -deps test.d

n8sh:
@FeepingCreature 
>Edit: tried that command line with this pr, and it seems to fix the issue, though it results in a possibly unrelated linker error.

I ran your script in the master branch, except without `-unittest -deps`.  I got what I presume is the same linker error as you, so it's not related to this pull request.

>Undefined symbols for architecture x86_64:
  "_D3std12experimental9scripting12__ModuleInfoZ", referenced from:
      _D4test12__ModuleInfoZ in test.o

wilzbach:
You need to use the freshly compiled Phobos library when doing _anything_ with master - so yes the "error" is unrelated.

jmdavis:
Man. I was just forced to do this sort of thing with dxml, because folks where getting weird linker errors when they built the unit tests for their own projects which used dxml. For some reason, not all of the symbols used by dxml's unit tests were available - though dub seems to take this a step stupider than dmd does on its own. dub seems to build the actual unittest target for all dependencies not just have problems with the imports being affected by the -unittest flag. This whole situation is a mess.

wilzbach:
> This whole situation is a mess.

Agreed and it looks like we need to make a decision pretty soon (the master branch off will happen is supposed to happen in two days).

JackStouffer:
Let's get an executive decision on this from @andralex and move on.

andralex:
As a simple consideration, any change that adds an adornment to 4000+ lines of code in a 300 KLOC project should be approached with suspicion that there probably are 10 wrong lines of code someplace else. It's nice to see that other commenters noticed that as well.

Neither the bug nor the PR lists the problem that this change is addressing, so I'll rely on the info in https://github.com/dlang/phobos/pull/6142/ and links from there. Far as I understand:

* We added `version(StdUnittest)` because we wanted certain things (e.g.imports, test-specific global data, unittests inside templates) to only be compiled when Phobos was being unittested, not the user code importing Phobos. For that, `version(unittest)` was deemed insufficient because it is active during user-level unittesting.
* Once `StdUnittest` started getting used (especially in `import`s), code in template unittests got out of sync with the versioned imports.
* There's also a DIP82 proposing a mechanism for preventing unittests in templates from being run.

There'd be a long discussion, but in short:

* Unittests in templates are a good feature because they test a template in conjunction with the user-defined types that it uses. (We've actually gotten a kudos from Matthias Felleisen himself for that.)
* In the rare cases when they shall be disabled, we already have a number of mechanisms for that (`version`, `static if`...) A language feature is not necessary; DIP82's motivation is tenuous.
* We should use `StdUnittest` with extreme caution and almost never for guarding imports (instead, imports should be moved down into the unittests needing them). At present there are 18 uses of `StdUnittest` in Phobos. If we get on them, perhaps we conclude there are 18 too many and get rid of it. That would be a great improvement.
* Please exercise good engineering sense when approaching problems. This huge change is not a one-time pain, it's an ongoing pain due to redundancy and visual noise.

CyberShadow:
> Once StdUnittest started getting used (especially in imports), code in template unittests got out of sync with the versioned imports.

That wasn't what I understood the main problem to be. My understanding was that Phobos unittests participate in building user code when `-unittest` is enabled, and that impacts compilation time. Sometimes the impact is egregious (e.g. with `std.regex`) and reaches the point where we need a mechanism to prevent Phobos unittests from being compiled at all, except when we're explicitly testing Phobos itself.

To demonstrate:

```bash
$ cat test.d
import std.regex;

$ time dmd -c test.d          
dmd -c test.d  0.21s user 0.01s system 98% cpu 0.226 total

$ time dmd -c -unittest test.d
dmd -c -unittest test.d  0.76s user 0.10s system 99% cpu 0.863 total
```

n8sh:
>Unittests in templates are a good feature because they test a template in conjunction with the user-defined types that it uses. (We've actually gotten a kudos from Matthias Felleisen himself for that.)

A benefit of using `version(StdUnittest)` instead of a change to dub or dmd is that we can selectively leave tests that like enabled.

quickfur:
@andralex:

On the whole, I agree that `version(StdUnittest)` is a crutch, and an ugly one, and that we should exercise caution in taking that route. OTOH, this to me is a sign that possibly a language enhancement is necessary.

*Please* let's not confuse this issue with template unittests, even though the two are related.  Your argument that template unittests being run in conjunction with user types is a good thing, is actually misrepresenting the real situation.  Yes, in the cases where the unittest *actually* tests the template with the given user types, this is something beneficial. However, the majority of unittests in Phobos are of the kind:
````
unittest
{
    auto testData = [ 1, 2, 3 ];
    auto range = templateBeingTested(testData);
    assert(range.equals(expectedOutput));
}
````
Most of the time, such unittests are put inside the template because of the ddoc'd unittest feature: it automatically turns into a code example attached to the template's docs, which is a wonderful thing, because that ensures that code examples in the docs are actually tested by the autotester and verified to be runnable (and have correct results).

However, since this is inside a template, that means the *same* unittest, testing the *same* type of test data, is instantiated *n* times, where *n* is the number of different types user code might call it with. That's `n-1` redundant unittest instantiations, and *in user code*.  Does a user project *really* care that `templateBeingTested` works correctly for an array of ints? Of course not -- the autotester has already tested and verified that this case works correctly, and for *all* supported OS/platforms.  None of the benefits you mentioned apply here: the unittest doesn't even instantiate the template with any user type; it *always* tests the template with an array of ints.  Worse yet, it redundantly repeats this test *n* times, as if once on the autotester wasn't already enough to verify that the template works for this particular case.  And to put this into more perspective: this is done *for every user project out there* that uses this template.  So we're essentially making D users pay for this `n*numDUsers` times, where once on the autotester ought to be enough already.

And this isn't even the issue we're trying to address here. The issue here is that some Phobos unittests are very expensive to run, like those in `std.regex`.  As @DmitryOlshansky has already pointed out, it causes a degradation in compilation time of *an order of magnitude*, and for what?  For running unittests that the D autotester has already run to death however many times it does every single time a PR is merged into Phobos.  Do these unittests actually add more value?  No.  All the user did was to `import std.regex` -- he hasn't even begun using the module yet, and already he's paying for (1) decreased compilation time (so much for "fast code fast"), (2) unittest bloat in his executable, (3) runtime penalty of executing Phobos tests that, ostensibly, has already been tested to death by the D autotester, and besides isn't even the responsibility of *user code* to test.

And this goes beyond just Phobos alone.  The way unittests are currently implemented, the same behaviour happens for *all* third party libraries that user code may import.  If some library author A ships a D library L that contains unittests, then *every single time* some downstream user U imports L and compiles his code with `-unittest`, he has to pay for decreased compilation time, unittest bloat, and runtime penalty for running tests that ought to have been the responsibility of library author A instead.  This seems to be an unreasonable cost when user U's intention is only "I want to run *my own* unittests in my *own* code". Currently, there is no way for him to do this -- either he doesn't compile with `-unittest` at all, or he compiles with `-unittest` and the very act of importing a module itself already causes library L's unittests to be compiled into his code -- unittests that he doesn't even care about because they don't concern his own code.

tl;dr: there needs to be a way for upstream library authors (including Phobos) to make it so that users compiling with `-unittest` don't have to pay for unittests that ostensibly said upstream library authors ought to already have run.  If an in-library solution (like this PR) is not acceptable, then surely a language enhancement is in order.

FeepingCreature:
Specifically, the problem I encountered is that Phobos unittests participate in user code **when `-unittest -deps` is enabled**, because the compiler *has* to semantically evaluate the bodies of unittests in order to determine whether they contain imports. As such, moving imports into unittests will still leave me with horrendous compile times and memory usage (ten gigabytes for hello world) with this combination of flags, because DMD is not smart enough to determine that it does not need to output dependency information for modules in an external library, and does not have any mechanism to inform it that dependencies from non-templated Phobos unittests are irrelevant to the program being compiled.

andralex:
@FeepingCreature can the ddoc unittests be versioned with `version(StdDdoc)`?

andralex:
@CyberShadow I agree the increase of build time in unittests is unpleasant. It seems the right long-term solution to that is making the `-unittest` flag more modular. Another possibility is curate Phobos imported files upon distribution. I don't think this solution is the right response.

andralex:
>tl;dr: there needs to be a way for upstream library authors (including Phobos) to make it so that users compiling with -unittest don't have to pay for unittests that ostensibly said upstream library authors ought to already have run. If an in-library solution (like this PR) is not acceptable, then surely a language enhancement is in order.

I agree. I think we need to experiment with making `-unittest` modular. I assume we'll run into issues, the question is how bad they are.

quickfur:
Actually, I already have an idea that's quite simple (barring unforeseen side-effects), that should address most, if not all, of the concerns raised here:

Basically, change the compiler so that `-unittest` only emits unittests that are either inside a template, or found in modules that are currently being compiled on the command-line.  I think something similar was done last year or the year before, where certain templates won't actually instantiate until you actually compile the module they're found in (sorry, I don't remember exactly what was done, so I'm a bit hazy on the details).

Presumably, any unittests that user code might be concerned about would be contained in a module that eventually would be passed to the compiler on the command-line, so they will end up being instantiated *somewhere*.  Third-party libraries (or Phobos itself) would be separately compiled, and presumably distributed as binaries, so presumably they won't end up on the command-line (at least the .d files won't), so library-specific unittests will be automatically left out.

Perhaps I should reserve some time to write up a DIP about this.

CyberShadow:
@quickfur Nice idea, though it won't work for e.g. `targetType "sourceLibrary"` Dub libraries, or libraries used with `rdmd`.

JackStouffer:
I think the best option is to revert StdUnittest. I don't think adding in this cruft/workaround is a viable alternative to fixing the compiler in this area.

I'm going to go head and close this, and open a new PR on stable that does that. I think this conversation has been stalled for too long, and we need to fix this before we ship code with StdUnittest in it.

JackStouffer:
https://github.com/dlang/phobos/pull/6202

PR #6160, opened 2018-02-11
posix.mak: Enforce whitespace before opening version parenthesis

ibuclaw:
Get everyone using the same style.

### DIFF
--- a/etc/c/odbc/sqltypes.d
+++ b/etc/c/odbc/sqltypes.d
@@ -175,11 +175,11 @@ alias GUID     = SQLGUID;
 alias BOOKMARK = uint;
 alias SQLWCHAR = ushort;
 
-version( UNICODE )
+version ( UNICODE )
 {
 alias SQLTCHAR = SQLWCHAR;
 }
 else
 {
 alias SQLTCHAR = SQLCHAR;
-} // end version( UNICODE )
+} // end version ( UNICODE )
--- a/etc/c/curl.d
+++ b/etc/c/curl.d
@@ -115,12 +115,12 @@ alias CURL = void;
 alias curl_socket_t = socket_t;
 
 /// jdrewsen - Would like to get socket error constant from std.socket by it is private atm.
-version(Windows)
+version (Windows)
 {
   import core.sys.windows.windows, core.sys.windows.winsock2;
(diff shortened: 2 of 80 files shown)


n8sh:
If this is being added to posix.max there should also be a line about it in https://github.com/dlang/dlang.org/blob/master/dstyle.dd. I recently checked it for guidance on `version` but couldn't find any.

etc/c/odbc/sqltypes.d
```diff
@@ -175,11 +175,11 @@ alias GUID     = SQLGUID;
 alias BOOKMARK = uint;
 alias SQLWCHAR = ushort;
 
-version( UNICODE )
+version ( UNICODE )
```
wilzbach: Could we use this opportunity to normalize these too?

wilzbach:
I'm all in favor of being consistent in the style, but may I ask why you picked this one?

```
> ~/dlang/phobos on master ± ag "version\(" | wc -l
714

> ~/dlang/phobos on master ± ag "version \(" | wc -l
545
```

ibuclaw:
> I'm all in favor of being consistent in the style, but may I ask why you picked this one?

You mean `version` style?  Because it is a heavily used feature that has at least three different styles, frequently intermixed in the same condition statement!

wilzbach:
> You mean version style? 

No I meant why `version(` and not `version )` - see the output of the grep command.

ibuclaw:
Ah, that would mean having its own regexp to catch that particular style.

I could use the same argument against the following also:
```
if ( cond )
switch ( cond )
while ( cond )
foreach ( var; array )
```

But to be really precise about this, we may need something smarter, as you'd also want to catch e.g:
```
if ( cond1 ||
     cond2 )
```

wilzbach:
Yeah, you want this to be a D-Scanner check, e.g.

https://github.com/dlang-community/D-Scanner/pull/450
https://github.com/dlang-community/D-Scanner/pull/348

However, I never managed to push these into D-Scanner (there were some controversies on how D-Scanner can access the tokens), but we now have a special branch `phobos` there into which we can push anything (even things Brian objects to).

n8sh:
(superseded by https://github.com/dlang/phobos/pull/6717)

PR #6161, opened 2018-02-11
Added examples for AscendingPageAllocator

jercaianu:
- moved both the StdDdoc and implementation under the struct definition
- added ddoc examples
- updated the paper details.


wilzbach:
Thanks a lot! How about only having the mixined methods in the `StdDdoc` block?
In other words, move the documentation for `empty`, `allocate`, `expand` etc. to the existing method?

jercaianu:
@wilzbach 
Thanks, moved the docs as you suggested

PR #6162, opened 2018-02-11
Fix Issue 11555 - std.algorithm.reverse should return the just-reversed range

wilzbach:


wilzbach:
I also used the opportunity to unify the first two overloads of `reverse` into one.

PR #6163, opened 2018-02-11
[Trivial] Should use '$(MAKE)' instead of 'make', inside posix.mak Makefile

dkgroot:
We cannot assume the name of the makefile processor has to be 'make', for example on freebsd, gnu make is called gmake.


wilzbach:
Urgh - dlang-bot seems to be running into API limits with the GitHub API :/

CyberShadow:
> @dkgroot dkgroot changed the title from [Trivial] Should use instead of make inside a Makefile to [Trivial] Should use instead $(MAKE) of make inside Makefile

In the future, please update the commit message as well.

dkgroot:
@CyberShadow Sorry about that !
Thanks for review and merge guys !

PR #6168, opened 2018-02-13
Make VariantN infer dtor safety from types

edi33416:


wilzbach:
(DScanner failure is unrelated. Fix: https://github.com/dlang/phobos/pull/6171)

std/variant.d
```diff
@@ -586,7 +586,9 @@ public:
     {
         ~this()
         {
-            fptr(OpID.destruct, &store, null);
+            // Infer the safety of the provided types
+            static if (AllowedTypes.length) { if (false) AllowedTypes var; }
```
wilzbach: - Typically `if (0)` is used at Phobos.
- Allman!

PR #6173, opened 2018-02-14
Work On Issue 10828 - Add Writer Versions of TimeZone toStrings

JackStouffer:
Rather straight-forward. Will be used in the SysTime versions.

### DIFF
--- a/std/datetime/timezone.d
+++ b/std/datetime/timezone.d
@@ -1460,16 +1460,32 @@ package:
       +/
     static string toISOString(Duration utcOffset) @safe pure
     {
+        import std.array : appender;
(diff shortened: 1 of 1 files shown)


std/datetime/timezone.d
```diff
@@ -1514,9 +1530,19 @@ package:
                         west).
       +/
     static string toISOExtString(Duration utcOffset) @safe pure
+    {
```
wilzbach: Error - file 'std/datetime/timezone.d' contains trailing whitespace at line 1533
posix.mak:514: recipe for target 'checkwhitespace' failed

PR #6175, opened 2018-02-14
Added SharedRegion

jercaianu:
Implemented the threadsafe version of the Region allocator, SharedRegion.
Allocations are lock-free, using the cas mechanism from core.atomic.

### DIFF
--- a/std/experimental/allocator/building_blocks/region.d
+++ b/std/experimental/allocator/building_blocks/region.d
@@ -20,7 +20,7 @@ the store and the limits. One allocation entails rounding up the allocation
 size for alignment purposes, bumping the current pointer, and comparing it
 against the limit.
 
-If $(D ParentAllocator) is different from $(D NullAllocator), $(D Region)
+If $(D ParentAllocator) is different from $(REF_ALTTEXT `NullAllocator`, NullAllocator, std,experimental,allocator,building_blocks,null_allocator), $(D Region)
 deallocates the chunk of memory during destruction.
 
 The $(D minAlign) parameter establishes alignment. If $(D minAlign > 1), the
@@ -53,42 +53,49 @@ struct Region(ParentAllocator = NullAllocator,
     {
         alias parent = ParentAllocator.instance;
     }
+
     private void* _current, _begin, _end;
 
+    pure nothrow @trusted @nogc
+    private void* roundedBegin() const
+    {
+        return cast(void*) roundUpToAlignment(cast(size_t) _begin, alignment);
+    }
+
+    pure nothrow @trusted @nogc
+    private void* roundedEnd() const
+    {
+        return cast(void*) roundDownToAlignment(cast(size_t) _end, alignment);
+    }
     /**
-    Constructs a region backed by a user-provided store. Assumes $(D store) is
-    aligned at $(D minAlign). Also assumes the memory was allocated with $(D
-    ParentAllocator) (if different from $(D NullAllocator)).
+    Constructs a region backed by a user-provided store.
+    Assumes the memory was allocated with $(D ParentAllocator)
+    (if different from $(REF_ALTTEXT `NullAllocator`, NullAllocator, std,experimental,allocator,building_blocks,null_allocator)).
 
     Params:
-    store = User-provided store backing up the region. $(D store) must be
-    aligned at $(D minAlign) (enforced with $(D assert)). If $(D
-    ParentAllocator) is different from $(D NullAllocator), memory is assumed to
-    have been allocated with $(D ParentAllocator).
-    n = Bytes to allocate using $(D ParentAllocator). This constructor is only
-    defined If $(D ParentAllocator) is different from $(D NullAllocator). If
-    $(D parent.allocate(n)) returns $(D null), the region will be initialized
-    as empty (correctly initialized but unable to allocate).
-    */
+        store = User-provided store backing up the region. If $(D
+        ParentAllocator) is different from $(REF_ALTTEXT `NullAllocator`, NullAllocator, std,experimental,allocator,building_blocks,null_allocator), memory is assumed to
+        have been allocated with $(D ParentAllocator).
+        n = Bytes to allocate using $(D ParentAllocator). This constructor is only
+        defined If $(D ParentAllocator) is different from $(REF_ALTTEXT `NullAllocator`, NullAllocator, std,experimental,allocator,building_blocks,null_allocator). If
+        $(D parent.allocate(n)) returns $(D null), the region will be initialized
+        as empty (correctly initialized but unable to allocate).
+        */
     this(ubyte[] store)
     {
-        store = cast(ubyte[])(store.roundUpToAlignment(alignment));
-        store = store[0 .. $.roundDownToAlignment(alignment)];
-        assert(store.ptr.alignedAt(minAlign));
-        assert(store.length % minAlign == 0);
         _begin = store.ptr;
         _end = store.ptr + store.length;
         static if (growDownwards)
-            _current = _end;
+            _current = roundedEnd();
         else
-            _current = store.ptr;
+            _current = roundedBegin();
     }
 
     /// Ditto
     static if (!is(ParentAllocator == NullAllocator))
     this(size_t n)
     {
-        this(cast(ubyte[])(parent.allocate(n.roundUpToAlignment(alignment))));
+        this(cast(ubyte[]) (parent.allocate(n.roundUpToAlignment(alignment))));
     }
 
     /*
@@ -97,7 +104,8 @@ struct Region(ParentAllocator = NullAllocator,
     */
 
     /**
-    If `ParentAllocator` is not `NullAllocator` and defines `deallocate`, the region defines a destructor that uses `ParentAllocator.delete` to free the
+    If `ParentAllocator` is not $(REF_ALTTEXT `NullAllocator`, NullAllocator, std,experimental,allocator,building_blocks,null_allocator) and defines `deallocate`,
+    the region defines a destructor that uses `ParentAllocator.deallocate` to free the
     memory chunk.
     */
     static if (!is(ParentAllocator == NullAllocator)
@@ -125,11 +133,11 @@ struct Region(ParentAllocator = NullAllocator,
     adjustment (if $(D alignment > 1)), an increment, and a comparison.
 
     Params:
-    n = number of bytes to allocate
+        n = number of bytes to allocate
 
     Returns:
-    A properly-aligned buffer of size $(D n) or $(D null) if request could not
-    be satisfied.
+        A properly-aligned buffer of size $(D n) or $(D null) if request could not
+        be satisfied.
     */
     void[] allocate(size_t n)
     {
@@ -157,11 +165,11 @@ struct Region(ParentAllocator = NullAllocator,
     Allocates $(D n) bytes of memory aligned at alignment $(D a).
 
     Params:
-    n = number of bytes to allocate
-    a = alignment for the allocated block
+        n = number of bytes to allocate
+        a = alignment for the allocated block
 
     Returns:
-    Either a suitable block of $(D n) bytes aligned at $(D a), or $(D null).
+        Either a suitable block of $(D n) bytes aligned at $(D a), or $(D null).
     */
     void[] alignedAllocate(size_t n, uint a)
     {
@@ -254,8 +262,8 @@ struct Region(ParentAllocator = NullAllocator,
     does nothing.
 
     Params:
-    b = Block previously obtained by a call to $(D allocate) against this
-    allocator ($(D null) is allowed).
+        b = Block previously obtained by a call to $(D allocate) against this
+        allocator ($(D null) is allowed).
     */
     pure nothrow @nogc
     bool deallocate(void[] b)
@@ -291,11 +299,11 @@ struct Region(ParentAllocator = NullAllocator,
     {
         static if (growDownwards)
         {
-            _current = _end;
+            _current = roundedEnd();
         }
         else
         {
-            _current = _begin;
+            _current = roundedBegin();
         }
         return true;
     }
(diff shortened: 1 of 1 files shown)


jercaianu:
ping @andralex @edi33416

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +890,385 @@ version(Posix) @system unittest
...
+    be satisfied.
+    */
+    void[] allocate(size_t n)
+    {
+        import core.atomic : cas, atomicLoad;
+
+        if (n == 0) return null;
+        auto rounded = goodAllocSize(n);
+
+        shared void *localCurrent;
```
wilzbach: `void*` - this isn't C ;-)
(also you could have declared these in the same line)

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +890,385 @@ version(Posix) @system unittest
...
+            do
+            {
+                localCurrent = atomicLoad(_current);
+                localNewCurrent = localCurrent + rounded;
+                if (localNewCurrent < localCurrent || localNewCurrent > _end)
+                    return null;
+            } while (!cas(&_current, localCurrent, localNewCurrent));
+
+            auto result = cast(void[]) localCurrent[0 .. n];
+        }
```
wilzbach: Don't repeat yourself!
Use the static if only for the bits that changed (or better: use `mixin` for the sign)

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +890,385 @@ version(Posix) @system unittest
...
+    b = Block previously obtained by a call to $(D allocate) against this
+    allocator ($(D null) is allowed).
+    */
+    bool deallocate(void[] b)
+    {
+        import core.atomic : cas, atomicLoad;
+
+        auto rounded = goodAllocSize(b.length);
+        shared void *localCurrent;
+        shared void *localNewCurrent;
```
wilzbach: `void*`

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +890,385 @@ version(Posix) @system unittest
...
+            localNewCurrent = localCurrent + rounded;
+            if (b.ptr == localCurrent)
+                return cas(&_current, localCurrent, localNewCurrent);
+        }
+        else
+        {
+            localNewCurrent = localCurrent - rounded;
+            if (b.ptr == localNewCurrent)
+                return cas(&_current, localCurrent, localNewCurrent);
+        }
```
wilzbach: Again DRY

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +890,385 @@ version(Posix) @system unittest
     assert((() nothrow @safe @nogc => alloc.owns(c))() == Ternary.yes);
     assert((() nothrow @safe @nogc => alloc.owns(null))() == Ternary.no);
 }
+
+/**
+The threadsafe version of the `Region` allocator.
+Allocations and deallocations are lock-free based using the `cas`
+functionality from `core.atomic`.
```
wilzbach: Use the `REF` for links

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +890,385 @@ version(Posix) @system unittest
...
+        shared void *localNewCurrent;
+        static if (growDownwards)
+        {
+            do
+            {
+                localCurrent = atomicLoad(_current);
+                localNewCurrent = localCurrent - rounded;
+                if (localNewCurrent > localCurrent || localNewCurrent < _begin)
+                    return null;
+            } while (!cas(&_current, localCurrent, localNewCurrent));
```
wilzbach: Ideally this would be just one-liner like `atomicOp!"=-"(_current, rounded)`.
How about creating a local version of `atomicOp` for your need of checking that you haven't under/overrun the region?

std/experimental/allocator/building_blocks/region.d
```diff
@@ -368,20 +368,29 @@ struct Region(ParentAllocator = NullAllocator,
...
+        assert((() pure nothrow @safe @nogc => a.empty)() ==  Ternary.yes);
+        const b = a.allocate(101);
+        assert(b.length == 101);
+        assert((() nothrow @safe @nogc => a.owns(b))() == Ternary.yes);
+
+        // Ensure deallocate inherits from parent allocators
+        auto c = a.allocate(42);
+        assert(c.length == 42);
+        assert((() nothrow @nogc => a.deallocate(c))());
+        assert((() pure nothrow @safe @nogc => a.empty)() ==  Ternary.no);
```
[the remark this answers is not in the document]
edi33416: They are. The lambdas only force the compiler to check for the attributes

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +890,385 @@ version(Posix) @system unittest
...
+    /**
+    Alignment offered.
+    */
+    alias alignment = minAlign;
+
+    /**
+    Allocates $(D n) bytes of memory. The allocation is served by atomically incrementing
+    a pointer which keeps track of the current used space.
+
+    To increase performance, allocations which exceed the region capacity could cause other
```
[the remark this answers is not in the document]
jercaianu: Taking a second look at this, I was wrong in my comments.
The cas loop is useful because it prevents intermitent failures by checking bounds inside the loop.

jercaianu:
@andralex @wilzbach 
Thanks for the review!
The cas loops, even though they may look repetitive, are necessary to prevent intermitent allocation failures caused by parallel allocations exceeding the region size.

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +890,385 @@ version(Posix) @system unittest
...
+
+/**
+The threadsafe version of the `Region` allocator.
+Allocations and deallocations are lock-free based using the `cas`
+functionality from `core.atomic`.
+*/
+shared struct SharedRegion(ParentAllocator = NullAllocator,
+    uint minAlign = platformAlignment,
+    Flag!"growDownwards" growDownwards = No.growDownwards)
+{
```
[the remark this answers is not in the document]
jercaianu: Currently, the destructor of both Region and SharedRegion deallocates the memory (if the case) from the ParentAllocator.
If the SharedRegion has its storage from the ParentAllocator, we can't know if the destructor will be nothrow, nogc.

Should I change the behavior of the destructor and force the caller to manually call deallocate on the Region inner storage?

std/experimental/allocator/building_blocks/region.d
jercaianu: Can you please help me with this?
I'm not sure how to use that for "core.atomic"
Thanks

std/experimental/allocator/building_blocks/region.d
```diff
@@ -53,26 +53,34 @@ struct Region(ParentAllocator = NullAllocator,
...
+        have been allocated with $(D ParentAllocator).
+        n = Bytes to allocate using $(D ParentAllocator). This constructor is only
+        defined If $(D ParentAllocator) is different from $(D NullAllocator). If
+        $(D parent.allocate(n)) returns $(D null), the region will be initialized
+        as empty (correctly initialized but unable to allocate).
+        */
     this(ubyte[] store)
     {
-        store = cast(ubyte[])(store.roundUpToAlignment(alignment));
+        static if (!is(ParentAllocator == NullAllocator))
```
jercaianu: @andralex 
If provided with storage from a ParentAllocator, the Region would incorrectly call 'deallocate' from its destructor.

This constructor here changes the "store" range parameter, which is later passed to the ParentAllocator for deallocation, in the destructor.

jercaianu: See below, I also added a test to reflect this change

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +890,385 @@ version(Posix) @system unittest
...
+
+    Params:
+    store = User-provided store backing up the region. $(D store) must be
+    aligned at $(D minAlign) (enforced with $(D assert)). If $(D
+    ParentAllocator) is different from $(D NullAllocator), memory is assumed to
+    have been allocated with $(D ParentAllocator).
+    n = Bytes to allocate using $(D ParentAllocator). This constructor is only
+    defined If $(D ParentAllocator) is different from $(D NullAllocator). If
+    $(D parent.allocate(n)) returns $(D null), the region will be initialized
+    as empty (correctly initialized but unable to allocate).
```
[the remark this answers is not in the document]
jercaianu: I added the REF macro for NullAllocator, but since the full module name is so long, it doesn't look too good cosmetically.

std/experimental/allocator/building_blocks/region.d
jercaianu: It would work, but this lock-free code is already rather difficult to understand.
Wouldn't it be better to keep it a bit longer, but more simple?

wilzbach:
> I added the REF macro for NullAllocator, but since the full module name is so long, it doesn't look too good cosmetically.

Search for `REF_ALTTEXT` in the phobos code.

std/experimental/allocator/building_blocks/region.d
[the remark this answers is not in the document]
andralex: we should assume all are nothrow and nogc

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +890,385 @@ version(Posix) @system unittest
...
+    */
+    static if (stateSize!ParentAllocator)
+    {
+        ParentAllocator parent;
+    }
+    else
+    {
+        alias parent = ParentAllocator.instance;
+    }
+    private shared void* _current, _begin, _end;
```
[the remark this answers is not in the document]
jpf91: The style guide says
> Also, names do not begin with an underscore ‘_’ unless they are private

This convention is also often used in languages which do have `private` (java, c#). It's useful to disambiguate when you have possible naming conflicts (private member variable `_foo` constructor with parameter `foo` initializing `_foo`, public property `foo` returning `_foo`). So I think this coding style is perfectly reasonable.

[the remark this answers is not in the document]
jmdavis: @wilzbach  It's _very_ common in C++, Java, C#, D, etc. to prefix private member variables with an underscore (another common convention is to prefix them with `m_`), and as jpf91 points out, it reduces problems with public properties. A lot of the programmers that I've worked with would consider it bad practice to name a private member variable without prefixing it with something like `_`  or `m_`. In my experience, folks who think that it's perfectly okay to name private member variables without adding something to them to differentiate them from public members are in the minority - and the vast majority of my programming experience is in languages which have private.

std/experimental/allocator/building_blocks/region.d
klickverbot: Do we guarantee that core.atomic.cas doesn't fail spuriously?

JackStouffer:
ping @andralex @edi33416 

Is this good to go?

std/experimental/allocator/building_blocks/region.d
edi33416: Good catch

std/experimental/allocator/building_blocks/region.d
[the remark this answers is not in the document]
edi33416: I believe that this is a good enforcement on the parent allocator: `deallocate` should not produce garbage while freeing and it should not throw; imho `deallocate` should follow up with the guidelines for destructors

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +939,388 @@ version(Posix) @system unittest
...
+    /**
+    If `ParentAllocator` is not $(REF_ALTTEXT `NullAllocator`, NullAllocator, std,experimental,allocator,building_blocks,null_allocator) and defines `deallocate`,
+    the region defines a destructor that uses `ParentAllocator.deallocate` to free the
+    memory chunk.
+    */
+    static if (!is(ParentAllocator == NullAllocator)
+        && hasMember!(ParentAllocator, "deallocate"))
+    ~this()
+    {
+        parent.deallocate(cast(void[]) _innerStorage);
```
edi33416: Looking at this, I don't think it is safe to assume that nobody else has a reference to `_innerStorage`, as we are passing a memory chunk allocated by the parent in `this(ubyte[])`.

Think of the following example:
```
auto buf = cast(ubyte[]) Mallocator.instance.allocate(42);
auto r1 = Region!(Mallocator)(buf);
auto r2 = Region!(Mallocator)(buf); // double free on ~this()
```

I believe we should take ownership of the memory chunk that gets passed into the region allocator, thus allowing us to safely deallocate the memory chunk.
In order to do this, we should define two `ctors` that take a memory chunk
* `this(ref ubyte[] store)` that takes an lvalue takes over the buffer
* `this(ubyte[] store)` that gets called by `this(size_t n)` which takes an rvalue

Another way would be to never `deallocate` a memory chunk that was passed to the region allocator, and only deallocate for regions that were created through `this(size_t n)`, thus leaving the resource management to the parent owning allocator.

What do you think?

std/experimental/allocator/building_blocks/region.d
```diff
@@ -878,3 +939,388 @@ version(Posix) @system unittest
...
+        }
+        tg.joinAll();
+
+        sort!((a, b) => a.ptr < b.ptr)(buf[0 .. 2 * numThreads]);
+        foreach (i; 0 .. 2 * numThreads - 1)
+        {
+            assert(buf[i].ptr + buf[i].length <= buf[i + 1].ptr);
+        }
+
+        assert(!a.deallocate(buf[1]));
```
edi33416: Maybe we should consider adding an `alignedDeallocate` method that is able to solve the issue of deallocating aligned buffers?

std/experimental/allocator/building_blocks/region.d
edi33416: I'm not saying that the above sample of using buf is correct, but I don't think this should crash, though this is debatable.
A trickier use case, with an unexpected behavior would be
```
auto buf = cast(ubyte[]) Mallocator.instance.allocate(42);
{
  auto r = Region!(Mallocator)(buf);
  // do stuff with region
}
// attempt to reuse buf here and avoid allocating again
// and fail since Region silently deallocated buf
```

std/experimental/allocator/building_blocks/region.d
andralex: @klickverbot I think it's LL/SC that may fail spuriously, cas doesn't.

andralex:
So I assume the autotester failure is spurious.

PR #6177, opened 2018-02-14
add opDispatch-based test to BitFlags enum

MartinNowak:
- allows for less brittle flag test
  (`flags.featureX` vs. `(flags & FlagsEnum.featureX)`)
- allows for simpler testing of combined (unsafe) flags
  (`flags.combined` vs.
   `(flags & FlagsEnum.combined) == BitFlags!(FlagsEnum, Yes.unsafe)(FlagsEnum.combined)`)

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -7691,13 +7691,15 @@ public:
         return opBinary!op(flag);
     }
 
-    bool opDispatch(string name)() const if (__traits(hasMember, E, name))
+    bool opDispatch(string name)() const
+    if (__traits(hasMember, E, name))
     {
         enum e = __traits(getMember, E, name);
         return (mValue & e) == e;
     }
 
-    void opDispatch(string name)(bool set) if (__traits(hasMember, E, name))
+    void opDispatch(string name)(bool set)
+    if (__traits(hasMember, E, name))
     {
         enum e = __traits(getMember, E, name);
         if (set)
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -7858,9 +7858,9 @@ public:
 {
     enum UnsafeEnum
     {
-        A,
-        B,
-        C,
+        A = 1,
+        B = 2,
+        C = 4,
         BC = B|C
     }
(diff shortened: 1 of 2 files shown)


std/typecons.d
```diff
@@ -7645,6 +7645,12 @@ public:
     {
         return opBinary!op(flag);
     }
+
+    bool opDispatch(string name)() const if (__traits(hasMember, E, name))
+    {
+        enum e = __traits(getMember, E, name);
+        return (mValue & e) == e;
```
MetaLang: Why this instead of `cast(bool)(mValue & e)`?

wilzbach:
```d
auto mValue = 2;
auto e = 3;

(mValue & e).writeln;
((mValue & e) == e).writeln;
```

https://run.dlang.io/is/l1LXwy

MetaLang: Ah, ok, I was thinking with the wrong intuition for what this syntax meant.

MartinNowak: A good optimizer should be able to lower this to identical code for single bit flags, but in particular multi-bit flags are tedius to test.
Good old `flags & Enum.combined` can be used when you want to test for any instead of all flags.
Unfortunately the docs for BitFlags are in bad shape, so I just added this to the all-explaining example.

wilzbach:
(rebased to reflect the solve the conflicts resulting from the recent `unittest` restructuring)

MartinNowak:
It just struck me, why would we restrict this to getting a property when we could also support setting a property. Please re-review and help with the wording of the changelog.

wilzbach:
> property. Please re-review and help with the wording of the changelog.

Nice! I appended a few ideas to your PR. Feel free to squash if you approve.
In general for unsafe enums I think the point that not the entire enum, but only the bits that are accessed by the combination should be stressed (I modified your new test to test for this too).

MartinNowak:
The updated FreeBSD32 tester hangs in the std.net.curl test.

PR #4936, opened 2016-12-08
Fix Issue 16745 - Add template helper for creating static arrays with the size inferred

John-Colvin:

PR #4090, opened 2016-03-17
Better static array support in std.array

rcorre:
Better static array support in `std.array`.

The main change I wanted to introduce is `array!n`, which functions like `array` but creates a static array.

For example:

    int[5] = only(1,2,3,4,5).array!5;

This also implements `staticArray`, which is a sort of library-level static array literal:

    auto i = staticArray(1,2,3); // i is an int[3], you don't have to deduce the size manually.

I'm mostly interested in `array!size`, but figured I'd throw in `staticArray` as well just to see what people think.

PR #6178, opened 2018-02-15
static array litteral: `[1,2].asStatic` + overloads for CT ranges, RT ranges, explicit types

timotheecour:
(moved from https://github.com/dlang/druntime/pull/2093)

## usage
```
// from litteral
auto a = [1,2].asStatic; // int[2]
auto a = [1,2].asStatic!byte; // byte[2]

// from range
auto a = 2.iota.asStatic!2; // int[2]
auto a = 2.iota.asStatic!(byte[2]); // byte[2]

// from alias
auto a = asStatic!(2.iota); // int[2]
auto a = asStatic!(byte, 2.iota); // byte[2]
```

## motivation
places like here: https://github.com/dlang/dub/pull/1377#discussion_r168262426

## question for reviewers
not sure how one wants to treat cases where ForEachType != ElementType, eg:
"ab∫".staticArray currently is `immutable(char[5])` instead of immutable(dchar[3])

## links
https://issues.dlang.org/show_bug.cgi?id=16745 Add template helper for creating static arrays with the size inferred
https://issues.dlang.org/show_bug.cgi?id=8008 Issue 8008 - Syntax for fixed size array literals like [1,2,3]s (edit)
https://issues.dlang.org/show_bug.cgi?id=481 Issue 481 - Fixed-length arrays with automatically computed length (edit)

https://github.com/dlang/phobos/pull/4936 Fix Issue 16745 - Add template helper for creating static arrays with the size inferred #4936
https://github.com/dlang/phobos/pull/4090 Better static array support in std.array #4090
https://github.com/dlang/phobos/pull/6214 Overload for static arrays for std.array.array #6214

## not related
https://issues.dlang.org/show_bug.cgi?id=9165 Issue 9165 - Auto conversion from dynamic array to fixed size array at return (edit)

---
This change is [Reviewable](https://reviewable.io/reviews/dlang/phobos/6178)

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -3980,6 +3980,7 @@ nothrow pure @safe unittest
     {
         [1, 2, val].asStatic.checkStaticArray!int(gold);
     }
+    checkNogc();
 
     [1, 2, val].asStatic!double.checkStaticArray!double(gold);
     [1, 2, 3].asStatic!int.checkStaticArray!int(gold);
--- a/std/array.d
+++ b/std/array.d
@@ -3933,7 +3933,7 @@ Constructs a static array from `a`.
 The type of elements can be specified implicitly (`[1,2].staticArray` of type int[2])
 or explicitly (`[1,2].staticArray!float` of type float[2]).
 When `a` is a range (not known at compile time), the number of elements has to be given as template argument
-(eg `myrange.staticArray!2`).
+(e.g. `myrange.staticArray!2`).
 Size and type can be combined (eg: `2.iota.staticArray!(byte[2])`).
 When the range `a` is known at compile time, it can also be specified as a
 template argument to avoid having to specify the number of elements
--- a/std/array.d
+++ b/std/array.d
@@ -82,6 +82,7 @@ import std.meta;
 import std.traits;
 
 import std.range.primitives;
+import std.conv : emplaceRef;
 public import std.range.primitives : save, empty, popFront, popBack, front, back;
 
 /**
@@ -3926,4 +3927,287 @@ unittest
     a ~= 'a'; //Clobbers here?
     assert(appS.data == "hellow");
     assert(appA.data == "hellow");
+}
+
+/++
+Params: a = The array elements
+
+Returns: A static array constructed from `a`. The type of elements can be
+specified implicitly (`int[2] a = staticArray(1,2);`) or explicitly
+(`float[2] a = staticArray!float(1,2)`).
+The result is an rvalue, therefore uses like
+`foo(staticArray(1, 2, 3))` may be inefficient because of the copies.
++/
+pragma(inline, true) U[T.length] staticArray(U = CommonType!T, T...)(T a) nothrow @safe pure @nogc
+{
+    return [a];
+}
+
+/// ditto
+// Workaround https://issues.dlang.org/show_bug.cgi?id=16779 (make alias to staticArray once fixed)
+pragma(inline, true) U[T.length] staticArrayCast(U, T...)(T a) nothrow @safe pure @nogc
+{
+    enum n = T.length;
+    U[n] ret = void;
+    static foreach (i; 0 .. n)
+    {
+        // TODO: If any of these throws, the destructors of the already-constructed elements is not called. https://github.com/dlang/phobos/pull/4936/files#r131709329
+        emplaceRef!U(ret[i], cast(U) a[i]);
+    }
+    return ret;
+}
+
+///
+nothrow pure @safe unittest
+{
+    auto a = staticArray(1, 2);
+    assert(is(typeof(a) == int[2]) && a == [1, 2]);
+
+    auto b = staticArrayCast!byte(1, 2);
+    assert(is(typeof(b) == byte[2]) && b == [1, 2]);
+}
+
+nothrow pure @safe unittest
+{
+    int val = 3;
+    staticArray(1, 2, val).checkStaticArray!int([1, 2, 3]);
+    staticArray(1, 2.0).checkStaticArray!double([1, 2.0]);
+    assert(!__traits(compiles, staticArray(1, "")));
+    staticArray().checkStaticArray!void([]);
+    staticArray!float(1, 2).checkStaticArray!float([1, 2]);
+    // auto a = staticArray!byte(1, 2);
+    staticArrayCast!byte(1, 2).checkStaticArray!byte([1, 2]);
+    staticArrayCast!byte(1, 129).checkStaticArray!byte([1, -127]);
+
+    staticArrayCast!(const(int))(1, 2).checkStaticArray!(const(int))([1, 2]);
+    staticArrayCast!(immutable(int))(1, 2).checkStaticArray!(immutable(int))([1, 2]);
+    staticArray([1]).checkStaticArray!(int[])([[1]]);
+}
+
+@system unittest
+{
+    version (Bug)
+    {
+        // NOTE: correctly issues a deprecation
+        int[] a2 = staticArray(1, 2);
+    }
+}
+
+/++
+Params: arr = The array elements
+
+Returns: A static array constructed from `arr`. The type of elements can be
+specified implicitly (`int[2] a = [1,2].asStatic;`) or explicitly
+(`float[2] a = [1,2].asStaticCast!float`).
++/
+pragma(inline, true) T[n] asStatic(T, size_t n)(auto ref T[n] arr) nothrow @safe pure @nogc
+{
+    return arr;
+}
+
+/// ditto
+U[n] asStaticCast(U, T, size_t n)(auto ref T[n] arr) nothrow @safe pure @nogc
+{
+    U[n] ret = void;
+    static foreach (i; 0 .. n)
+    {
+        emplaceRef!U(ret[i], cast(U) arr[i]);
+    }
+    return ret;
+}
+
+///
+nothrow pure @safe unittest
+{
+    auto a = [0, 1].asStatic;
+    assert(is(typeof(a) == int[2]) && a == [0, 1]);
+
+    auto b = [0, 1].asStaticCast!byte;
+    assert(is(typeof(b) == byte[2]) && b == [0, 1]);
+}
+
+nothrow pure @safe unittest
+{
+    int val = 3;
+    static immutable gold = [1, 2, 3];
+    [1, 2, val].asStatic.checkStaticArray!int([1, 2, 3]);
+
+    @nogc void checkNogc()
+    {
+        [1, 2, val].asStatic.checkStaticArray!int(gold);
+    }
+
+    [1, 2, val].asStaticCast!double.checkStaticArray!double(gold);
+    [1, 2, 3].asStaticCast!int.checkStaticArray!int(gold);
+
+    [1, 2, 3].asStaticCast!(const(int)).checkStaticArray!(const(int))(gold);
+    [1, 2, 3].asStaticCast!(const(double)).checkStaticArray!(const(double))(gold);
+    {
+        const(int)[3] a2 = [1, 2, 3].asStatic;
+    }
+
+    [1, 129].asStaticCast!byte.checkStaticArray!byte([1, -127]);
+
+}
+
+/++
+Params: a = input range of elements
+
+Returns: A static array constructed from `a`.
++/
+auto asStatic(size_t n, T)(T a) nothrow @safe pure @nogc
+{
+    // TODO: ElementType vs ForeachType
+    alias U = typeof(a[0]);
+    U[n] ret = void;
+    size_t i;
+    foreach (ref ai; a)
+    {
+        emplaceRef!U(ret[i++], ai);
+    }
+    assert(i == n);
+    return ret;
+}
+
+/// ditto
+auto asStaticCast(Un : U[n], U, size_t n, T)(T a) nothrow @safe pure @nogc
+{
+    U[n] ret = void;
+    size_t i;
+    foreach (ref ai; a)
+    {
+        emplaceRef!U(ret[i++], cast(U) ai);
+    }
+    assert(i == n);
+    return ret;
+}
+
+///
+nothrow pure @safe unittest
+{
+    import std.range : iota;
+
+    auto a = 2.iota.asStatic!2;
+    assert(is(typeof(a) == int[2]) && a == [0, 1]);
+    auto b = 2.iota.asStaticCast!(byte[2]);
+    assert(is(typeof(b) == byte[2]) && b == [0, 1]);
+}
+
+nothrow pure @safe unittest
+{
+
+    auto a = [1, 2].asStatic;
+    assert(is(typeof(a) == int[2]) && a == [1, 2]);
+
+    import std.range : iota;
+
+    2.iota.asStatic!2.checkStaticArray!int([0, 1]);
+    2.iota.asStaticCast!(double[2]).checkStaticArray!double([0, 1]);
+    2.iota.asStaticCast!(byte[2]).checkStaticArray!byte([0, 1]);
+}
+
+nothrow pure @system unittest
+{
+    import std.range : iota;
+
+    assert(isThrown!Error(2.iota.asStatic!1));
+    assert(isThrown!Error(2.iota.asStatic!3));
+    // NOTE: alternatives are not nothrow:
+    version (none)
+    {
+        import std.exception : assertThrown;
+
+        assertThrown!Error(2.iota.asStatic!3);
+        import std.exception : ifThrown;
+
+        assert(ifThrown!Error({ 2.iota.asStatic!1; return false; }(), true));
+    }
+}
+
+@system unittest
+{
+    version (Bug)
+    {
+        // https://issues.dlang.org/show_bug.cgi?id=16779
+        auto a2 = [1, 2, 3].asStatic!byte;
+        auto a3 = [1, 2, 3].asStatic!ubyte;
+
+        // NOTE: correctly issues a deprecation
+        int[] a2 = [1, 2].asStatic;
+    }
+}
+
+/++
+Params: arr = the compile time range
+
+Returns: A static array constructed from `arr`.
++/
+auto asStatic(alias arr)() nothrow @safe pure @nogc
+{
+    enum n = arr.length;
+    alias U = typeof(arr[0]);
+    U[n] ret = void;
+    static foreach (i; 0 .. n)
+    {
+        emplaceRef!U(ret[i], arr[i]);
+    }
+    return ret;
+}
+
+/// ditto
+auto asStaticCast(U, alias arr)() nothrow @safe pure @nogc
+{
+    enum n = arr.length;
+    U[n] ret = void;
+    static foreach (i; 0 .. n)
+    {
+        emplaceRef!U(ret[i], cast(U) arr[i]);
+    }
+    return ret;
+}
+
+///
+nothrow pure @safe unittest
+{
+    import std.range : iota;
+
+    enum a = asStatic!(2.iota);
+    assert(is(typeof(a) == int[2]) && a == [0, 1]);
+
+    enum b = asStaticCast!(byte, 2.iota);
+    assert(is(typeof(b) == byte[2]) && b == [0, 1]);
+}
+
+nothrow pure @safe unittest
+{
+    import std.range : iota;
+
+    enum a = asStatic!(2.iota);
+    asStatic!(2.iota).checkStaticArray!int([0, 1]);
+    asStaticCast!(double, 2.iota).checkStaticArray!double([0, 1]);
+    asStaticCast!(byte, 2.iota).checkStaticArray!byte([0, 1]);
+}
+
+void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
+{
+    assert(is(T1 == T[T1.length]));
+    assert(a == b);
+}
+
+// TODO: add this to assertThrown in std.exception
+bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow pure @system
+{
+    try
+    {
+        expression();
+        return false;
+    }
+    catch (T)
+    {
+        return true;
+    }
+    catch (Exception)
+    {
+        return false;
+    }
 }
--- a/std/array.d
+++ b/std/array.d
@@ -3952,7 +3952,7 @@ pragma(inline, true) T[n] staticArray(T, size_t n)(auto ref T[n] a) nothrow @saf
 }
 
 /// ditto
-U[n] staticArray(U, T, size_t n)(auto ref T[n] a)
+U[n] staticArray(U, T, size_t n)(auto ref T[n] a) nothrow
 if (!is(U == T))
 {
     import std.conv : emplaceRef;
@@ -4013,7 +4013,7 @@ if (isInputRange!T)
 }
 
 /// ditto
-auto staticArray(Un : U[n], U, size_t n, T)(T a) nothrow @safe pure @nogc
+auto staticArray(Un : U[n], U, size_t n, T)(T a) nothrow
 if (isInputRange!T)
 {
     import std.conv : emplaceRef;
--- a/std/array.d
+++ b/std/array.d
@@ -3930,12 +3930,14 @@ unittest
 
 /++
 Constructs a static array from `a`.
-The type of elements can be specified implicitly (`auto a = [1,2].asStatic;` of type int[2])
-or explicitly (`auto a = [1,2].asStatic!float` of type float[2]). When `a` is a range,
-the number of elements has to be given as template argument (eg `2.iota.asStatic!2`).
+The type of elements can be specified implicitly (`[1,2].asStatic` of type int[2])
+or explicitly (`[1,2].asStatic!float` of type float[2]).
+When `a` is a range (not known at compile time), the number of elements has to be given as template argument
+(eg `myrange.asStatic!2`).
 Size and type can be combined (eg: `2.iota.asStatic!(byte[2])`).
-Range `a` can also be specified as a template argument (eg: `asStatic!(2.iota)`
-or `asStatic!(double, 2.iota)`).
+When the range `a` is known at compile time, it can also be specified as a
+template argument to avoid having to specify the number of elements
+(eg: `asStatic!(2.iota)` or `asStatic!(double, 2.iota)`).
 
 Note: `foo([1, 2, 3].asStatic)` may be inefficient because of the copies involved.
 
@@ -4128,7 +4130,7 @@ private void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
 }
 
 // TODO: consider adding this to assertThrown in std.exception
-private bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow
+version(unittest) private bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow
 {
     try
     {
--- a/std/array.d
+++ b/std/array.d
@@ -4062,29 +4062,9 @@ nothrow pure @system unittest
 
     assert(isThrown!Error(2.iota.asStatic!1));
     assert(isThrown!Error(2.iota.asStatic!3));
-    // NOTE: alternatives are not nothrow:
-    version (none)
-    {
-        import std.exception : assertThrown;
 
-        assertThrown!Error(2.iota.asStatic!3);
-        import std.exception : ifThrown;
-
-        assert(ifThrown!Error({ 2.iota.asStatic!1; return false; }(), true));
-    }
-}
-
-@system unittest
-{
-    version (Bug)
-    {
-        // https://issues.dlang.org/show_bug.cgi?id=16779
-        auto a2 = [1, 2, 3].asStatic!byte;
-        auto a3 = [1, 2, 3].asStatic!ubyte;
-
-        // NOTE: correctly issues a deprecation
-        int[] a2 = [1, 2].asStatic;
-    }
+    // NOTE: correctly issues a deprecation
+    // int[] a2 = [1, 2].asStatic;
 }
 
 /// ditto
@@ -4129,7 +4109,20 @@ private void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
     assert(a == b);
 }
 
-// TODO: consider adding this to assertThrown in std.exception
+/+
+TODO: consider adding this to assertThrown in std.exception
+The alternatives are not nothrow (see https://issues.dlang.org/show_bug.cgi?id=12647)
+---
+import std.exception : assertThrown;
+assertThrown!Error(expr_that_throws);
+
+// or
+import std.exception : ifThrown;
+scope(failure) assert(0);
+assert(ifThrown!Error({ expr_that_throws; return false; }(), true));
+---
+Also, `isThrown` is easier to use than ifThrown and more flexible than assertThrown.
++/
 version(unittest) private bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow
 {
     try
--- a/std/array.d
+++ b/std/array.d
@@ -3956,6 +3956,7 @@ U[n] staticArray(U, T, size_t n)(auto ref T[n] a)
 if (!is(U == T))
 {
     import std.conv : emplaceRef;
+
     U[n] ret = void;
     static foreach (i; 0 .. n)
     {
@@ -3986,6 +3987,7 @@ nothrow pure @safe unittest
     {
         [1, 2, val].staticArray.checkStaticArray!int(gold);
     }
+
     checkNogc();
 
     [1, 2, val].staticArray!double.checkStaticArray!double(gold);
@@ -4003,10 +4005,12 @@ nothrow pure @safe unittest
 
 /// ditto
 auto staticArray(size_t n, T)(T a)
+if (isInputRange!T)
 {
     import std.conv : emplaceRef;
+
     // TODO: ElementType vs ForeachType
-    alias U = typeof(a[0]);
+    alias U = typeof(a.front);
     U[n] ret = void;
     size_t i;
     foreach (ref ai; a)
@@ -4019,8 +4023,10 @@ auto staticArray(size_t n, T)(T a)
 
 /// ditto
 auto staticArray(Un : U[n], U, size_t n, T)(T a) nothrow @safe pure @nogc
+if (isInputRange!T)
 {
     import std.conv : emplaceRef;
+
     U[n] ret = void;
     size_t i;
     foreach (ref ai; a)
@@ -4035,6 +4041,7 @@ auto staticArray(Un : U[n], U, size_t n, T)(T a) nothrow @safe pure @nogc
 nothrow pure @safe unittest
 {
     import std.range : iota;
+
     auto input = 2.iota;
     auto a = input.staticArray!2;
     static assert(is(typeof(a) == int[2]));
@@ -4070,6 +4077,7 @@ nothrow pure @system unittest
 
 /// ditto
 auto staticArray(alias a)()
+if (isInputRange!(typeof(a)))
 {
     // NOTE: without `cast`, getting error:
     // cannot deduce function from argument types !(length)(Result)
@@ -4078,6 +4086,7 @@ auto staticArray(alias a)()
 
 /// ditto
 auto staticArray(U, alias a)()
+if (isInputRange!(typeof(a)))
 {
     return .staticArray!(U[cast(size_t) a.length])(a);
 }
--- a/std/array.d
+++ b/std/array.d
@@ -4131,7 +4131,7 @@ nothrow pure @safe unittest
     asStatic!(byte, 2.iota).checkStaticArray!byte([0, 1]);
 }
 
-void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
+private void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
 {
     assert(is(T1 == T[T1.length]));
     assert(a == b);
--- a/std/array.d
+++ b/std/array.d
@@ -3938,7 +3938,7 @@ or explicitly (`float[2] a = [1,2].asStatic!float`). When `a` is a range,
 the number of elements has to be given as template argument (eg `2.iota.asStatic!2`).
 Size and type can be combined (eg: `2.iota.asStatic!(byte[2])`).
 Range `a` can also be specified as a template argument (eg: `asStatic!(2.iota)`
-or `asStatic!(double, 2.iota)`.
+or `asStatic!(double, 2.iota)`).
 
 Note: `foo([1, 2, 3].asStatic)` may be inefficient because of the copies involved.
 +/
--- a/std/array.d
+++ b/std/array.d
@@ -82,7 +82,6 @@ import std.meta;
 import std.traits;
 
 import std.range.primitives;
-import std.conv : emplaceRef;
 public import std.range.primitives : save, empty, popFront, popBack, front, back;
 
 /**
@@ -3947,6 +3946,7 @@ pragma(inline, true) U[T.length] staticArray(U = CommonType!T, T...)(T a) nothro
 // Workaround https://issues.dlang.org/show_bug.cgi?id=16779 (make alias to staticArray once fixed)
 pragma(inline, true) U[T.length] staticArrayCast(U, T...)(T a) nothrow @safe pure @nogc
 {
+    import std.conv : emplaceRef;
     enum n = T.length;
     U[n] ret = void;
     static foreach (i; 0 .. n)
@@ -4008,6 +4008,7 @@ pragma(inline, true) T[n] asStatic(T, size_t n)(auto ref T[n] arr) nothrow @safe
 /// ditto
 U[n] asStaticCast(U, T, size_t n)(auto ref T[n] arr) nothrow @safe pure @nogc
 {
+    import std.conv : emplaceRef;
     U[n] ret = void;
     static foreach (i; 0 .. n)
     {
@@ -4057,6 +4058,7 @@ Returns: A static array constructed from `a`.
 +/
 auto asStatic(size_t n, T)(T a) nothrow @safe pure @nogc
 {
+    import std.conv : emplaceRef;
     // TODO: ElementType vs ForeachType
     alias U = typeof(a[0]);
     U[n] ret = void;
@@ -4072,6 +4074,7 @@ auto asStatic(size_t n, T)(T a) nothrow @safe pure @nogc
 /// ditto
 auto asStaticCast(Un : U[n], U, size_t n, T)(T a) nothrow @safe pure @nogc
 {
+    import std.conv : emplaceRef;
     U[n] ret = void;
     size_t i;
     foreach (ref ai; a)
@@ -4144,6 +4147,7 @@ Returns: A static array constructed from `arr`.
 +/
 auto asStatic(alias arr)() nothrow @safe pure @nogc
 {
+    import std.conv : emplaceRef;
     enum n = arr.length;
     alias U = typeof(arr[0]);
     U[n] ret = void;
@@ -4157,6 +4161,7 @@ auto asStatic(alias arr)() nothrow @safe pure @nogc
 /// ditto
 auto asStaticCast(U, alias arr)() nothrow @safe pure @nogc
 {
+    import std.conv : emplaceRef;
     enum n = arr.length;
     U[n] ret = void;
     static foreach (i; 0 .. n)
@@ -4195,7 +4200,7 @@ void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
 }
 
 // TODO: add this to assertThrown in std.exception
-bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow pure @system
+private bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow pure @system
 {
     try
     {
--- a/std/array.d
+++ b/std/array.d
@@ -3949,7 +3949,7 @@ pragma(inline, true) T[n] asStatic(T, size_t n)(auto ref T[n] a) nothrow @safe p
 }
 
 /// ditto
-U[n] asStatic(U, T, size_t n)(auto ref T[n] a) nothrow @safe pure @nogc
+U[n] asStatic(U, T, size_t n)(auto ref T[n] a)
 if (!is(U == T))
 {
     import std.conv : emplaceRef;
@@ -3997,7 +3997,7 @@ nothrow pure @safe unittest
 }
 
 /// ditto
-auto asStatic(size_t n, T)(T a) nothrow @safe pure @nogc
+auto asStatic(size_t n, T)(T a)
 {
     import std.conv : emplaceRef;
     // TODO: ElementType vs ForeachType
@@ -4082,13 +4082,13 @@ nothrow pure @system unittest
 }
 
 /// ditto
-auto asStatic(alias a)() nothrow @safe pure @nogc
+auto asStatic(alias a)()
 {
     return .asStatic!(cast(size_t) a.length)(a);
 }
 
 /// ditto
-auto asStatic(U, alias a)() nothrow @safe pure @nogc
+auto asStatic(U, alias a)()
 {
     return .asStatic!(U[cast(size_t) a.length])(a);
 }
@@ -4122,7 +4122,7 @@ private void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
 }
 
 // TODO: consider adding this to assertThrown in std.exception
-private bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow pure @system
+private bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow
 {
     try
     {
--- a/std/array.d
+++ b/std/array.d
@@ -3929,18 +3929,19 @@ unittest
 }
 
 /++
-Params: a = The input elements
-
-Returns: A static array constructed from `a`.
-
-The type of elements can be specified implicitly (`int[2] a = [1,2].asStatic;`)
-or explicitly (`float[2] a = [1,2].asStatic!float`). When `a` is a range,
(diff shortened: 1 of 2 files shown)


JackStouffer:
There's no reason to break up std.array. I suggest you leave that change out.

std/array/static_array.d
```diff
@@ -0,0 +1,159 @@
...
+pragma(inline, true) U[T.length] staticArray(U = CommonType!T, T...)(T a) @nogc
+{
+    return [a];
+}
+
+// D20180214T185602: Workaround https://issues.dlang.org/show_bug.cgi?id=16779 (make alias to staticArray once fixed)
+pragma(inline, true) U[T.length] staticArrayCast(U, T...)(T a) @nogc
+{
+    enum n = T.length;
+    U[n] ret = void;
```
n8sh: If `U` is a struct you need to check if it has an elaborate destructor and in that case not void-initialize `ret` because the destructor for `ret[i]` will be run before `a[i]` is blitted over it. An alternative could be to use `emplace`.

timotheecour: done, used emplaceRef wherever applicable

timotheecour:
> There's no reason to break up std.array. I suggest you leave that change out.

added motivation for it in intro: std.array is a large package and `std.array.static_array` and `std.array.util` don't call each other, so that makes a perfect candidate for splitting.

PTAL, I've improved API to allow all cases (see updated top-level msg)

wilzbach:
> don't call each other, so that makes a perfect candidate for splitting.

So do many other methods in std.algorithm and std.range - the splitting seems to me a bit artificial. You need a lot stronger motivation for this and I really don't want Phobos to end up with one module for every function.

timotheecour:
PTAL, all comments addressed, i unsplit std.array (although i disagree with this change but whatever)

the only failure is unrelated:

```
std.socket.SocketOSException@std\socket.d(2801): Unable to connect socket: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
----------------
0x00411BC8
0x0040F05B
0x0040EFDD
0x0040EE78
0x0040C0DF
0x75BB343D in BaseThreadInitThunk
0x77B99832 in RtlInitializeExceptionChain
0x77B99805 in RtlInitializeExceptionChain
Connecting to www.dlang.org on port 80...
win32.mak:24: recipe for target 'htmlget' failed
gmake: *** [htmlget] Error 1
```

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
     assert(appS.data == "hellow");
     assert(appA.data == "hellow");
 }
+
+version(all)
+{
+
+import std.traits : CommonType, Unqual, isStaticArray;
```
JackStouffer: imports go at the top of modules

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
...
+import std.conv : emplaceRef;
+
+// @nogc: https://issues.dlang.org/show_bug.cgi?id=18439
+//nothrow @safe pure:
+nothrow:
```
JackStouffer: Please avoid using attributes like this, they are a huge pain later on and Andrei has expressed desire for removing all such uses from Phobos.

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
     assert(appS.data == "hellow");
     assert(appA.data == "hellow");
 }
+
+version(all)
```
JackStouffer: superfluous

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
...
+    2.iota.asStaticCast!(double[2]).assertSame!double([0, 1]);
+    2.iota.asStaticCast!(byte[2]).assertSame!byte([0, 1]);
+}
+
+///
+@system unittest
+{
+    import std.range;
+
+    assert(isThrown!Error(2.iota.asStatic!1));
```
JackStouffer: `assertThrown` doesn't work?

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
...
+}
+
+///
+@safe unittest
+{
+    import std.range;
+
+    enum a = asStatic!(2.iota);
+    asStatic!(2.iota).assertSame!int([0, 1]);
+    asStaticCast!(double, 2.iota).assertSame!double([0, 1]);
```
JackStouffer: Don't use undocumented functions in public examples.

Better to just manually enter in both tests from assertSame here for each call.

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
...
+///
+@system unittest
+{
+    import std.range;
+
+    assert(isThrown!Error(2.iota.asStatic!1));
+    assert(isThrown!Error(2.iota.asStatic!3));
+}
+
+/// ditto but with an alias
```
JackStouffer: `ditto` only works if it's the only thing in the ddoc-ed comment. If you want this to have its own docs, you have to include all the information because the ddox version shows all functions on separate pages, and not in sequential order like ddoc.

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
...
+import std.conv : emplaceRef;
+
+// @nogc: https://issues.dlang.org/show_bug.cgi?id=18439
+//nothrow @safe pure:
+nothrow:
+@safe:
+pure:
+
+/++
+Returns a static array constructed from `a`. The type of elements can be
```
JackStouffer: needs `Params:` and `Returns:` sections. same throughout

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
...
+specified implicitly (`int[2] a = staticArray(1,2);`) or explicitly
+(`float[2] a = staticArray!float(1,2)`).
+D20180214T203015: The result is an rvalue, therefore uses like
```
JackStouffer: > D20180214T203015

What is this? A date? What for?

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
...
+    [1, 2, 3].asStaticCast!(const(int)).assertSame!(const(int))(gold);
+    [1, 2, 3].asStaticCast!(const(double)).assertSame!(const(double))(gold);
+    {
+        const(int)[3] a2 = [1, 2, 3].asStatic;
+    }
+
+    [1, 129].asStaticCast!byte.assertSame!byte([1, -127]);
+
+    version (Bug)
+    {
```
wilzbach: This shouldn't be part of the __public__ documentation

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
...
+
+// Note: can't use `version (unittest)` because of https://issues.dlang.org/show_bug.cgi?id=18488
+void assertSame(T, T1, T2)(T1 a, T2 b) @nogc
+{
+    assert(is(T1 == T[T1.length]));
+    assert(a == b);
+}
+
+// TODO: add this to assertThrown in std.exception
+bool isThrown(T : Throwable = Exception, E)(lazy E expression) @system
```
wilzbach: FYI: There's already https://dlang.org/phobos/std_exception.html#ifThrown

```
auto foo() {
   throw new Exception("test");
   return true;
}

ifThrown(foo(), false).writeln;
```

https://run.dlang.io/is/ua4DpZ

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
...
+    import std.range;
+
+    enum a = asStatic!(2.iota);
+    asStatic!(2.iota).assertSame!int([0, 1]);
+    asStaticCast!(double, 2.iota).assertSame!double([0, 1]);
+    asStaticCast!(byte, 2.iota).assertSame!byte([0, 1]);
+}
+
+// Note: can't use `version (unittest)` because of https://issues.dlang.org/show_bug.cgi?id=18488
+void assertSame(T, T1, T2)(T1 a, T2 b) @nogc
```
wilzbach: This is a new public symbol. The tester failed for a _good_ reason. We don't want to use undocumented functionality in public examples.

std/array.d
```diff
@@ -3891,3 +3891,231 @@ unittest
...
+        emplaceRef!U(ret[i++], cast(U) ai);
+    }
+    assert(i == n);
+    return ret;
+}
+
+///
+@safe unittest
+{
+    import std.range;
```
wilzbach: ` : iota` - for public tests.

dukc:
Honestly I saw this only after doing my own which does pretty much the same: #6214. Otherwise I probably would have thought something else.

std/array.d
timotheecour: done

std/array.d
timotheecour: done

std/array.d
timotheecour: done

std/array.d
timotheecour: not nothrow

std/array.d
timotheecour: done

std/array.d
timotheecour: done (was meant as a simple GUID to cross-reference; but removed it)

std/array.d
timotheecour: done

std/array.d
timotheecour: not nothrow

std/array.d
timotheecour: removed its used from documented unittests

timotheecour:
> Honestly I saw this only after doing my own which does pretty much the same: #6214. Otherwise I probably would have thought something else.

I personally don't care which version gets in; but yours seems lacking functionality (See unitttests)

std/array.d
timotheecour: done i think

std/array.d
timotheecour: done

std/array.d
timotheecour: all comments addressed;

timotheecour:
@JackStouffer all comments addressed but the "requested changes" red cross is still showing

timotheecour:
ping

schveiguy:
@JackStouffer I thought that was deprecated?

JackStouffer:
Yes, it is deprecated; good.

JackStouffer:
I would never really use something like this, so I can't comment on the usefulness of the various overloads.

std/array.d
```diff
@@ -3852,3 +3853,286 @@ unittest
     assert(appS.data == "hellow");
     assert(appA.data == "hellow");
 }
+
+/++
+Params: a = The array elements
+
+Returns: A static array constructed from `a`.The type of elements can be
```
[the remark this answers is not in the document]
timotheecour: done

dukc:
@JackStouffer It does not help to mark `dynamicArray.asStatic` `@system`, since you could just as well do `int[2] variable = dynamicArray[]`. The problem lies in assignment of a static array to dynamic one, not in making one. We're barking the wrong tree here.

timotheecour:
ping: is there anything else blocking this? all comments have been addressed and only failure seems unrelated

dukc:
I was told once that new symbols or their overloads need Andrei's approval. He is already assigned here and thus has just probably forgotten to pop up. Email him! That has worked so far.

Andrei@erdani.com is the address.

timotheecour:
I was told once that new symbols or their overloads need Andrei's approval. He is already assigned here and thus has just probably forgotten to pop up. Email him! That has worked so far.

> ping @andralex

andralex:
I think we need to choose this, https://github.com/dlang/phobos/pull/6214, or a careful combination, but not both. Thoughts?

timotheecour:
This PR covers more use cases than #6214 (strict super set it seems), see top-level PR message + documented unittests

dukc:
I agree that this is better in the sense that it allows constructing an array from an arguent tuple, thus enabling easy initialization of static arrays without compromising `auto`, `@nogc` or needing to specify the size explicitly. We definitely should keep at least that part of this one.

Whether `aRange.asStatic!5` or `aRange.array!5` is better, is debatable. The difference is that `array` does not introduce a new symbol but an overload to existing one. You decide whether that's a good or a bad thing.

I would argue that `enumArray.range` is better of the compile time versions from syntax standpoint since it can be used UFCS-style. On the other hand, `asStatic!enumArray` is a whole specialized overload just for the purpose and thus is less likely to accidently introduce an abstraction cost to either the compile-time or the runtime version.

Also these are undeniably better tested than mine.

Should we open a forum discussion from this?

timotheecour:
> Whether aRange.asStatic!5 or aRange.array!5 is better, is debatable. The difference is that array does not introduce a new symbol but an overload to existing one. You decide whether that's a good or a bad thing.

obviously `[1,2].asStatic` can't be renamed `[1,2].array` so we're gonna need `asStatic` (or whatever other new name) if we are to support this use case; so the above point about not introducing a new symbol for `aRange.asStatic!5` is moot.

dukc:
> obviously [1,2].asStatic can't be renamed [1,2].array so we're gonna need asStatic (or whatever
> other new name) if we are to support this use case; so the above point about not introducing a
> new symbol for aRange.asStatic!5 is moot.

True if we want to keep that literal taking version, `[1, 2].asStatic`. But arguably it is unnecessary since you may as well write `staticArray(1, 2)`. It's a bit less UFCS-style, but at least personally I wouldn't find it a big deal since it does not change the order of arguments and would require using braces around them anyway, just different kinds of braces.

dukc:
Sounds like a good option, also. Either way, I'm happy.

dukc:
ping @JackStouffer ping @wilzbach ping @n8sh I believe your thoughts are also wanted here, see last Andrei's comment.

JackStouffer:
See my previous comment; I can't comment on the usefulness of either method because I generally have never found a need for such a function.

timotheecour:
ping, anything else blocking this?

n8sh:
If I was just choosing for myself I would pick this one.

std/array.d
```diff
@@ -3852,3 +3853,286 @@ unittest
...
+}
+
+void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
+{
+    assert(is(T1 == T[T1.length]));
+    assert(a == b);
+}
+
+// TODO: add this to assertThrown in std.exception
+bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow pure @system
```
wilzbach: Should be private for now then.

std/array.d
```diff
@@ -81,6 +81,7 @@ import std.meta;
 import std.traits;
 
 import std.range.primitives;
+import std.conv : emplaceRef;
```
wilzbach: Prefer local imports.

wilzbach:
Hmm, I'm not a fan of these `Cast` overloads. Can't we workaround this with e.g. this?

```d
pragma(inline, true) auto staticArray(T...)(T a) nothrow @safe pure @nogc
if (T.length > 0)
{
    alias U = CommonType!T;
    U[T.length] ret = [a];
    return ret;
}

/// ditto
template staticArray(U)
{
    // Workaround https://issues.dlang.org/show_bug.cgi?id=16779 (make alias to staticArray once fixed)
    pragma(inline, true) U[T.length] staticArray(T...)(T a) nothrow @safe pure @nogc
    //if (!is(CommonType!T == U))
    {
        enum n = T.length;
        U[n] ret = void;
        static foreach (i; 0 .. n)
        {
            // TODO: If any of these throws, the destructors of the already-constructed elements is not called. https://github.com/dlang/phobos/pull/4936/files#r131709329
            emplaceRef!U(ret[i], cast(U) a[i]);
        }
        return ret;
    }
}
```

I like `asStatic` and ideally that's the only symbol which we need to add.

wilzbach:
> Should we open a forum discussion from this?

That's almost never a bad idea.

std/array.d
timotheecour: done

std/array.d
timotheecour: done

timotheecour:
> I like asStatic and ideally that's the only symbol which we need to add.

done, PR now only adds asStatic

PTAL

std/array.d
```diff
@@ -3927,3 +3927,230 @@ unittest
...
+    [1, 129].asStatic!byte.checkStaticArray!byte([1, -127]);
+
+}
+
+/++
+Params: a = input range of elements
+
+Returns: A static array constructed from `a`.
++/
+auto asStatic(size_t n, T)(T a) nothrow @safe pure @nogc
```
wilzbach: Use ditto here too to group the overloads in the documentation.

std/array.d
```diff
@@ -3927,3 +3927,230 @@ unittest
...
+
+nothrow pure @safe unittest
+{
+
+    auto a = [1, 2].asStatic;
+    assert(is(typeof(a) == int[2]) && a == [1, 2]);
+
+    import std.range : iota;
+
+    2.iota.asStatic!2.checkStaticArray!int([0, 1]);
```
wilzbach: Shouldn't there be a test for something like `10.iota.asStatic!5` too?

std/array.d
```diff
@@ -3927,3 +3927,230 @@ unittest
...
+        int[] a2 = [1, 2].asStatic;
+    }
+}
+
+/++
+Params: arr = the compile time range
+
+Returns: A static array constructed from `arr`.
++/
+auto asStatic(alias arr)() nothrow @safe pure @nogc
```
wilzbach: Use ditto here to make the documentation more digestible for readers.

std/array.d
```diff
@@ -3927,3 +3927,230 @@ unittest
...
+{
+    import std.range : iota;
+
+    enum a = asStatic!(2.iota);
+    asStatic!(2.iota).checkStaticArray!int([0, 1]);
+    asStatic!(double, 2.iota).checkStaticArray!double([0, 1]);
+    asStatic!(byte, 2.iota).checkStaticArray!byte([0, 1]);
+}
+
+void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
```
wilzbach: Private and version(unittest)

std/array.d
```diff
@@ -3927,3 +3927,230 @@ unittest
...
+{
+    assert(is(T1 == T[T1.length]));
+    assert(a == b);
+}
+
+// TODO: add this to assertThrown in std.exception
+private bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow pure @system
```
wilzbach: Could be inlined in the single unittest that's using this or use version(unittest)

std/array.d
```diff
@@ -3927,3 +3927,230 @@ unittest
...
+    }
+}
+
+/++
+Params: arr = the compile time range
+
+Returns: A static array constructed from `arr`.
++/
+auto asStatic(alias arr)() nothrow @safe pure @nogc
+{
```
wilzbach: Shouldn't this entire block be in `static immutable = () {...}` to take full advantage of CTFE?

std/array.d
```diff
@@ -3927,3 +3927,230 @@ unittest
...
     assert(appA.data == "hellow");
 }
+
+/++
+Params: arr = The array elements
+
+Returns: A static array constructed from `arr`. The type of elements can be
+specified implicitly (`int[2] a = [1,2].asStatic;`) or explicitly
+(`float[2] a = [1,2].asStatic!float`).
+The result is an rvalue, therefore uses like
```
wilzbach: Rvalue/lvalue are compiler terms and shouldn't be used in user-facing docs.
Better something along the lines of creates a copy.
(BTW IIRC in same cases DMD is smart enough to infer a move.)

std/array.d
wilzbach: The docs should also highlight why/when this overload should be preferred.

n8sh:
>I like asStatic and ideally that's the only symbol which we need to add.

Doesn't that kill the ability to use this in a `@nogc` fashion? :-/

std/array.d
n8sh: As long as he can say the same thing without losing precision. The "users" here are programmers, not potato farmers :-)

std/array.d
timotheecour: these are in the non-documented unittests below:
    assert(isThrown!Error(2.iota.asStatic!1));
    assert(isThrown!Error(2.iota.asStatic!3));
isThrown is a private symbol (as reviews requested) so can't be used in documented unittest

std/array.d
timotheecour: replaced by:

```
auto asStatic(alias a)() nothrow @safe pure @nogc
{
    return .asStatic!(cast(size_t) a.length)(a);
}
```
not sure if that's what you meant but the newer form is indeed much better to avoid duplication

std/array.d
timotheecour: done

std/array.d
timotheecour: done

timotheecour:
>> I like asStatic and ideally that's the only symbol which we need to add.

> Doesn't that kill the ability to use this in a @nogc fashion? :-/

all the asStatic overloads are @nogc, not sure what you mean here

n8sh:
>all the asStatic overloads are `@nogc`, not sure what you mean here

I  mean that I have to construct a dynamic array first.

std/array.d
timotheecour: done

timotheecour:
no, look at unittests:
```
@nogc void checkNogc()
    {
        [1, 2, val].asStatic.checkStaticArray!int(gold);
    }
    checkNogc();
```

timotheecour:
all comments addressed, PTAL

std/array.d
wilzbach: FYI: that was before StdUnittest was reverted: https://github.com/dlang/phobos/pull/6202

(fair enough though, it was just a could be comment)

std/array.d
```diff
@@ -3927,3 +3927,213 @@ unittest
...
+Params: a = The input elements
+
+Returns: A static array constructed from `a`.
+
+The type of elements can be specified implicitly (`int[2] a = [1,2].asStatic;`)
+or explicitly (`float[2] a = [1,2].asStatic!float`). When `a` is a range,
+the number of elements has to be given as template argument (eg `2.iota.asStatic!2`).
+Size and type can be combined (eg: `2.iota.asStatic!(byte[2])`).
+Range `a` can also be specified as a template argument (eg: `asStatic!(2.iota)`
+or `asStatic!(double, 2.iota)`).
```
wilzbach: The body comes before the Params + Returns section (check the other methods in phobos)

std/array.d
```diff
@@ -3927,3 +3927,213 @@ unittest
...
     assert(appA.data == "hellow");
 }
+
+/++
+Params: a = The input elements
+
+Returns: A static array constructed from `a`.
+
+The type of elements can be specified implicitly (`int[2] a = [1,2].asStatic;`)
+or explicitly (`float[2] a = [1,2].asStatic!float`). When `a` is a range,
```
wilzbach: I think we shouldn't use `int[2] a` here as the entire point of asStatic was to be able to use `auto`. Maybe just say that the type is implicitly inferred to `int[2]` and leave the detailed explanation to the unittests?

std/array.d
```diff
@@ -3927,3 +3927,213 @@ unittest
...
+    return ret;
+}
+
+///
+nothrow pure @safe unittest
+{
+    import std.range : iota;
+
+    auto a = 2.iota.asStatic!2;
+    assert(is(typeof(a) == int[2]) && a == [0, 1]);
```
wilzbach: static

std/array.d
```diff
@@ -3927,3 +3927,213 @@ unittest
...
+    size_t i;
+    foreach (ref ai; a)
+    {
+        emplaceRef!U(ret[i++], cast(U) ai);
+    }
+    assert(i == n);
+    return ret;
+}
+
+///
```
wilzbach: You can give unittests a title, e.g. "Explicitly specify the size"

std/array.d
```diff
@@ -3927,3 +3927,213 @@ unittest
...
+    return .asStatic!(U[cast(size_t) a.length])(a);
+}
+
+///
+nothrow pure @safe unittest
+{
+    import std.range : iota;
+
+    enum a = asStatic!(2.iota);
+    assert(is(typeof(a) == int[2]) && a == [0, 1]);
```
wilzbach: static

std/array.d
```diff
@@ -3927,3 +3927,213 @@ unittest
...
+
+    enum a = asStatic!(2.iota);
+    asStatic!(2.iota).checkStaticArray!int([0, 1]);
+    asStatic!(double, 2.iota).checkStaticArray!double([0, 1]);
+    asStatic!(byte, 2.iota).checkStaticArray!byte([0, 1]);
+}
+
+private void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
+{
+    assert(is(T1 == T[T1.length]));
```
wilzbach: static

std/array.d
```diff
@@ -3927,3 +3927,213 @@ unittest
...
+
+/++
+Params: a = The input elements
+
+Returns: A static array constructed from `a`.
+
+The type of elements can be specified implicitly (`int[2] a = [1,2].asStatic;`)
+or explicitly (`float[2] a = [1,2].asStatic!float`). When `a` is a range,
+the number of elements has to be given as template argument (eg `2.iota.asStatic!2`).
+Size and type can be combined (eg: `2.iota.asStatic!(byte[2])`).
```
wilzbach: Is there any use case in which `3.iota.asStatic` would be preferred over `asStatic!(3.iota)`?

std/array.d
```diff
@@ -3927,3 +3927,213 @@ unittest
...
+
+Note: `foo([1, 2, 3].asStatic)` may be inefficient because of the copies involved.
++/
+pragma(inline, true) T[n] asStatic(T, size_t n)(auto ref T[n] a) nothrow @safe pure @nogc
+{
+    return a;
+}
+
+/// ditto
+U[n] asStatic(U, T, size_t n)(auto ref T[n] a) nothrow @safe pure @nogc
```
wilzbach: Explicitly specifying these attributes has the downside of erroring if a user has an unsafe destructor. The compiler already infers the attributes for you and you check this with your unittests, so what's the motivation for being explicit here?

std/array.d
```diff
@@ -3927,3 +3927,213 @@ unittest
...
+        emplaceRef!U(ret[i], cast(U) a[i]);
+    }
+    return ret;
+}
+
+///
+nothrow pure @safe unittest
+{
+    auto a = [0, 1].asStatic;
+    assert(is(typeof(a) == int[2]) && a == [0, 1]);
```
wilzbach: static

std/array.d
timotheecour: a == [0,1] can't be known at CT; i could split in 2 assert but not sure what's the advantage

std/array.d
timotheecour: when using a runtime range `r` we can't use `asStatic!r` ; I've changed document unittest to use:
```
    auto input = 2.iota;
    auto a = input.asStatic!2;
```
instead of
```
    auto a = 2.iota.asStatic!2;
```

std/array.d
timotheecour: done

std/array.d
timotheecour: done

std/array.d
timotheecour: done

std/array.d
timotheecour: done

std/array.d
```diff
@@ -3927,3 +3927,214 @@ unittest
     assert(appS.data == "hellow");
     assert(appA.data == "hellow");
 }
+
+/++
+Constructs a static array from `a`.
+The type of elements can be specified implicitly (`auto a = [1,2].asStatic;` of type int[2])
+or explicitly (`auto a = [1,2].asStatic!float` of type float[2]). When `a` is a range,
```
wilzbach: Inconsistency: you use once a semi-colon and in the next example you don't. I prefer it without the semi-colon, but I don't care as long as it's consistent.

wilzbach:
> a == [0,1] can't be known at CT; i could split in 2 assert but not sure what's the advantage

- `assert(a == [0, 1])` can be rewritten by the assert/writeln pipeline during the documentation build (applies only for public unittests)
- it's a bit easier to read if you separate CT asserts with runtime asserts

std/array.d
timotheecour: done, note that this rule isn't always observed eg:
```
auto uninitializedArray(T, I...)(I sizes) nothrow @system
std/complex.d:37:24:auto complex(R)(R re)  @safe pure nothrow @nogc
pure bool isLoggingEnabled()(LogLevel ll) @safe nothrow @nogc
std/math.d:646:23:auto conj(Num)(Num z) @safe pure nothrow @nogc
```

std/array.d
wilzbach: Oh I was asking this out of the perspective of a potential reader as the documentation doesn't make this clear.

I would do something like "if the input range is known at compile-time ... can be used. If the range content is dynamic and not known at compile-time, the size needs to be set explicitly with e.g. ..."

std/array.d
wilzbach: Well not all phobos code is created equal. There's a lot of old, ugly legacy code in Phobos (though it's not as bad as DMD).
Anyhow, fixes for the lines you observed are always welcome.

std/array.d
timotheecour: done (for cases in public unittest only since assert/writeln pipeline only applies to these)

std/array.d
timotheecour: done

timotheecour:
/cc @wilzbach @n8sh @JackStouffer @ZombineDev @andralex
all comments addressed, PTAL

wilzbach:
Cool! Apart from documentation nitpicks and a changelog entry, I don't see any blocking issue. I am starting to really like this proposed method. Thanks a lot for the hard work!

std/array.d
```diff
@@ -3927,3 +3927,220 @@ unittest
...
+    2.iota.asStatic!(byte[2]).checkStaticArray!byte([0, 1]);
+}
+
+nothrow pure @system unittest
+{
+    import std.range : iota;
+
+    assert(isThrown!Error(2.iota.asStatic!1));
+    assert(isThrown!Error(2.iota.asStatic!3));
+    // NOTE: alternatives are not nothrow:
```
wilzbach: That's a infamous bug in the inference of lazy parameters. The workaround is `scope(failure) assert(0);`

It's probably a good idea to either use that workaround or remove this block as it otherwise is dead code.

std/array.d
```diff
@@ -3927,3 +3927,220 @@ unittest
...
+    version (Bug)
+    {
+        // https://issues.dlang.org/show_bug.cgi?id=16779
+        auto a2 = [1, 2, 3].asStatic!byte;
+        auto a3 = [1, 2, 3].asStatic!ubyte;
+
+        // NOTE: correctly issues a deprecation
+        int[] a2 = [1, 2].asStatic;
+    }
+}
```
wilzbach: Not sure whether it's a good idea to include such dead blocks. Maybe open a Bugzilla issue to fix asStatic that depends on 16779 so that this isn't forgotten?

std/array.d
```diff
@@ -3927,3 +3927,220 @@ unittest
...
+
+private void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
+{
+    static assert(is(T1 == T[T1.length]));
+    assert(a == b);
+}
+
+// TODO: consider adding this to assertThrown in std.exception
+private bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow
```
wilzbach: The underlying compiler bug is https://issues.dlang.org/show_bug.cgi?id=12647

std/array.d
wilzbach: https://issues.dlang.org/show_bug.cgi?id=12647

std/array.d
timotheecour: done

std/array.d
timotheecour: ack

std/array.d
timotheecour: done (reworded ddoc)

std/array.d
timotheecour: done (added link in comment)

std/array.d
timotheecour: actually 16779 was for VRP on staticArray(T...)(T a) which I removed (i just have asStatic now), so removed that;
also, folded other dead block into a comment in preceding unittest.

__traits(deprecated) would be the optimal (see https://dlang.slack.com/archives/D8ZMKRGUV/p1519414575000349)

std/array.d
timotheecour: done

timotheecour:
addressed all remaining doc nitpicks and added changelog entry

std/array.d
timotheecour: => filed https://issues.dlang.org/show_bug.cgi?id=18618 Issue 18618 - templated functions should in general have their attributes inferred (edit)

std/array.d
```diff
@@ -3927,3 +3927,215 @@ unittest
...
+{
+    import std.range : iota;
+
+    enum a = asStatic!(2.iota);
+    asStatic!(2.iota).checkStaticArray!int([0, 1]);
+    asStatic!(double, 2.iota).checkStaticArray!double([0, 1]);
+    asStatic!(byte, 2.iota).checkStaticArray!byte([0, 1]);
+}
+
+private void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure @nogc
```
n8sh: Can this be `version(unittest)`?

timotheecour:
> +private void checkStaticArray(T, T1, T2)(T1 a, T2 b) nothrow @safe pure
@nogc
> Can this be version(unittest)?

@n8sh, I'm curious why I can't see your post on github (only via email) ?

timotheecour:
waiting for @andralex approval, @wilzbach already approved it

wilzbach:
Here's an idea to avoid being blocked on Andrei's approval for new symbols: we merge this as is to std.experimental and keep it there for one release. We then only need Andrei's approval once every two months when we merge everything from std.experimental to the standard.
This would also help in marking the new addition as unstable and help to avoid breakage if any issues are discovered.
We haven't done this before, but I think we could give it a try.
If this is going to experimental we could immediately merge this and imho having it wait in experimental is a lot better than it sitting in the PR queue forever.

wilzbach:
More ideas and a summary:

### 1) `std.experimental.staging`

As discussed a module in `std.experimental` for upcoming symbols.
There's a bit of a precedent with e.g. https://dlang.org/phobos/std_experimental_typecons.html

Pros:
- easy to find out which symbols are experimentally used
- easy to find which symbols are currently pending

Cons:
- source code needs to be modified when the symbol gets stabilized, but `alias` + deprecations could be used

It's debatable whether it should be `std.experimental.<moduleName>` or `std.experimental.staging` as both have slight advantages/disadvantages.
The one is a bit more explicit, but requires more typing and interaction from the user. As it also requires more effort on our side to create a module every time a new symbol is introduced, I thing a generic catch-all staging module is a good trade-off.

### 2) Mark it in the documentation as experimental

see e.g. https://github.com/rust-lang/rust/pull/44015

```
#[stable(feature = "indirect_hasher_impl", since = "1.22.0")]
impl<T: ?Sized + Hasher> Hasher for Box<T> {
```

Pros: 
- source code doesn't need to be modified when the symbol gets stabilized

Cons:
- accidental usage
- if someone uses an experimental symbol, this might not be clear to reviewers of the user-code

### 3) Only expose it via a `-experimental` flag (or `-version=StdExperimental`) for that matter

```
version(StdExperimental)
{
...
}
```

The flag would be added to the default build flag in the Phobos Makefiles, but the symbol wouldn't be accessible without this flag to the user.

Pros:
- source code doesn't need to be modified when the symbol gets stabilized
- explicit opt-in into experimental symbols once

Cons:
- a bit more overhead to maintain which symbols are currently unstable, but we could use an `@@@STABLE_2.082@@@` pattern like we do for deprecations (or just maintain a file in which unstable symbols are listed)
- might require a bit of documentation to explain that `-version(StdExperimental)` needs to be used (but that could be part of a default `$(UNSTABLE)` macro.

### 4) Put experimental functions into a `stdx` dub package

Example: https://github.com/dlang-stdx/collections

Pros:
- doesn't break user code if new symbols are added
- even easier to merge as it's completely decoupled from Phobos

Cons:
- a bit of overhead in creating the build tools
- might not be as visible as Phobos
- there are still some people who don't like dub (or don't want to/can't use dub)
- might introduce a bit of overhead in keeping it working with newer compiler versions (but there's the project tester for this)

--- 

Personally I would be fine with __any__ of these options as long as we make progress and don't let PRs die in the queue forever.

dukc:
I prefer option 2. If it's good enough for web api (https://developer.mozilla.org/en-US/docs/Web/API) I don't see why it wouldn't be for us. And probably causes the least "drag" in getting stuff in.

std/array.d
```diff
@@ -3927,3 +3927,215 @@ unittest
...
+{
+    return a;
+}
+
+/// ditto
+U[n] asStatic(U, T, size_t n)(auto ref T[n] a)
+if (!is(U == T))
+{
+    import std.conv : emplaceRef;
+    U[n] ret = void;
```
andralex: This function has a pernicious bug: if construction throws an exception, `ret` will be destroyed and destructors will be called for uninitialized objects. Recall that statically-sized arrays on the stack insert one destructor call for each member.

std/array.d
```diff
@@ -3927,3 +3927,215 @@ unittest
...
+    {
+        const(int)[3] a2 = [1, 2, 3].asStatic;
+    }
+
+    [1, 129].asStatic!byte.checkStaticArray!byte([1, -127]);
+
+}
+
+/// ditto
+auto asStatic(size_t n, T)(T a)
```
andralex: Template constraint?

std/array.d
```diff
@@ -3927,3 +3927,215 @@ unittest
...
+{
+    import std.conv : emplaceRef;
+    U[n] ret = void;
+    size_t i;
+    foreach (ref ai; a)
+    {
+        emplaceRef!U(ret[i++], cast(U) ai);
+    }
+    assert(i == n);
+    return ret;
```
andralex: This code duplicates implementation (and bug) from a different overload.

std/array.d
```diff
@@ -3927,3 +3927,215 @@ unittest
...
+    {
+        emplaceRef!U(ret[i++], ai);
+    }
+    assert(i == n);
+    return ret;
+}
+
+/// ditto
+auto asStatic(Un : U[n], U, size_t n, T)(T a) nothrow @safe pure @nogc
+{
```
andralex: constraint?

std/array.d
```diff
@@ -3927,3 +3927,215 @@ unittest
...
+
+    assert(isThrown!Error(2.iota.asStatic!1));
+    assert(isThrown!Error(2.iota.asStatic!3));
+
+    // NOTE: correctly issues a deprecation
+    // int[] a2 = [1, 2].asStatic;
+}
+
+/// ditto
+auto asStatic(alias a)()
```
andralex: constraint?

std/array.d
```diff
@@ -3927,3 +3927,215 @@ unittest
...
+{
+    return .asStatic!(cast(size_t) a.length)(a);
```
andralex: why is a cast needed?

std/array.d
```diff
@@ -3927,3 +3927,215 @@ unittest
...
+}
+
+/// ditto
+auto asStatic(U, alias a)()
```
andralex: constraint

std/array.d
```diff
@@ -3927,3 +3927,215 @@ unittest
...
+assertThrown!Error(expr_that_throws);
+
+// or
+import std.exception : ifThrown;
+scope(failure) assert(0);
+assert(ifThrown!Error({ expr_that_throws; return false; }(), true));
+---
+Also, `isThrown` is easier to use than ifThrown and more flexible than assertThrown.
++/
+version(unittest) private bool isThrown(T : Throwable = Exception, E)(lazy E expression) nothrow
```
andralex: nice

std/array.d
```diff
@@ -3927,3 +3927,215 @@ unittest
...
+
+Returns: A static array constructed from `a`.
++/
+pragma(inline, true) T[n] asStatic(T, size_t n)(auto ref T[n] a) nothrow @safe pure @nogc
+{
+    return a;
+}
+
+/// ditto
+U[n] asStatic(U, T, size_t n)(auto ref T[n] a)
```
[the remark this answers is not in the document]
timotheecour: I can replace with 
```
template staticArray(U...) if (U.length == 0 || U.length == 1 && is(U[0]))
{
    static if (U.length == 0)
    {
        pragma(inline, true) T[n] staticArray(T, size_t n)(auto ref T[n] a) nothrow @safe pure @nogc
        {
            return a;
        }
    }
    else
    {
        alias U0 = U[0];
        U0[n] staticArray(T, size_t n)(auto ref T[n] a)
        {
            import std.conv : emplaceRef;

            U0[n] ret = void;
            static foreach (i; 0 .. n)
            {
                emplaceRef!U0(ret[i], cast(U0) a[i]);
            }
            return ret;
        }
    }
}
```

but I don't see in what ways it improves over what I had:

```
pragma(inline, true) T[n] staticArray(T, size_t n)(auto ref T[n] a) nothrow @safe pure @nogc
{
    return a;
}

/// ditto
U[n] staticArray(U, T, size_t n)(auto ref T[n] a)
if (!is(U == T))
{
    import std.conv : emplaceRef;
    U[n] ret = void;
    static foreach (i; 0 .. n)
    {
        emplaceRef!U(ret[i], cast(U) a[i]);
    }
    return ret;
}
```

(more lines of code, more complexity, less readable IMO)

std/array.d
timotheecour: otherwise:
Error: template std.array.staticArray cannot deduce function from argument types !(length)(Result)

std/array.d
timotheecour: ack

std/array.d
```diff
@@ -3927,3 +3927,215 @@ unittest
...
+The type of elements can be specified implicitly (`[1,2].asStatic` of type int[2])
+or explicitly (`[1,2].asStatic!float` of type float[2]).
+When `a` is a range (not known at compile time), the number of elements has to be given as template argument
+(eg `myrange.asStatic!2`).
+Size and type can be combined (eg: `2.iota.asStatic!(byte[2])`).
+When the range `a` is known at compile time, it can also be specified as a
+template argument to avoid having to specify the number of elements
+(eg: `asStatic!(2.iota)` or `asStatic!(double, 2.iota)`).
+
+Note: `foo([1, 2, 3].asStatic)` may be inefficient because of the copies involved.
```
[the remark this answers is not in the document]
timotheecour: done

std/array.d
timotheecour: done

std/array.d
timotheecour: done

std/array.d
timotheecour: done

std/array.d
timotheecour: done

std/array.d
timotheecour: avoided this bug by making these overloads as `nothrow`, so a throwing ctor will cause CT error instead of potential runtime bug.

support for throwing constructors can always be addressed in a future PR (without breaking backward compatibility), maybe something along these lines:

```
U[n] staticArray(U, T, size_t n)(auto ref T[n] a)
if (!is(U == T))
{
     import std.conv : emplaceRef;

     U[n] ret = void;
    size_t i;
    try
     {
        for (; i < n; i++)
        {
            emplaceRef!U(ret[i], cast(U) a[i]);
        }
    }
    catch(Exception t) // catch(Throwable t) causes errors with D linter; but we really want Throwable here for correctness
    {
        for (; i < n; i++)
        {
            emplaceRef!U(ret[i], U.init);
        }
        throw t;
     }
     return ret;
 }
```

timotheecour:
PTAL, all comments addressed

/cc @andralex not sure how one wants to treat cases where ForEachType != ElementType, eg:
"ab∫".staticArray currently is immutable(char[5]) instead of immutable(dchar[3])

andralex:
@timotheecour on strings we should follow the rule of least astonishment - i.e. if you start with a narrow string you get a static string of the same narrowness. Qualifier on the characters should be the same, but no need to make the entire result qualified, i.e. `immutable(char)[5]`, not `immutable(char[5])`.

wilzbach:
Can we please merge this to `std.experimental` (or [similar](https://github.com/dlang/phobos/pull/6178#issuecomment-374574728))?

There are tons of places within Phobos where this would be very useful and we can work out the exact specific and nitpick the name while we get experience using it.

I think it goes without saying, but it's a very powerful tool for convenient `@nogc` usage.

std/array.d
```diff
@@ -3927,3 +3927,218 @@ unittest
...
+Returns: A static array constructed from `a`.
++/
+
+pragma(inline, true) T[n] staticArray(T, size_t n)(auto ref T[n] a) nothrow @safe pure @nogc
+{
+    return a;
+}
+
+/// ditto
+U[n] staticArray(U, T, size_t n)(auto ref T[n] a) nothrow
```
andralex: I'd accept the argument "let's leave it nothrow and worry about it later) if (a) this PR were urgent, and/or (b) the right code would be difficult to implement. Neither is applicable. The right code is (stylized):

```d
U[n] staticArray(U, T, size_t n)(auto ref T[n] a)
{
    union Temp { U[n] payload; }
    Temp result = void;
    size_t i = 0;
    static if (hasElaborateDestructor!U) scope (exit)
    {
        foreach (j; 0 .. i)
            destroy(result.payload[j]);
    }
    for (; i != n; ++i)
        emplaceRef!U(result.payload[i], cast(U) a[i]);
    return result.payload;
}
```

That's awkward and disables named return value optimization, so here's what I think would be a better approach:

```d
U[n] staticArray(U, T, size_t n)(auto ref T[n] a)
{
    U[n] result = void;
    size_t i = 0;
    static if (hasElaborateDestructor!U) scope (failure)
    {
        foreach (j; i .. n)
            emplace(result[j]); // initialize with T.init to make sure is destroyable
    }
    for (; i != n; ++i)
        emplaceRef!U(result[i], cast(U) a[i]);
    return result;
}
```

This is cute and hopefully not too much so - it comes from the opposite direction: in case of an exception, it initializes (in a nothrow manner) the uninitialized part of the result and then leaves the rest to the compiler. This allows NRVO, is efficient on the nothrow path, and is simple.

std/array.d
```diff
@@ -3927,3 +3927,218 @@ unittest
...
+if (!is(U == T))
+{
+    import std.conv : emplaceRef;
+
+    U[n] ret = void;
+    // TODO: ElementType vs ForeachType
+    static foreach (i; 0 .. n)
+    {
+        emplaceRef!U(ret[i], cast(U) a[i]);
```
andralex: A cast here is dangerous, especially since it's virtually undocumented. I understand the necessity is to create arrays of `short`, `float` etc. from literals. But we can't afford to allow innocent syntax like `[1, 2, 3].asStatic!Object` to disastrously pass. 

I'll "ask for changes" on account of this.

I recall you had a variant `asStaticCast` that took care of this. We may need to return to that, but probably we could find a better idea. We could special-case numbers at most, but not allow unchecked casts. Another possibility is to use `to`.

This is the kind of decision that can be deferred - I suggest you remove this overload entirely, get this in, then add something later.

std/array.d
```diff
@@ -3927,3 +3927,218 @@ unittest
...
+    return staticArray!(U[n], U, n)(a);
+}
+
+/// ditto
+auto staticArray(Un : U[n], U, size_t n, T)(T a) nothrow
+if (isInputRange!T)
+{
+    import std.conv : emplaceRef;
+
+    U[n] ret = void;
```
andralex: same discussion of nothrow

std/array.d
```diff
@@ -3927,3 +3927,218 @@ unittest
...
+    size_t i;
+    foreach (ref ai; a)
+    {
+        emplaceRef!U(ret[i++], cast(U) ai);
```
andralex: DEFINITELY no cast here

std/array.d
```diff
@@ -3927,3 +3927,218 @@ unittest
...
+    }
+    assert(i == n);
```
andralex: This is definitely unsatisfactory. Code shouldn't rely on bounds checking to be correct (see https://dlang.org/spec/arrays.html#bounds), and by the time the assert is even looked at the code has trampled well into infested territory.

Too short is not good, either, because there'd be portions of the array that are left uninitialized! This function is thoroughly problematic.

One solution I suggested to @n8sh: define an overload taking an `out size_t` informing the user on how many elements have been initialized, and initialize the rest to `U.init`. To address the other problem (input range too long), we could probably accept truncation. After all the user only seems to be interested in a fragment of the range. They could check the length externally if they cared. All told:

```d
auto staticArray(Un : U[n], U, size_t n, T)(T a, out size_t filled)
{
    U[n] result = void;
    size_t i = 0;
    // same exception safety mechanism as above
    foreach (ref ai; a)
    {
        if (i == n) break;
        emplace(...);
    }
    filled = i;
    return result;
}

auto staticArray(Un : U[n], U, size_t n, T)(auto ref T a)
{
    size_t unused;
    return staticArray!(Un, U, n, T)(a, unused);
}
```

dukc:
@andralex to speed things up, please tell us that once those are resolved, do you want to recheck the code yourself, or is it enough for the other reviewers to accept it?

andralex:
@dukc I'd like to take a last look, but the more good reviews the better.

std/array.d
andralex: Forgot to mention, some ranges return an rvalue from `front` so using `foreach` is problematic with those. It's simplest to use `for`.

andralex:
@timotheecour why the closure? this was a few touches away from merging.

timotheecour:
I may reopen if I find some time later, but right now I don't have time so I don't want to pollute the PR queue

andralex:
@timotheecour cool, thanks. Would you allow someone else to take it over? Perhaps @dukc ?

dukc:
I was actually considering. I also have more work now than before, but not so much that it's out of question. If this is still waiting the next time I decide to do something, I'll take this.

andralex:
@dukc thanks!

timotheecour:
@andralex @dukc

> @timotheecour cool, thanks. Would you allow someone else to take it over? Perhaps @dukc ?

 of course

PR #6179, opened 2018-02-15
Fix issue 18438 - Handle T* in hasStaticMember like hasMember

aliak00:
hasMember treats T* as T and has no failures. hasStaticMember fails with a T* because of it's use of __traits(getMember. Can be fixed by explicitly handling pointer T.

Either that or hasMember on a T* should maybe instead fail as hasStaticMember does?

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -3644,10 +3644,15 @@ template hasStaticMember(T, string member)
 {
     static if (__traits(hasMember, T, member))
     {
-        import std.meta : Alias;
-        alias sym = Alias!(__traits(getMember, T, member));
+        static if (isPointer!T)
+            alias U = PointerTarget!T;
+        else
+            alias U = T;
 
-        static if (__traits(getOverloads, T, member).length == 0)
+        import std.meta : Alias;
+        alias sym = Alias!(__traits(getMember, U, member));
+
+        static if (__traits(getOverloads, U, member).length == 0)
             enum bool hasStaticMember = __traits(compiles, &sym);
         else
             enum bool hasStaticMember = __traits(isStaticFunction, sym);
@@ -3757,6 +3762,7 @@ template hasStaticMember(T, string member)
         static @property int sp();
     }
 
+    static assert(!hasStaticMember!(S, "na"));
     static assert(!hasStaticMember!(S, "X"));
     static assert(!hasStaticMember!(S, "Y"));
     static assert(!hasStaticMember!(S, "Y.i"));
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -3804,6 +3811,32 @@ template hasStaticMember(T, string member)
...
+    static assert(!hasStaticMember!(P, "g"));
+    static assert( hasStaticMember!(P, "fp"));
+    static assert( hasStaticMember!(P, "gfp"));
+    static assert(!hasStaticMember!(P, "fpm"));
+    static assert(!hasStaticMember!(P, "m"));
+    static assert(!hasStaticMember!(P, "m2"));
+    static assert(!hasStaticMember!(P, "iom"));
+    static assert( hasStaticMember!(P, "iosf"));
+    static assert(!hasStaticMember!(P, "p"));
+    static assert( hasStaticMember!(P, "sp"));
```
aliak00: Maybe all these tests are unnecessary to just see if T* doesn't fail?

wilzbach: No more tests are almost always helpful.

jmdavis:
Honestly, I think that it would be cleaner if `hasMember` didn't work with `T*`, because the members are for `T`, not `T*`, just like it wouldn't make sense IMHO for it to work with something like `T[]`. However, if `hasMember` works with `T*`, then making it not work with `T*` risks breaking code, which we generally try to avoid, so as ugly as it is that `hasMember` works with `T*`, it probably is better to make `hasStaticMember` consistent and let it accept `T*`. However, I'm inclined to argue that the problem here is caused by a compiler bug - or at least an inconsistency in the current implementation - since `hasMember` uses `__traits(hasMember, ...)`, whereas hasStaticMember uses `__traits(getMember, ...)` and `__traits(getOverloads, ...)`, and why would the compiler allow `T*` to work with one of those but not the others?

So, this seems like a reasonable workaround, but I think that the inconsistency with `__traits` should be reported as a compiler bug.

aliak00:
> Honestly, I think that it would be cleaner if hasMember didn't work with T*

Agreed.

> However, I'm inclined to argue that the problem here is caused by a compiler bug - or at least an inconsistency in the current implementation...

I want to say that the bug would be hasMember working for T*. But, then again, D does not have different conventions for calling a field on T or T*, so I'm not sure what would be correct, or where the bug, if there is one, is.

aliak00:
`object.Exception@../tools/changed.d(211): Changelog entries should consist of one title line, a blank separator line, and a description.` ?? 

Should I do something to fix this?

PR #6183, opened 2018-02-15
std.datetime.timezone - fix a -dip1000 compilable issue; trivial

carblue:
A call to std.stdio.writefln!(char, string) in std/datetime/timezone.d within @safe unittest{...}, when compiled with -dip1000 via make -f posix.mak std/datetime/timezone.test,
raises following error (for some yet unknown reason not with switch -dip25):
std/datetime/timezone.d(552): Error: [manually/temporarily attributed @safe] function std.datetime.timezone.TimeZone.__unittest_L546_C22.testPZSuccess cannot call @system function std.stdio.writefln!(char, string).writefln

Thus the fix trivially replaces writefln by writeln.

### DIFF
--- a/std/datetime/timezone.d
+++ b/std/datetime/timezone.d
@@ -546,10 +546,10 @@ public:
     deprecated @safe unittest
     {
         import std.exception : assertNotThrown;
-        import std.stdio : writefln;
+        import std.stdio : writeln;
         static void testPZSuccess(string tzName)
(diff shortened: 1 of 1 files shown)


jmdavis:
Given that this is in a deprecated test, I don't really care, but otherwise, I'd argue that the root problem should be fixed rather than changing the code.

PR #6184, opened 2018-02-15
Expose the enforce overload on the docs

wilzbach:
Note this isn't ideal yet, because `enforce` will be displayed after the the following two overloads but it already improves the status quo and for now I don't have good ideas that can be easily done.


JackStouffer:
@wilzbach Can you somehow force jenkins to build?

PR #6185, opened 2018-02-15
std.regex: Alignment hotfix

kinke:
This fixes the unittests for LDC on ARM (with enabled optimizations).

The code used to store a class instance (whose size isn't padded) right before some buffer, without any padding inbetween. The buffer is thus susceptible to misalignment; e.g., `BacktrackingMatcher.dupTo()` casts the buffer from `void[]` to `size_t[]` (in `initExternalMemory()`), simply assuming an alignment >= `size_t.alignof`, which only holds if the prepended class instance size is a multiple of that (as `malloc()` returns a sufficiently aligned block already in this case).

### DIFF
--- a/std/regex/internal/ir.d
+++ b/std/regex/internal/ir.d
@@ -445,7 +445,8 @@ abstract class GenericFactory(alias EngineType, Char) : MatcherFactory!Char
 {
     import core.stdc.stdlib : malloc, free;
     import core.memory : GC;
-    enum classSize = __traits(classInstanceSize, EngineType!Char);
+    // round up to next multiple of size_t for alignment purposes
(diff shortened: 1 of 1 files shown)


joakim-noah:
The recent `std.regex` redesign in #5722 introduced this change that broke many tests on ARM. @DmitryOlshansky, this fixes it for me on Android/ARM.

DmitryOlshansky:
> The recent std.regex redesign in #5722 introduced this change that broke many tests on ARM. @DmitryOlshansky, this fixes it for me on Android/ARM.

LGTM. Indeed it ddin't occur to me that class instance size != align instance size. I think for structs it accounts for that.

DmitryOlshansky:
@kinke I think there could be more uses of classInstanceSize, might be worth it to double check.

PR #6186, opened 2018-02-16
Renaming the upcoming global import file to std.experimental.all

wilzbach:
I just tried to explain this to someone as he was confused by `scripting`.

`scripting` was used because it was the first name suggested by @andralex, but while I don't hope that this will stay in `experimental` for too long, it's probably a good idea to use `all` is it will immediately tell every reader what's going on. `scripting` is rather vague.

Credits: this has been suggested by @carun here https://github.com/dlang/phobos/pull/5916#discussion_r167300850

I know the name doesn't matter too much and we ideally can use `import std` soon, but it was a quick five minute change and will hopefully lead to less confusion.

See also: https://github.com/dlang/phobos/pull/5916 (PR that introduced `std.experimental.scripting`)

### DIFF
--- a/changelog/std-experimental-all.dd
+++ b/changelog/std-experimental-all.dd
@@ -1,10 +1,10 @@
-`import std.experimental.scripting` as a global convenience import
+`import std.experimental.all` as a global convenience import
 
-$(MREF std,experimental,scripting) allows convenient use of all Phobos modules
+$(MREF std,experimental,all) allows convenient use of all Phobos modules
 with one import:
 
 ---
-import std.experimental.scripting;
+import std.experimental.all;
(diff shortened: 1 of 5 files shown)


ZombineDev:
Agreed, `.scripting` is counter-effective - it's neither self-explanatory, nor as easy to type for one-off code snippets (the primary use case for the package).

wilzbach:
FreeBSD 32-bit is broken at the moment - see braddr/d-tester#70
We need to move forward, so I'm going to merge this PR.
For the record - it passes on Jenkins too:

image

But two project builds were "aborted" from some unknown, unrelated reason.

PR #6187, opened 2018-02-16
Make examples of BitFields more digestable for a reader

wilzbach:
Motivation: https://dlang.org/phobos/std_typecons.html#BitFlags

The idea:
- present examples in small chunks, s.t. a reader doesn't need to study 79 lines (!) to grasp the context
- explain _one_ feature per example
- order examples in chronical order (e.g. first introduce `&`, then use it)

This doesn't aim to be perfect, but better than the existing documentation.

Other remarks: Most variable names here use the non DStyle snake_case, e.g. `flags_empty` or `flags_AB`.
I left renaming them to another PR as this PR is purely focussed on reorganizing the content to be better digestable.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -7446,12 +7446,12 @@ template isBitFlagEnum(E)
     {
         enum isBitFlagEnum = (E.min >= 0) &&
         {
-            foreach (immutable flag; EnumMembers!E)
-            {
+            static foreach (immutable flag; EnumMembers!E)
+            {{
                 Base value = flag;
                 value &= value - 1;
                 if (value != 0) return false;
-            }
+            }}
             return true;
         }();
     }
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -7787,7 +7787,7 @@ public:
     assert(value == Enum.A);
 }
 
-/// You need to specify the `unsafe` parameter for enum with custom values
+/// You need to specify the `unsafe` parameter for enums with custom values
 @safe @nogc pure nothrow unittest
 {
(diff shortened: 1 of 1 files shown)


std/typecons.d
```diff
@@ -7446,12 +7446,12 @@ template isBitFlagEnum(E)
     {
         enum isBitFlagEnum = (E.min >= 0) &&
         {
-            foreach (immutable flag; EnumMembers!E)
-            {
+            static foreach (immutable flag; EnumMembers!E)
```
MetaLang: I feel that this change is a net negative. We don't want a user to have to puzzle out why the double braces are here and what they mean while also trying to learn what `BitFlags` is for and how to use it.

std/typecons.d
```diff
@@ -7647,90 +7655,152 @@ public:
...
+    assert(flags & flags_AB);
+    assert(flags & Enum.A);
 
     // Finally, you can of course get you raw value out of flags
-    auto value = cast(int) flags_A;
+    auto value = cast(int) flags;
     assert(value == Enum.A);
 }
 
+/// You need to specify the `unsafe` parameter for enum with custom values
```
MetaLang: `enum` -> `enums`

std/typecons.d
wilzbach: This isn't an example. It's the actual implementation. I just spotted a `static foreach` here and included it as a separate commit.

wilzbach:
Thanks a lot for the review, fix-up and quick-merge @MetaLang! Really appreciated :)

PR #6188, opened 2018-02-16
Add `TailShared` to `std.traits`

thaven:
While working on some thread-safe code, I felt an easy way to unshare just the head of a type was missing. But when looking at the docs of `core.atomic` I saw references to TailShared, which happens to do just that, but is private to that module.

Now I could propose to make it public there... but it actually it hasn't much to do with atomicity, so `core.atomic` did not feel like the right place to me. I'd look in `std.traits` first for a thing like that. But moving it there is also wrong, because druntime should not depend on phobos.

So I decided to let `std.traits` steal it from `core.atomic`. A bit ugly in implementation, but I think it results in the best API.

### DIFF
--- a/changelog/std-traits-tailshared.dd
+++ b/changelog/std-traits-tailshared.dd
@@ -1,0 +1,11 @@
+`TailShared` was added to `std.traits`
+
+$(REF TailShared, std, traits) changes a type's qualifiers such that its tail is shared, but its head isn't.
+
+---
+import std.traits;
+
+static assert(is(TailShared!int == int));
+static assert(is(TailShared!(int*) == shared(int)*));
+static assert(is(TailShared!(shared int**) == shared(int*)*));
+---
--- a/std/traits.d
+++ b/std/traits.d
@@ -128,6 +128,7 @@
  *           $(LREF OriginalType)
  *           $(LREF PointerTarget)
  *           $(LREF Signed)
+ *           $(LREF TailShared)
  *           $(LREF Unqual)
  *           $(LREF Unsigned)
  *           $(LREF ValueType)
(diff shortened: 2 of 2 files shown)


std/traits.d
```diff
@@ -7025,6 +7026,41 @@ template Unqual(T)
     static assert(is(Unqual!ImmIntArr == immutable(int)[]));
 }
 
+/**
+Construct a type with a shared tail, and if possible with an unshared head.
+
+If $(D T) does not have indirections, or $(D T) already has an unshared head and
+a shared tail, $(D TailShared) will alias itself away.
```
edi33416: Could you use backticks `` ` `` instead of `$(D` for single words? I know there is a mix between the two styles, but new ddocs should use `` ` ``.

std/traits.d
```diff
@@ -7025,6 +7026,41 @@ template Unqual(T)
...
+{
+    static assert(is(TailShared!(shared int) == int));
+    static assert(is(TailShared!(shared int*) == shared(int)*));
+
+    static class C { int i; }
+    static assert(is(TailShared!C == shared C));
+
+    static struct S3 { int* p; int i; }
+    static assert(!is(TailShared!S3 : S3));
+    static assert(is(TailShared!S3 : shared S3));
```
edi33416: Is there a reason you are using the `:` operator instead of `==` inside the is expression? Is there a difference between how `TailShared` works with `class`es and `struct`s?

thaven: There is indeed a difference between `class` and `struct` types here. A `class` is a reference type, and a `struct` is a value type. So, given `class C` and `struct S`, `shared(C)` means that both the instance itself and its reference are `shared`, while `shared(S)` means the struct instance is `shared`, and it may be referenced either through a `shared` pointer (`shared(S*)`) or an unshared pointer (`shared(S)*`).

So the `==` in the `assert` about a class is in fact the exception to the rule, as technically `TailShared!C` _should_ unshare the reference, but it doesn't, because it can't.

aG0aep6G: > Is there a reason you are using the : operator instead of == inside the is expression?

`TailShared!S3` is not `shared S3` but a wrapper that converts implicitly. And that's what `:` means. The point of the wrapper is that the (property-wrapped) fields get tail-shared types. You get `i` as an `int` instead of a `shared(int)`, and `p` as a `shared(int)*` instead of a `shared(int*)`

> So the == in the assert about a class is in fact the exception to the rule, as technically TailShared!C should unshare the reference, but it doesn't, because it can't.

It's true that `TailShared!C` would strip shared from the reference if it could, but we'd test that with `==`.

`:` is the exception. It's only needed when `TailShared!S` resolves to the wrapper. And the wrapper is only needed for a certain kind of structs, not all of them.

thaven: > `:` is the exception. It's only needed when `TailShared!S` resolves to the wrapper. And the wrapper is only needed for a certain kind of structs, not all of them.

By '`==` is the exception' I just meant that `is(TailShared!T == shared T)` if and only if T is a reference type (class or interface), while `is(TailShared!T : shared T)` for any type T.

changelog/std-traits-tailshared.dd
```diff
@@ -0,0 +1,11 @@
+`TailShared` was added to `std.traits`
```
ntrel: I think `std.typecons` is more apt.

LightBender:
@jmdavis This one deals with std.traits. It may be of interest to you.

PR #6189, opened 2018-02-16
typo. concret; -> concrete

btbytes:


wilzbach:
@quickfur @JackStouffer thanks for the quick action. Remember to approve new contributors at the auto-tester though - otherwise it won't be tested and thus no auto-merge too (I just did so for @btbytes)

PR #6190, opened 2018-02-16
Fix Issue 10828 - datetime toString functions should accept sink

JackStouffer:
Finishes out all of the remaining functions.

### DIFF
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -7791,45 +7791,67 @@ public:
         enough to uniquely identify the time zone.
 
         Time zone offsets will be in the form +HH:MM or -HH:MM.
+
+        Params:
+            writer = A `char` accepting
+            $(REF_ALTTEXT output range, isOutputRange, std, range, primitives)
+        Returns:
+            A `string` when not using an output range; `void` otherwise.
       +/
     string toISOExtString() @safe const nothrow
     {
+        import std.array : appender;
+        auto app = appender!string();
+        app.reserve(35);
         try
-        {
-            immutable adjustedTime = adjTime;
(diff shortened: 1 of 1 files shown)


std/datetime/systime.d
```diff
@@ -7928,45 +7950,67 @@ public:
...
+        app.reserve(35);
         try
-        {
-            immutable adjustedTime = adjTime;
-            long hnsecs = adjustedTime;
-
-            auto days = splitUnitsFromHNSecs!"days"(hnsecs) + 1;
+            toSimpleString(app);
+        catch (Exception e)
+            assert(0, "toSimpleString() threw.");
```
BBasile: This is not a good way to maintain the `nothrow`ability of the function. There is even no way to deal with the possible `AssertError` that can happen, assuming the rule of not catching Error is respected. This would be more reasonable not to catch and to remove nothrow.

jmdavis: This is perfectly fine so long as `toSimpleString` will never throw unless there's a bug in the code.  `assert(0)` is the standard way to indicate that a piece of code should be unreachable and that it's a serious bug if it is.

std.datetime does this in quite a few places where it's guaranteed that there will be no exception unless there's a bug in the code - e.g. when calling `format`, since `format` throws when it's given a bad format string, but the format string in question is always the same, and what's being fed to `format` is well-known. So, as long as it's guaranteed that `toSimpleString` is never going to throw as long as the code isn't buggy, then this is perfectly fine.

JinShil:
According to digger, this PR caused a regression: https://issues.dlang.org/show_bug.cgi?id=19123

PR #6191, opened 2018-02-16
SList implementation improvements

thaven:
- Optimize implementation of opBinary

- ~Allow concatenation of lists with different element types, if both
  element types have a common base type.~ (this change has been removed from the PR)

- Use separate common logic from insertFront and insertAfter, and use it in both of them.

- Explicitly forbid shared element types (did not work anyway, and would be bad usage of `shared` as well).

### DIFF
--- a/std/container/slist.d
+++ b/std/container/slist.d
@@ -929,3 +929,11 @@ Complexity: $(BIGOH n)
     s.reverse();
     assert(s[].equal([3, 2, 1]));
 }
+
+@safe unittest
+{
+    auto s = SList!int([4, 6, 8, 12, 16]);
+    auto d = s.dup;
+    assert(d !is s);
+    assert(d == s);
+}
--- a/std/container/slist.d
+++ b/std/container/slist.d
@@ -290,24 +290,28 @@ define $(D opBinary).
     SList opBinary(string op, Stuff)(Stuff rhs)
     if (op == "~" && is(typeof(SList(rhs))))
     {
-        auto toAdd = SList(rhs);
-        if (empty) return toAdd;
-        // TODO: optimize
-        auto result = dup;
-        auto n = findLastNode(result._first);
-        n._next = toAdd._first;
-        return result;
+        import std.range : chain, only;
+
+        static if (isInputRange!Stuff)
+            alias r = rhs;
+        else
+            auto r = only(rhs);
+
+        return SList(this[].chain(r));
     }
 
     /// ditto
     SList opBinaryRight(string op, Stuff)(Stuff lhs)
     if (op == "~" && !is(typeof(lhs.opBinary!"~"(this))) && is(typeof(SList(lhs))))
     {
-        auto toAdd = SList(lhs);
-        if (empty) return toAdd;
-        auto result = dup;
-        result.insertFront(toAdd[]);
-        return result;
+        import std.range : chain, only;
+
+        static if (isInputRange!Stuff)
+            alias r = lhs;
+        else
+            auto r = only(lhs);
+
+        return SList(r.chain(this[]));
     }
 
 /**
--- a/std/container/slist.d
+++ b/std/container/slist.d
@@ -135,6 +135,48 @@ struct SList(T)
         return n;
     }
 
+    private static auto createNodeChain(Stuff)(Stuff stuff)
+    if (isImplicitlyConvertible!(Stuff, T))
+    {
+        import std.range : only;
+        return createNodeChain(only(stuff));
+    }
+
+    private static auto createNodeChain(Stuff)(Stuff stuff)
+    if (isInputRange!Stuff && isImplicitlyConvertible!(ElementType!Stuff, T))
+    {
+        static struct Chain
+        {
+            Node* first;
+            Node* last;
+            size_t length;
+        }
+
+        Chain ch;
+
+        foreach (item; stuff)
+        {
+            auto newNode = new Node(null, item);
+            (ch.first ? ch.last._next : ch.first) = newNode;
+            ch.last = newNode;
+            ++ch.length;
+        }
+
+        return ch;
+    }
+
+    private static size_t insertAfterNode(Stuff)(Node* n, Stuff stuff)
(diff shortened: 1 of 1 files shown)


JackStouffer:
Oof. It would have been better to split these changes into different PRs. Can you at least separate these changes into different commits?

thaven:
Okay, split it up, and reduced the diff a bit by leaving out some code styling changes. Both changes for `opBinary` and `opBinaryRight` are not easily separated, so I left them in a single commit.

The `opBinary` optimization comes down to this change:
- originally the list was copied, then walked to find the last node, and then `rhs` was being inserted at the end.
- I changed that to: first chain ranges `this[]` and `rhs` (no list walking required for that), then use that combined range as the source for the new list.

Also removed the one-line change in the unittest for concatenation, as it turns out that didn't prove anything. Actually, the relaxed typing for concatenation would need a more complex test case. I can add that later, no time for that now.

thaven:
While working on the unittest I discovered that the relaxed typing for opBinary might have unwanted side effects, especially with stucts that have an `alias this`, so I ended up removing this change. I'll push the new commits in a few minutes.

JackStouffer:
Hmm, later I'm going to have to add to the docs that none of this works with transient ranges.

std/container/slist.d
```diff
@@ -135,6 +135,48 @@ struct SList(T)
...
+        return ch;
+    }
+
+    private static size_t insertAfterNode(Stuff)(Node* n, Stuff stuff)
+    {
+        auto ch = createNodeChain(stuff);
+
+        if (!ch.length) return 0;
+
+        ch.last._next = n._next;
```
JackStouffer: Doesn't this line not do anything useful, as `ch` goes out of scope at the end of the function?

std/container/slist.d
```diff
@@ -359,39 +401,15 @@ Returns: The number of elements inserted
 Complexity: $(BIGOH m), where $(D m) is the length of $(D stuff)
      */
     size_t insertFront(Stuff)(Stuff stuff)
-    if (isInputRange!Stuff && isImplicitlyConvertible!(ElementType!Stuff, T))
```
JackStouffer: Probably should combine the two constraints with an `||` rather than getting rid of them

std/container/slist.d
thaven: `ch` itself is going out of scope indeed, but this line is useful when inserting the new chain of nodes anywhere but the end of the list. Without this line we'd lose the tail of the list.

std/container/slist.d
thaven: Would certainly be correct. I chose to do it like it already was on `insertAfter`. Willing to change.

std/container/slist.d
```diff
@@ -58,7 +58,7 @@ struct SList(T)
     import std.exception : enforce;
     import std.range : Take;
     import std.range.primitives : isInputRange, isForwardRange, ElementType;
-    import std.traits : isImplicitlyConvertible;
+    import std.traits : CommonType, isImplicitlyConvertible;
```
wilzbach: `CommonType` isn't used here.

std/container/slist.d
wilzbach: Hmm, IIRC there were some issues with `isImplicitlyConvertible` and @jmdavis has been working on slowly fixing Phobos for this.
The problem essentially boils down to static arrays been implicitly convertible to strings:

```
static assert( isImplicitlyConvertible!(immutable(char)[4], string));
```

which means that the string slice could point to memory that is no longer on the heap...

std/container/slist.d
```diff
@@ -135,6 +135,48 @@ struct SList(T)
...
+
+        Chain ch;
+
+        foreach (item; stuff)
+        {
+            auto newNode = new Node(null, item);
+            (ch.first ? ch.last._next : ch.first) = newNode;
+            ch.last = newNode;
+            ++ch.length;
+        }
```
wilzbach: > Also optimize

Technically this is not an optimization because now a bit more work needs to be done in the special case of just inserting one node. I doubt that it matters though (and yes saving the allocation on the root node is a win).

std/container/slist.d
```diff
@@ -135,6 +135,48 @@ struct SList(T)
...
+
+        Chain ch;
+
+        foreach (item; stuff)
+        {
+            auto newNode = new Node(null, item);
+            (ch.first ? ch.last._next : ch.first) = newNode;
+            ch.last = newNode;
+            ++ch.length;
```
wilzbach: AFAICT you don't need that. Just check whether `first !is null`? (probably a good idea to add an `empty` method for this to Chain to encapsulate things)

std/container/slist.d
thaven: Oops, that's a leftover...

std/container/slist.d
thaven: the `length` field is there just so `insertFront` and `insertAfter` can return the number of elements inserted, as they are supposed to do.

std/container/slist.d
wilzbach: > I chose to do it like it already was on insertAfter. Willing to change.

I think it would be better to keep the constraints and use `isInputRange!Stuff || isImplicitlyConvertible!(Stuff, T)` as then the user will see an error message in his code and not in Phobos code.
Probably a good idea to do the same for `insertAfter` too.

thaven:
Ping @ZombineDev

thaven:
Test failed after squashing fixup commits. Failure does not seem to be related to this PR.

wilzbach:
Yep, the Jenkins failure is unrelated -> https://github.com/dlang/ci/issues/190

wilzbach:
@thaven thanks again for you PR and sorry that it got a bit lost in the queue. We're trying to give each PR enough attention, but this one clearly could have received more.

Anyhow, a ping like you just did now, often helps.
You can also reach us on #phobos at dlang.slack.com (it currently requires an invitation, but feel free to ping me with your email address if you want to join.)

PR #6192, opened 2018-02-16
Add Layout abstraction

andralex:
This is much more useful and usable than `Fields` or `RepresentationTypeTuple`, which lose the essential information of offsets in data fields.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -2681,9 +2681,6 @@ types. The presence, number, and types of additional fields are
 implementation-dependent and should not be relied upon in portable code. All these
 additional types have names that start with a double underscore.)
 
-$(LI If `T` is a statically-sized array, the layout is an iteration of the
-layouts of each field of the array. Note that this may create large tuples.)
-
 $(LI For all other types, the layout is $(D AliasSeq!(size_t(0), T)).)
 
 )
@@ -2699,16 +2696,7 @@ template Layout(T)
                                      Add!(offset, U[2 .. $]));
     }
 
-    static if (is(T == U[n], U, size_t n))
-    {
-        static if (n == 0)
-            alias Layout = AliasSeq!();
-        else static if (is(U == class) || is(U == interface))
-            alias Layout = AliasSeq!(size_t(0), U, Add!(U.sizeof, Layout!(U[n - 1])));
-        else
-            alias Layout = AliasSeq!(Layout!U, Add!(U.sizeof, Layout!(U[n - 1])));
-    }
-    else static if (!is(T == struct) && !is(T == union)
+    static if (!is(T == struct) && !is(T == union)
                     && !is(T == class) && !is(T == interface))
     {
         alias Layout = AliasSeq!(size_t(0), T);
--- a/std/traits.d
+++ b/std/traits.d
@@ -50,6 +50,7 @@
  *           $(LREF InterfacesTuple)
  *           $(LREF isInnerClass)
  *           $(LREF isNested)
+ *           $(LREF Layout)
  *           $(LREF MemberFunctionsTuple)
  *           $(LREF RepresentationTypeTuple)
  *           $(LREF TemplateArgsOf)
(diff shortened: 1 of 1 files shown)


andralex:
Eh, we should relax the style checker a bit:

```
std/traits.d(2803:19)[error]: Expected `{` instead of `;`
std/traits.d(2803:19)[warn]: Empty declaration
std/traits.d(2804:24)[error]: Expected `{` instead of `;`
std/traits.d(2804:24)[warn]: Empty declaration
std/traits.d(8357:1)[error]: Expected `}` instead of `EOF`
std/traits.d(8357:1)[error]: Expected `}` instead of `EOF
```

in reference to:

```
// Intentionally undefined and undocumented.
struct __Vtable(C);
struct __Bookkeeping(C);
```

cc @wilzbach

std/traits.d
```diff
@@ -2658,6 +2659,150 @@ template RepresentationTypeTuple(T)
...
+additional opaque data managed by the compiler. Usually these are placed at the
+front of the object. This data is represented in the layout as pointers to opaque
+types. The presence, number, and types of additional fields are
+implementation-dependent and should not be relied upon in portable code. All these
+additional types have names that start with a double underscore.)
+
+$(LI If `T` is a statically-sized array, the layout is an iteration of the
+layouts of each field of the array. Note that this may create large tuples.)
+
+$(LI For all other types, the layout is $(D AliasSeq!(size_t(0), T)).)
```
WalterBright: The documentation enumerates cases 1,2,3,4. However, the code is laid out as cases 3,4,(1,2). The code can be laid out in the same order as the documentation. Code comments should be added to say which case # they are dealing with.

std/traits.d
```diff
@@ -2658,6 +2659,150 @@ template RepresentationTypeTuple(T)
...
+    {
+        static if (n == 0)
+            alias Layout = AliasSeq!();
+        else static if (is(U == class) || is(U == interface))
+            alias Layout = AliasSeq!(size_t(0), U, Add!(U.sizeof, Layout!(U[n - 1])));
+        else
+            alias Layout = AliasSeq!(Layout!U, Add!(U.sizeof, Layout!(U[n - 1])));
+    }
+    else static if (!is(T == struct) && !is(T == union)
+                    && !is(T == class) && !is(T == interface))
```
WalterBright: Instead of all the nots, make the body of this the `else` clause which is case 4 "for all other types".

std/traits.d
```diff
@@ -2658,6 +2659,150 @@ template RepresentationTypeTuple(T)
...
+    {
+        alias Layout = AliasSeq!(size_t(0), T);
+    }
+    else
```
WalterBright: Make this case `else static if (is(T == struct) || ... )` to match the documentation.

std/traits.d
```diff
@@ -2658,6 +2659,150 @@ template RepresentationTypeTuple(T)
     static assert(R2.length == 2 && is(R2[0] == int) && is(R2[1] == immutable(Object)));
 }
 
+// Layout
+/**
+Returns the layout in memory of type `T` as an `AliasSeq`. For each field in
```
WalterBright: Should be `Params:` section and `Returns:` sections.

std/traits.d
```diff
@@ -2658,6 +2659,150 @@ template RepresentationTypeTuple(T)
...
+$(LI If `T` is a `class` or `interface` type, the layout is similar to the one
+return for `struct`s, with one difference: classes and interfaces have
+additional opaque data managed by the compiler. Usually these are placed at the
+front of the object. This data is represented in the layout as pointers to opaque
+types. The presence, number, and types of additional fields are
+implementation-dependent and should not be relied upon in portable code. All these
+additional types have names that start with a double underscore.)
+
+$(LI If `T` is a statically-sized array, the layout is an iteration of the
+layouts of each field of the array. Note that this may create large tuples.)
```
WalterBright: Does this mean that if there's a T[1000] array in there, there will be 2000 tuple elements? Not sure this is the best idea.

UplinkCoder:
This implementation will have rather bad compile-time performance.
Please count how many instances will be created for a `Layout` with 4,8,32 fields.

UplinkCoder:
sadly it does not seem to work
@andralex please try:
```
{
    struct S
    {
        uint u;
        char[64] c64;
        long l1, l2, l3, l4, l5, l6;
        union {
            S* s;
            S*[32] s32;
        }
    }

    pragma(msg, __traits(allMembers, Layout!S));
}
```

std/traits.d
```diff
@@ -2658,6 +2659,150 @@ template RepresentationTypeTuple(T)
...
+
+    class C1 { char[] a; union { S1 b; S1 * c; } }
+    alias L3 = Layout!C1;
+    static assert(L3.length == 12);
+    static assert(L3[0] == 0 && is(L3[1] == __Vtable!C1*));
+    static assert(L3[2] == size_t.sizeof && is(L3[3] == __Bookkeeping!C1*));
+    static assert(L3[4] == 2 * size_t.sizeof && is(L3[5] == char[]));
+    static assert(L3[6] == 4 * size_t.sizeof && is(L3[7] == int));
+    static assert(L3[8] == 4 * size_t.sizeof && is(L3[9] == S1*));
+    static assert(L3[10] == 4 + 4 * size_t.sizeof && is(L3[11] == float));
```
JohanEngelen: Ideas for tests missing:
- `align(1) struct S { byte a; int b; byte c; int d;}`
- `align(2) struct S { byte a; int b; byte c; int d;}`
- `struct S { byte a; align(1) int b; byte c; int d;}`
- Test `Layout!(const S2)`. Specify that Layout discards all qualifiers on the types, also of struct members. Test with `struct S1 { int a; const S2 b}` (note the inner const)
- Test with abstract class, class inheritance, interfaces.
- extern(C++) types
- `Layout!int` and `Layout!(int[])` and `Layout!(int[5])` (test errors or sane behavior)
- Sorting for union of these structs: `align(1) struct S1 { byte a; int b; char c; char d; }; align(1) struct S2 { short a; short b; long c; byte d; }; align(1) struct S3 { int a; double b; };`   (intentionally many different types, but avoiding pointers because `size_t` is variable).
- Perhaps important to take into account when testing that `size_t` can be 16bits.

JohanEngelen:
Ideas for tests missing:
- `align(1) struct S { byte a; int b; byte c; int d;}`
- `align(2) struct S { byte a; int b; byte c; int d;}`
- `struct S { byte a; align(1) int b; byte c; int d;}`
- Test `Layout!(const S2)`. Specify that Layout discards all qualifiers on the types, also of struct members. Test with `struct S1 { int a; const S2 b}` (note the inner const)
- Test with abstract class, class inheritance, interfaces.
- extern(C++) types
- `Layout!int` and `Layout!(int[])` and `Layout!(int[5])` (test errors or sane behavior)
- Sorting for union of these structs: `align(1) struct S1 { byte a; int b; char c; char d; }; align(1) struct S2 { short a; short b; long c; byte d; }; align(1) struct S3 { int a; double b; };`   (intentionally many different types, but avoiding pointers because `size_t` is variable).
- Perhaps important to take into account when testing that `size_t` can be 16bits.

std/traits.d
andralex: I respectfully disagree here. When getting into code, I found this pattern:

```d
if (expression)
{
    ... one-liner handling the simple case ...
}
else
{
    ... 70 lines handling the complex case ...
}
```

considerably easier to get into than the converse:

```d
if (expression)
{
    ... 70 lines handling the complex case ...
}
else
{
    ... one-liner handling the simple case ...
}
```

This is because whilst reading the 70 lines the other branch, also potentially as long, is still looming. If the simple case is eliminated early, I can understand it easily and then mentally dispense with it.

Also: 

* I find recursions easier to read if the limit condition is placed first
* I find loops with filtering "of all elements just process those fulfilling some condition..." easier to read if they eliminate the mismatches with `if (mismatch) continue;`

I do understand others believe differently! Live and let live is my principle.

std/traits.d
andralex: Yeah, this is tricky. The nice thing about expanding arrays is that it reveals these structures have identical layout:

```d
struct S1 { int r, g, b; }
struct S2 { int[3] colors; }
```

If arrays are not expanded, that complicates the structural equivalence algorithm. Nothing terrible, just inconvenient. But I think you're right the liability is too high. I'll make the change.

andralex:
This is great feedback you guys, thanks.

andralex:
@JohanEngelen I figure it's a mistake to eliminate qualifiers (with `Unqual`). Due to `immutable`, immutable and mutable structures do not have compatible layout - e.g. you can't seat one on top of the other. Will change it such that qualifiers are preserved.

Also: currently the layout is sorted by offset first, order of declaration second. But the order of declaration is irrelevant. Better find a canonical order - will sort by offset first, type second. Are there better ways to sort types than by `.stringof`?

andralex:
All: how do I compare that two layouts are identical? `is(Layout!A == Layout!B)` does not work because the instantiation of `Layout` is not a type. What is the right way to compare that two instantiations of a `template` are identical? Thx!

John-Colvin:
@andralex template instantiation is pure, so you can compare the template args for equality instead (recursively if necessary, as a template args could be a template). Use `std.traits.TemplateArgsOf` or some similar method.

I prefer to just use a `struct` instead of a `template` to avoid this problem altogether.

UplinkCoder:
> template instantiation is pure [...]
@John-Colvin only if you are playing nice :p

UplinkCoder:
@andralex I think we have a `isSame` template.
which should do some sort of compare for you.

John-Colvin:
@UplinkCoder I assume you mean https://dlang.org/spec/traits.html#isSame

UplinkCoder:
@andralex I'd like to see a potential usecase for this.

JackStouffer:
@andralex `static if` doesn't work?

std/traits.d
```diff
@@ -2658,6 +2659,142 @@ template RepresentationTypeTuple(T)
...
+                    else
+                        alias With =
+                            AliasSeq!(U[0], U[1], Merge!T.With!(U[2 .. $]));
+                }
+            }
+            static if (T.length <= 2)
+                alias Sort = T;
+            else
+                alias Sort = Merge!(Sort!(T[0 .. T.length / 2]))
+                    .With!(Sort!(T[T.length / 2 .. $]));
```
aliak00: If T.length = 6 (e.g. in `Layout!{int a, b, c;}` then is the merge doing `Merge!(0UL, int, 4UL).With!(int, 8UL, int)`? Should maybe check if T.length / 2 == odd or even and deal appropriately?

std/traits.d
```diff
@@ -2658,6 +2659,142 @@ template RepresentationTypeTuple(T)
...
+`T`, the result includes the offset of the field followed by the type of the
+field. Detailed information depends on the type's kind, as follows:
+
+$(UL
+$(LI For built-in types including primitive types, static and dynamic arrays,
+and pointers, the layout is $(D AliasSeq!(size_t(0), T)).)
+
+$(LI If `T` is a `struct` or `union` type, then `Layout!T` is an `AliasSeq`
+containing each field preceded by its offset. Fields that are in turn of
+`struct` or `union` types are decomposed transitively. The result is ordered
```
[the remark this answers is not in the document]
aliak00: Ah, mentioned below :)

std/traits.d
```diff
@@ -2658,6 +2659,142 @@ template RepresentationTypeTuple(T)
...
+in increasing offset order. (Two or more fields may have the same offset due to
+`union`s.) For a given offset, types are ordered lexicographically by their
+name.)
+
+$(LI If `T` is a `class` or `interface` type, the layout is similar to the one
+return for `struct`s, with one difference: classes and interfaces have
```
aliak00: *return -> returned

std/traits.d
```diff
@@ -2714,7 +2713,8 @@ template Layout(T)
                         alias With = U;
                     else static if (U.length == 0)
                         alias With = T;
-                    else static if (T[0] <= U[0])
+                    else static if (T[0] < U[0]
```
UplinkCoder: This may fail.

std/traits.d
```diff
@@ -2714,7 +2713,8 @@ template Layout(T)
...
+                            || T[0] == U[0] && T[1].stringof <= U[1].stringof)
```
UplinkCoder: I'd highly discourage anyone from using stringof for anything but debugging.
It is certainly not for use in the stdlib.

wilzbach:
> Eh, we should relax the style checker a bit:

https://github.com/dlang/phobos/pull/6197

Biotronic:
Why implement your own compile-time mergesort - what's wrong with `std.meta.staticSort`?
Also, is there a great benefit to having the result be one long `AliasSeq` instead of having a struct or template instance with `Type` and `offset` fields?
For comparison, here's how I would have implemented this (including names, since that's been requested):
https://gist.github.com/Biotronic/6182f24eabb444bca90f29d7a82bd428

jercaianu:
This will be a powerful tool for the allocators, allowing us to create at compile time a smart mapping between types and allocators. 
For each group of objects with identical layouts, we can use the same allocator, allowing for safe memory reuse.

std/traits.d
```diff
@@ -2658,6 +2659,157 @@ template RepresentationTypeTuple(T)
...
+    static assert(L3[4] == 2 * size_t.sizeof && is(L3[5] == char[]));
+    static assert(L3[6] == 4 * size_t.sizeof && is(L3[7] == S1*));
+    static assert(L3[8] == 4 * size_t.sizeof && is(L3[9] == int));
+    static assert(L3[10] == 4 + 4 * size_t.sizeof && is(L3[11] == float));
+}
+
+// Intentionally undefined and undocumented.
+struct __ClassMeta(C, uint index);
+
+unittest
```
JackStouffer: annotate with `@safe`

std/traits.d
```diff
@@ -2658,6 +2659,157 @@ template RepresentationTypeTuple(T)
...
+    {
+        uint u;
+        char[64] c64;
+        long l1, l2, l3, l4, l5, l6;
+        union {
+            S* s;
+            S*[32] s32;
+        }
+    }
+    static assert(Layout!S.length == 20);
```
stefan-koch-sociomantic: To me this looks like 10 members at most.

LightBender:
@jmdavis If you think there is anything here, you're welcome to use it, but you don't have to.

PR #6193, opened 2018-02-17
Add needed changes to get all tests passing with Musl

joakim-noah:
Only one assert trips when running the Phobos tests after this patch, [this one in `std.datetime.timezone`](https://github.com/dlang/phobos/blob/master/std/datetime/timezone.d#L781), but I don't know if that's because of the bare minimum install of Alpine I'm testing in or if it needs to be addressed. @yshui, what do you say?

### DIFF
--- a/std/experimental/allocator/building_blocks/region.d
+++ b/std/experimental/allocator/building_blocks/region.d
@@ -647,6 +647,12 @@ struct InSituRegion(size_t size, size_t minAlign = platformAlignment)
     assert((() nothrow @nogc => r2.deallocateAll())());
 }
 
+version(CRuntime_Musl)
+{
+    // sbrk and brk are disabled in Musl:
+    // https://git.musl-libc.org/cgit/musl/commit/?id=7a995fe706e519a4f55399776ef0df9596101f93
+    // https://git.musl-libc.org/cgit/musl/commit/?id=863d628d93ea341b6a32661a1654320ce69f6a07
+} else:
 private extern(C) void* sbrk(long) nothrow @nogc;
 private extern(C) int brk(shared void*) nothrow @nogc;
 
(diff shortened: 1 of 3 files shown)


std/socket.d
```diff
@@ -163,56 +163,14 @@ string formatSocketError(int err) @trusted
...
-        else version (Solaris)
-        {
-            auto errs = strerror_r(err, buf.ptr, buf.length);
-            if (errs == 0)
-                cs = buf.ptr;
-            else
-                return "Socket error " ~ to!string(err);
-        }
-        else version (CRuntime_Bionic)
+        else
```
joakim-noah: I see no reason to copy and paste this block for every non-Glibc Posix system.

std/stdio.d
```diff
@@ -44,38 +44,38 @@ version (CRuntime_Glibc)
     version = GCC_IO;
     version = HAS_GETDELIM;
 }
-
-version (OSX)
+else version (CRuntime_Bionic)
```
joakim-noah: Added `else`'s as made sense, and switched Bionic to use `getdelim`, as it was added with Android API 18.

std/experimental/allocator/building_blocks/region.d
```diff
@@ -647,6 +647,12 @@ struct InSituRegion(size_t size, size_t minAlign = platformAlignment)
     assert((() nothrow @nogc => r2.deallocateAll())());
 }
 
+version(CRuntime_Musl)
+{
+    // sbrk and brk are disabled in Musl:
+    // https://git.musl-libc.org/cgit/musl/commit/?id=7a995fe706e519a4f55399776ef0df9596101f93
+    // https://git.musl-libc.org/cgit/musl/commit/?id=863d628d93ea341b6a32661a1654320ce69f6a07
```
ZombineDev: Interesting, thanks for the research @joakim-noah

PR #6194, opened 2018-02-18
Update how AffixAllocator infers shared from parent

edi33416:
`AffixAllocator` used to only check if `Allocator.instance` is `shared`. If the parent `Allocator` doesn't have an `instance` field, the check will be false, even though the `Allocator` is shared.

### DIFF
--- a/std/experimental/allocator/building_blocks/affix_allocator.d
+++ b/std/experimental/allocator/building_blocks/affix_allocator.d
@@ -349,7 +349,13 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
     }
     else static if (is(typeof(Allocator.instance) == shared))
     {
+        static assert(stateSize!Allocator == 0);
         static shared AffixAllocator instance;
+        shared { mixin Impl!(); }
+    }
+    else static if (is(Allocator == shared))
+    {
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -499,3 +504,16 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
     assert(b.length == 100);
     assert((() nothrow @nogc => a.deallocate(b))());
 }
+
+@system shared unittest
+{
+    import std.experimental.allocator : processAllocator, RCISharedAllocator;
+
+    alias SharedAllocT = shared AffixAllocator!(RCISharedAllocator, int);
+    assert(is(RCISharedAllocator == shared));
```
ZombineDev: You should also add an assert which verifies that `SharedAllocT`'s methods are actually inferred as `shared`.

ZombineDev: I'm not sure how annotating the unittest with `shared` helps.

andralex:
@ZombineDev can you please add the assert?

edi33416:
I updated the unittest and added `static assert`s;

PR #6195, opened 2018-02-18
posix.mak - module specific -dip25/1000 for someModule.test and unittest rules

carblue:
@wilzbach
Discussion about this topic is also in
https://github.com/dlang/phobos/pull/6041
https://github.com/dlang/phobos/pull/6181

aa represents the current state of "phobos -dip1000 compilability".

I preferred associative array aa entries in the "modulename-notation" (e.g. aa[std.array]) instead of OS-dependant "pathname-notation" (didn't recall whether windows can cope with forward-slashes).
But this only defers the issue to the call-site (I use Linux, i.e. /): $(aa[$(subst /,.,$(basename $<))])

Used currently in 2 rules, e.g.
%.test : %.d $(LIB)
	    $(DMD) -od$$T $(DFLAGS) $(aa[$(subst /,.,$(basename $<))]) -main $(UDFLAGS) $(LIB) -defaultlib= -debuglib= $(LINKDL) -cov -run $< ;     \
	    RET=$$? ; rm -rf $$T ; exit $$RET                                                                   \
	  )

Positioning $(aa[$(subst /,.,$(basename $<))])  after $(DFLAGS) resulting in e.g. -dip25 -dip1000 always worked for me, the compiler picked up the last one.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -389,7 +389,7 @@ aa[std.system]=-dip1000
 aa[std.traits]=-dip1000
 aa[std.typecons]=-dip25 # cannot call @system function std.format.formattedWrite!(Appender!string, char, Nullable!int)
 aa[std.typetuple]=-dip1000
-aa[std.uni]=-dip1000
+aa[std.uni]=-dip25 # depends on PR 6041 merged
 aa[std.uri]=-dip1000
 aa[std.utf]=-dip1000
 aa[std.uuid]=-dip1000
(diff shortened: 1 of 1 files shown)


wilzbach:
Looks good and I'm aware that https://github.com/dlang/phobos/pull/6041 is blocking this.
(Though of course I hope that we can switch to `-dip1000` by default soon)

CC @CyberShadow @MartinNowak - are you okay with this approach too?

carblue:
This lacks reply from CyberShadow or MartinNowak for more than 2 weeks, thus closing this PR

PR #6199, opened 2018-02-19
Fix broken links in the std.parallelism.fold changelog entry

wilzbach:


acehreli:
Thanks Seb but the fix is wrong because the text should reference std.algorithm.fold, not std.parallelism.reduce. Thanks...

MartinNowak:
Doesn't show up in the changelog as that is generated on stable.

PR #6200, opened 2018-02-19
Fix Issue 18341 - Documentation for std.array.split is confusing/incorrect

wilzbach:
I used the opportunity to reorder the examples of `splitter` too and as they are public examples I converted them to DStyle (UFCS, ~~~no string lambdas~~~).

edit: seems like `splitter` has frame access problems with normal lambdas, so I reverted the string lambda -> lambda for now and opened [18470](https://issues.dlang.org/show_bug.cgi?id=18470) for now.

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -3660,31 +3660,31 @@ The number of seeds must be correspondingly increased.
 
 // splitter
 /**
-Lazily splits a range using an element as a separator. This can be used with
-any narrow string type or sliceable range type, but is most popular with string
-types.
+Lazily splits a range using an element or range as a separator.
+Separator ranges can be any narrow string type or sliceable range type.
 
 Two adjacent separators are considered to surround an empty element in
 the split range. Use `filter!(a => !a.empty)` on the result to compress
 empty elements.
 
-The predicate is passed to $(REF binaryFun, std,functional), and can either accept
-a string, or any callable that can be executed via `pred(element, s)`.
+The predicate is passed to $(REF binaryFun, std,functional) and accepts
+any callable function that can be executed via `pred(element, s)`.
 
-If the empty range is given, the result is an empty range. If a range with
-one separator is given, the result is a range with two empty elements.
+Notes:
+    If splitting a string on whitespace and token compression is desired,
+    consider using `splitter` without specifying a separator.
 
-If splitting a string on whitespace and token compression is desired,
-consider using `splitter` without specifying a separator (see fourth overload
-below).
+    If no separator is passed, the $(REF_ALTTEXT, unary, unaryFun, std,functional)
+    predicate `isTerminator` decides whether to accept an element of `r`.
 
 Params:
     pred = The predicate for comparing each element with the separator,
         defaulting to `"a == b"`.
     r = The $(REF_ALTTEXT input range, isInputRange, std,range,primitives) to be
-        split. Must support slicing and `.length`.
-    s = The element to be treated as the separator between range segments to be
-        split.
+        split. Must support slicing and `.length` or be a narrow string type.
+    s = The element (or range) to be treated as the separator
+        between range segments to be split.
+    isTerminator = The predicate for deciding where to split the range when no separator is passed
 
 Constraints:
     The predicate `pred` needs to accept an element of `r` and the
@@ -3695,10 +3695,15 @@ Returns:
     is a $(REF_ALTTEXT forward range, isForwardRange, std,range,primitives)
     or $(REF_ALTTEXT bidirectional range, isBidirectionalRange, std,range,primitives),
     the returned range will be likewise.
+    When a range is used a separator, bidirectionality isn't possible.
+
+    If an empty range is given, the result is an empty range. If a range with
+    one separator is given, the result is a range with two empty elements.
 
 See_Also:
  $(REF _splitter, std,regex) for a version that splits using a regular
-expression defined separator.
+expression defined separator and
+ $(REF _splitter, std,array) for a version that splits eagerly.
 */
 auto splitter(alias pred = "a == b", Range, Separator)(Range r, Separator s)
 if (is(typeof(binaryFun!pred(r.front, s)) : bool)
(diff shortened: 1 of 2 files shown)


bachmeil:
It looks like this fixes everything in my bug report - it's clear and correct AFAICT.

std/algorithm/iteration.d
```diff
@@ -3855,46 +3855,104 @@ if (is(typeof(binaryFun!pred(r.front, s)) : bool)
...
+    assert(a.splitter([0, 0]).equal([ [], [1] ]));
+}
+
+/// Custom predicate functions.
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+    import std.ascii : toLower;
+
+    //assert(equal(splitter!((a,b) => a.toLower == b)("abXcdxef", 'x'),
```
JackStouffer: did you forget to uncomment this?

std/algorithm/iteration.d
```diff
@@ -3660,31 +3660,40 @@ The number of seeds must be correspondingly increased.
...
     r = The $(REF_ALTTEXT input range, isInputRange, std,range,primitives) to be
         split. Must support slicing and `.length`.
-    s = The element to be treated as the separator between range segments to be
-        split.
+    s = The element (or element range) to be treated as the separator
+        between range segments to be split.
+    isTerminator = The predicate for deciding where to split the range when no separator is passed
 
 Constraints:
     The predicate `pred` needs to accept an element of `r` and the
```
wilzbach: So I tried to group the [endless `splitter` iteration](https://dlang.org/phobos/std_algorithm_iteration.html#.splitter) into one function, but if you think that busts the scope of this PR, I can also move the DDoc unification to another PR.

std/algorithm/iteration.d
```diff
@@ -3660,31 +3660,40 @@ The number of seeds must be correspondingly increased.
...
-If the empty range is given, the result is an empty range. If a range with
-one separator is given, the result is a range with two empty elements.
+Notes:
+    If splitting a string on whitespace and token compression is desired,
+    consider using `splitter` without specifying a separator.
 
-If splitting a string on whitespace and token compression is desired,
-consider using `splitter` without specifying a separator (see fourth overload
-below).
+    If an element range is used as a separator, it needs to be at least
```
JackStouffer: I'd just say "If a range"

std/algorithm/iteration.d
```diff
@@ -3660,31 +3660,40 @@ The number of seeds must be correspondingly increased.
...
+    a $(REF_ALTTEXT forward range, isForwardRange, std,range,primitives).
+    Moreover, when a subrange is used a separator, bidirectionality isn't possible.
```
JackStouffer: I'd move this sentence to the `Returns:` section

std/algorithm/iteration.d
```diff
@@ -3660,31 +3660,40 @@ The number of seeds must be correspondingly increased.
...
+
+    If no separator is passed, the $(REF_ALTTEXT, unary, unaryFun, std,functional)
+    predicate `isTerminator` decides whether to accept an element of `r`.
+
+Special_cases:
+    If the empty range is given, the result is an empty range. If a range with
```
JackStouffer: Same here, `Returns:`

std/algorithm/iteration.d
```diff
@@ -3660,31 +3660,40 @@ The number of seeds must be correspondingly increased.
...
+    one separator is given, the result is a range with two empty elements.
 
 Params:
     pred = The predicate for comparing each element with the separator,
         defaulting to `"a == b"`.
     r = The $(REF_ALTTEXT input range, isInputRange, std,range,primitives) to be
         split. Must support slicing and `.length`.
-    s = The element to be treated as the separator between range segments to be
-        split.
+    s = The element (or element range) to be treated as the separator
```
JackStouffer: "or range"

std/algorithm/iteration.d
```diff
@@ -3855,46 +3864,127 @@ if (is(typeof(binaryFun!pred(r.front, s)) : bool)
...
+
+    assert("a|b||c".splitter('|').equal([ "a", "b", "", "c" ]));
+    assert("hello  world".splitter(' ').equal([ "hello", "", "world" ]));
+
+    auto a = [ 1, 2, 0, 0, 3, 0, 4, 5, 0 ];
+    auto w = [ [1, 2], [], [3], [4, 5], [] ];
+    assert(a.splitter(0).equal(w));
+}
+
+/// Use a sub-range for splitting
```
JackStouffer: "Use a range for splitting"

std/algorithm/iteration.d
```diff
@@ -3855,46 +3864,127 @@ if (is(typeof(binaryFun!pred(r.front, s)) : bool)
...
 
-    // Predicate functions.
-    assert(equal(splitter!"a.toLower == b"("abXcdxef", 'x'),
+/// Custom predicate functions.
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+    import std.ascii : toLower;
+
+    assert("abXcdxef".splitter!"a.toLower == b"('x').equal(
```
JackStouffer: If you're changing the code, it'd be a good time to update this to a lambda. Same throughout

std/algorithm/iteration.d
```diff
@@ -3855,46 +3864,127 @@ if (is(typeof(binaryFun!pred(r.front, s)) : bool)
...
+
     w = [ [0], [1], [2] ];
-    assert(equal(splitter!"a.front == b"(w, 1), [ [[0]], [[2]] ]));
+    assert(equal(splitter!(a => a.front == 1)(w), [ [[0]], [[2]] ]));
+}
 
-    // Bidirectional ranges.
-    assert(equal(splitter("a|bc|def", '|').retro, [ "def", "bc", "a" ]));
+/// Empty and separator-only ranges.
+@safe unittest
```
JackStouffer: I'd move this up in-front of the range separator examples.

std/array.d
```diff
@@ -1443,20 +1443,30 @@ if (isInputRange!S && !isDynamicArray!S)
...
-$(D @safe), $(D pure) and $(D CTFE)-able.
+When no delimiter is provided, strings are split into an array of words,
+using whitespace as delimiter.
+Runs of whitespace are merged together (no empty words are produced).
+
+The `range` must be a $(REF_ALTTEXT forward _range, isForwardRange, std,_range,primitives).
+The separator can be a value of the same type as the elements in `range`
+or it can be another forward `range`.
+
+Note:
```
JackStouffer: this is already covered in the see also section

std/array.d
```diff
@@ -1443,20 +1443,30 @@ if (isInputRange!S && !isDynamicArray!S)
...
+    `split` allocates memory. The same effect can be achieved lazily
+    using $(REF splitter, std,algorithm,iteration).
 
 Params:
-    s = the string to split
+    s = the string to split by word
```
JackStouffer: It's only split by word when `sep` is not given

std/algorithm/iteration.d
wilzbach: Agreed, but it's not trivial (frame access issues). Hence, I moved this into a separate issue https://issues.dlang.org/show_bug.cgi?id=18470

std/array.d
wilzbach: This was explicitly mentioned in this bug report, but fair enough ...

PR #6201, opened 2018-02-20
Added writer version of toString to FormatSpec

JackStouffer:
Also adds a test.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -1596,22 +1596,60 @@ if (is(Unqual!Char == Char))
         return w.data;
     }
 
-    string toString()
+    /**
+     * Gives a string containing all of the member variables on their own
+     * line.
+     *
+     * Params:
+     *     writer = A `char` accepting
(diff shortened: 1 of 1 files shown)


std/format.d
```diff
@@ -1596,22 +1596,60 @@ if (is(Unqual!Char == Char))
...
-                "\nflHash = ", flHash,
-                "\nflSeparator = ", flSeparator,
-                "\nnested = ", nested,
-                "\ntrailing = ", trailing, "\n");
+        import std.array : appender;
+        auto app = appender!string();
+        app.reserve(200 + trailing.length);
+        toString(app);
+        return app.data;
+    }
```
wilzbach: I wonder whether we can add something like this to the compiler or druntime, e.g. if `toString` overload found, but `string toString` isn't defined, generate this boilerplate.

I'm aware that reserving the memory initially is a bit faster, but for user-code it usually doesn't matter.

PR #6205, opened 2018-02-21
Make AffixAllocator take into account RCISharedAllocator

edi33416:


std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -499,3 +510,12 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
     assert(b.length == 100);
     assert((() nothrow @nogc => a.deallocate(b))());
 }
+
+@system unittest
+{
+    import std.experimental.allocator : RCISharedAllocator, processAllocator;
```
wilzbach: `processAllocator` isn't used here ;-)

std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -499,3 +510,12 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
...
+
+    AffixAllocator!(RCISharedAllocator, uint) a;
+    auto buf = a.allocate(42);
+    assert(buf.length == 42);
```
wilzbach: Any chance we can replace the `processAllocator` with something like this?

```d
auto prevAllocator = processAllocator;
scope(exit) processAllocator = prevAllocator;

alias SCAlloc = StatsCollector!(Mallocator, Options.bytesUsed);
SCAlloc statsCollectorAlloc;
processAllocator = allocatorObject(statsCollectorAlloc);
...
assert(statsCollectorAlloc.bytesUsed == 100);
```

std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -69,11 +70,21 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
     static if (stateSize!Allocator)
     {
         Allocator _parent;
-        static if (is(Allocator == RCIAllocator))
+        static if (is(Allocator == RCIAllocator) || is(Allocator == RCISharedAllocator))
```
wilzbach: This behavior isn't mentioned in the docs above...

std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -69,11 +70,21 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
     static if (stateSize!Allocator)
     {
         Allocator _parent;
```
edi33416: This should probably be made private

edi33416: I think this, `Allocator _parent`, should be made private as we want the user to use the `parent()` method.

edi33416:
@jercaianu

edi33416:
I've moved the `parent()` logic inside the `mixin template Impl` as this also needs to infer the `shared` attribute.

std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -60,36 +63,39 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
...
+        static if (stateSize!Allocator)
         {
-            Allocator parent()
+            Allocator _parent;
+            static if (is(Allocator == RCIAllocator) || is(Allocator == RCISharedAllocator))
             {
-                if (_parent.isNull) _parent = theAllocator;
-                assert(alignment <= _parent.alignment);
-                return _parent;
+                Allocator parent()
```
jercaianu: Does this need to return by value? 
Shouldn't it return a reference to the parent?

std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -60,36 +63,39 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
...
+                {
+                    if (_parent.isNull)
+                    {
+                        static if (is(Allocator == RCIAllocator))
+                        {
+                            _parent = theAllocator;
+                        }
+                        else
+                        {
+                            _parent = processAllocator;
```
jercaianu: Since parent can be called by multiple threads, is this assignment guaranteed to be thread safe?

std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -520,3 +540,30 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
     static assert(is(typeof(a.allocate) == shared));
     assert(buf.length == 10);
 }
+
+@system unittest
```
jercaianu: I think it would be a good idea to also have a test in which we access this from multiple threads.

edi33416:
@n8sh could you have another look at the code? Changed sync to use only one variable

std/experimental/allocator/building_blocks/affix_allocator.d
```diff
@@ -60,61 +63,103 @@ struct AffixAllocator(Allocator, Prefix, Suffix = void)
...
+                    {
+                        parentLock(_lockVar);
+                        if (_parent.isNull)
+                        {
+                            auto fn = (() @trusted =>
+                                    cast(RCISharedAllocator function() @nogc nothrow pure @safe)
+                                    (&wrapProcAllocatorObject))();
+                            _parent = fn();
+                        }
+                        parentUnlock(_lockVar);
```
n8sh: To avoid deadlock instead put `scope(exit) parentUnlock(_lockVar);` right after the lock is acquired.

RazvanN7:
@edi33416 What is the state of this PR? Should we pull it in?

LightBender:
@thewilsonator if can rebase this and get it in we'll take it.
@edi33416 if you want to resurrect this we'll take it in.

PR #6206, opened 2018-02-21
SharedBitmappedBlock

jercaianu:
Added the shared implementation of the BitmappedBlock allocator.
Allocations are served by finding the first unset bit in the bitmap and atomically setting the bit.
Deallocations work in a similar manner.

I benchmarked the allocator by requesting 10k allocations (of 64 bytes) and deallocations, all repeated 10k times.
On single threaded benchmarks, I got the following results on my machine:
`SharedBitmappedBlock: 6s`
`BitmappedBlock: 9s`
`Mallocator: 11s`

Maybe we can improve the BitmappedBlock implementation.


jercaianu:
ping @andralex @edi33416

wilzbach:
> On single threaded benchmarks, I got the following results on my machine:

I know Phobos doesn't have a `benchmark` folder yet, but we should really start having one.
Could you maybe make the first step and add your script to it. We can always worry later about the details, but collecting benchmarks scripts from old PRs is a rather ungrateful and time-consuming task.

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -1746,3 +1760,350 @@ private struct BitVector
...
+            0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2,
+            2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,
+            2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3,
+            4, 4, 4, 4, 4, 4, 4, 4, 5, 5, 5, 5, 6, 6, 7
+        ];
+
+        for (size_t i = 0; i < _control.length; i++)
```
wilzbach: `foreach (i; 0 .. control.length)

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -1746,3 +1760,350 @@ private struct BitVector
...
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2,
+            2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,
+            2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3,
+            4, 4, 4, 4, 4, 4, 4, 4, 5, 5, 5, 5, 6, 6, 7
+        ];
```
wilzbach: How about generating this in CTFE?

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -1746,3 +1760,350 @@ private struct BitVector
...
+        ulong controlVal, newControlVal;
+        auto blockIndex = (b.ptr - _payload.ptr) / blockSize;
+        auto controlIndex = blockIndex / 64;
+        auto bitIndex = blockIndex % 64;
+
+        do
+        {
+            controlVal = atomicLoad(_control[controlIndex]);
+            newControlVal = controlVal & ~(1 << (63 - bitIndex));
+        } while (!cas(&_control[controlIndex], controlVal, newControlVal));
```
wilzbach: I can't stress it too often, but DRY!!
Ideally this should work:

```d
atomicOp!"&="(controlVal, ~(1 << (63 - bitIndex)));
```

It probably just requires a minor tweak in core.atomic.

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -1746,3 +1760,350 @@ private struct BitVector
...
+        enum numThreads = 100;
+        void[][numThreads] buf;
+        size_t count = 0;
+
+        void fun()
+        {
+            void[] b = a.allocate(63);
+            assert(b.length == 63);
+
+            lock.lock();
```
wilzbach: `scope(exit) lock.unlock` ?

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -1746,3 +1760,350 @@ private struct BitVector
...
+            assert(buf[i].ptr + a.goodAllocSize(buf[i].length) <= buf[i + 1].ptr);
+        }
+
+        foreach (i; 0 .. numThreads)
+        {
+            assert(a.deallocate(buf[i]));
+        }
+    }
+
+     auto fb = SharedBitmappedBlock!(64, platformAlignment, Mallocator)(1024 * 1024);
```
wilzbach: It's typically called `alloc` or similar. Why `fb`?

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -1746,3 +1760,350 @@ private struct BitVector
...
+    bool expand(ref void[] b, immutable size_t delta)
+    {
+        if (delta == 0)
+            return true;
+
+        immutable newLength = delta + b.length;
+        if (b is null || newLength > blockSize)
+            return false;
+
+        b = b.ptr[0 .. newLength];
```
wilzbach: Why `b.ptr`? `b = b[0 .. newLength]` should work too.
(Besides `.ptr` is unsafe...)

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -1746,3 +1760,350 @@ private struct BitVector
...
+                        break;
+                    }
+                    bitIndex += 8;
+                }
+            } while (!cas(&_control[i], controlVal, newControlVal));
+
+            auto blockIndex = bitIndex + 64 * i;
+            if (controlVal != ulong.max && blockIndex < _blocks)
+            {
+                size_t payloadBlockStart = cast(size_t) blockIndex * blockSize;
```
wilzbach: This could overflow. How about using `std.experimental.checkedint`?

std/experimental/allocator/building_blocks/bitmapped_block.d
jercaianu: Yes, I dont need the cas loop here, thanks!

std/experimental/allocator/building_blocks/bitmapped_block.d
jercaianu: If i declare it static, won't it be inlined in code?

std/experimental/allocator/building_blocks/bitmapped_block.d
jercaianu: if the newLength is greater than the previous length, it doesn't work without the ptr

std/experimental/allocator/building_blocks/bitmapped_block.d
jercaianu: An extra check here could be expensive.
Also the blockIndex check above itself prevents against overflow

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -1746,3 +1760,350 @@ private struct BitVector
...
+    private size_t totalAllocation(size_t capacity)
+    {
+        auto blocks = capacity.divideRoundUp(blockSize);
+        auto leadingUlongs = blocks.divideRoundUp(64);
+        import std.algorithm.comparison : min;
+        immutable initialAlignment = min(parentAlignment,
+            1U << trailingZeros(leadingUlongs * 8));
+        auto maxSlack = alignment <= initialAlignment
+            ? 0
+            : alignment - initialAlignment;
```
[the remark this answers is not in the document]
jercaianu: This is somehow covered by the asserts in the constructor which uses this and inherently byt all tests which use this method of creating a bitmappedblock

jercaianu:
@wilzbach 
Thanks for the comments!
Do you have any idea why there is the FreeBSD 32 failure?

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -1746,3 +1760,350 @@ private struct BitVector
...
+    /// Ditto
+    static if (chooseAtRuntime == theBlockSize)
+    this(ubyte[] data, uint blockSize)
+    {
+        this._blockSize = blockSize;
+        this(data);
+    }
+
+    /// Ditto
+    static if (!is(ParentAllocator == NullAllocator))
```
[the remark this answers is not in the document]
edi33416: If I got the question right, using `stateSize` would fail for stateless allocators.

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -1746,3 +1760,344 @@ private struct BitVector
...
+    @trusted void[] allocate(const size_t s)
+    {
+        import core.atomic : cas, atomicLoad, atomicOp;
+
+        if (s.divideRoundUp(blockSize) != 1)
+            return null;
+
+        // First zero bit position for all values in the 0 - 255 range
+        // for fast lookup
+        static ubyte[] firstZero = [
```
andralex: Make this `static immutable ubyte[256] firstZero = [ ... ];` so it goes into the rodata section. Also initializing with ctfe is nice and won't harm.

andralex: To initialize with ctfe:

```d
static ubyte[256] initFirstZero()
{
    ... do it ...
}
static immutable ubyte[256] firstZero = initFirstZero();
```

JackStouffer:
`git fetch origin && git rebase origin/master`

jercaianu:
@JackStouffer 
I'm so sorry, I messed up

PR #6207, opened 2018-02-21
Fix Issue 7879 - format of const class with non const toString()

JackStouffer:
So trying to make `Object.toString` const was a nightmare, so I added in this work-around to at least allow the code to compile.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -3868,6 +3868,26 @@ version(unittest)
     assert(s == "shared(std.format.C)");
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=7879
+@safe unittest
+{
+    class F
+    {
+        override string toString() const @safe
+        {
+            return "Foo";
+        }
+    }
+
+    const(F) c;
+    auto s = format("%s", c);
+    assert(s == "null");
+
+    const(F) c2 = new F();
+    s = format("%s", c2);
+    assert(s == "Foo", s);
+}
+
 // ditto
 private void formatValueImpl(Writer, T, Char)(auto ref Writer w, T val, const ref FormatSpec!Char f)
 if (is(T == interface) && (hasToString!(T, Char) || !is(BuiltinTypeOf!T)) && !is(T == enum))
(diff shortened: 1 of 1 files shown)


JackStouffer:
Changed code to match std.format's existing behavior for class formatting

wilzbach:
> So trying to make Object.toString const was a nightmare

Yeah, though we should really move things out of `object.d`.
A quick selection:

- https://issues.dlang.org/show_bug.cgi?id=9769
- https://issues.dlang.org/show_bug.cgi?id=9770
- https://issues.dlang.org/show_bug.cgi?id=9771
- https://issues.dlang.org/show_bug.cgi?id=9772

std/format.d
```diff
@@ -3830,6 +3841,53 @@ if (is(T == class) && !is(T == enum))
...
+    s = format("%s", c4);
+    assert(s == "shared(std.format.C)");
+}
+
+// https://issues.dlang.org/show_bug.cgi?id=7879
+@safe unittest
+{
+    class F
+    {
+        override string toString() const @safe
```
wilzbach: BTW we can't use `inout` here: https://issues.dlang.org/show_bug.cgi?id=18500

std/format.d
```diff
@@ -3738,7 +3736,20 @@ if (is(T == class) && !is(T == enum))
         put(w, "null");
     else
     {
-        static if (hasToString!(T, Char) > 1 || (!isInputRange!T && !is(BuiltinTypeOf!T)))
+        static if ((is(T == immutable) || is(T == const) || is(T == shared)) && hasToString!(T, Char) == 0)
```
wilzbach: I would prefer if we could use `== none` here.
See: https://github.com/dlang/phobos/pull/6209

std/format.d
```diff
@@ -3738,7 +3736,20 @@ if (is(T == class) && !is(T == enum))
...
+                put(w, "immutable(");
+            else static if (is(T == const))
+                put(w, "const(");
+            else static if (is(T == shared))
+                put(w, "shared(");
+
+            put(w, typeid(Unqual!T).name);
+            put(w, ')');
+        }
+        else static if (hasToString!(T, Char) > 1 || !isInputRange!T && !is(BuiltinTypeOf!T))
```
wilzbach: With https://github.com/dlang/phobos/pull/6209: `!overload.among(none, hasSomeToString)`

JackStouffer:
Moving ahead with this to shorten the queue.

JackStouffer:
@wilzbach Can I get a manual merge?

PR #6208, opened 2018-02-22
Fix Issue #18491 pass default values as parameter not template parameters in Logger

burner:
https://issues.dlang.org/show_bug.cgi?id=18491

### DIFF
--- a/std/experimental/logger/core.d
+++ b/std/experimental/logger/core.d
@@ -194,33 +194,17 @@ Example:
 log(LogLevel.warning, true, "Hello World", 3.1415);
 --------------------
 */
-void log(int line = __LINE__, string file = __FILE__,
-    string funcName = __FUNCTION__, string prettyFuncName = __PRETTY_FUNCTION__,
-    string moduleName = __MODULE__, A...)(const LogLevel ll,
-    lazy bool condition, lazy A args)
-if (args.length != 1)
+void log(string moduleName = __MODULE__, A...)(const LogLevel ll,
+    lazy bool condition, lazy A args, int line = __LINE__,
+    string file = __FILE__, string funcName = __FUNCTION__,
+    string prettyFuncName = __PRETTY_FUNCTION__)
 {
     static if (isLoggingActive)
     {
         if (ll >= moduleLogLevel!moduleName)
         {
-               stdThreadLocalLog.log!(line, file, funcName, prettyFuncName, moduleName)
-                (ll, condition, args);
-        }
-    }
-}
-
-/// Ditto
-void log(T, string moduleName = __MODULE__)(const LogLevel ll,
-    lazy bool condition, lazy T arg, int line = __LINE__, string file = __FILE__,
-    string funcName = __FUNCTION__, string prettyFuncName = __PRETTY_FUNCTION__)
-{
-    static if (isLoggingActive)
-    {
-        if (ll >= moduleLogLevel!moduleName)
-        {
-            stdThreadLocalLog.log!(T,moduleName)(ll, condition, arg, line, file, funcName,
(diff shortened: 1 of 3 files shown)


burner:
The style checker fails to consider default arguments in ddoc parameter counting. I rechecked by hand and all passable parameters have ddoc comments.

wilzbach:
> The style checker fails to consider default arguments in ddoc parameter counting. I rechecked by hand and all passable parameters have ddoc comments.

It's not even the style checker per se, but Ddoc.

```
Check that Ddoc runs without errors
../dmd/generated/linux/debug/64/dmd -conf= -I../druntime/import  -w -de -dip25 -m64 -fPIC -transition=complex -g -debug -defaultlib= -debuglib= generated/linux/debug/64/libphobos2.a -w -D -Df/dev/null -main -c -o- $(find etc std -type f -name '*.d') 2>&1
```

I think Ddoc expects documentation for parameters like `int line = __LINE__`.
See https://run.dlang.io/is/cW31iD

burner:
Not good. I have not the time nor the knowledge to fix ddoc to get this PR passing.

JackStouffer:
@burner All you have to do is add all of the parameters to the `Params` section.

n8sh:
Jenkins has a build failure for [libasync](https://github.com/etcimon/libasync) because libasync has its own logging functions that are just `nothrow` wrappers for std.experimental.logging and call `logImpl` with explicit `__LINE__`, etc. template parameters.

burner:
@n8sh thanks

burner:
@JackStouffer @wilzbach ping

std/experimental/logger/core.d
```diff
@@ -188,39 +188,27 @@ Params:
...
-        }
-    }
-}
-
-/// Ditto
-void log(T, string moduleName = __MODULE__)(const LogLevel ll,
-    lazy bool condition, lazy T arg, int line = __LINE__, string file = __FILE__,
-    string funcName = __FUNCTION__, string prettyFuncName = __PRETTY_FUNCTION__)
+void log(string moduleName = __MODULE__, A...)(const LogLevel ll,
+    lazy bool condition, lazy A args, int line = __LINE__,
```
JackStouffer: I'm surprised this is legal in D, as it's impossible to actually set `line` manually.

n8sh: It's only been allowed since 2.079 and it seems to have been added to the language [specifically for this function](https://dlang.org/changelog/2.079.0.html#default_after_variadic).

n8sh:
What's the way forward on this? To move the current `std.experimental.logger` to `stdx.experimental.logger` (like with allocators) and submit a PR to libasync to use that?

ZombineDev:
@n8sh sound like a plan.

burner:
@n8sh @ZombineDev I do not follow. Please elaborate.

n8sh:
@burner Example of what I was talking about: https://github.com/atilaneves/automem/pull/20
PRs like that were made for several/many? projects that had been using `std.experimental.allocator`.

burner:
Ok, so we are moving stuff to code.dlang.org and give a version number. So people who don't want to rewrite and link against that!? And after that we change upstream?

ZombineDev:
@burner yes, this is so they can upgrade to a newer version of stdx and the compiler independently. Very often people use `dmd`, `dmd-beta` and even `dmd-nightly` for CI testing, which makes breaking changes (for which a deprecation period can't be reasonably enforced) especially painful.

LightBender:
PR closed as stalled. If you wish to resurrect it you are welcome to do so. We can revisit this in Phobos 3.

PR #6209, opened 2018-02-23
Use enums for hasToString

wilzbach:
Because dealing with raw integers is really error-prone and makes looking at PR diffs really difficult.

BTW @JackStouffer was there any reason not to allow `(P, S)(ref put, S stringLike)` for the new `.put`-writer overloads. The sink overloads accept a format string or a format spec.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -3658,12 +3658,23 @@ if (is(AssocArrayTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     formatTest(e2, "[A, B, C]");
 }
 
+private enum HasToStringResult
+{
+    none,
+    hasSomeToString,
+    constCharSink,
+    constCharSinkFormatString,
+    constCharSinkFormatSpec,
+    customPutWriter,
+    customPutWriterFormatSpec,
+}
+
 private template hasToString(T, Char)
 {
     static if (isPointer!T && !isAggregateType!T)
     {
         // X* does not have toString, even if X is aggregate type has toString.
-        enum hasToString = 0;
+        enum hasToString = HasToStringResult.none;
     }
     else static if (is(typeof(
         {T val = void;
@@ -3676,7 +3687,7 @@ private template hasToString(T, Char)
         static assert(!__traits(compiles, val.toString(S(), f)));}
     )))
     {
-        enum hasToString = 6;
+        enum hasToString = HasToStringResult.customPutWriterFormatSpec;
     }
     else static if (is(typeof(
         {T val = void;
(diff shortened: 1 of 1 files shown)


JackStouffer:
> was there any reason not to allow (P, S)(ref put, S stringLike) for the new .put-writer overloads. The sink overloads accept a format string or a format spec.

It was never used in Phobos (outside of one use in bigint which immediately converted it to a single spec) so it seemed like an unnecessary complication.

JackStouffer:
Sorry, hit the wrong button

std/format.d
```diff
@@ -3542,12 +3542,23 @@ if (is(AssocArrayTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     formatTest(e2, "[A, B, C]");
 }
 
+enum HasToStringResult
```
JackStouffer: private

std/format.d
```diff
@@ -3631,30 +3645,30 @@ if (hasToString!(T, Char))
...
-    else static if (overload == 5)
+    else static if (overload == HasToStringResult.customPutWriter)
     {
         val.toString(w);
     }
     // not using the overload enum to not break badly defined toString overloads
     // e.g. defining the FormatSpec as ref and not const ref led this function
     // to ignore that toString overload
-    else static if (is(typeof(val.toString((const(char)[] s){}, f))))
+    else static if (overload == HasToStringResult.constCharSinkFormatSpec)
```
JackStouffer: This is a breaking change, see comment above

std/format.d
```diff
@@ -3692,30 +3706,27 @@ if (hasToString!(T, Char))
...
         val.toString(w, f);
     }
-    else static if (overload == 5)
+    else static if (overload == HasToStringResult.customPutWriter)
     {
         val.toString(w);
     }
-    // not using the overload enum to not break badly defined toString overloads
-    // e.g. defining the FormatSpec as ref and not const ref led this function
-    // to ignore that toString overload
```
wilzbach: @JackStouffer the `constCharSinkFormatSpec` overload is:

```
T val = void; val.toString((const(char)[] s){}, "%s");
```

JackStouffer: that's the `constCharSinkFormatString` overload

std/format.d
wilzbach: Why would this be a breaking change?

wilzbach:
CC @thewilsonator. I just revived this from the dead.

thewilsonator:
Auto-merge toggled on

PR #3774, opened 2015-10-26
fixup for string-like overloads

MartinNowak:
- undocument overloads for aliased/static strings
- make std.traits helpers phobos internal
- rename isStringLike to isConvertibleToString for clearer documentation
- rename peelStringLike to convertToString
- fix a few staticMap!(selectString) uses in std.path

PR #3770, opened 2015-10-24
fix Issue 15027 – rangified functions no longer work with string-like types

MartinNowak:
- [Issue 15027 – rangified functions no longer work with alias this'ed strings (e.g. DirEntry)](https://issues.dlang.org/show_bug.cgi?id=15027)

PR #6213, opened 2018-02-23
Revert "Issue 18244: std.math generic functions need to have sig constraints"

MetaLang:
Reverts dlang/phobos#6040

This seemingly breaks a fair amount of user code, so we should revert it until we can figure out a less intrusive fix.

https://issues.dlang.org/show_bug.cgi?id=18287
https://issues.dlang.org/show_bug.cgi?id=18473

Not sure if this is jumping the gun, but better safe than sorry.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -133,7 +133,6 @@ version (Win64)
 static import core.math;
 static import core.stdc.math;
 static import core.stdc.fenv;
-import std.range.primitives : isInputRange, ElementType;
 import std.traits; // CommonType, isFloatingPoint, isIntegral, isSigned, isUnsigned, Largest, Unqual
 
 version(LDC)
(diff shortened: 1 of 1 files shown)


ibuclaw:
Will there be any follow-up action taken?

MetaLang:
Not sure. @quickfur submitted the fix for 18244 but I don't know if he plans to try a different fix that doesn't break code.

quickfur:
I'm too busy to work on an alternative fix right now, but the plan is to fix this one way or another somehow.  Having unconstrained templates in Phobos is generally not a good idea.

MartinNowak:
Similar to what we had to do for string constraints.
[15027 – rangified functions no longer work with alias this'ed strings (e.g. DirEntry)](https://issues.dlang.org/show_bug.cgi?id=15027)
https://github.com/dlang/phobos/pull/3770
https://github.com/dlang/phobos/pull/3774

PR #6214, opened 2018-02-23
Overload for static arrays for std.array.array

dukc:
With static arrays, `auto` keyword has been hard to use so far.  This changes today!

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -382,39 +382,44 @@ if (isNarrowString!String)
 
 /**
  * Returns a static array from the stack with length of `s` and initializes
- * it with copies of the elements of range $(D r). The remainder of the
- * array is default-initialized.
+ * it with copies of first `s` elements of range $(D r). If $(D r) does not
+ * have `s` elements the rest of the array is default-initialized.
  *
  * Recommended particulary for complex initializations of static arrays
  * during compilation.
  *
  * Params:
  *      r = The range (or aggregate with $(D opApply) function) whose elements
- *      are copied into the stack-allocated array. Must be as long or shorter
- *      than `s`, otherwise causes an array overflow.
+ *      are copied into the stack-allocated array. It may be longer or shorter
+ *      than `s`.
+ *      rangeLength = outputs the amount of elements copied. Same as
+ *      $(D r).$(REF walkLength, std,range,primitives).
  * Returns:
  *      An initialized static array
  */
-ForeachType!Range[s] array(size_t s, Range)(Range r)
+ForeachType!Range[s] array(size_t s, Range)(Range r, out size_t rangeLength)
 if (isInputRange!Range && !isInfinite!Range)
 {
     alias E = ForeachType!Range;
 
     import std.conv : emplaceRef;
-    import std.range : padRight;
+    import std.algorithm.mutation : uninitializedFill;
+    import std.algorithm.comparison : min;
+    import std.range : take;
+
+    size_t i = 0;
 
     if (__ctfe)
     {
         // Compile-time version to avoid memcpy calls.
         Unqual!E[s] result;
-        size_t i;
-        foreach (e; r)
+        foreach (ref e; r.take(s))
         {
-
             result[i] = e;
             i++;
         }
 
+        rangeLength = i;
         return (() @trusted => cast(E[s]) result)();
     }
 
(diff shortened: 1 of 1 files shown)


dukc:
@andralex opinion needed because this is a new symbol overload.

std/array.d
```diff
@@ -380,6 +380,95 @@ if (isNarrowString!String)
...
+    import std.range : padRight;
+
+    if (__ctfe)
+    {
+        // Compile-time version to avoid memcpy calls.
+        Unqual!E[s] result;
+        size_t i;
+        foreach (e; r)
+        {
+
```
andralex: huh?

std/array.d
```diff
@@ -380,6 +380,95 @@ if (isNarrowString!String)
     )));
 }
 
+/**
+ * Returns a static array from the stack with length of s and initializes
+ * it with copies of the elements of range $(D r). The remainder of the
+ * array is default-initialized.
```
andralex: What happens if the range is longer than `s`?

std/array.d
```diff
@@ -380,6 +380,95 @@ if (isNarrowString!String)
...
+ *
```
andralex: The documentation should add that this is recommended particularly for nontrivial initialization of static arrays during compilation.

std/array.d
```diff
@@ -380,6 +380,95 @@ if (isNarrowString!String)
     )));
 }
 
+/**
+ * Returns a static array from the stack with length of s and initializes
```
andralex: s in backquotes

std/array.d
```diff
@@ -380,6 +380,95 @@ if (isNarrowString!String)
...
+ *
+ * Params:
+ *      r = The range (or aggregate with $(D opApply) function) whose elements
+ *      are copied into the stack-allocated array. Must be as long or shorter
+ *      than s.
+ * Returns:
+ *      An initialized static array
+ */
+ForeachType!Range[s] array(size_t s, Range)(Range r)
+if (isInputRange!Range && !isInfinite!Range)
```
andralex: You don't care whether the range is finite - you take a finite subset anyway.

std/array.d
dukc: A standard overflow. But i think you meant that I add that in comment.

std/array.d
dukc: That's if I change it to check that. Now it goes unchecked.

JackStouffer:
This somewhat overlaps with https://github.com/dlang/phobos/pull/6178

std/array.d
dukc: I copied that comment from regular `.array`. I think `emplaceRef` calls memCpy.

dukc:
Yea, I should probably look around before starting doing.

JackStouffer:
DIP1000 in Phobos has been "just around the corner" for over a year now. If the same safety issues are present in this implementation, I would close this.

dukc:
Well, we have Andrei around so I'll let him do that if he agrees with that.

dukc:
@andralex Review addressed and the overload you requested has been added. On the other hand, check last three comments above this before merging.

std/array.d
```diff
@@ -380,6 +380,113 @@ if (isNarrowString!String)
...
+ * it with copies of the elements of range $(D r). The remainder of the
+ * array is default-initialized.
+ *
+ * Recommended particulary for complex initializations of static arrays
+ * during compilation.
+ *
+ * Params:
+ *      r = The range (or aggregate with $(D opApply) function) whose elements
+ *      are copied into the stack-allocated array. Must be as long or shorter
+ *      than `s`, otherwise causes an array overflow.
```
andralex: Please no overflow. It's trivial and low cost to have a test here. In case of overflow, just fill the array and place the real length in `rangeLength`.

std/array.d
```diff
@@ -380,6 +380,113 @@ if (isNarrowString!String)
...
+ */
+ForeachType!Range[s] array(size_t s, Range)(Range r, out size_t rangeLength)
+if (isInputRange!Range && !isInfinite!Range)
+{
+    alias E = ForeachType!Range;
+
+    import std.conv : emplaceRef;
+    import std.algorithm.mutation : uninitializedFill;
+
+    rangeLength = 0;
```
andralex: `out` parameters are always zero-initialized

std/array.d
```diff
@@ -380,6 +380,113 @@ if (isNarrowString!String)
...
+    if (__ctfe)
+    {
+        // Compile-time version to avoid memcpy calls.
+        Unqual!E[s] result;
+        foreach (e; r)
+        {
+
```
andralex: Please no spurious empty lines. This particular one separates nothing in particular and adds negative readability value.

std/array.d
```diff
@@ -380,6 +380,113 @@ if (isNarrowString!String)
...
+    }
+
+    auto result = (() @trusted
+    {
+        Unqual!E[s] theArray = void;
+        return theArray;
+    }());
+
+
+    foreach (e; r)
```
andralex: This makes a copy for each element in the range.

std/array.d
```diff
@@ -380,6 +380,113 @@ if (isNarrowString!String)
...
+    {
+        emplaceRef!E(result[rangeLength], e);
+        rangeLength++;
```
andralex: Incrementing via pointer is slow, please use a local value and assign rangeLength at end

std/array.d
dukc: Take would be easy to add if the user wants it, but not to remove, so this makes it more-than-zero cost abstraction.

On the other hand, we can always add a non-checking equivalent later so I tend to agree with this. Regardless, I'll do it.

dukc:
@andralex All done now.

andralex:
I think we need to choose this, https://github.com/dlang/phobos/pull/6178, or a careful combination, but not both. Thoughts?

dukc:
It seems Cour already stated his opinion at #6178 so I decided it's a better place for discussion. See my answer there.

dukc:
Not needed anymore as #6178 appears to be getting in.

PR #6217, opened 2018-02-23
Disables Typedef.toString()

denizzzka:
It is need to avoid silently conversion of Typedef struct to string by std.conv.to and to prohibit string type payload leakage.

Discussion:
https://forum.dlang.org/post/kjzayldqnixehfprgslc@forum.dlang.org

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -6735,6 +6735,10 @@ struct Typedef(T, T init = T.init, string cookie=null)
             TD re() {return TD(Typedef_payload.re);}
             TD im() {return TD(Typedef_payload.im);}
         }
+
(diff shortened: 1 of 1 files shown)


MetaLang:
Sorry, I think I caused a misunderstanding. We do not want to add an `@disable`'d `toString` method for `Typedef`, but add a `toString` method that just delegates to the wrapped value.

Check out `std.typecons.Nullable`'s various overloads for inspiration:
https://github.com/dlang/phobos/blob/master/std/typecons.d#L2458

denizzzka:
Нou are right. Calling conv.to function is always a conscious action, and an unnoticed leak of a string will not be here.

PR #6218, opened 2018-02-23
Adds Typedef.toString()

denizzzka:
It is need to avoid silent conversion of Typedef struct itself to string by std.conv.to.

Discussion:
https://forum.dlang.org/post/kjzayldqnixehfprgslc@forum.dlang.org

Previous attempt: #6217

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -6736,6 +6736,35 @@ struct Typedef(T, T init = T.init, string cookie=null)
             TD im() {return TD(Typedef_payload.im);}
         }
     }
+
+    /**
+     * Convert wrapped value to a human readable string
+     */
+    string toString()
+    {
+        import std.array : appender;
+        auto app = appender!string();
+        auto spec = singleSpec("%s");
+        toString(app, spec);
+        return app.data;
+    }
+
+    /// ditto
+    void toString(W)(ref W writer, const ref FormatSpec!char fmt)
+    if (isOutputRange!(W, char))
+    {
+        formatValue(writer, Typedef_payload, fmt);
+    }
+
+    ///
+    @safe unittest
+    {
+        import std.conv : to;
+
+        int i = 123;
+        auto td = Typedef!int(i);
+        assert(i.to!string == td.to!string);
+    }
 }
 
 /**
@@ -6948,6 +6977,25 @@ template TypedefType(T)
     char[] s2 = cast(char[]) cs;
     const(char)[] cs2 = cast(const(char)[])s;
     assert(s2 == cs2);
+}
+
+@system unittest // toString
+{
+    import std.meta : AliasSeq;
+    import std.conv : to;
+
+    struct TestS {}
+    class TestC {}
+
+    static foreach (T; AliasSeq!(int, bool, float, double, real,
+                                 char, dchar, wchar,
+                                 TestS, TestC,
+                                 int*, int[], int[2], int[int]))
+    {{
+        T t;
+        Typedef!T td;
+        assert(t.to!string() == td.to!string());
+    }}
 }
 
 /**


denizzzka:
I am should use git squash or this is not necessary?

MetaLang:
Yes, please squash.

thewilsonator:
Yes do squash. Don't worry about the formatting, see https://github.com/dlang/phobos/pull/6220

wilzbach:
> Don't worry about the formatting, see #6220

FWIW: This is something different. `dmd-cxx` is the old C++ branch that didn't compile fro years and it slowly brought back to life.
Anyhow, while sticking to the [DStyle](https://dlang.org/dstyle.html) is a requirement for official code, it's not blocking your PR. Every maintainer can easily fix that for you if there's interest in the PR (i.e. the idea has been approved.)

----

Now back to the PR in question, is there any reason why you didn't add a writer-based `toString`?
`string toString` isn't really nice, because it always allocates which is slow.

@JackStouffer has been working hard on adding writer-based `toString` overloads, even for the mentioned `Nullable.toString`: https://github.com/dlang/phobos/pull/6198

MetaLang:
Sorry @denizzzka, I've been busy with other work the past few days and did not have time to properly review this. As @wilzbach pointed out, it would be preferably to add a writer-based overload of `toString` so there will be no allocation at all when possible. Take a look at the aforementioned pull request he linked to see how to do it.

denizzzka:
Changed for writer support and squashed

std/typecons.d
```diff
@@ -6736,6 +6736,31 @@ struct Typedef(T, T init = T.init, string cookie=null)
...
+     *     A `string` if `writer` and `fmt` are not set; `void` otherwise.
+     */
+    void toString(W)(ref W writer, const ref FormatSpec!char fmt)
+    if (isOutputRange!(W, char))
+    {
+        formatValue(writer, Typedef_payload, fmt);
+    }
+
+    /// ditto
+    string toString()
```
MetaLang: This overload should be documented and considered the "primary" one, with the other just marked as "ditto". Explain that this forwards to the wrapped value and the results between `<base type value>.to!string()` and `<Typedef'd type value>.to!string()` will be identical.

std/typecons.d
```diff
@@ -6736,6 +6736,31 @@ struct Typedef(T, T init = T.init, string cookie=null)
...
     }
+
+    /**
+     * Params:
+     *     writer = A `char` accepting
+     *     $(REF_ALTTEXT output range, isOutputRange, std, range, primitives)
+     *     fmt = A $(REF FormatSpec, std,format) which is used to represent
+     *     the value
+     * Returns:
+     *     A `string` if `writer` and `fmt` are not set; `void` otherwise.
```
MetaLang: This is incorrect, as each overload will show up separately in the documentation, and it will confuse readers if you say that this one could return `void` _or_ `string`.

std/typecons.d
```diff
@@ -6950,6 +6975,12 @@ template TypedefType(T)
     assert(s2 == cs2);
 }
 
+@safe unittest // toString
```
MetaLang: More unit tests are needed that test both overloads of `toString` with multiple different types of values.

Another good test to add would be something like the following to be reasonably sure that the results of `Typedef.toString` are the same as the base type:

```
import std.meta: AliasSeq;
import std.conv;

struct TestS {}
class TestC {}

static foreach (T; AliasSeq!(int, bool, float, double, real, ifloat, idouble, ireal,
                             cfloat, cdouble, creal, char, dchar, wchar, TestS, 
                             TestC, int*, int[], int[2], int[int]))
{{
    T t;
    Typedef!T td
    assert(t.to!string() == td.to!string());
}}
```

std/typecons.d
denizzzka: (This is copy and paste from Nullable, so its documentation is weak too)

std/typecons.d
denizzzka: I do not know how to describe this beautifully, just added an example.

denizzzka:
Something strange with CI check

wilzbach:
Not too strange though.
Just follow the advice of the message and add `@safe` to your unittest. Explicit annotation is required because it's too often forgotten.

denizzzka:
Oh, thank you, did not noticed by myself.

MetaLang:
Thanks @denizzzka!

PR #6219, opened 2018-02-25
std.complex: Minimally relax unittest for 64-bit reals

kinke:
I'm seeing a diff in the last mantissa bit for the imaginary parts with LDC on Win64 and its 64-bit reals.

### DIFF
--- a/std/complex.d
+++ b/std/complex.d
@@ -929,7 +929,15 @@ deprecated
     assert(coshisinh(3.0L) == complex(std.math.cosh(3.0L), std.math.sinh(3.0L)));
     auto z1 = coshisinh(1.234);
     auto z2 = std.math.coshisinh(1.234);
-    assert(z1.re == z2.re && z1.im == z2.im);
+    static if (real.mant_dig == 53)
+    {
+        assert(std.math.feqrel(z1.re, z2.re) >= real.mant_dig - 1 &&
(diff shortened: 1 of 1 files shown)


wilzbach:
> I'm seeing a diff in the last mantissa bit for the imaginary parts with LDC on Win64 and its 64-bit reals.

That's weird. AFAICT one of the reasons that Phobos has an implementation of all math algorithms from scratch is that they have reproducible results and don't rely on the varying implementation of the cruntime, s.t. these kind of patches aren't required ...

wilzbach:
FreeBSD 32-bit is broken at the moment - see braddr/d-tester#70
We need to move forward, so I'm going to merge this PR manually.

kinke:
> That's weird. AFAICT one of the reasons that Phobos has an implementation of all math algorithms from scratch is that they have reproducible results and don't rely on the varying implementation of the cruntime, s.t. these kind of patches aren't required ...

CTFE being another reason I imagine. I've similarly weakened a number of other unittests in the past, as they rely on `real` featuring a higher precision than `double` (most algorithms being implemented for real only), but that's only the case for x86(_64) and AArch64, all other architectures only have double-precision reals. [LDC additionally uses 64-bit reals for MSVC targets, as the MS runtime doesn't support x87 at all]. So DMD CI doesn't test 64-bit reals at all.

PR #6220, opened 2018-02-25
[dmd-cxx] Temporarily disable checkwhitespace to get auto-tester to build DMD's testsuite

wilzbach:
See also:

https://github.com/dlang/dmd/pull/7953#issuecomment-368275298

In short `dmd-cxx` is the C/C++ branch of dmd and interesting for bootstrapping on new OSes (though LDC's cross-compilation works too). This is a small step towards getting the testsuite on the auto-tester working again as running the dmd testsuite itself will hopefully reveal the real issue.


JackStouffer:
What's the status of this?

PR #6221, opened 2018-02-25
Add uClibc definitions to std.stdio

rracariu:


ZombineDev:
@wilzbach force merge pls

PR #6222, opened 2018-02-26
simplify changelog scripting example

timotheecour:


ZombineDev:
You need to target the stable branch where we renamed the module to `std.experimental.all`: https://github.com/dlang/phobos/blob/stable/changelog/std-experimental-all.dd

timotheecour:
done, PTAL

wilzbach:
Actually I used `reverseArgs` on purpose to show that `std.functional` is available too, but I don't have a strong opinion either way.

PR #6225, opened 2018-02-26
Fix awfully slow sicmp

DmitryOlshansky:
Turns out byDchar is ~2x slower then decodeFront,
time to either deprecate byDchar or make it fast.

<del>
2nd order of magnitude improvement is handling
ascii case w/o lookup tables.</del>

```
//1st module
import core.stdc.string;
import std.uni;
import std.stdio;
import bench_data;

extern(C) int strcasecmp(const char *s1, const char *s2);

void main() {
    auto timings = benchmark!(
        () => blackhole = sicmp(ascii1, ascii2),
        () => blackhole = sicmp(uni1, uni2),
        () => blackhole = strcasecmp(ascii1.ptr, ascii2.ptr)
    )(1_000_000);
    writefln("ASCII %s", timings[0]);
    writefln("Unicode %s", timings[1]);
    writefln("strcasecmp %s", timings[2]);
}

// 2nd module to hide data from optimizer
module bench_data;
string ascii1 = "This is some STring";
string ascii2 = "ThiS is some strIng";

string uni1 = "Некая СТроКа";
string uni2 = "некаЯ стРока";

int blackhole;
```

### DIFF
--- a/std/uni.d
+++ b/std/uni.d
@@ -7777,20 +7777,26 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
     && isInputRange!S2 && isSomeChar!(ElementEncodingType!S2))
 {
     import std.internal.unicode_tables : sTable = simpleCaseTable; // generated file
-    import std.utf : byDchar;
+    import std.utf : decodeFront;
+    import std.typecons : Yes;
+    static import std.ascii;
 
-    auto str1 = r1.byDchar;
-    auto str2 = r2.byDchar;
-
-    foreach (immutable lhs; str1)
+    while (!r1.empty)
     {
-        if (str2.empty)
+        immutable lhs = decodeFront!(Yes.useReplacementDchar)(r1);
+        if (r2.empty)
             return 1;
-        immutable rhs = str2.front;
-        str2.popFront();
+        immutable rhs = decodeFront!(Yes.useReplacementDchar)(r2);
         int diff = lhs - rhs;
         if (!diff)
             continue;
+        else if ((lhs | rhs) < 0x80)
+        {
+            auto lowL = std.ascii.toLower(lhs);
+            auto lowR = std.ascii.toLower(rhs);
+            if (!(lowL - lowR)) continue;
+            else return lowL - lowR;
+        }
         size_t idx = simpleCaseTrie[lhs];
         size_t idx2 = simpleCaseTrie[rhs];
         // simpleCaseTrie is packed index table
@@ -7816,7 +7822,7 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
         // one of chars is not cased at all
         return diff;
     }
-    return str2.empty ? 0 : -1;
+    return r2.empty ? 0 : -1;
 }
 
 ///
--- a/std/uni.d
+++ b/std/uni.d
@@ -7792,10 +7792,9 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
             continue;
         if ((lhs | rhs) < 0x80)
         {
-            auto lowL = std.ascii.toLower(lhs);
-            auto lowR = std.ascii.toLower(rhs);
-            if (!(lowL - lowR)) continue;
-            else return lowL - lowR;
+            immutable d = std.ascii.toLower(lhs) - std.ascii.toLower(rhs);
+            if (!d) continue;
+            return d;
         }
         size_t idx = simpleCaseTrie[lhs];
         size_t idx2 = simpleCaseTrie[rhs];
--- a/std/uni.d
+++ b/std/uni.d
@@ -7821,7 +7821,7 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
         // one of chars is not cased at all
         return diff;
     }
-    return -int(r2.empty);
+    return int(r2.empty) - 1;
 }
 
 ///
--- a/std/uni.d
+++ b/std/uni.d
@@ -7790,7 +7790,7 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
         int diff = lhs - rhs;
         if (!diff)
             continue;
-        else if ((lhs | rhs) < 0x80)
+        if ((lhs | rhs) < 0x80)
         {
             auto lowL = std.ascii.toLower(lhs);
             auto lowR = std.ascii.toLower(rhs);
@@ -7822,7 +7822,7 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
         // one of chars is not cased at all
         return diff;
     }
-    return r2.empty ? 0 : -1;
+    return -int(r2.empty);
 }
 
 ///


DmitryOlshansky:
I think @WalterBright was the original author of byDchar so maybe he has a clue of why the codegen is worse in that case (both are nothrow).

schveiguy:
for strings, `front` and `popFront` are both going to decode (where as `decodeFront` will only do it once). So I would expect at least 2x worse performance, but 10x seems much higher than that.

schveiguy:
I take it back, I didn't realize how byDchar (byUTF) works. It's only decoded once.

I can say that when I optimized iopipe's byLine, I found that avoiding using the struct members (i.e. only use local copies) during the function was a huge speedup. But I don't know if that matters here.

schveiguy:
Hm.. I wonder if in the dchar case, you can replace pos with an enum of 0, and it may make things much faster.

JackStouffer:
In this case, the speed up seems to come from avoiding the extra call to encode that's done in byUTF.

schveiguy:
Yep, there is an optimization from decoding from dchar, should be an equivalent optimization for encoding to dchar.

quickfur:
> I can say that when I optimized iopipe's byLine, I found that avoiding using the struct members (i.e. only use local copies) during the function was a huge speedup. But I don't know if that matters here.

That's rather disturbing.  Is this only a dmd problem, or does it also happen on ldc?

I've basically given up on dmd producing optimal code, so it doesn't even register on my radar these days when performance of generated code is concerned.

schveiguy:
> That's rather disturbing. Is this only a dmd problem, or does it also happen on ldc?

I think both. I don't think there's anything you can do about it.

Eseentially, instead of updating/reading the variable for "current position" in a buffer from the struct itself (which requires an indirection, and always stores/loads), you just declare a local variable, then save it back to the struct at the end.

Like the variable `newChecked` here, instead of using `this.checked` everywhere: https://github.com/schveiguy/iopipe/blob/master/source/iopipe/textpipe.d#L332

JackStouffer:
>Yep, there is an optimization from decoding from dchar, should be an equivalent optimization for encoding to dchar.

I get zero change in performance with this patch

```
+                            {
                                 dchar dc = () @trusted { return decodeFront!(Yes.useReplacementDchar)(r); }();
-                            fill = cast(ushort) encode!(Yes.useReplacementDchar)(buf, dc);
+                            }
+
+                            static if (is(C == dchar))
+                            {
+                                fill = 1;
+                                buf[0] = dc;
+                            }
+                            else
+                            {
+                                fill = cast(ushort) encode!(Yes.useReplacementDchar)(buf, dc);
+                            }
```

DmitryOlshansky:
After fixing *ehm* *completely incorrect* but fast version, I get this on LDC 1.8beta (best of ~3 runs):
```
// original
ASCII 214 ms, 234 μs, and 2 hnsecs
Unicode 195 ms, 28 μs, and 8 hnsecs
strcasecmp 5 ms, 877 μs, and 6 hnsecs

// decodeFront
ASCII 139 ms, 813 μs, and 1 hnsec
Unicode 133 ms, 514 μs, and 4 hnsecs
strcasecmp 5 ms, 752 μs, and 7 hnsecs

// decodeFront + ASCII handling
ASCII 127 ms, 495 μs, and 9 hnsecs
Unicode 130 ms, 685 μs, and 7 hnsecs
strcasecmp 5 ms, 857 μs, and 6 hnsecs

```

DmitryOlshansky:
> That's rather disturbing. Is this only a dmd problem, or does it also happen on ldc?

In my tests LLVM can actually optimize by pulling member of `this` struct into registers, though not always.
DMD almost never does it.

DmitryOlshansky:
For posterity DMD's "results":
```
// Before
▶ ./bench_uni                                                 
ASCII 736 ms, 878 μs, and 5 hnsecs
Unicode 1 sec, 313 ms, 676 μs, and 1 hnsec
strcasecmp 9 ms, 812 μs, and 9 hnsecs

//After 
▶ ./bench_uni 
ASCII 275 ms, 750 μs, and 3 hnsecs
Unicode 668 ms, 452 μs, and 6 hnsecs
strcasecmp 12 ms, 245 μs, and 5 hnsecs
```

DmitryOlshansky:
So we still take x10 more time to case-insensitively compare 2 ascii strings most likely due to not using SIMD.

n8sh:
If you're really concerned about ASCII performance we can get much closer to strcasecmp by putting a special case up at the top:
```d
    static if (isDynamicArray!S1 && isDynamicArray!S2
        && is(ElementEncodingType!S1 == ElementEncodingType!S2))
    {
        // Optimization for ASCII narrow strings that also helps with dstrings.
        // Could add optimizations for cases where only one of the two
        // arguments is a narrow string.
        size_t i = 0;
        for (immutable end = r1.length > r2.length
                ? r2.length : r1.length; i < end; ++i)
        {
            auto lhs = r1[i];
            auto rhs = r2[i];
            if ((lhs | rhs) >= 0x80) goto non_ascii_path;
            if (lhs == rhs) continue;
            auto lowDiff = std.ascii.toLower(lhs) - std.ascii.toLower(rhs);
            if (lowDiff) return lowDiff;
        }
        return 0;

        non_ascii_path:
        r1 = r1[i .. $];
        r2 = r2[i .. $];
        // Fall through to "while (!r1.empty)"
    }
```

Here are benchmark results on my machine. "new_sicmp_v1" is the version from this pull request, and "new_sicmp_v2" is the version from this pull request plus the above snippet.

Using `ldc2 -O2 -release`:
```
old_sicmp ASCII 292 ms, 983 μs, and 2 hnsecs
old_sicmp Unicode 297 ms and 140 μs
strcasecmp 29 ms, 857 μs, and 7 hnsecs
new_sicmp_v1 ASCII 57 ms, 17 μs, and 8 hnsecs
new_sicmp_v2 ASCII 30 ms, 866 μs, and 4 hnsecs
new_sicmp_v1 Unicode 116 ms, 711 μs, and 3 hnsecs
new_sicmp_v2 Unicode 128 ms, 62 μs, and 6 hnsecs
```

Using `dmd -O -inline -release`:
```
old_sicmp ASCII 671 ms, 773 μs, and 8 hnsecs
old_sicmp Unicode 1 sec, 208 ms, 11 μs, and 7 hnsecs
strcasecmp 32 ms, 919 μs, and 5 hnsecs
new_sicmp_v1 ASCII 84 ms, 394 μs, and 3 hnsecs
new_sicmp_v2 ASCII 38 ms, 703 μs, and 9 hnsecs
new_sicmp_v1 Unicode 282 ms, 52 μs, and 2 hnsecs
new_sicmp_v2 Unicode 286 ms, 652 μs, and 4 hnsecs
```

std/uni.d
```diff
@@ -7777,20 +7777,26 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
...
+        immutable rhs = decodeFront!(Yes.useReplacementDchar)(r2);
         int diff = lhs - rhs;
         if (!diff)
             continue;
+        else if ((lhs | rhs) < 0x80)
+        {
+            auto lowL = std.ascii.toLower(lhs);
+            auto lowR = std.ascii.toLower(rhs);
+            if (!(lowL - lowR)) continue;
+            else return lowL - lowR;
```
BBasile: The code generated is _slightly_ better (-2 MOVs and -1 SUB) with DMD for

```d
auto d = std.ascii.toLower(lhs) - std.ascii.toLower(rhs);
if (!d) continue;
else return d;
```

std/uni.d
```diff
@@ -7777,20 +7777,26 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
     && isInputRange!S2 && isSomeChar!(ElementEncodingType!S2))
 {
     import std.internal.unicode_tables : sTable = simpleCaseTable; // generated file
-    import std.utf : byDchar;
-
-    auto str1 = r1.byDchar;
-    auto str2 = r2.byDchar;
+    import std.utf : decodeFront;
+    import std.typecons : Yes;
+    static import std.ascii;
```
BBasile: Why not a selective and renamed import ?

DmitryOlshansky:
> If you're really concerned about ASCII performance we can get much closer to strcasecmp by putting a special case up at the top:

Indeed that would be nice. The biggest issue is that case-insensitive comparisons are often used in parsing text-mode machine protocols (for better or worse).
I'm not a fun of the result ... but, yes a fast-path + fallback to proper decode seems to be worth it.

Lastly we may want to upgrade the `icmp` as well, no reason for it to be slow on ASCII but in a separate PR.

std/uni.d
DmitryOlshansky: I like explicit std.ascii prefix whenever there is toLower/toUpper and friends

std/uni.d
```diff
@@ -7816,7 +7821,7 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
         // one of chars is not cased at all
         return diff;
     }
-    return str2.empty ? 0 : -1;
```
DmitryOlshansky: if r2.empty it should return 0 not -1.
a proper check is
```d
 int(r2.empty) - 1;
```

JackStouffer:
Ping @DmitryOlshansky

quickfur:
ping @DmitryOlshansky 

Auto tester failing, pls fix.

std/uni.d
n8sh: Pushed a fix.

DmitryOlshansky:
> Ping @DmitryOlshansky

Lost that, thanks for taking action @n8sh !
Github notifications are either massive or I miss comments... I haven't found balance yet.

PR #4951, opened 2016-12-13
Fix link format for changes in PR 4940 (std.numeric.gcd)

wilzbach:
Fixup for #4940, which was merged prematurely :/

PR #4411, opened 2016-06-05
Fix broken WEB links

wilzbach:
As mentioned in https://github.com/dlang/dlang.org/pull/1337 and https://github.com/dlang/dlang.org/pull/1333, yet another `sed` adventure!
Today's purge are the `WEB`, `HTTP`, `HTTPS` macros. 

Reason: We still have way too many custom ddoc macros to make simple links - creating a link should be simple!
Have a look at this selection:

```
A = <a href="$1">$+</a>
AHTTP = <a href="http://$1">$+</a>
AHTTPS = <a class="https" href="https://$1">$+</a>
ALOCAL = <a href="#$1">$+</a>

LINK = $(A $0, $0)
LINK2 = $(A $1, $+)

HTTP = $(LINK2 http://$1,$2)
HTTPS = $(LINK2 https://$1,$2)
WEB = $(HTTP $1,$2)
```

Some people might argue that `WEB` is great, because it's a bit shorter, but
- it leads to problems if you think it's a normal link and use `http://` (happens quite often, even here in Phobos)
- it discourages the fact that we are in 2016 and SSL should be everywhere. `WEB` always resolves to `http://`
- yet another macro that does the same thing

Or to quote Andrei from a recent forum discussion:

> Yah, I wouldn't disagree things can be vastly improved. My point there is to make sure we distinguish > what's trivially fixable (e.g. redundant macro definitions) from larger issues. Again, if you find identical macros across dd/ddoc files, or redundant macros that do the same thing in just slightly different ways, please fix or file. Thanks!
>
> Andrei
http://forum.dlang.org/post/ni4b13$21go$1@digitalmars.com

Hence this PR replaces all `WEB`, `HTTP` and `HTTPS` macros with `LINK2`. Ideally in the future we can use "markdown-like" links in ddoc - we all love them from Github, right?

Commands
----------------

This PR is done fully automated - again regular expressions to the help!

1) fix broken WEB links `sed 's/(WEB http:\/\//(WEB /' -i **/*.d`
2) HTTP links `sed 's/(WEB /(LINK2 http:\/\//' -i */.d`
3) single-line WEB links `sed 's/\$(WEB$/\$(LINK2\nhttp:\/\//' -i **/*.d`
4) multi-line links `perl -0777 -i -pe 's/\$\(WEB\n(\s*)/\$(LINK2\n\1http:\/\//g' **/*.d`

Should I add a simple grep to `travis.yml` or can we ensure - with the help of DAutotester - that those macros aren't used in the future?

PR #6226, opened 2018-02-26
fix bad url in std-file-readText.dd

timotheecour:


changelog/std-file-readText.dd
```diff
@@ -1,7 +1,7 @@
 readText now checks BOMs
 
 $(REF readText, std, file) now checks for a
-$(HTTP https://en.wikipedia.org/wiki/Byte_order_mark, BOM). If a BOM is present
+$(HTTPS en.wikipedia.org/wiki/Byte_order_mark, BOM). If a BOM is present
```
wilzbach: Ideally we phase out this entire HTTP/HTTPS non-sense, and just use LINK2, but Andrei has been against this in the past.

wilzbach:
Merged manually due to https://github.com/braddr/d-tester/issues/70

changelog/std-file-readText.dd
timotheecour: breaking down links like that is UGLY and contrary to having readable docs when they're un-rendered (eg we can't simply copy paste these links to/from a browser); and it's error prone as examplified here

wilzbach: See also:

https://github.com/dlang/phobos/pull/4411#issuecomment-226326952
https://github.com/dlang/phobos/pull/4951#issuecomment-266670581

In short:
- at least there's `WEB` anymore
- The Ddoc issue has been fixed, so in theory we could try this again, but I'm not sure whether it's worth the pain.

Though maybe we can can @andralex's approval this time?

andralex: Typing `https://` is awkward and integrating it in a macro is sensible. We could and should fix matters on the ddoc side; the issue of line breakage in macros is not singular to this one. Until then please leave HTTP and HTTPS where they are. Thanks.

PR #3264, opened 2015-05-07
Rename *Tuple symbols

WalterBright:
Does the following renames in std.traits:

isExpressionTuple => isExpressions
 ParameterTypeTuple => Parameters
 FieldTypeTuple => Fields
 ParameterDefaultValueTuple => ParameterDefaults

and in std.typetuple:

 TypeTuple => Arguments

The original names are kept and documented as aliases kept for legacy compatibility. I didn't do a global search/replace in Phobos to change the names, preferring to keep this PR simple. Over time, as various modules get worked on, the names can be changed.

PR #4968, opened 2016-12-19
Deprecate std.typetuple

wilzbach:
Split off #4954 that simply deprecates `std.typetuple : TypeTuple` in favor of `std.meta : AliasSeq`.
It also removes `typetuple` from the documentation build.

Note that the [documentation](https://dlang.org/phobos/std_typetuple.html) has stated the rename for quite a long time, but it didn't trigger deprecation warnings

PR #4954, opened 2016-12-13
Update std.traits to reflect renaming of TypeTuple to AliasSeq

wilzbach:
Revived #3756 as is (just rebased to latest Master).

PR #6227, opened 2018-02-27
Continue renaming std.traits.*Tuple as per #3264

ntrel:
Continue renaming of templates resolving to a sequence, but are named as `*Tuple`.
Some new names don't follow the pattern (shown in italic):

* ParameterIdentifierTuple -> *ParameterNames* (consistency with FieldNames, shorter)
* ParameterStorageClassTuple -> ParameterStorageClasses
* BaseClassesTuple -> BaseClasses
* BaseTypeTuple -> BaseTypes
* InterfacesTuple -> *BaseInterfaces* (consistency with above two, interfaces can inherit too)
* FieldNameTuple -> FieldNames
* MemberFunctionsTuple -> *MemberOverloads* (more precise as it takes a member name argument)
* RepresentationTypeTuple -> RepresentationTypes
* TransitiveBaseTypeTuple -> TransitiveBaseTypes
* isTypeTuple -> isTypes (to match the `isExpressionTuple` -> `isExpressions` change)

As #3264 didn't deprecate the legacy aliases, I haven't done so either. BTW that pull did make the following changes, which I think are inaccurate:
* ParameterTypeTuple -> Parameters
* FieldTypeTuple -> Fields

I would prefer those aliases were *ParameterTypes* and *FieldTypes*. Mentioning now as it may be a good time to fix those too.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -24,19 +24,20 @@
  *           $(LREF isUnsafe)
  *           $(LREF isFinal)
  *           $(LREF ParameterDefaults)
- *           $(LREF ParameterIdentifierTuple)
- *           $(LREF ParameterStorageClassTuple)
+ *           $(LREF ParameterNames)
+ *           $(LREF ParameterStorageClasses)
  *           $(LREF Parameters)
  *           $(LREF ReturnType)
  *           $(LREF SetFunctionAttributes)
  *           $(LREF variadicFunctionStyle)
  * ))
  * $(TR $(TD Aggregate Type _traits) $(TD
- *           $(LREF BaseClassesTuple)
- *           $(LREF BaseTypeTuple)
+ *           $(LREF BaseClasses)
+ *           $(LREF BaseInterfaces)
+ *           $(LREF BaseTypes)
  *           $(LREF classInstanceAlignment)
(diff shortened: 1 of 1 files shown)


JackStouffer:
This needs a changelog entry.

Also, I'd argue that deprecations should be added onto these aliases. Any objections?

MetaLang:
They probably weren't deprecated back in 2015 because we weren't as worried about breaking code. I agree with @JackStouffer.

wilzbach:
> Also, I'd argue that deprecations should be added onto these aliases. Any objections?

I tried to revive this too: https://github.com/dlang/phobos/pull/4954

Andrei closed it with:

> This seems to be a pass that someone made through a personal project after finding a nicer naming convention. I think it breaks too many things for too little benefit.

So yeah, it looks like we can't deprecate the aliases, but imho we should still do the renames.

BTW there was also https://github.com/dlang/phobos/pull/4968 which didn't get any interest at that time.

andralex:
No more grand renamings please.

JackStouffer:
Well, since Andrei is in charge of names I think it's safe to close this.

ntrel:
Would this be appropriate for `std2`?

PR #6232, opened 2018-02-27
Temporarily disable curl tests for FreeBSD 32-bit

wilzbach:
See also: https://github.com/braddr/d-tester/issues/70#issuecomment-368670639

I'm aware that this isn't ideal and `StdUnittest` would be better suited for this, but it keeps the diff small and `std.net.curl` is broken on FreeBSD 32-bit anyhow.
Moreover, I have the opinion that we should just drop the support for legacy platforms that almost no one actively uses, but well that's a different story.

### DIFF
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -164,6 +164,16 @@ import std.encoding : EncodingScheme;
 import std.traits : isSomeChar;
 import std.typecons : Flag, Yes, No, Tuple;
 
+// Curl tests for FreeBSD 32-bit are temporarily disabled.
+// https://github.com/braddr/d-tester/issues/70
+// https://issues.dlang.org/show_bug.cgi?id=18519
+version(unittest)
+version(FreeBSD)
+version(X86)
+    version = DisableCurlTests;
+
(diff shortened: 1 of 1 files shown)


JackStouffer:
Does this issue effect the beta? If so, we should make a final decision about what to do before it's released.

wilzbach:
> Does this issue effect the beta? If so, we should make a final decision about what to do before it's released.

`std.net.curl` is broken on FreeBSD 32-bit for on anything newer than FreeBSD 8. FreeBSD 9 was released in 2012 and support for FreeBSD 8 ended in 2015.
So as far as I can judge, it doesn't change the status quo at all, but I think it makes merging easier if I target `stable` and then merge it back to `master`.

BTW looking at the auto-tester report I see that somehow FreeBSD 32-bit didn't get started for this PR. 
CC @braddr

wilzbach:
The [workaround at dmd](https://github.com/dlang/dmd/pull/7952) was targeted at `master`, not `stable`. So `stable` is still broken. Going to rebase this again to `master` then ...

PR #6234, opened 2018-02-28
fix issue 18519 freebsd 11 + phobos + curl, timing out

rainers:
stop the TestServer waiting in accept() before thread_joinall

Windows is also affected if libcurl.dll is missing.

### DIFF
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -185,6 +185,7 @@ version(unittest)
     private:
         string _addr;
         Tid tid;
+        TcpSocket sock;
 
         static void loop(shared TcpSocket listener)
         {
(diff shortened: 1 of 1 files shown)


std/net/curl.d
```diff
@@ -214,21 +215,32 @@ version(unittest)
         import std.concurrency : spawn;
         import std.socket : INADDR_LOOPBACK, InternetAddress, TcpSocket;
 
+        tlsInit = true;
```
atilaneves: Is this ever _not_ true?

rainers: tlsInit stays false in all threads that did not start the TestServer.

atilaneves: I thought it was something like that but couldn't tell from the patch.

atilaneves:
LGTMn

wilzbach:
FYI this doesn't fix the problem for FreeBSD11. I just tested this with https://github.com/dlang/phobos/pull/6264

PR #5284, opened 2017-03-16
Improve docs for isSomeString and isNarrowString.

jmdavis:
The main purpose of these changes is to provide warnings about the fact
that isSomeString and isNarrowString are true for enums, but they also
improve the docs for them in general - including adding a description of
what a narrow string actual is, since that was lacking before.

I would have liked to have included some links to something which sanely explained code units, code points, and graphemes, but I couldn't find anything like that in a quick online search. I suppose that what I should really do is finally get around to writing up a nice introductory article to the range/auto-decoding mess and then provide a link to that.

PR #5137, opened 2017-02-15
Add std.traits.isStringLike to check for types similar to strings

wilzbach:
Follow-up to #5132 and the [NG discussion](http://forum.dlang.org/post/egudvlzgtwrqpdkmgwfb@forum.dlang.org).

The idea is that a common constraints combination like here with `isStringLike` can help to make the constraints easier to read.
It uses the definition as it is used e.g. by [`std.file`](https://github.com/dlang/phobos/blob/master/std/file.d#L235) though I have to admit that I am not quite sure whether to exclude `isConvertibleToString` types is a good idea as it excludes the following three types:

```d
// Static string arrays
char[10];
// string enums
enum StringEnum { a = "foo" }

// Aggregated strings
static struct Stringish
{
    string s;
    alias s this;
}
```

After all methods in `std.file` is usually just calls [`tempCString`](https://github.com/dlang/phobos/blob/master/std/internal/cstring.d#L89) which has these constraints for `From`:

```d
(isInputRange!From || isSomeString!From) &&
        isSomeChar!(ElementEncodingType!From)
```

PR #6235, opened 2018-02-28
Added writer overload of toString to Appender

JackStouffer:


std/array.d
```diff
@@ -3158,7 +3159,49 @@ if (isDynamicArray!A)
...
+        // Even though the Appender might represent a string,
+        // make a copy of the data anyway because we don't want
+        // manipulations of the toString result to change the
+        // data of the Appender
+        auto app = appender!string();
+        auto spec = singleSpec("%s");
+        // different reserve lengths because each element in a
+        // non-string-like array uses two extra characters for `, `
+        static if (isSomeString!A)
+            app.reserve(_data.arr.length + 10);
```
n8sh: Why not just `_data.arr.length` if the source is a string?

JackStouffer:
There's some sort of odd call to appender's toString happening somewhere in `conv.to`. Currently investigating.

JackStouffer:
I found a bug in `hasToString` that I need help fixing. @n8sh @wilzbach

Consider

```d
import std.format;
import std.range;

struct A(T)
{
    string toString() const { return ""; }
    void toString(W)(ref W w) const { put(w, ""); }
}

struct B(T)
{
    string toString()() const { return ""; }
    void toString(W)(ref W w) const { put(w, ""); }
}

void main()
{
    pragma(msg, hasToString!(A!(int), char)); // 1
    pragma(msg, hasToString!(B!(int), char)); // 5
}
```

It has to do with this line https://github.com/dlang/phobos/blob/8f0094593c18466095da3156d6d574c93c26466c/std/format.d#L3627

Because the first toString overload in `A` is not a template but a function, the attempted instantiation of the `toString` template inside `hasToString` always fails despite the fact that there is an existing `toString` template.

JackStouffer:
This needs to be fixed before we release 2.079

JackStouffer:
Needs https://github.com/dlang/phobos/pull/6247

JackStouffer:
Ready to go.

std/array.d
```diff
@@ -3173,7 +3174,51 @@ if (isDynamicArray!A)
...
+     *     is formatted.
+     * Returns:
+     *     A `string` if `writer` is not set; `void` otherwise.
+     */
+    string toString()() const
+    {
+        // Even though the Appender might represent a string,
+        // make a copy of the data anyway because we don't want
+        // manipulations of the toString result to change the
+        // data of the Appender
```
n8sh: If the element type is immutable how could manipulations of the toString result change the data of the Appender?

std/array.d
```diff
@@ -3173,7 +3174,51 @@ if (isDynamicArray!A)
...
+        auto app = appender!string();
+        auto spec = singleSpec("%s");
+        // different reserve lengths because each element in a
+        // non-string-like array uses two extra characters for `, `
+        static if (isSomeString!A)
```
wilzbach: FWIW we wanted to deprecate this trait...

https://github.com/dlang/phobos/pull/5284, https://github.com/dlang/phobos/pull/5137

std/array.d
wilzbach: We could also support `const(string) toString()() const` ...

JackStouffer: hmmm, you're right, I was thinking more of the `char[]` case. In any case, the code remains the same regardless to keep it dry and to use the speedup of Appender.

I'll just remove the comment.

std/array.d
```diff
@@ -3173,7 +3174,51 @@ if (isDynamicArray!A)
...
+            app.reserve(_data.arr.length + 15);
```
wilzbach: `Appender!(int[])()` has 18 character, so maybe `+ 20` or `+25` or maybe even better with `Unqual!(typeof(this)).stringof.length + 2`?

std/array.d
```diff
@@ -3173,7 +3174,51 @@ if (isDynamicArray!A)
...
+        else
+            app.reserve((_data.arr.length * 3) + 15);
```
wilzbach: Hmm, I assume this is the conservative estimate that of `X, ` being three characters which does make sense, though I think it would be nice to add a comment for future readers about this conservative estimate.

std/array.d
JackStouffer: `isSomeString` is far too ubiquitous to deprecate IMO. Johnathan already modified it to not recognize `enum`s, which I think was the main issue.

std/array.d
```diff
@@ -3173,7 +3174,47 @@ if (isDynamicArray!A)
...
+     *     w = A `char` accepting
+     *     $(REF_ALTTEXT output range, isOutputRange, std, range, primitives).
+     *     fmt = A $(REF FormatSpec, std, format) which controls how the array
+     *     is formatted.
+     * Returns:
+     *     A `string` if `writer` is not set; `void` otherwise.
+     */
+    string toString() const
+    {
+        auto app = appender!string();
```
wilzbach: In theory we could also do `const(string) toString()() const { return app.data }` for `appender!string`, but that's a potential optimization for the future.

JackStouffer: You still need to add `Unqual!(typeof(this)).stringof`

wilzbach: Fair point. Though that could be solved by allowing `toString` to return a range, but that's a story for a different PR.

JackStouffer:
That weird std.conv error is showing up again. I'll fix this when I get a chance.

std/uni.d
```diff
@@ -7102,7 +7102,7 @@ public: // Public API continues
     Returns:
         length of grapheme cluster
 +/
-size_t graphemeStride(C)(const scope C[] input, size_t index)
+size_t graphemeStride(C)(const scope C[] input, size_t index) @safe pure
```
JackStouffer: After going all the way down the rabbit hole, I found out that due to some compiler bug, DMD marked this as `@system` when it was instantiated from `std.formatRange` from `std.conv.textImpl`. No idea why. Manually adding the attributes fixes the whole issue.

PR #6236, opened 2018-02-28
Regression test for #18539 added

drug007:
Regression for 2.078.*. Fixed in future 2.079 (dmd-beta).

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -3030,6 +3030,17 @@ auto nullable(T)(T t)
     assert(c.canary == 0xA71FE);
 }
 
+// Regression test for issue 18539
+@safe unittest
+{
+    import std.math : approxEqual;
+
+    auto foo = nullable(2.0);
+    auto bar = nullable(2.0);
+
+    assert(foo.approxEqual(bar));
+}
+
 /**
 Just like $(D Nullable!T), except that the null state is defined as a
 particular value. For example, $(D Nullable!(uint, uint.max)) is an


wilzbach:
FYI: no need to push commits like "remove trailing whitespace" to the master tree. As long as we don't enable auto-merge-squash, imho such things should be squashed manually.
I also changed the commit message to reference the issue, s.t. it's automatically closed.

FeepingCreature:
@drug007 Ping!

I found this PR while trying to prepare for a PR to deprecate the `alias get this` in Nullable, since it causes no end of trouble for us at work. Since you went through the trouble to get a regression test for this into Phobos, I wanted to ask if this is a critical feature for you before I'd consider removing your test.

drug007:
First of all I appreciate you asked me.
I can't estimate consequences of `alias get this` removing except it will definitely break my code. What sort of problems do you try to solve?

FeepingCreature:
We're using `Nullable` as a stdlib-based "optional" type, and we keep running into issues where we accidentally use `Nullable!T` as `T`, or make some field `Nullable` and forget to add `isNull` checks, and instead of a nice informative compile error we get a runtime crash, potentially arbitrary time later, potentially in production. As I keep semi-joking, "Nullable adds pointer semantics to arbitrary types, up to and including the null pointer crash". It's just bad design. If the compiler can tell that an error may occur, it should require it to be handled, _especially_ if the alternative is only guarded by an AssertError.

drug007:
I agree that `Nullable` isn't great but how disabling `alias get this` helps here? Without `alias get this` you will be forced by compiler to use additional `get` so you can't use `Nullable!T` as `T`, but it won't force you to use `isNull` check properly. Also I wouldn't say that capability of `Nullable!T` to mimic `T` is disadvantage - in my case it is opposite. And in meta programming, the need to distinguish case of `Nullable`, where I need to use the extra `get` to get value, complicates the code, and IMHO is a disadvantage.
That's debatable of course, but I think that removing of `alias get this` do not improve `Nullable` in general but brings breaking changes. Using of other more appropriate data structures  would be better approach here.

FeepingCreature:
Without `alias get this` you will not convert `Nullable!T` to `T` **by accident.**

drug007:
In general I would prefer do not change my code of course, but your solution demands few efforts to fix my code and although it makes code uglier (and more safe, I agree with you) I'm fine with it.

FeepingCreature:
Thank you!

drug007:
Not at all! 
Just in case - did you consider using for example [optional](http://code.dlang.org/packages/optional) package?

FeepingCreature:
`optional` seems neat, but I'm not sure of the decision to make the equivalent of `.get` be `.front`. Why would `Optional!T` be foremost a range? It seems an unusual way to look at it.

lesderid:
> `optional` seems neat, but I'm not sure of the decision to make the equivalent of `.get` be `.front`. Why would `Optional!T` be foremost a range? It seems an unusual way to look at it.

It's so it can be used similar to how an option monad can be used in functional languages. Scala does it too. There's a good example at the top of this page: https://www.scala-lang.org/api/current/scala/Option.html.

FeepingCreature:
I get it, I just don't agree it's a good idea.

PR #6239, opened 2018-03-01
Update lockingTextWriter docs

JackStouffer:
Add note about buffering.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2981,7 +2981,7 @@ $(D Range) that locks the file and allows fast writing to it.
     }
 
     /**
-     * Output range which locks the file when created, and unlocks it when it goes
+     * Output range which locks the file when created, and unlocks the file when it goes
      * out of scope.
      *
      * Returns: An $(REF_ALTTEXT output range, isOutputRange, std, range, primitives)
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2981,8 +2981,17 @@ $(D Range) that locks the file and allows fast writing to it.
     }
 
     /**
+     * Output range which locks the file when created, and unlocks it when it goes
+     * out of scope.
+     *
      * Returns: An $(REF_ALTTEXT output range, isOutputRange, std, range, primitives)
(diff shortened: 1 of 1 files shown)


std/stdio.d
```diff
@@ -2981,8 +2981,16 @@ $(D Range) that locks the file and allows fast writing to it.
...
+     * out of scope.
+     *
      * Returns: An $(REF_ALTTEXT output range, isOutputRange, std, range, primitives)
-     * that locks the file and allows fast writing to it.
+     * which accepts string types, `ubyte[]`, individual character types, and
+     * individual `ubyte`s.
+     *
+     * Note: Writing either arrays of `char`s or `ubyte`s is much faster than
+     * writing each character individually a range. For large files,
+     * writing using an array acting as output buffer can result in a speed increase.
```
schveiguy: "individually a range" -> ??? Not sure what you meant here, but it needs rewording

"For large files, writing using an array acting as output buffer can result in a speed increase."

Isn't this redundant? You just said writing an array is "much faster". I don't get the point about "acting as an output buffer".

std/stdio.d
```diff
@@ -2981,8 +2981,17 @@ $(D Range) that locks the file and allows fast writing to it.
...
+     * Output range which locks the file when created, and unlocks it when it goes
+     * out of scope.
+     *
      * Returns: An $(REF_ALTTEXT output range, isOutputRange, std, range, primitives)
-     * that locks the file and allows fast writing to it.
+     * which accepts string types, `ubyte[]`, individual character types, and
+     * individual `ubyte`s.
+     *
+     * Note: Writing either arrays of `char`s or `ubyte`s is faster than
+     * writing each character individually from a range. For large files,
```
jondegenhardt: Sentence talks about the speed increase writing to files resulting from writing in chunks. A minor clarification would be that this applies to writing to `stdout` as well as files.

schveiguy: How about just "large amounts of data"?

jondegenhardt: Sounds goods to me.

PR #6241, opened 2018-03-01
Prefer an appender to a dynamic array

cedretaber:
benchmark:
```d
unittest
{
    import std.stdio, std.datetime.stopwatch;

    auto str = "a bb ccc dddd eeee fff gg h i jj kkk llll mmmm nnn oo p q rr sss tttt uuuu vvv ww x y zz";

    void use_an_array() { auto ret = split(str); }
    void use_an_appender() { auto ret = split2(str); }

    auto r = benchmark!(use_an_array, use_an_appender)(1_000_000);

    writefln("use_an_array: %s", r[0]);
    writefln("use_an_appender: %s", r[1]);
}
```

output:
```
use_an_array: 4 secs, 343 ms, 762 μs, and 4 hnsecs
use_an_appender: 2 secs, 882 ms, 757 μs, and 9 hnsecs
```

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -1472,7 +1472,7 @@ if (isSomeString!S)
 {
     size_t istart;
     bool inword = false;
-    S[] result;
+    auto result = appender!(S[]);
 
     foreach (i, dchar c ; s)
     {
@@ -1481,7 +1481,7 @@ if (isSomeString!S)
         {
             if (inword)
             {
-                result ~= s[istart .. i];
+                put(result, s[istart .. i]);
                 inword = false;
             }
         }
@@ -1495,8 +1495,8 @@ if (isSomeString!S)
         }
     }
     if (inword)
-        result ~= s[istart .. $];
-    return result;
+        put(result, s[istart .. $]);
+    return result.data;
 }
 
 ///


JackStouffer:
Is the `put` actually copying the data into the appender, or just copying the slice?

JackStouffer:
Nevermind. It seems like `~=` also just copied the slice, so this doesn't seem to be an issue.

```d
import std.stdio;
import std.conv;
import std.array;
import std.range;
import std.algorithm;
import std.string;

 void main()
 {
     char[] str = to!(char[])("Hello World!");
     auto res = str.split;
     res[0][0] = 'x';
     assert(str[0] == 'x');
}
```

JackStouffer:
Can you post the performance numbers using LDC as well?

cedretaber:
```
$ ldc2 --version
LDC - the LLVM D compiler (1.7.0):
  based on DMD v2.077.1 and LLVM 5.0.1
  built with LDC - the LLVM D compiler (0.17.5)
  Default target: x86_64-apple-darwin16.7.0
  Host CPU: skylake
  http://dlang.org - http://wiki.dlang.org/LDC

  ...

$ ldc2 test.d
$ ./test
use_an_array: 1 sec, 898 ms, 17 μs, and 7 hnsecs
use_an_appender: 1 sec, 558 ms, 382 μs, and 2 hnsecs
$ ldc2 -O test.d
$ ./test
use_an_array: 1 sec, 814 ms, 969 μs, and 3 hnsecs
use_an_appender: 993 ms, 264 μs, and 8 hnsecs
```

std/array.d
```diff
@@ -1481,7 +1481,7 @@ if (isSomeString!S)
         {
             if (inword)
             {
-                result ~= s[istart .. i];
+                result.put(s[istart .. i]);
```
jmdavis: Please don't use UFCS with `put`. Given that this code is using a specific output range, it's not as bad, but in general, it's just begging for problems when using `put` with strings.

jmdavis: The same goes below.

jmdavis:
Please squash the commits, and then I think that it's good to go.

cedretaber:
What can I do to be approved on auto-tester?

jmdavis:
> What can I do to be approved on auto-tester?

Someone has to add you to the auto-tester. I thought that Brad had to do that, but it looks like I can, and in theory, I just did. So, hopefully the auto-tester will test your PRs now.

PR #6242, opened 2018-03-02
[ASan] Handle the (invalid) return value of posix_memalign

JohanEngelen:
See https://github.com/ldc-developers/phobos/pull/58

With ASan enabled, posix_memalign returns -1 upon invalid parameters or when not enough memory is available. This bug is reported with compiler-rt here: https://bugs.llvm.org/show_bug.cgi?id=36510
Still, if we deal with the invalid -1 return value in Phobos, we can continue testing with ASan enabled.
It looks like this workaround is only needed on OSX.

@andralex

### DIFF
--- a/std/experimental/allocator/mallocator.d
+++ b/std/experimental/allocator/mallocator.d
@@ -243,7 +243,6 @@ version(LDC_AddressSanitizer)
         if (code == -1)
             return null;
 }
-
         if (code == ENOMEM)
             return null;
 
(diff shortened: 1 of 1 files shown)


andralex:
thx

JohanEngelen:
The issue is fixed in LLVM current trunk (LLVM 7), so this workaround could be removed for very new LLVM versions.

PR #6244, opened 2018-03-03
[dmd-cxx] Backport PIC logic from stable into dmd-cxx

ibuclaw:
This would be a prerequisite for https://github.com/dlang/dmd/pull/7953

Ping @wilzbach


ZombineDev:
Most CIs fail somewhere in druntime, so we need to fix it, before it can be green.

PR #6247, opened 2018-03-05
Fix Issue 18548 - std.format ignores templated toString if another to…

JackStouffer:
…String is not a template

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -3550,20 +3550,25 @@ private template hasToString(T, Char)
         enum hasToString = 0;
     }
     else static if (is(typeof(
-        {T val = void; FormatSpec!Char f; static struct S {void put(Char s){}}
-        S s; val.toString(s, f);
-        // force toString to take output range by ref
-        static assert(ParameterStorageClassTuple!(T.toString!(S))[0] == ParameterStorageClass.ref_);
-        static assert(ParameterStorageClassTuple!(T.toString!(S))[1] == ParameterStorageClass.ref_);
-        static assert(is(Parameters!(T.toString!(S))[1] == const));}
+        {T val = void;
+        const FormatSpec!Char f;
+        static struct S {void put(Char s){}}
+        S s;
+        val.toString(s, f);
+        // force toString to take parameters by ref
+        static assert(!__traits(compiles, val.toString(s, FormatSpec!Char())));
+        static assert(!__traits(compiles, val.toString(S(), f)));}
     )))
     {
         enum hasToString = 6;
     }
     else static if (is(typeof(
-        {T val = void; static struct S {void put(Char s){}}
-        S s; val.toString(s);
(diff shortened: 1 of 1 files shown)


std/format.d
```diff
@@ -3550,20 +3550,24 @@ private template hasToString(T, Char)
...
-        S s; val.toString(s, f);
+        {T val = void;
+        const FormatSpec!Char f;
+        static struct S {void put(Char s){}}
+        S s;
         // force toString to take output range by ref
-        static assert(ParameterStorageClassTuple!(T.toString!(S))[0] == ParameterStorageClass.ref_);
-        static assert(ParameterStorageClassTuple!(T.toString!(S))[1] == ParameterStorageClass.ref_);
-        static assert(is(Parameters!(T.toString!(S))[1] == const));}
+        static assert( __traits(compiles, val.toString(s, f)));
```
wilzbach: `static assert( __traits(compiles` is unnecessary, isn't it? Or in other words: `val.toString(s, f)` should work fine as it's already is `is(typeof(`.

std/format.d
```diff
@@ -3614,7 +3620,35 @@ private template hasToString(T, Char)
...
+    {
+        string toString() {return "";}
+        void toString(Writer)(ref Writer w) if (isOutputRange!(Writer, string)) {}
+    }
+    static struct H
+    {
+        string toString() {return "";}
+        void toString(Writer)(ref Writer w, const ref FormatSpec!char fmt)
+        if (isOutputRange!(Writer, string))
+        {}
```
wilzbach: Maybe add an additional test for a) `Writer` not being `ref` and b) `FormatSpec` not being `ref`?

JackStouffer:
Fixed

std/format.d
```diff
@@ -3623,6 +3670,12 @@ private template hasToString(T, Char)
     static assert(hasToString!(D, char) == 2);
     static assert(hasToString!(E, char) == 1);
     static assert(hasToString!(F, char) == 6);
+    static assert(hasToString!(G, char) == 5);
+    static assert(hasToString!(H, char) == 6);
+    static assert(hasToString!(I, char) == 6);
+    static assert(hasToString!(J, char) == 1);
+    static assert(hasToString!(K, char) == 4);
```
wilzbach: That was rather unexpected.
4 is:

```
else static if (is(typeof({ T val = void; FormatSpec!Char f; val.toString((const(char)[] s){}, f); })))
```

and:

```
static assert(isOutputRange!(typeof((const(char)[] s){}), char));
```

so it looks ok.

wilzbach:
Stable is currently broken at auto-tester: https://github.com/braddr/d-tester/issues/71

JackStouffer:
@wilzbach Can you merge stable back onto master?

wilzbach:
> @wilzbach Can you merge stable back onto master?

Sure - https://github.com/dlang/phobos/pull/6271

PR #6249, opened 2018-03-06
Added writer version of toString to std.datetime.interval.Interval

JackStouffer:


std/datetime/interval.d
```diff
@@ -1536,47 +1537,38 @@ public:
...
-        import std.format : format;
+        import std.array : appender;
+        auto app = appender!string();
         try
-            return format("[%s - %s)", _begin, _end);
+            toString(app);
         catch (Exception e)
-            assert(0, "format() threw.");
+            assert(0, "toString() threw.");
+        return app.data;
```
wilzbach: I think I have now seen this for the 20th time. Couldn't we add a toString function in std.format or somewhere else that is defined exactly when there is no toString but a output range toString?

PR #6250, opened 2018-03-06
std.variant: dummy This struct needs a size

WalterBright:


jmdavis:
@WalterBright BTW, would you consider that struct used as a UDA needs members if it's just used as a UDA (and thus is only used at compile time)? I suspect that using an empty struct is a fairly common thing for UDAs.

WalterBright:
The trouble in this particular case was the undefined struct was used as the key for an AA, leaving the semantic routine in a difficult spot. If it overlooks that it is undefined, and later it gets defined with having no members, then one has found a hole in the checking.

WalterBright:
Here's the issue:
```
std/variant.d(101): Error: struct `std.variant.This` unknown size
std/variant.d(103): Error: template instance `AliasSeq!(ReplaceType!(This, VariantN!(16LU, real, This[], This[int], This[This]), real, This[], This[int], This[This]))` recursive template expansion
std/variant.d(137): Error: alias `std.variant.VariantN!(16LU, real, This[], This[int], This[This]).VariantN.AllowedTypes` recursive alias declaration
std/typecons.d(8017): Error: template instance `std.typecons.ReplaceType!(This, VariantN!(16LU, real, This[], This[int], This[This]), This[This])` error instantiating
std/variant.d(103):        4 recursive instantiations from here: `ReplaceType!(This, VariantN!(16LU, real, This[], This[int], This[This]), real, This[], This[int], This[This])`
std/variant.d(137):        instantiated from here: `This2Variant!(VariantN!(16LU, real, This[], This[int], This[This]), real, This[], This[int], This[This])`
std/variant.d(1500):        instantiated from here: `VariantN!(16LU, real, This[], This[int], This[This])`
std/variant.d(1633):        instantiated from here: `Algebraic!(real, This[], This[int], This[This])`
```
and the `This[This]` is the specific problematic code.

WalterBright:
I redesigned the 0 check so it is done later when the AA is actually used, rendering this PR unnecessary.

PR #6251, opened 2018-03-07
Fix Issue 18168 - std.base64.encode output ranges @safe

JackStouffer:


JackStouffer:
huh, all the tests pass on my machine. currently investigating

JackStouffer:
Fixed.

std/base64.d
```diff
@@ -390,7 +390,7 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
      * Returns:
      *  The number of times the output range's $(D put) method was invoked.
      */
-    size_t encode(R1, R2)(in R1 source, auto ref R2 range)
+    size_t encode(R1, R2)(R1 source, auto ref R2 range)
```
n8sh: Since you have `auto s = source;` below is it necessary to remove `in`?

std/base64.d
```diff
@@ -405,23 +405,23 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
 
         immutable blocks = srcLen / 3;
         immutable remain = srcLen % 3;
-        auto      srcptr = source.ptr;
-        size_t    pcount;
+        auto s = source; // copy for out contract length check
```
n8sh: Because `source` can be a static array I think this has to be `auto s = source[];`

n8sh: If I'm not wrong about this then there probably needs to be a unittest where the source is a static array, since the tests didn't catch this.

std/base64.d
JackStouffer: `in` and `out` are basically obsolete and I want to remove them whenever possible.

n8sh: What if `R1` is a large static array?

n8sh: Ah, I see: even though the first condition is `isArray`, `ElementType` only works for dynamic arrays.

On that note the arguments should probably be changed from `(R1 source, auto ref R2 range)` to `(scope const(E)[] source, auto ref R2 range)` but that could be separate pull request.

JackStouffer:
Fixed

std/base64.d
```diff
@@ -390,9 +390,8 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
      * Returns:
      *  The number of times the output range's $(D put) method was invoked.
      */
-    size_t encode(R1, R2)(in R1 source, auto ref R2 range)
-        if (isArray!R1 && is(ElementType!R1 : ubyte) &&
-            !is(R2 == char[]) && isOutputRange!(R2, char))
+    size_t encode(E, R)(scope const(E)[] source, auto ref R range)
+    if (is(E : ubyte) && isOutputRange!(R, char))
```
wilzbach: char[] is now allowed. What's the motivation for this?

JackStouffer: `char[]` is now allowed to be the output range. I see no reason to disallow it if `isOutputRange!(R, char)` is `true`

JackStouffer:
@wilzbach `source` must always be a dynamic array now, so @n8sh's comment no longer applies.

std/base64.d
wilzbach: Ok, but maybe we should add a test for `char[]` then?

JackStouffer: `isOutputRange!(char[], char) == false` due to auto-decoding.

PR #6252, opened 2018-03-07
Fix Issue 18158 - std.file.getcwd should be usable in @safe

JackStouffer:


std/file.d
```diff
@@ -2865,7 +2865,7 @@ version(Windows) string getcwd()
         return ptr[0 .. n2].to!string;
     }
 }
-else version (Solaris) string getcwd()
+else version (Solaris) string getcwd() @trusted
```
JackStouffer: In this branch and for Posix, the function is simple enough where this can be applied without much worry.

std/file.d
```diff
@@ -2832,7 +2832,7 @@ version(Posix) @system unittest // input range of dchars
  * Get the current working directory.
  * Throws: $(D FileException) on error.
  */
-version(Windows) string getcwd()
+version(Windows) string getcwd() @trusted
```
JackStouffer: this is where this gets a little dangerous. I when through the code and I don't see any security holes. I tried using trusted lambdas here, but the issue is 90% of the function is unsafe.

wilzbach: You could use a ScopedAllocator for the allocation + once we fix checked, then most of this should be @safe.

PR #6254, opened 2018-03-08
Work On Issue 18184 - std.zip should be usable in @safe

JackStouffer:
This gets us 80% there. You can compress stuff completely `@safe`-ly. The rest relies on `std.zlib`.

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -77,10 +77,9 @@ module std.zip;
  */
 class ZipException : Exception
 {
-    this(string msg) @safe
-    {
-        super("ZipException: " ~ msg);
-    }
+    import std.exception : basicExceptionCtors;
+    ///
+    mixin basicExceptionCtors;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


std/zip.d
```diff
@@ -410,19 +412,26 @@ final class ZipArchive
      *
      * Returns: array representing the entire archive.
      */
-    void[] build()
+    void[] build() @safe pure
```
JackStouffer: The question here is whether anything that returns void[] can be safe in reality.

jmdavis: I don't know, but I think so. As long as operating on `void[]` gets flagged as `@system`, I think that marking this function as `@safe` so long as everything it's doing is `@safe` should be fine. The code that uses the `void[]` then has to worry about being `@trusted` appropriately, but this will have been vetted already, so they won't have to worry about it.

JackStouffer: Since stdx.allocator was recently changed to no longer return `void[]`, I wonder if it would be better to have a `@safe` overload which returned `ubyte[]` and a `@system` overload which casted the result of the first overload to `void[]`. That way everything is backwards compatible but you side-step some potential holes down the road.

aG0aep6G: > I wonder if it would be better to have a @safe overload which returned ubyte[] and a @system overload which casted the result of the first overload to void[].

The overload wouldn't have to be `@system`, and you wouldn't need a cast. Every array converts implicitly and `@safe`ly to `void[]`.

jmdavis: In general, I would argue that `ubyte[]` is better than `void[]` when you're dealing with raw file data and the like, but changing the return value would break code, and even if returning `void[]` were considered `@system` (and I don't think that it is), you can't overload on `@safety` attributes.

n8sh: >In general, I would argue that `ubyte[]` is better than `void[]` when you're dealing with raw file data and the like

I agree 100%.

>changing the return value would break code,

Do you have any examples of code that would break? `ubyte[]` implicitly converts to `void[]` so `void[] whatever = ...` will work fine, and if they were casting it to something else that will work just as well with `ubyte[]` as with `void[]`.

jmdavis: The issue is if `auto` is used, because then the variable will end up being typed as `ubyte[]` instead of `void[]`, and any code that expects `void[]` risks breaking. The easiest examples that I can think of the moment would be if the result were passed to a template whose template constraint expected `void[]` and didn't work with `ubyte[]` or if a function were overloaded on `ubyte[]` and `void[]`. Neither is particularly likely, but they're both possible. Given that `ubyte[]` implictly converts to `void[]` and that you can't really operate on `void[]` without casting it, a lot of the typical places where code would break won't, but they're still not the same, so I'd be very nervous about just swapping out the type. I expect that a lot of code would just work, but we can't guarantee that it all will.

JackStouffer:
Ping @jmdavis @CyberShadow

std/zip.d
```diff
@@ -410,19 +412,26 @@ final class ZipArchive
...
+
         uint i;
         uint directoryOffset;
 
         if (comment.length > 0xFFFF)
             throw new ZipException("archive comment longer than 65535");
 
+        static auto vals(T)(T a) @trusted
+        {
+            return a.values().sort!((x, y) => x.index < y.index).release;
```
n8sh: Here is the only use of `@trusted` in the PR. Why is this sorting not considered `safe` by the compiler, and why is it known to humans to be memory-safe?

JackStouffer:
@n8sh Fixed. The real issue is the AA methods in Druntime not being marked `@safe` when they are practically.

wilzbach:
> @n8sh Fixed. The real issue is the AA methods in Druntime not being marked @safe when they are practically.

That only applies for `.keys` and `.values` - the rest like `.byValue` is @safe since 2.078:

https://run.dlang.io/is/7nbeXd

(the respective Bugzilla issue https://issues.dlang.org/show_bug.cgi?id=14439)

std/zip.d
```diff
@@ -410,19 +412,26 @@ final class ZipArchive
...
+        static auto vals(T)(T a) @trusted
+        {
+            return a.values();
+        }
+
         // Compress each member; compute size
         uint archiveSize = 0;
         uint directorySize = 0;
-        auto directory = _directory.values().sort!((x, y) => x.index < y.index).release;
+        auto directory = vals(_directory).sort!((x, y) => x.index < y.index).release;
```
wilzbach: `byValue.array` is @safe - though in this case a ScopedAppender would be cool to have.

JackStouffer:
@wilzbach @n8sh Fixed

PR #6255, opened 2018-03-09
Fix Issue 18166 - std.array.replace should be usable in @safe for dst…

JackStouffer:
…rings

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -2268,7 +2268,14 @@ if (isInputRange!Range &&
             copy(stuff, retval[from .. from + stuff.length]);
 
         retval[from + stuff.length .. $] = subject[to .. $];
-        return cast(T[]) retval;
+        static if (is(T == const) || is(T == immutable))
+        {
+            return () @trusted { return cast(T[]) retval; } ();
+        }
+        else
+        {
+            return cast(T[]) retval;
+        }
     }
     else
     {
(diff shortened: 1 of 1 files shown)


std/array.d
```diff
@@ -2267,7 +2267,16 @@ if (isInputRange!Range &&
             copy(stuff, retval[from .. from + stuff.length]);
 
         retval[from + stuff.length .. $] = subject[to .. $];
-        return cast(T[]) retval;
+        static if (is(T == const) || is(T == immutable))
+        {
+            auto ret = () @trusted { return cast(T[]) retval; } ();
+            retval = null;
+            return ret;
```
wilzbach: Why not simply `return ()...`

JackStouffer: I was copying the behavior of `assumeUnique`

wilzbach: Well, `assumeUnique` is very old. It's in a9bc53273c7930c9b2cf442e3d8bc4e28d65ee4b from 2007 since git tracking started and presumably a lot older and I don't see any reason why this is necessary in D2

JackStouffer:
Fixed

std/array.d
```diff
@@ -2268,7 +2268,14 @@ if (isInputRange!Range &&
             copy(stuff, retval[from .. from + stuff.length]);
 
         retval[from + stuff.length .. $] = subject[to .. $];
-        return cast(T[]) retval;
+        static if (is(T == const) || is(T == immutable))
```
n8sh: Should this also check for `is(T == shared)`?

JackStouffer: I'm not sure. I'm hesitant to do so because who really knows what it means to cast a non-shared array to a shared array (the language spec sure doesn't). Line 2264 should probably be allocated differently to be shared at conception in that case.

PR #6257, opened 2018-03-10
Fix issue 18559

thewilsonator:
Still need to do atan, atan2: help much appreciated

DMD part: https://github.com/dlang/dmd/pull/7995
Druntime part: https://github.com/dlang/druntime/pull/2135

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -703,12 +703,10 @@ deprecated
  */
 
 real cos(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.cos(x); }
-//FIXME
 ///ditto
-double cos(double x) @safe pure nothrow @nogc { return cos(cast(real) x); }
-//FIXME
+double cos(double x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.cos(x); }
 ///ditto
-float cos(float x) @safe pure nothrow @nogc { return cos(cast(real) x); }
+float cos(float x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.cos(x); }
 
 @safe unittest
 {
@@ -737,12 +735,10 @@ float cos(float x) @safe pure nothrow @nogc { return cos(cast(real) x); }
  */
 
 real sin(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.sin(x); }
-//FIXME
 ///ditto
-double sin(double x) @safe pure nothrow @nogc { return sin(cast(real) x); }
-//FIXME
+double sin(double x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.sin(x); }
 ///ditto
-float sin(float x) @safe pure nothrow @nogc { return sin(cast(real) x); }
+float sin(float x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.sin(x); }
 
 ///
 @safe unittest
@@ -1099,10 +1095,16 @@ real acos(real x) @safe pure nothrow @nogc
 }
 
 /// ditto
-double acos(double x) @safe pure nothrow @nogc { return acos(cast(real) x); }
+double acos(double x) @safe pure nothrow @nogc
+{
+    return atan2(sqrt(1-x*x), x);
+}
 
 /// ditto
-float acos(float x) @safe pure nothrow @nogc  { return acos(cast(real) x); }
+float acos(float x) @safe pure nothrow @nogc
+{
+    return atan2(sqrt(1-x*x), x);
+}
 
 @system unittest
 {
@@ -1126,10 +1128,16 @@ real asin(real x) @safe pure nothrow @nogc
 }
 
 /// ditto
-double asin(double x) @safe pure nothrow @nogc { return asin(cast(real) x); }
+double asin(double x) @safe pure nothrow @nogc
+{
+    return atan2(x, sqrt(1-x*x));
+}
 
 /// ditto
-float asin(float x) @safe pure nothrow @nogc  { return asin(cast(real) x); }
+float asin(float x) @safe pure nothrow @nogc
+{
+    return atan2(x, sqrt(1-x*x));
+}
 
 @system unittest
 {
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -703,12 +703,10 @@ deprecated
...
 
 real cos(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.cos(x); }
-//FIXME
 ///ditto
-double cos(double x) @safe pure nothrow @nogc { return cos(cast(real) x); }
-//FIXME
+double cos(double x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.cos(x); }
 ///ditto
-float cos(float x) @safe pure nothrow @nogc { return cos(cast(real) x); }
+float cos(float x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.cos(x); }
```
JackStouffer: You haven't actually changed anything yet. `core.math.cos(x)` is defined as

```d
real cos(real x) @safe pure nothrow;
```

So the argument is still being converted to `real`. Same throughout.

thewilsonator: see dlang/druntime#2135

n8sh: Is there a reason `public import core.math : cos` wouldn't be better?

thewilsonator: Good point. I'll wait for the DMD/Druntime prs first though, since this needs them.

n8sh:
atan is straightforward:

https://run.dlang.io/is/PEy6sm <-- proof that this compiles
```d
float atan(float x) @nogc nothrow pure @safe
{
    import cmath = core.stdc.math;
    static if (__traits(compiles, cmath.atanf(x))) // Added in C99.
        return cmath.atanf(x);
    else
        return cmath.atan(x);
}

double atan(double x) @nogc nothrow pure @safe
{
    import cmath = core.stdc.math;
    return cmath.atan(x);
}
```

kinke:
> atan is straightforward:

Just forwarding to the C function firstly prevents CTFE and secondly makes the result depend on the C/platform implementation.

n8sh:
>Just forwarding to the C function firstly prevents CTFE

CTFE doesn't work for atan in the first place. https://run.dlang.io/is/TMxC2K
```d
void main()
{
    import std.math: atan;
    enum test = atan(12345.0L); // Doesn't compile.
}
```

>secondly makes the result depend on the C/platform implementation

If it is okay for C why not for D? Does the documentation somewhere promise identical cross-platform floating point results?

kinke:
> CTFE doesn't work for atan in the first place.

That's not a good argument. Many/most math functions are CTFE-able for LDC (incl. atan() except for non-Windows x86 architectures).

> Does the documentation somewhere promise identical cross-platform floating point results?

I don't think so (but think it should :)), although some unittests are very strict. E.g., that's the reason LDC can't use some LLVM intrinsics (pow() IIRC), as the various C implementations differ quite a bit.

n8sh:
>(incl. atan() except for non-Windows x86 architectures)

To be clear this means it doesn't work on `version(X86)` and it also doesn't work on `version(X86_64)`, right?

kinke:
Almost; it works for x86(_64) Windows MSVC targets too (where LDC doesn't use X87 `real`s). See https://github.com/ldc-developers/phobos/blob/ldc/std/math.d#L1156; if atan2() used its inline-asm version only for non-CTFE, it would work on all platforms.

n8sh:
>I don't think so (but think it should :)), although some unittests are very strict. E.g., that's the reason LDC can't use some LLVM intrinsics (pow() or exp() IIRC), as the various C implementations differ quite a bit.

Someone evaluating D for scientific computing asking "Why is D so much slower than C++ at trigonometry?" and being told "std.math is slow, you have to use core.stdc.math" was the motivation for this pull request. If bit-for-bit identical floating point behavior on different platforms comes at the price of performance it probably should be opt-in. Maybe have separate modules for fast math and portable math.

kinke:
AFAIK that stems from central algos like atan[2]() here always using the old FPU on x86(_64), even when fed with doubles and floats. So using the Cephes algos in Phobos for single- and double-precision ought to get rid of these performance issues (I've been disliking myself for too long). As all of these algos already support double-precision reals, exposing a double version is simply accomplished by generalizing the existing function taking a `real` (in various precisions due to different platforms) to a template. Then adding the single-precision constants enables the exposure of a further `atan(float)` overload.

If I find some time, I may have a go at this later this week.

kinke:
@klickverbot: The existing P/Q constants for both double and quadruple precision don't match the ones [here](https://github.com/jeremybarnes/cephes/blob/master/128bit/atanll.c). Where did you get them from?

kinke:
See https://github.com/dlang/phobos/pull/6272 for a start with atan().

quickfur:
In principle, we need to merge this. Barring, of course, implementation details.  Basically, if user code wants to use float/double, we should compute the result using the fastest way possible (e.g., on x86_64, use SSE instructions instead of x87), rather than promote to real, do a slow computation, only to truncate the result afterwards.  Of course, the primary implementation will be in the dmd PR, but druntime/phobos should be such that we compute at the precision level asked for by the user without implicit slowdowns.  If extra accuracy is desired, we already support passing in `real` and getting the result as `real`.

std/math.d
```diff
@@ -1382,10 +1390,18 @@ real cosh(real x) @safe pure nothrow @nogc
...
+    const double y = exp(x);
+    return (y + 1.0/y) * 0.5;
+}
 
 /// ditto
-float cosh(float x) @safe pure nothrow @nogc  { return cosh(cast(real) x); }
+float cosh(float x) @safe pure nothrow @nogc
+{
+    const float y = exp(x);
+    return (y + 1.0/y) * 0.5;
```
kinke: Use the `f` suffix for the literals, otherwise it'll be computed in double precision before being truncated to float.

std/math.d
```diff
@@ -1417,10 +1433,28 @@ real sinh(real x) @safe pure nothrow @nogc
 }
 
 /// ditto
-double sinh(double x) @safe pure nothrow @nogc { return sinh(cast(real) x); }
+double sinh(double x) @safe pure nothrow @nogc
+{
+    if (fabs(x) > double.mant_dig * LN2)
```
kinke: `LN2` is of type `real`, so better cast to double to prevent the comparison being performed in `real` precision.

std/math.d
```diff
@@ -1417,10 +1433,28 @@ real sinh(real x) @safe pure nothrow @nogc
...
-float sinh(float x) @safe pure nothrow @nogc  { return sinh(cast(real) x); }
+float sinh(float x) @safe pure nothrow @nogc
+{
+    if (fabs(x) > float.mant_dig * LN2)
+    {
+        return copysign(0.5 * exp(fabs(x)), x);
+    }
+    
+    const float y = expm1(x);
+    return 0.5 * y / (y+1) * (y+2);
```
kinke: Ditto for `LN2` and literals.

thewilsonator:
superseded by https://github.com/dlang/phobos/pull/7463

PR #6258, opened 2018-03-10
Add website link to author name for Jonathan M Davis.

jmdavis:


jmdavis:
> You may want to enable SSL on your website and link to the https:// URI, since browser vendors seem to be intent on deprecating unencrypted HTTP.

Getting HTTPS set up is on my todo list, but even if I make it HTTPS-only, it'll need to redirect from HTTP to HTTPS, so having the link here be HTTP shouldn't be a problem. Either way, I need to find time to work on getting HTTPS sorted out before it's going to work.

PR #6259, opened 2018-03-11
Mark IeeeFlags and FloatingPointControl as @trusted nothrow @nogc

ibuclaw:
Have added `pure` onto the flag getters, and removed `pure` from the flag setters (for now...).

Granted that `std.math.exp` is pure _and still_ raises an FP exception gives us at least an argument for adding pure to these structures also (or removing the `pure`  attribute from this and others functions, breaking everyone's code).

As well, a new private `FloatingPointControl.raiseException` method should be added too.  As this hack can now be defeated by the optimizer, however this wasn't the case in the past.
https://github.com/dlang/phobos/blob/fd9fdeb25d3f497bf68ba5ac29667adb3c912ddb/std/math.d#L1833-L1850

Edit: Although now I'm swayed in the direction that the above should just be removed and unittests updated.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -4921,6 +4921,8 @@ real remquo(real x, real y, out int n) @trusted nothrow @nogc  /// ditto
  */
 struct IeeeFlags
 {
+@trusted nothrow @nogc:
+
 private:
     // The x87 FPU status register is 16 bits.
     // The Pentium SSE2 status register is 32 bits.
@@ -4959,7 +4961,7 @@ private:
     }
 
 private:
-    static uint getIeeeFlags()
+    static uint getIeeeFlags() pure
     {
         version(InlineAsm_X86_Any)
         {
@@ -4991,11 +4993,12 @@ private:
         else
             assert(0, "Not yet supported");
     }
-    static void resetIeeeFlags() @nogc
+
+    static void resetIeeeFlags()
     {
         version(InlineAsm_X86_Any)
         {
-            asm pure nothrow @nogc
+            asm nothrow @nogc
             {
                 fnclex;
             }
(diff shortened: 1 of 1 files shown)


JackStouffer:
Pinging qualified reviewers @kinke @klickverbot

std/math.d
```diff
@@ -5225,7 +5233,7 @@ public:
     }
 
     /// Returns: the exceptions which are currently enabled (unmasked)
-    @property static ExceptionMask enabledExceptions() @nogc
+    @property static ExceptionMask enabledExceptions()
```
kinke: Should be pure as well.

std/math.d
```diff
@@ -5054,14 +5062,14 @@ struct FloatingPointControl
...
-    @property void rounding(RoundingMode newMode) @nogc
+    @property void rounding(RoundingMode newMode)
     {
         initialize();
         setControlState((getControlState() & (-1 - roundingMask)) | (newMode & roundingMask));
     }
 
     /// Returns: the currently active rounding mode
-    @property static RoundingMode rounding() @nogc
+    @property static RoundingMode rounding()
```
kinke: pure

std/math.d
```diff
@@ -4950,10 +4953,13 @@ else version(ARM)
...
 /// Set all of the floating-point status flags to false.
-void resetIeeeFlags() @nogc { IeeeFlags.resetIeeeFlags(); }
+void resetIeeeFlags() @trusted nothrow @nogc
+{
+    IeeeFlags.resetIeeeFlags();
+}
 
 /// Returns: snapshot of the current state of the floating-point status flags
-@property IeeeFlags ieeeFlags()
+@property IeeeFlags ieeeFlags() @trusted nothrow @nogc
```
kinke: pure; possibly `@safe` (same for `resetIeeeFlags()` above)

kinke:
LGTM otherwise, thanks.

std/math.d
```diff
@@ -4789,11 +4791,12 @@ private:
         else
             assert(0, "Not yet supported");
     }
-    static void resetIeeeFlags() @nogc
+
+    static void resetIeeeFlags()
```
n8sh: Should be `pure` per the specification:
https://dlang.org/spec/function.html#pure-functions
>8. As a concession to practicality, a pure function can also:
> * read **and write** the floating point exception flags
> ...

std/math.d
```diff
@@ -5280,18 +5288,18 @@ private:
     }
 
     // Clear all pending exceptions
-    static void clearExceptions() @nogc
+    static void clearExceptions()
```
n8sh: `pure`

ibuclaw:
Rebased and added pure onto more getters.  As you can see from @n8sh's comments, not the setters just yet.

PR #6261, opened 2018-03-11
std.traits.FieldNameTuple: clarify documentation and add testing for classes

JohanEngelen:
For me, the behavior for inherited fields was unexpected; otherwise this just adds tests that should have been there already.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -2527,8 +2527,10 @@ private enum NameOf(alias T) = T.stringof;
  * Get as an expression tuple the names of the fields of a struct, class, or
  * union. This consists of the fields that take up memory space, excluding the
  * hidden fields like the virtual function table pointer or a context pointer
- * for nested types. If $(D T) isn't a struct, class, or union returns an
- * expression tuple with an empty string.
+ * for nested types.
+ * Inherited fields (for classes) are not included.
+ * If $(D T) isn't a struct, class, or union, an
+ * expression tuple with an empty string is returned.
  */
 template FieldNameTuple(T)
 {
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -2560,6 +2562,15 @@ template FieldNameTuple(T)
     static struct StaticStruct2 { int a, b; }
     static assert(FieldNameTuple!StaticStruct2 == AliasSeq!("a", "b"));
 
+    static class StaticClass1 { }
+    static assert(is(FieldNameTuple!StaticClass1 == AliasSeq!()));
```
wilzbach: This test already exists four lines above?

std/traits.d
```diff
@@ -2527,8 +2527,10 @@ private enum NameOf(alias T) = T.stringof;
  * Get as an expression tuple the names of the fields of a struct, class, or
  * union. This consists of the fields that take up memory space, excluding the
  * hidden fields like the virtual function table pointer or a context pointer
- * for nested types. If $(D T) isn't a struct, class, or union returns an
- * expression tuple with an empty string.
+ * for nested types.
+ * Inherited fields (for classes) are not included.
+ * If $(D T) isn't a struct, class, or union, an
+ * expression tuple with an empty string is returned.
```
wilzbach: This is a _really bad_ behavior for this case, but I guess there's nothing we can do about this anymore.

std/traits.d
JohanEngelen: `class` vs. `struct`, so it's a new test.

std/traits.d
JohanEngelen: Yep, more exhaustive testing would have uncovered that. Already the next item, RepresentationTypeTuple, is also not testing inherited classes, or static classes, ...

PR #6264, opened 2018-03-11
Revert "Temporarily disable curl tests for FreeBSD 32-bit"

wilzbach:
Reverts dlang/phobos#6232

Let's see whether https://github.com/dlang/phobos/pull/6234 fixed the issue for FreeBSD.


wilzbach:
Nope, doesn't fix the problem. Sorry for the noise.

PR #6268, opened 2018-03-12
Cache and reuse std.regex engine to speed repeated matching

DmitryOlshansky:
Prime use case is line by line matching.

Some anecdotal evidence on @jondegenhardt benchmarks shows:

```
# Before
▶ ./find_regex '(aa)+(cc)+g' strings.txt               
106437 matches; 5538.671 milliseconds

# After
▶ ./find_regex '(aa)+(cc)+g' strings.txt
106437 matches; 1880.550 milliseconds
```
For roughly ~3x speedup.

See also:
https://issues.dlang.org/show_bug.cgi?id=18114

### DIFF
--- a/std/regex/internal/backtracking.d
+++ b/std/regex/internal/backtracking.d
@@ -172,6 +172,15 @@ final:
         backtracking.initExternalMemory(memBlock);
     }
 
+    override Matcher!Char rearm(in Char[] data)
+    {
+        merge[] = Trace.init;
+        exhausted = false;
+        s = Stream(data);
+        next();
+        return this;
+    }
+
     this(ref const RegEx program, Stream stream, void[] memBlock, dchar ch, DataIndex idx)
     {
         _refCount = 1;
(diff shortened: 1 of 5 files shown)


jondegenhardt:
Added [Regex performance enhancement for repeated matchFirst calls](https://issues.dlang.org/show_bug.cgi?id=18600) for this.

DmitryOlshansky:
Ready to roll!

jondegenhardt:
The final performance fix was included in LDC 1.10.0-beta1. For this release standard benchmark I used for the TSV Utilities improved as follows:
```
LDC 1.7.0 (before regression):  8.37 seconds
LDC 1.8.0 (after regression):  10.01 seconds
LDC 1.9.0 (first fixes):        9.44 seconds
LDC 1.10.0-beta1 (this PR):     5.85 seconds
```
First fixes: Phobos PR 5981, DMD PR 7599

The benchmark test used reads a TSV file line-by-line and checks individual fields for regex matches. A significant amount of processing time is IO, so the percentage gain on the regex portion is higher than the overall gain. The overall gain from LDC 1.7.0 is 30%.

Test was run on MacOS, MacMini with 16GB RAM, SSD drives. The file used was
2.7GB, 14 million lines. Test info can be found here:
https://github.com/eBay/tsv-utils-dlang/blob/master/docs/ComparativeBenchmarks2018.md

Great result!

PR #6269, opened 2018-03-12
Fix CircleCI

JackStouffer:


wilzbach:
Looks like the CI result from https://github.com/dlang/phobos/pull/6156 was outdated :/

PR #6271, opened 2018-03-12

wilzbach:


wilzbach:
FYI: I just run this script - https://github.com/wilzbach/git-tools/blob/master/merge_stable

PR #6272, opened 2018-03-12
Implement std.math.{atan[2],tan,exp[2],expm1} for single- and double-precision

kinke:
And make the x87 `real` version CTFE-able.

Based on Cephes like the existing quadruple and x87 implementations,
https://github.com/jeremybarnes/cephes/blob/master/cmath/atan.c &
https://github.com/jeremybarnes/cephes/blob/master/single/atanf.c

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -1123,10 +1123,11 @@ private T tanImpl(T)(T x) @safe pure nothrow @nogc
     {
         T y = floor(x / cast(T) PI_4);
         // Strip high bits of integer part.
-        enum numHighBits = (realFormat == RealFormat.ieeeDouble ? 3 : 4);
-        T z = ldexp(y, -numHighBits);
+        enum T highBitsFactor = (realFormat == RealFormat.ieeeDouble ? 0x1p3 : 0x1p4);
+        enum T highBitsInv = 1.0 / highBitsFactor;
+        T z = y * highBitsInv;
         // Compute y - 2^numHighBits * (y / 2^numHighBits).
-        z = y - ldexp(floor(z), numHighBits);
+        z = y - highBitsFactor * floor(z);
 
         // Integer and fraction part modulo one octant.
         int j = cast(int)(z);
@@ -2344,7 +2345,7 @@ private T expImpl(T)(T x) @safe pure nothrow @nogc
         xx = x * x;
         const T px = x * poly(xx, P);
         x = px / (poly(xx, Q) - px);
-        x = (cast(T) 1.0) + ldexp(x, 1);
+        x = (cast(T) 1.0) + (cast(T) 2.0) * x;
     }
 
     // Scale by power of 2.
@@ -3173,7 +3174,7 @@ private T exp2Impl(T)(T x) @nogc @safe pure nothrow
         const T xx = x * x;
         const T px = x * poly(xx, P);
         x = px / (poly(xx, Q) - px);
-        x = (cast(T) 1.0) + ldexp(x, 1);
+        x = (cast(T) 1.0) + (cast(T) 2.0) * x;
     }
 
     // Scale by power of 2.
--- a/std/math.d
+++ b/std/math.d
@@ -2498,7 +2498,29 @@ private T expImpl(T)(T x) @safe pure nothrow @nogc
  *    $(TR $(TD $(NAN))        $(TD $(NAN))       )
  *  )
  */
-real expm1(real x) @trusted pure nothrow @nogc
+real expm1(real x) @trusted pure nothrow @nogc // TODO: @safe
+{
+    version(InlineAsm_X86_Any)
+    {
+        if (!__ctfe)
+            return expm1Asm(x);
+    }
+    return expm1Impl(x);
+}
+
+/// ditto
+double expm1(double x) @safe pure nothrow @nogc { return expm1Impl(x); }
+
+///
+@safe unittest
+{
+    assert(isIdentical(expm1(0.0), 0.0));
+    assert(expm1(1.0).feqrel(1.71828) > 16);
+    assert(expm1(2.0).feqrel(6.3890) > 16);
+}
+
+version(InlineAsm_X86_Any)
+private real expm1Asm(real x) @trusted pure nothrow @nogc
 {
     version(D_InlineAsm_X86)
     {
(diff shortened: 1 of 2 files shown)


kinke:
There are unittest failures due to `hypot()` still only supporting reals etc. So I guess more or less all remaining std.math functions need to get float/double support in order to hopefully pass the druntime/Phobos unittests again.

I'll need an okay to proceed though, I don't want to waste any more time on this if this doesn't have a chance to be merged (backwards compatibility concerns).

kinke:
`compilable/ctfe_math.d` fails because DMD would need to be patched to recognize the `std.math.tan(float|double)` overloads as CTFE builtins in `builtin.d`. The non-asm version still isn't fully CTFE-able because primitives `copysign()` and `signbit()` aren't.

No unittest failures anymore [well, test coverage is minimal at the moment].

x86 timings are quite interesting. Linux, Intel i5-3550 @4GHz:
```D
static import std.math;
import std.stdio;

T atan(T)() { return std.math.atan(cast(T) 0.43685); }
T atan2(T)() { return std.math.atan2(cast(T) 0.43685, cast(T) 0.06912); }
T tan(T)() { return std.math.tan(cast(T) 0.43685); }
T exp(T)() { return std.math.exp(cast(T) 0.43685); }
T exp2(T)() { return std.math.exp2(cast(T) 0.43685); }
T expm1(T)() { return std.math.expm1(cast(T) 0.43685); }

enum N = 10_000_000;

void bench(alias F)()
{
    writeln(".: ", F.stringof);
    const durations = benchmark!(F!real, F!double, F!float)(N);
    writeln("real:   ", durations[0]);
    writeln("double: ", durations[1]);
    writeln("float:  ", durations[2]);
}

void main()
{
    import std.stdio;

    bench!(atan);
    bench!(atan2);
    bench!(tan);
    bench!(exp);
    bench!(exp2);

    writeln(".: ", expm1.stringof); // no float version
    const durations = benchmark!(expm1!real, expm1!double)(N);
    writeln("real:   ", durations[0]);
    writeln("double: ", durations[1]);
}
```

=>

```
.: atan(T)()
real:   430 ms, 58 μs, and 7 hnsecs
double: 268 ms, 79 μs, and 3 hnsecs
float:  113 ms, 737 μs, and 4 hnsecs
.: atan2(T)()
real:   427 ms, 635 μs, and 3 hnsecs
double: 480 ms, 229 μs, and 6 hnsecs
float:  266 ms, 752 μs, and 3 hnsecs
.: tan(T)()
real:   351 ms, 427 μs, and 5 hnsecs
double: 909 ms, 280 μs, and 5 hnsecs
float:  220 ms, 476 μs, and 4 hnsecs
.: exp(T)()
real:   223 ms and 639 μs
double: 544 ms, 363 μs, and 5 hnsecs
float:  390 ms, 93 μs, and 9 hnsecs
.: exp2(T)()
real:   226 ms, 165 μs, and 2 hnsecs
double: 619 ms, 915 μs, and 6 hnsecs
float:  513 ms, 462 μs, and 2 hnsecs
.: expm1(T)()
real:   234 ms, 788 μs, and 2 hnsecs
double: 197 ms, 129 μs, and 8 hnsecs
```

The `real` timings correspond to the DMD 2.079.0 timings for all 3 floating-point types. So from ~4x speed-up to ~2.6x slow-down. ;)

kinke:
Well I for one didn't expect such inconsistent results, or a ~4x slow-down of software-double-tan vs. software-float-tan etc.

kinke:
Here are the corresponding `core.stdc.tgmath` timings (Ubuntu 16.04):
```
.: atan(T)()
real:   403 ms, 803 μs, and 5 hnsecs
double: 685 ms, 848 μs, and 3 hnsecs
float:  58 ms, 113 μs, and 3 hnsecs
.: atan2(T)()
real:   510 ms, 110 μs, and 6 hnsecs
double: 903 ms, 576 μs, and 4 hnsecs
float:  165 ms and 114 μs
.: tan(T)()
real:   193 ms, 504 μs, and 1 hnsec
double: 716 ms, 779 μs, and 5 hnsecs
float:  83 ms and 817 μs
.: exp(T)()
real:   511 ms, 916 μs, and 1 hnsec
double: 690 ms, 991 μs, and 3 hnsecs
float:  87 ms, 717 μs, and 6 hnsecs
.: exp2(T)()
real:   547 ms, 701 μs, and 2 hnsecs
double: 129 ms, 991 μs, and 6 hnsecs
float:  123 ms and 33 μs
.: expm1(T)()
real:   471 ms and 943 μs
double: 110 ms, 476 μs, and 3 hnsecs
```

kinke:
Rough runtimes comparison for the single/double precision overloads:

benchmark

wilzbach:
Regarding the project tester failure:

```
core.exception.AssertError@source/mir/random/flex/internal/area.d(525): unittest failure
----------------
??:? _d_unittestp [0x617ef1]
source/mir/random/flex/internal/area.d:525 @safe void mir.random.flex.internal.area.__unittest_L441_C32() [0x5a65a1]
??:? void mir.random.flex.internal.area.__modtest() [0x5b2f9f]
??:? int core.runtime.runModuleUnitTests().__foreachbody2(object.ModuleInfo*) [0x6308e3]
??:? int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)).__lambda2(immutable(object.ModuleInfo*)) [0x616622]
??:? int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*))).__foreachbody2(ref rt.sections_elf_shared.DSO) [0x61dde1]
??:? int rt.sections_elf_shared.DSO.opApply(scope int delegate(ref rt.sections_elf_shared.DSO)) [0x61df74]
??:? int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*))) [0x61dd6d]
??:? int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)) [0x6165f9]
??:? runModuleUnitTests [0x6306b9]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll() [0x619e84]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x619e0b]
??:? _d_run_main [0x619d76]
??:? main [0x560021]
??:? __libc_start_main [0xe29d882f]
core.exception.AssertError@source/mir/random/flex/package.d(916): unittest failure
----------------
??:? _d_unittestp [0x617ef1]
source/mir/random/flex/package.d:916 void mir.random.flex.__unittest_L879_C26() [0x5c9871]
```

that's because the flex algorithm is quite sensitive to the floating point behavior and the tests use a special `fpEqual` function:

https://github.com/libmir/mir-random/blob/master/source/mir/random/flex/internal/area.d

kinke:
> that's because the flex algorithm is quite sensitive to the floating point behavior and the tests use a special fpEqual function

... which (only) for DMD and x86_64 is defined as exact equality (== operator). The failing test there seems to depend on `exp()`, which for DMD x86(_64) is defined as `exp2Asm(LOG2E*x)`, so I'd argue that exact equality is too exaggerated and `approxEqual()` a better fit, as used for all other compilers/platforms.
Thanks for checking. :)

kinke:
Statically unrolled `poly()` makes the Cephes version of `atan(double)` faster by ~31%, and `expm1(double)` by 19%; the double versions of other functions only show marginal improvements.

kinke:
I'm afraid I don't have any LDC timings. This patch cannot be simply cherry-picked due to existing mods in the touched parts; the main problem though is that the timings wouldn't be easily comparable, as e.g. the currently real-only `core.math.ldexp()` primitive is implemented differently (intrinsic for DMD, i.e., inlined x87 asm; for LDC, a [stupidly non-inlined](https://github.com/ldc-developers/ldc/issues/2552) call to druntime which forwards to the C `long double` version of the C runtime).

kinke:
Using `core.stdc.math.ldexp[l|f]` in `std.math.ldexp()` (instead of the `core.math.ldexp(real)` intrinsic using a supposedly slow `fscale` instruction) leads to a significant performance increase, up to almost 2x faster, for `tan(double)` and `exp2(float|double)`, i.e., exactly where the Cephes version lacks significantly behind in the chart above.

kinke:
[Chart and benchmark code updated.]

kinke:
Chart updated once more after getting rid of a few `ldexp()` calls, which makes `tan(double)` twice as fast on my machine (!).

kinke:
So on my machine and using the purely arbitrary input data above:

* The new Cephes double versions are 0.4x - 2x faster than the current versions, with a geometric mean of 0.88 (0.97 with C `ldexp`), i.e., overall slower than the current versions using the FPU, almost exclusively due to great performance of Phobos x87 inline asm for `std.math.exp2(real)` (twice as fast as the `long double` version in the C runtime!). Cephes almost always clearly beats the C runtime, except for `exp2(double)` (losing a lot there) and `expm1(double)`.
* The float versions are 0.5x - 3.8x faster than the current versions, with a geometric mean of 1.2 (1.4 with C ldexp). The C runtime is always and by far the winner though (possibly due to less handled special cases? - up to 12x as fast as the C double version!).

kinke:
I ported this to LDC, and the results are way better than expected: overall speed-up factor of >3 for the new double/float versions compared to the old ones, only beaten by the C runtime in 2 cases (`exp[2](float)`) and otherwise leading the pack. Here's the extended chart:

benchmark incl. LDC port

std/math.d
```diff
@@ -7384,6 +7649,20 @@ do
     }
 }
 
+/// ditto
+Unqual!(CommonType!(T1, T2)) poly(T1, T2, int N)(T1 x, ref const T2[N] A) @safe pure nothrow @nogc
+if (isFloatingPoint!T1 && isFloatingPoint!T2 && N > 0 && N <= 10)
+{
+    // statically unrolled version for up to 10 coefficients
```
ibuclaw: Why up to 10? Is that a certain sweet spot between compile time and runtime speed? Or just arbitrary.

std/math.d
```diff
@@ -1150,106 +1204,180 @@ real atan(real x) @safe pure nothrow @nogc
 {
     version(InlineAsm_X86_Any)
     {
-        return atan2(x, 1.0L);
+        if (!__ctfe)
+            return atan2Asm(x, 1.0L);
+    }
+    return atanImpl(x);
```
ibuclaw: Perhaps all functions could be split up this way between asm and impl. `floor`/`ceil` are two others where this could be applied too.

I forsee an opportunity to split std.math perhaps in the same way bigint is divided into generic and asm modules. Maybe even add ldc and gdc asm implementations also. ;-)

kinke:
@klickverbot: I posted and later updated the benchmark code [earlier](https://github.com/dlang/phobos/pull/6272#issuecomment-373879587). [Note that `tan(real)` uses the Cephes version, not upstream's x87 asm, that's why it's significantly slower than DMD there.]

std/math.d
kinke: Just an arbitrary guess wrt. sweet spot for code-size/inline-ability. I searched std.math for P and Q arrays, and the longest one had 9 coefficients. ;)

std/math.d
kinke: In the long run, I'd prefer std.math being split up into a package, with a dedicated module per function family (atan, exp, etc.). Editing that huge file would be much easier then.

TurkeyMan:
But... pow()?? :(

ibuclaw:
@TurkeyMan `pow()` is a template, and so there's already float and double implementations.

kinke:
Except that it internally [uses](https://github.com/dlang/phobos/blob/master/std/math.d#L6671) `real` precision. - This PR needs to be seen as a start, e.g., the whole `log()` family (log, log10, log2, log1p, logb` is still missing as well, as are the trivial one-liners currently only operating with `real`s. A simple `poly()` test indicates that normal arithmetic ops (at least multiply and add) for x87 `real` are slower than float/double by a factor of ~10, so in essence **all** functions in std.math should be extended to proper implementations in float/double, but that's quite a big task, and help is obviously appreciated, a `:(` is no help at all.

kinke:
@ibuclaw: unittests added.

std/math.d
```diff
@@ -1150,106 +1217,228 @@ real atan(real x) @safe pure nothrow @nogc
...
         // Rational form in x^^2.
-        const real z = x * x;
-        y = y + (poly(z, P) / poly(z, Q)) * z * x + x;
-
-        return (sign) ? -y : y;
+        const T z = x * x;
+        static if (realFormat == RealFormat.ieeeSingle)
+        {
+            y += ((( 8.05374449538e-2f * z
+                   - 1.38776856032E-1f) * z
```
ibuclaw: Mix between cases here

std/math.d
```diff
@@ -1150,106 +1217,228 @@ real atan(real x) @safe pure nothrow @nogc
...
+            2.2981886733594175366172E1L,
+            1.0000000000000000000000E0L,
+        ];
+    }
+    else static if (realFormat == RealFormat.ieeeDouble)
+    {
+        static immutable T[5] P = [
+            -6.485021904942025371773E1L,
+            -1.228866684490136173410E2L,
+            -7.500855792314704667340E1L,
```
ibuclaw: Upper E

std/math.d
```diff
@@ -1722,159 +1993,334 @@ auto sqrt(creal z) @nogc @safe pure nothrow
...
+
+    static if (F.realFormat == RealFormat.ieeeSingle)
+    {
+        xx = x * x;
+        x = ((((( 1.9875691500E-4f  * x
+                + 1.3981999507E-3f) * x
+                + 8.3334519073E-3f) * x
+                + 4.1665795894E-2f) * x
+                + 1.6666665459E-1f) * x
+                + 5.0000001201E-1f) * xx
```
ibuclaw: Although. Is this still needed when you have a static foreach poly implementation?

kinke: Not really I guess. I changed it accordingly; it's not like the D port has been a 1:1 copy of the C code anyway, so deviating some more is probably okay.

kinke:
[Btw this fixes the existing `real` implementations for double-precision reals, as used on most non-x86 architectures. Most implementations silently assumed the x87 constants and shared quadruple/extended-precision code path.]

kinke:
No changes to benchmark. I did check the optimized IR generated by LDC to make sure the calls are there inside a loop. I didn't use `-release`.

kinke:
@klickverbot: Thanks, looking forward to your results. I updated the LDC port => https://github.com/ldc-developers/ldc/pull/2624.

std/math.d
kinke: The DMD timings for the float-versions seem to suffer by about 10% after changing to `poly()`. I've seen similar slowdowns when adding a new variable (`const T xx = x*x`) instead of reusing an existing one (`z = x*x`) with DMD too though, so... ;)

kinke:
> Are we okay with merging this (after reviewing the diff) even if the performance results for DMD are a wash, given that it greatly improves the performance on LDC?

I think this should definitely go in, given that it also fixes the Cephes constants for 64-bit reals. For DMD and its suboptimal codegen, the x87 versions could still be preferred over the extended Cephes implementations, potentially even selectively where it pays off (i.e., `exp[2]()`). Performance numbers may need to be revised as Iain has changed the under/overflow behavior of `exp[2]()` in the meantime.

ibuclaw:
> Performance numbers may need to be revised as Iain has changed the Cephes under/overflow behavior of exp[2]() in the meantime.

Depending on how good ldc is at keeping track of data layout, either it would be a little faster, or no change at all.

kinke:
I was referring to DMD performance, it's clear that LDC shouldn't be affected with enabled optimizations.

kinke:
Rebased.

kinke:
Still TODO:

* CircleCI linter warnings (`no documented example`) - hopefully not for each new overload
* fix mir.random unittest for poject tester
* teach DMD about new overloads as additional CTFE builtins and adapt `test/compilable/test5227.d
` accordingly
* find a solution for missing `expm1(float)` (I didn't find an implementation in Cephes - maybe just use the double version)

wilzbach:
> CircleCI DScanner warnings (no documented example) - hopefully not for each new overload

Nope just for new public symbols.

```
std/math.d(924:6)[warn]: Public declaration 'tan' has no documented example.
std/math.d(1328:6)[warn]: Public declaration 'atan' has no documented example.
std/math.d(1575:6)[warn]: Public declaration 'atan2' has no documented example.
std/math.d(2188:6)[warn]: Public declaration 'exp' has no documented example.
std/math.d(2504:6)[warn]: Public declaration 'expm1' has no documented example.
std/math.d(2840:6)[warn]: Public declaration 'exp2' has no documented example.
```

Note that due to a DMD ddoc "unfeature" there can't be anything between the public symbols and public unittests. So probably just moving the private implementation blocks around will fix these errors.

kinke:
Remaining blocker: mir.random.

n8sh:
>Remaining blocker: mir.random.

@kinke I'm looking into this and I'm seeing a computation using 64-bit doubles whose result differs from the expected value in the second-to-last bit of the mantissa. The test *succeeds* when compiled with LDC2 with which it uses the variable-precision `llvm.exp.f#` intrinsic. Before I look closer, how sure are you that there isn't a rounding error in the lowest bit(s) of this implementation?

n8sh:
I assume it's more likely that the test in mir.random is wrong than this code is wrong, but if there actually is a mistake here then this unit test suite has really proven its value.

n8sh:
Ah, sorry for not reading the full comments before saying something that's already been covered... yeah changing the `Flex_fpEqual` seems to be the way to go.

n8sh:
>... which (only) for DMD and x86_64 is defined as exact equality (== operator). The failing test there seems to depend on `exp(double x)`, which for DMD x86(_64) is defined as `cast(double) exp2Asm(LOG2E * cast(real) x)`, so I'd argue that exact equality is too exaggerated and `approxEqual()` a better fit, as used for all other compilers/platforms.

BTW I checked and with LDC on X86_64 the tests still pass if `fpEqual` is defined as exact equality. Nevertheless I've created PR libmir/mir-random#85.

n8sh:
Time to squash and merge?

n8sh:
When you put it that way, perhaps not.

ibuclaw:
@WalterBright indeed, we touched the subject earlier here https://github.com/dlang/phobos/pull/6272#discussion_r175281673

kinke:
Rebased...

std/math.d
MartinNowak: Keeping similar things (e.g. docs) in files in sync would become more difficult and switching between files is quite an overhead. Might be a better idea to outsource some heavy implementations to `std/internal/math/` if module size is a concern.

MartinNowak:
@kinke We got the surprising bug report that `exp` no longer works at compile time.
https://forum.dlang.org/post/oocmwvbzxsekxykxavts@forum.dlang.org
Would be good to figure out what happened there and why this wasn't tested/found beforehand.

kinke:
I fixed that for LDC, as we do test it, but forgot to upstream it. If someone gets to it before I do, see https://github.com/ldc-developers/ldc/commit/95af69ef9c5d958c6df8c39b92ce46cc8a5375e8 (and add a test [here](https://github.com/dlang/dmd/blob/e4f19f3ea8a22bc8bb9cf13ff491d08f8dbc775d/test/compilable/test5227.d)).

PR #6274, opened 2018-03-13
Fix Issue 18178 - std.path should be usable in @safe

JackStouffer:


std/path.d
```diff
@@ -3956,23 +3956,31 @@ string expandTilde(string inputPath) nothrow
...
+
+            if (home == path)
+                return home;
 
             return combineCPathWithDPath(home, path, 1);
         }
 
         // Replaces the tilde from path with the path from the user database.
-        static string expandFromDatabase(string path) nothrow
+        static string expandFromDatabase(string path) nothrow @trusted
```
wilzbach: Hmmm that's quite a huge `@trusted` - can't we use a ScopedAllocator from stdx.allocator? This should make all the manual allocations safe.

std/path.d
```diff
@@ -3956,23 +3956,31 @@ string expandTilde(string inputPath) nothrow
...
 
             // Get HOME and use that to replace the tilde.
-            auto home = getenv("HOME");
-            if (home == null)
-                return path;
+            auto home = () @trusted {
+                auto homeStr = getenv("HOME");
+                if (homeStr == null)
+                    return path;
+                return homeStr[0 .. strlen(homeStr)].assumeUnique;
```
wilzbach: `assumeUnique` can be done outside of the @trusted lambda.

std/path.d
```diff
@@ -3947,7 +3947,7 @@ string expandTilde(string inputPath) nothrow
             string cp;
             if (char_pos < path.length)
                 // Append something from path
-                cp = cast(string)(c_path[0 .. end] ~ path[char_pos .. $]);
+                cp = text(c_path[0 .. end], path[char_pos .. $]);
```
wilzbach: IIRC `a ~ b` is faster than `text` because for the former the exact length is know and thus only one allocation happens whereas text uses an appender.
(I doubt that it matters here though)

std/path.d
JackStouffer: `assumeUnique` is `@system`

std/path.d
JackStouffer: `ScopedAllocator`, even with `GCAllocator` is also `@system`

JackStouffer:
Ping @wilzbach @CyberShadow

std/path.d
kyllingstad: I agree that this is far too much code to mark as `@trusted`.

I wonder if it's really necessary to use `malloc` here. How about using a static array buffer instead, and fall back to a GC-allocated array in the few cases where the username is uncommonly long?

kyllingstad: Ah, I only saw the `malloc` and didn't notice the `realloc` calls before now. But I actually think it's OK to use plain GC allocations for both `username` and `extra_memory`.

JackStouffer:
Reworked to use GC allocations.

JackStouffer:
Ping @kyllingstad @wilzbach

std/path.d
```diff
@@ -4017,24 +4014,27 @@ string expandTilde(string inputPath) nothrow
...
-                scope(exit) free(extra_memory);
+                char[] extra_memory;
 
                 passwd result;
                 while (1)
                 {
-                    extra_memory = cast(char*) realloc(extra_memory, extra_memory_size * char.sizeof);
-                    if (extra_memory == null)
-                        onOutOfMemoryError();
+                    extra_memory ~= new char[extra_memory_size];
```
kyllingstad: If you change this to

    extra_memory.length += extra_memory_size;

I think the array may be resized in place (i.e., without a complete reallocation) in some cases, whereas an append operation will always reallocate.

JackStouffer:
Fixed

PR #6275, opened 2018-03-13
std.math: fix feqrel for ibmExtended floats.

redstar:
`cast(double*)(&x)[MANTISSA_MSB]` is of type `double*`. Moving the
opening parenthesis gives the right result.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -7093,16 +7093,16 @@ if (isFloatingPoint!(X))
     alias F = floatTraits!(X);
     static if (F.realFormat == RealFormat.ibmExtended)
     {
-        if (cast(double*)(&x)[MANTISSA_MSB] == cast(double*)(&y)[MANTISSA_MSB])
+        if ((cast(double*)&x)[MANTISSA_MSB] == (cast(double*)&y)[MANTISSA_MSB])
         {
             return double.mant_dig
-            + feqrel(cast(double*)(&x)[MANTISSA_LSB],
-                    cast(double*)(&y)[MANTISSA_LSB]);
+            + feqrel((cast(double*)&x)[MANTISSA_LSB],
+                    (cast(double*)&y)[MANTISSA_LSB]);
         }
         else
         {
-            return feqrel(cast(double*)(&x)[MANTISSA_MSB],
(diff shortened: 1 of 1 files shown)


ibuclaw:
@redstar #5822 and #5823

ibuclaw:
@wilzbach I have a cross builder for ppc64le, haven't managed to get testsuite running yet, it would have to be in under qemu unless someone has hardware to run on. =)

redstar:
@wilzbach My development version of LDC uses `ibmExtended` right now. Nothing is committed yet and therefore not visible on the buildbot (http://buildbot.ldc-developers.org/waterfall).

ibuclaw:
Nice. I've disabled mine until the hosting provider gets more servers in. :-(

https://buildbot.dgnu.org/#/waterfall

wilzbach:
> Nice. I've disabled mine until the hosting provider gets more servers in. :-(

@ibuclaw FYI: if money is the bottleneck, Laeeth recently offered to sponsor a few ARM build servers for LDC:

> How much RAM is needed to compile LDC on ARM?    If more scaleway build servers would help, I can easily contribute up to 10 right now - in which case I would just sign up and give you or him the login details.

> https://www.scaleway.com/pricing/

> It doesn't make any direct difference to me commercially for the foreseeable future.  But I'd like to know D runs on Android and Linux on ARM.  Currently we use iOS devices for security reasons, but China is on Android so it would be nice to move to some kind of hardened Android (we are looking into Copperhead).  And maybe with the spec execution problems, future generations of ARM will become more appealing than Intel for servers.

I'm not sure whether this offer would apply to GDC or other architectures like PPC64 too, but otherwise the DLF might be able to help out monetary-wise too.

ibuclaw:
Cost isn't a problem. Scaleway have a lack of servers is the bottleneck. Also ping @andralex on slack and ask about D foundation funding for servers.

redstar:
@ibuclaw Is this the new buildbot ui? Updating from 0.8.x to 1.x is next on my agenda...

ibuclaw:
@redstar yes it is the new UI.

PR #6276, opened 2018-03-14
Remove a cast from std.process

wilzbach:


wilzbach:
>  I think we decided that if we want to at least make autodecoding nothrow, then we need to enforce that string types contain only valid UTF-8, as anything else is an error.

I thought that effort didn't go anywhere?
https://wiki.dlang.org/DIP76

> Actually, isn't this one of those places where we'd want to do preemptive UTF-8 validation before putting bytes in a string?

Hmm, do we do this somewhere else?

CyberShadow:
> Hmm, do we do this somewhere else?

Well, `readText` is the canonical example.

CyberShadow:
> I thought that effort didn't go anywhere?

Yeah, I don't think we decided we didn't want to do this, either... it seems to keep coming up from time to time, along with the general autodecoding discussions.

PR #5013, opened 2017-01-04
Fix 16191 - std/digest/digest.d should be renamed to package.d

wilzbach:
CC @schveiguy @jpf91 

From Bugzilla:

> The fully qualified name of the digest function template (https://dlang.org/phobos/std_digest_digest.html#digest) is std.digest.digest.digest. This is because std.digest is a package, with a module named digest in it, and the function digest inside that.

> I think it's reasonable to have std.digest.digest. Even though there's repetition, it's clear that the first digest is a module, the second is a function name. But there's no reason (especially with the advent of package.d) to have a std.digest.digest module.

I heard somewhere that `std/digest/digest` comes from a time where `package.d` wasn't supported yet.
The rename is a separate commit, s.t. git detects it correctly and can retain the history.

PR #6277, opened 2018-03-15
Remove the empty std.digest.digest site

wilzbach:
https://dlang.org/phobos/std_digest_digest.html

std.digest.digest was moved to std.digest in https://github.com/dlang/phobos/pull/5013

### DIFF
--- a/std/digest/digest.d
+++ b/std/digest/digest.d
@@ -2,20 +2,34 @@ module std.digest.digest;
 
 import _newDigest = std.digest;
 
-// scheduled for deprecation in 2.077
-// See also: https://github.com/dlang/phobos/pull/5013#issuecomment-313987845
+// @@@DEPRECATED_2.084@@@
+deprecated("import std.digest instead of std.digest.digest. std.digest.digest will be removed in 2.084")
 alias isDigest = _newDigest.isDigest;
+deprecated("import std.digest instead of std.digest.digest. std.digest.digest will be removed in 2.084")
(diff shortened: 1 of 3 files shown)


wilzbach:
Oh it seems like the `deprecation` warnings have never been added. I also added this to this PR and opt-ed for 2.084 as it already has been part of the changelog in 2.076 and the documentation hasn't been existent since 2.076.

https://github.com/dlang/phobos/pull/5013
https://dlang.org/changelog/2.076.0.html#std-digest-package

std/digest/digest.d
```diff
@@ -2,20 +2,34 @@ module std.digest.digest;
 
 import _newDigest = std.digest;
 
-// scheduled for deprecation in 2.077
-// See also: https://github.com/dlang/phobos/pull/5013#issuecomment-313987845
+// @@@DEPRECATED_2.084@@@
+deprecated("std.digest.digest is deprecated. Please use std.digest. Removal in 2.084. Removal in 2.084")
```
JackStouffer: You repeated the removal time.

Can we use a sentence that helps the user by explicitly telling them what to do? Like this

```d
deprecated("import std.digest instead of std.digest.digest. std.digest.digest will be removed in 2.084")
```

wilzbach: > Can we use a sentence that helps the user by explicitly telling them what to do? Like this

Good idea. Done.

PR #6280, opened 2018-03-15
Reduce size of slow testcase in std.algorithm.setops

FeepingCreature:
This test takes an unfeasibly large amount of time to run. (I've never managed to run it to completion.)

The only reason why it appears to pass is that it completes instantly, because canFind never evaluates the Cartesian range at all due to DMD bug #12486. When I fixed it in DMD PR https://github.com/dlang/dmd/pull/8013, the test started timing out.

By the way, how do I get reviewers for that DMD PR?

### DIFF
--- a/std/algorithm/setops.d
+++ b/std/algorithm/setops.d
@@ -504,7 +504,7 @@ if (!allSatisfy!(isForwardRange, R1, R2, RR) ||
 
     assert(canFind(N4, tuple(1, 2, 3, 4)));
     assert(canFind(N4, tuple(4, 3, 2, 1)));
-    assert(canFind(N4, tuple(10, 31, 7, 12)));
+    assert(canFind(N4, tuple(10, 3, 7, 2)));
 }
 
 // Issue 9878


std/algorithm/setops.d
```diff
@@ -504,7 +504,7 @@ if (!allSatisfy!(isForwardRange, R1, R2, RR) ||
 
     assert(canFind(N4, tuple(1, 2, 3, 4)));
     assert(canFind(N4, tuple(4, 3, 2, 1)));
-    assert(canFind(N4, tuple(10, 31, 7, 12)));
+    assert(canFind(N4, tuple(10, 3, 7, 2)));
```
n8sh: Even better could that line be removed outright? Does it cover anything not already tested by
```d
assert(canFind(N4, tuple(1, 2, 3, 4)));
assert(canFind(N4, tuple(4, 3, 2, 1)));
```
?

FeepingCreature: I guess "larger numbers than 1 in every dimension"?

quickfur:
Though, BTW, a more fundamental problem here is how variadic `cartesianProduct` is implemented. The current implementation for infinite ranges causes an exponential number of template expansions, which will quickly cause system RAM to run out as the number of arguments grow. There ought to be a way of implementing it such that it's less template-heavy.

PR #6278, opened 2018-03-15
posix.mak: module-specific -dip25/1000 for someModule.test and unitte…

carblue:
st rules

Somehow I managed to screw my git push-connection to closed PR https://github.com/dlang/phobos/pull/6195
So here it is again, updated.

This PR is not expected to be merged, possibly wrong to merge unchanged (I don't know; and it's useful temporarily only anyway).
"I'm not aware of the complete picture of the build process and all the consequences of such a phobos.posix.mak in CI: It may perhaps prevent merging a legitimate fix in one file because it touches yet non(-dip1000/-unit-)tested code of another phobos module, now failing due to -dip1000. "

It's main purpose is to document the state/infos/progress of -dip1000 compilable phobos and preferrably to be used locally by anybody interested. I will refer to this posix.mak in my "fix -dip1000 compilable" PRs.

It includes a global associative array 'aa' with module-specific (key) and value for -dip25/1000, according to how master's modules are compilable, updated regularly acc. progress.
Thus it's also the TODO-list for the "few" remaining -dip25 only compilable modules.
Members with write access can write into carblue:dip1000_3; I'll pay attention not to overwrite when pushing updates.
WIP_carblue will be my mark what I'm working on.
If there is a champion who wants to contribute for a difficult issue: It's the std.stdio. write family of functions, that needs fixing, currently triggering errors from several modules:
Just set -dip1000 for the aa[module] and go down to the rabbit hole.
I do these tests, each with -dip1000 and -dip25 as well (because it may happen to work for -dip1000, but not for -dip25 any more):
make -j6 -f posix.mak publictests     and the next 2 ones

aa is included in 2 rules for unittesting/compiling, currently; suitable as is for Linux (at least):

make     -f posix.mak path/to/someModule.test    e.g. make -f posix.mak std/uni.test
make -j6 -f posix.mak unittest

I preferred associative array aa entries in the "modulename-notation" (e.g. aa[std.array]) instead of OS-dependant "pathname-notation" (didn't recall whether windows can cope with forward-slashes).
But this only defers the issue to the call-site (I use Linux, i.e. /): $(aa[$(subst /,.,$(basename $<))])

Used currently in 2 rules, e.g.
%.test : %.d $(LIB)
T=mktemp -d /tmp/.dmd-run-test.XXXXXX &&
(
[...]

PR #6281, opened 2018-03-15
Work On Issue 18172 - add scope to pointers in getopt to allow @safe

JackStouffer:


std/getopt.d
```diff
@@ -459,6 +459,19 @@ GetoptResult getopt(T...)(ref string[] args, T opts)
     }
 }
 
+version(DIP1000)
```
JackStouffer: I'm not sure how we want to handle DIP1000 regressions, but we need a way to ensure things stay `@safe`

std/getopt.d
```diff
@@ -700,7 +713,16 @@ private void getoptImpl(T...)(ref string[] args, ref configuration cfg,
         else
         {
             // it's an option string
-            auto option = to!string(opts[0]);
+            static if (is(typeof(opts[0]) == string))
+            {
+                // allocate a new string so as to remove the
+                // scope restriction
+                string option = "" ~ opts[0];
```
JackStouffer: this is due to the fact that `to!(string)(string)` acts as a pass through.

klickverbot: .idup?

std/getopt.d
wilzbach: By https://github.com/dlang/phobos/pull/6278 and thus gradually enabling -dip1000?

JackStouffer: But this `unittest` needs to be annotated with `@safe` when `-dip1000` is passed and `@system` when it's not to prevent any DIP1000 regressions. #6278 Doesn't have a mechanism for that.

wilzbach: Argh fair enough - there still might be issues as long as `-deps` compiles all unittests of its dependencies :/
Anyhow, for the time being #6278 (or similar could set `DIP1000` by hand).

std/getopt.d
```diff
@@ -700,7 +713,16 @@ private void getoptImpl(T...)(ref string[] args, ref configuration cfg,
         else
         {
             // it's an option string
-            auto option = to!string(opts[0]);
+            static if (is(typeof(opts[0]) == string))
+            {
+                // allocate a new string so as to remove the
+                // scope restriction
```
wilzbach: That really sucks. Safety shouldn't result in a performance overhead :/
Especially because in this case
- `getoptImpl` doesn't store anything, but modifies `configuration` and `rslt`
- it's a GC allocated string

wilzbach:
Here's an idea how to programmatically detect `-dip1000`:

```d
void main()
{

    enum isDIP1000 = __traits(compiles, (){
        void foo() @safe {
            int x;
            int* p;
            p = &x;
        }
    });
    pragma(msg, isDIP1000);
}
```

normal: https://run.dlang.io/is/MWDG8p
-dip1000: https://run.dlang.io/is/dKU7Mk

Of course that could be fancier. 

(based on the recent forum thread: https://forum.dlang.org/post/ayemvxctmpqqsokbmqeb@forum.dlang.org)

JackStouffer:
If people want to tackle the DIP1000 returning a struct issue, I'd be glad to see their solution. Until then, I'm closing this as this isn't going to go anywhere.

PR #6282, opened 2018-03-15
Add error message for static assert in bitmanip.d

crimaniak:
Static assert, checking the number of bits for a field with the type
bool, did not have an error message.


std/bitmanip.d
```diff
@@ -107,7 +107,7 @@ private template createAccessors(
 
         static if (is(T == bool))
         {
-            static assert(len == 1);
+            static assert(len == 1, "'"~name~"' definition problem: type 'bool' is allowed only for single-bit fields");
```
JackStouffer: spaces around `~`

crimaniak: fixed, but in fact, this file has complete anarchy.

wilzbach: Yes we try to gradually create consistency whenever it comes up. Though there's the idea of using dfmt to format the git diff automatically.

wilzbach:
So we use backticks since a while for colorized output. So it would be great to use them here as well instead of the normal single quotes. Also don't forget to squash everything into one commit (though we could do that for you if you run into problems)

Thanks.

JackStouffer:
Backticks work for static assert? I thought it was only for DMD's messages?

crimaniak:
Sorry guys, I can not find the documentation. Where can I read, how and in what messages do the backticks work?

wilzbach:
I don't know of any documentation documenting this :/
However, it does work:

https://run.dlang.io/is/AM3P5i

crimaniak:
About squashing: I read some documentation but I never did it before, only local before any push, so this time it will better if you just use merge with squashing.

wilzbach:
So I squashed everything for you and split the message in two lines as it was over 120 chars (the reason for the CircleCI failure).

crimaniak:
Thanks for fixing my commit and message text! (I am not native English speaker).

JackStouffer:
I think that’s true for half of the D contributors :)

wilzbach:
... even for myself ;-)

PR #6286, opened 2018-03-16
Push deprecation of std.exception.enforceEx

wilzbach:
In https://github.com/dlang/phobos/pull/6086 we prepared `enforce` to be a super set of `enforceEx` and thus allow a deprecation of it.

See also: https://dlang.org/changelog/2.079.0.html#std-exception-enforce

> However, if enforce can't fully replace enforceEx, then it needs to stick around for at least one more release so that we can have a release where enforce can fully replace enforceEx.

This has happened, so we finally can start the deprecation rodeo.

The targetted deprecation version is 2.084 as it's marked as a legacy symbol in the docs for quite a while now.

### DIFF
--- a/std/exception.d
+++ b/std/exception.d
@@ -636,16 +636,14 @@ T errnoEnforce(T, string file = __FILE__, size_t line = __LINE__)
     return value;
 }
 
-// @@@DEPRECATED_2.084@@@
+// @@@DEPRECATED_2.089@@@
 /++
-    $(RED Deprecated. Please use $(LREF enforce) instead. This function will be removed 2.084.)
+    $(RED Deprecated. Please use $(LREF enforce) instead. This function will be removed 2.089.)
 
     If $(D !value) is $(D false), $(D value) is returned. Otherwise,
     $(D new E(msg, file, line)) is thrown. Or if $(D E) doesn't take a message
     and can be constructed with $(D new E(file, line)), then
     $(D new E(file, line)) will be thrown.
-
-    This is legacy name, it is recommended to use $(D enforce!E) instead.
 
     Example:
     --------------------
@@ -654,7 +652,7 @@ T errnoEnforce(T, string file = __FILE__, size_t line = __LINE__)
     enforceEx!DataCorruptionException(line.length);
     --------------------
  +/
-//deprecated("Please use enforce instead")
+deprecated("Use `enforce`. `enforceEx` will be removed with 2.089.")
 template enforceEx(E : Throwable)
 if (is(typeof(new E("", string.init, size_t.init))))
 {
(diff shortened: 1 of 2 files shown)


CyberShadow:
What does `errnoEnforce` have to do with #6086, or anything? Do you mean `enforceEx`?

changelog/std-exception-errnoEnforce.dd
```diff
@@ -0,0 +1,20 @@
...
+With 2.079 $(REF enforce, std,exception) became a complete super set of
+$(REF errnoEnforce, std,exception)
+
+$(H3 Corrective action)
+
+Replace:
+
+---
+import std.exception : errnoEnforce;
+errnoEnforce(enforce(errno != EINVAL, "Invalid environment variable name: '" ~ name ~ "'");
```
CyberShadow: Not sure what you meant here but this is not valid code.

changelog/std-exception-errnoEnforce.dd
```diff
@@ -0,0 +1,20 @@
...
+---
+
+with:
+
+---
+import std.exception : enforce;
+enforce(enforce(errno != EINVAL, "Invalid environment variable name: '" ~ name ~ "'");
```
CyberShadow: Ditto

std/mmfile.d
```diff
@@ -124,7 +124,7 @@ class MmFile
 
         // Adjust size
         stat_t statbuf = void;
-        errnoEnforce(fstat(fd, &statbuf) == 0);
+        enforce(fstat(fd, &statbuf) == 0);
```
CyberShadow: All of these changes are bogus.

`errnoEnforce` is useful in that it catches `errno` and prints a useful error message depending on its value.

wilzbach:
> errnoEnforce is useful in that it catches errno and prints a useful error message depending on its value.

Well `alias errnoEnforce = enforce!ErrnoException` (see https://github.com/dlang/phobos/pull/6289), but you are right I mixed things up  (sorry - I should really go to bed earlier) and this PR should only be about triggering the overdue deprecation of `enforceEx`.

We could deprecate `errnoEnforce`  as well, but that's a different discussion (-> https://github.com/dlang/phobos/pull/6289).

std/exception.d
```diff
@@ -645,16 +645,14 @@ T errnoEnforce(T, string file = __FILE__, size_t line = __LINE__)
...
-
     Example:
     --------------------
     auto f = enforceEx!FileMissingException(fopen("data.txt"));
     auto line = readln(f);
     enforceEx!DataCorruptionException(line.length);
     --------------------
  +/
-//deprecated("Please use enforce instead")
+deprecated("Use `enforce`. `enforceEx` will be removed with 2.084.")
```
JackStouffer: this should be 2.089 if we follow https://github.com/dlang/DIPs/pull/108

wilzbach:
> this should be 2.089 if we follow dlang/DIPs#108

Urgh that's very long, but it will be even longer if this doesn't get merged, so I changed the version ...

wilzbach:
The deprecation has already been approved in #6086

PR #6288, opened 2018-03-16
Deprecate binaryReverseArgs in favor of reverseArgs

wilzbach:
`reverseArgs` was added in https://github.com/dlang/phobos/commit/29f3cc23f76c97a2782848033136d2d7ec7ca57e, but isn't in Phobos and [very seldom on GitHub](https://github.com/search?l=D&q=binaryReverseArgs&type=Code&utf8=%E2%9C%93) - considering that almost all matches are either a copy of Phobos or a copy of the same Rosetta solution. For the latter `writefln!"..."` would be the idiomatic code anyhow.

### DIFF
--- a/std/functional.d
+++ b/std/functional.d
@@ -36,7 +36,7 @@ $(TR $(TH Function Name) $(TH Description)
         $(TD Creates a function that binds the first argument of a given function
         to a given value.
     ))
-    $(TR $(TD $(LREF reverseArgs), $(LREF binaryReverseArgs))
+    $(TR $(TD $(LREF reverseArgs))
         $(TD Predicate that reverses the order of its arguments.
     ))
     $(TR $(TD $(LREF toDelegate))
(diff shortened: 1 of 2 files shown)


std/functional.d
```diff
@@ -537,11 +537,28 @@ alias equalTo = safeOp!"==";
 template reverseArgs(alias pred)
 {
     auto reverseArgs(Args...)(auto ref Args args)
-        if (is(typeof(pred(Reverse!args))))
+    if (is(typeof(pred(Reverse!args))))
     {
         return pred(Reverse!args);
     }
 }
+//
```
JackStouffer: did you mean to use `///` here?

wilzbach: Yes. I copied the public tests from binaryReverseArgs as they could be kept.

wilzbach:
BTW would you be okay with removing deprecated symbols from the "module quick overview booktable"?
Imho we shouldn't advertise deprecated symbols there.

JackStouffer:
>BTW would you be okay with removing deprecated symbols from the "module quick overview booktable"?

Yeah that sounds like a good practice, and something that I should add to the DIP.

wilzbach:
So we need @andralex's approval for this simple deprecation or can we move ahead directly?

andralex:
no need but thx

PR #6290, opened 2018-03-16
Fix Issue 18624 - wrong behavior of getSymbolsByUDA if one of the symbols having th…

Spoov:
…e UDA is a function

The last change to getSymbolsByUDA introduced a bug, causing it to ignore some symbols

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -7877,7 +7877,8 @@ private template getSymbolsByUDAImpl(alias symbol, alias attribute, names...)
             else static if (isFunction!member)
             {
                 enum hasSpecificUDA(alias member) = hasUDA!(member, attribute);
-                alias getSymbolsByUDAImpl = Filter!(hasSpecificUDA, __traits(getOverloads, symbol, names[0]));
+                alias overloadsWithUDA = Filter!(hasSpecificUDA, __traits(getOverloads, symbol, names[0]));
+                alias getSymbolsByUDAImpl = AliasSeq!(overloadsWithUDA, tail);
             }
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -8032,6 +8033,21 @@ private template getSymbolsByUDAImpl(alias symbol, alias attribute, names...)
...
+    enum Attr;
+    struct A
+    {
+        @Attr void a();
+        @Attr void a(int n);
+              void b();
+        @Attr void c();
+    }
+
+    static assert(getSymbolsByUDA!(A, Attr).length == 3);
```
wilzbach: Why don't you check the result directly?
(would be more robust)

A simple way would be `getSymbolsByUDA!(A, Attr).stringof == "tuple(a, a, c)"`

wilzbach:
As this is a regression it would be great if you could open an issue for it and reference it in the commit message (`git commit --amend`).

Also it should be rebased to stable:

```
git rebase --onto upstream/stable upstream/master
```

See the message of dlang-bot for details.

wilzbach:
See also https://github.com/dlang/phobos/pull/6292 - that's why linking to the PR is a good idea ...

Spoov:
This is my first Pull Request / Issue here so i didn't tough of doing it. I hope i remember doing it next time :)

wilzbach:
No worries. Nobody is perfect - especially on the first run.
Thanks a lot for your contribution! This will be part of of 2.079.1 - it will be released ~ 1.4.18

PR #6291, opened 2018-03-16
Add public examples to std.conv and enable dscanner check

JackStouffer:


std/conv.d
```diff
@@ -4179,20 +4193,10 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) &&
 string text(T...)(T args)
 if (T.length > 0) { return textImpl!string(args); }
 
-// @@@DEPRECATED_2018-06@@@
```
JackStouffer: https://github.com/dlang-community/D-Scanner/issues/566

PR #6292, opened 2018-03-17
Fix issue 18624

Biotronic:
getSymbolsByUDA produces wrong result if one of the symbols having the UDA is a function


wilzbach:
See also https://github.com/dlang/phobos/pull/6290 - that's why linking to the PR is a good idea ...

wilzbach:
#6290 is already targeted at stable, so I think it makes sense to focus the efforts on that PR.

PR #6297, opened 2018-03-19
Fix documentation for Take, Cycle, and Choose

JackStouffer:


std/range/package.d
```diff
@@ -1538,37 +1570,6 @@ private struct ChooseResult(R1, R2)
         }
 }
 
-///
-@safe nothrow pure @nogc unittest
```
JackStouffer: https://github.com/dlang-community/D-Scanner/issues/566

std/range/package.d
```diff
@@ -1371,6 +1371,38 @@ if (isInputRange!(Unqual!R1) && isInputRange!(Unqual!R2) &&
...
 }
 
+///
+@safe nothrow pure @nogc unittest
+{
+    import std.algorithm.comparison : equal;
+    import std.algorithm.iteration : filter, map;
+
+    auto data1 = only(1, 2, 3, 4).filter!(a => a != 3);
+    auto data2 = only(5, 6, 7, 8).map!(a => a + 1);
```
wilzbach: I know that this is just copy/pasted, but looking at this I'm not sure whether it's a good idea to use `only` just to be `@nogc`. People are a lot more familiar with arrays.

PR #6298, opened 2018-03-19
Added public example for assumeSorted

JackStouffer:


std/range/package.d
```diff
@@ -10328,24 +10328,24 @@ if (isInputRange!(Unqual!R))
...
-    assert(equal(p, [0, 1, 2, 3, 4]));
-    p = assumeSorted(a).lowerBound(6);
-    assert(equal(p, [ 0, 1, 2, 3, 4, 5]));
-    p = assumeSorted(a).lowerBound(6.9);
-    assert(equal(p, [ 0, 1, 2, 3, 4, 5, 6]));
+
+    int[] a = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+    auto p = assumeSorted(a);
+
+    assert(equal(p.lowerBound(4), [0, 1, 2, 3]));
```
wilzbach: Nit: FWIW people were complaining on the state of D survey about inconsistency in the docs. I think the majority uses the UFCS ( `p1.equal(X)`), so maybe a good idea to stick to that for public examples?
(I know that this is only copied)

PR #6299, opened 2018-03-19
Add documentation for StoppingPolicy and Fix docs for Lockstep

JackStouffer:


std/range/package.d
```diff
@@ -4787,48 +4818,6 @@ private:
     StoppingPolicy _stoppingPolicy;
 }
 
-string lockstepReverseFailMixin(Ranges...)(bool withIndex)
```
JackStouffer: https://github.com/dlang-community/D-Scanner/issues/566

std/range/package.d
```diff
@@ -4435,6 +4435,37 @@ enum StoppingPolicy
...
+{
+    import std.algorithm.comparison : equal;
+    import std.exception : assertThrown;
+    import std.range.primitives;
+    import std.typecons : tuple;
+
+    auto a = [1, 2, 3];
+    auto b = [4, 5, 6, 7];
+
+    auto shortest = zip(StoppingPolicy.shortest, a.save, b.save);
```
wilzbach: No need for `.save` here. Arrays are views only and the ptr + length is copied automatically. Also `zip` doesn't use `ref`

JackStouffer:
Fixed

PR #6300, opened 2018-03-19
Add examples for TransverseOptions, isTwoWayCompatible, SearchPolicy

JackStouffer:


std/range/package.d
```diff
@@ -6028,7 +6028,25 @@ enum TransverseOptions
...
+    auto arr = [[1, 2], [3, 4, 5]];
+
+    auto r1 = arr.frontTransversal!(TransverseOptions.assumeJagged);
+    assert(r1.equal([1, 3]));
+
+    // throws on construction
+    assertThrown!Exception(arr.frontTransversal!(TransverseOptions.enforceNotJagged));
+
+    auto r2 = arr.frontTransversal!(TransverseOptions.assumeNotJagged);
+    assert(r1.equal([1, 3]));
```
wilzbach: Hmm, this option is more or less useless. Should we really document it?

std/range/package.d
```diff
@@ -9676,20 +9705,33 @@ enum SearchPolicy
...
+}
+
+///
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+
+    auto a = assumeSorted([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]);
+    auto p1 = a.upperBound!(SearchPolicy.binarySearch)(3);
+    assert(equal(p1, [4, 5, 6, 7, 8, 9]));
```
wilzbach: Nit: FWIW people were complaining on the state of D survey about inconsistency in the docs. I think the majority uses the UFCS ( `p1.equal(X)`), so maybe a good idea to stick to that for public examples?

JackStouffer:
Fixed

PR #6302, opened 2018-03-19
fix issue 18631 - Allow choice to work with const arrays

BBasile:


std/random.d
```diff
@@ -2017,7 +2017,7 @@ Returns:
     return a `ref` to the $(D range element), otherwise it will return
     a copy.
  */
-auto ref choice(Range, RandomGen = Random)(auto ref Range range,
+auto ref choice(Range, RandomGen = Random)(inout ref Range range,
```
JackStouffer: Shouldn't this be `inout auto ref`?

Because I don't think this will compile anymore `choice([1, 2, 3, 4]);`

BBasile: `inout` stands for `auto` in this case. I've added a test for a call by value since your comment.

jmdavis: Honestly, this seems like a compiler bug. I've never heard of `inout ref` doing anything similar to `auto ref`, and that behavior does not match what happens with `const` or `immutable`. Also, the way `inout` is designed, it's always supposed to be used on the return value in addition to one of the parameters. So, even if it's true that `inout ref` is supposed to act like `auto ref` with regards to rvalues, it would then be a bug that `inout` isn't used on the return type.

JackStouffer:
Oh, just saw your comment @jmdavis. @BBasile Can you add `inout` to the return type?

std/random.d
BBasile: The problem is that i use an overload, not the function changed. Actually the fix is wrong...

BBasile:
Okay done.

JackStouffer:
Geez inout is so ugly.

jmdavis:
> Geez inout is so ugly.

In most cases, with templated functions, you can just not bother with `const` or `inout` or whatever, and everything works, but as soon as `ref` gets involved, things get messier. :(

wilzbach:
(manually merging as Darwin_64_32 is currently broken)

jmdavis:
This PR is a breaking change, because it assumes that `length` and `opSlice` work with `const`, and they frequently do not.

quickfur:
Any suggestions as to how to fix it, short of reverting the whole thing?  Preferably, we would not have to undo everything done here.

jmdavis:
This is trying to fix a corner cases in ranges. Either it will work to just slice the array before passing it in, or a separate overload for dynamic arrays will be required. Regardless, marking a range as `inout` or `const` is almost never appropriate. I didn't look over this thoroughly enough when I glanced over it before, so I didn't notice the problem like I should have. This should never have been merged.

https://github.com/dlang/phobos/pull/6331

schveiguy:
> Geez inout is so ugly.

The `inout(ElementType!Range)` was not required, you could have just put `inout`. It's a storage class, just like all the others, and so allows for type inference.

schveiguy:
Yes, `inout` (and `const`) needs to have buy-in from all the types you are trying to wrap, or you need to introspect and forward them separately. In the case where it is supported, unfortunately, you can't have it inferred.

The downside of not doing this, is that you lose the "temporary const" guarantee that it provides.

schveiguy:
Since it's pretty accepted that you can slice an object and get a range from it, Another way to fix this may be (not a full implementation, needs some tweaking):

```D
auto ref choice(R)(auto ref R r) if (!isRandomAccessRange!R && isRandomAccessRange!(typeof(r[])))
{
    return choice(r[]);
}
```

It should work for containers as well, such as `Array!T`

PR #6304, opened 2018-03-19
Fix issue 18634 - support for delegate comparators in rbtree

chvor:
RedBlackTree.opEquals() was preventing the use of delegates by explicitly forcing the equal() check to go through a function.

### DIFF
--- a/std/container/rbtree.d
+++ b/std/container/rbtree.d
@@ -1075,7 +1075,7 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
 
         auto thisRange = this[];
         auto thatRange = that[];
-        return equal!(function(Elem a, Elem b) => !_less(a,b) && !_less(b,a))
+        return equal!((Elem a, Elem b) => !_less(a,b) && !_less(b,a))
                      (thisRange, thatRange);
     }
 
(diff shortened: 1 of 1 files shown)


BBasile:
This issue needs to reference an issue of D's bugzilla. The point is that it would automatically appear in the changelog if picked in the stable branch for a release.

BBasile:
Thanks but you must change the commit message, (git amend) the exact same thing with the git commit message will be fine.

std/container/rbtree.d
```diff
@@ -2064,3 +2064,13 @@ if ( is(typeof(binaryFun!less((ElementType!Stuff).init, (ElementType!Stuff).init
     static assert(!__traits(compiles, *t3.front.p = 4));
     assert(*t3.front.p == 1);
 }
+
+// make sure the comparator can be a delegate
+@safe pure unittest
+{
+    import std.algorithm.comparison : equal;
+
+    auto t = new RedBlackTree!(int, delegate (a, b) => a > b);
```
wilzbach: It's always a good idea to use the more common style:

```
~/dlang/phobos on master [A11] ± ag "delegate\(" |wc
    384    2599   31451

~/dlang/phobos on master [A11] ± ag "delegate \(" |wc
      9      91     795
```

wilzbach:
Thanks a lot for your first PR to Phobos and welcome to the community!
If you ever need any help, you can also find it on #d on IRC, #phobos on Slack or simply pinging us here.

std/container/rbtree.d
chvor: Thanks!

chvor:
@BBasile and @wilzbach, thanks for the advice and the warm welcome!

PR #6306, opened 2018-03-20
Fixed AscendingPageAllocator race condition

jercaianu:
The race here was that while 'extendMemoryProtection' was called, some other thread could access memory before the calls to mprotect/VirtualProtect could finish, resulting in segfault.

Fix here was easy, moving 'extendMemoryProtection' under the lock.


jercaianu:
Do not merge please, it's not completely fixed

JackStouffer:
Ping @Darredevil @edi33416

edi33416:
LGTM

PR #6308, opened 2018-03-20
Add examples to std.datetime.date and enable dscanner check

JackStouffer:


std/datetime/date.d
```diff
@@ -3681,6 +3694,29 @@ private:
...
+    import core.time : days, seconds;
+
+    auto dt = DateTime(2000, 6, 1, 10, 30, 0);
+
+    assert(dt.date == Date(2000, 6, 1));
+    assert(dt.timeOfDay == TimeOfDay(10, 30, 0));
+    assert(dt.dayOfYear == 153);
+    assert(dt.dayOfWeek == DayOfWeek.thu);
+
+    dt += days(10) + seconds(100);
```
wilzbach: `10.days` + `100.seconds` would be idiomatic

std/datetime/date.d
```diff
@@ -8098,6 +8134,28 @@ package:
...
+@safe pure unittest
+{
+    import core.time : days;
+
+    auto d = Date(2000, 6, 1);
+
+    assert(d.dayOfYear == 153);
+    assert(d.dayOfWeek == DayOfWeek.thu);
+
+    d += days(10);
```
wilzbach: 10.days would be idiomatic

std/datetime/date.d
```diff
@@ -9407,6 +9465,22 @@ package:
...
 }
 
+///
+@safe pure unittest
+{
+    import core.time : minutes, seconds;
+
+    auto t = TimeOfDay(12, 30, 0);
+
+    t += minutes(10) + seconds(100);
```
wilzbach: 10.minutes + 100.seconds would be idiomatic.

PR #6310, opened 2018-03-20
Add public examples to std.variant + DScanner check

wilzbach:


std/variant.d
```diff
@@ -1215,6 +1215,51 @@ public:
     }
 }
 
+///
+@system unittest
+{
+    Variant a; // Must assign before use, otherwise exception ensues
```
JackStouffer: Shouldn't these be examples of `VariantN`?

wilzbach: GitHub's comment matching algorithm doesn't seem to be ideal :/
Anyhow, I now use `alias Var = VariantN!(maxSize!(int, real, double));` for the `VariantN` examples.

JackStouffer:
@wilzbach Please version out the symlink tests for Windows.

wilzbach:
> @wilzbach Please version out the symlink tests for Windows.

Yeah I discovered the same, sorry -> https://github.com/dlang/phobos/pull/6326

(though I still don't know why auto-tester allowed the other PR to be merged)

PR #6311, opened 2018-03-20
Split-up .dscanner.ini blacklists into multiple lines

wilzbach:
To avoid merge conflicts and ease editing the big lists.

Requires https://github.com/burner/inifiled/pull/10 (the underlying library used by DScanner)

(I used one check as an example)


wilzbach:
I was too annoyed with the constant merge conflicts, so I updated `phobos` branch of DScanner to include burner/inifiled#10 (see https://github.com/dlang-community/D-Scanner/pull/569).

-> we can move ahead with this

CC @JackStouffer

PR #6314, opened 2018-03-20
Add public examples to std.uri + enable DScanner check

wilzbach:


std/uri.d
```diff
@@ -287,6 +294,13 @@ if (isSomeChar!Char)
     return r;
 }
 
+///
+@safe unittest
+{
+    assert("foo&/".encode == "foo&/");
```
JackStouffer: This should be an example for `decode`

wilzbach: There was a small typo before but this unittest is for `encode`

std/uri.d
```diff
@@ -287,6 +294,13 @@ if (isSomeChar!Char)
     return r;
 }
 
+///
+@safe unittest
+{
+    assert("foo&/".decode == "foo&/");
```
JackStouffer: This doesn't really show anything off...

PR #6317, opened 2018-03-21
Combine docs for std.range.repeat

JackStouffer:


std/range/package.d
```diff
@@ -3183,13 +3183,17 @@ pure @safe nothrow unittest
 }
 
 /**
-Create a range which repeats one value forever.
+Create a range which repeats one value.
 
 Params:
-    value = the value to repeat
+    value = the _value to repeat
+    n = the number of times to repeat `value`
```
wilzbach: You are a bit inconsistent here (_value vs. `value`)

JackStouffer: It's the difference between referring to the parameter value (\`value\`) vs the concept of a value (_value).

JackStouffer:
Attempting to restart the doc tester

PR #6318, opened 2018-03-21
Add public examples to std.typecons + DScanner check

wilzbach:
~~~Postponed the DScanner check until https://github.com/dlang/phobos/pull/6311 is in.~~~

edit: now with DScanner check

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2083,6 +2083,28 @@ if (is(T == class) || is(T == interface) || isDynamicArray!T || isAssociativeArr
     return ret;
 }
 
+///
+@system unittest
+{
+    class C
+    {
+        int payload;
+        this(int p) { payload = p; }
+    }
+    const c = new C(1);
+
+    auto c2 = c.rebindable;
+    assert(c2.payload == 1);
+    // passing Rebindable to rebindable
+    c2 = c2.rebindable;
+
+    c2 = new C(2);
+    assert(c2.payload == 2);
+
+    const c3 = c2.get;
+    assert(c3.payload == 2);
+}
+
 /**
 This function simply returns the $(D Rebindable) object passed in.  It's useful
 in generic programming cases when a given object may be either a regular
@@ -2097,6 +2119,24 @@ Returns:
 Rebindable!T rebindable(T)(Rebindable!T obj)
 {
     return obj;
+}
+
+// TODO: remove me once the rebindable overloads have been joined
+///
+@system unittest
+{
+    class C
+    {
+        int payload;
+        this(int p) { payload = p; }
+    }
+    const c = new C(1);
+
+    auto c2 = c.rebindable;
+    assert(c2.payload == 1);
+    // passing Rebindable to rebindable
+    c2 = c2.rebindable;
+    assert(c2.payload == 1);
 }
 
 @system unittest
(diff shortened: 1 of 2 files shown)


std/typecons.d
```diff
@@ -3966,6 +3986,60 @@ class AutoImplement(
...
+
+         // Workaround https://issues.dlang.org/show_bug.cgi?id=2525
+        abstract int foo();
+        abstract int bar();
+    }
+
+    template fallback(T, alias func)
+    {
+        import std.format : format;
+        enum fallback = q{
```
JackStouffer: add a comment here explaining this line so less experienced users understand

std/typecons.d
```diff
@@ -5497,6 +5607,27 @@ enum RefCountedAutoInitialize
...
+    {
+        int a = 42;
+    }
+
+    RefCounted!(Foo, RefCountedAutoInitialize.yes) rcAuto;
+    RefCounted!(Foo, RefCountedAutoInitialize.no) rcNoAuto;
+
+    assert(rcAuto.refCountedPayload.a == 42);
+
+    assertThrown!AssertError((){ return rcNoAuto.refCountedPayload; }().a);
```
JackStouffer: why the anonymous function?

std/typecons.d
```diff
@@ -6778,6 +6909,23 @@ struct Typedef(T, T init = T.init, string cookie=null)
     }
 }
 
+///
+@safe unittest
```
JackStouffer: I'd remove the examples from the doc string

wilzbach: There was actually a bit of value in them. So I kept them, but moved them to `unittest` blocks (see the second commit).

wilzbach: (I had to rebase due to a merge conflict, so I squashed everything)

std/typecons.d
```diff
@@ -2068,35 +2068,55 @@ if (is(T == class) || is(T == interface) || isDynamicArray!T || isAssociativeArr
 Convenience function for creating a $(D Rebindable) using automatic type
 inference.
 
+If a $(LREF Rebindable) obj is passed to `rebindable`, it is returned without modifications.
```
MetaLang: IMO the documentation should not contain shortenings like this. "obj" can be removed without affecting the clarity of the text.

std/typecons.d
```diff
@@ -3966,6 +3986,63 @@ if (is(Interface == interface) && is(BaseClass == class))
...
+    {
+        protected PackageSupplier default_, fallback;
+
+        this(PackageSupplier default_, PackageSupplier fallback)
+        {
+            this.default_ = default_;
+            this.fallback = fallback;
+        }
+
+         // Workaround https://issues.dlang.org/show_bug.cgi?id=2525
```
MetaLang: I hate to reference a bug in public facing documentation. Who will remember to remove this when the bug is fixed?

std/typecons.d
```diff
@@ -2068,35 +2068,55 @@ if (is(T == class) || is(T == interface) || isDynamicArray!T || isAssociativeArr
...
+        return ret;
+    }
+    else static if (__traits(isSame, TemplateOf!T, Rebindable))
+    {
+        return obj;
+    }
+    else
+    {
+        static assert(0, "Unknown type used.");
+    }
```
MetaLang: Don't mix changes in functionality in with a documentation PR. Please split this off into a separate one.

std/typecons.d
```diff
@@ -5014,9 +5122,6 @@ if (!isMutable!Target)
     {
         int reflesh();
     }
-    // does not have structural conformance
```
MetaLang: Why remove these?

std/typecons.d
```diff
@@ -4742,12 +4855,14 @@ if (is(T == class) || is(T == interface))
...
- * Supports structural based typesafe conversion.
- *
- * If $(D Source) has structural conformance with the $(D interface) $(D Targets),
- * wrap creates internal wrapper class which inherits $(D Targets) and
- * wrap $(D src) object, then return it.
- */
+Supports structural based typesafe conversion.
+
+If `Source` has structural conformance with the `interface` `Targets`,
+wrap creates internal wrapper class which inherits `Targets` and
```
MetaLang: "creates _an_"
"and _wraps_ _the_ `src` object, then _returns_ it."

Would fix them myself, but on my phone.

std/typecons.d
wilzbach: This was copy/pasted from the other documentation header, but yeah good idea to fix.

std/typecons.d
wilzbach: Me too, but I thought we somehow should point out that these redundancy isn't on purpose -> removed then.

std/typecons.d
wilzbach: Hmm this is needed to be able to avoid duplicating the unittest for `rebindable` is has two different overloads and I thought that's trivial enough, but I can reverted it for now then.

std/typecons.d
```diff
@@ -3966,6 +4006,62 @@ if (is(Interface == interface) && is(BaseClass == class))
...
+        protected PackageSupplier default_, fallback;
+
+        this(PackageSupplier default_, PackageSupplier fallback)
+        {
+            this.default_ = default_;
+            this.fallback = fallback;
+        }
+
+        abstract int foo();
+        abstract int bar();
```
wilzbach: For the record: this is a workaround against  https://issues.dlang.org/show_bug.cgi?id=2525

std/typecons.d
```diff
@@ -2099,6 +2121,24 @@ Rebindable!T rebindable(T)(Rebindable!T obj)
...
+        int payload;
+        this(int p) { payload = p; }
+    }
+    const c = new C(1);
+
+    auto c2 = c.rebindable;
+    assert(c2.payload == 1);
+    // passing Rebindable to rebindable
+    c2 = c2.rebindable;
+    assert(c2.payload == 2);
```
JackStouffer: fails

JackStouffer:
de-prioritizing until test failures are fixed.

PR #6320, opened 2018-03-21
Add public examples to std.file + enable DScanner check

wilzbach:


std/file.d
```diff
@@ -1213,14 +1284,14 @@ if (isConvertibleToString!R)
     auto currTime = Clock.currTime();
 
     write(deleteme, "a");
-    scope(exit) { assert(exists(deleteme)); remove(deleteme); }
+    scope(exit) deleteme.exists && deleteme.remove;
```
JackStouffer: why remove the assert?

std/file.d
```diff
@@ -1585,6 +1675,31 @@ if (isConvertibleToString!R)
...
+    SysTime modificationTime = SysTime(DateTime(2011, 10, 4, 0, 0, 30), 10.hnsecs);
+    setTimes(deleteme, accessTime, modificationTime);
+
+    SysTime accessTimeResolved, modificationTimeResolved;
+    getTimes(deleteme, accessTimeResolved, modificationTimeResolved);
+
+    assert(accessTime == accessTimeResolved);
+    assert(modificationTime == modificationTimeResolved);
+
+    import std.datetime : Clock, seconds;
```
JackStouffer: Everything before this line isn't related to `timeLastModified` and should be removed IMO

std/file.d
wilzbach: Removes throws and exception if the file doesn't exist

std/file.d
```diff
@@ -1492,6 +1563,28 @@ if (isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R) &&
...
+    setTimes(deleteme, accessTime, modificationTime);
+
+    SysTime accessTimeResolved, modificationTimeResolved;
+    getTimes(deleteme, accessTimeResolved, modificationTimeResolved);
+
+    assert(accessTime == accessTimeResolved);
+    import std.stdio;
+    modificationTime.writeln;
+    modificationTimeResolved.writeln;
+    assert(modificationTime == modificationTimeResolved);
```
wilzbach: This is failing on OSX. Maybe OSX doesn't have so fine-grained timestamps?

wilzbach: Looks like it:

```
> modificationTime.writeln;
> modificationTimeResolved.writeln;
2018-Oct-04 00:00:30.000001
2018-Oct-04 00:00:30
```

PR #6324, opened 2018-03-22
Added examples, improved docs, and changed style of std.traits.QualifierOf

JackStouffer:


std/traits.d
```diff
@@ -421,16 +421,40 @@ template ImmutableOf(T)
...
+ */
 template QualifierOf(T)
 {
-         static if (is(T == shared(const U), U)) alias QualifierOf = SharedConstOf;
-    else static if (is(T ==        const U , U)) alias QualifierOf = ConstOf;
-    else static if (is(T == shared(inout U), U)) alias QualifierOf = SharedInoutOf;
-    else static if (is(T ==        inout U , U)) alias QualifierOf = InoutOf;
-    else static if (is(T ==    immutable U , U)) alias QualifierOf = ImmutableOf;
-    else static if (is(T ==       shared U , U)) alias QualifierOf = SharedOf;
-    else                                         alias QualifierOf = MutableOf;
```
wilzbach: I think this was purposefully formatted like this to allow a brief overview on all covered cases.

JackStouffer: I can revert it but the reason I changed it is it doesn't technically conform to the style guide.

wilzbach: I don't feel strongly either way, maybe we wait one or two days, s.t. someone who feels strongly can raise their voice?

JackStouffer:
Going to move forward with this.

PR #6325, opened 2018-03-22
Added examples and improved docs of std.traits.extractParameterStorag…

JackStouffer:
…eClassFlags

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -1199,12 +1199,14 @@ template ParameterStorageClassTuple(func...)
     static assert(pstc[2] == STC.none);
 }
 
-/*****************
- * Convert string tuple Attribs to ParameterStorageClass bits
- * Params:
- *      Attribs = string tuple
- * Returns:
- *      ParameterStorageClass bits
+/**
+Convert the result of `__traits(getParameterStorageClasses)`
+to $(LREF ParameterStorageClass) `enum`s.
+
+Params:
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -1199,12 +1199,14 @@ template ParameterStorageClassTuple(func...)
...
+/**
+ * Convert the result of `__traits(getParameterStorageClasses)`
+ * to $(LREF ParameterStorageClass) `enum`s.
+ *
  * Params:
- *      Attribs = string tuple
+ *     Attribs = The return value of `__traits(getParameterStorageClasses)`
  * Returns:
- *      ParameterStorageClass bits
+ *     The bitwise OR of the equivalent $(LREF ParameterStorageClass) `enum`s.
```
wilzbach: BTW when you restructure comments, you can remove the leading stars as we agreed on it not being the [DStyle for Phobos](https://dlang.org/dstyle.html#phobos_documentation).
(this is not required, just an FYI)

std/traits.d
```diff
@@ -1234,6 +1236,30 @@ template extractParameterStorageClassFlags(Attribs...)
     }();
 }
 
+///
+@safe unittest
+{
+    static void func(ref int ctx, out real result)
+    {
+    }
```
wilzbach: FYI: It's really okay to do out everything in one line in such cases e.g. `func(...) {}

PR #6326, opened 2018-03-22
Fix std.file unittests for auto-tester

wilzbach:
So apparently auto-tester is currently broken because `std.file.symlink` is Posix-only. This is a hotfix for the newly added tests.
I'm still impressed that https://github.com/dlang/phobos/pull/6320 passed on auto-tester with "8 pass" and not "8 pass, 2 pending" or "8 pass, 2 fail".
CC @braddr

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -2189,19 +2189,23 @@ if (isConvertibleToString!R)
     import std.conv : octal;
 
     auto f = deleteme ~ "file";
-    scope(exit) f.remove;
+    version(Posix)
+    {
+        scope(exit) f.remove;
 
-    assert(!f.exists);
-    assertThrown!FileException(f.setAttributes(octal!777));
+        assert(!f.exists);
+        assertThrown!FileException(f.setAttributes(octal!777));
 
-    f.write(".");
-    auto attributes = f.getAttributes;
-    assert(!attributes.attrIsDir);
-    assert(attributes.attrIsFile);
+        f.write(".");
+        auto attributes = f.getAttributes;
+        assert(!attributes.attrIsDir);
+        assert(attributes.attrIsFile);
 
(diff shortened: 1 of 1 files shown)


std/file.d
```diff
@@ -2737,17 +2748,21 @@ bool attrIsSymlink(uint attributes) @safe pure nothrow @nogc
...
     assert(!source.exists);
     assertThrown!FileException(source.getLinkAttributes.attrIsSymlink);
 
-    target.write("target");
-    target.symlink(source);
-    assert(source.readText == "target");
-    assert(source.isSymlink);
-    assert(source.getLinkAttributes.attrIsSymlink);
-    assert(!source.getAttributes.attrIsSymlink);
+    // symlinking isn't available on Windows
```
CyberShadow: This does not look good in an example of `attrIsSymlink`, which *is* available on Windows.

wilzbach: It's a hotfix to get auto-tester passing again (the examples were added today with #6320, so we still have quite some time to improve them until they are part of the released docs).

Anyhow, I didn't know a better way to show off `attrIsSymlink` in a cross-platform way.
Just `version(Posix) unittest` isn't a good idea either, because then code on dlang.org doesn't work when copy/pasting locally

std/file.d
```diff
@@ -2196,7 +2201,9 @@ if (isConvertibleToString!R)
 
     f.setAttributes(octal!777);
     attributes = f.getAttributes;
-    assert((attributes & 1023) == octal!777);
+
+    version(Posix)
+        assert((attributes & 1023) == octal!777);
```
wilzbach: Again, here I'm not sure what's the best way to represent an easy-to-graph cross-platform example of `setAttributes`, but for now I disabled the failing check.

PR #6328, opened 2018-03-22
Fix Issue 18525: std.algorithm.mutate.remove should work on mutable strings

John-Colvin:


std/algorithm/mutation.d
```diff
@@ -1923,6 +1924,71 @@ if (s == SwapStrategy.stable
...
 
+/// ditto
+// SwapStrategy s is ignored here, it's just always stable,
+// unstable doesn't really work with variable length encoding
+auto remove
+(SwapStrategy s = SwapStrategy.stable, Range, Offset...)
+(Range range, Offset offsets)
+if (isNarrowString!Range
+    && isMutable!(typeof(range[0]))
+    && Offset.length >= 1)
```
wilzbach: Please use `static if` internally - the user doesn't care about the implementation details.

std/algorithm/mutation.d
```diff
@@ -2068,6 +2148,45 @@ if (isBidirectionalRange!Range
     return result;
 }
 
+///ditto
+// SwapStrategy s is ignored here, it's just always stable,
+// unstable doesn't really work with variable length encoding
+Range remove(alias pred, SwapStrategy s = SwapStrategy.stable, Range)
+(Range range)
+if (isNarrowString!Range
+    && isMutable!(typeof(range[0])))
```
wilzbach: Should be done with `static if`

std/algorithm/mutation.d
```diff
@@ -2068,6 +2148,45 @@ if (isBidirectionalRange!Range
...
+        auto doRemove = pred_(decode(range, charIdx));
+        auto encodedLen = charIdx - start;
+        if (doRemove)
+            charShift += encodedLen;
+        else
+            foreach (i; start .. charIdx)
+                range[i - charShift] = range[i];
+    }
+
+    return range[0 .. $ - charShift];
```
wilzbach: this looks very similar to the range-based `remove`. Why can't this be combined?

std/algorithm/mutation.d
```diff
@@ -2012,6 +2078,20 @@ if (s == SwapStrategy.stable
...
+{
+    char[] chars = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h'];
+    remove(chars, 4);
+    assert(chars == ['a', 'b', 'c', 'd', 'f', 'g', 'h', 'h']);
+
+    char[] bigChars = "∑œ∆¬é˚˙ƒé∂ß¡¡".dup;
+    assert(remove(bigChars, tuple(4, 6), 8) == ("∑œ∆¬˙ƒ∂ß¡¡"));
+
+    import std.exception : assertThrown;
+    assertThrown(remove(bigChars.dup, 1, 0));
```
wilzbach: FWIW there's a PR to deprecate this because it's really error-prone: https://github.com/dlang/phobos/pull/6154/files

std/algorithm/mutation.d
John-Colvin: I guess there might be a way to merge them, but the major difference is that for auto-decoded strings you can't just `move` elements around.

std/algorithm/mutation.d
John-Colvin: Ok, makes sense. Whichever one gets merged first will have to adapt to the other.

std/algorithm/mutation.d
John-Colvin: You mean combine with the non-string overloads? I imagine one overload that takes offsets and one that takes a predicate.

wilzbach: > You mean combine with the non-string overloads?

Yep, the rest is implementation detail.

>  I imagine one overload that takes offsets and one that takes a predicate.

Yup, but these should use `static if` internally.

John-Colvin: Given `A(T) if (conditionA!T)` and `B(T) if (conditionB!T)` to merge in to one template, is there a strong preference to do

`A(T) if (conditionA!T || conditionB!T)
{
    static if (conditionA!T) { ... }
    else { static assert(conditionB!T); ... }
}`

or

`A(T)
{
    static if (conditionA!T) { ... }
    else static if (conditionB!T) { ... }
    else static assert(false, "Can't do A with ` ~ T.stringof);
}`

in phobos.

wilzbach: I think there exists no strong guideline on this yet.
However, in the past we went more often with `static assert(0)`due to the better error messages and because it avoids the bulky constraints.
So ideally it's something like:

```d
A(T)
if (simplifiedCommonCondition)
{
    static if (conditionA!T) { ... }
    else static if (conditionB!T) { ... }
    else static assert(false, "Can't do A with ` ~ T.stringof);
}
```

John-Colvin: I've made an attempt at this.

std/algorithm/mutation.d
```diff
@@ -1923,93 +2047,61 @@ if (s == SwapStrategy.stable
...
-    assert(arr == [1,2]);
+        auto mark = charIdx;
+        while (dcharIdx < from + delta)
+            skipOne();
+        charShift += charIdx - mark;
+    }
 
-}
+    while (charIdx < range.length)
+        copyBackOne();
```
n8sh: Here there are no more characters to be removed we can copy the remaining characters without calling `stride`.

std/algorithm/mutation.d
```diff
@@ -1787,13 +1788,143 @@ Returns:
...
+    {
+        static assert(isMutable!(typeof(range[0])),
+                "Elements must be mutable to remove");
+        static assert(s == SwapStrategy.stable,
+                "Only stable removing can be done for character arrays");
+        return removeStableString(range, offset);
+    }
+    else
+    {
+        static assert (isBidirectionalRange!Range,
```
n8sh: The CircleCI style checker won't allow this to be merged until you change this line and line 1807 so there's no space between the `assert` and the `(`.

std/algorithm/mutation.d
John-Colvin: Good spot. Done.

std/algorithm/mutation.d
John-Colvin: done

JackStouffer:
can't review atm but this should be squished before it's merged.

John-Colvin:
@JackStouffer done

std/algorithm/mutation.d
```diff
@@ -1787,13 +1788,143 @@ Returns:
...
-    && hasLength!Range
-    && Offset.length >= 1)
+if (Offset.length >= 1)
+{
+    static if (isNarrowString!Range)
+    {
+        static assert(isMutable!(typeof(range[0])),
+                "Elements must be mutable to remove");
+        static assert(s == SwapStrategy.stable,
+                "Only stable removing can be done for character arrays");
```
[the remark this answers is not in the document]
John-Colvin: The documentation talks about moving elements from the end of the array for `unstable`, which this doesn't do. I'm tempted to take your view that `unstable` is the "relaxed" option, but  it's documented to mean something slightly different here.

Perhaps we could go ahead with this restriction for now and make a bugzilla entry to discuss lifting it?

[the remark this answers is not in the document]
n8sh: >Perhaps we could go ahead with this restriction for now and make a bugzilla entry to discuss lifting it?

I think that makes sense.

PR #6329, opened 2018-03-22
Fix Issue 18648 - Document the type of std.stdio.stdin and friends

wilzbach:
A start.

### DIFF
--- a/std/experimental/all.d
+++ b/std/experimental/all.d
@@ -10,8 +10,8 @@ module std.experimental.all;
 
     int len;
     const r = 6.iota
-              .filter!(a => a % 2) // 0 2 4
-              .map!(a => a * 2) // 0 4 8
+              .filter!(a => a % 2) // 1 3 5
+              .map!(a => a * 2) // 2 6 10
               .tee!(_ => len++)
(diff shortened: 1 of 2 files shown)


std/experimental/all.d
```diff
@@ -10,8 +10,8 @@ module std.experimental.all;
 
     int len;
     const r = 6.iota
-              .filter!(a => a % 2) // 0 2 4
-              .map!(a => a * 2) // 0 4 8
+              .filter!(a => a % 2) // 1 3 5
+              .map!(a => a * 2) // 2 6 10
```
quickfur: What's this all about?

quickfur: Oh nevermind, I see why now.

std/stdio.d
```diff
@@ -4717,18 +4719,75 @@ alias stdin = makeGlobal!(StdFileHandle.stdin);
...
+                  .sum;
+
+        assert(r == 18);
+    }
+}
+
+///
+@safe unittest
+{
+    void main() {
```
JackStouffer: brace style

PR #6334, opened 2018-03-23
Modernize std.demangle

JackStouffer:
* makes `@safe pure nothrow`
* Adds params and returns and enables the dscanner check
* Adds unittest examples
* removes some dead code


std/demangle.d
```diff
@@ -18,72 +18,73 @@
...
+ */
+string demangle(string name) @safe pure nothrow
 {
-    this()
-    {
-        super("MangleException");
-    }
+    import core.demangle : demangle;
+    auto ret = demangle(name);
+    return ret.idup;
```
rainers: What's wrong with `assumeUnique`? ok, the name doesn't fit, but casting to string should be fine.

If you want to keep `idup`, I'd recommend not to make a copy if the name was not demangled (as the documentation "returns its argument" can be interpreted).

std/demangle.d
```diff
@@ -18,72 +18,73 @@
...
+    // int in module a
+    assert(demangle("_D1a1bi") == "int a.b");
+    // char array in module test
+    assert(demangle("_D4test3fooAa") == "char[] test.foo");
+}
 
-int main()
+/**
+This program reads standard in and writes it to standard out,
+pretty-printing any found D mangled names.
```
rainers: I'm not sure this is a good example for the documentation:
* it's style is very much C/D1
* it doesn't handle UTF8 correctly
* when running the example in the browser, there is no output

wilzbach: An "D2-like" example:

```d
@safe unittest
{
      import std.ascii, std.algorithm, std.conv, std.demangle, std.functional, std.stdio;
    "aaa  ...  _D9main4testFifZv"
        .chunkBy!(a => std.ascii.isAlphaNum(a) || a == '_')
        .map!(a => a[1].pipe!(to!string, demangle))
        .joiner
        .writeln;
}
```

"D2-like", because demangle should accept a range and not only string to avoid the `to!string` allocation.

If we still want to support stdin/stdout, that would be sth. like:

```d
void main()
{
    import std.ascii, std.algorithm, std.conv, std.demangle, std.functional, std.stdio;
    stdin.byLineCopy
        .map!(l => l.chunkBy!(a => std.ascii.isAlphaNum(a) || a == '_').map!(a => a[1].pipe!(to!string, demangle)).joiner)
        .copy(stdout.lockingTextWriter);
}
```

std/demangle.d
```diff
@@ -18,72 +18,51 @@
...
 {
-    string buffer;
-    bool inword;
-    int c;
+    import std.ascii;
+    import std.algorithm;
+    import std.conv;
+    import std.demangle;
+    import std.functional;
+    import std.stdio;
```
wilzbach: What's the reason for using individual lines here? The imports are now longer than the actual example :O

JackStouffer: I'm not against it in principal, but the vast majority of code in Phobos uses one import per-line style.

wilzbach: I don't have a strong opinion either, but the vast majority of Phobos code uses selective imports too.
I just checked dlang.org and we use one-line imports for most examples there (except for the ones with selective imports).
That being said, I don't feel strongly about this.

std/demangle.d
```diff
@@ -18,72 +18,51 @@
...
-                buffer.length = 0;
-                buffer ~= cast(char) c;
-            }
-            else
-                write(cast(char) c);
-        }
+        stdin.byLineCopy
+            .map!(
+                l => l.chunkBy!(a => std.ascii.isAlphaNum(a) || a == '_')
+                    .map!(a => a[1].pipe!(to!string, demangle)).joiner
```
wilzbach: Isn't our style guide is to align on the UFCS dot?

.dscanner.ini
```diff
@@ -219,7 +218,7 @@ object_const_check="-std.algorithm.searching,-std.array,-std.bitmanip,-std.concu
...
...egion,-std.experimental.allocator.building_blocks.scoped_allocator,-std.experimental.allocator.building_blocks.segregator,-std.experimental.allocator.building_blocks.stats_collector,-std.experimental.allocator.common,-std.experimental.allocator.gc_allocator,-std.experimental.allocator.mallocator,-std.experimental.allocator.showcase,-std.experimental.allocator.typed,-std.experimental.checkedint,-std.experimental.logger.core,-std.experimental.logger.filelogger,-std.file,-std.format,-std.functional,-std.getopt,-std.internal.cstring,-std.internal.math.biguintcore,-std.internal.math.biguintnoasm,-std.internal.scopebuffer,-std.internal.test.dummyrange,-std.internal.unicode_tables,-std.json,-std.math,-std.mathspecial,-std.meta,-std.mmfile,-std.net.curl,-std.net.isemail,-std.numeric,-std.outbuffer,-std.parallelism,-std.path,-std.random,-std.range,-std.range.interfaces,-std.range.primitives,-std.regex,-std.regex.internal.backtracking,-std.regex.internal.ir,-std.regex.internal.kickstart,-std.regex.internal.parser,-std.regex.internal.thompson,-std.signals,-std.socket,-std.stdio,-std.string,-std.traits,-std.typecons,-std.uni,-std.uri,-std.utf,-std.uuid,-std.variant,-std.xml,-std.zip,-std.zlib"
```
wilzbach: FYI: Once this is merged, we should will split this in a multi-line statement. I'm happy to make the PR for it though.

PR #6335, opened 2018-03-23
Suggest using cache before tee.

Hackerpilot:


std/algorithm/iteration.d
```diff
@@ -128,7 +128,7 @@ rather than re-evaluated.
 This can be a useful function to place in a chain, after functions
 that have expensive evaluation, as a lazy alternative to $(REF array, std,array).
 In particular, it can be placed after a call to `map`, or before a call
-to `filter`.
+to `filter` or `tee`.
```
wilzbach: Why not use references while you are at it?

> to $(REF filter, std,_range) or $(REF tee, std,_range)

PR #6336, opened 2018-03-24
std.math: Remove special cases that attempt to raise FE overflow/underflow flag

ibuclaw:
Because they now no longer work, and to be honest they shouldn't be force raising a floating point exception anyway, as that arguably violates `pure`.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -1834,24 +1834,13 @@ real exp(real x) @trusted pure nothrow @nogc
         else
             static assert(0, "Not implemented for this architecture");
 
-        // Special cases. Raises an overflow or underflow flag accordingly,
-        // except in the case for CTFE, where there are no hardware controls.
+        // Special cases.
         if (isNaN(x))
             return x;
         if (x > OF)
-        {
-            if (__ctfe)
-                return real.infinity;
-            else
-                return real.max * copysign(real.max, real.infinity);
-        }
+            return real.infinity;
         if (x < UF)
-        {
-            if (__ctfe)
-                return 0.0;
-            else
-                return real.min_normal * copysign(real.min_normal, 0.0);
-        }
+            return 0.0;
 
         // Express: e^^x = e^^g * 2^^n
         //   = e^^g * e^^(n * LOG2E)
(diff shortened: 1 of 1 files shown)


kinke:
My findings with LDC are that the over-/underflow bits are set correctly for the x87 `real` implementation in inline assembly. The Cephes versions also do unless optimizations are enabled (but for DMD, they are still set). The MS C runtime function doesn't seem to set them.

n8sh:
>Because they now no longer work, and to be honest they shouldn't be force raising a floating point exception anyway, as that arguably violates `pure`.

I am given to understand that the specification explicitly allows this.

https://dlang.org/spec/function.html#pure-functions

>8. As a concession to practicality, a pure function can also:
> * read and write the floating point exception flags
> * read and write the floating point mode flags, as long as those flags are restored to their initial state > upon function entry
> *  perform impure operations in statements that are in a ConditionalStatement controlled by a DebugCondition.

ibuclaw:
> The Cephes single/double-precision versions also do unless optimizations are enabled

Right, unoptimized unittest run passes, but optimized release library has a different behaviour. And so it is in everyone's best interest to build unittest with at least `-O2`.g

ibuclaw:
> I am given to understand that the specification explicitly allows this.

Then you should probably approve and merge #6259. Then we'll mark the FP exception setters as pure, then we'll add a `raiseException` function which implements this properly.

In any case, we only have FP flag setting for these three functions, and only check that one of them "kind of"  works. This is missing in so many other functions in std.math (eg: floor doesn't raise FE_INEXACT) that exp/expm1 are the exception and not the rule. ;-)

n8sh:
>Then you should probably approve and merge #6259. 

Would that I could, but I lack that power.

andralex:
@n8sh check again :)

JackStouffer:
Ping @kinke @n8sh. Does this look good?

JackStouffer:
This is a breaking change, but since this was never documented behavior I'm going to move forward with this.

ibuclaw:
If we have a raiseException that sets FP exception flags in a way that is `pure` (and according to the spec it is allowed) then these can be re-added and deterministic, as it won't be at the mercy of the optimizer.

PR #6338, opened 2018-03-25
std.typecons: Fix a -dip1000 compilable issue

carblue:
https://dlang.org/phobos/std_typecons.html#Proxy
"template Proxy(alias a)
Creates a proxy for the value a that will forward all operations while disabling implicit conversions. The aliased item a must be an lvalue."

There is this unittest in std/typecons.d:
```
@system unittest                  <== line 6579
{
    // bug14213, using function for the payload
    static struct S
    {
        int foo() { return 12; }
        mixin Proxy!foo;          <== line 6585
    }
    static class C
    {
        int foo() { return 12; }
        mixin Proxy!foo;
    }
    S s;
    assert(s + 1 == 13);
    C c = new C();
    assert(s * 2 == 24);
}
```

I don't understanf, why a non-lvalue returning function is used/(usable at all) here as template argument, but it compiles for S and C with -dip25.
With -dip1000 it compiles for C only.
Strange.
This quasi-fix just comments-out the offending lines to be -dip1000 compilable, but as said, without understanding much of what the implications are.

aa[std.typecons]=-dip1000
Errors when running: make -f posix.mak std/typecons.test
...
```
    ../dmd/generated/linux/release/64/dmd -od$T -conf= -I../druntime/import  -w -de -dip25 -m64 -fPIC -transition=complex -O -release -dip1000  -main -unittest generated/linux/release/64/libphobos2.a -defaultlib= -debuglib= -L-ldl -cov -run std/typecons.d ;     \
  )
std/typecons.d(5953): Error: circular typeof definition
std/typecons.d(5953): Error: alias `std.typecons.__unittest_L6579_C9.S.Proxy!(foo).ValueType` recursive alias declaration
std/typecons.d(6585): Error: mixin `std.typecons.__unittest_L6579_C9.S.Proxy!(foo)` error instantiating
```

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -6829,6 +6829,8 @@ mixin template Proxy(alias a)
     bool* b = Name("a") in names;
 }
 
+// excludes struct S; it's 'mixin Proxy!foo' doesn't compile with -dip1000
+version(DIP1000) {} else
 @system unittest
 {
     // bug14213, using function for the payload
@@ -6837,15 +6839,19 @@ mixin template Proxy(alias a)
         int foo() { return 12; }
         mixin Proxy!foo;
     }
+    S s;
+    assert(s + 1 == 13);
+    assert(s * 2 == 24);
+}
+
+@system unittest
+{
     static class C
     {
         int foo() { return 12; }
         mixin Proxy!foo;
     }
-    S s;
-    assert(s + 1 == 13);
     C c = new C();
-    assert(s * 2 == 24);
 }
 
 // Check all floating point comparisons for both Proxy and Typedef,
--- a/dip1000.mak
+++ b/dip1000.mak
@@ -46,7 +46,7 @@ aa[std.stdio]=-dip25 #    TODO
 aa[std.string]=-dip1000
 aa[std.system]=-dip1000
 aa[std.traits]=-dip1000
-aa[std.typecons]=-dip25 # cannot call @system function std.format.formattedWrite!(Appender!string, char, Nullable!int)
+aa[std.typecons]=-dip1000 -version=DIP1000 # merged https://github.com/dlang/phobos/pull/6338; COMPROMISE: check the reason for non-dip1000: static struct S. mixin Proxy!foo;
 aa[std.typetuple]=-dip1000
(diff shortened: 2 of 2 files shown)


std/typecons.d
```diff
@@ -6582,17 +6582,17 @@ mixin template Proxy(alias a)
...
     }
     static class C
     {
         int foo() { return 12; }
         mixin Proxy!foo;
     }
-    S s;
-    assert(s + 1 == 13);
+//    S s;
+//    assert(s + 1 == 13);
```
wilzbach: This removes the test for the normal code path too, so if we really decide we don't want to test this for DIP1000, I think it should be sth. like:

```
unittest
{
    static class C
...
}

version(DIP1000) {} else
unittest
{
   static struct S
}
```

WalterBright: There isn't a standard DIP1000 version (perhaps there should be), but in any case the reasons these tests are commented out should be made clear in the code.

std/typecons.d
```diff
@@ -6829,6 +6829,18 @@ mixin template Proxy(alias a)
...
+@system unittest
+{
+    // bug14213, using function for the payload
+    static class C
+    {
+        int foo() { return 12; }
+        mixin Proxy!foo;
+    }
+    C c = new C();
+}
```
wilzbach: The duplication wouldn't have been necessary:

```d
@system unittest
{
    // bug14213, using function for the payload
    static struct S
    {
        int foo() { return 12; }
        mixin Proxy!foo;
    }
    S s;
    assert(s + 1 == 13);
    assert(s * 2 == 24);
}

version(DIP1000) {} else
@system unittest
{
    static class C
    {
        int foo() { return 12; }
        mixin Proxy!foo;
    }
    C c = new C();
}

carblue: @wilzbach Yes, I was aware of the duplicated class C: My idea was to simply delete my original version(DIP1000) addition again, once the reason for failing struct S mixin Proxy!foo is found.

wilzbach: Deleting `version(DIP1000) {} else` is even easier ;-)

PR #6339, opened 2018-03-25
[WIP] Explore string interpolation library

marler8997:
A prototype implementation for interpolated strings to explore their viability as a library.  Comments/suggestions welcome.

### DIFF
--- a/std/experimental/interpolate.d
+++ b/std/experimental/interpolate.d
@@ -1,0 +1,230 @@
+//
+//
+// TODO: move into std/experimental/typecons.d
+//
+//
+
+/**
+Interpolates the given string into a code that creates a tuple of string literals and expressions.
+The code returned should be passed to `mixin` in order to create the tuple.
+
+Example:
+---
+int a = 42;
+
+writeln(mixin(interp("a is $(a)")));
+// same as: writeln(mixin(`std.typecons.tuple("a is ", a).expand`));
+
+// Output:
+// a is 42
+---
+*/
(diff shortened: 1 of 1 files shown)


std/experimental/interpolate.d
```diff
@@ -0,0 +1,222 @@
...
+writeln(mixin(interp("a is $(a)")));
+// same as: writeln(mixin(`tuple("a is ", a).expand`));
+
+// Output:
+// a is 42
+---
+*/
+string interp(string str, string file = __FILE__, size_t line = __LINE__) pure @safe
+{
+    return "tuple(" ~ interpToCommaExpression(str, file, line) ~ ").expand";
```
ntrel: I think we could avoid the public import by (essentially) doing this:
```d
return q{() @property {import std.typecons : tuple; return tuple(...);}};
```

marler8997: great idea...changed

marler8997:
Cleaning up old PRs that aren't going anywhere

Geod24:
@marler8997 : Could you possibly turn this into a dub package ?

marler8997:
Yeah that's possible, the purpose of it was a proof of concept to see if interpolated strings should be implemented as a library or as part of the language.  If it works well as a library then we could make a dub package.  It has some problems though, for example if we want to support arbitrary expressions inside the string then we really need to include the DMD lexer...another reason why it would make sense to put it in the compiler instead of as a library.

PR #6340, opened 2018-03-25
std.regex.internal.tests: Fix a -dip1000 compilable issue

carblue:
aa[std.regex.internal.tests]=-dip1000
Error when running: make -f posix.mak std/regex/internal/tests.test
```
std/regex/internal/tests.d(463): Error: @safe function std.regex.internal.tests.__unittest_L55_C7 cannot call @system function std.regex.internal.tests.__unittest_L55_C7.ct_tests
```

### DIFF
--- a/std/regex/internal/tests.d
+++ b/std/regex/internal/tests.d
@@ -10,6 +10,8 @@ import std.conv, std.exception, std.meta, std.range,
 
 import std.uni : Escapables; // characters that need escaping
 
+debug(std_regex_test) import std.stdio;
+
 @safe unittest
 {//sanity checks
     regex("(a|b)*");
@@ -449,11 +451,9 @@ import std.uni : Escapables; // characters that need escaping
                     v ,": ", tvd.pattern));
                 if (c == 'y')
                 {
-                    import std.stdio;
                     auto result = produceExpected(m, tvd.format);
-                    if (result != tvd.replace)
-                        writeln("ctRegex mismatch pattern #", v, ": ", tvd.pattern," expected: ",
-                                tvd.replace, " vs ", result);
+                    assert(result == tvd.replace, text("ctRegex mismatch pattern #", v,
+                        ": ", tvd.pattern," expected: ", tvd.replace, " vs ", result));
                 }
             }
         }}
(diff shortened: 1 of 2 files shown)


std/regex/internal/tests.d
```diff
@@ -452,7 +452,7 @@ import std.uni : Escapables; // characters that need escaping
                     import std.stdio;
```
wilzbach: That is optional now. Though it looks like it's the only `std.stdio` import in this module even though there are six more `debug(std_regex_test) writeln`, so maybe put `debug(std_regex_test) import std.stdio;`at the module level?

DmitryOlshansky: I think CT testsuit is horribly slow to build so I wanted to print all errors originally. Make it an assert then to avoid getting bugs silently.

carblue:
@wilzbach 
It's up to You how to assess the -dip setting:
Being absolutely strict, it's stll -dip25, but as -debug=std_regex_test isn't used (except setting that manually), in practice it may be seen as -dip1000

carblue:
Oh, I didn't read Dmitry's request for changes when commiting; I'll have a look at that first

std/regex/internal/tests.d
```diff
@@ -449,11 +451,11 @@ import std.uni : Escapables; // characters that need escaping
                     v ,": ", tvd.pattern));
                 if (c == 'y')
                 {
-                    import std.stdio;
                     auto result = produceExpected(m, tvd.format);
-                    if (result != tvd.replace)
-                        writeln("ctRegex mismatch pattern #", v, ": ", tvd.pattern," expected: ",
+                    if (result != tvd.replace) {
```
wilzbach: CI complains about K&R style -> Allman:

```
Enforce Allman style
grep -nrE '(if|for|foreach|foreach_reverse|while|unittest|switch|else|version) .*{$' $(find etc std -name '*.d'); test $? -eq 1
std/regex/internal/tests.d:455:                    if (result != tvd.replace) {
posix.mak:544: recipe for target 'style_lint' failed
```

wilzbach:
`debug` is allowed to break out of `@nogc`, `@safe`,`pure` and `nothrow` - see https://github.com/dlang/dlang.org/pull/2209

PR #6344, opened 2018-03-25
Make more unittests in std.bitarray public

wilzbach:
There are already many good examples existing in std.bitmanip. They just need to the see light of the D world.

### DIFF
--- a/std/bitmanip.d
+++ b/std/bitmanip.d
@@ -245,52 +245,73 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
 Allows creating bit fields inside $(D_PARAM struct)s and $(D_PARAM
 class)es.
 
-Example:
-
-----
-struct A
-{
-    int a;
-    mixin(bitfields!(
-        uint, "x",    2,
-        int,  "y",    3,
-        uint, "z",    2,
-        bool, "flag", 1));
-}
-A obj;
-obj.x = 2;
-obj.z = obj.x;
-----
-
-The example above creates a bitfield pack of eight bits, which fit in
-one $(D_PARAM ubyte). The bitfields are allocated starting from the
-least significant bit, i.e. x occupies the two least significant bits
-of the bitfields storage.
-
-The sum of all bit lengths in one $(D_PARAM bitfield) instantiation
-must be exactly 8, 16, 32, or 64. If padding is needed, just allocate
-one bitfield with an empty name.
-
-Example:
-
-----
-struct A
-{
-    mixin(bitfields!(
-        bool, "flag1",    1,
-        bool, "flag2",    1,
-        uint, "",         6));
-}
-----
-
 The type of a bit field can be any integral type or enumerated
 type. The most efficient type to store in bitfields is $(D_PARAM
 bool), followed by unsigned types, followed by signed types.
+
+See_Also: $(REF BitFlags, std,typecons)
 */
 
 template bitfields(T...)
 {
     enum { bitfields = createStorageAndFields!T.result }
+}
+
+/**
+Create a bitfield pack of eight bits, which fit in
+one $(D_PARAM ubyte). The bitfields are allocated starting from the
+least significant bit, i.e. x occupies the two least significant bits
+of the bitfields storage.
+*/
+@safe unittest
+{
+    struct A
+    {
+        int a;
+        mixin(bitfields!(
+            uint, "x",    2,
+            int,  "y",    3,
+            uint, "z",    2,
+            bool, "flag", 1));
+    }
+
+    A obj;
+    obj.x = 2;
+    obj.z = obj.x;
+
+    assert(obj.x == 2);
+    assert(obj.y == 0);
+    assert(obj.z == 2);
+}
+
+/**
+The sum of all bit lengths in one $(D_PARAM bitfield) instantiation
+must be exactly 8, 16, 32, or 64. If padding is needed, just allocate
+one bitfield with an empty name.
+*/
+@safe unittest
+{
+    struct A
+    {
+        mixin(bitfields!(
+            bool, "flag1",    1,
+            bool, "flag2",    1,
+            uint, "",         6));
+    }
+
+}
+
+/// enums can be used too
+@safe unittest
+{
+    enum ABC { A, B, C }
+    struct EnumTest
+    {
+        mixin(bitfields!(
+                  ABC, "x", 2,
+                  bool, "y", 1,
+                  ubyte, "z", 5));
+    }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


wilzbach:
Interesting - DScanner just segfaulted while trying to iterate through directories:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff6fa6fb6 in __readdir (dirp=0x0) at ../sysdeps/posix/readdir.c:44
44	../sysdeps/posix/readdir.c: No such file or directory.
#0  0x00007ffff6fa6fb6 in __readdir (dirp=0x0) at ../sysdeps/posix/readdir.c:44
#1  0x0000000000a89169 in std.file.DirIteratorImpl.next() (this=...) at std/file.d:4029
#2  0x0000000000a89628 in std.file.DirIteratorImpl.popFront() (this=...) at std/file.d:4133
#3  0x0000000000a898e9 in std.file.DirIterator.popFront() (this=...) at std/file.d:4155
#4  0x00000000008c8da1 in dsymbol.modulecache.ModuleCache.addImportPaths() (this=0x7ffff7ee9530, root=...) at dsymbol/src/dsymbol/modulecache.d:133
#5  0x00000000008c8d95 in dsymbol.modulecache.ModuleCache.addImportPaths() (this=0x7ffff7ee9530, root=...) at dsymbol/src/dsymbol/modulecache.d:132
#6  0x00000000008c8d95 in dsymbol.modulecache.ModuleCache.addImportPaths() (this=0x7ffff7ee9530, root=...) at dsymbol/src/dsymbol/modulecache.d:132
#7  0x00000000008c8bcb in dsymbol.modulecache.ModuleCache.addImportPaths() (this=..., paths=...) at dsymbol/src/dsymbol/modulecache.d:136
#8  0x0000000000a565ed in D main (args=...) at src/main.d:151
#9  0x0000000000a75784 in rt.dmain2._d_run_main() (this=0x7fffffffe3e0) at src/rt/dmain2.d:508
#10 0x0000000000a755cc in rt.dmain2._d_run_main() (this=0x7fffffffe3e0, dg=...) at src/rt/dmain2.d:469
#11 0x0000000000a756d9 in rt.dmain2._d_run_main() (this=0x7fffffffe3e0) at src/rt/dmain2.d:508
#12 0x0000000000a755cc in rt.dmain2._d_run_main() (this=0x7fffffffe3e0, dg=...) at src/rt/dmain2.d:469
#13 0x0000000000a75530 in _d_run_main (argc=7, argv=0x7fffffffe4e8, mainFunc=0xa56010 <D main>) at src/rt/dmain2.d:528
#14 0x0000000000a63a0e in main ()
```

https://github.com/dlang/phobos/blob/4c4c37d0b464df05e2f5298d4ddc9b6a9f16bb1b/std/file.d#L4029

std/bitmanip.d
```diff
@@ -718,32 +739,26 @@ struct DoubleRep
     enum uint bias = 1023, signBits = 1, fractionBits = 52, exponentBits = 11;
 }
 
+/// test reading
```
JackStouffer: I would just say "Reading", as the doc user doesn't know this is a test.

std/bitmanip.d
```diff
@@ -1216,6 +1236,7 @@ public:
         return this;
     }
 
+    ///
     @system unittest
     {
         BitArray b;
```
JackStouffer: should use foreach, or just simplify to use `==` with another BitArray

std/bitmanip.d
```diff
@@ -1278,6 +1299,7 @@ public:
         return this;
     }
 
+    ///
     @system unittest
     {
         __gshared size_t x = 0b1100011000;
```
JackStouffer: remove __gshared

std/bitmanip.d
```diff
@@ -1536,6 +1564,7 @@ public:
         return _ptr[0 .. dim];
     }
 
+    ///
     @system unittest
```
JackStouffer: the opCast overload is supposed to show casting to `sizet[]`, but it shows casting to `void[]`

std/bitmanip.d
```diff
@@ -1566,6 +1595,7 @@ public:
         return result;
     }
 
+    ///
     @system unittest
```
JackStouffer: What on Earth is `opCom` and why is it associated with `~`?

std/bitmanip.d
```diff
@@ -852,6 +867,7 @@ public:
         return cast(bool) bt(_ptr, i);
     }
 
+    ///
     @system unittest
     {
         void Fun(const BitArray arr)
```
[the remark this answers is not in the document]
wilzbach: No need for `static` here. `BitArray` isn't templated,  so `void Fun` is only instantiated once.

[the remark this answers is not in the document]
JackStouffer: it was more about the function pointer to the unittest context and the name `Fun` vs `fun`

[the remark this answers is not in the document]
wilzbach: Changed.

std/bitmanip.d
wilzbach: Seems to be a real thing: https://github.com/dlang/dmd/blob/25bc0213a7455b8ccbb9a4cc869c86ecc24c6777/src/dmd/id.d#L205

(and it is called from the tests)

wilzbach: Looks like DMD supports explicitly overloading ~ with `opCom`:

https://github.com/dlang/dmd/blob/25bc0213a7455b8ccbb9a4cc869c86ecc24c6777/src/dmd/expression.d#L5756

wilzbach: See also: https://issues.dlang.org/show_bug.cgi?id=18665

std/bitmanip.d
```diff
@@ -1536,20 +1561,22 @@ public:
...
-        assert(v.length == a.dim * size_t.sizeof);
+        assert(v.length == a.length / size_t.sizeof);
+        assert(v.length == a.dim);
     }
 
     /***************************************
      * Support for unary operator ~ for $(D BitArray).
      */
-    BitArray opCom() const pure nothrow
+    BitArray opUnary(string op)() const pure nothrow
```
JackStouffer: Unfortunately we can't just remove `opCom` if we can avoid it. If possible with the current implementation of overloading in DMD, we should have both `opCom` and `opUnary` side by side for one release per the proposed deprecation process, and then deprecate `opCom`.

wilzbach:
(passing the CIs again)

wilzbach:
Because I initially thought it's a good idea that the documentation updates appear soon on dlang.org and not in May. Sorry about this - I think all other PRs targeted `master`.

PR #6345, opened 2018-03-25
Automatically clone dmd+druntime if missing

wilzbach:
This shrinks the setup of a D development environment down to:

```sh
git clone https://github.com/dlang/phobos
make -f posix.mak AUTO_BOOTSTRAP=1
```

We use the same logic at dlang.org for quite a while now:
https://github.com/dlang/dlang.org/blob/c88c08ab28c145c912cd04e9298c63740e492b18/posix.mak#L159

CC @CyberShadow

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -37,6 +37,11 @@ QUIET:=@
 DEBUGGER=gdb
 GIT_HOME=https://github.com/dlang
 DMD_DIR=../dmd
+DRUNTIME_PATH = ../druntime
+
+# Auto-cloning missing directories
+$(shell [ ! -d $(DMD_DIR) ] && git clone ${GIT_HOME}/dmd $(DMD_DIR))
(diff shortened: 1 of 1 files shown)


JackStouffer:
Where does the `AUTO_BOOTSTRAP` come into play? It looks like you're downloading the repos regardless of the flags set.

wilzbach:
It's for building the DMD on a system that doesn't have a D compiler installed:

https://github.com/dlang/dmd/blob/master/src/posix.mak#L127

wilzbach:
Ping @CyberShadow - seems like no one except for you is brave enough to approve this. Would you mind?

CyberShadow:
I'm kind of a bit worried that all this convenience stuff will result in growing complexity, and doing too many things on behalf of the user will make it harder for them to diagnose problems when something breaks

wilzbach:
Fair enough. Closing then.

PR #6346, opened 2018-03-25
make `group`, `chain`, and `choose` compatible with `RefRange`

aG0aep6G:
The first three of probably many functions that need to be fixed in order to resolve [issue 18657](https://issues.dlang.org/show_bug.cgi?id=18657).

Related forum discussion: <https://forum.dlang.org/post/p96gs4$22fp$1@digitalmars.com>.

I'm not sure if this is the right direction. The alternative would be to add assignment to the range interface, and remove `opAssign` from `RefRange`. That would be less busywork, and being able to use assignment with ranges might be better for everyone's sanity. But it would also be more of a breaking change.

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -1684,9 +1684,12 @@ if (isInputRange!R)
     static if (isForwardRange!R)
     {
         ///
-        @property typeof(this) save() {
+        @property typeof(this) save()
+        {
+            import std.algorithm.mutation : move;
             typeof(this) ret = this;
-            ret._input = this._input.save;
+            auto saved = this._input.save;
+            move(saved, ret._input);
             ret._current = this._current;
             return ret;
         }
@@ -1805,6 +1808,15 @@ if (isInputRange!R)
     assert(r.equal([ tuple(5, 1u) ]));
     assert(s.equal([ tuple(4, 3u), tuple(5, 1u) ]));
     assert(t.equal([ tuple(3, 1u), tuple(4, 3u), tuple(5, 1u) ]));
+}
+
+pure @safe unittest // issue 18657
+{
+    import std.algorithm.comparison : equal;
+    import std.range : refRange;
+    auto r = refRange(&["foo"][0]).group;
+    assert(equal(r.save, "foo".group));
+    assert(equal(r, "foo".group));
 }
 
 // Used by implementation of chunkBy for non-forward input ranges.
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -1524,7 +1524,11 @@ private struct ChooseResult(R1, R2)
         @property auto save()
         {
             auto result = this;
-            actOnChosen!((ref r) { r = r.save; })(result);
+            actOnChosen!((ref r) {
+                import std.algorithm.mutation : move;
+                auto saved = r.save;
+                move(saved, r);
+            })(result);
             return result;
         }
 
@@ -1610,6 +1614,14 @@ private struct ChooseResult(R1, R2)
                         return choose(false, Slice1.init, r[begin .. end]);
                 })(this, begin, end);
         }
+}
+
+pure @safe unittest // issue 18657
+{
+    import std.algorithm.comparison : equal;
+    auto r = choose(true, refRange(&["foo"][0]), "bar");
+    assert(equal(r.save, "foo"));
+    assert(equal(r, "foo"));
 }
 
 /**
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -972,10 +972,12 @@ if (Ranges.length > 0 &&
             static if (allSatisfy!(isForwardRange, R))
                 @property auto save()
                 {
+                    import std.algorithm.mutation : move;
                     typeof(this) result = this;
                     foreach (i, Unused; R)
                     {
-                        result.source[i] = result.source[i].save;
+                        auto saved = result.source[i].save;
+                        move(saved, result.source[i]);
                     }
                     return result;
                 }
@@ -1378,6 +1380,14 @@ pure @safe nothrow @nogc unittest
     immutable(Foo)[] a;
     immutable(Foo)[] b;
     assert(chain(a, b).empty);
+}
+
+pure @safe unittest // issue 18657
+{
+    import std.algorithm.comparison : equal;
+    auto r = refRange(&["foo"][0]).chain("bar");
+    assert(equal(r.save, "foobar"));
+    assert(equal(r, "foobar"));
 }
 
 /**


std/algorithm/iteration.d
```diff
@@ -1502,9 +1502,12 @@ if (isInputRange!R)
...
     {
         ///
-        @property typeof(this) save() {
+        @property typeof(this) save()
+        {
+            import std.algorithm.mutation : move;
             typeof(this) ret = this;
-            ret._input = this._input.save;
+            auto saved = this._input.save;
+            move(saved, ret._input);
```
n8sh: >I'm not sure if this is the right direction.

I'm not sure either. If we're doing it this way though we could at least put this in a branch that checks for `hasElaborateAssign` and otherwise use the normal behavior.

aG0aep6G: > we could at least put this in a branch that checks for hasElaborateAssign and otherwise use the normal behavior.

What would that accomplish? It looks like `move` boils down to `target = source;` anyway when there's nothing special to be done.

https://github.com/dlang/phobos/blob/5b5e5281a88d24fd31fa316c17850c960ae41664/std/algorithm/mutation.d#L1376-L1379

n8sh: >What would that accomplish?

It avoids unnecessary function calls and unnecessary data copying and potentially reduces template bloat. Compare the generated assembly for `aaaCopyImplAAA` and `bbbCopyImplBBB` with compile options "-O -inline": https://run.dlang.io/is/1aofWf
```d
struct Inner
{
    uint meaninglessField;
    @property typeof(this) save() { return this; }
}

struct ExampleStruct
{
    Inner exampleField;
    @property typeof(this) aaaCopyImplAAA()
    {
        typeof(this) ret = this;
        ret.exampleField = this.exampleField.save;
        return ret;
    }
    @property typeof(this) bbbCopyImplBBB()
    {
        import std.algorithm.mutation : move;
        typeof(this) ret = this;
        auto saved = this.exampleField.save;
        move(saved, ret.exampleField);
        return ret;
    }        
}

void main()
{
}
```
aaaCopyImplAAA:
```
onlineapp.ExampleStruct.aaaCopyImplAAA():
		push	RBP
		mov	RBP,RSP
		mov	EAX,[RDI]
		pop	RBP
		ret
```
bbbCopyImplBBB:
```
onlineapp.ExampleStruct.bbbCopyImplBBB():
		push	RBP
		mov	RBP,RSP
		sub	RSP,010h
		mov	EAX,[RDI]
		mov	-8[RBP],EAX
		mov	ECX,[RDI]
		mov	-4[RBP],ECX
		lea	RSI,-4[RBP]
		lea	RDI,-8[RBP]
		call	  pure nothrow @nogc void std.algorithm.mutation.moveImpl!(onlineapp.Inner).moveImpl(ref onlineapp.Inner, ref onlineapp.Inner)@PLT32
		mov	EAX,-8[RBP]
		mov	RSP,RBP
		pop	RBP
		ret
		add	[RCX],AH

pure nothrow @nogc void std.algorithm.mutation.moveImpl!(onlineapp.Inner).moveImpl(ref onlineapp.Inner, ref onlineapp.Inner):
		push	RBP
		mov	RBP,RSP
		sub	RSP,010h
		mov	-010h[RBP],RBX
		mov	RCX,RDI
		mov	RDX,RSI
		cmp	RDX,RCX
		jne	L20
		mov	RBX,-010h[RBP]
		mov	RSP,RBP
		pop	RBP
		ret
L20:		cmp	RDX,RCX
		je	L32
		mov	EBX,[RDX]
		mov	[RCX],EBX
		mov	RBX,-010h[RBP]
		mov	RSP,RBP
		pop	RBP
		ret
L32:		mov	R8D,055Fh
		lea	RCX,_TMP0@PC32[RIP]
		mov	EAX,040h
		mov	RDX,RAX
		mov	-8[RBP],RDX
		lea	RDX,_TMP0@PC32[RIP]
		mov	EDI,027h
		mov	RSI,RDX
		mov	RDX,-8[RBP]
		call	  _d_assert_msg@PLT32
		mov	RBX,-010h[RBP]
		mov	RSP,RBP
		pop	RBP
		ret
```

JackStouffer: Always compare ASM from LDC and never from DMD unless you're working on DMD's backend. DMD's optimizer is out of date, and therefore should not be catered to.

quickfur: Yes, please don't "optimize" code based on what dmd does.  Due to dmd's inferior optimizer, this tends to greatly uglify code needlessly, where ldc/gdc would produce code better than dmd anyway, even when written in the original "unoptimized" form.

We should rather improve dmd's optimizer than make Phobos a rats' nest of workarounds for it.

n8sh: Gotcha, and LDC handles this example nicely. (The output is still littered with unused functions produced by template expansion, but I believe that's being actively worked on.)

wilzbach: > The output is still littered with unused functions produced by template expansion

`-L--gc-sections` should be able to remove unused sections.
(see e.g. https://sourceware.org/binutils/docs/ld/Options.html)

aG0aep6G:
This has been open for three weeks, and the only discussion has been about optimizing for DMD or not (consensus: not). Ping?

std/algorithm/iteration.d
```diff
@@ -1502,9 +1502,12 @@ if (isInputRange!R)
...
             ret._current = this._current;
```
JackStouffer: shouldn't this line also invoke refRange's opAssign?

aG0aep6G: > shouldn't this line also invoke refRange's opAssign?

`_current` is the `front` element. It's not the range itself.

aG0aep6G:
> In my opinion, it would be better to update the range spec and go from there.

I wouldn't mind that at all.

MartinNowak:
> This has been open for three weeks, and the only discussion has been about optimizing for DMD or not (consensus: not). Ping?

This is because this is a complex and deep rooted topic that needs proper understanding, also related to unsupported unique ownership ranges.
FWIW using copy construction + move to avoid a broken opAssign is a hack that might bite us in the long-run. It's also problematic to leak such workarounds into the standard library, where people look up how to implement certain behaviors.
Whether the design issue is within refRange, the algorithms, the range spec, or the language someone still needs to exactly point out. And btw, it's the purpose of refRange to mutate the original range, so maybe it's not even reasonable to implement `save`.

Thanks for your work on this fix, but I think a deeper understanding of the problem would be more appropriate than simply fixing a few symptoms.

JackStouffer:
I've reconsidered based on @jmdavis's comments.

I think the option that will cause the least pain will keeping in the spirt of what `RefRange` is supposed to do, is to remove `RefRange`'s forward range capabilities.

The range semantics still need to be defined, but that doesn't solve the issue of breaking something about `RefRange`. By removing `save` we keep the current intended behavior and remove the buggy behavior.

aG0aep6G:
> I think the option that will cause the least pain will keeping in the spirt of what RefRange is supposed to do, is to remove RefRange's forward range capabilities.

Maybe. I'm not so sure if the original goal of RefRange is worthwhile, though.

What the opAssign achieves is that the RefRange binds more tightly to its target than a pointer or class reference would. It's more like a `ref` parameter, which also cannot be rebound.

I think this behavior is actually somewhat surprising. I'd expect `refRange(&foo)` to work like `&foo`, just with the necessary adjustments to make it a proper range itself. I've had uses for that part of RefRange. I don't think I ever had a use for the stronger non-rebindable aspect. But it's of course possible that I just wasn't aware of relying on it.

Example:

```d
import std.range: refRange;
import std.stdio: writeln;

void main()
{
    string s = "foo";
    
    auto p = &s;
    auto r = refRange(&s);
    
    auto p2 = p;
    auto r2 = r;
    
    p2 = ["bar"].ptr; /* Rebind p2. */
    writeln(*p); /* Still "foo". */
    
    r2 = refRange(["bar"].ptr); /* Can't actually rebind r2. */
    writeln(r); /* Surprisingly(?), it's "bar" now. */
}
```

> The range semantics still need to be defined, but that doesn't solve the issue of breaking something about RefRange. By removing save we keep the current intended behavior and remove the buggy behavior.

You also lose the intended behavior of RefRange being a forward range. That's ok if it was a mistake in the first place, but does it buy us anything beyond that? Seems mostly a separate issue to me.

If non-surprising assignment becomes a requirement for ranges, we have to fix RefRange in that regard. And if funky opAssign overloads get deliberately allowed,  generic code can't assume anything about assignment, and we have to do what this PR does. In both cases, it doesn't really matter if RefRange has `save` or not.

One scenario where `RefRange.save` would become relevant is if we say that forward ranges must have non-surprising assignment, but input ranges don't have to. Doesn't really make sense to me, but that's the one case I can think of where removing `save` from RefRange would have a meaningful impact.

MartinNowak:
> I'm not so sure if the original goal of RefRange is worthwhile, though.

My point was, `refRange` is a very important primitive, in particular with unique ownership to circumvent the copy vs. move problem. It's prolly also in widespread use.
If we need to fix sth. about this then we need to consider the whole ecosystem, including fallout, future directions, and learning hurdle/complexity.

> I'd expect refRange(&foo) to work like &foo, just with the necessary adjustments to make it a proper range itself.

It this a valid expectation though? Pointers to ranges are valid InputRanges (try `isInputRange!(typeof(&rng))`.
Ironically where that breaks down is for `auto saved = ptrRng.save; ptrRng = saved;`, so RefRange just seems to fix that bit, thus supporting `isForwardRange!(typeof(refRange(&rng)))`.
If `RefRange` behaved like a pointer, why would we need it in the first place?

Intuitively it seems there is a need for at least 2 use-cases:
- Using refRange to consume (i.e. empty) a range with a normally preserving algorithm.
  Particularly useful for forward ranges that only support `O(N)` `popFrontN`.
  e.g. `refRange(&fwdRng).until!(v => v > 10).sum`
- Using refRange to workaround internal copying of ranges, expecting that it preserves the current range just as passing a normal copyable one.
  e.g. `refRange(&uniqRng).sum; assert(!uniqRng.empty);`

There might be other use-cases I'm not aware of right now!

Please re-read your bug report https://issues.dlang.org/show_bug.cgi?id=18657#c0, it's a bit one-sided and uses complex examples to argue for "doing the obvious thing".

Sorry to play bouncer here, I fully understand the urge to just fix sth., but this problem seems important enough to warrant an open mindset.

thewilsonator:
> A little ping after three weeks of silence isn't being pushy, is it?

Not at all ;) sorry I've only just got around to this. This needs a rebase.

aG0aep6G:
> This needs a rebase.

Done. But this needs a decision more than it needs a rebase.

aG0aep6G:
Happy birthday, PR #6346! :tada:

Still waiting on someone to decide if this is the way to fix [issue 18657](https://issues.dlang.org/show_bug.cgi?id=18657).

aG0aep6G:
Thank you for pulling this, @thewilsonator. I hope you're aware that this was just the first step towards fixing the Bugzilla issue.

Here's the next batch: #6935. More will be needed.

WalterBright:
This sounds like exactly the sort of "Generality Creep" identified by Andrei https://digitalmars.com/d/archives/digitalmars/D/Generality_creep_324867.html

Perhaps RefRange needs to be fundamentally rethought instead of changing every range algorithm.

jmdavis:
> This sounds like exactly the sort of "Generality Creep" identified by Andrei https://digitalmars.com/d/archives/digitalmars/D/Generality_creep_324867.html
> 
> Perhaps RefRange needs to be fundamentally rethought instead of changing every range algorithm.

Honestly, I think that it was a huge mistake to add `RefRange` in the first place (and I'm the one who added it). It's trying to fight against how ranges work to make certain use cases work better, and while it works in some situations, it's ultimately a mess.

PR #6347, opened 2018-03-26
Fix spurious build failures in std.file by using more unique test directories

wilzbach:
Since we added a whole bunch of new test to std.file, auto-tester sometimes fails:

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3092231&isPull=true

So it's probably a good idea to use unique directories like `deleteme` or
in the case of the `tempDir` examples, generating a unique file.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -4398,7 +4398,7 @@ enum SpanMode
     import std.algorithm.iteration : map;
     import std.path : buildPath, relativePath;
 
-    auto root = tempDir.buildPath("root");
+    auto root = deleteme ~ "root";
     scope(exit) root.rmdirRecurse;
     root.mkdir;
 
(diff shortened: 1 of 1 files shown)


std/file.d
```diff
@@ -5105,8 +5105,15 @@ string tempDir() @trusted
 ///
 @safe unittest
 {
+    import std.ascii : letters;
+    import std.conv : to;
     import std.path : buildPath;
-    auto myFile = tempDir.buildPath("my_tmp_file");
```
JackStouffer: Why not `auto myFile = deleteme ~ "my_tmp_file";`?

Seems like this is more complicated than it needs to be.

wilzbach: Because this is the public example for `tempDir` - wouldn't make sense to not use it.

JackStouffer: `auto myFile = tempDir.buildPath(deleteme);`?

wilzbach: `deleteme = "/tmp/<randomid>.<pid>`

PR #6348, opened 2018-03-26
Fix Issue 18663 - std.random.isSeedable has false positives

n8sh:
Example of error:
https://run.dlang.io/is/0nQtvT
```d
void main()
{
    import std.random : isSeedable, Random;
    static assert(isSeedable!(Random, ulong)); // Passes erroneously.
    Random rnd;
    rnd.seed(ulong.max); // Compile-time failure.
}
```

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -178,7 +178,8 @@ template isSeedable(Rng, SeedType)
         is(typeof(
         {
             Rng r = void;              // can define a Rng object
-            r.seed(SeedType.init);     // can seed a Rng
+            SeedType s = void;
+            r.seed(s); // can seed a Rng
         }));
 }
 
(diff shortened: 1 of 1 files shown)


JackStouffer:
What's the cause of the false positive? I don't see any obvious reason why the old code fails and the new code works.

n8sh:
>What's the cause of the false positive? I don't see any obvious reason why the old code fails and the new code works.

It seems as though `ulong.init` gets treated here like a literal `0` and so is implicitly converted to `uint`.

JackStouffer:
That sounds like a language bug, as it should be the literal `0UL`. Can you please submit an issue for this, and then add a comment to the code with the issue number?

JackStouffer:
DMD's constant folding rewrites that code to the earlier code due to constant folding; creating `seed` can be completely elided.

PR #6350, opened 2018-03-26
[trivial] Fixed >120 character line in std.datetime.timezone

JackStouffer:


JackStouffer:
Gah, it needs to be on one line. I don't want to mess with the parsing code. Closing.

aG0aep6G:
> Gah, it needs to be on one line. I don't want to mess with the parsing code. Closing.

You could split it like this:
```
... foo ` ~
    `bar ...
```

JackStouffer:
Moving forward with this to free up the PR queue. Thanks @aG0aep6G for the advice.

PR #6351, opened 2018-03-27
[Trivial]std.functional: -dip1000 compilable by @safe => @system unit…

carblue:
…test

aa[std.functional]=-dip1000
Errors when running: make -f posix.mak std/functional.test
...
```
std/functional.d(1214): Error: @safe function std.functional.__unittest_L1197_C7 cannot call @system function std.functional.__unittest_L1197_C7.memoize!(median).memoize
std/functional.d(1215): Error: @safe function std.functional.__unittest_L1197_C7 cannot call @system function std.functional.__unittest_L1197_C7.memoize!(median).memoize
std/functional.d(1268): Error: @safe function std.functional.__unittest_L1240_C7 cannot call @system function std.functional.__unittest_L1240_C7.memoize!(pickFirst).memoize
std/functional.d(1269): Error: @safe function std.functional.__unittest_L1240_C7 cannot call @system function std.functional.__unittest_L1240_C7.memoize!(pickFirst).memoize
```

Fixing memoize to be @safe requires a lot of @trusted usage, thus I abstain from that

### DIFF
--- a/dip1000.mak
+++ b/dip1000.mak
@@ -26,7 +26,7 @@ aa[std.encoding]=-dip1000
 aa[std.exception]=-dip1000 # merged https://github.com/dlang/phobos/pull/6323; a workaround for https://issues.dlang.org/show_bug.cgi?id=18637
 aa[std.file]=-dip25 # probably already fixed (std.uni); currently: undefined symbol  pure nothrow @nogc return @safe std.uni.SliceOverIndexed!(std.uni.Grapheme).SliceOverIndexed std.uni.SliceOverIndexed!(std.uni.Grapheme).SliceOverIndexed.opSlice()
 aa[std.format]=-dip25 # @system function std.range.primitives.put
-aa[std.functional]=-dip25 # DROP: cannot call @system function std.functional.__unittest_L1216_C7.memoize!(pickFirst).memoize
+aa[std.functional]=-dip1000 # merged https://github.com/dlang/phobos/pull/6351
 aa[std.getopt]=-dip1000
 aa[std.json]=-dip1000
 aa[std.math]=-dip1000
--- a/std/functional.d
+++ b/std/functional.d
@@ -1194,7 +1194,7 @@ template memoize(alias fun, uint maxSize)
 }
 
 // 16079: memoize should work with arrays
-@safe unittest
+@system unittest // not @safe with -dip1000 due to memoize
 {
     int executed = 0;
     T median(T)(const T[] nums) {
@@ -1237,7 +1237,7 @@ template memoize(alias fun, uint maxSize)
 }
 
 // 16079: memoize should work with classes
-@safe unittest
+@system unittest // not @safe with -dip1000 due to memoize
 {
     int executed = 0;
     T pickFirst(T)(T first)


JackStouffer:
The right change here would have been making memoize `@safe`

carblue:
@JackStouffer Why "would have been"? It still can be done and then go back to @safe unittest.
Here are the preliminary changes required (?) that I collected while trying to fix memoize, and still that wasn't the end; some @safe temporary only :
```
diff --git a/std/functional.d b/std/functional.d
index 4a1ff7c..cba77ae 100644
--- a/std/functional.d
+++ b/std/functional.d
@@ -1019,7 +1019,7 @@ template memoize(alias fun)
     import std.traits : ReturnType;
     // alias Args = Parameters!fun; // Bugzilla 13580
 
-    ReturnType!fun memoize(Parameters!fun args)
+    ReturnType!fun memoize(Parameters!fun args) @safe
     {
         alias Args = Parameters!fun;
         import std.typecons : Tuple;
@@ -1037,7 +1037,7 @@ template memoize(alias fun, uint maxSize)
 {
     import std.traits : ReturnType;
     // alias Args = Parameters!fun; // Bugzilla 13580
-    ReturnType!fun memoize(Parameters!fun args)
+    ReturnType!fun memoize(Parameters!fun args) @safe
     {
         import std.traits : hasIndirections;
         import std.typecons : tuple;
@@ -1054,9 +1054,9 @@ template memoize(alias fun, uint maxSize)
             static assert(maxSize < size_t.max - (8 * size_t.sizeof - 1));
 
             enum attr = GC.BlkAttr.NO_INTERIOR | (hasIndirections!Value ? 0 : GC.BlkAttr.NO_SCAN);
-            memo = (cast(Value*) GC.malloc(Value.sizeof * maxSize, attr))[0 .. maxSize];
+            memo = (() @trusted => (cast(Value*) GC.malloc(Value.sizeof * maxSize, attr))[0 .. maxSize])();
             enum nwords = (maxSize + 8 * size_t.sizeof - 1) / (8 * size_t.sizeof);
-            initialized = (cast(size_t*) GC.calloc(nwords * size_t.sizeof, attr | GC.BlkAttr.NO_SCAN))[0 .. nwords];
+            initialized = (() @trusted => (cast(size_t*) GC.calloc(nwords * size_t.sizeof, attr | GC.BlkAttr.NO_SCAN))[0 .. nwords])();
         }
 
         import core.bitop : bt, bts;
@@ -1064,23 +1064,23 @@ template memoize(alias fun, uint maxSize)
 
         size_t hash;
         foreach (ref arg; args)
-            hash = hashOf(arg, hash);
+            hash = 99;// TODO 99 is just an arbitrary setting to get around @system errors from: hashOf(arg, hash);
         // cuckoo hashing
         immutable idx1 = hash % maxSize;
-        if (!bt(initialized.ptr, idx1))
+        if (!(() @trusted => bt(&initialized[0], idx1))())
         {
             emplace(&memo[idx1], args, fun(args));
-            bts(initialized.ptr, idx1); // only set to initialized after setting args and value (bugzilla 14025)
+            (() @trusted => bts(&initialized[0], idx1))(); // only set to initialized after setting args and value (bugzilla 14025)
             return memo[idx1].res;
         }
         else if (memo[idx1].args == args)
             return memo[idx1].res;
         // FNV prime
         immutable idx2 = (hash * 16_777_619) % maxSize;
-        if (!bt(initialized.ptr, idx2))
+        if (!(() @trusted => bt(&initialized[0], idx2))())
         {
             emplace(&memo[idx2], memo[idx1]);
-            bts(initialized.ptr, idx2); // only set to initialized after setting args and value (bugzilla 14025)
+            (() @trusted => bts(&initialized[0], idx2))(); // only set to initialized after setting args and value (bugzilla 14025)
         }
         else if (memo[idx2].args == args)
             return memo[idx2].res;
```

JackStouffer:
Because allowing a regression in safety to pass the test suite is not desirable. Better than DIP1000 error until memoize compiles in `@safe` code.

The goal here is not just to get DIP1000 to compile.

carblue:
I believe, that was pretended safety only in these 2 unittest. The changes above show, where -dip1000 stumbles upon calls which are @system, and I wonder why it compiled at all with -dip25 as @safe unittest.

WalterBright:
> The goal here is not just to get DIP1000 to compile.

To some extent, it is. DIP1000 cannot be realistically used at all until Phobos compiles with it. It's a bit of all or nothing. If some compromises are made in the process, those can be addressed afterwards. Phobos still is not fully @safe anyway.

PR #6352, opened 2018-03-27
Fix Issue 8048 - Missing head function in std.net.curl

vlasebian:
I have implemented head functionality for HTTP. A sample output for running head("www.example.com") is the following:
```
HTTP/1.1 200 OK
accept-ranges: bytes
cache-control: max-age=604800
content-type: text/html
date: Tue, 27 Mar 2018 10:59:56 GMT
etag: "1541025663"
expires: Tue, 03 Apr 2018 10:59:56 GMT
last-modified: Fri, 09 Aug 2013 23:54:35 GMT
server: ECS (dca/2469)
x-cache: HIT
content-length: 1270

```

### DIFF
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -1,5 +1,4 @@
 // Written in the D programming language.
-
 /**
 Networking client functionality as provided by $(HTTP _curl.haxx.se/libcurl,
 libcurl). The libcurl library must be installed on the system in order to use
@@ -59,6 +58,10 @@ uploads file from file system to URL.)
 )
 $(TR $(TDNW $(LREF get)) $(TD $(D
 get("dlang.org")) returns a char[] containing the dlang.org web page.)
+)
+$(TR $(TDNW $(LREF head)) $(TD $(D
+head("dlang.org")) returns a char[] containing the header of the dlang.org
+web page.)
 )
 $(TR $(TDNW $(LREF put)) $(TD $(D
 put("dlang.org", "Hi")) returns a char[] containing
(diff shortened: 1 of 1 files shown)


std/net/curl.d
```diff
@@ -525,6 +528,54 @@ if (isCurlConn!Conn)
...
+ * A T[] range containing the head of the resource pointed to by the URL.
+ *
+ * Throws:
+ *
+ * $(D CurlException) on error.
+ *
+ * See_Also: $(LREF HTTP.Method)
+*/
+
+T[] head(T = char)(const (char)[] url, HTTP conn = HTTP())
```
JackStouffer: remove space after const

std/net/curl.d
```diff
@@ -525,6 +528,54 @@ if (isCurlConn!Conn)
...
+if ( (is(T == char) || is(T == ubyte)) )
```
JackStouffer: remove spaces around braces

std/net/curl.d
```diff
@@ -525,6 +528,54 @@ if (isCurlConn!Conn)
     }
 }
 
+/** HTTP head
```
wilzbach: Maybe the other symbols aren't ideal either, but I think a more descriptive summery should be used here.

std/net/curl.d
```diff
@@ -1069,12 +1121,23 @@ private auto _basicHTTP(T)(const(char)[] url, const(void)[] sendData, HTTP clien
             import std.conv : to;
             content.reserve(value.to!size_t);
         }
+        header ~= cast(ubyte[])(key ~ ": " ~ value ~ "\n");
```
wilzbach: Doesn't this add a lot of overhead if you aren't interested in head? Allocations aren't for free

CyberShadow:
I'm not sure about this.

The canonical way to read headers with curl is using the callbacks (`onReceiveStatusLine` and `onReceiveHeader`). Getting the raw HTTP headers in raw string form is a pretty specific use case for non-interactive uses.

std/net/curl.d
```diff
@@ -1069,12 +1121,23 @@ private auto _basicHTTP(T)(const(char)[] url, const(void)[] sendData, HTTP clien
...
     enforce(statusLine.code / 100 == 2, new HTTPStatusException(statusLine.code,
             format("HTTP request returned status code %d (%s)", statusLine.code, statusLine.reason)));
 
+    if (client.method == HTTP.Method.head)
+    {
+        import std.conv : to;
+
+        auto tmpHeader = _decodeContent!T(header.data, client.p.charset);
+        auto statusFormat = to!(typeof(tmpHeader))(format("HTTP/%s.%s %s %s\n", statusLine.majorVersion,
+                statusLine.minorVersion, statusLine.code, statusLine.reason));
```
CyberShadow: Oh no, so this doesn't actually even return the headers / status line as it was sent by the HTTP server - it just builds a string that looks like one!

Sorry, I can't imagine there being a justifiable use case for this.

CyberShadow:
@vlasebian Could you please go into more detail what problem this change would help you solve? I think there is probably a better way to achieve that.

jercaianu:
ping @vlasebian

vlasebian:
Sorry for the late response. I don't need the `head` method for a specific purpose, but someone posted an issue about it and I tried to solve it. For uniformity, I think it would be a great idea to have a `head` functionality.

Do you think it would be fine to make this a wrapper around `responseHeaders()' from `struct HTTP`?
If not, how would you recommend solving this?

CyberShadow:
> Do you think it would be fine to make this a wrapper around responseHeaders() from struct HTTP?

I'm really not sure about that. Doing this properly would mean introducing a new type which stores the results. It also introduces a "cognitive dissonance": both GET and HEAD HTTP methods return headers; `head` would return just the headers (because there is no body), but `get` returns only the body, without the headers (even though they are sent). The solution is to use the `HTTP` type, which allows doing both types of requests and receiving both the headers and body symmetrically.

> If not, how would you recommend solving this?

I suggest closing the issue as WONTFIX, at least until we have a better picture of what problem the user is trying to solve.

PR #6355, opened 2018-03-27
Fix Issue 18669 - isNestedFunction does not check if parameter is a f…

JackStouffer:
…unction

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -7097,19 +7097,49 @@ if (T.length == 1)
 }
 
 /**
-Determines whether function $(D f) requires a context pointer.
+Determines if `f` is a function that requires a context pointer.
+
+Params:
+    f = The type to check
+Returns
+    A `bool`
 */
 template isNestedFunction(alias f)
 {
-    enum isNestedFunction = __traits(isNested, f);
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -7097,19 +7097,49 @@ if (T.length == 1)
...
 {
-    enum isNestedFunction = __traits(isNested, f);
+    enum isNestedFunction = __traits(isNested, f) && isSomeFunction!(f);
 }
 
+///
 @safe unittest
 {
-    static void f() { }
-    void g() { }
```
JackStouffer: changed up this example because it's not clear to the documentation user that the example happens in a function context

std/traits.d
```diff
@@ -7097,19 +7097,49 @@ if (T.length == 1)
 }
 
 /**
-Determines whether function $(D f) requires a context pointer.
+Determines whether function `f` requires a context pointer.
+
+Params:
+    f = a function
```
n8sh: If evaluating `isNestedFunction` for a non-function parameter is okay then the parameter docs should say so. If it's not okay then it should be a template constraint rather than part of the body.

JackStouffer:
@n8sh Fixed

PR #6356, opened 2018-03-27
Simplified std.datetime.package overview

JackStouffer:
Many people have remarked that `std.datetime` is confusing: https://github.com/wilzbach/state-of-d-2018/blob/master/12h:%20Which%20parts%20of%20Phobos%20(D%27s%20Standard%20Library)%2C%20if%20any%2C%20do%20you%20find%20difficult%20to%20work%20with%3F

This is the first step to improving that situation.

This PR:

1. Gives links to symbols based on usage in a clear table
2. Adds the most common usage of datetime in examples
3. Removes a large amount of text that is repeated elsewhere. I believe this text was either self-explanitory or gone over on each of the respective function docs

### DIFF
--- a/std/datetime/package.d
+++ b/std/datetime/package.d
@@ -1,102 +1,80 @@
 // Written in the D programming language
 
 /++
-    Module containing Date/Time functionality.
+    $(SCRIPT inhibitQuickIndex = 1;)
 
-    This module provides:
+    Phobos provides the following functionality for time:
+
+    $(DIVC quickindex,
+    $(BOOKTABLE ,
+    $(TR $(TH Functionality) $(TH Symbols)
+    )
+    $(TR
+        $(TD Points in Time)
+        $(TD
+            $(REF_ALTTEXT Date, Date, std, datetime, date)$(NBSP)
+            $(REF_ALTTEXT TimeOfDay, TimeOfDay, std, datetime, date)$(NBSP)
+            $(REF_ALTTEXT DateTime, DateTime, std, datetime, date)$(NBSP)
+            $(REF_ALTTEXT SysTime, SysTime, std, datetime, systime)$(NBSP)
+        )
+    )
+    $(TR
+        $(TD Timezones)
+        $(TD
+            $(REF_ALTTEXT TimeZone, TimeZone, std, datetime, timezone)$(NBSP)
+            $(REF_ALTTEXT UTC, UTC, std, datetime, timezone)$(NBSP)
+            $(REF_ALTTEXT LocalTime, LocalTime, std, datetime, timezone)$(NBSP)
+            $(REF_ALTTEXT PosixTimeZone, PosixTimeZone, std, datetime, timezone)$(NBSP)
+            $(REF_ALTTEXT WindowsTimeZone, WindowsTimeZone, std, datetime, timezone)$(NBSP)
+            $(REF_ALTTEXT SimpleTimeZone, SimpleTimeZone, std, datetime, timezone)$(NBSP)
+        )
+    )
+    $(TR
+        $(TD Intervals and Ranges of Time)
+        $(TD
+            $(REF_ALTTEXT Interval, Interval, std, datetime, interval)$(NBSP)
+            $(REF_ALTTEXT PosInfInterval, PosInfInterval, std, datetime, interval)$(NBSP)
+            $(REF_ALTTEXT NegInfInterval, NegInfInterval, std, datetime, interval)$(NBSP)
+        )
+    )
+    $(TR
+        $(TD Durations of Time)
+        $(TD
+            $(REF_ALTTEXT Duration, Duration, core, time)$(NBSP)
+            $(REF_ALTTEXT weeks, weeks, core, time)$(NBSP)
+            $(REF_ALTTEXT days, days, core, time)$(NBSP)
+            $(REF_ALTTEXT hours, hours, core, time)$(NBSP)
+            $(REF_ALTTEXT minutes, minutes, core, time)$(NBSP)
+            $(REF_ALTTEXT seconds, seconds, core, time)$(NBSP)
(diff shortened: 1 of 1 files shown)


JackStouffer:
This is just another step along the way to making std.datetime more accessible. I also want to add top level examples to all datetime modules to quickly introduce the concepts of the module.

std/datetime/package.d
```diff
@@ -109,6 +82,43 @@ auto restoredTime = SysTime.fromISOExtString(timeString);
...
+Construct a specific point in time in the UTC timezone and
+add two days.
+ */
+@safe unittest
+{
+    import std.datetime.systime : SysTime;
+    import std.datetime.timezone : UTC;
+    import core.time : days;
+
+    auto st = SysTime(DateTime(2018, 1, 1, 12, 30, 10), UTC());
```
wilzbach: Maybe use an actual TimeZone here, s.t. it's non-trivial?

wilzbach:
Haha ironically auto-tester is running in a spurious failure in std.datetime in this PR on FreeBSD:

```
timelimit: sending warning signal 15
gmake[1]: *** [posix.mak:380: unittest/std/datetime/systime.run] Error 143
```

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3094297&isPull=true

JackStouffer:
What's up with MREF throwing out malformed HTML?

http://dtest.dlang.io/artifact/website-774b335033cedcb7eec8178408bfa3ad5ba2c011-56a6b22e12c72884486cc8bd4920b952/web/phobos-prerelease/std_datetime.html

std/datetime/package.d
```diff
@@ -1,98 +1,71 @@
...
-        $(REF DateTimeException,std,_datetime,date) is an alias for
-        $(REF TimeException, core,time), so you don't need to worry about
-        core.time functions and std.datetime functions throwing different
-        exception types (except in the rare case that they throw something other
-        than $(REF TimeException, core,time) or
-        $(REF DateTimeException,std,_datetime,date)).
+    This functionality is separated into the following modules
+
+    $(UL
+        $(LI $(MREF std, datetime, date))
```
wilzbach: You need to use `_datetime` here. While opt-out of automatic keyword highlighting is available since 2.079, it hasn't been applied to dlang.org yet:

https://dlang.org/changelog/2.079.0.html#fix18361

wilzbach: Let's turn the switch -> https://github.com/dlang/dlang.org/pull/2307

std/datetime/package.d
```diff
@@ -1,98 +1,73 @@
...
-             $(REF DateTime,std,_datetime,date).)
-        $(LI Types to represent intervals of time.)
-        $(LI Types to represent ranges over intervals of time.)
-        $(LI Types to represent time zones (used by
-             $(REF SysTime,std,_datetime,systime)).)
-        $(LI A platform-independent, high precision stopwatch type:
-             $(LREF StopWatch))
-        $(LI Benchmarking functions.)
-        $(LI Various helper functions.)
+    Phobos provides the following functionality for time:
```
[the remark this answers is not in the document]
JackStouffer: This is referring to time the concept and not time on the clock.

JackStouffer:
I don't think much information is lost. A lot of what's written here is covered in the documentation for the modules and for the types.

n8sh:
>I don't think much information is lost. A lot of what's written here is covered in the documentation for the modules and for the types.

The only thing that leaps out to me as needing module-level explanation is a table of accepted time unit strings. (For ease of learning units probably should have been an enumerated type rather than raw strings, but documentation is for what exists.)

JackStouffer:
>The only thing that leaps out to me as needing module-level explanation is a table of accepted time unit strings.

I think that would go on the individual module docs.

@jmdavis Ping.

std/datetime/package.d
```diff
@@ -1,98 +1,76 @@
...
+
+    $(DIVC quickindex,
+    $(BOOKTABLE ,
+    $(TR $(TH Functionality) $(TH Symbols)
+    )
+    $(TR
+        $(TD Points in Time)
+        $(TD
+            $(REF_ALTTEXT Date, Date, std, _datetime, date)$(NBSP)
```
wilzbach: Preventing auto-highlighting is no longer needed :)
see e.g. https://github.com/dlang/dlang.org/pull/2307

std/datetime/package.d
```diff
@@ -1,98 +1,76 @@
...
-        core.time functions and std.datetime functions throwing different
-        exception types (except in the rare case that they throw something other
-        than $(REF TimeException, core,time) or
-        $(REF DateTimeException,std,_datetime,date)).
+    $(UL
+        $(LI $(MREF std, _datetime, date))
+        $(LI $(MREF std, _datetime, timezone))
+        $(LI $(MREF std, _datetime, systime))
+        $(LI $(MREF std, _datetime, interval))
+        $(LI $(MREF std, _datetime, stopwatch))
```
wilzbach: Maybe add a short, one-line description of each of these packages, s.t. the user knows what the difference betwen date and systime is?

jmdavis:
> I don't think much information is lost. A lot of what's written here is covered in the documentation for the modules and for the types.

Some of the information that was here should be in the package documentation as part of an overview, and some if really just belongs in the modules. The result of these changes is that we pretty much just end up with a table listing the types and a distinct lack of any kind of explanation.

I guess that I'll let Seb merge this when his requested changes are done, but pretty the only way that I'm going to be happy with this is if I take the time to go over the module documentation for each module in std.datetime (including package.d) and rework it. In the interim, at least these changes get rid of the reference to std.datetime being a module, and the table cleans up what information is kept even if just about everything else is lost in the process. Unfortunately, I simply don't have the time to spend on reworking the docs right now. I guess that I'll try to tackle that during dconf.

PR #6357, opened 2018-03-28
make buildNormalizedPath() @safe

WalterBright:
Shrink the @trusted to the smallest bit of the function that needs it. The rest is to follow the convention that `result` is for the return value.

### DIFF
--- a/std/path.d
+++ b/std/path.d
@@ -1725,21 +1725,23 @@ if (Ranges.length >= 2 &&
     Returns: The assembled path.
 */
 immutable(C)[] buildNormalizedPath(C)(const(C[])[] paths...)
-    @trusted pure nothrow
+    @safe pure nothrow
 if (isSomeChar!C)
 {
     import std.array : array;
+    import std.exception : assumeUnique;
 
-    const(C)[] result;
+    const(C)[] chained;
     foreach (path; paths)
     {
-        if (result)
-            result = chainPath(result, path).array;
+        if (chained)
+            chained = chainPath(chained, path).array;
         else
-            result = path;
+            chained = path;
     }
-    result = asNormalizedPath(result).array;
-    return cast(typeof(return)) result;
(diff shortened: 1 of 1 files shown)


std/path.d
```diff
@@ -1725,21 +1725,22 @@ if (Ranges.length >= 2 &&
...
+        if (chained)
+            chained = chainPath(chained, path).array;
         else
-            result = path;
+            chained = path;
     }
-    result = asNormalizedPath(result).array;
-    return cast(typeof(return)) result;
+    auto result = asNormalizedPath(chained).array;  // .array returns a copy, so we know it is unique
+    return () @trusted { return assumeUnique(result); } ();
```
aG0aep6G: I realize that this has become an accepted way to do `@trusted`. But I think it actually goes against the idea behind it.

The `@trusted` snippet doesn't ensure safety by itself. It relies on the rest of the code not messing with `result`. That means, you can trigger undefined behavior by editing `@safe` code. Exactly what should not be possible.

WalterBright: Good point.

std/path.d
```diff
@@ -1725,21 +1725,23 @@ if (Ranges.length >= 2 &&
...
+            chained = chainPath(chained, path).array;
         else
-            result = path;
+            chained = path;
     }
-    result = asNormalizedPath(result).array;
-    return cast(typeof(return)) result;
+    auto result = asNormalizedPath(chained);
+    // .array returns a copy, so we know it is unique
+    return () @trusted { return assumeUnique(result.array); } ();
```
quickfur: Hmm. If `.array` returns a unique array, how come the compiler can't infer it?  I would have thought that it should be inferred as `pure`, and the compiler would have been able to implicitly convert it to immutable without a `@trusted` escape.

WalterBright: Pure isn't enough to infer the result is immutable.

PR #6360, opened 2018-03-28
[WIP] std.stdio: make std.stdio -dip1000 compatible

wilzbach:
This is an experiment on what's needed to make `std.stdio` -dip1000.

It currently just slaps `@trusted` over the `formattedWrite` calls, which is absolutely wrong.
However, when std.format gets upgraded to dip1000, only those five places need to be looked at.
Apart from that with the `scope` annotations from this PR, std.stdio would be ready.

CC @carblue


wilzbach:
Superseded by https://github.com/dlang/phobos/pull/6931

PR #6361, opened 2018-03-28
Add ASCII fast path to sicmp & icmp

n8sh:
Followup to #6225. Add fast path for ASCII string comparison with fallback to proper decoding in `sicmp` and `icmp`.

Below benchmarks were calculated as in https://github.com/dlang/phobos/pull/6225#issue-171419109. "sicmp_6225" is PR 6225 and "sicmp_new" is PR 6225 plus this PR.

Benchmark ``using ldc2 -O2 -release`` (LDC v1.7.0):
```
strcasecmp 29 ms, 857 μs, and 7 hnsecs
sicmp_6225 ASCII 57 ms, 17 μs, and 8 hnsecs
sicmp_new ASCII 30 ms, 866 μs, and 4 hnsecs
sicmp_6225 Unicode 116 ms, 711 μs, and 3 hnsecs
sicmp_new Unicode 128 ms, 62 μs, and 6 hnsecs
```

Benchmark using ``dmd -O -inline -release`` (DMD v2.078.3):
```
strcasecmp 32 ms, 919 μs, and 5 hnsecs
sicmp_6225 ASCII 84 ms, 394 μs, and 3 hnsecs
sicmp_new ASCII 38 ms, 703 μs, and 9 hnsecs
sicmp_6225 Unicode 282 ms, 52 μs, and 2 hnsecs
sicmp_new Unicode 286 ms, 652 μs, and 4 hnsecs
```

### DIFF
--- a/std/uni.d
+++ b/std/uni.d
@@ -7787,6 +7787,46 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
     import std.typecons : Yes;
     static import std.ascii;
 
+    static if ((isDynamicArray!S1 || isRandomAccessRange!S1)
+        && (isDynamicArray!S2 || isRandomAccessRange!S2)
+        && !(isInfinite!S1 && isInfinite!S2)
+        && __traits(compiles,
+            {
+                size_t s = size_t.sizeof / 2;
+                r1 = r1[s .. $];
+                r2 = r2[s .. $];
+            }))
+    {{
+        // ASCII optimization for dynamic arrays & similar.
+        size_t i = 0;
+        static if (isInfinite!S1)
+            immutable end = r2.length;
+        else static if (isInfinite!S2)
+            immutable end = r1.length;
+        else
+            immutable end = r1.length > r2.length ? r2.length : r1.length;
+        for (; i < end; ++i)
+        {
+            auto lhs = r1[i];
+            auto rhs = r2[i];
+            if ((lhs | rhs) >= 0x80) goto NonAsciiPath;
+            if (lhs == rhs) continue;
+            auto lowDiff = std.ascii.toLower(lhs) - std.ascii.toLower(rhs);
+            if (lowDiff) return lowDiff;
+        }
+        static if (isInfinite!S1)
+            return 1;
+        else static if (isInfinite!S2)
+            return -1;
+        else
+            return (r1.length > r2.length) - (r2.length > r1.length);
+
+    NonAsciiPath:
+        r1 = r1[i .. $];
+        r2 = r2[i .. $];
+        // Fall through to standard case.
+    }}
+
     while (!r1.empty)
     {
         immutable lhs = decodeFront!(Yes.useReplacementDchar)(r1);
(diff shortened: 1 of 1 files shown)


std/uni.d
```diff
@@ -7787,6 +7787,41 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
...
+            immutable end = r2.length;
+        else static if (isInfinite!S2)
+            immutable end = r1.length;
+        else
+            immutable end = r1.length > r2.length ? r2.length : r1.length;
+        for (; i < end; ++i)
+        {
+            auto lhs = r1[i];
+            auto rhs = r2[i];
+            if ((lhs | rhs) >= 0x80) goto non_ascii_path;
```
wilzbach: Nit: I couldn't many precedents for using `_` in labels (there are a few `L_backtrack`).
IIRC Andrei requested once that labels should start with UpperCase too, so it would be `NonAsciiPath`, but I'm not entirely sure on this either. We should probably put it to the DStyle once clarified.

std/uni.d
```diff
@@ -7787,6 +7787,41 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
...
+            if (lhs == rhs) continue;
+            auto lowDiff = std.ascii.toLower(lhs) - std.ascii.toLower(rhs);
+            if (lowDiff) return lowDiff;
+        }
+        return (r1.length > r2.length) - (r2.length > r1.length);
```
wilzbach: :+1:

std/uni.d
```diff
@@ -7787,6 +7787,41 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
...
     static import std.ascii;
 
+    static if ((isDynamicArray!S1 || isRandomAccessRange!S1)
+        && (isDynamicArray!S2 || isRandomAccessRange!S2)
+        && !(isInfinite!S1 && isInfinite!S2)
+        && __traits(compiles,
+            {
+                size_t s = size_t.sizeof / 2;
+                r1 = r1[s .. $];
+                r2 = r2[s .. $];
```
wilzbach: How about `allSatisfy!(hasSlicing, S1, S2)` ?

DmitryOlshansky:
Looks good overall, maybe we need to devise a non-decoding case-comparison tables for non-ASCII.

A bit uneasy about the ASCII case going off rails on the first non-ASCII codepoint, maybe attempt to improve on it by doing single decodeFront + part of the old logic and continue the fast loop?

std/uni.d
n8sh: I think for infinite ranges slicing doesn't necessarily produce the same type so I need to check assignment works.

std/uni.d
```diff
@@ -7787,6 +7787,41 @@ if (isInputRange!S1 && isSomeChar!(ElementEncodingType!S1)
...
+            auto lhs = r1[i];
+            auto rhs = r2[i];
+            if ((lhs | rhs) >= 0x80) goto non_ascii_path;
+            if (lhs == rhs) continue;
+            auto lowDiff = std.ascii.toLower(lhs) - std.ascii.toLower(rhs);
+            if (lowDiff) return lowDiff;
+        }
+        return (r1.length > r2.length) - (r2.length > r1.length);
+
+        non_ascii_path:
```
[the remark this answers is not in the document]
n8sh: Fixed.

std/uni.d
n8sh: Fixed.

n8sh:
>maybe attempt to improve on it by doing single decodeFront + part of the old logic and continue the fast loop?

That would make it not much different from the primary loop (which as of #6225 includes an ASCII optimization which we would still benefit from after breaking out of the ASCII-only case). My intuition is that not doing this is part of what makes the fast path faster.

Obviously on any data consisting solely of ASCII, de-simplifying the fast loop could only decrease performance. On data consisting of all non-ASCII characters trying the ASCII loop more times will also be a waste. Making a benchmark-driven improvement would require a representative corpus of strings containing a mixture of ASCII & non-ASCII characters, since the fastest way would depend on the frequency of non-ASCII and the characteristics of the distribution. (For instance, the fastest way might be to optimize for alternation between long runs of ASCII characters and long runs of non-ASCII character. Or it might not. We need to measure using real input data to find out.)

DmitryOlshansky:
> That would make it not much different from the primary loop (which as of #6225 includes an ASCII optimization which we would still benefit from after breaking out of the ASCII-only case). My intuition is that not doing this is part of what makes the fast path faster.

I think it might be due to decodeFront being slower then just peeking at 1 byte of each string with `[i]`. Would be interesting to check.

DmitryOlshansky:
> since the fastest way would depend on the frequency of non-ASCII and the characteristics of the distribution. (For instance, the fastest way might be to optimize for alternation between long runs of ASCII characters and long runs of non-ASCII character. Or it might not. We need to measure using real input data to find out.)

I used to run std.uni.isXYZ on a dumps of wikipedia title for 4 different languages (of different groups).
Might still be worth it: https://github.com/DmitryOlshansky/gsoc-bench-2012
look at -wiki files.

wilzbach:
Oh wait I just thought more about it and there could be a point made that this slight decrease for the normal path is not nice:

```
sicmp_6225 Unicode 116 ms, 711 μs, and 3 hnsecs
sicmp_new Unicode 128 ms, 62 μs, and 6 hnsecs
...
sicmp_6225 Unicode 282 ms, 52 μs, and 2 hnsecs
sicmp_new Unicode 286 ms, 652 μs, and 4 hnsecs
```

So are we okay with this slight decrease for the unicode case?

DmitryOlshansky:
> So are we okay with this slight decrease for the unicode case?

~ 2ms is quite close to statistic error (at least as I run tests on my machine) so not really.

DmitryOlshansky:
I meant “not really a problem”.

wilzbach:
The auto-tester failure on Windows 32_64 unrelated:

```
The system cannot find the path specified.
failed to execute '\"Program Files (x86)"\"Microsoft Visual Studio 10.0"\VC\bin\amd64\cl.exe /c /nologo runnable\extra-files\cpp_abi_tests.cpp /Fotest_results\runnable\cpp_abi_tests.cpp.obj'
Makefile:133: recipe for target 'test_results/runnable/cpp_abi_tests.d.out' failed
gmake[1]: *** [test_results/runnable/cpp_abi_tests.d.out] Error 1
gmake[1]: *** Waiting for unfinished jobs....
```

See also: https://github.com/braddr/d-tester/issues/71

PR #6332, opened 2018-03-23
std.container.rbtree: Fix -dip1000 compilable issues

carblue:
https://github.com/dlang/phobos/blob/master/dip1000.mak with
aa[std.container.rbtree]=-dip1000
Errors when running: make -f posix.mak std/container/rbtree.test
```
std/container/rbtree.d(817): Error: @safe function std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.__unittest_L813_C39 cannot call @system constructor std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.this
std/container/rbtree.d(991): Error: @safe function std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.__unittest_L988_C39 cannot call @system constructor std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.this
std/container/rbtree.d(1056): Error: @safe function std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.__unittest_L1054_C39 cannot call @system constructor std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.this
std/container/rbtree.d(1111): Error: @safe function std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.__unittest_L1109_C39 cannot call @system constructor std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.this
std/container/rbtree.d(1173): Error: @safe function std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.__unittest_L1171_C39 cannot call @system constructor std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.this
std/container/rbtree.d(1222): Error: @safe function std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.__unittest_L1220_C39 cannot call @system constructor std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.this
std/container/rbtree.d(1266): Error: @safe function std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.__unittest_L1264_C39 cannot call @system constructor std.container.rbtree.RedBlackTree!(int, "a < b", false).RedBlackTree.this
[...]

PR #6362, opened 2018-03-28
std.container.rbtree: Fix -dip1000 compilable issues (alternative)

wilzbach:
Alternative version to: https://github.com/dlang/phobos/pull/6332

Still not perfect and I don't know why this is disallowed:

```
auto rbt4 = redBlackTree(chain(["hello"], ["world"]));
assert(equal(rbt4[], ["hello", "world"]));
```

but this is allowed:

```
auto rbt5 = redBlackTree!true(chain([0, 1], [5, 7, 5]));
assert(equal(rbt5[], [0, 1, 5, 5, 7]));
```

### DIFF
--- a/dip1000.mak
+++ b/dip1000.mak
@@ -94,7 +94,7 @@ aa[std.container.array]=-dip1000
 aa[std.container.binaryheap]=-dip1000
 aa[std.container.dlist]=-dip1000
 aa[std.container.package]=-dip1000
-aa[std.container.rbtree]=-dip25 # DROP
+aa[std.container.rbtree]=-dip1000 -version=DIP1000
 aa[std.container.slist]=-dip25 # -dip1000 -version=DIP1000   depends on an update (no insertFront's code duplication in constructor) and merge of https://github.com/dlang/phobos/pull/6295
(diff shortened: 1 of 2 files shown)


schveiguy:
is scope transitive? If so, then it's going to be impossible to work with this. 

For example, the ctor that accepts a stack-allocated list of elems must be declared scope (the array is on the stack after all), but the elements themselves may not be if they point at non-scope data (such as strings).

wilzbach:
@WalterBright you need to approve PRs at druntime & Phobos before/when adding auto-merge.

schveiguy:
> scope is not transitive

It sure seems like it is. If I mark a parameter `scope`, then all the references appear to be assumed `scope`:

```D
@safe:
void noscope(string s) {}

void test(scope string[] s) {
   noscope(s[0]); // Error: scope variable arr assigned to non-scope parameter s calling testscope.noscope
}
```

How do we identify that the strings inside `s` are not `scope`?

wilzbach:
Superseded by https://github.com/dlang/phobos/pull/6931

PR #6363, opened 2018-03-28
dip1000.mak: Lift std.conv

carblue:
"Magically" :-) std.conv got -dip1000 in the meantime

I'll go over to std.typecons and std.socket, then continue with std.algorithm.iteration:
This was a complicated beast when I made a break there, touching so many other files.

I didn't include in dip1000.mak, that std.container.util depends on both rbtree and slist to be dip1000,
because it may interfere with dip1000.mak s in the PR pipeline


wilzbach:
FYI: We also have this dashboard (https://github.com/dlang/phobos/projects/6), but I'm not so sure whether you can modify items there.
Maybe @WalterBright or @andralex can invite you to team though ;-)

carblue:
I'm not very comfortable with just commenting out lines in std.typecons to get it to compile with -dip1000 (especially as I don't yet fully understand Proxy), but I think, as a compromise, it may be done.
I'll start to explicitly state compromises or "merged PR URLs" in dip1000.mak's comments: I look at those as TODOs, once the dip1000 transition is done, to have a look into those again, thus those compromises don't get lost.

No, I don't think to be able to modify the dashboard

PR #6366, opened 2018-03-28
Fix localtime

FeepingCreature:
`localtime` returns a pointer to a global shared data structure. This is terrible. `localtime` is terrible. `gmtime` is also terrible. Do not call `localtime`. Do not call `gmtime`. Do not interact with `localtime` or `gmtime` in any way, except if the way is "I decided to remove `localtime` and replace it with `localtime_r`".

This PR removes `localtime` and replaces it with `localtime_r`.

### DIFF
--- a/std/datetime/timezone.d
+++ b/std/datetime/timezone.d
@@ -281,9 +281,12 @@ public:
 
                 static void testTM(in SysTime st)
                 {
-                    import core.stdc.time : localtime, tm;
+                    import core.stdc.time : tm;
+                    import core.sys.posix.time : localtime_r;
+
                     time_t unixTime = st.toUnixTime();
-                    tm* osTimeInfo = localtime(&unixTime);
+                    tm osTimeInfo = void;
+                    localtime_r(&unixTime, &osTimeInfo);
                     tm ourTimeInfo = st.toTM();
 
                     assert(ourTimeInfo.tm_sec == osTimeInfo.tm_sec);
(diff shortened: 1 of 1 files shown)


FeepingCreature:
Ah, fixed windows not building. Windows localtime does it right: it returns a pointer to a threadlocal variable.

... I just wrote "Windows does it right." In my heart, a penguin is crying.

FeepingCreature:
ping.

std/datetime/timezone.d
```diff
@@ -874,17 +877,23 @@ public:
...
+        import core.stdc.time : tm;
+
         time_t unixTime = stdTimeToUnixTime(stdTime);
 
         version(Posix)
         {
-            tm* timeInfo = localtime(&unixTime);
+            import core.sys.posix.time : localtime_r;
+
+            tm timeInfo;
```
n8sh: `= void`?

FeepingCreature: Is void supposed to be used whenever initialization is not required, or is it just a performance hack meant for inner loops?

n8sh: In general writing or omitting void doesn't matter, since the optimizer can determine if writing the variable is unnecessary and omit it, but here it cannot. In any case it serves to document intention.

jmdavis: Yeah, it has nothing to do with inner loops specifically. It's merely a way to avoid having the variable default-initialized when you know that it's unnecessary. And in this case, there's no way that the optimizer can optimize out the default initialization, because the variable is set in a C function. Ideally, each time that this code declares a variable just to pass it to a C function to set it without looking at its value, `= void` would be used to avoid the default initialization.

wilzbach:
Should we maybe connect this issue with a Bugzilla issue, s.t. it appears on the changelog and people who potentially experience problems would know where to start looking?

FeepingCreature:
I'll go look for one or file one.

std/datetime/timezone.d
FeepingCreature: Sure... I guess.

jmdavis:
I'd say that this looks good to go except that the commits should be squashed.

FeepingCreature:
Squashed.

wilzbach:
(the Jenkins/Project tester failure is unfortunately a permanent bug in dub -> https://github.com/dlang/ci/issues/190)

PR #6367, opened 2018-03-28
Fix 18689 - std.format should throw FormatException on bad format spec

JackStouffer:
This is not a breaking change, as any use of try/catch or collectException will continue to work, even if they manually specified the type.

This change just makes sense logically to me.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -936,7 +936,7 @@ uint formattedRead(R, Char, S...)(auto ref R r, const(Char)[] fmt, auto ref S ar
 
     int[4] sa2;
     input = `[1,2,3]`;
-    assertThrown(formattedRead(input, "[%(%s,%)]", &sa2));
+    assertThrown!FormatException(formattedRead(input, "[%(%s,%)]", &sa2));
 }
 
 @system pure unittest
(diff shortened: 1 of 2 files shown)


std/format.d
```diff
@@ -64,24 +64,31 @@ import std.range.primitives;
...
 {
-    @safe pure nothrow
+    @safe @nogc pure nothrow
     this()
     {
         super("format error");
     }
 
-    @safe pure nothrow
+    @safe @nogc pure nothrow
```
wilzbach: Interesting that the compiler accepts this (though it doesn't make so much sense, does it?)

std/format.d
```diff
@@ -5440,7 +5447,9 @@ if (isInputRange!Range && isIntegral!T && !is(T == enum) && isSomeChar!(ElementT
                 || is(Unqual!(ElementEncodingType!Range) == ubyte))
             return rawRead!T(input);
         else
-            throw new Exception("The raw read specifier %r may only be used with narrow strings and ranges of bytes.");
+            throw new FormatException(
+                "The raw read specifier %r may only be used with narrow strings and ranges of bytes."
+            );
     }
 
     enforce(find(acceptedSpecs!T, spec.spec).length,
```
wilzbach: How about using `enforceFmt` here?

std/format.d
JackStouffer: Exception doesn't anything. `new` does that. You can easily use stdx.allocator.make and free to throw and catch this exception.

wilzbach:
The auto-tester failure on Windows 32_64 unrelated:

```
The system cannot find the path specified.
failed to execute '\"Program Files (x86)"\"Microsoft Visual Studio 10.0"\VC\bin\amd64\cl.exe /c /nologo runnable\extra-files\cpp_abi_tests.cpp /Fotest_results\runnable\cpp_abi_tests.cpp.obj'
Makefile:133: recipe for target 'test_results/runnable/cpp_abi_tests.d.out' failed
gmake[1]: *** [test_results/runnable/cpp_abi_tests.d.out] Error 1
gmake[1]: *** Waiting for unfinished jobs....
```

See also: https://github.com/braddr/d-tester/issues/71

PR #6368, opened 2018-03-28
Accomodate Enum init deprecation

marler8997:
PR https://github.com/dlang/dmd/pull/7996 is modifying the compiler to comply with the language specification:

> The value of an EnumMember is given by its AssignExpression. If there is no AssignExpression and it is the first EnumMember, its value is EnumBaseType.init.

The current behavior is to assign the value `0` to the first enum member when no assign expression is given.  The referenced PR introduces a deprecation message when an enum's base type's `.init` member is not 0.  This applies to the types `double`, `char` and `real`.  This allows applications to maintain the current behavior for a time until the deprecation is over and the behavior will change to match the specification.  Note that this means you cannot currently create an enum with base type `double`, `char` or `real` without a deprecation message unless you give it's first member an explicit assignment.  Since phobos treats deprecations as errors, this PR provides explicit assignments in these cases during this deprecation period.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -3814,8 +3814,8 @@ void writeln(T...)(T args)
     scope(exit) { std.file.remove(deleteme); }
 
     enum EI : int    { A, B }
-    enum ED : double { A, B }
-    enum EC : char   { A, B }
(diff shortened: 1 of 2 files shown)


JackStouffer:
@WalterBright PRs need at least one approval to merge.

PR #6370, opened 2018-03-28
Fix Issue 18615 - Rebindable!A calls A.opEquals

SimonN:
The issue was that Rebindable!A always compared two rebindable
references by 'a is b', not by A.opEquals. With this fix, two rebindable
references, or a rebindable and a raw reference, always compare with
A.opEquals.

It's my first PR for the D ecosystem. I'll be happy improve the PR based on your feedback, then squash everything once it's good.

Reviewers:
* I haven't considered `opCmp` or `toHash`. I hope that `Rebindable!A.toHash` is automatically forwarded to `A.toHash` via the `alias this`.
* I haven't put any qualifiers on the `Rebindable.opEquals` because `.object.opEquals` doesn't allow any.
* I compare against `cast(const(U))stripped`, not against `T original`. That means that `shared const U` cannot be compared, but at least all tests pass. When I tried to take `T rhs` and compare against `T original`, I got errors:
```
std/typecons.d(1997): Error: none of the overloads of opEquals are callable using argument types (shared(const(Data)), shared(Data)), candidates are:
../druntime/import/object.d(140):        object.opEquals(Object lhs, Object rhs)
../druntime/import/object.d(165):        object.opEquals(const(Object) lhs, const(Object) rhs)
std/typecons.d(2317): Error: template instance `std.typecons.UnqualRef!(shared(Data))` error instantiating
```


wilzbach:
>> I hope that Rebindable!A.toHash is automatically forwarded to A.toHash via the alias this.
> I've verified it does.

For reference: https://run.dlang.io/is/zAwrtO

std/typecons.d
```diff
@@ -2064,6 +2075,46 @@ if (is(T == class) || is(T == interface) || isDynamicArray!T || isAssociativeArr
...
 
+@system unittest // issue 18615; object.opEquals is not @safe
+{
+    class CustomOpEq
+    {
+        int x;
+        override bool opEquals(Object rhsObj)
+        {
+            const rhs = cast(const(CustomOpEq)) rhsObj;
+            return this.x == rhs.x;
```
wilzbach: FYI: in general this should be:

```d
if (auto rhs = cast(const(CustomOpEq)) rhsObj)
    return this.x == rhs.x;
else
    return false;
```

(though obviously that's not necessary for this test)

wilzbach:
The only question that we need to ask ourselves is can this break code and which cases?
I would argue that could that depended/depends on:

```d
class A { int x; }
A a = new A();
auto ca = a.rebindable;
auto cb = b.rebindbale;
assert (ca != cb); 
```

is fundamentally broken and we shouldn't cater for it.

std/typecons.d
SimonN: Thanks, good call, I'll use your code. All unittests should be written like examples even when this particular one isn't exported as documentation.

SimonN:
@wilzbach and @n8sh, I've included your feedback. With 50 lines of unittests, the tests are lengthy compared with how little code I added in Rebindable -- but since equality and `toHash` weren't tested at all before, it should be okay.

SimonN:
Rebased onto `master` and squashed into 1 commit. Ready to be merged. :-)

wilzbach:
> The only question that we need to ask ourselves is can this break code and which cases?
> I would argue that could that depended/depends on:
> ..
> is fundamentally broken and we shouldn't cater for it.

Still it's a potentially breaking change and should be handled/assessed with great care.
What do other people think about this?

(would be nice to get another pairs of eyes on this one)

n8sh:
>Still it's a potentially breaking change and should be handled/assessed with great care.
What do other people think about this?

I think it is unreasonable for code to rely on behavior that contradicts the documentation, and I don't think that is something we should in general cater to.

PR #6372, opened 2018-03-28
Use MIPS32 instead of MIPS

rracariu:
According to https://dlang.org/spec/version.html#predefined-versions there is no `MIPS` version.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -145,12 +145,12 @@ version(DigitalMars)
     version = INLINE_YL2X;        // x87 has opcodes for these
 }
 
-version (X86)    version = X86_Any;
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -149,7 +149,7 @@ version (X86)    version = X86_Any;
 version (X86_64) version = X86_Any;
 version (PPC)    version = PPC_Any;
 version (PPC64)  version = PPC_Any;
-version (MIPS)   version = MIPS_Any;
+version (MIPS32)   version = MIPS_Any;
```
JackStouffer: Fix version alignment

rracariu:
done, thank guys!

PR #6373, opened 2018-03-28
Split DScanner's unused_variable_check check into multiple lines

wilzbach:
Split the most-important DScanner checks into multiple lines.
Maybe it's a good idea to set the DScanner checks that we never intend to fix to disable
or just visually group them with doc comments.


JackStouffer:
Until the issues of false positive’s with ASM statements are resolved a lot of these are going to stay blacklisted for a while :/

wilzbach:
Yeah, there are also these check which probably will probably never get fixed:
:
- `could_be_immutable_check` <- doesn't seem to be ready (a lot of false positives)
- `imports_sortedness`  <- I once tried to push this, but it got reverted
- `style_check` <- to many public violations that can't be fixed easily
- `undocumented_declaration_check` <- way way too many :/

PR #6374, opened 2018-03-29
Add public examples to std.bitmanip + enable DScanner check

wilzbach:


std/bitmanip.d
```diff
@@ -296,6 +262,51 @@ template bitfields(T...)
     enum { bitfields = createStorageAndFields!T.result }
 }
 
+/**
+Creates a bitfield pack of eight bits, which fit in
+one $(D_PARAM ubyte). The bitfields are allocated starting from the
```
JackStouffer: backticks

PR #6375, opened 2018-03-29
posix.mak: Don't rebuild Phobos when running dscanner

wilzbach:
At the moment every execution `make -f posix.mak dscanner` will rebuild Phobos first which is absolutely unnecessary and super annoying.
After all, the dscanner binary only needs to be built once and doesn't need to be rebuilt if a line in Phobos has been changed.

There seems to be no clean way to tell Make to avoid building its order-only dependencies if the final target doesn't need a rebuild anyhow.
So I went with a quick one-line workaround.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -536,7 +536,10 @@ $(DSCANNER_DIR)/dsc: | $(DSCANNER_DIR) $(DMD) $(LIB)
 style: publictests style_lint
 
 # runs static code analysis with Dscanner
-dscanner: | $(DSCANNER_DIR)/dsc
+dscanner:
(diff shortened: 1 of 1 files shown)


CyberShadow:
Why this or order-only dependency instead of a dependency-less target?

wilzbach:
> Why this or order-only dependency instead of a dependency-less target?

As explained order-only dependencies are still rebuilt if their sources change (imho that's a Make bug, but well that's its behavior).
A dependency-less target without the `make` line wouldn't ensure that the Phobos library exists and has been built.

wilzbach:
> This is fine, then, but I think such unobvious things need to be explained in comments in the makefile itself.

OK. Good point. Added an explanatory comment.

PR #6377, opened 2018-03-29
std.algorithm.iteration: Fix a -dip1000 compilable issue

carblue:
aa[std.algorithm.iteration]=-dip1000
Errors when running: make -f posix.mak std/algorithm/iteration.test
...
```
    ../dmd/generated/linux/release/64/dmd -od$T -conf= -I../druntime/import  -w -de -dip25 -m64 -fPIC -transition=complex -O -release -dip1000  -main -unittest generated/linux/release/64/libphobos2.a -defaultlib= -debuglib= -L-ldl -cov -run std/algorithm/iteration.d ;     \
  )
std/algorithm/iteration.d(4245): Error: @safe function std.algorithm.iteration.__unittest_L4227_C7 cannot call @system function std.algorithm.comparison.equal!().equal!(Result, string[]).equal
std/algorithm/iteration.d(5525): Error: static assert:  is(typeof(sum(R)(R r) if (isInputRange!R && !isInfinite!R && is(typeof(r.front + r.front)))((__error)[])) == double) is false
```
The remaining errors are from std/container/slist.d only

### DIFF
--- a/dip1000.mak
+++ b/dip1000.mak
@@ -59,8 +59,8 @@ aa[std.zlib]=-dip1000
 
 aa[std.algorithm.comparison]=-dip1000
 aa[std.algorithm.internal]=-dip1000
-aa[std.algorithm.iteration]=-dip25 #    WIP_carblue
-aa[std.algorithm.mutation]=-dip25 # depends on std.container.slist (https://github.com/dlang/phobos/pull/6295)
+aa[std.algorithm.iteration]=-dip25 # depends on std.container.slist (to be updated https://github.com/dlang/phobos/pull/6295)
+aa[std.algorithm.mutation]=-dip25 #  depends on std.container.slist (to be updated https://github.com/dlang/phobos/pull/6295)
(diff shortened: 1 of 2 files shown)


std/range/package.d
```diff
@@ -9806,7 +9806,7 @@ if (isInputRange!Range)
     {
         return predFun(rhs, lhs);
     }
-    private Range _input;
+    /*private*/ Range _input;
```
wilzbach: That's not really a proper fix and we shouldn't revert to such hacks :/

WalterBright: Why does making it public make it safe?

wilzbach: I just tried this locally and if I set `@trusted:` to `std.container.slist`, I can compile + test std.algorithm.iteration successfully with -dip1000 (e.g. `pmake std/algorithm/iteration.test`)
-> So I suggest to drop this specific change.

carblue: @wilzbach Done as suggested

PR #6378, opened 2018-03-29
Fix Issue 18452 - std.zip has size limit of 2 GB

johnsilver97:
Because data member was indexed by an integer, the upper limit of a file which could be archived was 2GB. I replaced every occurrence with uint and manually tested on my machine with files larger than 2GB. 
Adding a test could be problematic, however we can experiment with memory mapping larger than 2GB.
Do you think it is an overkill?.
Thanks, 
Andrei.

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -764,7 +764,7 @@ final class ZipArchive
         if (de.flags & 1)
             throw new ZipException("encryption not supported");
 
-        int i;
+        uint i;
         i = de.offset + 30 + namelen + extralen;
         if (i + de.compressedSize > endrecOffset)
             throw new ZipException("invalid directory entry 5");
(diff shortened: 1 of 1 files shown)


DmitryOlshansky:
That bumps it to 4G I guess, but I think there is a Zip format extension for 64-bit file sizes?

johnsilver97:
Thanks for the replies @DmitryOlshansky, @n8sh!
@n8sh Could you give me some pointers on what to follow and how zip should behave with the isZip64 flag set?

johnsilver97:
Sorry for the notification, i messed up git rebase.

n8sh:
This is a good PR so I'll go ahead and tag this ``auto-merge``. Further updates can be submitted separately.

RazvanN7:
It looked like there were some spurious failures in jenkins. I restarted the tests, hopefully that will fix the problem

wilzbach:
@RazvanN7 sadly they aren't spurious, but they are unrelated (see https://github.com/dlang/ci/issues/190). So you need to "manually" merge for now :/

PR #6379, opened 2018-03-29
Fix issue 18691 - memory errors in std.regex.Captures

DmitryOlshansky:
Instead of a messy code that ended up with refcount at the wrong place
provide a clean sealed fixed-size array with required semantics and
build on top of that.

### DIFF
--- a/std/regex/internal/ir.d
+++ b/std/regex/internal/ir.d
@@ -899,3 +899,214 @@ struct CharMatcher {
             return trie[ch];
     }
 }
+
+// Internal non-resizeble array, switches between inline storage and CoW
+// POD-only
+struct SmallFixedArray(T, uint SMALL=3)
+if (!hasElaborateDestructor!T)
+{
+    import core.stdc.stdlib : malloc, free;
+    static struct Payload
+    {
+        size_t refcount;
+        T[0] placeholder;
+        inout(T)* ptr() inout { return placeholder.ptr; }
+    }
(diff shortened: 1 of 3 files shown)


std/regex/internal/ir.d
```diff
@@ -899,3 +899,214 @@ struct CharMatcher {
             return trie[ch];
     }
 }
+
+// Internal non-resizeble array, switches between inline storage and CoW
+// POD-only
+struct SmallFixedArray(T, uint SMALL=3)
+if (!hasElaborateDestructor!T)
+{
+    import core.stdc.stdlib : malloc, free;
```
n8sh: How about `core.memory : pureMalloc, pureFree`? Or would there be no benefit because other stuff is impure?

DmitryOlshansky: That’s a good idea. Indeed it won’t make other things pure but worth a shot.

std/regex/package.d
```diff
@@ -681,18 +627,6 @@ public:
 
     ///A hook for compatibility with original std.regex.
     @property ref captures(){ return this; }
-
-    typeof(this) opAssign()(auto ref Captures rhs)
```
JackStouffer: don't you want to manually overload this to copy the Store?

DmitryOlshansky: Why would I - store is copied + bitblitter by default, no?

PR #6382, opened 2018-03-29
Fix Issue 15768 - std.stdio.trustedStdout accesses __gshared data wit…

JackStouffer:
Running the following code

```d
import std.algorithm;
import std.parallelism;
import std.range;
import std.stdio;

void main()
{
    stderr = File("/dev/null", "w");
    
    foreach(t; 1_000_000.iota.parallel)
    {
        stderr.write("aaa");
    }
}
```

dies 100% of the time with v2.079.0, but with this fix it works every time I've tried.

I don't think there can be a test in the test suite for this, so please review carefully.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -4708,7 +4708,18 @@ enum StdFileHandle: string
 
 /** The standard input stream.
 
-    Returns: stdin as a $(LREF File).
+    Returns:
+        stdin as a $(LREF File).
+
+    Note:
+        The returned $(LREF File) wraps $(REF stdin,core,stdio), and
+        is therefore thread global. Reassigning `stdin` to a different
+        `File` must be done in a single-threaded or locked context in
+        order to avoid race conditions.
+
+        All reading from `stdin` automatically locks the file globally,
+        and will cause all other threads calling `read` to wait until
+        the lock is released.
 */
 alias stdin = makeGlobal!(StdFileHandle.stdin);
 
@@ -4735,7 +4746,18 @@ alias stdin = makeGlobal!(StdFileHandle.stdin);
 /**
     The standard output stream.
 
-    Returns: stdout as a $(LREF File).
+    Returns:
+        stdout as a $(LREF File).
+
+    Note:
+        The returned $(LREF File) wraps $(REF stdout,core,stdio), and
+        is therefore thread global. Reassigning `stdout` to a different
+        `File` must be done in a single-threaded or locked context in
+        order to avoid race conditions.
+
+        All writing to `stdout` automatically locks the file globally,
+        and will cause all other threads calling `write` to wait until
+        the lock is released.
 */
 alias stdout = makeGlobal!(StdFileHandle.stdout);
 
@@ -4787,7 +4809,18 @@ alias stdout = makeGlobal!(StdFileHandle.stdout);
 /**
     The standard error stream.
 
-    Returns: stderr as a $(LREF File).
+    Returns:
+        stderr as a $(LREF File).
+
+    Note:
+        The returned $(LREF File) wraps $(REF stderr,core,stdio), and
+        is therefore thread global. Reassigning `stderr` to a different
+        `File` must be done in a single-threaded or locked context in
+        order to avoid race conditions.
+
+        All writing to `stderr` automatically locks the file globally,
+        and will cause all other threads calling `write` to wait until
+        the lock is released.
 */
 alias stderr = makeGlobal!(StdFileHandle.stderr);
 
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -369,6 +369,7 @@ Hello, Jimmy!
  */
 struct File
 {
+    import core.atomic : atomicOp, atomicStore, atomicLoad;
     import std.range.primitives : ElementEncodingType;
     import std.traits : isScalarType, isArray;
     enum Orientation { unknown, narrow, wide }
@@ -376,7 +377,7 @@ struct File
     private struct Impl
     {
         FILE * handle = null; // Is null iff this Impl is closed by another File
-        uint refs = uint.max / 2;
+        shared uint refs = uint.max / 2;
         bool isPopened; // true iff the stream has been created by popen()
         Orientation orientation;
     }
@@ -397,7 +398,7 @@ struct File
     {
         assert(_p);
         _p.handle = handle;
-        _p.refs = refs;
+        atomicStore(_p.refs, refs);
         _p.isPopened = isPopened;
         _p.orientation = Orientation.unknown;
         _name = name;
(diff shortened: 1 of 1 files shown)


std/stdio.d
```diff
@@ -843,12 +844,12 @@ Throws: $(D ErrnoException) on failure if closing the file.
     void detach() @safe
     {
         if (!_p) return;
-        if (_p.refs == 1)
+        if (atomicLoad(_p.refs) == 1)
```
n8sh: This is a potential resource leak. Example:

Two threads have references. Both check count, and see that it is 2. Both skip this branch and so the file is not closed. What needs to instead happen is unconditional `atomic!"-="` and check the result of that.

So in the new example:
Two threads have references. Both atomically decrement. One sees `0` as the new value and the other sees `1` as the new value. Whichever sees `0` then calls close.

std/stdio.d
```diff
@@ -513,14 +514,16 @@ Throws: $(D ErrnoException) in case of error.
...
-        else if (_p.refs == 1)
         {
-            closeHandles();
-        }
-        else
-        {
-            _p.refs--;
-            _p = cast(Impl*) enforce(malloc(Impl.sizeof), "Out of memory");
+            atomicOp!"-="(_p.refs, 1);
+            if (atomicLoad(_p.refs) == 0)
```
n8sh: You can save an atomic operation: `if (atomicOp!"-="(_p.refs, 1) == 0)`

std/stdio.d
```diff
@@ -887,8 +881,7 @@ Throws: $(D ErrnoException) on error.
         if (!_p) return; // succeed vacuously
         scope(exit)
         {
-            assert(_p.refs);
-            if (!--_p.refs)
+            if (atomicLoad(_p.refs) == 0)
```
andralex: This is different from the existing code - it doesn't decrement anymore.

jercaianu:
@JackStouffer 
For testing, can't we redirect stdout to a file, write something in there from a couple of threads and then restore stdout?

wilzbach:
> For testing, can't we redirect stdout to a file, write something in there from a couple of threads and then restore stdout?

Well, the test that Jack posted above allows to perfectly reproduce this bug. It just takes ~0.3s on my system. Maybe we can have a special folder (e.g. `stresstests`), s.t. these tests aren't run by default for the `unittest` target?

JackStouffer:
Added test under a new version branch.

std/stdio.d
```diff
@@ -3427,6 +3425,24 @@ void main()
...
+    // issue 15768
+    @system unittest
+    {
+        import std.parallelism : parallel;
+        import std.range : iota;
+
+        auto deleteme = testFilename();
+        stderr = File(deleteme, "w");
+
+        foreach(t; 1_000_000.iota.parallel)
```
wilzbach: std/stdio.d:3439:        foreach(t; 1_000_000.iota.parallel)
posix.mak:544: recipe for target 'style_lint' failed

JackStouffer:
Can anyone that uses Windows confirm or deny that 13727 is fixed by this?

JackStouffer:
I'd need to have the test be run on windows on the auto tester. And I'd assume we don't want to run that for every PR.

wilzbach:
Well, we can definitely run it once here for validation?
(also the test case that Walter wanted to add it in dlang/dmd#5747 only uses 1000 runs.)

anon17:
13727 has a C test, it doesn't depend on phobos.

anon17:
Also don't remove comments about unsafe reassignment. You only fixed that reading them is thread-safe after single publishing, but concurrent writing is still subject to race condition.

JackStouffer:
How so? Writing is done with LockingTextWriter, which uses OS file locks.

schveiguy:
I think he means writing the handle itself, not writing data to the handle.

But the warning needs to be reworded -- the handle is subject to race conditions after it's initialized, for reading OR writing. It has nothing to do with the issue this PR is fixing.

JackStouffer:
What specific race conditions are you talking about?

schveiguy:
Let me back up a second here -- this PR is fixing an issue, but it doesn't appear to be the issue identified by the bug report's description. It appears that the test case given in the bug report has nothing to do with a race condition on modifying `stdout`, but improperly implemented reference counting inside `File`.

My recommendation: split the bug into another report, as there are really 2 bugs here -- one is the `File` refcount implementation, and the other is the race condition you could have from `trustedStdout` returning `stdout` without locking.

I'm not sure we even want to add a mutex lock to stdout. Most of the time, it doesn't change ever, or if it does, you are doing it early on when only one thread exists. Maybe just updating the documentation to reflect the rules is the proper fix.

schveiguy:
> What specific race conditions are you talking about?

Classic race condition with shared data. 2 or more threads writing the `stdout` handle without locking, or 1 thread writing the `stdout` handle while another thread is calling `trustedStdout` to read it without locking. In order to properly provide a way to modify the `stdout` handle, we would need locking.

The other option is to document that you can only change the handles from a single thread, or you get race/undefined behavior.

JackStouffer:
The test case shows the assignment outside of the threaded code because the bug in the bug report is specifically about the ref counting in File not working in a multi-threaded environment, and how that's an issue for stdout/in/err. It could possibly be an issue that a thread could re-assign stdout to another file instance in multi-threaded code, if that's what you're referring to.

In that case, the nature of stdout/in/err being global due to the nature of C libraries should be documented, and the fact that calls to write/ln auto lock the file. I really wouldn't call writing to stdout a "race condition" in the normal usage of the phrase because there are locks around writing, and just like in any well designed multi-threaded program you'll need some queue equivalent to make sure data is written to the file in the correct order. Once a chunk of data starts to be written to the file, that entire chunk will successfully write.

schveiguy:
> It could possibly be an issue that a thread could re-assign stdout to another file instance in multi-threaded code, if that's what you're referring to.

Yes.

> I really wouldn't call writing to stdout a "race condition" in the normal usage of the phrase because there are locks around writing

Again, I'm not talking about writing data to stdout, but reassigning stdout itself to a different handle (e.g. `stdout = File("output.txt");`). There are no protections around this.

The original warning was worded poorly anyway -- it said reassigning stdout was dangerous due to bug 15768, but really any use of any of the standard handles (regardless of assignment) was dangerous.

In any case, your new information is good, so I think it's solved.

schveiguy:
The title of the bug report is the problem (it's talking about `trustedStdout` which simply returns stdout). I'll just change it, so it doesn't confuse people.

anon17:
`$(REF stdin,core,stdout)` you probably meant `$(REF stdin,core,stdio)`. Same for stderr.

PR #6384, opened 2018-03-29
Make std.socket DIP1000 compatible

wilzbach:
From @WalterBright:

> How about simply making std.socket @System so it will compile with -dip1000 and move on?

https://github.com/dlang/phobos/pull/6204#discussion_r177646821


std/socket.d
```diff
@@ -2477,13 +2477,19 @@ public:
             assert(!errorSet.isSet(testPair[1]));
 
             ubyte[1] b;
-            testPair[0].send(b[]);
+            // std.socket isn't fully @safe anymore - see https://github.com/dlang/phobos/pull/6204
```
klickverbot: This comment seems odd/wrong – std.socket is just as safe as before, it's just that the `send`/`receive` arguments aren't marked as `scope`, so *client code* isn't safe anymore with dip1000.

wilzbach: Ok. Tried to improve the wording.

PR #6385, opened 2018-03-30
Add more public examples to std.exception + enable DScanner check

wilzbach:


std/exception.d
```diff
@@ -660,6 +660,15 @@ if (is(typeof(new E("", string.init, size_t.init))))
     }
 }
 
+///
+deprecated
```
n8sh: If we're telling people not to use this it seems weird to have an example of how to use it.

wilzbach: Fair point. Let's fix DScanner then: https://github.com/dlang-community/D-Scanner/pull/576

PR #5729, opened 2017-09-10
Change AliasSeq(0, ...) uses of foreach to static foreach

wilzbach:
I started to have a quick look at introducing more `static foreach` at Phobos. 
Sadly even the naive idea to mark all `foreach` loops which definitely are at Compile-Time - e.g. with `AliasSeq` doesn't work:

```d
sed -E "s/foreach(.*)AliasSeq/static foreach\1AliasSeq/" -i **/*.d
```

or in simpler words with a simplified example from Phobos (most of the uses of `foreach` + `AliasSeq` are like this):

```
import std.stdio, std.meta;
void main(string[] args)
{
    foreach (S; AliasSeq!(string, wstring, dstring))
    {
        auto s = S.stringof;
        writeln(s);
    }
}
```
https://run.dlang.io/gist/6ae32becc7ffaa1296a1a4752cebc294?compiler=dmd

And now with `static foreach` this fails due to the symbol being in the outer scope:

```
import std.stdio, std.meta;
void main(string[] args)
{
    static foreach (S; AliasSeq!(string, wstring, dstring))
    {
        auto s = S.stringof; // Error: declaration foo.main.s is already defined
        writeln(s);
    }
}
```

FWIW [Local Declarations](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1010.md#local-declarations) would solve this and FYI @tgehr mentioned that [he has a working implementation of `__local` already](http://forum.dlang.org/post/ooo9kt$1dnf$1@digitalmars.com).

PR #6386, opened 2018-03-30
Fix Issue 14637 - Array operations should work on tuples

wilzbach:
Other operations that have been missed?

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -565,7 +565,7 @@ if (distinctFieldNames!(Specs))
     enum areBuildCompatibleTuples(Tup1, Tup2) = isTuple!Tup2 && is(typeof(
     {
         static assert(Tup1.Types.length == Tup2.Types.length);
-        foreach (i, _; Tup1.Types)
+        static foreach (i, _; Tup1.Types)
             static assert(isBuildable!(Tup1.Types[i], Tup2.Types[i]));
     }));
 
@@ -575,7 +575,7 @@ if (distinctFieldNames!(Specs))
         U u = U.init;
         T t = u;
     }));
-    /+ Helper for partial instanciation +/
+    /+ Helper for partial instantiation +/
     template isBuildableFrom(U)
     {
         enum isBuildableFrom(T) = isBuildable!(T, U);
@@ -587,6 +587,8 @@ if (distinctFieldNames!(Specs))
          * The types of the `Tuple`'s components.
          */
         alias Types = staticMap!(extractType, fieldSpecs);
+
+        private alias _Fields = Specs;
 
         ///
         static if (Specs.length == 0) @safe unittest
@@ -768,6 +770,17 @@ if (distinctFieldNames!(Specs))
             return field[] == rhs.field[];
         }
 
+        /// ditto
+        bool opEquals(R...)(auto ref R rhs)
+        if (R.length > 1 && areCompatibleTuples!(typeof(this), Tuple!R, "=="))
+        {
+            static foreach (i, _; Types)
+                if (field[i] != rhs[i])
+                    return false;
+
+            return true;
+        }
+
         ///
         static if (Specs.length == 0) @safe unittest
         {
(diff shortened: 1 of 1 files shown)


std/typecons.d
```diff
@@ -768,6 +768,12 @@ if (distinctFieldNames!(Specs))
             return field[] == rhs.field[];
         }
 
+        bool opEquals(R...)(R rhs) const
+        if (R.length > 1 && areCompatibleTuples!(typeof(this), Tuple!R, "=="))
+        {
+            return field[] == tuple(rhs).field[];
+        }
```
wilzbach: This is required for `static assert(t.fieldNames == tuple("foo", "bar"));`

std/typecons.d
```diff
@@ -565,7 +565,7 @@ if (distinctFieldNames!(Specs))
     enum areBuildCompatibleTuples(Tup1, Tup2) = isTuple!Tup2 && is(typeof(
     {
         static assert(Tup1.Types.length == Tup2.Types.length);
-        foreach (i, _; Tup1.Types)
+        static foreach (i, _; Tup1.Types)
```
n8sh: Since all that's happening in this is a static assert, is there a reason to prefer `static foreach` over `foreach`?

wilzbach: Style/clarity (see e.g. https://github.com/dlang/phobos/pull/5729)

std/typecons.d
```diff
@@ -768,6 +770,12 @@ if (distinctFieldNames!(Specs))
             return field[] == rhs.field[];
         }
 
+        bool opEquals(R...)(R rhs) const
```
andralex: auto ref may be helpful here

std/typecons.d
```diff
@@ -768,6 +770,12 @@ if (distinctFieldNames!(Specs))
...
+        if (R.length > 1 && areCompatibleTuples!(typeof(this), Tuple!R, "=="))
```
andralex: Does `areCompatibleTuples` also include the length check?

std/typecons.d
```diff
@@ -768,6 +770,12 @@ if (distinctFieldNames!(Specs))
...
+        {
+            return field[] == tuple(rhs).field[];
```
andralex: This is wasteful if `field[] == rhs.field[]` works, which would be more efficient. Why is there a need to create a temp?

std/typecons.d
```diff
@@ -836,6 +844,26 @@ if (distinctFieldNames!(Specs))
...
+            else
+                return Tuple!(_Fields, T)(expand, t);
+        }
+
+        /// ditto
+        auto opBinaryRight(string op, T)(T t)
+        if (op == "~")
+        {
+            static if (isTuple!T)
+                return Tuple!(AliasSeq!(T._Fields, _Fields))(t.expand, expand);
```
andralex: I wonder how this doesn't create an ambiguity when concatenating two tuples.

std/typecons.d
```diff
@@ -836,6 +844,26 @@ if (distinctFieldNames!(Specs))
             assert(tup1 > tup2);
         }
 
+        /// Allow tuple concatenation
+        auto opBinary(string op, T)(T t)
+        if (op == "~")
+        {
+            static if (isTuple!T)
+                return Tuple!(AliasSeq!(_Fields, T._Fields))(expand, t.expand);
```
MetaLang: I think the use of `AliasSeq` here is redundant, but I'm not completely sure without checking it.

std/typecons.d
MetaLang: Same here with redundant `AliasSeq`.

std/typecons.d
wilzbach: Because `R` isn't a tuple - without it I get:

```
std/typecons.d(776): Error: no property field for tuple (string, string)
```

std/typecons.d
wilzbach: Yes.

std/typecons.d
andralex: Got it, thanks. Well the template constraint would at best describe what the requirements on R are; currently it relies on `areCompatibleTuples`, which is utterly undocumented.

std/typecons.d
wilzbach: Yeah, but then DMD can no longer find a matching `opEquals`

std/typecons.d
wilzbach: https://dlang.org/phobos/std_typecons.html#.Tuple.opEquals no?

std/typecons.d
JackStouffer: Can you file a bug for `auto ref opEquals`?

std/typecons.d
```diff
@@ -768,6 +770,16 @@ if (distinctFieldNames!(Specs))
             return field[] == rhs.field[];
         }
 
+        bool opEquals(R...)(auto ref const R rhs) const
```
wilzbach: BTW shouldn't the `opEquals` overloads above also use `const`?

std/typecons.d
wilzbach: (with the latest constraint + body changes, I can no longer reproduce this -> changed to `const auto ref`)

std/typecons.d
```diff
@@ -768,6 +770,16 @@ if (distinctFieldNames!(Specs))
             return field[] == rhs.field[];
         }
 
+        bool opEquals(R...)(const auto ref R rhs) const
```
MetaLang: I don't think `const` is a good idea here:

```
struct Test(T...)
{
    T field;
    
    this (T t) { field = t; }
    
    bool opEquals(R...)(const auto ref R rhs) const
    {
        static foreach (i, _; T)
            if (field[i] != rhs[i])
            	return false;
            
        return true;
    }
}

struct Bad
{
    int a;
    
    bool opEquals(Bad b)
    {
        return a == b.a;
    }
}

void main()
{
	import std.meta;
    auto t = Test!(int, Bad, string)(1, Bad(1), "asdf");
    assert(t == AliasSeq!(1, Bad(1), "asdf"));
}
```

wilzbach: Urgh. That really sucks. Well, we could duplicate the `opEquals` yet another time, but for now I just removed both `const`'s and added your example to the source code, s.t. we don't regress against this.
Thanks!

std/typecons.d
```diff
@@ -836,6 +848,26 @@ if (distinctFieldNames!(Specs))
             assert(tup1 > tup2);
         }
 
+        /// Allow tuple concatenation
```
JackStouffer: Params, Returns

std/typecons.d
```diff
@@ -768,6 +770,16 @@ if (distinctFieldNames!(Specs))
             return field[] == rhs.field[];
         }
 
+        bool opEquals(R...)(auto ref R rhs)
```
JackStouffer: not documented

wilzbach:
Added a nice error message if the two concatenated tuples have non-distinct fields.

PR #6387, opened 2018-03-30
std.container.slist: Fix -dip1000 compilable issues

carblue:
A new attempt since https://github.com/dlang/phobos/pull/6295

aa[std.container.slist]=-dip1000
Errors when running: make -f posix.mak std/container/slist.test
...
```
    ../dmd/generated/linux/release/64/dmd -od$T -conf= -I../druntime/import  -w -de -dip25 -m64 -fPIC -transition=complex -O -release -dip1000 -version=DIP1000  -main -unittest generated/linux/release/64/libphobos2.a -defaultlib= -debuglib= -L-ldl -cov -run std/container/slist.d ;     \
  )
std/container/slist.d(280): Error: @safe function std.container.slist.SList!int.SList.__unittest_L278_C11 cannot call @system constructor std.container.slist.SList!int.SList.__ctor!int.this
std/container/slist.d(282): Error: @safe function std.container.slist.SList!int.SList.__unittest_L278_C11 cannot call @system constructor std.container.slist.SList!int.SList.__ctor!int.this
std/container/slist.d(27): Error: template instance `std.container.slist.SList!int` error instantiating
std/container/slist.d(801): Error: template instance `std.container.slist.SList!string` error instantiating
```
Changing to:
size_t insertFront(Stuff)(scope Stuff stuff)
if (isInputRange!Stuff && isImplicitlyConvertible!(ElementType!Stuff, T))

compiles everything except:  auto sl = SList!string(["a", "b", "d"]);
```
std/container/slist.d(365): Error: scope variable item may not be copied into allocated memory
std/container/slist.d(154): Error: template instance `std.container.slist.SList!string.SList.insertFront!(Range)` error instantiating
std/container/slist.d(250):        instantiated from here: __ctor!(Range)
std/container/slist.d(801):        instantiated from here: SList!string
```
Thus I'm going to try with a dedicated constructor (or ?) for the string types, probably including dup-ing consructor arguments. We'll see.

### DIFF
--- a/std/container/slist.d
+++ b/std/container/slist.d
@@ -354,7 +354,7 @@ Returns: The number of elements inserted
 
 Complexity: $(BIGOH m), where $(D m) is the length of $(D stuff)
      */
-    size_t insertFront(Stuff)(Stuff stuff)
+    size_t insertFront(Stuff)(scope Stuff stuff)
     if (isInputRange!Stuff && isImplicitlyConvertible!(ElementType!Stuff, T))
     {
         initialize();
@@ -791,7 +791,7 @@ Complexity: $(BIGOH n)
     assert(s.insertAfter(r, 5) == 1);
     assert(s == SList!int(1, 2, 5, 3, 4));
 }
-
+/+
 @safe unittest
 {
     import std.algorithm.comparison : equal;
@@ -804,7 +804,7 @@ Complexity: $(BIGOH n)
     sl.insertAfter(take(sl[], 2), "c"); // insert after "b"
     assert(equal(sl[], ["a", "b", "c", "d", "e"]));
 }
-
++/
 @safe unittest
 {
     import std.range.primitives;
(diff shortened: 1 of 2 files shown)


std/container/slist.d
```diff
@@ -354,15 +354,18 @@ Returns: The number of elements inserted
...
     if (isInputRange!Stuff && isImplicitlyConvertible!(ElementType!Stuff, T))
     {
         initialize();
         size_t result;
         Node * n, newRoot;
         foreach (item; stuff)
         {
-            auto newNode = new Node(null, item);
+            static if (isSomeString!T)
+                auto newNode = new Node(null, item.dup);
```
wilzbach: Didn't we agree that this is bad because we duplicate memory without the user explicitly agreeing too.
Why not `version(DIP1000) else` out the test for now?  And/or report the DMD error I reduced here: https://github.com/dlang/phobos/pull/6295#issuecomment-375813328

klickverbot: Also, wouldn't special-casing strings this way break all other reference types anyway?

PR #6390, opened 2018-03-30
Followup to addition of apply

andralex:
* inline code words easiest as backticks
* prefer https to http
* apply should not be "opinionated" about value vs reference, just transparently pass arguments through

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -3445,9 +3445,9 @@ if (is (typeof(nullValue) == T))
 /**
 Unpacks the content of a $(D Nullable), performs an operation and packs it again. Does nothing if isNull.
 
-When called on a $(D Nullable), `apply` will unpack the value contained in the $(D Nullable),
-pass it to the function you provide and wrap the result in another $(D Nullable) (if necessary).
-If the Nullable is null, `apply` will return null itself.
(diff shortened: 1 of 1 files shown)


andralex:
see also https://github.com/dlang/phobos/pull/6182
cc @FeepingCreature

andralex:
CI shouldn't fail... cc @MartinNowak

std/typecons.d
```diff
@@ -3441,28 +3441,29 @@ if (is (typeof(nullValue) == T))
...
 }
 
+// apply
 /**
 Unpacks the content of a $(D Nullable), performs an operation and packs it again. Does nothing if isNull.
 
-When called on a $(D Nullable), $(D apply) will unpack the value contained in the $(D Nullable),
-pass it to the function you provide and wrap the result in another $(D Nullable) (if necessary).
-If the Nullable is null, $(D apply) will return null itself.
+When called on a `Nullable`, `apply` will unpack the value contained in the `Nullable`,
```
wilzbach: I think we should continue the automation alternative here. At the moment we only converted std.algorithm automatically

wilzbach:
Fix for the Project Tester incoming -> https://github.com/dlang/ci/pull/189
I can't do much about auto-tester sorry.

PR #6391, opened 2018-03-30
Markdownify Phobos

wilzbach:
$(D word) -> `word`

-----

Done with: https://github.com/wilzbach/dscanner-fix/blob/master/d_code_markdown.d

Continuation of the first experiment in https://github.com/dlang/phobos/pull/5970

Replacement of #5801 

CC @andralex 

(the script doesn't replace everything, it doesn't try to deal with a few tricky instances. They can be manually looked afterwords. `D_PARAM` isn't touched either)


std/container/package.d
```diff
@@ -222,279 +222,279 @@ $(TR
...
+    $(TDNW `c.removeKey(k)`)
+    $(TDNW `log n$(SUBSCRIPT c)`)
+    $(TD Removes an element from `c` by using its key `k`.
          The key's type is defined by the _container.)
 )
 $(TR
-    $(TDNW $(D ))
-    $(TDNW $(D ))
+    $(TDNW ``)
+    $(TDNW ``)
```
wilzbach: This one is scary.

etc/c/odbc/sql.d
```diff
@@ -15,7 +15,7 @@ typically not a part of the distribution packages of other compilers
 such as `gdc` and `ldc`). To use ODBC, install it per the vendor- and
 platform-specific instructions and then use the appropriate command-line
 flags (e.g. for dmd, `-L-lodbc` on Posix and `-Lodbc32.lib` on Windows) to link
-with the ODBC library. On Windows, using $(D pragma(lib, "odbc32")) in D
+with the ODBC library. On Windows, using `pragma(lib, "odbc32")` in D
```
andralex: Only single-word (no spaces) inline code should be in backquotes. This is because many editors reflow comments automatically, and we don't support backquoted code spanning multiple lines.

wilzbach: Hmm, smart editors should be able to handle that, but restricted to single-words for now then.

CyberShadow: You mean, smart editors should be able to handle that once someone writes the code for it? I don't think anything is smart enough to do this correctly right now.

wilzbach:
@JackStouffer rebased though I guess this will easily run into a conflict again...

JackStouffer:
@wilzbach Rebase and then ping me on slack, and then I'll merge.

wilzbach:
Thanks! & rebased.

wilzbach:
jenkins is still failing due to https://github.com/dlang/ci/issues/190

CyberShadow:
DAutoTest shows a lot of diffs, but the diffs reveal that previously the HTML was really broken in some places - text was outright missing. Was a macro previously accidentally undefined? @wilzbach Any idea what happened there?

wilzbach:
> DAutoTest shows a lot of diffs, but the diffs reveal that previously the HTML was really broken in some places 

-> https://github.com/dlang/dlang.org/pull/2326

PR #6392, opened 2018-03-30
Allow to add additional DFLAGS when building Phobos

wilzbach:
https://github.com/dlang/dmd/pull/8089 for Phobos, s.t. downstream packers can set their custom flags and don't need to patch `posix.mak` like e.g. https://github.com/dlang/phobos/pull/5966

It's also useful for quickly applying a flag for local testing:

```
make -f posix.mak std/algorithm/iteration.test DFLAGS="-debug=MyDebug"
```

For reference, environment variables aren't read, so global `DFLAGS` won't start to influence this Makefile, e.g.

```
DFLAGS="-debug=MyDebug" make -f posix.mak std/algorithm/iteration.test
```

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -120,19 +120,20 @@ else
 endif
 
 # Set DFLAGS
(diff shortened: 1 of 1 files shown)


wilzbach:
FYI the analog [DMD PR](https://github.com/dlang/dmd/pull/8089) has been merged a few days ago now.

PR #6393, opened 2018-03-31
Add public examples to std.random + enable DScanner check

wilzbach:
+ a few small documentation updates

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -109,6 +109,9 @@ import std.traits;
 ///
 @safe unittest
 {
+    import std.algorithm.comparison : among, equal;
+    import std.range : iota;
+
     // seed a random generator with a constant
     auto rnd = Random(42);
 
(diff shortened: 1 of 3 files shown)


std/random.d
```diff
@@ -2397,40 +2584,7 @@ Returns:
...
-
-foreach (e; randomCover(a, rndGen))  // DANGEROUS!! rndGen gets copied by value
-{
-    writeln(e);
-}
-
-foreach (e; randomCover(a, rndGen))  // ... so this second random cover
-{                                    // will output the same sequence as
-    writeln(e);                      // the previous one.
-}
```
wilzbach: This is probably the only "controversial" bit. While std.random is far from perfect, this particular nastiness isn't there anymore as now the overloads are `auto ref`:

https://run.dlang.io/is/6N3Nb9

std/traits.d
```diff
@@ -4727,12 +4727,15 @@ template CommonType(T...)
...
 }
+
+///
 @safe unittest
 {
     static assert(is(CommonType!(3) == int));
     static assert(is(CommonType!(double, 4, float) == double));
     static assert(is(CommonType!(string, char[]) == const(char)[]));
     static assert(is(CommonType!(3, 3U) == uint));
+    static assert(is(CommonType!(double, int) == double));
```
wilzbach: Moved this line from the `uniformDistribution`.

std/random.d
```diff
@@ -452,6 +536,28 @@ Always $(D false) (random generators are infinite ranges).
...
+    auto rnd = CPP11LCG(42);
+    auto n = rnd.front; // same for each run
+    assert(n == 2027382);
+}
+
+/// Declare your own linear congruential engine
+@safe unittest
+{
+    // glibc's LCG
+    alias GLibcLCG= LinearCongruentialEngine!(uint, 1103515245, 12345, 2_147_483_648);
```
JackStouffer: space around =

wilzbach:
So I bumped into https://issues.dlang.org/show_bug.cgi?id=15147 (seeded PRNGs of std.random are platform-dependent).
This means even if the seed is known, they return different values on different platforms. This is really nasty.
I guess I either have to fix this or I can't `assert` on the value :/

JackStouffer:
Just version out the asserts for now. This is a documentation PR, so document the bugs.

wilzbach:
> Just version out the asserts for now. This is a documentation PR, so document the bugs.

Ok, there are now 9 "FIXME" `version` block - is this what you expected?

std/random.d
```diff
@@ -121,9 +124,41 @@ import std.traits;
...
+
+    // Select a random subsample from a range
+    assert(10.iota.randomSample(3, rnd2).equal([7, 8, 9]));
+
+    // Cover all elements in an array in random order
+    version(X86_64) // FIXME
+    assert(10.iota.randomCover(rnd2).equal([7, 4, 2, 0, 1, 6, 8, 3, 9, 5]));
+
+    // Shuffle an array
+    version(X86_64) // FIXME
```
JackStouffer: I would replace all instances of `FIXME` with `Issue XXXXXX`

wilzbach: Sure. Done.

std/random.d
```diff
@@ -121,9 +124,41 @@ import std.traits;
...
+    // Select a random subsample from a range
+    assert(10.iota.randomSample(3, rnd2).equal([7, 8, 9]));
+
+    // Cover all elements in an array in random order
+    version(X86_64) // Issue 15147
+    assert(10.iota.randomCover(rnd2).equal([7, 4, 2, 0, 1, 6, 8, 3, 9, 5]));
+
+    // Shuffle an array
+    version(X86_64) // Issue 15147
+    assert([0, 1, 2, 4, 5].randomShuffle(rnd2).equal([2, 0, 4, 5, 1]));
```
ibuclaw, line 161: This doesn't work for X32 targets which print the same as 32.
```
[4, 8, 7, 3, 5, 9, 2, 6, 0, 1]
[4, 2, 5, 0, 1]
```

PR #6394, opened 2018-03-31
std.random: add a module-level booktable

wilzbach:
image

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -3,6 +3,57 @@
 /**
 Facilities for random number generation.
 
+$(SCRIPT inhibitQuickIndex = 1;)
+$(BOOKTABLE,
+$(TR $(TH Category) $(TH Functions))
+$(TR $(TD Uniform sampling) $(TD
+    $(LREF uniform)
+    $(LREF uniform01)
+    $(LREF uniformDistribution)
+))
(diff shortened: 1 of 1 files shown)


std/random.d
```diff
@@ -3,6 +3,57 @@
...
+    $(LREF dice)
+))
+$(TR $(TD Range sampling) $(TD
+    $(LREF randomCover)
+    $(LREF randomSample)
+))
+$(TR $(TD Default Random Engines) $(TD
+    $(LREF rndGen)
+    $(LREF Random)
+    $(LREF unpredictableSeed)
```
wilzbach: `unpredictableSeed` isn't really an engine, but it didn't fit anywhere else

std/random.d
```diff
@@ -3,6 +3,57 @@
...
+))
+$(TR $(TD Xorshift Engines) $(TD
+    $(LREF Xorshift)
+    $(LREF XorshiftEngine)
+    $(LREF Xorshift32)
+    $(LREF Xorshift64)
+    $(LREF Xorshift96)
+    $(LREF Xorshift128)
+    $(LREF Xorshift160)
+    $(LREF Xorshift192)
```
wilzbach: Not sure whether it's really worth listing all variants in the quick table

PR #6395, opened 2018-03-31
Add public examples to std.csv + enable DScanner check

wilzbach:
Another one.


std/csv.d
```diff
@@ -424,6 +366,101 @@ if (isInputRange!Range && is(Unqual!(ElementType!Range) == dchar)
...
+    string text = "a,b,c\nHello,65,63.63\nWorld,123,3673.562";
+    auto records = text.csvReader!int(["b"]);
+
+    assert(records.equal!equal([
+        [65],
+        [123],
+    ]));
+}
+
+/// Read from header of different order
```
JackStouffer: "Read while rearranging the columns by specifying a header with a different order"

std/csv.d
```diff
@@ -424,6 +366,101 @@ if (isInputRange!Range && is(Unqual!(ElementType!Range) == dchar)
...
+
+    auto records = text.csvReader!Layout(["b","c","a"]);
+    assert(records.equal([
+        Layout(65, 2.5, "Hello"),
+        Layout(123, 7.5, "World")
+    ]));
+}
+
+/**
+The header can also be left empty if the input contains a header but
```
JackStouffer: >The header can also be left empty if the input contains a header **row** ~~but~~ **and** all columns should be iterated

PR #6359, opened 2018-03-28
Prepare std.algorithm.{searching,sorting} for -dip1000

wilzbach:
- with this fix std.algorithm.searching only depends on std.container.slist to be @safe (see e.g. https://github.com/dlang/phobos/pull/6295)
- std.algorithm.sorting still has one issue with dip1000 afterwards

Here's the reduced issue of std.algorithm.sorting that still remains:

```
ref front(T)(T[] a) {
    return a[0];
}

struct Merge(Rs...)
{
    Rs source;
    auto ref front()
    {
        foreach (i, _; Rs)
            return source[i].front;
    }

}

auto merge(Rs...)(Rs rs)
{
    return Merge!Rs(rs);
}

void main() @safe
{
    struct DummyType
    {
        int payload;
        ref inout(typeof(payload)) front() inout
        {
            return payload;
        }
    }
    DummyType d;
    d.merge([1.]);
}
```

```
test.d(8): Error: function type pure nothrow @nogc return @safe double() has return but does not return any indirections
```

PR #6396, opened 2018-03-31
std.range.package.d: -dip1000 compilable

carblue:
aa[std.range.package]=-dip1000
Errors when running: make -f posix.mak std/range/package.test
...
```
    ../dmd/generated/linux/release/64/dmd -od$T -conf= -I../druntime/import  -w -de -dip25 -m64 -fPIC -transition=complex -O -release -dip1000  -main -unittest generated/linux/release/64/libphobos2.a -defaultlib= -debuglib= -L-ldl -cov -run std/range/package.d ;     \
  )
std/range/package.d(1703): Error: reference to local variable a assigned to non-scope parameter _param_1 calling std.range.chooseAmong!(RefAccessRange, RefAccessRange).chooseAmong
std/range/package.d(1703): Error: reference to local variable b assigned to non-scope parameter _param_2 calling std.range.chooseAmong!(RefAccessRange, RefAccessRange).chooseAmong
std/range/package.d(5089): Error: scope variable r1 assigned to non-scope parameter _param_0 calling std.range.zip!(NonSliceableRandomAccess, NonSliceableRandomAccess).zip
std/range/package.d(5089): Error: scope variable r2 assigned to non-scope parameter _param_1 calling std.range.zip!(NonSliceableRandomAccess, NonSliceableRandomAccess).zip
std/range/package.d(5090):        while evaluating: static assert(isRandomAccessRange!(typeof(__error)))
```
Maybe, the changes to the affected unittests are too drastic in that they aren't @nogc and don't operate with static arrays any more.
The PR shows, what's working without changing std.range algorithms.

---
Starting a countdown: The last 5 (6?) phobos modules, that haven't -dip1000 set (apart from those in the PR pipeline):
std.format
std.outbuffer
std.path
(std.algorithm.searching   https://github.com/dlang/phobos/pull/6359 states, it depends on std.container.slist only)
std.algorithm.sorting   see also https://github.com/dlang/phobos/pull/6359
std.experimental.logger.filelogger

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -1713,25 +1713,25 @@ if (Ranges.length >= 2
     assert(c[0] == 42);
 }
 
-@safe nothrow pure @nogc unittest
+@safe nothrow pure /*@nogc*/ unittest
 {
     static struct RefAccessRange
     {
         int[] r;
-        ref front() @property { return r[0]; }
-        ref back() @property { return r[$ - 1]; }
+        scope ref front() @property { return r[0]; }
+        scope ref back() @property { return r[$ - 1]; }
         void popFront() { r = r[1 .. $]; }
         void popBack() { r = r[0 .. $ - 1]; }
         auto empty() @property { return r.empty; }
-        ref opIndex(size_t i) { return r[i]; }
+        scope ref opIndex(size_t i) { return r[i]; }
         auto length() @property { return r.length; }
         alias opDollar = length;
         auto save() { return this; }
     }
     static assert(isRandomAccessRange!RefAccessRange);
     static assert(isRandomAccessRange!RefAccessRange);
-    int[4] a = [4, 3, 2, 1];
-    int[2] b = [6, 5];
+    int[] a = [4, 3, 2, 1];
+    int[] b = [6, 5];
     auto c = chooseAmong(0, RefAccessRange(a[]), RefAccessRange(b[]));
 
     void refFunc(ref int a, int target) { assert(a == target); }
@@ -5052,21 +5052,21 @@ nothrow pure @system unittest
     assert(z2.front == tuple(0,1));
 }
 
-@nogc nothrow pure @safe unittest
+/*@nogc*/ nothrow pure @safe unittest
 {
     // Test zip's `back` and `length` with non-equal ranges.
     static struct NonSliceableRandomAccess
     {
         private int[] a;
-        @property ref front()
+        @property scope ref front()
         {
             return a.front;
         }
-        @property ref back()
+        @property scope ref back()
         {
             return a.back;
         }
-        ref opIndex(size_t i)
+        scope ref opIndex(size_t i)
         {
             return a[i];
         }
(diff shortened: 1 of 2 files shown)


std/range/package.d
```diff
@@ -1681,25 +1681,25 @@ if (Ranges.length >= 2
...
         alias opDollar = length;
         auto save() { return this; }
     }
     static assert(isRandomAccessRange!RefAccessRange);
     static assert(isRandomAccessRange!RefAccessRange);
-    int[4] a = [4, 3, 2, 1];
-    int[2] b = [6, 5];
+    int[] a = [4, 3, 2, 1];
+    int[] b = [6, 5];
     auto c = chooseAmong(0, RefAccessRange(a[]), RefAccessRange(b[]));
```
wilzbach: 1) `[]` is done here to create a slice of the static array (it's not needed for dynamic arrays, they are already slices)
2) It feels like things like this here should be possible with DIP1000 :/

However, I looked at this problem myself two days ago and I couldn't find a better solution either.
So I think `version(DIP1000) {} else` at the top of the unittest might be better, because then we know that this still needs a look at, but aren't blocked by it.
What do you or others think?

carblue: 1. I tried both, a and a[] etc. and they both work; don't know why I added superfluous [] again, probably to be minimally invasive, expecting a change back to static arrays

carblue: Yes, in order not to lose the @nogc and static array testing, "version(DIP1000) {} else" will be better, but I'll wait for a go

carblue: Maybe https://github.com/D-Programming-Deimos/openssl needs an update to openssl 1.0.2g? That's what I have on my Kubuntu system since the day before yesterday

wilzbach: How is this related? (and most likely - I think the Deimos packages aren't super-actively maintained)

carblue: It relates to my reasoning why continuous-integration/jenkins/pr-merge might have failed:
...
Fetching openssl 1.1.6+1.0.1g (getting selected version)...
Package diet-ng can be upgraded from 1.4.3 to 1.4.5.
Package taggedalgebraic can be upgraded from 0.10.8 to 0.10.11.
Package stdx-allocator can be upgraded from 2.77.0 to 2.77.1.
Package vibe-d can be upgraded from 0.8.3-alpha.3 to 0.8.3.
Package memutils can be upgraded from 0.4.9 to 0.4.10.
Package vibe-core can be upgraded from 1.4.0-alpha.1 to 1.4.0.
Package eventcore can be upgraded from 0.8.27 to 0.8.34.
Use "dub upgrade" to perform those changes.
Non-optional dependency openssl of botan not found in dependency tree!?.
script returned exit code 2

wilzbach: Oh that's an old Dub bug, see e.g.

https://github.com/dlang/dub/issues/1180
https://github.com/dlang/dub/issues/1112
https://github.com/dlang/dub/issues/986

Unfortunately Jenkins with all the projects it tests is quite "vulnerable" to random failures.
OTOH it has caught tons of regressions, so nothing is for free :/

Anyhow I restarted the job.
BTW feel free to ping me on Slack or the people in #phobos when you run into this again.

thewilsonator:
#6765 

I'm closing this one cos' its logged ~840 hours on the autotester in a single shot.

PR #6398, opened 2018-03-31
Fix Issue 13880 - nothrow @nogc std.algorithm.reduce on fixed-size arrays

wilzbach:
The same could probably be added to quite a few ranges that use seeds though some already use `assert`.
Maybe there could be a case made to just use `assert` here too?

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -2793,7 +2793,13 @@ if (fun.length >= 1)
 
         static if (isInputRange!R)
         {
-            enforce(!r.empty, "Cannot reduce an empty input range w/o an explicit seed value.");
+            // no need to throw if range is statically known to be non-empty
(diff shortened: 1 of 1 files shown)


std/algorithm/iteration.d
```diff
@@ -3165,6 +3178,15 @@ The number of seeds must be correspondingly increased.
     assert(data.length == 0);
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=13880
+// reduce shouldn't throw if the length is statically known
+pure nothrow @safe @nogc unittest
+{
+    import std.algorithm.comparison : min;
+    int[5] arr;
+    cast(void) reduce!min(arr);
```
JackStouffer: Add test for an `int[0]` with a seed.

wilzbach: Added.

PR #6399, opened 2018-03-31
Add example to chain explaining char corner cases

JackStouffer:
See https://issues.dlang.org/show_bug.cgi?id=17141


wilzbach:
> Non-optional dependency diet-ng of vibe-d:http not found in dependency tree!?.

Urgh that one seems to be the newest DUB failure on the Project Tester.
I created an issue to track this and search for fixes/solutions: https://github.com/dlang/ci/issues/190

PR #5322, opened 2017-04-04
Allocators should take a ubyte[] memory block instead of a void[] one

edi33416:
Allocators that rely on taking a block of unused memory to manage and allocate
from it should take a ubyte[] instead of a void[] in order to prevent users
from accidentally passing valid data and cause memory corruption/leaks.

PR #6400, opened 2018-03-31
Qualify Region, SharedRegion and SbrkRegion allocators

nordlow:
This qualifies `Region` and `SharedRegion` members `this`, `allocate` and `alignedAllocate` to be `pure`. It also qualifies its `allocate` and `alignedAllocate` to be @trusted.

This is in line with qualifiers of [pureMalloc](https://dlang.org/library/core/memory/pure_malloc.html).

It further qualifies `SbrkRegion` `allocate` and `alignedAllocate` to be `nothrow @nogc`.

AFAICT, `this(ubyte[] store)` shouldn't be `@trusted`.

I've also unified the formatting of the qualifiers to come after the function head.


nordlow:
@andralex Are there other allocators that needs to be qualified?

wilzbach:
@nordlow how about adding a test to prevent regressions?

nordlow:
I added `SharedRegion` aswell.

std/experimental/allocator/building_blocks/region.d
```diff
@@ -81,7 +79,7 @@ struct Region(ParentAllocator = NullAllocator,
         $(D parent.allocate(n)) returns $(D null), the region will be initialized
         as empty (correctly initialized but unable to allocate).
         */
-    this(ubyte[] store)
+    this(ubyte[] store) pure nothrow @nogc
```
edi33416: You are right that `this(ubyte[] store)` should not be trusted.

The problem with it is that if you use the GC to provide the store,
you can not allocate objects that store pointers to other GC allocated objects
because the GC will not perform a scan on it. Please see #5322 for more details.

I've read your [forum post](https://forum.dlang.org/thread/jfgpngdudtprzznrckwf@forum.dlang.org?page=1)
and I am afraid that you might get bit by this issue, so please use with care until we get the fix in.
I've made a few attempts to fix this, see #5403 and #5882, but those have hit a stall.

@andralex and I have talked about this issue on a regular basis and we are thinking of a solution.

std/experimental/allocator/building_blocks/region.d
```diff
@@ -56,14 +56,12 @@ struct Region(ParentAllocator = NullAllocator,
 
     private void* _current, _begin, _end;
 
-    pure nothrow @trusted @nogc
-    private void* roundedBegin() const
+    private void* roundedBegin() const pure nothrow @trusted @nogc
```
edi33416: I'm not sure if this makes it easier to read the function attributes or not.
As @andralex said, we need to come up with a consistent way of writing those;
there are currently a few styles in phobos.

std/experimental/allocator/building_blocks/region.d
```diff
@@ -749,18 +741,18 @@ version(Posix) struct SbrkRegion(uint minAlign = platformAlignment)
...
     */
-    size_t goodAllocSize(size_t n) shared
+    size_t goodAllocSize(size_t n) shared const pure nothrow @safe @nogc
     {
         return n.roundUpToMultipleOf(alignment);
     }
 
     /// Ditto
-    void[] allocate(size_t bytes) shared
+    void[] allocate(size_t bytes) shared nothrow @nogc
```
edi33416: This should also be marked as `@trusted` as it falls in line with how [Mallocator](https://github.com/dlang/phobos/blob/master/std/experimental/allocator/mallocator.d#L32) works.

std/experimental/allocator/building_blocks/region.d
```diff
@@ -784,7 +776,7 @@ version(Posix) struct SbrkRegion(uint minAlign = platformAlignment)
     }
 
     /// Ditto
-    void[] alignedAllocate(size_t bytes, uint a) shared
+    void[] alignedAllocate(size_t bytes, uint a) shared nothrow @nogc
```
edi33416: ditto

std/experimental/allocator/building_blocks/region.d
```diff
@@ -830,7 +822,7 @@ version(Posix) struct SbrkRegion(uint minAlign = platformAlignment)
             || assert(0);
```
edi33416: I'm guessing you have missed `expand`s function attributes reordering

std/experimental/allocator/building_blocks/region.d
```diff
@@ -875,7 +866,7 @@ version(Posix) struct SbrkRegion(uint minAlign = platformAlignment)
     bool deallocate(void[] b) shared
```
edi33416: Missed reordering?

edi33416:
> Are there other allocators that need to be qualified?

Most of the allocators in the module still need more qualifiers.
Ideally, the allocators should stick to the following list of attributes:

- allocate, alignedAllocate, allocateAll: `pure, @safe, nothrow`; should infer `@nogc` or not, depending if it uses the `GCAllocator`
- expand: `pure, @safe, nothrow`; should infer `@nogc` or not, depending if it uses the `GCAllocator`
- goodAllocSize, empty, owns, resolveInternalPointer:  `pure, @safe, nothrow, @nogc`
- alignment: `pure, @safe, nothrow, @nogc`; this is an enum in all the current allocators
- reallocate, alignedReallocate: `nothrow`; may be `pure, @nogc`
- deallocate, deallocateAll: `pure, nothrow, @nogc`

@andralex could you please double check the list above?

std/experimental/allocator/building_blocks/region.d
nordlow: I'm leaving it as is, then.

std/experimental/allocator/building_blocks/region.d
nordlow: So which of the attributes `pure nothrow @nogc` should I remove then?

std/experimental/allocator/building_blocks/region.d
nordlow: Done.

std/experimental/allocator/building_blocks/region.d
nordlow: Done.

std/experimental/allocator/building_blocks/region.d
nordlow: Done.

std/experimental/allocator/building_blocks/region.d
nordlow: Done.

nordlow:
One more round @edi33416?

std/experimental/allocator/building_blocks/region.d
wilzbach: We already had this discussion before and thus even put it in the [DStyle](https://dlang.org/dstyle.html#phobos_attributes)

> Attributes should be listed in alphabetical ordering, e.g. const @nogc nothrow pure @safe (the ordering should ignore the leading @).

Though without automation no one really felt the need to change things and adapt his/her own style to it.

n8sh:
Since this has been approved and all of @edi33416's comments have been resolved other than the ordering of function attributes and https://github.com/dlang/phobos/pull/6400#discussion_r178534291 which does not seem to call for any change to this PR in particular  I am tagging this for auto-merging.

edi33416:
Yep, all good. Nice!

nordlow:
Thanks!

PR #6402, opened 2018-04-01
Make some traits iterative using static foreach

nordlow:
This reduces the number of template instantiations needed and is measured to be at least as fast as the recursive variants [here](https://github.com/nordlow/phobos-next/blob/c34a0442f785bfe9941c02ae6ff9cb6fb7886537/benchmarks/iterative-traits/source/app.d#L41).

The refactored traits are: `genericIndexOf`, `allSatisfy`, `anySatisfy`, `isExpression` and `allSameType`.

There a probably more traits that could benefit from this refactoring. Proposals are welcome.

if we added a `static break;` to exit a `static foreach` prematurely, would that speed up the compilation even further, @WalterBright?


wilzbach:
> if we added a static break; to exit a static foreach prematurely, would that speed up the compilation even further

Yeah that would be really cool.
FWIW this was even proposed in DIP1010:

https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1010.md#static-break-and-static-continue

It also comes up in the NG from time to time:

https://forum.dlang.org/post/mailman.2099.1515541202.9493.digitalmars-d-learn@puremagic.com

std/meta.d
```diff
@@ -301,29 +301,19 @@ template staticIndexOf(alias T, TList...)
 
 // [internal]
 private template genericIndexOf(args...)
-if (args.length >= 1)
+    if (args.length >= 1)
```
wilzbach:
```
std/meta.d(304:9)[warn]: If constraints should have the same indentation as the function
```

std/meta.d
```diff
@@ -301,29 +301,19 @@ template staticIndexOf(alias T, TList...)
...
     {
-        alias head = OldAlias!(tuple[0]);
-        alias tail = tuple[1 .. $];
-
-        static if (isSame!(e, head))
-        {
-            enum index = 0;
-        }
-        else
+        static if (is(typeof(genericIndexOf) == void) && // not yet defined
```
wilzbach: Nice trick :+1:

std/meta.d
```diff
@@ -301,29 +301,19 @@ template staticIndexOf(alias T, TList...)
...
 // [internal]
 private template genericIndexOf(args...)
-if (args.length >= 1)
+    if (args.length >= 1)
 {
-    alias e     = OldAlias!(args[0]);
-    alias tuple = args[1 .. $];
-
-    static if (tuple.length)
+    static foreach (idx, argi; args[1 .. $])
```
wilzbach: Nit: Why not "arg" instead of "args" (it's a typical convention to denote the array with a "s" postfix.

std/traits.d
```diff
@@ -6857,17 +6857,20 @@ template isInstanceOf(alias S, alias T)
...
-            isExpressions!(T[$/2 .. $]);
-    else static if (T.length == 1)
-        enum bool isExpressions =
-            !is(T[0]) && __traits(compiles, { auto ex = T[0]; });
-    else
-        enum bool isExpressions = true; // default
+    static foreach (Ti; T)
+    {
+        static if (!is(typeof(isExpressions) == bool) && // not yet defined
+                   !is(T[0]) && __traits(compiles, { auto ex = T[0]; }))
```
wilzbach: Shouldn't this be `Ti` here?

std/meta.d
andralex: wow

std/meta.d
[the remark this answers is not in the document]
nordlow: Done.

nordlow: Done.

std/meta.d
nordlow: Done.

std/meta.d
nordlow: Done.

std/traits.d
nordlow: Correct. Fixed. Thanks!

nordlow:
Should be ready now. Can you find any other traits that can be made iterative?

nordlow:
Ready now!

std/traits.d
```diff
@@ -8402,13 +8405,18 @@ private template getSymbolsByUDAImpl(alias symbol, alias attribute, names...)
...
+                   !is(T[idx] == T[idx + 1]))
+        {
+            enum bool allSameType = false;
+        }
     }
-    else
+    static if (!is(typeof(allSameType) == bool))
     {
-        enum bool allSameType = is(T[0] == T[1]) && allSameType!(T[1..$]);
+        enum bool allSameType = true;
```
wilzbach: Hmm looks like DScanner's check is bogus here:

```
std/traits.d(8419:19)[warn]: Variable "allSameType" has the same name as a variable defined on line 8414.
```

I don't know why the other very similar code is accepted though.
However, you can ignore this module for this check by adding the following to the .dscanner.ini

```
label_var_same_name_check="-std.traits"
```

(Ideally preceded by a reference to a DScanner issue)

nordlow: Done.

nordlow:
Thanks, @wilzbach.

nordlow:
I don't know why Win32 fails...

andralex:
@nordlow try wine

wilzbach:
There's an ICE on Linux too (when run with -D):

```
core.exception.AssertError@dmd/cond.d(378): Assertion failure
----------------
??:? _d_assertp [0x74903d]
dmd/cond.d:378 pure nothrow @nogc @safe void dmd.cond.StaticForeach.prepare(dmd.dscope.Scope*).__require() [0x51392f]
dmd/cond.d:377 void dmd.cond.StaticForeach.prepare(dmd.dscope.Scope*) [0x5137dc]
dmd/attrib.d:1083 _ZN24StaticForeachDeclaration7includeEP5Scope [0x50c2df]
dmd/doc.d:1042 _ZN11emitComment11EmitComment5visitEP17AttribDeclaration [0x55811b]
dmd/parsetimevisitor.d:75 _ZN16ParseTimeVisitorI10ASTCodegenE5visitEP24StaticForeachDeclaration [0x6437f2]
dmd/attrib.d:1143 _ZN24StaticForeachDeclaration6acceptEP7Visitor [0x50c48d]
dmd/doc.d:1091 void dmd.doc.emitComment(dmd.dsymbol.Dsymbol, dmd.root.outbuffer.OutBuffer*, dmd.dscope.Scope*) [0x557798]
dmd/doc.d:796 void dmd.doc.emitMemberComments(dmd.dsymbol.ScopeDsymbol, dmd.root.outbuffer.OutBuffer*, dmd.dscope.Scope*) [0x55767c]
dmd/doc.d:916 _ZN11emitComment11EmitComment4emitEP5ScopeP7DsymbolPKc [0x557c57]
dmd/doc.d:960 _ZN11emitComment11EmitComment5visitEP11Declaration [0x557ded]
dmd/parsetimevisitor.d:47 _ZN16ParseTimeVisitorI10ASTCodegenE5visitEP16AliasDeclaration [0x643492]
dmd/declaration.d:929 _ZN16AliasDeclaration6acceptEP7Visitor [0x535d61]
dmd/doc.d:1091 void dmd.doc.emitComment(dmd.dsymbol.Dsymbol, dmd.root.outbuffer.OutBuffer*, dmd.dscope.Scope*) [0x557798]
dmd/doc.d:796 void dmd.doc.emitMemberComments(dmd.dsymbol.ScopeDsymbol, dmd.root.outbuffer.OutBuffer*, dmd.dscope.Scope*) [0x55767c]
dmd/doc.d:439 _Z10gendocfileP6Module [0x556783]
dmd/mars.d:911 int dmd.mars.tryMain(ulong, const(char)**) [0x5f8097]
dmd/mars.d:1098 _Dmain [0x5f8d92]
```

reported -> https://issues.dlang.org/show_bug.cgi?id=18721

wilzbach:
@nordlow you could also try to split this PR into smaller ones, e.g. the documentation ICE comes from `allSameType`

wilzbach:
Oh and I think the persistent Windows access violation means that there are now too many symbols in the OMF object file (the maximum is 32768).
So I assume another "split-up" of `phobos.lib` is required :/
See also: https://issues.dlang.org/show_bug.cgi?id=4904

> Copyright (C) Digital Mars 2000-2007

We really need to switch to a more modern, actively maintained linker ...

nordlow:
@wilzbach, can I run ddoc generation on phobos from the makefile?

wilzbach:
> @wilzbach, can I run ddoc generation on phobos from the makefile?

Well you need to clone dlang.org first, e.g.

```
cd ~/dlang
git clone https://github.com/dlang/dlang.org
make -f posix.mak phobos-prerelease -j4
```

and then check `web/phobos-prerelease`

Though for the failure you don't need any macros, just `dmd -D <file>` should be enough.

> reported -> https://issues.dlang.org/show_bug.cgi?id=18721

BTW my noobish attempt: https://github.com/dlang/dmd/pull/8128 (though the ICE I reduced looks a bit different)

andralex:
status?

nordlow:
Made a merge against master to see if something has changed (for the better).

nordlow:
I did a merge against master to see if the linking error still occurs on Windows. What's the best way of fixing it?

nordlow:
Build passes. Ok to merge @andralex?

wilzbach:
Please rebase instead of merging for a slicker git history.

wilzbach:
Regarding the linker error: looks like you are lucky and symbols have been moved around in the last six months. As long as auto-tester is green, everything should be okay.

.dscanner.ini
```diff
@@ -272,7 +272,7 @@ has_public_example="-etc.c.curl,\
...
 ; Checks for labels with the same name as variables
-label_var_same_name_check="-std.algorithm.iteration,-std.algorithm.sorting,-std.array,-std.bigint,-std.bitmanip,-std.conv,-std.encoding,-std.experimental.allocator.building_blocks.segregator,-std.experimental.typecons,-std.format,-std.internal.digest.sha_SSSE3,-std.parallelism,-std.process,-std.typecons,-std.utf"
+label_var_same_name_check="-std.algorithm.iteration,-std.algorithm.sorting,-std.array,-std.bigint,-std.bitmanip,-std.conv,-std.encoding,-std.experimental.allocator.building_blocks.segregator,-std.experimental.typecons,-std.format,-std.internal.digest.sha_SSSE3,-std.parallelism,-std.process,-std.typecons,-std.utf,-std.traits"
```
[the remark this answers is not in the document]
wilzbach: I can't recall that this would have been fixed on the dscanner side. Though opening an issue there might help.

nordlow:
Thanks.

Hackerpilot:
This introduced an undocumented change: https://issues.dlang.org/show_bug.cgi?id=19560

I'm not really sure that I want to call it a regression because the new behavior makes more sense than the old, and the old behavior wasn't really specified.

andralex:
Good catch. I think what was not specified and becomes specified should be documented carefully but not count as a regression.

PR #6403, opened 2018-04-01
Add std.traits.isStronglyPure

SirSireesh:
See relevant NG discussion: https://forum.dlang.org/thread/p9m6lb$iud$1@digitalmars.com

D currently allows functions to be pure even though they are not strictly pure.
To be strictly pure, a function must satisfy the following:
* Each parameter:
  - is `immutable`, OR
  - can be converted automatically to `immutable` (i.e. has no mutable indirections) AND is passed by value
* The return type:
  - is `immutable`, OR
  - can be converted automatically to `immutable`

Andrei has said that `const` will not count for now

I'm also unsure on how to unittest the second overload for isStronglyPure,  since overloading of functions within the local scope is not allowed. Please help with this. Thanks!

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -9107,27 +9107,31 @@ if (F.length == 1 && isCallable!F)
 
         //check if the function is a member function
         static if (isType!(__traits(parent, F)) && !hasFunctionAttributes!(F, "immutable"))
-            return false;
-
-        alias STC = ParameterStorageClass;
-
-        static foreach (i, param; Parameters!F)
         {
-            if (!is(param == immutable)
-                    && (isRefType!param || hasAliasing!param
-                    || ParameterStorageClassTuple!F[i] == STC.ref_
-                    || ParameterStorageClassTuple!F[i] == STC.out_))
-                return false;
-        }
-
-        alias rt = ReturnType!F;
-
-        if (is(rt == immutable))
-            return true;
-        else if (isRefType!rt || hasAliasing!rt || hasFunctionAttributes!(F, "ref"))
             return false;
+        }
         else
-            return true;
+        {
+            alias STC = ParameterStorageClass;
+
+            foreach (i, param; Parameters!F)
+            {
+                 if (!is(param == immutable)
+                        && (isRefType!param || hasAliasing!param
+                        || ParameterStorageClassTuple!F[i] == STC.ref_
+                        || ParameterStorageClassTuple!F[i] == STC.out_))
+                    return false;
+            }
+
+            alias rt = ReturnType!F;
+
+            if (is(rt == immutable))
+                return true;
+            else if (isRefType!rt || hasAliasing!rt || hasFunctionAttributes!(F, "ref"))
+                return false;
+            else
+                return true;
+        }
     }
 
     private bool isRefType(T)()
--- a/std/traits.d
+++ b/std/traits.d
@@ -22,6 +22,7 @@
  *           $(LREF FunctionTypeOf)
  *           $(LREF isSafe)
  *           $(LREF isUnsafe)
+ *           $(LREF isStronglyPure)
  *           $(LREF isFinal)
  *           $(LREF ParameterDefaults)
  *           $(LREF ParameterIdentifierTuple)
@@ -9079,4 +9080,114 @@ enum isCopyable(S) = __traits(isCopyable, S);
     static assert(isCopyable!C1);
     static assert(isCopyable!int);
     static assert(isCopyable!(int[]));
+}
+
+/++
+Determines if `F` is a strongly `pure` function. This is true when :
+    $(UL
+        $(LI Each parameter is `immutable`, or can be converted automatically to `immutable`
+        (i.e. has no mutable indirections) and is passed by value)
+        $(LI The return type is `immutable` or can be implicitly converted to `immutable`)
+    )
+Params:
+    F = The function to be tested
+
+Returns:
+    `true` if the function is strongly `pure`, `false` if otherwise.
+Note:
+    Member functions must be marked `immutable`
+++/
+template isStronglyPure(F...)
+if (F.length == 1 && isCallable!F)
+{
+    bool isStronglyPure()
+    {
+        if (!hasFunctionAttributes!(F, "pure"))
+            return false;
+
+        //check if the function is a member function
+        static if (isType!(__traits(parent, F)) && !hasFunctionAttributes!(F, "immutable"))
+            return false;
+
+        alias STC = ParameterStorageClass;
+
+        static foreach (i, param; Parameters!F)
+        {
+            if (!is(param == immutable)
+                    && (isRefType!param || hasAliasing!param
+                    || ParameterStorageClassTuple!F[i] == STC.ref_
+                    || ParameterStorageClassTuple!F[i] == STC.out_))
+                return false;
+        }
+
+        alias rt = ReturnType!F;
+
+        if (is(rt == immutable))
+            return true;
+        else if (isRefType!rt || hasAliasing!rt || hasFunctionAttributes!(F, "ref"))
+            return false;
+        else
+            return true;
+    }
+
+    private bool isRefType(T)()
+    {
+        if (is(T == class) || is(T == interface)
+                || isDelegate!T || isAssociativeArray!T
+                || isPointer!T)
+            return true;
+        else if (isDynamicArray!T && !is(T : immutable(U)[], U))
+            return true;
+        else
+            return false;
+    }
+}
+
+///
+@safe pure unittest
+{
+    int f(int) pure;
+    void g(ref int) pure;
+    void j(immutable (void function(ref int) pure)) pure;
+    static assert( isStronglyPure!f);
+    static assert(!isStronglyPure!g); //takes args by mutable reference
+    static assert( isStronglyPure!j); //can call a function which is not strongly pure
+}
+
+@safe pure unittest
+{
+    void h(out int) pure;
+
+    struct Test
+    {
+        int[string] aa;
+    }
+
+    void i(Test) pure;
+
+    Test k(int) pure;
+
+    string l(int[]) pure;
+
+    ref int m(int) pure;
+
+    string n(int) pure;
+
+    struct S
+    {
+        int x;
+        int f() pure
+        {
+            return x++;
+        }
+    }
+
+
+    static assert(!isStronglyPure!h);
+    static assert(!isStronglyPure!i); //has parameter Test with mutable reference member aa
+    static assert(!isStronglyPure!k);
+    static assert(!isStronglyPure!l);
+    static assert(!isStronglyPure!m);
+    static assert( isStronglyPure!n);
+    static assert(!isStronglyPure!(S.f)); //S.f should be immutable
 }
--- a/changelog/std-traits-isStronglyPure.dd
+++ b/changelog/std-traits-isStronglyPure.dd
@@ -1,0 +1,15 @@
+Addition: `std.traits.isStronglyPure` 
+
+Strongly `pure` functions are `pure` functions of which parameters and return value cannot
+reach any mutable state. Practically this means the layout of all parameters and that of
+the result contains nonpointer data passed (or returned) by value, in conjunction
+with `immutable` indirectly-accessed data. This is in keeping with the
+$(LINK2 $(ROOT_DIR)spec/function.html#pure-functions, D language spacification of pure functions).
+
+$(REF isStronglyPure, std, traits) yields `true` if a the function passed to it satisfies
+these conditions. Example:
+-------
+static assert(isStronglyPure!((int x) => x));
+static assert(!isStronglyPure((ref int x) => x));
+-------
+


std/traits.d
```diff
@@ -8590,3 +8591,119 @@ enum isCopyable(S) = is(typeof(
...
+ + `true` if the function is strongly `pure`, `false` if otherwise.
+ + ++/
+template isStronglyPure(F...)
+if (F.length == 1 && isCallable!F)
+{
+    bool isStronglyPure()
+    {
+        alias STC = ParameterStorageClass;
+
+        if (!(functionAttributes!F & FunctionAttribute.pure_))
```
MetaLang: This can be more succinctly written as `!hasAttributes!(F, "pure")`

std/traits.d
```diff
@@ -8590,3 +8591,119 @@ enum isCopyable(S) = is(typeof(
...
+            return false;
+
+        alias params = Parameters!F;
+
+        static foreach (i; 0 .. arity!F)
```
MetaLang: You could instead write `static foreach (i, param; Parameters!F)`

std/traits.d
```diff
@@ -8590,3 +8591,119 @@ enum isCopyable(S) = is(typeof(
     static assert(isCopyable!int);
     static assert(isCopyable!(int[]));
 }
+
+/++
+ + Determines if `F` is a strongly `pure` function. To be strongly `pure`, a function must satisfy the following:
```
BBasile: To prevent redundancies, i'd go for something like this instead:
`Determines if a function is string `pure`", which is the case when...`

std/traits.d
```diff
@@ -8590,3 +8591,119 @@ enum isCopyable(S) = is(typeof(
...
+ + $(UL
+ +   $(LI Each parameter is `immutable`, OR can be converted automatically to `immutable`
```
BBasile: Lower case OR and AND, maybe. Readers are assumed to know a bit boolean logic, no ?

std/traits.d
```diff
@@ -8590,3 +8591,119 @@ enum isCopyable(S) = is(typeof(
...
+    {}
+
+    static assert( isStronglyPure!f);
+    static assert( isStronglyPure!g);
+    static assert(!isStronglyPure!h);
+    static assert(!isStronglyPure!i); //has parameter Test with mutable reference member aa
+}
+
+/++
+ + Determines if a function is a strongly `pure` function. To be strongly `pure`, a function must satisfy the following:
```
BBasile: Same comments regarding the DDOC as for the first overload.

std/traits.d
```diff
@@ -8590,3 +8591,119 @@ enum isCopyable(S) = is(typeof(
...
+ {}
+ void f(out int x) pure
+ {}
+ static assert ( isStronglyPure!"f");
+ static assert (!isStronglyPure!("f", int));
+ -----------------
+ + Returns:
+ + `true` if the function is strongly `pure`, `false` if otherwise.
+ + ++/
+bool isStronglyPure(string funcName, T...)()
```
BBasile: Is is really useful ? I don't think. People are supposed to use `getOverloads` by themselves. Many traits for functions in this module don't have this helper.

std/traits.d
```diff
@@ -8590,3 +8591,119 @@ enum isCopyable(S) = is(typeof(
...
+ + )
+ + This version is useful if a function has multiple overloads. The correct overload can be selected by specifying
+ + the argument types as the template arguments
+ +
+ + Params:
+ + funcName = The name of the function to be tested
+ + T = The types of the arguments accepted by the overload that should be tested
+ +
+ + Example:
+ -----------------
```
BBasile: Use commented unittest instead of the old-fashioned `Example` section please.

wilzbach:
I'm not entirely sure what the use case for this trait will be, but for those who don't know and as it's a bit related:
Vibe.d has the concept of `isStronglyIsolated` and `isWeaklyIsolated` (combined with `makeIsolated` and `assumeIsolated`):

https://github.com/vibe-d/vibe-core/blob/master/source/vibe/core/concurrency.d#L960
https://github.com/vibe-d/vibe-core/blob/master/source/vibe/core/concurrency.d#L310

std/traits.d
wilzbach: Also please combine the overloads with `/// ditto`, s.t. they are visually grouped on the docs.

BBasile: The parameters differ, i think this is why OP has copied the whole stuff.

std/traits.d
scarface-one: Done

std/traits.d
scarface-one: Done

std/traits.d
scarface-one: Considering the fact that it does not work for local functions, I think its best to get rid of this overload.

scarface-one:
I've removed the second overload for the following reasons:
- It does not work with local functions 
- Also makes it harder to unittest.
- None of the other functions in `std.traits` do this type of overloading as far as I can tell

std/traits.d
```diff
@@ -8590,3 +8591,83 @@ enum isCopyable(S) = is(typeof(
...
+
+        static foreach (i, param; Parameters!F)
+        {
+            static if (is(param == immutable))
+            {}
+            else static if (isRefType!param || hasAliasing!param
+                    || ParameterStorageClassTuple!F[i] == STC.ref_
+                    || ParameterStorageClassTuple!F[i] == STC.out_)
+                return false;
+        }
```
wilzbach: Looks like we're missing a nice `hasParametersAttribute!(param, "ref", "out")` (a-la `hasFunctionAttributes`) API :/

std/traits.d
```diff
@@ -8590,3 +8591,83 @@ enum isCopyable(S) = is(typeof(
...
+    {}
+
+    void j(immutable (void function(ref int) pure)) pure
+    {}
+
+    static assert( isStronglyPure!f);
+    static assert( isStronglyPure!g);
+    static assert(!isStronglyPure!h);
+    static assert(!isStronglyPure!i); //has parameter Test with mutable reference member aa
+    static assert( isStronglyPure!j); //can call a function which is not strongly pure
```
wilzbach: FYI: this is the publicly displayed example and while it does show off must of the functionality we have the informal consensus that public example snippets should ideally be 10 lines or less, because that's what a lazy reader can grasp within "one mind read".

-> how about splitting it up?

(you can also of course add more tests that aren't public or "hide" some of these)

scarface-one: Done

wilzbach:
> Merge branch 'std.traits.isStronglyPure' of github.com:

Please don't use merge commits on your branch - instead `rebase` is recommended:

https://wiki.dlang.org/Starting_as_a_Contributor#Rebasing
https://wiki.dlang.org/Starting_as_a_Contributor#Squashing
https://wiki.dlang.org/Starting_as_a_Contributor#A_file_that_I_made_a_change_on_was_modified_by_a_different_merged_PR.2C_and_now_my_PR_can.27t_be_merged.2C_now_what.3F

We also prefer all "fixup" commits to be squashed, but this can wait until the end (i.e. when this function has been approved) and we can also easily do this for you.

BTW also notice the bot's message:

> If your PR contains non-trivial changes, please reference a Bugzilla issue or create a manual changelog.

In this case a changelog entry is preferred as it's a new symbol.

std/traits.d
```diff
@@ -8590,3 +8591,85 @@ enum isCopyable(S) = is(typeof(
...
+ +   $(LI Each parameter is `immutable`, or can be converted automatically to `immutable`
+ +   (i.e. has no mutable indirections) and is passed by value)
+ +   $(LI The Return Type is `immutable` or can be impicitly converted to `immutable`)
+ + )
+ + Params:
+ + F = The function to be tested
+ +
+ + Returns:
+ + `true` if the function is strongly `pure`, `false` if otherwise.
+ + ++/
```
wilzbach: No leading stars or pluses please.
Have a look at https://dlang.org/dstyle.html#phobos_documentation (i.e. of sections indentation is also nice)

std/traits.d
```diff
@@ -8590,3 +8591,85 @@ enum isCopyable(S) = is(typeof(
     static assert(isCopyable!int);
     static assert(isCopyable!(int[]));
 }
+
+/++
+ + Determines if `F` is a strongly `pure` function. This is true when :
+ + $(UL
+ +   $(LI Each parameter is `immutable`, or can be converted automatically to `immutable`
+ +   (i.e. has no mutable indirections) and is passed by value)
+ +   $(LI The Return Type is `immutable` or can be impicitly converted to `immutable`)
```
wilzbach: - implicitly
- return type in lower case

std/traits.d
```diff
@@ -8590,3 +8591,85 @@ enum isCopyable(S) = is(typeof(
     static assert(isCopyable!int);
     static assert(isCopyable!(int[]));
 }
+
+/++
+ + Determines if `F` is a strongly `pure` function. This is true when :
```
wilzbach: How about:

> when the following requirements are fulfilled
> when the following rules are met

n8sh:
Is this the same as the definition of "strong purity" that the compiler uses in deciding whether certain optimizations are legal?

scarface-one:
I have no clue what happened! I'm going to close this PR and start over

wilzbach:
@scarface-one no need to close this PR. It's better to keep it, s.t. the discussion isn't split over multiple places and it's easier for people to follow. I will reopen this and fix (aka "rebase") it for you and squash all commits for you.

scarface-one:
Thanks, I panicked a bit after that. I'll add changelog and fix the docs now

CyberShadow:
> no need to close this PR

One reason in favor of recreating this is that now all the code owners that were accidentally assigned are now following this discussion, and would need to manually unsubscribe to avoid getting irrelevant notifications in their inbox.

scarface-one:
See https://forum.dlang.org/post/mailman.2001.1522655154.3374.digitalmars-d@puremagic.com
Essentially:
```d
int[] f(int x) pure
{
    return [x];
}
immutable var = f(1); // works
```
This was previously incorrectly reported to not be strongly pure. I've fixed this, but I think checking return type is unnecessary. I think that if all of a function's parameters are `immutable` or implicitly convertible to `immutable`,  then its return type can always be converted to `immutable`.

Can someone give a case where this may not be true?

scarface-one:
>Obviously, rewriting this to
>
>```d
>auto var1 = f(1);
>var1[0] = 0;
>auto var2 = var1;
>```
>would be wrong.

Makes sense. Then I think `isStronglyPure` should require that the function's return type is explicitly annotated with `immutable`

scarface-one:
I've changed it to require that the return type is strictly `immutable` or `void`. I'll squash the commits later

scarface-one:
If `f` was marked strongly pure, then
```d
auto var1 = f(1);
var1[0] = 0;
auto var2 = var1;
```
becomes correct. On the other hand if `f`'s return type was marked `immutable` then this would not be allowed. Which is why I think it should strictly be `immutable`

scarface-one:
Ping. Should I revert this to the old behaviour (allow implicit conversion to `immutable`) or is the new stricter behaviour preferable?
I think the strict behaviour should be preferred, since would be weird if `isStronglyPure!F` was true but the compiler was not treating `F` like a strongly pure function. e.g. are there in above discussions

std/traits.d
```diff
@@ -8590,3 +8591,92 @@ enum isCopyable(S) = is(typeof(
     static assert(isCopyable!int);
     static assert(isCopyable!(int[]));
 }
+
+/++
+Determines if `F` is a strongly `pure` function. This is true when the following conditions are fulfilled:
+$(UL
+    $(LI Each parameter is `immutable`, or can be converted implicitly to `immutable`
+    (i.e. has no mutable indirections) and is passed by value)
```
andralex: `ref` parameters are also okay if `immutable`

scarface-one: > Each parameter is `immutable`, or

std/traits.d
```diff
@@ -8590,3 +8591,92 @@ enum isCopyable(S) = is(typeof(
...
+            if (is(param == immutable))
+            {}
+            else if (isRefType!param
+                    || hasAliasing!param
+                    || ParameterStorageClassTuple!F[i] == STC.ref_
+                    || ParameterStorageClassTuple!F[i] == STC.out_)
+                return false;
+        }
+
+        if (!is(ReturnType!F == immutable) && !is(ReturnType!F == void))
```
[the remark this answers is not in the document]
scarface-one:
```d
ref int f(immutable ref int x) pure
{
    return *(new int(x));
}
```
Should this count as strongly pure?
`T` and `ref T` are currently allowed for return types as long as 
- none of the mutable parameters implicitly convert to the return type or
- the return type is implicitly convertible to `immutable`
The above counts as strongly pure under these conditions

std/traits.d
```diff
@@ -8593,3 +8593,101 @@ enum isCopyable(S) = is(typeof(
...
+        static foreach (i, param; Parameters!F)
+        {
+            if (is(param == immutable))
+            {}
+            else if (isRefType!param || hasAliasing!param
+                    || ParameterStorageClassTuple!F[i] == STC.ref_
+                    || ParameterStorageClassTuple!F[i] == STC.out_)
+                return false;
+            if (is(param : rt))
+                immutablert = false;
```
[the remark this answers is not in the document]
scarface-one: It checks if any of the parameters are implicitly convertible to the return type. This is necessary because : 
```d
int[] foo(int[] a) pure
{
    return [1, 2, 3, 4, 5];
}

void main()
{
    int[] a = [1, 2];
    immutable b = foo(a); //fails, since b may become a slice of a 
    immutable c = foo([1, 2]); //ok
}
```
I decided its best to play safe and not call such functions strongly pure. Should they count as strongly pure?

std/traits.d
```diff
@@ -8593,3 +8593,101 @@ enum isCopyable(S) = is(typeof(
...
+    }
+
+    int[] l(int[] x) pure
+    {
+        return [1];
+    }
+
+    static assert(!isStronglyPure!h);
+    static assert(!isStronglyPure!i); //has parameter Test with mutable reference member aa
+    static assert( isStronglyPure!k);
```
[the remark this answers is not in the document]
scarface-one: No, since `immutable a = k(1)` will work, regardless of what you pass to k.

std/traits.d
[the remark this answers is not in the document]
scarface-one: Thanks for clearing this up!

scarface-one:
Ok I think I've addressed all the comments except Andrei's. He said that `ref` return types should not be allowed, but I'm not sure how to check if the return type is `ref`. The following code passes:

```d
ref int f(int x)
{
    return *(new int(x));
}
immutable a = f(1);
immutable b = f(1);
static assert(a == b);
```

I think as long as the return type has no mutable indirections, this should be allowed as strongly pure. Am I correct?

scarface-one:
Ok I figured out how to check if return type is `ref`. It applies as a function's attributes, as in the docs. `ref` return types are now not allowed unless marked immutable. Just a summary of what counts as strongly pure:
- The parameters are all `immutable` or implicitly convertible to `immutable`
- The return type is `immutable` or implicitly convertible to `immutable`

Types may be implicitly converted to `immutable` if they are not `ref` of pseudo-ref types and, if they are aggregate types, the same applies recursively to all subtypes

scarface-one:
ping @andralex is this good to go now?

andralex:
@klickverbot all cool?

scarface-one:
ping @klickverbot

scarface-one:
ping

scarface-one:
I just realised that the following code passes:
```d
struct S
{
    int x;
    int f() pure
    {
        return x++;
    }
}

static assert (isStronglyPure!(S.f));
```
The correct thing to do would be to ensure that member functions are marked as `const`, but I'm not sure how one can find out if a function is a member of a `struct` or a `class` or a free function. Does anyone know how this can be done?

jmdavis:
> The correct thing to do would be to ensure that member functions are marked as const

Actually, it would have to be `immutable`, since strongly `pure` requires that the parameters be `immutable` or implicitly convertible to `immutable`.

scarface-one:
> Actually, it would have to be `immutable`, since strongly `pure` requires that the parameters be `immutable` or implicitly convertible to `immutable`.

Yes I meant `immutable`. I changed it so member functions are now required to be marked `immutable`.

scarface-one:
ping. This is ready for another review

scarface-one:
ping

std/traits.d
```diff
@@ -8593,3 +8594,115 @@ enum isCopyable(S) = is(typeof(
...
+++/
+template isStronglyPure(F...)
+if (F.length == 1 && isCallable!F)
+{
+    bool isStronglyPure()
+    {
+        if (!hasFunctionAttributes!(F, "pure"))
+            return false;
+
+        //check if the function is a memebr of a struct or a class
```
andralex: typo

scarface-one: fixed

scarface-one:
CI failures seem unrelated, although I couldn't fix the "unreachable code" warning

scarface-one:
ping @klickverbot

wilzbach:
Try rebasing to upstream/master. That might help with the CIs.

ZombineDev:
@klickverbot can you take another look?

scarface-one:
ping @klickverbot

scarface-one:
Ping on this. @klickverbot

thewilsonator:
Ping @dnadlinger

thewilsonator:
(I'll add this to the PRs that need decisions made list for the DConf hackathon if nothing is done about it in the meantime.)

scarface-one:
Was any progress made on this?

andralex:
status?

andralex:
cc @RazvanN7

RazvanN7:
@andralex I take it you are in favor of this addition?

dkorpel:
> Andrei has said that `const` will not count for now

As of https://github.com/dlang/dmd/pull/12890, dmd does count `const` for strongly `pure`.

RazvanN7:
This PR implements strong purity based on some exaggerated constraints. The compiler checks for strong purity are a bit different:

- the compiler does not care about the return type of a strongly pure function, whereas this PR requires the return type to be immutable or implicitly convertible to immutable.
- the compiler requires that the parameters of a strongly pure function do not have any mutable indirections, whereas this PR requires that the parameters are immutable or implicitly convertible to immutable.

Any implementation of `isStronglyPure` should match what the compiler is doing which leads to the question on whether we should expose this via the `__traits` interface or have a third party tool implementation via dmd as a library.

andralex:
> @andralex I take it you are in favor of this addition?

It's a good trait to have, and at best it would be 100% with what the compiler assumes.

atilaneves:
> Any implementation of `isStronglyPure` should match what the compiler is doing which leads to the question on whether we should expose this via the `__traits` interface or have a third party tool implementation via dmd as a library.

I agree. I also don't know what this is for.

RazvanN7:
Closing this on the basis that insufficient rationale was provided to pursue this further.

PR #6404, opened 2018-04-02
Fix Issue 18704 - std/datetime/stopwatch.d(452) assertion failure on auto-tester

wilzbach:
Looks like an optimization issue. With `asm`, DMD gets pretty conservative.

Though ideally we should have something like https://github.com/dlang/phobos/pull/5416 built-in into the `benchmark` method.

### DIFF
--- a/std/datetime/stopwatch.d
+++ b/std/datetime/stopwatch.d
@@ -445,7 +445,14 @@ Duration[fun.length] benchmark(fun...)(uint n)
 
     int a;
     void f0() nothrow {}
-    void f1() nothrow { auto b = to!string(a); }
+    void f1() nothrow @trusted {
+        // do not allow any optimizer to optimize this function away
+        import core.thread : getpid;
+        import core.stdc.stdio : printf;
+        auto b = getpid.to!string;
+        if (getpid == 1) // never happens, but prevents optimization
+            printf("%p", &b);
+    }
     auto r = benchmark!(f0, f1)(1000);
     assert(r[0] >= Duration.zero);
     assert(r[1] > Duration.zero);


jmdavis:
Did you try just slapping `pragma(inline, false)` on it? That would certainly make the intent clearer, and from what I've seen in benchmarking that I've done, it's prevented functions from being optimized out.

wilzbach:
Hmm so I experimented with a few variants and `pragma(inline` isn't a magic stop for the optimizer - and empty `asm` is neither though:

- `pragma(inline` - https://run.dlang.io/is/ukWPbD
- empty `asm` - https://run.dlang.io/is/vRtZBd

However once the `asm` block uses any instructions, the optimizer stops (e.g.  https://run.dlang.io/is/Jafl6E)

Now, just to be sure and avoid any failures for the LDC people with a more advanced optimizer, I used the `getpid` trick as this can never, ever be optimized away (see e.g. https://run.dlang.io/is/enbg5l)

wilzbach:
Just saw this again on the auto-tester - this time on Linux_64_64:

```
core.exception.AssertError@std/datetime/stopwatch.d(452): unittest failure
----------------
posix.mak:380: recipe for target 'unittest/std/datetime/stopwatch.run' failed
make[1]: *** [unittest/std/datetime/stopwatch.run] Error 1
make[1]: *** Waiting for unfinished jobs....
0.346s PASS debug64 std.datetime
8.206s PASS debug64 std.parallelism
make[1]: Leaving directory '/home/braddr/sandbox/at-client/pull-3105522-Linux_64_64/phobos'
posix.mak:298: recipe for target 'unittest-debug' failed
```

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3105522&isPull=true

PR #5879, opened 2017-11-21
Make allocate @safe

edi33416:
For allocators that implement their own `allocate` method, this should be a `pure nothrow @safe` function.
Allocators that are building on top of such allocators should infer the function attributes from their parents.

This PR is a subset of #5330, as this approach will provide us with better granularity. More smaller PRs to come :)

PR #6405, opened 2018-04-02
Make mmap_allocator allocate @safe

edi33416:

### DIFF
--- a/std/experimental/allocator/mmap_allocator.d
+++ b/std/experimental/allocator/mmap_allocator.d
@@ -28,15 +28,16 @@ struct MmapAllocator
     version(Posix)
     {
         /// Allocator API.
+        nothrow @nogc @safe
         void[] allocate(size_t bytes) shared
         {
             import core.sys.posix.sys.mman : mmap, MAP_ANON, PROT_READ,
                 PROT_WRITE, MAP_PRIVATE, MAP_FAILED;
             if (!bytes) return null;
-            auto p = mmap(null, bytes, PROT_READ | PROT_WRITE,
-                MAP_PRIVATE | MAP_ANON, -1, 0);
+            auto p = (() @trusted => mmap(null, bytes, PROT_READ | PROT_WRITE,
+                MAP_PRIVATE | MAP_ANON, -1, 0))();
             if (p is MAP_FAILED) return null;
-            return p[0 .. bytes];
+            return (() @trusted => p[0 .. bytes])();
         }
 
         /// Ditto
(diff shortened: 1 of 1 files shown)


jercaianu:
Looks fine to me

std/experimental/allocator/mmap_allocator.d
```diff
@@ -28,6 +28,7 @@ struct MmapAllocator
     version(Posix)
     {
         /// Allocator API.
+        nothrow @trusted @nogc
         void[] allocate(size_t bytes) shared
         {
             import core.sys.posix.sys.mman : mmap, MAP_ANON, PROT_READ,
```
wilzbach: On a second look, it might be better to use a `@trusted` lambda around mmap as `@trusted` on functions is a really anti-pattern that we soon might automatically check against.

std/experimental/allocator/mmap_allocator.d
```diff
@@ -72,10 +74,10 @@ struct MmapAllocator
...
 }
 
-@system unittest
+nothrow @safe @nogc unittest
 {
     alias alloc = MmapAllocator.instance;
     auto p = alloc.allocate(100);
     assert(p.length == 100);
-    () nothrow @nogc { alloc.deallocate(p); }();
+    () @trusted { alloc.deallocate(p); }();
```
aG0aep6G: Also need to nullify `p` in the `@trusted` section. Else you have a dangling pointer in `@safe` code.

std/experimental/allocator/mmap_allocator.d
edi33416: Can we make make `mmap` and `VirtualAlloc` pure?

edi33416:
@wilzbach  @aG0aep6G Can we make make mmap and VirtualAlloc pure?

wilzbach:
@edi33416 have a look at https://github.com/dlang/druntime/pull/1718

std/experimental/allocator/mmap_allocator.d
```diff
@@ -72,10 +74,10 @@ struct MmapAllocator
...
 }
 
-@system unittest
+nothrow @safe @nogc unittest
 {
     alias alloc = MmapAllocator.instance;
     auto p = alloc.allocate(100);
     assert(p.length == 100);
-    () nothrow @nogc { alloc.deallocate(p); }();
+    () @trusted { alloc.deallocate(p); p = null; }();
```
andralex: @aG0aep6G good point, operated the change myself

PR #6406, opened 2018-04-02
Make null_allocator allocate @safe

edi33416:


wilzbach:
> Maybe put this at top level for the whole struct:
> pure nothrow @ safe @ nogc

FWIW `dmd` has bugs in this transitive induction, e.g. https://issues.dlang.org/show_bug.cgi?id=18672

std/experimental/allocator/building_blocks/null_allocator.d
```diff
@@ -79,7 +80,7 @@ struct NullAllocator
 {
     assert(NullAllocator.instance.alignedAllocate(100, 0) is null);
     assert(NullAllocator.instance.allocateAll() is null);
-    auto b = NullAllocator.instance.allocate(100);
+    auto b = (() nothrow @safe @nogc => NullAllocator.instance.allocate(100))();
```
wilzbach: `pure`?

std/experimental/allocator/building_blocks/null_allocator.d
```diff
@@ -23,6 +23,7 @@ struct NullAllocator
     //size_t goodAllocSize(size_t n) shared const
     //{ return .goodAllocSize(this, n); }
     /// Always returns $(D null).
+    pure nothrow @safe @nogc
     void[] allocate(size_t) shared { return null; }
     /// Always returns $(D null).
     void[] alignedAllocate(size_t, uint) shared { return null; }
```
n8sh: Why not `alignedAllocate` too?

std/experimental/allocator/building_blocks/null_allocator.d
edi33416: This is not `pure` because it is using `instance` which is global data. I believe `pure` should accept  `stateless` global data, but it doesn't (for the time being at least)

edi33416:
I moved all the attributes at top level and cleaned up the unittest

edi33416:
@wilzbach is the Win32 failing test related to the transitive induction bug?

PR #6407, opened 2018-04-02
Make bitmapped_block allocate @safe

edi33416:

### DIFF
--- a/std/experimental/allocator/building_blocks/bitmapped_block.d
+++ b/std/experimental/allocator/building_blocks/bitmapped_block.d
@@ -303,7 +303,8 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
     Allocations greater than 64 blocks require a multiword search through the
     metadata.
     */
-    @trusted void[] allocate(const size_t s)
+    pure nothrow @trusted @nogc
+    void[] allocate(const size_t s)
     {
         const blocks = s.divideRoundUp(blockSize);
         void[] result = void;
@@ -509,9 +510,10 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
         return available;
     }
 
+    pure nothrow @safe @nogc
     private void[] smallAlloc(uint blocks)
     {
-        assert(blocks >= 2 && blocks <= 64, text(blocks));
+        assert(blocks >= 2 && blocks <= 64);
         void[] result;
         foreach (i; _startIdx .. _control.rep.length)
         {
@@ -544,6 +546,7 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
         return null;
     }
 
+    pure nothrow @safe @nogc
     private void[] hugeAlloc(size_t blocks)
     {
         assert(blocks > 64);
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -806,7 +809,7 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
     uint blockSize = 64;
     auto a = BitmappedBlock!(chooseAtRuntime, 64)(cast(ubyte[])(r.allocateAll()), blockSize);
     static assert(hasMember!(InSituRegion!(10_240, 64), "allocateAll"));
-    const b = a.allocate(100);
+    const b = (() pure nothrow @safe @nogc => a.allocate(100))();
```
wilzbach: Why do we need this lambda? The `unittest` is still @system

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -816,7 +819,7 @@ struct BitmappedBlock(size_t theBlockSize, uint theAlignment = platformAlignment
 
     auto a = BitmappedBlock!(64, 64)(new ubyte[10_240]);
     assert((() nothrow @safe @nogc => a.empty)() == Ternary.yes);
-    const b = a.allocate(100);
+    const b = (() pure nothrow @safe @nogc => a.allocate(100))();
```
wilzbach: ditto - @system test

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -1216,13 +1223,13 @@ struct BitmappedBlockWithInternalPointers(
...
         if (!r.ptr) return r;
-        immutable block = (r.ptr - _heap._payload.ptr) / _heap.blockSize;
+        immutable block = (() @trusted => (r.ptr - _heap._payload.ptr) / _heap.blockSize)();
         immutable blocks =
             (r.length + _heap.blockSize - 1) / _heap.blockSize;
         if (!ensureRoomForAllocStart(block + blocks))
         {
             // Failed, free r and bailout
-            _heap.deallocate(r);
+            () @trusted { _heap.deallocate(r); }();
```
aG0aep6G: Nullify `r` in the `@trusted` section.

n8sh: For what purpose?

std/experimental/allocator/building_blocks/bitmapped_block.d
n8sh: It looks like it's so if `allocate` stops being `@safe` in the future this unittest will fail.

wilzbach: Sorry for being so unclear, I wanted to ask: "Why not make the unittest @safe?"
(we/I periodically grep for "@system unittest" for this list - https://github.com/dlang/phobos/projects/5)

std/experimental/allocator/building_blocks/bitmapped_block.d
aG0aep6G: Otherwise `r` is a dangling pointer in `@safe` code. Can't have dangling pointers in `@safe` code.

std/experimental/allocator/building_blocks/bitmapped_block.d
edi33416: All those attributed lambdas were added with each method that got attributed. We will be able to do a cleanup (i.e. add the attributes to the unittest itself) once all the methods have their attributes. I don't like the looks either :smile:

PR #6409, opened 2018-04-02
Add public examples to many functions to std.math

JackStouffer:


std/math.d
```diff
@@ -3724,18 +3750,18 @@ real log10(real x) @safe pure nothrow @nogc
...
+ * log(1 + x).
  *
  *  $(TABLE_SV
  *  $(TR $(TH x)            $(TH log1p(x))     $(TH divide by 0?) $(TH invalid?))
  *  $(TR $(TD $(PLUSMN)0.0) $(TD $(PLUSMN)0.0) $(TD no)           $(TD no))
  *  $(TR $(TD -1.0)         $(TD -$(INFIN))    $(TD yes)          $(TD no))
- *  $(TR $(TD $(LT)-1.0)    $(TD $(NAN))       $(TD no)           $(TD yes))
- *  $(TR $(TD +$(INFIN))    $(TD -$(INFIN))    $(TD no)           $(TD no))
+ *  $(TR $(TD $(LT)-1.0)    $(TD -$(NAN))      $(TD no)           $(TD yes))
+ *  $(TR $(TD +$(INFIN))    $(TD +$(INFIN))    $(TD no)           $(TD no))
```
JackStouffer: these where incorrect

std/math.d
```diff
@@ -4826,8 +4923,11 @@ version(Posix)
  * Returns the integer portion of x, dropping the fractional portion.
  *
  * This is also known as "chop" rounding.
+ *
+ * Returns:
+ *     A `real`.
  */
-real trunc(real x) @trusted nothrow @nogc
+auto trunc(real x) @trusted
```
JackStouffer: changed to `auto` to auto infer attributes like pure based on implementation without using template

std/math.d
```diff
@@ -4870,6 +4970,15 @@ real trunc(real x) @trusted nothrow @nogc
...
 }
 
+///
+@safe pure unittest
+{
+    assert(trunc(0.01) == 0);
+    assert(trunc(0.49) == 0);
+    assert(trunc(0.5) == 0);
+    assert(trunc(1.5) == 1);
+}
```
wilzbach: Looks like the remainder isn't available on Win32?

```
std\math.d(4908): Error: static assert:  "lround not implemented"
```

std/math.d
```diff
@@ -4809,12 +4905,13 @@ long lround(real x) @trusted nothrow @nogc
     version (Posix)
         return core.stdc.math.llroundl(x);
     else
-        assert(0, "lround not implemented");
+        static assert(0, "lround not implemented");
```
wilzbach: Looks like the remainder isn't available on Win32, but still run?

```
std\math.d(4908): Error: static assert:  "lround not implemented"
```

Maybe we should roll a custom, "slow" implementation of it?

std/math.d
```diff
@@ -4536,24 +4585,46 @@ if (is(typeof(rfunc(F.init)) : F) && isFloatingPoint!F)
...
+ *
  * If the return value is not equal to x, the FE_INEXACT
  * exception is raised.
- * $(B nearbyint) performs
- * the same operation, but does not set the FE_INEXACT exception.
+ *
+ * $(LREF nearbyint) performs the same operation, but does
+ * not set the FE_INEXACT exception.
  */
 real rint(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.rint(x); }
```
wilzbach: So we have `lround`, `nearbyint`, `round` and `rint` which all perform almost the same function. That's not very "abstracted high-level" over C  :/

std/math.d
```diff
@@ -3724,18 +3750,18 @@ real log10(real x) @safe pure nothrow @nogc
...
+ * For very small x, log1p(x) will be more accurate than
+ * log(1 + x).
  *
  *  $(TABLE_SV
  *  $(TR $(TH x)            $(TH log1p(x))     $(TH divide by 0?) $(TH invalid?))
  *  $(TR $(TD $(PLUSMN)0.0) $(TD $(PLUSMN)0.0) $(TD no)           $(TD no))
  *  $(TR $(TD -1.0)         $(TD -$(INFIN))    $(TD yes)          $(TD no))
- *  $(TR $(TD $(LT)-1.0)    $(TD $(NAN))       $(TD no)           $(TD yes))
- *  $(TR $(TD +$(INFIN))    $(TD -$(INFIN))    $(TD no)           $(TD no))
+ *  $(TR $(TD $(LT)-1.0)    $(TD -$(NAN))      $(TD no)           $(TD yes))
```
wilzbach: Could you add an explicit test for this? (i.e. that it's a negative nan)

std/math.d
n8sh: I feel the downside of not documenting the return type outweighs the benefit of attribute inference here.

JackStouffer: I added a note in the docs about the return type

std/math.d
JackStouffer: I have a test for this on line 3798, but because `-real.nan != -real.nan` there's really no way to test this

n8sh: Use `is`.

std/math.d
```diff
@@ -4812,62 +4919,88 @@ long lround(real x) @trusted nothrow @nogc
...
     }
 }
 
-/****************************************************
- * Returns the integer portion of x, dropping the fractional portion.
- *
- * This is also known as "chop" rounding.
- */
-real trunc(real x) @trusted nothrow @nogc
+version(StdDdoc)
```
JackStouffer: Ok, so apparently `ret` in ASM doesn't work with an `auto` return type. So we now have this mess.

PR #6410, opened 2018-04-02
Add AlignedBlockList

jercaianu:
Implemented AlignedBlockList, an allocator which holds a list of other allocators and allows for constant time deallocations, using aligned allocations.

Internally, this holds a doubly linked list of other allocators. Each node contains an allocator, followed by its payload. Each node will be allocated by the parent allocator at addresses multiple of 'theAlignment'.
Whenever a deallocation happens, the buffer candidate for deallocation will be rounded to the nearest multiple of 'theAlignment', in order to find in constant time the allocator from which the buffer originated.

This will play a very important part in creating a fast and safe allocator.

Thanks,
Alex


jercaianu:
ping @andralex @edi33416

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
+// Written in the D programming language.
+/**
```
wilzbach: Please provide a description here.

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+Source: $(PHOBOSSRC std/experimental/allocator/building_blocks/_aligned_block_list.d)
```
wilzbach: Preventing Ddoc's auto-hightlighting is no longer required - https://github.com/dlang/dlang.org/pull/2307

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+    }
+
+    AlignedBlockNode *root;
+    int numNodes;
+    enum maxNodes = 50;
+
+    static if (isShared)
+    SpinLock lock = SpinLock(SpinLock.Contention.brief);
+
+    private void moveToFront(AlignedBlockNode *tmp)
```
wilzbach: Document

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+            shared(size_t) bytesUsed;
+            uint keepAlive;
+        }
+        else
+        {
+            size_t bytesUsed;
+        }
+    }
+
+    AlignedBlockNode *root;
```
wilzbach: DStyle

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+    int numNodes;
+    enum maxNodes = 50;
+
+    static if (isShared)
+    SpinLock lock = SpinLock(SpinLock.Contention.brief);
```
wilzbach: Indent (DStyle)

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+/**
+Source: $(PHOBOSSRC std/experimental/allocator/building_blocks/_aligned_block_list.d)
+*/
+module std.experimental.allocator.building_blocks.aligned_block_list;
+
+import std.experimental.allocator.common;
+import std.experimental.allocator.building_blocks.null_allocator;
+
+// Common function implementation for thread local and shared AlignedBlockList
+private mixin template AlignedBlockListImpl(bool isShared)
```
wilzbach: Better use std.typecons.Flag + document

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+    import std.traits : hasMember;
+    import std.typecons : Ternary;
+    static if (isShared)
+    import core.internal.spinlock : SpinLock;
+
+private:
+    struct AlignedBlockNode
+    {
+        AlignedBlockNode* next, prev;
+        Allocator bAlloc;
```
wilzbach: "b" for block? It would be good if we either rename it to `alloc` or `blockAllocator` to avoid confusion and needless wonderings when looking at the code.

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+
+    enum ulong alignment = theAlignment;
+
+    static if (hasMember!(ParentAllocator, "owns"))
+    Ternary owns(void[] b)
+    {
+        return parent.owns(b);
+    }
+
+    bool deallocate(void[] b)
```
wilzbach: Document

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+    {
+        static if (isShared)
+        import core.atomic : atomicOp;
```
wilzbach: Move this to where the import is actually used

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+        AlignedBlockNode *node = cast(AlignedBlockNode*) ptr;
+        if (node.bAlloc.deallocate(b))
+        {
+            static if (isShared)
+            {
+                atomicOp!"-="(node.bytesUsed, b.length);
+            }
+            else
+            {
+                node.bytesUsed -= b.length;
```
wilzbach: You might want to consider using your own local function of `atomicOp` that either uses the real `atomicOp` or uses direct assignment - depending on `isShared`

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+    {
+        import std.typecons : Ternary;
+        import std.traits : hasMember;
+
+        /**
+        Returns a chunk of memory of size `n`
+        It finds the first node in the `AlignedBlockNode` list which has available memory,
+        and moves it to the front of the list.
+
+        All empty nodes which cannot return new memory, are removed from the list.
```
wilzbach: needs Params + Returns

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+            SharedAscendingPageAllocator,
+            1 << 16);
+    void[][totalAllocs] buf;
+
+    SuperAllocator a;
+    a.parent = SharedAscendingPageAllocator(4096 * 1024);
+
+    void fun()
+    {
+        auto rnd = Random();
```
wilzbach: No. Use a fixed seed and a fixed engine, everything else might lead to random coverage changes.

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+
+    SuperAllocator a;
+    auto pageAlloc = AscendingPageAllocator(4096 * 4096);
+    a.allocatorForSize!4096 = &pageAlloc;
+    a.allocatorForSize!2048.parent = &pageAlloc;
+    a.allocatorForSize!1024.parent = &pageAlloc;
+    a.allocatorForSize!512.parent = &pageAlloc;
+    a.allocatorForSize!256.parent = &pageAlloc;
+
+    auto rnd = Random();
```
wilzbach: Seed + fix the generator

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+`AlignedBlockList` represents a wrapper around a chain of allocators, allowing for fast deallocations
+and preserving a low degree of fragmentation.
+The allocator holds internally a doubly linked list of `Allocator` objects, which will serve allocations
+in a most-recently-used fashion. Most recent allocators used for `allocate` calls, will be
+moved to the front of the list.
+
+Although allocations are in theory served in linear searching time, `deallocate` calls take
+$(BIGOH 1) time, by using aligned allocations. `ParentAllocator` must implement `alignedAllocate`
+and it must be able to allocate `theAlignment` bytes at the same alignment. Each aligned allocation
+done by `ParentAllocator` will contain metadata for an `Allocator`, followed by its payload.
```
wilzbach: Needs a ddoc Params section.

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+            root.bytesUsed += result.length;
+        }
+
+        return result;
+    }
+
+    size_t goodAllocSize(const size_t n)
+    {
+        Allocator a = null;
+        return a.goodAllocSize(n);
```
wilzbach: Wouldn't this segfault?

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+                lock.unlock();
+            }
+
+            auto result = tmp.bAlloc.allocate(n);
+            if (result.length == n)
+            {
+                static if (isShared)
+                {
+                    atomicOp!"+="(tmp.bytesUsed, n);
+                    lock.lock();
```
wilzbach: You never unlock this

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,629 @@
...
+// Common function implementation for thread local and shared AlignedBlockList
+private mixin template AlignedBlockListImpl(bool isShared)
+{
+    import std.traits : hasMember;
+    import std.typecons : Ternary;
+    static if (isShared)
+    import core.internal.spinlock : SpinLock;
+
+private:
+    struct AlignedBlockNode
```
n8sh: `static struct`?

std/experimental/allocator/building_blocks/aligned_block_list.d
jercaianu: If goodAllocSize does not use internal allocator state (and it shouldn't) it won't segfault.
Currently, this would work with all allocators in experimental

std/experimental/allocator/building_blocks/aligned_block_list.d
jercaianu: A bit above I have a scope(exit) lock.unlock()

jercaianu:
@wilzbach Thanks for comments

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,680 @@
...
+        mixin AlignedBlockListImpl!true;
+    }
+}
+
+///
+@system unittest
+{
+    import std.experimental.allocator.building_blocks.region : SharedRegion;
+    import std.experimental.allocator.building_blocks.ascending_page_allocator : SharedAscendingPageAllocator;
+    import std.std.experimental.allocator.building_blocks.null_allocator : NullAllocator;
```
wilzbach: std/experimental/allocator/building_blocks/aligned_block_list.d(512): Error: module `null_allocator` is in file 'std/std/experimental/allocator/building_blocks/null_allocator.d' which cannot be read
import path[0] = ../druntime/import
posix.mak:350: recipe for target 'generated/linux/debug/32/unittest/std/experimental/allocator/building_blocks/aligned_block_list.o' failed
make[1]: *** [generated/linux/debug/32/unittest/std/experimental/allocator/building_blocks/aligned_block_list.o] Error 1

-> You have a `std.std`

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,690 @@
...
+    }
+
+    // Root of the internal doubly linked list
+    AlignedBlockNode* root;
+
+    // Number of active nodes
+    int numNodes;
+
+    // If the numNodes exceeds this limit, we will start deallocating nodes
+    enum maxNodes = 50;
```
n8sh: I'm curious as to where this number came from. Was it just a gut feeling? Did you make a decision to bias towards "probably not too high" or "probably not too low"?

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,690 @@
...
+    }
+
+    // Removes a node from the list, including its payload
+    // The payload is deallocated by calling 'parent.deallocate'
+    void removeNode(AlignedBlockNode* tmp)
+    {
+        auto next = tmp.next;
+        if (tmp.prev) tmp.prev.next = tmp.next;
+        if (tmp.next) tmp.next.prev = tmp.prev;
+        assert(parent.deallocate((cast(void*) tmp)[0 .. alignment]));
```
n8sh: In release mode assertions are omitted and this code won't be run.

Example: https://run.dlang.io/is/m8gzG3
```d
import std.stdio;
auto foo(ref int x)
{
    return x &= 1;
}
void main(string[] args)
{
    int x = 255;
    assert(foo(x));
    writeln(x);
}
```

This prints 1 when run with no compiler arguments but prints 255 with `-release`.

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,690 @@
...
+        }
+
+        return true;
+    }
+
+public:
+    static if (stateSize!ParentAllocator) ParentAllocator parent;
+    else alias parent = ParentAllocator.instance;
+
+    enum ulong alignment = theAlignment;
```
n8sh: I don't see how this could be correct if `theAlignment` is the size of each node and memory is allocated at an offset from the start of the node.

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,690 @@
...
+in a most-recently-used fashion. Most recent allocators used for `allocate` calls, will be
+moved to the front of the list.
+
+Although allocations are in theory served in linear searching time, `deallocate` calls take
+$(BIGOH 1) time, by using aligned allocations. `ParentAllocator` must implement `alignedAllocate`
+and it must be able to allocate `theAlignment` bytes at the same alignment. Each aligned allocation
+done by `ParentAllocator` will contain metadata for an `Allocator`, followed by its payload.
+
+Params:
+    Allocator = the allocator which is used to manage each node
```
n8sh: The documentation does an insufficient job of explaining what needs to be true of the Allocator!

n8sh:
It isn't clear to me that the `shared` version is correct. Does using a spinlock prevent stale reads of the `root` pointer and prevent stale reads of `AlignedBlockNode`'s `next`/`prev` pointers?

std/experimental/allocator/building_blocks/aligned_block_list.d
jercaianu: Thanks, good catch!

std/experimental/allocator/building_blocks/aligned_block_list.d
jercaianu: Since the metadata for each node and the payload are adjacent in memory, I considered the alignment to be for the whole block (metadata + payload).
Inside the payload, an `Allocator` can manage the allocations with its own alignment.

std/experimental/allocator/building_blocks/aligned_block_list.d
jercaianu: Yes, you're right, I haven't explained here too much about the `Allocator`.
`Allocator` can be anything, as long as it has a constructor which receives `ubyte[]` and its `ParentAllocator` is the `NullAllocator`.

For the moment, these are the only constraints which come to mind.
Thanks!

jercaianu:
@n8sh 
Thanks for the review!
The list only is changed when the lock is taken. Only under the lock new nodes are added/removed and root updated.
To increase performance, `allocate` happens when the lock is not taken. Each node has `keepAlive`, which basically acts as a refcount to make sure no node is removed while it is used for allocations, outside the lock.

Besides the unittests, I stress tested this and I haven't found any races.

std/experimental/allocator/building_blocks/aligned_block_list.d
jercaianu: Depending on what the caller passes, `maxNodes` should be different.
The deallocation of the blocks provided by `ParentAllocator` happen lazily inside `allocate`. 
I didn't want to delete nodes too often, but also I didn't want to have high memory usage so I hardcoded something for my tests.
I'll investigate here a bit more to see if I can get rid of this, or maybe pass it as a parameter somehow.
Thanks!

std/experimental/allocator/building_blocks/aligned_block_list.d
jercaianu: At the same time, I think it's not great for the caller to expect each allocation to have an alignment, but we silently add some padding before.
I think it's better for the alignment to be (metadata + payload), but clearly specify that and expose the length of the metadata.

n8sh: Because other allocators in `std.experiment.allocator` use `alignment` to specifically mean the alignment of memory returned by `allocate`, use a different name here and everything will be fine. Otherwise, templated code that uses design-by-introspection can leap to a wrong conclusion -- the compiler doesn't read the DDOC, so a comment will not prevent that!

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -130,7 +130,7 @@ public:
     static if (stateSize!ParentAllocator) ParentAllocator parent;
     else alias parent = ParentAllocator.instance;
 
-    enum ulong alignment = theAlignment;
+    enum ulong alignment = ParentAllocator.alignment;
```
n8sh: If `alignment` means the alignment of the memory returned by `allocate`, it would appear this should be `Allocator.alignment` instead of `ParentAllocator.alignment`.

Also maybe it should have a guard like:
```d
static if (hasStaticallyKnownAlignment!Allocator)
    enum alignment = Allocator.alignment;
```
The existence of `std.experimental.allocator.common : hasStaticallyKnownAlignment` implies that not all allocators have statically known alignments, but this may be obsolete or not applicable to the specific allocators required here.

std/experimental/allocator/building_blocks/aligned_block_list.d
andralex: bool is OK for a private artifact

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,692 @@
...
+        {
+            size_t bytesUsed;
+        }
+    }
+
+    // Root of the internal doubly linked list
+    AlignedBlockNode* root;
+
+    // Number of active nodes
+    int numNodes;
```
andralex: `uint`

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,692 @@
...
+
+    // If the numNodes exceeds this limit, we will start deallocating nodes
+    enum maxNodes = 64;
```
andralex: `enum uint`

std/experimental/allocator/building_blocks/aligned_block_list.d
```diff
@@ -0,0 +1,692 @@
...
+        return true;
+    }
+
+public:
+    static if (stateSize!ParentAllocator) ParentAllocator parent;
+    else alias parent = ParentAllocator.instance;
+
+    enum ulong alignment = Allocator.alignment;
+
+    static if (hasMember!(ParentAllocator, "owns"))
```
andralex: document

std/experimental/allocator/building_blocks/aligned_block_list.d
edi33416: Wouldn't `size_t` be better?

andralex: `size_t` is size of objects in memory. For counts we can use that of course, but just incidentally.

PR #6411, opened 2018-04-03
make/makeArray: take advantage of allocators with smth. faster than allocate+memset zero

n8sh:
Some allocators can allocate zeroed memory more efficiently than allocate + memset.  This pull request takes advantage of this in `std.experimental.allocator : make` and `makeArray`.

This PR adds non-public `allocateZeroed` methods to `Mallocator` and `MmapAllocator` and `GCAllocator`. Various derived allocators could also implement such methods when their parent allocators support them, but that work is left to a future PR (if this PR is merged).

Ignoring whitespace changes when reviewing this PR should make `make` clearer.

### DIFF
--- a/std/experimental/allocator/common.d
+++ b/std/experimental/allocator/common.d
@@ -511,6 +511,18 @@ version(unittest)
         assert(b2.length == 2);
         assert(b2.ptr + b2.length <= b1.ptr || b1.ptr + b1.length <= b2.ptr);
 
+        // Test allocateZeroed
+        static if (hasMember!(A, "allocateZeroed"))
+        {{
+            auto b3 = a.allocateZeroed(8);
+            if (b3 !is null)
+            {
+                assert(b3.length == 8);
+                foreach (e; cast(ubyte[]) b3)
+                    assert(e == 0);
+            }
+        }}
+
         // Test alignedAllocate
         static if (hasMember!(A, "alignedAllocate"))
         {{
--- a/std/experimental/allocator/gc_allocator.d
+++ b/std/experimental/allocator/gc_allocator.d
@@ -111,6 +111,13 @@ struct GCAllocator
         return ((n + 4095) / 4096) * 4096;
     }
 
+    package pure nothrow @trusted void[] allocateZeroed(size_t bytes) shared const
+    {
+        if (!bytes) return null;
+        auto p = GC.calloc(bytes);
+        return p ? p[0 .. bytes] : null;
+    }
+
     /**
     Returns the global instance of this allocator type. The garbage collected
     allocator is thread-safe, therefore all of its methods and `instance` itself
--- a/std/experimental/allocator/mallocator.d
+++ b/std/experimental/allocator/mallocator.d
@@ -64,6 +64,15 @@ struct Mallocator
         return true;
     }
 
+    @trusted @nogc nothrow pure
+    package void[] allocateZeroed(size_t bytes) shared
+    {
+        import core.memory : pureCalloc;
+        if (!bytes) return null;
+        auto p = pureCalloc(1, bytes);
+        return p ? p[0 .. bytes] : null;
+    }
+
     /**
     Returns the global instance of this allocator type. The C heap allocator is
     thread-safe, therefore all of its methods and `it` itself are
--- a/std/experimental/allocator/mmap_allocator.d
+++ b/std/experimental/allocator/mmap_allocator.d
@@ -48,6 +48,21 @@ struct MmapAllocator
             if (b.ptr) munmap(b.ptr, b.length) == 0 || assert(0);
             return true;
         }
+
+        // Anonymous mmap might be zero-filled on all Posix systems but
+        // not all commit to this in the documentation.
+        version(linux)
+            // http://man7.org/linux/man-pages/man2/mmap.2.html
+            package alias allocateZeroed = allocate;
+        else version(NetBSD)
+            // http://netbsd.gw.com/cgi-bin/man-cgi?mmap+2+NetBSD-current
+            package alias allocateZeroed = allocate;
+        else version(Solaris)
+            // https://docs.oracle.com/cd/E88353_01/html/E37841/mmap-2.html
+            package alias allocateZeroed = allocate;
+        else version(AIX)
+            // https://www.ibm.com/support/knowledgecenter/en/ssw_aix_71/com.ibm.aix.basetrf1/mmap.htm
+            package alias allocateZeroed = allocate;
     }
     else version(Windows)
     {
@@ -71,6 +86,8 @@ struct MmapAllocator
         {
             return b.ptr is null || VirtualFree(b.ptr, 0, MEM_RELEASE) != 0;
         }
+
+        package alias allocateZeroed = allocate;
     }
 }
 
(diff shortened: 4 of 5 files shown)


std/experimental/allocator/common.d
```diff
@@ -512,6 +512,18 @@ version(unittest)
...
 
+        // Test allocateZeroed
+        static if (hasMember!(A, "allocateZeroed"))
+        {{
+            auto b3 = a.allocateZeroed(8);
+            if (b3 !is null)
+            {
+                assert(b3.length == 8);
+                foreach (e; cast(ubyte[]) b3)
+                    assert(e is 0);
```
wilzbach: Out of interest: why `is` and not `==`?

std/experimental/allocator/package.d
```diff
@@ -1121,40 +1121,49 @@ propagates the exception.
...
     {
-        static if (is(T == class)) return emplace!T(m, args);
-        else
-        {
-            // Assume cast is safe as allocation succeeded for `stateSize!T`
-            auto p = () @trusted { return cast(T*) m.ptr; }();
-            emplaceRef(*p, args);
-            return p;
-        }
+        auto m = alloc.allocateZeroed(T.sizeof ? T.sizeof : 1);
```
wilzbach: You don't like the `stateSize` that is used in the else branch?

std/experimental/allocator/package.d
```diff
@@ -1121,40 +1121,49 @@ propagates the exception.
...
+        return (() @trusted => cast(T*) m.ptr)();
```
wilzbach: We should probably add a test that `isInitAllZeroBits` returns false when for `@disable this`.
(there's none atm)

std/experimental/allocator/common.d
n8sh: No reason.

std/experimental/allocator/package.d
n8sh: It didn't seem to do anything applicable. It would sometimes give us 0 instead of `T.sizeof` but we don't ever want 0.

std/experimental/allocator/package.d
n8sh: For instance, in `uninitializedFillDefault` we want to know what's in `T.init` and we don't care if the default constructor is disabled.

n8sh: (Check added to condition)

std/experimental/allocator/package.d
```diff
@@ -1484,10 +1494,18 @@ exception if the copy operation throws.
 T[] makeArray(T, Allocator)(auto ref Allocator alloc, size_t length)
 {
     if (!length) return null;
-    auto m = alloc.allocate(T.sizeof * length);
-    if (!m.ptr) return null;
-    alias U = Unqual!T;
-    return () @trusted { return cast(T[]) uninitializedFillDefault(cast(U[]) m); }();
+    static if (isInitAllZeroBits!T && is(typeof(alloc.allocateZeroed(length))))
```
wilzbach: This also needs to check that the element is actually constructable (i.e. doesn't have `@disable this;`)

n8sh: That wouldn't change the result because the other branch (i.e. the current implementation) uses `uninitializedFillDefault` which ensures the array is filled with bitwise copies of `T.init` and doesn't care about `@disable this()`.

std/experimental/allocator/package.d
edi33416: How about `max(T.sizeof, 1)` to be consistent with the behaviour in the `else` branch?

std/experimental/allocator/gc_allocator.d
```diff
@@ -111,6 +111,13 @@ struct GCAllocator
         return ((n + 4095) / 4096) * 4096;
     }
 
+    package pure nothrow @trusted void[] allocateZeroed(size_t bytes) shared
```
edi33416: Which `package`? Are you planning to make those public in a future PR?

std/experimental/allocator/mallocator.d
```diff
@@ -65,6 +65,15 @@ struct Mallocator
         return true;
     }
 
+    @trusted @nogc nothrow pure
+    package void[] allocateZeroed(size_t bytes) shared
```
edi33416: ditto

std/experimental/allocator/mmap_allocator.d
```diff
@@ -48,6 +48,21 @@ struct MmapAllocator
             if (b.ptr) munmap(b.ptr, b.length) == 0 || assert(0);
             return true;
         }
+
+        // Anonymous mmap might be zero-filled on all Posix systems but
+        // not all commit to this in the documentation.
+        version(linux)
+            // http://man7.org/linux/man-pages/man2/mmap.2.html
+            package alias allocateZeroed = allocate;
```
edi33416: ditto

std/experimental/allocator/gc_allocator.d
n8sh: I think it would make sense, but this PR still works if it's entirely under the hood.

n8sh:
>How about max(T.sizeof, 1) to be consistent with the behaviour in the else branch?

For some reason I can't reply to that comment directly, but I've changed it to be the same.

std/experimental/allocator/package.d
n8sh: OK sure

std/experimental/allocator/common.d
```diff
@@ -512,6 +512,18 @@ version(unittest)
         assert(b2.length == 2);
         assert(b2.ptr + b2.length <= b1.ptr || b1.ptr + b1.length <= b2.ptr);
 
+        // Test allocateZeroed
+        static if (hasMember!(A, "allocateZeroed"))
+        {{
+            auto b3 = a.allocateZeroed(8);
```
wilzbach: Looks like the recent GCAllocator PR (#6432) broke this:

```
std/experimental/allocator/common.d(518): Error: `shared` mutable method `std.experimental.allocator.gc_allocator.GCAllocator.allocateZeroed` is not callable using a `shared` `const` object
std/experimental/allocator/common.d(518):        Consider adding `const` or `inout` to std.experimental.allocator.gc_allocator.GCAllocator.allocateZeroed
```

std/experimental/allocator/gc_allocator.d
wilzbach: Looks like this has to be `const` due to #6432

JackStouffer:
Looks like there's still discussion, so removing this label.

wilzbach:
@JackStouffer FYI: there are no open discussions that need to be addressed in this PR. While @edi33416, @nordlow and @n8sh would like to make `allocateZeroed` public, this discussion can and should happen in another PR as adding a new public symbol to Phobos is always a complicated / time-consuming process.

(Anyhow this currently can't be merged due to the missing `const`, but I just added that one, so this PR should be passing soon)

edi33416:
@n8sh could you do a rebase so we can then toggle the auto-merge? Cheers

n8sh:
done

wilzbach:
That's a known segfault issue with std.file.dirEntries :/
I restarted.

JinShil:
According to digger, this PR caused a regression: https://issues.dlang.org/show_bug.cgi?id=18871

n8sh:
@JinShil There's an open issue for the relevant DMD bug: https://issues.dlang.org/show_bug.cgi?id=18819

nordlow:
See https://forum.dlang.org/post/ybdhdcfyraukhhpgojpf@forum.dlang.org, @wilzbach ?

std/experimental/allocator/common.d
nordlow: Is this still a problem?

PR #6413, opened 2018-04-03
Fix Issue 17806 - processAllocator getter will override set value if …

edi33416:
…it was set before getter was called at least once

### DIFF
--- a/std/experimental/allocator/package.d
+++ b/std/experimental/allocator/package.d
@@ -226,6 +226,18 @@ module std.experimental.allocator;
 public import std.experimental.allocator.common,
     std.experimental.allocator.typed;
 
+// Fix issue 17806: this should always be the first unittest in this module
+// in order to ensure that we use the `processAllocator` setter before the getter
+@system unittest
+{
+    import std.experimental.allocator.mallocator: Mallocator;
+    import std.experimental.allocator.gc_allocator : GCAllocator;
+    auto newAlloc = sharedAllocatorObject(Mallocator.instance);
+    processAllocator = newAlloc;
+    assert(processAllocator is newAlloc);
+    processAllocator = sharedAllocatorObject(GCAllocator.instance);
+}
+
 // Example in the synopsis above
 @system unittest
 {
@@ -1038,10 +1050,10 @@ allocator can be cast to `shared`.
 
 /// Ditto
 nothrow @system @nogc
-@property void processAllocator(RCISharedAllocator a)
+@property void processAllocator(ref RCISharedAllocator a)
 {
     assert(!a.isNull);
-    _processAllocator = a;
+    processAllocator() = a;
 }
 
 @system unittest


std/experimental/allocator/package.d
```diff
@@ -226,6 +226,22 @@ module std.experimental.allocator;
 public import std.experimental.allocator.common,
     std.experimental.allocator.typed;
 
+shared static this()
+{
+    import std.experimental.allocator.gc_allocator : GCAllocator;
+    _processAllocator = sharedAllocatorObject(GCAllocator.instance);
+}
+
+unittest
```
wilzbach: std/experimental/allocator/package.d(235:1)[warn]: A unittest should be annotated with at least @safe or @system

std/experimental/allocator/package.d
```diff
@@ -226,6 +226,22 @@ module std.experimental.allocator;
 public import std.experimental.allocator.common,
     std.experimental.allocator.typed;
 
+shared static this()
```
wilzbach: object.Error@src/rt/minfo.d(371): Cyclic dependency between module std.experimental.allocator and std.parallelism
std.experimental.allocator* ->
std.algorithm.comparison ->
std.parallelism* ->
std.file ->
std.experimental.checkedint ->
std.experimental.allocator.common ->
std.experimental.allocator*

std/experimental/allocator/package.d
```diff
@@ -226,6 +226,18 @@ module std.experimental.allocator;
 public import std.experimental.allocator.common,
     std.experimental.allocator.typed;
 
+// Fix issue 17806: this should always be the first unittest in this module
+// in order to ensure that we use the `processAllocator` setter before the getter
+@system unittest
+{
+    import std.experimental.allocator.mallocator: Mallocator;
```
wilzbach: spacing between `:`
From the CI:

```
std/experimental/allocator/package.d:233:    import std.experimental.allocator.mallocator: Mallocator;
```

std/experimental/allocator/package.d
```diff
@@ -226,6 +226,18 @@ module std.experimental.allocator;
...
+    import std.experimental.allocator.gc_allocator : GCAllocator;
+    auto newAlloc = sharedAllocatorObject(Mallocator.instance);
+    processAllocator = newAlloc;
+    assert(processAllocator is newAlloc);
+    processAllocator = sharedAllocatorObject(GCAllocator.instance);
```
wilzbach: Maybe not implicitly assume that it was the GCAllocator before and just reset it to the old value?

edi33416: This would require the setter to return the previous allocator, which it currently doesn't. In order to change the API to do this, we'd need to guard with a `lock`. We can regard this as a best effort.

n8sh: >This would require the setter to return the previous allocator, which it currently doesn't.

That would also be a potential landmine since for basic types the result of `(a = b)` is the post-modification value of `a`, not the pre-modification value of `a`.

PR #6415, opened 2018-04-03
Make Mallocator use pureMalloc, pureFree and pureRealloc

edi33416:


edi33416:
Would the same changes apply to `AlignedMallocator`?

I am not sure if `core.sys.posix.stdlib.posix_memalign` and `_aligned_malloc` (DMD Win64 uses Microsoft's std C lib) are `pure`

nordlow:
Are there no plans for wrapping `calloc` aswell, say, in a member named `allocateZeros` like at

https://github.com/nordlow/phobos-next/blob/master/src/pure_mallocator.d#L36

in use at

https://github.com/nordlow/phobos-next/blob/master/src/container_traits.d#L544

andralex:
@nordlow we had such in an early iteration of the allocators framework, but dropped it. Not many D objects are bitwise zeros. What use cases are problematic?

nordlow:
No problems, just pontentially better performance now that we have `isInitAllZeroBits` and `isAllZeroBits` in Phobos which checks if a default value is zero bits only.

n8sh:
>Are there any plans for wrapping calloc aswell, say, in a member named, say, `allocateZeros` or `zeroallocate`

There's already PR https://github.com/dlang/phobos/pull/6411. Even more than `calloc` it's motivated by `mmap` memory being zero by default. The PR doesn't make those methods public (yet) but it uses them internally in `make`/`makeArray` when appropriate.

n8sh:
>Not many D objects are bitwise zeros

Zero-initialization is not uncommon in data structures: for example the D runtime's associative array implementation has an internal array of `struct Bucket { size_t hash; void* entry; }` that it allocates using `GC.calloc`. I imagine the most common case will be arrays of pointers. Maybe also allocating buffers, although in most cases buffers don't need their contents to be default initialized.

nordlow:
Great, @n8sh! Stars are aligning now!

PR #6417, opened 2018-04-04
Add moveFile

marler8997:
Added the `moveFile` function to move a single file that can work across drives and filesystems.

For windows this means adding a flag to the `MoveFile` function, for posix it means calling rename and falling back to copy on failure.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -887,7 +887,8 @@ version(Windows) private void writeImpl(const(char)[] name, const(FSChar)* namez
 }
 
 /***************************************************
- * Rename file `from` _to `to`.
+ * Rename file `from` _to `to`. Note that this may fail if `from` and `to` are on
+ * different drives/filesystems, use `moveFile` to support these scenarios.
  * If the target file exists, it is overwritten.
  * Params:
  *    from = string or range of characters representing the existing file name
@@ -914,7 +915,7 @@ if ((isInputRange!RF && !isInfinite!RF && isSomeChar!(ElementEncodingType!RF) ||
     else
         enum string t = null;
 
-    renameImpl(f, t, fromz, toz);
+    moveFileImpl(f, t, fromz, toz, No.canCopy);
 }
 
 /// ditto
@@ -939,24 +940,96 @@ if (isConvertibleToString!RF || isConvertibleToString!RT)
 @safe unittest
 {
     auto t1 = deleteme, t2 = deleteme~"2";
-    scope(exit) foreach (t; [t1, t2]) if (t.exists) t.remove();
+    scope(exit) t2.remove();
 
     t1.write("1");
     t1.rename(t2);
     assert(t2.readText == "1");
+    assert(!t1.exists);
 
     t1.write("2");
     t1.rename(t2);
     assert(t2.readText == "2");
+    assert(!t1.exists);
 }
 
-private void renameImpl(const(char)[] f, const(char)[] t, const(FSChar)* fromz, const(FSChar)* toz) @trusted
+@safe unittest
+{
+    import std.utf : byWchar;
+
+    auto t1 = deleteme, t2 = deleteme~"2";
+    scope(exit) t2.remove();
+
+    write(t1, "1");
+    rename(t1, t2);
+    assert(readText(t2) == "1");
+    assert(!t1.exists);
+
+    write(t1, "2");
+    rename(t1, t2.byWchar);
+    assert(readText(t2) == "2");
+    assert(!t1.exists);
+}
+
+
+/***************************************************
+Move file `from` _to `to`.
+If the target file exists, it is overwritten. Unlike `rename` this function supports
+moving files accross drives/filesystems.
+Params:
+   from = string or range of characters representing the existing file name
+   to = string or range of characters representing the target file name
+Throws: `FileException` on error.
+*/
+void moveFile(RF, RT)(RF from, RT to)
+if ((isInputRange!RF && !isInfinite!RF && isSomeChar!(ElementEncodingType!RF) || isSomeString!RF)
+    && !isConvertibleToString!RF &&
+    (isInputRange!RT && !isInfinite!RT && isSomeChar!(ElementEncodingType!RT) || isSomeString!RT)
+    && !isConvertibleToString!RT)
+{
+    // Place outside of @trusted block
+    auto fromz = from.tempCString!FSChar();
+    auto toz = to.tempCString!FSChar();
+
+    static if (isNarrowString!RF && is(Unqual!(ElementEncodingType!RF) == char))
+        alias f = from;
+    else
+        enum string f = null;
+
+    static if (isNarrowString!RT && is(Unqual!(ElementEncodingType!RT) == char))
+        alias t = to;
+    else
+        enum string t = null;
+
+    moveFileImpl(f, t, fromz, toz, Yes.canCopy);
+}
+
+///
+@safe unittest
+{
+    auto t1 = deleteme, t2 = deleteme~"2";
+    scope(exit) foreach (t; [t1, t2]) if (t.exists) t.remove();
+
+    t1.write("1");
+    t1.moveFile(t2);
+    assert(t2.readText == "1");
+    assert(!t1.exists);
+
+    t1.write("2");
+    t1.moveFile(t2);
+    assert(t2.readText == "2");
+    assert(!t1.exists);
+}
+
+private void moveFileImpl(const(char)[] f, const(char)[] t, const(FSChar)* fromz,
+    const(FSChar)* toz, Flag!"canCopy" canCopy) @trusted
 {
     version(Windows)
     {
         import std.exception : enforce;
 
-        const result = MoveFileExW(fromz, toz, MOVEFILE_REPLACE_EXISTING);
+        const result = MoveFileExW(fromz, toz, MOVEFILE_REPLACE_EXISTING |
+            (canCopy ? MOVEFILE_COPY_ALLOWED : 0));
         if (!result)
         {
             import core.stdc.wchar_ : wcslen;
@@ -970,14 +1043,23 @@ private void renameImpl(const(char)[] f, const(char)[] t, const(FSChar)* fromz, 
 
             enforce(false,
                 new FileException(
-                    text("Attempting to rename file ", f, " to ", t)));
+                    text("Attempting to ", canCopy ? "move" : "rename" , " file ", f, " to ", t)));
         }
     }
     else version(Posix)
     {
         static import core.stdc.stdio;
-
-        cenforce(core.stdc.stdio.rename(fromz, toz) == 0, t, toz);
+        auto result = core.stdc.stdio.rename(fromz, toz);
+        if (result != 0)
+        {
+            if (canCopy && result == EXDEV)
+            {
+                copyImpl(f, t, fromz, toz, PreserveAttributes.yes);
+                removeImpl(f, fromz);
+            }
+            else
+                cenforce(false, t, toz);
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


jmdavis:
Given that `rename` works with directories, and copying and then deleting the original would require recursively copying the directory, I think that it's a mistake to add copying behavior to `rename`. And since `std.file.copy` only works with files, this code as-is will fail if you try to move a directory, and the target is on another filesystem. As such, I think that we should add a separate function called either `move` or `moveFile` which only operates on files but which will copy and delete if `rename` fails, leaving `rename` to work with both files and directories so long as the the target is on the same filesystem. It will also be more user-friendly to call `move` or `moveFile` than to call `rename!(CanCopy.yes)`, especially when you consider that many folks will just want to use the new version by default rather than dealing with the fact that `rename` could fail due to the target being on a different filesystem.

std/file.d
```diff
@@ -946,13 +947,14 @@ if (isConvertibleToString!RF || isConvertibleToString!RT)
     assert(t2.readText == "2");
 }
 
-private void renameImpl(const(char)[] f, const(char)[] t, const(FSChar)* fromz, const(FSChar)* toz) @trusted
+private void renameImpl(const(char)[] f, const(char)[] t, const(FSChar)* fromz, const(FSChar)* toz, Flag!"canCopy" canCopy) @trusted
```
wilzbach: std/file.d(950:116)[warn]: Line is longer than 120 characters

changelog/std-file-can-copy.dd
```diff
@@ -0,0 +1,7 @@
+Added Flag!"canCopy" to std.file.rename
+
+Added Flag!"canCopy" to std.file.rename:
```
wilzbach: Use `REF` here, s.t. it's clickable, e.g. `$(REF rename, std,file)`

std/file.d
marler8997: going to refactor this based on @jmdavis  comments (no need to review this code since it's all going to change)

std/file.d
```diff
@@ -946,13 +947,96 @@ if (isConvertibleToString!RF || isConvertibleToString!RT)
...
+Params:
+   from = string or range of characters representing the existing file name
+   to = string or range of characters representing the target file name
+Throws: `FileException` on error.
+*/
+void moveFile(RF, RT)(RF from, RT to)
+if ((isInputRange!RF && !isInfinite!RF && isSomeChar!(ElementEncodingType!RF) || isSomeString!RF)
+    && !isConvertibleToString!RF &&
+    (isInputRange!RT && !isInfinite!RT && isSomeChar!(ElementEncodingType!RT) || isSomeString!RT)
+    && !isConvertibleToString!RT)
```
wilzbach: We really need a good `isStringLike` trait :/
See e.g. https://github.com/dlang/phobos/pull/5137

std/file.d
```diff
@@ -946,13 +947,96 @@ if (isConvertibleToString!RF || isConvertibleToString!RT)
...
+
+///
+@safe unittest
+{
+    auto t1 = deleteme, t2 = deleteme~"2";
+    scope(exit) foreach (t; [t1, t2]) if (t.exists) t.remove();
+
+    t1.write("1");
+    t1.moveFile(t2);
+    assert(t2.readText == "1");
```
wilzbach: assert(!t1.exists);

std/file.d
```diff
@@ -946,13 +947,96 @@ if (isConvertibleToString!RF || isConvertibleToString!RT)
...
+
+    t1.write("2");
+    t1.moveFile(t2);
+    assert(t2.readText == "2");
```
wilzbach: assert(!t1.exists);

std/file.d
```diff
@@ -946,13 +947,96 @@ if (isConvertibleToString!RF || isConvertibleToString!RT)
...
+        alias t = to;
+    else
+        enum string t = null;
+
+    moveFileImpl(f, t, fromz, toz, Yes.canCopy);
+}
+
+/// ditto
+void moveFile(RF, RT)(auto ref RF from, auto ref RT to)
+if (isConvertibleToString!RF || isConvertibleToString!RT)
```
wilzbach: IIRC this overload was only added for legacy compatibility. So I'm not sure whether new symbols need it.
Or in other words: do we want to support aliased-strings for this new symbol.

std/file.d
```diff
@@ -946,13 +947,96 @@ if (isConvertibleToString!RF || isConvertibleToString!RT)
...
+    t1.write("1");
+    t1.moveFile(t2);
+    assert(t2.readText == "1");
+
+    t1.write("2");
+    t1.moveFile(t2);
+    assert(t2.readText == "2");
+}
+
+private void moveFileImpl(const(char)[] f, const(char)[] t, const(FSChar)* fromz, const(FSChar)* toz, Flag!"canCopy" canCopy) @trusted
```
wilzbach: std/file.d(1032:118)[warn]: Line is longer than 120 characters

std/file.d
jmdavis: IMHO, it's a mistake to add any new code that supports such implicit conversions, and either way, `isConvertibleToString` should not be used in any new code. It's going to be deprecated once Phobos doesn't use it anymore, because it's generally unsafe (though this particular case doesn't suffer from that problem).

WalterBright:
I don't see the value of adding this. It can already be done just by simple combination of existing functions in std.file?

marler8997:
Yes you can do the same thing with a rename and then catch the exception and fall back to copy. Closing

marler8997:
Yes you can do the same thing with a rename and then catch the exception and fall back to copy. Closing

wilzbach:
> I don't see the value of adding this. It can already be done just by simple combination of existing functions in std.file?

How?

The point was that rename doesn't work across file systems (which requires copying).
It wasn't added as a parameter due to:

>  It will also be more user-friendly to call move or moveFile than to call rename!(CanCopy.yes),

jmdavis:
> The point was that rename doesn't work across file systems (which requires copying).
> It wasn't added as a parameter due to:
>> It will also be more user-friendly to call move or moveFile than to call rename!(CanCopy.yes),

It also makes no sense to add a parameter to `copy`, because trying to move a directory via copying becomes a complete mess is a case where the application really should be in direct control over what happens (especially if anything fails during the copy). It can be debated as to whether having `moveFile` instead of using `rename`, catching the `FileException` on move failure, and then calling `copy` followed by `remove` is worth it, but I think that it's clear that changing `rename` to support copying would be a mistake given that it supports directories (and that's not even getting into the issues of `rename` then not being atomic).

CyberShadow:
What is the use case for this?

This function is useful if you are writing a generic file manager tool or a D reimplementation of coreutils, when you don't know where the source and destination might be. But if you are writing a program that knows what files it's operating with, then you probably also know where they are located on the filesystem, and already know whether you need to rename or copy+delete.

I can imagine that this could be useful in a situation where e.g. a program needs to move some data in or out of a cache, where the cache might or might not be on another filesystem, but this invariably brings up questions like what to do with directories, symlinks, timestamps, extended file attributes, etc. etc. that it's better to leave the user to address these concerns explicitly in their program.

jmdavis:
> What is the use case for this?

Honestly, this is what I'd want in most cases where I'd move a file. I rarely care whether the source and destination are on the same filesystem - especially when I'm using something like zfs, where you frequently end up with quite a few filesystems on your system, because datasets get used heavily, and those are treated like separate partitions with regards to moving. It matters a lot less if you're just using actual partitions, but if I'm moving a file, whether the source and destination are on the same filesystem is almost always an implementation detail. So, in practice, I find the fact that `rename` doesn't act like the  unix `mv` utility to be incredibly annoying - to the point that it's not infrequent that I just use std.process to call the unix utility when copying or moving files around, because they're user-friendly in a way that std.file really isn't.

If we want to argue that it's simple enough to do create a `moveFile` wrapper around `rename` yourself that it's not worth putting in Phobos, then fine, but personally, I would find a function like `moveFile` to be extremely useful in comparison to `rename`. I've found that using `rename` is a bug in the making, because the application rarely has control over what filesystems files are on unless it's just playing in a directory specific to it and where it controls all of the files.

marler8997:
Here's on example that exists in D's tools.

If you have 2 drives on your windows box (i.e. `C:\` and `D:\`) then you can cause `rdmd` to fail.

```
C:\> rdmd -oftest test.d
D:\> rdmd -oftest test.d
```

One of these will fail because it will cause `rdmd` to copy the temporary output file from it's cache to the users current directoy, which will be another drive in one of these cases.  This is because `rdmd` uses `rename` without catching the exception and falling back to copy.  Using `moveFile` instead would have prevented this bug and provided a simple API for developers to use instead of having to do this all the time:
```D
try { rename(from, to); } catch (FileException) { copy(from, to); remove(from); }
```

wilzbach:
(reopening as there's clearly not a common consensus)

CyberShadow:
> It matters a lot less if you're just using actual partitions, but if I'm moving a file, whether the source and destination are on the same filesystem is almost always an implementation detail.

Exactly, so the implementation of your code should follow the "implementation" of your filesystem layout.

> So, in practice, I find the fact that rename doesn't act like the unix mv utility to be incredibly annoying - to the point that it's not infrequent that I just use std.process to call the unix utility when copying or moving files around,

This is bad advice. Command-line tools do a lot of magic, DWIM stuff that are unsuitable for writing correct and reliable programs.

> because they're user-friendly in a way that std.file really isn't.

`std.file`, or any part of Phobos, should NOT be "user-friendly" - it should be an API that does exactly what the documentation says it does, and nothing more.

WalterBright:
The try/catch code is 6 lines. But this PR is 103 lines. I suggest putting in the 6 lines as an example in the documentation, it is not worth its own function.

Phobos should provide orthogonal building blocks that can be combined by the user. There should be n functions, not n*n (an enumeration of all the combinations). std.algorithm is based on this idea.

marler8997:
@WalterBright please stop using "lines of code" as a reason not to integrate features. It's one of the worst arguments you can give.  You can tell that most of the lines modified in this PR is testing/documentation.  Using this argument is a "cop out" that people give when they want to dismiss an idea without providing good reason.  By using this argument, you are encouraging people not to write tests... not to fully document their changes...not to include examples...and encouraging them to try to be "clever" in their implementation by reducing line count even if it means the code is more difficult to understand/maintain.  This is a bad precedent to set and an example of poor leadership in my opinion (something that I think the D is severely lacking in).

In any case, I think the current state of phobos violates the principle of a good interface as defined by Scott Meyers, namely, easy to use correctly, and hard to use incorrectly.  In most cases programs want this functionality:
```D
try {
    rename(from, to);
} catch (FileException) {
    copy(from, to);
    remove(from);
}
```
but they use this one:
```D
rename(from, to);
```

I think it's obvious why.  That being said, it's not my code so I'll leave it up to the maintainers to decide what interface is best.

jmdavis:
> Exactly, so the implementation of your code should follow the "implementation" of your filesystem layout.

You're assuming that the program you're running has anything to do with the filesystem layout. The same program could be run on multiple computers where each has a different filesystem layout, and it's not necessarily the case that the program knows which files it will be operating on ahead of time, let alone where they are.

> This is bad advice. Command-line tools do a lot of magic, DWIM stuff that are unsuitable for writing correct and reliable programs.

The reality of the matter is that the command-line tools for stuff like copying and moving usually do what I need, whereas std.file often doesn't. `rename` is a prime example. I almost never care whether the source and target are on the same filesystem, and I frequently have no way of knowing whether they are. `mv` deals with that quite nicely. `rename` does not. If we want to argue that the try-catch required to work around the fact that `rename` is not actually guaranteed to be able to move the file is simple enough that adding `moveFile` isn't worth it, then fine. But `rename`'s behavior has caused nothing but problems for me.

I have no problem with `rename` existing, and I'm not arguing that it should not be possible to have low-level control or that it never makes sense to be that strict about what gets done when moving the file, but the current result is that std.file does a very poor job of solving the use cases that I typically have, and if I use `rename`, I almost always have to fix it later. Honestly, what I really need to do is put together a library that wraps std.file and actually works more like `cp` and `mv`, because for the most part, those have the semantics that I need. If you don't like those semantics and don't want to use functions that act like that, that's fine, but there is no way that I'm going to agree that the way that `rename` works is actually desirable behavior in the average use case or that the way that `mv` works is undesirable. I think that it's clear that we fundamentally disagree on what constitutes a good move function.

> std.file, or any part of Phobos, should NOT be "user-friendly" - it should be an API that does exactly what the documentation says it does, and nothing more.

Having a function do exactly what it's documented to do and having it be user-friendly do not inherently conflict.

marler8997:
This isn't something I want to spend time fighting for...closing

PR #6418, opened 2018-04-04
Added examples to std.datetime.systime

JackStouffer:


std/datetime/systime.d
```diff
@@ -31,6 +31,38 @@ $(TR $(TD Conversion) $(TD
...
+    assert(st == SysTime(DateTime(2018, 1, 1, 10, 30, 0), UTC()));
+}
+
+/// Make a specific point in time in the New York timezone
+@safe unittest
+{
+    import std.datetime.timezone : SimpleTimeZone;
+
+    auto ny = SysTime(
+        DateTime(2018, 1, 1, 10, 30, 0),
```
n8sh: Maybe add a comment saying what time this corresponds to. It's kind of belaboring the obvious, but I think it wouldn't hurt.

std/datetime/systime.d
```diff
@@ -31,6 +31,38 @@ $(TR $(TD Conversion) $(TD
...
+        new immutable SimpleTimeZone(-5.hours, "America/New_York")
+    );
+
+    // ISO standard time strings
+    assert(st.toISOString() == "20180101T103000-05:00");
+    assert(st.toISOExtString() == "2018-01-01T10:30:00-05:00");
```
wilzbach: std/datetime/systime.d(62): Error: undefined identifier `st`, did you mean import `std`?
std/datetime/systime.d(63): Error: undefined identifier `st`, did you mean import `std`?

DMD is kinda dumb here - it should be `ny`

std/datetime/systime.d
```diff
@@ -31,6 +31,41 @@ $(TR $(TD Conversion) $(TD
...
+/// Make a specific point in time in the New York timezone
+@safe unittest
+{
+    import core.time : hours;
+    import std.datetime.date : DateTime;
+    import std.datetime.timezone : SimpleTimeZone;
+
+    auto ny = SysTime(
+        DateTime(2018, 1, 1, 10, 30, 0),
+        new immutable SimpleTimeZone(-5.hours, "America/New_York")
```
jmdavis: Honestly, I'm against calling a timezone "America/New_York" when it's clearly not actually "America/New_York". For the actual "America/New_York", you'd need to use `PosixTimeZone`. I guess that I'll just tweak it later though.

Either way, @wilzbach, please don't merge PRs just a few hours after they've been put up unless they're truly trivial or need to be pushed through ASAP to fix the autotester or something similar. There isn't anything seriously wrong with this PR, but when you merge stuff that fast, not much of anyone else has a chance to look at it. We don't want PRs languishing, but pushing them through as fast as possible is just going to cause other problems, and in the case of std.datetime, I would very much like to be looking at PRs before they're merged. If I'm not getting around to it in a timely manner, that's one thing, but it looks like you added the auto-merge label for this one less than an hour after it was posted.

wilzbach: I didn't - I realized this myself:

> wilzbach removed the auto-merge  label about 7 hours ago

(though to be fair for other modules I did merge "new example" PRs quickly before, but this is more due to our PR queue exploding + sadly not many people interested in reviewing such PRs + it takes them ~ 1 month to go "live" on dlang.org anyhow)

jmdavis: It's been complained about before when PRs are merged in under a day, because then most folks don't have a chance to review the PR. And pushing PRs through as quickly as possible risks letting stuff through that shouldn't get through, which is exactly the sort of thing that led to Andrei insisting on reviewing all PRs that added public symbols. At the same time, we do have PRs that sit around for too long, and that causes problems, and they do need to be reviewed and dealt with. So, there is a balance to be struck, but pushing stuff through just to get it out of the way or because we can always change it later is just risking problems, and waiting a day or so before pulling the trigger isn't going to make the PR queue balloon.

With regards to std.datetime in particular, I'm very interested in reviewing any changes before they go through. Jack Stouffer has been doing good work with making some improvements to it, but I still want to have the chance to see stuff before it gets merged in. If a PR for std.datetime is sitting around for several days without my saying anything, then it's not unreasonable to push it through if others have reviewed it and think that it looks good, but this was merged mere hours after it was posted.

PR #6419, opened 2018-04-04
[trivial] Fix EndianSwapper to work with qualified types

tchaloupka:
Found by accident, one should be able to call nativeToBigEndian with const or immutable floating points too.

I'm not sure if this should be handled at `FloatingPointTypeOf` template instead so I've chosen less impacting way.

### DIFF
--- a/std/bitmanip.d
+++ b/std/bitmanip.d
@@ -2793,9 +2793,9 @@ if (canSwapEndianness!T)
     Unqual!T value;
     ubyte[T.sizeof] array;
 
-    static if (is(FloatingPointTypeOf!T == float))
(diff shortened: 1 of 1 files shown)


std/bitmanip.d
```diff
@@ -2256,6 +2256,10 @@ unittest
     double d = 123.45;
     ubyte[8] swappedD = nativeToBigEndian(d);
     assert(d == bigEndianToNative!double(swappedD));
+
+    const double cd = 123.45;
+    ubyte[8] swappedCD = nativeToBigEndian(cd);
+    assert(cd == bigEndianToNative!double(swappedCD));
```
wilzbach: Add a test for `float` too, s.t. regressions can be prevented?

PR #6421, opened 2018-04-05
std.algorithm.searching: Implement `none` as shorthand for `!any`.

FeepingCreature:
When `any` is invoked with a long condition in ufcs form, it can be a bit hard to read the logic, ie. `if (!someArray.any!(element => someConditionTest(element) && element > 2))` - there's a lot of space between that negating ! and the actual test. In comparison, `if (someArray.none!(element => someConditionTest(element) && element > 2))` is read more easily and quickly.

### DIFF
--- a/std/algorithm/searching.d
+++ b/std/algorithm/searching.d
@@ -155,9 +155,8 @@ are true.
 
 /++
 Checks if $(I _any) of the elements verifies `pred`.
-`!any` can be used to verify that $(I none) of the elements verify
-`pred`.
 This is sometimes called `exists` in other languages.
+`! any!pred` is equivalent to `none!pred`.
  +/
 template any(alias pred = "a")
 {
@@ -179,19 +178,15 @@ template any(alias pred = "a")
     import std.ascii : isWhite;
     assert( all!(any!isWhite)(["a a", "b b"]));
     assert(!any!(all!isWhite)(["a a", "b b"]));
+    assert(none!(all!isWhite)(["a a", "b b"]));
 }
 
 /++
 `any` can also be used without a predicate, if its items can be
-evaluated to true or false in a conditional statement. `!any` can be a
-convenient way to quickly test that $(I none) of the elements of a range
-evaluate to true.
+evaluated to true or false in a conditional statement.
  +/
 @safe unittest
 {
-    int[3] vals1 = [0, 0, 0];
-    assert(!any(vals1[])); //none of vals1 evaluate to true
-
     int[3] vals2 = [2, 0, 2];
     assert( any(vals2[]));
     assert(!all(vals2[]));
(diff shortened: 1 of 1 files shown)


std/algorithm/searching.d
```diff
@@ -207,6 +202,38 @@ evaluate to true.
     assert(any!"a == 2"(a));
 }
 
+/++
+Checks if $(I _none) of the elements verify `pred`. This is a convenience wrapper for `!any!pred`.
```
wilzbach: Auto-highlighting has been removed recently - `_`  is no longer necessary :)
https://github.com/dlang/dlang.org/pull/2307

std/algorithm/searching.d
```diff
@@ -207,6 +202,38 @@ evaluate to true.
...
+ +/
+template none(alias pred = "a")
+{
+    /++
+    Returns `true` if and only if $(I _no) value `v` found in the
```
wilzbach: Auto-highlighting has been removed recently - `_`  is no longer necessary :)
https://github.com/dlang/dlang.org/pull/2307

std/algorithm/searching.d
```diff
@@ -155,9 +155,8 @@ are true.
 
 /++
 Checks if $(I _any) of the elements verifies `pred`.
-`!any` can be used to verify that $(I none) of the elements verify
-`pred`.
 This is sometimes called `exists` in other languages.
+`! any!pred` is equivalent to `none!pred`.
```
wilzbach: `See_Also: $(LREF all), $(LREF none)` ?

JackStouffer:
> just compose using lambdas etc

Indeed, in your own program you can do

```d
alias none = x => !any(x);

void main()
{
    auto a = [0, 0, 0];
    assert(a.none);
}
```

Closing due to rejection.

FeepingCreature:
Seems a bit silly to say in the docs "this is the same as `none`" and then not just providing `none`. But eh, it's your library.

Though of course, both `any` and `all` are just trivial wrappers around `find`.

andralex:
>Seems a bit silly to say in the docs "this is the same as none" and then not just providing none.

The relevant quote is: [!any can be used to verify that none of the elements verify pred](https://dlang.org/phobos/std_algorithm_searching.html#any). That seems a reasonable thing to say. Clearly one could go on and define none as its own primitive - e.g. Ruby and Perl have `unless` as the negation of `if`.

>Though of course, both any and all are just trivial wrappers around find.

Roger. There's been some decision making there, and one good argument was that `range.find!(pred).empty` was frequently needed and somewhat awkward.

FeepingCreature:
In my opinion, the analogous reason for `none` is the distance between the verb and the negation if you're using UFCS, since you can't do expression.(!any). :) So either the negation ends up all the way at the start of the ufcs chain, or you have to write the awkward `.all(!pred)`.

WebFreak001:
boo! `none` is very common and when chaining calls and long lines it get's really confusing quickly

PR #6423, opened 2018-04-05
Added params and returns to std.array functions

JackStouffer:
Can't enable the dscanner check because of https://github.com/dlang-community/D-Scanner/issues/601

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -451,14 +451,16 @@ if (isInputRange!Range)
 /**
 Construct a range iterating over an associative array by key/value tuples.
 
-Params: aa = The associative array to iterate over.
+Params:
+    aa = The associative array to iterate over.
 
 Returns: A $(REF_ALTTEXT forward range, isForwardRange, std,_range,primitives)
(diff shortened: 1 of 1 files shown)


std/array.d
```diff
@@ -451,14 +451,16 @@ if (isInputRange!Range)
...
+Params:
+    aa = The associative array to iterate over.
 
 Returns: A $(REF_ALTTEXT forward range, isForwardRange, std,_range,primitives)
 of Tuple's of key and value pairs from the given associative array. The members
 of each pair can be accessed by name (`.key` and `.value`). or by integer
 index (0 and 1 respectively).
 */
-auto byPair(AA : Value[Key], Value, Key)(AA aa)
+auto byPair(AA)(AA aa)
```
JackStouffer: The previous function signature required `Value, Key` to be in the params section, so I simplified it

n8sh: Actually -- it doesn't matter. Strike the previous message.

PR #6424, opened 2018-04-05
Fix Issue 18327 - std.random.XorshiftEngine is parameterized by UIntType but only works with uint

n8sh:
Backported from libmir/mir-random#80.


std/random.d
```diff
@@ -1443,10 +1566,19 @@ alias Xorshift    = Xorshift128;                            /// ditto
         [5783321UL, 393427209, 1947109840, 565829276, 1006220149, 971147905,
         1436324242, 2800460115, 1484058076, 3823330032],
         [0UL, 246875399, 3690007200, 1264581005, 3906711041, 1866187943, 2481925219,
```
n8sh: Pre-existing unittests test the output of `XorshiftEngine` and verify the sequence produced has not changed.

wilzbach:
> I didn't touch any of those so I'll just come back later and see if dscanner has stopped failing.

Looks like a bug in Dscanner. I will hopefully find some time on the weekend to look into that one, but that shouldn't stop other reviewers from looking over this PR...

std/random.d
```diff
@@ -1349,39 +1482,17 @@ if (isUnsigned!UIntType)
     {
         return this;
     }
-
-
-    /**
-     * Compares against $(D_PARAM rhs) for equality.
-     */
-    bool opEquals(ref const XorshiftEngine rhs) const @safe pure nothrow @nogc
```
n8sh: This struct's custom `opEquals` appeared to be the same as default struct equality so I removed it.

std/random.d
```diff
@@ -1349,39 +1482,17 @@ if (isUnsigned!UIntType)
...
-    {
-        return seeds_ == rhs.seeds_;
-    }
-
-
-  private:
-    static void sanitizeSeeds(ref UIntType[size] seeds) @safe pure nothrow @nogc
```
n8sh: Deleted because it was incorporated into `seed` like in @9il's reworked version of XorshiftGenerator.

n8sh:
Found a workaround for the DScanner bug.

std/random.d
```diff
@@ -1183,86 +1186,230 @@ alias Mt19937_64 = MersenneTwisterEngine!(ulong, 64, 312, 156, 31,
...
+            cast(UIntType) 362_436_069,
+            cast(UIntType) 521_288_629,
+            cast(UIntType)  88_675_123,
+            cast(UIntType)   5_783_321][0 .. N];
+    }
     else
     {
-        static assert(false, "Phobos Error: Xorshift has no instantiation rule for "
-                             ~ to!string(bits) ~ " bits.");
+        UIntType[N] _state = ()
```
BBasile: This whole delegate is not covered. Maybe some tests for UIntType as `ulong` must be added.

n8sh: This PR tests Xorshift64_64 and Xorshift128_64 with UIntType = ulong. Could the coverage report be confused because this delegate is only run at compile time?

BBasile: Sorry i missed the version(unittest) declarations and how they're used. I don't know if the comment about cov decreasing was relevant then.

thewilsonator:
@n8sh please rebase.

PR #6425, opened 2018-04-06
zlib: remove switch from in front of super()

WalterBright:
needed for https://github.com/dlang/dmd/pull/8143

### DIFF
--- a/std/zlib.d
+++ b/std/zlib.d
@@ -75,9 +75,9 @@ enum
 
 class ZlibException : Exception
 {
-    this(int errnum)
-    {   string msg;
-
+    private static string getmsg(int errnum) nothrow @nogc pure @safe
+    {
+        string msg;
         switch (errnum)
         {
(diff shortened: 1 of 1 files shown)


wilzbach:
@WalterBright the segfault comes from Phobos -> https://issues.dlang.org/show_bug.cgi?id=18667
It's not blocked, it's a spurious segfault that appears on every ~10th run and a restart makes it go away.
Yeah I know it's not perfect, but apparently no one had had time to dive into 18667.

std/zlib.d
```diff
@@ -75,9 +75,9 @@ enum
 
 class ZlibException : Exception
 {
-    this(int errnum)
-    {   string msg;
-
+    private static string getmsg(int errnum) nothrow @nogc pure
```
wilzbach: Could be `@safe` too (though the entire module isn't @safe yet anyhow)

WalterBright: I agree. Done.

PR #6429, opened 2018-04-06
Work On Issue 18693 - Remove rndtonl from std.math

JackStouffer:
This function 

1. Results in link errors if you're not using DM's C runtime
2. Is non-standard
3. Is covered by `round` and friends

Side note: all of these platform specific math functions which share the C names feel a lot like PHP.  This is not how these would be designed if std.math was to be written today, so any instance of these that we can remove/redesign, the better.

### DIFF
--- a/.dscanner.ini
+++ b/.dscanner.ini
@@ -252,7 +252,6 @@ has_public_example="-etc.c.curl,\
 -std.internal.scopebuffer,\
 -std.internal.test.dummyrange,\
 -std.json,\
--std.math,\
 -std.mathspecial,\
 -std.mmfile,\
 -std.net.curl,\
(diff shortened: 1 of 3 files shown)


JackStouffer:
Looks like ddox is mis-parsing the unittest block and including the next function signature

wilzbach:
> Both at the time of deprecation and removal, a changelog entry must be made. This
> changelog entry should have a short motivation for the deprecation (or removal)
> and should describe which steps can be taken by the user to upgrade their codebase.

(from https://github.com/dlang/DIPs/pull/108)

wilzbach:
@andralex could we quickly get your approval on this one? Thanks!

JackStouffer:
`assert(rndtonl(1.0) is -real.nan);`

does this function even work at all

JackStouffer:
>does this function even work at all

Nope.

PR #4733, opened 2016-08-15
Make GCAllocator const -> makeArray pure

wilzbach:
This is a follow-up to the recent `@safe` improvement to `makeArray` of `std.allocator`.

The biggest problem of the `GCAllocator` is it's sharedness which makes it impossible to write `pure` code even though the GCAllocator is used by default. Hence I propose to make the `instance`  `const` and thus a lot of methods `pure`.

PR #6432, opened 2018-04-07
Make GCAllocator usable in pure functions

nordlow:
Qualify members of `GCAllocator` as `const` to make it usable in pure contexts.

Because `new` can be used in pure contexts I see no reason why `GCAllocator` shouldn't be that aswell.

### DIFF
--- a/std/experimental/allocator/gc_allocator.d
+++ b/std/experimental/allocator/gc_allocator.d
@@ -27,7 +27,7 @@ struct GCAllocator
     deallocate) and `reallocate` methods are `@system` because they may
     move memory around, leaving dangling pointers in user code.
     */
-    pure nothrow @trusted void[] allocate(size_t bytes) shared
+    pure nothrow @trusted void[] allocate(size_t bytes) shared const
     {
         if (!bytes) return null;
         auto p = GC.malloc(bytes);
@@ -35,7 +35,7 @@ struct GCAllocator
     }
 
     /// Ditto
-    pure nothrow @trusted bool expand(ref void[] b, size_t delta) shared
+    pure nothrow @trusted bool expand(ref void[] b, size_t delta) shared const
     {
         if (delta == 0) return true;
         if (b is null) return false;
@@ -58,7 +58,7 @@ struct GCAllocator
     }
 
     /// Ditto
-    pure nothrow @system bool reallocate(ref void[] b, size_t newSize) shared
+    pure nothrow @system bool reallocate(ref void[] b, size_t newSize) shared const
     {
         import core.exception : OutOfMemoryError;
         try
@@ -76,7 +76,7 @@ struct GCAllocator
 
     /// Ditto
     pure nothrow @trusted @nogc
-    Ternary resolveInternalPointer(const void* p, ref void[] result) shared
+    Ternary resolveInternalPointer(const void* p, ref void[] result) shared const
     {
         auto r = GC.addrOf(cast(void*) p);
         if (!r) return Ternary.no;
@@ -86,7 +86,7 @@ struct GCAllocator
 
     /// Ditto
     pure nothrow @system @nogc
-    bool deallocate(void[] b) shared
+    bool deallocate(void[] b) shared const
     {
         GC.free(b.ptr);
         return true;
@@ -94,7 +94,7 @@ struct GCAllocator
 
     /// Ditto
     pure nothrow @safe @nogc
-    size_t goodAllocSize(size_t n) shared
+    size_t goodAllocSize(size_t n) shared const
     {
         if (n == 0)
             return 0;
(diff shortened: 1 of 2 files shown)


std/experimental/allocator/gc_allocator.d
```diff
@@ -1,6 +1,6 @@
 // Written in the D programming language.
 /**
-D's built-in garbage-collected allocator.
+   D's built-in garbage-collected allocator.
```
wilzbach: Unrelated

nordlow: Fixed.

nordlow:
Had to update a `static assert` aswell.

std/experimental/allocator/gc_allocator.d
```diff
@@ -117,31 +117,31 @@ struct GCAllocator
     are `shared`.
     */
 
-    static shared GCAllocator instance;
+    static shared const GCAllocator instance;
```
n8sh: Why not immutable? Does it matter one way or the other?

nordlow: No, AFAICT. An immutable instance is automatically shared, but since GCAllocator contains no data I don't think it matters in this case.

n8sh: That's bizarre.

std/experimental/allocator/typed.d
```diff
@@ -22,7 +22,7 @@ import std.traits : isPointer, hasElaborateDestructor;
 import std.typecons : Flag, Yes, No;
 
 /**
-Allocation-related flags dictated by type characteristics. `TypedAllocator`
+   Allocation-related flags dictated by type characteristics. `TypedAllocator`
```
n8sh: No reason for indent

nordlow: Fixed.

std/experimental/allocator/gc_allocator.d
nordlow: @andralex @edi33416 ?

nordlow:
Thanks.

wilzbach:
Sorry for being away over the weekend, but I thought we couldn't do this:

> The current code models that reality. Yes, we did make a concession for new to allow pure functions to allocate, but I don't want to use that as precedent to essentially model things that don't exist.

> I suggest we leave the low-level allocators alone. Then use a cast to force makeArray be pure.

https://github.com/dlang/phobos/pull/4733#issuecomment-243412290

(I tried the same ~ two years ago)

nordlow:
Thanks, anyway.

wilzbach:
In case I wasn't clear: I was trying to say that except @andralex changed his mind, I think we need to revert this :/

andralex:
After working on allocators and collection for a good while, I got to the conclusion the allocation funcs must be pure, so let's keep this in.

nordlow:
Great, `MmapAllocator` will be next!

std/experimental/allocator/gc_allocator.d
edi33416: The issue seems to be that when you have `static immutable GCAllocator instance` the following
`is(typeof(GCAllocator.instance) == shared)` is `false`, which is odd. 

If you have a look at [Segregator](https://github.com/dlang/phobos/blob/master/std/experimental/allocator/building_blocks/segregator.d#L260) it compares the small and large allocators against `shared` in order to decide how to `mixin` the `Impl` template struct.

With `immutable`, the is expression is `false` and thus the `shared` attribute is not applied.
I don't know why it's false; maybe a compiler bug? @andralex

nordlow: I had the same problems in the static assert. I agree, it looks like a compiler bug.

edi33416: @nordlow @n8sh so this isn't a compiler bug, it was just a misunderstanding.

The is expression `is(T == shared)` checks to see if `T` has `shared` among his type qualifiers, so it is ok to return `false` for `immutable`. `immutable` isn't `shared`, but it is convertible to it.

So the check should be `is(T : shared(T))`. There are a couple of allocators that are affected by this incomplete check; from the top of my head I know of `AffixAllocator` and `Segregator`, though there may be others.

I propose we add a `is(T == immutable)` to those, instead of using `:`, as this is, imho, simpler and clearer. @nordlow do you want to have a go at it?

nordlow: Sure. Ok to add it [this](https://github.com/dlang/phobos/pull/6440)?

edi33416: Probably a new PR would be better

nordlow: Which allocators should we update?

nordlow: Ok, I'll try that.

nordlow: @edi33416, can you have a go at this? My time is currently very spare.

edi33416: @nordlow I'll have at it after DConf

nordlow: Thanks.

PR #6435, opened 2018-04-08
Make references to input ranges a link to isInputRange

JackStouffer:
Unfortunately we can't control when people are introduced to the concept of a range. Ideally they read Ali's chapter on it or go through the tour before they start going through the docs. But, a lot of users will have these terms thrown at them without introduction (from their perspective).

This PR makes mentions of input ranges into links to `isInputRange`. Also, the major range types now have links to the header of std.range which contains lots of range tutorials.

### DIFF
--- a/std/algorithm/setops.d
+++ b/std/algorithm/setops.d
@@ -930,7 +930,8 @@ alias nWayUnion = multiwayMerge;
 alias NWayUnion = MultiwayMerge;
 
 /**
-Computes the union of multiple ranges. The input ranges are passed
+Computes the union of multiple ranges. The
+$(REF_ALTTEXT input ranges, isInputRange, std,range,primitives) are passed
 as a range of ranges and each is assumed to be sorted by $(D
 less). Computation is done lazily, one union element at a time.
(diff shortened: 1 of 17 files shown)


std/algorithm/setops.d
```diff
@@ -1117,8 +1118,9 @@ SetDifference!(less, R1, R2) setDifference(alias less = "a < b", R1, R2)
 }
 
 /**
-Lazily computes the intersection of two or more input ranges $(D
-ranges). The ranges are assumed to be sorted by `less`. The element
+Lazily computes the intersection of two or more
+$(REF_ALTTEXT input ranges, isInputRange, std,range,primitives)
+`ranges`. The ranges are assumed to be sorted by `less`. The element
```
n8sh: input ranges ranges

JackStouffer: That was in the original, and I assume it's referring to the parameter `ranges` of the function, hence the backticks.

JackStouffer: The parameter for the ctor is `input` which is a seperate ddoc. The parameter for this ddoc block is on line 1231 and is `Rs ranges`

PR #6437, opened 2018-04-09
Added toBytes property to std.bigint.BigInt

JonathanWilbur:
Per issue 13804, I created a `toBytes` property for `std.bigint.BigInt`. This generates an arbitrarily-long array of unsigned bytes that represents the signed, native-endian binary representation of the `BigInt`.

https://issues.dlang.org/show_bug.cgi?id=13804


JonathanWilbur:
That punycode commit you see there was from a previous pull request. I removed it in a subsequent commit.

JonathanWilbur:
Oh, also, the main rationale--the one I suspect everybody has for wanting this property--is the ability to convert large integers to arrays of bytes for ASN.1 encoding in RSA data structures. That's why I made it, FWIW.

Biotronic:
Have you looked at `BigUInt` (in std/internal/math/biguintcore.d)? Almost every function in `BigInt` is just a call to a function of `BigUint`, and I suggest this should be as well.

`BigUint` would then have a `toBytes` function that doesn't deal with signs, and `BigInt`'s `toBytes` implementation would take that value and do 2's complement if necessary. `BigUint.data` is simply a `uint[]` and a simple cast should do most of what this PR does.

std/bigint.d
```diff
@@ -1025,6 +1025,220 @@ public:
...
+    {
+        assert(this.uintLength > 0u);
+    }
+    out (ret)
+    {
+        assert(ret.length > 0u);
+    }
+    body
+    {
+        ubyte[] ret;
```
BBasile: Use an `appender!(ubyte[])` i think

std/bigint.d
```diff
@@ -1025,6 +1025,220 @@ public:
...
     }
 
+    /*
+        Increments an arbitrary-length integer, represented as an array of
+        bytes, using the native endianness.
+
+        This is just used by `toBytes` to convert unsigned bytes of the
+        BigInt to two's complement form.
+    */
+    private static void incrementBytes (ref ubyte[] value) nothrow pure @safe
```
BBasile: Should take an output range of ubyte (confer with other comment proposing to use an Appender)

std/bigint.d
```diff
@@ -1025,6 +1025,220 @@ public:
...
+        if (value.length == 0u)
+        {
+            value = [ 0x01u ];
+            return;
+        }
+
+        version (BigEndian)
+        {
+            // This loop adds one to an arbitrary length array of bytes.
+            for (size_t i = (value.length - 1); i < 0u; i--)
```
BBasile: style: there's no ambiguity with precedence so value.length-1 can live without parens.

std/bigint.d
```diff
@@ -1025,6 +1025,220 @@ public:
...
+                if (value[i] != 0xFFu)
+                {
+                    value[i]++;
+                    break;
+                }
+
+                value[i] = 0x00u;
+
+                // If the array of bytes is maxed out, append a next byte set to one.
+                if (i == (value.length - 1))
```
BBasile: style: there's no ambiguity with precedence so value.length-1 can live without parens.

std/bigint.d
```diff
@@ -1025,6 +1025,220 @@ public:
...
+    ubyte[] toBytes() const @system @property nothrow
+    in
+    {
+        assert(this.uintLength > 0u);
+    }
+    out (ret)
+    {
+        assert(ret.length > 0u);
+    }
+    body
```
BBasile: use `do` instead of `body`. The latter will be deprecated.

std/bigint.d
```diff
@@ -1025,6 +1025,220 @@ public:
...
+        }
+
+        if (this.sign) BigInt.incrementBytes(ret);
+
+        version (BigEndian)
+        {
+            size_t startOfNonPadding = 0u;
+            if (this >= 0)
+            {
+                for (size_t i = 0u; i < (ret.length - 1); i++)
```
BBasile: style: there's no ambiguity with precedence so ret.length-1 can live without parens.

std/bigint.d
```diff
@@ -1025,6 +1025,220 @@ public:
...
+                {
+                    if (ret[i] != 0x00u) break;
+                    if (!(ret[i+1] & 0x80u)) startOfNonPadding++;
+                }
+                ret = ret[startOfNonPadding .. $];
+                if (ret[$-1] & 0x80u) ret = (0x00u ~ ret);
+            }
+            else
+            {
+                for (size_t i = 0u; i < (ret.length - 1); i++)
```
BBasile: style: there's no ambiguity with precedence so ret.length-1 can live without parens.

BBasile: by the way `foreach(i; 0.. N)` tends to be used more nowadays in D.

std/bigint.d
```diff
@@ -1025,6 +1025,220 @@ public:
...
+        assert(this.uintLength > 0u);
+    }
+    out (ret)
+    {
+        assert(ret.length > 0u);
+    }
+    body
+    {
+        ubyte[] ret;
+        ret.length = (this.uintLength << 2); // Multiply by 4
```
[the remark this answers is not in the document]
JonathanWilbur: Neat! I didn't know the compiler did this!

std/bigint.d
JonathanWilbur: Well, I subsequently set the length of the array to the length that will be needed, then dump the `uint`s in that array. How would `Appender` improve that?

std/bigint.d
JonathanWilbur: Can you clarify? What should the signature of this method be?

BBasile: a template constraint taking only an output range of ubyte

std/bigint.d
BBasile: okay, forget the appender thing.

std/bigint.d
Biotronic: So something like `private static void incrementBytes(Range)(Range result, ubyte[] value) if (isOutputRange!Range)`? You still need to pass the original data in there somehow.

BBasile: Ok. Let's forget the outrange thing in this review and concentrate about the style.

std/bigint.d
JonathanWilbur: That's a good tip; definitely a cleaner way to do read-only loops. However, that won't work when you have to set the values in an array, because `i` in your example would be passed in by value, so setting `i` would not affect the original array from which the slice was taken. (And indeed, I have to do that in my code, particularly when converting an array of bytes to two's complement form.)

I don't know how slicing works (base+bound vs. big dumb copy vs. copy-on-write), but if it is the latter two of those three, then changes to the slice will not modify the original as well. But I'm not qualified to assert anything else. 😝

std/bigint.d
JonathanWilbur: Good to know. I must have missed the memo.

std/bigint.d
```diff
@@ -1025,6 +1025,193 @@ public:
...
+        if (value.length == 0u)
+        {
+            value = [ 0x01u ];
+            return;
+        }
+
+        version (BigEndian)
+        {
+            // This loop adds one to an arbitrary length array of bytes.
+            for (size_t i = value.length - 1; i < 0u; i--)
```
wilzbach: This would be `foreach_reverse (0u ... value.length - 1)` or `foreach_reverse (e; size_t(0) .. 10)` if you want `size_t` - though I think in this case even `int` is fine.
Also note that `i < 0` can never be true and that IIRC `BigEndian` isn't tested on any CIs.

std/bigint.d
```diff
@@ -1025,6 +1025,193 @@ public:
...
+                {
+                    value = (0x01u ~ value);
+                    return;
+                }
+            }
+        }
+        else version (LittleEndian)
+        {
+            // This loop adds one to an arbitrary length array of bytes.
+            for (size_t i = 0u; i < value.length; i++)
```
wilzbach: - `foreach (e; 0 .. value.length)`
- `foreach (e; size_t(0) .. value.length)`
- `foreach (i, ref e; value)` <- the best

std/bigint.d
```diff
@@ -1025,6 +1025,193 @@ public:
...
+    }
+
+    /**
+        Converts the `BigInt` to a sequence of bytes that represents the
+        native-endian representation of that number.
+    */
+    ubyte[] toBytes() const pure @system @property nothrow
+    in
+    {
+        assert(this.uintLength > 0u);
```
wilzbach: Needs a message

std/bigint.d
```diff
@@ -1025,6 +1025,193 @@ public:
...
+    }
+    out (ret)
+    {
+        assert(ret.length > 0u);
```
wilzbach: Needs a message

std/bigint.d
```diff
@@ -1025,6 +1025,193 @@ public:
...
+        assert(b == [ 0x00u, 0x01u ]);
+
+        ubyte[] c = [ 0xFDu, 0xFFu, 0xFFu ];
+        incrementBytes(c);
+        assert(c == [ 0xFEu, 0xFFu, 0xFFu ]);
+    }
+
+    /**
+        Converts the `BigInt` to a sequence of bytes that represents the
+        native-endian representation of that number.
```
wilzbach: Needs "Returns:"

std/bigint.d
```diff
@@ -1025,6 +1025,193 @@ public:
...
+    out (ret)
+    {
+        assert(ret.length > 0u);
+    }
+    do
+    {
+        ubyte[] ret = this.data.toBytes;
+        if (this.sign)
+        {
+            for (size_t u = 0u; u < ret.length; u++)
```
wilzbach: foreach

std/bigint.d
```diff
@@ -1025,6 +1025,193 @@ public:
...
+        bytes, using the native endianness.
+
+        This is just used by `toBytes` to convert unsigned bytes of the
+        BigInt to two's complement form.
+    */
+    private static void incrementBytes (ref ubyte[] value) nothrow pure @safe
+    {
+        if (value.length == 0u)
+        {
+            value = [ 0x01u ];
```
wilzbach: Uncovered

JonathanWilbur:
@wilzbach , I do not have a machine for testing on `BigEndian`. Actually, I assumed that one of the hosts in the D Auto-Tester fleet would be big endian...

JonathanWilbur:
Also, I don't think that Circle CI failure is my fault: it says that Dscanner failed. I didn't change anything about Dscanner.

jmdavis:
> Actually, I assumed that one of the hosts in the D Auto-Tester fleet would be big endian...

dmd only supports x86 and x86_64, so it doesn't support Big Endian at all, much as the language does. To do anything with Big Endian and D, you need either ldc or gdc and a Big Endian machine that they support. And since they're maintained separately from dmd, we can't test PRs against them. So, unfortunately, anything we've done with Big Endian in Phobos is done with the hope that it works and that when the changes finally get merged into gdc or ldc, if they're broken for Big Endian, someone will report it. Fortunately, not much code in Phobos cares about endianness.

wilzbach:
> Also, I don't think that Circle CI failure is my fault: it says that Dscanner failed. I didn't change anything about Dscanner.

Yep, not your fault. The GC is segfaulting spuriously, but so far no one knows why (or has cared to investigate).
The issue is: https://issues.dlang.org/show_bug.cgi?id=18720

JonathanWilbur:
So what is the next step? Does somebody else need to review this? Do I need to change something?

Biotronic:
I still hold that the correct location for most of this code is in std/internal/math/biguintcore.d. It should handle the unsigned stuff, and std/bigint.d should only do the 2's complement when that's necessary.

JonathanWilbur:
@Biotronic , I guess you missed it, but I already moved most of it to biguintcore.d. Please review it again.

JonathanWilbur:
Just following up on this, @wilzbach (or anybody else), is there a reason this has not been merged yet? Do I need to change anything?

I should also add that the auto-tester appears to run on and off, so it looks like this is still in testing even though it has passed before. There was one strange failure a few days ago, which appears to come from:

```
make[1]: *** [generated/linux/debug/64/unittest/std/algorithm/searching.o] Killed
make[1]: Leaving directory `/media/ephemeral0/sandbox/at-client/pull-3115292-Linux_64_64/phobos'
```

which I don't think has anything to do with my commit.

wilzbach:
Thanks for the ping and sorry for not clarifying this. New symbols need the following:
- a changelog entry
- approval from @andralex
- full ddoc documentation headers

Some CI have unfortunately spurious failures. While we are trying to weed them out, it's not always easy, but you can safely ignore such a failure.

wilzbach:
BTW at the end we need to squash all commits into one, so I recommend to use rebases instead of merge commits as then the final squashing is easier (though it's not a blocking issue if you have troubles with this. We can help you out with this if you need help and force-push over your branch at the end though I'm this case you used the master branch, so you might want to make sure that your punycode commit is safely stored in another branch)

std/bigint.d
```diff
@@ -1025,6 +1025,212 @@ public:
...
+    }
+
+    /**
+        Override for the `cast` operator, so the `BigInt` can be cast to a
+        `ubyte[]`.
+
+        Returns: a `ubyte[]` array, representing the native-endian
+            representation of that number.
+    */
+    T opCast(T : ubyte[])() pure nothrow @system const
```
n8sh: I am a bit wary of an `opCast` that allocates. Can someone else comment on whether that's normal for Phobos?

Biotronic: Searching for `opCast` in Phobos yields many results, but none of them seem to be allocating.

JonathanWilbur: I am a little confused. What is the alternative? This `opCast()` override just calls the `toBytes()` accessor.

Biotronic: The alternative is to not have the `opCast` at all, and force the programmer to call `toBytes()` to get the byte representation. The advantage of this is the allocation is more explicit, while `cast(byte[])myBigint` may look like it's not doing anything scary - after all, for some possible implementations, it would simply return a different view of the same data.

std/internal/math/biguintcore.d
```diff
@@ -409,6 +409,72 @@ public:
...
+        Converts the `BigUint` to a sequence of bytes that represents the
+        native-endian unsigned representation of that number. The sign is
+        completely ignored, so -1,000,000 will produce the same output as
+        1,000,000. The integer will be encoded on the fewest possible bytes.
+        In other words, there will be no leading zeroes.
+
+        Returns: the native-endian unsigned representation of the `BigInt` in
+            the form of a `ubyte[]` array.
+    */
+    ubyte[] toBytes() const pure @system @property nothrow
```
n8sh: Instead of allocating a new buffer would it be feasible to simply return a slice of `this.data`? Is [DIP1000's `scope` qualifier for function return values](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md#scope-function-returns) useful for that?

Biotronic: Nope. `toBytes` is 2s complement, while `BigInt` uses signed magnitude.

n8sh: For positive numbers.

std/bigint.d
JonathanWilbur: Ah. I've never actually thought about that, but that makes sense! I will remove the `opCast` override then.

std/internal/math/biguintcore.d
JonathanWilbur: I suppose that the internal array of `uint`s that is `this.data` could be cast to a `ubyte` array if it is positive, and a slice returned, but I will have to test that. But yeah, when it's negative, that goes out the door, as Biotronic said.

PR #6438, opened 2018-04-09
Add a wide fromStringz function.

GilesBathgate:
This adds a fromStringz function to complement the existing fromStringz function but for works for wide strings.

### DIFF
--- a/std/string.d
+++ b/std/string.d
@@ -212,21 +212,14 @@ class StringException : Exception
     $(RED Important Note:) The returned array is a slice of the original buffer.
     The original data is not changed and not copied.
 +/
-
-inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow
-{
-    return fromStringz!char(cString);
-}
-
-/++ Ditto +/
-inout(Char)[] fromStringz(Char)(inout(Char)* cString) @nogc @system pure nothrow
+Char[] fromStringz(Char)(Char* cString) @nogc @system pure nothrow
 if (isSomeChar!Char)
 {
     import core.stdc.stddef : wchar_t;
 
-    static if (is(Char == char))
+    static if (is(Unqual!Char == char))
         import core.stdc.string : cstrlen = strlen;
-    else static if (is(Char == wchar_t))
+    else static if (is(Unqual!Char == wchar_t))
         import core.stdc.wchar_ : cstrlen = wcslen;
     else
         static size_t cstrlen(const Char* s)
@@ -243,12 +236,12 @@ if (isSomeChar!Char)
 ///
 @system pure unittest
 {
-    assert(fromStringz(null) == null);
-    assert(fromStringz("foo") == "foo");
     assert(fromStringz("foo\0"c.ptr) == "foo"c);
     assert(fromStringz("foo\0"w.ptr) == "foo"w);
-    assert(fromStringz("福\0"w.ptr) == "福"w);
     assert(fromStringz("foo\0"d.ptr) == "foo"d);
+
+    assert(fromStringz("福\0"c.ptr) == "福"c);
+    assert(fromStringz("福\0"w.ptr) == "福"w);
     assert(fromStringz("福\0"d.ptr) == "福"d);
 }
 
@@ -260,6 +253,24 @@ if (isSomeChar!Char)
     assert(fromStringz(b) == null);
     dchar* c = null;
     assert(fromStringz(c) == null);
+
+    const char* d = "foo\0";
+    assert(fromStringz(d) == "foo");
+
+    immutable char* e = "foo\0";
+    assert(fromStringz(e) == "foo");
+
+    const wchar* f = "foo\0";
+    assert(fromStringz(f) == "foo");
+
+    immutable wchar* g = "foo\0";
+    assert(fromStringz(g) == "foo");
+
+    const dchar* h = "foo\0";
+    assert(fromStringz(h) == "foo");
+
+    immutable dchar* i = "foo\0";
+    assert(fromStringz(i) == "foo");
 
     immutable wchar z = 0x0000;
     // Test some surrogate pairs
--- a/std/string.d
+++ b/std/string.d
@@ -206,16 +206,38 @@ class StringException : Exception
     Params:
         cString = A null-terminated c-style string.
 
-    Returns: A D-style array of `char` referencing the same string.  The
-    returned array will retain the same type qualifiers as the input.
+    Returns: A D-style array of `char`, `wchar` or `dchar` referencing the same
+    string. The returned array will retain the same type qualifiers as the input.
 
     $(RED Important Note:) The returned array is a slice of the original buffer.
     The original data is not changed and not copied.
 +/
 
-inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow {
-    import core.stdc.string : strlen;
-    return cString ? cString[0 .. strlen(cString)] : null;
+inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow
+{
+    return fromStringz!char(cString);
+}
+
+/++ Ditto +/
+inout(Char)[] fromStringz(Char)(inout(Char)* cString) @nogc @system pure nothrow
+if (isSomeChar!Char)
+{
+    import core.stdc.stddef : wchar_t;
+
+    static if (is(Char == char))
+        import core.stdc.string : cstrlen = strlen;
+    else static if (is(Char == wchar_t))
+        import core.stdc.wchar_ : cstrlen = wcslen;
+    else
+        static size_t cstrlen(const Char* s)
+        {
+            const(Char)* p = s;
+            while (*p)
+                ++p;
+            return p - s;
+        }
+
+    return cString ? cString[0 .. cstrlen(cString)] : null;
 }
 
 ///
@@ -223,6 +245,34 @@ inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow {
 {
     assert(fromStringz(null) == null);
     assert(fromStringz("foo") == "foo");
+    assert(fromStringz("foo\0"c.ptr) == "foo"c);
+    assert(fromStringz("foo\0"w.ptr) == "foo"w);
+    assert(fromStringz("福\0"w.ptr) == "福"w);
+    assert(fromStringz("foo\0"d.ptr) == "foo"d);
+    assert(fromStringz("福\0"d.ptr) == "福"d);
+}
+
+@system pure unittest
+{
+    char* a = null;
+    assert(fromStringz(a) == null);
+    wchar* b = null;
+    assert(fromStringz(b) == null);
+    dchar* c = null;
+    assert(fromStringz(c) == null);
+
+    immutable wchar z = 0x0000;
+    // Test some surrogate pairs
+    // high surrogates are in the range 0xD800 .. 0xDC00
+    // low surrogates are in the range  0xDC00 .. 0xE000
+    // since UTF16 doesn't specify endianness we test both.
+    foreach (wchar[] t; [[0xD800, 0xDC00], [0xD800, 0xE000], [0xDC00, 0xDC00],
+            [0xDC00, 0xE000], [0xDA00, 0xDE00]])
+    {
+        immutable hi = t[0], lo = t[1];
+        assert(fromStringz([hi, lo, z].ptr) == [hi, lo]);
+        assert(fromStringz([lo, hi, z].ptr) == [lo, hi]);
+    }
 }
 
 /++


GilesBathgate:
<sigh> wchar_t is 16bit on windows and 32bit on posix. I will have to fix this PR.

GilesBathgate:
If this is any good I can add a dchar version too.

std/string.d
```diff
@@ -221,10 +221,52 @@ inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow {
...
+    }
+    else version (Posix)
+    {
+        size_t strlen16(const wchar* s)
+        {
+            const(wchar)* p = s;
+            while (*p)
+                p++;
+            return p - s;
+        }
```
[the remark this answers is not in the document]
wilzbach: My personal opinion: Correctness (and no potential weird corner-cases) > performance (and I doubt that this can be optimized much further by the stdc implementation)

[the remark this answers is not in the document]
GilesBathgate: Ok I agree, but as an aside, there is this: https://code.woboq.org/userspace/glibc/wcsmbs/wcslen.c.html  Is this faster? If so I don't really understand why ;)

std/string.d
```diff
@@ -221,10 +221,44 @@ inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow {
...
+    Returns: A D-style array of `wchar` referencing the same string.  The
+    returned array will retain the same type qualifiers as the input.
+
+    $(RED Important Note:) The returned array is a slice of the original buffer.
+    The original data is not changed and not copied.
++/
+
+inout(wchar)[] fromStringz(inout(wchar)* cString) @nogc @system pure nothrow
+{
+    static size_t strlen16(const wchar* s)
```
n8sh: Some libraries have optimized architecture-specific versions of `wcslen`. You could use that when C's `wchar_t` defined in `core.stdc.stddef` is the same width as D's `wchar`.

GilesBathgate: Please read https://github.com/dlang/phobos/pull/6438#discussion_r180185217 , unless I have misunderstood, using wcslen as I had coded originally was dismissed by @wilzbach in favour of consistency.

std/string.d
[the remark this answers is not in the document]
n8sh: It is manual loop unrolling.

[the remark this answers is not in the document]
GilesBathgate: This is even more fun: http://www.stdlib.net/~colmmacc/strlen.c.html (not for wide string though)

std/string.d
n8sh: I agree that we don't want `version (Windows)`, but in my opinion it would be perfectly safe to write `static if (wchar_t.sizeof == wchar.sizeof)`. @wilzbach what do you say?

wilzbach: _Optimization is the root of all evil_ :P

I just think that it's really hard to optimize such a three-line loop and that we should value correctness over potential weird corner cases.
So before we go down the route of calling a runtime function that might behave differently, we should be sure it's worth it.
That's just my opinion though.

wilzbach: (OTOH of course the case could be made that we should be able to trust the C runtime)

GilesBathgate: @wilzbach Would it be worth adding another test this new function, for example, I wanted to be sure that it works correctly when surrogate pairs are used. I think my assumption is correct that surrogate pairs are always non-zero.

n8sh: A surrogate pair consists of a high surrogate in the half-open range `D800 .. DC00` followed by a low surrogate in the half-open range `DC00 .. E000`. So yes, surrogate pairs are always non-zero.

n8sh: Regarding how an optimized implementation can be faster than a naïve implementation at the expense of clarity and brevity, some read until the pointer is properly aligned then switch to full-word reads or vector operations.

wilzbach: @GilesBathgate FWIW adding more tests is almost never a bad idea ;-)
@n8sh understood. I was just stating my opinion. I don't feel strongly about this and wouldn't object if Giles prefers to go with your proposal.

n8sh: @wilzbach Likewise!

std/string.d
```diff
@@ -257,6 +257,16 @@ inout(wchar)[] fromStringz(inout(wchar)* cString) @nogc @system pure nothrow
     assert(fromStringz(foo) == null);
     assert(fromStringz("foo"w.ptr) == "foo"w);
     assert(fromStringz("foo\0"w.ptr) == "foo"w);
+    assert(fromStringz("福\0"w.ptr) == "福"w);
+}
+
+@system pure unittest
+{
+    // Test for all surrogate pairs
+    // (its a total of 1048576 itterations but its fast enough)
```
n8sh: This seems excessive for something that's going to be run every time. Either test for specific failure scenarios, or just test endpoints and midpoints.

std/string.d
```diff
@@ -257,6 +257,16 @@ inout(wchar)[] fromStringz(inout(wchar)* cString) @nogc @system pure nothrow
...
+    for (wchar i = 0xD800; i < 0xDC00; ++i)
+        for (wchar j = 0xDC00; j < 0xE000; ++j)
+            assert(fromStringz([j, i, i, j, '\u0000'].ptr) == [j, i, i, j]);
```
n8sh: FYI a reversed surrogate pair isn't valid UTF-16.

GilesBathgate: UTF16 has no specified endianness, so while the sequence as a whole is never valid, it covers both little and big endian sequences.

std/string.d
wilzbach: Hmm, we have similar tests in std.algorithm (e.g. remove) or std.datetime and 1m iterations shouldn't take too long. It probably is still faster than a single test in std.concurrency
I would only worry about this if the execution time is really noticable/measurable.

GilesBathgate: @n8sh I agree it is excessive, but since it takes ~0.2ms to run I thought it would be better to be comprehensive. It is also hard for me to imagine corner cases in which wsclen and strlen16 would behave differently so this was to cover as many cases as possible given your suggestion.

n8sh: Why merely all surrogate pairs? Why not all sequences of two consecutive `wchar`s? A single test is sufficient to verify that the custom `wcslen` inside `fromStringz` is not performing auto-decoding. What are the subsequent million checks testing?

GilesBathgate: >What are the subsequent million checks testing?

The speed of electrons?

GilesBathgate: @n8sh You are right I will consider revising the test, but having worked in the industry for 15 years I think the world would be a better place with more comprehensive tests that saving some CPU cycles ;)

std/string.d
```diff
@@ -221,10 +221,66 @@ inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow {
 ///
 @system pure unittest
 {
-    assert(fromStringz(null) == null);
```
n8sh: I've also run into the problem of `null` literals matching multiple overrides. Is there a way to get the D compiler to resolve plain `null` to a specific override when there are multiple potentially matching functions without writing a `cast` every time?

GilesBathgate: Create an override that accepts void?

n8sh: For this particular case I've found what looks like a solution, changing `toStringz` to a templated function with a default template argument:

https://run.dlang.io/is/yJSJhX
```d
import std.stdio;

inout(Char)[] fromStringz(Char = char)(inout(Char)* cString)
if (is(Char == char) || is(Char == wchar) || is (Char == dchar))
{
	import core.stdc.wchar_ : wchar_t;
    static if (Char.sizeof == 1)
        import core.stdc.string : cstrlen = strlen;
    else static if (Char.sizeof == wchar_t.sizeof)
        import core.stdc.wchar_ : cstrlen = wcslen;
    else
        static size_t cstrlen(const wchar* s)
    	{
        	const(wchar)* p = s;
       		while (*p)
                ++p;
        	return p - s;
        }
    return cString ? cString[0 .. cstrlen(cString)] : null;
}

void main()
{
    string s = "abc\0"c;
    wstring w = "abc\0"w;
    assert(fromStringz(s.ptr) == "abc"c);
    assert(fromStringz(w.ptr) == "abc"w);
    assert(fromStringz(null) is null); // Can use plain `null`!
    writeln("Tests passed.");
}
```

GilesBathgate: That's a nice solution, it also implements dchar too. @wilzbach Any concerns about modifying the existing code as opposed to adding a new override?

jmdavis: You should be able to use `typeof(null)` if you want a function to explicitly accept `null`, though if a default template argument solves the problem, that seems cleaner than having to add an overload just for `null`.

n8sh: @jmdavis Thanks, that may solve the problem more generally for me.

std/string.d
```diff
@@ -206,23 +206,60 @@ class StringException : Exception
...
+    import core.stdc.wchar_ : wchar_t;
+
+    static if (Char.sizeof == 1)
+        import core.stdc.string : cstrlen = strlen;
+    else static if (Char.sizeof == wchar_t.sizeof)
+        import core.stdc.wchar_ : cstrlen = wcslen;
+    else
+        static size_t cstrlen(const Char* s)
+        {
+            const(wchar)* p = s;
```
n8sh: Typo from my post: should be `const(Char)*`.

std/string.d
```diff
@@ -206,23 +206,60 @@ class StringException : Exception
...
 
     $(RED Important Note:) The returned array is a slice of the original buffer.
     The original data is not changed and not copied.
 +/
 
-inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow {
-    import core.stdc.string : strlen;
-    return cString ? cString[0 .. strlen(cString)] : null;
+inout(Char)[] fromStringz(Char = char)(inout(Char)* cString)
+        if (is(Char == char) || is(Char == wchar) || is(Char == dchar))
```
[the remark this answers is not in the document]
GilesBathgate: hmm, I used dfmt to check style. Don't know why it added the indent.

std/string.d
```diff
@@ -206,23 +206,73 @@ class StringException : Exception
...
+inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow
+{
     import core.stdc.string : strlen;
+
     return cString ? cString[0 .. strlen(cString)] : null;
 }
 
+/++ Ditto +/
+inout(Char)[] fromStringz(Char)(inout(Char)* cString) @nogc @system pure nothrow
+if (is(Char == wchar) || is(Char == dchar))
```
n8sh: If we can write`fromStringz!wchar` it would be nice if we could also write `fromStringz!char`. Was there another consideration that led you to keep the `char` version of `fromStringz` as a non-templated overload?

GilesBathgate: @n8sh Backwards compatibility. `fromStringz("foo");` (as per the original unit tests) does not compile with the template version. I tried to come up with a solution that worked as you originally suggested. Perhaps you will have more luck.

jmdavis: Honestly, I don't see much reason to care about `fromStringz("foo")`. The entire point of `fromStringz` is to convert from a null-terminated string to a dynamic array. As such, aside from `"foo"` being an easy way to get a null-terminated string for a test, I see no value in accepting it. It would never make sense to pass a string literal to `fromStringz` in real code.

GilesBathgate: @jmdavis True but I would hate to break everyone's code who is using it incorrectly.

jmdavis: I would be _very_ surprised if anyone actually uses `fromStringz` with a string literal. It's completely pointless. If you want to turn a string literal into a `string`, just use it as a `string`. It's already a `string`.

GilesBathgate: Will d choose non-deprecated function over a deprecated one during override/template matching?

GilesBathgate: I am happy to change the implementation to a single template function that breaks the API (and unittests) if @jmdavis @wilzbach, and @burner are happy with that. Otherwise I will leave it as is because the current solution doesn't break anything. Or unless @n8sh can suggest a work-around.

n8sh: I'm surprised there was an issue with string literals. Anyway, it doesn't seem necessary to consolidate the existing `fromStringz` into a template if it makes anything more difficult.

GilesBathgate: @n8sh Well if you can get the following to compile (without modifying main)

```d
inout(Char)[] fromStringz(Char = char)(inout(Char)* cString)
if (is(Char == char) || is(Char == wchar) || is (Char == dchar))
{
    import core.stdc.stddef : wchar_t;
    static if (Char.sizeof == 1)
        import core.stdc.string : cstrlen = strlen;
    else static if (Char.sizeof == wchar_t.sizeof)
        import core.stdc.wchar_ : cstrlen = wcslen;
    else
        static size_t cstrlen(const Char* s)
    	{
        	const(Char)* p = s;
       		while (*p)
                ++p;
        	return p - s;
        }

    return cString ? cString[0 .. cstrlen(cString)] : null;
}

void main()
{
    auto a = fromStringz(null);
    auto b = fromStringz("foo");
    auto c = fromStringz("foo\0"c.ptr);
    auto d = fromStringz("foo\0"w.ptr);
    auto d = fromStringz("foo\0"d.ptr);
}
```

aG0aep6G: > Well if you can get the following to compile (without modifying main)

I haven't really followed the discussion, and maybe this falls short in some regard. But string literals convert implicitly to pointers, so you can get that main to compile by providing concrete overloads instead of a template.

For the null literal, provide an overload that takes `typeof(null)`. That seems somewhat pointless to me, though. Why would a user write `fromStringz(null)` instead of `null`, `[]`, or `""`?

```d
inout(Char)[] fromStringzImpl(Char = char)(inout(Char)* cString)
if (is(Char == char) || is(Char == wchar) || is (Char == dchar))
{
    import core.stdc.stddef : wchar_t;
    static if (Char.sizeof == 1)
        import core.stdc.string : cstrlen = strlen;
    else static if (Char.sizeof == wchar_t.sizeof)
        import core.stdc.wchar_ : cstrlen = wcslen;
    else
        static size_t cstrlen(const Char* s)
        {
            const(Char)* p = s;
            while (*p)
                ++p;
            return p - s;
        }

    return cString ? cString[0 .. cstrlen(cString)] : null;
}

alias fromStringz = fromStringzImpl!char;
alias fromStringz = fromStringzImpl!wchar;
alias fromStringz = fromStringzImpl!dchar;

inout(char)[] fromStringz(inout typeof(null) cString)
{
    return fromStringzImpl!char(null);
}

void main()
{
    auto a = fromStringz(null);
    auto b = fromStringz("foo");
    auto c = fromStringz("foo\0"c.ptr);
    auto d = fromStringz("foo\0"w.ptr);
    auto e = fromStringz("foo\0"d.ptr);
}
```

GilesBathgate: @aG0aep6G Thanks for your input, I didn't see your comment until I pushed my solution. I think my version its simpler, but similar. I welcome any comments.

GilesBathgate:
For anyone wondering, I would rather not have a home-rolled `cstrlen` fallback function. Why re-implement `memchr` right?

Unfortunately, on windows, a 32bit `wcslen` function doesn't exist, and on posix a 16bit `wcslen` function doesn't exist.

wilzbach:
@JackStouffer why did you add the @andralex label here?
AFAICT this is just a generalization of the existing symbol and imho doesn't require new symbol approval.

std/string.d
```diff
@@ -206,23 +206,73 @@ class StringException : Exception
...
+    string. The returned array will retain the same type qualifiers as the input.
 
     $(RED Important Note:) The returned array is a slice of the original buffer.
     The original data is not changed and not copied.
 +/
 
-inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow {
-    import core.stdc.string : strlen;
-    return cString ? cString[0 .. strlen(cString)] : null;
+inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow
```
andralex: Would deleting this work? We can just rely on the template. Some breakages may occur in odd code that takes address but I think that's broken anyway after adding the template.

jmdavis:
>@JackStouffer why did you add the @andralex label here?
AFAICT this is just a generalization of the existing symbol and imho doesn't require new symbol approval.

IIRC, this wasn't originally a template, so it did involve adding a new function, whereas now, it templatizes an existing one.

std/string.d
jmdavis: I argued that before, but the argument was the tests that used string literals break without this. Personally, I think that the tests should just be fixed and the non-templated overload removed. Passing an actual string to `fromStringz` is just broken, and there's no need to pass string literals except for the tests.

GilesBathgate: >I think that's broken anyway after adding the template.

I am happy to remove the old method. The intention was to **not** break old code. IIRC The two testcase breakages were passing `null`, **AND** passing a plain string literal without an `a` suffix i.e "foo", vs "foo"a

GilesBathgate:
>IIRC, this wasn't originally a template

@jmdavis indeed, at one point I thought it should be called `fromWString` so that would have been a new symbol.

std/string.d
jmdavis: We don't want to break code, but given that the entire point of `fromString` is to take a null-terminated string and convert it to a D string, I don't see how it's a valid use case to pass it either a string literal or a `null` literal. They might have made sense as a simple way to write a test, but it's not something that real code would do.

std/string.d
```diff
@@ -206,23 +206,73 @@ class StringException : Exception
...
+{
+    return fromStringz!char(cString);
+}
+
+/++ Ditto +/
+inout(Char)[] fromStringz(Char)(inout(Char)* cString) @nogc @system pure nothrow
```
andralex: I think there's an opportunity here - drop the `inout` and let the template work the qualifier out.

std/string.d
GilesBathgate: >I would be very surprised if anyone actually uses fromStringz with a string literal. It's completely pointless.

@jmdavis Yes this was the jist of your argument before. I completely agree, but I was leaving it up to someone to provide a definitive decision.

std/string.d
```diff
@@ -206,23 +206,66 @@ class StringException : Exception
...
 +/
+Char[] fromStringz(Char)(Char* cString) @nogc @system pure nothrow
+if (isSomeChar!Char)
+{
+    import core.stdc.stddef : wchar_t;
 
-inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow {
-    import core.stdc.string : strlen;
-    return cString ? cString[0 .. strlen(cString)] : null;
+    static if (is(Char == char))
```
n8sh: Now that you've removed the `inout`, this should be changed to either `is(Unqual!Char == char)` (using `std.traits : Unqual`) or `Char.sizeof == char.sizeof`.

n8sh: And same below.

GilesBathgate:
@andralex I've removed `inout` as you requested, added `Unqual!Char` as @n8sh requested, and removed the old function as @jmdavis requested. } // good to go

n8sh:
All review comments have been addressed so I am labeling this `72h no objection -> merge`.

PR #6440, opened 2018-04-09
Qualify MmapAllocator

nordlow:
Because `mmap` and `munmap` potentially sets errno I guess we need to wrap them in pure variants `pureMmap` and `pureUnmap` similar to `PureMallocator` calling `pureMalloc`.

Ok to move forward with this, @andralex?


n8sh:
>Because `mmap` and `munmap` potentially sets errno I guess we need to wrap them in pure variants `pureMmap` and `pureUnmap` similar to `PureMallocator` calling `pureMalloc` and alikes.

`mmap` cannot be pure in general because you can use it to write to read from and write to files. The way to do it would be to locally purify a `private` version of `mmap` [like `fakePureMalloc` inside `core.memory`](https://github.com/dlang/druntime/blob/b3dbfc267ab53488a9e285f5315cc7375ff66f4d/src/core/memory.d#L912-L921) (which is what `pureMalloc` uses internally) with the promise that we won't ever use it except with the `MAP_ANON` flag.
```d
extern (C) private pure @system @nogc nothrow
{
    pragma(mangle, "fakePureErrnoImpl") ref int fakePureErrno();

    pragma(mangle, "malloc") void* fakePureMalloc(size_t);
    pragma(mangle, "calloc") void* fakePureCalloc(size_t nmemb, size_t size);
    pragma(mangle, "realloc") void* fakePureRealloc(void* ptr, size_t size);

    pragma(mangle, "free") void fakePureFree(void* ptr);
}
```

nordlow:
Where do you want me to be these purified variants?

n8sh:
They should be `private` so put them in this file with an explanatory comment.

nordlow:
Done.

std/experimental/allocator/mmap_allocator.d
```diff
@@ -66,15 +65,48 @@ struct MmapAllocator
...
+extern (C) private @system @nogc nothrow
+{
+    ref int fakePureErrnoImpl()
+    {
+        import core.stdc.errno : errno;
+        return errno();
+    }
+}
+
+extern (C) private pure @system @nogc nothrow
```
n8sh: An explanatory comment for why it's valid to apply `pure` is necessary in the source code.

nordlow: Done.

wilzbach:
Mind the CI failures:

```
src/etc/linux/memoryerror.d:(.text.fakePureErrnoImpl+0x0): multiple definition of `fakePureErrnoImpl'
generated/freebsd/release/32/libphobos2.so.0.80.o:std/typetuple.d:(.text.fakePureErrnoImpl+0x0): first defined here
```

So I guess you need to pick a different and unique name.

n8sh:
>I still need guidance on the Windows variant.

@nordlow The same idea, except with `GetLastError`/`SetLastError` instead of `errno`.

nordlow:
Thanks!

PR #6444, opened 2018-04-11
Make RCIAllocators pure

edi33416:


aG0aep6G:
How can `processAllocator` be `pure`? It accesses global mutable state, doesn't it? Isn't that the one thing that `pure` functions are not supposed to do?

aG0aep6G:
> How can processAllocator be pure? It accesses global mutable state, doesn't it? Isn't that the one thing that pure functions are not supposed to do?

To add some weight to that, here's code that compiles with this PR, but shouldn't:

```d
bool f() pure @safe
{
    import std.experimental.allocator;
    auto pa = processAllocator();
    processAllocator() = RCISharedAllocator.init;
    return pa is RCISharedAllocator.init;
}
void main()
{
    assert(f() == f()); /* Fails. Must pass with pure function. */
}
```

PR #6447, opened 2018-04-11
Removed scope wide @trusted use in std.typecons

JackStouffer:


wilzbach:
@klickverbot thanks a lot for opening the issue about this!
AFAICT the idea behind this is/was to remove all @trusted scopes, s.t. DScanner will prevent new ones from being added.
Of course, this doesn't change the numerous @trusted holes in Phobos, but makes it easier to follow-up on this and remove @trusted on a per-function basis.

PR #6450, opened 2018-04-12
Remove version(unittest) imports to avoid extra imports in user code.

schveiguy:
Having imports in `version(unittest)` blocks means that user code being unit tested will also import these files incurring a compile-time penalty.

Should help with dlang/dmd#8124 a bit, at least for the `version(unittest)` portion.

I will submit more PRs for the rest of Phobos, I wanted to make these PRs small and reviewable.

### DIFF
--- a/std/algorithm/sorting.d
+++ b/std/algorithm/sorting.d
@@ -640,9 +640,12 @@ if (isRandomAccessRange!Range && hasLength!Range && hasSlicing!Range && hasAssig
             // Loop invariant
             version(unittest)
             {
-                import std.algorithm.searching : all;
-                assert(r[0 .. lo].all!(x => !lt(p, x)));
-                assert(r[hi + 1 .. r.length].all!(x => !lt(x, p)));
+                // this used to import std.algorithm.all, but we want to save
+                // imports when unittests are enabled if possible.
+                foreach (x; r[0 .. lo])
+                    assert(!lt(p, x));
+                foreach (x; r[hi+1 .. r.length])
+                    assert(!lt(x, p));
             }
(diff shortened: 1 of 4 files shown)


std/algorithm/sorting.d
```diff
@@ -640,9 +640,12 @@ if (isRandomAccessRange!Range && hasLength!Range && hasSlicing!Range && hasAssig
             // Loop invariant
             version(unittest)
             {
-                import std.algorithm.searching : all;
-                assert(r[0 .. lo].all!(x => !lt(p, x)));
-                assert(r[hi + 1 .. r.length].all!(x => !lt(x, p)));
+                // this used to import std.algorithm.all, but we want to save
+                // imports when unittests are enabled if possible.
```
schveiguy: I added the comment here to prevent regression.

schveiguy:
ping @marler8997 here's the first one.

JackStouffer:
```
std/array.d(557:1)[warn]: A unittest should be annotated with at least @safe or @system
std/array.d(576:1)[warn]: A unittest should be annotated with at least @safe or @system
```

schveiguy:
@JackStouffer fixed

schveiguy:
FYI @marler8997 I have a few more modules to do, they are the tougher/more controversial ones. Will get to them next week.

std/algorithm/sorting.d
wilzbach: StdUnittest was reverted because apparently `-deps` includes _all_ unittests and thus usage of StdUnittest caused breakage for projects using `-deps`

PR #6452, opened 2018-04-13
Trim as much version(unittest) as possible from allocator

schveiguy:
Most of the `version(unittest)` stuff here is not needed at all (there's no reason to guard unittests with `version(unittest)`).

A couple of the changes are not trivial, however.

See #6450 for more details.

### DIFF
--- a/std/experimental/allocator/common.d
+++ b/std/experimental/allocator/common.d
@@ -612,7 +612,7 @@ version(unittest)
         // unnecessary import of std.experimental.allocator
         import std.experimental.allocator : RCIAllocator, RCISharedAllocator;
         static assert(is(RCAllocInterface == RCIAllocator)
-            || is (RCAllocInterface == shared RCISharedAllocator));
+            || is (RCAllocInterface == RCISharedAllocator));
 
         import std.conv : text;
         import std.math : isPowerOf2;
--- a/std/experimental/allocator/common.d
+++ b/std/experimental/allocator/common.d
@@ -481,7 +481,6 @@ Forwards each of the methods in `funs` (if defined) to `member`.
 
 version(unittest)
 {
-    import std.experimental.allocator : RCIAllocator, RCISharedAllocator;
 
     package void testAllocator(alias make)()
     {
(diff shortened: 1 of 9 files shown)


std/experimental/allocator/mallocator.d
```diff
@@ -399,24 +394,27 @@ version(CRuntime_DigitalMars)
 {
     void* m;
 
+    @nogc nothrow
+        size_t addr() { return cast(size_t) m; }
```
schveiguy: I moved this in the unittest, even though the version block had no imports, as I can't imagine there's a good reason for adding `addr` to the global namespace as a public symbol. The loss of UFCS is mitigated by the fact that we always were calling this on `m`, so it actually gets shorter.

std/experimental/allocator/common.d
```diff
@@ -608,9 +607,13 @@ version(unittest)
...
 
     package void testAllocatorObject(RCAllocInterface)(RCAllocInterface a)
-    if (is(RCAllocInterface == RCIAllocator)
-        || is (RCAllocInterface == shared RCISharedAllocator))
     {
+        // this used to be a template constraint, but moving it inside prevents
+        // unnecessary import of std.experimental.allocator
+        import std.experimental.allocator : RCIAllocator, RCISharedAllocator;
+        static assert(is(RCAllocInterface == RCIAllocator)
+            || is (RCAllocInterface == shared RCISharedAllocator));
```
edi33416: While you are at it, you could also remove the `shared` keyword as this was before declaring `RCISharedAllocator` as a `shared struct`.

std/experimental/allocator/mallocator.d
edi33416: The loss of UFCS might make this less intuitive. I know I had to double check what `addr` does. What is the gain in changing the function signature?

Do the function attributes and the function definition have different indents or is it just the web display?

schveiguy: If I didn't change the function (and just made it a static), then it would become `addr(m)`. I changed it to m_addr to make it more obvious what it is.

The attributes were indented differently. This is what my editor did. But I have redone the function into a lambda, which looks quite nice.

schveiguy:
Fixed nits, let me know what you think on the `m_addr` thing.

edi33416:
I think this solves the (my :smile: ) "problem" perfectly

edi33416:
Don't you have to call `m_addr` or does it get called by the compiler? I can see that the `Win32` build is failing because of this
```
std\experimental\allocator\mallocator.d(402): Error: incompatible types for `(__lambda1) & (15)`: `uint delegate() pure nothrow @nogc @safe` and `int`
```

schveiguy:
Hah! the tests passed my mac, forgetting that this was a windows-specific test.

schveiguy:
I thought there was a way to alias a function lambda so that it becomes a real function? Oh well, will revert back to a function, but we don't need the attributes, they should be inferred.

schveiguy:
@edi33416 fixed. Not as pretty, but still short. Hopefully it passes this time, I tested something similar on my mac.

PR #6453, opened 2018-04-13
More removal of version(unittest)

schveiguy:
See #6450

### DIFF
--- a/std/internal/cstring.d
+++ b/std/internal/cstring.d
@@ -41,18 +41,6 @@ module std.internal.cstring;
 import std.range;
 import std.traits;
 
-version(unittest)
-@property inout(C)[] asArray(C)(inout C* cstr) pure nothrow @nogc @trusted
-if (isSomeChar!C)
-in { assert(cstr); }
-do
-{
-    size_t length = 0;
-    while (cstr[length])
-        ++length;
-    return cstr[0 .. length];
-}
-
 /**
 Creates temporary 0-terminated $(I C string) with copy of passed text.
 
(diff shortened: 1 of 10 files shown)


std/internal/cstring.d
```diff
@@ -41,18 +41,6 @@ module std.internal.cstring;
 import std.range;
 import std.traits;
 
-version(unittest)
-@property inout(C)[] asArray(C)(inout C* cstr) pure nothrow @nogc @trusted
```
schveiguy: Putting a trusted function like this in the global namespace is not a good idea.

std/socket.d
```diff
@@ -122,14 +122,10 @@ else
 
 version(unittest)
 {
-    static assert(is(uint32_t == uint));
-    static assert(is(uint16_t == ushort));
```
schveiguy: Note, `uint32_t` and `uint16_t` are nowhere else in this file. I have no idea what these were for so I removed them. If they are critical to have, they should go somewhere else.

std/socket.d
```diff
@@ -454,6 +450,7 @@ class Protocol
 version(CRuntime_Bionic) {} else
 @safe unittest
 {
+    // import std.stdio : writefln;
```
schveiguy: writefln is used in commented out lines below. I put in this comment because before it was covered by the version(unittest) block.

schveiguy:
ping @wilzbach what is this circleci error in typecons? I can't figure out if it's something I did or not, but I didn't touch typecons.

JackStouffer:
Random error, ignore it for the time being.

schveiguy:
> Random error

Ugh. Now doc builder too.

schveiguy:
Doc failure seems repeatable, but I don't know what it means. @wilzbach or @CyberShadow ?

CyberShadow:
Er, I'm not sure @wilzbach worked on the spec testing stuff, maybe he knows?

Might be exposing a latent compiler bug through some edge case.

wilzbach:
Looks really weird:

```
---
void main() {
import std.stdio;
        struct S { int x; float y; }

        int foo(S s) { return s.x; }

        foo( S(1, 2) ); // set field x to 1, field y to 2
        
}
---
/usr/bin/ld: error in __stdin.o(.eh_frame); no .eh_frame_hdr table will be created.
/usr/bin/ld: __stdin.o: invalid string offset 1096055152 >= 528 for section `.strtab'
/usr/bin/ld: unknown operator '(' in complex symbol
/usr/bin/ld: final link failed: Invalid operation
collect2: error: ld returned 1 exit status
Error: linker exited with status 1
```

I can try more testing tomorrow, but for now I have no idea why this is failing.

schveiguy:
@wilzbach thanks for looking into it.

wilzbach:
Regarding the DAutoTest failure -> https://github.com/dlang/dlang.org/pull/2345

PR #6454, opened 2018-04-16
Added Params and Returns to std.functional

JackStouffer:


std/functional.d
```diff
@@ -1528,11 +1591,19 @@ template forward(args...)
         static int foo(int n) { return 1; }
         static int foo(ref int n) { return 2; }
     }
+
+    // with forward
```
JackStouffer: also made this example clearer

JackStouffer:
Same doc failure as other PRs

std/functional.d
```diff
@@ -637,6 +661,11 @@ deprecated
 
 /**
 Negates predicate `pred`.
+
+Params:
+    pred = A string or a callable
+Returns:
+    A function which calls `pred` and returns its value negated.
```
n8sh: Although a predicate by definition is a function x => bool, it might be better to say "the logical negation of its value" instead of "its value negated".

JackStouffer:
@n8sh Fixed

JackStouffer:
Ping

PR #6456, opened 2018-04-16
Add alignedAllocate for AllocatorList

jercaianu:
`alignedAllocate` mimics the logic of the regular `allocate`, with the exception that if a new node is requested, its size would be at least `allocationSize + alignment - 1`, to make sure that the new aligned allocation would fit.

### DIFF
--- a/std/experimental/allocator/building_blocks/allocator_list.d
+++ b/std/experimental/allocator/building_blocks/allocator_list.d
@@ -196,6 +196,60 @@ struct AllocatorList(Factory, BookkeepingAllocator = GCAllocator)
         return null;
     }
 
+    /**
+    Allocate a block of size `s` with alignment `a`. First tries to allocate
+    from the existing list of already-created allocators. If neither can
+    satisfy the request, creates a new allocator by calling `make(s + a - 1)`
+    and delegates the request to it. However, if the allocation fresh off a
+    newly created allocator fails, subsequent calls to `alignedAllocate`
+    will not cause more calls to `make`.
+    */
+    static if (hasMember!(Allocator, "alignedAllocate"))
+    void[] alignedAllocate(size_t s, uint theAlignment)
+    {
+        import std.algorithm.comparison : max;
+        import core.checkedint : addu;
+
+        if (theAlignment == 0 || s == 0)
+            return null;
+
+        for (auto p = &root, n = *p; n; p = &n.next, n = *p)
+        {
+            auto result = n.alignedAllocate(s, theAlignment);
+            if (result.length != s) continue;
+            // Bring to front if not already
+            if (root != n)
+            {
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/building_blocks/allocator_list.d
```diff
@@ -196,6 +196,57 @@ struct AllocatorList(Factory, BookkeepingAllocator = GCAllocator)
...
+        if (root && root.empty == Ternary.yes)
+        {
+            // no can do
+            return null;
+        }
+
+        bool overflow = false;
+        size_t maxSize = adds(s - 1, cast(size_t) theAlignment, overflow);
+        assert(!overflow, "Requested size is too large");
+        if (overflow)
```
n8sh: Because of the `assert(!overflow)` above, [by the language specification the compiler may assume that `overflow` is never true](https://dlang.org/spec/contracts.html#assert_contracts) (although I do not believe any D compiler currently does this):
>The compiler is free to assume the assert expression is true and optimize subsequent code accordingly.

n8sh:
Aside from the issue with `assert` this PR looks completely good.

std/experimental/allocator/building_blocks/allocator_list.d
```diff
@@ -196,6 +196,57 @@ struct AllocatorList(Factory, BookkeepingAllocator = GCAllocator)
...
+        // Can't allocate from the current pool. Check if we just added a new
+        // allocator, in that case it won't do any good to add yet another.
+        if (root && root.empty == Ternary.yes)
+        {
+            // no can do
+            return null;
+        }
+
+        bool overflow = false;
+        size_t maxSize = adds(s - 1, cast(size_t) theAlignment, overflow);
```
edi33416: What if `s` is `0`? Would that be an issue, or would this overflow and thus return `null`? Maybe add a check at function entry?

edi33416: Also, could you add two tests that allocate `0` and `size_t.max`?

std/experimental/allocator/building_blocks/allocator_list.d
jercaianu: For debug builds, I assume that the assert will not be optimized.
However, in case an overflow happens, I have the check right below the assert which should never be optimized away.

std/experimental/allocator/building_blocks/allocator_list.d
jercaianu: The allocation requests are passed to `Allocator` which has to correctly treat these cases.

jercaianu: Also, added test for size `0`, but for `size_t.max`, the assert would trigger if I add a test.

PR #6457, opened 2018-04-16
Fix Issue 18766 - std.typecons.Tuple.toHash could be improved

n8sh:
Current std.typecons.Tuple has a custom `toHash` that is just the sum of the hashes of the tuple's fields. This PR instead combines values using ~~the runtime's default hash function~~ the equivalent of [`boost::hash_combine`](https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine).

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -1174,11 +1174,19 @@ if (distinctFieldNames!(Specs))
             Returns:
                 A `size_t` representing the hash of this `Tuple`.
          */
-        size_t toHash() const nothrow @trusted
+        size_t toHash() const nothrow @safe
         {
             size_t h = 0;
-            foreach (i, T; Types)
-                h += typeid(T).getHash(cast(const void*)&field[i]);
+            static foreach (i, T; Types)
+            {{
+                const k = typeid(T).getHash((() @trusted => cast(const void*) &field[i])());
+                static if (i == 0)
+                    h = k;
+                else
+                    // As in boost::hash_combine
+                    // https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine
+                    h ^= k + 0x9e3779b9 + (h << 6) + (h >>> 2);
+            }}
             return h;
         }
 


n8sh:
On the other hand it might be even better to delete the custom `toHash`.

JackStouffer:
If it has a custom opEquals it needs a toHash

n8sh:
I see.

n8sh:
Updated to use the equivalent of [`boost::hash_combine`](https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine).

std/typecons.d
```diff
@@ -1177,8 +1177,16 @@ if (distinctFieldNames!(Specs))
         size_t toHash() const nothrow @trusted
         {
             size_t h = 0;
-            foreach (i, T; Types)
-                h += typeid(T).getHash(cast(const void*)&field[i]);
+            static foreach (i, T; Types)
+            {{
+                const k = typeid(T).getHash(cast(const void*) &field[i]);
```
JackStouffer: Should use trusted escape here instead of on the whole function

n8sh: done

std/typecons.d
```diff
@@ -1174,11 +1174,19 @@ if (distinctFieldNames!(Specs))
...
          */
-        size_t toHash() const nothrow @trusted
+        size_t toHash() const nothrow @safe
         {
             size_t h = 0;
-            foreach (i, T; Types)
-                h += typeid(T).getHash(cast(const void*)&field[i]);
+            static foreach (i, T; Types)
+            {{
+                const k = typeid(T).getHash((() @trusted => cast(const void*) &field[i])());
```
edi33416: Shouldn't this call `T.toHash` if it's defined?

std/typecons.d
```diff
@@ -1174,11 +1174,19 @@ if (distinctFieldNames!(Specs))
...
+                static if (i == 0)
+                    h = k;
+                else
+                    // As in boost::hash_combine
+                    // https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine
+                    h ^= k + 0x9e3779b9 + (h << 6) + (h >>> 2);
```
edi33416: Maybe we could use `boost::hash_combine` to enhance how the hash of an `AA` is computed by [`hashOf`](https://github.com/dlang/druntime/blob/master/src/core/internal/hash.d#L170)?
Currently that just does a `^=`

std/typecons.d
```diff
@@ -1174,11 +1174,19 @@ if (distinctFieldNames!(Specs))
             Returns:
                 A `size_t` representing the hash of this `Tuple`.
          */
-        size_t toHash() const nothrow @trusted
+        size_t toHash() const nothrow @safe
```
edi33416: Can this also be `pure`? I believe it should be.

std/typecons.d
n8sh: Yes, it should:

```d
class TypeInfo_Struct : TypeInfo
{
//...
    override size_t getHash(scope const void* p) @trusted pure nothrow const
    {
        assert(p);
        if (xtoHash)
        {
            return (*xtoHash)(p);
        }
        else
        {
            import core.internal.traits : externDFunc;
            alias hashOf = externDFunc!("rt.util.hash.hashOf",
                                        size_t function(const(void)[], size_t) @trusted pure nothrow @nogc);
            return hashOf(p[0 .. initializer().length], 0);
        }
    }
//...
}
```

std/typecons.d
n8sh: It can't because the base signature of `TypeInfo.getHash` doesn't include `pure`.

std/typecons.d
n8sh: We can't in that case because the AA's hash needs to be independent of traversal order.

wilzbach: Out of interest: why not use `std.digest` (e.g. CRC) and use the nicer `.put` API?

PR #6458, opened 2018-04-16
Fix Issue 18675 - Missing toHash in std/experimental/checkedint.d

vlasebian:
I am trying to implement `toHash` functionality in `std.experimental.checkedint.Checked`. I have written the function and some tests, I hope I am on the right track.

### DIFF
--- a/std/experimental/checkedint.d
+++ b/std/experimental/checkedint.d
@@ -472,6 +472,48 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
         assert(!MyHook.thereWereErrors);
     }
 
+    // toHash
+    /**
+
+    Generates a hash for `this`. If `Hook` defines `hookToHash`, the call
+    immediately returns `hook.hookToHash(payload)`. If `Hook` does not
+    implement `hookToHash`, but it has state, a hash will be generated for
+    the `Hook` using the built-in function and it will be xored with the
+    hash of the `payload`. Otherwise, the hash of the payload is returned.
+
+    */
+    size_t toHash() const nothrow @safe
+    {
+        static size_t hashFunc(T)(T objToHash)
+        {
+            return (() @trusted => typeid(objToHash).getHash(&objToHash))();
+        }
+
+        static if (hasMember!(Hook, "hookToHash"))
+        {
+            return hook.hookToHash(payload);
+        }
+        else static if (stateSize!Hook > 0)
+        {
+            static if (hasMember!(typeof(payload), "toHash"))
+            {
+                return payload.toHash() ^ hashFunc(hook);
+            }
+            else
+            {
+                return hashFunc(payload) ^ hashFunc(hook);
+            }
+        }
+        else static if (hasMember!(typeof(payload), "toHash"))
+        {
+            return payload.toHash();
+        }
+        else
+        {
+            return hashFunc(payload);
+        }
+    }
+
     // opCmp
     /**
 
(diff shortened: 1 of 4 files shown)


std/experimental/checkedint.d
```diff
@@ -472,6 +472,23 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
         assert(!MyHook.thereWereErrors);
     }
 
+    // toHash
+    /** Generates a hash for `this`. If `Hook` defines `hookToHash`, the call
+      immediately returns `hook.hookToHash(get)`. Otherwise, if payload type
+      defines `toHash`, call is redirected to payload's `toHash`, else the
+      built-in function `getHash` is called.
+    */
+
```
JackStouffer: Remove line

std/experimental/checkedint.d
```diff
@@ -472,6 +472,23 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
+    size_t toHash()
+    {
+        static if (hasMember!(Hook, "hookToHash"))
+            return hook.hookToHash(payload);
+        else static if (hasMember!(typeof(payload), "toHash"))
+            return payload.toHash();
+        else
+            return (() @trusted => typeid(payload).getHash(&payload))();
```
JackStouffer: The problem with doing it this way is `checked!Warn(42).toHash == checked!Abort(42).toHash;`

See https://github.com/dlang/phobos/pull/6457/files for some idea of how to use the hash of the payload + the hash of the Hook

n8sh:
@vlasebian What is the motivation for adding an explicit `toHash` to `Checked`? The auto-generated hash function seems to work fine.

https://run.dlang.io/is/3jJ4Co
```d
void main()
{
    import std.experimental.checkedint : Checked, Abort, Warn;

    Checked!(int, Abort) a = 123;
    Checked!(int, Abort) b = 456;
    Checked!(int, Warn) c = 123;

    assert(hashOf(a) != hashOf(b), "Different values & same hook => different hash");
    b = 123;
    assert(hashOf(a) == hashOf(b), "Same value & same hook => same hash");
    assert(hashOf(a) != hashOf(c), "Same value & different hook => different hash");
}
```

jercaianu:
@n8sh 
To enable Checked to be used with associative arrays. When trying to to use Checked with AA you get the following:
`Error: AA key type Checked should have size_t toHash() const nothrow @safe if opEquals defined`

std/experimental/checkedint.d
edi33416: I believe this is the right way to do it in order to be consistent with `Checked.opEquals`.
I added an assert to the example provided by @n8sh [here](https://run.dlang.io/is/BmOoYK)

imho, delegating to the hook the behaviour of the has is the most transparent to the user

n8sh: >imho, delegating to the hook the behaviour of the has is the most transparent to the user

If the desired behavior is that `Checked`s with the same payload but different hooks have the same hash code, then it is never correct for a hook to have a custom hash function.

n8sh: Although with implicit numerical promotion it is not possible for `a == b` to entail `hashOf(a) == hashOf(b)`.

https://run.dlang.io/is/gRxZib
```d
void main()
{
    immutable int a = -1;
    immutable long b = -1;
    immutable float c = -1;
    assert(a == b);
    assert(a == c);
    assert(hashOf(a) != hashOf(b));
    assert(hashOf(a) != hashOf(c));
    
    import std.experimental.checkedint : Checked;
    immutable ca = Checked!int(a);
    immutable cb = Checked!long(b);
    assert(ca == cb);
    assert(ca == a && ca == b);
    assert(cb == a && cb == b);
    assert(hashOf(ca) != hashOf(cb) || hashOf(ca) != hashOf(a)
           || hashOf(cb) != hashOf(b), "Cannot all be true.");
    
    //assert(ca == c); // Bonus (doesn't compile).
}
```

vlasebian:
Thank you for taking time to review my PR. I will make the requested changes soon.

wilzbach:
@vlasebian just referencing the issue in your commit message is enough for bug fixes.
Have a look at https://github.com/dlang-bots/dlang-bot#automated-references

std/experimental/checkedint.d
```diff
@@ -472,6 +472,23 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
+      defines `toHash`, call is redirected to payload's `toHash`, else the
+      built-in function `getHash` is called.
+    */
+
+    size_t toHash()
+    {
+        static if (hasMember!(Hook, "hookToHash"))
+            return hook.hookToHash(payload);
+        else static if (hasMember!(typeof(payload), "toHash"))
+            return payload.toHash();
```
wilzbach: Uncovered

std/experimental/checkedint.d
```diff
@@ -473,20 +473,35 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
+    is checked for a `toHash` function. Otherwise, the build-in function
+    `getHash` is called.
+
+    */
+    size_t toHash() const nothrow @safe
     {
         static if (hasMember!(Hook, "hookToHash"))
-            return hook.hookToHash(payload);
+        {
+            static if (stateSize!Hook > 0)
```
edi33416: This isn't consistent with the docs above
> If `Hook` defines `hookToHash`, the call immediately returns `hook.hookToHash(payload)`
The docs don't mention anything about the state of the hook and how this matters in the hash.

This should just forward to the hook, regardless of state or not.
An example for when this would break is the following:
If your hook is a reference counted object, then two references to the same `hook` will
have different hashes because of the reference count value.

Just forward to `hook.hookToHash` and let the hook do all the work; for the previous example,
only the hook can decide to ignore the reference count.

vlasebian:
I'm closing this PR because I have messed up my history and I'll open another PR.

wilzbach:
Please avoid opening a new PR in the future and use rebase instead. This way the history and conversation is kept in one way place. If you have problems with this, ping us on #phobos on Slack or ask for help in the IRC channel.

PR #6459, opened 2018-04-17

MartinNowak:


JackStouffer:
Is this necessary? There are no changes.

MartinNowak:
> Is this necessary? There are no changes.

Yes, we the stable branch to be fully merged back, so we can check for unmerged commits in scripts.
I also just fixed sth. in a phobos changelog entry so it's no longer empty ;).

PR #6460, opened 2018-04-17
Add deallocateAll for SharedRegion

jercaianu:
Also removed global nogc, because if the parent is the GC, the ctor which receives `size_t` will not be nogc

### DIFF
--- a/std/experimental/allocator/building_blocks/region.d
+++ b/std/experimental/allocator/building_blocks/region.d
@@ -290,7 +290,7 @@ struct Region(ParentAllocator = NullAllocator,
     Deallocates all memory allocated by this region, which can be subsequently
     reused for new allocations.
     */
-    bool deallocateAll() @safe pure nothrow @nogc
+    bool deallocateAll() pure nothrow @nogc
     {
         static if (growDownwards)
         {
@@ -955,7 +955,6 @@ shared struct SharedRegion(ParentAllocator = NullAllocator,
     uint minAlign = platformAlignment,
     Flag!"growDownwards" growDownwards = No.growDownwards)
 {
-nothrow @nogc:
     static assert(minAlign.isGoodStaticAlignment);
     static assert(ParentAllocator.alignment >= minAlign);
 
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/building_blocks/region.d
```diff
@@ -1114,6 +1113,24 @@ nothrow @nogc:
         return false;
     }
 
+    /**
+    Deallocates all memory allocated by this region, which can be subsequently
+    reused for new allocations.
+    */
+    bool deallocateAll() @trusted pure nothrow @nogc
```
edi33416: This should not be `@trusted`. Deallocation is a `@system` operation from the pov of the allocator. The user can make it trusted through his own mechanisms, such as reference counting or other external meta data that can guarantee there are no more references to the allocated memory.

std/experimental/allocator/building_blocks/region.d
```diff
@@ -1114,6 +1113,24 @@ nothrow @nogc:
...
+    {
+        import core.atomic : atomicStore;
+        static if (growDownwards)
+        {
+            atomicStore(_current, cast(shared(void*)) roundedEnd());
```
n8sh: Do we need the default `MemoryOrder.seq` here rather than `MemoryOrder.rel`?

std/experimental/allocator/building_blocks/region.d
edi33416: I've just noticed the same is true for `Region.deallocateAll`. Could you remove `@safe` from there as well?
@nordlow I hope this change won't break your code.

std/experimental/allocator/building_blocks/region.d
jercaianu: Honestly, I don't know if it will work the same with `MemoryOrder.rel`.
I'd rather not add such an optimization to a function which will be rarely used, at the risk of creating a hard to detect bug.
What are your thoughts on this?
Thanks!

std/experimental/allocator/building_blocks/region.d
jercaianu: Good observation, thanks!

jercaianu:
@edi33416 done

edi33416:
Cool. I'll add the `auto-merge` label as I understand that @n8sh is ok with the chosen `MemoryOrder`

PR #6461, opened 2018-04-17
In std.conv.emplaceInitializer use memset instead of memcpy when setting all 0s

n8sh:
In std.conv.emplaceInitializer use memset instead of memcpy when setting all 0s.

~~This PR also moves `isInitAllZeroBits` from std.experimental.allocator.common to ~~std.conv~~ **EDIT: `std.traits`** to avoid making everything that uses `std.conv` also depend on `std.experimental.allocator.common`. The visibility remains `package`.~~

EDIT2: The above remark is obsolete. This PR uses the new `__traits(isZeroInit)`.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -4360,15 +4360,23 @@ if (is(UT == Unqual!UT))
 }
 
 //emplace helper functions
-private void emplaceInitializer(T)(ref T chunk) @trusted pure nothrow
+private void emplaceInitializer(T)(scope ref T chunk) @trusted pure nothrow
 {
     static if (!hasElaborateAssign!T && isAssignable!T)
         chunk = T.init;
     else
     {
-        import core.stdc.string : memcpy;
-        static immutable T init = T.init;
-        memcpy(&chunk, &init, T.sizeof);
+        static if (__traits(isZeroInit, T))
+        {
+            import core.stdc.string : memset;
+            memset(&chunk, 0, T.sizeof);
+        }
+        else
+        {
+            import core.stdc.string : memcpy;
+            static immutable T init = T.init;
+            memcpy(&chunk, &init, T.sizeof);
+        }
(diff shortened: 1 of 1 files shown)


std/conv.d
```diff
@@ -6604,3 +6622,158 @@ if ((radix == 2 || radix == 8 || radix == 10 || radix == 16) &&
         }
     }
 }
+
+/+
+Can the representation be determined at compile time to consist of nothing but
+zero bits? Padding between a struct's fields is not considered.
++/
+private template isAllZeroBits(T, T value)
```
n8sh: This code is moved unaltered from another module.

n8sh:
Currently causes dmd to crash.

std/conv.d
n8sh: Makes sense.

JackStouffer:
The issue with DMD should be reduced with dustmite.

n8sh:
@JackStouffer I've sent a reduced version of the failing code to @UplinkCoder.

wilzbach:
Could you please open an issue with the reduced code to the bugtracker, s.t. everyone has access to it and it doesn't get lost?
Thanks!

n8sh:
@wilzbach Sure, https://issues.dlang.org/show_bug.cgi?id=18819

n8sh:
The compiler crash has been fixed so this isn't blocked anymore.

@edi33416 moved `isInitAllZeroBits!T` to `std.traits`. Visibility remains `package`.

n8sh:
Updated to use improved `isInitAllZeroBits` from https://github.com/dlang/phobos/pull/6537.

n8sh:
Updated to use new `__traits(isZeroInit)` from https://github.com/dlang/dmd/pull/8583. Removing WIP label.

PR #6462, opened 2018-04-18
Allocator list does not return null if two empty nodes are found

jercaianu:
When an allocation is requested, if two empty nodes are detected, `allocate` and `alignedAllocate` return `null`.
This will not work with allocators which always return fresh memory. Such allocators may not hold alive objects (are empty), but don't have available fresh memory.
I also added a test which this PR fixes.


CyberShadow:
Thanks, I also ran into this bug:
https://issues.dlang.org/show_bug.cgi?id=18853

PR #6464, opened 2018-04-19
Fix issue 18778

Hackerpilot:
Check the return value of `fmt.writeUpToNextSpec` to determine if there are more format specifiers to format before handling the separator.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -3332,11 +3332,24 @@ if (isInputRange!T)
         for (;;)
         {
             auto fmt = FormatSpec!Char(f.nested);
-            fmt.writeUpToNextSpec(w);
-            if (f.flDash)
-                formatValue(w, val.front, fmt);
-            else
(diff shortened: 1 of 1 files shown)


Hackerpilot:
To investigate the CI failure I used digger to set up a test environment on my laptop. I saw the same error that the CI showed. Then I reverted my changes and still saw the same error.

wilzbach:
Sorry the failures are related to your changes. Try running `make -f posix.mak std/format.test`
The problem is hidden in the `assertCTFEable`. A better error shows up when you remove that:

```
core.exception.AssertError@std/format.d(3096): expected = `[73 74 72 69 6e 67]`, result = `[73 74 72 69 6e 67 ]`
```

So you write too much whitespace.

Hackerpilot:
There must have been a cache invalidation issue when I last re-ran the tests. Reproduced and fixed the issue.

PR #6466, opened 2018-04-20
Added examples to std.path

JackStouffer:


std/path.d
```diff
@@ -4076,6 +4094,17 @@ string expandTilde(string inputPath) nothrow
     }
 }
 
+///
+@system unittest
+{
+    version (Posix)
+    {
+        import std.process : environment;
+        environment["HOME"] = "dmd/test";
```
CyberShadow: This changes HOME for the remaining unit tests, too.

PR #6467, opened 2018-04-21
Fix Issue 18675 - Missing toHash in std/experimental/checkedint.d

vlasebian:
This is a clean version of #6458

### DIFF
--- a/changelog/std-experimental-checkedint-toHash.dd
+++ b/changelog/std-experimental-checkedint-toHash.dd
@@ -1,0 +1,21 @@
+`toHash` function has been added to `std.experimental.checkedint.Checked`
+
+Custom hashing can be implemented in a `Hook`. If no hashing is specified, then
+the built-in function will be used. Checked can now be used with associative
+arrays.
+
+-------
+import std.experimental.checkedint;
+
+static struct MyHook
+{
+    static size_t hookToHash(T)(T payload)
+    {
+        auto h = typeid(payload).getHash(&payload);
+        return h + 0x9e3779b9 + (h << 6) + (h >> 2);
+    }
+}
+
+auto h = checked!MyHook(142).toHash();
+-------
+
--- a/std/experimental/checkedint.d
+++ b/std/experimental/checkedint.d
@@ -186,6 +186,11 @@ $(TR $(TD `onUpperBound`) $(TD If defined, $(D hook.onUpperBound(value, bound))
 binary operators `+=`,  `-=`, `*=`, `/=`, `%=`, `^^=`, `&=`, `|=`, `^=`, `<<=`, `>>=`,
 and `>>>=` is larger than the largest value representable by `T`.)
 )
+$(TR $(TD `hookToHash`) $(TD If defined, $(D hook.hookToHash(payload))
+(where `payload` is a reference to the value wrapped by Checked) is forwarded
+to when `toHash` is called on a Checked type. Custom hashing can be implemented
+in a `Hook`, otherwise the built-in hashing is used.)
+)
 )
 
 Source: $(PHOBOSSRC std/experimental/checkedint.d)
@@ -508,6 +513,41 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
         assert(checked!MyHook2(uint(-42)) == a);
         // Hook on left hand side takes precedence, so no errors
         assert(!MyHook.thereWereErrors);
+    }
+
+    // toHash
+    /**
+    Generates a hash for `this`. If `Hook` defines `hookToHash`, the call
+    immediately returns `hook.hookToHash(payload)`. If `Hook` does not
+    implement `hookToHash`, but it has state, a hash will be generated for
+    the `Hook` using the built-in function and it will be xored with the
+    hash of the `payload`.
+    */
+    size_t toHash() const nothrow @safe
+    {
+        static if (hasMember!(Hook, "hookToHash"))
+        {
+            return hook.hookToHash(payload);
+        }
+        else static if (stateSize!Hook > 0)
+        {
+            static if (hasMember!(typeof(payload), "toHash"))
+            {
+                return payload.toHash() ^ hashOf(hook);
+            }
+            else
+            {
+                return hashOf(payload) ^ hashOf(hook);
+            }
+        }
+        else static if (hasMember!(typeof(payload), "toHash"))
+        {
+            return payload.toHash();
+        }
+        else
+        {
+            return .hashOf(payload);
+        }
     }
 
     // opCmp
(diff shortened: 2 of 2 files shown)


std/experimental/checkedint.d
```diff
@@ -472,6 +472,48 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
         assert(!MyHook.thereWereErrors);
     }
 
+    // toHash
+    /**
+
```
edi33416: Please delete empty line

std/experimental/checkedint.d
```diff
@@ -472,6 +472,48 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
+    Generates a hash for `this`. If `Hook` defines `hookToHash`, the call
+    immediately returns `hook.hookToHash(payload)`. If `Hook` does not
+    implement `hookToHash`, but it has state, a hash will be generated for
+    the `Hook` using the built-in function and it will be xored with the
+    hash of the `payload`. Otherwise, the hash of the payload is returned.
+
```
edi33416: Please delete empty line

std/experimental/checkedint.d
```diff
@@ -472,6 +472,48 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
+    */
+    size_t toHash() const nothrow @safe
```
edi33416: Please remove the function attributes as they should be inferred from `Hook` and/or `payload`

std/experimental/checkedint.d
```diff
@@ -3063,3 +3105,90 @@ version(unittest) private struct CountOpBinary
...
+        static if (size_t.sizeof == 4)
+        {
+            static size_t hashMask = 0xFFFF_0000;
+        }
+        else
+        {
+            static size_t hashMask = 0xFFFF_0000_FFFF_0000;
+        }
+
+        static size_t hookToHash(T)(T payload)
```
edi33416: Mark `hookToHash` as `@nogc nothrow @trusted` and then you can check that `toHash` correctly infers the attributes from his `Hook`. In order to check for the attributes add a `lambda` inside the assert, like so
`assert((() @nogc nothrow @safe => x.toHash() == x.toHash())());`

This defines a `lambda function` using the short syntax. You can read more about lambdas [here](https://dlang.org/spec/expression.html#lambdas)

std/experimental/checkedint.d
```diff
@@ -3063,3 +3105,90 @@ version(unittest) private struct CountOpBinary
...
+    assert(z.toHash() == t.toHash());
+    assert(z.toHash() != x.toHash());
+    assert(z.toHash() != w.toHash());
+
+    const long c = 0xF0F0F0F0;
+    const long d = 0xF0F0F0F0;
+
+    assert(checked!Hook1(c).toHash() != checked!Hook2(c));
+    assert(checked!Hook1(c).toHash() != checked!Hook1(d));
+
```
edi33416: Could you split this in two unittests? Then, you can use the `///` syntax to add this unittest to the [docs](https://dlang.org/spec/ddoc.html#using_ddoc_to_generate_examples).

std/experimental/checkedint.d
jercaianu: `toHash` is used by the associative array and the hashing function has to be `const nothrow @safe`. This will not work with `aa`, unless we force the `Hook` to respect the same attributes.
Instead of letting `toHash` infer attributes from the `Hook`, I think it's better for `toHash` to force the `Hook` to be `const nothrow @safe`.

edi33416: Cool. Thank you for the insight. I hadn't look at `AA`. Let's leave this as it is then.

I think we should still add a ddoc unittest that builds an AA of Checked

changelog/std-experimental-checkedint-toHash.dd
```diff
@@ -0,0 +1,21 @@
...
+arrays.
+
+-------
+import std.experimental.checkedint;
+
+static struct MyHook
+{
+    static size_t hookToHash(T)(T payload)
+    {
+        auto h = (() @trusted => typeid(payload).getHash(&payload))();
```
wilzbach: Not sure whether we want to advertise this `@trusted` in the public changelog.

std/experimental/checkedint.d
```diff
@@ -472,6 +472,46 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
+    implement `hookToHash`, but it has state, a hash will be generated for
+    the `Hook` using the built-in function and it will be xored with the
+    hash of the `payload`. Otherwise, the hash of the payload is returned.
+    */
+    size_t toHash() const nothrow @safe
+    {
+        static size_t hashFunc(T)(T objToHash)
+        {
+            return (() @trusted => typeid(objToHash).getHash(&objToHash))();
+        }
```
wilzbach: Why not `hashOf`?

std/experimental/checkedint.d
```diff
@@ -472,6 +472,46 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
+        else static if (stateSize!Hook > 0)
+        {
+            static if (hasMember!(typeof(payload), "toHash"))
+            {
+                return payload.toHash() ^ hashFunc(hook);
+            }
+            else
+            {
+                return hashFunc(payload) ^ hashFunc(hook);
```
wilzbach: Not covered

std/experimental/checkedint.d
```diff
@@ -472,6 +472,46 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
+            }
+        }
+        else static if (hasMember!(typeof(payload), "toHash"))
+        {
+            return payload.toHash();
```
wilzbach: Not covered

wilzbach:
Mind the CI failures on Windows:

```
std\experimental\checkedint.d(492): Error: `@safe` function `std.experimental.checkedint.Checked!(long, Hook1).Checked.toHash` cannot call `@system` function `std.experimental.checkedint.__unittest_L3108_C9.Hook1.hookToHash!(const(long)).hookToHash`
```

vlasebian:
I split the unittest and added an entry for `hookToHash` in the Hook Table. Also I changed the function a bit, because I couldn't think of a case when the payload has its own `toHash` function implemented (because checkedint always wraps an integer) so if the hook does not implement `hookToHash` the built-in function will be used.

JackStouffer:
@vlasebian 

> because I couldn't think of a case when the payload has its own toHash function implemented (because checkedint always wraps an integer) 

This is not the case. `Checked!(Checked!int, ProperCompare)` is totally valid.

vlasebian:
I made some changes. Hope it's ok now.

n8sh:
Rebased & replaced a use of `TypeInfo.getHash` with `hashOf`.

n8sh:
Noting for the record that I disagree with this in concept (as discussed in the other PR) but there's nothing wrong with the implementation.

PR #6468, opened 2018-04-21
[trivial] Fix `Nullable.get` docs

ntrel:
* Fix `datum` typo.
* Clarify that `Nullable!T.get` doesn't always throw (e.g. in release mode).
* Add preconditions for `Nullable!(T, nullValue).get`.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2839,9 +2839,9 @@ Params:
 }
 
 /**
-Gets the value if not null. If `this` is in the null state and the optional
-parameter `fallback` is passed, then `fallback` is returned, otherwise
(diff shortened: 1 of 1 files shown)


std/typecons.d
```diff
@@ -3446,6 +3446,7 @@ Params:
 Gets the value. `this` must not be in the null state.
 This function is also called for the implicit conversion to `T`.
 
+Preconditions: `isNull` must be `false`.
```
wilzbach: Why not use an in contract for this instead of just documenting it?

wilzbach:
(Rebased to restart Jenkins)

PR #6469, opened 2018-04-21
fix issue 18789 - std.stdio messes up UTF conversions on output

aG0aep6G:


aG0aep6G:
I think I'm soon ready to give up on this. Latest surprise: [Microsoft's `fwide`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fwide) is useless ("unimplemented"). So `orientation_` is bogus on that platform. With Microsoft, the distinction is ANSI/Unicode, not narrow/wide.

I can probably split off the wchar -> char case, though. That doesn't involve any file mode madness, as far as I can see. I'm going to look into that tomorrow.

aG0aep6G:
> I can probably split off the wchar -> char case, though. That doesn't involve any file mode madness, as far as I can see. I'm going to look into that tomorrow.

#6474

aG0aep6G:
Rebooted without out the stuff that has already been merged and the debugging commits. Let's see where it breaks this time around.

aG0aep6G:
I think we need to clarify what `LockingTextWriter`'s mission statement is. Is it (A) print text in a readable form (to a terminal)? Or is it (B) write UTF-8 always?

If A, we should always enforce wide/Unicode mode, transcode to `wchar_t`, and call `FPUTWC`. This would fix `writeln("Hello, wörld!");` on `MICROSOFT_STDIO` Windows, without needing `chcp 65001` or similar nonsense. I don't know if `DIGITAL_MARS_STDIO` Windows can be made to work, because it doesn't seem to have a Unicode mode.

If B, we should always enforce binary mode, transcode to UTF-8, and call `FPUTC`.

If neither, what is `LockingTextWriter`'s mission?

aG0aep6G:
> I think we need to clarify what `LockingTextWriter`'s mission statement is.

Ignoring that, I've rewritten the unittest. Might pass now.

aG0aep6G:
I think this is finally ready for review. buildkite says druntime is failing, but that seems unrelated.

aG0aep6G:
> I think this is finally ready for review.

Still waiting for someone to take a look at this.

std/stdio.d
```diff
@@ -3640,6 +3667,69 @@ void main()
     }
     assert(std.file.readText!string(deleteme).stripLeft("\uFEFF") == "foobar");
 }
+@safe unittest // char/wchar -> wchar_t
```
thewilsonator, line 3670: Is there any reason for this unittest to not just be `@system` , what with all the `@trusted` lambdas?

thewilsonator (approved):
Otherwise looks good.

std/stdio.d
aG0aep6G: The thing that's actually being tested is `writer.put(s)`. And we might want to ensure that that's `@safe`. The `@trusted` parts are just setup for the test.

But truth be told, I just started with `@safe` and added `@trusted` as needed. I can change it to `@system` if you like that better.

thewilsonator: fair enough.

aG0aep6G:
I guess auto-merge doesn't work when buildkite is red? As far as I can tell, this is the error in druntime:

generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std5regex8internal9kickstart__T7ShiftOrTaZQl6__ctorMFNcNeKSQCiQChQCe2ir__T5RegexTaZQjAkZSQDmQDlQDiQDc__TQCvTaZQDb: error: undefined reference to '_D3std5range__T11SortedRangeTAkVAyaa6_61203c3d2062VEQByQBx18SortedRangeOptionsi0ZQCo6lengthMFNaNbNdNiNfZm'

I can't reproduce that locally. And I don't see how this PR could cause that undefined reference. What do I do now?

wilzbach:
Yeah it's a known issue that sometimes the wrong druntime gets checked out

PR #6470, opened 2018-04-21
[Trivial] Annotates std.random unittests with `@nogc` when already possible

BBasile:


std/random.d
```diff
@@ -1154,21 +1154,22 @@ alias Mt19937_64 = MersenneTwisterEngine!(ulong, 64, 312, 156, 31,
...
-@safe pure nothrow unittest //11690
+@safe @nogc pure nothrow unittest //11690
 {
     alias MT(UIntType, uint w) = MersenneTwisterEngine!(UIntType, w, 624, 397, 31,
                                                         0x9908b0df, 11, 0xffffffff, 7,
                                                         0x9d2c5680, 15,
                                                         0xefc60000, 18, 1812433253);
 
-    ulong[] expectedFirstValue = [3499211612uL, 3499211612uL,
+    enum ulong[] expectedFirstValue = [3499211612uL, 3499211612uL,
```
JackStouffer: These should be `static ulong[]` to avoid copying.

std/random.d
```diff
@@ -1431,7 +1432,7 @@ alias Xorshift    = Xorshift128;                            /// ditto
     static assert(Xorshift32.min == 1);
 
     // Result from reference implementation.
-    auto checking = [
+    enum ulong[][] checking = [
```
JackStouffer: These should be `static ulong[][]` to avoid copying.

PR #6471, opened 2018-04-22
char[] and wchar[] should be output ranges

schveiguy:
There's no reason for this not to work. `put` jumps through hoops to helpfully transcode between character types just so it can work with any kind of output ranges that accept characters, but it had some specific arbitrary limitations for narrow strings that prevented them from working. This removes the restrictions, so putting to a `char[]` or `wchar[]` is now allowed. All widths work.

I'm not 100% sure of the paths that are taken, maybe they could use some improvement. But at least it will now work.

This enables nice optimizations such as easily allocating buffers on the stack to use with `formattedWrite`

### DIFF
--- a/std/base64.d
+++ b/std/base64.d
@@ -391,7 +391,7 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
      *  The number of times the output range's `put` method was invoked.
      */
     size_t encode(E, R)(scope const(E)[] source, auto ref R range)
-    if (is(E : ubyte) && isOutputRange!(R, char))
+    if (is(E : ubyte) && isOutputRange!(R, char) && !is(R == char[]))
     out(result)
     {
         assert(result == encodeLength(source.length), "The number of put is different from the length of Base64");
--- a/std/range/primitives.d
+++ b/std/range/primitives.d
@@ -467,21 +467,31 @@ void put(R, E)(ref R r, E e)
 }
 
 /**
- * Because of auto-decoding, the `front` of a `string` is a `dchar`,
- * so using `put` with `char` arrays is disallowed. In order to fill
- * any `char` type array, use $(REF byCodeUnit, std, utf).
+ * It's also possible to `put` any width strings or characters into narrow
+ * strings -- put does the conversion for you.
+ *
+ * Note that putting the same width character as the target buffer type is
+ * `nothrow`, but transcoding can throw a $(REF UTFException, std, utf).
  */
-@safe pure nothrow unittest
+@safe pure unittest
 {
-    import std.utf : byCodeUnit;
-
     // the elements must be mutable, so using string or const(char)[]
     // won't compile
     char[] s1 = new char[13];
-    auto r1 = s1.byCodeUnit;
+    auto r1 = s1;
+    put(r1, "Hello, World!"w);
+    assert(s1 == "Hello, World!");
+}
+
+@safe pure nothrow unittest
+{
+    // same thing, just using same character width.
+    char[] s1 = new char[13];
+    auto r1 = s1;
     put(r1, "Hello, World!");
     assert(s1 == "Hello, World!");
 }
+
 
 @safe pure nothrow @nogc unittest
 {
--- a/std/range/primitives.d
+++ b/std/range/primitives.d
@@ -253,7 +253,7 @@ enum bool isInputRange(R) =
 puts the whole raw element `e` into `r`. doPut will not attempt to
 iterate, slice or transcode `e` in any way shape or form. It will $(B only)
 call the correct primitive (`r.put(e)`,  $(D r.front = e) or
-`r(0)` once.
+`r(e)` once.
 
 This can be important when `e` needs to be placed in `r` unchanged.
 Furthermore, it can be useful when working with `InputRange`s, as doPut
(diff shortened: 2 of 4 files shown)


std/base64.d
```diff
@@ -391,7 +391,7 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
      *  The number of times the output range's `put` method was invoked.
      */
     size_t encode(E, R)(scope const(E)[] source, auto ref R range)
-    if (is(E : ubyte) && isOutputRange!(R, char))
+    if (is(E : ubyte) && isOutputRange!(R, char) && !is(R == char[]))
```
schveiguy: The other overload had this check, so I'm just thinking it was forgotten but the compiler didn't complain because R was never an output range of char.

std/range/primitives.d
```diff
@@ -272,6 +272,20 @@ private void doPut(R, E)(ref R r, auto ref E e)
...
+        r = r[1 .. $];
+    }
+    else static if (isNarrowString!R && isNarrowString!E && is(typeof(r[] = e)))
+    {
+        // slice assign. Note that this is a duplicate from put, but because
+        // putChar uses doPut exclusively, we have to copy it here.
+        immutable len = e.length;
+        r[0 .. len] = e;
+        r = r[len .. $];
+    }
```
schveiguy: This made it work, but I'm open to moving stuff around. I'm not quite sure what the point of doPut is, but I think this follows the rules above (it doesn't slice or modify `e` at all).

std/range/primitives.d
```diff
@@ -486,9 +500,9 @@ if (isSomeChar!E)
     ref const(wchar)[] wstringInit();
     ref const(dchar)[] dstringInit();
 
-    enum csCond = !isDynamicArray!R && is(typeof(doPut(r, cstringInit())));
-    enum wsCond = !isDynamicArray!R && is(typeof(doPut(r, wstringInit())));
-    enum dsCond = !isDynamicArray!R && is(typeof(doPut(r, dstringInit())));
+    enum csCond = is(typeof(doPut(r, cstringInit())));
+    enum wsCond = is(typeof(doPut(r, wstringInit())));
+    enum dsCond = is(typeof(doPut(r, dstringInit())));
```
schveiguy: Hard to tell, but I think that the `isDynamicArray!R` restrictions were specifically for narrow strings. If anyone wants to shed some light as to why they were there, I'd appreciate it.

std/base64.d
JackStouffer: >but the compiler didn't complain because R was never an output range of char

I specifically asked it to be removed IIRC in a different PR for this reason

thewilsonator:
Nice!

You should add tests for formatted write since that is the motivation for this PR.
Please also check if this also fixes 18472, I suspect it does.

schveiguy:
I can put in a test for `formattedWrite` (I did verify locally that this works), but I don't see how I can check 18472, as that seems to have a `TypeInfo` error on top of the `put` error. How would I test it?

thewilsonator:
The typeinfo was fixed.

schveiguy:
still happens on my branch, what PR fixed the TypeInfo thing?

thewilsonator:
#8173

schveiguy:
Ah, that's a DMD PR I updated your comments to reflect that. Will update my local DMD and see if it helps.

thewilsonator:
Ah whoops, i forgot this is phobos.

schveiguy:
Nope still fails: Error: TypeInfo cannot be used with -betterC. I think @JinShil says in the bug report that the PR helped in one case, but not the formatting case.

schveiguy:
Wait, I didn't realize that went into stable. Doesn't seem to be in master yet. Will re-check.

JackStouffer:
I’ll be able to review this tomorrow. I’m very interested in this change of it works.

std/range/primitives.d
JackStouffer: On line 379, if you change the static if, you should be able to consolidate this code with the existing code.

schveiguy: So that is what I did originally (I just removed the !isNarrowString). But the problem is that if the char widths don't match, it defers to `putChar`, but `putChar` never looks back at `put` again, it only goes to `doPut`. So I had to copy that code here.

Or maybe you mean something different?

I'm not sure the reason for this whole `put` trifecta, it's really hard to tell the purpose for all of these. So I figured this was the least disruptive.

schveiguy:
@JackStouffer ok, I didn't notice those.

std/range/primitives.d
JackStouffer: Just add the condition on line 281 as an OR to the condition on line 379

JackStouffer: Oh, I see the issue. Yeah you'll have to add an `is(typeof(r[0]) == typeof(e[0]))`

schveiguy: But then it won't work for putting one type of character into another type of buffer. For example, putting a `wchar` into a `char[]`. Right now, that does work.

schveiguy:
@JackStouffer fixed the example.

schveiguy:
@JackStouffer didn't see your message from before sorry. I'll take a look and see if there's a way to shuffle this around.

schveiguy:
@JackStouffer I looked again at this. I'd much rather have `put` handle the slice assign, since it's already handling the slice assign for a normal array, but the problem is that `putChar` can try transforming a single character (code unit) into an array, and it only calls `doPut`. Your suggestion is what I tried first, and it didn't work because of this limitation. If `putChar` would call `put` again, then we could handle it there, but it simply doesn't go that way (maybe it should?).

Not understanding the reason behind the `put`/`doPut` split, and in light of the fact that there is so much code that depends on the static introspection which I'm not sure is all tested in unittests here, I didn't want to mess too much with how this works. This was the least bad option.

So let me know if there is anything else stopping this from being merged, or if you can think of a better way to rearrange this.

schveiguy:
@JackStouffer ping, any comment here, or can we dismiss the stale review?

schveiguy:
Anyone with a concern here? I think this is ready to be merged.

n8sh:
>Anyone with a concern here?

Not a concern, but I'd like to know why it was disabled to begin with.

schveiguy:
I think it was preserving the original behavior. Here is a really OLD version of `std.range` that I think even predates github, that shows how the original put function looked:

https://github.com/dlang/phobos/blob/9d3ae0ad700856aa2d0896ffba77a8e115098614/std/range.d#L270

From what I can tell, at that point `put` would do *nothing* for parameters `char[]` and  `char[]`.

At this point, it was fixed to be an error, and then `char[]` became not an output range of `char[]`:

https://github.com/dlang/phobos/commit/cdc546892dea3a2c333e78b5c930ecae7e9cb982

So I would say I don't think it was a PLAN to make it not work, it just happened to work that way and was formalized.

PR #6472, opened 2018-04-22
fix issue 17084 - Can't sort array of structs with alias this

BBasile:
This fix is actually more a workaround. I propose it because it's verifiable by hand that `doesPointTo` is well `@nogc`.

### DIFF
--- a/std/exception.d
+++ b/std/exception.d
@@ -1088,6 +1088,10 @@ Params:
     source = The source object
     target = The target object
 
+Bugs:
+    The function is explicitly annotated `@nogc` because inference could fail,
+    see $(LINK2 https://issues.dlang.org/show_bug.cgi?id=17084, issue 17084).
+
 Returns: `true` if `source`'s representation embeds a pointer
 that points to `target`'s representation or somewhere inside
 it.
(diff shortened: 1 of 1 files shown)


wilzbach:
Hmm, how about using [`assumeNoGC`](https://forum.dlang.org/post/kvacxfalxqtoumatxtsx@forum.dlang.org) for the line the compiler can't infer correctly?
(though it looks like this should be better fixed in dmd)

BBasile:
I don't know. It uses a delegate, meaning an indirection. AssumeNoGC is rather used when the attribute cant even be put at all.

std/exception.d
```diff
@@ -1300,6 +1304,31 @@ bool mayPointTo(S, T)(auto ref const shared S source, ref const shared T target)
     assert(b.doesPointTo(*aLoc)); // b points to where a is pointing
 }
 
+
+version(unittest)
+{
+    // 17084 : bug doesn't happen if these declaration are local (even static)
+    struct Page17084
```
edi33416: Can those be made `private` or is this what you meant by `local`?

BBasile: they can be made `private`. by 
>local even static

 i meant inside the unittest block. Will make the comment less ambiguous.

edi33416:
@BBasile could you squash the commits?

Also, what about the `doesPointTo` overload for *shared objects*? Would that also require `@nogc` or is it able to infer it?

BBasile:
> what about the doesPointTo overload for shared objects? 

Added a test, it works without explicit `@nogc`.

PR #6473, opened 2018-04-22
Issue 18471

PetruGurita:
Added std.experimental.checkedint support for constructing or assigning a Checked!int with a custom hook that defines min and max. These defined min and max values are no longer ignored. The constructor and the assignment value are compared to these values and if defined, the `onLowerBound` and `onUpperBound` methods are triggered when `value < min` or `value > max` .

### DIFF
--- a/std/experimental/checkedint.d
+++ b/std/experimental/checkedint.d
@@ -293,26 +293,22 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
     {
         static if (isIntegral!U)
         {
-            auto val = rhs;
+            payload = rhs;
         }
         else
         {
-            auto val = rhs.payload;
+            payload = rhs.payload;
         }
 
         static if (hasMember!(Hook, "min") && hasMember!(Hook, "onLowerBound") &&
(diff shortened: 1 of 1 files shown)


RazvanN7:
Please add a commit containing the string "Fix Issue 18471 - " followed by the title of the issue so that the bot links the PR with the issue. Also, the title of the PR should be the same as the commit as just mentioned.

Please add a link to this PR on the bug post.

std/experimental/checkedint.d
```diff
@@ -292,9 +292,24 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
             is(typeof(Checked!(T, Hook)(rhs.get))))
     {
         static if (isIntegral!U)
+        {
```
RazvanN7: Please do only the necessary changes. Adding brackets here does not add any value to this PR

wilzbach:
Also
- add a changelog entry
- update the documentation on the top
- add more examples/test for more tricky cases
- add at least one public example (the ones with `///`)
- mind the CI failures

RazvanN7:
Closing due to inactivity.

PR #6474, opened 2018-04-23
fix conversion from wchar to char in LockingTextWriter.put

aG0aep6G:
Spin-off from #6469. I'm hoping that this will be relatively easy to get to work on the different platforms.

Instead of throwing a UTFException on unpaired high surrogates, a replacement character could be written. But throwing is what happens currently for unpaired low surrogates, so I went with that.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2926,8 +2926,7 @@ is empty, throws an `Exception`. In case of an I/O error throws
             }
             else static if (c.sizeof == 2)
             {
-                import std.utf : encode;
-                import std.utf : decodeFront;
+                import std.utf : decodeFront, encode;
 
                 if (orientation_ <= 0)
                 {
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2861,6 +2861,9 @@ is empty, throws an `Exception`. In case of an I/O error throws
             {
                 if (p.handle) FUNLOCK(p.handle);
             }
+            file_ = File.init;
+                /* Destroy file_ before possibly throwing. Else it wouldn't be
+                destroyed, and its reference count would be wrong. */
             rbuf16ShouldBeEmpty();
         }
 
@@ -3473,7 +3476,6 @@ void main()
             assertThrown!UTFException(writer.put(dchar('y')));
             assertThrown!UTFException(writer.put(surr));
         } ());
-        f.close(); // No idea why this is needed.
     }
     assert(std.file.readText!string(deleteme) == "x");
 
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2926,7 +2926,7 @@ is empty, throws an `Exception`. In case of an I/O error throws
             }
             else static if (c.sizeof == 2)
             {
-                import std.utf : decodeFront, encode;
+                import std.utf : encode;
 
                 if (orientation_ <= 0)
                 {
@@ -2942,11 +2942,13 @@ is empty, throws an `Exception`. In case of an I/O error throws
                     }
                     else // standalone or low surrogate
                     {
-                        immutable wchar[2] rbuf = [highSurrogate, c];
-                        wstring str = rbuf[highSurrogate == '\0' ? 1 : 0 .. $];
-                            // Skipping the high surrogate when there's none.
-                        highSurrogate = '\0';
-                        immutable dchar d = decodeFront(str);
+                        dchar d = c;
+                        if (highSurrogate != '\0')
+                        {
+                            immutable wchar[2] rbuf = [highSurrogate, c];
+                            d = rbuf[].front;
+                            highSurrogate = 0;
+                        }
                         char[4] wbuf;
                         immutable size = encode(wbuf, d);
                         foreach (i; 0 .. size)
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2831,6 +2831,16 @@ is empty, throws an `Exception`. In case of an I/O error throws
 
         // the file's orientation (byte- or wide-oriented)
         int orientation_;
+
+        // A buffer for when we need to transcode.
+        wchar[2] rbuf16;
+        size_t rbuf16Filled = 0;
+        void rbuf16ShouldBeEmpty() @safe
+        {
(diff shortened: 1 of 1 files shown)


aG0aep6G:
> I'm hoping that this will be relatively easy to get to work on the different platforms.

Yup. All green.

std/stdio.d
```diff
@@ -3419,6 +3445,49 @@ void main()
...
+        assertThrown!UTFException(() {
+            auto writer = f.lockingTextWriter();
+            writer.put('x');
+            writer.put(surr);
+            assertThrown!UTFException(writer.put(char('y')));
+            assertThrown!UTFException(writer.put(wchar('y')));
+            assertThrown!UTFException(writer.put(dchar('y')));
+            assertThrown!UTFException(writer.put(surr));
+        } ());
+        f.close(); // No idea why this is needed.
```
schveiguy: What happens if you don't include this?

std/stdio.d
```diff
@@ -3419,6 +3445,49 @@ void main()
...
+        writer.put("\U0001F608"w);
+    }
+    assert(std.file.readText!string(deleteme) == "\U0001F608");
+
+    // Test invalid input: unpaired high surrogate
+    {
+        immutable wchar surr = "\U0001F608"w[0];
+        auto f = File(deleteme, "w");
+        assertThrown!UTFException(() {
+            auto writer = f.lockingTextWriter();
```
schveiguy: Took me a few minutes to understand that the first assertThrown above is testing the throw from the lockingTextWriter dtor. Can you put in a comment to that effect?

std/stdio.d
```diff
@@ -2907,25 +2918,39 @@ is empty, throws an `Exception`. In case of an I/O error throws
...
+                    {
+                        rbuf16[rbuf16Filled] = c;
+                        ++rbuf16Filled;
+                        wchar[] str = rbuf16[0 .. rbuf16Filled];
+                        rbuf16Filled = 0;
+                        immutable dchar d = decodeFront(str);
+                        char[4] wbuf;
+                        immutable size = encode(wbuf, d);
+                        foreach (i; 0 .. size)
+                            trustedFPUTC(wbuf[i], handle_);
```
schveiguy: This seems like a lot of acrobatics for this purpose. A few comments here:

1. You only need a buffer of 1 wchar.
2. You don't need a "size" for checking if there are 0 or 1 wchars in there, just store a `wchar(0)` when you aren't using it.
3. From my experience with iopipe, you get better performance when you avoid using members to do the decoding.  This would be better off being duplicated in the function which takes whole wstrings, and you can then use a local to do all the decoding/encoding.

std/stdio.d
aG0aep6G: The `readText` below sees an empty file.

And now I see why: LockingTextWriter holds a reference counted File. When the destructor throws, the File won't be destroyed and its reference count won't be decremented. So the file only gets closed at the end of the program.

Fixed by destroying `file_` explicitly before throwing.

std/stdio.d
aG0aep6G: Added a comment at the end of the block.

std/stdio.d
aG0aep6G: Addressed points 1 and 2. Hopefully, the code's clearer now.

I think I'll leave point 3 for another PR.

std/stdio.d
schveiguy: Ah, glad you figured that out!

std/stdio.d
schveiguy: Good progress on the simplification. One more thing, I thought it might look better actually if you handle the surrogate pair explicitly instead of the ?: operator
```D
else
{
   dchar d = c; // default to just translate directly
   if(highSurrogate)
   {
      immutable wchar[2] buf = [highSurrogate, c];
      d = buf[].front; // doesn't this just work? Do we need to use decodeFront?
   }
   // .. all the rest of the stuff you have
}
```

schveiguy: It would be cool if std.utf provided a way to decode N code units directly. For example, if `decode(highSurrogate, c)` just worked.

schveiguy:
I suppose the difference between wchar and char is that LockingTextWriter is actually writing char, so there is a one-to-one mapping. When you are writing wchar, you can't just stop in the middle of a surrogate pair because nothing has been written. In any case, there is no denying that unicode is messy when it comes to streaming.

std/stdio.d
aG0aep6G: Your code misses the case where c is an unpaired low surrogate. But encode throws on that, so that works. Done.

aG0aep6G:
> I suppose the difference between wchar and char is that LockingTextWriter is actually writing char, so there is a one-to-one mapping. When you are writing wchar, you can't just stop in the middle of a surrogate pair because nothing has been written. When you are writing char, you can write everything given.

Yeah, I wouldn't validate in the char -> char case. Better just pass it through.

But there's also the char -> wchar_t case. There, the situation is the same as here, and I should check for unused chars in the buffer before starting a new code point.

schveiguy:
OK, I think this looks good. I'll give it a bit of time, and then merge it. @wilzbach does that label do it automatically?

JackStouffer:
No, it's manual.

schveiguy:
Not sure why the bot isn't merging, but I'll do it via auto-tester. Ping @wilzbach

schveiguy:
Auto-merge toggled on

wilzbach:
Because the bot only merges if all CI pass and Jenkins failed here.

schveiguy:
Ah, ok. Jenkins wasn't marked as required, so I assumed it would merge.

wilzbach:
Yeah the default behavior has been changed a few month ago as it's better if a human judges spurious CI failures and also it's better if we don't have them at all.
In this case, the failure was in DCD with this PR hopefully fixing this
https://github.com/dlang-community/DCD/pull/470
(and I also temporarily removed DCD from the tester until the PR is part of the next release - https://github.com/dlang-community/DCD/issues/455)

PR #6477, opened 2018-04-26
fix issue 18801 - std.stdio.File doesn't work with MSVCRT's UTF-8 mode

aG0aep6G:


std/stdio.d
```diff
@@ -2835,13 +2838,32 @@ is empty, throws an `Exception`. In case of an I/O error throws
...
+
+            version (MICROSOFT_STDIO)
+            {
+                // Microsoft doesn't implement fwide. Instead, there's the
+                // concept of ANSI/UNICODE mode. fputc doesn't work in UNICODE
+                // mode; fputwc has to be used. So that essentially means
+                // "wide-oriented" for us.
+                immutable int mode = _setmode(f.fileno, _O_TEXT);
+                    // Set some arbitrary mode to obtain the previous one.
+                _setmode(f.fileno, mode); // Restore previous mode.
```
aG0aep6G: This is a bit awkward, but I couldn't find a function that just gets the mode.

aG0aep6G:
I can't make sense of this.

The Win64 tester apparently sees the mode as `_O_WTEXT` when it should be `_O_U8TEXT`. But then it writes a UTF-8 BOM anyway. BUT THEN it encodes "foo" in UTF-16. And all that happens without D's File or LockingTextWriter being involved at all. It's just `fopen` and `fputwc`.

I'm open to suggestions.

wilzbach:
(rebased)

aG0aep6G:
Rebased properly (hopefully). I fully expect this to still fail in the same nonsensical way, though.

aG0aep6G:
Looks like the new Windows testing setup turned this green. With the nonsensical failures gone, this might be good to go.

There is a failure on buildkite, though: Apparently, CyberShadow/ae segfaults. I'm not sure how to address that. Is it caused by this PR? Do I need to investigate it? @CyberShadow?

Geod24:
https://github.com/CyberShadow/ae/issues/46

aG0aep6G:
Sweet. All green.

thewilsonator (approved):

PR #6478, opened 2018-04-26
Fix std.math.log1p() unittest

kinke:
`log1p(-2.0)` returns +NaN for the non-x87 software implementation, and so does the MSVC runtime function. The x87 asm implementation and libm on Linux return -NaN. So just ignore the sign.


ibuclaw:
Can't you just use copysign?

ibuclaw:
http://pubs.opengroup.org/onlinepubs/9699919799/functions/log1p.html

Only that it returns NaN.

PR #6479, opened 2018-04-26
std.math: Change fmin/fmax semantics wrt. NaNs

kinke:
If one of the arguments is a NaN, return the other. These are the C99 semantics; the previous simple ternary expression is covered by `std.algorithm.min/max`.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -6922,8 +6922,13 @@ real fdim(real x, real y) @safe pure nothrow @nogc { return (x > y) ? x - y : +0
 
 /**
  * Returns the larger of x and y.
+ *
+ * If one of the arguments is a NaN, the other is returned.
  */
-real fmax(real x, real y) @safe pure nothrow @nogc { return x > y ? x : y; }
+real fmax(real x, real y) @safe pure nothrow @nogc
+{
+    return (y > x || isNaN(x)) ? y : x;
+}
 
 ///
 @safe pure nothrow @nogc unittest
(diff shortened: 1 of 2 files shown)


std/math.d
```diff
@@ -6922,8 +6922,10 @@ real fdim(real x, real y) @safe pure nothrow @nogc { return (x > y) ? x - y : +0
 
 /**
  * Returns the larger of x and y.
+ *
+ * If one of the arguments is a NaN, the other is returned.
  */
-real fmax(real x, real y) @safe pure nothrow @nogc { return x > y ? x : y; }
+real fmax(real x, real y) @safe pure nothrow @nogc { return core.stdc.math.fmaxl(x, y); }
```
ibuclaw: Could we use isNaN here to allow the compiler to inline the tests? (Not a problem for compilers that treat the extern math functions as builtins).

kinke: Done.

kinke:
I think this makes perfect sense:

* Matches C semantics (same function name).
* Matches LLVM minnum & maxnum intrinsics semantics (i.e., LDC semantics).
* I'm perfectly fine with `std.algorithm.min/max` using the faster single comparison, as I guess that's what everybody uses. These different semantics justify the fmin/fmax existence IMO. I don't recall ever calling fmin/fmax myself in my code.

MartinNowak:
> This is not as simple as it seems

It never is :see_no_evil:.

What's the cost of this extra comparison? Maybe check glibc's fmax for optimizations.
Handling `fdim` is missing (and prolly others). What about code breakage?

ibuclaw:
Just a note on the "C semantics":  https://busybox.net/~landley/c99-draft.html#F.9.9

The equivalence of `isgreaterequal(x, y)` would be the removed `x !< y` operator, and so we would still need to check for isNaN on both sides I think - maybe even to sign bit also?

kinke:
> Handling fdim is missing

fdim already returns 0 if any arg is a NaN (and still conforms to its definition as `fmax(x-y, 0)`), see the tests:
```D
assert(fdim(real.nan, 2.0) == 0.0);
assert(fdim(2.0, real.nan) == 0.0);
```

kinke:
Rebasing onto master had unpleasant side effects (wrongly referenced issues & requested reviews), sorry about that. Is that a blocker or something else still to do?

wilzbach:
Not at all, dlang-bot probably just missed an event and the changelog is built from the commits, not what dlang-bot shows. That's merely a convenience tool.

PR #6485, opened 2018-04-30
fix issue 18804 - std.algorithm.mutation.copy puts whole source range…

aG0aep6G:
… into target range when it should put elements

### DIFF
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -414,7 +414,8 @@ if (!areCopyCompatibleArrays!(SourceRange, TargetRange) &&
     }
     else
     {
-        put(target, source);
+        foreach (element; source)
+            put(target, element);
         return target;
     }
 }
@@ -533,6 +534,24 @@ $(LINK2 http://en.cppreference.com/w/cpp/algorithm/copy_backward, STL's `copy_ba
         assert(a1[] == "123");
         assert(a2[] == "123789");
     }}
+}
+
+@safe unittest // issue 18804
+{
+    static struct NullSink
+    {
+        void put(E)(E) {}
+    }
+    int line = 0;
+    struct R
+    {
+        int front;
+        @property bool empty() { return line == 1; }
+        void popFront() { line = 1; }
+    }
+    R r;
+    copy(r, NullSink());
+    assert(line == 1);
 }
 
 /**


dukc:
Looks good.

But I don't have NPOW since I reported the bug.

JackStouffer:
Something definitely looks wrong here. Put is supposed to compile down to this loop on its own. Can you explain why this change modifies anything?

aG0aep6G:
> Something definitely looks wrong here. Put is supposed to compile down to this loop on its own. Can you explain why this change modifies anything?

`NullSink` is an output range for everything. In particular, it's an output range for `int` elements, and at the same time it's an output range for `R` elements. `put` has to decide which to choose. As per [the documentation](https://dlang.org/phobos/std_range_primitives.html#.put), it does so by trying in order:

1. `r.doPut(e);` - i.e., interpret the argument itself as an output element,
2. [...]
3. [...]
4. `for (; !e.empty; e.popFront()) put(r, e.front);` - i.e., interpret the argument as an input range of output elements.

With `NullSink`, the first step always wins. So generic code like `copy` has to do the loop itself.

dukc:
In fact, I thought it was dmd optimizer not noticing the range has side effects. Fortunately Ag0Aep was alert.

n8sh:
This is why functions shouldn't use the same name for "do X to Y" and "do X to every element of Y". (See also: `put`, `insertBack`, `replace`.) If D made more use of runtime polymorphism then problems like this would bite people a lot more in practice since it would be easy to write code that does the wrong thing but still compiles.

JackStouffer:
This isn't an issue with copy. This is an issue with `NullSink`; it's breaking the rules.

If any `put` accepts an input range, it's supposed to consume the entire range, or throw/error if it runs out of space before the range is done. `NullSink` is accepting the range and doing nothing. This fix should change `NullSink` to have two overloads, one which accepts all types of input ranges, and calls `each`. The other does nothing.

aG0aep6G:
> If any put accepts an input range, it's supposed to consume the entire range

I was going to agree with you and say that the range concept is ambiguous there, because an input range can be interpreted as a single element or as a range of elements. But looking over the documentation for `put` and `isOutputRange`, I'm not so sure anymore. Seems that's only the case for arrays, not for all input ranges.

The description of [`std.range.primitives.put`](https://dlang.org/phobos/std_range_primitives.html#.put) has three snippets that might be relevant:

1) `r.doPut(e);`
That is, if an output range accepts an `E` specifically, then it's an output range for `E` elements. This is the obvious case.
2) `r.doPut([ e ]);`
That is, if it accepts `E[]`, then it's also an output range for `E` elements. At the same time, it's an output range for `E[]` elements, as per point 1.
3) `for (; !e.empty; e.popFront()) put(r, e.front);`
That is, std.range.primitives.put will unwrap input ranges until it finds an element type that works for the output range. This is not about input ranges being passed to output ranges.

So there's an ambiguity with arrays. They can be seen as elements themselves, or as arrays of elements. Fortunately, NullSink can do the same for both: nothing.

But other input ranges can apparently only be elements, not ranges of elements.

dukc:
@JackStouffer This was supposed to be merged almost a month ago. I believe you should either merge or answer Ag0aep6g and remove the no objection -> merge label.

dukc:
And I think it's better like this, instead of modifying `NullSink`. The reason? Consider you are copying to `void[]`. What would you except 'copy' to do? Personally I would except it to copy the elements. That's what it is for. If I wanted to copy the range ref only, I'd assign it to 'target[0]'.

aG0aep6G:
> I am resetting the 72 hour merge clock on this.

That was 5 days ago. Ready to go?

PR #6487, opened 2018-04-30
[Trivial] truncl is pure in core/stdc/math -> remove specialization

dkgroot:
[Trivial] truncl is defined as pure in core/stdc/math and this was updated a couple of weeks ago to also be true for dragonfly, therefor this different definition can be remove and we can use the default.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -4948,77 +4948,61 @@ long lround(real x) @trusted nothrow @nogc
     }
 }
 
-version(StdDdoc)
+/**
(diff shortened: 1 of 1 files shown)


JackStouffer:
The StdDdoc version should be merged with the actual code, since there are no more differences between different platforms.

JackStouffer:
This shouldn't go on stable, as stable is for bug fixes or regressions. Please change the branch back to master.

dkgroot:
Sorry about that... I misunderstood

PR #6489, opened 2018-05-01
Static array UFCS-style

dukc:
This is #6178 rebooted, so most of this is work of @timotheecour  , not mine. It tries to address the last review @andralex made of it, and also some feedback made to #6214, which is my pull request aiming for same functionality.

### DIFF
--- a/changelog/std-array-asStatic.md
+++ b/changelog/std-array-asStatic.md
@@ -1,29 +1,23 @@
-Added `staticArray` to construct a static array from input array / range / CT range.
+Added `staticArray` to construct a static array from array / input range. Includes a length-inferring compile-time variant.
 
-The type of elements can be specified implicitly (`[1,2].staticArray` of type int[2])
-or explicitly (`[1,2].staticArray!float` of type float[2]).
-When `a` is a range (not known at compile time), the number of elements has to be given as template argument
-(eg `myrange.staticArray!2`).
-Size and type can be combined (eg: `2.iota.staticArray!(byte[2])`).
+The type of elements can be specified implicitly so that $(D [1, 2].staticArray) results in `int[2]`,
+or explicitly, e.g. $(D [1, 2].staticArray!float) returns `float[2]`.
+When `a` is a range whose length is not known at compile time, the number of elements must be given as template argument
+(e.g. `myrange.staticArray!2`).
+Size and type can be combined, if the source range elements are implicitly
+convertible to the requested element type (eg: `2.iota.staticArray!(long[2])`).
 When the range `a` is known at compile time, it can also be specified as a
 template argument to avoid having to specify the number of elements
-(eg: `staticArray!(2.iota)` or `staticArray!(double, 2.iota)`).
-
-Note: `foo([1, 2, 3].staticArray)` may be inefficient because of the copies involved.
+(e.g.: `staticArray!(2.iota)` or `staticArray!(double, 2.iota)`).
 
 ---
-auto a1 = [0, 1].staticArray;
-static assert(is(typeof(a1) == int[2]));
-assert(a1 == [0, 1]);
-
-auto a2 = [0, 1].staticArray!byte;
-static assert(is(typeof(a2) == byte[2]));
-
 import std.range : iota;
-auto input = 2.iota;
-auto a3 = input.staticArray!2;
-auto a4 = input.staticArray!(byte[2]);
 
-auto a5 = staticArray!(2.iota);
-auto a6 = staticArray!(double, 2.iota);
+auto input = 3.iota;
+auto a = input.staticArray!2;
+static assert(is(typeof(a) == int[2]));
+assert(a == [0, 1]);
+auto b = input.staticArray!(long[4]);
+static assert(is(typeof(b) == long[4]));
+assert(b == [0, 1, 2, 0]);
 ---
--- a/changelog/std-array-asStatic.md
+++ b/changelog/std-array-asStatic.md
@@ -1,0 +1,29 @@
+Added `staticArray` to construct a static array from input array / range / CT range.
+
+The type of elements can be specified implicitly (`[1,2].staticArray` of type int[2])
+or explicitly (`[1,2].staticArray!float` of type float[2]).
+When `a` is a range (not known at compile time), the number of elements has to be given as template argument
+(eg `myrange.staticArray!2`).
+Size and type can be combined (eg: `2.iota.staticArray!(byte[2])`).
+When the range `a` is known at compile time, it can also be specified as a
+template argument to avoid having to specify the number of elements
+(eg: `staticArray!(2.iota)` or `staticArray!(double, 2.iota)`).
+
+Note: `foo([1, 2, 3].staticArray)` may be inefficient because of the copies involved.
+
+---
+auto a1 = [0, 1].staticArray;
+static assert(is(typeof(a1) == int[2]));
+assert(a1 == [0, 1]);
+
+auto a2 = [0, 1].staticArray!byte;
+static assert(is(typeof(a2) == byte[2]));
+
+import std.range : iota;
+auto input = 2.iota;
+auto a3 = input.staticArray!2;
+auto a4 = input.staticArray!(byte[2]);
+
+auto a5 = staticArray!(2.iota);
+auto a6 = staticArray!(double, 2.iota);
+---
--- a/std/array.d
+++ b/std/array.d
@@ -4033,3 +4033,218 @@ unittest
     assert(appS.data == "hellow");
     assert(appA.data == "hellow");
 }
+
+/++
+Constructs a static array from `a`.
+The type of elements can be specified implicitly (`[1,2].staticArray` of type int[2])
+or explicitly (`[1,2].staticArray!float` of type float[2]).
+When `a` is a range (not known at compile time), the number of elements has to be given as template argument
+(e.g. `myrange.staticArray!2`).
+Size and type can be combined (eg: `2.iota.staticArray!(byte[2])`).
+When the range `a` is known at compile time, it can also be specified as a
+template argument to avoid having to specify the number of elements
+(eg: `staticArray!(2.iota)` or `staticArray!(double, 2.iota)`).
+
+Note: `staticArray` returns by value, so expressions involving large arrays may be inefficient.
+
+Params: a = The input elements
+
+Returns: A static array constructed from `a`.
++/
+
+pragma(inline, true) T[n] staticArray(T, size_t n)(auto ref T[n] a) nothrow @safe pure @nogc
+{
+    return a;
+}
+
+/// ditto
+U[n] staticArray(U, T, size_t n)(auto ref T[n] a) nothrow
+if (!is(U == T))
+{
+    import std.conv : emplaceRef;
+
+    U[n] ret = void;
+    // TODO: ElementType vs ForeachType
+    static foreach (i; 0 .. n)
+    {
+        emplaceRef!U(ret[i], cast(U) a[i]);
+    }
+    return ret;
+}
+
+/// static array from array
+nothrow pure @safe unittest
+{
+    auto a = [0, 1].staticArray;
+    static assert(is(typeof(a) == int[2]));
+    assert(a == [0, 1]);
+
+    auto b = [0, 1].staticArray!byte;
+    static assert(is(typeof(b) == byte[2]));
+    assert(b == [0, 1]);
+}
+
+nothrow pure @safe unittest
+{
+    int val = 3;
+    static immutable gold = [1, 2, 3];
+    [1, 2, val].staticArray.checkStaticArray!int([1, 2, 3]);
+
+    @nogc void checkNogc()
+    {
+        [1, 2, val].staticArray.checkStaticArray!int(gold);
+    }
+
+    checkNogc();
+
+    [1, 2, val].staticArray!double.checkStaticArray!double(gold);
+    [1, 2, 3].staticArray!int.checkStaticArray!int(gold);
+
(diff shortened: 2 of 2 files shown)


dukc:
Circleci crashed. Trying to restart...

dukc:
Anybody know how I can fix the Circleci seqfault?

std/array.d
```diff
@@ -3980,3 +3980,227 @@ unittest
...
+/// ditto
+auto staticArray(size_t n, T)(T a)
+if (isInputRange!T)
+{
+    alias U = ElementType!T;
+    return staticArray!(U[n], U, n)(a);
+}
+
+/// ditto
+auto staticArray(size_t n, T)(T a, out size_t rangeLength)
```
wilzbach: Do we need `rangeLength`? (i.e. is there any good use case?)

std/array.d
```diff
@@ -3980,3 +3980,227 @@ unittest
...
+        return (() @trusted => cast(U[n]) ret)();
+    }
+
+    auto ret = (() @trusted
+    {
+        Unqual!U[n] theArray = void;
+        return theArray;
+    }());
+
+    for (auto iter = a.take(n); !iter.empty; iter.popFront())
```
wilzbach: Why not `foreach`?

std/array.d
```diff
@@ -3980,3 +3980,227 @@ unittest
...
+    import std.range : take;
+    import std.conv : emplaceRef;
+
+    size_t i;
+
+    if (__ctfe)
+    {
+        // Compile-time version to avoid unchecked memory access.
+        Unqual!U[n] ret;
+        for (auto iter = a.take(n); !iter.empty; iter.popFront())
```
wilzbach: Why not `foreach`?

std/array.d
```diff
@@ -3980,3 +3980,227 @@ unittest
...
+        {
+            ret[i] = iter.front;
+            i++;
+        }
+
+        rangeLength = i;
+        return (() @trusted => cast(U[n]) ret)();
+    }
+
+    auto ret = (() @trusted
```
wilzbach: Does this need to be `@trusted`?

std/array.d
```diff
@@ -3980,3 +3980,227 @@ unittest
...
+
+Params:
+    a = The input elements. If there are less elements than the specified length of the static array,
+    the rest of it is default-initialized. If there are more than specified, the first elements
+    up to the specified length are used.
+    rangeLength = outputs the number of elements used from `a` to it. Optional.
+
+Returns: A static array constructed from `a`.
++/
+
```
wilzbach: Nit: we typically don't have an empty line here.

std/array.d
dukc: Andrei wanted that feature in #6214 review, and becuse it's so similar I assume the same wish applies here too.

std/array.d
dukc: It was a ref foreach in #6178, but Andrei pointed out that it woud not work with rvalue elements, and suggested this. Would foreach without ref be better? I did not think.

std/array.d
```diff
@@ -3980,3 +3980,226 @@ unittest
...
+/// static array from array literal
+nothrow pure @safe unittest
+{
+    auto a = [0, 1].staticArray;
+    static assert(is(typeof(a) == int[2]));
+    assert(a == [0, 1]);
+}
+
+pragma(inline, true) U[n] staticArray(U, T, size_t n)(auto ref T[n] a) nothrow @safe pure @nogc
+if (!is(T == U) && is(T : U))
```
andralex: Do we need the first test? `is(T : U)` should be sufficient.

std/array.d
```diff
@@ -3980,3 +3980,226 @@ unittest
...
+
+    auto ret = (() @trusted
+    {
+        Unqual!U[n] theArray = void;
+        return theArray;
+    }());
+
+    for (auto iter = a.take(n); !iter.empty; iter.popFront())
+    {
+        emplaceRef!U(ret[i++], iter.front);
```
andralex: If an exception is thrown, we're toast - uninitialized values will be incorrectly destroyed. See @timotheecour 's PR where I posted code that can be used to fix that.

std/array.d
dukc: It is. Tried without, there was ambiquity with the first overload in this pr.

dukc:
@andralex exceptions and destructor thing addressed.

Good catch btw! I think no-one else noticed.

std/array.d
andralex: Oh, indeed. I missed the negation on the first `is`...

std/array.d
```diff
@@ -3980,3 +3980,232 @@ unittest
...
+auto staticArray(size_t n, T)(T a, out size_t rangeLength)
+if (isInputRange!T)
+{
+    alias U = ElementType!T;
+    return staticArray!(U[n], U, n)(a, rangeLength);
+}
+
+/// ditto
+auto staticArray(Un : U[n], U, size_t n, T)(T a) nothrow
+if ((isInputRange!T) && is(ElementType!T : U))
```
andralex: redundant parens

std/array.d
```diff
@@ -3980,3 +3980,232 @@ unittest
...
+/// ditto
+auto staticArray(size_t n, T)(T a, out size_t rangeLength)
+if (isInputRange!T)
+{
+    alias U = ElementType!T;
+    return staticArray!(U[n], U, n)(a, rangeLength);
+}
+
+/// ditto
+auto staticArray(Un : U[n], U, size_t n, T)(T a) nothrow
```
andralex: wait can't this throw if copying objects throws?

std/array.d
```diff
@@ -3980,3 +3980,232 @@ unittest
...
+if ((isInputRange!T) && is(ElementType!T : U))
+{
+    size_t extraStackSpace;
+    return staticArray!(Un, U, n)(a, extraStackSpace);
+}
+
+/// ditto
+auto staticArray(Un : U[n], U, size_t n, T)(T a, out size_t rangeLength)
+if ((isInputRange!T) && is(ElementType!T : U))
```
andralex: redundant parens

std/array.d
dukc: Indeed. That's a leftover from Timothee. I think I also need to add a test to prevent errors like this.

dukc:
@andralex  "Looked" once more.

dukc:
@wilzbach CircleCI fails to compile but my local version and the autotester does not. What is happening?

dukc:
I think the issue is that `object._ArrayDtor!(CopyPreventer)` does not have a scope attribute in it's parameter. It's either in DRuntime or dmd itself, so fixing it here is out of question. I just wrapped the function call complained about in `@trusted`. I try to produce a test case so I can file a bugzilla Issue to refer to in comment.

Otherwise, ready for review! @andralex

std/array.d
```diff
@@ -3980,3 +3980,270 @@ unittest
...
+        return (() @trusted => cast(U[n]) ret)();
+    }
+
+    auto ret = (() @trusted
+    {
+        Unqual!U[n] theArray = void;
+        return theArray;
+    }());
+
+    if (true)
```
klickverbot: `if (true)`?

std/array.d
```diff
@@ -3980,3 +3980,270 @@ unittest
...
+
+/// ditto
+auto staticArray(Un : U[n], U, size_t n, T)(scope T a, out size_t rangeLength)
+if (isInputRange!T && is(ElementType!T : U))
+{
+    import std.algorithm.mutation : uninitializedFill;
+    import std.range : take;
+    import std.conv : emplaceRef;
+
+    size_t i;
```
klickverbot: I'd declare these `i`s locally next to the loops to make it clearer what the scope is. Stylistic choice, though.

changelog/std-array-asStatic.md
```diff
@@ -0,0 +1,25 @@
...
+or explicitly ($(D [1, 2].staticArray!float) of type `float[2]`).
+When `a` is a range with length not known at compile time, the number of elements must be given as template argument
+(eg `myrange.staticArray!2`).
+Size and type can be combined, if source range elements are implicitly
+convertible to the requested element type (eg: `2.iota.staticArray!(long[2])`).
+When the range `a` is known at compile time, it can also be specified as a
+template argument to avoid having to specify the number of elements
+(eg: `staticArray!(2.iota)` or `staticArray!(double, 2.iota)`).
+
+Note: $(D fun([1, 2, 3].staticArray)) may be inefficient because of the copies involved.
```
klickverbot: Shouldn't we be able to make sure that RVO happens?

std/array.d
```diff
@@ -3980,3 +3980,270 @@ unittest
...
+}
+
+// Tests that code compiles when there is an elaborate destructor and exceptions
+// are thrown. Unfortunately can't test that memory is initialized
+// before having a destructor called on it.
+// @system required because of issue 18872.
+@system nothrow unittest
+{
+    // exists only to allow doing something in the destructor
+    static int preventersDestroyed = 0;
```
klickverbot: Shouldn't this be compared against the expected value?

std/array.d
```diff
@@ -3980,3 +3980,270 @@ unittest
...
+    {
+        auto staticArray = normalArray.staticArray!5;
+    }
+    catch (Exception e){}
+}
+
+
+nothrow pure @safe unittest
+{
+
```
klickverbot: Superfluous whitespace.

std/array.d
```diff
@@ -3980,3 +3980,270 @@ unittest
...
+    assert(copiedAmount == 2);
+}
+
+/// ditto
+auto staticArray(alias a)()
+if (isInputRange!(typeof(a)))
+{
+    // NOTE: without `cast`, getting error:
+    // cannot deduce function from argument types !(length)(Result)
+    return .staticArray!(cast(size_t) a.length)(a);
```
klickverbot: Can this be `size_t(a.length)` to avoid accidentally introducing bugs?

std/array.d
dukc: It's for the scope(exit) inside it.

changelog/std-array-asStatic.md
dukc: Ask Walter, not me. Because the comment was here already I thought there may be a reason for it and decided to leave it there..

std/array.d
dukc: I originally did, but it was larger than I excepted. That tells me that the value is implementatation-dependant and thus better to not be tested.

dukc:
Thanks for the comments. All explained and/or fixed.

std/array.d
dukc: However I decided to add `assert (false)` after the static array initializer to ensure the copy preventing postblit works.

dukc:
@andralex  Tests pass and it's ready for another try.

dukc:
ping @andralex

changelog/std-array-asStatic.md
wilzbach: I would remove the note from the changelog as I think it's clear that in the general case static arrays need to be copied and such details belong into the detailed documentation of the function.

See also: https://godbolt.org/g/GEsh3E

std/array.d
```diff
@@ -3980,3 +3980,270 @@ unittest
...
+    [1, 2, 3].staticArray!int.checkStaticArray!int(gold);
+
+    [1, 2, 3].staticArray!(const(int)).checkStaticArray!(const(int))(gold);
+    [1, 2, 3].staticArray!(const(double)).checkStaticArray!(const(double))(gold);
+    {
+        const(int)[3] a2 = [1, 2, 3].staticArray;
+    }
+
+    [cast(byte) 1, cast(byte) 129].staticArray.checkStaticArray!byte([1, -127]);
+
```
wilzbach: unnecessary empty line.

std/array.d
```diff
@@ -3980,3 +3980,270 @@ unittest
...
+/// ditto
+auto staticArray(size_t n, T)(scope T a)
+if (isInputRange!T)
+{
+    alias U = ElementType!T;
+    return staticArray!(U[n], U, n)(a);
+}
+
+/// ditto
+auto staticArray(size_t n, T)(scope T a, out size_t rangeLength)
```
wilzbach: I'm still not sure whether it's a good idea to expose the `rangeLength` API if there's no good use case for it. Or did you manage to find one?

dukc: I believe it was wanted for cases like reading input to a fixed-size buffer. So that one can get the length of the input actually in the buffer without using `tee` or `enumerate`.

changelog/std-array-asStatic.md
dukc: Oh, i didn't notice that static array was being fed to a function. I thought he meant that [1, 2, 3] might be unnecessarily copied when initializing. Sounded unlikely, but I thought that Timothee might know something I do not.

But since it means that feeding a static array to a function involves a copy, I will remove that.

But looking at that godbolt example, why does LDC make `bar` to copy the static array when passing it to `foo` when it does not use it for anything else? Can't it rely on `foo` to copy it if it mutates it?

dukc:
Okay, addressed.

dukc:
ping @andralex

std/array.d
wilzbach: Ah right. Forgot that using `void` makes every unsafe. Sorry.

dukc:
@wilzbach Docs corrected. But I don't want to squash, because I don't know how to squash commits by others. I don't want to accidently remove all credit to Timothee!

I let you to do that, if you want.

wilzbach:
> But I don't want to squash, because I don't know how to squash commits by others. I don't want to accidently remove all credit to Timothee!
> I let you to squash them, if you want. But please do not remove Timothee.

Okay squashed them down to two commits (i.e. all your commits into one and commits of Timothee into one commit).

wilzbach:
Auto-merge toggled on

PR #6491, opened 2018-05-02
Only incorporate VariantN doc tests in doc builds.

dhasenan:
The tests are for documentation and are located inside a template. This means they'll run every time you instantiate the VariantN template.

The unfortunate effect of this commit is that the documentation tests will not be executed, allowing them to get out of date. The obvious way to deal with this is to require every documentation test to run and to add `-version=StdDdoc` when compiling for the test target.

### DIFF
--- a/std/variant.d
+++ b/std/variant.d
@@ -705,7 +705,7 @@ public:
     }
 
     ///
-    version(unittest)
+    version(StdDdoc)
     @system unittest
     {
         Variant a;
@@ -738,7 +738,7 @@ public:
     }
 
     ///
-    version(unittest)
+    version(StdDdoc)
     @system unittest
     {
         Variant a = 5;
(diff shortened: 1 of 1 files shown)


wilzbach:
`StdDdoc` isn't set when the test are run on the auto-tester which essentially means they are disabled, so a copy of these tests probably needs to be moved outside to ensure that no regressions occur...

dhasenan:
The problem with merely copying those tests out is that they're not tested. Should we have another version flag for documentation tests, one that's enabled for both documentation and unittest builds?

wilzbach:
An alternative would be to pick a type list that's unique and `static if` the tests.
An example from stdx.checkedint:

```d
static if (is(T == int) && is(Hook == void)) @system unittest
```

thewilsonator:
This is in upstream

PR #6492, opened 2018-05-02
Improve joiner performance for random-access separators

wilzbach:
As mentioned in Jon's talk today, `joiner` is __really slow__.
This small improvement that was laying around for far too long.

```
std.joiner      = 5 secs, 180 ms, 193 μs, and 7 hnsecs
std.joiner.alias = 5 secs, 106 ms, and 2 μs
new.joiner      = 3 secs, 168 ms, 560 μs, and 6 hnsecs
foreach         = 2 secs, 537 ms, 962 μs, and 8 hnsecs
```
(ldc 1.9.0)

Benchmark code: https://gist.github.com/wilzbach/ffd5d20639766a831fd4b1962572897a

FWIW dmd is playing in a different universe here:

```
> dmd -O -inline -release ./joiner.d && ./joiner
std.joiner      = 57 secs, 834 ms, 289 μs, and 3 hnsecs
std.joiner.alias = 54 secs, 836 ms, 908 μs, and 6 hnsecs
new.joiner      = 44 secs, 936 ms, 706 μs, and 5 hnsecs
foreach         = 23 secs, 732 ms, 820 μs, and 3 hnsecs
```

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -2161,18 +2161,66 @@ if (isInputRange!RoR && isInputRange!(ElementType!RoR)
     {
         private RoR _items;
         private ElementType!RoR _current;
-        private Separator _sep, _currentSep;
+        static if (isRandomAccessRange!Separator)
+        {
+            static struct CurrentSep
+            {
+                private Separator _sep;
+                private size_t sepIndex;
+                private size_t sepLength; // cache the length for performance
+                auto front() { return _sep[sepIndex]; }
+                void popFront() { sepIndex++; }
+                auto empty() { return sepIndex >= sepLength; }
+                auto save()
+                {
+                    auto copy = this;
+                    copy._sep = _sep;
+                    return copy;
+                }
+                void reset()
+                {
+                    sepIndex = 0;
+                }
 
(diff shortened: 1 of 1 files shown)


wilzbach:
BTW I think @jondegenhardt who discovered this might be interested in this.

(The issue is https://issues.dlang.org/buglist.cgi?quicksearch=joiner&list_id=221332, but I didn't reference it here because there's still a quite a difference between the naive loop and joiner.)

wilzbach:
Thanks for the review and feedback @jondegenhardt!
Finally got around addressing the minor things.

jondegenhardt:
@wilzbach I don't see this in the 2.081.0 Beta-1 release notes. Seems worthy of inclusion, even if there is more to be gained.

wilzbach:
It's part of the 2.081 Beta-1 release:

image

However, it doesn't show up in the release notes because I didn't reference [18643](https://issues.dlang.org/show_bug.cgi?id=18463) as fixed by this PR because as you said yourself there's still more to gain.
It probably makes sense to add a special changelog entry anyway as it's "free advertisement" and showing that we slowly fix our issues (will do so tonight).
Thanks a lot for the ping and reminder!

PR #6493, opened 2018-05-03
Fix Issue 18822 - Compiling byGrapheme Fails

JackStouffer:
So, what's happening is that the `toString` of `Appender` now uses `formatValue`, which calls `formatRange`, which calls `graphemeStride`. `Grapheme` uses `TrieBuilder`, which uses `std.conv.text` for an error message which uses `Appender`, and you can see the problem.

If the `Appender.toString` code wasn't already released, I'd suggest removing the format specifier from it, as it would be a much cleaner fix. My guess is that other such return inference issues exist that we can't detect.

The other troubling issue is that the test suite didn't pick this up. The error only appears when compiling a file outside of the Phobos directory. Which strengthens my hypothesis about hidden inference problems.

### DIFF
--- a/std/uni.d
+++ b/std/uni.d
@@ -4086,9 +4086,8 @@ public:
     */
     void putValue(Key key, Value v)
     {
-        import std.conv : text;
         auto idx = getIndex(key);
-        enforce(idx >= curIndex, text(errMsg, " ", idx));
+        enforce(idx >= curIndex, errMsg);
         putAt(idx, v);
     }
 


JackStouffer:
Even more troubling than the auto-tester is that the project tester didn't pick this up either :/

std/uni.d
```diff
@@ -4086,9 +4086,8 @@ public:
     */
     void putValue(Key key, Value v)
     {
-        import std.conv : text;
         auto idx = getIndex(key);
-        enforce(idx >= curIndex, text(errMsg, " ", idx));
+        enforce(idx >= curIndex, errMsg);
```
wilzbach: Can't we do `errMsg ~ " " ~ idx`?
It's `lazy` after all :)

JackStouffer: idx is a integer.

JackStouffer:
Pinging random reviewers as this is a critical bug fix and should probably warrant it's own point release

@wilzbach @ZombineDev @n8sh @quickfur

wilzbach:
> The other troubling issue is that the test suite didn't pick this up. The error only appears when compiling a file outside of the Phobos directory. Which strengthens my hypothesis about hidden inference problems.

In which project did you encounter this? Maybe it's enough to add the project to the project tester?

PR #6494, opened 2018-05-04
fix issue 13741 - std.traits.moduleName does not work with functions that have parameters

BBasile:
~~The problem was that the template accessed the symbol using `T.stringof` instead of `__traits(identifier, T)`. I use the FQN which doesn't have this issue and decompose it until it can be used to mix an import statement.~~

_This description is out of date, see beyond_

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -489,12 +489,14 @@ template packageName(alias T)
 {
     import std.algorithm.searching : startsWith;
 
+    enum bool isNotFunc = !isSomeFunction!(T);
+
     static if (__traits(compiles, parentOf!T))
         enum parent = packageName!(parentOf!T);
     else
         enum string parent = null;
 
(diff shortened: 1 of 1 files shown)


BBasile:
There might other bugs caused by the same usage of `.stringof` on template parameters. The question is then rather if something can be done in DMD. Relevant issue is [17546](https://issues.dlang.org/show_bug.cgi?id=17546)

MetaLang:
I don't like how this turned a template with a constant number of expansions into N expansions, even though N will almost always be small.

BBasile:
@MetaLang, i 've made a much better fix. It doesn't lead to N new template instances. Just one to prevent `T.stringof` on functions, which was never required anyway, since `T.stringof` is just used to detected the word `package` and `module`.

PR #3786, opened 2015-10-30
Add isInputRangeOf template constraint

atilaneves:
Idiomatic D uses input ranges where possible. Right now however, transiotining a function from taking `T[]` to a range is more cumbersome than necessary. This PR introduces a new template contraint so that:

    void func(int[] numbers) { /*...*/ }

Can easily be converted to 
    
    void func(R)(R range) if(isInputRangeOf!(R, int)) { /*...*/ }

Instead of manually typing out

    void func(R)(R range) if(isInputRange!R && is(ElementType!R == int)) { /*...*/ }

PR #6495, opened 2018-05-05
Optional element type parameter to isInputRange

atilaneves:
Optional element type parameter to `isInputRange`.

### DIFF
--- a/std/csv.d
+++ b/std/csv.d
@@ -326,7 +326,7 @@ Throws:
 */
 auto csvReader(Contents = string,Malformed ErrorLevel = Malformed.throwException, Range, Separator = char)(Range input,
                  Separator delimiter = ',', Separator quote = '"')
-    if (isInputRange!(Range, char)
+    if (isInputRange!(Range, dchar)
     && isSomeChar!(Separator)
     && !is(Contents T : T[U], U : string))
 {
--- a/std/csv.d
+++ b/std/csv.d
@@ -326,7 +326,7 @@ Throws:
 */
 auto csvReader(Contents = string,Malformed ErrorLevel = Malformed.throwException, Range, Separator = char)(Range input,
                  Separator delimiter = ',', Separator quote = '"')
-if (isInputRange!Range && is(Unqual!(ElementType!Range) == dchar)
+    if (isInputRange!(Range, char)
     && isSomeChar!(Separator)
     && !is(Contents T : T[U], U : string))
 {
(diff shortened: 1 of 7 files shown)


std/csv.d
```diff
@@ -326,7 +326,7 @@ Throws:
 */
 auto csvReader(Contents = string,Malformed ErrorLevel = Malformed.throwException, Range, Separator = char)(Range input,
                  Separator delimiter = ',', Separator quote = '"')
-if (isInputRange!Range && is(Unqual!(ElementType!Range) == dchar)
+    if (isInputRange!(Range, dchar)
```
wilzbach: > Constraints on declarations should have the same indentation level as their declaration:

https://dlang.org/dstyle.html#phobos_declarations

std/range/primitives.d
```diff
@@ -161,16 +161,19 @@ See_Also:
...
  */
-enum bool isInputRange(R) =
+enum bool isInputRange(R, E = void) =
     is(typeof(R.init) == R)
     && is(ReturnType!((R r) => r.empty) == bool)
     && is(typeof((return ref R r) => r.front))
     && !is(ReturnType!((R r) => r.front) == void)
-    && is(typeof((R r) => r.popFront));
+    && is(typeof((R r) => r.popFront))
+    && (is(E == void) || is(Unqual!(ElementType!R) == E));
```
n8sh: Why `is(Unqual!(ElementType!R) == E))` instead of `is(R : E)` (like in #3786)? Is there an example of where this way does the right thing and the other way does the wrong thing?

jmdavis: Implicit conversions are generally a terrible idea in generic code. At minimum, if a template constraint accepts an implicit conversion, then it needs to force the implicit conversion, or you risk bugs - some of which result in compile-time errors, and some of which just result in weird, buggy behavior (e.g. if some of the function uses the object as-is, and other parts trigger the implicit conversion, which can happen easily if some of the stuff done to the object works without the conversion and other stuff only works if the conversion is done). So, having a commonly-used trait use `:` instead of `==` is just begging for problems. Anyone using `:` needs to know that that's what they're doing and program accordingly and _not_ have that be the default.

Using `Unqual` is already potentially risky. It's needed so often that it's probably okay, but it's not hard to write code that works with mutable and not other levels of constness such that just letting `const` and `immutable` elements through will cause problems some of the time - though at least in those cases, it's virtually guaranteed to be a compilation error. So, the risk of problems is probably outweighed by the usability gained by then working in all of the cases where you don't care about constness.

But `:` should be used in template constraints with extreme caution.

atilaneves:
@JackStouffer Not quite - that PR added a new name. This one just adds functionality to `isInputRange`. It behaves as before when no explicit T is used.

PR #6496, opened 2018-05-05
fix Issue 13507 - std.range.enumerate with BigInt indexes too

Voultapher:
Seems to work with modified existing unit test. Didn't find a good way to test enumerate range with a `BigInt` worthy size.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -229,6 +229,7 @@ Authors: $(HTTP erdani.com, Andrei Alexandrescu), David Simcha,
 module std.range;
 
 public import std.array;
+public import std.bigint : BigInt;
 public import std.range.interfaces;
 public import std.range.primitives;
 public import std.typecons : Flag, Yes, No;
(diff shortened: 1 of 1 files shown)


Voultapher:
As discussed with @wilzbach, the feature could be implemented, in a minimal working way. Yet, trying to test it revealed ergonomic issues with other library components. For example, getting an enumerate range to a point where `BigInt` is necessary, should be possible with:

```d
const auto upperBound = BigInt("1_000_000_000_000_000_000_000");
auto range = iota(upperBound).enumerate!BigInt;
range.popFrontN(upperBound - 10);
```

For this to work, `iota!BigInt` would have to return a range than has a length and can be sliced. However `hasLength` is defined and constrained to a length function that returns `size_t`, and this is only one of the problems faced. Short of changing vast parts of phobos or duplicating a lot of code to support `BigInt`, I don't see a way to implement this ergonomically.

MetaLang:
`However hasLength is defined and constrained to a length function that returns size_t, and this is only one of the problems faced.`

I've got an idea for how to fix `hasLength`, but what are the other problems? If you can list them we can go through one by one and come up with a solution.

jmdavis:
> I've got an idea for how to fix hasLength, but what are the other problems? If you can list them we can go through one by one and come up with a solution.

If you want to change `hasLength`, you're going to need @andralex's approval. After repeated discussions on the topic (and after reverting changes in some cases), we've stuck with `size_t` for `length`, because allowing anything else for `length` complicates life considerably. The fact that `size_t` is explicitly `size_t` and not anything else is very much on purpose.

Voultapher:
This was tagged as a Bootcamp issue, now it has evolved into a wider design discussion. I'd be glad if someone more experienced with the codebase could take over.

Voultapher:
This has not seen any progress in years so I'm closing it.

PR #6497, opened 2018-05-05
Fix issue 15732 - std.function partial does not work with function / delegate references

Zevenberge:
Flipped the static if statements in partial. It now checks whether we get a template or a struct/class with opCall overloaded. Else it assumes we get a function.

The branches in static ifs are to give better compile errors.

Bugzilla #15732

### DIFF
--- a/std/functional.d
+++ b/std/functional.d
@@ -716,15 +716,9 @@ Returns:
  */
 template partial(alias fun, alias arg)
 {
-    static if (is(typeof(fun) == delegate) || is(typeof(fun) == function))
-    {
-        import std.traits : ReturnType;
-        ReturnType!fun partial(Parameters!fun[1..$] args2)
-        {
-            return fun(arg, args2);
-        }
-    }
-    else
+    import std.traits : isCallable;
+    // Check whether fun is a user defined type which implements opCall or a template.
+    static if (is(typeof(fun) == struct) || is(typeof(fun) == class) || __traits(isTemplate, fun))
     {
         auto partial(Ts...)(Ts args2)
         {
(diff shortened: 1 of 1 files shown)


wilzbach:
Hey thanks a lot for your contribution.
Could you please reference the Bugzilla issue, s.t. it gets mentioned in the changelog and auto-closed?
(see e.g. https://github.com/dlang-bots/dlang-bot#automated-references for instructions)

Also mind the CI failure about trailing whitespace.

Zevenberge:
I think I resolved all the issues that D-Bot and CI highlighted.

std/functional.d
```diff
@@ -716,15 +716,9 @@ Returns:
...
-        import std.traits : ReturnType;
-        ReturnType!fun partial(Parameters!fun[1..$] args2)
-        {
-            return fun(arg, args2);
-        }
-    }
-    else
+    import std.traits : isCallable;
+    // Check whether fun is a user defined type which implements opCall or a template (void).
+    static if (is(typeof(fun) == struct) || is(typeof(fun) == class) || is(typeof(fun) == void))
```
MetaLang: Instead of doing the check for void hack, there is now an [isTemplate](https://dlang.org/spec/traits.html#isTemplate) trait to check directly.

std/functional.d
```diff
@@ -747,6 +741,35 @@ template partial(alias fun, alias arg)
             }
         }
     }
+    else static if (!isCallable!fun)
+    {
+        static assert(false, "Cannot apply partial to a non-callable '" ~ fun.stringof ~ "'.");
+    }
+    else // Assume fun is callable and uniquely defined.
+    {
+        static if (Parameters!fun.length == 0)
```
MetaLang: I think this case is unnecessary. If they don't pass any arguments, just alias `partial` to `fun`.

Zevenberge: This check states that the function to be partially invoked should have arguments. I will update the error message to reflect this.
As far as I see there is no support for partially invoking functions with multiple partial arguments.

std/functional.d
```diff
@@ -716,15 +716,9 @@ Returns:
...
-        import std.traits : ReturnType;
-        ReturnType!fun partial(Parameters!fun[1..$] args2)
-        {
-            return fun(arg, args2);
-        }
-    }
-    else
+    import std.traits : isCallable;
+    // Check whether fun is a user defined type which implements opCall or a template.
+    static if (is(typeof(fun) == struct) || is(typeof(fun) == class) || __traits(isTemplate, fun))
```
MetaLang: Looks like I missed this on my first time through. If `fun` is a struct or class, there should also be a check that they are actually callable. In fact, your check for `isCallable!fun` should either be moved up as a static assert at the top of the template, or added as a template guard.

After you make these changes we can get this merged, finally.

Zevenberge: Sorry for taking so long. I experimented a bit and the submitted solution (check for the presence of opCall) is the only one that passes all existing tests and gives a compile error on types that are not callable. I'm still not 100% satisfied with the check, though, as cases that obviously should not compile (e.g. a class with opCall(int, int) being called with string as a first argument) still compile. However, checking that quickly escalates the complexity, which is probably why partial for functors is implemented with templates in the first place.

n8sh:
@MetaLang I think you misunderstood. That isn't checking the number of partial arguments but the number of arguments that `foo` is capable of accepting.

PR #6498, opened 2018-05-06
Fix issue 18832 — std.typecons.tuple comparison broken for floating point elements

Bolpat:
This code compiles:
```d
import std.typecons : tuple;
auto tup = tuple(double.nan);
assert(tup > tup);
assert(!(tup < tup));
```
The reason is that `Tuple`s `opCmp` assumes comparison on the fields are total.


Bolpat:
Reopening after figuring out the issues. Basically it's [this](https://forum.dlang.org/thread/wfkjpxpfsypvkuvtctzb@forum.dlang.org).

PR #6499, opened 2018-05-06
Fix issue 18832 — std.typecons.tuple comparison broken for floating point elements

Bolpat:
This code compiles:
```d
import std.typecons : tuple;
auto tup = tuple(double.nan);
assert(tup > tup);
assert(!(tup < tup));
```
The reason is that `Tuple`s `opCmp` assumes comparison on the fields are total.

(Cannot reopen #6498 due to GitHub [issue](https://github.com/isaacs/github/issues/361))

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -781,13 +781,13 @@ if (distinctFieldNames!(Specs))
         {
             Tuple!(int, string) t1 = tuple(1, "test");
             Tuple!(double, string) t2 =  tuple(1.0, "test");
-            //Ok, int can be compared with double and
-            //both have a value of 1
+            // Ok, int can be compared with double and
+            // both have a value of 1
             assert(t1 == t2);
         }
 
         /**
-         * Comparison for ordering.
+         * Lexical comparison for ordering.
          *
          * Params:
          *     rhs = The `Tuple` to compare against. It must meet the criteria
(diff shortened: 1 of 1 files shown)


Bolpat:
I don't know what's wrong with the linter. It says ```*** Error in `/home/circleci/phobos/generated/dscanner-032ac7e3ed5ea7df5e097badcfcd91d3cb8f18da/dsc': free(): invalid pointer: 0x00007ffff7fe4a00 ***``` Maybe it suffices to restart the CI.

Bolpat:
The linter should be revised. It rejects auto functions because of no return statement even if they contain a `mixin` statement.

wilzbach:
> The linter should be revised. It rejects auto functions because of no return statement even if they contain a mixin statement.

Are you referring to DScanner? Please open an issue with a testcase there then? Thanks!

std/typecons.d
```diff
@@ -805,48 +800,69 @@ if (distinctFieldNames!(Specs))
...
-
+        auto opCmp(R)(R rhs) { mixin(opCmpImpl); return 0; }
         /// ditto
-        int opCmp(R)(R rhs) const
-        if (areCompatibleTuples!(typeof(this), R, "<"))
-        {
-            foreach (i, Unused; Types)
-            {
-                if (field[i] != rhs.field[i])
+        auto opCmp(R)(R rhs) const { mixin(opCmpImpl); return 0; }
```
wilzbach: Should we have a constraint here? (it used to have one)

wilzbach: and what's the motivation for using a mixin? I don't see any reason

std/typecons.d
```diff
@@ -1190,136 +1206,139 @@ if (distinctFieldNames!(Specs))
...
+                        formatElement(writer, field[i], f);
+                    }
                 }
+                writer(footer);
+            }
+            else
+            {
+                throw new FormatException(
+                    "Expected '%s' or '%(...%)' or '%(...%|...%)' format specifier for type '" ~
+                        Unqual!(typeof(this)).stringof ~ "', not '%" ~ fmt.spec ~ "'.");
```
wilzbach: AFAICT this isn't related to this PR. Please move it into a separate PR. Thanks!

std/typecons.d
Bolpat: Yes, I forgot the constraint.
Rationale for the mixin: The functions cannot be unified (e.g. with `inout`), so this was an attempt to reduce non-trivial code duplication.

Bolpat:
I'll close this PR in favor of splitting the minor changes off into separate one.
It will be reopened and force pushed after the new PR is merged.

std/typecons.d
Bolpat: I'll do.

wilzbach:
Please squash instead of merging.

Bolpat:
I can do squash and rebase manually, but doesn't Github have the squash and merge feature any more? Some time ago, I learned that manual rebasing wasn't necessary any more.

wilzbach:
Yes it does, but people abused it here, which is we disabled it for dlang-bot

Bolpat:
Technically, there are two distinct changes: Fixing table in the documentation of `toString` and the fix itself. I realized the table is srewed up after the merge of #6502.

wilzbach:
> Technically, there are two distinct changes

Then please separate them in two PRs (or move the doc change fix in another PR). Thanks!

std/typecons.d
```diff
@@ -1328,6 +1349,48 @@ if (distinctFieldNames!(Specs))
...
+        static if (Types.length == 0)
+        @safe unittest
+        {
+            import std.format : format, FormatException;
+            import std.exception : assertThrown;
+
+            // enum tupStr = tuple(1, 1.0).toString; // toString is *impure* for double (issue 17628)
+            // static assert(tupStr == `Tuple!(int, double)(1, 1)`);
+
+            Tuple!(int, double)[3] tupList = [ tuple(1, 1.0), tuple(2, 4.0), tuple(3, 9.0) ];
```
wilzbach: Wasn't this already part of #6502?

Bolpat: It should have been. Now, it's in #6504.

std/typecons.d
```diff
@@ -800,48 +800,69 @@ if (distinctFieldNames!(Specs))
...
-                    return field[i] < rhs.field[i] ? -1 : 1;
-                }
-            }
-            return 0;
-        }
-
-        /// ditto
-        int opCmp(R)(R rhs) const
-        if (areCompatibleTuples!(typeof(this), R, "<"))
+        auto opCmp(this L, R)(R rhs)
```
wilzbach: Misses the template constraint

std/typecons.d
```diff
@@ -1185,6 +1206,7 @@ if (distinctFieldNames!(Specs))
             return h;
         }
 
+
```
wilzbach: Unrelated

std/typecons.d
Bolpat: Corrected.

std/typecons.d
Bolpat: Removed.

Bolpat:
There is more stuff to be done to fix the ordering.

wilzbach:
Why do you constantly keep closing this PR? That's more noise than keeping it open.

Bolpat:
For the last one: I thought I wouldn't be there some time.
I've added `isPartiallyOrdered` in `std.traits` in #6513. The issue boils down to D not having real support for partially ordered types (*partially* meaning *possibly non-total*, not the mathematical term).
@wilzbach Should I reopen now or when #6513 is merged (if it is not rejected)?

PR #6501, opened 2018-05-07
Remove template wide trusted in std.array

JackStouffer:
Pulls the call to `toUTF32` out of a call to `@trusted`. Reduces the scope of the trusted application.


schveiguy:
What I mean is that putting `@safe` right on it documents that it's safe. It always should be (as the only parameter allowed is a `char[]` or `wchar[]` array type), and the unittest makes sure it is, but that's not obvious from the documentation.

PR #6502, opened 2018-05-07
Typecons minor issues

Bolpat:
Fixes some minor issues.
I was asked to split this off from #6499.
Note that a large portion of the changes are white space only due to indentation.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -452,11 +452,6 @@ function. If `obj` is a `Tuple`, the individual members are
 accessible with the syntax `obj[0]` for the first field, `obj[1]`
 for the second, and so on.
 
-The choice of zero-based indexing instead of one-base indexing was
-motivated by the ability to use value tuples with various compile-time
-loop constructs (e.g. $(REF AliasSeq, std,meta) iteration), all of which use
-zero-based indexing.
-
 See_Also: $(LREF tuple).
 
 Params:
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -1204,6 +1204,10 @@ if (distinctFieldNames!(Specs))
         /**
          * Formats `Tuple` with either `%s`, `%(inner%)` or `%(inner%|sep%)`.
          *
+         * Params:
+         *     sink = A `char` accepting delegate
+         *     fmt = A $(REF FormatSpec, std,format)
+         *
          * $(TABLE2 Formats supported by Tuple,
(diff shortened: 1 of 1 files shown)


std/typecons.d
```diff
@@ -1204,6 +1204,11 @@ if (distinctFieldNames!(Specs))
         /**
          * Formats `Tuple` with either `%s`, `%(inner%)` or `%(inner%|sep%)`.
          *
+         * Params:
+         *     writer = A `char` accepting
```
wilzbach: Ehm that will be confusing to people because sink and writer are defined terms.
- sink = lambda output
- writer = output range

FYI: Since 2.079.0 toString supports OutputRange writers -> https://dlang.org/changelog/2.079.0.html#toString

tl;dr: keep the name

std/typecons.d
```diff
@@ -1212,6 +1217,8 @@ if (distinctFieldNames!(Specs))
          * $(TROW $(P `%(inner%|sep%)`), $(P The format `inner` is one format, that is applied
          *      on all fields of the `Tuple`. The inner format must be compatible to all
          *      of them.)))
+         *
+         * Example:
```
wilzbach: As you already touch this code, it would be a good idea to move the examples from text into an actual, tested unittest block.

Bolpat: Okay; that's trivial as the unittest block exists and the example is just duplication.

std/typecons.d
Bolpat: So, not change to output range `toString`, right?

wilzbach: Oh you can definitely add a `toString` overload that uses a writer output range, but just renaming the variable might be very confusing.

wilzbach:
Squashed the undo commit & left this part of the unittest at its original place, s.t. it doesn't show up in the public documentation:

```diff
-    // enum tupStr = tuple(1, 1.0).toString; // toString is *impure*.	
-    //static assert(tupStr == `Tuple!(int, double)(1, 1)`);
```

Bolpat:
Linter bullshits again:
`std/typecons.d(2437:9)[warn]: Left side of logical or is identical to right side.`
Is not true. Does restarting the CI help?

wilzbach:
> Does restarting the CI help?

Yep. BTW the issue for this is: https://github.com/dlang-community/D-Scanner/issues/593

PR #6504, opened 2018-05-08
Fix Misplaced Table

Bolpat:
The table explaining format specifiers for `Tuple` has been placed in the parameter explanation.
I moved the table so this is not the case.
The table contained unescaped commas, which are escaped now.

This is also an addition to #6502.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -1204,65 +1204,23 @@ if (distinctFieldNames!(Specs))
         /**
          * Formats `Tuple` with either `%s`, `%(inner%)` or `%(inner%|sep%)`.
          *
-         * Params:
-         *     sink = A `char` accepting delegate
-         *     fmt = A $(REF FormatSpec, std,format)
-         *
          * $(TABLE2 Formats supported by Tuple,
          * $(THEAD Format, Description)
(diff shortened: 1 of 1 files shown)


Bolpat:
I'd not ditto `string` returning and `delegate` taking overloads, as they do completely different things. It's easier to explain them separately, e.g. `string toString()()` doesn't care about format specifiers.

Sparate examples are a good idea. The example is older than executable examples, that's the reason.

wilzbach:
> I'll split it into a working and possible error section.

Thanks.

> @wilzbach the dlang-bot removed the auto-merge label.

Yep, I'm aware. (After all, I did programmed the bot to do so). It's a security feature, s.t. unreviewed code can't be sneaked into Phobos.
-> I re-applied auto-merge

ntrel:
@Bolpat:
> The table contained unescaped commas, which are escaped now.
> ```$(P The format `inner` is applied the expanded `Tuple`$(COMMA) so```

I don't think that is necessary, `P` is defined to accept multiple arguments:
```
P = $(T p, $0)
...
T=<$1>$+</$1>
```

PR #6507, opened 2018-05-12
Remove deprecated std.string funcs inPattern, squeeze, countChars, removeChars and munch

BBasile:


BBasile:
[The author of the project tester item that uses them is notified](https://github.com/PhilippeSigaud/Pegged/issues/246).

wilzbach:
IIRC pegged is no longer in development, so I am not sure whether this notification will help :/

BBasile:
> Sorry

No problem i didn't know it work like that. Just one think: the problem is that it leads to undocumented public symbols which our linter doesn't like much.

jmdavis:
> No problem i didn't know it works like that.

It's how it's been working, though there's been enough confusion over it that I'm probably going to need to change how that's handled so that the documentation indicates when the symbol is going to be completely removed, and then there can be a non-public comment indicating when the documentation is supposed to be removed. But I haven't switched over to that yet, and this was deprecated using the approach that we've been using for several years now. So, what's listed in the documentation is the date for removal from the docs, not the final removal date.

> the problem is that it leads to undocumented public symbols which our linter doesn't like much.

Are you talking about something in the circleci stuff? Well, it's going to need to be able to handle undocumented, deprecated symbols, because that's a standard part of the deprecation process. Symbols are deprecated and documented for one year, then deprecated and undocumented for a year, and then finally removed. So, they're going to sit there with no public documentation for about a year as a normal part of the process.

PR #6508, opened 2018-05-15
Fix Issue 18847 - std.allocator: Region uses .parent before it can be set

CyberShadow:
Also includes similar fixes for KKRegion and BitmappedBlock as per issue comment.

### DIFF
--- a/std/experimental/allocator/building_blocks/bitmapped_block.d
+++ b/std/experimental/allocator/building_blocks/bitmapped_block.d
@@ -156,7 +156,7 @@ private mixin template BitmappedBlockImpl(bool isShared, bool multiBlock)
         this(data);
     }
 
-    static if (!is(ParentAllocator == NullAllocator))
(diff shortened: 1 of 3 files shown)


edi33416:
CC @jercaianu

std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -165,14 +165,33 @@ private mixin template BitmappedBlockImpl(bool isShared, bool multiBlock)
         assert(_blocks * blockSize >= capacity);
     }
 
+    static if (!is(ParentAllocator == NullAllocator) && stateSize!ParentAllocator)
+    this(ParentAllocator parent, size_t capacity)
+    {
+        this.parent = parent;
+        size_t toAllocate = totalAllocation(capacity);
+        auto data = cast(ubyte[])(parent.allocate(toAllocate));
+        this(data);
```
wilzbach: Shouldn't there be `assert(_blocks * blockSize >= capacity);` too here?

PR #6511, opened 2018-05-16
std.math: Merge ARM and AArch64 versions into ARM_Any

ibuclaw:
All code around FPSCR, minus the inline assembler interacting with the register, is common between them.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -5588,7 +5588,7 @@ public:
             auto oldState = getControlState();
             // If exceptions are not supported, we set the bit but read it back as zero
             // https://sourceware.org/ml/libc-ports/2012-06/msg00091.html
-            setControlState(oldState | (divByZeroException & allExceptions));
+            setControlState(oldState | divByZeroException);
             immutable result = (getControlState() & allExceptions) != 0;
             setControlState(oldState);
             return result;


std/math.d
```diff
@@ -5495,22 +5493,6 @@ nothrow @nogc:
                                  | inexactException,
         }
     }
-    else version(ARM)
```
JackStouffer: These two branches are not equal. The first branch is missing the `subnormalException` definition.

ibuclaw: Someone doesn't read commit messages.

JackStouffer: Nope :)

joakim-noah:
In fact, it appears to be the converse, [these](http://docs.roguewave.com/totalview/2017.1/html/index.html#page/Reference_Guides/PowerFPSCRRegister_2.html) [pages](https://chromium.googlesource.com/android_tools/+/e91e44dc29517115794a72eea554616d067925ae/ndk/platforms/android-L/arch-arm64/usr/include/machine/fenv.h) note an IDE bit on AArch64, though strangely I can't find any of the ARMv8 exception control bits documented on ARM's website.

ibuclaw:
Hmm, I was looking in glibc and found no mention. Though now that I look at the diagram for FPCR/FPSR again.

http://repo.or.cz/glibc.git/blob/HEAD:/sysdeps/aarch64/fpu/fpu_control.h

I can see IDC and IDE documented (I initially wasn't reading it vertically), however they are always preserved/ignored (see reserved masks), presumably because IEEE standard takes precedence in glibc though.

ibuclaw:
> though strangely I can't find any of the ARMv8 exception control bits documented on ARM's website.

Indeed, it's only documented as reserved space.

ibuclaw:
@joakim-noah - just removed the first commit then.

std/math.d
```diff
@@ -5600,17 +5583,7 @@ public:
             return true;
         else version(MIPS_Any)
             return true;
-        else version(AArch64)
-        {
-            auto oldState = getControlState();
-            // If exceptions are not supported, we set the bit but read it back as zero
-            // https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/aarch64/fpu/feenablxcpth.c
-            setControlState(oldState | (divByZeroException & allExceptions));
```
joakim-noah: Two notes on this: `setControlstate` is not pure, so I had to pull in the required assembly in ldc-developers/phobos#65 to get this to build. Also, I don't see the need for `allExceptions` here, `divByZeroException` is all that's going to come out of the logical and.

ibuclaw: Actually, as per the concession in the language spec, `setControlState()` can be pure also. As the FPControl itself saves state in `initialize()` and clears and restores state in `~this()`.

ibuclaw: True about the `AND` thing, depends how much you want to change in one go, considering the one-pr, one-change rule.

joakim-noah: These are all static member functions though, and I'd rather just clean up the unnecessary `&` here too.

ibuclaw: Ah... I was only looking at the public members.  The private static functions can be pure also.  We _know_ that the internal implementation doesn't have any side effects.

joakim-noah: It is possible to call the static member function `setControlState()` with non-exception bits, in which case it must reset them before exiting to be considered `pure`, which it can't do because it would defeat its purpose. However, it must have the `pure` attribute to be called from `hasExceptionTraps()`, so I can't call it there. As you see with my linked pull, my workaround is not to use `setControlState()`, but to just pull in the ARM32/64 assembly it uses, which I know is being used in a `pure` way internally to `hasExceptionTraps`, at least according to the spec.

Anyway, since the main Phobos branch doesn't implement the ARM32/64 paths anyway, this all just a placeholder for now, so it's fine. We can figure it out if and when we ever upstream the actual ARM code.

PR #6512, opened 2018-05-16
std.math: Fix mask values of inexact and invalid on Mips

ibuclaw:
They were assigned to each other wrongly. FYI @kinke.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -5530,11 +5530,11 @@ nothrow @nogc:
     {
         enum : ExceptionMask
         {
-            inexactException      = 0x0800,
+            inexactException      = 0x0080,
             divByZeroException    = 0x0400,
             overflowException     = 0x0200,
             underflowException    = 0x0100,
-            invalidException      = 0x0080,
+            invalidException      = 0x0800,
             severeExceptions   = overflowException | divByZeroException
                                  | invalidException,
             allExceptions      = severeExceptions | underflowException


ibuclaw:
By the way, every time I get round to updating this, it always seems to take me half an hour or so to remember where these values come from.

Maybe we should demystify them somehow, either by improving documentation, assigning using the appropriate conversation, by moving the logic to the get/set control state functions (in which case these values become the same as fenv FE_XXX flags), or by joining the two enums in ieeeFlags and FPControl together so they are in one place private to std.math.

For mips, the fpcr masks are just the FE_XXX flags bit shifted by 5. Not all CPUs have such trivial mappings though.

Any thoughts or shall we just leave as-is?

JinShil:
I've never used MIPS, and very seldom use Phobos or these kinds of math APIs.  I'm also not on Team Phobos, but if I were reviewing this, I'd like to see a link to the source of this information so that I could verify it independently.  In terms of how to model the code, I always try to mimic how it is in the architecture's documentation so it is easy to cross-reference, but if you have several architectures in one file then you'll likely see inconsistencies depending on how the vendor chose to write their docs, so there may not be a winning formula.  IMO, it would be sufficient to add a link to the docs (page number/section number if it's a big PDF file) so one can verify it independently.

joakim-noah:
Like Mike, I really don't use `std.math`, just been porting some of it for ARM32/64. Looking back in git history, it says Don added both structs at the same time, with commit a8cb0ed more than 8 years back, not sure why as they seem to do similar things. Maybe one was for more low-level access while the other would make sure to reset the FPU state automatically?

You're right that these constants are replicated in `core.stdc.fenv` and here, and I see that these flags you fixed are set correctly there. I'm all for setting them once, rather than this repetition in two modules.

PR #6514, opened 2018-05-18
fix issue 18415 - Typedef ignores `@disabled` default constructor

BBasile:
The problem is only fixable for `struct` and `union` and not `class`, since `Typedef` is a `struct`.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -7118,6 +7118,12 @@ struct Typedef(T, T init = T.init, string cookie=null)
 {
     private T Typedef_payload = init;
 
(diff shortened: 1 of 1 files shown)


BBasile:
For a class that has its default `this` disabled you can still declare a variable without init. For a struct this is not the case. Since the `Typedef` of a `class` is a `struct` this would en up with this scenario: 

```d
class A{@disable this();}
alias TA  = Typedef!A;

A a; // ok;
TA ta; // compile error, because Typedef wraps in a struct with default construction disabled.
```

MetaLang:
Ah, makes perfect sense. Thanks @BBasile.

PR #6516, opened 2018-05-19
Fix Issue 18142 - checkedint opOpAssign doesn't accept a checkedint

andreiv05:
I added a template constraint to `opOpAssign` so that now can accept a `Checked` type as `Rhs`. Secondly, if `Rhs` is of type `Checked`, I call `opOpAssign` recursively with `rhs.get`, so that I can get to the value that `Checked` wraps up.

### DIFF
--- a/std/experimental/checkedint.d
+++ b/std/experimental/checkedint.d
@@ -916,20 +916,25 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
     Returns: A reference to `this`.
     */
     ref Checked opOpAssign(string op, Rhs)(const Rhs rhs) return
-    if (isIntegral!Rhs || isFloatingPoint!Rhs || is(Rhs == bool))
+    if (isIntegral!Rhs || isFloatingPoint!Rhs || is(Rhs == bool) ||
+        is(typeof(Checked!(Rhs, Hook)(rhs))))
     {
-        static assert(is(typeof(mixin("payload" ~ op ~ "=rhs")) == T));
-
-        static if (hasMember!(Hook, "hookOpOpAssign"))
+        static assert(is(typeof(mixin("payload" ~ op ~ "=rhs")) == T) ||
+            is(typeof(Checked!(Rhs, Hook)(rhs))));
+        static if (is(Rhs == Checked!(X, Y), X, Y))
+        {
+            return this.opOpAssign!op(rhs.get);
+        }
+        else static if (hasMember!(Hook, "hookOpOpAssign"))
         {
             hook.hookOpOpAssign!op(payload, rhs);
+            return this;
         }
         else
         {
             alias R = typeof(get + rhs);
             auto r = opBinary!op(rhs).get;
             import std.conv : unsigned;
-
             static if (ProperCompare.hookOpCmp(R.min, min.get) < 0 &&
                 hasMember!(Hook, "onLowerBound"))
             {
@@ -951,8 +956,8 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
                 }
             }
             payload = cast(T) r;
+            return this;
         }
-        return this;
     }
 
     ///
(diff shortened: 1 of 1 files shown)


BBasile:
Please, don't forget to test the changes locally before opening a PR. Sometimes it's possible to miss problems related to a specific platform but here it seems that you should have seen the warning, even with `-wi`. Testing a simple phobos change is often as simple as `dmd <themodule.d> -main -unittest` and then run the result.

johnsilver97:
Sorry about that, there was a case when `Hook` defined `hookOpOpAssign` that I haven't noticed, I think it should be fine now. Thanks for the help!

std/experimental/checkedint.d
```diff
@@ -980,6 +985,17 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
+    @system unittest
+    {
+        auto x1 = Checked!int(10);
+        auto x2 = Checked!int(10);
+        x1 += x2;
+        assert(x1.get == 20);
+        auto x3 = Checked!(Checked!int)(10);
+        x1 += x3;
+        assert(x1.get == 30);
+    }
```
wilzbach: Sadly this needs to be outside of the template as otherwise it would be instantiated for every template.
That's the reason for `static if (is(T == int) && is(Hook == void)) unittest` btw. 
Also we typically add a link to the respective bugzilla issue like:

```
// https://issues.dlang.org/show_bug.cgi?id=18142
unittest
...
```

std/experimental/checkedint.d
```diff
@@ -916,20 +916,25 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
         {
             hook.hookOpOpAssign!op(payload, rhs);
+            return this;
         }
         else
         {
             alias R = typeof(get + rhs);
             auto r = opBinary!op(rhs).get;
             import std.conv : unsigned;
-
```
wilzbach: Unrelated change. Please avoid.

std/experimental/checkedint.d
```diff
@@ -916,20 +916,25 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
-    if (isIntegral!Rhs || isFloatingPoint!Rhs || is(Rhs == bool))
+    if (isIntegral!Rhs || isFloatingPoint!Rhs || is(Rhs == bool) ||
+        is(typeof(Checked!(Rhs, Hook)(rhs))))
     {
-        static assert(is(typeof(mixin("payload" ~ op ~ "=rhs")) == T));
-
-        static if (hasMember!(Hook, "hookOpOpAssign"))
+        static assert(is(typeof(mixin("payload" ~ op ~ "=rhs")) == T) ||
+            is(typeof(Checked!(Rhs, Hook)(rhs))));
+        static if (is(Rhs == Checked!(X, Y), X, Y))
```
wilzbach: Why let the compiler do the infersion again?
You already know that its `Checked!(Rhs, Hook)`, no?

std/experimental/checkedint.d
```diff
@@ -980,6 +985,17 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
         x += 1;
         assert(MyHook.thereWereErrors);
     }
+
+    @system unittest
+    {
+        auto x1 = Checked!int(10);
+        auto x2 = Checked!int(10);
+        x1 += x2;
+        assert(x1.get == 20);
```
wilzbach: `.get` isn't necessary here.

std/experimental/checkedint.d
```diff
@@ -980,6 +985,17 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
+        auto x3 = Checked!(Checked!int)(10);
+        x1 += x3;
+        assert(x1.get == 30);
```
wilzbach: Would be great to add two more tests:
- different hook
- different type, but one that coerces to int (e.g. `short`)

RazvanN7:
The issue has been fixed by: https://github.com/dlang/phobos/pull/6685

PR #6517, opened 2018-05-20
Fix Issue 18683 - no toHash in rbtree

vlasebian:
I'm trying to implement `toHash` functionality for RedBlackTree. I used [this](https://github.com/n8sh/phobos/blob/307ac7e4628c0cf4bb7f4e6efdfa62076754a6f2/std/typecons.d#L1177) as a reference, hope it's not a problem.

Offtopic: it would be nice to have an example like the one [here](https://digitalmars.com/d/archives/digitalmars/D/learn/RedBlackTree_and_myClass_79072.html) on how to use rbtrees in the docs, it helped me a lot.

### DIFF
--- a/std/container/rbtree.d
+++ b/std/container/rbtree.d
@@ -1101,47 +1101,12 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
      */
     override size_t toHash() nothrow @safe
     {
-        import std.traits : hasMember;
-
-        auto thisRange = this[];
-
-        static if (hasMember!(Elem, "toHash"))
-        {
-            size_t hash = 0;
-
-            static if (is(T == class) || is(T == interface))
-            {
-                foreach (ref e; thisRange)
-                {
-                    hash += (e is null ? 0 : e.toHash()) + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
-                }
-            }
-            else
-            {
-                foreach (ref e; thisRange)
-                {
-                    hash += e.toHash() + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
-                }
-            }
-
-            return hash;
-        }
-        else
-        {
-            static size_t hashOf(T)(ref T objToHash)
-            {
-                return (() @trusted => typeid(objToHash).getHash(&objToHash))();
-            }
-
-            size_t hash = 0;
-
-            foreach (ref e; thisRange)
-            {
-                hash += hashOf(e) + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
-            }
-
-            return hash;
-        }
+        size_t hash = cast(size_t) 0x6b63_616c_4264_6552UL;
+        foreach (ref e; this[])
+            // As in boost::hash_combine
+            // https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine
+            hash += .hashOf(e) + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
+        return hash;
     }
 
     static if (doUnittest) @system unittest
--- a/std/container/rbtree.d
+++ b/std/container/rbtree.d
@@ -1095,6 +1095,167 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
     }
 
     /**
+     * Generates a hash for the tree. Note that with a custom comparison function
+     * it may not hold that if two rbtrees are equal, the hashes of the trees
+     * will be equal.
+     */
+    override size_t toHash() nothrow @safe
+    {
+        import std.traits : hasMember;
+
+        auto thisRange = this[];
+
+        static if (hasMember!(Elem, "toHash"))
+        {
+            size_t hash = 0;
+
+            static if (is(T == class) || is(T == interface))
+            {
+                foreach (ref e; thisRange)
+                {
+                    hash += (e is null ? 0 : e.toHash()) + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
+                }
+            }
+            else
+            {
+                foreach (ref e; thisRange)
+                {
+                    hash += e.toHash() + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
+                }
+            }
+
+            return hash;
+        }
+        else
+        {
+            static size_t hashOf(T)(ref T objToHash)
+            {
+                return (() @trusted => typeid(objToHash).getHash(&objToHash))();
+            }
+
+            size_t hash = 0;
+
+            foreach (ref e; thisRange)
+            {
+                hash += hashOf(e) + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
+            }
+
+            return hash;
+        }
+    }
+
+    static if (doUnittest) @system unittest
+    {
+        auto t1 = new RedBlackTree(1,2,3,4);
+        auto t2 = new RedBlackTree(1,2,3,4);
+        auto t3 = new RedBlackTree(1,2,3,5);
+        auto t4 = new RedBlackTree(1,2,3,4,5);
+
+        assert(t1.toHash() == t2.toHash);
+
+        assert(t1.toHash() != t3.toHash);
+        assert(t2.toHash() != t3.toHash);
+
+        assert(t3.toHash() != t4.toHash);
+        assert(t1.toHash() != t4.toHash);
+
+        // empty tree
+        auto t5 = new RedBlackTree();
+        auto t6 = new RedBlackTree();
+
+        assert(t5.toHash() == t6.toHash());
+
+        auto t7 = new RedBlackTree!string("red", "black");
+        auto t8 = new RedBlackTree!string("white", "black");
+        auto t9 = new RedBlackTree!string("red", "black");
+
+        assert(t7.toHash() == t9.toHash());
+        assert(t7.toHash() != t8.toHash());
+
+        static struct MyInt
+        {
+            int x;
+
+            this(int init_x)
+            {
+                x = init_x;
+            }
+
+            size_t toHash() const @safe nothrow
+            {
+                return typeid(x).getHash(&x) ^ 0xF0F0_F0F0;
+            }
+
+            int opCmp(const MyInt that) const
+            {
+                return (x > that.x) - (x < that.x);
+            }
+
+            bool opEquals(const MyInt that) const
+            {
+                return (this.x == that.x);
+            }
+        }
+
+        auto rbt1 = new RedBlackTree!MyInt(MyInt(1), MyInt(2), MyInt(3), MyInt(4));
+        auto rbt2 = new RedBlackTree!MyInt(MyInt(1), MyInt(2), MyInt(3), MyInt(4));
+
+        assert(rbt1.toHash() == rbt2.toHash());
+        assert(rbt1.toHash() != t1.toHash());
+
+        auto rbt3 = new RedBlackTree!MyInt(MyInt(4), MyInt(2), MyInt(3), MyInt(4));
+
+        assert(rbt1.toHash() != rbt3.toHash());
+
+        class MyInt2
+        {
+            int x;
+
+            this(int init_x)
+            {
+                x = init_x;
+            }
+
+            override size_t toHash() const @safe nothrow
+            {
+                return typeid(x).getHash(&x) ^ 0xF0F0_F0F0;
+            }
+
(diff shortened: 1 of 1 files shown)


std/container/rbtree.d
```diff
@@ -1094,6 +1094,110 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
         assert(t1 != o);  // pathological case, must not crash
     }
 
+    /**
+     * Generates a hash for a Red Black Tree
+     *
+     *
```
BBasile: This block comment is unnecessary big and quite pointless. 

> for a Red Black Tree

The reader knows that already, it's an overridden member function after all.

std/container/rbtree.d
```diff
@@ -1094,6 +1094,110 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
...
+     */
+    override size_t toHash() nothrow @safe
+    {
+        import std.traits : hasMember;
+
+        auto thisRange = this[];
+
+        static if (hasMember!(Elem, "toHash"))
+        {
+            size_t hash = 0;
```
BBasile: The initializer is useless (although the official position is to keep them for readability).

std/container/rbtree.d
```diff
@@ -1094,6 +1094,110 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
...
+     *
+     *
+     */
+    override size_t toHash() nothrow @safe
+    {
+        import std.traits : hasMember;
+
+        auto thisRange = this[];
+
+        static if (hasMember!(Elem, "toHash"))
```
n8sh: What's the reason for having this branch instead of always using `typeid(x).getHash(&x)`?

vlasebian: If the type of the elements in the rbtree implements `toHash` then this hashing function will be used in computing the hash of a node in the tree (I have written a unittest [here](https://github.com/dlang/phobos/pull/6517/files/63c9859cd2e024017a69a3d6aa82f1b33e7b8633#diff-5ce425da0dc54bc1f90d170bf3ae1d8fR1165) for this case).

std/container/rbtree.d
vlasebian: I fixed the comment.

std/container/rbtree.d
```diff
@@ -1094,6 +1094,108 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
...
+            foreach (ref e; thisRange)
+            {
+                hash += e.toHash() + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
+            }
+
+            return hash;
+        }
+        else
+        {
+            static size_t hashOf(T)(T objToHash)
```
n8sh: I think we want to pass the argument by reference.

std/container/rbtree.d
```diff
@@ -1094,6 +1094,108 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
...
+
+        auto thisRange = this[];
+
+        static if (hasMember!(Elem, "toHash"))
+        {
+            size_t hash = 0;
+
+            foreach (ref e; thisRange)
+            {
+                hash += e.toHash() + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
```
n8sh: If `is(T == class)` instead of `e.toHash()` we need to write `(e is null ? 0 : e.toHash())`, unless we know the tree disallows null elements.

vlasebian: How can I check if the tree disallows null elements? I have made the suggested changes, and I have written this test:
```
        class MyInt2
        {
            int x;

            this(int init_x)
            {
                x = init_x;
            }

            override size_t toHash() const @safe nothrow
            {
                return typeid(x).getHash(&x) ^ 0xF0F0_F0F0;
            }

            int opCmp(const MyInt2 that) const
            {
                return (x > that.x) - (x < that.x);
            }

            bool opEquals(const MyInt2 that) const
            {
                return (this.x == that.x);
            }
        }

        auto rbt4 = new RedBlackTree!MyInt2(null); // this is okay
        auto rbt5 = new RedBlackTree!MyInt2(new MyInt2(1), null, new MyInt2(3), null); // this generates a sigsev
```

The output is:
```
/tmp/dmd_run5P9L1u(+0x1eaa8c)[0x5642324f3a8c]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x110c0)[0x7f5ef63b10c0]
/tmp/dmd_run5P9L1u(+0x6a5d2)[0x5642323735d2]
/tmp/dmd_run5P9L1u(+0x67e37)[0x564232370e37]
/tmp/dmd_run5P9L1u(+0x6a56c)[0x56423237356c]
/tmp/dmd_run5P9L1u(+0x691ce)[0x5642323721ce]
/tmp/dmd_run5P9L1u(+0x51d60)[0x56423235ad60]
/tmp/dmd_run5P9L1u(+0x1da1ae)[0x5642324e31ae]
/tmp/dmd_run5P9L1u(+0x1eaae0)[0x5642324f3ae0]
/tmp/dmd_run5P9L1u(+0x1dc53f)[0x5642324e553f]
/tmp/dmd_run5P9L1u(+0x1e2c7e)[0x5642324ebc7e]
/tmp/dmd_run5P9L1u(+0x1e2e11)[0x5642324ebe11]
/tmp/dmd_run5P9L1u(+0x1e2c0a)[0x5642324ebc0a]
/tmp/dmd_run5P9L1u(+0x1dc516)[0x5642324e5516]
/tmp/dmd_run5P9L1u(+0x1ea8b6)[0x5642324f38b6]
/tmp/dmd_run5P9L1u(+0x1ddd85)[0x5642324e6d85]
/tmp/dmd_run5P9L1u(+0x1ddd0c)[0x5642324e6d0c]
/tmp/dmd_run5P9L1u(+0x1ddc77)[0x5642324e6c77]
/tmp/dmd_run5P9L1u(+0x1da9b4)[0x5642324e39b4]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf1)[0x7f5ef58fe2e1]
/tmp/dmd_run5P9L1u(+0x4d5ea)[0x5642323565ea]
Error: program killed by signal 11
posix.mak:399: recipe for target 'std/container/rbtree.test' failed
make: *** [std/container/rbtree.test] Error 1
```

n8sh: >How can I check if the tree disallows null elements?

In principle we could tell this by seeing if the comparison predicate accepted null values, but I think it's fine to just assume that if `is(T == class) || is(T == interface)` then `null` is a possibility.

>I have made the suggested changes
>Error: program killed by signal 11

It's segfaulting because you still need to do:
```d
hash += (e is null ? 0 : e.toHash()) + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
```
instead of 
```d
hash += e.toHash() + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
```
You don't need to do the same in the other branch.

std/container/rbtree.d
```diff
@@ -1094,6 +1094,108 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
...
+    /**
+     * Generates a hash for the tree
+     */
+    override size_t toHash() nothrow @safe
+    {
+        import std.traits : hasMember;
+
+        auto thisRange = this[];
+
+        static if (hasMember!(Elem, "toHash"))
```
n8sh: Also: this should be `static if (hasMember!(Elem, "toHash") || is(Elem == interface ))` so this will work with interfaces. Demonstration:

https://run.dlang.io/is/tt68Vd
```d
interface I {}
void main()
{
    static assert(__traits(hasMember, I, "toHash")); // Fails!
}
```

n8sh:
This PR is a great step forward and will be merged once it handles `null` objects gracefully. Some problems will remain though. With a custom comparison function it still may not hold that `if(a == b) a.toHash() == b.toHash()`. Consider:

https://run.dlang.io/is/iKD5BJ
```d
void main()
{
    import std.container.rbtree : RedBlackTree;
    static bool unsignedLess(int a, int b)
    {
        import std.math : abs;
        return abs(a) < abs(b);
    }
    
    auto tree1 = new RedBlackTree!(int, unsignedLess)(1, 2, 3, 4);
    auto tree2 = new RedBlackTree!(int, unsignedLess)(1, -2, 3, -4);
    assert(tree1 == tree2); // Passes because unsignedLess is used for equality! Probably undesirable.
    assert(tree1.toHash() == tree2.toHash(), "Fails before PR #6517. Still fails after PR #6517.");
}
```
Since this is an issue with the design of `RedBlackTree.opEquals` this isn't a reason to reject this PR, but in the mean time the limitation should be documented.

n8sh:
>Is it ok if I wrote this in the ddoc comment of the function

Yeah that makes sense.

>Also, I made the change for the is(T == class) case but I still get a sigsev 

You need to use a custom comparison function that is not a virtual function of `MyInt2`, because calling the `opCmp` virtual function will segfault if the receiver is `null`. Here it will also segfault if the argument is `null` since it tries to read one of the argument's fields.

An example of the `less` function you could use is:
```d
bool nullSafeLess(scope const MyInt2 a, scope const MyInt2 b)
{
     return a is null || (b !is null && a < b);
}
```

std/container/rbtree.d
```diff
@@ -1094,6 +1094,155 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
...
+    {
+        import std.traits : hasMember;
+
+        auto thisRange = this[];
+
+        static if (hasMember!(Elem, "toHash") || is(Elem == interface))
+        {
+            size_t hash = 0;
+
+            static if (is(T == class))
```
n8sh: Should be `static if (is(T == class) || is(T == interface))`

vlasebian:
I made some changes, hope it's better now.

std/container/rbtree.d
wilzbach: FWIW there's no official position on either position yet, I think it's more that PRs that remove them don't contain much value and imho sometimes in more complicated code it does help with readability.

std/container/rbtree.d
```diff
@@ -1094,6 +1094,167 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
...
+                }
+            }
+
+            return hash;
+        }
+        else
+        {
+            static size_t hashOf(T)(ref T objToHash)
+            {
+                return (() @trusted => typeid(objToHash).getHash(&objToHash))();
```
wilzbach: Wouldn't it be better to use `hashOf` here as `hashOf` can do more optimizations at compile-time?

std/container/rbtree.d
```diff
@@ -1094,6 +1094,167 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
...
+
+        static if (hasMember!(Elem, "toHash"))
+        {
+            size_t hash = 0;
+
+            static if (is(T == class) || is(T == interface))
+            {
+                foreach (ref e; thisRange)
+                {
+                    hash += (e is null ? 0 : e.toHash()) + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
```
wilzbach: The hash number could have been at least a magic constant, though repeating the hashing code three times isn't really great either. I would have gone with an inlineable function or mixin.

std/container/rbtree.d
```diff
@@ -1094,6 +1094,167 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
...
+
+            size_t hash = 0;
+
+            foreach (ref e; thisRange)
+            {
+                hash += hashOf(e) + 0x9e3779b9 + (hash << 6) + (hash >>> 2);
+            }
+
+            return hash;
+        }
```
wilzbach: I don't think we need all these special cases here anymore as `hashOf` is smart enough.

1) `hashOf` will prefer a custom `toHash` implementation if existent
2) `hashOf` does check for classes which are `null`
3) `hashOf` supports seeding

std/container/rbtree.d
```diff
@@ -1094,6 +1094,167 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
...
+        else
+        {
+            static size_t hashOf(T)(ref T objToHash)
+            {
+                return (() @trusted => typeid(objToHash).getHash(&objToHash))();
+            }
+
+            size_t hash = 0;
```
wilzbach: Why `0` as initial seed and not a more random constant?

n8sh:
Rebased and implemented @wilzbach's suggestions.

PR #6518, opened 2018-05-20
Fix issue 18884 - getSymbolsByUDA fails on AliasSeq members

Biotronic:


std/traits.d
```diff
@@ -8391,6 +8391,16 @@ template getSymbolsByUDA(alias symbol, alias attribute)
     static assert(getSymbolsByUDA!(A, Attr).stringof == "tuple(a, a, c)");
 }
 
+// #18884: getSymbolsByUDA fails on AliasSeq members
+@safe unittest
+{
+    struct X {
```
BBasile: even in the tests use Allman brace styles, please

std/traits.d
```diff
@@ -8391,6 +8391,16 @@ template getSymbolsByUDA(alias symbol, alias attribute)
...
+        alias A = AliasSeq!(ulong, uint);
+    }
+
+    assert(__traits(compiles, getSymbolsByUDA!(X, X)));
```
BBasile: could be a `static assert`. More important insert of checking that it compiles you could check the expected results.

PR #6519, opened 2018-05-21
Correct documentation for std.stdio.write.

veelo:
Remove errors in the description: It writes to stdout not args[0], calling without arguments does compile and it does not use std.conv.to.
Use similar description as std.stdio.writef.


dukc:
I was going to say that this pr needlessly discards information, but then saw your rationale. It's good, LGTM.

PR #6522, opened 2018-05-26
fix issue 18470 - std.algorithm.splitter has frame access problems for custom preds

BBasile:


std/algorithm/iteration.d
```diff
@@ -3781,8 +3781,16 @@ if (is(typeof(binaryFun!pred(r.front, s)) : bool)
         {
             static size_t lastIndexOf(Range haystack, Separator needle)
             {
+                // issue 18470 : we need to "copy" the custom pred.
```
BBasile: "recontextualize" rather than "copy".

BBasile:
For me the bug is more that the same lambda is used in two different scopes.

BBasile:
This is not a compiler bug, the problem is that the range is static.

std/algorithm/iteration.d
```diff
@@ -3758,7 +3758,7 @@ if (is(typeof(binaryFun!pred(r.front, s)) : bool)
     import std.algorithm.searching : find;
     import std.conv : unsigned;
 
-    static struct Result
+     struct Result
```
JackStouffer: remove extra space

std/algorithm/iteration.d
```diff
@@ -4104,6 +4104,23 @@ if (is(typeof(binaryFun!pred(r.front, s)) : bool)
     assert(s.empty);
 }
 
+@safe unittest // issue 18470
+{
+    import std.algorithm;
+    import std.range;
```
JackStouffer: change to specific function imports

std/algorithm/iteration.d
```diff
@@ -4104,6 +4104,23 @@ if (is(typeof(binaryFun!pred(r.front, s)) : bool)
...
+    const w = [[0], [1], [2]];
+    assert(w.splitter!((a, b) => a.front() == b)(1).equal([[[0]], [[2]]]));
+}
+
+@safe unittest // issue 18470
+{
+    import std.algorithm;
```
JackStouffer: ditto

BBasile:
Addressed review comments and squashed to delete the first solution.

PR #6523, opened 2018-05-26
[trivial] Add mean to the doc index of std.algorithm and std.algorithm.iteration

complistic-gaff:
Add missing links to `std.algorithm.iteration : mean` from the indexs of both `std.algorithm` package and the `std.algorithm.iteration`.


wilzbach:
Thanks a lot for noticing and submitting the PR!

PR #6420, opened 2018-04-04
Add public examples to all functions in std.math

JackStouffer:
Finishes everything except `rndtonl` which is blocked by issue 18693

PR #6525, opened 2018-05-27
std.math fixes for ARM

joakim-noah:


std/math.d
```diff
@@ -5058,8 +5059,8 @@ real remquo(real x, real y, out int n) @trusted nothrow @nogc  /// ditto
         assert(remainder(-5.1, 3.0).feqrel(0.9) > 16);
         assert(remainder(0.0, 3.0) == 0.0);
 
-        assert(remainder(1.0, 0.0) is -real.nan);
-        assert(remainder(-1.0, 0.0) is -real.nan);
+        assert(isNaN(remainder(1.0, 0.0)));
+        assert(isNaN(remainder(-1.0, 0.0)));
```
joakim-noah: [Discussed in a previous comment on the PR when it was added](https://github.com/dlang/phobos/pull/6420#discussion_r190309079).

std/math.d
```diff
@@ -5252,7 +5253,7 @@ public:
     assert(ieeeFlags == f);
 }
 
-version(D_HardFloat) @safe unittest
```
joakim-noah: I versioned this out sometime back, but looking around it appears that some soft-float implementatons do support checking these IEEE flags. I guess that's why there's two different structs `IeeeFlags` and `FloatingPointControl`, the former possibly supporting software floating-point implementations, while the latter exclusively for hardware? If so, this test block should be run on soft-float too.

wilzbach:
Yeah I think DScanner seems to just skips the versioned unittest. Quite stupid :/
I can have a look at the code, but that sadly won't be before the weekend. I guess the workaround for now could be to put the version block inside the unittest.

n8sh:
As mentioned above the CircleCI failure is not related to anything wrong with this PR.

JackStouffer:
Yes, this same problem cropped up in the std.path `unittest`s, and moving the version inside as the easiest solution.

joakim-noah:
I'll wait for the Dscanner fix, don't like putting it inside.

joakim-noah:
ping, can be merged.

PR #6526, opened 2018-05-27
Added tie, aka. assignalbe tuples

Bolpat:
With `tie` you can deconstruct tuples in a convenient manner, that serves when functions
return tuples, you want to swap values or simultaneously iterate variables.

Note, however, that swapping with `tie` is basically using intermediate variables,
in contrast to $(REF swap, std, algorithm, mutation).

You can use `$` or `null` on the left-hand side to ignore right-hand side
values. Use `tie` on the right-hand side for heterogeneous tuples; for
type-homogeneous tuples, static arrays are fine, too.

```d
int x = 2, y = 3;

// Swapping.
tie[ x, y ] = tie[ y, x ];
assert(x == 3 && y == 2);
// Simultaneous iteration.
tie[ x, y ] = tie[ y + 1, x * 2 ];
assert(x == 3 && y == 6);
// If all types support an operator op, you can use op=, too.
tie[ x, y ] += tie[ 2, 1 ];
assert(x == 5 && y == 7);

// $ is used for discarding.
tie[ $, x, $, y, $ ] = tie[ 0, 1, 2, 3, 4 ];
assert(x == 1 && y == 3);

static auto returnsTuple(int v) { return tuple(v * 2, v + 2); }
static int[3] returnsStaticArray(int v) { return [ v - 1, v, v + 1 ]; }

// Deconstruction.
tie[ x, y ] = returnsTuple(3);
assert(x == 6 && y == 5);
tie[ x, $, y ] = returnsStaticArray(x);
assert(x == 5 && y == 7);
```

`tie` supports non-copyable and must-initialize types.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -11,6 +11,7 @@ $(TR $(TD Tuple) $(TD
     $(LREF isTuple)
     $(LREF Tuple)
     $(LREF tuple)
+    $(LREF tie)
     $(LREF reverse)
 ))
 $(TR $(TD Flags) $(TD
@@ -64,7 +65,8 @@ Authors:   $(HTTP erdani.org, Andrei Alexandrescu),
            $(HTTP bartoszmilewski.wordpress.com, Bartosz Milewski),
            Don Clugston,
            Shin Fujishiro,
-           Kenji Hara
+           Kenji Hara,
+           Quirin Schroll
  */
 module std.typecons;
 
@@ -461,6 +463,8 @@ See_Also: $(LREF tuple).
 
 Params:
     Specs = A list of types (and optionally, member names) that the `Tuple` contains.
+Limitations:
+    `Tuple` does not support non-copyable types.
 */
 template Tuple(Specs...)
 if (distinctFieldNames!(Specs))
@@ -1988,11 +1992,17 @@ private template ReverseTupleSpecs(T...)
     the given arguments.
 
     Params:
-        Names = An optional list of strings naming each successive field of the `Tuple`.
-                Each name matches up with the corresponding field given by `Args`.
+        Names = An optional list of strings naming each successive field of the `Tuple`
+                or a list of types that the elements are being casted to.
+                For a list of names,
+                each name matches up with the corresponding field given by `Args`.
                 A name does not have to be provided for every field, but as
                 the names must proceed in order, it is not possible to skip
                 one field and name the next after it.
+                For a list of types,
+                there must be exactly as many types as parameters.
+
+    Limitations: `tuple` does not support non-copyable types.
 */
 template tuple(Names...)
 {
@@ -2004,8 +2014,9 @@ template tuple(Names...)
     Returns:
         A new `Tuple` with its type inferred from the arguments given.
      */
-    auto tuple(Args...)(Args args)
+    auto tuple(Args...)(auto ref Args args)
     {
+        import std.functional : forward;
         static if (Names.length == 0)
         {
             // No specified names, just infer types from Args...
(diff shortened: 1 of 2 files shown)


Bolpat:
I don't know why CircleCI fails. It says
```
std/typecons.d(2662:9)[warn]: Left side of logical or is identical to right side.
```
My commit does not touch that line. In the mentioned line, there is no logical or either.

n8sh:
For the benefit of people who aren't coming from the C++ world, this PR seems to be mirroring [`std::tie` which was added in C++11](http://en.cppreference.com/w/cpp/utility/tuple/tie).

Bolpat:
One thing `tie` cannot do, is declaring variables. There is a work-in-progress first-class tuple DIP that suggests introducing `auto (a, b) = returnsTuple()` and `(auto a, auto b) = returnsTuple()` to declare when unpacking.

This could be done with a much more general language change: `out` declarations. If a function `void f(out int x)` has a `out` parameter, one can then use
```d
f(out int x);
```
to declare a variable `x`. For templates, we'd need `auto out`.

With that, `tie` can fully replace first-class tuples.

In current D, as any value can be moved, you can always declare a variable beforehand and use `tie` to move something into it. With DIP 1014, this could be false in general.

Bolpat:
I have some questions to you who are reading this:
1. Should `tie` support array literals on the right-hand side?
2. Should `tie` support rubber-discard? You'd use `tie[x, $..$, y] = returns4Tuple()` instead of `tie[x, $, $, y] = ...`. Obviously, you can use only one rubber-discard.
3. Should `tie` support tuple unpacking on the last position on the left-hand side? Say, you have `int x` and `Tuple!(int, int) t` use `tie[x, t] = returns3Tuple()`.
4. Should `tie` support tuple unpacking on any position on the left-hand side? Say, you have `int x, y` and `Tuple!(int, int) t` use `tie[x, t, y] = returns4Tuple()`.

Comments about
1. It would be easy to implement assigning from array literals, i.e. `tie[ ... ] = [ ... ]`; the problem being implicit conversion inside the right-hand side literal might be overseen.
2. This abuses `$` even more. In that context, it means something completely different than with ranges. I think, `$` for discard is acceptable, it's hard to misinterpret. The following are even worse.
3. Could be difficult to understand. I tried an approach enforcing `tie[x, t..$] = ...`, but that necessitates pointers. It'd be easy to decay `t..$` into `t`, making `..$` optional.
4. See 3.

dukc:
This sure is a big decision, if we are to allow this in principle. Except a bloody debate. A lot of great use cases, definitely.

ping @andralex

std/typecons.d
```diff
@@ -2094,6 +2100,343 @@ enum isTuple(T) = __traits(compiles,
...
+        return s[0 .. $ - 1];
+    }
+    enum string hw = ctfeFunction(); // evaluated at compile-time
+    static assert(hw == "Hello, World!");
+}
+
+/// `tie` supports moving. It can be observed using an explicit destructor.
+pure nothrow @nogc @safe
+unittest
+{
```
dukc: Thats one long unittest. Can it be split?

std/typecons.d
```diff
@@ -2094,6 +2100,343 @@ enum isTuple(T) = __traits(compiles,
...
+    assert(x == 2);
+    assert(mtuple[0].isInit); // ... from here
+    assert(mtuple[1].isInit); // destroyed
+    // Tuples do not support piecewise move extraction
+}
+
+/// `tie` supports non copyable and must-initialize types.
+pure nothrow @nogc @safe
+unittest
+{
```
dukc: Again, a long unittest.

std/typecons.d
```diff
@@ -2094,6 +2100,343 @@ enum isTuple(T) = __traits(compiles,
...
+    tie[ nc ] = cast(NoCopy[1]) [ move(nc2) ];
+    assert(nc == 3 && nc2.isInit);
+    tie[ nc, nc2 ] = f();
+    assert(nc == 1 && nc2 == 2);
+
+    NoCopy[2] ncarray = [ NoCopy(1), NoCopy(2) ];
+    static assert(!__traits(compiles, {
+        tie[ nc, $ ] = ncarray; // this copies!
+    }));
+    // These are supposed to work; see Limitations.
```
dukc: Hmm, if we have nothing against them working, why unittest that they won't?

Bolpat: As soon as the bug is fixed, they won't compile and the documentation can be adapted.
If that's not intended, I'll remove it.

Bolpat:
I cannot merge because on the active master
```
dmd -main -unittest -run std/typecons.d
```
fails with linker error:
```
Error 42: Symbol Undefined __D3std4conv16testEmplaceChunkFNaNbNiNfAvkkZv
```

Bolpat:
Sorry, wrong button click.

std/typecons.d
Bolpat: I have split it.

std/typecons.d
Bolpat: Yes, in contrast to the other, is one is concerned about must-initialize and non-copyable types. Splitting it means duplicating the struct. The unit test is long, splitting it makes it longer. There's nothing to be left out. If you're dealing with non-copyable types, there's enough struggle. At least, the examples here will be rich.

std/typecons.d
dukc: Good rationale.

Bolpat:
Closing, no answers in years.

PR #6527, opened 2018-05-28
Adds const to Typedef.toString

denizzzka:


jmdavis:
What happens if the type being wrapped does not have a `toString` which is `const`? Glancing at the code, I expect that it won't compile. So, at minimum, that case must be tested for, but it's probably the case that it's not possible to make `Typedef.toString` `const` without getting into the whole mess of duplicating it and using static ifs to put `const` on it or not based on whether it can work as `const`, which gets ugly fast.

Bolpat:
Use `tostring(this Me)()`. Then `Me` becomes the type of the object including qualifiers. That's the temple version of `inout`. For the language reference, see [Template This Parameter](https://dlang.org/spec/template.html#template_this_parameter).

n8sh:
This issue isn't uncommon. It seems that the best thing might be for `const` to be inferred like `@nogc` and `nothrow`, although I am sure this has been discussed before.

Bolpat:
@n8sh In generic code, inferring things is always the way to go. There is still a difference between attributes and `const`: The latter is a type qualifier. I'm not sure if inferring it for templates *by default* is good. The necessity is low as it can be done on demand.

std/typecons.d
```diff
@@ -7177,7 +7177,7 @@ struct Typedef(T, T init = T.init, string cookie=null)
     /**
      * Convert wrapped value to a human readable string
      */
-    string toString()
+    string toString() const
```
Bolpat: Use `string toString(this T)()`

std/typecons.d
```diff
@@ -7187,7 +7187,7 @@ struct Typedef(T, T init = T.init, string cookie=null)
     }
 
     /// ditto
-    void toString(W)(ref W writer, const ref FormatSpec!char fmt)
+    void toString(W)(ref W writer, const ref FormatSpec!char fmt) const
```
Bolpat: Use `void toString(this T, W)(ref W writer, const ref FormatSpec!char fmt)`

denizzzka:
@Bolpat thanks!

MetaLang:
Can confirm what @jmdavis said; it just won't work. I also have not thought to try `toString(this T)()` though, which probably _will_ work. Good idea @Bolpat.

denizzzka:
This PR can be merged?

PR #6528, opened 2018-05-29
Check for overflow in std.experimental.allocator.makeArray

n8sh:


std/experimental/allocator/package.d
```diff
@@ -1539,14 +1539,26 @@ exception if the copy operation throws.
...
+    static if (T.sizeof <= 1)
+    {
+        const nAlloc = length * T.sizeof;
+    }
+    else
+    {
+        import core.checkedint : mulu;
+        bool overflow;
+        const nAlloc = mulu(length, T.sizeof, overflow);
+        if (overflow) return null;
```
wilzbach: Why not use `checkedint`?
It's what it was made for...

wilzbach: e.g.

```d
scope(failure) return null;
return new int[(Checked!(size_t, Throw)(size_t.max) * size_t.max).get];
```

https://run.dlang.io/is/HiyHyD

n8sh: I feel that using a non-`@nogc` solution in `std.experimental.allocator` would defeat the purpose for many use cases.

JackStouffer: Yes, until there's a `checkedint` hook that just does an `assert(0)` on error it's better to just use the `core.checkedint` functions

JackStouffer: Because the use case here is to return `null` when an overflow is found, I don't think `Checked` is really usable here.

PR #6530, opened 2018-05-30
[Trivial] Improve documentation of std.array.replace and replaceInto

JohanEngelen:
Note that `replace` doesn't have a `sink` parameter. And for `replaceInto`, `sink` must always be defined.


JohanEngelen:
Can someone please retrigger CircleCI out-of-mem failure?

ZombineDev:
Sure, done.

PR #6531, opened 2018-05-30
Fix documentation for round()

stephentgrammer:
Fix function signature and clarify documentation for `round()`.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -4882,14 +4882,11 @@ static if (real.mant_dig >= long.sizeof * 8)
 }
 
 /*******************************************
- * Return the value of x rounded to the nearest integer.
- * If the fractional part of x is exactly 0.5, the return value is
- * rounded away from zero.
- *
  * Returns:
- *     A `real`.
+ *     A `real`, which is the value of x rounded to the nearest integer. If the fractional
+ *     part of x is exactly 0.5, the it is rounded away from zero.
  */
-auto round(real x) @trusted nothrow @nogc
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -4882,14 +4882,11 @@ static if (real.mant_dig >= long.sizeof * 8)
...
 
 /*******************************************
- * Return the value of x rounded to the nearest integer.
- * If the fractional part of x is exactly 0.5, the return value is
- * rounded away from zero.
- *
  * Returns:
- *     A `real`.
+ *     A `real`, which is the value of x rounded to the nearest integer. If the fractional
+ *     part of x is exactly 0.5, the it is rounded away from zero.
```
ZombineDev: "...exactly 0.5, ~the~ it is rounded..."

Hackerpilot:
Looks like the CI fails because if you use `real` instead of `auto` it can no longer determine that the function is `pure` for the posix-specific unittest.

JackStouffer:
Yes, I changed it to auto so we didn't need to add two overloads.

@stephentgrammer in the future please avoid changes that don't add to the quality of code; it just represents churn.

stephentgrammer:
Sounds good. Thanks for your work guys!

Hackerpilot:
@JackStouffer If it weren't for the problem with the compiler failing to figure out `pure`, explicitly specifying the return type would be an improvement.

JackStouffer:
Yes, but the implementation is what we got, and having `auto` in this case is the cleaner option.

PR #6515, opened 2018-05-18
Remove deprecated std.c package

JinShil:
According to the `@@@DEPRECATED_2017-06@@@` header comment, this is overdue for removal.  Mostly this is just house-cleaning, but somewhat motivated by discussion at https://github.com/dlang/druntime/pull/2179

PR #6532, opened 2018-05-31
Temporarily disable dscanner if_else_same_check for std.typecons

JinShil:
This check appears to be preventing Circle CI from passing at https://github.com/dlang/phobos/pull/6515

The failure is due to a dscanner bug documented at https://github.com/dlang-community/D-Scanner/issues/593

Unfortunately, like others, I am not able to reproduce the problem locally.   I tried to simply re-initiate Circle CI, but it still failed.

### DIFF
--- a/.dscanner.ini
+++ b/.dscanner.ini
@@ -19,7 +19,8 @@ object_const_check="enabled"
 ; Checks for .. expressions where the left side is larger than the right.
 backwards_range_check="enabled"
 ; Checks for if statements whose 'then' block is the same as the 'else' block
-if_else_same_check="enabled"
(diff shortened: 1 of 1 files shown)


JinShil:
I recommend #6533 over this PR, but I will leave that decision to the committers.

wilzbach:
Changed it to remove the Phobos PR URL with the CircleCI advertising as it's not the only PR this occurred and the DScanner issue is the real information gain here and I also changed the check to only exclude std.typecons, but I haven't had time to verify this with the [bug reproduction script](https://github.com/dlang-community/D-Scanner/issues/593#issuecomment-393389272).

PR #6535, opened 2018-06-01
Fix Issue 18933 - std.range.assumeSorted should not nest SortedRange!(SortedRange!(...), pred)

n8sh:


std/range/package.d
```diff
@@ -10899,7 +10899,15 @@ cost $(BIGOH n), use $(REF isSorted, std,algorithm,sorting).
 auto assumeSorted(alias pred = "a < b", R)(R r)
 if (isInputRange!(Unqual!R))
 {
-    return SortedRange!(Unqual!R, pred)(r);
+    // Avoid senseless `SortedRange!(SortedRange!(...), pred)` nesting.
+    static if (is(R == SortedRange!Args, Args...))
+    {
+        return SortedRange!(Unqual!(typeof(r._input)), pred)(r._input);
```
JackStouffer: why not add another branch which just returns the input if the `pred` is the same when checked by `__traits(isSame)`?

n8sh: Added.

std/range/package.d
```diff
@@ -10899,7 +10899,20 @@ cost $(BIGOH n), use $(REF isSorted, std,algorithm,sorting).
 auto assumeSorted(alias pred = "a < b", R)(R r)
 if (isInputRange!(Unqual!R))
 {
-    return SortedRange!(Unqual!R, pred)(r);
+    // Avoid senseless `SortedRange!(SortedRange!(...), pred)` nesting.
+    static if (is(R == SortedRange!(RRange, RPred, Etc), RRange, alias RPred, Etc...))
```
wilzbach: Why `Etc`?
AFAICT SortedRange only has `Range` + `pred`:

https://github.com/dlang/phobos/blob/f8a17a7898fd2a995f2f3fb1aac58f00f1ef827b/std/range/package.d#L10278-L10279

n8sh: I didn't want to have to update this if that template ever gained additional parameters. It may have made more sense when this is was `SortedRange!(RRange, Etc)`.

n8sh: Removed the unnecessary `Etc`.

std/range/package.d
```diff
@@ -10899,7 +10899,20 @@ cost $(BIGOH n), use $(REF isSorted, std,algorithm,sorting).
...
-    return SortedRange!(Unqual!R, pred)(r);
+    // Avoid senseless `SortedRange!(SortedRange!(...), pred)` nesting.
+    static if (is(R == SortedRange!(RRange, RPred), RRange, alias RPred))
+    {
+        static if (isInputRange!R && __traits(isSame, pred, RPred))
+            // If the predicate is the same and we don't need to cast away
+            // constness for the result to be an input range.
+            return r;
+        else
+            return SortedRange!(Unqual!(typeof(r._input)), pred)(r._input);
```
wilzbach: This is never covered. Sth. like this should cover it:

```
[0, 1, 2].assumeSorted!((a, b) => a == b).assumeSorted!((a, b) => a + 1 == b +1);
```

n8sh: Fixed.

PR #6536, opened 2018-06-02
Fix Issue 18752 - std.file.read runnable example fails

wilzbach:
The problem is that run.dlang.io automatically adds `import std.stdio` and thus
we silently print to stdio instead of writing to a file (we should really deprecated
std.file.write because of this!).

Anyhow, I think the easiest solution here is to use the fully qualified symbol.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -323,7 +323,7 @@ if (isInputRange!R && isSomeChar!(ElementEncodingType!R) && !isInfinite!R &&
         remove(deleteme);
     }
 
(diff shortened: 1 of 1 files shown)


CyberShadow:
> we should really deprecated
> std.file.write because of this

FWIW, the replacement is `toFile`, though there's a bit of asymmetry because it's in `std.stdio` and not `std.file` like `read`, and it uses C FILE APIs.

PR #6537, opened 2018-06-03
fix issue 18937 - [REG 2.080.0] std.experimental.allocator: compiling…

aG0aep6G:
… `make` needs an unreasonable amount of memory for structs that contain static arrays

### DIFF
--- a/std/experimental/allocator/common.d
+++ b/std/experimental/allocator/common.d
@@ -829,7 +829,7 @@ private template isAllZeroBits(T, T value)
 
 @nogc nothrow pure @safe unittest // large static arrays
 {
-    import std.meta: Repeat;
+    import std.meta : Repeat;
     enum n = 16 * 1024;
 
     static assert(isAllZeroBits!(ubyte[n], (ubyte[n]).init));
--- a/std/experimental/allocator/common.d
+++ b/std/experimental/allocator/common.d
@@ -781,13 +781,13 @@ private bool bitwiseIdentical(T)(T a, T b)
     static struct NeverEq
     {
         int x;
-        bool opEquals(NeverEq other) { return false; }
+        bool opEquals(NeverEq other) const { return false; }
     }
 
     static struct AlwaysEq
     {
         int x;
-        bool opEquals(AlwaysEq other) { return true; }
+        bool opEquals(AlwaysEq other) const { return true; }
     }
 
     static foreach (x; AliasSeq!(-1, 0, 1, 2, "foo", NeverEq(0)))
@@ -906,7 +906,7 @@ private template isAllZeroBits(T, T value)
     static struct AlwaysEq
     {
         int x;
-        bool opEquals(AlwaysEq other) { return true; }
+        bool opEquals(AlwaysEq other) const { return true; }
     }
     static assert(AlwaysEq(0) == AlwaysEq(0));
     static assert(AlwaysEq(0) == AlwaysEq(1));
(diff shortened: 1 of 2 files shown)


aG0aep6G:
Adding this PR to the test case from [issue 18937](https://issues.dlang.org/show_bug.cgi?id=18937):

`for dmd in dmd2.079 dmd2.080 dmd2.git; do for v in -version=small -version=large -version=vlarg; do /usr/bin/time -f "$dmd $v: time: %E, memory: %M KiB" $dmd -c -unittest $v test.d; done; done`:

```
dmd2.079 -version=small: time: 0:00.11, memory: 40288 KiB
dmd2.079 -version=large: time: 0:00.11, memory: 40584 KiB
dmd2.079 -version=vlarg: time: 0:00.11, memory: 40768 KiB
dmd2.080 -version=small: time: 0:00.45, memory: 133164 KiB
dmd2.080 -version=large: time: 0:04.62, memory: 1342576 KiB
Error: out of memory
Command exited with non-zero status 1
dmd2.080 -version=vlarg: time: 1:30.94, memory: 3713420 KiB
dmd2.git -version=small: time: 0:00.17, memory: 45272 KiB
dmd2.git -version=large: time: 0:00.15, memory: 50016 KiB
dmd2.git -version=vlarg: time: 0:00.25, memory: 69864 KiB
```

So, still a bit worse than pre-regression, but much more tolerable.

aG0aep6G:
> Couldn't we put the check for static arrays into a CTFE'd function? That is, never unpack static arrays on the template level; rather make the condition (pseudocode) enum isAllZeroBits = isAllZeroBits!(value[0]) && allElementsEqual(value);?

That didn't occur to me. Seems to work, but it's a tiny bit slower than the template stuff.

I tested this:

```d
        enum isAllZeroBits = ()
        {
            static if (value.length == 0) return true;
            else static if (.isAllZeroBits!(typeof(value[0]), value[0]))
            {
                foreach (e; value)
                {
                    if (e !is value[0]) return false;
                }
                return true;
            }
            else return false;
        }();
```

And got these stats:

```
dmd2.git -version=small: time: 0:00.13, memory: 45472 KiB
dmd2.git -version=large: time: 0:00.16, memory: 50832 KiB
dmd2.git -version=vlarg: time: 0:00.28, memory: 73108 KiB
```

The difference really isn't large, though. I wouldn't mind changing it to this, if you prefer it over the template version. Maybe this version will benefit from NewCTFE?

aG0aep6G:
Found a small problem: Comparing static arrays with `==` is no good, because an element's opEquals might give a result that has nothing to with its bits (e.g. always true or always false). And `is` gives a deprecation warning saying it compares by reference, not by values. It actually seems to do the right thing during CTFE, but the deprecation is probably going to become an error at some point, so I don't want to add code that triggers it.

So I've added a `bitwiseIdentical` function which steps into static arrays to do `is` on the elements. It would be nice if there was a simpler way to do this, but I can't see one.

With this, the template version lost the slight advantage it had in the `T.init` case. So I've switched to the CTFE `foreach` version. It performs better on values that are not `T.init`. Also added tests for funky opEquals and nested static arrays.

Current stats for the test case of issue 18937:

```
dmd2.git -version=small: time: 0:00.14, memory: 45472 KiB
dmd2.git -version=large: time: 0:00.17, memory: 51692 KiB
dmd2.git -version=vlarg: time: 0:00.32, memory: 75480 KiB
```

aG0aep6G:
Ping? This is a regression fix, and I'd like to move forward here.

PR #6539, opened 2018-06-03
Add getTimesPosix

FreeSlave:
I'm not sure about the order of arguments in getTimesPosix.
Maybe you want it to follow getTimesWin-like approach when the additional argument is the first among SysTime arguments.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -51,6 +51,7 @@ $(TR $(TD Attributes) $(TD
 ))
 $(TR $(TD Timestamp) $(TD
           $(LREF getTimes)
+          $(LREF getTimesPosix)
           $(LREF getTimesWin)
           $(LREF setTimes)
           $(LREF timeLastModified)
@@ -1374,6 +1375,56 @@ version (StdDdoc)
                         out SysTime fileModificationTime)
     if (isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R) &&
         !isConvertibleToString!R);
+
+    /++
+     $(BLUE This function is Posix-Only.)
+
+     Get access/modified/change times of file `name`.
+
+     This is the same as `getTimes` except that it also gives you the file
+     status change time.
+
+     Params:
+     name             = File _name to get times for.
+     accessTime       = Time the file was last accessed.
+     modificationTime = Time the file was last modified.
+     changeTime       = Time the file status was last changed.
+
+     Throws:
+     $(LREF FileException) on error.
+     +/
+    void getTimesPosix(R)(R name,
+              out SysTime accessTime,
+              out SysTime modificationTime,
+              out SysTime changeTime)
+    if (isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R) &&
+        !isConvertibleToString!R);
+}
+else version(Posix)
+{
+    void getTimesPosix(R)(R name,
+              out SysTime accessTime,
+              out SysTime modificationTime,
+              out SysTime changeTime)
+    if (isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R) &&
+        !isConvertibleToString!R)
+    {
+        stat_t statbuf = void;
+        fillStatBuf(name, statbuf);
+
+        accessTime = statTimeToStdTime!'a'(statbuf);
+        modificationTime = statTimeToStdTime!'m'(statbuf);
+        changeTime = statTimeToStdTime!'c'(statbuf);
+    }
+
+    void getTimesPosix(R)(auto ref R name,
+              out SysTime accessTime,
+              out SysTime modificationTime,
+              out SysTime changeTime)
+    if (isConvertibleToString!R)
+    {
+        return getTimesPosix!(StringTypeOf!R)(name, accessTime, modificationTime);
+    }
 }
 else version (Windows)
 {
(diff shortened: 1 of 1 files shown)


std/file.d
```diff
@@ -1463,6 +1519,68 @@ version(Windows) @system unittest
...
+            auto diffa = accessTime2 - currTime;
+            auto diffm = modificationTime2 - currTime;
+            auto diffc = changeTime2 - currTime;
+            scope(failure) writefln("[%s] [%s] [%s] [%s] [%s] [%s] [%s]",
+                                    accessTime2, modificationTime2, changeTime2, currTime, diffa, diffm, diffc);
+
+            //There is no guarantee that the access time will be updated.
+            assert(abs(diffa) <= leeway + sleepTime);
+            assert(abs(diffm) <= leeway);
+            assert(abs(diffc) <= leeway);
```
CyberShadow: Are these tests meaningful with `sleepTime < leeway`?

std/file.d
```diff
@@ -1463,6 +1519,68 @@ version(Windows) @system unittest
...
+        }
+
+        assert(accessTime1 <= accessTime2);
+        assert(modificationTime1 <= modificationTime2);
```
CyberShadow: Why `<=` and not `<` ? AFAIK the lowest precision of filesystem timestamps is FAT32 with 2 seconds, which should still almost always pass this test with `<`.

CyberShadow:
Pinging @andralex as he requested that his approval is obtained for any new Phobos symbols.

Summary: this adds `getTimesPosix` to complement `getTimesWin`. Currently, there is no D-ish way to get the "changed" timestamp of a file on POSIX, other than first wrapping it in a `DirEntry` and querying that.

wilzbach:
Why not do the following to have a unified API for all OSes?

```d
Times getTimes(R)(R name)
...
struct Times
{
    SysTime accessTime;
    SysTime modificationTime;
    version(Windows) SysTime creationTime;
    version(Posix) SysTime statusChangeTime;
}
```

This will probably be a lot less confusing than having getTimes, getTimesWin and getTimesPosix which all only marginally differ.

CyberShadow:
That's pretty much what `DirEntry` already is.

wilzbach:
Cool! So we have a successful precedent?

CyberShadow:
Not sure what you mean by that :)

I guess we *could* go in the other direction and instead of adding `getTimesPosix` to complement `getTimesWin`, deprecate `getTimesWin` in favor of `DirEntry`. Is that what you meant?

CyberShadow:
BTW, perhaps that could also be a good direction for a solution for #5755? Add a `DirEntry` constructor that takes a name and `stat_t`.

wilzbach:
Oh sorry for not being so specific. I was dumply thinking that DirEntry only works for directories, but yeah of course it does for files too.
So I was actually thinking about adding a `getTimes` function that only takes a name and returns a DirEntry-like object (or maybe even a DirEntry) and yep then we could deprecate getTimesWin which would be an additional win in my opinion.
Anyhow if we decide to add an additional constructor overload to DirEntry, we probably want to add the convenience `dirEntry`  too as that's a common pattern throughout Phobos for nice UFCS usage.

CyberShadow:
> So I was actually thinking about adding a getTimes function that only takes a name and returns a DirEntry-like object (or maybe even a DirEntry)

My point was that would be entirely redundant because you can use `DirEntry` itself in exactly the same way:

```d
auto de = DirEntry(fileName);
writeln(de.timeLastModified);
writeln(de.timeLastAccessed);
```
etc.


> Anyhow if we decide to add an additional constructor overload to DirEntry, we probably want to add the convenience dirEntry too as that's a common pattern throughout Phobos for nice UFCS usage.

Funny of you to say that, as there *was* a `dirEntry` function not too long ago, which got deprecated in favor of the constructor.

In any case, constructors are UFCS-able, so...

```d
	string fn;
	auto de = fn.DirEntry;
```

std/file.d
FreeSlave: I don't know, I just copied tests from getTimes.

std/file.d
FreeSlave: Again, it's just copy-paste from getTimes. Not very good, but getTimesWin also has this copy-paste

FreeSlave:
If we deprecate getTimesWin in favor of DirEntry, we can deprecate getTimes too.

FreeSlave:
> BTW, perhaps that could also be a good direction for a solution for #5755? Add a DirEntry constructor that takes a name and stat_t.

Such constructor can be added, but I hardly see it as a solution. The user may want to avoid using DirEntry.

CyberShadow:
> The user may want to avoid using DirEntry.

Why though?

FreeSlave:
> Why though?

On Posix It's not nothrow, not pure, etc.

andralex:
All - please educate me. Why not use `DirEntry` for this?

FreeSlave:
In this case DirEntry is ok. I just thought we need a mirror function to complement getTimesWin.

I would like you to give some attention to #5755 which is also related to time functions and more important in my opinion.

jmdavis:
All this does over `getTimes` is provide the value for `st_ctim`, which I very much doubt that very many people want, and honestly, I question the value of either `getTimes` or `getTimesWin` over `DirEntry`, but I believe that they predate being able to get a `DirEntry` for a file without calling `dirEntries`.

andralex:
Cool, I approved #5755

wilzbach:
Why was this merged?
I thought there was a consensus that this is the wrong direction and that we should rather deprecate getTimesWin instead of introducing new platform specific functions in a supposedly abstracting standard library.

Also this has never been approved by @andralex and for good reasons all new public symbols need to be approved by Andrei.

PR #6541, opened 2018-06-04
Remove invalid Markdown backticks in std.path

wilzbach:
To fix https://github.com/dlang/dmd/pull/8325

### DIFF
--- a/std/path.d
+++ b/std/path.d
@@ -13,7 +13,7 @@
     To differentiate between these cases, use $(REF isDir, std,file) and
     $(REF exists, std,file).
 
-    Note that on Windows, both the backslash (``\``) and the slash (``/``)
+    Note that on Windows, both the backslash ($(D `\`)) and the slash ($(D `/`))
     are in principle valid directory separators.  This module treats them
     both on equal footing, but in cases where a $(I new) separator is
     added, a backslash will be used.  Furthermore, the $(LREF buildNormalizedPath)
@@ -141,8 +141,8 @@ else static assert(0, "unsupported platform");
 
 /** Determines whether the given character is a directory separator.
 
-    On Windows, this includes both ``\`` and ``/``.
-    On POSIX, it's just ``/``.
+    On Windows, this includes both $(D `\`) and $(D `/`).
+    On POSIX, it's just $(D `/`).
 */
 bool isDirSeparator(dchar c)  @safe pure nothrow @nogc
 {
@@ -1425,7 +1425,7 @@ private auto _withDefaultExtension(R, C)(R path, C[] ext)
     preceding segments will be dropped.
 
     On Windows, if one of the path segments are rooted, but not absolute
-    (e.g. ``\foo``), all preceding path segments down to the previous
+    (e.g. $(D `\foo`)), all preceding path segments down to the previous
     root will be dropped.  (See below for an example.)
 
     This function always allocates memory to hold the resulting path.
@@ -2639,7 +2639,7 @@ if (isRandomAccessRange!R && isSomeChar!(ElementType!R) ||
     ---
 
     On Windows, an absolute path starts at the root directory of
-    a specific drive.  Hence, it must start with ``d:\`` or ``d:/``,
+    a specific drive.  Hence, it must start with $(D `d:\`) or $(D `d:/`),
     where `d` is the drive letter.  Alternatively, it may be a
     network path, i.e. a path starting with a double (back)slash.
     ---
@@ -2849,14 +2849,14 @@ if (isConvertibleToString!R)
     taken to be the current working directory.  If specified,
     `base` must be an absolute path, and it is always assumed
     to refer to a directory.  If `path` and `base` refer to
-    the same directory, the function returns ``.``.
+    the same directory, the function returns $(D `.`).
 
     The following algorithm is used:
     $(OL
         $(LI If `path` is a relative directory, return it unaltered.)
         $(LI Find a common root between `path` and `base`.
             If there is no common root, return `path` unaltered.)
-        $(LI Prepare a string with as many ``../`` or ``..\`` as
+        $(LI Prepare a string with as many $(D `../`) or $(D `..\`) as
             necessary to reach the common root from base path.)
         $(LI Append the remaining segments of `path` to the string
             and return.)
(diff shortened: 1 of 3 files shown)


wilzbach:
I'm not sure I understand, here's the diff from DAutoTest of the resulting HTML:

```diff
-    Note that on Windows, both the backslash (<span class="d_inlinecode donthyphenate notranslate"></span>\<span class="d_inlinecode donthyphenate notranslate"></span>) and the slash (<span class="d_inlinecode donthyphenate notranslate"></span>/<span class="d_inlinecode donthyphenate notranslate"></span>)
+    Note that on Windows, both the backslash (<span class="d_inlinecode donthyphenate notranslate">\</span>) and the slash (<span class="d_inlinecode donthyphenate notranslate">/</span>)
     are in principle valid directory separators.  This module treats them
     both on equal footing, but in cases where a <i>new</i> separator is
     added, a backslash will be used.  Furthermore, the <a href="#buildNormalizedPath"><span class="d_inlinecode donthyphenate notranslate">buildNormalizedPath</span></a>
```

wilzbach:
BTW with a quick grep I found more instances of double-backticks.

jmdavis:
> I'm not sure I understand, here's the diff from DAutoTest of the resulting HTML:

Are you referring to my comment about using

    $(D `\`)

instead? I mean that both slashes are supposed to be strings. Originally, they would have been

    $(D `/`)

and

    $(D `\`)

where the backticks make them strings where the backslash doesn't have to be escaped - e.g. `$(D "\\")`. However, someone came along and mass-replaced `$(D ... )` with backslashes, which then screwed up any case where backslahes were used as string delimeters. The way to fix this is not to get rid of the "extra" backticks but to stop using backticks instead of `$(D ...)`. If ddoc can't handle backtick strings in `$(D ...)` anymore, then that's a serious bug in ddoc that needs to be fixed.

> BTW with a quick grep I found more instances of double-backticks.

Any case where backticks were used as string delimiters would have been screwed up by any attempt to mass-replace `$(D ...)` with backticks, so it doesn't surprise me at all if we ended up with double backticks in a number of places.

dgileadi:
As the DAutoTest output above shows the consecutive backticks are not currently producing the desired output—they surround the slash with two empty code spans instead of wrapping it in a code span.

Markdown [collapses consecutive backticks](https://daringfireball.net/projects/markdown/syntax#code), so my upcoming markdown support will treat \`\`\\\`\` and \`\\\` identically and would thus produce the correct output. The problem in dlang/dmd#8325 would be fixed by the upcoming Markdown support for collapsing consecutive backticks. However the `markdown` branch doesn't yet support consecutive backticks but does support backslash escapes, and that caused the failure. It seems to me that this PR is the correct fix for dlang/dmd#8325, and also has the happy effect of producing the desired output in `master`.

jmdavis:
> Markdown collapses consecutive backticks, so my upcoming markdown support will treat ``\`` and `\` identically and would thus produce the correct output.

When you say that, do you mean that two backticks in a row are treated the same as one backtick or that two backticks in a row will result in the output having a single backtick? What we want here is to treat these as strings - the equivalent of `"/"` and `"\\"` except that it uses backticks for string delimiters instead of double quotes so that the escaping isn't necessary. If there's a clean way to do that with escaping the backticks rather than using `$(D...)`, then fine, but these should be string literals with backticks in the final result.

Either way, whatever you do with markdown support, `$(D...)` with backticks in it still needs to treat them as backticks and not markdown, or it will break all of the existing inline code samples that use backticks for string literals.

dgileadi:
>> Markdown collapses consecutive backticks, so my upcoming markdown support will treat \'\'\\'\' and \'\\' identically and would thus produce the correct output.

> When you say that, do you mean that two backticks in a row are treated the same as one backtick or that two backticks in a row will result in the output having a single backtick? What we want here is to treat these as strings - the equivalent of "/" and "\\" except that it uses backticks for string delimiters instead of double quotes so that the escaping isn't necessary. If there's a clean way to do that with escaping the backticks rather than using $(D...), then fine, but these should be string literals with backticks in the final result.

Yes, under Markdown multiple backticks in a row will be treated the same as a single backtick for creating a code span, provided the span is closed with the same number of backticks. If you want to output a backtick character you can backslash-escape it like \\`.

Note that for the documentation this PR addresses in std.path it appears that the backticks are not currently displayed (see the red line in the DAutoTest output), and from the context of the paragraph I don't believe that the original author wanted a backtick character to display. Rather I believe they intended to make code spans that displayed a backslash and a forward slash.

> Either way, whatever you do with markdown support, $(D...) with backticks in it still needs to treat them as backticks and not markdown, or it will break all of the existing inline code samples that use backticks for string literals.

So far as I can tell, $(D...) is not treated specially in ddoc processing—backticks inside it are treated just the same as backticks outside it. So today $(D \`stuff in backticks\`) would be double-wrapped by two code spans.

aG0aep6G:
> I don't believe that the original author wanted a backtick character to display. Rather I believe they intended to make code spans that displayed a backslash and a forward slash.

Nope. The backticks were supposed to be displayed. @CyberShadow changed it from

    '\'

to

    `\`

in #1267.

The PR description says: "Fixed a few char/string literals so they're in valid D syntax". A backslash on its own isn't valid D syntax. But a backslash in backticks is a string literal.

That was back in 2.063. The backticks were displayed then. DDoc only got backtick support with 2.067 (https://github.com/dlang/dmd/commit/26c5de42214c788ab42d96694a994c3eb8ae976a).

dgileadi:
Nice research. In that case maybe it has been broken since 2.067?

aG0aep6G:
> In that case maybe it has been broken since 2.067?

Yearp.

wilzbach:
So what's the consensus here: do we go with `` `/` `` or `` $(D `/`) ``?
I don't really mind and it seems to me to be a purely aesthetic choice. I just want one solution to get merged, s.t. the DMD PR is no longer blocked.

CyberShadow:
> I don't really mind and it seems to me to be a purely aesthetic choice.

It's not. The intended result is that in the rendered documentation, a string literal is displayed. The contents of the backticks is not D source code, so it should not be highlighted as such. You broke it in #6391, so, please take the opportunity to fix it.

This should have been caught by DAutoTest's HTML diff before #6391 was merged, and to this day I still don't understand why it didn't work there, or why #6391 was merged before ensuring that the changes were correct.

wilzbach:
Changed it to `` $(D `/`) `` then.

std/path.d
```diff
@@ -1425,7 +1425,7 @@ private auto _withDefaultExtension(R, C)(R path, C[] ext)
     preceding segments will be dropped.
 
     On Windows, if one of the path segments are rooted, but not absolute
-    (e.g. ``\foo``), all preceding path segments down to the previous
+    (e.g. `\foo`), all preceding path segments down to the previous
```
jmdavis: $(D `\foo`)

RazvanN7:
ping @wilzbach

wilzbach:
Finally got around addressing the final nits. We should be good to go here now?

PR #6542, opened 2018-06-04
Remove a few cases of underscore escaping

wilzbach:
A start at removing the no longer needed underscore escaping from Phobos.
This was done partially by a script to get an initial set of replacements:

```d
import std.experimental.all;
void main(string[] args)
{
    foreach (file; dirEntries(".", "*.d", SpanMode.depth))
    {
        auto text = file.readText;
        auto packageName = file.baseName.stripExtension;
        auto moduleName = file.dirName.baseName;
        foreach (rep; [packageName, moduleName])
            text = text.replace("_"~rep, rep);
        text.toFile(file);
    }
}
```

And then manually adding the replacements to the git index.
I might have missed a few instances (or skipped the ones which were a bit more complicated),
but we can easily find them later with the `DDOC_AUTO_PSYMBOL_SUPPRESS` macro).

See also:
- https://github.com/dlang/dlang.org/pull/2307
- https://dlang.org/changelog/2.079.0.html#fix18361

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -1,7 +1,7 @@
 // Written in the D programming language.
 /**
 This is a submodule of $(MREF std, algorithm).
-It contains generic _comparison algorithms.
+It contains generic comparison algorithms.
 
 $(SCRIPT inhibitQuickIndex = 1;)
 $(BOOKTABLE Cheat Sheet,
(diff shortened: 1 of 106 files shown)


schveiguy:
I was starting to look through all these, but there's too many! Luckily we have the doc preview, I can see only a few html files changed, and all the changes are fixes (removing unintentional underscores).

PR #6544, opened 2018-06-04
Fix issue 18934 - make std.variant support getting types that have const members

schveiguy:
The root cause is simply that the wrong static if check is used for something that should be a move semantic.

The two places that I could see this affects are assigning a `Variant` to another `Variant`, and the `get` function.

I'm 99% sure that this is correct. There's no way to get a reference to the data, so the fact that technically const data is changing doesn't matter. It's more like the same memory is just being used for something different.

This is reinforced by the fact that you can simply assign another value and it works (even without this change).

### DIFF
--- a/std/variant.d
+++ b/std/variant.d
@@ -299,7 +299,14 @@ private:
                 if (targetType != typeid(T))
                     continue;
 
-                static if (is(typeof(*cast(T*) target = *src)) ||
+                // SPECIAL NOTE: variant only will ever create a new value with
+                // tryPutting (effectively), and T is ALWAYS the same type of
+                // A, but with different modifiers (and a limited set of
+                // implicit targets). So this checks to see if we can construct
+                // a T from A, knowing that prerequisite. This handles issues
+                // where the type contains some constant data aside from the
+                // modifiers on the type itself.
+                static if (is(typeof(delegate T() {return *src;})) ||
                            is(T ==        const(U), U) ||
                            is(T ==       shared(U), U) ||
                            is(T == shared const(U), U) ||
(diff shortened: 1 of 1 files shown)


std/variant.d
```diff
@@ -299,7 +299,10 @@ private:
                 if (targetType != typeid(T))
                     continue;
 
-                static if (is(typeof(*cast(T*) target = *src)) ||
+                // note, variant only will ever create a new value with
+                // tryPutting (effectively), so it's OK to assume move
+                // semantics.
+                static if (is(typeof(delegate T() {return *src;})) ||
```
MartinNowak: Returning a value isn't a move, but a copy construction, isn't it? So you're changing the check from assignable to copy-constructible.

schveiguy: I guess I don't know the correct terminology. Whatever it is where you pass an rvalue parameter to a function or return a value, that's what I was trying to say. I thought that was "move semantics".

But I can see that if `A` is not the same type as `T`, but `T` can be constructed from an `A`, it would still pass. Going to do some testing to make sure this makes sense.

schveiguy: Yeah, it doesn't work if you try to get a copy-constructible type from the variant. I forgot the `T` here is some constancy flavor of `A`, not the `T` passed into `get`.

So I think it's safe to say we don't have to worry about that case.

I'll change the comment to be clearer.

schveiguy:
@MartinNowak updated the comment, let me know if it's clearer or not.

wilzbach:
Probably it ran in an unlucky time. See https://github.com/dlang/dmd/commit/137ebc2265aad3880b8b5c3e7e09bfe5a6dc773f

I guess a restart helps.

schveiguy:
Looks like this didn't make it into 2.080.1

Should I re-target master, or can this still go into stable?

MartinNowak:
> Should I re-target master, or can this still go into stable?

Fixes can always go into stable.

schveiguy:
@MartinNowak thx!

PR #6545, opened 2018-06-04
Issue 18948 - std.uni.toLower and std.uni.toUpper should work with random access ranges

JackStouffer:


std/uni.d
```diff
@@ -8997,20 +8997,23 @@ private alias UpperTriple = AliasSeq!(toUpperIndex, MAX_SIMPLE_UPPER, toUpperTab
 private alias LowerTriple = AliasSeq!(toLowerIndex, MAX_SIMPLE_LOWER, toLowerTab);
 
 // generic toUpper/toLower on whole string, creates new or returns as is
-private S toCase(alias indexFn, uint maxIdx, alias tableFn, alias asciiConvert, S)(S s) @trusted pure
-if (isSomeString!S)
+private ElementEncodingType!S[] toCase(alias indexFn, uint maxIdx, alias tableFn, alias asciiConvert, S)(S s)
+if (isSomeString!S || (isRandomAccessRange!S && hasLength!S && hasSlicing!S && isSomeChar!(ElementType!S)))
```
wilzbach: We really need to fix our stringLike trait mess :/

jmdavis:
The issue there is that whether you really want something "string-like" really depends on what you're doing. In general, what you really want is a range of characters. Then, depending on what your function requires, it may then require a forward, bi-directional, or random-access range. And it may or may not require length or slicing - all of the normal questions when writing a range-based function. There's nothing special about strings in this regard.

I don't think that "string-like" really comes into play until you start discussing adding appending and concatenation to the list so that you're actually dealing with a type that really, truly acts like a  string. And of course, auto-decoding complicates the whole thing further.

But honestly, I think that having longer template constraints such as

    if (isSomeString!S || (isRandomAccessRange!S && hasLength!S && hasSlicing!S && isSomeChar!(ElementType!S)))

is really just a consequence of ranges in general, and there's nothing special about strings here. It's more complicated than it should be thanks to the issue with narrow strings and auto-decoding, but restricting functions based on whether they support the full set of operations that a string does rarely actually makes sense. Rather, you restrict it based on what that particular function needs, and that's just normal range stuff. If we could get rid of auto-decoding, then we could clean up some of this stuff, but I don't think that it's at all reasonable to expect to be able to get rid of the longer portion of the template constraints for range-based code.

The place where something "string-like" might actually come into play is when you're actually trying to operate on strings and not ranges, and for the most part, we should be operating on ranges, because that's more flexible.

JackStouffer:
tuple expansion for `enumerate` is failing in dustmite for some reason

JackStouffer:
meh, it's faster to use a raw for loop anyway

JackStouffer:
string-like ranges are a separate topic, and as long as range interfaces are provided with whatever library string type finally gets created, it really doesn't matter

BBasile:
causes this regression https://issues.dlang.org/show_bug.cgi?id=18993

std/uni.d
```diff
@@ -8997,20 +8997,24 @@ private alias UpperTriple = AliasSeq!(toUpperIndex, MAX_SIMPLE_UPPER, toUpperTab
...
 {
-    import std.array : appender;
+    import std.array : appender, array;
     import std.ascii : isASCII;
+    import std.utf : byDchar;
 
-    foreach (i, dchar cOuter; s)
+    auto r = s.byDchar;
+    for (size_t i; !r.empty; ++i, r.popFront())
```
tgehr: This change is incorrect. `i` has to increase by the encoding length of the character.

PR #5510, opened 2017-06-25
Add writeNewFile - check file for existence and open file atomically

FreeSlave:
Check this idea.
Probably the name is not the best, suggest the others.

PR #6546, opened 2018-06-05
File non string open mode

FreeSlave:
That adds a new way to create and open File object - instead of using C standard library style string modes it allows to use symbolic flags. This is a more granular approach and allows to open file in 'Create New' mode to ensure the file did not exist before opening. Note that C11 standard has 'x' flag that does the same thing, but C11 is not always implemented.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -469,15 +469,37 @@ Throws: `ErrnoException` if the file could not be opened.
     /// Flags for file open mode.
     enum OpenMode
     {
-        read = 1 << 0, /// Open in read mode.
-        write = 1 << 1, /// Open in write mode, dont' truncate.
-        truncate = 1 << 2, /// Open in write mode, create if needed, truncate if exists.
-        append = 1 << 3, /// Open in write mode, create if needed. Append to the end on writing.
-        createNew = 1 << 4, /// Open in write mode, ensure the file did not exist.
+        /// Open in read mode.
+        read = 1 << 0,
+        /**
+         * Open in write mode, dont' truncate.
+         * Create a file if `existingOnly` flag is not provided.
+         */
+        write = 1 << 1,
+        /**
+         * Open in write mode, truncate if exists.
+         * Create a file if `existingOnly` flag is not provided.
+         */
+        truncate = 1 << 2,
+        /**
+         * Open in write mode. Append to the end on writing.
+         * Create a file if `existingOnly` flag is not provided.
+         *
+         * Note that it has a special meaning on Posix:
+         * the file opened in append mode will have write operations
+         * happening at the end of the file regardless of manual seek position changing.
+         */
+        append = 1 << 3,
+        /**
+         * Open in write mode. Create file only if it does not exist, error otherwise.
+         * The check for existence and the file creation is an atomical operation.
+         * Use this flag when need to ensure that the file with such name did not exist.
+         */
+        createNew = 1 << 4,
         /**
          * Don't create file if it does not exist when opening in write mode.
          * This flag is not necessary when opening a file in read-only mode.
-         * This flag can't be used together with OpenMode.createNew flag.
+         * This flag can't be used together with `OpenMode.createNew` flag.
          */
         existingOnly = 1 << 5,
     }
@@ -646,6 +668,17 @@ Throws: `ErrnoException` if the file could not be opened.
         }
     }
 
+    /**
+    Ditto, but takes the open mode in the symbolic form rather than in the string one.
+    It allows to open in modes that are not possible when specifying via string,
+    e.g. to create a new file only if the file with provided name does not already exist.
+
+    Params:
+        name = range or string representing the file _name
+        openMode = flags for the open mode
+
+    Throws: `ErrnoException` if the file could not be opened. On Windows it can also throw `WindowsException`.
+    */
     this(string name, OpenMode openMode) @safe
     {
         import std.exception : enforce;
@@ -688,7 +721,7 @@ Throws: `ErrnoException` if the file could not be opened.
 
         assert(std.file.read(deleteme) == "foobar");
 
-        f = File(deleteme, OpenMode.truncate);
+        f = File(deleteme, OpenMode.truncate | OpenMode.existingOnly);
         f.write("baz");
         f.close();
 
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -512,7 +512,8 @@ Throws: `ErrnoException` if the file could not be opened.
     private static string modezFromOpenMode(OpenMode openMode) @safe
     {
         const bool anyWrite = hasAnyWriteFlag(openMode);
-        if (openMode & OpenMode.read) {
+        if (openMode & OpenMode.read)
+        {
             if (anyWrite)
             {
                 if (openMode & OpenMode.append)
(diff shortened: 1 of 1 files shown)


std/stdio.d
```diff
@@ -466,6 +466,283 @@ Throws: `ErrnoException` if the file could not be opened.
...
+         */
+        existingOnly = 1 << 5,
+    }
+
+    private static bool hasAnyWriteFlag(OpenMode openMode) @safe nothrow pure @nogc
+    {
+        return (openMode & (OpenMode.write | OpenMode.truncate | OpenMode.append | OpenMode.createNew)) != 0;
+    }
+
+    private static const(char)[4] modezFromOpenMode(OpenMode openMode) @trusted
```
BBasile: this is `@safe`

std/stdio.d
```diff
@@ -466,6 +466,283 @@ Throws: `ErrnoException` if the file could not be opened.
...
+            else if ((openMode & OpenMode.read) != 0 && !hasAnyWriteFlag(openMode))
+                mode |= O_RDONLY;
+
+            auto fd = _open_osfhandle(cast(intptr_t) handle, mode);
+        }
+        errnoEnforce(fd >= 0, "Cannot open Windows HANDLE");
+        return fdToFp(fd, openMode);
+    }
+
+    private static FILE* openFile(string name, OpenMode openMode) @trusted
```
BBasile: this is `@safe` since inside calls are trusted.

std/stdio.d
```diff
@@ -466,6 +466,283 @@ Throws: `ErrnoException` if the file could not be opened.
...
+    {
+        return (openMode & (OpenMode.write | OpenMode.truncate | OpenMode.append | OpenMode.createNew)) != 0;
+    }
+
+    private static const(char)[4] modezFromOpenMode(OpenMode openMode) @trusted
+    {
+        const bool anyWrite = hasAnyWriteFlag(openMode);
+        char[4] modez;
+        size_t i = 0;
+        if (openMode & OpenMode.read) {
```
BBasile: curly braces on their own lines

std/stdio.d
```diff
@@ -466,6 +466,283 @@ Throws: `ErrnoException` if the file could not be opened.
...
+            } else {
+                modez[i++] = 'r';
+            }
+        } else if (anyWrite) {
+            if (openMode & OpenMode.append)
+                modez[i++] = 'a';
+            else
+                modez[i++] = 'w';
+        }
+        assert(i < modez.length);
```
BBasile: boundscheck will catch this too automatically and before.

std/stdio.d
```diff
@@ -466,6 +466,283 @@ Throws: `ErrnoException` if the file could not be opened.
...
+         * Don't create file if it does not exist when opening in write mode.
+         * This flag is not necessary when opening a file in read-only mode.
+         * This flag can't be used together with `OpenMode.createNew` flag.
+         */
+        existingOnly = 1 << 5,
+    }
+
+    private static bool hasAnyWriteFlag(OpenMode openMode) @safe nothrow pure @nogc
+    {
+        return (openMode & (OpenMode.write | OpenMode.truncate | OpenMode.append | OpenMode.createNew)) != 0;
```
BBasile: how about `with(OpenMode) return...` for a smaller return statement.

std/stdio.d
```diff
@@ -466,6 +466,283 @@ Throws: `ErrnoException` if the file could not be opened.
...
+        else
+            fd = core.sys.posix.fcntl.open(namez, mode);
+        errnoEnforce(fd >= 0);
+        return fd;
+    }
+
+    version(Windows) private static HANDLE openHandle(string name, OpenMode openMode) @trusted
+    {
+        import std.internal.cstring : tempCString;
+        import core.sys.windows.windows;
```
BBasile: selective import, as you did for the Posix version of the function

std/stdio.d
```diff
@@ -466,6 +466,283 @@ Throws: `ErrnoException` if the file could not be opened.
...
+
+    private static const(char)[4] modezFromOpenMode(OpenMode openMode) @trusted
+    {
+        const bool anyWrite = hasAnyWriteFlag(openMode);
+        char[4] modez;
+        size_t i = 0;
+        if (openMode & OpenMode.read) {
+            if (anyWrite) {
+                if (openMode & OpenMode.append) {
+                    modez[i++] = 'a';
```
BBasile: This whole block has a poor style. You could return string literals, taking advantage of the automatic null sentinel, eg:

```d
        if (openMode & OpenMode.read)
        {
            if (anyWrite)
            {
                if (openMode & OpenMode.append)
                    return "a+"; // or "a+\0\0" if necessary
                else if (openMode & OpenMode.existingOnly)
                    return "r+";
                else if (openMode & OpenMode.truncate)
                    return "rw";
                else
                   return "w+";
            }
            else return "r";
        }
 // etc....
```

std/stdio.d
FreeSlave: It's too many symbols. But will do if it's really required.

BBasile: I think you don't have the choice if you want tthis PR to become green Circle CI (still) fails because of this.

CyberShadow:
Hmm, I don't think this will fly (mainly with @andralex).

This adds a lot of new code, while being able to use enums instead of strings is only a superficial improvement.

std.stdio is specifically designed to be built on top of C stdio. Repeating its work by reimplementing things that belong in C stdio is a tough sell as well. I think a more orthogonal approach would be to e.g. implement the x file mode in DigitalMars' libc, which would then make it available on Win32 D.

Is there anything else that I missed?

BBasile:
@CyberShadow  I expected such as reaction, however my review was purely technical. And @FreeSlave, you can still create a DUB package for this, so that work is not lost, in case final decision is "no no no".

FreeSlave:
> Is there anything else that I missed?

I think all useful combinations are covered with string mode considering there's a support for x and I believe it existed in glibc for a long time before C11. But I'm not sure about unix-like platforms that don't use glibc and also Microsoft libraries (fixing dmc C library is not enough, because user may link to Microsoft C library, right?)

CyberShadow:
> But I'm not sure about unix-like platforms that don't use glibc 

Along the lines of the above, it is the burden of said libc libraries to support C11.

> and also Microsoft libraries (fixing dmc C library is not enough, because user may link to Microsoft C library, right?)

They may, yes, if they use `-m64` or `-m32mscoff`.

FreeSlave:
I'm not fond of the fact that the public interface of D standard libary depends on the special features of the C library implementation. But I guess std.stdio is just designed that way. If others say no, I'll take the idea of making the dub package. Fortunately std.stdio has fdopen and windowsHandleOpen publicly available, so it's possible to implement such functionality without changing the phobos code.

CyberShadow:
> Fortunately std.stdio has fdopen and windowsHandleOpen publicly available, so it's possible to implement such functionality without changing the phobos code.

Yes, I added them when I needed to implement something similar to this PR in my own library:
https://github.com/CyberShadow/ae/blob/7ba3063ebff069793dba01fb193ce4d1e667ecd6/sys/file.d#L1022

FreeSlave:
Btw, I made this PR only because you agreed that "Specifying flags via a string is an ancient API anti-pattern" when discussing https://github.com/dlang/phobos/pull/5510

schveiguy:
> We should definitely not use them for std.io should that ever materialize.

FYI: https://github.com/MartinNowak/io uses enums, but also uses strings as a convenience feature to specify the mode (but all is done at compile-time). I like the strings simply because they are short and to the point vs. an enum bitfield where you have to repeat the name constantly.

See: https://martinnowak.github.io/io/std/io/file/mode.html

FreeSlave:
Are we waiting for someone else's decision or I can close this?

CyberShadow:
You can try pinging Andrei/Walter, as they are the ones who have the final say.

FreeSlave:
@andralex can you tell if we need this stuff in phobos?

andralex:
cc @atilaneves

atilaneves:
I think that a dub package would be preferable, so closing.

PR #6548, opened 2018-06-06
Musl: last remaining tweaks

joakim-noah:
`tzname` is unset by default on Musl, and @jercaianu added more code below `SbrkRegion`, so need to version out that part alone for Musl now. With this pull, all Phobos tests pass with Musl on Alpine.

### DIFF
--- a/std/datetime/timezone.d
+++ b/std/datetime/timezone.d
@@ -781,7 +781,11 @@ public:
 
     @safe unittest
     {
-        assert(LocalTime().dstName !is null);
+        // tzname, called from dstName, isn't set by default for Musl.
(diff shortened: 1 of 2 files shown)


std/datetime/timezone.d
```diff
@@ -781,7 +781,10 @@ public:
 
     @safe unittest
     {
-        assert(LocalTime().dstName !is null);
+        version(CRuntime_Musl)
```
wilzbach: A comment here would probably help people reading this in the future.

joakim-noah: Done, this thread seems to confirm it's not set by default, TooTallNate/node-time#94.

PR #6550, opened 2018-06-06
Add an Assert hook to CheckedInt

wilzbach:
Okay I forgot that I created https://github.com/dlang/phobos/pull/5928 a while ago and it seems we now have three options to fix https://issues.dlang.org/show_bug.cgi?id=18024

1) https://github.com/dlang/phobos/pull/5928 - remove Warn from Abort
2) https://github.com/dlang/phobos/pull/6369 - use @trusted Warn
3) this PR - add Assert and `alias Abort(T) = Checked!(Checked!(T, Abort), Warn)`

All options have their pros/cons, but mostly they don't differ so much, so we should really start to agree on one option..

See also: https://github.com/dlang/phobos/pull/6369 and https://github.com/dlang/phobos/pull/5928

### DIFF
--- a/std/experimental/checkedint.d
+++ b/std/experimental/checkedint.d
@@ -54,6 +54,9 @@ $(BOOKTABLE ,
     $(TR $(TD $(LREF Throw)) $(TD
         fails every incorrect operation by throwing an exception.
     ))
+    $(TR $(TD $(LREF Assert)) $(TD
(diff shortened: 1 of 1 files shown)


RazvanN7:
This has been fixed by: https://github.com/dlang/phobos/pull/7909

PR #6551, opened 2018-06-08
bigint: sink param should be scope

WalterBright:
Sinks should never leak their parameter.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -2046,7 +2046,7 @@ void formatValue(Writer, T, Char)(auto ref Writer w, auto ref T val, const ref F
    {
        int x, y;
 
-       void toString(scope void delegate(const(char)[]) @safe sink,
+       void toString(scope void delegate(scope const(char)[]) @safe sink,
                      FormatSpec!char fmt) const
        {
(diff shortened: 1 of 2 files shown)


std/format.d
```diff
@@ -3656,15 +3656,15 @@ private template hasToString(T, Char)
...
     }
-    else static if (is(typeof({ T val = void; val.toString((const(char)[] s){}, "%s"); })))
+    else static if (is(typeof({ T val = void; val.toString((scope const(char)[] s){}, "%s"); })))
     {
         enum hasToString = 3;
     }
-    else static if (is(typeof({ T val = void; val.toString((const(char)[] s){}); })))
+    else static if (is(typeof({ T val = void; val.toString((scope const(char)[] s){}); })))
     {
         enum hasToString = 2;
```
wilzbach: I'm not sure whether we can modify these overload checks without a deprecation period.

WalterBright: Because of contravariance, it should still work with the non-scope versions.

wilzbach:
This is in master now.

PR #6553, opened 2018-06-08
LockingTextWriter() should not leak its parameter

WalterBright:


wilzbach:
This is technically a breaking change. Are we sure we want to do this in such a popular module without a deprecation period?

WalterBright:
It shouldn't be a breaking change because the constraint is on LockingTextWriter(), not its parameter. BTW it's hopeless moving forward with dip1000 without this change.

WalterBright:
> inferred anyway?

The calls to formattedWrite() are such a circular thicket of delegates and templates the compiler gives up on it.

schveiguy:
> the compiler gives up on it.

Really? I didn't know that could happen. That's a bit concerning actually, since usually you depend on the compiler to determine attributes for templates.

WalterBright:
More specifically, it gives up when the inference graph is circular. There's a bugzilla issue on that posted by Andrei, but I don't have a link handy for it.
Follow the formatting templates, they go in a circle :-) there is no bottom.

schveiguy:
> it gives up when the inference graph is circular

Ah! That makes more sense, and is a different kind of problem. :)

PR #6558, opened 2018-06-08
Appender: should be scope since it copies its arguments

WalterBright:


dukc:
This does not convince me. The element itself is copied, but if it has mutable indirection it's contents are not.

You do not want to put a pointer to a local into a global array.

RazvanN7:
This function is now an alias to put, therefore this change is no longer necessary.

PR #6561, opened 2018-06-08
all put() functions should have 'scope' parameters

WalterBright:


WalterBright:
I've been replacing `in` with `scope const`.

Geod24:
Yes, but what is the plan for `in` within the language ? `in` was supposed to mean `const scope`, now it just means `const`, so it's essentially useless ATM.

WalterBright:
Yes, it is useless at the moment. No particular plans for it at the moment.

PR #6567, opened 2018-06-09
uuid: add scope

WalterBright:


std/uuid.d
```diff
@@ -1117,13 +1117,13 @@ public struct UUID
  * for strings and wstrings. It's always possible to pass wstrings and dstrings
  * by using the ubyte[] function overload (but be aware of endianness issues!).
  */
-@safe pure nothrow @nogc UUID sha1UUID(in char[] name, const UUID namespace = UUID.init)
+@safe pure nothrow @nogc UUID sha1UUID(scope const(char)[] name, scope const UUID namespace = UUID.init)
```
wilzbach: What effect does `scope` on a struct with only statically sized members have?
I thought it gets copied...

PR #6570, opened 2018-06-09
primitives: add scope

WalterBright:


wilzbach:
```
std/range/primitives.d(600): Error: `@safe` function `std.range.primitives.__unittest_L593_C7` cannot call `@system` function `std.range.primitives.put!(int[][], int[]).put`
std/range/primitives.d(610): Error: `@safe` function `std.range.primitives.__unittest_L605_C7` cannot call `@system` function `std.range.primitives.put!(int[][], int[]).put`
```

Maybe we should be a bit more careful with this one?

wilzbach:
rebased to master

PR #6576, opened 2018-06-10
RandomCover use a `@nogc` buffer to store already chosen results

ghost:
The life time of the array storing the already chosen results is deterministic so there's no need to use a built-in array, instead a small dedicated array is used. This change is far from making the `randomCover` family of functions `@nogc` but at least it avoids an alloc (fundamentally, issue 14001 is blocked by exceptions in `uniform` and in the different RNGs).

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -2644,10 +2644,12 @@ private struct RandomCoverChoices
     private immutable size_t _length;
     private immutable bool hasPackedBits;
 
+    void opAssign(T)(T) @disable;
+
     this(this) pure nothrow @nogc @trusted
     {
         import core.memory : pureMalloc;
-        import core.stdc.string : memmove;
+        import core.stdc.string : memcpy;
         import core.exception : onOutOfMemoryError;
 
         if (!hasPackedBits && buffer !is null)
@@ -2655,7 +2657,7 @@ private struct RandomCoverChoices
             void* nbuffer = pureMalloc(_length);
             if (nbuffer is null)
                 onOutOfMemoryError();
-            buffer = memmove(nbuffer, buffer, _length);
+            buffer = memcpy(nbuffer, buffer, _length);
         }
     }
 


std/random.d
```diff
@@ -2634,6 +2634,78 @@ do
...
+
+    this(this) nothrow @nogc @trusted @property
+    {
+        import core.stdc.stdlib : malloc;
+        import core.stdc.string : memmove;
+
+        void* nbuffer = malloc(_length);
+        if (nbuffer is null)
+            assert(0, emsg);
+        buffer = memmove(nbuffer, buffer, _length);
```
wilzbach: why not `makeArray`?

BBasile:
> Could we at least maybe use stdx.allocator?

I thought to this and concluded that it's not a good idea to use the allocators in phobos until their are promoted to `std`. (note that they are not used at all in phobos).

> There shouldn't be a need to code a custom struct every time we want sth. to be `@nogc`

Yeah. Obviously i'd use `std.container.array.Array` if it hasn't this hilarious constraint that is, every type is allowed as element but not `bool`.

wilzbach:
> I thought to this and concluded that it's not a good idea to use the allocators in phobos until their are promoted to std. (note that they are not used at all in phobos).

FWIW that shouldn't be a problem and for example std.file is using std.experimental.checkedint already.
As long as it's used internally, we should be good. Imho it's even good if we start to use it internally because only by intensive use we can detect flaws.

> Obviously i'd use std.container.array.Array if it hasn't this hilarious constraint that is, every type is allowed as element but not bool.

Ouch. Does anyone know about the history of this constraint?
Was the idea that [BitArray](https://dlang.org/phobos/std_bitmanip.html#BitArray) is a replacement?

aG0aep6G:
> Yeah. Obviously i'd use std.container.array.Array if it hasn't this hilarious constraint that is, every type is allowed as element but not bool.

The constraint is just on the first template, because there's another one that's specialized for `bool`: https://dlang.org/phobos/std_container_array.html#.Array.2

It would probably be less confusing if there was just one template, with the specialization inside.

BBasile:
Array!bool is not `@nogc` enough anyway.

BBasile:
The new custom type can be made more useful by allowing bit packed array of bool for small length (up to size_t * 8 elements to cover.

std/random.d
```diff
@@ -2634,6 +2634,104 @@ do
...
+    private immutable bool hasPackedBits;
+
+    this(this) nothrow @nogc @trusted
+    {
+        import core.stdc.stdlib : malloc;
+        import core.stdc.string : memmove;
+
+        if (!hasPackedBits && buffer !is null)
+        {
+            void* nbuffer = malloc(_length);
```
BBasile: this branch is not covered anymore since the switch to the bit field. There should be a test with 65 or more elements to cover...

BBasile:
@wilzbach, maybe you could reconsider your POV. Now that a field of bits is used the custom type is really interesting for small (<= 64) ranges to cover.

std/random.d
```diff
@@ -2634,6 +2634,104 @@ do
...
+private struct RandomCoverChoices
+{
+    private static immutable emsg = "No memory available to store random cover choices";
+    private void* buffer;
+    private immutable size_t _length;
+    private immutable bool hasPackedBits;
+
+    this(this) nothrow @nogc @trusted
+    {
+        import core.stdc.stdlib : malloc;
```
n8sh: [`core.memory : pureMalloc`, `pureCalloc`, `pureFree`](https://dlang.org/phobos/core_memory.html#.pureMalloc)?

n8sh: Although since you `assert(0)` if they fail, it would be justifiable to call "locally `pure`ified" `malloc`/etc. directly.

std/random.d
```diff
@@ -2634,6 +2634,104 @@ do
...
+        import core.stdc.string : memmove;
+
+        if (!hasPackedBits && buffer !is null)
+        {
+            void* nbuffer = malloc(_length);
+            if (nbuffer is null)
+                assert(0, emsg);
```
n8sh: In places like this it seems like the convention is to use [`onOutOfMemoryError`](https://dlang.org/phobos/core_exception.html#.onOutOfMemoryError):
```d
if (nbuffer is null)
{
    onOutOfMemoryError();
    assert(0);
}
```
...but this has the downside that your custom message would be lost.

std/random.d
```diff
@@ -2634,6 +2634,104 @@ do
...
+            buffer = memmove(nbuffer, buffer, _length);
+            if (buffer is null)
```
n8sh: This check doesn't make sense. Cut & paste error?

std/random.d
BBasile: i've moved to `onOutOfMemoryErrorNoGC` which is not pure so this is not an option anymore.
Now there's a dilemma : _pure allocs and asserts_ or _static exceptions and impure_

std/random.d
```diff
@@ -2634,6 +2634,103 @@ do
...
+        if (!hasPackedBits && buffer !is null)
+        {
+            void* nbuffer = malloc(_length);
+            if (nbuffer is null)
+                onOutOfMemoryErrorNoGC();
+            buffer = memmove(nbuffer, buffer, _length);
+        }
+    }
+
+    this(size_t value) nothrow @nogc @trusted @property
```
n8sh: Parameter name should hint at purpose.

std/random.d
n8sh: `onOutOfMemoryError` is `@nogc nothrow pure`, so why not use it instead of `onOutOfMemoryErrorNoGC`?

BBasile:
Actually i think this could close 14001. While `RandomCover` is still not `@nogc` every suggestion made in there is implemented.

std/random.d
```diff
@@ -2634,6 +2634,103 @@ do
...
+- constructed with an invariable length
+- small length means 0 alloc and bit field (if up to 32(x86) or 64(x64) choices to cover)
+- bigger length means non-GC heap allocation(s) and dealloc. +/
+private struct RandomCoverChoices
+{
+    private void* buffer;
+    private immutable size_t _length;
+    private immutable bool hasPackedBits;
+
+    this(this) pure nothrow @nogc @trusted
```
n8sh: I believe an explicit (or disabled) `opAssign` is also required.

std/random.d
```diff
@@ -2634,6 +2634,103 @@ do
...
+        import core.memory : pureMalloc;
+        import core.stdc.string : memmove;
+        import core.exception : onOutOfMemoryError;
+
+        if (!hasPackedBits && buffer !is null)
+        {
+            void* nbuffer = pureMalloc(_length);
+            if (nbuffer is null)
+                onOutOfMemoryError();
+            buffer = memmove(nbuffer, buffer, _length);
```
n8sh: You can use `memcpy` here instead of `memmove` because the destination and source cannot overlap.

PR #6578, opened 2018-06-12
Fixed typos

bausshf:


aG0aep6G:
Those are not typos. "Iff" with two Fs is a thing: https://en.wikipedia.org/wiki/If_and_only_if

PR #6579, opened 2018-06-12
Expose experimental packages on Dub

wilzbach:
I often had the need to use the latest version of a package in Phobos, but without requiring my library users to upgrade their compiler.

This simple `dub.sdl` allows the usage of the checkedint module via DUB.

Example
-------

```d
#!/usr/bin/env dub
/++dub.sdl:
dependency "phobos:checkedint" path=".."
+/

void main(string[] args)
{
    import stdx.checkedint; // From latest Phobos
    import std.stdio; // DMD's Phobos
    writeln("checkedint: ", 2.checked + 3);
}
```

Motivation
----------

Of course, the main motivation is `std.experimental.allocator`, but that's a bit more complicated.

Future
------

Also, like for [dmd's dub package](http://code.dlang.org/packages/dmd) it will only be exposed as `~master` due to D's versioning not matching SemVer and thus the DUB registry rejecting the versions.

I will add a preview for the `std.experimental.allocator` package within the next days, but maybe someone already has an opinion on this?

### DIFF
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -134,6 +134,15 @@ publictests()
     make -f posix.mak -j$N publictests DUB=$DUB BUILD=$BUILD
 }
 
+# test stdx dub package
+dub_package()
+{
+    pushd test
+    dub -v --single dub_stdx_checkedint.d
+    dub -v --single dub_stdx_allocator.d
+    popd
+}
+
 case $1 in
     install-deps) install_deps ;;
     setup-repos) setup_repos ;;
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ obj/
 .vscode
 
 *.html
+
+.dub/
+libphobos_*
+source/
--- a/dub.sdl
+++ b/dub.sdl
@@ -1,6 +1,50 @@
 name "phobos"
 license "BSL-1.0"
+description "D Standard Library"
+authors "DLang Community"
 copyright "Copyright © 1999-2018, The D Language Foundation"
+
 targetType "library"
 sourcePaths "std" "etc"
 targetPath "generated"
+
+subPackage {
+  name "checkedint"
+  targetType "library"
+  preGenerateCommands `rdmd --eval='
+	auto pkgDir = environment.get("DUB_PACKAGE_DIR");
+	auto destDir = pkgDir.buildPath("source", "stdx");
+	mkdirRecurse(destDir);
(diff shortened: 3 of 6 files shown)


wilzbach:
Okay finally got around adding support for the allocator package too.
It's a bit ugly, because DUB searches through `sourcePaths` _before_ it evaluates `preGenerateCommands` (breaks kinda the points of having a command that accordingly evaluated before project generation, but well I found a workaround).

As the new [experimental collections library](https://github.com/dlang-stdx/collections) depends on the Phobos's master version of stdx-allocator, it would be great to start this experiment as soon as possible.

I added two simple dub test scripts that are tested on CircleCi, but I think only by testing this in practice we will gain experience with it.
As mentioned earlier, as of now only `~master` will be exposed, so we won't give out any guarantees or "stable" versions.

wilzbach:
https://github.com/dlang/dlang.org/pull/2391 is required for DAutoTest to pass.

ZombineDev:
As far as I can see the main complexity in the implementation of this PR is because we want to map `std.experimental` to `stdx`. But do we really want that? Also do we really need a separate dub subpackage for each phobos module? In contrast, #6587 in contrast looks much more appealing.

wilzbach:
> But do we really want that?

Yes, otherwise you will run into tons of conflicts.
There's sadly no one to tell DMD that it should prefer the user-supplied version of `std.experimental.allocator` nor no way to tell the linker how it should resolve the duplicate symbol conflicts.

> Also do we really need a separate dub subpackage for each phobos module?

Yep, Phobos is quite large (~65M) and people are typically interested in only a single module.
Also, and that's the main motivation by taking just a single module the chances of it compiling with the last "stable" DMD release or even an older release are a lot higher than trying to compile entire Phobos.
The experimental packages currently change the most and as mentioned https://github.com/dlang-stdx/collections currently only builds with DMD nightly. With this PR, we can build it with a released DMD and LDC.

>  In contrast, #6587 in contrast looks much more appealing.

Well, that one is just for auto-completion in your favorite editor.
It doesn't help in exposing Phobos modules on code.dlang.org.

wilzbach:
Okay finally managed to rebase this PR. Any reason not to start this experiment?
It would really help for dub packages that depend on the latest, greatest allocator code.

ZombineDev:
> Any reason not to start this experiment?

I don't see any!

PR #6580, opened 2018-06-12
Fix Issue 18595 & Issue 18596: replace MindstdRand0 w/SplitMix & templatize unpredictableSeed!UIntType

n8sh:
Slicing the elephant into smaller pieces. This PR differs from https://github.com/dlang/phobos/pull/6388 in that it omits entirely system entropy.

The PR's visible change is that one may now write `unpredictableSeed!ulong` and get a `ulong`. The PR's under-the-hood change is that instead of thread-local MinstdRand0 `unpredictableSeed` uses a global SplitMix.

**Q: Why prefer global shared (atomic) instance to thread-locals?**
**A:** It gives us an easy guarantee that the same `ulong` seed will not be produced by more than 1 thread except if `unpredictableSeed!ulong` is called 2^^64 times during the life of the application. The question should rather be, what advantage would thread-local storage give us here?

**Q: What about cryptography?**
**A:** This PR changes nothing about the current situation. `unpredictableSeed` is not in any sense unpredictable to an adversary willing to use mathematics.

**Q: What about speed?**
**A:** This shouldn't be a main consideration, but this implementation happens to be faster. The new `unpredictableSeed` takes about 1/3 the time with `ldc2 -O` and about 1/2 the time with `dmd -O -inline`. The story is even better if you compare the new `unpredictableSeed!ulong` to calling the old `unpredictableSeed` twice to produce a 64-bit value.

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -1083,7 +1083,7 @@ alias Mt19937_64 = MersenneTwisterEngine!(ulong, 64, 312, 156, 31,
     assert(n == 6597103971274460346);
 
     // Seed with an unpredictable value
-    gen.seed(unpredictableSeed);
+    gen.seed(unpredictableSeed!ulong);
     n = gen.front; // different across runs
 }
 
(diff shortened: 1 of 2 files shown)


thewilsonator:
@n8sh does this need a changelog entry?

n8sh:
@thewilsonator I'll add one.

thewilsonator:
Thanks!

n8sh:
Added.

std/random.d
```diff
@@ -1529,6 +1523,29 @@ version (AnyARC4Random)
...
+            result = (result ^ x) * m;
+        }
+        import core.thread : getpid, Thread;
+        import core.time : MonoTime;
+
+        updateResult(cast(ulong) cast(void*) Thread.getThis());
+        updateResult(cast(ulong) getpid());
+        updateResult(cast(ulong) MonoTime.currTime.ticks);
+        result = (result ^ (result >>> 47)) * m;
+        return result ^ (result >>> 47);
```
kinke: With the LLVM optimizer, this whole function is optimized to calling `Thread.getThis()`, `getpid()` and `MonoTime.currTime().ticks()` (for potential side effects) and then returning an undefined value - due to the final value of `result` depending on its initial value, which is explicitly undefined.

The intent probably was to make the boostrap seed somewhat more random, but this doesn't work here. Would it be okay to initialize `result` with 0?

kinke: Quoting Johan:
> Using an uninitialized value is UB in D. https://dlang.org/spec/declaration.html#void_init

n8sh: >Would it be okay to initialize `result` with 0?

Yes.

PR #5683, opened 2017-08-09
Fix issue 17732: SysTime.init.toString() segfaults. (alternate fix)

jmdavis:
This adds a special TimeZone type internal to SysTime which _timezone is
directly initialized to so that SysTime.init will work without
segfaulting but will still be uinquely identifiable with the is
operator.

The new TimeZone type - InitTimeZone - is internal to SysTime and can
only be obtained by the timezone getter on SysTime.init. It acts the
same as UTC except that it is not special-cased by the to*String
functions and thus will print out its timezone as +00:00 instead of z,
which is perfectly legitimate per the spec. And as such, there are no
extra checks anywhere because of this change, and this should have zero
impact on any existing code beyond the fact that SysTime.init will now
work with functions that require its TimeZone object rather than
segfaulting when the code attempts to use it.

Unlike previous attempts along these lines, this does not make it so
that SysTime.init has NaN behavior such that any operation (other than
assignment) on an an uninitialized SysTime would result in SysTime.init,
and the timezone setter property does not set the SysTime to
SysTime.init if it's passed this TimeZone. So, unfortunately, it _is_
possible to have other SysTime values with the special TimeZone, but it
was deemed unnecessarily complex for too little benefit to add the NaN
behavior. And regardless, SysTime.init is still uniquely identifiable
via the is operator. It's just that it can't technically be uniquely
identified by the timezone getter, which was never a supported feature
anyway.

PR #6581, opened 2018-06-13
Fix Issue 12507 - SysTime.init.toString should not segfault

n8sh:
Unbelievably, [this bug was reported on April 2, 2014](https://issues.dlang.org/show_bug.cgi?id=12507). I doubt it is of great interest to anyone precisely what is displayed by `SysTime.init.toString`, but it shouldn't cause a segfault.

### DIFF
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -2342,7 +2342,8 @@ public:
         {
             import std.utf : toUTFz;
             timeInfo.tm_gmtoff = cast(int) convert!("hnsecs", "seconds")(adjTime - _stdTime);
-            auto zone = _timezone is null ? "" : (timeInfo.tm_isdst ? _timezone.dstName : _timezone.stdName);
+            auto zone = _timezone is null ? "" :
+                (timeInfo.tm_isdst ? _timezone.dstName : _timezone.stdName);
             timeInfo.tm_zone = zone.toUTFz!(char*)();
         }
 
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -2336,13 +2336,13 @@ public:
         timeInfo.tm_year = dateTime.year - 1900;
         timeInfo.tm_wday = dateTime.dayOfWeek;
         timeInfo.tm_yday = dateTime.dayOfYear - 1;
-        timeInfo.tm_isdst = _timezone.dstInEffect(_stdTime);
+        timeInfo.tm_isdst = _timezone ? _timezone.dstInEffect(_stdTime) : 0;
 
         version(Posix)
         {
             import std.utf : toUTFz;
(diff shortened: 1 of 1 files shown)


std/datetime/systime.d
```diff
@@ -2336,13 +2336,13 @@ public:
...
         timeInfo.tm_yday = dateTime.dayOfYear - 1;
-        timeInfo.tm_isdst = _timezone.dstInEffect(_stdTime);
+        timeInfo.tm_isdst = _timezone ? _timezone.dstInEffect(_stdTime) : 0;
 
         version(Posix)
         {
             import std.utf : toUTFz;
             timeInfo.tm_gmtoff = cast(int) convert!("hnsecs", "seconds")(adjTime - _stdTime);
-            auto zone = (timeInfo.tm_isdst ? _timezone.dstName : _timezone.stdName);
+            auto zone = _timezone is null ? "" : (timeInfo.tm_isdst ? _timezone.dstName : _timezone.stdName);
```
quickfur: Nitpick: wrap the long line.

schveiguy:
Technically, this should be fixed by making the default timezone printable. But that ran into compiler errors: #5683

quickfur:
Ugh.  I hate compiler bugs. :-(

wilzbach:
CC @jmdavis.
Should we go with this solution as the proper one seems to be out of reach for the foreseeable future?

jmdavis:
> Should we go with this solution as the proper one seems to be out of reach for the foreseeable future?

Since I have no idea how to get anyone to actually fix the bug blocking the correct solution, and nothing I've tried thus far has managed to work around the bug, we probably should go with a worse solution in the short term.

However, these changes don't cover all of the places that `_timezone` is used, and I really think that some comments are required so that it will be easier to remove the unnecessary changes when the blocking bug is finally fixed. It will probably be quicker for me to just do the PR myself than to explain what I think should be done. I probably should have done it a while ago, but I keep hoping that the blocking bug will be fixed, and I hate adding null checks to fix this. Unfortunately, that `Rebindable` in `SysTime` has historically run into a number of bugs.

I'll put something together and either have it up in a little bit here or in the morning.

n8sh:
>It will probably be quicker for me to just do the PR myself than to explain what I think should be done.

Pushing to this PR is enabled for this reason. No need to waste time explaining small things, just change them.

jmdavis:
I went with a hacked up version of the "correct" solution that introduces `InitTimeZone` for `SysTime.init` but uses a helper function that does the null check everywhere that a function would be called on `_timezone` and returns `InitTimeZone()` instead. Once the compiler bug has been fixed, the helper function can be removed, but in the interim, that's the closest that I can get to implementing the "correct" solution given the compiler bug that's blocking it:

https://github.com/dlang/phobos/pull/6591

PR #6583, opened 2018-06-15
Added opEquals to Nullable(T, T nullValue)

bausshf:
Right now comparing a `Nullable(T, T nullValue)` with another `Nullable(T, T nullValue)` will cause an `AssertError`, unlike comparing `Nullable(T)` with another `Nullable(T)`. They should both have same behavior.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -3431,6 +3431,36 @@ Params:
     assert(!npi.isNull);
 }
 
+	/**
+      If they are both null, then they are equal. If one is null and the other
+      is not, then they are not equal. If they are both non-null, then they are
+      equal if their values are equal.
+      */
+    bool opEquals()(auto ref const(typeof(this)) rhs) const
+    {
+        if (isNull)
+            return rhs._isNull;
+        if (rhs._isNull)
+            return false;
+        return _value == rhs._value;
+    }
+
+    /// Ditto
+    bool opEquals(U)(auto ref const(U) rhs) const
+    if (is(typeof(this.get == rhs)))
+    {
+        return isNull ? false : rhs == _value;
+    }
+
+@safe unittest
+{
+	Nullable!(uint, uint.max) val3;
+	Nullable!(uint, uint.max) val4;
+	assert(val3.isNull);
+	assert(val4.isNull);
+	assert(val3 == val4);
+}
+
 /**
 Gets the value. `this` must not be in the null state.
 This function is also called for the implicit conversion to `T`.


std/typecons.d
```diff
@@ -3431,6 +3431,36 @@ Params:
...
+        if (isNull)
+            return rhs._isNull;
+        if (rhs._isNull)
+            return false;
+        return _value == rhs._value;
+    }
+
+    /// Ditto
+    bool opEquals(U)(auto ref const(U) rhs) const
+    if (is(typeof(this.get == rhs)))
```
quickfur: This doesn't seem right. Shouldn't it be:
````
if (is(typeof(this.get) == typeof(rhs)))
````
?

Comment (bausshf) [reply std/typecons.d]: That is possible, it's taken from `Nullable(T)` which means the issue will exist within both types.

Will change it for both if that would be correct. Will wait for a more official answer to that.

Comment (schveiguy) [reply std/typecons.d]: No, it's right, it's asking can `this.get() == rhs` compile.

You wouldn't want the other way, because then e.g. `Nullable!int` couldn't be compared to `1L`

File: std/typecons.d [hunk 2]
```diff
@@ -3431,6 +3431,36 @@ Params:
...
+      equal if their values are equal.
+      */
+    bool opEquals()(auto ref const(typeof(this)) rhs) const
+    {
+        if (isNull)
+            return rhs.isNull;
+        if (rhs.isNull)
+            return false;
+        return _value == rhs._value;
+    }
```
schveiguy: This can be simplified to return `_value == rhs._value`

std/typecons.d
schveiguy: I think this is correct, but it may be surprising in some cases. For example:
```D
enum NullValue = int.max;
Nullable!(int, NullValue) n;

assert(n == NullValue); // fails
```

However, I also don't know that we want to make this true either. I think the proposed behavior is at least consistent with the other Nullable.

std/typecons.d
schveiguy: And by that I mean, the whole function can reduce to that :) I realize reading my comment may not be clear.

std/typecons.d
```diff
@@ -3431,6 +3431,32 @@ Params:
...
 }
 
+    /**
+      If they are both null, then they are equal. If one is null and the other
+      is not, then they are not equal. If they are both non-null, then they are
+      equal if their values are equal.
+      */
+    bool opEquals()(auto ref const(typeof(this)) rhs) const
+    {
+        return _value == rhs._value;
```
schveiguy: Ugh, I hate to do this to you, but actually there needs to be one static if: if `_value` is a floating point, then this won't work, but `isNull` will. So you need to duplicate what `isNull` does in that instance.

std/typecons.d
```diff
@@ -3452,7 +3452,7 @@ Params:
         return isNull ? false : rhs == _value;
     }
 
-@safe unittest
+@system unittest
```
schveiguy: This should not be necessary. And in fact, there are other places it's `@safe`. Something is wrong here, it appears the compiler isn't properly inferring safety.

schveiguy:
At line 3483, you need to change the assertThrown call. This is a documented unittest, so I'd suggest:
```D
    // even though -1 is the null value, comparing with any int is false
    // because ni.isNull is true.
    assert(ni != -1);
```

schveiguy:
OK, doing some digging, it appears there is a compiler error causing it to give up inferring attributes for some reason. We may not be able to fix this without a compiler change. Problem is, I don't know why it's not doing the inference, so it may be hard to make a reduced case.

schveiguy:
The compiler bug has something to do with a circular dependency. The compiler is thinking it can call the second form of opEquals with typeof(this).

I was able to silence the error by changing the template constraint:

```D
    bool opEquals(U)(auto ref const(U) rhs) const
    if (!is(typeof(rhs) == const(typeof(this))) && is(typeof(this.get == rhs)))
```

I'm not sure why it's not required for the version that doesn't take a nullable sentinel value.

std/typecons.d
FeepingCreature: At least should be `__traits(compiles, this.get == rhs)` to avoid ambiguity, right?

schveiguy: Can you explain the ambiguity? I'm always lost when it comes to the differences between `__traits(compiles` and `is(typeof`

FeepingCreature: I mean, to avoid the confusion with `is(typeof() == typeof())`.

schveiguy: But there *are* subtle differences. I don't remember what they are, but someone (maybe Timon or Stefan?) has outlined why you should prefer one or the other, I don't have the link to the NG post handy. I'd prefer leaving it the way it is for this PR, since it's already that way for the other version of `Nullable`, and we can explore changing the `is(typeof` to `__traits(compiles` across the whole module in another PR.

std/typecons.d
n8sh: We can just do:
```d
bool opEquals()(auto ref const(typeof(this)) rhs) const
{
    // We check __traits(compiles) in case opEquals is not CTFE-able.
    static if (__traits(compiles, { static assert(nullValue == nullValue); }))
        return _value == rhs._value;
    else
        // Note that it is wrong to use this branch for floating point T
        // when nullValue is not NaN.
        return _value is rhs._value;
}
```

dukc: @n8sh No. That else block would malfunction if `typeof(_value)` is a class with custom opEquals and neither of the operands are null. Otherwise, no strict opinions on this.

bausshf:
Sorry, I have been a lot busy with a lot other stuff, but I will get back to the rest of the modifications needed later today since I got some time on my hand.

PR #6587, opened 2018-06-17
add a DUB description

BBasile:
This addition is a convenience for the IDE users, so that we have phobos in the project explorer (or whatever it is called...) rather than a way to build and test (even if this works).


wilzbach:
Good idea. Thanks!

PR #6590, opened 2018-06-18
Fix Issue 13300 - Removed pure from Appender.reserve and Appender.ensureAddable

trikko:
Pureness of those methods prevent use of  "std.array : array" function with ranges if elements have a non pure postblit.
Added a test to check that code still works with pure code / postblit as well.

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -2967,7 +2967,7 @@ if (isDynamicArray!A)
             ensureAddable(newCapacity);
         }
     }
-    
+
 
     /**
      * Returns: the capacity of the array (the maximum number of elements the
@@ -3340,14 +3340,14 @@ if (isDynamicArray!A)
     app.put(r[]);
 }
 
-unittest // Issue 13300
-{ 
+@system unittest // Issue 13300
+{
     // Issue 13300: appender must work with non pure postblit
     import std.array : array;
 
     struct Simple13300
     {
-        @disable this(); 
+        @disable this();
         this(this) { import std.stdio : stdout; stdout; }  // Not pure!
 
         private:
@@ -3370,11 +3370,11 @@ unittest // Issue 13300
 {
     // Issue 13300: appender must work with pure code as well
     import std.array : array;
-   
+
     struct Simple13300(bool disableDefaultCtor)
     {
         pure:
-        static if(disableDefaultCtor) @disable this();
+        static if (disableDefaultCtor) @disable this();
         this(this) {  }
 
         private:


trikko:
It solves https://issues.dlang.org/show_bug.cgi?id=13300
Pureness of both method should not be forced but detected by compiler, I think.

std/array.d
```diff
@@ -2955,7 +2955,7 @@ if (isDynamicArray!A)
      * Params:
      *     newCapacity = the capacity the `Appender` should have
      */
-    void reserve(size_t newCapacity) @safe pure nothrow
+    void reserve(size_t newCapacity) @safe nothrow
```
jmdavis: If the issue is that whether the function works as `pure` or not based on the template argument, then the function shouldn't be marked with `@safe` or `nothrow` either. All such attributes should be inferred. So, you should remove all attributes from both functions and ensure that they're tested appropriately in the unit tests, not just `pure`.

std/array.d
```diff
@@ -3339,6 +3340,69 @@ if (isDynamicArray!A)
     app.put(r[]);
 }
 
+unittest // Issue 13300
```
jmdavis: For better or worse, all `unittest` blocks are supposed to be marked with `@safe` or `@system`. So, if the tests work with `@safe`, then the `unittest` block should be marked as `@safe`, and if they don't, then it should be marked as `@system`.

std/array.d
```diff
@@ -3339,6 +3340,69 @@ if (isDynamicArray!A)
...
+{ 
+    // Issue 13300: appender must work with non pure postblit
+    import std.array : array;
+
+    struct Simple13300
```
jmdavis: I don't think that it makes much difference in this case, but any struct or class declared inside a function which isn't declared as `static` has access to the function's scope and can sometimes screw with tests. So, it would be better to marke each of these structs as `static`.

std/array.d
```diff
@@ -3339,6 +3340,69 @@ if (isDynamicArray!A)
...
+    {
+        @disable this(); 
+        this(this) { import std.stdio : stdout; stdout; }  // Not pure!
```
jmdavis: Please just use locally declared function that's not `pure` rather than importing std.stdio. Otherwise, you're impacting compile times for little benefit by importing a module that you don't need.

std/array.d
```diff
@@ -3339,6 +3340,69 @@ if (isDynamicArray!A)
...
+
+        private:
+        this(int tmp) { }
```
jmdavis: What is the purpose of disabling `this()` and declaring a constructor? At first glance, it looks to me like they're completely unnecessary and that simply using the `init` value in the range would be fine.

std/array.d
```diff
@@ -2967,6 +2967,7 @@ if (isDynamicArray!A)
             ensureAddable(newCapacity);
         }
     }
+
```
n8sh: Please `amend` this away. Then this PR can be merged.

std/array.d
```diff
@@ -3339,6 +3340,69 @@ if (isDynamicArray!A)
...
+        this(int tmp) { }
+    }
+
+    struct Range13300
+    {
+        @property front() { return Simple13300(0); }
+        void popFront() { count++; }
+        @property empty() { return count > 3; }
+        size_t count;
+    }
```
jmdavis: I'd suggest that you just use `std.range.range.only` instead and not declare a range type just for this. Something like `assert(only(Simple1330.init, Simple1330.init).array.length == 2);` should work just as well with less code.

std/array.d
n8sh: Actually, I'll just make the changes for CircleCI. Then please squash.

std/array.d
```diff
@@ -3339,6 +3340,69 @@ if (isDynamicArray!A)
...
+    // Test ensureAddable workaround for disable @this().
+    {
+        Range13300!true r;
+        assert(r.array.length == 4);
+    }
+
+    {
+        Range13300!false r;
+        assert(r.array.length == 4);
+    }
```
jmdavis: Similarly here, why are you doing all of this stuff with `@disable this();`? Isn't this just supposed to be testing that the code is still `pure` if the postblit is `pure`? It seems to me that this code should be simplified so that that's what it's testing and use `only` to shorten it. All of this could just be a few lines of code. And of course, it needs to test that it's `@safe` and `nothrow` with those attributes removed as mentioned above.

trikko: On ensureAddable() there's a workaround for struct with default ctor disabled. That's where the bug pop out. So I check both with and without default ctor disabled to be sure it works in both cases.

std/array.d
trikko: See comment below

std/array.d
trikko: I think this require a bigger decision because probably we should remove all attributes from a lot of places. This is just a fix for bug 13300.

std/array.d
trikko: I don't know how to modify global state from inside a unittest block in a trivial way. Probably I'm missing something.

std/array.d
trikko: See comment below

n8sh:
@trikko When you squash this PR, also rename it so it starts with "Fix Issue 13300 - " so dlang-bot will automatically connect it to the issue.

PR #6591, opened 2018-06-18
Fix issue 12507: SysTime.init.toString() segfaults.

jmdavis:
This adds a special TimeZone type internal to SysTime which _timezone is
directly initialized to so that SysTime.init will work without
segfaulting but will still be uniquely identifiable with the is
operator. Or at least, _timezone is _supposed_ to be directly
initialized to it, but issue# 17740 currently prevents that. So, the
private accessor timezoneHack has been added for the moment, and any
place where _timezone would have a function called on it, timezoneHack
is called instead. timezoneHack does the null check and returns
InitTimeZone() for SysTime.init. Once issue# 17740 has been fixed,
timezoneHack will be unnecessary and should be removed in favor of
directly initializing _timezone with InitTimeZone().

The new TimeZone type - InitTimeZone - is internal to SysTime and can
only be obtained by the timezone getter on SysTime.init. It acts the
same as UTC except that it is not special-cased by the to*String
functions and thus will print out its timezone as +00:00 instead of z,
which is perfectly legitimate per the spec. And as such, if _timezone
were directly initialized with InitTimeZone(), there would be no extra
checks due to this change, and everything would just work. However,
until issue# 17740 is fixed, there will be an extra null check any time
that a function is called on _timezone, because timezoneHack will be
called instead, and it does the null check.

Unlike previous attempts along these lines, this does not make it so
that SysTime.init has NaN behavior such that any operation (other than
assignment) on an an uninitialized SysTime would result in SysTime.init,
and the timezone setter property does not set the SysTime to
SysTime.init if it's passed this TimeZone. So, unfortunately, it _is_
possible to have other SysTime values with the special TimeZone, but it
was deemed unnecessarily complex for too little benefit to add the NaN
behavior. And regardless, SysTime.init is still uniquely identifiable
via the is operator. It's just that it can't technically be uniquely
identified by the timezone getter, which was never a supported feature
anyway.

### DIFF
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -2003,7 +2003,14 @@ public:
     @property bool dstInEffect() @safe const nothrow
     {
         return _timezone.dstInEffect(_stdTime);
-        // This function's unit testing is done in the time zone classes.
+    }
+
+    // This function's full unit testing is done in the time zone classes, but
+    // this verifies that SysTime.init works correctly, since historically, it
+    // has segfaulted due to a null _timezone.
+    @safe unittest
+    {
+        assert(!SysTime.init.dstInEffect);
     }
 
 
@@ -2014,6 +2021,14 @@ public:
     @property Duration utcOffset() @safe const nothrow
     {
         return _timezone.utcOffsetAt(_stdTime);
+    }
+
+    // This function's full unit testing is done in the time zone classes, but
+    // this verifies that SysTime.init works correctly, since historically, it
+    // has segfaulted due to a null _timezone.
+    @safe unittest
+    {
+        assert(SysTime.init.utcOffset == Duration.zero);
     }
 
 
@@ -2342,7 +2357,7 @@ public:
         {
             import std.utf : toUTFz;
             timeInfo.tm_gmtoff = cast(int) convert!("hnsecs", "seconds")(adjTime - _stdTime);
-            auto zone = (timeInfo.tm_isdst ? _timezone.dstName : _timezone.stdName);
+            auto zone = timeInfo.tm_isdst ? _timezone.dstName : _timezone.stdName;
             timeInfo.tm_zone = zone.toUTFz!(char*)();
         }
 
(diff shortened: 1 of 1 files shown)


std/datetime/systime.d
```diff
@@ -9031,22 +9069,64 @@ private:
...
+
+        override long tzToUTC(long adjTime) @safe const nothrow @nogc { return 0; }
+
+        override Duration utcOffsetAt(long stdTime) @safe const nothrow @nogc { return Duration.zero; }
+
+    private:
+
+        this() @safe immutable pure
+        {
+            super("SysTime.init's timezone", "SysTime.init's timezone", "SysTime.init's timezone");
```
n8sh: If there's a chance that this string is ever going to be displayed in an application, I suggest instead something like `"uninitialized timezone"`.

jmdavis: Why? That seems less clear to me. The problem isn't that the timezone was uninitialized (especially if the compiler bug is finally fixed). The problem is that `SysTime.init` is being used instead of giving the variable an actual, useful value.

n8sh: >Why?

Because "SysTime.init's timezone" it's meaningless to the end user of the application. If it's only for debugging purposes and `TimeZone.getTimeZone` that's fine. Otherwise it needs to be written for humans.

n8sh: Are there any legitimate uses of `SysTime.init`?

jmdavis: Not really. It indicates that the `SysTime` hasn't been initialized yet, and you could use it if you needed to subtract from midnight, January 1st, 1 AD, but realistically, it only gets used when someone is operating on a `SysTime` that hasn't been properly initialized, and occasionally, such values get printed (which currently segfaults), so having that not blow up would be nice, but really, all `SysTime.init` is there for is to be the default value, not to be actually used for anything beyond that.

std/datetime/systime.d
```diff
@@ -9031,22 +9069,64 @@ private:
...
+        assert(SysTime.init.toISOExtString() == "0001-01-01T00:00:00+00:00");
+        assert(SysTime.init.toSimpleString() == "0001-Jan-01 00:00:00+00:00");
+        assert(SysTime.init.toString() == "0001-Jan-01 00:00:00+00:00");
     }
-    +/
 
 
     long  _stdTime;
     Rebindable!(immutable TimeZone) _timezone;
+    //Rebindable!(immutable TimeZone) _timezone = InitTimeZone();
```
n8sh: What is this?

jmdavis: As the comment above states, this is what `_timezone` is supposed to be set to but can't currently be set to until the compiler bug is fixed.

schveiguy:
I would suggest a different approach -- change `_timezone` to an accessor, and add something like `_tzStorage` to back it. I don't like the name "timezoneHack" and there's a lot of work here modifying code that doesn't need to be there (and will need to be modified when the compiler is fixed). Plus review is much simpler both here and on the other end (right now, I have to search through the file for instances of `_timezone` to make sure you got them all).

jmdavis:
I don't really like having to add a setter like that, but it should end up being inlined, and naming the accessor `_timeline` does reduce the number of changes nicely. So, I've updated the code accordingly. Now, hopefully that compiler bug will actually be fixed at some point here...

jmdavis:
Are there any objections to the current version of these changes, or is this good to go?

schveiguy:
Thanks, sorry I dropped the ball on this. Looks good.

PR #6594, opened 2018-06-18
Fix issue 19003: don't call toString on T.init

FeepingCreature:
Avoid actually calling `toString` when speculatively formatting to check for format exceptions.

`Args.init` is not a constructed value; hence it is not safe to call methods on it.

Two-step solution:

1. avoid actually calling `toString` when writing into `NullSink` (why bother?)

2. call `formattedWrite` to write into `NullSink` in `checkFormatException`.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -3769,38 +3769,46 @@ private template hasToString(T, Char)
     static assert(hasToString!(L, char) == 0);
 }
 
+// Like NullSink, but toString() isn't even called at all. Used to test the format string.
+private struct NoOpSink
+{
+    void put(E)(scope const E) pure @safe @nogc nothrow {}
+}
+
 // object formatting with toString
 private void formatObject(Writer, T, Char)(ref Writer w, ref T val, const ref FormatSpec!Char f)
 if (hasToString!(T, Char))
 {
     enum overload = hasToString!(T, Char);
 
+    enum noop = is(Writer == NoOpSink);
+
     static if (overload == 6)
     {
-        val.toString(w, f);
+        static if (!noop) val.toString(w, f);
     }
     else static if (overload == 5)
     {
-        val.toString(w);
+        static if (!noop) val.toString(w);
     }
     // not using the overload enum to not break badly defined toString overloads
     // e.g. defining the FormatSpec as ref and not const ref led this function
     // to ignore that toString overload
     else static if (is(typeof(val.toString((scope const(char)[] s){}, f))))
     {
-        val.toString((scope const(char)[] s) { put(w, s); }, f);
+        static if (!noop) val.toString((scope const(char)[] s) { put(w, s); }, f);
     }
     else static if (is(typeof(val.toString((scope const(char)[] s){}, "%s"))))
     {
-        val.toString((const(char)[] s) { put(w, s); }, f.getCurFmtStr());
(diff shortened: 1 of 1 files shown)


FeepingCreature:
ping

FeepingCreature:
ping

FeepingCreature:
I didn't want the change to become too verbose; this way it's structurally easy to parse. I don't know how else I'd do it given that the final else should still always run.

std/format.d
```diff
@@ -4004,7 +4008,27 @@ version(unittest)
...
+{
+    struct S
+    {
+        int i;
+
+        @disable this();
+
+        invariant { assert(this.i); }
+
+        this(int i) @safe in { assert(i); } body { this.i = i; }
```
Geod24: `s/body/do/`

FeepingCreature:
ping

n8sh:
>1. avoid actually calling toString when writing into NullSink (why bother?)

Perhaps because someone is calling it for a side-effect (like timing a formatting operation) or because someone is calling it to make sure that the code runs. Neither of those happen if the code is blocked out with `static if`. If `NullSink` has any purpose at all, then you probably shouldn't be doing this.

FeepingCreature:
It seemed kind of redundant to implement a CompileErrorTestSink { NullSink ns; alias ns this; }, but sure.

n8sh:
It's not obvious at a cursory glance: can you explain why the "invalid format" check still works with the code behind `static if (!noop)` not being called or compiled?

FeepingCreature:
The "invalid format" check has largely nothing to do with toString(); rather, it checks for the format string being valid. However, like the Nullable but more so, this relies on T.init being a valid value that you can call methods on. (I'm increasingly becoming convinced that the case of structs with invalid T.init was never intended - so I really don't understand why you can @disable this() in the first place...) That said, the failure of T.init.toString() does not tell us much about formatting an arbitrary T - a toString() that simply always throws an exception is not a very common problem. So this PR assumes that invalid T.inits are more common than toStrings that always throw.

FeepingCreature:
@n8sh Replaced `NullSink` with `NoOpSink`, which is a copypaste of `NullSink`.

FeepingCreature:
oops, tiny fix - readded a comment I accidentally removed

FeepingCreature:
@n8sh note that it won't merge despite that label due to the lack of approving review. What do?

Geod24:
Well there is also a CircleCI failure

FeepingCreature:
Yeah, that sucks for CircleCI.

It has nothing to do with this PR though.

Geod24:
> It has nothing to do with this PR though.

A rebase would help then

> edit: Oh you mean it won't merge cause a required check is failing?

Yes, and people are less eager to review a PR which is red.

FeepingCreature:
> A rebase would help then

Would it? It's been red for about a week at this point.

wilzbach:
> Would it? It's been red everywhere for about a week at this point.

No need to waste resources on this. See #ci or check https://github.com/dlang/phobos/pull/6626

wilzbach:
Merged the CircleCi PR and restarted it. It should pass now. Jenkins is still a PITA, but we're planning to switch to Buildkite.
Unfortunately for now Buildkite doesn't offer an easy way to have public logs, but (1) it's passing and (2) I send out an invite to @FeepingCreature which should allow him to view the logs and restart a build if it should ever fail in the future.

PR #6595, opened 2018-06-18
Add a changelog entry for the performance improvement of std.algorithmm.iteration.joiner

wilzbach:
Follow-up to https://github.com/dlang/phobos/pull/6492

Preview:

image

### DIFF
--- a/changelog/std-algorithm-iteration-joiner.dd
+++ b/changelog/std-algorithm-iteration-joiner.dd
@@ -1,0 +1,19 @@
+The performance of `std.algorithm.iteration.joiner` has been improved
+
+$(H4 DMD)
+
+$(CONSOLE
+> dmd -O -inline -release ./joiner.d && ./joiner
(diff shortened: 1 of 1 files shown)


ZombineDev:
@wilzbach feel free to force merge as Jenkins is still marked as required.

Geod24:
Can we make that gist live under the `dlang` organisation, instead of your personal account ?

wilzbach:
I'm not sure there's an easy way to do so :/
We could create a special benchmark repository, but I'm not sure it's worth the effort. 
If you are worried about the link being dead in the future, we could link to the respective PR instead (though of course you never know what will happen to GitHub in the MS-age).

Geod24:
> (though of course you never know what will happen to GitHub in the MS-age)

Well... Doesn't that reasoning applies to `gist` as well ? :)
Do you see any way to add it [here](https://gist.github.com/dlang/) ?

PR #6596, opened 2018-06-19
Remove alias this from Final

IgorStepanov:
As we declared in [DIP66](https://wiki.dlang.org/DIP66), classes and structs shouldn't mix alias this and opDispatch. Unfortunately, this issue has got in public code. Fortunately, this is an experemental code, and I hope we may fix it. This fix is need to [dmd#8378](https://github.com/dlang/dmd/pull/8378)

### DIFF
--- a/std/experimental/typecons.d
+++ b/std/experimental/typecons.d
@@ -821,10 +821,7 @@ else
 {
     struct Final
     {
-        import std.typecons : Proxy;
-
         private T final_value;
-        mixin Proxy!final_value;
 
         /**
          * Construction is forwarded to the underlying type.
@@ -1008,9 +1005,6 @@ pure nothrow @safe unittest
     auto sboth = Final!S(42, "foo", 3.14);
     assert(sboth.i == 42);
 
-    sboth.i = 24;
-    assert(sboth.i == 24);
-
     struct NestedS
     {
         int i;
@@ -1021,7 +1015,7 @@ pure nothrow @safe unittest
     static assert(!__traits(compiles, Final!NestedS(6)));
     auto s = makeFinal(NestedS(6));
     assert(s.i == 6);
-    assert(s.get == 30);
+    assert(s.get == 48);
 
     class NestedC
     {
@@ -1034,7 +1028,7 @@ pure nothrow @safe unittest
 
     auto c = makeFinal(new NestedC(6));
     assert(c.i == 6);
-    assert(c.get == 30);
+    assert(c.get == 48);
 }
 
 pure nothrow @safe unittest


IgorStepanov:
Someone, please help. Why continuous-integration/jenkins/pr-merge doesn't pass? Is it PR trouble, or checker?

wilzbach:
ci.dlang.io (the Project Tester) is unfortunately down, so it has nothing to do with your PR, sorry.

IgorStepanov:
What does it do, btw? I see many changes from my last contributing :)

wilzbach:
It runs the unittest / test suite of the top40 projects on DUB for the dmd / druntime / phobos of the respective PR.

>  I see many changes from my last contributing :)

BTW great to have you back!
Yep, lot's of things happened. We also have a Slack space for discussions now. It might be a quicker way to ask a few questions to get you up to date to the latest changes. At the moment it's invite-only and I don't have your email address, so if you want to join, just send me a short "hello" via mail ;-)

IgorStepanov:
>so if you want to join, just send me a short "hello" via mail ;-)

Done

IgorStepanov:
@andralex Hello. Three years ago we decided to ban alias this and opDispatch in the same class. Then we considered it a lesser evil. What do you think about this now?

IgorStepanov:
Why do we use Proxy, if we anyway use alias this later, BTW? Alias this should allow access to overloaded operators and members.

wilzbach:
Unfortunately @andralex is on the road and will be until the end of the month.
I haven't been around when opDispatch / multiple alias this was discussed, so I don't feel comfortable with merging this.
Has anyone been around when this was discussed?
@jmdavis @ZombineDev @quickfur maybe?

jmdavis:
I think that the real question here is whether `Final` is supposed to create a type that implicitly converts to the type that's being wrapped (or at least can implicitly convert to the `const` variant of the type being wrapped). Personally, I would have expected that conversion to work. After all, this is trying to emulate head-`const`, and it would be pretty weird for a head-`const` value to not implicitly convert to a fully `const` value.

As such, I would think that the better approach would be to fix the implementation so that it doesn't use `Proxy` - which could be a bit of work, but `Proxy` specifically is trying _not_ to have implicit conversions, whereas I would expect `Final` to have them.

IgorStepanov:
@wilzbach @jmdavis I've hacked this trouble via `opDot` (Does anyone remember if? :o) 
Looks like we don't change interface of `Final` in this PR.

wilzbach:
I'm not sure whether using `opDot` is a good idea.
`opDot` has been scheduled for deprecation [since a very long time](https://stackoverflow.com/questions/9880064/d2-what-are-semantics-of-opdot).

jmdavis:
From what I can tell, what really needs to happen is that `Final` needs to be fixed so that it doesn't use `Proxy`.

jmdavis:
`opDot` is not acceptable, because it's going away at some point here. The only reason that it's still around is basically that no one has actually bothered to deprecate it. Nothing should be using it at this point.

If `Final` cannot do what it needs to do with just `alias this`, then we need to figure out what is actually required to make it work. If that means that `opDispatch` is required, then that's an argument for not making it illegal to have both `alias this` and `opDispatch` in the same type. If it would work with multiple `alias this`es instead of using `opDispatch`, then that's an argument for implementing multiple `alias this`es before making it illegal to have a type have both `opDispatch` and `alias this`. I don't see how it can be considered acceptable for `Final` to require explicit casts to convert to the fully `const` variant of the type it's wrapping. The completely violates the concept of head-`const`, which is what `Final` is for.

wilzbach:
> defer solution to time when opDot will be deprecated.

I'm sorry, but I just submitted the [deprecation PR for `opDot`](https://github.com/dlang/dmd/pull/8406/files).

IgorStepanov:
@jmdavis Maybe we may disallow changing of a struct fields?
If I remove `opDot` all be fine, except the one case:
When Final wraps a struct (only struct, not class), we will not able to change the fields:
```
auto val = Final!S(42);
val.i = 24;
assert(val.i == 24); // OK -> Wrong
```
It looks correct: struct is not reference type and head const should be applied to fields.

D `Final!Class` is similar C++ head const`Class * const`:
```
Class * const obj = new Class();
obj->a = 42; // OK
obj = new Class(); // NG
```

D `Final!Struct` should be similar C++ `const Struct`, because D struct is a value-type:
```
const Struct obj;
obj.a = 42; //NG
obj = Struct(); //NG
```

jmdavis:
What I would expect for struct members that were value types would be that it would not be legal to mutate them (since they should be `const` as they're directly in the struct and thus part of the "head" that's `const` in head-`const`) but that you could mutate any indirections that a non-value type member had - e.g. a member of type `int*` could not have the pointer itself mutated, whereas what it pointed to could be. However, I don't know how that currently works with `Final` or what is actually possible, since `Final` is basically trying to hack up the type system to add something that it doesn't inherently support. If `Final` could be made to work that way without `opDot`, then that sounds reasonable. Regardless, if you can mutate what is directly inside a struct wrapped by `Final`, then that would definitely be violating head-`const`.

`Final` needs to prevent the mutation of anything directly inside what it's wrapping while allowing the mutation of anything referred to by what's wrapped, and it needs to be implicitly convertible to the fully `const` version of what it's wrapping.

IgorStepanov:
Now Final defines property, which returns r-value stored value.
```
struct Final(T)
{
        T final_value; 
        inout(T) final_get() inout
        {
                return final_value;
        }
        alias final_get this;
}
```
If Final stores reference type, you may change it via alias this:
```
class Class
{
    int i;
    this (int i) {this.i = i;}
}
auto val = Final!Class(new Class(42));
val.final_get().i = 30; // OK, changed i by reference
                                  // final_get() returns copy of final_value, but it is a reference

val.i = 20; // [1] similar previous

struct Struct
{
    int i;
    Class c;
}

auto val2 = Final!Struct(Struct(24, new Class(42)));
val2.c.i = 99; // [2] OK, because .final_get() returns a copy of struct, which contains a copy of reference
val2.i = 24; // [3] Wrong, we just change a copy
```

Now `[3]` is works via `opDispatch` and you may change a struct field.
This PR breaks `[3]` but stay correct `[1]` and `[2]`.

wilzbach:
> What I would expect for struct members that were value types would be that it would not be legal to mutate them

AFAICT it should be legal to mutate the members.
The idea of `Final` is to imitate `final` from D1/Java (e.g. https://en.wikipedia.org/wiki/Final_(Java)).
See also the first example here: https://dlang.org/phobos/std_experimental_typecons.html#.Final

> This PR breaks [3] but stay correct [1] and [2].

Hmm, we do need [3] too. So I guess `opDispatch` and `alias this` need to be allowed to work for the same struct?

> Maybe we may disallow changing of a struct fields?

AFAICT that wouldn't be `Final` anymore, but more like the C++ head const. This does have a valid use case, but that's not the use case `Final` is supposed to tackle (e.g. ensuring that a variable can only be set once).

jmdavis:
If you can set the members of a struct, then you can set the the value of the struct more than once, because the members are part of the struct. Classes are different, because `Final` contains a reference to the class, not the class object itself. Java doesn't _have_ structs, so using it as an example for how structs should work with `Final` doesn't really work.

I don't see any difference between Java's `final` and head-`const`.

IgorStepanov:
>So I guess opDispatch and alias this need to be allowed to work for the same struct?

Three years ago was a big discussion about it: https://forum.dlang.org/post/m74pg8$gl4$1@digitalmars.com

The main objection: alias this is a subtyping. Inheritance is also subtyping. And subtyping shoud have a higher priorty than opDispatch, or at least cause conflict when the same field can be accessed through opDispatch and alias this.

In current case we see that opDispatch silently steals method from alias this. In this case it is wanted behaviour. But it another cases it can be unexpected behaviour. In other words, Final hacks typesystem for the own goals.
Maybe, in future, we may re-allow opDispatch and alias this works together, but opDispatch will get a lower priority than inheritance and alias this. If we do this exchange now, we silently break existing code. For example we break Final without any error message. Thus we need to disallow works opDispatch and alias this together now.

wilzbach:
> I am not happy with the fact that I break the existing code,

I'm pretty sure that your DMD PR needs to trigger a deprecation first.
After all, [people may depend on this in their projects](https://forum.dlang.org/post/tdpixlovhxebwatjaxmt@forum.dlang.org) too and we can't break their projects from one DMD version to the next.
For more details on this, there's currently a DIP in its final evaluation stage:
https://github.com/dlang/DIPs/blob/master/DIPs/DIP1013.md

jmdavis:
> I'm pretty sure that your DMD PR needs to trigger a deprecation first.

I don't know if that's true or not for std.experimental. Certainly, it makes no sense whatsoever to apply the normal deprecation period to code in std.experimental. As I understand it, we have broken code in std.experimental.allocator with no warning (which is why there's the dub package which isn't as up-to-date). But I don't know what the actual policy with regards to std.experimental is.

wilzbach:
> I don't know if that's true or not for std.experimental. Certainly, it makes no sense whatsoever to apply the normal deprecation period to code in std.experimental

Oh I wasn't talking about std.experimental, but the compiler changes that @IgorStepanov plans to do in dmd#8378 which might affect other projects (unfortunately the Project Tester is currently offline).
And yes AFAICT we can break things freely in std.experimental and have repeatedly done so in the past.

jmdavis:
Yeah, I misread what you wrote.

IgorStepanov:
Ok, I've removed the limitation from dmd PR and I will add deprecation in another PR.

RazvanN7:
@IgorStepanov @wilzbach @jmdavis This PR does not seem to break any code. Is it ok to move forward?

PR #6598, opened 2018-06-22
Document deprecation of TimeZone.getTimeZone

Alexibu:
There was documentation refering to the now deprecated method of getting one of the of the timezone implementations via the static TimeZone.getTimeZone.


jmdavis:
Okay. I made a couple of fix-ups to what `name` said on `TimeZone` and `LocalTime`, since it was incorrect with regards to the TZ Database. This is the first time that I've tried editing a PR via github, so hopefully nothing went wrong. It at least seems to be okay.

PR #6600, opened 2018-06-24
dip1000: add scope to std.conv

WalterBright:


n8sh:
Rebased.

wilzbach:
No loner required as Phobos now compiles with -dip1000.

PR #6601, opened 2018-06-24
dip1000: std.datetime.system replace 'in' with 'scope const'

WalterBright:
There are more @trusted annotations here than should be. However, the code is such a horrific tangle of templates, it so far has proved impractical to straighten it out without rewriting big chunks of Phobos. So, in order to do it piece by piece, I have to temporarily use @trusted.

### DIFF
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -419,7 +419,7 @@ public:
                        given $(REF DateTime,std,datetime,date) is assumed to
                        be in the given time zone.
       +/
-    this(in DateTime dateTime, immutable TimeZone tz = null) @safe nothrow
+    this(scope const DateTime dateTime, immutable TimeZone tz = null) @safe nothrow
     {
         try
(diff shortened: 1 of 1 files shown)


WalterBright:
@jmdavis they do have indirections, because they were failing when compiled with dip1000. I don't know what they are, because it is not easy to penetrate the layers to find out. I would appreciate your help with this when you can - replace @trusted with @safe one by one and figure out why they fail.

jmdavis:
`SysTime` does have an indirection (it holds a `Rebindable(immutable TimeZone)`), but `DateTime` definitely doesn't. It just contains a `Date` and a `TimeOfDay` which contain integral types for the year, month, day, hour, minute, and second. So, if any `DateTime`s have to be marked with `scope` anywhere, then that sounds like a bug in the compiler. But clearly, I'll have to find time soon to look over std.datetime to strip out any `scope`s that should be unnecessary and make code `@safe` instead of `@trusted`.

PR #6602, opened 2018-06-24
dip1000: std.datetime.date replace 'in' with 'scope const'

WalterBright:


jmdavis:
> On second thought, Date, TimeOfDay and DateTime are all simple PODs without indirections. We should remove in or scope const when they're used as parameters as it's completely pointless. The only cases where scope const matters, AFAICS, are all the from*String functions.

Yes, but that can be done in another pass. This at least ensures that the -dip1000 stuff that Walter is doing works.

ZombineDev:
> Yes, but that can be done in another pass.

Alright, fair enough.

PR #6605, opened 2018-06-24
pick fix for 18993 from master and fix conflict

BBasile:
There's been a change last week with the Appender used (reserve) so the commit cant be picked  exactly from stable branch and i don't know how you do usually but it looks like the same hash cant be preserved.


Temtaime:
@BBasile thanks !

wilzbach:
We merge `stable` back into `master` with a merge commit to avoid future merge conflicts.

I created https://github.com/dlang/phobos/pull/6606

For reference, I use this script for merging stable back to master https://github.com/wilzbach/git-tools/blob/master/merge_stable
It can be run by anyone.

BBasile:
Thanks i expected this.

PR #6607, opened 2018-06-24
[RFC] Fix Issue 13683 - More precise error message for wrong lambda

wilzbach:
So I'm interested on what the consensus on this issue is.
We haven't managed to improve DMD's error message over the last four year and letting Phobos emit better error messages isn't too hard.
So should we do this for a few important functions?

### DIFF
--- a/std/algorithm/searching.d
+++ b/std/algorithm/searching.d
@@ -120,8 +120,10 @@ template all(alias pred = "a")
     Performs (at most) $(BIGOH range.length) evaluations of `pred`.
      +/
     bool all(Range)(Range range)
-    if (isInputRange!Range && is(typeof(unaryFun!pred(range.front))))
+    if (isInputRange!Range)
     {
+        static assert(is(typeof(unaryFun!pred(range.front))),
+                "`" ~ pred.stringof[1..$-1] ~ "` isn't a unary predicate function for range.front");
         import std.functional : not;
 
         return find!(not!(unaryFun!pred))(range).empty;


FeepingCreature:
This seems like there's room for a language enhancement similar to in():

    if (isInputRange!Range)
    if (is(typeof(unaryFun!pred(range.front))),
        "`" ~ pred.stringof[1..$-1] ~ "` isn't a unary predicate function for range.front"))
    {
    }

Where DMD would add the relevant messages when all templates fail to match.

thewilsonator:
Thats a really good idea, [draft DIP](https://github.com/thewilsonator/DIPs/blob/template-constraints/DIPs/DIP1xxx.md) for it.

thewilsonator:
And an [implementation(https://github.com/thewilsonator/dmd/tree/template-constraint-dip) with [test file](https://github.com/thewilsonator/dmd/blob/template-constraint-dip/test/fail_compilation/diptemplateconstraint.d).
Example 
```D
void foo()()
if (0, "zero")
if (1)
if (1, "one")
{
}

void foo()()
if (0, "zero")
if (0)
if (5, "fi"~"ve")
if (4)
{
}
void main()
{
    foo();
}
```
Outputs 

```
fail_compilation/diptemplateconstraint(31): Error: template `foo` cannot deduce function from argument types !()(), candidates are:
fail_compilation/diptemplateconstraint.d(17):        foo()()
        not satisfied: zero
            satisfied: 1
satisfied: one
fail_compilation/diptemplateconstraint.d(24):        foo()()
        not satisfied: zero
        not satisfied: 0
            satisfied: five
            satisfied: 4
```

BBasile:
@thewilsonator, the problem here is more to get the error message that's gagged, for example with this

```d
void main()
{
    [1,2,3].each!((x) => writeln(x.length));
}
```

nothing tells that the error is `x.length`.

thewilsonator:
Indeed given:
```d
bool all(Range)(Range range)
     if (is(typeof(unaryFun!pred(range.front)))
     if (isInputRange!Range)
```

with the current implementation you would get:
```
test.d(6): Error: template `std.algorithm.searching.all` cannot deduce function from argument types !((x) => writeln(x.length))(int[]), candidates are:
...
std/algortithm/searching.d(123):        all(ange)(Range range)
    not satisfied: is(typeof(unaryFun!pred(range.front))
...
```
I hope to get to
->
```
test.d(6): Error: template `std.algorithm.searching.all` cannot deduce function from argument types !((x) => writeln(x.length))(int[]), candidates are:
std/algortithm/searching.d(123):        all(ange)(Range range)
    not satisfied: is(typeof(unaryFun!pred(range.front))
        errors in: `unaryFun!pred(range.front)`
            [insert (speculative) errors]
```

but I'm not quite there yet, but I'm making progress.

BBasile:
I think that changing functions in std.algorithm that take a lambda is a nice step. What i don't like in the DIP proposed by @thewilsonator is that in a first time there's a non trivial language change (i.e we have to update dparse etc...) and it still not fixes the issue which is supposed to come in a second step.

Since it's a RFC:

1/ make the lambda error more smart
2/ maybe the DIP for constraints.

not the opposite.

BBasile:
To be perfectly clear: in a first time it's better to have still messy messages but at least in one of them the right error when the templated lambda was tried (e.g x.length like in my example).

thewilsonator:
Sorry i should have used the example 
```d
bool all(Range)(Range range)
     if (is(typeof(unaryFun!pred(range.front)),"`" ~ pred.stringof[1..$-1] ~ "` isn't a unary predicate function for range.front")
     if (isInputRange!Range)
```
which is the same functionality as this RFC, except that the overload resolution remains unchanged (not that is matters in this case). If this doesn't allay your concerns, please post them on the DIP draft review, it will be more more useful there.

To your actual comments: 

IIUC, you are saying this shouldn't wait for the DIP to go through to improve the error message? I agree with that, but I'm not in a position to approve this PR.

w.r.t the DIP, I'm reluctant to stipulate what must be done other than the necessary language changes, but the implementation will (assuming I can get it to work) include speculative errors for `is(typeof(...))` constraints.

BBasile:
I think that a library solution is possible quickly eg. in 2.084. The DIP solution would take at least 18 months and would sucks a bit in regard of the required changes in the ecosystem. My final comment is that: let's patch std.algorithm in a first time, at the condition that the patches are able to output a useful error message.

thewilsonator:
I agree, but someone else needs to approve it.

thewilsonator:
I think this can also be applied to wherever the constraint on the lambda is not required for correct overload resolution.

wilzbach:
I will add more if I get around finding a bit of free time in the next weeks. Things are quite busy these days.

BBasile:
Does someone wants to make the same for all std.algorithm functions ? Issue is not resolved until done more widely.

thewilsonator:
That would be good, the only catch is that it does change overload resolution in the general case. It's fine here because there is only one.

BBasile:
Yes of course, reviewers would check these details in the bigger PR. I just left the message to notigy people that there's more todo.

wilzbach:
I'm not really sure whether we should do this for entire Phobos. Fixing DMD would be right way to go :/

PR #6610, opened 2018-06-25
Remove const scope from value types in std.datetime.date.

jmdavis:
None of these uses of const scope should be required, because Date,
TimeOfDay, and DateTime are all value types with no indirections.

### DIFF
--- a/std/datetime/date.d
+++ b/std/datetime/date.d
@@ -164,7 +164,7 @@ public:
             date = The date portion of $(LREF DateTime).
             tod  = The time portion of $(LREF DateTime).
       +/
-    this(scope const Date date, scope const TimeOfDay tod = TimeOfDay.init) @safe pure nothrow @nogc
+    this(Date date, TimeOfDay tod = TimeOfDay.init) @safe pure nothrow @nogc
     {
         _date = date;
         _tod = tod;
@@ -233,7 +233,7 @@ public:
             $(TR $(TD this &gt; rhs) $(TD &gt; 0))
             )
      +/
-    int opCmp(scope const DateTime rhs) const @safe pure nothrow @nogc
+    int opCmp(DateTime rhs) const @safe pure nothrow @nogc
     {
         immutable dateResult = _date.opCmp(rhs._date);
 
@@ -473,7 +473,7 @@ public:
         Params:
             date = The Date to set this $(LREF DateTime)'s date portion to.
       +/
-    @property void date(scope const Date date) @safe pure nothrow @nogc
+    @property void date(Date date) @safe pure nothrow @nogc
     {
         _date = date;
     }
@@ -526,7 +526,7 @@ public:
             tod = The $(REF TimeOfDay,std,datetime,date) to set this
                   $(LREF DateTime)'s time portion to.
       +/
-    @property void timeOfDay(scope const TimeOfDay tod) @safe pure nothrow @nogc
+    @property void timeOfDay(TimeOfDay tod) @safe pure nothrow @nogc
     {
         _tod = tod;
     }
@@ -593,7 +593,7 @@ public:
 
     @safe unittest
     {
-        static void testDT(DateTime dt, int year, scope const DateTime expected, size_t line = __LINE__)
+        static void testDT(DateTime dt, int year, DateTime expected, size_t line = __LINE__)
         {
             dt.year = year;
             assert(dt == expected);
@@ -637,7 +637,7 @@ public:
 
     @safe unittest
     {
-        assertThrown!DateTimeException((scope const DateTime dt){dt.yearBC;}(DateTime(Date(1, 1, 1))));
+        assertThrown!DateTimeException((DateTime dt){dt.yearBC;}(DateTime(Date(1, 1, 1))));
 
         auto dt = DateTime(1999, 7, 6, 12, 30, 33);
         const cdt = DateTime(1999, 7, 6, 12, 30, 33);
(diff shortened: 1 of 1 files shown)


Geod24:
Just a thought: Would it make sense to keep it to prevent accidental, incompatible additions to those structs ? `scope const`, even on value type, makes it very obvious what those will be used for. I suppose that was the intent behind the original `in`, it's a shame that it was never properly implemented.

jmdavis:
> I suppose that was the intent behind the original in, it's a shame that it was never properly implemented.

The intent behind the original `in` was that at the time, I didn't understand `in` properly, and someone told me that I should use it, so I used it in several places in std.datetime where I shouldn't have. IMHO, it adds zero value here. The `scope` is pointless, because there are no indirections, and these structs will definitely never have any indirections. There's no reason to ever add them. The only change that I've ever even considered would be to add fractional seconds to `DateTime`, and even if I did that, it wouldn't add any indirections. And the `const` is pointless, because the argument is copied, and it has no indirections. All it would do would be to prevent accidentally mutating it within the function, but because it's a copy, it really doesn't matter whether it is mutated within the function.

Geod24:
You're right that both `const` and `scope` are pointless if you are familiar with the implementation of both the function receiving the argument and the arguments themselves.
The advantage I see with `scope const` is that you can just go "I don't care" as soon as you see it on a parameter list.
That being say, it's a stylistic issue, so up to Phobos devs.

n8sh:
>The advantage I see with `scope const` is that you can just go "I don't care" as soon as you see it on a parameter list.

@Geod24 right, it serves as documentation. @ZombineDev please leave this open for a bit longer for discussion.

jmdavis:
How on earth does it serve as documentation? The types in question have no indirections and are clearly value types. Anyone using the code should understand that, or they're going to have problems in general. It's ridiculous to be putting `scope` on them. Should we start marking `int` with `scope` just because it's a parameter? `scope` serves no purpose here.

n8sh:
Aside from serving as documentation, it also acts as a safeguard. If some future developer were to change `Date` to have indirections (which *could* reasonably occur: maybe someone decides that `Date` should support multiple calendar systems), the original intention will remain intact and any changes that would violate it will not compile.

jmdavis:
Then what, we're going to put `scope` on every parameter ever? It does _nothing_ here, and it makes no sense for `Date` to ever contain any indirections.

jmdavis:
The only reason that any of these parameters are marked with `scope` is because they were marked with `in` previously, and Walter is changing all of the `in`s in Phobos to `scope const`. The only reason that any of these were marked with `in` was because I didn't understand `in` properly when I originally wrote this code. `scope` does nothing here and is completely unnecessary. And this code has never been released with `scope` on it.

Thanks to what -dip1000 makes `scope` do, adding indirections to a type will now _always_ risk breaking code. Slapping `scope` everywhere to try and avoid that in some cases doesn't really fix anything. It just reduces the damage in some cases. And anyone who is changing a type to add indirections must now take that into account and deal with it appropriately. Vandalizing parameters with unnecessary attributes just in case the type gets changed later to have indirections is not the answer. Lots of code will have to be changed in that case anyway.

n8sh:
>How on earth does it serve as documentation?

It lets a developer, by looking at the auto-generated documentation, understand *something* about how the functions may be used without needing to understand everything about the implementation of each data type.

>The only reason that any of these were marked with `in` was because I didn't understand `in` properly when I originally wrote this code.

From Phobos and other code I saw in the wild I got the impression that this kind of over-specification was encouraged style - perhaps mistakenly! Anyway to minimize confusion this should be consistent with the rest of Phobos. If redundant `in` (or redundant `scope cost` post-Walter Bright's changes) is generally added to parameters then functions in `std.datetime` shouldn't stand out as an exception, and if it's not done elsewhere then it would be weird to do it here.

jmdavis:
> It lets a developer, by looking at the auto-generated documentation, understand something about how the functions may be used without needing to understand everything about the implementation of each data type.

It also gives the false impression that the types involved have indirections which actually makes understanding what's going on harder.

> From Phobos and other code I saw in the wild I got the impression that this kind of over-specification was encouraged style - perhaps mistakenly!

It's never been the Phobos style to put attributes on stuff when the attributes do nothing. The only case I can think of where it could be argued that "over"-specification is used is when we mark templated functions with attributes rather than having them be inferred in cases where the attributes don't depend on the template arguments and thus don't need to be inferred, and being explicit in those cases makes the documentation clearer. But we don't even normally do stuff like put `const` on parameters that don't require them (though with how much in Phobos is templated, `const` often can't be used anyway).

> Anyway to minimize confusion this should be consistent with the rest of Phobos. If redundant in (or redundant scope cost post-Walter Bright's changes) is generally added to parameters then functions in std.datetime shouldn't stand out as an exception, and if it's not done elsewhere then it would be weird to do it here.

Phobos in general does not put `scope` or `in` on value types. The fact that `in` was ever used here is an oddity and was a mistake on my part that arguably should have gotten caught in the original code review but was not, and I should have just removed it ages ago, but I haven't generally done much in the way of maintenance on std.datetime that wasn't necessary, so I never got around to cleaning up stuff like this like I should have. IMHO, with the introduction of a `scope` that actually does something, it becomes that much more important to make sure that the attributes aren't lying.

Also, as far as code breakage concerns go, using `scope` where it's currently unnecessary but would be necessary if a type were changed later is just masking the code breakage that would occur upon such a change and would make it harder for Phobos developers to realize that they were likely breaking user code with the change, since it's unlikely that only Phobos would be affected. If Phobos code breaks because of something now needing `scope` that didn't before, we want to catch that so that we know to either not make the change that is going to make `scope` required or so that we can do whatever needs to be done with regards to the deprecation cycle and the changelog.

PR #6611, opened 2018-06-25

MartinNowak:


wilzbach:
Looks like we first need to fix DScanner:

```
std/experimental/allocator/common.d(0:0)[warn]: Use `enum bool empty = false;` to define an infinite range.
[Inferior 1 (process 2292) exited with code 01]
```

wilzbach:
DScanner PR: https://github.com/dlang-community/D-Scanner/pull/672

wilzbach:
Auto-merge toggled on

PR #6612, opened 2018-06-26
[Experiment] Run the testsuite in parallel

wilzbach:
I'm just interested to see what would break if parallelism is used at the auto-tester.
At the moment there doesn't seem to be one:

- https://github.com/braddr/at-client/blob/master/src/do_test_phobos.sho
- https://github.com/braddr/at-client/blob/master/src/setup_env.sh

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -621,7 +621,17 @@ $(TESTS_EXTRACTOR): $(TOOLS_DIR)/tests_extractor.d | $(LIB)
 .PHONY : auto-tester-build
 auto-tester-build: all checkwhitespace
 
+UNAME:= $(shell uname -s)
+JOBS=1
+ifeq ($(UNAME), Darwin)
+	JOBS=$(shell sysctl -n hw.ncpu)
+endif
+ifeq ($(UNAME), Linux)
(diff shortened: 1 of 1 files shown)


braddr:
The auto-tester's used parallelization since approximately day 1.  The windows builds don't see a lot of it due to walter's make.  That's one of the reasons it uses gmake for the dmd tests, which do use parallelization on every host.  Each hosts's config file sets PARALLELISM -- yes, # cores can be detected, but that doesn't always mean every core should be used.  Most hosts in the fleet do n+1, but a couple do n/2 or whatever due to being shared resources.

https://github.com/braddr/at-client/blob/master/src/setup_env.sh#L18

wilzbach:
Fair enough. Sorry.

braddr:
Nothing to apologize for.  If they weren't already doing it, parallelism would be a big boost.  And it might still help the other ci's.  It's worth noting that the windows builds _are_ significantly slower (4-5x) on identical hardware.  For instance, win-farm-1 and -2 and farm-3 and -4 are all the same hardware.  The posix builds get much better utilization across the full build than the windows builds do.

PR #6613, opened 2018-06-26
Expose a copy of std.conv with the DUB package of stdx.allocator

wilzbach:
This little trick is needed as `emplaceRef` isn't public (imho it should be, but that's an issue for a different day).
With this patch, the [collections](https://github.com/dlang-stdx/collections) library builds with a stable DMD/LDC compiler as it now can use the `phobos:allocator` package.

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -36,8 +36,13 @@ subPackage {
 		auto destFile = file.replace(srcDir, destDir);
 		destFile.dirName.mkdirRecurse;
 		file.readText.replace("std.experimental.allocator", "stdx.allocator")
+			.replace("std.conv", "stdx.allocator.conv")
 		 	.toFile(destFile);
 	}
+	pkgDir.buildPath("std", "conv.d")
+		  .readText
+		  .replace("std.conv", "stdx.allocator.conv")
+		  .toFile(destDir.buildPath("conv.d"));
 	'`
 }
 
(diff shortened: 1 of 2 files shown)


ZombineDev:
We should probably split `std.conv` into a package where `to` and `emplace` are different modules. That way we could expose only `std.conv.to` on dub, as the rest of `std.conv` has many dependencies that would also need to be exposed.

dub.sdl
```diff
@@ -36,8 +36,13 @@ subPackage {
 		auto destFile = file.replace(srcDir, destDir);
 		destFile.dirName.mkdirRecurse;
 		file.readText.replace("std.experimental.allocator", "stdx.allocator")
+			.replace("std.conv", "stdx.allocator.conv")
 		 	.toFile(destFile);
 	}
+	pkgDir.buildPath("std", "conv.d")
+		  .readText
+		  .replace("std.conv", "stdx.allocator.conv")
+		  .toFile(destDir.buildPath("conv.d"));
```
ZombineDev: I'm surprised that this works, and I think this can be quite brittle in the future - e.g. `emplace` may depend on a fix or new feature in some other module. But given that we have a CI to check for such issues, I'd say we're fine for this experimental feature. But these sorts of hacks are not going to be sustainable, so sooner or later we should look into a more future-proof solution.

PR #6614, opened 2018-06-26
[WIP] Fix Issue 19027 - iota(int.min, int.max).length is incorrect

wilzbach:


schveiguy:
We should do what the working version (the one with a step) does. I'm concerned about casting to `size_t` for both items, since the types could potentially not fit in a `size_t`, even though the length fits in a `size_t`.

For example, `iota(cent.max - uint.max, cent.max)` should work (theoretically if `cent` worked), but would return 0 with your code.

The working version uses `std.conv.unsigned` to make sure the correct conversion happens.

PR #6615, opened 2018-06-26
Speedup text() wtext() and dtext() by allowing them to add chars to i…

complistic-gaff:
Speedup text() wtext() and dtext() by allowing them to add chars to it's own appender rather than using to! then coping them back.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -4211,6 +4211,25 @@ if (T.length > 0) { return textImpl!dstring(args); }
     assert(dtext(42, ' ', 1.5, ": xyz") == "42 1.5: xyz"d);
 }
 
+@safe unittest
+{
+    char  c = 'h';
+    wchar w = '你';
+    dchar d = 'እ';
+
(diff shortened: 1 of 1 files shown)


n8sh:
The Win_32_64 test for DMD [fails without any log](https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3221725&dataid=22863018&isPull=true), so I can't tell if the error is related to this PR or is just the test machine being unreliable. Make a no-diff `git commit --amend` to trigger the auto-tester again?

schveiguy:
> The Win_32_64 test for DMD fails without any log

I restarted it, we'll see what happens.

schveiguy:
FYI @n8sh you can, as a user with commit rights, do some things with the auto-tester. On the top left of the page when you go there is a Login button, and then when you are in there, there is a "deprecate" link on a test. If you deprecate the test, then it will rerun without having to push any new code.

PR #6617, opened 2018-06-28
Fix issue 16639 - Review std.json wrt this article on JSON edge cases and ambiguities

tylerknott:
The test corpus provided at https://github.com/nst/JSONTestSuite/ revealed some issues with the std.json.parseJSON function. Since addressing some of the issues required parseJSON to reject input it previously accepted, I have added a new JSONOptions.strictParsing flag so callers can opt-in to the stricter parsing.

The issues, and how I've addressed them, are listed below (approximately from most severe to least):

Silently dropping ASCII NUL characters from strings:
n_string_unescaped_crtl_char.json

This is the most serious problem I found while fixing the test cases. The current implementation of parseJSON() uses a helper function called peekChar() which can store the next character to handle in a variable of type Char (an alias of the character type). Unfortunately it was using 0 to indicate it has not read a character yet so if an ASCII NUL (which will have the value 0) is present in the text and someone reads it with peekChar() then it will effectively be skipped over, which was happening in string and whitespace parsing.

I changed peekChar() to use a Nullable!Char as the temporary storage for the next character to disambiguate the case where there is no pending unconsumed character from the case where there is a pending unconsumed ASCII NUL. In strict mode JSON with unescaped ASCII NULs in strings will throw an exception while in non-strict mode the JSON will be accepted with the NUL included in the string value.

Failure to accept ASCII DEL (0x7f) unescaped in strings:
y_string_unescaped_char_delete.json
y_string_with_del_character.json

These were the only test cases that std.json rejected that it should have accepted. This issue was addressed by changing the string parsing logic to explicitly check for character values < 0x20 instead of using std.ascii.isControl (which also returned true for 0x7f), with a special exception for ASCII NULs in non-strict mode as mentioned above.

Parsing "true", "false", and "null" tokens case-insensitively:
n_structure_capitalized_True.json

In strict mode those tokens are now parsed case-sensitively.

Accepting control characters other than ' ', '\t', '\r', and '\n' as whitespace:
n_structure_null-byte-outside-string.json
n_structure_whitespace_formfeed.json

In strict mode only the listed characters are accepted as whitespace, while non-strict mode continues to use std.ascii.isWhite with an additional exception for ASCII NUL for a similar reason as the n_string_unescaped_ctrl_char.json case (the skipWhitespace() function used peekChar() so it didn't handle ASCII NULs consistently; non-strict mode after my changes is actually more permissive than the previous behavior but it is at least consistently permissive).

Silently accepting empty data:
n_structure_no_data.json

In strict mode an exception is now thrown instead of returning an empty value.

Failure to enforce that numbers beginning with 0 cannot have any additional digits in the non-fractional part:
n_number_-01.json
n_number_neg_int_starting_with_zero.json
n_number_with_leading_zero.json

An additional check is now performed in strict mode when the whole part of a number begins with zero to ensure trailing digits are not present.

Failure to check for trailing characters after parsing:
n_array_comma_after_close.json
n_array_extra_close.json
n_multidigit_number_then_00.json
n_object_trailing_comment.json
n_object_trailing_comment_open.json
n_object_trailing_comment_slash_open_incomplete.json
n_object_trailing_comment_slash_open.json
n_object_with_trailing_garbage.json
n_string_with_trailing_garbage.json
n_structure_array_trailing_garbage.json
n_structure_array_with_extra_array_close.json
n_structure_close_unopened_array.json
n_structure_double_array.json
n_structure_number_with_trailing_garbage.json
n_structure_object_followed_by_closing_object.json
n_structure_object_with_trailing_garbage.json
n_structure_trailing_#.json

An additional check is now performed in strict mode to ensure any trailing characters after the initial JSON value are only whitespace.

In addition to the above issues, parseJSON() will throw ConvException for numbers out of the range of double/long/ulong which was not previously documented. I have updated the ddoc comment to reference that exception.

I have attached a patch with the test harness I used to run the tests. It can be applied to the JSONTestSuite repository linked above, and I plan to submit it as a pull request to that repository after this issue is handled.
[test_harness.zip](https://github.com/dlang/phobos/files/2143375/test_harness.zip)

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -6,7 +6,7 @@ JavaScript Object Notation
 Copyright: Copyright Jeremie Pelletier 2008 - 2009.
 License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
 Authors:   Jeremie Pelletier, David Herberth
-References: $(LINK http://json.org/)
+References: $(LINK http://json.org/), $(LINK http://seriot.ch/parsing_json.html)
 Source:    $(PHOBOSSRC std/json.d)
 */
 /*
@@ -77,6 +77,7 @@ enum JSONOptions
     specialFloatLiterals = 0x1, /// encode NaN and Inf float values as strings
     escapeNonAsciiChars = 0x2,  /// encode non ascii characters with an unicode escape sequence
     doNotEscapeSlashes = 0x4,   /// do not escape slashes ('/')
+    strictParsing = 0x8,        /// Strictly follow RFC-8259 grammar when parsing
 }
 
 /**
@@ -694,7 +695,8 @@ struct JSONValue
 
 /**
 Parses a serialized string and returns a tree of JSON values.
-Throws: $(LREF JSONException) if the depth exceeds the max depth.
+Throws: $(LREF JSONException) if string does not follow the JSON grammar or the depth exceeds the max depth,
+        $(LREF ConvException) if a number in the input cannot be represented by a native D type.
 Params:
     json = json-formatted string to parse
     maxDepth = maximum depth of nesting allowed, -1 disables depth checking
@@ -703,8 +705,8 @@ Params:
 JSONValue parseJSON(T)(T json, int maxDepth = -1, JSONOptions options = JSONOptions.none)
 if (isInputRange!T && !isInfinite!T && isSomeChar!(ElementEncodingType!T))
 {
-    import std.ascii : isWhite, isDigit, isHexDigit, toUpper, toLower;
-    import std.typecons : Yes;
+    import std.ascii : isDigit, isHexDigit, toUpper, toLower;
+    import std.typecons : Nullable, Yes;
     JSONValue root;
     root.type_tag = JSON_TYPE.NULL;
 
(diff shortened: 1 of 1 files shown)


Geod24:
Thanks for the contribution !
Could you add some test cases for the issues you've addressed, so it won't regress ?

tylerknott:
Sure. I'll update the unit tests to cover these cases. Also, the style check failed the "Check for package wide std.algorithm imports" test. Is that supposed to fail for cases like this where the import from std.algorithm is local within a function?

Geod24:
> Is that supposed to fail for cases like this where the import from std.algorithm is local within a function?

It looks like that, even if the import is selective.

tylerknott:
It looks like the auto-tester failure on the Darwin_64_64 platform is unrelated to my changes (the std.datetime.stopwatch and std.datetime.interval modules don't use std.json at all). Is there any way to re-try that check?

wilzbach:
Auto-tester constantly restarts itself and invalidates all builds whenever a PR at DMD, druntime or Phobos is merged. So these failures should disappear quite quickly. There's also a button "deprecate" if you login over there to manually invalidate a build, but as mentioned there's typically no need.

PR #6618, opened 2018-06-28
Clean up use of scope in std.datetime.systime.

jmdavis:
This removes unnecessary uses of scope, adds a lot of necessary uses of
scope, and adds basic tests that ensure that each of the functions in
std.datetime.systime compile when given scope arguments. It also enables
some previously commented out tests involving immutable SysTimes (IIRC,
they didn't use to compile due to compiler bugs related to
Rebindable!(immutable TimeZone), but they now compiler, so they should
be enabled).

### DIFF
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -419,7 +419,7 @@ public:
                        given $(REF DateTime,std,datetime,date) is assumed to
                        be in the given time zone.
       +/
-    this(scope const DateTime dateTime, immutable TimeZone tz = null) @safe nothrow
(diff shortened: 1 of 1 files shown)


WalterBright:
I'll approve if:
```
dmd std/datetime/systime.d -unittest -main -dip1000
systime
```
works!

wilzbach:
> dmd std/datetime/systime.d -unittest -main -dip1000

That's already checked by our CI:

https://github.com/dlang/phobos/blob/master/dip1000.mak#L105

PR #6619, opened 2018-06-28
Fix issue 19037: use move/moveEmplace in Nullable

FeepingCreature:
Makes Nullable usable with every type, aside the crazies who have pointers to their members. opAssign of the underlying type will no longer be called.

This enables Nullable on, for instance, structs with `@disable this();` and a `SysTime`.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -3269,6 +3269,80 @@ auto nullable(T)(T t)
 
     assert(foo.approxEqual(bar));
 }
+// bugzilla issue 19037
+@safe unittest
+{
+    import std.datetime : SysTime;
+
+    struct Test
+    {
+        bool b;
+
+        nothrow invariant { assert(b == true); }
+
+        SysTime _st;
+
+        static bool destroyed;
+
+        @disable this();
+        this(bool b) { this.b = b; }
+        ~this() @safe { destroyed = true; }
+
+        // mustn't call opAssign on Test.init in Nullable!Test, because the invariant
+        // will be called before opAssign on the Test.init that is in Nullable
+        // and Test.init violates its invariant.
+        void opAssign(Test rhs) @safe { assert(false); }
+    }
+
+    {
+        Nullable!Test nt;
+
+        nt = Test(true);
+
+        // destroy value
+        Test.destroyed = false;
+
+        nt.nullify;
+
+        assert(Test.destroyed);
+
+        Test.destroyed = false;
+    }
+    // don't run destructor on T.init in Nullable on scope exit!
+    assert(!Test.destroyed);
+}
+// check that the contained type's destructor is called on assignment
+@system unittest
+{
+    struct S
+    {
+        // can't be static, since we need a specific value's pointer
+        bool* destroyedRef;
+
+        ~this()
+        {
+            if (this.destroyedRef)
+            {
+                *this.destroyedRef = true;
+            }
+        }
+    }
+
+    Nullable!S ns;
+
+    bool destroyed;
+
+    ns = S(&destroyed);
+
+    // reset from rvalue destruction in Nullable's opAssign
+    destroyed = false;
+
+    // overwrite Nullable
+    ns = S(null);
+
+    // the original S should be destroyed.
+    assert(destroyed == true);
+}
 
 /**
 Just like `Nullable!T`, except that the null state is defined as a
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2569,7 +2569,12 @@ Practically `Nullable!T` stores a `T` and a `bool`.
  */
 struct Nullable(T)
 {
-    private T _value = T.init;
+    private union DontCallDestructorT
+    {
+        T payload;
+    }
+
+    private DontCallDestructorT _value = DontCallDestructorT.init;
 
     private bool _isNull = true;
 
@@ -2581,7 +2586,7 @@ Params:
  */
     this(inout T value) inout
     {
-        _value = value;
+        _value.payload = value;
         _isNull = false;
     }
 
@@ -2596,14 +2601,14 @@ Params:
             return rhs._isNull;
         if (rhs._isNull)
             return false;
-        return _value == rhs._value;
+        return _value.payload == rhs._value.payload;
     }
 
     /// Ditto
     bool opEquals(U)(auto ref const(U) rhs) const
     if (is(typeof(this.get == rhs)))
     {
-        return _isNull ? false : rhs == _value;
+        return _isNull ? false : rhs == _value.payload;
     }
 
     ///
@@ -2689,7 +2694,7 @@ Params:
         if (isNull)
             put(writer, "Nullable.null");
         else
-            formatValue(writer, _value, fmt);
+            formatValue(writer, _value.payload, fmt);
     }
 
     //@@@DEPRECATED_2.086@@@
@@ -2702,7 +2707,7 @@ Params:
         }
         else
         {
-            sink.formatValue(_value, fmt);
+            sink.formatValue(_value.payload, fmt);
         }
     }
 
@@ -2717,7 +2722,7 @@ Params:
         }
         else
         {
-            sink.formatValue(_value, fmt);
+            sink.formatValue(_value.payload, fmt);
         }
     }
 
@@ -2760,9 +2765,9 @@ Forces `this` to the null state.
     void nullify()()
     {
         static if (is(T == class) || is(T == interface))
-            _value = null;
+            _value.payload = null;
         else
-            .destroy(_value);
+            .destroy(_value.payload);
         _isNull = true;
     }
 
@@ -2787,16 +2792,18 @@ Params:
     {
         import std.algorithm.mutation : moveEmplace, move;
 
+        // the lifetime of the value in copy shall be managed by
+        // this Nullable, so we must avoid calling its destructor.
+        auto copy = DontCallDestructorT(value);
+
         if (_isNull)
         {
-            // trusted since _value is known to be T.init here.
-            // the lifetime of value shall be managed by this Nullable;
-            // hence moveEmplace resets it to T.init.
-            () @trusted { moveEmplace(value, _value); }();
+            // trusted since payload is known to be T.init here.
+            () @trusted { moveEmplace(copy.payload, _value.payload); }();
         }
         else
         {
-            move(value, _value);
+            move(copy.payload, _value.payload);
         }
         _isNull = false;
     }
@@ -2837,13 +2844,13 @@ Returns:
     {
         enum message = "Called `get' on null Nullable!" ~ T.stringof ~ ".";
         assert(!isNull, message);
-        return _value;
+        return _value.payload;
     }
 
     /// ditto
     @property get(U)(inout(U) fallback) inout @safe pure nothrow
     {
-        return isNull ? fallback : _value;
+        return isNull ? fallback : _value.payload;
     }
 
 ///
@@ -3475,7 +3482,9 @@ Params:
  */
     void opAssign()(T value)
     {
-        _value = value;
+        import std.algorithm.mutation : swap;
+
+        swap(value, _value);
     }
 
 /**
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2590,6 +2590,17 @@ Params:
         _isNull = false;
     }
 
+    static if (is(T == struct) && hasElaborateDestructor!T)
+    {
+        ~this()
+        {
+            if (!_isNull)
+            {
+                destroy(_value.payload);
+            }
+        }
+    }
+
     /**
       If they are both null, then they are equal. If one is null and the other
       is not, then they are not equal. If they are both non-null, then they are
@@ -3349,6 +3360,28 @@ auto nullable(T)(T t)
 
     // the original S should be destroyed.
     assert(destroyed == true);
+}
+// check that the contained type's destructor is still called when required
+@system unittest
+{
+    bool destructorCalled = false;
+
+    struct S
+    {
+        bool* destroyed;
+        ~this() { *this.destroyed = true; }
+    }
+
+    {
+        Nullable!S ns;
+    }
+    assert(!destructorCalled);
+    {
+        Nullable!S ns = Nullable!S(S(&destructorCalled));
+
+        destructorCalled = false; // reset after S was destroyed in the NS constructor
+    }
+    assert(destructorCalled);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ZombineDev:
On a first glance the changes look good. Can you add one more test cases for payload types for which the use of `move` / `moveEmplace` is necessary?

FeepingCreature:
Related question: should 

    Nullable!T nt = T(1);
    nt = T(2);

call opAssign on the T in the Nullable?

FeepingCreature:
Testcase added. I'll just stick with "always move" for now.

FeepingCreature:
ping

std/typecons.d
```diff
@@ -2801,9 +2784,18 @@ succeeds, `this` becomes non-null.
 Params:
     value = A value of type `T` to assign to this `Nullable`.
  */
-    void opAssign()(T value)
+    @trusted void opAssign()(T value)
```
wilzbach: How can we be sure that this is always `@safe`?
Wouldn't it be better to make move/moveEmplace due the inference (and rather fix that than blasting things with @trusted here)?

FeepingCreature: Because `move` is @safe and `moveEmplace` is only unsafe because it calls memcpy, which we know to be safe because the target variable is only initialized to `.init` and no methods have been called on it yet, or `.destroy` has been called on it; semantically it's uninitialized memory.

FeepingCreature:
The assumptions used in this code are documented in https://github.com/dlang/dlang.org/pull/2410 .

Note that DMD does not play ball yet - PR incoming.

FeepingCreature:
Uh, as far as I can tell these aren't my failures.

wilzbach:
CircleCi is unrelated -> https://github.com/dlang/phobos/pull/6626, but not sure why DAutoTest failed, but as it seems to be offline at the moment, it could be related to this.

FeepingCreature:
Note that the current state of this PR does not make very much sense without https://github.com/dlang/dlang.org/pull/2410 and https://github.com/dlang/dmd/pull/8462 , which formalize the assumption gleamed from `moveEmplace`.

FeepingCreature:
Hm, though given that T.init.~this is supposed to be valid anyways apparently, I guess this should go ahead regardless.

FeepingCreature:
Re CircleCI, I really don't know what's going on there. `std.uni` does not even use `Nullable`...

wilzbach:
Don't worry about this. CircleCi is currently broken as someone managed to change the -dip1000 semantics and break std.uni (see https://github.com/dlang/phobos/pull/6626).
We're working on a replacement for ci.dlang.io (https://github.com/dlang/ci/pull/225#issuecomment-403338519).
Sadly both issues are known.

FeepingCreature:
@n8sh There's a potential problem with this PR.

As opposed to before, Nullable now actually depends on the assumption that calling a destructor on T.init is a valid operation. This may make problems if people use structs with invariants that check for null or 0 and, say, a SysTime.

I hate to delay my own PR, but wanna wait for CircleCI to come back so we can see?

n8sh:
@FeepingCreature Merging anything but trivial changes without waiting for CircleCI would be imprudent. How do you mean this PR makes `Nullable` depend on the validity of calling a destructor on `T.init` any more than it already did?  It seems like this PR only removes a place where T's destructor would formerly have been called.

FeepingCreature:
Previously, in `opAssign` for instance since it'd be implemented as `this.value_ = value`, the value of `value` would still be valid (:smile:) at opAssign exit. With move/moveEmplace, it's reset to T.init and runs into the T.init.~this controversy.

n8sh:
@FeepingCreature you could avoid resetting `value` to `T.init` by using `std.algorithm.mutation  : swap`.

```d
void opAssign()(T value)
{
    static if (!hasElaborateAssign!T && !hasElaborateDestructor!T)
    {
        _value = value;
    }
    else
    {
        import std.algorithm.mutation : swap;
        swap(value, _value);
    }
    _isNull = false;
}
```

That way if `_value` and `value` were both valid on function entry, they'll still both be valid on function exit.

FeepingCreature:
True, but ... initially _value is not valid, so that'd still be equivalent to resetting value to T.init the first time it's assigned, and throw anyway?

n8sh:
The small benefit to `swap` is that if you construct `Nullable!T` with `this(inout T value) inout`, later calls to `opAssign` will not cause anything to be reset to `T.init` unlike with `move`, but you're right that doesn't do anything to solve the problem of a default constructed `Nullable!T`. Music chairs indeed.

The only way I know of to prevent a struct's destructor from being called is by passing it around inside a union. We could use `moveEmplace`/`move` in `opAssign` without worrying about the argument's destructor if the argument type were `union { T value; }`, but that would require callers to use different syntax. Maybe some kind of implicit conversion magic could avoid that ugliness.

FeepingCreature:
Unions! That's it!

std/typecons.d
```diff
@@ -2803,7 +2791,21 @@ Params:
...
+        auto copy = DontCallDestructorT(value);
+
+        if (_isNull)
+        {
+            // trusted since _value is known to be T.init here.
+            () @trusted { moveEmplace(copy.t, _value.t); }();
+        }
+        else
+        {
+            move(copy.t, _value.t);
```
n8sh: In this branch we know `_value` is initialized so why avoid calling the postblit? It seems to me like this might cause something we don't want, like a `RefCounted` argument having its count drop to zero and deallocating its storage after we assign it to `_value`.

n8sh: Ah wait! The postblit gets called by `auto copy = DontCallDestructorT(value)`.

FeepingCreature:
Oops: I'd forgotten to implement explicit destruction where required. In now.

std/typecons.d
```diff
@@ -3275,6 +3288,50 @@ auto nullable(T)(T t)
...
+    nt.nullify;
+    // don't run destructor on T.init in Nullable!
+}
+// check that the contained type's destructor is still called when required
+@system unittest
+{
+    bool destructorCalled = false;
+    struct S
+    {
+        bool* destroyed;
```
wilzbach: Simply make this static?

std/typecons.d
```diff
@@ -3275,6 +3288,50 @@ auto nullable(T)(T t)
...
+    import std.datetime : SysTime;
+
+    struct Test
+    {
+        bool b;
+        @disable this();
+        this(bool b) { this.b = b; }
+        nothrow invariant { assert(b == true); }
+        void opAssign(Test test2) @safe { this.b = test2.b; this._st = test2._st; }
+        ~this() @safe { }
```
wilzbach: space between variables and declarations.

std/typecons.d
```diff
@@ -3275,6 +3288,50 @@ auto nullable(T)(T t)
...
+@safe unittest
+{
+    import std.datetime : SysTime;
+
+    struct Test
+    {
+        bool b;
+        @disable this();
+        this(bool b) { this.b = b; }
+        nothrow invariant { assert(b == true); }
```
wilzbach: Ehm a bit of spacing wouldn't hurt here.

std/typecons.d
```diff
@@ -3275,6 +3288,50 @@ auto nullable(T)(T t)
...
+        void opAssign(Test test2) @safe { this.b = test2.b; this._st = test2._st; }
+        ~this() @safe { }
+        SysTime _st;
```
wilzbach: How about grouping the variables?

std/typecons.d
```diff
@@ -3275,6 +3288,50 @@ auto nullable(T)(T t)
...
+{
+    import std.datetime : SysTime;
+
+    struct Test
+    {
+        bool b;
+        @disable this();
+        this(bool b) { this.b = b; }
+        nothrow invariant { assert(b == true); }
+        void opAssign(Test test2) @safe { this.b = test2.b; this._st = test2._st; }
```
wilzbach: This is typically named `rhs`

std/typecons.d
```diff
@@ -3275,6 +3288,50 @@ auto nullable(T)(T t)
...
+        SysTime _st;
+    }
+    Nullable!Test nt;
+    // mustn't call opAssign on Test.init in nt, because the invariant will be called
+    // on the Test.init that has to be in Nullable (because Test() is disabled),
+    // and Test.init violates its invariant.
+    nt = Test(true);
+    // destroy value
+    nt.nullify;
+    // don't run destructor on T.init in Nullable!
```
wilzbach: You don't test for this

std/typecons.d
```diff
@@ -3275,6 +3288,50 @@ auto nullable(T)(T t)
...
+        this(bool b) { this.b = b; }
+        nothrow invariant { assert(b == true); }
+        void opAssign(Test test2) @safe { this.b = test2.b; this._st = test2._st; }
+        ~this() @safe { }
+        SysTime _st;
+    }
+    Nullable!Test nt;
+    // mustn't call opAssign on Test.init in nt, because the invariant will be called
+    // on the Test.init that has to be in Nullable (because Test() is disabled),
+    // and Test.init violates its invariant.
```
wilzbach: Would be better to have an actual `assert` that checks that opAssign hasn't been called (instead of relying on the invariant).

std/typecons.d
```diff
@@ -2570,25 +2570,13 @@ Practically `Nullable!T` stores a `T` and a `bool`.
...
-    {
-        private T _value = T.init;
-    }
-    else
-    {
-        static assert(false,
-                      "Cannot construct " ~ typeof(this).stringof ~
-                      ": type has no default constructor and overloaded assignment."
-        );
+        T t;
```
wilzbach: If you look at the code just `.t` is a bit weird.
Maybe `payload`?

std/typecons.d
FeepingCreature: The invariant effectively "tests" for it. I'll add a comment.

std/typecons.d
FeepingCreature: Oh okay. Will add checks.

std/typecons.d
FeepingCreature: done.

std/typecons.d
FeepingCreature: fixed

std/typecons.d
FeepingCreature: fixed

std/typecons.d
FeepingCreature: fixed

FeepingCreature:
core.exception.AssertError@/var/lib/jenkins/dlang_projects/weka-io/mecca/src/mecca/reactor/io/signals.d(246): Assert: 3 < 5 Signal count incorrect

    $ git clone --depth 1 https://github.com/weka-io/mecca.git
    ...
    $ cd mecca
    $ grep Nullable . -R
    $

FeepingCreature:
buildkite has been running for two hours now...? DAutoTest also seems to be stuck.

wilzbach:
Nay, DAutoTest is just pretty busy and buildkite doesn't run with full resources yet as its experimental (it only uses two cheap VPS of mine for all build jobs).
Anyhow, could you please stop pinging us about the CI being in process?
That's __normal__ and it __doesn't block__ your PR from being merged (for example auto-tester will constantly invalidate its results whenever sth. new got merged and rebuild your PR on all hosts). Once this has the auto-merge label, it will have superior priority for the CIs.

Feel free to discuss CI problems in #ci on Slack

FeepingCreature:
Happy One Month Anniversary of this PR!

A merge tag would be a really good present...

CyberShadow:
I think this change is partially incorrect. The reason is that this moves `Nullable` and makes assertions about the language in a direction that is incompatible with supporting non-copyable objects.

Consider the `opAssign` declaration:

```d
Nullable opAssign()(T value)
```

To support non-copyable objects, we'd just use `move` to move the value into the store payload. This is the way that many such constructs use. Instead, `Nullable`'s implementation does something ugly and copies it into a weird `union`.

More importantly, there is a unit test which verifies that during an `opAssign` call, the destructor was never called on a default-initialized (`.init`) value.

Well, this is a problem. D does not have move semantics, so `move` can't actually tell the language to skip destruction of `value`; all it does is populate it with the default value, which will still have its destructor called. 

So, the destructor *will* be called on `value`. The choice is between:
1. Letting `move` put `.init` in it, therefore the destructor will be called on the default value
2. Copying it, therefore the destructor will be called on the old copy of the value.

The unit test checks that the implementation does 2 and not 1. This is a problem if we want `Nullable` to support non-copyable types.

One conclusion we can draw from the above is that, because D lacks true move semantics, it should always be valid to call the destructor on the default value; otherwise, such objects will be incompatible with `move` and everything that uses it. So, I think the unit test is incorrect.

BTW, without move semantics, I don't see any way to support both non-copyable and init-less types :(

FeepingCreature:
It's been six years, so I don't remember too well. I think I was trying to make Nullable act like Rebindable here? This is before I realized that Rebindable really needed to be its own thing, with RebindableNullable on top.

But... Nullable specifically *can* skip destruction, via the union hack, because it knows for a fact if its value field is populated or not. If it's not doing that, it probably should.

CyberShadow:
> But... Nullable specifically _can_ skip destruction, via the union hack, because it knows for a fact if its value field is populated or not. If it's not doing that, it probably should.

OK, but you can't do the same thing for `opAssign`'s parameter. It's unavoidable in general.

FeepingCreature:
Yeah I'd forgotten about the parameter. Fwiw, I think this is not the only place that assumes that `T.init` is a valid value.

CyberShadow:
Some amendments to the above:

- `.init` and default construction are not the same thing, so this is about default constructor.
- Apparently things are even worse and `move` doesn't care if a type has `@disable this()`, it will still put `.init` there and let the language call the destructor on it.

CyberShadow:
> Fwiw, I think this is not the only place that assumes that `T.init` is a valid value.

The problem is actually the inverse - the code and tests try to avoid calling the destructor on `T.init` at great cost (making copies of the object). I think it's better to allow calling the destructor on `T.init`.

FeepingCreature:
I think from a language design perspective this is some sort of "pick two out of three" thing between "borrow checker", "field mutation" and "move semantics". Pretty sure D has just design-overloaded itself in this matter.

CyberShadow:
Well, here's the PR: https://github.com/dlang/phobos/pull/8874

CyberShadow:
I am not a language designer but move semantics (i.e. invoked explicitly using [some kind of new syntax](https://en.cppreference.com/w/cpp/language/move_assignment)) seem like a purely additive improvement.

PR #6620, opened 2018-06-29
Remove unnecessary conflicts in std.stdio.

jmdavis:
fopen and popen wrap C functions that exist in druntime, and it's not
uncommon for folks to end up accidentally trying to use the private
functions in std.stdio instead of the ones from druntime - which then
tends to result in questions on D.Learn. There's no reason that either
of these private functions needs to be named the same as the C function
that it wraps. It just causes confusion when folks accidentally try to
use them instead of the C functions.

So, this changes them to _fopen and _popen so that there will be no such
conflict, and the error messages when folks try to call the C functions
but do so incorrectly will not mention std.stdio's internals.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -427,7 +427,7 @@ Throws: `ErrnoException` if the file could not be opened.
         import std.conv : text;
         import std.exception : errnoEnforce;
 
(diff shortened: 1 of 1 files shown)


jmdavis:
Honestly, it also makes the code clearer, because then it erases any confusion over which `fopen` is being used.

PR #6621, opened 2018-06-29
no break after throwing exception

Sandman83:
Compiler yields a warning about "statement not reachable"

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -1984,7 +1984,7 @@ private
 
             bool onStandardMsg(ref Message msg)
             {
-                foreach (i, t; Ops)
+                static foreach (i, t; Ops)
                 {
                     alias Args = Parameters!(t);
                     auto op = ops[i];
@@ -2003,13 +2003,13 @@ private
                     }
                     else
                     {
-                        if(i + 1 == Ops.length)
+                        if (i + 1 == Ops.length)
                         {
                             return false;
                         }
                     }
                 }
-                assert(0);
+                assert(0, "Received an unknown message type");
             }
 
             bool onLinkDeadMsg(ref Message msg)
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -1980,7 +1980,7 @@ private
                 enum timedWait = false;
             }
 
-            static assert(Ops.length); 
+            static assert(Ops.length);
 
             bool onStandardMsg(ref Message msg)
             {
@@ -2005,11 +2005,11 @@ private
                     {
                         if(i + 1 == Ops.length)
                         {
-                            return false; 
+                            return false;
                         }
                     }
                 }
-                assert(0); 
+                assert(0);
             }
 
             bool onLinkDeadMsg(ref Message msg)
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -1980,6 +1980,8 @@ private
                 enum timedWait = false;
             }
 
+            static assert(Ops.length); 
+
             bool onStandardMsg(ref Message msg)
             {
                 foreach (i, t; Ops)
@@ -1999,8 +2001,15 @@ private
                             return true;
                         }
                     }
+                    else
+                    {
+                        if(i + 1 == Ops.length)
+                        {
+                            return false; 
+                        }
+                    }
                 }
-                return false;
+                assert(0); 
             }
 
             bool onLinkDeadMsg(ref Message msg)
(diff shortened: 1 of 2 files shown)


Geod24:
Which compiler did you try?

Sandman83:
DMD64 D Compiler v2.080.1

wilzbach:
CI fail due to trailing whitespace. Could you please add a unittest on how to reproduce the error to your PR, s.t. this is testable by us and doesn't regress (and ideally connect it with a Bugzilla issue - see the dlang-bot comment for details). Thanks!

Sandman83:
Ok. This was not obvious: I tried to write a unittest, however, I had no luck. It seems to compile well, using dmd, and to yield a warning using dub. 

Given this

```
struct S
{
	@trusted void error(string msg)
	{
		throw new Exception("");
	}

	void fun2(){}

	void fun1()
	{
		error("");
		fun2; 
	}
}
```

`
dmd source/app.d -profile -main
`

works well, however

`
dub build --build=profile
`

yields a warning "statement is not reachable" at the fun2 call in the last line and refuses to compile. 

It affects only the profile build, using together with dub. I tried all by default existing profiles, which dub mentions on its documentation site https://code.dlang.org/docs/commandline


dmd --version
DMD64 D Compiler v2.080.1

dub --version
DUB version 1.9.0, built on May  2 2018


So... I think you can just reject my PR, as it does not concern the real bug. 
But something is going wrong here. I think, it is a feature, that while building via dub it is recognized, that the fun2 call is not reachable (although, only in one of all possible build types). But how can this feature be achieved at all, if even dmd doesn't recognize this circumstance? And if dmd is meant to recognize such circumstances, then, maybe, the example above could be used to find another bug, which has still to be reported?

Geod24:
Added your test case to an existing issue: https://issues.dlang.org/show_bug.cgi?id=13165#c3

std/concurrency.d
```diff
@@ -1986,8 +1988,15 @@ private
                             return true;
                         }
                     }
+                    else
+                    {
+                        if (i + 1 == Ops.length)
+                        {
+                            return false;
+                        }
```
wilzbach: @Sandman83 why did you add this check?
`i` can never be bigger than `Ops.length - 1`...

Sandman83: If I remember correctly, compiler complained about, that the function doesn't return a value otherwise

std/concurrency.d
```diff
@@ -1967,9 +1967,11 @@ private
                 enum timedWait = false;
             }
 
+            static assert(Ops.length);
+
             bool onStandardMsg(ref Message msg)
             {
-                foreach (i, t; Ops)
+                static foreach (i, t; Ops)
```
Basile-z: is this newly `static` foreach run with more iterations than a one at some point ? 
I ask because the two declarations that follow should create an error in this case.

Sandman83: Ah... I see. I didn't have any case so far... But nevertheless, it would be wrong then.

Sandman83:
> There's probably a problem [here](https://github.com/dlang/phobos/pull/6621/files#r250245833) to solve

Should I simply close the commit, then? As you pointed out, the static foreach is out of place. But then, none of the changes in std.concurrency are needed?

Basile-z:
I just wanted to avoid any accidental merge after seeing the declarations in the non scoped static foreach.
I'm not saying that you should close.

jblachly:
What's the story with this fix?

`std.regex`:

https://github.com/dlang/phobos/blob/0e8d1a038b222917b7b8ecad3aa0ed1bda5aebbb/std/regex/internal/parser.d#L845

triggers this warning in the recent DMD2 (`DMD64 D Compiler v2.084.1`)

`/Library/D/dmd/src/phobos/std/regex/internal/parser.d(845,13): Warning: statement is not reachable`

ljmf00:
I guess this should be merge. Probably its waiting for @MartinNowak to review, otherwise, people can't use some regex features on DMD compiler.

ljmf00:
@wilzbach can you resolve this?

wilzbach:
> @wilzbach can you resolve this?

Rebased, but it looks like this the `break` has already been removed in master.

std/concurrency.d
wilzbach: That's weird and shouldn't be the case. Do you have a test case?

Sandman83: This was the case while compiling with -profile, if I remember correctly.
Can't reproduce it right now, however...

ljmf00:
@wilzbach sure, it was on May 15 #7004. Sorry. So, it should be closed?

PR #6622, opened 2018-06-29
[stable] Fix std.experimental.allocator.common.effectiveAlignment()

kinke:
It was unable to handle alignments > 2^31 bytes, e.g., returning 2 for the 64-bit pointer 0x2FA_00000000. Return a `size_t` instead of a `uint`. Note that it's only called in one place, an assertion in a ctor of the
`BitmappedBlockImpl` mixin, which handles `size_t` fine.

This fixes sporadic `std.experimental.allocator.building_blocks.bitmapped_block` unittest failures for LDC CI on Win64 (something like 1 out of 100 runs failing).

### DIFF
--- a/std/experimental/allocator/building_blocks/bitmapped_block.d
+++ b/std/experimental/allocator/building_blocks/bitmapped_block.d
@@ -99,7 +99,7 @@ private mixin template BitmappedBlockImpl(bool isShared, bool multiBlock)
         auto leadingUlongs = blocks.divideRoundUp(64);
         import std.algorithm.comparison : min;
         immutable initialAlignment = min(parentAlignment,
-            1U << trailingZeros(leadingUlongs * 8));
+            1U << min(31U, trailingZeros(leadingUlongs * 8)));
         auto maxSlack = alignment <= initialAlignment
             ? 0
             : alignment - initialAlignment;
--- a/std/experimental/allocator/common.d
+++ b/std/experimental/allocator/common.d
@@ -287,9 +287,9 @@ Returns the effective alignment of `ptr`, i.e. the largest power of two that is
 a divisor of `ptr`.
 */
 @nogc nothrow pure
-package uint effectiveAlignment(void* ptr)
+package size_t effectiveAlignment(void* ptr)
 {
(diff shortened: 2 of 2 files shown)


std/experimental/allocator/building_blocks/bitmapped_block.d
```diff
@@ -99,7 +99,7 @@ private mixin template BitmappedBlockImpl(bool isShared, bool multiBlock)
         auto leadingUlongs = blocks.divideRoundUp(64);
         import std.algorithm.comparison : min;
         immutable initialAlignment = min(parentAlignment,
-            1U << trailingZeros(leadingUlongs * 8));
+            1U << min(31U, trailingZeros(leadingUlongs * 8)));
```
kinke: This was the only other occurrence of dangerous `1U <<` in the allocator package.

wilzbach:
The CircleCi error can be safely ignored. https://github.com/dlang/phobos/pull/6623 is the cherry-pick from master.

std/experimental/allocator/common.d
```diff
@@ -287,16 +287,19 @@ Returns the effective alignment of `ptr`, i.e. the largest power of two that is
 a divisor of `ptr`.
 */
 @nogc nothrow pure
-package uint effectiveAlignment(void* ptr)
+package size_t effectiveAlignment(void* ptr)
 {
-    return 1U << trailingZeros(cast(size_t) ptr);
+    return (cast(size_t) 1) << trailingZeros(cast(size_t) ptr);
```
wilzbach: Isn't it possible for trailingZeros to return `64`?

```
assert(trailingZeros(0) == 64);
```

So maybe `trailingZeros` needs to be fixed to accept `size_t` too?

kinke: Right, good point, so for 32-bit `size_t(0)`, trailingZeros() would indeed return 64 as well. Here, it doesn't matter, as `effectiveAlignment(null)` returns `size_t(1) << 64U => 1` for both 32/64-bit size_t.

ZombineDev: Style nit: prefer `size_t(1)`

PR #6625, opened 2018-07-03

MartinNowak:


wilzbach:
Looks like dlang/dmd#8408 broke std.uni - we really need to re-enable the -dip1000 testing on auto-tester, s.t. it is run on dmd too.

PR #6443, opened 2018-04-11
Only run -dip1000 checks on CircleCi on not on the auto-tester

wilzbach:
It seems that there are still many issues with -dip1000 and for the time being it's apparentely better to only test the separate `.test` targets with -dip1000.
This is not ideal as -dip1000 is now only checked on Linux 64-bit, but before it wasn't run on Windows neither and -dip1000 is not very platform dependent.

This should hopefully unblock https://github.com/dlang/dmd/pull/8124

The only downside to this is that now a PR at dmd will no longer run Phobos with -dip1000 checks. We can either
- hope that no one will regress the CircleCi dip1000-enabled `*.test`
- let dmd's CircleCi run the `*.test` targets too
- fix the mangling issues with -dip1000 and re-enable the checking that this PR disables

PR #6626, opened 2018-07-04
[WIP] Re-enable -dip1000 on auto-tester + fix CircleCi

wilzbach:
Currently CircleCi fails (see e.g. https://github.com/dlang/phobos/pull/6625), because possible https://github.com/dlang/dmd/pull/8408 broke the -dip1000 build of std.uni and it has been noticed, because the auto-tester didn't run the tests with -dip1000 (only CircleCi at Phobos does).
This was reverted at https://github.com/dlang/phobos/pull/6443 to unblock a now closed DMD PR.

The main problem here is that I'm still not sure how we can properly fix `std.uni` again.


schveiguy:
I still don't think testing with DIP1000 makes sense unless we are doing the whole of Phobos.

ZombineDev:
> I still don't think testing with DIP1000 makes sense unless we are doing the whole of Phobos.

Doing the whole is intractable, so we need a way to ensure that our incremental progress doesn't regress.

FeepingCreature:
CircleCI should be marked "not required" for the moment. A falsely failing test is worse than no test.

wilzbach:
Okay, I will now merge this PR as CircleCi has been red for two weeks now and no one has come up with a better solution. I will open a PR for to revert the `@trusted` hack in `std.uni`, s.t. it's not forgotten and a better solution can be found.

wilzbach:
>  I will open a PR for to revert the @trusted hack in std.uni, s.t. it's not forgotten and a better solution can be found.

https://github.com/dlang/phobos/pull/6638

PR #6630, opened 2018-07-07
fix issue 18580 - std.conv.to!(const ubyte[])(void[]) should work

BBasile:
This PR allows the conversion from `void[]` to `immutable` and `const` `ubyte[]`. This will be useful for example to transform the result of `std.file.read` in `@safe` functions.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -1528,7 +1528,8 @@ Array-to-array conversion (except when target is a string type)
 converts each element in turn by using `to`.
  */
 private T toImpl(T, S)(S value)
-if (!isImplicitlyConvertible!(S, T) &&
+if (!(isArray!S && is(Unqual!(ElementType!S) == void)) &&
+    !isImplicitlyConvertible!(S, T) &&
     !isSomeString!S && isDynamicArray!S &&
     !isExactSomeString!T && isArray!T)
 {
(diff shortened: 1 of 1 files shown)


std/conv.d
```diff
@@ -1555,6 +1556,17 @@ if (!isImplicitlyConvertible!(S, T) &&
     }
 }
 
+/**
+Void array to const array of ubyte conversion. Converting to a mutable array
+is not possible because source may contain pointers.
+ */
+private T toImpl(T, S)(S value) @trusted
+if (isArray!S && is(Unqual!(ElementType!S) == void) &&
+    isArray!T && is(ElementType!T == const ubyte) || is(ElementType!T == immutable ubyte))
```
BBasile: Maybe that all basic types should work

std/conv.d
```diff
@@ -1555,6 +1556,17 @@ if (!isImplicitlyConvertible!(S, T) &&
...
+{
+   return cast(const(ubyte)[]) value.dup;
```
ZombineDev: `.to!T` should not allocate behind the user's back. I suggest you drop `.dup` and allow only conversion to `const ubyte[]`, which is safe if the input is not `shared`.

BBasile: There was not real valid to dup anyway since the result is not modifiable.

BBasile:
I'm gonna close this. It was only motivated for usage in D:YAML actually, which has to trust std.file.read.
Maybe it's better to make the file function safe.

PR #6631, opened 2018-07-10
[WIP] Add rcstring to std.experimental.collections

wilzbach:
Still needs a bit of work, but it should soon be ready for early testing.

Let's see how much of this breaks on the CIs and how the rendered docs look.


std/experimental/collections/rcstring.d
```diff
@@ -0,0 +1,833 @@
...
+
+    Params:
+         allocator = a $(REF RCIAllocator, std,experimental,allocator) or
+                     $(REF RCISharedAllocator, std,experimental,allocator)
+                     allocator object
+         s = input string
+
+    Complexity: $(BIGOH m), where `m` is the number of bytes of the input string.
+    */
+    this()(string s)
```
n8sh: Could this be `scope const char[]`? Same question for other constructors and functions below taking `string`, `wstring`, or `dstring`.

n8sh:
Are there plans for a small string optimization?

PR #5015, opened 2017-01-04
Removed auto-decoding from the floating point version of std.conv.parse

JackStouffer:
New version of https://github.com/dlang/phobos/pull/4674

13.5x faster! [Code here](https://gist.github.com/JackStouffer/0cf93d3cc861d699b334e18ea67fb66c#file-test-d).

```
# new
$ ldc2 -O5 -release test.d && ./test
Total time: 387 ms, 506 μs, and 8 hnsecs

# old
$ ldc2 -O5 -release test.d && ./test
Total time: 5 secs, 238 ms, 59 μs, and 4 hnsecs
```

PR #6633, opened 2018-07-11
Extend std.conv.parse for strings to full ieeeQuadruple precision and other fixes

joakim-noah:
This pull was tested with ldc on linux/x64, Android/AArch64, and Android/ARM, exercising 80-bit, 128-bit, and 64-bit reals. The new decimal test was off by one bit on the first and last platforms, but strangely matched exactly for the most precise, AArch64. Putting this up early to see what the CI says on other platforms, ~will~ have also extended the hex parser ~if all goes well~.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -3466,8 +3466,6 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
     {
         version (CRuntime_Microsoft)
             ld1 = 0x1.FFFFFFFFFFFFFFFEp-16382L; // strtold currently mapped to strtod
-        else version (CRuntime_Bionic)
-            ld1 = 0x1.FFFFFFFFFFFFFFFEp-16382L; // strtold currently mapped to strtod
         else
             ld1 = strtold(s.ptr, null);
     }
--- a/std/conv.d
+++ b/std/conv.d
@@ -2777,29 +2777,24 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
         enforce(!p.empty, bailOut());
         if (toLower(p.front) == 'i')
             goto case 'i';
-        enforce(!p.empty, bailOut());
         break;
     case '+':
         p.popFront();
         enforce(!p.empty, bailOut());
         break;
     case 'i': case 'I':
+        // inf
         p.popFront();
-        enforce(!p.empty, bailOut());
-        if (toLower(p.front) == 'n')
-        {
-            p.popFront();
-            enforce(!p.empty, bailOut());
-            if (toLower(p.front) == 'f')
-            {
-                // 'inf'
-                p.popFront();
-                static if (isNarrowString!Source)
-                    source = cast(Source) p;
-                return sign ? -Target.infinity : Target.infinity;
-            }
-        }
-        goto default;
+        enforce(!p.empty && toUpper(p.front) == 'N',
+               bailOut("error converting input to floating point"));
+        p.popFront();
+        enforce(!p.empty && toUpper(p.front) == 'F',
+               bailOut("error converting input to floating point"));
+        // skip past the last 'f'
+        p.popFront();
+        static if (isNarrowString!Source)
+            source = cast(Source) p;
+        return sign ? -Target.infinity : Target.infinity;
     default: {}
     }
 
@@ -2816,55 +2811,97 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
         }
 
         isHex = p.front == 'x' || p.front == 'X';
+        if (isHex) p.popFront();
     }
+    else if (toLower(p.front) == 'n')
+    {
+        // nan
+        p.popFront();
+        enforce(!p.empty && toUpper(p.front) == 'A',
+               bailOut("error converting input to floating point"));
+        p.popFront();
+        enforce(!p.empty && toUpper(p.front) == 'N',
+               bailOut("error converting input to floating point"));
+        // skip past the last 'n'
+        p.popFront();
+        static if (isNarrowString!Source)
+            source = cast(Source) p;
+        return typeof(return).nan;
+    }
+
+    /*
+     * The following algorithm consists of 2 steps:
+     * 1) parseDigits processes the textual input into msdec and possibly
+     *    lsdec/msscale variables, followed by the exponent parser which sets
+     *    exp below.
+     *    Hex: input is 0xaaaaa...p+000... where aaaa is the mantissa in hex
+     *    and 000 is the exponent in decimal format with base 2.
+     *    Decimal: input is 0.00333...p+000... where 0.0033 is the mantissa
+     *    in decimal and 000 is the exponent in decimal format with base 10.
+     * 2) Convert msdec/lsdec and exp into native real format
+     */
 
     real ldval = 0.0;
     char dot = 0;                        /* if decimal point has been seen */
     int exp = 0;
     ulong msdec = 0, lsdec = 0;
     ulong msscale = 1;
+    bool sawDigits;
 
-    if (isHex)
+    enum { hex, decimal }
+
+    // sets msdec, lsdec/msscale, and sawDigits by parsing the mantissa digits
+    void parseDigits(alias FloatFormat)()
     {
-        /*
-         * The following algorithm consists of mainly 3 steps (and maybe should
-         * be refactored into functions accordingly)
-         * 1) Parse the textual input into msdec and exp variables:
-         *    input is 0xaaaaa...p+000... where aaaa is the mantissa in hex and
-         *    000 is the exponent in decimal format.
-         * 2) Rounding, ...
-         * 3) Convert msdec and exp into native real format
-         */
+        static if (FloatFormat == hex)
+        {
+            enum uint base = 16;
+            enum ulong msscaleMax = 0x1000_0000_0000_0000UL; // largest power of 16 a ulong holds
+            enum ubyte expIter = 4; // iterate the base-2 exponent by 4 for every hex digit
+            alias checkDigit = isHexDigit;
+            /*
+             * convert letter to binary representation: First clear bit
+             * to convert lower space chars to upperspace, then -('A'-10)
+             * converts letter A to 10, letter B to 11, ...
+             */
+            alias convertDigit = (int x) => isAlpha(x) ? ((x & ~0x20) - ('A' - 10)) : x - '0';
+            sawDigits = false;
+        }
+        else static if (FloatFormat == decimal)
+        {
+            enum uint base = 10;
+            enum ulong msscaleMax = 10_000_000_000_000_000_000UL; // largest power of 10 a ulong holds
+            enum ubyte expIter = 1; // iterate the base-10 exponent once for every decimal digit
+            alias checkDigit = isDigit;
+            alias convertDigit = (int x) => x - '0';
+            // Used to enforce that any mantissa digits are present
+            sawDigits = startsWithZero;
+        }
+        else
+            static assert(false, "Unrecognized floating-point format used.");
 
-        // Used to enforce that any mantissa digits are present
-        bool anydigits = false;
-
-        p.popFront();
         while (!p.empty)
         {
             int i = p.front;
-            while (isHexDigit(i))
+            while (checkDigit(i))
             {
-                anydigits = true;
-                /*
-                 * convert letter to binary representation: First clear bit
-                 * to convert lower space chars to upperspace, then -('A'-10)
-                 * converts letter A to 10, letter B to 11, ...
-                 */
-                i = isAlpha(i) ? ((i & ~0x20) - ('A' - 10)) : i - '0';
-                if (msdec < (ulong.max-16)/16)
+                sawDigits = true;        /* must have at least 1 digit   */
+
+                i = convertDigit(i);
+
+                if (msdec < (ulong.max - base)/base)
                 {
-                    // base 16: Y = ... + y3*16^3 + y2*16^2 + y1*16^1 + y0*16^0
-                    msdec = msdec * 16 + i;
+                    // For base 16: Y = ... + y3*16^3 + y2*16^2 + y1*16^1 + y0*16^0
+                    msdec = msdec * base + i;
                 }
-                else if (msscale < 0x1000_0000_0000_0000UL)
+                else if (msscale < msscaleMax)
                 {
-                    lsdec = lsdec * 16 + i;
-                    msscale *= 16;
+                    lsdec = lsdec * base + i;
+                    msscale *= base;
                 }
                 else
                 {
-                    exp += 4;
+                    exp += expIter;
                 }
                 exp -= dot;
                 p.popFront();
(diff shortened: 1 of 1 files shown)


std/conv.d
```diff
@@ -3142,9 +3142,9 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
             while (isDigit(i))
             {
                 sawDigits = true;        /* must have at least 1 digit   */
-                if (msdec < (0x7FFFFFFFFFFFL-10)/10)
```
joakim-noah: I'm unsure why this was made to max out at 47 bits, worry that it might overflow 64-bit reals, with their 53-bit mantissas, below? Tested this on Android/ARM with `long.max` assigned to a 64-bit real, no problem there.

std/conv.d
```diff
@@ -3142,9 +3142,9 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
...
+                if (msdec < (long.max-10)/10)
                     msdec = msdec * 10 + (i - '0');
-                else if (msscale < (0xFFFFFFFF-10)/10)
+                else if (msscale < 10_000_000_000_000_000_000UL)
```
joakim-noah: Stop at the largest power of 10 that a `ulong` can hold.

std/conv.d
```diff
@@ -3340,6 +3340,10 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
     }
     r = to!real(to!string(real.max));
     assert(to!string(r) == to!string(real.max));
+
+    real pi = 3.1415926535897932384626433832795028841971693993751;
+    string full_precision = "3.1415926535897932384626433832795028841971693993751";
+    assert(feq(parse!real(full_precision), pi, 2*real.epsilon));
```
joakim-noah: This finally exercises the extended precision code with `lsdec` above, which [codecov.io says wasn't covered](https://codecov.io/gh/dlang/phobos/src/master/std/conv.d#L3147).

std/conv.d
wilzbach: I think it makes sense to add this as a comment in the source for future readers.

joakim-noah:
Added a WIP commit to extend the hex parser to ieeeQuadruple precision, worked fine on linux/x64 and Android/AArch64, running it through the CI now.

std/conv.d
```diff
@@ -3113,7 +3112,14 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
         }
         else
             static assert(false, "Floating point format of real type not supported");
++/
+        ldval = msdec;
+        if (msscale != 1)               /* if stuff was accumulated in lsdec */
+            ldval = ldval * msscale + lsdec;
```
joakim-noah: @jpf91, why do all the bit-packing above only for hex strings? Instead, I commented it out and simply assigned the mantissa to the real, as done for decimal strings below. I ignore rounding for now, which is done separately for full 80-bit reals and 64-bit reals above: I'm sure we could add it back in later once we decide how to handle this.

joakim-noah:
The Jenkins and Circle build problems are unrelated.

joakim-noah:
Merged the two commits extending the parser and simply got rid of all the bit-packing and rounding the hex parser alone was doing. Now that they're pretty similar, going to refactor them into a single parser for both hex and decimal strings.

joakim-noah:
Refactored the floating-point string parser so hex and decimal strings use the same, more compact code path, probably less performant but I'm not sure that matters. I'll add some more tests and extend others and this should be done.

joakim-noah:
Alright, this should be ready to go, once we know it passes the CI. I looked at [issue 8523, CTFE parsing of hex floating-point strings](https://issues.dlang.org/show_bug.cgi?id=8523) gets a little farther now, but still craps out because of the `ldexp` call at the end, lack of D source. I suppose that could be done manually, as done for powers of 10 just below, if someone wanted to finally fix that issue.

joakim-noah:
The Jenkins failure is because tsv-utils checks for a different error message on bad input, I'll submit a pull to that repo to change that test once this pull is approved. Circle failure still unrelated.

jondegenhardt:
> The Jenkins failure is because tsv-utils checks for a different error message on bad input, I'll submit a pull to that repo to change that test once this pull is approved.

I was waiting for this to happen. That is, for exception text to change across compiler/phobos versions. I've finessed it so far, but it'll be more difficult in this case. The way the tsv-utils tests are setup they don't know about version, but I still need to test against different compiler versions. It may be necessary to add support for multiple compilers, which will make any PR a bit more complicated. And, I don't really want to drop this test either.

Even though what `tsv-utils` is reporting is a change in the exception text, it looks like the text msg in the modified exception is a bug. From the text failure, it seems that a `const char[] val` consisting of characters "F2" is being called as `val.to!double`. This is failing, and the text added to the exception is:

    no digits seen for input "[70, 50]".

This is likely generated from [parse in conv.d](https://github.com/dlang/phobos/blob/master/std/conv.d#L2765). However, the original "F2" input is getting corrupted. The message should instead be reporting:

    no digits seen for input "F2".

For my tools, including the input text in the exception message would be a benefit, but only if it's correct. What's being tested in this case is a command line call like:

    $ tsv-filter --eq 2:1 input1.tsv

This tells `tsv-filter` to keep lines where field 2 is equal to one. However, the file has a header and values on the first line don't convert to numbers. So, the full interaction is:

    $ tsv-filter --eq 2:1 input1.tsv
    Error [tsv-filter]: Could not process line or field: no digits seen
      File: input1.tsv Line: 1
      Is this a header line? Use --header to skip.

It would certainly be an improvement if the output showed the value that was not interpreted:

    $ tsv-filter --eq 2:1 input1.tsv
    Error [tsv-filter]: Could not process line or field: no digits seen for input "F2".
      File: input1.tsv Line: 1
      Is this a header line? Use --header to skip.

But, corrupted text would be confusing.

joakim-noah:
>It would certainly be an improvement if the output showed the value that failed to parse

Yes, I changed the parser to call the `bailOut` function for that error, which also prints the input but hasn't been amended for the `representation` call at function start, as done elsewhere in #5015. I'll add the necessary cast and let's see if that's better.

> I'd like to encourage performance testing these changes as part of PR testing and due diligence

OK, I'll check.

joakim-noah:
Added a commit to fix that error message and a doc comment, both regressed in #5015. Benchmarking yields inconsistent results, I used this slightly amended version of the gist used to benchmark #5015, comparing with stock phobos and after applying this pull:
```
import std.stdio;
import std.conv;
import std.datetime;

__gshared char[] test = ['1', '2', '3', '4', '5', '.', '6', '7', '8', '9'];
__gshared char[] test1 = ['0', 'x', '3', '4', '5', '.', 'F', 'A', 'P', '9'];

void main()
{
    enum n = 10_000_000;

    double result, result1;
    StopWatch sw;
    Duration sum;
    TickDuration last = TickDuration.from!"seconds"(0);

    foreach(i; 0 .. n)
    {
        sw.start();

        result = parse!double(test);
        result1 = parse!double(test1);

        sw.stop();

        test = ['1', '2', '3', '4', '5', '.', '6', '7', '8', '9'];
        test1 = ['0', 'x', '3', '4', '5', '.', 'F', 'A', 'P', '9'];
        auto time = sw.peek() - last;
        last = sw.peek();

        sum += time.to!Duration();
    }

    writeln("Total time: ", sum);
}
```
With ldc 1.11 master and the `-O3` flag, I get a fastest time of 2.99 seconds in a linux/x64 VPS originally, fastest run of 4.59 secs after this pull. However, dmd with `-O -inline` consistently has this pull faster, best time of 3.42 secs for this pull, 4.49 secs without (not recompiling phobos there as I always do with ldc, just modifying the `std.conv` from the latest dmd 2.081.1 release, don't think that makes a difference). Android/ARM with ldc 1.11 has this pull slower, whereas Android/AArch64 with ldc 0.17 ltsmaster and this `std.conv.parse` copy-pasted in has this pull consistently faster.

If you'd like to use the above benchmark or some other one on other platforms and compare, let me know what you find.

joakim-noah:
Some more data points: I tried running that benchmark on linux/x64 with only the first two commits from this pull, ie before the refactoring commit, got a best run of 3.34 secs with dmd 2.081.1 and 2.50 secs with ldc 1.11. With ldc 0.17 running natively on Android/AArch64, I get best runs of 6.61 s originally, and a tie of 4.03 s both with and without the refactoring commit.

This implies the refactoring commit slows down dmd on linux/x64 too, but has no effect on Android/AArch64 for some reason.

jondegenhardt:
I should be able to plug this into my standard benchmarks, but it'll be a couple days before I get a chance to do so. Also, I should be able to change the tsv-utils tests, but again, it'll be a couple days before I get to it.

jondegenhardt:
I ran my `tsv-summarize` and `tsv-filter` (numeric filter) benchmarks on MacOS, these are the ones using string-to-double conversion significantly (See: [MacOS top-4](https://github.com/eBay/tsv-utils/blob/master/docs/ComparativeBenchmarks2018.md#macos-top-4-in-each-benchmark) and [Data files and Command Lines](https://github.com/eBay/tsv-utils/blob/master/docs/ComparativeBenchmarks2018.md#data-files)).

I built both current LDC and current LDC plus this PR and rebuilt both tools. Unfortunately wasn't able to use LTO and PGO, this would take a bit more work on my part. Both builds use `-release -O3 -boundscheck=off`.  I ran each benchmark three times. I also include the times from the last tsv-utils prebuilt binaries which are built using LTO and PGO. The 3 times from each build are below. Times in seconds.

Benchmark | Tool  |  tsv-utils release binaries | current LDC | LDC w/ PR 6633 
----------- | ----  | ----------------- | ------------ | -----------------
Numeric row filter | tsv-filter | 3.51, 3.65, 3.49 | 5.74, 5.73, 5.73 | 7.08, 7.05, 7.10
Summary Statistics | tsv-summarize | 9.98, 9.97, 9.93 | 30.03, 30.05, 30.03 | 44.34, 43.37, 44.18

This shows a significant performance hit due to this PR. Clearly times all times are dramatically impacted by not using LTO and PGO in the builds. However, the delta from this PR is significant enough that it's unlikely to be made up by turning LTO and PGO on.

Not that the delta due to conversion by itself is percentage-wise significantly more than the deltas shown above. These programs spend a significant amount of time doing I/O and other computations, they are not just doing string conversions.

jondegenhardt:
Here are the numbers with LTO & PGO. Quite a bit better than I expected, but still a hit to `tsv-summarize`. I ran each six times.

Benchmark | Tool  |  current LDC master | current LDC master w/ PR 6633 
----------- | ----  | -------------------- | ---------------------------
Numeric row filter | tsv-filter | 3.61, 3.59, 3.61, 3.61, 3.62, 3.60 | 3.66, 3.66, 3.66, 3.66, 3.67, 3.67
Summary Statistics | tsv-summarize | 11.93, 11.95, 11.97, 11.89, 11.91, 11.97 | 12.82, 12.79, 12.85, 12.94, 12.82, 12.82

About a 7% impact on `tsv-summarize`, and about 1% on `tsv-filter`. 
Note: `tsv-summarize` has slowed down over recent LDC releases prior to this PR. The tsv-utils prebuilt binaries are build with LDC 1.5 and run the Summary Statistics test in about 9.93.

jondegenhardt:
Sure, I'll try the first two commits and drop the third.

jondegenhardt:
Yes, the first two commits are faster. It's very significant when not using LTO/PGO. It's also significant when using LTO/PGO. For `tsv-summarize` it's faster than stock by 12%, faster than the full PR by 18%. It'd be great if you can preserve this.

Times with `-release -O3 -boundscheck=off`, but no LTO, no PGO (3 runs):

Benchmark | Tool  |  current LDC master | current LDC master w/ PR 6633 | current LDC master w/ PR 6633 first 2 commits
----------- | ----  | ----------------- | ------------ | -----------------
Numeric row filter | tsv-filter | 5.75, 5.74, 5.73 | 7.10, 7.10, 7.07 | 5.69, 5.67, 5.68
Summary Statistics | tsv-summarize | 30.07, 30.18, 30.03 | 44.14, 44.14, 44.17 | 29.51, 29.46, 29.52

Times with LTO & PGO (3 runs):

Benchmark | Tool  |  current LDC master | current LDC master w/ PR 6633 | current LDC master w/ PR 6633 first 2 commits
----------- | ----  | ----------------- | ------------ | -----------------
Numeric row filter | tsv-filter | 3.59, 3.61, 3.60 | 3.66, 3.67, 3.66 | 3.53, 3.54, 3.54
Summary Statistics | tsv-summarize | 11.96, 11.93, 11.93 | 12.92, 12.89, 12.82 | 10.42, 10.44, 10.46

joakim-noah:
OK, that coincides with my own microbenchmark results above. Honestly, the LTO/PGO results with the refactoring included are not much worse, I think the simplicity from refactoring is worth a possible 1-7% performance hit on certain platforms/compilers. However, I'll do whatever the consensus is from phobos reviewers, can simply drop the third refactoring commit if prefered.

jondegenhardt:
The 1-7% with LTO/PGO is compared to current stock. However, the delta between the 2nd commit and full PR is 18-20% with LTO/PGO.

JohanEngelen:
My opinion on the performance aspect: the refactoring is something separate from the extension of functionality. So I'd remove that 3rd commit for now, and perhaps put up a second PR where the refactoring can be discussed in isolation.  The performance hit is very large, and LTO/PGO are not always an option.
(it'd be good to add a comment in the code that it's a perf sensitive function, so that future changes check performance changes)
(btw, great that you got the perf boost in the first 2 commits!)

jondegenhardt:
I agree with @JohanEngelen regarding splitting the refactoring from the functionality change.

joakim-noah:
I investigated the performance drop from refactoring the parsers into one code path and tried to fix it by templatizing it, see last commit. I didn't bother formatting it because I just wanted to get the idea across, I don't know if it's idiomatic either. It seems to gain back the performance of the first two commits while not having the parser code replicated as before.

std/conv.d
```diff
@@ -2816,374 +2828,113 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
...
-        bool anydigits = false;
-        // Number of mantissa digits (digit: base 16) we have processed,
-        // ignoring leading 0s
-        uint ndigits = 0;
+            /*
+             * convert letter to binary representation: First clear bit
+             * to convert lower space chars to upperspace, then -('A'-10)
+             * converts letter A to 10, letter B to 11, ...
+             */
+            i = isAlpha(i) ? ((i & ~0x20) - ('A' - 10)) : i - '0';
```
joakim-noah: This `isAlpha` check is unnecessarily done for the decimal digit parser, just wasn't sure how to statically change the way the conversion is done, aliased lambdas?

std/conv.d
```diff
@@ -2816,342 +2828,82 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
...
-            {
-                e = e * 10 + p.front - '0';
-            }
-            p.popFront();
-            ndigits = 1;
-        }
-        exp += (sexp) ? -e : e;
-        enforce(ndigits, new ConvException("Error converting input"~
-                        " to floating point"));
+    enum { hex, decimal }
```
joakim-noah: Ran into an issue when I originally tried to use a named `enum` for the floating-point format, but had to change it like this because it wouldn't compile:
```
diff --git a/std/conv.d b/std/conv.d
index 2b37dfde2..931d0241c 100644
--- a/std/conv.d
+++ b/std/conv.d
@@ -2850,12 +2850,12 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
     ulong msscale = 1;
     bool sawDigits;
 
-    enum FloatFormat { Hex, Decimal }
+    enum { hex, decimal }
 
     // sets msdec, lsdec/msscale, and sawDigits by parsing the mantissa digits
-    void parseDigits(FloatFormat digitType)()
+    void parseDigits(alias FloatFormat)()
     {
-        static if (digitType == FloatFormat.Hex)
+        static if (FloatFormat == hex)
         {
             enum uint base = 16;
             enum ulong msscaleMax = 0x1000_0000_0000_0000UL; // largest power of 16 a ulong holds
@@ -2869,7 +2869,7 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
             alias convertDigit = (int x) => isAlpha(x) ? ((x & ~0x20) - ('A' - 10)) : x - '0';
             sawDigits = false;
         }
-        else static if (digitType == FloatFormat.Decimal)
+        else static if (FloatFormat == decimal)
         {
             enum uint base = 10;
             enum ulong msscaleMax = 10_000_000_000_000_000_000UL; // largest power of 10 a ulong holds
@@ -2929,9 +2929,9 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
     }
 
     if (isHex)
-        parseDigits!FloatFormat.Hex;
+        parseDigits!hex;
     else
-        parseDigits!FloatFormat.Decimal;
+        parseDigits!decimal;
 
     // Have we seen any mantissa digits so far?
     enforce(sawDigits, bailOut("no digits seen"));
```
I looked it up in bugzilla, probably some variation on [bug 13768](https://issues.dlang.org/show_bug.cgi?id=13768), and found [this workaround](https://github.com/libmir/mir-algorithm/blob/master/source/mir/ndslice/slice.d#L198) in [`mir-algorithm`](https://github.com/libmir/mir-algorithm/blob/master/source/mir/ndslice/slice.d#L388), ie if I added `alias fooDec = FloatFormat.Decimal;` and parametrized on that `parseDigits!fooDec;`, it works. Too much boilerplate at that point, so just used the anonymous enums.

jondegenhardt:
@joakim-noah - I started looking through the diffs, it looks like this PR makes a least a few small functional changes, for example, parsing `-nan`, whereas before the leading minus sign was not supported. Or perhaps I'm misreading this.

I suggest putting a list of these changes in the PR to help the reviewers. It'd also be useful for the change log.

I'll only looked through the start of the file, I'll take a more complete look later. I'm not an approver though, so it doesn't really matter when I get to it. I'm assuming I'll need to change `tsv-utils` tests for new error messages. I'm unlikely to get to this before the weekend.

joakim-noah:
No, there shouldn't be any functional changes, other than the highlighted one in the title of this pull, extending the precision to 128 bits on AArch64, and of course the new error messages.

`-NaN` was supported before, I simply added a test case to make sure it was covered. The parser merely broke out of that first switch statement after parsing the sign before and handled the `NaN` text later, whereas I combined all such non-numeric edge cases into the first switch statement now.

Strangely, the new error messages don't cause tsv-utils to fail on Jenkins anymore, not sure why.

jondegenhardt:
Thanks! It was that first switch statement I was looking at. From a review perspective, the first question I had was if this was a behavior change. I thought it might be part of "and other fixes" in the PR title.

At to the `tsv-utils` tests, Jenkins is broken and the `tsv-utils` tests are being aborted before being run, but recorded as success. If the error message has changed, they will start reporting failure at some point.

joakim-noah:
The other fixes are broken out into separate commits, like adding tests or fixing doc comments. I haven't been able to load that Jenkins results webpage in awhile- always shows a blue line that goes almost to the right edge and freezes the last 10 times I tried- didn't investigate the tsv-utils build log back when it last loaded.

jondegenhardt:
> Anyhow, as long as the Buildkite status check is green, there's nothing to worry about regarding the "Project Tester" ;-)

Okay, then we do need to check the Buildkite logs. I was expecting the tsv-utils tests to fail due to the change in text generated in the error cases. Unless the latest update does not change the text in the exception thrown. I'll look at the Buildkite logs when I get a chance, probably within a day.

wilzbach:
It could be something related to: https://github.com/dlang/ci/pull/245 (I will trigger a rebuild once that's merged)

jondegenhardt:
Yeah, looks like the tsv-utils tests aren't being run in current buildkite config. The tests finish in 0 seconds, which says they didn't run.

std/conv.d
```diff
@@ -2816,342 +2811,97 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
         }
 
         isHex = p.front == 'x' || p.front == 'X';
+        if (isHex) p.popFront();
     }
+    else if (toLower(p.front) == 'n')
+    {
+        // nan
```
joakim-noah: Modified the refactoring to move this `NaN` check out of the switch statement I'd moved it to above before, which is in keeping with the way it was done before this pull, ie optimizing for the presumably more common non-`NaN` case by not checking for it until later.

jondegenhardt:
fyi - For tsv-utils I've decided on an approach to support versioned test results. This is what I'll use to handle cases like this going forward. Shouldn't be take too long for me to get it in place, but I can't guarantee when it'll be done. Feel free to ping me if a status update is needed. At the moment this PR looks like it's also waiting on other reviews and potentially other buildkite project tests.

joakim-noah:
@n8sh or @9il, this pull is ready, should be straightforward to review, particularly if you go commit by commit.

9il:
Greate work. Would be awesome to see a new Mir library with BetterC analog of this PR

wilzbach:
As we now switched fully to Buildkite and the logs are still not public, I (as an experiment) created a dummy account with readonly permissions for those who haven't got an account at Buildkite yet:

user: dummy@dlang.io
password: dlangrocks

(one can't do much with this account, except for being a troll and changing the password. The only effect would be that dummy account can no longer be used.)

joakim-noah:
@klickverbot or @ibuclaw, review needed: you guys did a lot of the work to get quadruple-precision working. This should finish it up, along with the initial commit in dlang/druntime#2257.

ibuclaw:
Looks reasonable at a glance. Has the performance regression been sufficiently dealt with?

joakim-noah:
Yes, see [comment above](https://github.com/dlang/phobos/pull/6633#issuecomment-407692632), parsing should always be faster after this pull is applied now, which wasn't always the case with previous iterations of this PR.

jondegenhardt:
The `tsv-utils` tests now pass. The version just released now recognizes both the old and new error message text as correct results.

joakim-noah:
Thanks, much more involved diff than I thought, buildkite passing now.

OK, just need someone to approve now.

jondegenhardt:
@joakim-noah You're welcome. A change I had to make eventually. It would have been an involved task for someone unfamiliar with the build setup.

std/conv.d
```diff
@@ -3169,21 +2919,29 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
...
             else
-            {
                 break;
-            }
         }
-        enforce(sawDigits, new ConvException("no digits seen"));
+
+        // Have we seen any mantissa digits so far?
+        enforce(sawDigits, bailOut("no digits seen"));
+        static if (FloatFormat == hex)
```
joakim-noah: Rebased and made one small tweak, moved this check into `parseDigits()` so it becomes a static check, instead of being done at runtime before.

joakim-noah: Thought updating might get DAutoTest to work, but looks like it's just broken right now for every pull.

PR #6634, opened 2018-07-17
Fix Issue 19041 - errnoEnforce: example does not compile

wilzbach:
Not happy about the fix.

### DIFF
--- a/std/exception.d
+++ b/std/exception.d
@@ -622,15 +622,19 @@ private void bailOut(E : Throwable = Exception)(string file, size_t line, scope 
     $(D new ErrnoException(msg)) is thrown.  It is assumed that the last
     operation set `errno` to an error code corresponding with the failed
     condition.
-
-    Example:
-    --------------------
-    auto f = errnoEnforce(fopen("data.txt"));
-    auto line = readln(f);
-    enforce(line.length); // expect a non-empty line
-    --------------------
  +/
 alias errnoEnforce = enforce!ErrnoException;
+
+///
+@system unittest
+{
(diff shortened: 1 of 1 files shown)


std/exception.d
```diff
@@ -622,16 +622,19 @@ private void bailOut(E : Throwable = Exception)(string file, size_t line, scope
...
-    enforce(line.length); // expect a non-empty line
-    --------------------
  +/
 alias errnoEnforce = enforce!ErrnoException;
 
+///
+@system unittest
+{
+    import core.stdc.stdio : fgets, fopen;
+    auto f = errnoEnforce(fopen(__FILE__, "r"));
```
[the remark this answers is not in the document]
wilzbach: Sure - I was just copying the existing, not-working example ;-)

std/exception.d
```diff
@@ -622,16 +622,19 @@ private void bailOut(E : Throwable = Exception)(string file, size_t line, scope
...
 alias errnoEnforce = enforce!ErrnoException;
 
+///
+@system unittest
+{
+    import core.stdc.stdio : fgets, fopen;
+    auto f = fopen(__FILE__, "r").errnoEnforce;
+    char[100] buf;
+    auto line = fgets(buf.ptr, buf.length, f);
+    enforce(line[0] != 0); // expect a non-empty line
```
ZombineDev: `line != null` - see: https://en.cppreference.com/w/c/io/fgets

ZombineDev: Yeah, of course.

PR #6635, opened 2018-07-17
Issue 19055 - [404 Not Found] std.digest.digest

wilzbach:
This also requires a dlang.org update.


schveiguy:
result looks good. Not sure what the circle ci problems are. I know I've seen that message before...

wilzbach:
See: https://github.com/dlang/phobos/pull/6626

PR #6636, opened 2018-07-17
Fix Issue 19082 - Cannot inline "...Slides.numberOfFullFrames", "...Slides.gap"

wilzbach:


std/range/package.d
```diff
@@ -8587,13 +8587,11 @@ public:
             *      gap: (7 - 4) % 2 = 3 % 2 = 1
             *      end: 7 - 1 = 6
             */
-            pragma(inline, true);
```
n8sh: Why not replace it with `pragma(inline)` instead of removing it outright? I believe that `pragma(inline)` means the same as `pragma(inline, true)` except it doesn't cause a compilation error when the function can't be inlined, which seems exactly what we want here.

klickverbot: Nope, as per the spec, leaving out the argument implies default behaviour.

wilzbach:
DAutotests failures are unrelated.

PR #6638, opened 2018-07-19
Revert @trusted hotfix for std.uni

wilzbach:
Follow-up to https://github.com/dlang/phobos/pull/6626
As many modules depend on `std.uni`, we would have needed to revert almost all of the existing `-dip1000` work.

The `@trusted` shouldn't be necessary here, but so far no one has come up with a good fix for std.uni :/

### DIFF
--- a/std/uni.d
+++ b/std/uni.d
@@ -2170,20 +2170,23 @@ public struct InversionList(SP=GcPolicy)
 
     /**
         Get range that spans all of the $(CODEPOINT) intervals in this $(LREF InversionList).
+    */
+    @property auto byInterval() scope
+    {
+        // TODO: change this to data[] once the -dip1000 errors have been fixed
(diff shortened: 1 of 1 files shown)


std/uni.d
```diff
@@ -2170,20 +2170,23 @@ public struct InversionList(SP=GcPolicy)
 
     /**
         Get range that spans all of the $(CODEPOINT) intervals in this $(LREF InversionList).
+    */
+    @property auto byInterval() scope
+    {
+        // TODO: change this to data[] once the -dip1000 errors have been fixed
+        // see e.g. https://github.com/dlang/phobos/pull/6638
+        import std.array : array;
+        return Intervals!(typeof(data.array))(data.array);
```
wilzbach: Looks like this part of `-dip1000` is still under work, so I presume it's best to do the ugly allocation for now, s.t. Walter can continue to work on DIP1000 and this stays @safe.

PR #6640, opened 2018-07-20
Add -betterC testsuite for Phobos

wilzbach:
Split of https://github.com/dlang/phobos/pull/5952 that only contains the testsuite (and a first minimal test) as #5952 is currently blocked (it tries to fix `Tuple` for betterC and there are still some unresolved issues with `toHash`).

This PR should be non-controversial and thus us allow to finally get started with a betterC testsuite.

```
make -f posix.mak betterC
```

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -424,6 +424,18 @@ unittest/%.run : $(ROOT)/unittest/test_runner
 	 BUILD=debug $(MAKE) -f $(MAKEFILE) $(basename $<).debug_with_debugger
 
 ################################################################################
+# Run separate -betterC tests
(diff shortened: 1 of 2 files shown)


wilzbach:
> I think we should also have baremetal tests for things that don't depend even on libc. 

Are you thinking about sth. like https://github.com/dlang/phobos/pull/6641?

> @wilzbach also do you think we could build them on buildkite with ldc?

Yes, that should be possible (though this will might require approval from @andralex).
However, before I set up more CIs jobs, it's probably a good idea to figure out how such a testsuite would look like.

PR #6641, opened 2018-07-21
[RFC] Add a minimal baremetal testsuite to Phobos

wilzbach:
Similarly to https://github.com/dlang/phobos/pull/6640, this will allow to add tests to Phobos for functionality that doesn't depend on libc.
It currently contains a minimal Linux x86 runtime.

Future
------

We probably want to add a special UDA (e.g. `@test-baremetal` and `@test-betterc`) to Phobos and thus extract such tests automatically (instead of keeping them separate).

Another possible route is to combine this with the testsuite runner of DMD or at least have support for minimal things like `TEST_OUTPUT`.

CC @ZombineDev

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -427,13 +427,41 @@ unittest/%.run : $(ROOT)/unittest/test_runner
 # Run separate -betterC tests
 ################################################################################
 
-test/betterC/%.run: test/betterC/%.d $(DMD) $(LIB)
+test/betterC/%.run: test/betterC/%.d $(DMD) $(DRUNTIME)
 	mkdir -p $(ROOT)/unittest/betterC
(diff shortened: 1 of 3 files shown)


wilzbach:
> We probably want to add a special UDA 

https://github.com/dlang/tools/pull/357 (also supports just strings, e.g. `@("betterc")`)

andralex:
What is the use case?

wilzbach:
To test that some functions in Phobos don't depend on druntime at all which is important for people like @JinShil who would love to use parts of Phobos in their baremetal development.
Anyhow, this isn't very important to me and we managed to get `@betterC`, so I'm closing this for now ...

PR #6643, opened 2018-07-26
Add Quote to make passing multiple tuples easier

yshui:


wilzbach:
Finally our CIs are giving sensible error messages :)

```
std/meta.d(86:8)[warn]: Public declaration 'Quote' has no documented example.
[Inferior 1 (process 2276) exited with code 01]
```

std/meta.d
```diff
@@ -77,6 +78,17 @@
 
 module std.meta;
 
+/**
+ * Create a wrapper struct for a sequence of zero or more aliases. This
+ * can be used to pass multiple sequences as template arguments, because
+ * this prevent sequences from being coalesced together.
```
wilzbach: Mind the D style - https://dlang.org/dstyle.html#phobos_documentation

...
- Every public function should have a Ddoc description and documented Params: and Returns: sections (if applicable):
- Documentation comments should not use more than two stars /** or two pluses /++ in the header line.
- Either Block comments (/**) or nesting block comments (/++) should be used except when the ddoc comment is a ditto comment such as /// Ditto
- Documentation comments should not have leading stars on each line.
...

wilzbach: I know. At some point we agreed that's nice to have documentation written in the same format, but we didn't touch all the existing code (mainly to have clearer git blames), but we try to use the "consistent" format for at least new symbols.

yshui:
@wilzbach It'd better if the linter can output all the warnings at once.

wilzbach:
> @wilzbach It'd better if the linter can output all the warnings at once.

This message is related to new code you pushed - how should the linker have known about this before?
But yeah, there's currently a two-stage linting (because some simple things weren't possible with DScanner), we yes we could at least run all "tests" in stage 1 and not immediately abort.
Anyhow, on Posix you can run it locally:

```
make -f posix.mak style
```

(there are sub make targets too)

yshui:
@wilzbach ping.

yshui:
Looks like the dmd version on auto-tester is too old. Should build with dmd 2.080.1

yshui:
Whoops, there is a change in 2.081 that breaks `(A)[0]` syntax. This builds with 2.080.1

aliak00:
Why is it called "Quote"? Is this supposed to be a "`PackedAliasSeq`"?

Have you seen: https://github.com/yshui/phobos/blob/1c4a6ed23624cd439ad0f95042dd445a2e51b3c2/std/meta.d#L1762 ?

It would also be nice to see some use-cases in unittests to get a bit of a feel on how it can be used practically.

yshui:
@aliak00 whoops, don't know how I missed that...

ZombineDev:
IMO, the correct terminology is `AliasTuple`, for the same reason we renamed `TypeTuple` to `AliasSeq`:
1. It is an alias to a sequence of arbitrary compile-time entities (not only types), like symbols, values, overload sets, etc.
2. Tuples don't auto-expand while "sequence" implies a loose list-like structure.

`AliasTuple` doesn't auto-expand (i.e. the tuple part) and is an alias to a sequence of arbitrary CT entities.

yshui:
I'm fine with whatever names you come up with.

Question is should I adapt `Pack` and make it public, or should I add a new template?

aliak00:
My vote would go to adapting pack and fixing up the API. I have something similar in a library for some ideas:

https://github.com/aliak00/bolts/blob/a6d9e7830c915e11667fc1ad2e80f323cb314cdc/source/bolts/meta.d#L108

There's also this forum post:

https://forum.dlang.org/post/mnobngrzdmqbxomulpts@forum.dlang.org

LightBender:
@jmdavis this one is for you to consider in Phobos 3.

PR #3476, opened 2015-07-07
std.process: use vfork when possible (issue 14770)

dsp:
vfork(2) is a memory efficient version of fork(2). As long as we always
call execv* or _exit and don't temper with the parents process memory
we should be good to use vfork instead of fork.

It seems "correct" to use posix_spawn, but this requires changes to Druntime, etc. This change is intended as an incremental change that decreases memory copying while keep code changes minimal.

PR #6644, opened 2018-07-30
[WIP] Issue 14770 - Use vfork on Linux when available

kubo39:
## Purpose

vfork (or posix_spawn) sometimes very useful when we need subprocess (e.g, https://about.gitlab.com/2018/01/23/how-a-fix-in-go-19-sped-up-our-gitaly-service-by-30x/ Go uses clone(CLONE_VFORK, ...) internally.)
~~posix_spawn has some problems (cannnot change working directory, and so on), so I decide to use vfork.~~

posix_spawn seems good, but posix_spawn requires glibc 2.24+ and changes to druntime, so I started to use vfork.

## TODO:

- [ ] Investigte singal handling.
- [ ] Update CHANGELOG. (Since I still investigate)

## Related things

refs: https://github.com/dlang/phobos/pull/3476

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -490,8 +490,8 @@ private Pid spawnProcessImpl(scope const(char[])[] args,
             }
             return false;
         }
-        if (!(config & Config.inheritFDs) &&
-            (config & Config.detached) &&
+        if ((config & Config.inheritFDs) &&
+            !(config & Config.detached) &&
             !hasPrivilege())
             id = core.sys.posix.unistd.vfork();
         else
--- a/std/process.d
+++ b/std/process.d
@@ -457,7 +457,44 @@ private Pid spawnProcessImpl(scope const(char[])[] args,
             close(pidPipe[1]);
     }
 
-    auto id = core.sys.posix.unistd.fork();
+    pid_t id;
+    version (linux)
+    {
+        bool hasPrivilege()
+        {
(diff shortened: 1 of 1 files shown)


CyberShadow:
The [manual page](http://man7.org/linux/man-pages/man2/vfork.2.html) says that we can't call anything except `exec` or `exit`, which means we can't close file descriptors or change the working directory, so I'm not sure this is suitable for `spawnProcess` etc.

kubo39:
That means, the parent process is suspended until the vforked child successfully execute _exit(2) or exec family functions.
I think it's not problem for this.

schveiguy:
No, look back at the "Standard Description":

>    (From POSIX.1) The vfork() function has the same effect as fork(2),
       except that **the behavior is undefined** if the process created by
       vfork() either modifies any data other than a variable of type pid_t
       used to store the return value from vfork(), or returns from the
       function in which vfork() was called, **or calls any other function
       before successfully calling _exit(2) or one of the exec(3) family of
       functions.**

(emphasis by me)

kubo39:
Ah, sorry, it's undefined behavior. However It doesn't mean not suitable for `spawnProcess` though it is dangerous.

kubo39:
Today druntime has a posix_spawn() function. Altough posix_spawn() has no portable way to close file descriptors between fork-exec, it's better than vfork().

PR #6645, opened 2018-07-31
Allow @betterC tests in Phobos

wilzbach:
Open questions:
- Should we remove the `test/betterC` folder? (the @betterC test extraction is imho a better replacement)

Requires: https://github.com/dlang/tools/pull/357

### DIFF
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -73,7 +73,7 @@ setup_repos()
         git checkout -f FETCH_HEAD
     fi
 
-    for proj in dmd druntime ; do
+    for proj in dmd druntime tools ; do
         if [ $base_branch != master ] && [ $base_branch != stable ] &&
             ! git ls-remote --exit-code --heads https://github.com/dlang/$proj.git $base_branch > /dev/null; then
             # use master as fallback for other repos to test feature branches
@@ -124,12 +124,6 @@ coverage()
 publictests()
 {
     source "$(CURL_USER_AGENT=\"$CURL_USER_AGENT\" bash ~/dlang/install.sh dmd-$HOST_DMD_VER --activate)"
-
-    # checkout a specific version of https://github.com/dlang/tools
-    if [ ! -d ../tools ] ; then
-        clone https://github.com/dlang/tools.git ../tools master
-    fi
-    git -C ../tools checkout 6ad91215253b52e6ecfc39fe1854815867c66f23
 
     make -f posix.mak -j$N publictests DUB=$DUB BUILD=$BUILD
 }
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -126,6 +126,10 @@ publictests()
     source "$(CURL_USER_AGENT=\"$CURL_USER_AGENT\" bash ~/dlang/install.sh dmd-$HOST_DMD_VER --activate)"
 
     make -f posix.mak -j$N publictests DUB=$DUB BUILD=$BUILD
+
+    # run -betterC tests
+    make -f posix.mak test_extractor # build in single-threaded mode
+    make -f posix.mak -j$N betterc
 }
 
 # test stdx dub package
--- a/posix.mak
+++ b/posix.mak
@@ -286,6 +286,7 @@ SHARED=$(if $(findstring $(OS),linux freebsd),1,)
 
 TESTS_EXTRACTOR=$(ROOT)/tests_extractor
 PUBLICTESTS_DIR=$(ROOT)/publictests
+BETTERCTESTS_DIR=$(ROOT)/betterctests
 
 ################################################################################
 # Rules begin here
@@ -422,18 +423,6 @@ unittest/%.run : $(ROOT)/unittest/test_runner
 # ddd in this case is a graphical frontend to gdb
 %.debug : %.d
 	 BUILD=debug $(MAKE) -f $(MAKEFILE) $(basename $<).debug_with_debugger
-
-################################################################################
-# Run separate -betterC tests
-################################################################################
-
-test/betterC/%.run: test/betterC/%.d $(DMD) $(LIB)
-	mkdir -p $(ROOT)/unittest/betterC
-	$(DMD) $(DFLAGS) -of$(ROOT)/unittest/betterC/$(notdir $(basename $<)) -betterC $(UDFLAGS) \
-		-defaultlib= -debuglib= $(LINKDL) $<
-	./$(ROOT)/unittest/betterC/$(notdir $(basename $<))
-
-betterC: $(subst .d,.run,$(wildcard test/betterC/*.d))
 
 ################################################################################
 # More stuff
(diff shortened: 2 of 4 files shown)


.circleci/run.sh
```diff
@@ -73,7 +73,7 @@ setup_repos()
         git checkout -f FETCH_HEAD
     fi
 
-    for proj in dmd druntime ; do
+    for proj in dmd druntime tools ; do
```
wilzbach: The auto-tester and other CI behave like this already anyhow.

(Also locking tools to a specific version was done initially because it seemed as a good idea at the time, but in practice it just made updates more annoying.)

posix.mak
```diff
@@ -626,10 +616,47 @@ $(TESTS_EXTRACTOR): $(TOOLS_DIR)/tests_extractor.d | $(LIB)
...
+# Check and run @betterC tests
+# ----------------------------
+#
+# Extract @betterC tests of a module and run them in -betterC
+#
+#   make -f posix.mak std/format.betterc
+################################################################################
+
+betterc-phobos-tests: $(addsuffix .betterc,$(D_MODULES))
+betterc: betterc-phobos-tests betterc-run-tests
```
wilzbach: We could add an alias `betterC: betterc` if wanted.

ZombineDev:
I guess having a separate folder is good primary for integration test or test code that doesn't fit into any particular `std.*` module.

posix.mak
```diff
@@ -626,10 +616,47 @@ $(TESTS_EXTRACTOR): $(TOOLS_DIR)/tests_extractor.d | $(LIB)
...
+# Run separate -betterC tests
+################################################################################
+
+test/betterC/%.run: test/betterC/%.d $(DMD) $(LIB)
+	mkdir -p $(ROOT)/unittest/betterC
+	$(DMD) $(DFLAGS) -of$(ROOT)/unittest/betterC/$(notdir $(basename $<)) -betterC $(UDFLAGS) \
+		-defaultlib= -debuglib= $(LINKDL) $<
+	./$(ROOT)/unittest/betterC/$(notdir $(basename $<))
+
+betterc-run-tests: $(subst .d,.run,$(wildcard test/betterC/*.d))
```
wilzbach: There's probably not much value in keeping these separate tests anymore as `@betterC` is almost as powerful and we avoid lots of redundant declarations.

std/algorithm/comparison.d
```diff
@@ -116,7 +118,7 @@ if (isExpressionTuple!values)
 }
 
 ///
-@safe unittest
+@safe @betterC unittest
```
wilzbach: This won't be displayed on dlang.org yet, but the next step would be to figure out how to get Ddoc / Ddox to do so.
(it will be a lot easier for Ddox though.)

wilzbach:
>> Should we remove the test/betterC folder? (the @BetterC test extraction is imho a better replacement)

> I guess having a separate folder is good primary for integration tests or test code that doesn't fit into any particular std.* module.

Yep, but do we have examples where we would actually need to do this for Phobos?
AFAICT the main goal is just to ensure that most of Phobos eventually works with `-betterC` and I think the `@betterC` works well in this regard. For everything else, I guess DMD's testsuite is better suited?

wilzbach:
With dlang/tools#357 merged, this should be green now :)

std/internal/attributes.d
```diff
@@ -0,0 +1,4 @@
+module std.internal.attributes;
+
+// Internal attributes to be to e.g. annotate unittest blocks
```
WalterBright: What does this attribute mean? That it only works with betterC, or that it works with regular D or betterC?

ZombineDev:
@WalterBright, the intention is similar to other languages attributes like `pure` - `@betterC` `unittest`s need to execute successfully in normal D and with `-betterC`.

posix.mak
```diff
@@ -626,14 +630,51 @@ $(TESTS_EXTRACTOR): $(TOOLS_DIR)/tests_extractor.d | $(LIB)
...
+	@$(DMD) $(DFLAGS) $(NODEFAULTLIB) -betterC $(UDFLAGS) -run $(BETTERCTESTS_DIR)/$(subst /,_,$<)
+
+################################################################################
+# Run separate -betterC tests
+################################################################################
+
+test/betterC/%.run: test/betterC/%.d $(DMD) $(LIB)
+	mkdir -p $(ROOT)/unittest/betterC
+	$(DMD) $(DFLAGS) -of$(ROOT)/unittest/betterC/$(notdir $(basename $<)) -betterC $(UDFLAGS) \
+		-defaultlib= -debuglib= $(LINKDL) $<
```
ZombineDev: `-defaultlib= -debuglib=` -> `$(NODEFAULTLIB)`

posix.mak
ZombineDev: Yep, I'm ok with removing this target.

posix.mak
```diff
@@ -601,13 +602,26 @@ style_lint: dscanner $(LIB)
...
 	$(DMD) $(DFLAGS) $(NODEFAULTLIB) $(LIB) -w -D -Df/dev/null -main -c -o- $$(find etc std -type f -name '*.d') 2>&1
 
+################################################################################
+# Build dub from source if no dub is available
+################################################################################
+
+DUB_PATH=$(abspath $(shell which $(DUB)))
+ifeq (,$(DUB_PATH))
+	DUB_PATH=$(DUB_DIR)/bin/dub
+	DUB=$(DUB_DIR)/bin/dub
```
ZombineDev: `DUB=$(DUB_PATH)` ?

posix.mak
```diff
@@ -601,13 +602,26 @@ style_lint: dscanner $(LIB)
...
+endif
+
+$(DUB): | $(DUB_DIR) $(LIB)
+	cd $(DUB_DIR) && DFLAGS="$(DFLAGS) $(LIB) $(NODEFAULTLIB) $(LINKDL)" DMD=$${PWD}/$(DMD) ./build.sh
```
ZombineDev: Why is `$(NODEFAULTLIB)` necessary here?

ZombineDev: Ah, I guess because you want to use the version of druntime and phobos that were just built.

wilzbach: Well, I have to build `dub` from source, because auto-tester doesn't come with a `dub` version (kinda silly) and also no globally installed dmd (`HOST_DMD` is only exposed when building dmd and my PR to fix that has remained unanswered: https://github.com/braddr/at-client/pull/9).

wilzbach:
This is sadly blocked until I figure out a reliable way to use or build `dub` on the auto-tester.
At the moment, cloning `dub` and building it seems to work, but due to make's parallel execution this seems to happen in many parallel steps and some of them fail :/

wilzbach:
> This is sadly blocked until I figure out a reliable way to use or build dub on the auto-tester.

Ah well, I want to start adding `@betterC` test and stop regressions, so I removed the last experimental commit that adds `dub` to the auto-tester (and will move this into a new PR) and just enabled the betterC on CircleCi.
This is a bit problematic, because the CircleCi tests aren't run on dmd nor druntime, but thanks to our new Buildkite setup I will later today setup Buildkite, s.t. it also runs all testsuites and we can prevent such problems.

wilzbach:
> This is a bit problematic, because the CircleCi tests aren't run on dmd nor druntime, but thanks to our new Buildkite setup I will later today setup Buildkite, s.t. it also runs all testsuites and we can prevent such problems.

With https://github.com/dlang/phobos/pull/6669 and https://github.com/dlang/ci/pull/287 merged, Buildkite will now run the `buildkite-test` target on every PR.
This means that the `@betterC` tests added here will also be run on every PR at dmd, druntime and of course Phobos and thus I think we can finally move on with this (:

PR #6646, opened 2018-08-01
Move common parts of win{32,64} into win.mak

wilzbach:
Let's see whether the auto-tester Windows CIs are happy now.


wilzbach:
Urgh auto-tester's still patching the win{32/64}.mak files :/

https://github.com/braddr/at-client/blob/master/src/diff-phobos-win64.diff

Sorry I thought at some point it stopped doing that...

PR #6647, opened 2018-08-01
Move common parts of win{32,64} into win.mak [minimal]

wilzbach:
Minimal variant of https://github.com/dlang/phobos/pull/6646 to avoid running into merge conflicts.


wilzbach:
Argh. Looks like this isn't going to work for now :/

PR #6649, opened 2018-08-02
Fix Issue 19135: Rename std.json.JSON_TYPE to JsonType with lowercase attributes

FeepingCreature:
JSON_TYPE is an egregious violation of the [D Style Guide](https://dlang.org/dstyle.html).
Rewrite std.json.JSON_TYPE to be CamelCase with lowercase members, with underscore attached to avoid keyword collision.
Deprecate the old type via alias.

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -39,7 +39,7 @@ import std.traits;
     long x;
     if (const(JSONValue)* code = "code" in j)
     {
-        if (code.type() == JSON_TYPE.INTEGER)
+        if (code.type() == JSONType.integer)
             x = code.integer;
         else
             x = to!int(code.str);
@@ -83,19 +83,32 @@ enum JSONOptions
 /**
 JSON type enumeration
 */
-enum JSON_TYPE : byte
+enum JSONType : byte
 {
     /// Indicates the type of a `JSONValue`.
-    NULL,
-    STRING,  /// ditto
-    INTEGER, /// ditto
-    UINTEGER,/// ditto
-    FLOAT,   /// ditto
-    OBJECT,  /// ditto
-    ARRAY,   /// ditto
-    TRUE,    /// ditto
-    FALSE    /// ditto
+    null_,
+    string,   /// ditto
+    integer,  /// ditto
+    uinteger, /// ditto
+    float_,   /// ditto
+    array,    /// ditto
+    object,   /// ditto
+    true_,    /// ditto
+    false_,   /// ditto
+    /// These symbols will be deprecated after 2.082.
+    NULL = null_,
+    STRING = string,
+    INTEGER = integer,
+    UINTEGER = uinteger,
+    FLOAT = float_,
+    ARRAY = array,
+    OBJECT = object,
+    TRUE = true_,
+    FALSE = false_,
 }
+
+/// This alias will be deprecated after 2.082.
+alias JSON_TYPE = JSONType;
 
 /**
 JSON value node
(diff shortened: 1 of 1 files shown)


FeepingCreature:
Thank you, fixed.

The test failures are weird. Deprecated enum attributes is explicitly merged and supported on master, but they behave as if they're using an outdated DMD.

wilzbach:
Well, as you correctly realized there's no support for deprecations on enums in libdparse which is used by dscanner.

BBasile:
Will be unblocked by #6650.

FeepingCreature:
Thanks @BBasile ! I'm not sure about the test failures - as far as I can see, druntime runs into a benchmark failure due to the deprecation warnings from this change, but I can't exactly change druntime *first*.

BBasile:
Isn't this a `-de` VS `-dw` issue ? If so the command line can be changed first, then this phobos PR will pass then finally update again the druntime to use the new identifier and switch back to `-de`. Sounds like a plan, doesn't it ?

FeepingCreature:
@BBasile possibly better approach: I've disabled the deprecation warnings for the moment, with a nice fat TODO. Order of operations: merge this, fix druntime, enable deprecations.

wilzbach:
No, we don't move away from `-de`. Is there for a very good reason as otherwise people forgot to update the other core repositories and then other people complain that we use outdated stuff in our core repositories.

I highly recommend you to read [DIP1013](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1013.md) as your PR doesn't follow it at all and it also explains why `-de` is used:

> If the deprecation is occurring because the symbol(s) are being replaced by new symbols, both the old and the new symbol(s) MUST be available without a deprecation message in at least one release to allow users to build their code without issue on both the stable and master branches.

FeepingCreature:
Alright, gonna stick with the current version then. Can I leave the TODO in?

std/json.d
```diff
@@ -83,20 +83,31 @@ enum JSONOptions
...
+    false_,   /// ditto
+    NULL = null_,
+    STRING = string,
+    INTEGER = integer,
+    UINTEGER = uinteger,
+    FLOAT = float_,
+    ARRAY = array,
+    OBJECT = object,
+    TRUE = true_,
+    FALSE = false_,
```
wilzbach: I would probably add comments that this starts the deprecation cycle in 2.082 (same for JSON_TYPE  below)

FeepingCreature: added

wilzbach:
@FeepingCreature as mentioned you need to send Andrei a mail as he gets too many GitHub pings these days.

FeepingCreature:
I'm not entirely sure if the right answer to "Andrei gets too many notifications on Github" is "get his attention on this Github issue via a different medium on which he is not _yet_ spammed to death"...

FeepingCreature:
I couldn't find his email, so I just tweeted at him. ~optimism~

FeepingCreature:
I mean, should we not be using it? If std.json is not supposed to be used, it should not be in std.

andralex:
@FeepingCreature we should remove it only when there's a replacement.

wilzbach:
FWIW [std.data.json](https://github.com/s-ludwig/std_data_json) has been out there for quite a while (and is already a lot better than std.json), but Sönke has more pressing things to work on than finalizing std.data.json.

JinShil:
Actually, everything under "Data Formats" at https://dlang.org/phobos/index.html should just be moved to DUB.

FeepingCreature:
Yeah part of the point of having a standard library package is that it becomes the [Schelling point](https://en.wikipedia.org/wiki/Focal_point_(game_theory\)) for maintenance.

JohanEngelen:
Wouldn't it have been better to also add an underscore to `string`? The nameclash is annoying and results in confusing errors. I just had to fix code like this (heavily simplified):
```d
import std.conv;
import std.json;

void foo(ref string obj, JSONValue jv) {
    with (JSON_TYPE) {
        obj = jv.integer.to!string();
    }
}
```

jacob-carlborg:
> Why `object`

Because that's the name of the "main" module that is always imported. Might not be necessary, that's why I said "possibly".

PR #6650, opened 2018-08-02
Update D-Scanner to support the new enum member attributes

BBasile:
To unblock #6649, which uses the new upstream feature.


BBasile:
After more updates due to additional problem revealed in D-Scanner this PR finally passes the tests, ping @wilzbach for getting quickly merged.

PR #4114, opened 2016-03-23
Avoid brute force fd close() on Linux

mrkline:
The current Posix implementation of spawnProcess calls close on every fd from 3 to the max fd count (as gathered by `rlimit()`). Given that the process doesn't have that many descriptors open
in the *vast* majority of cases, this is quite pessimal.

Instead, check `/proc/self/fd` to see which descriptors are actually open, and close those.
A similar solution is probably possible for BSD and friends.

PR #6652, opened 2018-08-03
Do not call async-signal-safe library function between fork-exec

kubo39:
malloc() is not async-singal-safe function, so stop call between fork-exec.

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -339,7 +339,7 @@ version(Posix) private enum InternalError : ubyte
     noerror,
     exec,
     chdir,
-    getrlimit,
+    sysconf,
     doubleFork,
 }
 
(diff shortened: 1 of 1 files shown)


kubo39:
hmm,, I'm not sure why removed code needed.., Anyway only glibc malloc can be used here because it releases the locks at pthread_atfork.
I think alloca has same problem. At least, FreeBSD has closefrom(2) system call, this could be useful.

CyberShadow:
> hmm,, I'm not sure why removed code needed..,

A good way to discover the answer to such questions is to use "git blame". It will point you to the relevant commit. If the commit message is not enough, you can find the corresponding pull request, and Bugzilla issue number, which should have an explanation and some discussion.

kubo39:
@CyberShadow I checked https://github.com/dlang/phobos/pull/4114, (is this right?)

If perfromance is required as a requrement, it's ok for me. However I think it's better to supplement with comments.

CyberShadow:
> @CyberShadow I checked #4114, (is this right?)

No; actually, looking at it, this code was there from the very beginning.

The problem is that we need to make sure child processes do not inherit any file descriptors they shouldn't. In theory, all such file descriptors should have been opened with the `CLOEXEC` flag, but in practice I think that's not realistic, as all code (incl. in libraries) must remember to use that flag.

kubo39:
So, I just remove the code using malloc(), still closes all fds except forkPipeOut.

CyberShadow:
Oh, right, the `poll` code removed here was introduced in #4114.

> Anyway only glibc malloc can be used here because it releases the locks at pthread_atfork.

`malloc` here is the libc `malloc`, so, is there still a problem?

kubo39:
The problem is that it depends on implementation. I'm not sure whether jemalloc/tcmalloc/bionic-libc works like glibc does.

kubo39:
circle ci failed, but I suppose it is spurious.

wilzbach:
Yes it's due to the recent dub registry outage :/

kubo39:
rebased.

n8sh:
>glibc malloc can be used here because it releases the locks at pthread_atfork.

Can we use `version (CRuntime_Glibc)`?

kubo39:
I'm not sure that it is safe when LD_PRELOAD is used. If we need a faster implementation, it would be better  to port cpython's `_close_open_fds_safe` function.

kubo39:
I checked cpython's subprocess `_close_open_fds_maybe_unsafe` and commented,

> f these, the one most likely to ever cause a problem is opendir() as it performs an internal
> malloc().  Practically this should not be a problem.  The Java VM makes the
> same calls between fork and exec in its own UNIXProcess_md.c implementation.

I'm not sure still this should not be a problem, but jemalloc seems work well.
So I close this.

PR #6655, opened 2018-08-06
[stable] Remove superfluous core.stdc.* dependencies in std.{conv,random}

kinke:
For WebAssembly etc.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -2737,7 +2737,6 @@ Target parse(Target, Source)(ref Source source)
 if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum) &&
     isFloatingPoint!Target && !is(Target == enum))
 {
-    import core.stdc.math : HUGE_VAL;
     import std.ascii : isDigit, isAlpha, toLower, toUpper, isHexDigit;
     import std.exception : enforce;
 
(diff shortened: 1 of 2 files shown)


std/conv.d
```diff
@@ -3240,7 +3239,7 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
         }
     }
   L6: // if overflow occurred
-    enforce(ldval != HUGE_VAL, new ConvException("Range error"));
+    enforce(ldval != real.infinity, new ConvException("Range error"));
```
kinke: `HUGE_VAL` is `double.infinity`, `ldval` is a `real`.

std/random.d
```diff
@@ -2182,8 +2182,7 @@ do
         immutable T u = (rng.front - rng.min) * factor;
         rng.popFront();
 
-        import core.stdc.limits : CHAR_BIT;  // CHAR_BIT is always 8
```
atilaneves: What if on some platform, someday, it isn't 8?

kinke: Then `T.sizeof` wouldn't be in 8-bit bytes anymore, and all hell breaks loose I guess.

wilzbach:
(The DAutoTest failure is unrelated and can be ignored.)

PR #6659, opened 2018-08-13
Update conv.d

shove70:
https://issues.dlang.org/show_bug.cgi?id=17712#c15


BBasile:
There are other ways to fix one of the case (https://issues.dlang.org/show_bug.cgi?id=17712#c11) which i would prefer. It's delicate because while there's another test case, this is still **only** a workaround for one of the case.

n8sh:
>There are other ways to fix it (https://issues.dlang.org/show_bug.cgi?id=17712#c11) which i would prefer. It's delicate because while there's another test case, this is still only a workaround for one of the case of a compiler bug.

@BBasile I'm sympathetic to this but Issue #17712 has been open for a year without movement. I'm going to mark this `72h no objection -> merge`. If another PR is made during that time it can go ahead. Otherwise I think the right thing to do is merge this, and if another solution is made in the future it can replace it.

PR #6660, opened 2018-08-13
Fix Issue 19164 - malloc may be considered pure when failure results in program exit (no need to reset errno)

n8sh:
The `core.memory.pureMalloc` wrapper reads errno, calls malloc, then resets errno. This is unnecessary if when the allocation fails the program terminates with `assert(false)` or `onOutOfMemoryError()`.

### DIFF
--- a/std/container/array.d
+++ b/std/container/array.d
@@ -251,10 +251,8 @@ private struct RangeT(A)
 struct Array(T)
 if (!is(Unqual!T == bool))
 {
-    import core.memory : pureMalloc, pureRealloc, pureFree;
-    private alias malloc = pureMalloc;
-    private alias realloc = pureRealloc;
-    private alias free = pureFree;
+    import core.memory : free = pureFree;
+    import std.internal.memory : enforceMalloc, enforceRealloc;
     import core.stdc.string : memcpy, memmove, memset;
 
     import core.memory : GC;
@@ -321,8 +319,7 @@ if (!is(Unqual!T == bool))
 
                 static if (hasIndirections!T)
                 {
-                    auto newPayloadPtr = cast(T*) malloc(nbytes);
-                    newPayloadPtr || assert(false, "std.container.Array.length failed to allocate memory.");
+                    auto newPayloadPtr = cast(T*) enforceMalloc(nbytes);
                     auto newPayload = newPayloadPtr[0 .. newLength];
                     memcpy(newPayload.ptr, _payload.ptr, startEmplace * T.sizeof);
                     memset(newPayload.ptr + startEmplace, 0,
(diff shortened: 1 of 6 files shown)


std/container/array.d
```diff
@@ -334,8 +342,9 @@ if (!is(Unqual!T == bool))
                 }
                 else
                 {
-                    _payload = (cast(T*) realloc(_payload.ptr,
-                            nbytes))[0 .. newLength];
+                    auto newPayloadPtr = cast(T*) realloc(_payload.ptr, nbytes);
+                    newPayloadPtr || assert(false, "std.container.Array.length failed to allocate memory.");
+                    _payload = newPayloadPtr[0 .. newLength];
```
n8sh: The other branch using `malloc` already had `newPayloadPtr || assert(false, ...)` but this branch using `realloc` didn't.

wilzbach: Maybe we should trigger an `onOutOfMemoryError()` like its done in `trustedRealloc`?

std/container/array.d
```diff
@@ -231,6 +231,12 @@ private struct RangeT(A)
     }
 }
 
+extern (C) private pure @system @nogc nothrow
+{
+    pragma(mangle, "malloc") void* fakePureMalloc(size_t);
+    pragma(mangle, "realloc") void* fakePureRealloc(void* ptr, size_t size);
+}
```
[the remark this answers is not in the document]
n8sh: I'm hesitant to make a new module. Is there an existing module in `std.internal` that strikes you as appropriate?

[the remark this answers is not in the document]
wilzbach: > Is there an existing module in std.internal that strikes you as appropriate?

Not really, that's why I didn't suggest a specific one.

> I'm hesitant to make a new module.

It should be okay, because the user will never see it and we can always freely reorganize things. The only annoying part is to add it to `posix.mak` and `win{32,64}.mak`

std/container/array.d
n8sh: I should reverse this and go back to `assert` per https://github.com/dlang/druntime/pull/2276#discussion_r210530402

n8sh:
https://github.com/dlang/druntime/pull/2276 seems to have stalled so I'll implement @wilzbach's suggestion to create a module `std.internal.memory` to hold the fail-fast allocation functions.

thewilsonator:
Thanks.

std/container/array.d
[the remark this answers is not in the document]
n8sh: Added `std.internal.memory`.

std/internal/memory.d
```diff
@@ -0,0 +1,71 @@
...
+{
+    auto result = fakePureRealloc(ptr, size);
+    if (!result)
+    {
+        mixin (onOutOfMemoryError);
+    }
+    return result;
+}
+
+// Purified for local use only.
```
thewilsonator: `private` Otherwise LGTM.

std/container/array.d
n8sh: This now uses either `assert(0)` or `onOutOfMemoryError` depending on whether version flag `D_Exceptions` is set.

PR #6661, opened 2018-08-13
Remove duplicate definitions of DOUBLEPAIR_MSB & DOUBLEPAIR_LSB for IBM doubledouble

n8sh:


n8sh:
The git diff cuts off too soon but here you can see `DOUBLEPAIR_MSB` and `DOUBLEPAIR_LSB` are defined twice, before and after the version blocks:
https://github.com/dlang/phobos/blob/a86b2d05cae3cac89e73fd29b8925a1a822b6edf/std/math.d#L378-L399

PR #6664, opened 2018-08-15
Fix issue 19162 - [REG: 2.079.0] Public Import Overlapping Names Conflict Resolution

look-at-me:
This shouldn't have been committed to begin with until after the actual issue had been fixed. A breaking change which allows abs() to be instigated for all types irregardless if they have comparison and negative operators.

@wilzbach Per your commit.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -711,6 +711,22 @@ static foreach (Num; AliasSeq!(cfloat, cdouble, creal, ifloat, idouble, ireal))
     }
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=19162
+@safe unittest
+{
+    struct Vector(T, int size)
+    {
+        T x, y, z;
+    }
+
+    static auto abs(T, int size)(auto ref const Vector!(T, size) v)
+    {
+        return v;
+    }
+    Vector!(float, 3) v;
+    assert(abs(v) == v);
+}
+
 /*
  * Complex conjugate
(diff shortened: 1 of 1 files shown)


wilzbach:
Argh it looks like https://issues.dlang.org/show_bug.cgi?id=18250 and https://issues.dlang.org/show_bug.cgi?id=18251 are still in effect:

```
std/math.d(629): Deprecation: use of imaginary type `ireal` is deprecated, use `real` instead
std/math.d(629): Deprecation: use of imaginary type `ireal` is deprecated, use `real` instead
std/math.d(629): Deprecation: use of imaginary type `ireal` is deprecated, use `real` instead
std/math.d(629): Deprecation: use of imaginary type `ireal` is deprecated, use `real` instead
```

wilzbach:
I pushed another commit with the `TestComplex` workaround (the same was used for e.g. `std.format`) - https://github.com/dlang/phobos/commit/a972e266ede75ce98f58b23735cbec51150b5c45#diff-e28dc7a93ec122fed96f51899af668a0R2608

wilzbach:
Rebased this onto stable, s.t. we still can get it in for 2.082.
@ others: what do you think about the `version(TestComplex)` workaround?
It does fix this issue and we have done it for `std.format` too as it looks like it's the only solution for the complex type deprecation we have at the moment :/

PR #6665, opened 2018-08-15
WIP: Add non-throwing variants of parse() and to()

nordlow:
Putting it out in the open just to get early feedback.

See also: https://forum.dlang.org/post/hotgouygobclmesixayw@forum.dlang.org

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -2707,7 +2707,7 @@ if (isSomeString!Source && !is(Source == enum) &&
 }
 
 ///
-@safe unittest
+@safe pure unittest
 {
     enum EnumType : bool { a = true, b = false, c = a }
 
--- a/std/conv.d
+++ b/std/conv.d
@@ -2698,7 +2698,7 @@ if (isSomeString!Source && !is(Source == enum) &&
     auto result = tryParse!(Target, Source)(s);
     if (!result.isNull)
     {
-        return result.get(Target.init); // TODO remove `Target.init` parameter?
+        return result.get;    // safe to call when !result.isNull
     }
     else
     {
--- a/std/conv.d
+++ b/std/conv.d
@@ -2636,21 +2636,46 @@ do
     assert(parse!uint(str) == 0);
 }
 
-/**
- * Takes a string representing an `enum` type and returns that type.
- *
- * Params:
- *     Target = the `enum` type to convert to
- *     s = the lvalue of the range to _parse
- *
- * Returns:
- *     An `enum` of type `Target`
- *
- * Throws:
- *     A $(LREF ConvException) if type `Target` does not have a member
- *     represented by `s`.
- */
-Target parse(Target, Source)(ref Source s)
+struct Expected(Result, Error)
+{
+    this(Result result) @trusted
+    {
+        this.result = result;
+        this._hasResult = true;
+    }
+    static typeof(this) asError(Error error) @trusted
+    {
+        typeof(this) that = void;
+        that.error = error;
+        that._hasResult = false;
+        return that;
+    }
+
+    ~this() @trusted
+    {
+        if (_hasResult)
+        {
+            destroy(error);
+        }
+        else
+        {
+            destroy(result);
+        }
+    }
+
+    @property hasResult() const { return _hasResult; }
+
+    union
+    {
+        Result result;
+        Error error;
+    }
+    bool _hasResult;            // TODO don't waste bool here
+}
+
+import std.typecons : Nullable;
+
+Nullable!(Target) tryParse(Target, Source)(ref Source s)
 if (isSomeString!Source && !is(Source == enum) &&
     is(Target == enum))
 {
@@ -2671,12 +2696,51 @@ if (isSomeString!Source && !is(Source == enum) &&
     if (longest_match > 0)
     {
         s = s[longest_match .. $];
-        return result ;
+        return typeof(return)(result);
     }
 
-    throw new ConvException(
-        Target.stringof ~ " does not have a member named '"
-        ~ to!string(s) ~ "'");
+    return typeof(return).init;
+}
+
+///
+@safe pure nothrow @nogc unittest
+{
+    enum EnumType : bool { a = true, b = false, c = a }
+    auto str = "a";
+    auto expected = tryParse!EnumType(str);
+    // TODO how can this compile in a nothrow function when `get` may throw?:
+    assert(!expected.isNull &&
+           expected.get == EnumType.a);
+}
+
+/**
+ * Takes a string representing an `enum` type and returns that type.
+ *
+ * Params:
+ *     Target = the `enum` type to convert to
+ *     s = the lvalue of the range to _parse
+ *
+ * Returns:
+ *     An `enum` of type `Target`
+ *
+ * Throws:
+ *     A $(LREF ConvException) if type `Target` does not have a member
+ *     represented by `s`.
+ */
+Target parse(Target, Source)(ref Source s)
+if (isSomeString!Source && !is(Source == enum) &&
+    is(Target == enum))
+{
+    auto result = tryParse!(Target, Source)(s);
+    if (!result.isNull)
+    {
+        return result.get(Target.init); // TODO remove `Target.init` parameter?
+    }
+    else
+    {
+        throw new ConvException(Target.stringof ~ " does not have a member named '"
+                                ~ to!string(s) ~ "'");
+    }
 }
 
 ///
--- a/std/conv.d
+++ b/std/conv.d
@@ -2636,43 +2636,6 @@ do
     assert(parse!uint(str) == 0);
 }
 
-struct Expected(Result, Error)
-{
-    this(Result result) @trusted
-    {
-        this.result = result;
-        this._hasResult = true;
-    }
-    static typeof(this) asError(Error error) @trusted
-    {
-        typeof(this) that = void;
-        that.error = error;
-        that._hasResult = false;
-        return that;
-    }
-
-    ~this() @trusted
-    {
-        if (_hasResult)
-        {
-            destroy(error);
-        }
-        else
-        {
-            destroy(result);
-        }
-    }
-
-    @property hasResult() const { return _hasResult; }
-
-    union
-    {
-        Result result;
-        Error error;
-    }
-    bool _hasResult;            // TODO don't waste bool here
-}
-
 import std.typecons : Nullable;
 
 Nullable!(Target) tryParse(Target, Source)(ref Source s)
--- a/std/conv.d
+++ b/std/conv.d
@@ -2669,11 +2669,14 @@ if (isSomeString!Source && !is(Source == enum) &&
 {
     enum EnumType : bool { a = true, b = false, c = a }
 
-    auto str = "a";
-    auto expected = tryParse!EnumType(str);
+    auto a_str = "a";
+    auto a_result = tryParse!EnumType(a_str);
+    assert(!a_result.isNull &&
+           a_result.get == EnumType.a);
 
-    assert(!expected.isNull &&
-           expected.get == EnumType.a);
+    auto z_str = "z";
+    auto z_result = tryParse!EnumType(z_str);
+    assert(z_result.isNull);
 }
 
 /**
--- a/std/conv.d
+++ b/std/conv.d
@@ -1,7 +1,7 @@
 // Written in the D programming language.
 
 /**
-A one-stop shop for converting values from one type to another.
+   A one-stop shop for converting values from one type to another.
 
 $(SCRIPT inhibitQuickIndex = 1;)
 $(BOOKTABLE,
@@ -52,6 +52,7 @@ public import std.ascii : LetterCase;
 import std.meta;
 import std.range.primitives;
 import std.traits;
+import std.typecons : Nullable;
 
 // Same as std.string.format, but "self-importing".
 // Helps reduce code and imports, particularly in static asserts.
@@ -2635,8 +2636,6 @@ do
     auto str = "0=\x00\x02\x55\x40&\xff\xf0\n\x00\x04\x55\x40\xff\xf0~4+10\n";
     assert(parse!uint(str) == 0);
 }
-
-import std.typecons : Nullable;
 
 Nullable!(Target) tryParse(Target, Source)(ref Source s)
 if (isSomeString!Source && !is(Source == enum) &&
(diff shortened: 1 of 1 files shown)


wilzbach:
I like the idea (and I'm generally in favor of using more `Nullable` / `Optional` in Phobos).
However a few concerns for now:
- should we really promote a different type and not use [`Nullable`](https://dlang.org/phobos/std_typecons.html#Nullable). I'm aware of the problem that Nullable doesn't allow one to store a custom error message. I guess at the very least the API should be similar (maybe we can generalize the Nullable concept like we did with ranges). Also there are lots of nice goodies in `Nullable` like `apply` (the analog of Folly's  `then`)
- `expected.result` (`.get` in Nullable) should automatically `throw` when accessed and I think we might be better off with sth. like `get!ConvException`. Only `Nullable.get(<fallback>)` is `nothrow`

nordlow:
Sounds good to me.

Is `Nullable.apply` nothrow?

wilzbach:
>  Is Nullable.apply nothrow?
> Can we feed Nullable.apply a lambda and still be nothrow?


Yes and even `@nogc` e.g.

https://github.com/dlang/phobos/blob/bb769cfaf51f32a9f20567f885361fe828e9f8a3/std/typecons.d#L3807-L3823

See also: https://dlang.org/changelog/2.080.0.html#std-typecons-nullable-apply

nordlow:
Ok, I'm all in on using `Nullable`. Lets make this happen!

jacob-carlborg:
I don’t think we should use Nullable, instead implement a proper Optional type. I’ve an implementation here [1] that implements the range interface and safe access. 

[1] https://github.com/jacob-carlborg/dlp/blob/master/source/dlp/core/optional.d

nordlow:
Can you briefly describe why we need an Optional instead of a Nullable?

jacob-carlborg:
> Can you please describe why we need an Optional instead of a Nullable? Are there any other reasons than the above mentioned?

No, it's mostly because it has a range interface and safe access interface. I also think it's a better, more descriptive name.

nordlow:
How the is the interface `@safer`? 

AFAICT `Optional.get` is as (un)safe as `Nullable.get` is. Both may result in unexpected behaviour when compiled without asserts.

jacob-carlborg:
> How the is the interface `@safer`?

I didn't say `@safe`, I said "safe access", without the at sign. "safe" might not be the best name but what I mean is you can do this:

```
class Bar { int a; }
class Foo { Bar bar; }
auto foo = none!Foo;
auto a = foo.bar.a; // safe access, "a" will be an optional int
auto b = a.or(3); // unwraps "a" if it has a value, otherwise returns 3
auto c = a.map!(i => i * 4); // it also supports the range interface
```

If the standard algorithms are used, the lambda will only be called if the optional has a value. When using `map` it's basically like `apply` for `Nullable` but with a standard interface.

nordlow:
So lets try adding your `opDispatch` logic to `Nullable` then!

ZombineDev:
How about until we all settle on the return type, we simply use `auto`? That way the exact return type becomes an implementation detail and in future, when we have made the necessary changes to `Nullable` or we have added an new type, we could switch from `auto` to an explicit return type.

ZombineDev:
Another option would be add these functions first to `std.experimental`, so we would be free to change the return type later.

jacob-carlborg:
> So lets try adding your opDispatch to Nullable then!

Please no. I suggest we deprecate `Nullable` and add a new implementation called `Optional`. The name `Nullable` is poorly chosen. The name `Nullable` suggests it should only be used for value types which otherwise cannot be assigned `null`. But `Optional` is supposed to be used for all types, even those that can be assigned `null`. An optional pointer would mean that the pointer might not point to something, a plain pointer would mean that it always points to something, it can never be `null` (I know that this is not enforced by the language).

aliak00:
I agree `Nullable` can be used here but it's says nothing about "this value may be present" in a truely generic sense - null is a very valid part of some value type domains - e.g. pointers. 

An optional is explicit in intent and it'd be a much nicer interface with an Optional. Also, once we start using `Nullable` in phobos for stuff like this it's going to be very hard to go back :/ 

So super big +1 for putting this in std.experimental until we have an Optional type.

jacob-carlborg:
Nothing outside `std.experimental` is allowed to use `std.experimental`. Perhaps create an internal function marked with `package` to be able to share it?

nordlow:
You mean an internal function `_tryParse()` marked with package inside std.experimental.typecons?

nordlow:
Incidentally...: https://forum.dlang.org/post/vyzbihhgrvmiqacakezk@forum.dlang.org

wilzbach:
> Nothing outside std.experimental is allowed to use std.experimental.

Nope, that's not true. You can't return experimental types, but Phobos __is allowed__ to use things from experimental internally. We do so with `checkedint` and the `allocator` packages in a few places already.

nordlow:
So, @wilzbach and @jacob-carlborg...should I put `tryParse`and `tryTo`

- in `std.experimental` and call them from `std.typecons` or
- put it `std.typecons` and mark the new functions with `package`?

jacob-carlborg:
I will let @wilzbach decide.

wilzbach:
> I will let @wilzbach decide.

I would go with `experimental` because they are then at least partially exposed whereas `package` functions are entirely hidden from the user.

wilzbach:
> You mean an internal function. say tryParse(), marked with package inside std.experimental.typecons?

Why `stdx.typecons` and not `std.experimental.conv`?

nordlow:
Ahh, my mistake..,of course they should go into `std.experimental.conv`.

jacob-carlborg:
> I would go with experimental because they are then at least partially exposed whereas package functions are entirely hidden from the user.

What I meant was to put all the logic in a package function inside `std.conv` and expose it in `std.experimental.conv.tryParse` and `std.conv.parse`.

std/conv.d
```diff
@@ -2636,21 +2637,7 @@ do
...
- *
- * Returns:
- *     An `enum` of type `Target`
- *
- * Throws:
- *     A $(LREF ConvException) if type `Target` does not have a member
- *     represented by `s`.
- */
-Target parse(Target, Source)(ref Source s)
+Nullable!(Target) tryParse(Target, Source)(ref Source s)
```
andralex: Initiative is good but I think we'd need an `Expected` facility to make it right: https://www.youtube.com/watch?v=nVzgkepAg5Y

aliak00: With a `noexcept` Expected facility this would be even better!
We can do this with SumType: https://run.dlang.io/is/82Q0DR

FeepingCreature:
I have only one plea.

If we introduce an `Optional` as a possible alternative to `Nullable`. Then whatever you do, _please for the love of God_ do not `alias get this`! It was an error with Nullable, though I suppose it did introduce feature parity with pointers in the field of unexpected crashes...

jacob-carlborg:
Yeah, `get` should be rarely used.

jmdavis:
> Yeah, get should be rarely used.

Actually, I think that his point was the opposite. Rather, `get` should be used explicitly to access the value rather than having it implicitly convert to the type it's wrapping, because the `alias this` makes it very easy to accidentally convert to the wrapped type, which of course blows up in your face if the `Nullable` doesn't contain a value. He's brought this up on several occasions previously and wanted to deprecate the `alias this` on `Nullable`, but a compiler bug prevented it.

jacob-carlborg:
Yes, I agree. I meant `get` should not be used, regardless if it's called directly or through `alias this`. Functions like `getOrElse`, `map` or `filter` should be used instead.

nordlow:
I've added a very basic version of `Expected` [here](https://github.com/nordlow/phobos-next/blob/master/src/expected.d). Would you be so kind to add some comments?

FeepingCreature:
I can't seem to add comments to an ordinary file. Could you open a mini-PR and link it here?

FeepingCreature:
The logic of using ranges is, I think, that a type which may either contain one value or no value, ie. Nullable, is sort of a special case of a type which may contain any number of values, ie. ranges.

jacob-carlborg:
> Is it ok to add Expected to this PR for now in order for you to make comments in it?

It would be better to have a separate PR. Would it make sense to build a new Optional type as well and have it be based on each other?

FeepingCreature:
Maybe `VariantN` should be the backing type in any case. I don't know if it has open issues.

jmdavis:
> Optional is not supposed to act like a pointer, in my opinion. Perhaps Nullable is.

Honestly, the way `Nullable` works is pretty much exactly what I'd expect from an `Optional` type except maybe for the issue of `alias this`. I really don't see any difference between a type emulating a pointer and one that just optionally holds a value.

Now, having a type that either holds a value or hold some sort of error type as proposed here _is_ something different, but in its case, it basically holds a value regardless. It's just that it could be the target type or the error type.

> Maybe VariantN should be the backing type in any case. I don't know if it has open issues.

Using std.variant does tend to cause subtle problems, but `union` has at least some of those problems on its own already. So, if the solution is going to contain a `union`, then maybe it would make more sense to use `VariantN`. Personally, when I've created types like `Expected` in the past, I've just used separate variables for the value and the error, since it's simpler and avoids the problems you typically get with unions. However, it does have the downside of making the struct take up more space. I don't know which approach is ultimately better.

FeepingCreature:
> I've just used separate variables for the value and the error

This runs into the problems that I've recently fixed in Nullable (by using a `union` :smile:) involving structs whose `.init` state is not a valid type instance, where destruction of the sum type leads to a false destructor call on the field, possibly causing crashes. So I definitely recommend using a `union`.

aliak00:
A union seems like a pretty good approach to back an Expected type - [here](https://run.dlang.io/is/oeiwXB)'s a dirty proof of concept based on [SumType](https://github.com/pbackus/sumtype) by @pbackus.

> I really don't see any difference between a type emulating a pointer and one that just optionally holds a value.

1) The semantics are completely different. Is "no value" a valid value? (this is a valid question for a pointer type, but not for optional). Who owns the value (valid for pointer, but not for an optional value type, barring indirections which applies to any value type). 2) safety - a pointer does not force you to check for presence in anyway. An optional, depending on how it's implemented, does. 3) intent - what does a pointer return mean? An optional is clear as day - ok this overlaps with 1.

While you can treat a pointer the same sometime, it becomes conventional and not enforceable.

FeepingCreature:
I played with it a bit, and added the Turducken testcase: [code](https://run.dlang.io/is/r6L73L). Seems to eat it fine.

pbackus:
I'm working on [a fleshed-out version of `Expected`](https://github.com/pbackus/expected). Comments welcome (though it's still pretty rough).

std/conv.d
nordlow: I believe we should use a Result type. FYI, @pbackus

andralex:
Just a note - returning either Nulltable or Optional from parsing functions makes it impossible to distinguish the cause of error when there could be more than one (e.g. end of stream vs. overflow). With Expected the result contains arbitrarily rich information in case of failure.

nordlow:
> Just a note - returning either Nulltable or Optional from parsing functions makes it impossible to distinguish the cause of error when there could be more than one (e.g. end of stream vs. overflow). With Expected the result contains arbitrarily rich information in case of failure.

Is Expected different from Result?

thewilsonator:
Result is usually
```
struct Result(T) { T thing; bool error; } 
```

whereas expected is more like
```
struct Expected(T)
{
     union { T thing; struct { void* errorObject; TypeInfo ti;} }
     bool error;
}
```

pbackus:
`Result` is typically a union of an expected value and an error value (e.g. in [Rust][1], [F#][2], [OCaml][3]).

`Expected(T, E)` is just another name for `Result(T, E)`; `Expected(T)` is equivalent to `Result(T, E)` with `E` fixed to some specific type, such as `Exception`.

Personally I would recommend against using the name "expected". The most common names for this type in functional-programming languages are "result" and "either".

[1]: https://doc.rust-lang.org/std/result/
[2]: https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/results
[3]: https://ocamlbook.org/error-handling/#the-functional-way--option-and-result-types

andralex:
One issue with `Result!(T, E)` is that it commits to the error type, which in our case means we need to use heap allocation for the exception. I wish we had a few more options for the error case:

* `Exception`
* a lambda
* perhaps a string describing the error

Wrt names I find `Result` is better than `Either` because the latter gives equal importance to the branches, whereas `Result` can be easier thought as "normal path and contingency path".

nordlow:
> One issue with `Result!(T, E)` is that it commits to the error type, which in our case means we need to use heap allocation for the exception. I wish we had a few more options for the error case:

How does `Result!(T, E)` limit `E` to be a heap-allocated exception?

andralex:
@nordlow I just meant it limits the error to one type. I think it would be better (without adding other types to Result) to allow not only allocated exceptions but lambdas that throw exceptions and strings from which exceptions can be constructed.

In brief: `Result!(T, Exception)` allocates the exception eagerly even if not thrown. A better approach would be `Result!T` that makes the exception construction lazy as well.

pbackus:
@andralex If we want to store information from which an exception can be constructed, rather than an actual `Exception`, then there's no reason to use `Exception` as the error type. Instead, we can use the type of that information directly; e.g.,

```d
alias ExceptionArgs = Tuple!(string, "msg", string, "file", size_t, "line");

Result!(T, ExceptionArgs) fun(...)
{
    // ...
    if (ok)
        return typeof(return)(T(...));
    else
        return typeof(return)(ExceptionArgs("no good", __FILE__, __LINE__));
}
```

There's plenty of room for polish here, of course, but the basic approach ought to work.

andralex:
@pbackus understood, though exposing the arguments all the way to the user-facing API is not really nice (what if you change them later etc).

Users should have a simple mechanism to get some uniform `Result!T` type they can they inspect - is this a `T`, and if not, what was the problem? Give me an integral error code, or a string description, or an `Exception` object I can throw around.

nordlow:
> Users should have a simple mechanism to get some uniform `Result!T` type they can they inspect - is this a `T`, and if not, what was the problem? Give me an integral error code, or a string description, or an `Exception` object I can throw around.

Note that SIL's `Result` type by design doesn't take the error type as template parameter exactly because of the same reason.

pbackus:
This sounds like a mechanism-vs-policy issue. Certainly it would be ideal if we could live in a world where all functions reported errors in some uniform way, and it might be a good idea to adopt such a policy for Phobos, but we do not live in such a world, and I do not think that a general-purpose `Result` implementation is the right place to encode such a policy. Rather, `Result` should provide the mechanism, and leave the policy to its users.

LightBender:
This idea would be better pursued as a low-level API design for Phobos 3.

PR #6666, opened 2018-08-15
Fix issue 19171 - Fix invalid assert in Array!bool slicing

schveiguy:


std/container/array.d
```diff
@@ -1664,8 +1664,9 @@ if (is(Unqual!T == bool))
         /// ditto
         Range opSlice(size_t low, size_t high)
         {
+            // Note: indexes start at 0, which is equivalent to _a
```
n8sh: I would suggest rewording this comment as "`// Note: indices are offsets from _a`" or just removing it.

schveiguy: Sorry, didn't get to this. It is simply an internal comment to try and express the issue that was there before, so probably not worth a followup.

PR #6667, opened 2018-08-15
Fix dxml regression in multiple arguments skipOver

wilzbach:
See https://github.com/dlang/phobos/pull/6143 and https://github.com/dlang/ci/pull/286

### DIFF
--- a/std/algorithm/searching.d
+++ b/std/algorithm/searching.d
@@ -4073,8 +4073,9 @@ template skipOver(alias pred = (a, b) => a == b)
             {
                 import std.algorithm.iteration : map;
                 import std.algorithm.searching : minElement;
+                import std.range : only;
                 // Shortcut opportunity!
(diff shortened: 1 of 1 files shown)


std/algorithm/searching.d
```diff
@@ -4353,6 +4354,19 @@ template skipOver(alias pred = (a, b) => a == b)
...
+@safe unittest
+{
+    import std.utf : byCodeUnit;
+
+    bool stripStartsWith(Text)(ref Text text, string needle)
+    {
+        return text.skipOver(needle.byCodeUnit());
+    }
+    auto text = "<xml></xml>"d.byCodeUnit;
+    assert(stripStartsWith(text, "<xml>"));
```
n8sh: It seems like you should verify that the right characters were skipped.

wilzbach: Done. The error was an compilation error, but added the verification anyhow as it doesn't hurt.

PR #5268, opened 2017-03-12
Make `each` capable of early stopping

andralex:
`Flag` to the rescue...

PR #6668, opened 2018-08-15
Add a changelog entry for each's early stopping

wilzbach:
See also: https://github.com/dlang/phobos/pull/5268

Preview:

image

### DIFF
--- a/changelog/std-algorithm-iteration-each-early-stopping.dd
+++ b/changelog/std-algorithm-iteration-each-early-stopping.dd
@@ -1,0 +1,13 @@
+`std.algorithm.iteration.each` is now capable of early-stopping
+
+$(REF each, std,algorithm,iteration) is now capable of exiting early.
+When a `No.each` $(REF Flag, std,typecons) is returned from the function that
+is called by `each`, the iteration will be aborted early.
+Analogously, returning `Yes.each` will continue the iteration.
+For example:
+
+---
+auto arr = [10, 20, 30];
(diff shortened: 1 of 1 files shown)


changelog/std-algorithm-iteration-each-early-stopping.dd
```diff
@@ -0,0 +1,19 @@
+`std.algorithm.iteration.each` is now capable of early-stopping
+
+$(REF each, std,algorithm,iteration) is now capable of exiting early.
+When a `No.each` $(REF Flag, std,typecons) is passed to `each`, the iteration will be aborted early.
+For example:
```
jondegenhardt: (Minor) I had to read this a couple times to get it. Perhaps better would be if the wording describes what the iteration function returns rather than what is passed to `each`. A slightly different example showing both yes and no cases might also do this. eg.
```
auto arr = [10, 20, 30];
arr.each!((n) { arr ~= n; return (n == 20) ? No.each : Yes.each; });
assert(arr == [10, 20, 30, 10, 20]);
```

wilzbach: Good idea. I tried to rephrase it.

jondegenhardt:
LGTM

wilzbach:
Ah damn. It looks like this was merged shortly after the 2.082 branch-off has happened.
Thus, I cherry-picked it for stable: https://github.com/dlang/phobos/pull/6676

PR #6670, opened 2018-08-16
Fix Issue 19085 - std.experimental.allocator.makeArray should work with void

wilzbach:
Not really convinced that we should encourage people to use `void`, but it's not too hard to support `void` here as this implies uninitialized memory.

### DIFF
--- a/std/experimental/allocator/common.d
+++ b/std/experimental/allocator/common.d
@@ -832,6 +832,7 @@ private template isAllZeroBits(T, T value)
     else static if (is(typeof(value is 0)))
         enum isAllZeroBits = value is 0;
     else static if (isStaticArray!(typeof(value)))
+    {
         enum isAllZeroBits = ()
         {
 
@@ -846,6 +847,7 @@ private template isAllZeroBits(T, T value)
             }
             else return false;
         }();
+    }
     else static if (is(typeof(value) == struct) || is(typeof(value) == union))
         enum isAllZeroBits = ()
         {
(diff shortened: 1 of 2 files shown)


std/experimental/allocator/package.d
```diff
@@ -1489,6 +1489,11 @@ private T[] uninitializedFillDefault(T)(T[] array) nothrow
             memset(array.ptr, 0xff, T.sizeof * array.length);
         return array;
     }
+    else static if (is(T == void))
+    {
+        // don't initialize void
+        return array;
+    }
```
John-Colvin: Just because it's `void[]` doesn't mean it shouldn't be initialised at all.

I think the correct change is to change `std.experimental.allocator.common.isInitAllZeroBits` to evaluate to `true` for `void` (and `void[N]`).

wilzbach: Hmm, DMD does seem to initialize `void[]` to zeros - https://run.dlang.io/is/Q4lNNs

n8sh:
Not quite right. This fails if added to the unittest:
```d
    static struct StructWithFixedSizeVoidArray
    {
        void[10] b;
    }
    static assert(isInitAllZeroBits!StructWithFixedSizeVoidArray);
```

ZombineDev:
@John-Colvin can you offer some more motivation for treating `void` like `ubyte`? What use cases does it add that `ubyte` doesn't cover? Although the current language implementation allows `new void[](10)`, I'm not sure it is a good thing, even for generic programming.

n8sh:
After this latest problem I'm proposing that DMD expose `__traits(isZeroInit, T)` so we can use it instead of `isInitAllZeroBits!T`. PR is at https://github.com/dlang/dmd/pull/8583.

std/experimental/allocator/common.d
```diff
@@ -831,7 +831,10 @@ private template isAllZeroBits(T, T value)
         enum isAllZeroBits = value is null;
     else static if (is(typeof(value is 0)))
         enum isAllZeroBits = value is 0;
+    else static if (is(Unqual!(ElementType!T) == void[]))
```
n8sh: This is unnecessary. Removing it doesn't cause the new unittests to fail.

PR #6674, opened 2018-08-20
[stable] Fix Issue 18682: Forward Nullable.toHash to contained value, if non-null.

FeepingCreature:
This was always wrong, however it has become urgent in 2.082 now that Nullable contains a union and can no longer generate an implicit toHash to call the contained type's toHash.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2675,6 +2675,11 @@ Params:
         assert(e != 12);
     }
 
+    size_t toHash() const @safe nothrow
+    {
+        return _isNull ? 0 : typeid(T).getHash(&_value.payload);
+    }
+
     /**
      * Gives the string `"Nullable.null"` if `isNull` is `true`. Otherwise, the
      * result is equivalent to calling $(REF formattedWrite, std,format) on the
@@ -3382,6 +3387,21 @@ auto nullable(T)(T t)
         destructorCalled = false; // reset after S was destroyed in the NS constructor
     }
     assert(destructorCalled);
+}
+
+// check that toHash on Nullable is forwarded to the contained type
+@safe unittest
+{
+    struct S
+    {
+        size_t toHash() const @safe pure nothrow { return 5; }
+    }
+
+    Nullable!S s1 = S();
+    Nullable!S s2 = Nullable!S();
+
+    assert(typeid(Nullable!S).getHash(&s1) == 5);
+    assert(typeid(Nullable!S).getHash(&s2) == 0);
 }
 
 /**


FeepingCreature:
Rebased to `stable`.

std/typecons.d
```diff
@@ -2675,6 +2675,11 @@ Params:
         assert(e != 12);
     }
 
+    size_t toHash() const @safe nothrow
+    {
+        return _isNull ? 0 : typeid(T).getHash(&_value.payload);
```
n8sh: When we reach release 2.082 ` typeid(T).getHash(&_value.payload)` could be replaced with `.hashOf(_value.payload)` to avoid looking up at runtime information known at compile time and to enable inlining.

FeepingCreature: Oh, nice!

FeepingCreature: Ah! .hashOf is @system...

n8sh: >`toHash` is not marked as `const` in `Object`.
Yeah. It's a mess.

n8sh: Casting away `const` is what `typeid(Object).getHash(&_value.payload)` does too. But because it is a `@trusted` function there's no warning when using it in `@safe` code.

[object.d#L1700-1716](https://github.com/dlang/druntime/blob/430585650930797369eeb5b3b142faba10572f10/src/object.d#L1700-L1716):
```d
class TypeInfo_Class : TypeInfo
{
    override string toString() const { return info.name; }

    override bool opEquals(Object o)
    {
        if (this is o)
            return true;
        auto c = cast(const TypeInfo_Class)o;
        return c && this.info.name == c.info.name;
    }

    override size_t getHash(scope const void* p) @trusted const
    {
        auto o = *cast(Object*)p;
        return o ? o.toHash() : 0;
    }
```

andralex: @dkgroot I'm not familiar with the terminology widening/narrowing (more with upcast/downcast). Upcasts are implicit and fine, forced downcasts from `void*` or unrelated types are not safe for obvious reasons (incompatible layouts etc). So you're asking about casting a reference to base class to a reference to derived class? That shouldn't be unsafe in D because it returns null when the cast doesn't work.

dkgroot: @andralex :
My question was partially unrelated to this PR, more a general dlang quesion (Maybe asking inside a PR Review is not the right place). In this PR the issue is centered around casting away const, which i think could/should be a considered allowable inside a @safe block.

The definition of "widening/narrowing" cast explained in the question above, they are similar to up/down casts but mostly pertain to class inheritance.

ZombineDev: @dkgroot 
Type qualifier casts are orthogonal to class widening/narrowing casts and are always unsafe when you're stripping them off (`const`/`immitable` -> mutable, `shared` -> non-`shared`). `immutable` is implicitly convertible to `const` on the premise that casting `const` away is undefined behavior. A `const` reference may actually point to an `immuable` object stored in read-only memory, so you can imagine what sort of damage mutating such object may cause.

ZombineDev: @FeepingCreature the `@safe`-ty of this method should depend on whether `.hashOf(_value.payload)` is safe - `hashOf` will be our source of truth. So you can ahead and use it, by removing the expicit `@safe` attribute and letting the compiler infer the safety.

FeepingCreature: How would it be inferred? toHash is a regular function.

FeepingCreature: update: As far as I can see, `.toHash` simply does not work at all, since the class `hashOf` in druntime is itself defined `@safe` but does an unsafe operation on line 615.

n8sh: >How would it be inferred? toHash is a regular function. Inference only happens for template functions.

Inferrence happens for `Nullable(T).toHash` because `Nullable` is templated. For instance:

https://run.dlang.io/is/0pWg2Z
```d
import std.stdio;

struct Nullable(T)
{
    T val;
    bool isNull;
    alias val this;
    
    this(A...)(auto ref A args) if (is(typeof(T(A)) : A)) { isNull = false; val = T(A); }
    
	size_t toHash() const nothrow
    {
        return isNull ? 0 : .hashOf(val);
    }
}

void main() @safe
{
    writeln(Nullable!int.init.toHash());
}
```

FeepingCreature:
ping

ZombineDev:
Fair enough. I opened https://issues.dlang.org/show_bug.cgi?id=19197 to remind us for the potential for improvement.

FeepingCreature:
Thank you!

std/typecons.d
```diff
@@ -3384,6 +3389,21 @@ auto nullable(T)(T t)
...
+    {
+        size_t toHash() const @safe pure nothrow { return 5; }
+    }
+
+    Nullable!S s1 = S();
+    Nullable!S s2 = Nullable!S();
+
+    assert(typeid(Nullable!S).getHash(&s1) == 5);
+    assert(typeid(Nullable!S).getHash(&s2) == 0);
+}
```
kinke: Taking the address of locals in a `@safe` unittest? https://run.dlang.io/is/6VLvzX
How did this pass the CI tests?!

kinke: Pinging @MartinNowak; this is a blocker for the final (obviously).

FeepingCreature: Were the tests run with DIP1000? It works with that.

kinke: Omg... https://github.com/dlang/phobos/blob/master/dip1000.mak - unused by LDC, we use CMake. This is the first module failing to compile without it.

PR #6678, opened 2018-08-24
std.experimental.logger: Change LogLevel documentation

ljmf00:
Fix this issue on documentation:
image

### DIFF
--- a/std/experimental/logger/core.d
+++ b/std/experimental/logger/core.d
@@ -685,19 +685,27 @@ that logger is called.
 */
 enum LogLevel : ubyte
 {
-    all = 1, /** Lowest possible assignable `LogLevel`. */
-    trace = 32, /** `LogLevel` for tracing the execution of the program. */
-    info = 64, /** This level is used to display information about the
-                program. */
-    warning = 96, /** warnings about the program should be displayed with this
-                   level. */
-    error = 128, /** Information about errors should be logged with this
-                   level.*/
-    critical = 160, /** Messages that inform about critical errors should be
-                    logged with this level. */
-    fatal = 192,   /** Log messages that describe fatal errors should use this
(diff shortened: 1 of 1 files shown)


ZombineDev:
This actually seems like a bug in [dpl-docs](http://dpldocs.info/experimental-docs/std.experimental.logger.core.LogLevel.html)' ddoc-parsing engine. As per https://dlang.org/spec/ddoc#parsing:
> 1. [..] If the documentation comment appears on the same line to the right of a declaration, it applies to that.

I filed a bug: https://github.com/adamdruppe/adrdox/issues/30. Usually @adamdruppe I quite responsive, so I prefer to wait and see if this is something that could be easily fixed on his side, otherwise I would be fine with merging your PR as it doesn't negatively affect the other documentation engines and it would be could if documentation rendered correctly on dpl-docs too. My preference is for fixing this on the adrdox, as this bug probably affects other things too.

adamdruppe:
On Fri, Aug 24, 2018 at 06:48:28PM +0000, Petar Kirov wrote:
> I filed a bug: https://github.com/adamdruppe/adrdox/issues/30. Usually @adamdruppe I quite responsive, so I prefer to wait and see if this is something that could be easily fixed on his side

Yes, this should be an easy fix for me when I am back from vacation,
I can do a fix on my end for the adrdox bug Monday or Tuesday.

burner:
in the phobos docs it look corrent

ljmf00:
> in the phobos docs it look corrent

Yeah, this is just happening in [dpldocs](http://dpldocs.info/experimental-docs/std.experimental.logger.core.LogLevel.html) .

ljmf00:
According to D Documentation Generator specs, the documentation is well written. So I will close this issue

PR #6679, opened 2018-08-25
fix issue 11052 - pathSplitter cannot be converted to string[], and to!() fails

BBasile:
`to` now works on `pathSplitter`result using the cast operator overloading. Since the function returns a lazy range there's very little chances that anyone ever used  `cast` on it (which would stop working with this change).


BBasile:
Actually this is more an enhancement; `to` has no overload that converts input ranges to arrays.

BBasile:
This is useless actually.

PR #6680, opened 2018-08-25
fix issue 11431 - std.file.slurp fails with Windows newlines

BBasile:


std/file.d
```diff
@@ -5098,13 +5098,15 @@ slurp(Types...)(string filename, scope const(char)[] format)
...
     import std.stdio : File;
+    import std.string : stripRight;
 
     auto app = appender!(typeof(return))();
     ElementType!(typeof(return)) toAdd;
     auto f = File(filename);
     scope(exit) f.close();
     foreach (line; f.byLine())
     {
+        line = stripRight(line, ['\r']);
```
CyberShadow: `['\r']` doesn't compile in `@nogc`. Maybe just use `"r"`?

BBasile: Yes, fortunately the right overload exists.

PR #6681, opened 2018-08-25
Fix issue 11959 - Set private symbols declared in version(unittest) blocks

BBasile:
A check in D-Scanner is planned for this.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -5125,7 +5125,7 @@ version(unittest) private class __conv_EmplaceTestClass
 version(unittest)
 {
     //Ambiguity
-    struct __std_conv_S
+    private struct __std_conv_S
     {
         int i;
         this(__std_conv_SS ss)         {assert(0);}
@@ -5135,7 +5135,7 @@ version(unittest)
             return s;
         }
     }
-    struct __std_conv_SS
+    private struct __std_conv_SS
     {
         int j;
         __std_conv_S s;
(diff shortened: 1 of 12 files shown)


std/experimental/checkedint.d
```diff
@@ -2660,6 +2660,7 @@ if (isIntegral!T && T.sizeof >= 4)
 
 version(unittest) private struct CountOverflows
 {
+private:
```
ZombineDev: The symbol (`CountOverflows`) defined in the unittest block is already `private`. Perhaps this is a false positive on DScanner's side.

std/experimental/checkedint.d
```diff
@@ -2685,6 +2686,7 @@ version(unittest) private struct CountOverflows
 
 version(unittest) private struct CountOpBinary
 {
+private:
```
ZombineDev: ditto

ZombineDev:
@wilzbach the style target fails with:
```
make -C ../druntime -f posix.mak MODEL=64 DMD=../dmd/generated/linux/release/64/dmd OS=linux BUILD=release
make[1]: *** ../druntime: No such file or directory.  Stop.
posix.mak:469: recipe for target '../druntime/generated/linux/release/64/libdruntime.a' failed
make: *** [../druntime/generated/linux/release/64/libdruntime.a] Error 2
```

I think this a config problem on the Buildkite's side.

PR #6682, opened 2018-08-26
Add a getter and setter for booleans in JSON

Erikvv:
Works the same as `integer`, `floating` etc. 

Seems obvious you need this to use the module in any real-world usage.

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -231,6 +231,38 @@ struct JSONValue
     }
 
     /***
+     * Value getter/setter for boolean stored in JSON.
+     * Throws: `JSONException` for read access if `this.type` is not
+     * `JSONType.true_` or `JSONType.false_`.
+     */
+    @property bool boolean() const pure @safe
+    {
+        if (type == JSONType.true_) return true;
+        if (type == JSONType.false_) return false;
+
+        throw new JSONException("JSONValue is not a boolean type");
+    }
+    /// ditto
+    @property bool boolean(bool v) pure nothrow @safe @nogc
+    {
+        assign(v);
+        return v;
+    }
+    ///
+    @safe unittest
+    {
+        JSONValue j = true;
+        assert(j.boolean == true);
+
+        j.boolean = false;
+        assert(j.boolean == false);
+
+        j.integer = 12;
+        import std.exception : assertThrown;
+        assertThrown!JSONException(j.boolean);
+    }
+
+    /***
      * Value getter/setter for `JSONType.object`.
      * Throws: `JSONException` for read access if `type` is not
      * `JSONType.object`.


std/json.d
```diff
@@ -230,6 +230,43 @@ struct JSONValue
         return store.floating;
     }
 
+    /***
+     * Value getter/setter for boolean stored in JSON
```
ZombineDev: Missing full stop at the end of the sentence.

std/json.d
```diff
@@ -230,6 +230,43 @@ struct JSONValue
...
+     * Throws: `JSONException` for read access if `this.type` is not
+     * `JSONType.true_` or `false_`.
```
ZombineDev: Prefer spelling out the full name:
```
`JSONType.true_` or `JSONType.false_`.
```

ZombineDev:
Since this adds a new public symbol, @andralex will need to sign off on this PR before it can be merged. On the other hand the newly added `@property` is done analogously to the other ones so process should be straightforward.

std/json.d
Erikvv: Done

std/json.d
Erikvv: Done. I find the comment has too much implementation details but it's the same in the rest of the module.

std/json.d
```diff
@@ -230,6 +230,43 @@ struct JSONValue
         return store.floating;
     }
 
+    /***
+     * Value getter/setter for boolean stored in JSON.
+     * Throws: `JSONException` for read access if `this.type` is not
+     * `JSONType.true_` or `JSONType.false_`.
+     */
+    @property inout(bool) boolean() inout pure @safe
```
andralex: Why inout? This should suffice: `@property bool boolean() const pure @safe`. That ensures no surprise with functions that initialize an `auto` with the result of the call.

std/json.d
```diff
@@ -230,6 +230,43 @@ struct JSONValue
...
+    {
+        switch (this.type)
+        {
+            case JSONType.true_:
+                return true;
+            case JSONType.false_:
+                return false;
+            default:
+                throw new JSONException("JSONValue is not a boolean type");
+        }
```
andralex: A three-branch switch seems a bit much to me.

```
if (type == JSONType.true_) return true;
if (type == JSONType.false_) return false;
throw ...
```

But... six vs. half a dozen.

Erikvv: Done.

std/json.d
Erikvv: You're right I changed it to const. I'll make another PR to change the other getters as well.

Erikvv: #6683

PR #6683, opened 2018-08-27
Remove inout from JSON valuetype getters

Erikvv:
After feedback of #6682


n8sh:
You've unintentionally included https://github.com/dlang/phobos/pull/6682. Other than that, thumbs up.

PR #6684, opened 2018-08-28
Fix issue 19200 - Variant operators don't overload correctly

Biotronic:


std/variant.d
```diff
@@ -1021,64 +1021,15 @@ public:
...
-    // }
-    ///ditto
-    VariantN opUShr(T)(T rhs) { return opLogic!(T, ">>>")(rhs); }
-    // ///ditto
-    // VariantN opUShr_r(T)(T lhs)
-    // {
-    //     return VariantN(lhs).opLogic!(VariantN, ">>>")(this);
-    // }
+    VariantN opBinary(string op, T)(T rhs)
+    if ((op == "&" || op == "|" || op == "^" || op == ">>" || op == "<<" || op == ">>>") &&
```
ZombineDev: It looks like the style target fails because of trailing whitespace:
```
Running DScanner
generated/dscanner-e4ee46366d21eaa613da13c52f3f1a66684cdbaa/dsc --config .dscanner.ini --styleCheck etc std -I.
Check for trailing whitespace
grep -nr '[[:blank:]]$' etc std ; test $? -eq 1
std/variant.d:1025:    if ((op == "+" || op == "-" || op == "*" || op == "/" || op == "^^") &&
std/variant.d:1030:    if ((op == "&" || op == "|" || op == "^" || op == ">>" || op == "<<" || op == ">>>") &&
posix.mak:561: recipe for target 'style_lint' failed
make: *** [style_lint] Error 1
rm generated/linux/release/64/publictests/.directory
🚨 Error: The command exited with status 2
```

std/variant.d
```diff
@@ -3028,3 +2979,19 @@ if (isAlgebraic!VariantType && Handler.length > 0)
...
+    {
+        static int opBinaryRight(string op : "|", T)(T rhs)
+        {
+            return 3;
+        }
+    }
+
+    S s;
+    Variant v;
+    auto b = v | s;
```
ZombineDev: Can you assert that the result is 3?

n8sh: Added.

PR #6686, opened 2018-08-29
WIP: Add Expected

nordlow:
As ref see @andralex's [talk](https://www.youtube.com/watch?v=nVzgkepAg5Y) and https://forum.dlang.org/thread/hdflitsqoemcrwmvmewc@forum.dlang.org

Code has TODOs that should be resolved.

Putting this out early to get feedback.

Implementation doesn't depend on other modules and compiles (in isolation from typecons) very fast (~46 ms in my machine) (with DMD -debug -unittest).

**Update:** Are we sure we shouldn't name `Expected` as `Result` instead?

**Update**: For reference, also see: https://github.com/pbackus/expectations

### DIFF
--- a/std/experimental/typecons.d
+++ b/std/experimental/typecons.d
@@ -1118,6 +1118,8 @@ private struct Unexpected(E)
 struct Expected(T, E = Exception)
 if (!isInstanceOf!(Unexpected, T)) // an `Unexpected` cannot be `Expected` :)
 {
+    import std.algorithm.mutation : moveEmplace;
+
     // TODO ok for default construction to initialize
     // - _expectedValue = T.init (zeros)
     // - _ok = true (better to have _isError so default is zero bits here aswell?)
@@ -1126,7 +1128,7 @@ if (!isInstanceOf!(Unexpected, T)) // an `Unexpected` cannot be `Expected` :)
     this()(auto ref T expectedValue) @trusted
     {
         // TODO reuse opAssign?
-        _expectedValue = expectedValue;       // TODO use `moveEmplace` here aswell?
+        moveEmplace(expectedValue, _expectedValue);
         _ok = true;
     }
 
@@ -1134,7 +1136,7 @@ if (!isInstanceOf!(Unexpected, T)) // an `Unexpected` cannot be `Expected` :)
     this(Unexpected!E unexpectedValue) @trusted
     {
         // TODO reuse opAssign?
-        _unexpectedValue = unexpectedValue; // TODO use `moveEmplace` here aswell?
+        moveEmplace(unexpectedValue, _unexpectedValue);
         _ok = false;
     }
 
@@ -1143,7 +1145,6 @@ if (!isInstanceOf!(Unexpected, T)) // an `Unexpected` cannot be `Expected` :)
     {
         // TODO is this ok?:
         clear();
-        import std.algorithm.mutation : moveEmplace;
         moveEmplace(expectedValue, _expectedValue);
         _ok = true;
     }
@@ -1152,7 +1153,6 @@ if (!isInstanceOf!(Unexpected, T)) // an `Unexpected` cannot be `Expected` :)
     void opAssign(E unexpectedValue) @trusted
     {
         clear();
-        import std.algorithm.mutation : moveEmplace;
         moveEmplace(unexpectedValue, _unexpectedValue);
         _ok = false;
     }
--- a/std/experimental/typecons.d
+++ b/std/experimental/typecons.d
@@ -1131,7 +1131,7 @@ if (!isInstanceOf!(Unexpected, T)) // an `Unexpected` cannot be `Expected` :)
         _ok = true;
     }
     /// ditto
-    this(const scope ref T expectedValue)
+    this(ref T expectedValue)
     {
         _expectedValue = expectedValue; // TODO do we need to care about `_expectedValue` init value??
         _ok = true;
@@ -1144,7 +1144,7 @@ if (!isInstanceOf!(Unexpected, T)) // an `Unexpected` cannot be `Expected` :)
         _ok = false;
     }
     /// ditto
-    this(const scope ref Unexpected!E unexpectedValue)
+    this(ref Unexpected!E unexpectedValue)
     {
         _unexpectedValue = unexpectedValue; // TODO do we need to care about `_unexpectedValue` init value??
         _ok = false;
--- a/std/experimental/typecons.d
+++ b/std/experimental/typecons.d
@@ -1074,3 +1074,258 @@ pure nothrow @system unittest
         arr.ptr++
     ));
 }
+
+import std.traits : isInstanceOf;
+
+/** Wrapper type for an unexpected value of type `E`.
+ */
+private struct Unexpected(E)
+{
+    E value;
+    alias value this;
+}
+
+/** Union (sum) type of either an expected (most probable) value of type `T` or
+ * an unexpected value of type `E` (being an instance of type `Unexpected!E`).
+ *
+ * `E` is typically an error code (for instance C/C++'s' `errno` int) or a
+ * subclass of `Exception` (which is the default).
+ *
+ * See_Also: https://www.youtube.com/watch?v=nVzgkepAg5Y
+ * See_Also: https://github.com/dlang/phobos/pull/6665
+ *
+ * TODO https://dlang.org/phobos/std_typecons.html#.apply
+ *
+ * TODO I'm not convinced about the naming
+ * - `Expected`: instead call it something that tells us that it can be either expected or unexpected?
+ * - `Unexpected`: if so why shouldn't we have a similar value wrapper `Expected`?
+ *
+ * TODO we could get around the `Unexpected` wrapper logic by instead expressing
+ * construction in static constructor functions, say,:
+ * - static typeof(this) fromExpectedValue(T expectedValue)
+ * - static typeof(this) fromUnexpectedValue(E unexpectedValue)
+ *
+ * TODO swap
+ *
+ * TODO which functions should be `nothrow`?
+ *
+ * TODO later on: remove _ok when `_expectedValue` and ` _unexpectedValue` can store this state
+ * "collectively" for instance when both are pointers or classes (use trait
+ * `isAddress`)
+ *
+ * TODO ok to default `E` to `Exception`?
+ */
+struct Expected(T, E = Exception)
+if (!isInstanceOf!(Unexpected, T)) // an `Unexpected` cannot be `Expected` :)
+{
+    // TODO ok for default construction to initialize
+    // - _expectedValue = T.init (zeros)
+    // - _ok = true (better to have _isError so default is zero bits here aswell?)
+
+    /// Construct from expected value `expectedValue.`
+    this(T expectedValue) @trusted
+    {
+        // TODO reuse opAssign?
+        _expectedValue = expectedValue;       // TODO use `moveEmplace` here aswell?
+        _ok = true;
+    }
+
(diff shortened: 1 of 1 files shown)


std/experimental/typecons.d
```diff
@@ -1074,3 +1074,258 @@ pure nothrow @system unittest
...
+ */
+struct Expected(T, E = Exception)
+if (!isInstanceOf!(Unexpected, T)) // an `Unexpected` cannot be `Expected` :)
+{
+    // TODO ok for default construction to initialize
+    // - _expectedValue = T.init (zeros)
+    // - _ok = true (better to have _isError so default is zero bits here aswell?)
+
+    /// Construct from expected value `expectedValue.`
+    this(T expectedValue) @trusted
```
edi33416: What if `T` is not copyable? Does it make sense for such a `T` to exist in this context?
Also, you should probably pass a `ref T` or an `auto ref T` to avoid a postblit call when passing the argument.

edi33416: Shouldn't just mark the functions as `@trusted`. The safety should be inferred from `T` and probably `E` as well.

std/experimental/typecons.d
```diff
@@ -1074,3 +1074,258 @@ pure nothrow @system unittest
...
+    {
+        // TODO reuse opAssign?
+        _expectedValue = expectedValue;       // TODO use `moveEmplace` here aswell?
```
edi33416: I'd probably use `moveEmplace`.
Because it's part of an union, you must be careful and default initialize `_expectedValue` with `T.init`.

Take the following scenario:
If `T` has a `ptr` and `opAssign` does some checks on `ptr` (like for reference counting) then you have UB.

std/experimental/typecons.d
nordlow: Done for ctor and `opAssign`.

Should we use `auto ref` for `Unexpected` too?

nordlow: Ahh, I see if assignment/copying of `T` and `E` are not @safe this covers that fact, right?

Do you have a suggestion for a fix?

std/experimental/typecons.d
nordlow: Done.

std/experimental/typecons.d
```diff
@@ -1074,3 +1074,258 @@ pure nothrow @system unittest
...
+    // - _expectedValue = T.init (zeros)
+    // - _ok = true (better to have _isError so default is zero bits here aswell?)
+
+    /// Construct from expected value `expectedValue.`
+    this()(auto ref T expectedValue) @trusted
+    {
+        // TODO reuse opAssign?
+        moveEmplace(expectedValue, _expectedValue);
+        _ok = true;
+    }
```
pbackus: `ref` + `moveEmplace` means that this constructor will reset any lvalue passed to it to `T.init`. Probably not what you want.

std/experimental/typecons.d
```diff
@@ -1074,3 +1074,258 @@ pure nothrow @system unittest
...
+
+    /// Assign from expected value `expectedValue.`
+    void opAssign()(auto ref T expectedValue) @trusted
+    {
+        // TODO is this ok?:
+        clear();
+        moveEmplace(expectedValue, _expectedValue);
+        _ok = true;
+    }
```
pbackus: See above.

std/experimental/typecons.d
```diff
@@ -1074,3 +1074,258 @@ pure nothrow @system unittest
...
+    {
+        if (hasExpectedValue)
+        {
+            return expectedValue;
+        }
+        else
+        {
+            return elseWorkFun(); // TODO is this correct
+        }
+    }
```
pbackus: Because of [issue 5710](https://issues.dlang.org/show_bug.cgi?id=5710), this method needs to be an external function that takes an explicit "self" parameter.

std/experimental/typecons.d
```diff
@@ -1074,3 +1074,258 @@ pure nothrow @system unittest
...
+        alias fn = unaryFun!fun;
+        if (hasExpectedValue)
+        {
+            return typeof(return)(fn(_expectedValue));
+        }
+        else
+        {
+            return typeof(return)(Unexpected!E(_unexpectedValue));
+        }
+    }
```
pbackus: See above.

std/experimental/typecons.d
```diff
@@ -1074,3 +1074,258 @@ pure nothrow @system unittest
...
+        assert(_ok);
+        clear();
+    }
+
+private:
+    union
+    {
+        T _expectedValue;         // TODO do we need to default-initialize this somehow?
+        Unexpected!E _unexpectedValue;
+    }
```
pbackus: Because of [issue 19122](https://issues.dlang.org/show_bug.cgi?id=19122), this needs to be a named union, not an anonymous one.

std/experimental/typecons.d
nordlow: Fixed.

std/experimental/typecons.d
nordlow: ditto

std/experimental/typecons.d
nordlow: Do you mean that `valueOr` needs to be an external function?

std/experimental/typecons.d
nordlow: I guess `apply` needs to be external aswell?

std/experimental/typecons.d
nordlow: How do I alleviate this?

```D
union _data
{
    ...
}
```

fails....

std/experimental/typecons.d
pbackus: Yes. Sorry for any confusion; this is my first time using Github's code review interface.

nordlow: Ok, thanks. What do you mean by explicit "self" parameter?

pbackus: Meaning it takes the `Expected` as a normal parameter instead of via the implicit `this` parameter:

```d
// reduced a bit for brevity
T valueOr(alias elseFun)(Expected!T self)
    if (is(typeof(elseFun()) == T))
{
    if (self.hasExpectedValue) return self.expectedValue;
    else return elseFun();
}
```

std/experimental/typecons.d
pbackus:
```d
union Data
{
    // ...
}

Data _data;
```

std/experimental/typecons.d
nordlow: Done.

std/experimental/typecons.d
nordlow: Done.

nordlow:
Ping. I'm awaiting input now, especially from @andralex

std/experimental/typecons.d
pbackus: Apologies for missing this. Yes, it does.

std/experimental/typecons.d
```diff
@@ -1074,3 +1074,307 @@ pure nothrow @system unittest
...
+    private void clear() @trusted
+    {
+        release();
+        static if (isAddress!T)
+        {
+            _store.expectedValue = null;
+        }
+    }
+
+    /// Release any memory used to store contents.
```
pbackus: Misleading comment. This function calls `destroy` on the contents, which may free memory, or may free other resources, depending on what type is being stored. The memory used by the `Expected` instance itself is not released.

std/experimental/typecons.d
```diff
@@ -1074,3 +1074,307 @@ pure nothrow @system unittest
...
+        else
+        {
+            static if (!is(E == class))
+            {
+                static if (hasElaborateDestructor!E)
+                {
+                    destroy(_store.unexpectedValue);
+                }
+            }
+            destroy(_store.unexpectedValue);
```
pbackus: If this destroy call is correct, the one above it is unnecessary. If the one above is correct, this one is a bug. Either way, something is wrong here.

std/experimental/typecons.d
```diff
@@ -1074,3 +1074,307 @@ pure nothrow @system unittest
...
+
+    /// Pop (clear) current value.
+    void popFront()
+    {
+        assert(_ok);
+        clear();
+    }
+
+private:
+    union Store // TODO can we be sure this union is zero-initialized before constructor call?
```
pbackus: This question is answered in the D language specification: https://dlang.org/spec/struct.html#default_union_init

pbackus:
@nordlow, can you comment at all on the API and design questions I raised in [my initial comment on this PR](https://github.com/dlang/phobos/pull/6686#issuecomment-417945948)? Having heard only silence thus far, I am left with the impression that very little thought has gone into the design of this version of `Expected`, which is not encouraging.

RazvanN7:
ping @nordlow . What is the status of this and how should we proceed?

nordlow:
> @nordlow, can you comment at all on the API and design questions I raised in [my initial comment on this PR](https://github.com/dlang/phobos/pull/6686#issuecomment-417945948)? Having heard only silence thus far, I am left with the impression that very little thought has gone into the design of this version of `Expected`, which is not encouraging.

Has this been superceeded by a Result that reuses SumType?

pbackus:
> Can this be superceeded by a `Result` expressed in terms of the recently added `SumType`?

Yes, now that `SumType` is in Phobos, it would make sense to use it for something like this.

RazvanN7:
@nordlow OK to close?

nordlow:
> @nordlow OK to close?

Yes.

andralex:
Cool. Is there an open proposal for `Result`?

I agree with @pbackus about no necessity for a second template parameter. However, I'd like to retain (as an optimization) the possibility that the exception is not allocated before being thrown. For example, the parameters to constructing the exception could be saved (most often, a `string`) and the exception should be allocated if and only if the user wants it.

I'm unclear what name is best. `Expected` is present in C++ lore and being proposed for standardization. Doing something similar here just under a different name seems gratuitous.

pbackus:
I have actually come around on the idea of a second template parameter. It removes a hard dependency on the D runtime in general and on `Object` in particular, and makes the type easier to integrate with e.g. C functions that return error codes (one can use `Expected!(T, int)` directly instead of wrapping the `int` in a class instance). Even for those who don't mind using exceptions, there is still value in being able to write `Expected!(T, MyCustomException)` (e.g., if one wants to use an `Exception` subclass that defines `pure`/`nothrow`/`@safe`/etc. methods).

Re: naming, here are some names used in other languages:

* __C++__: [`expected`](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0323r7.html) has been proposed for standardization; [`result` and `outcome`](https://www.boost.org/doc/libs/1_70_0/libs/outcome/doc/html/index.html) are also used in Boost.
* __Rust__: [`Result`](https://doc.rust-lang.org/std/result/).
* __Scala__: [`Try`](https://www.scala-lang.org/api/current/scala/util/Try.html) and [`Either`](https://www.scala-lang.org/api/2.12.0/scala/util/Either.html).
* __Haskell__: [`Either`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Either.html).
* __OCaml__: [`result`](https://ocamlbook.org/error-handling/#the-functional-way--option-and-result-types)
* __F#__: [`Result`](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/results)

"Result" and "either" are probably the most familiar names to programmers with functional-programming experience. "Expected" is a C++-ism. In general, I think it is a good idea to do some research before taking naming cues from C++ (see also `std::accumulate` and `std::transform`, which are C++-isms for `fold`/`reduce` and `map`).

PR #6688, opened 2018-08-30
Document getopt output on stderr.

veelo:
Document `defaultGetoptFormatter` as an alternative to `defaultGetoptPrinter` for printing on stderr.

### DIFF
--- a/std/getopt.d
+++ b/std/getopt.d
@@ -1617,6 +1617,9 @@ foreach (it; opt)
 }
 ------------
 
+For printing on `stderr` instead, see the documentation of the function
+`defaultGetoptFormatter`.
+
 Params:
     text = The text to printed at the beginning of the help output.
     opt = The `Option` extracted from the `getopt` parameter.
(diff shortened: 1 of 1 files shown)


std/getopt.d
```diff
@@ -1632,6 +1635,13 @@ void defaultGetoptPrinter(string text, Option[] opt)
 in the manner described in the documentation of function
 `defaultGetoptPrinter`.
 
+The following prints `text` and `Option`s `opt` on `stderr`:
+
+------------
+import std.stdio : stderr;
+defaultGetoptFormatter(stderr.lockingTextWriter, text, opt);
```
ZombineDev: Can you move this to documented unittest?

veelo: You mean the existing one on row 445 or a new one?

I was actually about to close this PR as I reverted from doing this in my code. Now I think it is better to print an error on `stderr` referencing "-h", and continue printing the help on `stderr`.

Do you agree on closing, or do you find the PR helpful?

veelo:
The opinion of [stackoverflow](https://stackoverflow.com/questions/2199624/should-the-command-line-usage-be-printed-on-stdout-or-stderr) and [stackexchange](
https://unix.stackexchange.com/questions/8813/should-the-usage-message-go-to-stderr-or-stdout): options should go to `stdout` when requested by `-h`/`--help`. On error, a message to `stderr` is important, but options would normally not be part of that -- a reference to "--help" would suffice.

Closing.

PR #6689, opened 2018-09-01
switch Digital Mars copyright to D Language Foundation

WalterBright:


std/bitmanip.d
```diff
@@ -34,7 +34,7 @@ $(TR $(TD Tagging) $(TD
 ))
 )
 
-Copyright: Copyright Digital Mars 2007 - 2011.
+Copyright: Copyright The D Language Foundation" 2007 - 2011.
```
n8sh: Does the `"` do something in ddoc or should it be removed?

ibuclaw: Looks like a bad find/replace to me.

BBasile: The "Copyright" section doesn't even appear in the generated doc. But at the bottom the footer is already correct.

BBasile: Shortcut for new [bitmanip doc](http://dtest.dlang.io/artifact/website-748d1fb3b3acc80ef7bc5c69516165dc709b330b-d4c7dd79f99f3fa706c7f547db77e590/web/phobos-prerelease/std_bitmanip.html). No copyright at all at the top.

PR #6691, opened 2018-09-04
only run betterC tests on buildkite

MartinNowak:
- to avoid duplication of lengthy unittests (also run on auto-tester)


MartinNowak:
Mmh, hardly reduces test time, phobos unittest build time has become quite crazy since splitting up all tests (and adding more of 'em).

PR #6692, opened 2018-09-04
Implement allocateZeroed for building-block allocators

n8sh:
Followup to https://github.com/dlang/phobos/pull/6411

>Some allocators can allocate zeroed memory more efficiently than allocate + memset.
>[...]
>This PR adds non-public `allocateZeroed` methods to `Mallocator` and `MmapAllocator` and `GCAllocator`. Various derived allocators could also implement such methods when their parent allocators support them, but that work is left to a future PR

### DIFF
--- a/std/experimental/allocator/building_blocks/allocator_list.d
+++ b/std/experimental/allocator/building_blocks/allocator_list.d
@@ -190,6 +190,33 @@ struct AllocatorList(Factory, BookkeepingAllocator = GCAllocator)
         return null;
     }
 
+    static if (hasMember!(Allocator, "allocateZeroed"))
+    package(std) void[] allocateZeroed()(size_t s)
+    {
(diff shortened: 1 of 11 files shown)


std/experimental/allocator/building_blocks/allocator_list.d
```diff
@@ -190,6 +190,33 @@ struct AllocatorList(Factory, BookkeepingAllocator = GCAllocator)
...
+
+        // Add a new allocator
+        if (auto a = addAllocator(s))
+        {
+            auto result = a.allocateZeroed(s);
+            assert(owns(result) == Ternary.yes || !result.ptr);
+            return result;
+        }
+        return null;
+    }
```
wilzbach: There's a bit of duplication with `allocate` as essentially only the called subfunction differs. 
Maybe generate the string for both and mix it in?

PR #6693, opened 2018-09-05
Fix Issue 19226 - std.typecons.Nullable(T, T nullValue) doesn't fully handle non-self-equal nullValue

n8sh:
`Nullable(T, T nullValue)` special-cases floating point numbers but this doesn't account for all types where `nullValue != nullValue`. For instance:

https://run.dlang.io/is/hiLncI

```d
struct S
{
    float f;
}

void main()
{
    import std.typecons : Nullable;

    alias N = Nullable!(S, S.init);
    assert(N(S.init).isNull); // Fails!
}
```

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -3542,7 +3542,7 @@ Returns:
         }
         //Need to use 'is' if T is a float type
         //because NaN != NaN
-        else static if (isFloatingPoint!T)
+        else static if (__traits(isFloating, T) || __traits(compiles, { static assert(!(nullValue == nullValue)); }))
         {
             return _value is nullValue;
         }
@@ -3561,6 +3561,14 @@ Returns:
 
     ni = 0;
     assert(!ni.isNull);
+}
+
+@system unittest
+{
+    assert(typeof(this).init.isNull, typeof(this).stringof ~
+        ".isNull does not work correctly because " ~ T.stringof ~
+        " has an == operator that is non-reflexive and could not be" ~
+        " determined before runtime to be non-reflexive!");
 }
 
 // https://issues.dlang.org/show_bug.cgi?id=11135
@@ -3735,6 +3743,25 @@ if (is (typeof(nullValue) == T))
     assert(a == 8);
     a.nullify();
     assert(a.isNull);
+}
+
+@nogc nothrow pure @safe unittest
+{
+    // issue 19226 - fully handle non-self-equal nullValue
+    static struct Fraction
+    {
+        int denominator;
+        bool isNaN() const
+        {
+            return denominator == 0;
+        }
+        bool opEquals(const Fraction rhs) const
+        {
+            return !isNaN && denominator == rhs.denominator;
+        }
+    }
+    alias N = Nullable!(Fraction, Fraction.init);
+    assert(N.init.isNull);
 }
 
 @safe unittest


n8sh:
Opened bug report https://issues.dlang.org/show_bug.cgi?id=19227 *`S.init is S.init` failing for struct with float member*.

schveiguy:
Is there a codegen bug on Linux 32-bit? Seems odd that the test case would only fail on that platform.

n8sh:
It seems like this must be a bug.

std/typecons.d
```diff
@@ -3445,7 +3445,7 @@ Returns:
         }
         //Need to use 'is' if T is a float type
         //because NaN != NaN
-        else static if (isFloatingPoint!T)
+        else static if (__traits(compiles, { static assert(!(nullValue == nullValue)); }))
```
MetaLang: Is there no simpler way to write this?

n8sh: I wrote that instead of `else static if (nullValue != nullValue)` in case `opEquals` isn't CTFE-able. Is it okay if that causes compilation to fail?

MetaLang: I figured that was the reason. Unfortunately I don't think we have something like `if (CanExecAtCompileTime!(expr)`.

The problem is with requiring T.opEqual to be CTFE-able. I don't have a lot of time to think about it, but my first instinct would be to move the check to runtime time in this case.

```
...
else static if (isFloatingPoint!T) { ...}
else
{
    if (nullValue != nullValue)
        return _value is nullValue;
    else
        return _value == nullValue;
}
```

Though that's obviously not ideal.

n8sh: I think it would be sufficient to add a unittest inside the template along the lines of:

```d
@system unittest
{
    assert(typeof(this)(nullValue).isNull, typeof(this).stringof ~ 
        ".isNull does not work correctly because " ~ typeof(T).stringof ~
        " has an == operator that is non-reflexive and could not be " ~
        " determined before runtime to be non-reflexive!");
}
```

std/typecons.d
```diff
@@ -3476,7 +3476,7 @@ Returns:
         }
         //Need to use 'is' if T is a float type
         //because NaN != NaN
-        else static if (isFloatingPoint!T)
+        else static if (__traits(isFloating, T) || __traits(compiles, { static assert(!(nullValue == nullValue)); }))
```
n8sh: Q: Why is there the seemingly-extraneous `__traits(isFloating, T)`?
A: So this patch doesn't change the behavior of `Nullable!(double, -0.0)`.

n8sh:
The wrong code generation on 32-bit linux was still happening last I checked so I expect this not to go through.

12345swordy:
3 days have been passed already.

wilzbach:
> The wrong code generation on 32-bit linux was still happening last I checked so I expect this not to go through.

Yup, it's still there.

n8sh:
I've been unable to replicate the Linux_32 / Linux_64_32 problem locally so far so I'm unable to tell what the problem is or make a useful bug report.

n8sh:
I am changing the test case for this PR. Whatever is wrong with 32-bit linux is a separate issue.

n8sh:
I am changing the test case for this PR. Whatever is wrong with 32-bit linux is a separate issue.

n8sh:
I am changing the test case for this PR. Whatever is wrong with 32-bit linux is a separate issue.

PR #6695, opened 2018-09-06
Fix issue 19229 - formattedWrite destructively iterates over forward …

Biotronic:
…ranges

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -3920,6 +3920,11 @@ if (is(T == class) && !is(T == enum))
             {
                 formatObject(w, val, f);
             }
+            else static if (isForwardRange!T)
+            {
+                auto tmp = val.save;
+                formatRange(w, tmp, f);
+            }
             else static if (isInputRange!T)
             {
                 formatRange(w, val, f);
@@ -3944,7 +3949,6 @@ if (is(T == class) && !is(T == enum))
     // class range (issue 5154)
     auto c = inputRangeObject([1,2,3,4]);
     formatTest( c, "[1, 2, 3, 4]" );
-    assert(c.empty);
     c = null;
     formatTest( c, "null" );
 }
@@ -4084,6 +4088,11 @@ if (is(T == interface) && (hasToString!(T, Char) || !is(BuiltinTypeOf!T)) && !is
         {
             formatObject(w, val, f);
         }
+        else static if (isForwardRange!T)
+        {
+            auto tmp = val.save;
+            formatRange(w, tmp, f);
+        }
         else static if (isInputRange!T)
         {
             formatRange(w, val, f);
@@ -4164,6 +4173,11 @@ if ((is(T == struct) || is(T == union)) && (hasToString!(T, Char) || !is(Builtin
     static if (hasToString!(T, Char))
     {
         formatObject(w, val, f);
+    }
+    else static if (isForwardRange!T)
+    {
+        auto tmp = val.save;
+        formatRange(w, tmp, f);
     }
     else static if (isInputRange!T)
     {
@@ -6673,4 +6687,12 @@ char[] sformat(Char, Args...)(return scope char[] buf, scope const(Char)[] fmt, 
     cmp = "0100,000";
     tmp  = format("%08,d", 100_000);
     assert(tmp == cmp, tmp);
+}
+
+// Issue 19229
+@safe unittest
+{
+    import std.algorithm.sorting : sort;
+    auto a = [sort([1,2,3])];
+    assert(format("%s", a) == format("%s", a));
 }


thewilsonator:
You've somehow got a diff of `std.traits` against `std.format`, for a total of ~13k lines.

BBasile:
Yes, it looks like you have pasted std.format (patched) in std.traits.

Biotronic:
I'm impressed. Not in a good way, but impressed nonetheless. Fixed now.

std/format.d
```diff
@@ -3920,6 +3920,11 @@ if (is(T == class) && !is(T == enum))
             {
                 formatObject(w, val, f);
             }
+            else static if (isForwardRange!T)
+            {
+                auto tmp = val.save;
+                formatRange(w, tmp, f);
```
BBasile: can't you pass `val.save` as parameter directly, i.e w/o tmp ?

Biotronic: It's a ref parameter, so no.

Biotronic:
The test that failed is this:
```
    // class range (issue 5154)
    auto c = inputRangeObject([1,2,3,4]);
    formatTest( c, "[1, 2, 3, 4]" );
    assert(c.empty);
```
Turns out, `inputRangeObject` is a forward range if possible.

I just removed the assert that `formatTest` should leave the range exhausted, but I'm not sure if this is the correct behavior.

schveiguy:
Looks like this is also breaking `dxml` ping @jmdavis

schveiguy:
So here's an interesting observation -- previously, outputting an input range object would simply iterate the range. Yes it exhausted the range, but no allocations happened.

Now, outputting an input range object (that is a forward range) will *allocate a new object* to output it. We are trading correctness for performance, but I suppose that's the price you pay if you want to have an object represent your ranges.

std/format.d
```diff
@@ -6674,3 +6688,11 @@ char[] sformat(Char, Args...)(return scope char[] buf, scope const(Char)[] fmt,
...
     assert(tmp == cmp, tmp);
 }
+
+// Issue 19229
+@safe unittest
+{
+    import std.algorithm.sorting : sort;
+    auto a = [sort([1,2,3])];
+    assert(format("%s", a) == format("%s", a));
+}
```
jmdavis: This is fundamentally broken. If you pass a range to a function by value, it's copied. And the semantics of copying a range are unspecified, which means that unless you know for a fact that copying the range is equivalent to calling `save`, doing _anything_ with the range after copying it is broken. What it does depends entirely on how the range is defined and will vary from range to range. If you copy a range - including by passing it to a function - you _must_ call `save` on it if you then want to do anything else with it. So, for this example to be valid, `save` needs to be called on `a` (and if it's not yet the case that it's guaranteed that the evaluation order is left to right, then it actually needs to be called on both uses of `a` in the expression). As such, this is not fixing an actually bug. The bug is in the example.

schveiguy: If you add save to a, it will still fail.

schveiguy: > unless you know for a fact that copying the range is equivalent to calling `save`

In this case, the range is an array, so it's exactly equivalent.

jmdavis: True. Just calling `save` on the outer layer is not enough. You need to recursively call `save`. It's the same problem, just one level deeper. If you want to continue to use the ranges after passing them, you need to have `save`d them, because the function may iterate over them. In the case of `format`, it's definitely going to iterate over them, making it that much more critical that `save` be called, but the issue is the same regardless of which function you pass the range to. If you want an independent copy, then you need to call `save`, and any time a range is copied, you can't use the original anymore unless you know that copying it is the same as calling `save` on it. The fact that this is dealing with a range of ranges rather than just one level of ranges doesn't change that. It just makes it more annoying to deal with.

schveiguy: > You need to recursively call save

That's exactly what this PR does.

schveiguy: Note that in order to save recursively *before* calling formattedWrite, you have to duplicate the array. This means that an array whose items are ranges is not saveable unless you reallocate it.

The PR's semantics are much better -- no allocation required.

jmdavis: Except that calling `save` inside the function like this is already too late. It may happen to work with this particular test, but in the general case it won't. The range was already copied when it was passed to `formatElement`, resulting in wildly different semantics depending on how the range is implemented. Even if we want to make a change like this, `save` needs to be called on the inner range when it's passed to `formatElement` and not deeper in the call chain like it is here. You always need to call `save` before the copy is made, not after.

But even then, in the general case, when you pass a range of ranges to a function, you can't assume that you can then continue to do anything with the inner ranges without having `save`d them first, because what happens is highly dependent on how the ranges are implemented and what the function does. So, while making `formattedWrite` work in this particular case might ultimately be valuable, it's just working around the issue in this particular case. It's still a bug in general to pass a range of ranges to a function and then continue to do anything with the original ranges without having recursively saved them first. It will work in some cases to be sure, but not in the general case.

So, I'd definitely argue that this is not fixing a bug but rather is working around a bug in the caller, though maybe it's worth it in this particular case.

In any case, if we are going to do this, it needs to be calling `save` inside `formatRange` when it calls `formatElement` rather than calling it inside `formatValueImpl`. It's already way too late by the time it gets this deep.

Also, the tests need to include a variety of range types, not just a `SortedRange`. At minimum, it needs to be testing an inner type that is clearly a value type range and one that's clearly a reference type range.

berni44:
I closed the corresponding issue as "wontfix" because I think, this is part of something much greater, that cannot be fixed in one PR. We need to redefine the way, ranges work, and then we can take care of such things. Not to get it wrong: I'd prefere if the example given in the issue would work, I just think this should be done by implementing a greater concept, probably when going to D3.

For the time being, I think it's better to not add this, because then the user can choose between using `save` to keep a copy and faster performance. In other words, I think, this PR should be closed.

LightBender:
The issue was closed as "WONTFIX". Closing this PR.

PR #6696, opened 2018-09-08
[ARM] [GDC] Make hasExceptionTraps pure on ARM

jpf91:
To fix the ARM build on GDC, I guess LDC is affected as well? ping @klickverbot @JohanEngelen @kinke @ibuclaw 

I had to move the code to an external `_impl` function, because of this bug:
https://issues.dlang.org/show_bug.cgi?id=19235

An alternative might be inlining the `setControlState` code, but this implementation here is closer to the original code.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -6063,6 +6063,20 @@ nothrow @nogc:
     else
         static assert(false, "Not implemented for this architecture");
 
+    version (ARM_Any)
+    {
+        static bool hasExceptionTraps_impl() @safe
+        {
+            auto oldState = getControlState();
+            // If exceptions are not supported, we set the bit but read it back as zero
+            // https://sourceware.org/ml/libc-ports/2012-06/msg00091.html
+            setControlState(oldState | divByZeroException);
+            immutable result = (getControlState() & allExceptions) != 0;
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -6075,13 +6089,11 @@ public:
...
         {
-            auto oldState = getControlState();
-            // If exceptions are not supported, we set the bit but read it back as zero
-            // https://sourceware.org/ml/libc-ports/2012-06/msg00091.html
-            setControlState(oldState | divByZeroException);
-            immutable result = (getControlState() & allExceptions) != 0;
-            setControlState(oldState);
-            return result;
+            // The hasExceptionTraps_impl function is basically pure,
+            // as it restores all global state
```
n8sh: This is the same reasoning used in [`core.memory.pureMalloc`](https://dlang.org/phobos/core_memory.html#.pureMalloc) etc.
>Purity is achieved by saving and restoring the value of errno, thus behaving as if it were never changed.

kinke:
> I guess LDC is affected as well?

Yeah, we have a workaround (inlined asm) in place, this one here is better, thx.

PR #6697, opened 2018-09-09
Replace isInitAllZeroBits with new __traits(isZeroInit)

n8sh:
Requires https://github.com/dlang/dmd/pull/8583.

Also remove `isInitAllOneBits` and replace with direct check for `char`/`wchar`. `isInitAllOneBits` had the same problem with unreasonable memory usage as the old `isInitAllZeroBits` (https://issues.dlang.org/show_bug.cgi?id=18937).


std/experimental/allocator/package.d
```diff
@@ -1567,7 +1566,7 @@ T[] makeArray(T, Allocator)(auto ref Allocator alloc, size_t length)
         if (overflow) return null;
     }
 
-    static if (isInitAllZeroBits!T && hasMember!(T, "allocateZeroed"))
```
n8sh: This line previously had a bug.

PR #6699, opened 2018-09-09
Fix Issue 19197 - Replace instances of typeid(T).getHash(..) with hashOf

n8sh:
>Since core.internal.hash.hashOf is a template, there are a number potential benefits:
>* better performance (e.g. more opportunities for inlining, less indirections for value types)
>* less dependence on TypeInfo (which we want to phase out in the long term)
>* -betterC friendly (no need to link druntime and phobos to your library / app, in order to use simple templates like std.typecons.Nullable)

Additionally, using `hashOf` instead of `TypeInfo.getHash` allows correct function attribute inference.


n8sh:
This PR is failing CI because code that was unjustifiably `@trusted` is no longer `@trusted`, resulting in correct warnings about `toHash` being non-`@safe` (and warnings are treated as errors).

n8sh:
As a partial solution I'm going to fall back to `TypeInfo.getHash` when `hashOf` is not `@safe` and `nothrow`.

tibi77:
I have worked on this as well. My implementation is filing at line 2682 because the hashes differ with the ones that are in the unittest.

std/typecons.d
```diff
@@ -1175,7 +1175,12 @@ if (distinctFieldNames!(Specs))
             size_t h = 0;
             static foreach (i, T; Types)
             {{
-                const k = typeid(T).getHash((() @trusted => cast(const void*) &field[i])());
+                static if (__traits(compiles, h = .hashOf(field[i])))
+                    const k = .hashOf(field[i]);
+                else
+                    // Workaround for when .hashOf is not both @safe and nothrow.
+                    // BUG: Improperly casts away `shared`!
+                    const k = typeid(T).getHash((() @trusted => cast(const void*) &field[i])());
```
edi33416: Any reason not to add some `static if`s that check the type qualifiers and cast to the correct pointer type?

n8sh: That doesn't work :-/. If it did I wouldn't need to cast at all, I could just write ` &field[i]`.

Example:
https://run.dlang.io/is/ov8L8P
```d
void main()
{
    shared int a;
    int b = typeid(a).getHash(&a); // can't TypeInfo.getHash a shared argumennt
}
```

n8sh: To be clear, I don't think we should be doing this type of casting in `toHash`, but this was already being done in `Tuple` & `Unique` & `Typedef` so removing it could cause previously-compiling code using those templates to fail.

edi33416: Got it. Thanks for clarifying it for me.

std/typecons.d
```diff
@@ -2679,7 +2684,11 @@ Params:
 
     size_t toHash() const @safe nothrow
```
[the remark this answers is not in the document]
edi33416: Two questions pop in my mind:
- shouldn't it infer everything, including safety and thrownness?
- if we are enforcing `@safe nothrow`, why not go all the way and add `pure @nogc`? In the end, computing the hash must not produce any garbage and should always have the same behaviour.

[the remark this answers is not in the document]
n8sh: That would make this incompatible with many objects since the Object.toHash isn't `pure` or `@nogc`. It would also prevent this from working with TypeInfo because TypeInfo.getHash isn't `pure` or `@nogc`.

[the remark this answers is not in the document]
edi33416: Cool. So then, imho this PR looks good. Probably in the future we could make an attempt to make `Object.toHash` and `TypeInfo.getHash` also `pure @nogc`.

edi33416:
@n8sh seems like this PR also solved [this ](https://issues.dlang.org/show_bug.cgi?id=15641) regression. Can you please get this into `stable`?

PR #6700, opened 2018-09-10
Allow no-arg splitter function to work on ranges that aren't standard strings

schveiguy:
See bug report for more info.

Relevant discussion: https://forum.dlang.org/post/pn498r$2928$1@digitalmars.com

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -5001,53 +5001,113 @@ private struct SplitterResult(alias isTerminator, Range)
 }
 
 /++
-Lazily splits the string `s` into words, using whitespace as the delimiter.
+Lazily splits the character-based range `s` into words, using whitespace as the
+delimiter.
 
-This function is string specific and, contrary to
+This function is character-range specific and, contrary to
 `splitter!(std.uni.isWhite)`, runs of whitespace will be merged together
 (no empty tokens will be produced).
 
 Params:
-    s = The string to be split.
+    s = The character-based range to be split. Must be a string, or a
+    random-access range of character types.
 
 Returns:
     An $(REF_ALTTEXT input range, isInputRange, std,range,primitives) of slices of
-    the original string split by whitespace.
+    the original range split by whitespace.
  +/
-auto splitter(C)(C[] s)
-if (isSomeChar!C)
+auto splitter(Range)(Range s)
+if (isSomeString!Range ||
+    isRandomAccessRange!Range && hasLength!Range && hasSlicing!Range &&
+    !isConvertibleToString!Range &&
+    isSomeChar!(ElementEncodingType!Range))
 {
     import std.algorithm.searching : find;
     static struct Result
     {
     private:
         import core.exception : RangeError;
-        C[] _s;
+        Range _s;
         size_t _frontLength;
 
-        void getFirst() pure @safe
+        void getFirst()
         {
             import std.uni : isWhite;
+            import std.traits : Unqual;
 
-            auto r = find!(isWhite)(_s);
-            _frontLength = _s.length - r.length;
+            static if (is(Unqual!(ElementEncodingType!Range) == wchar) &&
+                       is(Unqual!(ElementType!Range) == dchar))
+            {
+                // all unicode whitespace characters fit into a wchar. However,
+                // this range is a wchar array, so we will treat it like a
+                // wchar array instead of decoding each code point.
+                _frontLength = _s.length; // default condition, no spaces
+                foreach (i; 0 .. _s.length)
+                    if (isWhite(_s[i]))
+                    {
+                        _frontLength = i;
+                        break;
+                    }
+            }
+            else static if (is(Unqual!(ElementType!Range) == dchar) ||
+                            is(Unqual!(ElementType!Range) == wchar))
+            {
+                // dchar or wchar range, we can just use find.
+                auto r = find!(isWhite)(_s.save);
+                _frontLength = _s.length - r.length;
+            }
+            else
+            {
+                // need to decode the characters until we find a space. This is
+                // ported from std.string.stripLeft.
+                static import std.ascii;
+                static import std.uni;
+                import std.utf : decodeFront;
+
+                auto input = _s.save;
+                size_t iLength = input.length;
+
+                while (!input.empty)
+                {
+                    auto c = input.front;
+                    if (std.ascii.isASCII(c))
+                    {
+                        if (std.ascii.isWhite(c))
+                            break;
+                        input.popFront();
+                        --iLength;
+                    }
+                    else
+                    {
+                        auto dc = decodeFront(input);
+                        if (std.uni.isWhite(dc))
+                            break;
+                        iLength = input.length;
+                    }
+                }
+
+                // sanity check
+                assert(iLength <= _s.length);
+
+                _frontLength = _s.length - iLength;
+            }
         }
 
     public:
-        this(C[] s) pure @safe
+        this(Range s)
         {
-            import std.string : strip;
-            _s = s.strip();
+            import std.string : stripLeft;
+            _s = s.stripLeft();
             getFirst();
         }
 
-        @property C[] front() pure @safe
+        @property auto front()
         {
             version(assert) if (empty) throw new RangeError();
             return _s[0 .. _frontLength];
         }
 
-        void popFront() pure @safe
+        void popFront()
         {
             import std.string : stripLeft;
             version(assert) if (empty) throw new RangeError();
@@ -5055,7 +5115,7 @@ if (isSomeChar!C)
             getFirst();
         }
 
-        @property bool empty() const @safe pure nothrow
+        @property bool empty() const
         {
             return _s.empty;
         }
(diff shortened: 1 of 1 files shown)


std/algorithm/iteration.d
```diff
@@ -5034,28 +5039,28 @@ if (isSomeChar!C)
         }
 
     public:
-        this(C[] s) pure @safe
+        this(Range s)
         {
-            import std.string : strip;
-            _s = s.strip();
+            import std.string : stripLeft;
+            _s = s.stripLeft();
```
schveiguy: Note, I realized that strip isn't necessary -- the `getFirst` function is going to skip any trailing whitespace anyway.

std/algorithm/iteration.d
```diff
@@ -5034,28 +5039,28 @@ if (isSomeChar!C)
...
             getFirst();
         }
 
-        @property C[] front() pure @safe
+        @property front()
```
wilzbach: Why did you remove the return type here?

schveiguy: It's `auto` return now. I could put `Range`, but that may be more confusing.

Plus, I know this is a template type, but not sure if attributes will be inferred if it's not `auto`, and I can't specify `pure @safe` anymore, due to the possibility of the underlying range not being one of those.

schveiguy: In any case, the type is voldemort, it has no docs except in the doc of the function. So including a type has no effect on documentation. One has to look at the implementation to see what it is returning exactly, and the answer is right below anyway.

schveiguy:
Hm... I just realized something -- this isn't going to treat unicode whitespace as whitespace, just ones that fit into `char`. This is due to a) the way `find` works, and b) questionable `char` promotion to `dchar`.

While I can probably cobble together something that properly fixes this, I could also just document the limitation. Seeing as it didn't compile before, it's not something that would break any code.

schveiguy:
> I could also just document the limitation

Thinking about this some more, I think this is the wrong approach. All the information is there to ascertain whether the item has unicode white space on the front. I will do what std.string.stripLeft does anyway.

Don't merge this, I'll work on getting it to work with unicode.

schveiguy:
OK, I've implemented the necessary code to make it work with unicode spaces. Trickier than I would like. `byCodeUnit` should really compare to a `string` without having to convert.

std/algorithm/iteration.d
schveiguy: More info -- this must be inferred, but I was wrong that it's not inferred without an `auto` return. Otherwise, the other functions wouldn't be inferred, and the `pure @safe` unittest I added below would fail to compile.

So I can add the return type back in, if that's what gets this merged.

wilzbach: I was just curious because I think dscanner has a rule (not enabled here) that enforces that the return type is either defined or auto.

schveiguy: Done.

std/algorithm/iteration.d
```diff
@@ -5001,61 +5001,107 @@ private struct SplitterResult(alias isTerminator, Range)
...
+                       is(ElementType!Range == immutable(dchar)))
+            {
+                // dchar range, we can just use find.
+                auto r = find!(isWhite)(_s.save);
+                _frontLength = _s.length - r.length;
+            }
+            else
+            {
+                // need to decode the characters until we find a space. This is
+                // ported from std.string.stripLeft.
```
n8sh: Is it really necessary to proliferate the evil that is auto-decoding?

>One of the worst things about autodecoding is it is special, it *only* steps in for strings. Fortunately, however, that specialness enabled us to save things with byCodePoint and byCodeUnit.

schveiguy: This overload of `splitter` is special. It splits on *runs* of whitespace, which isn't possible with any of the other overloads. It's also one of the more used functions. That it doesn't work with arbitrary ranges of char is an artificial limitation.

In order for `byCodeUnit` to be useful in all places `string`s are useful, it has to have similar support from Phobos.

Besides, in this case, you are requesting auto-decoding, because you want to split on unicode whitespace.

std/algorithm/iteration.d
n8sh: Likewise the `stripLeft` in `popFront`.

n8sh: Nevermind, you are correct.

std/algorithm/iteration.d
n8sh: If your custom type is intended to behave like a string then like a string its `front` should return an entire `dchar` and its `popFront` should remove an entire `dchar`.

>In order for `byCodeUnit` to be useful in all places strings are useful, it has to have similar support from Phobos.

Isn't the entire point of `byCodeUnit` that it is a non-auto-decoding type whose elements are individual `char`s and _not_ `dchar`s?

schveiguy: > Isn't the entire point of `byCodeUnit` that it is a non-auto-decoding type whose elements are individual `char`s and _not_ `dchar`s?

Yes, but that doesn't mean you can't use it to decode `dchar`s.

Consider that if you have a `byCodeUnit` range, `bcu = strip(bcu)` still works and strips unicode spaces. Why shouldn't `splitter` enjoy the same benefits?

Autodecoding doesn't mean NO decoding, it just means I have to opt in to use it. Using this form of `splitter` is opting in.

> If your custom type is intended to behave like a string then like a string its `front` should return an entire `dchar` and its `popFront` should remove an entire `dchar`.

It should be `string`-like in all respects _except_ this one. That is, I want to use it as an array, but treat it like a string for printing, processing, using with `strip`, `splitter`, etc.

schveiguy: > in this case, you are requesting auto-decoding

I said this wrong. In this case you are requesting _decoding_.

schveiguy: > What about giving it an inverse of `byCodeUnit` and explicitly opting in to decoding, like the reverse is done for strings?

No, because what I am ultimately developing (not here, but this is a requirement) is that `splitter` returns the *exact* range type I pass in. If I have to change to another range type, then I lose the extra information I've added to my specialized range.

It's also an odd extra step, when the guts of this function, `stripLeft`, work fine on a `char`-based range.

> If I want my range _not_ to be cross-element decoded what should I use to split on runs of consecutive whitespace?

I'm not understanding this. The elements returned by `splitter` will not turn into auto-decoded ranges, so how does this not satisfy your requirements?

std/algorithm/iteration.d
```diff
@@ -5001,61 +5001,107 @@ private struct SplitterResult(alias isTerminator, Range)
...
         {
             import std.uni : isWhite;
 
-            auto r = find!(isWhite)(_s);
-            _frontLength = _s.length - r.length;
+            static if (is(ElementType!Range == dchar) ||
+                       is(ElementType!Range == const(dchar)) ||
+                       is(ElementType!Range == immutable(dchar)))
+            {
+                // dchar range, we can just use find.
```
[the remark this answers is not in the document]
schveiguy: Hm... is there not a danger of new whitespace characters being added to the spec at a future date?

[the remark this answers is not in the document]
schveiguy: Note that stripLeft and stripRight only take a shortcut for ascii, otherwise it decodes.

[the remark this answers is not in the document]
n8sh: For new whitespace characters to be a problem they would have to be outside the basic multilingual plane. No whitespace has been added to Unicode since 2002, and no whitespace characters outside the "General Punctuation" / "Supplemental Punctuation" blocks have been added since 1999.

std/algorithm/iteration.d
n8sh: I don't mean the return type, I mean the performance cost.

std/algorithm/iteration.d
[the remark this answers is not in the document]
n8sh: >Note that `stripLeft` and `stripRight` only take a shortcut for ascii, otherwise it decodes.

The "shortcut" in `stripLeft` is a pessimization for auto-decoding types. It uses `.front` so for `string` and `wstring` it performs decoding once and if the result is non-ASCII it performs decoding a second time. That's among the things I fix in PR #6727.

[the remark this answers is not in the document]
schveiguy: Hm.. I get confused with all the template constraints. I thought `!isConvertibleToString` would disqualify normal string arrays. I actually don't know what it means when something passes or fails that check!

So I thought that overload was only for non-auto-decoding strings, which would actually not make it a pessimization.

I will review your other PR.

However, in this case, we are just using `find` on auto-decoding strings (i.e. element type is `dchar`), which should do the right thing.



> For new whitespace characters to be a problem they would have to be outside the basic multilingual plane. No whitespace has been added to Unicode since 2002, and no whitespace characters outside the "General Punctuation" / "Supplemental Punctuation" blocks have been added since 1999.

OK, good evidence that we can take the risk. I'll update for `wchar`.

std/algorithm/iteration.d
schveiguy: I suppose the right question to answer is, how do I check for whitespace using `std.uni.isWhite` without decoding?

n8sh: Since `std.string.stripLeft` works like this already when given non-string ranges I'm coming to believe this might be correct.

schveiguy:
Added special branch for `wstring` and also updated `dchar` branch to accept ranges of `wchar`.

std/algorithm/iteration.d
```diff
@@ -5083,7 +5143,7 @@ if (isSomeChar!C)
     static foreach (S; AliasSeq!(string, wstring, dstring))
     {{
         import std.conv : to;
-        S a = " a     bcd   ef gh ";
+        S a = " a  \u2028   bcd   ef gh ";
```
schveiguy: Added this to make sure branch of `wstring` catches unicode spaces.

PR #6701, opened 2018-09-10
Revert "Fix Issue 19236 - Replace runtime `typeid(T).initializer().ptr is null` checks with compile-time `__traits(isZeroInit, T)`"

ibuclaw:
Reverts dlang/phobos#6698

Compiler implementation was not tested properly


ibuclaw:
Regression tracked down.

PR #6703, opened 2018-09-10
Revert "In std.conv.emplaceInitializer use memset instead of memcpy when setting all 0s"

ibuclaw:
Reverts dlang/phobos#6461

This was merged too eagerly, and the compiler implementation was not tested against other back-ends.


ibuclaw:
Regression tracked down.

PR #6704, opened 2018-09-10
std.typecons: Unittest isn't safe

ibuclaw:
It takes the address of local variables.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -3390,7 +3390,7 @@ auto nullable(T)(T t)
 }
 
 // check that toHash on Nullable is forwarded to the contained type
-@safe unittest
+@system unittest
 {
     struct S
     {


Geod24:
How did it compile before ?

ZombineDev:
Taking the address is `@safe`, as long as you don't escape it under `-dip1000`. @ibuclaw my guess is that you and the @ldc-developers didn't see this file: https://github.com/dlang/phobos/blob/0ac79f5a8214333dc58ec9774f386efda9de3fe3/dip1000.mak#L49

ZombineDev:
@ibuclaw the reason for this mess is that we can't fix everything preventing the use of `-dip1000` all at once in phobos, so we decided to go with incremental approach until all modules compile. After we're done, we'll switch to all in.
Side note: this approach is not `-dip1000` specific - most large scale language changes would need to follow the same incremental tactic. So in the future we need to better collaborate on such initiatives.

ibuclaw:
@wilzbach then get the library building entirely with dip1000 then.

This file is completely unworkable.
https://github.com/dlang/phobos/blob/0ac79f5a8214333dc58ec9774f386efda9de3fe3/dip1000.mak

If I read the makefile right, you're building release library without dip1000 anyway, so it's not a security issue.

ZombineDev:
@wilzbach as mentioned before, we care about the public symbols' safety, not about some random code snippets in our unittests. We use unittest to guarantee the safety our public symbols, but in this case the safety aspect of the test in question had nothing to do with the public symbol.

PR #6707, opened 2018-09-15
Added an enum for file attributes on Windows

bausshf:
Also updated relevant documentation for "remove" which may indirectly be tied to "setAttributes" and the "FileAttribute" enum.

This will help people who get "misleading" error messages on Windows when ex. removing a file, because the documentation now specifies what you can do etc.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -48,6 +48,7 @@ $(TR $(TD Attributes) $(TD
           $(LREF getLinkAttributes)
           $(LREF getSize)
           $(LREF setAttributes)
+          $(LREF FileAttribute)
 ))
 $(TR $(TD Timestamp) $(TD
           $(LREF getTimes)
@@ -1008,8 +1009,11 @@ Delete file `name`.
 Params:
     name = string or range of characters representing the file _name
 
-Throws: $(LREF FileException) on error.
- */
+Throws:
+    $(LREF FileException) on error.
+
+    On Windows it may also give an "Access denied" error, which may require you to call $(LREF setAttributes) beforehand with $(LREF FileAttribute).normal
+*/
 void remove(R)(R name)
 if (isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R) &&
     !isConvertibleToString!R)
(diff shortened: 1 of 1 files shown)


std/file.d
```diff
@@ -1008,8 +1009,11 @@ Delete file `name`.
 Params:
     name = string or range of characters representing the file _name
 
-Throws: $(LREF FileException) on error.
- */
+Throws:
+    $(LREF FileException) on error.
+
+    On Windows it may also give an "Access denied" error, which may require you to call $(LREF setAttributes) beforehand with $(LREF FileAttribute).normal
+*/
```
CyberShadow: Please do not mix unrelated changes in the same commit / pull request. Remove this change from this pull request, and add it to a separate one.

std/file.d
```diff
@@ -2304,6 +2308,39 @@ if (isConvertibleToString!R)
     }
 }
 
+version (Windows)
+{
+    /// File attributes that can be used on Windows usng ex. $(LREF setAttributes)
```
CyberShadow: "usng"

std/file.d
```diff
@@ -2304,6 +2308,39 @@ if (isConvertibleToString!R)
...
+    enum FileAttribute
+    {
+      /// A file or directory that is an archive file or directory. Applications typically use this attribute to mark files for backup or removal.
+      archive = 0x20,
```
CyberShadow: Why hard-code these constants instead of reusing the ones from `core.sys.windows`?

std/file.d
```diff
@@ -2304,6 +2308,39 @@ if (isConvertibleToString!R)
     }
 }
 
+version (Windows)
+{
+    /// File attributes that can be used on Windows usng ex. $(LREF setAttributes)
+    enum FileAttribute
+    {
+      /// A file or directory that is an archive file or directory. Applications typically use this attribute to mark files for backup or removal.
```
CyberShadow: Phobos style defines indentation to be 4 spaces.

std/file.d
```diff
@@ -2304,6 +2308,39 @@ if (isConvertibleToString!R)
     }
 }
 
+version (Windows)
```
CyberShadow: Why only define the Windows attributes?

CyberShadow: Windows-only symbol documentation needs to be in a `version(StdDdoc)`, otherwise it won't be visible on dlang.org.

RazvanN7:
ping @bausshf

PR #6708, opened 2018-09-15
std.math: Remove support for IBM extended reals

ibuclaw:
POWER9 supports 128-bit IEEE, and the old long double format is in the middle of being be phased out in glibc.

We have no desire to support legacy, especially as our support is incomplete.  Better to just drop this type and only focus on completing 128-bit IEEE support.

Best reviewed ignoring whitespace changes.

https://github.com/dlang/phobos/pull/6708/files?w=1

FYI @redstar.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -240,10 +240,8 @@ version(LittleEndian)
 }
 else
 {
-    static assert(real.mant_dig == 53 || real.mant_dig == 106
-               || real.mant_dig == 113,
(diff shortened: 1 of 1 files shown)


redstar:
Obviously we struggled a long time to implement this. And I also saw that the glibc plan is to switch to IEEE doubles. So I can support Iain's view: let's remove the legacy and focus on 128-bit IEEE doubles (which are used on other platforms, too).

PR #6711, opened 2018-09-17
Fix issue 19257: std.array.join: only remove const if you can actually assign const to non-const values

FeepingCreature:
For instance:

    const Class[][] array;

`array.join` will try to `Unqual` the `const Class`, then build a `Class[]` to return. But since `Class` is a reference type, you cannot assign `const Class` to `Class`, and `join` fails.

Only use `Unqual` if `Unqual!T` is assignable `T`; otherwise return `T[]` instead.

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -1917,7 +1917,15 @@ if (isInputRange!RoR &&
     isInputRange!(Unqual!(ElementType!RoR)))
 {
     alias RetType = typeof(return);
-    alias RetTypeElement = Unqual!(ElementEncodingType!RetType);
+    alias ConstRetTypeElement = ElementEncodingType!RetType;
+    static if (isAssignable!(Unqual!ConstRetTypeElement, ConstRetTypeElement))
+    {
+        alias RetTypeElement = Unqual!ConstRetTypeElement;
+    }
+    else
+    {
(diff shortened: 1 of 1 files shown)


BBasile:
Open a bugzilla issue for the problem you fix, reference it in the git commit, as usual.
Also since you target stable, is it a regression ?
Please try to take care to the formal aspect of the PR; since you're not a first time committer you should know that.

FeepingCreature:
I didn't realize that features could not be added in stable..?

wilzbach:
> How exactly do you expect people to know these things...? 

It's here (https://wiki.dlang.org/Starting_as_a_Contributor), but typically dlang-bot will also warn you (and provide instructions) if you would have linked your PR with a regression,

> Also that features needed a bug report?

Just have a look at dlang-bot in front of you:

image

wilzbach:
> Under semver, minor versions can add functionality.

While we don't 100% adhere to SemVer, stable would be the "patch" branch which thus typically only contains regressions and other important bug fixes whereas minor versions are more or less new D releases and are developed on master.

FeepingCreature:
Again sorry for the review spam, I don't know a way to change base without it.

FeepingCreature:
ping

FeepingCreature:
ping

PR #6712, opened 2018-09-18
Fix issue 19240 - std.typecons.Tuple should check for reserved identi…

Biotronic:
…fiers


wilzbach:
CI fails due to this line:

https://github.com/dlang/tools/blob/c3ee76ed0ac34a3d1bfc301480a6b25ee17fd331/changed.d#L431

```
changed.d(431): Error: template instance `std.typecons.Tuple!(string, "name", string, "headline", string, "path")` does not match template declaration `Tuple(Specs...) if (distinctFieldNames!Specs)`
```

PR #6716, opened 2018-09-21
Fix issue 19256: prevent JSONValue const conversion

FeepingCreature:
Fix https://issues.dlang.org/show_bug.cgi?id=19256
Reorder JSONValue's union members to make dmd realize that JSONValue contains const references and disallows reassigning const(JSONValue) to JSONValue, which would break constness.
This works around https://issues.dlang.org/show_bug.cgi?id=12885 .

Sample code illustrating the problem:

```
    const JSONValue innerObj = JSONValue(["foo": JSONValue(1)]);

    assert(innerObj["foo"] == JSONValue(1));

    // Why can I do this??
    JSONValue value = innerObj;

    value["foo"] = JSONValue(2);

    assert(innerObj["foo"] == JSONValue(1)); // fails
```

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -119,12 +119,12 @@ struct JSONValue
 
     union Store
     {
+        JSONValue[string]               object;
+        JSONValue[]                     array;
         string                          str;
         long                            integer;
         ulong                           uinteger;
         double                          floating;
-        JSONValue[string]               object;
-        JSONValue[]                     array;
     }
     private Store store;
     private JSONType type_tag;
@@ -772,6 +772,26 @@ struct JSONValue
     string toPrettyString(in JSONOptions options = JSONOptions.none) const @safe
     {
         return toJSON(this, true, options);
+    }
+}
+
+// regression test for issue 19256
+@safe unittest
+{
+    const JSONValue object = JSONValue(["foo": JSONValue(1)]);
+
+    // Usually, this should fail, as object is const.
+    static if (__traits(compiles, { JSONValue copy = object; }))
+    {
+        // Even if it works, for instance because JSONValue has gained a copy constructor,
+        // we should be fine as long as the constructor actually copies the contained references.
+        JSONValue copy = object;
+
+        // However, with issue 19256 we are able to affect the const original through the mutable copy.
+        copy["foo"] = JSONValue(2);
+
+        // Check for that.
+        assert(object["foo"] == JSONValue(1));
     }
 }
 


CyberShadow:
Test, please? You can use `!__traits(compiles, ...)` to test for accepts-invalid bugs.

FeepingCreature:
Test added.

FeepingCreature:
Regarding auto-tester failures, see https://github.com/dlang/druntime/pull/2304

FeepingCreature:
druntime PR merged, retrying.

BBasile:
About buildkite i don't know if you have seen but this PR breaks a non trivial amount of projects (vibe, dls, dub, ...). Ask for an invite to wilzbach or Martin Nowak or andralex in order to see the logs.

const_json_projs

FeepingCreature:
I am not sure that things *can* be done. D allows you to write straight-up _wrong_ code involving JSONValue right now. Presumably that's what they're doing.

BBasile:
Contacting people so that they change their code is what i had in mind.

BBasile:
Almost. After looking more carefully at the logs it's always vibe-d (but several places) excepted for the project dls.

CyberShadow:
Perhaps leaving this to a compiler fix would be preferred after all, as then it could go through a deprecation phase.

@WalterBright How likely are we to get a fix for https://issues.dlang.org/show_bug.cgi?id=12885 in DMD?

FeepingCreature:
Ping? "Not very likely"?

MartinNowak:
> @WalterBright How likely are we to get a fix for https://issues.dlang.org/show_bug.cgi?id=12885 in DMD?

I'd say we should fix that issue (https://issues.dlang.org/show_bug.cgi?id=12885#c3), though it will take some deprecation period. So this looks like a reasonable workaround in the meantime.

MartinNowak:
Build failure of vibe.d (https://buildkite.com/dlang/phobos/builds/397#dccb2203-9142-4358-9840-e650356a4ccb/99-140).

<details>
<summary>Error</summary>
<pre>
vibe-core-1.4.3/vibe-core/source/vibe/core/args.d(68,27): Error: function `vibe.core.args.fromValue!bool.fromValue(JSONValue val)` is not callable using argument types `(const(JSONValue))`
vibe-core-1.4.3/vibe-core/source/vibe/core/args.d(68,27):        cannot pass argument `*pv` of type `const(JSONValue)` to parameter `JSONValue val`
vibe-core-1.4.3/vibe-core/source/vibe/core/args.d(298,12): Error: template instance `vibe.core.args.readOption!bool` error instantiating
vibe-core-1.4.3/vibe-core/source/vibe/core/core.d(692,70): Deprecation: `std.algorithm.setops.No` is not visible from module `core`
vibe-core-1.4.3/vibe-core/source/vibe/core/args.d(68,27): Error: function `vibe.core.args.fromValue!string.fromValue(JSONValue val)` is not callable using argument types `(const(JSONValue))`
vibe-core-1.4.3/vibe-core/source/vibe/core/args.d(68,27):        cannot pass argument `*pv` of type `const(JSONValue)` to parameter `JSONValue val`
vibe-core-1.4.3/vibe-core/source/vibe/core/core.d(1328,13): Error: template instance `vibe.core.args.readOption!string` error instantiating
</pre>
</details>

MartinNowak:
Here is a trivial PR to vibe.d https://github.com/vibe-d/vibe-core/pull/100, seems like a rare occurrence of sth. that breaks with this workaround, so it might still be reasonable to go ahead with this once a new vibe.d version is released.

MartinNowak:
Should work now with the fixed vibe-d and good to go for 2.085.0 from my side. Seemed to risky to put in 2.084.0 without beta.

FeepingCreature:
ping

thewilsonator:
Ack, waiting on https://github.com/dlang/phobos/pull/7011 for green tests.

thewilsonator:
Can you please rebase this so the circle error goes away?

thewilsonator:
Hmm vibe.d still fails.

FeepingCreature:
I don't follow why. *Should* D take the const-convertible-ness of a union from the first field? I guess not, but it works today, and then JSONValue behaves correctly.

schveiguy:
> I'd rather see it codified as "well this sucks, but it's the best we can do."

It's not memory safe in `@safe` code. If D is no longer serious about memory safety, then we can leave it. I guess another option is to disable using such types in `@safe` code.

> SumType can "fix" this by just reordering its union by `hasIndirections`. It's not a good solution but it is _a_ solution.

SumType will suffer exactly the same fate as `JSONValue` the more it is used. Eventually there will be critical projects that will break with your suggested change because they accidentally implicitly cast away const, and it will become untouchable.

And there is no way to deprecate it, unless we want to rename the type.

The only path forward is to deprecate the current memory-unsafe behavior, and then fix it.

> [librebindable](https://code.dlang.org/packages/rebindable) uses a totally different approach (involving a lot more casting).

Rebindable in general is a horrendous hack, making up for a lack of tail-modifiers in the language. I'm not convinced it's sound type-wise.

FeepingCreature:
I mean, if the bug is fixed, JSONValue and SumType will break regardless, surely?

> Rebindable in general is a horrendous hack, making up for a lack of tail-modifiers in the language. I'm not convinced it's sound type-wise.

I believe that librebindable is typesafe, or rather, can be used to build typesafe abstractions if used correctly. Phobos Rebindable definitely wouldn't be typesafe, especially if it worked for structs, because it exposes its value by ref, leading to pointer leaks that allow observing immutable changes. My primary argument for librebindable being sound is that it does not do this.

schveiguy:
> I mean, if the bug is fixed, JSONValue and SumType will break regardless, surely?

Yes, but the big difference is that the compiler can deprecate the behavior first, notifying anyone using it for a (probably long-ish) period of time. A library cannot hook the implicit cast.

schveiguy:
> I believe that librebindable is typesafe, or rather, can be used to build typesafe abstractions if used correctly.

I honestly have not read much into it. The *idea* of deconstifying copied data is sound (i.e. deconstifying the head), but what I'm more worried about is if it's sound *within the current model*. That is, will amazingly obtuse optimizations suddenly happen because they are sound within the model, but not with the hacks performed here?

FeepingCreature:
> Yes, but the big difference is that the compiler can deprecate the behavior first, notifying anyone using it for a (probably long-ish) period of time. A library cannot hook the implicit cast.

Ah, now I get what you're saying. Yeah if there's any hope of the bug being fixed, I agree that's better than this PR.

LightBender:
Underlying issue has been fixed. You can reopen if it is still relevant. Closing this PR.

FeepingCreature:
Underlying issue has not been fixed?

thewilsonator:
https://issues.dlang.org/show_bug.cgi?id=12885 was allegedly fixed by https://github.com/dlang/dmd/pull/16594 
Was that an incomplete fix?

FeepingCreature:
Oh! You are correct. It's gated behind `-preview=fixImmutableConv`. I'll add a comment to 19256.

thewilsonator:
Thanks for the investigation!

PR #6720, opened 2018-09-26
fix issue 19208 - swapEndian doesn't support floating point types

BBasile:
- Refactor existing union with a more elegant static expression.
- Set the function parameter `const`  because parameter is always a value and this can reduce how many instances are generated.

### DIFF
--- a/std/bitmanip.d
+++ b/std/bitmanip.d
@@ -2710,7 +2710,8 @@ public:
     Swaps the endianness of the given integral, float or character.
   +/
 T swapEndian(T)(const T val) @safe pure nothrow @nogc
-if (canSwapEndianness!T)
+if (isIntegral!T || isSomeChar!T || isBoolean!T ||
+    (isFloatOrDouble!T && size_t.sizeof == 8))
 {
     static if (val.sizeof == 1)
         return val;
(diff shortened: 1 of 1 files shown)


BBasile:
buildkite seems to run the `Style` check using an older DMD, leading to failures : 

> std/conv.d(4369): Error: unrecognized trait `isZeroInit`

ping @wilzbach.

std/bitmanip.d
```diff
@@ -2707,17 +2707,24 @@ public:
 }
 
 /++
-    Swaps the endianness of the given integral value or character.
+    Swaps the endianness of the given integral, float or character.
   +/
-T swapEndian(T)(T val) @safe pure nothrow @nogc
-if (isIntegral!T || isSomeChar!T || isBoolean!T)
+T swapEndian(T)(const T val) @safe pure nothrow @nogc
+if (canSwapEndianness!T)
```
n8sh: It appears that return type `T` may not work properly for floats. From the documentation of `bigEndianToNative`:
```d
/++
...
    Taking a `ubyte[n]` helps prevent accidentally using a swapped value
    as a regular one (and in the case of floating point values, it's necessary,
    because the FPU will mess up any swapped floating point values. So, you
    can't actually have swapped floating point values as floating point values).
+/
```

To really test if this PR works, instead of testing `swapEndian(0.5f).swapEndian() == 0.5f` perhaps check `swapEndian(0x1.00007ep+0f) == 0x1.01007ep-1f`. Although it should be asked how exact the FPU was expected "mess up" swapped values so it can be verified this doesn't happen.

BBasile: Yes indeed. I did that because it's hard to get two exact representations of floats when swapped without cheating.

BBasile: swapping a valid float value that leads to NaN is out of topic i think, even if a true topic. The operation might be valid for serialization purpose.

klickverbot: I can't see how it is valid – just use a ubyte buffer on the serialised side.

BBasile: It depends on the serializer.

n8sh: If deserializing a messed-up float can result in it getting changed, then this doesn't work for either serialization or deserialization.

BBasile: Let's stop speaking about serialization, i don't know why i put this off topic point on the table. I'll just change the test, as initially suggested.

n8sh: What I think we're driving at is - under what circumstance would this function be useful? All it seems like now is a trap that appears useful for a certain purpose but will sometimes give garbage output.

BBasile: About your first comment @n8sh : I've added an overload for x86 that doesn't pass the parameter in ST(0).

PR #6722, opened 2018-09-29
Fix Issue 19275 - std.process: redirecting output in a non-console application fails

rainers:
MS runtime up to VS2013 used pseudo handle _NO_CONSOLE_FILENO in that case, see https://github.com/shihyu/learn_c/blob/master/vc_lib_src/src/ioinit.c#L285.

No idea how to test a non-console application in phobos.

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -706,12 +706,13 @@ private Pid spawnProcessImpl(scope const(char)[] commandLine,
     static void prepareStream(ref File file, DWORD stdHandle, string which,
                               out int fileDescriptor, out HANDLE handle)
     {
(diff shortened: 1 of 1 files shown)


rainers:
> If this fixes the linked bug report, please prepend "Fix " to the commit message.

Oops, I somehow omitted it. Changed now.

> There are some stand-alone tests in Phobos and Druntime (see test directory)

I don't see anything from the phobos/test folder running on the auto-tester. I could mimick druntime, but I'm not sure it is easily done as the build-server would also run the test as a console application with redirection.

PR #6723, opened 2018-09-30
Fix Issue 19289 - std.range.transposed with enforceNotJagged not throwing

crocopaw:
see https://forum.dlang.org/post/tmqppgmuevecltmxygrd@forum.dlang.org

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -7219,6 +7219,17 @@ if (isForwardRange!RangeOfRanges &&
     this(RangeOfRanges input)
     {
         this._input = input;
+        static if (opt == TransverseOptions.enforceNotJagged)
+        {
+            import std.exception : enforce;
+
+            if (empty) return;
+            immutable commonLength = _input.front.length;
+            foreach (e; _input)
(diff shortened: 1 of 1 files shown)


std/range/package.d
```diff
@@ -7219,6 +7219,17 @@ if (isForwardRange!RangeOfRanges &&
     this(RangeOfRanges input)
     {
         this._input = input;
+        static if (opt == TransverseOptions.enforceNotJagged)
+        {
+            import std.exception : enforce;
+
+            if (empty) return;
+            immutable commonLength = _input.front.length;
+            foreach (e; _input)
```
n8sh: `_input.front`'s length is unnecessarily enforced to be equal to itself.

n8sh:
Linux_64_32 failure appears unrelated.

std/range/package.d
crocopaw: I just copied the code from transverse. If you've got an idea how to remove this redundancy you should do this at transverse as well.

PR #6726, opened 2018-10-10
Fix issue 19297: handle comparison of signed, unsigned and floating JSONValues as in D

FeepingCreature:
JSON does not have any notion of multiple numeric types. (JS only uses double.)

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -629,27 +629,64 @@ struct JSONValue
         // Default doesn't work well since store is a union.  Compare only
         // what should be in store.
         // This is @trusted to remain nogc, nothrow, fast, and usable from @safe code.
-        if (type_tag != rhs.type_tag) return false;
 
         final switch (type_tag)
         {
-        case JSONType.string:
-            return store.str == rhs.store.str;
         case JSONType.integer:
-            return store.integer == rhs.store.integer;
+            switch (rhs.type_tag)
+            {
+                case JSONType.integer:
+                    return store.integer == rhs.store.integer;
+                case JSONType.uinteger:
+                    return store.integer == rhs.store.uinteger;
+                case JSONType.float_:
+                    return store.integer == rhs.store.floating;
+                default:
+                    return false;
+            }
         case JSONType.uinteger:
-            return store.uinteger == rhs.store.uinteger;
+            switch (rhs.type_tag)
+            {
+                case JSONType.integer:
+                    return store.uinteger == rhs.store.integer;
+                case JSONType.uinteger:
+                    return store.uinteger == rhs.store.uinteger;
+                case JSONType.float_:
+                    return store.uinteger == rhs.store.floating;
+                default:
+                    return false;
+            }
         case JSONType.float_:
-            return store.floating == rhs.store.floating;
+            switch (rhs.type_tag)
+            {
+                case JSONType.integer:
+                    return store.floating == rhs.store.integer;
+                case JSONType.uinteger:
+                    return store.floating == rhs.store.uinteger;
+                case JSONType.float_:
+                    return store.floating == rhs.store.floating;
+                default:
+                    return false;
+            }
+        case JSONType.string:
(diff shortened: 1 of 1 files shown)


n8sh:
The desired change in issue #19297 is that unsigned integers and signed integers with the same value should be considered equal. However, the change as written also does two other things not mentioned by issue #19297:

1. Each negative signed integer is equal to a corresponding large unsigned integer (just as with standard D comparison of `long` and `ulong`). Equality of negative and non-negative integers may not be what the PR intended. Combined with this PR making it possible for an integer/uinteger to equal a floating point value this also breaks the transitive property of equality.
2. Integers are truncated to `double.mant_dig` bits of precision before comparison to floating point values (just as with standard D comparison of `long` and `double`). This breaks the transitive property of equality since two large integers not equal two each other may be equal to the same floating point value.

Additionally, changing `==` to numeric equality as proposed in issue #19297 has the inevitable consequences that:

3. `a == b` no longer implies `a.type == b.type`. 
4. Two `JSONValue`s may be equal but a property (`integer`/`uinteger`/`floating`) that is accessible on one may throw an Exception on the other.

I would like to hear your thoughts on whether these are okay.

(Note: since JSONValue already isn't usable as a key in an associative array this PR doesn't need to add a `toHash` consistent with the changes to `opEquals`, but otherwise it would.)

FeepingCreature:
`a.type` is broken anyhow. JSON has no concept of `uinteger` to begin with. So I'm inclined to not care about that.

Good point about very large integers no longer preserving transitivity of equality. I'm open to solutions! Maybe integers not representable as double could compare false to double JSONValues? Similarly with signed and unsigned - types should only compare equal if they're within each other's intersection of domains.

n8sh:
@FeepingCreature That sounds reasonable.

CyberShadow:
I don't think it's possible to unequivocally define "equality". Some goals that seem sensible would be to do what D's `==` does, or what JavaScript's `==` or `===` do.

* JavaScript's `==` evaluates to `true` for comparisons like `1 == "1"`, so that seems out-of-scope for `std.json`.
* JavaScript's `===` is quite similar to D's `is`, except `is` will return false for different references to the same data. The behavior before this PR is close to `===`, however as has been mentioned, JS has no notion of integer types, so it's not possible to compare them adequately.
* D's `==` is close to the behavior of std.json with this PR, with the added property that type errors now result in `false`. Notably, `{ auto i = uniform!uint; assert(int(i) == uint(i)); }` will always hold in D, so I think that's a good goal. Equality of integers of different value ranges is part of a larger problem including different widths, ordered comparison, checked conversion and overflows, and seems like a problem better left to e.g. `std.checkedint` or `std.conv`.

FeepingCreature:
It was too ugly. Reduced the PR to just signed and unsigned ("toString equality").

FeepingCreature:
No. I'm going back to my initial approach of comparing signed, unsigned and floating directly.

@n8sh I agree with all your issues; however, they are already issues with D. It's shit that JSONValue.integer/uinteger/floating may crash for a JSONValue that compares equal to one where they don't crash, but - who does that anyways? That aside: D has made the decision that different numeric types are comparable. D has made the decision that `-1L == ulong.max`. D has made the decision that `0 == 0.0`. We may not like these decisions, but as it stands they're the "law of the land", and imo it is more important in this to be internally consistent than "correct" by some subjective numerical standard.

n8sh:
@FeepingCreature Reasonable. I'd say the greatest priority is for a `JSONValue` to behave like one would expect of a JSON value, and your PR brings things closer to that.

>who does that anyways?

Now that you mention it, what use case motivated this PR to begin with?

FeepingCreature:
> Now that you mention it, what use case motivated this PR to begin with?

JSON serialization from a type with `ulong` fields, unittested and compared with `JSONValue`s constructed with integer literals.

It took me a while to realize why `JSONValue(5678)` was not equal to `JSONValue(5678)`.

FeepingCreature:
ping

PR #6727, opened 2018-10-10
Add ASCII fast path to stripLeft & avoid unnecessary decoding

n8sh:
Similar to https://github.com/dlang/phobos/pull/6361. Code for benchmarks in next post.

**string benchmark, ldc2:**

| compiler                | function      | data      | time                                |
|-------------------------|---------------|-----------|-------------------------------------|
| ldc2 -O2 -release       | old_stripLeft | ascii     | 1 sec, 65 ms, 71 μs, and 8 hnsecs   |
| ldc2 -O2 -release       | new_stripLeft | ascii     | 492 ms, 570 μs, and 2 hnsecs        |
| ldc2 -O2 -release       | old_stripLeft | non_ascii | 1 sec, 852 ms, and 350 μs           |
| ldc2 -O2 -release       | new_stripLeft | non_ascii | 1 sec, 209 ms, 13 μs, and 6 hnsecs  |

**wstring benchmark, ldc2:**

| compiler                | function      | data        | time                                |
|-------------------------|---------------|-------------|-------------------------------------|
| ldc2 -O2 -release       | old_stripLeft | w_ascii     | 1 sec, 165 ms, 298 μs, and 7 hnsecs |
| ldc2 -O2 -release       | new_stripLeft | w_ascii     | 780 ms, 142 μs, and 9 hnsecs        |
| ldc2 -O2 -release       | old_stripLeft | w_non_ascii | 864 ms, 82 μs, and 6 hnsecs         |
| ldc2 -O2 -release       | new_stripLeft | w_non_ascii | 591 ms, 621 μs, and 9 hnsecs        |

**string benchmark, dmd:**

| compiler                | function      | data      | time                                |
|-------------------------|---------------|-----------|-------------------------------------|
| dmd -O -inline -release | old_stripLeft | ascii     | 1 sec, 287 ms, 162 μs, and 2 hnsecs |
| dmd -O -inline -release | new_stripLeft | ascii     | 650 ms, 753 μs, and 4 hnsecs        |
| dmd -O -inline -release | old_stripLeft | non_ascii | 2 secs, 139 ms, 37 μs, and 1 hnsec  |
| dmd -O -inline -release | new_stripLeft | non_ascii | 1 sec, 738 ms, 224 μs, and 4 hnsecs |

**wstring benchmark, dmd:**

| compiler                | function      | data      | time                                |
|-------------------------|---------------|-----------|-------------------------------------|
| dmd -O -inline -release | old_stripLeft | w_ascii     | 1 sec, 649 ms, 661 μs, and 2 hnsecs |
| dmd -O -inline -release | new_stripLeft | w_ascii     | 1 sec, 231 ms, 832 μs, and 5 hnsecs |
| dmd -O -inline -release | old_stripLeft | w_non_ascii | 1 sec, 674 ms, 585 μs, and 8 hnsecs |
| dmd -O -inline -release | new_stripLeft | w_non_ascii | 967 ms, 158 μs, and 2 hnsecs        |

### DIFF
--- a/std/string.d
+++ b/std/string.d
@@ -2937,26 +2937,90 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
 {
     static import std.ascii;
     static import std.uni;
-    import std.utf : decodeFront;
 
-    while (!input.empty)
+    static if (is(Unqual!(ElementEncodingType!Range) == dchar)
+        || is(Unqual!(ElementEncodingType!Range) == wchar))
     {
-        auto c = input.front;
-        if (std.ascii.isASCII(c))
+        // Decoding is never needed for dchar. It happens not to be needed
+        // here for wchar because no whitepace is outside the basic
+        // multilingual plane meaning every whitespace character is encoded
+        // with a single wchar and due to the design of UTF-16 those wchars
+        // will not occur as part of the encoding of multi-wchar codepoints.
+        static if (isDynamicArray!Range)
         {
-            if (!std.ascii.isWhite(c))
-                break;
-            input.popFront();
+            foreach (i; 0 .. input.length)
+            {
+                if (!std.uni.isWhite(input[i]))
+                    return input[i .. $];
+            }
+            return input[$ .. $];
         }
         else
         {
-            auto save = input.save;
-            auto dc = decodeFront(input);
-            if (!std.uni.isWhite(dc))
-                return save;
+            while (!input.empty)
+            {
+                if (!std.uni.isWhite(input.front))
+                    break;
+                input.popFront();
+            }
+            return input;
         }
     }
-    return input;
+    else
+    {
+        static if (isDynamicArray!Range)
+        {
+            // ASCII optimization for dynamic arrays.
+            size_t i = 0;
+            for (const size_t end = input.length; i < end; ++i)
+            {
+                auto c = input[i];
+                if (c >= 0x80) goto NonAsciiPath;
+                if (!std.ascii.isWhite(c)) break;
+            }
+            input = input[i .. $];
+            return input;
+
+        NonAsciiPath:
+            input = input[i .. $];
+            // Fall through to standard case.
+        }
+
+        static if (ElementType!Range.sizeof > ElementEncodingType!Range.sizeof)
+        {
+            // Type performs its own decoding.
+            while (!input.empty)
+            {
+                if (!std.uni.isWhite(input.front))
+                    break;
+                input.popFront();
+            }
+            return input;
+        }
+        else
+        {
+            // Type doesn't perform its own decoding.
+            import std.utf : decodeFront, UseReplacementDchar;
+            while (!input.empty)
+            {
+                auto c = input.front;
+                if (std.ascii.isASCII(c))
+                {
+                    if (!std.ascii.isWhite(c))
+                        break;
+                    input.popFront();
+                }
+                else
+                {
+                    auto save = input.save;
+                    auto dc = decodeFront!(UseReplacementDchar.yes)(input);
+                    if (!std.uni.isWhite(dc))
+                        return save;
+                }
+            }
+            return input;
+        }
+    }
 }
 
 ///
@@ -2967,6 +3031,8 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
            "hello world     ");
     assert(stripLeft("\n\t\v\rhello world\n\t\v\r") ==
            "hello world\n\t\v\r");
+    assert(stripLeft(" \u2028hello world") ==
+           "hello world");
     assert(stripLeft("hello world") ==
            "hello world");
     assert(stripLeft([lineSep] ~ "hello world" ~ lineSep) ==
@@ -2978,6 +3044,8 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
     import std.utf : byChar;
     assert(stripLeft("     hello world     "w.byChar).array ==
            "hello world     ");
+    assert(stripLeft("     \u2022hello world     ".byChar).array ==
+           "\u2022hello world     ");
 }
 
 auto stripLeft(Range)(auto ref Range str)


n8sh:
Benchmark using `ldc2 -O2 -release`:
```
old_stripLeft ascii 904 ms, 698 μs, and 9 hnsecs
new_stripLeft ascii 566 ms, 287 μs, and 8 hnsecs
old_stripLeft non_ascii 861 ms, 773 μs, and 3 hnsecs
new_stripLeft non_ascii 850 ms, 440 μs, and 9 hnsecs
```

Benchmark using `dmd -O -inline -release`:
```
old_stripLeft ascii 1 sec, 264 ms, 510 μs, and 4 hnsecs
new_stripLeft ascii 600 ms, 595 μs, and 1 hnsec
old_stripLeft non_ascii 1 sec, 156 ms, 651 μs, and 7 hnsecs
new_stripLeft non_ascii 1 sec, 215 ms, 262 μs, and 4 hnsecs
```

Benchmark code:

<details>
<summary>stripBench.d</summary>

```d
//1st module
import std.traits;
import std.range;

auto old_stripLeft(Range)(Range input)
if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
    !isInfinite!Range && !isConvertibleToString!Range)
{
    static import std.ascii;
    static import std.uni;
    import std.utf : decodeFront;

    while (!input.empty)
    {
        auto c = input.front;
        if (std.ascii.isASCII(c))
        {
            if (!std.ascii.isWhite(c))
                break;
            input.popFront();
        }
        else
        {
            auto save = input.save;
            auto dc = decodeFront(input);
            if (!std.uni.isWhite(dc))
                return save;
        }
    }
    return input;
}

auto new_stripLeft(Range)(Range input)
if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
    !isInfinite!Range && !isConvertibleToString!Range)
{
    static import std.ascii;
    static import std.uni;
    import std.utf : decodeFront;

    static if ((isDynamicArray!Range || (isRandomAccessRange!Range && hasLength!Range))
        && __traits(compiles,
            {
                size_t s = input.length / 2;
                input = input[s .. $];
            }))
    {{
        // ASCII optimization for dynamic arrays & similar.
        size_t i = 0;
        for (const size_t end = input.length; i < end; ++i)
        {
            auto c = input[i];
            if (c >= 0x80) goto NonAsciiPath;
            if (!std.ascii.isWhite(c)) break;
        }
        input = input[i .. $];
        return input;

    NonAsciiPath:
        input = input[i .. $];
        // Fall through to standard case.
    }}

    while (!input.empty)
    {
        auto c = input.front;
        if (std.ascii.isASCII(c))
        {
            if (!std.ascii.isWhite(c))
                break;
            input.popFront();
        }
        else
        {
            auto save = input.save;
            auto dc = decodeFront(input);
            if (!std.uni.isWhite(dc))
                return save;
        }
    }
    return input;
}

void main(string[] args)
{
    import core.stdc.string;
    import std.stdio;
    import bench_data;

    auto timings = benchmark!(
        ()
        {
            blackhole += old_stripLeft(ascii_1)[0];
        },
        ()
        {
            blackhole += new_stripLeft(ascii_1)[0];
        },
        ()
        {
            blackhole += old_stripLeft(non_ascii_1)[0];
        },
        ()
        {
            blackhole += new_stripLeft(non_ascii_1)[0];
        },
    )(50_000_000);
    writefln("old_stripLeft ascii %s", timings[0]);
    writefln("new_stripLeft ascii %s", timings[1]);
    writefln("old_stripLeft non_ascii %s", timings[2]);
    writefln("new_stripLeft non_ascii %s", timings[3]);
 }
```
</details>


<details>
<summary>bench_data.d</summary>

```d
// 2nd module to hide data from optimizer
module bench_data;

immutable string ascii_1 = "\n\t\v\rhello world\n\t\v\r";
immutable string non_ascii_1 = "\u2020\u2028hello world\u2020\u2028";

int blackhole;
```
</details>

burner:
could you add a testcase that combines an ascii and non-ascii whitespace, so that the break out statement are correct.

n8sh:
@burner Added.

std/string.d
```diff
@@ -2939,6 +2939,30 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
...
     import std.utf : decodeFront;
 
+    static if ((isDynamicArray!Range || (isRandomAccessRange!Range && hasLength!Range))
+        && !is(Unqual!(ElementEncodingType!Range) == dchar)
+        && __traits(compiles,
+            {
+                size_t s = input.length / 2;
+                input = input[s .. $];
+            }))
+    {{
```
BBasile: Enlighten me, why double curly braces here ?

n8sh: To create a scope. `static if` curly braces don't do that so I need a second pair.

BBasile: Sorry i must be tired but i don't see why a scope is needed here.

n8sh: Here it's not needed, I just prefer it.

n8sh:
Reworked. Added optimizations for `wstring` and `dstring` and avoided duplicate decoding for `string`.

std/string.d
```diff
@@ -2937,26 +2937,96 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
...
-                break;
-            input.popFront();
+            size_t i = 0;
+            for (const size_t end = input.length; i < end; ++i)
+            {
+                if (!std.uni.isWhite(input[i]))
+                    break;
+            }
+            input = input[i .. $];
+            return input;
```
schveiguy: How about:
```D
foreach(i; 0 .. input.length)
{
    if (!std.uni.isWhite(input[i]))
        return input[i .. $];
}
return input[$ .. $];
```

std/string.d
```diff
@@ -2937,26 +2937,96 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
...
     {
-        auto c = input.front;
-        if (std.ascii.isASCII(c))
+        // Decoding is never needed for dchar. It happens not to be needed
+        // here for wchar because no whitepace is outside the basic
+        // multilingual plane meaning every whitespace character is encoded
+        // with a single wchar and due to the design of UTF-16 those wchars
+        // will not occur as part of the encoding of multi-wchar codepoints.
+        static if (isDynamicArray!Range || (isRandomAccessRange!Range && hasLength!Range
+            && __traits(compiles, { size_t s = input.length / 2; input = input[s .. $]; })))
```
schveiguy: I don't think the traits compiles thing is necessary. Just use `hasSlicing`.

std/string.d
```diff
@@ -2937,26 +2937,96 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
...
+                input.popFront();
+            }
+            return input;
+        }
+    }
+    else
+    {
+        static if (isDynamicArray!Range || (isRandomAccessRange!Range && hasLength!Range
+            && __traits(compiles, { size_t s = input.length / 2; input = input[s .. $]; })))
+        {{
```
schveiguy: 2 braces not needed here. Only variable introduced is `i`, and it's not even necessary if you contain it in the for loop.

std/string.d
```diff
@@ -2937,26 +2937,96 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
...
+            // ASCII optimization for dynamic arrays & similar.
+            size_t i = 0;
+            for (const size_t end = input.length; i < end; ++i)
```
schveiguy: The suggested modification in the other branch makes this cleaner as well.

std/string.d
```diff
@@ -2937,26 +2937,96 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
...
+            {
+                auto c = input[i];
+                if (c >= 0x80) goto NonAsciiPath;
```
schveiguy: Can we avoid a `goto` here?

std/string.d
```diff
@@ -2937,26 +2937,96 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
...
+            }
+            input = input[i .. $];
+            return input;
+
+        NonAsciiPath:
+            input = input[i .. $];
+            // Fall through to standard case.
+        }}
+
+        static if (ElementType!Range.sizeof > 1)
```
schveiguy: How about `is(ElementType!Range != ElementEncodingType!Range)`. The sizeof thing requires one to deduce that you only got here if the element encoding type is char.

std/string.d
```diff
@@ -2937,26 +2937,96 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
...
+                    if (!std.ascii.isWhite(c))
+                        break;
+                    input.popFront();
+                }
+                else
+                {
+                    auto save = input.save;
+                    auto dc = decodeFront!(UseReplacementDchar.yes)(input);
+                    if (!std.uni.isWhite(dc))
+                        return save;
```
schveiguy: This section of code isn't tested by the given unittests.

std/string.d
```diff
@@ -2937,26 +2937,96 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
...
+                    break;
+                input.popFront();
+            }
+            return input;
+        }
+        else
+        {
+            // Type doesn't perform its own decoding. We impose decoding on it
+            // which is questionable but is how this function has worked up to
+            // now.
```
schveiguy: Why is this questionable?

std/string.d
n8sh: I need to also check that the result of the slice can be assigned back to input.

schveiguy: `hasSlicing` implies that `input[x .. y]` returns `typeof(input)`. All you have to do is check that `input` is assignable to `input`.

std/string.d
n8sh: I don't think so.

std/string.d
n8sh: Seem to be nearly the same, so I'll use your suggestion.

std/string.d
n8sh: With your rewrite I don't need to check even that.

std/string.d
n8sh: Added unittest.

std/string.d
n8sh: Comment removed.

ZombineDev:
> One thing that isn't crystal clear to me is whether the random access usage is needed for anything but arrays, as we don't know the mechanics of generic random-access ranges (and whether indexing is faster than front/popFront).

I haven't had the time to review this PR, just wanted to mention that if you're iterating sequentially, indexing with a CPU predictable index variable (e.g. `i++`) should be the same as using the range primitives for arrays, however for other ranges like ndslice, it's faster to use `popFront` as the indexing will repeat a noticable amount of work - suming the offsets for each dimensions O(N), while with popFront that's amortized O(1), where N is the number of dimensions.

n8sh:
Changed the PR to only use an index variable for arrays.

schveiguy:
> indexing with a CPU predictable index variable (e.g. `i++`) should be the same as using the range primitives for arrays

I would think indexing and incrementing would be less expensive than `front` and `popFront` for arrays, since `popFront` updates both the pointer and length. But maybe that is made up for by the fact that you are indexing with a constant?

std/string.d
schveiguy: Hm... I was thinking this too could be rewritten:

```D
foreach (i; 0 .. input.length)
{
    auto c = input[i];
    if (c >= 0x80)
    {
        input = input[i .. $];
        goto NonAsciiPath;
    }
    if (!std.ascii.isWhite)
        return input[i .. $];
}
return input[$ .. $];
NonAsciiPath:
```
While it doesn't get rid of the goto, the loop looks cleaner to me.

nordlow:
This is a great idea. There are more functions in Phobos that would benefit from ASCII-optimized paths. I manage to find these

- `find`
- `canFind`
- `startsWith`
- `endsWith`
- `findSplitBefore`
- `findSplitAfter`
- `findSplit`

.

std/string.d
n8sh: The rewrite makes the ASCII loop slower while avoiding a call to `_d_arraybounds`. When processing `"\n\t\v\rhello world\n\t\v\r"`, with `ldc2 -O2` the function overall takes ~5% longer and with `dmd -O -inline` it takes ~30% longer. Obviously when `-release` is added and no bounds checks are performed the comparison gets worse.

schveiguy: This has to be a cache line thing, because the major work done in the loop (calling isWhite and checking if c is non-ascii) is identical in both cases. In any case, it's only marginally more readable, so the original is fine.

schveiguy:
ping @wilzbach I can't figure out what went wrong in the buildkite. Seems to be consistent, but there is no real error message except that some script failed:

```
[ERROR] /var/lib/buildkite-agent/builds/buildkite-agent-01-1/dlang/phobos/build/dlang-dub/test/interactive-remove.sh:10 command failed
[ERROR] Script failure.
```

Is this a false error?

wilzbach:
Yes.

schveiguy:
Auto-merge toggled on

schveiguy:
OK, thanks. Just used the old mechanism instead.

PR #6729, opened 2018-10-16
Ddoc: fix botched backticks

aG0aep6G:


n8sh:
In the past @andralex requested that code sections longer than a single token still be enclosed with `$D(...)` instead of backticks. Anyone know if that still holds?

andralex:
The only reason is that refactoring may break backticked code across a line break, and sadly ddoc does not support that.

std/algorithm/sorting.d
```diff
@@ -3669,8 +3670,8 @@ if (isRandomAccessRange!(Range1) && hasLength!Range1 &&
 /**
 Copies the top `n` elements of the
 $(REF_ALTTEXT input range, isInputRange, std,range,primitives) `source` into the
-random-access range `target`, where `n $D(
-target.length). Elements of `source` are not touched. If $(D
+random-access range `target`, where `n = target.length`.
+Elements of `source` are not touched. If $(D
 sorted) is `true`, the target is sorted. Otherwise, the target
```
wilzbach:
```suggestion
Elements of `source` are not touched. If `sorted`
is `true`, the target is sorted. Otherwise, the target
```

PR #6731, opened 2018-10-18
Fix Issue 19312 - Reduce template bloat in std.math by using const arguments

n8sh:
Labeling variably-typed arguments of template functions as const causes the same type to be inferred when the argument is const, non-const, and immutable.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -576,7 +576,7 @@ template isDeprecatedComplex(T)
  *     The absolute value of the number.  If floating-point or integral,
  *     the return type will be the same as the input;
  */
-auto abs(Num)(Num x)
+auto abs(Num)(const Num x)
 if ((is(Unqual!Num == short) || is(Unqual!Num == byte)) ||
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -6302,34 +6303,34 @@ version (D_HardFloat) @safe unittest // rounding
  * Returns:
  *  `true` if $(D_PARAM x) is Nan.
  */
-bool isNaN(X)(X x) @nogc @trusted pure nothrow
+bool isNaN(X)(const X x) @nogc @trusted pure nothrow
```
n8sh: `const` here causes an error in CTFE for `isNaN(real.nan)`. Removing the `const` allows it to work fine. Bizarre.

ZombineDev: My guess is that `isNaN` (along with other functions) is evaluated as a built-in at compile-time and such built-ins are matched by their mangled name.

Geod24: Yep. Although some of those builtins are outdated and can (will!) removed at some point, because the real builtins are in `core`.

PR #6732, opened 2018-10-20
Fix Issue 8025 - returned response body for std.net.curl.del

tibi77:


std/net/curl.d
```diff
@@ -801,6 +801,7 @@ if (isCurlConn!Conn)
                                 text("No filename specified to delete for URL ", url));
         conn.addCommand("DELE " ~ t[1]);
         conn.perform();
+	return "DELE" ~ t[1];
```
CyberShadow: I think it's better to return `null` rather than dummy data. Also, let's fix the indentation.

Testing the new "suggestions" feature:

```suggestion
        return null;
```

std/net/curl.d
```diff
@@ -776,13 +776,13 @@ if ( isCurlConn!Conn && (is(T == char) || is(T == ubyte)) )
  *
  * See_Also: $(LREF HTTP.Method)
```
CyberShadow: Please add a "Returns:" clause as in the other functions returning a value. E.g.:

```
 * Returns:
 * A `T[]` range containing the response body for this request (`null` in case of FTP).
```

wilzbach:
Ping @tibi77. Any chance you get around finishing this?

PR #4012, opened 2016-02-21
Fix issue# 15710: Replace std.utf.validate.

jmdavis:
This introduces std.utf.isValidUnicode, which returns bool rather than
throwing and schedules validate for deprecation in 2.072 (the release
after isValidUnicode would be introduced).

PR #6734, opened 2018-10-20
Fix issue 15710 - Added new method to validate an UTF string without throwing error

ioanas96:


ioanas96:
I have seen the discussion on the issues list about redesigning how decode works internally. On what status is this? Are there any solutions or any new ideas?

jmdavis:
I honestly don't think that we should be adding a simple wrapper like this. Anyone who wants to do this can do it themselves trivially, and it's arguably even detrimental to add such a wrapper since it hides a performance problem. What needs to happen is a rewrite of the internals of `decode` so that it's possible to reuse them to add a function which just asks whether a range of characters is a valid range of code points or not and doesn't involve exceptions at all. I did that previously with this PR: https://github.com/dlang/phobos/pull/4012. However, Andrei did not like the approach. He thought that it was too hacky and wanted a more fundamental rewrite that was cleaner. I expect that I'll get back to it eventually, but it's very hard code to work on, and it hasn't been high on my todo list.

But even if we were willing to add a simple wrapper (which we really try to avoid in Phobos), this PR does not work with ranges in general (just strings) and has no documentation aside from a couple of basic ddoc-ed unit tests. It also has no real tests to verify that it's working properly with various inputs, just a couple of really basic examples.

n8sh:
Non-throwing string validation is useful because exception-handling is expensive. Just catching the exception doesn't fix this.

>It also has no real tests to verify that it's working properly with various inputs, just a couple of really basic examples.

I consider the test coverage adequate for a function that delegates to an existing function with its own tests.

n8sh:
>I have seen the discussion on the issues list about redesigning how decode works internally. On what status is this?

`decode!(UseReplacementDchar.yes)` is non-throwing. I've added a commit to this PR that makes `isValidUTF`  make use of it.

n8sh:
>simply comparing the decoded code point against the replacement character does not tell you whether the code point was replaced (meaning the code point in the input was bad) or whether the input already contained the replacement character and thus was just fine.

Fair point. My belief was that `\uFFFF` has no purpose other than in-band signaling of Unicode errors, so even if it appears "naturally" in the string it merely means that somewhere upstream there had been an error. But maybe this is not so?

>Also, again, if we're going to add a function like this, it needs to work on ranges in general, not just strings.

That's a nice-to-have, but not a necessity as the existing `validate` only accepts strings.

jmdavis:
> Fair point. My belief was that \uFFFF has no purpose other than in-band signaling of Unicode errors, so even if it appears "naturally" in the string it merely means that somewhere upstream there had been an error. But maybe this is not so?

Regardless of whether there was a bad code point in the data at some point previously in the program, there isn't one now. For instance, a program could be purposefully replacing bade code points with the replacement characters in one place (or have data from another program which did) and then have other data which it was joined to it where it needs to validate it. The replacement character is valid Unicode and should not be treated as invalid Unicode. That distinction needs to be kept. We don't know what exactly the program is going to be doing, and we shouldn't assume that it doesn't need to retain the distinction. This exact issue was actually a problem when validating XML in dxml (which unfortunately is a bit more complicated, since the XML spec was stupid enough to treat even more characters as invalid for no apparent reason, but not being able to tell whether a character was already the replacement character or not was definitely a problem and something that I had to work around).

> That's a nice-to-have, but not a necessity as the existing validate only accepts strings.

We should not be adding string-only functions, especially when it's relatively trivial to make them work with ranges.

n8sh:
I have come to agree regarding `\uFFFD` as [non-characters are valid in Unicode strings and UTFs](https://www.unicode.org/faq/private_use.html#nonchar8). Besides which, treating it as invalid would cause asymmetry between `validate` and `isValidUTF`.

>**Q: Are noncharacters invalid in Unicode strings and UTFs?**
>A: Absolutely not. Noncharacters do not cause a Unicode string to be ill-formed in any UTF. This can be seen explicitly in the [table above](https://www.unicode.org/faq/private_use.html#nonchar_codes), where every noncharacter code point has a well-formed representation in UTF-32, in UTF-16, and in UTF-8. An implementation which converts noncharacter code points between one UTF representation and another must preserve these values correctly. The fact that they are called "noncharacters" and are not _intended_ for open interchange does not mean that they are somehow illegal or invalid code points which make strings containing them invalid.

>**Q: So how should libraries and tools handle noncharacters?**
>A: Library APIs, components, and tool applications (such as low-level text editors) which handle all Unicode strings should also handle noncharacters. Often this means simple pass-through, the same way such an API or tool would handle a reserved unassigned code point. Such APIs and tools would not normally be expected to interpret the semantics of noncharacters, precisely because the intended use of a noncharacter is internal. But an API or tool should also not arbitrarily filter out, convert, or otherwise discard the value of noncharacters, any more than they would do for private-use characters or reserved unassigned code points.

jmdavis:
The more I think about this, the more I'm inclined to think that what we really want is public functions equivalent to `decode` and `decodeFront` which return whether the code point is valid or not, and then `isValidUTF` can be built on top of them. Then other code (such as what I did with dxml) can take advantage of them, since there are plenty of cases where someone needs to check a single code point without checking the entire range. Of course, the core problem is still reworking `decodeImpl` in a satisfactory manner, and that's far from easy (which is why I have yet to go back and finish reworking my PR for `isValidUTF`). There was a post in the newsgroup a month or two back about someone who had an implementation that checked for Unicode correctness using simd (in C or C++ IIRC) that was really fast, and Andrei showed great interest in us adapting it. At the time, I wasn't all that excited about it, primarily because it would mean that the logic wouldn't be shared with `decodeImpl`, but given the difficulties in cleanly reworking `decodeImpl` to work for all three use cases and the speed gains, maybe it would be better to not share code anyway. But I don't think that the code that was linked to was Boost, so I'm not sure that it could be used, meaning that it would probably have to be rewritten from scratch based on the idea.

Regardless, the core problem here is figuring out how to get the `decodeImpl` side of things working cleanly for `isValidUTF` to work. The way it's currently written just simply isn't compatible. If we can rework it to have something like `startsWithValidCodePoint` (or `decodeValid`, or whatever a good function name would be), then we're golden, but that's not easy.

jmdavis:
Although there are other use cases where having something more like `tryDecode` which returned a `Nullable!dchar` (or an `Expected!dchar` if we had an `Expected` type as has been discussed elsewhere), so some I don't know if a function simply returning `bool` is really the best idea in general. For `isValidUTF`, that's basically what we want though.

Basically, we have some core functionality here that's fairly complex that needs to be very efficient but which also needs subtly different APIs depending on the use case, and meeting all of that without being overly complex in terms of the API or the implementation - and without losing out on efficiency - isn't exactly easy. Thus far, we've managed to partially provide the functionality that we should without much code duplication, but adding further flexibility has proven difficult. And Unicode is nasty to begin with. :|

RazvanN7:
I'm going to close this due to lack of activity.

PR #6735, opened 2018-10-20
Fix Issue 7006 - Enabled pow to accept negative exponents for integrals

cristiancreteanu:


ibuclaw:
Why should negative exponents be allowed when you can't represent any of them as an integer?

std/math.d
```diff
@@ -7525,10 +7525,13 @@ if (isFloatingPoint!(F) && isIntegral!(G))
 typeof(Unqual!(F).init * Unqual!(G).init) pow(F, G)(F x, G n) @nogc @trusted pure nothrow
 if (isIntegral!(F) && isIntegral!(G))
 {
-    if (n<0) return x/0; // Only support positive powers
```
n8sh: Why does this runtime check exist instead of requiring the exponent to be unsigned?

dukc: This is a good reason to: https://forum.dlang.org/post/qobojhyjmxzvxorlhhnj@forum.dlang.org

std/math.d
```diff
@@ -7525,10 +7525,13 @@ if (isFloatingPoint!(F) && isIntegral!(G))
 typeof(Unqual!(F).init * Unqual!(G).init) pow(F, G)(F x, G n) @nogc @trusted pure nothrow
```
joakim-noah: The documentation above will need to be corrected too, if this is to be accepted.

berni44:
@RazvanN7 IMHO, this one can be closed - the bug has been fixed in #7322

RazvanN7:
@berni44 Thanks for the heads up.

PR #6736, opened 2018-10-21
Fix Issue 19197 - Replace instances of typeid(T).getHash(..) with hashOf

tibi77:


std/typecons.d
```diff
@@ -1175,7 +1175,7 @@ if (distinctFieldNames!(Specs))
             size_t h = 0;
             static foreach (i, T; Types)
             {{
-                const k = typeid(T).getHash((() @trusted => cast(const void*) &field[i])());
+                const k = hashOf((() @trusted => cast(const void*) &field[i])());
```
n8sh: This is wrong: you're hashing a pointer to the member rather than the member itself. `hashOf` takes its argument by reference (or by value in cases where passing by value is equivalent and faster).

n8sh:
Closed due to https://github.com/dlang/phobos/pull/6699 having been merged.

PR #6738, opened 2018-10-22
Fix issue 4957 - std.concurrency rejects structs with Tid

Geod24:
```
In addition to the fix which makes it recurse in struct,
the instantiation was changed to not use recursive template instantiation,
exception for struct's tupleof.
```

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -72,13 +72,26 @@ import std.traits;
 
 private
 {
-    template hasLocalAliasing(T...)
+    bool hasLocalAliasing(Types...)()
     {
-        static if (!T.length)
-            enum hasLocalAliasing = false;
-        else
-            enum hasLocalAliasing = (std.traits.hasUnsharedAliasing!(T[0]) && !is(T[0] == Tid)) ||
-                                     std.concurrency.hasLocalAliasing!(T[1 .. $]);
+        // Works around "statement is not reachable"
+        bool doesIt = false;
+        static foreach (T; Types)
+        {
+            static if (is(T == Tid))
+            { /* Allowed */ }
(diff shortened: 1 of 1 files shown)


bithavoc:
I hope this gets merged, this and https://issues.dlang.org/show_bug.cgi?id=5570 made me feel that Dlang was not stable enough a few years back. I've seen more activity in these bug reports these days, I'm feeling tempted to try Dlang again ☕️ .

andrej-mitrovic:
> the instantiation was changed to not use recursive template instantiation

But doesn't the old code short-circuit, and yours does not? Then again maybe the performance is ultimately better because of a lack of recursive template instantiations.. ¯\_(ツ)_/¯

n8sh:
>Is there a way to short-circuit without recursive instantiation ?

In my experience you can write `doesIt |= ...; if (doesIt) return true;` without triggering an unreachable code check.

Geod24:
Don't think this has much influence, because by the time the code is run at CTFE all we have is `doesIt |= [true|false]` so we're just assigning from a constant. Evaluating the template is what we should really short-circuit, but we can't do that with CTFE.

n8sh:
For the benefit of dlang-bot I edited the commit message to start "Fix Issue 4957" instead of "Fix 4957", but that's been clobbered. Please make that change.

Geod24:
I did already, but dlang-bot doesn't seem to realize this.

Geod24:
@n8sh : Ah, commit message, not PR title. Sorry. Thanks for the fix!

wilzbach:
Yep that's needed to aggregate the changelog entries over commit messages

PR #6741, opened 2018-10-29
Removed an unittest of minimallyInitializedArray that can fail randomly

dukc:
This test assumed that there would happen to be other bytes than zeroes in the area that was allocated. Since random pieces of memory often are just zeroes, this one probably was responsible of some of the random autotester failures.

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -708,8 +708,12 @@ if (isDynamicArray!T && allSatisfy!(isIntegral, I))
 
     auto arr = minimallyInitializedArray!(int[])(42);
     assert(arr.length == 42);
-    // Elements aren't necessarily initialized to 0
-    assert(!arr.equal(0.repeat(42)));
+
+    // Elements aren't necessarily initialized to 0, so don't do this:
+    // assert(arr.equal(0.repeat(42)));
+    // If that is needed, initialize the array normally instead:
+    auto arr2 = new int[42];
+    assert(arr2.equal(0.repeat(42)));
 }
 
 @safe pure nothrow unittest


std/array.d
```diff
@@ -708,8 +708,8 @@ if (isDynamicArray!T && allSatisfy!(isIntegral, I))
 
     auto arr = minimallyInitializedArray!(int[])(42);
     assert(arr.length == 42);
-    // Elements aren't necessarily initialized to 0
-    assert(!arr.equal(0.repeat(42)));
+    // Elements aren't necessarily initialized to 0, so don't do this:
+    // assert(arr.equal(0.repeat(42)));
```
Geod24: Having leftover comment feels a bit sloppy. Could this be turned into a snippet that does the right thing (e.g. initialize the array) ?

dukc:
I can't interprete what DAutoTest is complaining about. Can anyone tell me?

std/array.d
dukc: Good idea

Geod24:
It's not related to this PR, unfortunately.

dukc:
Can't be. If arrays were initialized, that test I just removed would have always failed!

Geod24:
@wilzbach :  I doubt `hasIndirection!int` would yield `true`.

dukc:
I was just going to say the same

std/array.d
```diff
@@ -708,8 +708,12 @@ if (isDynamicArray!T && allSatisfy!(isIntegral, I))
 
     auto arr = minimallyInitializedArray!(int[])(42);
     assert(arr.length == 42);
-    // Elements aren't necessarily initialized to 0
-    assert(!arr.equal(0.repeat(42)));
+
+    // Elements aren't necessarily initialized to 0, so don't do this:
+    // assert(arr.equal(0.repeat(42)));
+    // If that is needed, initialize array like this instead:
+    auto arr2 = 0.repeat(42).array;
```
n8sh: Is there a reason to recommend this instead of `new int[42]`?

dukc: No, but I didn't consider it worse either, unless I count the fact that `new int[42]` is closer to `minimallyInitializedArray!int(42)` than my line is.

n8sh: In that case I think lines 712 through 715 should be replace with simply:
```d
    // Because an int does not contain any pointers the elements of `arr`
    // are not initialized to zero.
```
But if you disagree this can still go forward.

dukc: Mathias wanted an example of how to initialize the array correctly, so I'm not deleting it.

I changed the initialization to what you suggested instead.

PR #6743, opened 2018-10-31
Decrease template bloat for string functions

n8sh:
Use const(E)[] instead of E[] when it is valid and doesn't change the return type.

### DIFF
--- a/std/range/primitives.d
+++ b/std/range/primitives.d
@@ -2192,7 +2192,7 @@ the first argument using the dot notation, `array.save` is
 equivalent to `save(array)`. The function does not duplicate the
 content of the array, it simply returns its argument.
  */
-@property T[] save(T)(T[] a) @safe pure nothrow @nogc
+@property inout(T)[] save(T)(return scope inout(T)[] a) @safe pure nothrow @nogc
 {
     return a;
 }
@@ -2213,7 +2213,7 @@ equivalent to `popFront(array)`. For $(GLOSSARY narrow strings),
 `popFront` automatically advances to the next $(GLOSSARY code
 point).
 */
-void popFront(T)(ref T[] a) @safe pure nothrow @nogc
+void popFront(T)(scope ref inout(T)[] a) @safe pure nothrow @nogc
 if (!isNarrowString!(T[]) && !is(T[] == void[]))
 {
     assert(a.length, "Attempting to popFront() past the end of an array of " ~ T.stringof);
@@ -2236,7 +2236,7 @@ if (!isNarrowString!(T[]) && !is(T[] == void[]))
 }
 
 /// ditto
-void popFront(C)(ref C[] str) @trusted pure nothrow
+void popFront(C)(scope ref inout(C)[] str) @trusted pure nothrow
 if (isNarrowString!(C[]))
 {
     import std.algorithm.comparison : min;
@@ -2334,7 +2334,7 @@ the first argument using the dot notation, `array.popBack` is
 equivalent to `popBack(array)`. For $(GLOSSARY narrow strings), $(D
 popFront) automatically eliminates the last $(GLOSSARY code point).
 */
-void popBack(T)(ref T[] a) @safe pure nothrow @nogc
+void popBack(T)(scope ref inout(T)[] a) @safe pure nothrow @nogc
 if (!isNarrowString!(T[]) && !is(T[] == void[]))
 {
     assert(a.length);
@@ -2357,7 +2357,7 @@ if (!isNarrowString!(T[]) && !is(T[] == void[]))
 }
 
 /// ditto
-void popBack(T)(ref T[] a) @safe pure
+void popBack(T)(scope ref inout(T)[] a) @safe pure
 if (isNarrowString!(T[]))
 {
     import std.utf : strideBack;
@@ -2401,8 +2401,15 @@ equivalent to `front(array)`. For $(GLOSSARY narrow strings), $(D
 front) automatically returns the first $(GLOSSARY code point) as _a $(D
 dchar).
 */
-@property ref T front(T)(T[] a) @safe pure nothrow @nogc
+@property ref T front(T)(return scope T[] a) @safe pure nothrow @nogc
 if (!isNarrowString!(T[]) && !is(T[] == void[]))
+// We would have preferred to write the function template
+// ---
+//     @property ref inout(T) front(T)(return scope inout(T)[] a)
+//        if (/* same constraint */)
+// ---
+// as that would cause fewer distinct functions to be generated with
+// IFTI, but that caused a linker error in the test suite on Win32_64.
 {
     assert(a.length, "Attempting to fetch the front of an empty array of " ~ T.stringof);
     return a[0];
@@ -2430,7 +2437,7 @@ if (!isNarrowString!(T[]) && !is(T[] == void[]))
 }
 
 /// ditto
-@property dchar front(T)(T[] a) @safe pure
+@property dchar front(T)(scope const(T)[] a) @safe pure
 if (isNarrowString!(T[]))
 {
     import std.utf : decode;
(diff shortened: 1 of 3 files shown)


std/range/primitives.d
```diff
@@ -2236,7 +2236,7 @@ if (!isNarrowString!(T[]) && !is(T[] == void[]))
 }
 
 /// ditto
-void popFront(C)(ref C[] str) @trusted pure nothrow
+void popFront(C)(scope ref C[] str) @trusted pure nothrow
```
n8sh: Unfortunately, even though `char[]` is implicitly convertible to `const(char)[]`, if I change `ref C[] str` to `ref const(C)[] str` the template will fail to match.

burner:
I assumed you grep'ed for the usage of decodeImpl to make sure CastForDecode is used when needed?

n8sh:
In effect yes. The only place it's not used is in a unittest.

burner:
Wouldn't it be better if decodeImpl would call CastForDecode, because currently we have to make sure that people remember calling CastForDecode if necessary.

n8sh:
Can you suggest a good way to do that? I'm not enthusiastic about adding a wrapper function.

burner:
sorry, strike my last comment. I misread the call side.

std/range/primitives.d
```diff
@@ -2213,7 +2213,7 @@ equivalent to `popFront(array)`. For $(GLOSSARY narrow strings),
 `popFront` automatically advances to the next $(GLOSSARY code
 point).
 */
-void popFront(T)(ref T[] a) @safe pure nothrow @nogc
+void popFront(T)(scope ref inout(T)[] a) @safe pure nothrow @nogc
```
jmdavis: `inout` makes no sense if the function isn't returning anything. I'm honestly surprised that it even compiles. Why add it? The same goes for the various other pop functions below.

std/range/primitives.d
```diff
@@ -2192,7 +2192,7 @@ the first argument using the dot notation, `array.save` is
 equivalent to `save(array)`. The function does not duplicate the
 content of the array, it simply returns its argument.
  */
-@property T[] save(T)(T[] a) @safe pure nothrow @nogc
+@property inout(T)[] save(T)(return scope inout(T)[] a) @safe pure nothrow @nogc
```
jmdavis: Given that this is already a templated function, what does adding `inout` add here (or anywhere in this entire PR)? Unless I misunderstand what the compiler does, it doesn't even save any template instantiations. it just makes the function signature more complicated for no real benefit. The primary benefit of `inout` is allowing a single function to work with `const`, `immutable`, and mutable when it's _not_ a template.

std/utf.d
```diff
@@ -1065,6 +1065,16 @@ if (isSomeChar!C)
 /// Whether or not to replace invalid UTF with $(LREF replacementDchar)
 alias UseReplacementDchar = Flag!"useReplacementDchar";
 
+// Reduce distinct instantiations of decodeImpl.
+private template CastForDecode(T)
+{
+    static if (isDynamicArray!T && is(T : E[], E) && __traits(isArithmetic, E)
+            && !is(E == shared))
+        alias CastForDecode = const(Unqual!E)[];
```
jmdavis: Honestly, this seems pretty convoluted, though all of the other ways I can think of doing it would make it considerably longer. However, at least please make the `static if ` a single line so that it's not broken up without braces.

std/utf.d
```diff
@@ -1065,6 +1065,16 @@ if (isSomeChar!C)
 /// Whether or not to replace invalid UTF with $(LREF replacementDchar)
 alias UseReplacementDchar = Flag!"useReplacementDchar";
 
+// Reduce distinct instantiations of decodeImpl.
+private template CastForDecode(T)
```
jmdavis: Please rename to something like `TypeForDecode`. It's named like it's a function, when its result is a type.

n8sh:
@jmdavis The purpose of this PR is not to improve compile times by reducing the absolute number of template uses, it's to improve the cache locality of the final executable by decreasing the number of distinct functions created by template instantiation. Look at the generated assembly for this code:

https://run.dlang.io/is/fMHThw
```d
void foo(T)(ref T[] a)
{
    a = a[1 .. $];
}

void bar(T)(ref inout(T)[] a)
{
    a = a[1 .. $];
}

void main()
{
    int[] a = [1, 2, 3];
    const(int)[] b = [4, 5, 6];
    immutable(int)[] c = [7, 8, 9];
    foo(a);
    foo(b);
    foo(c);
    bar(a);
    bar(b);
    bar(c);
}
```

The generated code has 3 separate `foo` functions that are each called once, while there is 1 single `bar` function that is called 3 times.

std/utf.d
n8sh: Done.

std/utf.d
n8sh: Done.

std/range/primitives.d
n8sh: It results in fewer distinct functions in the final executable. See https://github.com/dlang/phobos/pull/6743#issuecomment-435770952

std/range/primitives.d
n8sh: It means "you can use the exact same compiled function regardless of the storage class of `a`'s elements."  It results in fewer distinct functions in the final executable.  See https://github.com/dlang/phobos/pull/6743#issuecomment-435770952.

quickfur: Why not just `const(T)[]` then?  `inout` is basically const inside the function body, the only difference is that the return value may be implicitly converted to either immutable or mutable since the function cannot have modified it.  Since nothing is being returned here, `inout` is misleading. Just use `const` instead.

n8sh:
I agree with not using `inout` unnecessarily. I don't think there is anywhere in this PR where I wrote `inout(T)` when I could have written `const(T)` instead. If you see any such places please point them out!

std/range/primitives.d
quickfur: Argh... I see why now.  The compiler sees `const(T)` and tries to match *that* to the parameter type, rather than matching `T` to the parameter type first then implicitly converting to `const(T)`.  I'm tempted to call this a compiler bug / language defect, but I'm not sure I want to open the can of worms that is how IFTI performs its black magic.

As for why `inout` works at all, I'm not sure I want to know the real reason. But I guess it will have to suffice for now, so that this PR doesn't devolve into a tarpit of how/whether/when the compiler could be fixed to allow passing `T` to `const(T)` in a template parameter.

quickfur:
I see that I was mistaken about the one use of `inout` I was concerned about.  IFTI rears its ugly head again, and I think using `inout` in that case is a necessary evil, at least for the time being.  (I'm almost certain the fact that `inout` works where `const` doesn't is probably an oversight / unintended consequence of how IFTI is currently implemented, but I think a discussion of that will be far beyond the scope of this PR. So let's just leave it as `inout` for now.)

thewilsonator:
Note the the style checker is red (why I've no idea) so this will need to be merged manually.

jmdavis:
It's red, because it won't build due to build errors, not because of actual style errors (I don't know if there are actually style errors or not). I don't know what the deal with it is right now, but either way, I don't seem to have the permission to merge pull requests manually anymore (I used to - but the button is greyed out, and presumably, if I tried to do it via the command line or via the auto-tester, I'd get rejected, since the button here is greyed out).

thewilsonator:
Removing auto-merge, this fails on win64 with duplicate combat or something.

jmdavis:
Oh, lovely. A windows-only compiler bug. Gotta love those.

thewilsonator:
@n8sh this is all green, is it ready to go?

n8sh:
Yes!

thewilsonator:
Aaaand its red again. Bummer.

thewilsonator:
Yep, that went green.

thewilsonator:
Since `style` is guaranteed to fail, feel free to abuse `auto-merge` to get priority to reduce this.

thewilsonator:
And now `style` fails, WTF?

std/range/primitives.d
```diff
@@ -2401,8 +2401,15 @@ equivalent to `front(array)`. For $(GLOSSARY narrow strings), $(D
...
-@property ref T front(T)(T[] a) @safe pure nothrow @nogc
+@property ref T front(T)(return scope T[] a) @safe pure nothrow @nogc
 if (!isNarrowString!(T[]) && !is(T[] == void[]))
+// We would have preferred to write the function template
+// ---
+//     @property ref inout(T) front(T)(return scope inout(T)[] a)
+//        if (/* same constraint */)
+// ---
+// as that would cause fewer distinct functions to be generated with
+// IFTI, but that caused a linker error in the test suite on Win32_64.
```
thewilsonator: Its great that you found the source of the error, did you submit a bug for it?

PR #6744, opened 2018-11-01
std.math: Use SPARC_Any instead of SPARC64

ibuclaw:
Posted and reviewed on gcc-patches, enables building with SPARC32 targets.


n8sh:
The aforementioned post: https://gcc.gnu.org/ml/gcc-patches/2018-10/msg02002.html

PR #6746, opened 2018-11-01
Move std.numeric.CustomFloat.ToBinary to top-level

RazvanN7:
Normally, when a symbol is resolved, the alias this scope of the aggregate should be searched before moving to the upper scopes. However, due to a dmd bug, the alias this scope is checked after all the upper scopes are searched. [1] fixes the dmd bug, but adds a new transition switch. Compiling phobos with the new transition switch fails because ToBinary is resolved through the alias this. This is the correct behavior according to the new lookup rule. This PR makes phobos compliant with the new compiler transition switch.

[1] https://github.com/dlang/dmd/pull/8885

### DIFF
--- a/std/numeric.d
+++ b/std/numeric.d
@@ -145,6 +145,27 @@ if (((flags & flags.signed) + precision + exponentWidth) % 8 == 0 && precision +
     auto p = Probability(0.5);
 }
 
+// Facilitate converting numeric types to custom float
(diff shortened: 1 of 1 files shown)


ZombineDev:
@RazvanN7 please rebase in order to restart the CI.

wilzbach:
(rebased)

PR #6748, opened 2018-11-04
fix Issue 19337 - cannot call std.algorithm.sort twice

MartinNowak:
- introduced with #6535 which stripped nested SortedRange types
  in assumeSorted, thus causing a mismatch between the return type
  of sort (SortedRange!(SortedRange!R)) and assumeSorted SortedRange!R
- fix by stripping nested SortedRange types in the SortedRange!R
  template instantiation


MartinNowak:
I switched to use [std.traits.isInstanceOf](https://dlang.org/phobos/std_traits.html#isInstanceOf) to improve the readability of the docs.

PR #6750, opened 2018-11-05
Simplify isImplicitlyConvertible trait

edi33416:
`std.traits.isImplicitlyConvertible` should just use the `is` expression syntax [0].
This PR reduces and simplifies the implementation.

[0] - https://dlang.org/spec/expression.html#is_expression

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -4972,16 +4972,7 @@ template ImplicitConversionTargets(T)
 /**
 Is `From` implicitly convertible to `To`?
  */
-template isImplicitlyConvertible(From, To)
-{
-    enum bool isImplicitlyConvertible = is(typeof({
-        void fun(ref From v)
-        {
-            void gun(To) {}
-            gun(v);
-        }
-    }));
-}
+enum bool isImplicitlyConvertible(From, To) = is(From : To);
 
 ///
 @safe unittest


edi33416:
@wilzbach can you please have a look at buildkite? It's failing with
```
~> make -f posix.mak style
--
  | posix.mak:41: ../dmd/src/osmodel.mak: No such file or directory
  | awk: fatal: cannot open file `../dmd/VERSION' for reading (No such file or directory)
  | awk: fatal: cannot open file `../dmd/VERSION' for reading (No such file or directory)
  | make: *** No rule to make target '../dmd/src/osmodel.mak'.  Stop.
```

It looks like it's expecting a `dmd/VERSION` that doesn't exist.

BBasile:
The change looks good but does anyone know why it was not written like proposed in this PR in first place ? After searching in the DMD changelog it appears that the is syntax using colon has always existed.

edi33416:
> The change looks good but does anyone know why it was not written like proposed in this PR in first place ? After searching in the DMD changelog it appears that the is syntax using colon has always existed.

I had the same concern, but I can't think of a counter-example to the colon syntax.
If we can find one, that would be great news! Then we can just add a test to prevent another attempt like this one, in the future.

andralex:
@ZombineDev ah, so it was me. There might be some historical rationale for it, but it's very nice to have `isImplicitlyConvertible` simply map to `is`. I'll approve this. @edi33416 what's the thing with buildkite?

edi33416:
> [...] I'll approve this. @edi33416 what's the thing with buildkite?
@andralex as far as I can tell, it's an unrelated issue. @wilzbach can you also have a look please?

Can I also please get an invite to the `buildkite` org?

PR #6755, opened 2018-11-07
[stable] Fix opBinary[Right]`CustomFloat op CustomFloat` ambiguity

thewilsonator:
For dlang/dmd#8922

### DIFF
--- a/std/numeric.d
+++ b/std/numeric.d
@@ -553,7 +553,7 @@ public:
     ///ditto
     T opCast(T)() if (__traits(compiles, get!T )) { return get!T; }
 
-    /// Convert the CustomFloat to a real and perform the relavent operator on the result
+    /// Convert the CustomFloat to a real and perform the relevant operator on the result
     real opUnary(string op)()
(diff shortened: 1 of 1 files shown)


thewilsonator:
This is for stable so the DMD PR can build. Please review carefully.

std/numeric.d
```diff
@@ -568,16 +568,28 @@ public:
     }
 
     /// ditto
+    // Define an opBinary `CustomFloat op CustomFloat` so that those below
```
edi33416: Since you are here, can you please also fix the type in `opBinary`s doc?
Line 556, from `relavent` to `relevant`.

Thanks

thewilsonator: Done.

edi33416: Nice. Thanks

thewilsonator:
Thanks! I don't have merge rights, could you please do it.

RazvanN7:
This needs to be in master in order to unblock https://github.com/dlang/dmd/pull/8922

PR #6757, opened 2018-11-08
Simplify isInnerClass trait

edi33416:
This PR reduces and simplifies the implementation of `std.traits.isInnerClass`. This should also improve the compilation time.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -2513,11 +2513,8 @@ Returns:
 template isInnerClass(T)
 if (is(T == class))
 {
-    import std.meta : staticIndexOf;
-
     static if (is(typeof(T.outer)))
-        enum isInnerClass = __traits(isSame, typeof(T.outer), __traits(parent, T))
-                         && (staticIndexOf!(__traits(allMembers, T), "outer") == -1);
+        enum isInnerClass = __traits(isSame, typeof(T.outer), __traits(parent, T));
     else
         enum isInnerClass = false;
 }


std/traits.d
```diff
@@ -2513,11 +2513,8 @@ Returns:
 template isInnerClass(T)
 if (is(T == class))
 {
-    import std.meta : staticIndexOf;
-
     static if (is(typeof(T.outer)))
-        enum isInnerClass = __traits(isSame, typeof(T.outer), __traits(parent, T))
-                         && (staticIndexOf!(__traits(allMembers, T), "outer") == -1);
```
edi33416: `is(typeof(T.outer))` already checks if there's an `outer` member, which is what `staticIndexOf` does again.

edi33416: Related question: any reason to prefer `is(...)` to `__traits(hasMember, T, "outer")` ?

n8sh: Isn't the `staticIndexOf... == -1` verifying that there is *not* a member named "outer"?

edi33416: You are right. That is what I was trying to say, but I had a poor wording.

Now actually, I'm not sure what's supposed to be doing, because [staticIndexOf](https://dlang.org/phobos/std_meta.html#.staticIndexOf) checks if the first template argument is in the remainder of the template arguments list.

If it does indeed check if there is *no* `outer` member, won't the entire check evaluate to `false`, since it's part of the `is(typeof(T.outer))` branch path?

andralex: The arguments of `staticIndexOf` are REVERSED and that's a bug that should be caught in unittesting. @edi33416 can you please do that in this pr?

andralex: Can use `!hasMember!(T, "outer")`, see https://dlang.org/library/std/traits/has_member.html

UplinkCoder: using `__traits` directly will in most cases be faster than going through a template

TurkeyMan:
Why is this pending? What should the final PR look like? I need to repeat the proper implementation in druntime...

TurkeyMan:
Is `__traits(isSame, typeof(T.outer), __traits(parent, T)) && !__traits(hasMember, T, "outer")` the proper expression?

thewilsonator:
```
enum isInnerClass = __traits(isSame, typeof(T.outer), __traits(parent, T))
                          && (staticIndexOf!("outer", __traits(allMembers, T)) == -1);
```
will work. I'm not quite sure there is some confusion over wether it is correct to remove the `staticIndexOf `, `__traits(getMember, "outer")` may work.

TurkeyMan:
It doesn't really look like there's confusion... it appears @andralex requested the change and approved it?

std/traits.d
TurkeyMan: I tried this suggestion; it didn't work.
`__traits(hasMember, T, "outer") == true` for inner classes... even though it's not an implicit member.
I went with this: https://github.com/dlang/druntime/pull/2398/files#diff-f1322962163674c1ee8c7a018193ab45R160
It works.

TurkeyMan:
Yeah nar, the discussion here is all wrong.
This is the best solution I could manifest:
https://github.com/dlang/druntime/pull/2398/files#diff-f1322962163674c1ee8c7a018193ab45R160

thewilsonator:
Looks like we should have an _actual_ \_\_trait that does this.

UplinkCoder:
@thewilsonator Yes.

It should not be hard to add.

TurkeyMan:
I dun a different PR: https://github.com/dlang/phobos/pull/6757

edi33416:
> Why is this pending? What should the final PR look like? I need to repeat the proper implementation in druntime...

The call `staticIndexOf!(__traits(allMembers, T), "outer") == -1` (1) is wrong.
The check should be `staticIndexOf!("outer", __traits(allMembers, T)) == -1` (2).

[staticIndexOf](https://dlang.org/library/std/meta/static_index_of.html) has the *needle* as the first template argument and the *haystack* as the last one. Oddly enough, though, this *bugged* call seems to work.

I've tried to come up with unittest that fails with the current implementation (bugged call to `staticIndexOf`) so then the change to the correct call will also be documented through a test, but I didn't manage to write one that failed with (1), but passed with (2)

edi33416:
> Yeah nar, the discussion here is all wrong.
> This is the best solution I could manifest:
> https://github.com/dlang/druntime/pull/2398/files#diff-f1322962163674c1ee8c7a018193ab45R160
> 
> ```d
> template isInnerClass(T) if (is(T == class))
> {
>     static if (is(typeof(T.outer)))
>     {
>         template hasOuterMember(T...)
>         {
>             static if (T.length == 0)
>                 enum hasOuterMember = false;
>             else
>                 enum hasOuterMember = T[0] == "outer" || hasOuterMember!(T[1 .. $]);
>         }
>         enum isInnerClass = __traits(isSame, typeof(T.outer), __traits(parent, T)) && !hasOuterMember!(__traits(allMembers, T));
>     }
>     else
>         enum isInnerClass = false;
> }
> ```

Not sure if this is the right way to go, as `hasOuterMember` is the implementation of [genericIndexOf](https://github.com/dlang/phobos/blob/master/std/meta.d#L303) (which is what `staticIndexOf` calls), except that this generates a template recursion, whereas `genericIndexOf` does a `static foreach`.

TurkeyMan:
Yeah, I changed it to use `foreach`.
I just spent an hour extensively comparing `static foreach` and CTFE `foreach` expansions.
I couldn't create a case where `static foreach` was faster than CTFE `foreach` where it was possible to implement with both.

std/traits.d
```diff
@@ -2513,11 +2513,8 @@ Returns:
...
+        enum isInnerClass = __traits(isSame, typeof(T.outer), __traits(parent, T));
```
andralex: @TurkeyMan this implementation is really simple and uses built-in features, would this be the better of the two?

TurkeyMan: Ummm, this test is incomplete (as discussed in the first comment thread above). This test raises false positives when a type has an 'outer' member, which we need to negate with the search for an outer member. A real `__traits` would be better, but in lieu of that, it probably should still work properly.

TurkeyMan: I'm also not sure what you mean by 'built in features'? My test is identical, except with an extra clause that catches the false-positive case.

PR #6758, opened 2018-11-09
Typo in REF macro parameter list in each doc.

mdparker:
`$(REF Flag, std.typecons)`!"each"` should have a comma between the package names, not a dot. It was causing the text to render as `std.typecons..Flag!"each"`.


andralex:
Thanks!

PR #6761, opened 2018-11-12
Remove duplicate of Unqual implementation

edi33416:
druntime's `core.internal.traits.Unqual` is a copy of the same code. This PR removes the code duplication and the required double maintenance of `Unqual`.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -7349,26 +7349,8 @@ Removes all qualifiers, if any, from type `T`.
  */
 template Unqual(T)
 {
-    version (none) // Error: recursive alias declaration @@@BUG1308@@@
-    {
-             static if (is(T U ==     const U)) alias Unqual = Unqual!U;
-        else static if (is(T U == immutable U)) alias Unqual = Unqual!U;
-        else static if (is(T U ==     inout U)) alias Unqual = Unqual!U;
-        else static if (is(T U ==    shared U)) alias Unqual = Unqual!U;
-        else                                    alias Unqual =        T;
-    }
-    else // workaround
-    {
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -7349,26 +7349,8 @@ Removes all qualifiers, if any, from type `T`.
...
-        else static if (is(T U == shared inout       U)) alias Unqual = U;
-        else static if (is(T U == shared       const U)) alias Unqual = U;
-        else static if (is(T U == shared             U)) alias Unqual = U;
-        else static if (is(T U ==        inout const U)) alias Unqual = U;
-        else static if (is(T U ==        inout       U)) alias Unqual = U;
-        else static if (is(T U ==              const U)) alias Unqual = U;
-        else                                             alias Unqual = T;
-    }
+    import core.internal.traits : CoreUnqual = Unqual;
+    alias Unqual = CoreUnqual!(T);
```
jacob-carlborg: You can probably simplify this with a public import.

andralex:
See how this works with ddoc:

```d
public import core.internal.traits : Unqual;
```

std/traits.d
andralex: Also try at top level: 

```d
import core.internal.traits : CoreUnqual = Unqual;
alias Unqual = CoreUnqual;
```

Look at the dox

andralex: Oops documentation discusses type `T` so alias or public import are not too good. Will approve as is.

wilzbach:
We don't we do this properly and move the entire traits package to druntime? Would be a good opportunity to clean it up too...

PR #6762, opened 2018-11-12
Remove duplicate of hasElaborateDestructor implementation

edi33416:
druntime's `core.internal.traits.hasElaborateDestructor` is an **almost** copy of the same code, as you can [see](https://github.com/dlang/druntime/blob/master/src/core/internal/traits.d#L167).

The druntime implementation accepts a list of `T`. I think that the Phobos alias should do the same. What do you think?

This PR removes the code duplication and the required double maintenance of `hasElaborateDestructor`.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -3749,20 +3749,8 @@ template hasElaborateAssign(S)
  */
 template hasElaborateDestructor(S)
 {
-    import std.meta : anySatisfy;
-    static if (isStaticArray!S && S.length)
-    {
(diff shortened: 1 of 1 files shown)


edi33416:
@n8sh I was thinking we should add the Phobos ddocs to the druntime implementations. Do you have a better solution than just copy-pasting the doc string?

thewilsonator:
Not off the top of my head.

thewilsonator:
@edi33416 the FreeBSD tester is dead, could you please give this a kick? I'll re-add auto merge.

PR #6763, opened 2018-11-12
Remove duplicate of anySatisfy implementation

edi33416:
druntime's `core.internal.traits.anySatisfy` is a copy of the same code. This PR removes the code duplication and the required double maintenance of `anySatisfy`.


n8sh:
@nordlow Yup: https://github.com/dlang/druntime/pull/2338

PR #6764, opened 2018-11-13
Fix Issue 19396 - ScopeBuffer can't be used in betterC with inline

rracariu:
fixing https://issues.dlang.org/show_bug.cgi?id=19396

### DIFF
--- a/std/internal/scopebuffer.d
+++ b/std/internal/scopebuffer.d
@@ -12,6 +12,7 @@ module std.internal.scopebuffer;
 
 import core.stdc.stdlib : realloc;
 import std.traits;
+import std.internal.attributes : betterC;
 
 /**************************************
  * ScopeBuffer encapsulates using a local array as a temporary buffer.
@@ -128,7 +129,7 @@ if (isAssignable!T &&
         this.bufLen = cast(uint) buf.length;
     }
 
-    @system unittest
+    @system @betterC unittest
     {
         ubyte[10] tmpbuf = void;
         auto sbuf = ScopeBuffer!ubyte(tmpbuf);
@@ -289,7 +290,7 @@ if (isAssignable!T &&
     }
 }
 
-@system unittest
+@system @betterC unittest
 {
     import core.stdc.stdio;
     import std.range;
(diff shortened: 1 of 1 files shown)


thewilsonator:
Oh, that makes more sense. 

Did you really mean to do this PR from your mater branch?

rracariu:
1. Again, this is `betterC` - there is no `throw` and no kinds of exception in `betterC` - no matter if ref-counted or GCed. So no, -dip1008 will not and cannot fix this! 
2. What thousandths lines of code, do you look at a different commit?  This one has 17 lines, most of them are unittest changes making sure this code compiles and runs in `betterC`

If this is not acceptable I can just close this PR and mind my own business, this is getting hilarious!

thewilsonator:
Also there is no druntime in `-betterC`.

thewilsonator:
(Remind me about this if I forget, the failing buildkite/style means this won't self-merge)

ZombineDev:
Using druntime *templates* in `D_BetterC` user code is totally fine as long as they don't cause link-time dependency on libdruntime / libphobos.
So the way to go is to make `onOutOfMemoryError` a template and put `version (D_Exceptions) { /* previous code */ } else assert(0, "some msg");` inside of it.

rracariu:
@ZombineDev see https://github.com/dlang/druntime/pull/2351

wilzbach:
> What thousandths lines of code, 

Well this is kind of a precedent and there are many other cases in Phobos were errors are thrown and which could also be patched like this here, but my opinion was just that the root cause (errors in -betterC) should be fixed or at least the next best thing `onOutOfMemoryError` directly as it has been proposed now.

wilzbach:
> (Remind me about this if I forget, the failing buildkite/style means this won't self-merge)

This should fix Buildkite: https://github.com/dlang/ci/pull/340

n8sh:
As https://github.com/dlang/druntime/pull/2361 has been merged this PR should no longer be necessary. If we manually mark issue 19396 as fixed will it show up in the changelog for the release?

rracariu:
I can change the import and keep the unittest for betterC

n8sh:
Keeping the unittests makes sense.

wilzbach:
> If we manually mark issue 19396 as fixed will it show up in the changelog for the release?

No. The changelog is generated by parsing the git commit messages.

PR #6765, opened 2018-11-14
make std.range.package compile with -dip1000

thewilsonator:
rebases #6396


wilzbach:
Error is still the same:

```
std/range/package.d(5122): Error: scope variable `r1` assigned to non-scope parameter `_param_0` calling std.range.zip!(NonSliceableRandomAccess, NonSliceableRandomAccess).zip
std/range/package.d(5122): Error: scope variable `r2` assigned to non-scope parameter `_param_1` calling std.range.zip!(NonSliceableRandomAccess, NonSliceableRandomAccess).zip
std/range/package.d(5123):        while evaluating: `static assert(isRandomAccessRange!(typeof(__error)))`
```

wilzbach:
No longer required.

PR #6769, opened 2018-11-15
Fix Issue 19403 - Make std.string.stripLeft on char array nogc nothrow

n8sh:


std/string.d
```diff
@@ -2986,20 +2986,19 @@ if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) &&
             // Fall through to standard case.
         }
 
-        static if (ElementType!Range.sizeof > ElementEncodingType!Range.sizeof)
+        static if (isNarrowString!Range)
         {
-            // Type performs its own decoding.
-            while (!input.empty)
+            import std.utf : decode, UseReplacementDchar;
```
edi33416: Duplicate of line 3002. Maybe just move the imports out of the `static if { .. } else { .. }` blocks, since it's in both.

thewilsonator:
Gah, what is Circle doing, or rather not doing?

std/string.d
n8sh: Done.

PR #6773, opened 2018-11-17
Remove MmapAllocator's dependence on private function fakePureErrno in core.memory

n8sh:


ZombineDev:
What's the motivation for doing so? `fakePureErrno` was added to druntime specifically for `std.experimental.allocator` and to avoid code duplication like this.

n8sh:
I want to change this because it's brittle and it means that what on the surface appears to be a private implementation detail in DRuntime can't be changed without breaking Phobos.

n8sh:
Changing it also means MmapAllocator could be used in betterC.

edi33416:
> I want to change this because it's brittle and it means that what on the surface appears to be a private implementation detail in DRuntime can't be changed without breaking Phobos.

I agree that this is not ideal, but imho code duplication only makes matters worse. One will need to know there are two `fakePureErrno` functions that need fixing if, for ex., there's a bug.

> Changing it also means MmapAllocator could be used in betterC.

Maybe I'm missing something, but it's still using `druntime`, isn't it? (`core.stdc.errno`)

n8sh:
>Maybe I'm missing something, but it's still using druntime, isn't it? (`core.stdc.errno`)

Generally not. The definition is usually something like:
```d
version (CRuntime_DigitalMars)
{
    extern (C)
    {
        ref int _errno();
        alias errno = _errno;
    }
}
```
The `extern (C)` means during linking the symbol being looked for will be `_errno`, which is provided by the C runtime. Only for the fallback case will the executable depend on the D runtime:

```d
version (CRuntime_DigitalMars) // ...
else version (CRuntime_Microsoft) // ...
else version (CRuntime_Glibc) // ...
else version (CRuntime_Musl) // ...
else version (FreeBSD) // ...
else version (DragonFlyBSD) // ...
else version (CRuntime_Bionic) // ...
else version (CRuntime_UClibc) // ...
else version (Darwin) // ...
else version (Solaris) // ...
else
{
    ///
    @property int errno() { return getErrno(); }
    ///
    @property int errno(int n) { return setErrno(n); }

    extern (C)
    {
        private int getErrno();      // for internal use
        private int setErrno(int);   // for internal use
    }
}
```

n8sh:
Here's a quick demonstration.
https://run.dlang.io/is/NUm2zr
```d
extern (C):
int main()
{
    import core.stdc.errno : errno;
    int result = cast(int) errno;
    // If you uncomment the next two lines linking fails with -betterC.
    // import core.memory : pureMalloc;
    // result += cast(int) pureMalloc(1);
    return result;
}
```

n8sh:
@ZombineDev For avoiding code duplication, would you consider it adequate if `fakePureErrno` were made `package(std)` and moved to the recently-added `std.internal.memory`? Nothing else in Phobos currently needs this but it would be in an obvious common location.

ZombineDev:
List of things that should be moved to druntime:

* `std.internal.memory` -> `core.memory`
* `std.conv.emplace` -> `core.lifetime`
* `std.experimental.allocator` -> `core.memory.allocator`

Why? So we can reimplement the built-in AAs and dynamic arrays as custom allocator-aware templates (as opposed to opaque `extern (C)` functions.

RazvanN7:
@PetarKirov I take it that this should be closed.

LightBender:
Closing this PR per @RazvanN7.

PR #6776, opened 2018-11-19
std.algorithm.cmp with default ordering can use memcmp for all size 1 unsigned types (instead of just char)

n8sh:


std/algorithm/comparison.d
```diff
@@ -619,6 +619,28 @@ Note:
 auto cmp(R1, R2)(R1 r1, R2 r2)
 if (isInputRange!R1 && isInputRange!R2)
 {
+    static if (isDynamicArray!R1 && isDynamicArray!R2 &&
+        is(R1 : E1[], E1) && is(R2 : E2[], E2) // just to get E1 & E2
+        && __traits(isUnsigned, E1) && __traits(isUnsigned, E2)
+        && E1.sizeof == E2.sizeof && !is(E1 == __vector) && !is(E2 == __vector))
```
n8sh:
```console
Running DScanner
generated/dscanner-e4ee46366d21eaa613da13c52f3f1a66684cdbaa/dsc --config .dscanner.ini --styleCheck etc std -I.
std/algorithm/comparison.d(625:56)[error]: Expected `(` instead of `)`
std/algorithm/comparison.d(625:56)[error]: Primary expression expected
```
This is complaining about `!is(E1 == __vector)`. It looks like DScanner needs to be fixed.

std/algorithm/comparison.d
```diff
@@ -619,6 +619,34 @@ Note:
...
+        enum endianForMemcmp = ElementEncodingType!R1.sizeof == 1;
+
+    static if (endianForMemcmp
+        && isDynamicArray!R1 && isDynamicArray!R2
+        && is(R1 : E1[], E1) && is(R2 : E2[], E2) // just to get E1 & E2
+        && __traits(isUnsigned, E1) && __traits(isUnsigned, E2)
+        && E1.sizeof == E2.sizeof
+        && (is(Unqual!E1 == char) == is(Unqual!E2 == char)) // Auto-decoding needs to match.
+        && !is(Unqual!E1 == wchar) && !is(Unqual!E2 == wchar) // Can't memcmp UTF-16 to get order.
+        && !is(E1 == __vector) && !is(E2 == __vector))
```
thewilsonator: Maybe scanner is expecting `!is(E1 == __vector(T1[N1),T1,N1)  && !is(E2 == __vector(T2[N2),T2,N2))`?

std/algorithm/comparison.d
thewilsonator: https://github.com/dlang-community/D-Scanner/issues/716

std/algorithm/comparison.d
```diff
@@ -644,66 +672,33 @@ if (isInputRange!R1 && isInputRange!R2)
...
-                    auto p1 = r1.ptr, p2 = r2.ptr,
-                        pEnd = p1 + min(r1.length, r2.length);
-                    for (; p1 != pEnd; ++p1, ++p2)
-                    {
-                        if (*p1 != *p2) return threeWayInt(int(*p1), int(*p2));
-                    }
-                    return threeWay(r1.length, r2.length);
-                }();
-            }
+                    if (*p1 != *p2) return cast(int) *p1 - cast(int) *p2;
```
stefan-koch-sociomantic: you should cache `*p1` and `*p2` 
the additional dereference may be expensive

thewilsonator:
(Blocked on dscanner - so I stop relooking at this PR)

wilzbach:
> I'm against this change. We should be removing C dependencies from Phobos, not finding more ways to plug them in. The language should be capable enough without them.

`memcmp` is already used in this function.

radcapricorn:
@wilzbach As I said, "removing, not finding **more** ways" ;)

n8sh:
Benchmark of this PR's change:

| range       | compiler         | speedup |
|-------------|------------------|---------|
| ubyte[16]   | dmd -O -inline   | 2.5x    |
| ubyte[128]  | dmd -O -inline   | 10.5x   |
| ubyte[16]   | ldc2 -O2         | 2.4x    |
| ubyte[128]  | ldc2 -O2         | 13.5x   |

For wider types whether this was an improvement depended on length, so I've restricted it to just byte-length types.

@radcapricorn instead of using `memcmp` directly I changed this PR to use `core.internal.string.dstrcmp` which internally uses `memcmp`. Does that satisfy your concerns?

std/algorithm/comparison.d
n8sh: @stefan-koch-sociomantic  I've verified that the compiler performs this optimization without needing any hint from the programmer. Both `cmp1` and `cmp1` in the example below are compiled to the same machine code by `ldc -O2`:

https://run.dlang.io/is/cz5hay
```d
import std.algorithm : min;

int cmp1(const dchar[] r1, const dchar[] r2) @trusted
{
    auto p1 = r1.ptr, p2 = r2.ptr, pEnd = p1 + min(r1.length, r2.length);
    for (; p1 != pEnd; ++p1, ++p2)
    {
        if (*p1 != *p2)
            return cast(int)*p1 - cast(int)*p2;
    }
    return int(r1.length > r2.length) - int(r1.length < r2.length);
}

int cmp2(const dchar[] r1, const dchar[] r2) @trusted
{
    auto p1 = r1.ptr, p2 = r2.ptr, pEnd = p1 + min(r1.length, r2.length);
    for (; p1 != pEnd; ++p1, ++p2)
    {
        const v1 = *p1, v2 = *p2; // manually saving result of *p1 and *p2
        if (v1 != v2)
            return cast(int) v1 - cast(int) v2;
    }
    return int(r1.length > r2.length) - int(r1.length < r2.length);
}

void main()
{
}
```

PR #6779, opened 2018-11-22
update dscanner

BBasile:


BBasile:
no idea.

thewilsonator:
Well that makes two of us. Excluding it will remove a large swathe of red from the CIs maybe even make them green.

wilzbach:
Why shouldn't it be run against etc?
Some checks are still useful and for the ones we don't care, they are simply disabled, e.g. https://github.com/dlang/phobos/blob/master/.dscanner.ini

Maybe a regression in Dscanner's selective filtering?

thewilsonator:
The CircleCI log is full of stuff like 
```
etc/c/zlib.d(268:23)[warn]: Parameter strm is never used.
etc/c/zlib.d(268:33)[warn]: Parameter flush is never used.
etc/c/zlib.d(381:26)[warn]: Parameter strm is never used.
etc/c/zlib.d(420:23)[warn]: Parameter strm is never used.
etc/c/zlib.d(420:33)[warn]: Parameter flush is never used.
...
etc/c/sqlite3.d(62:5)[warn]: Parameter zOptName isn't documented in the `Params` section.
etc/c/sqlite3.d(62:5)[warn]: A public function needs to contain a `Returns` section.
etc/c/sqlite3.d(64:18)[warn]: Parameter N isn't documented in the `Params` section.
etc/c/sqlite3.d(64:48)[warn]: Parameter N is never used.
etc/c/sqlite3.d(69:5)[warn]: A public function needs to contain a `Returns` section.
etc/c/sqlite3.d(85:5)[warn]: Parameter  isn't documented in the `Params` section.
etc/c/sqlite3.d(85:5)[warn]: A public function needs to contain a `Returns` section.
```

ZombineDev:
@thewilsonator style checks for `etc.*` should be ignored in https://github.com/dlang/phobos/blob/master/.dscanner.ini.

BBasile:
And what's going on buildkite ? Often when i take a look at PRs that have failures i can see an error saying the there's no VERSION file or that there's no `../druntime` folder.

BBasile:
There are more fixes requires in dparse, will reopen in a few days.

thewilsonator:
OK

PR #6780, opened 2018-11-23
[trivial] remove empty asm instructions from std.math

BBasile:


n8sh:
An empty statement after a label at the end of a block makes it look like those lines were added deliberately to work around some limitation.

BBasile:
ping @WalterBright, since you wrote the code we're talking about, why did you add an additional empty asm instruction after the x86 code for poly ?

PR #6781, opened 2018-11-23
Fix Issue 19429 - indexOf("a", "b") should be nothrow/@nogc

CyberShadow:
Please see the individual commits' commit messages for details.

### DIFF
--- a/std/string.d
+++ b/std/string.d
@@ -766,20 +766,28 @@ private template _indexOfStr(CaseSensitive cs)
 
         static if (isSomeString!Range)
         {
-            import std.algorithm.searching : find;
-
-            const(Char1)[] balance;
-            static if (cs == Yes.caseSensitive)
+            static if (is(Char1 == Char) && cs == Yes.caseSensitive)
             {
-                balance = find(s, sub);
+                import std.algorithm.searching : countUntil;
+                return s.representation.countUntil(sub.representation);
             }
             else
             {
-                balance = find!
-                    ((a, b) => toLower(a) == toLower(b))
-                    (s, sub);
+                import std.algorithm.searching : find;
+
+                const(Char1)[] balance;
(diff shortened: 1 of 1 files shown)


CyberShadow:
This is a non-trivial change, let's wait at least 24 hours for reviews.

std/string.d
```diff
@@ -871,20 +871,44 @@ private template _indexOfStr(CaseSensitive cs)
...
+        in size_t startIdx)
+@safe
+if (isSomeChar!Char1 && isSomeChar!Char2)
+{
+    if (startIdx < s.length)
+    {
+        ptrdiff_t foundIdx = indexOf(s[startIdx .. $], sub);
+        if (foundIdx != -1)
+        {
+            return foundIdx + cast(ptrdiff_t) startIdx;
```
UplinkCoder: you cloud do foundIdx += cast(ptrdiff_t) startIdx;

and then return foundIdx at the end, which would safe one ret instruction and some branch predictor pain,

CyberShadow: Also, this is not my code. The function was duplicated in order to allow it to have its own inferred attributes, due to case sensitivity no longer being variable at runtime.

UplinkCoder: Hmm dmd amazes es always :)

UplinkCoder: my suggestion was different though
```
    ptrdiff_t foundIdx = -1;
    if (startIdx < s.length)
    {
        foundIdx = indexOf(s[startIdx .. $], sub);
        if (foundIdx != -1)
        {
            foundIdx += cast(ptrdiff_t) startIdx;
        }    
    }
    return foundIdx;
```

CyberShadow: Tried that too, still more instructions...

CyberShadow:
BuildKite failure looks unrelated.

PR #6777, opened 2018-11-21
Remove `AliasSeq` when following `__traits` is directly aliasable

BBasile:
~~There's just one remaining `AliasSeq!(__traits` but it's not aliasable (`getAliasThis`).~~ 

edit: it will be after fix for blocker merged.

PR #6783, opened 2018-11-27
update dscanner

BBasile:
Necessary to fix some issues encountered while linting (`__vector`, empty asm instructions) and also to support `__traits` as type.


thewilsonator:
Need to Dscanner rules to exclude the following warning rules in the following files 
```
etc/c/odbc/sql.d(609:27)[warn]: Parameter HandleType is never used.
etc/c/odbc/sqlext.d(1688:28)[warn]: Parameter hdbc is never used.
etc/c/odbc/sqlucode.d(42:18)[warn]: Parameter hstmt is never used.
etc/c/zlib.d(268:23)[warn]: Parameter strm is never used.
etc/c/curl.d(1979:22)[warn]: A public function needs to contain a `Returns` section.
etc/c/curl.d(1979:22)[warn]: Parameter curl isn't documented in the `Params` section.
etc/c/sqlite3.d(62:5)[warn]: Parameter zOptName isn't documented in the `Params` section.
etc/c/sqlite3.d(62:5)[warn]: A public function needs to contain a `Returns` section.
etc/c/sqlite3.d(64:48)[warn]: Parameter N is never used.
std/windows/charset.d(33:45)[warn]: Parameter s is never used.
std/windows/syserror.d(32:15)[warn]: Parameter errCode is never used.
```
The following should be fixed or ignored
```
std/internal/memory.d(56:58)[warn]: Parameter nmemb is never used.
std/internal/memory.d(56:72)[warn]: Parameter size is never used.
std/internal/memory.d(57:71)[warn]: Parameter size is never used.
std/stdio.d(196:24)[warn]: Parameter size is never used.
std/stdio.d(196:37)[warn]: Parameter n is never used.
std/stdio.d(735:35)[warn]: Parameter handle is never used.
std/stdio.d(735:63)[warn]: Parameter stdioOpenmode is never used.
```

BBasile:
There's a crash that was supposed to be fixed to fix for good anyway.

thewilsonator:
Oh, I see that now.

wilzbach:
BTW it could be that `.dscanner.ini` wasn't copied over on Buildkite, so https://github.com/dlang/ci/pull/351 might fix the filtering rules.

wilzbach:
> std/windows/charset.d(33:45)[warn]: Parameter s is never used.
> std/windows/syserror.d(32:15)[warn]: Parameter errCode is never used.

That looks like sth. that could be fixed on the Dscanner side too as the never-used check should ignore declarations without a body.

BBasile:
> BTW it could be that .dscanner.ini wasn't copied over on Buildkite, so dlang/ci#351 might fix the filtering rules.

No it's filtering that was not complete (but why it worked before ?). Now that rules are adjusted there just 1 remaining warning.

> That looks like sth. that could be fixed on the Dscanner side too as the never-used check should ignore declarations without a body.

Opens an issue for this. For now things will be filtered.

BBasile:
Ah yes i see now, declarations in `version(Stddoc)` blocks are problematic

wilzbach:
> > That looks like sth. that could be fixed on the Dscanner side too as the never-used check should ignore declarations without a body.
> 
> Opens an issue for this. For now things will be filtered.

https://github.com/dlang-community/D-Scanner/issues/727

.dscanner.ini
```diff
@@ -458,19 +458,18 @@ unused_variable_check="-std.algorithm.comparison,\
 -std.range,\
 -std.range.interfaces,\
 -std.range.primitives,\
--std.regex.internal.backtracking,\
--std.regex.internal.tests,\
--std.regex.internal.thompson,\
+-std.regex.internal.,\
```
wilzbach: Needs `std.path` here until https://github.com/dlang-community/D-Scanner/issues/727 is fixed.

due to:

```
std/path.d(2659:26)[warn]: Parameter path is never used.
```

BBasile:
I think it's because the adaptation to correct DIP1009 is incomplete. missing func body is not tested.
Gotta fix that in D-Scanner rather than add all these new rules.

BBasile:
I think manual merge is required here, buildkite cant pass for now since it uses the old d-scanner hash.

thewilsonator:
Done.

BBasile:
Thanks. Will rebase #6777 now.

PR #6784, opened 2018-11-29
Fix Issue 19444: hasRawAliasing loops on class with static array

radcapricorn:


thewilsonator:
Please target stable.

radcapricorn:
Does it not auto-close the bugzilla when targeting stable, or just needs some time?

thewilsonator:
No, only when it is merged back into master.

wilzbach:
Yup, someone needs to run [`merge_stable`](https://github.com/wilzbach/git-tools/blob/master/merge_stable) or we finally add this to dlang-bot, s.t. it's done automatically.

wilzbach:
Though thinking about this, we should probably change the Bugzilla branch to `stable` as that's the release branch.

PR #6787, opened 2018-12-01
[deduplication] `RefCounted` should use `moveEmplace`...

radcapricorn:
...as per a comment in its implementation.


radcapricorn:
Don't think that's needed:

https://github.com/dlang/phobos/blob/0a4740ac3da2efff02e73cdea75a95c5ae52a8f7/std/typecons.d#L6075-L6076

Should mark `nothrow pure` though...

PR #6790, opened 2018-12-04
[WIP] Android and AArch64: tweak tests that trip

joakim-noah:
I'm still looking at some alignment issues with std.variant because Android/x64 uses an IEEE Quad 128-bit 'long double`, will finish this up once I track that down.

I'm targeting stable because we'll just pull these into the final release of ldc 1.3 when the final point release of dmd 2.083 is out. Only the `/proc/cpuinfo` change has the potential to affect the dmd release, since most everything else is versioned out for the official releases.


thewilsonator:
Also did you mean to create a branch for this pr rather that from your `stable`?

std/math.d
```diff
@@ -4110,7 +4110,7 @@ real log(real x) @safe pure nothrow @nogc
 ///
 @safe pure nothrow @nogc unittest
 {
-    assert(log(E) == 1);
+    assert(feqrel(log(E), 1) >= real.mant_dig - 1);
```
joakim-noah: The assert for equality only trips on Android, but not on linux/glibc/AArch64. Since it's a documented test, I can version this only for IEEE Quadruples like the `std.complex` test above and keep this simpler assert for readability if wanted.

std/math.d
```diff
@@ -4616,6 +4616,14 @@ real hypot(real x, real y) @safe pure nothrow @nogc
 
     static assert(2*(SQRTMAX/2)*(SQRTMAX/2) <= real.max);
 
+    static if (floatTraits!(real).realFormat == RealFormat.ieeeQuadruple)
+    {
+        // This static assert overflows when cross-compiling from lower-precision
+        // floating-point reals at compile-time to higher precision at runtime.
```
joakim-noah: As noted, this isn't an issue just for Quadruples, but also hits when say the compile-time real is 64-bit and runtime is 80-bit, as I just tested with ldc on Android/ARM cross-compiling to linux/x64 (a Win/MSVC target works however, as it also uses 64-bit reals).

I'm unsure how to check the compile-time real's precision however, so I just settled for disabling this test whenever compiling to a Quad runtime real. Anyone know how to get the compile-time real's precision?

joakim-noah:
I don't have a stable branch on my Phobos fork, deleted it a while back and just pushed this now for this pull, after which I'll delete it again. :wink:

std/math.d
joakim-noah: @ibuclaw or @jpf91, any input?

kinke: As noted in the forum, there's no way of detecting the host's compile-time precision. Compiled code shouldn't have to know, and an assert makes perfect sense to guarantee the precision is sufficiently high. It clearly isn't with current LDC in all cross-compilation scenarios, as we don't have a software `real_t` yet.

So maybe weakening the `static assert` to a runtime `enum <actual-ct-value> = ...; assert(...)` would be a solution for this problem, potentially restricted to LDC only.

ibuclaw: Doesn't affect gdc, as we have used soft float since the start.

ibuclaw: These static asserts assume IEEE though. Non-IEEE reals could (and infact do in the case of ibm real) throw a compile time assert.

joakim-noah: Alright, since this doesn't affect dmd, because it only supports Intel, or gdc, because it uses a wider-precision soft-float at compile-time, I'll keep this in ldc: removed here.

thewilsonator:
This good to go?

joakim-noah:
Almost, but no hurry now that the point release shipped.

std/variant.d
```diff
@@ -102,6 +102,23 @@ struct This;
 
 private alias This2Variant(V, T...) = AliasSeq!(ReplaceType!(This, V, T));
 
+// copied from std.traits and modified the base case
```
kinke: You could reduce the code duplication by simply checking for an empty tuple (and returning the pointer-alignment then, which seems like a valid guess, although `max(size_t.alignof, real.alignof)` would be safer), and otherwise use the `std.traits` version.

std/variant.d
```diff
@@ -167,7 +184,7 @@ private:
     // state
     ptrdiff_t function(OpID selector, ubyte[size]* store, void* data) fptr
         = &handler!(void);
-    union
+    align(maxAlignment!(AllowedTypes)) union
     {
         ubyte[size] store;
```
kinke: If you move the `align()` down here, you won't change the alignment for `AllowedTypes = { int[2], short[4] }` for 64-bit targets (so that it would still be 8-bytes pointer-aligned due to the pointers array). It doesn't really matter, as the function pointer field makes sure the whole `VariantN` struct is still at least pointer-aligned and that the union's offset is 8; it'd just be good practice.

std/traits.d
```diff
@@ -4759,7 +4759,7 @@ template TemplateArgsOf(T : Base!Args, alias Base, Args...)
 }
 
 
-private template maxAlignment(U...)
+package template maxAlignment(U...)
```
joakim-noah: This protection plus the modification was why I copied it before.

std/variant.d
joakim-noah: Made the `std.traits` version `package`, so I can reuse it like you said.

As for setting the maximum alignment of `real` or `size_t`, that hits the Android/x64 segfaults I mentioned in the ldc pull, when using `align(16)` for the union. It's using 128-bit alignment then, as the `real` alignment is always larger than 64-bit `size_t` there, and then segfaults when trying to `MOVAPS` a 64-bit aligned location into the FP register `$xmm0`. With `size_t` alignment, it uses a `MOVUPS` instead.

Another case where LLVM isn't expecting that `__float128` alignment?

std/variant.d
joakim-noah: Don't really want to think about these alignment issues, just moved it down like you said: works on Android/x64.

joakim-noah:
Single dmd testsuite failure on linux seems spurious, since it passes everywhere else. Once the yaml-d issue is resolved, should be ready to go: removed the WIP title.

std/variant.d
kinke: > and then segfaults when trying to MOVAPS a 64-bit aligned location into the FP register $xmm0

That's exactly what shouldn't happen when aligning the union at 16 bytes. What that does though is change the min size of the union to 16 bytes, which might be unexpected for some test.

I mentioned `max(size_t.alignof, real.alignof)` for an empty `AllowedTypes` tuples (=> arbitrary types) as that should make sure it can be used as `real` too on platforms like Android x64. I'm pretty sure you'll hit alignment issues otherwise, just like the ones you currently get when using that max, for some bizarre reason.

thewilsonator:
This trips an assert in `dyaml`
`../.dub/packages/dyaml-0.7.0/dyaml/source/dyaml/node.d(188,9): Error: static assert:  "Unexpected YAML value size"`
```d
struct Node
{
    ...
    alias Value = Algebraic!(YAMLNull, YAMLMerge, bool, long, real, ubyte[], SysTime, string,
                         Node.Pair[], Node[], YAMLObject);
    ...
    Value value_;
    ...
    static assert(Value.sizeof <= 24, "Unexpected YAML value size");
}
```
because value is, with this change effectively 
```d
struct 
{
    void* functionPointer;
    align(16) ubyte[16];
}
```
for 32 bytes.

kinke:
That's a bug in dyaml then. ;) - The layout can be optimized to `struct VariantN { union { ... }; void* functionPointer }`, that'll save padding.

std/variant.d
joakim-noah: So we're clear, [the original failing test](https://github.com/ldc-developers/phobos/blob/fd5facfe0456d8f7ae16af8415f1ea01606c917f/std/variant.d#L1317) passes now: it's [the other tests that start to fail](https://github.com/ldc-developers/phobos/blob/fd5facfe0456d8f7ae16af8415f1ea01606c917f/std/variant.d#L39) if I use `real.alignof` when no types are passed to `Variant`, as [I mentioned before](https://github.com/ldc-developers/ldc/pull/2917#issuecomment-445751275). Looking into it more in a debugger and after hacking that first test block down to just two lines, `Variant b = 42; assert(b.type == typeid(int));`, it loads a 64-bit-aligned `_D3std7variant__T8VariantNVmi32ZQp6__initZ` and tries to `MOVAPS` the first 128-bit word and fails.

If I use this pull with `size_t.alignof` instead, all the `std.variant` tests pass, and it loads the FP registers from that same 64-bit aligned initializer with `MOVUPS` instructions instead.

I honestly don't care how this stuff is aligned, as long as the tests pass. :wink:

kinke: > it loads a 64-bit-aligned _D3std7variant__T8VariantNVmi32ZQp6__initZ

That seems to be the issue then - our (LDC) init symbols seem to violate the alignment.

kinke: Well that would have been quite surprising, so I checked - and it looks fine: https://run.dlang.io/is/q45ykS

```LL
@onlineapp.VariantN.__init = constant %onlineapp.VariantN zeroinitializer, comdat, align 16
```

So I guess it could be a linker problem.

std/variant.d
```diff
@@ -102,6 +102,20 @@ struct This;
...
+// We can't just use maxAlignment because no types might be specified
+// to VariantN, so handle that here and then pass along the rest.
+private template maxVariantAlignment(U...)
+if (isTypeTuple!U)
+{
+    static if (U.length == 0)
+        enum maxVariantAlignment = size_t.alignof;
+    else
+    {
+        import std.traits : maxAlignment;
```
kinke: There's a global `import std.traits` at the top of the file.

joakim-noah: Removed.

std/variant.d
kinke: Sorry, I wasn't really focused - `movups` is of course used for `align(8)`. `movaps` should be fine, and the reduced testcase suggests LDC does the right thing by requesting 16-bytes alignment for the init symbol -> I guess it's really a linker issue, wouldn't be the first of this kind.

joakim-noah: Alright, trying it.

BBasile:
I looks like a problem with buildkite. Since your last force-push i have updated D:YAML and relaunched the item manually, with success. So did i after dlang-tour new tag and this strange error started appearing for no reasons.

Maybe amend to get a new commit hash and force push will fix this ?

BBasile:
The problem is really related to dlang-tour, forced push won't help.

thewilsonator:
Yeah dlang/tour fails on other build as well. Is this still WIP?

joakim-noah:
Looks like the problem is the shallow clone, strange that that would affect the submodule. Just tried it and not reproducible for me on my Android tablet, running the same two commands for dlang-tour plus its English version.

std/variant.d
joakim-noah: Works on the auto-tester, could the Android/x64 linker be why `std.variant` passed its tests when I used `align(16)` and linked that module alone, but failed when I linked it into the test runner? If you think it's more correct this way, fine with me.

joakim-noah:
> Is this still WIP?

Looks done to me.

wilzbach:
> Looks like the problem is the shallow clone, strange that that would affect the submodule. Just tried it and not reproducible for me on my Android tablet, running the same two commands for dlang-tour plus its English version.

See https://github.com/dlang/ci/pull/363 which should hopefully fix Buildkite.
Anyhow, the failure is unrelated.

std/variant.d
kinke: Ah, very nice then.

joakim-noah:
Alright, fixed the Buildkite issue, but the build agents seem flakier and slow now: some hardware changes going on right now?

thewilsonator:
Yeah I noticed that on a few dmd PRs recently, no idea.

wilzbach:
@MartinNowak recently changed all agents to be auto-provisioned depending on the current load.

joakim-noah:
OK, ready to merge, CI passes.

PR #6792, opened 2018-12-05
inure against windows.d changes by casting to LPCWSTR

WalterBright:


CyberShadow:
I can't sign off on this.

What happens if you define `FSChar` as `WCHAR` instead?

CyberShadow:
@thewilsonator Are you OK with such noisy / intrusive changes being forced onto D code?

WalterBright:
> What happens if you define FSChar as WCHAR instead?

Then existing code won't compile, as wchar_t is becoming a different type from wchar.

CyberShadow:
Well, right, but it won't compile if you don't change anything either. What else would need to be changed if you change `FSChar` to `WCHAR`, is it possibly a smaller diff than this?

WalterBright:
> FSChar is not public and is not in any public functions' signatures.

Looks like you're right.

std/mmfile.d
```diff
@@ -219,7 +219,7 @@ class MmFile
 
             if (filename != null)
             {
-                hFile = CreateFileW(filename.tempCStringW(),
+                hFile = CreateFileW(cast(LPCWSTR) filename.tempCStringW(),
```
CyberShadow: Change `tempCStringW` to return `LPWSTR` instead of `wchar*`?

std/mmfile.d
```diff
@@ -242,7 +242,7 @@ class MmFile
 
             int hi = cast(int)(size >> 32);
             hFileMap = CreateFileMappingW(hFile, null, flProtect,
-                    hi, cast(uint) size, null);
+                    hi, cast(uint) size, cast(LPCWSTR) null);
```
CyberShadow: Why is it necessary to cast `null`?

std/process.d
```diff
@@ -3092,7 +3093,7 @@ private:
         import std.array : array;
         LPWSTR lpCommandLine = (to!(wchar[])(line) ~ "\0"w).ptr;
```
CyberShadow: `to!(WCHAR[])`? Will that work?

thewilsonator:
> Are you OK with such noisy / intrusive changes being forced onto D code?

Ideally no, but this PR is a stop-gap until I can figure out why https://github.com/dlang/dmd/pull/8958 doesn't work and fix it, any help is much appriciated.

std/mmfile.d
WalterBright: It's not necessary, it's just a reminder in case someone wants to add the parameter.

std/mmfile.d
WalterBright: Perhaps, but I don't know what other things tempCStringW might be used for. It's defined in internal/cstring.d, and may be worth changing.

std/mmfile.d
CyberShadow: I don't get it, but OK.

Maybe subjective, but I don't like using `cast` unless it's strictly necessary, because of how powerful it is and how many kinds of things it can do. So perhaps `LPCWSTR(null)` or `LPCWSTR.init` or `/*LPCWSTR*/ null`.

std/mmfile.d
CyberShadow: I think it would be a very safe bet to assume that `tempCStringW` was created to help interfacing with Unicode Windows APIs.

std/process.d
WalterBright: It'll probably fail to concat with a wchar[] string. I was going to fix dmd to implicitly convert string literals to wchar_t[], but those efforts have hit a brick wall with the bikeshedding over a switch to enable the new behaviors. My goal with this PR anyway is to have it compile regardless of the switch setting.

CyberShadow: > It'll probably fail to concat with a wchar[] string.

Generally: Argh.

Specifically: you can still change this one to `WCHAR(0)`.

> My goal with this PR anyway is to have it compile regardless of the switch setting.

I think changes to make things pedantically correct are fine whether or not the compiler change happens, as long as it doesn't add more code or uglify the existing one.

thewilsonator: > I was going to fix dmd to implicitly convert string literals to wchar_t[]

No need it already does https://github.com/dlang/dmd/blob/master/test/compilable/implicitconv.d#L5

std/mmfile.d
WalterBright: I understand, I'll yank it.

CyberShadow:
> ok, we'll see how it fares

My guess is that it is not going to just work.

And that brings up a bigger problem, which I think is a conceptual blocker for this whole endeavor:

How is Phobos (std.algorithm etc.) supposed to treat this new `wchar_t` type?

Is it supposed to treat it like any D char (and `wchar_t[]` like any D character range)? Then, at least the definition of `StringTypeOf` will need to be updated to recognize it.

Or, is it supposed to be an alien C type that D-ish parts of Phobos won't touch? Then, that will make interop with C APIs that use `wchar_t` less smooth, and explicit casts will be required.

WalterBright:
My opinion about `wchar_t` is it's a stupid idea, and a complete failure for C and C++ (which is why they had to try again with `char16_t` and `char32_t`). D got it right.

Therefore, D users should avoid `wchar_t` as much as possible, and only use it when interfacing to C++ code that uses it. They can use a cast as the last step to get the type right. And in fact, the only reason to use it at all is to match the name mangling C++ uses for it.

So if `wchar_t` fails to work with some Phobos functions, don't use it. (My attitude about `c_long` and `c_ulong` is the same. Don't use them in normal D code.)

That leaves the Windows API, which Microsoft erred by using `wchar_t`. People should use `WCHAR` to interface to Windows, and only cast code to WCHAR at the last minute.

thewilsonator:
`std\internal\cstring.d(178): Error: undefined identifier WCHAR`

CyberShadow:
So did `to!(WCHAR[])` work for you (with the compiler change)?

std/process.d
```diff
@@ -3090,9 +3091,9 @@ private:
     {
         import std.algorithm.iteration : map;
         import std.array : array;
-        LPWSTR lpCommandLine = (to!(wchar[])(line) ~ "\0"w).ptr;
+        auto lpCommandLine = (to!(WCHAR[])(line) ~ '\0').ptr;
```
thewilsonator: Presumably.

PR #6793, opened 2018-12-05
Fix issue 19466 - functionLinkage documentation omits some values

jsatellite:
std.traits.functionLinkage documentation doesn't show the complete list of possible linkage attributes. See https://issues.dlang.org/show_bug.cgi?id=19466


thewilsonator:
Thanks!

PR #6795, opened 2018-12-06
Fix Issue 10706 setops functions need to require SortedRanges

burner:
https://issues.dlang.org/show_bug.cgi?id=10706

* largestPartialIntersection
* largestPartialIntersectionWeighted
* multiwayMerge
* multiwayUnion
* setDifference
* setIntersection
* setSymmetricDifference

Now all require SortedRanges to be passed, as that is what they expect
implicitly anyway.

A new public symbol `isSortedRange` was added to std.range.

### DIFF
--- a/changelog/pr6795.dd
+++ b/changelog/pr6795.dd
@@ -1,0 +1,33 @@
+Fix Issue 10706: std.algorithm.setops functions require SortedRange arguments.
+
+The functions
+
+$(UL
+    $(LI largestPartialIntersection)
+    $(LI largestPartialIntersectionWeighted)
+    $(LI multiwayMerge)
+    $(LI multiwayUnion)
+    $(LI setDifference)
+    $(LI setIntersection)
+    $(LI setSymmetricDifference)
+)
+
+of $(MREF std,algorithm,setops) require that the elements of the passed
+Range/Ranges are sorted.
+This is easily overlooked as stated in
+https://issues.dlang.org/show_bug.cgi?id=10706
+
+Due to this change, ranges passed to any of the above listed function need to
+be of type $(REF SortedRange, std,algorithm,sorting).
+A `SortedRange` can either be obtainted by $(REF sort, std,algorithm,sorting)
+or $(REF assumeSorted, std,range).
+
+-------
+import std.algorithm : equal, sort;
+import std.range : assumeSorted;
+
+// sets
+auto a = sort([ 1, 2, 4, 5, 7, 9 ]);
+auto b = assumeUnique([ 0, 1, 2, 4, 7, 8 ]);
+auto intersection = setIntersection(assumeSorted(a), assumeSorted(b));
+-------
--- a/std/algorithm/setops.d
+++ b/std/algorithm/setops.d
@@ -53,7 +53,8 @@ import std.functional : unaryFun, binaryFun;
 import std.traits;
 import std.meta : AliasSeq, staticMap, allSatisfy, anySatisfy;
 
-import std.algorithm.sorting : Merge;
+import std.algorithm.sorting : Merge, sort;
+import std.range : ElementEncodingType, SortedRange, assumeSorted;
 import std.typecons : No;
 
 // cartesianProduct
@@ -643,13 +644,13 @@ void largestPartialIntersection
 
     // Figure which number can be found in most arrays of the set of
     // arrays below.
-    double[][] a =
+    auto a =
     [
-        [ 1, 4, 7, 8 ],
-        [ 1, 7 ],
-        [ 1, 7, 8],
-        [ 4 ],
-        [ 7 ],
+        assumeSorted([ 1.0, 4.0, 7.0, 8.0 ]),
+        assumeSorted([ 1.0, 7.0 ]),
+        assumeSorted([ 1.0, 7.0, 8.0]),
+        assumeSorted([ 4.0 ]),
+        assumeSorted([ 7.0 ]),
     ];
     auto b = new Tuple!(double, uint)[1];
     // it will modify the input range, hence we need to create a duplicate
@@ -666,13 +667,13 @@ void largestPartialIntersection
     // 1.0 occurs in 3 inputs
 
     // multiset
-    double[][] x =
+    auto x =
     [
-        [1, 1, 1, 1, 4, 7, 8],
-        [1, 7],
-        [1, 7, 8],
-        [4, 7],
-        [7]
+        assumeSorted([1.0, 1.0, 1.0, 1.0, 4.0, 7.0, 8.0]),
+        assumeSorted([1.0, 7.0]),
+        assumeSorted([1.0, 7.0, 8.0]),
+        assumeSorted([4.0, 7.0]),
+        assumeSorted([7.0])
     ];
     auto y = new Tuple!(double, uint)[2];
     largestPartialIntersection(x.dup, y);
@@ -726,13 +727,13 @@ void largestPartialIntersectionWeighted
 
     // Figure which number can be found in most arrays of the set of
     // arrays below, with specific per-element weights
-    double[][] a =
+    auto a =
     [
-        [ 1, 4, 7, 8 ],
-        [ 1, 7 ],
-        [ 1, 7, 8],
-        [ 4 ],
-        [ 7 ],
+        assumeSorted([ 1.0, 4.0, 7.0, 8.0 ]),
+        assumeSorted([ 1.0, 7.0 ]),
+        assumeSorted([ 1.0, 7.0, 8.0]),
+        assumeSorted([ 4.0 ]),
+        assumeSorted([ 7.0 ]),
     ];
     auto b = new Tuple!(double, uint)[1];
     double[double] weights = [ 1:1.2, 4:2.3, 7:1.1, 8:1.1 ];
@@ -742,14 +743,14 @@ void largestPartialIntersectionWeighted
     // 4.0 occurs 2 times -> 4.6 (2 * 2.3)
     // 7.0 occurs 3 times -> 4.4 (3 * 1.1)
 
-   // multiset
-    double[][] x =
+    // multiset
+    auto x =
     [
-        [ 1, 1, 1, 4, 7, 8 ],
-        [ 1, 7 ],
-        [ 1, 7, 8],
-        [ 4 ],
-        [ 7 ],
+        assumeSorted([ 1.0, 1.0, 1.0, 4.0, 7.0, 8.0 ]),
+        assumeSorted([ 1.0, 7.0 ]),
+        assumeSorted([ 1.0, 7.0, 8.0]),
+        assumeSorted([ 4.0 ]),
+        assumeSorted([ 7.0 ]),
     ];
     auto y = new Tuple!(double, uint)[1];
     largestPartialIntersectionWeighted(x, y, weights);
@@ -762,13 +763,13 @@ void largestPartialIntersectionWeighted
     import std.conv : text;
     import std.typecons : tuple, Tuple, Yes;
 
-    double[][] a =
+    auto a =
         [
-            [ 1, 4, 7, 8 ],
-            [ 1, 7 ],
-            [ 1, 7, 8],
-            [ 4 ],
-            [ 7 ],
+            assumeSorted([ 1.0, 4.0, 7.0, 8.0 ]),
+            assumeSorted([ 1.0, 7.0 ]),
+            assumeSorted([ 1.0, 7.0, 8.0]),
+            assumeSorted([ 4.0 ]),
+            assumeSorted([ 7.0 ]),
         ];
     auto b = new Tuple!(double, uint)[2];
     largestPartialIntersection(a, b, Yes.sortOutput);
@@ -781,13 +782,13 @@ void largestPartialIntersectionWeighted
     import std.conv : text;
     import std.typecons : tuple, Tuple, Yes;
 
-    string[][] a =
+    auto a =
         [
-            [ "1", "4", "7", "8" ],
-            [ "1", "7" ],
-            [ "1", "7", "8"],
-            [ "4" ],
-            [ "7" ],
+            assumeSorted([ "1", "4", "7", "8" ]),
+            assumeSorted([ "1", "7" ]),
+            assumeSorted([ "1", "7", "8"]),
+            assumeSorted([ "4" ]),
+            assumeSorted([ "7" ]),
         ];
     auto b = new Tuple!(string, uint)[2];
     largestPartialIntersection(a, b, Yes.sortOutput);
@@ -800,14 +801,14 @@ void largestPartialIntersectionWeighted
 
     // Figure which number can be found in most arrays of the set of
     // arrays below, with specific per-element weights
-    double[][] a =
+    auto a =
         [
-            [ 1, 4, 7, 8 ],
-            [ 1, 7 ],
-            [ 1, 7, 8],
-            [ 4 ],
-            [ 7 ],
-            ];
+            assumeSorted([ 1.0, 4.0, 7.0, 8.0 ]),
+            assumeSorted([ 1.0, 7.0 ]),
+            assumeSorted([ 1.0, 7.0, 8.0]),
+            assumeSorted([ 4.0 ]),
+            assumeSorted([ 7.0 ]),
+        ];
     auto b = new Tuple!(double, uint)[1];
     double[double] weights = [ 1:1.2, 4:2.3, 7:1.1, 8:1.1 ];
     largestPartialIntersectionWeighted(a, b, weights);
@@ -825,6 +826,22 @@ void largestPartialIntersectionWeighted
     const Array!T arrayTwo = Array!T([ T(1,2), T(3,4) ] );
 
     assert(arrayOne == arrayTwo);
+}
+
+
+private MultiwayMerge!(less, RangeOfRanges) multiwayMergeSortedRange
+(alias less = "a < b", RangeOfRanges)
+(RangeOfRanges ror)
+{
+    return typeof(return)(ror);
+}
+
+deprecated("multiwayMerge requires the passed ranges to be sorted")
+private MultiwayMerge!(less, RangeOfRanges) multiwayMergeUnknownRange
+(alias less = "a < b", RangeOfRanges)
+(RangeOfRanges ror)
+{
+    return typeof(return)(ror);
 }
 
 // MultiwayMerge
@@ -930,7 +947,14 @@ MultiwayMerge!(less, RangeOfRanges) multiwayMerge
 (alias less = "a < b", RangeOfRanges)
 (RangeOfRanges ror)
 {
-    return typeof(return)(ror);
+    static if (isSortedRange!(ElementEncodingType!(RangeOfRanges)))
+    {
+        return multiwayMergeSortedRange!less(ror);
+    }
+    else
+    {
+        return multiwayMergeUnknownRange!less(ror);
+    }
 }
 
 ///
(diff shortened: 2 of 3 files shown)


thewilsonator:
LGTM, given the possible fixing breakage that this will incur, this needs a changelog.

std/range/package.d
```diff
@@ -10923,6 +10923,21 @@ that break its sorted-ness, `SortedRange` will work erratically.
     a.sort.sort!"a > b";
 }
 
+/**
+Determines whether `T` is a $(D SortedRange).
```
wilzbach: You could use LREF here.

burner: fixed

burner:
changelog added

burner:
changelog added

thewilsonator:
Nice idea!

burner:
I will do the suggested change, but SortedRange will check for sortedness in debug. So this is still a breaking change.

But from 
> Assuming for the sake of argument that this should go ahead
it sounds like an opinion has already been formed regarding this PR. If this would be my first PR it would be my last. It's the tone that makes the music.

n8sh:
I'm sorry it came across that way. I think the red X makes the tone seem more negative than intended.

burner:
@n8sh I didn't notice the red X until you just mentioned it.

Anyway, what is the way forward with this.

n8sh:
>Anyway, what is the way forward with this.

You said you intend to do the suggested change, so when that is done this should be ready to go ahead.

burner:
I already added the overloads.

thewilsonator:
> I already added the overloads.

Could you please point me to them.

> But I can't mark them as deprecated as phobos will fail to build.

Bummer.

wilzbach:
> But I can't mark them as deprecated as phobos will fail to build.

Well, all tests that use them need to be marked as deprecated too then.

burner:
Ok, I'll mark the tests as deprecated and see what happens.

burner:
All tests that rely on the non SortedRange input have been marked deprecated.
I can't mark the non SortedRange version of the functions deprecated as they call each other.

I'm inclined to add a comment saying the will go away at some point in the future.

changelog/pr6795.dd
```diff
@@ -0,0 +1,31 @@
...
+
+The functions
+
+* largestPartialIntersection
+* largestPartialIntersectionWeighted
+* multiwayMerge
+* multiwayUnion
+* setDifference
+* setIntersection
+* setSymmetricDifference
```
wilzbach: This is ddoc, not markdown. See DAutolist. Until ddoc doesn't support markdown, you'll have to use `$(UL)` and `$(LI)`

changelog/pr6795.dd
```diff
@@ -0,0 +1,31 @@
...
+
+of $(REF std,algorithm,setops) require that the elements of the passed
```
wilzbach: MREF

changelog/pr6795.dd
burner: I knew that at some point.
Should be fixed now

n8sh:
>I can't mark the non SortedRange version of the functions deprecated as they call each other.

Can't each non-SortedRange function just call the SortedRange equivalent?

burner:
I hope to wrap sub calls in assumeSorted, by that doesn't work as I would have to reconstruct RangeOfRanges which could be everything.
I think this is as good as it gets right now.

thewilsonator:
@andralex You've got 3 days to give input on this.

n8sh:
>Stuff won't break

If that's true why are there changes inside unittest blocks?

n8sh:
>I can't mark the non SortedRange version of the functions deprecated as they call each other.

I think they can all be marked `deprecated` without a problem except for `multiwayMerge`. I'll push that change to this PR.

burner:
thank you

std/algorithm/setops.d
```diff
@@ -888,39 +1013,81 @@ struct MultiwayMerge(alias less, RangeOfRanges)
...
 (alias less = "a < b", RangeOfRanges)
 (RangeOfRanges ror)
+if (isSortedRange!(ElementType!(RangeOfRanges)))
 {
-    return typeof(return)(ror);
+    return multiwayMergeImpl!less(ror);
+}
+
+/// Ditto
+//deprecated("The use with non `SortedRange` ranges is deprecated. Please use sort or assumeSorted.")
```
thewilsonator: Why are these commented out?

std/algorithm/setops.d
```diff
@@ -672,33 +737,73 @@ Params:
...
 
-    if (tgt.empty) return;
-    alias InfoType = ElementType!Range;
-    bool heapComp(InfoType a, InfoType b)
-    {
-        return weights[a[0]] * a[1] > weights[b[0]] * b[1];
-    }
-    topNCopy!heapComp(group(multiwayMerge!less(ror)), tgt, sorted);
+/// Ditto
+//deprecated("The use with non `SortedRange` ranges is deprecated. Please use sort or assumeSorted.")
```
thewilsonator: Why are these commented out?

std/algorithm/setops.d
n8sh: If you  un-comment them then Phobos doesn't compile due to deprecation settings. Maybe `pragma(msg)` could be used here?

burner:
anything else?

std/algorithm/setops.d
```diff
@@ -929,89 +1099,156 @@ MultiwayMerge!(less, RangeOfRanges) multiwayMerge
...
+    [
+        [9.0, 8.0, 8.0, 8.0, 7.0, 6.0],
+        [9.0, 8.0, 6.0],
+        [9.0, 8.0, 5.0]
+    ];
+    auto witness2 = [9, 8, 7, 6, 5];
+    assert(equal(multiwayUnion!order(c), witness2));
+}
+
+private auto mulitwayUnionImpl(alias less = "a < b", RangeOfRanges)(RangeOfRanges ror)
```
andralex: Ditto - probably not needed?

std/algorithm/setops.d
```diff
@@ -1109,14 +1377,23 @@ SetDifference!(less, R1, R2) setDifference(alias less = "a < b", R1, R2)
...
-    int[] b = [2, 4];
+    auto a = assumeSorted([1, 2, 3, 4, 5]);
+    auto b = assumeSorted([2, 4]);
     foreach (ref e; setDifference(a, b))
         e = 0;
     assert(equal(a, [0, 2, 0, 4, 0]));
 }
 
+/// Ditto
+private auto setDifferenceImpl(alias less = "a < b", R1, R2)
```
andralex: Same question about the necessity of this

std/algorithm/setops.d
```diff
@@ -1255,40 +1566,60 @@ if (Rs.length >= 2 && allSatisfy!(isInputRange, Rs) &&
     assert(equal(setIntersection(d, e), [1, 1, 7]));
 }
 
+private auto setIntersectionImpl(alias less = "a < b", Rs...)(Rs ranges)
```
andralex: Same question about the necessity of this

std/algorithm/setops.d
```diff
@@ -1396,39 +1729,75 @@ public:
...
+    static assert(isForwardRange!(typeof(setSymmetricDifference(a, b))));
+
+    //mutisets
+    auto c = [1, 1, 1, 1, 2, 2, 2, 4, 5, 6];
+    auto d = [1, 1, 2, 2, 2, 2, 4, 7, 9];
+    assert(equal(setSymmetricDifference(c, d), setSymmetricDifference(d, c)));
+    assert(equal(setSymmetricDifference(c, d), [1, 1, 2, 5, 6, 7, 9]));
+}
+
+private auto setSymmetricDifferenceImpl(alias less = "a < b", R1, R2)
```
andralex: Same question about the necessity of this

std/range/package.d
```diff
@@ -10923,6 +10923,33 @@ that break its sorted-ness, `SortedRange` will work erratically.
...
+Determines whether `T` is a $(LREF SortedRange).
+
+Params:
+    T = type to test
+
+Returns:
+`true` if `T` is a $(LREF SortedRange);
+`false` otherwise
+*/
+template isSortedRange(T) {
```
andralex: Several issues with this:

* New public symbol added as a small part of an unrelated PR
* Wrong API - the predicate should be present (sorted by what?)
* Not actually needed in the PR so it needs to be justified separately (my take: unneeded)

std/algorithm/setops.d
```diff
@@ -593,29 +594,80 @@ duplicate in between calls).
...
-    {
-        static int opIndex(ElementType!(ElementType!RangeOfRanges)) { return 1; }
-    }
-    return largestPartialIntersectionWeighted!less(ror, tgt, UnitWeights(),
-            sorted);
+    largestPartialIntersectionImpl(ror, tgt, sorted);
+}
+
+/// ditto
+void largestPartialIntersection
```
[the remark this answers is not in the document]
burner: yes, the one that is not requiring a SortedRange should be removed eventually.

std/algorithm/setops.d
```diff
@@ -648,6 +700,18 @@ void largestPartialIntersection
     assert(y[1] == tuple(1.0, 6u));
 }
 
+private void largestPartialIntersectionImpl
+(alias less = "a < b", RangeOfRanges, Range)
```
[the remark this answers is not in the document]
burner: I'm don't think so, as the code would have to construct a new RangeOfRanges where is(ElementType!RangeOfRanges == SortedRange).
The problem is that RangeOfRanges can be any type of range, which would need to be recreated.

RazvanN7:
@burner I know this is old, but any chance of applying @andralex 's feedback and then tackle another round of review?

burner:
I'll take another look

RazvanN7:
ping @burner

burner:
yeah I know, shame on me. I promise I'll get to it. :disappointed:

RazvanN7:
@burner sorry for poking you like this, but have you made any progress on this?

burner:
@RazvanN7 please don't feel sorry, I have not. But I just opened up the std.algorithm.setup. Thank you for being my naggy feeling in the back of my mind !

burner:
@RazvanN7 I actually got it done I think.

std/range/package.d
RazvanN7: I think this addition is needed since there are other places that check whether a range is sorted ([1] is an example and there might be others), however, I adding it in this PR is not the right way to go. Regarding the presence of the predicate, I think that there are cases where you don't really care what the predicate is, but we should indeed accommodate the case where the predicate is also accounted for (in [1], it is also checked that the predicate is the default one).

What I suggest is to make a PR that adds `isSortedRange` so that others in the community take a look at it. Once it is merged we can come back to this one. How does that sound @burner ?

[1] https://github.com/dlang/phobos/blob/master/std/algorithm/searching.d#L1597

changelog/pr6795.dd
```diff
@@ -0,0 +1,33 @@
...
+or $(REF assumeSorted, std,range).
+
+-------
+import std.algorithm : equal, sort;
+import std.range : assumeSorted;
+
+// sets
+auto a = sort([ 1, 2, 4, 5, 7, 9 ]);
+auto b = assumeUnique([ 0, 1, 2, 4, 7, 8 ]);
+auto intersection = setIntersection(assumeSorted(a), assumeSorted(a));
```
benjones, line 32: I assume you meant assumeSorted(b)?

burner: Indeed, thank you, good catch.

burner:
@RazvanN7 should we take another look at this.

RazvanN7 (commented):
Besides the comments, a few general observations:

- why don't we just deprecate string predicates? They are ugly, provide no benefit and just look at how many hoops we have to jump to support them. If comparing predicates for equality is the answer, we can compare simple lambda functions with [__traits(isSame)](https://dlang.org/spec/traits.html#isSame)
- I see that you are using both `sort` and `assumeSorted` to obtain sorted ranges from already sorted arrays. Why not use only `assumeSorted? It should be more efficient that way. Now that being said:
- I possible counter argument for this addition is the fact that there might be situations where you know the range is correctly sorted (even though it is not a SortedRange), but you don't want to pay the penalty of creating a SortedRange. I admit that this should be a very rare case, but with this addition we are closing the door to such scenarios, whereas, with the status quo it's on the user to make sure about the correctness of the provided data. Also, consider, that users might want to use their own version of a sorted range - this becomes impossible without wrapping their type in a SortedRange.

std/range/package.d
```diff
@@ -10785,7 +10785,7 @@ enum SortedRangeOptions
 */
 struct SortedRange(Range, alias pred = "a < b",
     SortedRangeOptions opt = SortedRangeOptions.assumeSorted)
-if (isInputRange!Range && !isInstanceOf!(SortedRange, Range))
+if (isInputRange!Range && !isInstanceOf!(isInputRange, Range))
```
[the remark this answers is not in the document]
burner: yes, I don't know how that happend. Reverted now

changelog/pr6795.dd
```diff
@@ -0,0 +1,33 @@
+Fix Issue 10706: std.algorithm.setops functions require SortedRange arguments.
```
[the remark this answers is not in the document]
burner: with the current implementation, it is only breaking when dmd is called with -de

tgehr:
FWIW: I would prefer if this did not go ahead. We are not really doing this kind of thing with any other invariants and SortedRange is not even able to properly enforce its own invariant due to mutable aliasing. In fact, I would also prefer to have binary search functions in std.algorithm that do not require a sorted range.

burner:
> FWIW: I would prefer if this did not go ahead. We are not really doing this kind of thing with any other invariants and SortedRange is not even able to properly enforce its own invariant due to mutable aliasing. In fact, I would also prefer to have binary search functions in std.algorithm that do not require a sorted range.

I think that is a terrible argument. This PR makes the usage of the public facing api of phobos less error prone. To not do this because other parts of phobos not do this, yet, or because the already existing and used abstraction SortedRange has some problem, that hardly any user or dev cares about is the wrong approach in my opinion. 

> In fact, I would also prefer to have binary search functions in std.algorithm that do not require a sorted range.

andralex:
IIRC all a user needs to do to transform a range known to be sorted into a `SortedRange` is insert an `assumeSorted`, which performs a cheap (but surprisingly valuable statistically) test of the range's sortedness.

As long as does not force the library into a memory safety issue, being mindful of breaking sortedness by means of mutable aliasing is user's responsibility. Sortedness of a `SortedRange` is in many ways opt-in, not a guarantee.

andralex:
Looking through the existing code, I found an issue with `SortedRange`. The entire type is parameterized by `SortedRangeOptions`, but that parameter is only used by the constructor. 

1. Parameterizing the entire type by a constructor parameter is wasteful
2. Even in the constructor, that parameter can be a simple run-time parameter that costs virtually nothing to test.
3. In my opinion the `log(n)` checking (checkRoughly) should be the default or the only option. Doing no checking at all really legitimizes a cast and we already have strict checking with `isSorted`.

tgehr:
> IIRC all a user needs to do to transform a range known to be sorted into a `SortedRange` is insert an `assumeSorted`, which performs a cheap (but surprisingly valuable statistically) test of the range's sortedness.

iota(n).assumeSorted. :clown_face:

tgehr:
> > FWIW: I would prefer if this did not go ahead. We are not really doing this kind of thing with any other invariants and SortedRange is not even able to properly enforce its own invariant due to mutable aliasing. In fact, I would also prefer to have binary search functions in std.algorithm that do not require a sorted range.
> 
> I think that is a terrible argument. This PR makes the usage of the public facing api of phobos less error prone. To not do this because other parts of phobos not do this, yet,

The issue is that this way to enforce invariants results in less than great UX.

> or because the already existing and used abstraction SortedRange has some problem,

It provides a false sense of security.

> that hardly any user or dev cares about

Hardly any user or dev cares for the SortedRange design.

> is the wrong approach in my opinion.

What this change does is cause people to roll their own versions of those functions. It's already the case with binary search.

pbackus:
> In my opinion the log(n) checking (checkRoughly) should be the default or the only option. Doing no checking at all really legitimizes a cast and we already have strict checking with isSorted.

I would be very surprised to learn that a function with "assume" in its name does any checking at all by default. We don't do any checking for [`assumeUnique`][1] or [`assumeSafeAppend`][2], for example, and even [`assumeUTF`][3] only enables its checks in debug mode.

[1]: https://dlang.org/library/std/exception/assume_unique.html
[2]: https://dlang.org/library/object/assume_safe_append.html
[3]: https://dlang.org/library/std/string/assume_utf.html

jamesragray:
For what it is worth. I have not found the assume sorted mechanism ergonomic or useful, ever. Hence since (a) finding out that using it does some calculations wasting time; (b) being forced to use it even more, our team will no longer use the Phobos versions of these functions if this goes through. Perhaps one can add a way of opting out? e.g. add an overload or a flag that doesn't require this.

andralex:
> > In my opinion the log(n) checking (checkRoughly) should be the default or the only option. Doing no checking at all really legitimizes a cast and we already have strict checking with isSorted.
> 
> I would be very surprised to learn that a function with "assume" in its name does any checking at all by default. We don't do any checking for [`assumeUnique`](https://dlang.org/library/std/exception/assume_unique.html) or [`assumeSafeAppend`](https://dlang.org/library/object/assume_safe_append.html), for example, and even [`assumeUTF`](https://dlang.org/library/std/string/assume_utf.html) only enables its checks in debug mode.

A fair point. There's no meaningful checking we can do in the first two anyway - if we could, probably the story would be different. `assumeUTF` is a more relevant precedent, so perhaps `assumeSorted` could limit its checking to debug mode.

andralex:
> For what it is worth. I have not found the assume sorted mechanism ergonomic or useful, ever. Hence since (a) finding out that using it does some calculations wasting time; (b) being forced to use it even more, our team will no longer use the Phobos versions of these functions if this goes through. Perhaps one can add a way of opting out? e.g. add an overload or a flag that doesn't require this.

First it needs to be understood we're looking at like 20 comparisons for one million elements. The utility of those is high. I can't find the paper I looked at back in the day, but perhaps this (and/or its references) may be a good starting point: https://cs-people.bu.edu/sofya/pubs/TestingSortednessEncyclopedia.pdf

I don't see the utility of a rewrite. Would it consist of one-liners that append `.assumeSorted` at the end of the ranges involved?

In your application, would it be feasible and useful to keep ranges that are known to be sorted as `SortedRange`s? That would have the same properties as the original range plus the information that it is sorted.

burner:
I think I'm missing something but assumeSorted doesn't check as far as I see. The SortedRange ctor only checks on checkRoughly and checkStrictly which is not the case. The default is "assumeSorted" !?

andralex:
> I think I'm missing something but assumeSorted doesn't check as far as I see. The SortedRange ctor only checks on checkRoughly and checkStrictly which is not the case. The default is "assumeSorted" !?

Indeed, and per my previous answer that seems to be a problem (one of several of `SortedRange`), `checkStrictly` is useless because it overlaps `isSorted`; I consider no checking specious, and `checkRoughly` is the sweet spot.

jamesragray:
> > For what it is worth. I have not found the assume sorted mechanism ergonomic or useful, ever. Hence since (a) finding out that using it does some calculations wasting time; (b) being forced to use it even more, our team will no longer use the Phobos versions of these functions if this goes through. Perhaps one can add a way of opting out? e.g. add an overload or a flag that doesn't require this.
> 
> First it needs to be understood we're looking at like 20 comparisons for one million elements. The utility of those is high. I can't find the paper I looked at back in the day, but perhaps this (and/or its references) may be a good starting point: https://cs-people.bu.edu/sofya/pubs/TestingSortednessEncyclopedia.pdf
> 
Fair point, but still not needed (in my opinion) in most cases.  I think that it is probably true that most common case to want to do say a binary search is on an array. You are forcing the user to wrap their type for a tiny bit of safety.

> I don't see the utility of a rewrite. Would it consist of one-liners that append `.assumeSorted` at the end of the ranges involved?
>
I haven't studied the code carefully but I would expect one could simply copy the phobos code removing the template constraints, so that SortedRange is not needed.

> In your application, would it be feasible and useful to keep ranges that are known to be sorted as `SortedRange`s? 
I guess this might be reasonable in some cases, but I do not think Phobos should be forcing users do this.

LightBender:

PR #6796, opened 2018-12-08
canFind: added example to search for multiple needles in a hay stack.

carun:


std/algorithm/searching.d
```diff
@@ -2555,6 +2555,18 @@ template canFind(alias pred="a == b")
     ];
     assert(!canFind(words, "bees"));
     assert( canFind!((string a, string b) => a.startsWith(b))(words, "bees"));
+
+}
+
+/// Search for mutliple items in an array of items (searching needles in a hay stack)
```
wilzbach: Typo: multiple
Also the description is wrong. You are searching for items in an array of array of items. Though that will be confusing to people.
Maybe something like:

Search for needles in an array of hay stacks

carun: @wilzbach I force pushed. Does the new changeset look OK?

std/algorithm/searching.d
```diff
@@ -2557,6 +2557,17 @@ template canFind(alias pred="a == b")
     assert( canFind!((string a, string b) => a.startsWith(b))(words, "bees"));
 }
 
+/// Search for mutliple items in an array of items (search for needles in an array of hay stacks)
```
wilzbach: Still a typo here: multiple

Also it's probably a good idea to use multiple lines here as the soft limit is 80.

carun: Ah, I can't even spell multiple properly! I need to enable spell check "only" for comments.

wilzbach: Oh and sorry for just noticing now, haystack is one word: http://dictionary.cambridge.org/english/haystack

PR #6797, opened 2018-12-08
SHA on Win64: use SSSE3 version by adjusting register usage

rainers:
druntime benchmarks gcbench/con*.* perform very badly on Win64 (10 times slower in comparison to Win32), so they don't really measure GC performance. This fixes it.

### DIFF
--- a/std/digest/sha.d
+++ b/std/digest/sha.d
@@ -101,11 +101,7 @@ module std.digest.sha;
     hash1 = sha1.finish();
 }
 
-version (Win64)
-{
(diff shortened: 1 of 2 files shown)


thewilsonator:
This be `version`ed to ssse3 or newer, no?

rainers:
> This should be `version`ed to ssse3 or newer, no?

Both runtime and compile time checks are already done here: https://github.com/dlang/phobos/blob/master/std/digest/sha.d#L213

PR #6802, opened 2018-12-11
Make decodeImpl safe

edi33416:


thewilsonator:
`@safe` unittests?

edi33416:
> `@safe` unittests?

I rushed into it :laughing:

edi33416:
@thewilsonator I just did another push because I forgot to order alphabetically the `unittest` attributes

edi33416:
> I oppose these changes. They remove pointer arithmetic from `decodeImpl` in order to make it `@safe`, and the reason that pointer arithmetic is used is to avoid unnecessary bounds checking and thus improve performance. `decodeImpl` is a private function and currently allows the caller to deal with `@safe`. Why are these changes being made?

I just saw the code and thought "that's odd, this can be made `@safe`".
I rushed and I didn't take into account the performance hit.

Closing this

jmdavis:
That's understandable, but this is code where you're going to need to be very careful when messing with it, because it's highly performance sensitive, and it's one of the most used functions in all of Phobos (since it's at the core of a most string processing). Any changes made to it need to be very well thought through, and anything that changes the logic at all probably needs to be performance tested before being merged to ensure that it doesn't hurt performance. It doesn't help that it needs to be able to operate on dynamic arrays in a particular way and generic ranges in another (not just for performance - calling `front` or `popFront` here on a string would be recursive) or that it needs to work with both exceptions and with the Unicode replacement character. All in all, it's a very tricky piece of code.

edi33416:
> That's understandable, but this is code where you're going to need to be very careful when messing with it, because it's highly performance sensitive, and it's one of the most used functions in all of Phobos (since it's at the core of most string processing). Any changes made to it need to be very well thought through, and anything that changes the logic at all probably needs to be performance tested before being merged to ensure that it doesn't hurt performance. [...]

Is there a benchmarking test (suite, ideally) for `std.utf`? It would be nice if we'd have some `StdBenchUTF` `unittest`s

jmdavis:
> Is there a benchmarking test (suite, ideally) for std.utf? It would be nice if we'd have some StdBenchUTF unittests.

Not AFAIK. This isn't code that gets changed often. I recall Dmitry having something for std.uni, so that _might_ be applicable even if it's not specifically for std.utf, but the last time I seriously reworked `decodeImpl` was several years ago, and whatever I may have had for performance testing specifically is long gone. There are quite a few unit tests of course, but there isn't anything committed that's specifically for benchmarking.

PR #6803, opened 2018-12-12
Fix Issue 19213 - goto skips declaration of variable in std.algorithm…

edi33416:
….iteration.joiner

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -2706,21 +2706,24 @@ if (isInputRange!RoR && isInputRange!(ElementType!RoR))
 
         private enum popFrontEmptyElements = q{
             // Skip over empty subranges.
-            if (_items.empty) goto end;
-            while (_items.front.empty)
+            while (!_items.empty && _items.front.empty)
             {
                 _items.popFront();
-                if (_items.empty) goto end;
             }
-            // We cannot export .save method unless we ensure subranges are not
-            // consumed when a .save'd copy of ourselves is iterated over. So
-            // we need to .save each subrange we traverse.
-            static if (isForwardRange!RoR && isForwardRange!(ElementType!RoR))
-                _current = _items.front.save;
+            if (!_items.empty)
+            {
+                // We cannot export .save method unless we ensure subranges are not
+                // consumed when a .save'd copy of ourselves is iterated over. So
+                // we need to .save each subrange we traverse.
+                static if (isForwardRange!RoR && isForwardRange!(ElementType!RoR))
+                    _current = _items.front.save;
+                else
+                    _current = _items.front;
+            }
             else
-                _current = _items.front;
-        end:
-            assert(1); // required to avoid 'EOF instead of statement' error
+            {
+                _current = typeof(_current).init;
+            }
         };
 
         static if (isForwardRange!RoR && isForwardRange!(ElementType!RoR))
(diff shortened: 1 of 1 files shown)


edi33416:
Changes to `popBackEmptyElements` were made to be consistent with the changes required by `popFrontEmptyElements`.

Note: the regression can be fixed just with the changes done in `popFrontEmptyElements`

std/algorithm/iteration.d
```diff
@@ -2706,21 +2706,24 @@ if (isInputRange!RoR && isInputRange!(ElementType!RoR))
...
             {
                 _items.popFront();
-                if (_items.empty) goto end;
             }
-            // We cannot export .save method unless we ensure subranges are not
-            // consumed when a .save'd copy of ourselves is iterated over. So
-            // we need to .save each subrange we traverse.
-            static if (isForwardRange!RoR && isForwardRange!(ElementType!RoR))
-                _current = _items.front.save;
+            if (!_items.empty)
```
wilzbach: This is sub-optimal as it will be executed one-more time than strictly necessary.

std/algorithm/iteration.d
```diff
@@ -2706,21 +2706,24 @@ if (isInputRange!RoR && isInputRange!(ElementType!RoR))
...
+                    _current = _items.front.save;
+                else
+                    _current = _items.front;
+            }
             else
-                _current = _items.front;
-        end:
-            assert(1); // required to avoid 'EOF instead of statement' error
+            {
+                _current = typeof(_current).init;
```
wilzbach: Why is this required?

std/algorithm/iteration.d
```diff
@@ -2800,32 +2803,36 @@ if (isInputRange!RoR && isInputRange!(ElementType!RoR))
...
+                    }
                 }
                 else
                 {
-                    if (reachedFinalElement)
-                        _current = _items.back;
-                    else
-                        _currentBack = _items.back;
+                    _current = typeof(_current).init;
+                    _currentBack = typeof(_currentBack).init;
```
wilzbach: Why is this required?

std/algorithm/iteration.d
```diff
@@ -2800,32 +2803,36 @@ if (isInputRange!RoR && isInputRange!(ElementType!RoR))
...
                 {
                     _items.popBack();
-                    if (_items.empty) goto end2;
                 }
-                checkFinalElement;
-                // We cannot export .save method unless we ensure subranges are not
-                // consumed when a .save'd copy of ourselves is iterated over. So
-                // we need to .save each subrange we traverse.
-                static if (isForwardRange!RoR && isForwardRange!(ElementType!RoR))
+                if (!_items.empty)
```
wilzbach: This is sub-optimal as it will be executed one-more time than strictly necessary.

std/algorithm/iteration.d
wilzbach: It should be faster to do sth. this:

```
while (!_items.empty) {
   if (!_items.front.empty) {
      _items.popFront();
   } else {
     // <- static if code here
     break;
   }
}
```

edi33416: I still need the outside `if (!_items.empty) { } else { }` block, or the compiler will complain with `Error: one path skips field _current`.

This is because, in the case of an empty ROR, `_current` won't be initialised.

std/algorithm/iteration.d
edi33416: The compiler requires that `_current` is initialised on all possible paths.

std/algorithm/iteration.d
edi33416: Please see above comments

std/algorithm/iteration.d
edi33416: Please see above comments

std/algorithm/iteration.d
```diff
@@ -2948,6 +2955,20 @@ if (isInputRange!RoR && isInputRange!(ElementType!RoR))
...
+{
+    static struct Rec
+    {
+        int foo;
+    }
+    static assert(__traits(compiles, {
+        auto results = [[1,2], [3,4]]
+            .map!(q => q.map!Rec.chunkBy!"a.foo")
+            .joiner;
+    }));
```
wilzbach: No need to use `__traits(compiles` here. Use the range directly and even `assert` that its still iterated correctly.
Also, it would be good to add a test for iterating this range backwards and checking the result.

std/algorithm/iteration.d
```diff
@@ -2948,6 +2955,20 @@ if (isInputRange!RoR && isInputRange!(ElementType!RoR))
     assert(!equal(js2, js));
 }
 
+// Bugzilla issue 19213
```
wilzbach: Use links to Bugzilla.

std/algorithm/iteration.d
```diff
@@ -2948,6 +2955,15 @@ if (isInputRange!RoR && isInputRange!(ElementType!RoR))
     assert(!equal(js2, js));
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=19213
+@system unittest
+{
+    auto results = [[1,2], [3,4]].map!(q => q.chunkBy!"a").joiner;
+    int i = 1;
+    foreach(ref e; results)
```
wilzbach: Style target fails here due to missing whitespace after foreach. Why not use `equals`?
Also, we should probably test the bidirectional code too (use e.g. `retro`)

edi33416: The result is not a bidirectional range, so I can't use `retro`.

PR #6807, opened 2018-12-16
[trivial] Fix `dirEntries` docs

kubo39:
Use `std.process.executeShell` instead of old system function.


ZombineDev:
Auto-merge toggled on

PR #6808, opened 2018-12-17
Fix Issue 18755 - std.typecons.Rebindable breaks @safe-ty

edi33416:


std/typecons.d
```diff
@@ -2086,9 +2086,16 @@ if (is(T == class) || is(T == interface) || isAssociativeArray!T)
...
     {
-        stripped = cast(U) another;
+        // If `T` defines `opCast` we must infer the safety
+        static if (hasMember!(T, "opCast"))
+        {
+            // This will allow the compiler to infer the safety of `T.opCast!U`
+            // without generating any runtime cost
+            if (false) { stripped = cast(U) another; }
+        }
+        () @trusted { stripped = cast(U) another; }();
```
n8sh: What is the point of keeping this lambda? Is there a reason not to just remove `@trusted` from the function?

ZombineDev: For the case where there's not `opCast` member. In which case casting would be un-`@safe`, but we trust that it's done correctly.

std/typecons.d
```diff
@@ -2260,6 +2268,23 @@ if (is(T == class) || is(T == interface) || isDynamicArray!T || isAssociativeArr
...
+        auto opCast(T)() @system immutable pure nothrow
+        {
+            *(cast(uint*) 0xdeadbeef) = 0xcafebabe;
+            return T.init;
+        }
+    }
+
+    static assert(!__traits(compiles, () @safe {
+        auto r = Rebindable!(immutable Foo)(new Foo);
+    }));
```
ZombineDev: Can you also add for completeness:
```D
    static assert(!__traits(compiles, () @system {
        auto r = Rebindable!(immutable Foo)(new Foo);
    }));
```

edi33416: Done

std/typecons.d
edi33416: I feel that the same should happen for the `opAssign` overloads, that are also `@trusted`, as they are subject to the same issue

If you agree, I'll open a separate PR for those.

ZombineDev:
@n8sh anything to add from your side? I think this PR is good to go.

std/typecons.d
ZombineDev: Yes, we can explore that in a separate PR, but I'm not sure if we can do anything for e.g. this one:

https://github.com/dlang/phobos/blob/dbca8521b5164c2ba95401d9bc784ec653e1a157/std/typecons.d#L2101-L2104

n8sh:
@ZombineDev seems fine to me.

PR #6809, opened 2018-12-17
Replace imports of core.sys.windows.windows to speed up compilation

n8sh:
Importing `core.sys.windows.windows.d` is slow and replacing it can speed up compilation. See https://github.com/dlang/druntime/pull/2400#issuecomment-447933446


ZombineDev:
The autotester fails with:
```
std\format.d(4155): Deprecation: Symbol `core.sys.windows.windef.HRESULT` is not visible from module `format` because it is privately imported in module `winbase`
```

PR #6810, opened 2018-12-18
Fix Issue 18913 - Cannot move static array of non-copyable type

edi33416:


edi33416:
While working at this, I've noticed that `move` doesn't `null`ify references.
This, imho, is a big issue. See code snippet
```
@safe
void main()
{
  int[] x = [1, 2, 3];
  int[] nx;

  move(x, nx); // nx should take ownership, but it just points to the same address
  
  assert(x == nx); // oho
  x[0] = 2; // should fail, but, instead, it silently modifies nx's content
  assert(nx == [1, 2, 3]); // fails because of bug
}
```

std/algorithm/mutation.d
```diff
@@ -1420,9 +1420,17 @@ void moveEmplace(T)(ref T source, ref T target) @system
...
+        {
+            for (size_t i = 0; i < source.length; ++i)
+                move(source[i], target[i]);
+        }
+        else
+        {
+            // Primitive data (including pointers and arrays) or class -
+            // assignment works great
+            target = source;
+        }
```
ZombineDev: I would write the whole thing as:
```D
static if (is(T == struct))
{
    // ...
}
else static if (isStaticArray!T)
    moveEmplaceAll(source[], target[]);
else
    target = source;

```

edi33416: Can't use `moveEmplaceAll` as `source[]` fails the `isInputRange` check

ZombineDev: Yes, unfortunately the `isInputRange` constraint is too strict for `move[Emplace]{All|Some}` functions, as a range, whose elements are not copy-able is the primary use-case for `move`-ing.

If we adjust `isInputRange` like so 
```diff
@@ -168,7 +168,10 @@
 enum bool isInputRange(R) =
     is(typeof(R.init) == R)
     && is(ReturnType!((R r) => r.empty) == bool)
-    && is(typeof((return ref R r) => r.front))
+    && (
+            is(typeof((return ref R r) => r.front)) ||
+            is(typeof((return ref R r) => &r.front()))
+       )
     && !is(ReturnType!((R r) => r.front) == void)
     && is(typeof((R r) => r.popFront));
```

all tests pass.

PR #6813, opened 2018-12-21
logger: print log level string.

carun:
Log message currently doesn't have log level, thread ID, etc. Log messages without this information is mostly useless.

However, printing the thread ID requires `std.concurrency` to be `@safe` which is not possible at the moment. So I've added only the log level to the message string.

### DIFF
--- a/std/experimental/logger/filelogger.d
+++ b/std/experimental/logger/filelogger.d
@@ -131,8 +131,9 @@ class FileLogger : Logger
 
         auto lt = this.file_.lockingTextWriter();
         systimeToISOString(lt, timestamp);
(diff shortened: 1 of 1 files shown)


std/experimental/logger/filelogger.d
```diff
@@ -128,11 +128,11 @@ class FileLogger : Logger
         import std.string : lastIndexOf;
         ptrdiff_t fnIdx = file.lastIndexOf('/') + 1;
         ptrdiff_t funIdx = funcName.lastIndexOf('.') + 1;
-
```
thewilsonator: Not needed

PR #6818, opened 2018-12-27
Add LibcurlPath version option

ThomasMader:
Adds possibility to specify absolute path to libcurl.
The TZDatabaseDir version option already does the same for the tzdata directory.
This commit enables the same fix for the path to libcurl. It's needed for Operating systems like NixOS without a search path for dynamic libs.
See https://issues.dlang.org/show_bug.cgi?id=15391 for more information.
Additionally I removed the Makefile part of the TZDatabaseDir solution
because the necessary compiler arguments can be easily given via DFLAGS, which is also compiler independent and doesn't use the /tmp dir which leads to problems on NixOS.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -22,10 +22,6 @@
 #
 # make std/somemodule.test => only builds and unittests std.somemodule
 #
-# make TZ_DATABASE_DIR=path to the TZDatabase directory => This is useful to
-# overwrite the hardcoded path to the TZDatabase directory needed
-# for std/datetime/timezone.d
-
 ################################################################################
 # Configurable stuff, usually from the command line
 #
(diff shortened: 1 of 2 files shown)


ZombineDev:
It would be good to make a list of all version configurations like this one in the readme file.

ZombineDev:
Alright @ThomasMader, can you please update the wiki?

ThomasMader:
I added it to https://wiki.dlang.org/Building_under_Posix#Build_phobos only as it is not needed on Windows and Nix doesn't really run on Windows natively yet.

Thanks for merging.

ZombineDev:
Thank you as well, @ThomasMader!

PR #6820, opened 2018-12-30
Fix Issue 19526 - make RefCounted work in betterC

n8sh:
Array needs https://github.com/dlang/druntime/pull/2426 to work in betterC.


n8sh:
Have to remove `@betterC` from some of the tests because unlike normal unittests they are in a different module can't refer to private members.

n8sh:
Removed changes for std.container.array.Array because it needs TypeInfo.

PR #6822, opened 2019-01-02
Synchronize dmd-cxx with gdc-stable

ibuclaw:
GDC's github repo is dead.  Until GCC 10 development starts next April, I'll be using this branch as the "upstream" to sync from for platform and bug fixes in GCC 9.  Starting with synchronizing the druntime fork (minus the gdc-specific bits).


ibuclaw:
I think that I forgot to check for file addition/removals.  Namely `std.c`, one minute...

ibuclaw:
That's better, lots of deletions.

Merging so I can move onto next set of backports....

PR #6825, opened 2019-01-03
fix quoting in win64.mak: add them only when invoking command

rainers:
not when setting variable. This avoids having to skillfully quote variables with spaces in them: use `"CC=%cl64%"` instead of `"CC=\"%cl64%\""`

Not sure how the existing misquoting was supposed to work. Using quotes with the command is agnostic to using `/` instead of `\` as a path separator, too.

See also https://github.com/dlang/druntime/pull/2438

### DIFF
--- a/etc/c/zlib/win64.mak
+++ b/etc/c/zlib/win64.mak
@@ -3,9 +3,9 @@
 MODEL=64
 VCDIR=\Program Files (x86)\Microsoft Visual Studio 10.0\VC
 
-CC="$(VCDIR)\bin\amd64\cl"
-LD="$(VCDIR)\bin\amd64\link"
-LIB="$(VCDIR)\bin\amd64\lib"
+CC=$(VCDIR)\bin\amd64\cl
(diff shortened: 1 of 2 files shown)


rainers:
Now also quoting $(MAKE)

ZombineDev:
It looks like https://github.com/braddr/at-client/blob/master/src/diff-phobos-win64.diff needs to be updated. Or even better - removed entirely ;)

wilzbach:
CC @braddr

rainers:
> Or even better - removed entirely ;)

https://github.com/braddr/at-client/pull/10

rainers:
With the diffs removed from the autotester, this should be unblocked.

kinke:
> This avoids having to skillfully quote variables

Lol, one way to put it. ;)

wilzbach:
Looks like the patching is still done by the Auto-Tester:

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3546293&isPull=true

PR #6830, opened 2019-01-10
std.random: make dip1000 compliant

WalterBright:


thewilsonator:
It looks like you've missed a couple.

cc @n8sh 

(is mir-random dip1000 compatible?)

n8sh:
@thewilsonator I confess that I don't understand dip1000. Take this compilation error message for this PR:

>std/random.d(3687): Error: returning RandomSample(0LU, 0LU, nan, Result(0, 0), 0LU, cast(Skip)0, null).this(r, n, r.length(), rng) escapes a reference to parameter rng, perhaps annotate with return

But parameter rng is already annotated with return:

https://github.com/dlang/phobos/blob/f542fb9795c26b565972afde049b7f739513cc41/std/random.d#L3684-L3688

Is dip1000 supported by the compiler yet? Is this error message itself an error or is it justified from dip1000's intended design?

WalterBright:
> Is dip1000 supported by the compiler yet? Is this error message itself an error or is it justified from dip1000's intended design?

I found the actual problem with `std.random` and corrected it. (Shouldn't pass `&rng` to a parameter that is `auto ref`, it doesn't make much sense.) There's another problem I found and corrected in the compiler, but won't make a PR for it until tomorrow.

WalterBright:
Spoke too soon :-( Trouble is, I don't know how the std.random code is supposed to work.

thewilsonator:
Seems to be all green now.

WalterBright:
Yeah, I fixed it. Please approve, as I need it for the dmd PR.

wilzbach:
> (is mir-random dip1000 compatible?)

I don't think there's a single non-trivial D lib that's compatible with dip1000 as of now.

WalterBright:
druntime is. But we can't realistically expect anything else to be until Phobos is.

PR #6831, opened 2019-01-10
make OnlyResult dip1000 compliant

WalterBright:


thewilsonator:
This has a bunch of cascade errors.

cc @BBasile any ideas why dscanner is giving a bunch or errors in an unchanged file>

WalterBright:
I'm going to abandon this until I figure out what is wrong with https://github.com/dlang/phobos/pull/6830

PR #6832, opened 2019-01-10
Fix issue 19572: Prevent Appender from implicitly calling struct constructors to cast away const.

FeepingCreature:
Prevent Appender from implicitly calling struct constructors to cast away const.

Sample code:

```
struct Struct
{
    int value;

    int fun() const { return 23; }

    alias fun this;
}

void main()
{
    import std.array : Appender;

    Appender!(Struct[]) appender;

    appender.put(const(Struct)(42));

    auto result = appender.data[0];

    assert(result.value != 23);
}
```

What happens is that `Appender` checks that it can convert `const(Struct)` implicitly to `Struct`, but it does so via `cast(Struct)`, which then invokes `Struct(const(Struct).fun)` implicitly.

With this fix, instead of checking `const(Struct)` vs `Struct`, we check `Unqual` against `Unqual` and then use the dmd-fixed `cast()` to remove constness, before relying on implicit conversion to do the rest.

Note that this relies on https://github.com/dlang/dmd/pull/9223 and stems from the same investigation.

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -3339,7 +3339,7 @@ if (isDynamicArray!A)
     private template canPutItem(U)
     {
         enum bool canPutItem =
-            isImplicitlyConvertible!(U, T) ||
+            isImplicitlyConvertible!(Unqual!U, Unqual!T) ||
             isSomeChar!T && isSomeChar!U;
     }
     private template canPutConstRange(Range)
@@ -3384,7 +3384,7 @@ if (isDynamicArray!A)
             immutable len = _data.arr.length;
 
             auto bigData = (() @trusted => _data.arr.ptr[0 .. len + 1])();
-            emplaceRef!(Unqual!T)(bigData[len], cast(Unqual!T) item);
+            emplaceRef!(Unqual!T)(bigData[len], cast() item);
             //We do this at the end, in case of exceptions
             _data.arr = bigData;
         }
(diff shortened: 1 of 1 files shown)


FeepingCreature:
To emphasize: it is expected that this build fails while the DMD fix is not yet merged.

std/array.d
```diff
@@ -3657,6 +3657,26 @@ if (isDynamicArray!A)
...
+        alias fun this;
+    }
+
+    Appender!(Struct[]) appender;
+
+    appender.put(const(Struct)(42));
+
+    auto result = appender.data[0];
+
+    assert(result.value != 23);
```
thewilsonator: Apparently this fails.

FeepingCreature: Correct, since it depends on the ongoing DMD PR.

thewilsonator: Oh whoops. That needs a changelog but otherwise looks good.

wilzbach: Changelog will be automatically generated as it mentions an issue

thewilsonator: No, without a changelog it gets a generic `issue NNNNN was fixed` entry. A specific changelog entry is for highlighting things that may be of interest, e.g. w.r.t changing behaviour such as this.

wilzbach: (I was talking about this specific PR here which doesn't need a detailed changelog.)

FeepingCreature:
DMD fix merged, seems green now.

PR #6833, opened 2019-01-14
[non-DMD] std.random seed bootstrap: don't count on reading an uninitialized variable being treated as an ordinary read

n8sh:


n8sh:
Re-pushing to fix typo in comment

kinke:
Thx, should probably have been `stable`, but oh well. ;)

PR #6834, opened 2019-01-14
dip1000: add scope to SortedRange.opSlice()

WalterBright:


WalterBright:
This is just whack-a-mole. Abandoning.

thewilsonator:
Looks like you're having about as much fun as Manu. I recommend trying to find the bottom of the rabbit hole (where ever that is) and filling it back up as you go out.

WalterBright:
A major problem I have with Phobos is there is no bottom to the rabbit hole in Phobos, as it all goes around in a circle of dependencies.

PR #6836, opened 2019-01-18
Add minimum HPPA support to phobos

ibuclaw:
So it at least compiles...


thewilsonator:
`make: *** [dscanner] Segmentation fault`, Oy, that doesn't look good. @BBasile

BBasile:
@thewilsonator unfortunately D-Scanner doesn't crash locally. I even tested the exact PR. It looks like you'll have to run in in gdb, like in the past iirc.

thewilsonator:
Hmm, bummer. @ibuclaw you could try giving it another kick to see if it as a random failure.

BBasile:
Still the results from 9 hours ago. amend + force push is necessary.

PR #6839, opened 2019-01-22
alias std.array.Appender.opOpAssign to put

Herringway:
We can eliminate an unnecessary `__traits(compiles, ...)` by doing this and it simplifies things for us. It doesn't look like this change is feasible for RefAppender though.

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -3467,10 +3467,11 @@ if (isDynamicArray!A)
     }
 
     /**
-     * Appends `rhs` to the managed array.
+     * Appends to the managed array.
      * Params:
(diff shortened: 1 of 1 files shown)


std/array.d
```diff
@@ -3467,16 +3467,13 @@ if (isDynamicArray!A)
     }
 
     /**
-     * Appends `rhs` to the managed array.
+     * Appends to the managed array.
      * Params:
      *     op = the assignment operator `~`
```
Basile-z: I think that this parameter would not have to be documented because the "colon value" (which appears in the generated docs) 
- is self explanatory 
- cannot change.

PR #6842, opened 2019-01-28
Reduce std/uri compilation time by 10x

edi33416:
Time dropped from
```
time dmd -c std/uri.d                                                                                                                                           
dmd -c std/uri.d  0,32s user 0,03s system 95% cpu 0,363 total
```
to
```
time dmd -c std/uri.d                                                                                      
dmd -c std/uri.d  0,03s user 0,00s system 99% cpu 0,027 total
```

### DIFF
--- a/std/uri.d
+++ b/std/uri.d
@@ -378,7 +378,7 @@ if (isSomeChar!Char)
  * Returns:
  *      A string encoded using www-form-urlencoding.
  */
-package string urlEncode(scope string[string] values) @safe pure
+package string urlEncode()(scope string[string] values) @safe pure
 {
     if (values.length == 0)
         return "";


edi33416:
> Doesn't this just increase compilation time by about the same amount every time this is used, though?

Why would it?

Won't this just get instantiated (by the compiler) at most once, at the first use and then it will be there for future references?

It doesn't have any template arguments that would trigger multiple instantiations based on their type.

Am I missing something?

wilzbach:
Closing as this is [a dead-end](https://github.com/dlang/phobos/pull/6843#issuecomment-458669246). We should try to fix the compiler or compilation strategies instead.

That being said there are a few parts of Phobos (e.g. `std.uni`) which are notoriously expensive to import. Fixing the import time of these modules (and thus of all modules that import them), would be a nice improvement until we can fix the root problem.

PR #6843, opened 2019-01-28
Reduce std/complex compilation time by 10x

edi33416:
Time dropped from
```
time dmd -c std/complex.d
dmd  -c std/complex.d  0,31s user 0,04s system 99% cpu 0,352 total
```
to
```
time dmd -c std/complex.d
dmd  -c std/complex.d  0,02s user 0,00s system 98% cpu 0,020 total
```

### DIFF
--- a/std/complex.d
+++ b/std/complex.d
@@ -871,7 +871,7 @@ deprecated
     x87 $(I fsincos) instruction when possible, this function is no faster
     than calculating cos(y) and sin(y) separately.
 */
-Complex!real expi(real y)  @trusted pure nothrow @nogc
+Complex!real expi()(real y)  @trusted pure nothrow @nogc
 {
     import std.math : cos, sin;
     return Complex!real(cos(y), sin(y));
@@ -904,7 +904,7 @@ deprecated
     `coshisinh` is included here for convenience and for easy migration of code
     that uses $(REF _coshisinh, std,math).
 */
-Complex!real coshisinh(real y) @safe pure nothrow @nogc
+Complex!real coshisinh()(real y) @safe pure nothrow @nogc
 {
     static import std.math;
     if (std.math.fabs(y) <= 0.5)


quickfur:
This is ... interesting ... Generally, I like the idea of templatizing functions to reduce the cost of importing/using the module to a pay-as-you-go model, and also to reap the benefits of attribute inference, but in this case, the focus on reducing compilation times seems misplaced:  basically, you're deferring the compilation of said functions from Phobos build time to user code build time (when user code actually calls said functions).

Before this change, there would be a miniscule increase in Phobos build time, but thereafter any user code that calls these functions could just call them with no more additional cost, whereas after this change, the Phobos build time improves (by a miniscule amount), but the build time of *every user project* that calls these functions would increase because now the compiler has to instantiate the templates. And this is done *per compilation* of user code, rather than the one-time cost of compiling Phobos before shipping.  Of course, the total increase in user project compilation time will also be miniscule, but that just shows that this whole emphasis on reducing compilation times by fractions of seconds is misplaced.

Shouldn't we rather be looking at how to improve the actual *code* in the function body so that it compiles faster, rather than merely inserting `()` everywhere just to defer the compilation cost in order to optimize the (broken) metric?

wilzbach:
The only cost that we might try to optimize is the cost you pay for doing `import std.complex` (i.e. the real overhead cost for the user).

edi33416:
@quickfur @klickverbot @wilzbach  you are right; my metric was broken

Given a simple `test.d` file
```
import std.complex;

void main(string[] args)
{
}
```

The number of imported modules dropped from **16**
```
$ dmd -v -c test.d | grep '^import\ '
import    object        (/home/edis/workspace/dlang/code/druntime/import/object.d)
import    core.internal.hash    (/home/edis/workspace/dlang/code/druntime/import/core/internal/hash.d)
import    core.internal.traits  (/home/edis/workspace/dlang/code/druntime/import/core/internal/traits.d)
import    core.internal.convert (/home/edis/workspace/dlang/code/druntime/import/core/internal/convert.d)
import    std.complex   (/home/edis/workspace/dlang/code/phobos/std/complex.d)
import    std.traits    (/home/edis/workspace/dlang/code/phobos/std/traits.d)
import    std.meta      (/home/edis/workspace/dlang/code/phobos/std/meta.d)
import    std.functional        (/home/edis/workspace/dlang/code/phobos/std/functional.d)
import    std.format    (/home/edis/workspace/dlang/code/phobos/std/format.d)
import    core.vararg   (/home/edis/workspace/dlang/code/druntime/import/core/vararg.d)
import    core.stdc.stdarg      (/home/edis/workspace/dlang/code/druntime/import/core/stdc/stdarg.d)
import    core.stdc.stdlib      (/home/edis/workspace/dlang/code/druntime/import/core/stdc/stdlib.d)
import    core.stdc.config      (/home/edis/workspace/dlang/code/druntime/import/core/stdc/config.d)
import    core.stdc.stddef      (/home/edis/workspace/dlang/code/druntime/import/core/stdc/stddef.d)
import    std.exception (/home/edis/workspace/dlang/code/phobos/std/exception.d)
import    std.range.primitives  (/home/edis/workspace/dlang/code/phobos/std/range/primitives.d)
```
to **8**
```
$ dmd -v -c test.d | grep '^import\ '
import    object        (/home/edis/workspace/dlang/code/druntime/import/object.d)
import    core.internal.hash    (/home/edis/workspace/dlang/code/druntime/import/core/internal/hash.d)
import    core.internal.traits  (/home/edis/workspace/dlang/code/druntime/import/core/internal/traits.d)
import    core.internal.convert (/home/edis/workspace/dlang/code/druntime/import/core/internal/convert.d)
import    std.complex   (/home/edis/workspace/dlang/code/phobos/std/complex.d)
import    std.traits    (/home/edis/workspace/dlang/code/phobos/std/traits.d)
import    std.meta      (/home/edis/workspace/dlang/code/phobos/std/meta.d)
import    std.functional        (/home/edis/workspace/dlang/code/phobos/std/functional.d
```

This being said, I find it an interesting discovery, but I don't know if it's compelling enough to move the instantiation costs to the user.
IMHO, it is. What do you guys think?

wilzbach:
Closing as this is [a dead-end](https://github.com/dlang/phobos/pull/6843#issuecomment-458669246). We should try to fix the compiler or compilation strategies instead.

That being said there are a few parts of Phobos (e.g. `std.uni`) which are notoriously expensive to import. Fixing the import time of these modules (and thus of all modules that import them), would be a nice improvement until we can fix the root problem.

andralex:
After some thinking I agree with @wilzbach - picture trying to teach people "to accelerate compilation, add empty parens to function definitions"

andralex:
Will take this to @WalterBright - the compiler has no business compiling traditional function definitions while the module is being imported.

quickfur:
@andralex  In fact, it does *not* accelerate compilation. If anything, it *slows down* compilation (albeit only by a miniscule, probably unnoticeable amount) because now the compiler has to instantiate it as a template instead of just compiling it straight. Unless the function is never called, in which case it's kind of pointless: you might as well just delete the function instead.

Templatizing functions have other benefits, and I have been doing it more lately, but compilation speed is hardly one of them.

andralex:
Oh, I thought that was compilation while std.complex is being imported, not compiled. I think this is indeed a no-action-at all topic then - compiling std.compilex needs to pull those dependencies because it must produce code. Apologies for the confusion. cc @edi33416

andralex:
Oops, I was mistaken about being mistaken. In fact @edi33416 did measure an acceleration in compilation when std.complex is simply being imported (see his comment below the opener). This is definitely a performance bug in the compiler. Again, the compiler should have no business producing code from a module while the module is merely being imported.

PR #6844, opened 2019-01-29
Add support for RISC-V 32 & 64 bit

davidlt:
This allows libphobos to compiled on RISC-V targets.

I can run hello world and word count application on riscv64 using Fedora/RISCV disk images. D is used from GCC 9 upstream.

Signed-off-by: David Abdurachmanov <david.abdurachmanov@gmail.com>


kinke:
The asm parts are GDC specific and should thus not land here upstream. Upstream `std.math` is free from GDC/LDC specifics.

ibuclaw:
(Eventually we should split std.math into asm and generic parts, same as std.bigint).

But yes, the IASM parts are ok to commit to mainline and don't require upstreaming.

wilzbach:
CI is failing as apparently Dscanner can't handle the GDC specific ASM blocks:

```
generated/dscanner-383fcb84d892e5169c134e282878ee2c51e4265f/dsc --config .dscanner.ini --styleCheck etc std -I.
std/math.d(5608:17)[error]: identifier or `align` expected
std/math.d(5608:45)[warn]: Empty asm instruction
std/math.d(5639:17)[error]: identifier or `align` expected
std/math.d(5639:49)[warn]: Empty asm instruction
std/math.d(6232:17)[error]: identifier or `align` expected
std/math.d(6232:41)[warn]: Empty asm instruction
std/math.d(6274:17)[error]: identifier or `align` expected
std/math.d(6274:46)[warn]: Empty asm instruction
posix.mak:556: recipe for target 'dscanner' failed
```

ErnyTech:
GDC uses the GNU AS assembly syntax while DMD uses an Intel like syntax, so you need to convert the assembly blocks to the syntax used by DMD

ibuclaw:
No point having asm code that dmd will never support IMO. Converting it to dmd style is worse than current style.

n8sh:
Rebased onto master and used `mixin` to sidestep D-scanner.

kinke:
Thx. - With LDC now supporting GDC-style asm too, I withdraw my previous concerns. :]

PR #6845, opened 2019-01-29
std/exception lazily import module deps

edi33416:
Using Dgame's [version](https://forum.dlang.org/post/debrounbfusnlfejxnuq@forum.dlang.org) of Daniel Nielsen's [import idiom](https://dlang.org/blog/2017/02/13/a-new-import-idiom/) we can see a drop in the number of imported modules for a simple used program.

Given the following `test.d` used file
```
import std.exception;

void main(string[] args)
{
}
```

the number of imports drops from **9**
```
$ dmd -v -c test.d | grep '^import\ '
import    object	(/home/edis/workspace/dlang/code/druntime/import/object.d)
import    core.internal.hash	(/home/edis/workspace/dlang/code/druntime/import/core/internal/hash.d)
import    core.internal.traits	(/home/edis/workspace/dlang/code/druntime/import/core/internal/traits.d)
import    core.internal.convert	(/home/edis/workspace/dlang/code/druntime/import/core/internal/convert.d)
import    std.exception	(/home/edis/workspace/dlang/code/phobos/std/exception.d)
import    std.range.primitives	(/home/edis/workspace/dlang/code/phobos/std/range/primitives.d)
import    std.traits	(/home/edis/workspace/dlang/code/phobos/std/traits.d)
import    std.meta	(/home/edis/workspace/dlang/code/phobos/std/meta.d)
import    std.functional	(/home/edis/workspace/dlang/code/phobos/std/functional.d)
```
to a total of **5**
```
$ dmd -v -c test.d | grep '^import\ '
import    object	(/home/edis/workspace/dlang/code/druntime/import/object.d)
import    core.internal.hash	(/home/edis/workspace/dlang/code/druntime/import/core/internal/hash.d)
import    core.internal.traits	(/home/edis/workspace/dlang/code/druntime/import/core/internal/traits.d)
import    core.internal.convert	(/home/edis/workspace/dlang/code/druntime/import/core/internal/convert.d)
import    std.exception	(/home/edis/workspace/dlang/code/phobos/std/exception.d)
```

You can see that this is the minimum number of imports that the user will pay for, when he's not using anything from `std.exception`, as the first 4 come by default through `object`

### DIFF
--- a/std/exception.d
+++ b/std/exception.d
@@ -42,6 +42,58 @@ $(TR $(TD Other) $(TD
  +/
 module std.exception;
 
+private
+{
+
+template isModuleImport(string import_)
+{
+    enum isModuleImport = __traits(compiles, { mixin("import ", import_, ";"); });
+}
+
+template isSymbolInModule(string module_, string symbol)
+{
+    static if (isModuleImport!module_) {
+        enum import_ = module_ ~ ":" ~ symbol;
+        enum isSymbolInModule = __traits(compiles, { mixin("import ", import_, ";"); });
+    } else {
+        enum isSymbolInModule = false;
+    }
+}
+
+template FailedSymbol(string symbol, string module_)
+{
+    auto FailedSymbol(Args...)(auto ref Args args)
+    {
+        static assert(0, "Symbol \"" ~ symbol ~ "\" not found in " ~ module_);
+    }
+}
+
+struct FromImpl(string module_)
+{
+    template opDispatch(string symbol)
+    {
+        static if (isSymbolInModule!(module_, symbol)) {
+            mixin("import ", module_, "; alias opDispatch = ", symbol, ";");
+        } else {
+            static if (module_.length == 0) {
+                enum opDispatch = FromImpl!(symbol)();
+            } else {
+                enum import_ = module_ ~ "." ~ symbol;
+                static if (isModuleImport!import_) {
+                    enum opDispatch = FromImpl!(import_)();
+                } else {
+                    alias opDispatch = FailedSymbol!(symbol, module_);
+                }
+            }
+        }
+    }
+}
+
+enum from = FromImpl!null();
+enum _std = from.std;
+
+}
+
 /// Synopis
 @system unittest
 {
@@ -108,9 +160,6 @@ module std.exception;
     immutable res = assumeUnique(str);
     assert(res == "immutable");
 }
-
-import std.range.primitives;
-import std.traits;
 
 /++
     Asserts that the given expression does $(I not) throw the given type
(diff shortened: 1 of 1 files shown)


std/exception.d
```diff
@@ -1666,7 +1719,8 @@ class ErrnoException : Exception
         errorHandler.
 +/
 //lazy version
-CommonType!(T1, T2) ifThrown(E : Throwable = Exception, T1, T2)(lazy scope T1 expression, lazy scope T2 errorHandler)
+_std.traits.CommonType!(T1, T2) ifThrown(E : Throwable = Exception, T1, T2)(lazy scope T1 expression,
```
dnadlinger: I might write this as am explicit eponymous template instead.

Geod24:
> from should go in object.d

Errrrr... While it sounds like a nice idea at first, `object_.d` is host of symbols which are essential to the language (there are some [old exceptions](https://dlang.org/library/object/assume_safe_append.html)). `From` is a nice idiom, but by no way mandatory, so putting it in `object_.d` would only achieve convenience.

Basile-z:
@Geod24 i agree that object.d is not ideal. So let's think the problem differently : this idiom is useful then assuming that it will be used let's put it in a dedicated module. The module still has to be imported but so what. What should be avoided is to put several copy across the library. Ideally the new module should be a zero cost import : only templates, only cool stuff...

wilzbach:
For the protocol, I strongly believe that before we go and push the import idiom as best practice, we should really evaluate how hard it would be to make selective or static imports lazy (and of course how much overhead we will pay for this from idiom). Or you know [implement this idiom in the compiler](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1005.md).

std/exception.d
```diff
@@ -42,6 +42,58 @@ $(TR $(TD Other) $(TD
...
+
+template FailedSymbol(string symbol, string module_)
+{
+    auto FailedSymbol(Args...)(auto ref Args args)
+    {
+        static assert(0, "Symbol \"" ~ symbol ~ "\" not found in " ~ module_);
+    }
+}
+
+struct FromImpl(string module_)
```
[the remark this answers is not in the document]
andralex: The idea here is to keep this private until we develop some experience with it. Then, yes, we can publish it at an upper level.

std/exception.d
andralex: @dnadlinger wie meinst du das?

andralex:
>For the protocol, I strongly believe that before we go and push the import idiom as best practice, we should really evaluate how hard it would be to make selective or static imports lazy as this would immediately have a huge impact on _all_ D code out there (and of course how much overhead we will pay for this from idiom). Or you know [implement this idiom in the compiler](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1005.md).

Clearly this implementation could and should inform compiler improvements. Until then it is definitely welcome.

andralex:
I think the implementation should be put in its own small module (for now), to be usable without other cascading imports. That somewhat defeats its purpose because we want to open fewer files, not more. But for now it'll do.

So this should go in std/internal/autoimport.d, then other Phobos modules would import this and then use it. If our experience is positive, we move it and publish.

RazvanN7:
It seems that now the import idiom will be merged in druntime. I assume that we will start to use it in phobos, however, this PR is now obsolete.

PR #6846, opened 2019-01-31
std.traits: use assumePurein the public example

wilzbach:
Otherwise nothing gets tested.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -2446,6 +2446,19 @@ if (is(T == function))
         enum attrs = functionAttributes!T | FunctionAttribute.pure_;
         return cast(SetFunctionAttributes!(T, functionLinkage!T, attrs)) t;
     }
+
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -2435,17 +2435,30 @@ if (is(T == function))
...
 }
 
-///
+/// Use with extreme caution as this can easily result in Undefined Behavior
 @safe unittest
 {
     alias ExternC(T) = SetFunctionAttributes!(T, "C", functionAttributes!T);
 
-    auto assumePure(T)(T t)
+    auto assumePure(T)(T t) @trusted
```
ntrel: This is reinforcing the dangerous idea of exposing unsafe functions as trusted. Shouldn't the example have `g()` trusted instead? Then we can remove the doc-comment warning.

PR #6848, opened 2019-01-31
Use moved core.lifetime.move{,Emplace} for std.algorithm.mutation.move{,Emplace}

wilzbach:
In https://github.com/dlang/druntime/pull/2442, `doesPointTo` wasn't ported as it's quite big (https://github.com/dlang/druntime/pull/2447).

That's why one test which test this behavior would fail now. Should we simply move `doesPointTo` to druntime as proposed, s.t. we don't need to maintain two implementations?

### DIFF
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -1059,11 +1059,15 @@ Params:
 */
 void move(T)(ref T source, ref T target)
 {
-    // test @safe destructible
-    static if (__traits(compiles, (T t) @safe {}))
-        trustedMoveImpl(source, target);
-    else
-        moveImpl(source, target);
+    import core.lifetime : coreMove = move;
+    return coreMove(source, target);
+}
+
+@safe unittest//Issue 6217
+{
+    import std.algorithm.iteration : map;
+    auto x = map!"a"([1,2,3]);
+    x = move(x);
 }
 
 /// For non-struct types, `move` just performs `target = source`:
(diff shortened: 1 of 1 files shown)


std/algorithm/mutation.d
```diff
@@ -2960,12 +2667,12 @@ if (isBlitAssignable!T && !is(typeof(lhs.proxySwap(rhs))))
     swap(a, b);
 
     //Note: we can catch an error here, because there is no RAII in this test
-    import std.exception : assertThrown;
-    void* p, pp;
-    p = &p;
-    assertThrown!Error(move(p));
-    assertThrown!Error(move(p, pp));
-    assertThrown!Error(swap(p, pp));
+    //import std.exception : assertThrown;
```
edi33416: Why not delete instead of commenting?

wilzbach: I wanted to see whether this is the only issue.
As explained, the core.lifetime functions are technically different and using them is a breaking change.

LightBender:
@thewilsonator this one is for you.

0xEAB:
This one is painful because `core.lifetime` didn’t do its homework and breaks compatibility with existing code.
See: <https://github.com/dlang/dmd/blob/35f1146a1b537e9a361fa28098a1cfc9a384178f/druntime/src/core/lifetime.d#L2192> 

Which is particularly painful to fix/emulate downstream because of the `if (false) { nonTrustedImpl(); } trustedImpl();` `@safe`ty hacks used in the code that are inaccessibly from the outside (`private` “impl” functions).

0xEAB:
```
std/typecons.d(3988): Error: cannot read uninitialized variable `payload` in CTFE
            () @trusted { moveEmplace(value, _value.payload); }();
```
Also looks like the druntime implementation now lacks a bit of CTFE-ability that the Phobos one has.

0xEAB:
I figure, emulating the previous behavior might be a hopeless endeavor.

Everything is built around `moveEmplaceImpl` (the incompatible function in question), so in order to sneak in a wrapper around that, pretty much everything else this PR tries to remove has to stay in with `moveEmplaceImpl` calls directed to the emulator.

AFAICT a more reasonable way forward would be to fix `core.lifetime` by restoring the missing behavior of `emplaceMove()` from Phobos. Though, I’m afraid that would be a good chunk of work.

PR #6852, opened 2019-02-02
std.regex: Fix reassignment of ctRegex wrapper structs

kinke:
Previously, reassigning such a wrapper resulted in the `static immutable Regex!Char` global for that pattern (`staticRe`) to be overwritten by the global for the new pattern (!). Made possible via `alias this` and the logical-const hack.

Proposed fix: keep on allowing reassignments (people are already using them), but just point to another pattern's global. So use a common Wrapper struct type wrapping a pointer instead of an empty struct type per pattern.

Bug reported in https://github.com/ldc-developers/ldc/issues/2961.

### DIFF
--- a/std/regex/package.d
+++ b/std/regex/package.d
@@ -423,6 +423,16 @@ if (isSomeString!(S))
 }
 
 
+private struct CTRegexWrapper(Char)
+{
+    private immutable(Regex!Char)* re;
+
+    // allow code that expects mutable Regex to still work
+    // we stay "logically const"
+    @property @trusted ref getRe() const { return *cast(Regex!Char*) re; }
+    alias getRe this;
+}
+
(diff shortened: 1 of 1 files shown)


std/regex/package.d
```diff
@@ -457,6 +460,17 @@ template ctRegexImpl(alias pattern, string flags=[])
...
 
+@safe unittest
+{
+    auto re = ctRegex!`foo`;
+    assert(matchFirst("foo", re));
+
+    // test reassignment
+    re = ctRegex!`bar`;
+    assert(matchFirst("bar", re));
+    assert(!matchFirst("bar", ctRegex!`foo`));
```
kinke: This one currently fails; `ctRegex!"foo"` behaves as `ctRegex!"bar"` after the reassignment.

PR #6853, opened 2019-02-04
Fix issues 19644, 19647 - std.range.takeOne.opSlice bugs

jondegenhardt:
This PR fixes a pair of bugs in the `std.range.takeOne.opSlice` implementation. This implementation is used when the range passed to `takeOne` does not support slicing.

Issue 19644 is that the implementation incorrectly asserts when the slice `end` passed to `opSlice` equals the length. This was fixed by checking `end <= length` rather than `end < length`. In addition, the text of the assert messages generated was changed to be more consistent with similar `opSlice` asserts in `package.d`.

Issue 19647 is that the implementation returns a non-empty range if an empty slice (e.g. `[0 .. 0]` is requested. This is fixed by changing the `empty` argument passed to the `takeOne` constructor function from `false` to `true`.

These issues were identified while writing code coverage unit tests for this implementation. There are additional unit tests adding code coverage beyond the specifics of these two bugs.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -2655,8 +2655,15 @@ if (isInputRange!R)
             }
             auto opSlice(size_t m, size_t n)
             {
-                assert(m <= n && n < length, "Attempting to index a takeOne out of bounds");
-                return n > m ? this : Result(_source, false);
+                assert(
+                    m <= n,
+                    "Attempting to slice a takeOne range with a larger first argument than the second."
+                );
+                assert(
+                    n <= length,
+                    "Attempting to slice using an out of bounds index on a takeOne range."
+                    );
+                return n > m ? this : Result(_source, true);
             }
             // Non-standard property
             @property R source() { return _source; }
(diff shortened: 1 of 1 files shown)


std/range/package.d
```diff
@@ -2688,16 +2695,124 @@ pure @safe nothrow @nogc unittest
     struct NonForwardRange
     {
         enum empty = false;
-        int front() { return 42; }
+        @property int front() { return 42; }
```
wilzbach: The `@property` is useless here (same for the ones following). What was the motivation for adding it to the existing test?

std/range/package.d
```diff
@@ -2688,16 +2695,124 @@ pure @safe nothrow @nogc unittest
...
+
+    thrown = false;
+    try s.popFront;
+    catch (AssertError) thrown = true;
+    assert(thrown);
+
+    thrown = false;
+    try s.popBack;
+    catch (AssertError) thrown = true;
+    assert(thrown);
```
wilzbach: This happens fairly often. Why not move it in a function?
Sth. like e.g. works:

```d
import core.exception : AssertError;

void assertThrown(E)(lazy E exp)
{
    bool thrown;
    try
    {
        exp();
    }
    catch (AssertError)
    {
        thrown = true;
    }
    assert(thrown);
}

void main()
{
    assertThrown(assert(0));
    assertThrown(1 + 1);
}
```

https://run.dlang.io/is/rkYXbv

std/range/package.d
jondegenhardt: Standardization. Just about all the other front() routines in the file are `@property`. It was a superfluous change. I can undo if desired.

std/range/package.d
jondegenhardt: Nice idea. I do think there is a small disadvantage of using a relatively uncommon language feature (lazy function parameter evaluation), but I agree that it would save some space in the code. After looking at a few of these, I'm starting to think more unit tests should be covering the assert cases.

std/range/package.d
jondegenhardt: The `@property` attributes have been removed.

std/range/package.d
jondegenhardt: Turns out the it's not necessary to write a function. These tests can be written by passing `AssertError` as a template parameter to `assertThrown`. For example:
```
   assertThrown!AssertError(s.front);
```
These unit tests have been rewritten in this style.

n8sh:
Merging as @wilzbach's review has been addressed.

PR #6854, opened 2019-02-05
optimize several std.ascii functions

Basile-z:
- most isXXXX can be tweaked to generate a single `CMP`
- `toLower` and `toUper` can use the `XOR` trick (that's not really an opt, since `ADD` and `XOR` have the same timing _in fine_) 
- add `pragma(inline)` because despite of the optimization DMD backend will push/pop RBP

See the [codegen](https://d.godbolt.org/z/noNUEv) of the current VS the new, including the infuriating PUSH/POP, but that's another piece of cake to fix that...

### DIFF
--- a/std/ascii.d
+++ b/std/ascii.d
@@ -178,9 +178,10 @@ bool isAlpha(dchar c) @safe pure nothrow @nogc
     Params: c = The character to test.
     Returns: Whether `c` is a lowercase ASCII letter (a .. z).
   +/
+pragma(inline, true)
 bool isLower(dchar c) @safe pure nothrow @nogc
 {
-    return c >= 'a' && c <= 'z';
+    return c  - 'a' <=  'z' - 'a' ;
 }
 
 ///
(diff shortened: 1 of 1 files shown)


jondegenhardt:
Do you have performance benchmarks comparing the new and old forms, or perhaps references to other studies or systems that have adopted these forms? I'm not doubting there is a solid basis for these modifications. Just that it'd be useful to have them available for review as part the PR.

Basile-z:
No I haven't performed any benchmark. Note that on compiler explorer you can select LDC2 or GDC and see that their backends already did the right thing with the former function so this PR is really just to help a bit DMD.

jondegenhardt:
Thanks. I was just checking that as well. Thanks for publishing the compiler explorer view, that was very helpful. One thing that caught my eye was that the new form is relying on wrap-around arithmetic, which is less obvious at the source level. So if LDC/GDC are optimizing correctly, the better tradeoff may be to have the more obvious source code form.

quickfur:
I agree with @wilzbach, we haven't been optimizing for dmd's subpar optimizer, and we shouldn't start to.  Let's keep such optimizations to those that also affect ldc/gdc please.

PR #6856, opened 2019-02-08
Make currTime/currStdTime use gettimeofday on generic posix

n8sh:
This is to decrease work required to port D to new operating systems.

### DIFF
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -211,23 +211,7 @@ public:
             static import core.stdc.time;
             enum hnsecsToUnixEpoch = unixTimeToStdTime(0);
 
-            version (OSX)
-            {
-                static if (clockType == ClockType.second)
-                    return unixTimeToStdTime(core.stdc.time.time(null));
-                else
-                {
-                    import core.sys.posix.sys.time : gettimeofday, timeval;
-                    timeval tv = void;
-                    // Posix gettimeofday called with a valid timeval address
-                    // and a null second parameter doesn't fail.
-                    gettimeofday(&tv, null);
-                    return convert!("seconds", "hnsecs")(tv.tv_sec) +
-                           tv.tv_usec * 10 +
-                           hnsecsToUnixEpoch;
-                }
-            }
-            else version (linux)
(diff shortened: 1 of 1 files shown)


n8sh:
>You really don't want `gettimeofday` to be used unless there's no other choice, and with this change, that's not going to get caught. 

I try to strike a balance here by emitting a message:
```d
version (OSX) { /+ macOS does not have clock_gettime. +/ }
else
{
    pragma(msg, __FUNCTION__ ~ " has no specialization " ~
        "for target operating system indicating whether " ~
        "clock_gettime is available so instead using " ~
        "gettimeofday. Consider opening an enhancement " ~
        "request at https://issues.dlang.org/.");
}
```

jmdavis:
> I try to strike a balance here by emitting a message:

The reality of the matter is that it's highly unlikely that anyone is going to see that. The build just puts out too much output - especially when running the tests. Walter is very firm about us not assuming that OS-specific stuff is platform independent, which is why druntime doesn't use `version(Posix)` much even though that could theoretically reduce the number of declarations significantly. It just makes it too easy to have problems that get missed. You really want folks porting D to a new platform to be forced to look at this stuff and make the code correct for their platform rather than assuming that something will just work across platforms. It's true that that can make the work more tedious, but it makes it much easier to catch problems while porting - many of which would probably go unnoticed for a long time otherwise.

If someone wants to port this code to a new platform, and they're lazy enough to want to just use `gettimeofday` instead of trying to figure out the best way to do it on the new platform, then it will be trivial enough to copy the OSX version, whereas if this doesn't produce an error, even if someone wants to take the time to get it right, there's a high chance that they'll miss it if this does not result in a compilation error. The pragma does increase the chances that the problem will be caught, but in the process, you lose the guarantee that it will be caught, and with all of the output you get with a typical build, I'd be surprised if anyone noticed it - especially when you consider that a newer platform is going to have even fewer people looking at the build output.

RazvanN7:
@n8sh I agree with @jmdavis here. A pragma is not sufficient to warn the user. Should we close this?

PR #6858, opened 2019-02-09
add -preview=fieldwise to build

WalterBright:
Upgrade to spec compliant struct compares

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -135,7 +135,7 @@ endif
 
 # Set DFLAGS
 DFLAGS=
-override DFLAGS+=-conf= -I$(DRUNTIME_PATH)/import $(DMDEXTRAFLAGS) -w -de -preview=dip1000 -preview=dtorfields $(MODEL_FLAG) $(PIC)
+override DFLAGS+=-conf= -I$(DRUNTIME_PATH)/import $(DMDEXTRAFLAGS) -w -de -preview=dip1000 -preview=dtorfields -preview=fieldwise $(MODEL_FLAG) $(PIC)
 ifeq ($(BUILD),debug)
 override DFLAGS += -g -debug
 else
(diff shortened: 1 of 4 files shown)


WalterBright:
Blocked by https://github.com/dlang/dmd/pull/9338

ZombineDev:
This change will require upgrading the host compiler version to what would become 2.085.0.

wilzbach:
> This change will require upgrading the host compiler version to what would become 2.085.0.

No, DMD latest is always used to build Phobos and druntime. The host compiler is only necessary for DMD and some build scripts.

ZombineDev:
Never mind, for some reason I read `dlang/dmd`, instead of `dlang/phobos` 😄

thewilsonator:
> Blocked by [dlang/dmd#9338](https://github.com/dlang/dmd/pull/9338)

9338 was merged, but the is still all red, is it still blocked?

ZombineDev:
Rebased

thewilsonator:
Thanks!

thewilsonator:
`dsymbol/src/dsymbol/symbol.d(198): Deprecation: Cannot use `alias this` to partially initialize variable `part` of type `SymbolOwnership`. Use `part.ptr``

ZombineDev:
@thewilsonator that issues comes from here:
https://github.com/dlang/phobos/blob/aef8bc7a28b528066c7c1bd8981ec36305b518a9/posix.mak#L550
In other words, building phobos with this flag causes the flag to be passed to dscanner, which is a tool built by the `style` target for linting D code.
@Basile-z can you look into this?

The other errors seemed to be "statement not reachable" when building `std.variant`.

ZombineDev:
After restarting CircleCI, the Dscanner segfault went away (so as expected it was an intermittent failure). The new error is:
```
std/variant.d(270): Warning: statement is not reachable
```

MoonlightSentinel:
Needs a rebase

RazvanN7:
@WalterBright rebase this and let's see if we can get it in.

RazvanN7 (approved):

PR #6863, opened 2019-02-14
fix to follow the correct promotion rules of the shift expressions

Basile-z:
This is required to make https://github.com/dlang/dmd/pull/9361 pass so a decision must be took for the front-end change, even before its tests are green.


thewilsonator:
cc @n8sh

PR #6866, opened 2019-02-17
SortedRange.opSlice() should be return scope

WalterBright:
Progress to -dip1000 compatibility.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -10493,7 +10493,7 @@ if (isInputRange!Range && !isInstanceOf!(SortedRange, Range))
 
     /// Ditto
     static if (hasSlicing!Range)
-        auto opSlice(size_t a, size_t b)
+        auto opSlice(size_t a, size_t b) return scope @trusted
         {
             assert(
                 a <= b,


WalterBright:
buildkite/phobos fails with:
```
generated/linux/release/64/unittest/test_runner rt.typeinfo.ti_wchar
--
  | generated/linux/release/64/unittest/test_runner etc.linux.memoryerror
  | 0.000s PASS release64 etc.linux.memoryerror
  | done
  | make[1]: Leaving directory '/var/lib/buildkite-agent/builds/ci-agent-8331da0f-b683-48a8-9b92-9391bb7a3ec9-7/dlang/phobos/druntime'
  | 🚨 Error: The command exited with status 2
```
What does that mean?

WalterBright:
Blocking https://github.com/dlang/dmd/pull/9220

wilzbach:
No, the error is:

```
generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std5range__T11SortedRangeTAkVAyaa6_61203c3d2062ZQBk__T10lowerBoundVEQCrQCq12SearchPolicyi3TyiZQBoMFNaNbNiNfyiZSQEiQEh__TQEeTQDuVQDva6_61203c3d2062ZQFf: error: undefined reference to '_D3std5range__T11SortedRangeTAkVAyaa6_61203c3d2062ZQBk7opSliceMFNaNbNiNfmmZSQCwQCv__TQCsTQCiVQCja6_61203c3d2062ZQDt'
--
```

We went throw this numerous times. `scope` breaks the binary compatibility and thus the linking fails.
The problem here is that on Buildkite the druntime benchmark target (which is not run on the auto-tester) doesn't recompile Phobos.

tl;dr: it's safe to merge (in terms of this particular Buildkite error)

std/range/package.d
```diff
@@ -10493,7 +10493,7 @@ if (isInputRange!Range && !isInstanceOf!(SortedRange, Range))
 
     /// Ditto
     static if (hasSlicing!Range)
-        auto opSlice(size_t a, size_t b)
+        auto opSlice(size_t a, size_t b) return scope @trusted
```
wilzbach: Shouldn't the compiler be able to infer the safety here automatically?

aG0aep6G: What's more, `Range.opSlice` might be unsafe. This can't be `@trusted`.

aG0aep6G: > What's more, `Range.opSlice` might be unsafe. This can't be `@trusted`.

https://github.com/dlang/phobos/pull/6869

jondegenhardt: I read this PR earlier and was confused by this as well. The `opSlice` being called could easily be exported by a range that I wrote, not even a range in Phobos.

WalterBright: It's not always able to, often due to circular references. Phobos is full of circular templates.

jondegenhardt: Consider adding comments to the code with an explanation of why it's needed. If appropriate, the circumstances under which it could be removed. If it's a long explanation, add it to the PR and reference it from the code. Otherwise future readers of the code will be left wondering.

PR #6867, opened 2019-02-17
Mark etc.c.sqlite3 as deprecated

ibuclaw:
(I'm going to start this debate by opening a pull request).

Bindings are for 3.10 which is outdated for some years now, the current upstream version of sqlite3 is 3.28.

For up-to-date bindings, there are many external repositories in code.dlang.org that provide them, such as d2sqlite or derelict.

If no one comes forward to maintain it, they will be removed.

### DIFF
--- a/etc/c/sqlite3.d
+++ b/etc/c/sqlite3.d
@@ -34,6 +34,9 @@ module etc.c.sqlite3;
 
 import core.stdc.stdarg : va_list;
 
+// @@@DEPRECATED_2.090@@@
+deprecated("etc.c.sqlite3 is unmaintained, and will be removed in 2.090"):
+
 extern (C) __gshared nothrow:
 
 /**


CyberShadow:
The bindings work fine even against newer SQLite versions. What's the problem? Has the C interface received any critical updates that make updating the bindings important?

I personally rely heavily on this module. Removing it from Phobos will complicate building a lot of my software. Thus, I vote against removing it.

Also, I'll argue that SQLite is useful enough that it's worth keeping it in Phobos. At this point, SQLite has become the universal standard for embedded databases, being the standard database format on Android and JavaScript (WebSQL).

CyberShadow:
Plenty of use in the wild:
https://github.com/search?l=D&q=%22etc.c.sqlite3%22&type=Code

ibuclaw:
To paraphrase what has been said about every other module inside `etc/c/*`:

Why have C bindings in Phobos when a dub package will more than suffice, especially when there lacks any standard high level API that uses them.

CyberShadow:
> Why have C bindings in Phobos when a dub package will more than suffice, especially when there lacks any standard high level API that uses them.

- This applies to new modules, but not existing ones.
- Not everyone can use or wants to use Dub.
- I don't think there is such a thing as a "standard" high level API. You can write a wrapper in many ways, but generally, a binding in just one way. (The only other way I can think of is to load the library at runtime, which is what Derelict does, but that is rarely a hard requirement.)
- You can still use bindings directly, without a "high level wrapper", especially in `-betterC` programs.

ibuclaw:
> This applies to new modules, but not existing ones.

I'll have to check, but I'd expect this was also the case at the time of introduction too.

Usually things get accepted on the basis that a follow-up will occur at some indeterminate point in the future where a new std phobos module will start using them.  5 years on however, and they still remain unused.

This module has found no purpose in phobos, it is therefore a failed module.  The bindings should be relegated at best to deimos/derelict/dub, and at worse druntime.

CyberShadow:
> This module has found no purpose in phobos, it is therefore a failed module.

There is no logic in that statement.

Repeating myself:

>  https://github.com/search?l=D&q=%22etc.c.sqlite3%22&type=Code

This module is actively used in D code, even in new projects. As far as I can see, that is sufficient reason to end this discussion and close this PR.

CyberShadow:
@ibuclaw Are you confusing `etc.c` with `std.internal`, by chance?

CyberShadow:
> This module has found no purpose in phobos, it is therefore a failed module.

I think I now understand what you're trying to say, though I wouldn't use that choice of wording. If we assume that the intended purpose of the `etc` package is solely to provide bindings so that they can be wrapped elsewhere in Phobos, then, yes, `etc.c.sqlite3` hasn't fulfilled that purpose. But, that didn't prevent from the module being successfully used directly in many D projects. Moving it elsewhere would make things a bit tidier, but I don't see how the cost of such a breaking change would justify the gain.

I can even find D code that uses `etc.c.curl` without `std.net.curl`, so I think existing `etc` modules should be evaluated on their own merit, and not whether they've fulfilled some other role within Phobos.

CyberShadow:
> and at worse druntime

I recall we've moved a few C-like modules from Phobos to Druntime previously (with aliases lasting for a long time), so we could do it here as well, though I don't see much point other than making things a tiny bit tidier.

wilzbach:
> . 5 years on however, and they still remain unused.
>
> This module has found no purpose in phobos, it is therefore a failed module. The bindings should be relegated at best to deimos/derelict/dub

I couldn't say it better. That's where these bindings belong.

CyberShadow:
If we decide to move forward with eventually removing this module, I think the first step should be to de-document it, so that newer projects (see e.g. the Hunt framework) don't begin depending on it.

jondegenhardt:
> For up-to-date bindings, there are many external repositories in code.dlang.org that provide them, such as d2sqlite or derelict.

Putting on my corporate developer hat - It is highly valuable to have libraries containing basic functionality that come from relatively well known and trusted sources. Libraries shipped with the compiler are one the best sources. There are of course many good packages in code.dlang.org, but overall the quality is quite mixed and hard to discern for individual modules without doing your own investigation (time consuming). The rating system introduced on code.dlang.org last year was a valuable improvement, but there is still a way to go.

I don't have an informed opinion on whether SQLite falls into this category of basic functionality, but it feels like it does. My suggestion when evaluating depreciation is to consider it from this perspective. That is, is there a common need for a trusted, maintained form of this functionality, and if so, where does it live and how do developers find it? These are obviously difficult tradeoff choices due to the limited developer resources available.

> Not everyone can use or wants to use Dub.

That's been my experience. Specifically, Dub is not flexible enough to support many of the things I've found necessary to do at build time. This does make it more difficult to use code.dlang.org packages. This feels very solvable though, and Dub has been getting materially better over the last year or so.

wilzbach:
This needs definitely @andralex's approval.
My opinion on this is that we at least should undocument it if we can't maintain it properly.

CyberShadow:
>  My opinion on this is that we at least should undocument it if we can't maintain it properly.

If the problem is just that, I wouldn't mind maintaining it (I'm already in CODEOWNERS for it, which is how I was notified of this PR). Currently there's no way to subscribe to new bugs filed against a Phobos module, but checking, there aren't any against this module. Anyway, as far as I can tell, there's nothing wrong with the current version, other than that it doesn't have the latest binding definitions which the great majority of programs don't need anyway. I can look into updating the bindings myself if it would help with your perception of the situation.

CyberShadow:
> I personally rely heavily on this module. Removing it from Phobos will complicate building a lot of my software. Thus, I vote against removing it.

BTW, I realized after posting that that I could just copy-paste this module into the library which also has the high-level wrapper around this module, so I'm retracting that statement.

andralex:
For every action there should be a benefit. What is the benefit of this initiative?

ErnyTech:
I intend to update etc.c.sqlite3 to the latest version of sqlite to keep the support within the phobos

I think it is better to update what we already have than to remove it, also to maintain support without breaking the projects of others

wilzbach:
(rebased to master to fix the CIs)

wilzbach:
> For every action there should be a benefit. What is the benefit of this initiative?

- People don't use stuff that's outdated/unsupported and get a bad image of D not being production ready
- People don't run into bugs because they use unsupported old bindings
- Less maintenance work when porting D to other platforms
- Bindings should be generated automatically with DPP or dstep not maintained by hand by us (bad reference)

CyberShadow:
There is nothing wrong with these bindings. They work just fine.

CyberShadow:
> Bindings should be generated automatically with DPP or dstep not maintained by hand by us (bad reference)

I agree with this in general, but I don't think such ports would ever be perfect, as they would miss qualifiers such as `@nogc` (applicable to everything that doesn't call back into D code), `pure`, or `@safe` / `@system`. These are applied appropriately to `core.stdc`. Therefore, manual ports such as this one will probably have a higher quality (barring porting mistakes).

CyberShadow:
> People don't run into bugs because they use unsupported old bindings

- SQLite has a stable ABI, which allows old header files / bindings to continue working with newer library binaries.
- As already mentioned, there are no bugs filed against this module.

Can you provide an example of a bug that people would run into?

jmdavis:
> For up-to-date bindings, there are many external repositories in code.dlang.org that provide them, such as d2sqlite or derelict.

Even if we all agree that these bindings should be removed from Phobos (and thus should be deprecated), unless any of those projects provides compatible bindings such that all that is required is to grab that project and change the imports in your own code, then these bindings need to be made available elsewhere (presumably in deimos) with just the package name changed (if these bindings had come along later, that's where they would have been anyway). Then any deprecation message can point to the correct place to get the code, and anyone who uses these bindings can easily update their code.

wilzbach:
Well, we can just move them in the undead package. Would that work for everyone?
The upside is that people can actually pin their version or upgrade immediately - based on their preference, but not on the language release cycle.

CyberShadow:
> Well, we can just move them in the undead package. Would that work for everyone?

No? The arguments above still stand.

Note that https://github.com/dlang/phobos/pull/6997 exists which fixes the reason why this PR was opened.

CyberShadow:
If we decide to remove this module from Phobos, the process should be as follows:

- Merge #6997 first
- Create a Deimos package now, containing a copy of this module
  - (alternatively, move it to UndeaD, but Deimos seems more appropriate)
- Undocument it (remove from dlang.org), to prevent it from being used in new projects
- After a period, deprecate it
- After a period, remove it

However, as #6997 fixes the reason why this module was proposed for deprecation, I don't think there are remaining strong arguments to go forward with it.

CyberShadow:
I believe that a large part of Python's success is its commitment to backwards compatibility. Other than when crossing major version bumps, modules in its standard library generally stay in the standard library, no matter how poor their design has turned out to be in retrospect. Considering this, the arguments above, and now that #6997 has been merged, I really don't see how removing Phobos modules without providing a replacement adds value to D. As such, I hope I'm not doing something overly disagreeable by closing this pull request; but if I am, apologies, and please do reopen (or let me know so that I do that) and we can continue this discussion.

PR #3398, opened 2015-06-10
Add etc.c.odbc

andralex:
Just a little testing with iODBC - could folks on other OSs give it a shot? There are plenty of simple samples online, I went through this:

```D
import etc.c.odbc.sql;
import etc.c.odbc.sqlext;
import std.stdio;

int main() {
  SQLHENV env;
  SQLCHAR driver[256];
  SQLCHAR attr[256];
  SQLSMALLINT driver_ret;
  SQLSMALLINT attr_ret;
  SQLUSMALLINT direction;
  SQLRETURN ret;

  SQLAllocHandle(SQL_HANDLE_ENV, SQL_NULL_HANDLE, &env);
  SQLSetEnvAttr(env, SQL_ATTR_ODBC_VERSION, cast(void *) SQL_OV_ODBC3, 0);

  direction = SQL_FETCH_FIRST;
  while(SQL_SUCCEEDED(ret = SQLDrivers(env, direction,
                       driver.ptr, driver.sizeof, &driver_ret,
                       attr.ptr, attr.sizeof, &attr_ret))) {
    direction = SQL_FETCH_NEXT;
    printf("%s - %s\n", driver.ptr, attr.ptr);
    if (ret == SQL_SUCCESS_WITH_INFO) printf("\tdata truncation\n");
  }

  return 0;
}
```

Running with `rdmd -L-lodbc odbctest.d` should print the installed drivers.

PR #6868, opened 2019-02-17
Mark etc.c.odbc as deprecated

ibuclaw:
Publicly import core.sys.windows.sql instead.

Set removal date for one year in the future (February 2020).


CyberShadow:
CC @andralex as the original author (https://github.com/dlang/phobos/pull/3398).

John-Colvin:
Won’t this immediately break things for unixodbc or iodbc users on non-windows?

ibuclaw:
All have `extern (Windows):` guards at the top of the modules. So non-windows are excluded anyway.

wilzbach:
(rebased to master to fix the CIs)

RazvanN7:
@ibuclaw Should we go on with this deprecation or remove it directly? Either way, can you update the deprecation comment so we can get this in?

ibuclaw:
Bumped deprecation date to 2022, two years after the original remove date, _sigh_.

RazvanN7 (approved):

PR #6869, opened 2019-02-17
revert bad @trusted

aG0aep6G:
Range.opSlice might be unsafe. `_input[a .. b]` can't be @trusted.

Partially reverts #6866. CC @WalterBright @wilzbach


aG0aep6G:
The auto-tester doesn't like this. Full revert then: https://github.com/dlang/phobos/pull/6870

PR #6870, opened 2019-02-18
Revert "SortedRange.opSlice() should be return scope"

aG0aep6G:
Reverts dlang/phobos#6866

Range.opSlice might be unsafe. `_input[a .. b]` can't be `@trusted`.

I tried reverting just the `@trusted` attribute in #6869, but that doesn't seem to work (auto-tester fails).

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -10493,7 +10493,7 @@ if (isInputRange!Range && !isInstanceOf!(SortedRange, Range))
 
     /// Ditto
     static if (hasSlicing!Range)
-        auto opSlice(size_t a, size_t b) return scope @trusted
+        auto opSlice(size_t a, size_t b)
         {
             assert(
                 a <= b,


WalterBright:
There is a reason for this - a chicken-and-egg dependence on https://github.com/dlang/dmd/pull/9220 Neither works without the other. @trusted breaks the stalemate.

WalterBright:
Do not pull this.

wilzbach:
Yeah this clearly needs more discussion. Sorry I was merge-friendly because Walter set the other PR as blocking and the CI failure was somewhat unrelated.

aG0aep6G:
> There is a reason for this - a chicken-and-egg dependence on [dlang/dmd#9220](https://github.com/dlang/dmd/pull/9220) Neither works without the other. @trusted breaks the stalemate.

It's not clear whether you're going to remove `@trusted` again once dlang/dmd#9220 is in. If that's the plan, you should have said so in #6866. If the attribute is going to stay, that's not acceptable.

> Yeah this clearly needs more discussion.

#6866 would have needed more discussion.

WalterBright:
If pulling this results in phobos failing to compile with -dip1000, then this is the wrong solution.

aG0aep6G:
> If pulling this results in phobos failing to compile with -dip1000, then this is the wrong solution.

This was a revert to avoid releasing a safety regression. Maybe it can be fixed more nicely, but I don't know how to.

Unfortunately, Phobos seems to have moved past the point where this single revert worked (cf. auto-tester results for this PR). I have no idea how to avoid the regression now, so I'm closing this.

Here's a Bugzilla entry instead: https://issues.dlang.org/show_bug.cgi?id=19777

wilzbach:
> then this is the wrong solution..

No, now we have a yet another safety hole in Phobos which is another reason to stop using it.

PR #6871, opened 2019-02-18
range.only() returns its parameters

WalterBright:


thewilsonator:
Fails on ae 
```
/var/lib/buildkite-agent/builds/ci-agent-fad97297-da9f-4236-bad4-3b606332cc80-4/dlang/phobos/build/distribution/bin/../imports/std/range/package.d(9712,57): Error: returning `OnlyResult(0LU, 0LU, null).this(_param_0, _param_1, _param_2, _param_3)` escapes a reference to parameter `_param_3`, perhaps annotate with `return`
--
  | utils/path.d(182,21): Error: template instance `std.range.only!(string, string, string, typeof(null))` error instantiating
```

@CyberShadow whats going on here?

CyberShadow:
Exactly as the error message says - `std.range.only!(string, string, string, typeof(null))` now fails to instantiate.

Make sure `only("", null)` works.

std/range/package.d
```diff
@@ -9706,7 +9706,7 @@ Returns:
 
 See_Also: $(LREF chain) to chain ranges
  */
-auto only(Values...)(auto ref Values values)
+auto only(Values...)(return scope auto ref Values values)
 if (!is(CommonType!Values == void) || Values.length == 0)
 {
     return OnlyResult!(CommonType!Values, Values.length)(values);
```
CyberShadow: With `values` as `"", null`:

```
Error: returning OnlyResult(0LU, 0LU, null).this(_param_0, _param_1) escapes a reference to parameter _param_1, perhaps annotate with return
```

WalterBright:
I think I know what is wrong, working on it.

PR #6872, opened 2019-02-19
Fix issue 19686 - sgn is too greedy

Biotronic:


wilzbach:
A unittest should have been added here to avoid regressions in the future... ;-)

PR #6873, opened 2019-02-20
Fix issue 19689: don't use recursive template parameter iteration in std.traits.

FeepingCreature:
Pull inner Impl templates out into std.traits so they're only instantiated once.
Use anySatisfy, staticMap instead of manually recursing.

Recursive template parameter iteration completely fails for large types. By using `staticMap`, we take advantage of its logarithmic recursion. `anySatisfy` is also inherently more efficient.

It's shorter too.

For an example of the problems the current approach causes, consider

```
import std.algorithm;
import std.format;
import std.range;

struct VeryLargeType
{
    static foreach (i; 500.iota)
    {
        mixin(format!"int v%s;"(i));
    }
}

void main()
{
    VeryLargeType a, b;
    move(a, b);
}
```

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -165,7 +165,7 @@
  */
 module std.traits;
 
-import std.meta : AliasSeq, allSatisfy;
+import std.meta : AliasSeq, allSatisfy, anySatisfy;
 import std.functional : unaryFun;
 
 // Legacy inheritance from std.typetuple
@@ -2613,7 +2613,7 @@ have a context pointer.
 */
 template hasNested(T)
 {
-    import std.meta : anySatisfy, Filter;
+    import std.meta : Filter;
 
     static if (isStaticArray!T && T.length)
         enum hasNested = hasNested!(typeof(T.init[0]));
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -2902,42 +2874,55 @@ template RepresentationTypeTuple(T)
...
+    struct VeryLargeType
     {
-        static if (T.length == 0)
+        import std.format : format;
+        import std.range : iota;
+
+        static foreach (i; 500.iota)
         {
-            enum Impl = false;
+            mixin(format!"int v%s;"(i));
```
wilzbach: How does this affect the compilation time of the unittests?

FeepingCreature: Compilation time of just `std.traits`, averaged over 10 runs:

branch | time | change
---------|------|-----
master | 2.475s +- 28ms | 0%
PR with unittest | 2.589s +- 50ms | 5% slower
PR without unittest | 2.411s +- 22ms | 7% faster

wilzbach: So it has a well-noticable effect.
Can you reproduce the error also with lower n?

FeepingCreature: No, because it only happens when we hit DMD's template recursion limit, which is 500: see src/dmd/dsymbolsem.d:5817.

FeepingCreature: ping

FeepingCreature:
> Line 3613 or 3614

Ahaha yep, what the hell. Fixed. (Looks like a copypaste error from ~2012.)

PR #6876, opened 2019-02-22
Also expose 'parent' when it's abstract

thaven:
When generating methods to implement e.g. an interface, it is also very useful to have an alias to the (abstract) function being overridden. E.g. to have access to UDAs.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -5089,7 +5089,10 @@ private static:
 
         static string generateTestMethod(C, alias fun)() @property
         {
-            return "return this.boilerplate[0 .. a0];";
+            return `
+                static assert(__traits(parent, parent).stringof == "I_10");
+                return this.boilerplate[0 .. a0];
+            `;
         }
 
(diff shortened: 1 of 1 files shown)


thaven:
I added a separate test now (and removed the static assert that I originally added in the existing test). The test is a stripped-down version of the actual code I need this change for.

It seems it actually works for non-overloaded functions, but fails for overloaded ones. I think I'll need to loop through the overload set to find the correct overload to alias.

thaven:
It turns out the problem with overloading is an existing bug and also applies to the `self` alias...

thaven:
Created Issue 19715 about the `self` and `parent` aliases.

Could this PR move on without that fix (by temporarily commenting out an overload in the test + adding a TODO referencing the issue)?

thewilsonator:
Yes, fixing senate issues separately is usually a good idea.

thaven:
Rebased this old PR on current master. Let's see how the tests turn out now. (previous run, long ago, just got stuck somewhere)

thewilsonator (approved):

PR #6877, opened 2019-02-24
adjust setops.cartesianProduct() for dip1000

WalterBright:


WalterBright:
Needed for https://github.com/dlang/dmd/pull/9220

PR #6880, opened 2019-02-27
Fix issue 19703 - std.typecons.ReplaceType wrongly evaluates alias this

Biotronic:


std/typecons.d
```diff
@@ -8653,6 +8653,15 @@ replacement.
 */
 template ReplaceType(From, To, T...)
 {
+    template isAliasThis(T, U)
+    {
+        static foreach (at; __traits(getAliasThis, T))
+            static if (is(typeof(__traits(getMember, T, at)) : U))
+                enum tmp = true;
+
+        enum isAliasThis =__traits(compiles, assert(tmp));
```
thewilsonator: Is this the best way to do this? I suppose it does doesn't really matter if it is made private.

Biotronic: Probably not. This would be better:
```
    template isAliasThis(T, U) {
        enum checkAlias(string name) = is(typeof(__traits(getMember, T, name)) : U);
        enum isAliasThis = anySatisfy!(checkAlias, __traits(getAliasThis, T));
    }
```

thewilsonator:
Hmm, 
```
/home/circleci/phobos/std/concurrency.d:2096: undefined reference to `_D3std11concurrency__T4ListTSQBbQBa7MessageZQw7opSliceMFNaNbNiNjZSQCmQCl__TQCbTQBzZQCj5Range'
```

Biotronic:
I agree that's weird, but on the other hand I've never seen circleci *not* fail in some unexplained way, so I'm loath to take it too seriously. I also have a hard time seeing how this PR would cause that error.

wilzbach:
Superseded by https://github.com/dlang/phobos/pull/7121

PR #6884, opened 2019-03-01
Use double and float versions of sin,cos from core.stdc.math instead …

Alexibu:
…af real version.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -751,10 +751,10 @@ deprecated
 real cos(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.cos(x); }
 //FIXME
 ///ditto
-double cos(double x) @safe pure nothrow @nogc { return cos(cast(real) x); }
+double cos(double x) @safe pure nothrow @nogc { return core.stdc.math.cos(x); }
 //FIXME
 ///ditto
-float cos(float x) @safe pure nothrow @nogc { return cos(cast(real) x); }
+float cos(float x) @safe pure nothrow @nogc { return core.stdc.math.cosf(x); }
 
 ///
 @safe unittest
@@ -793,10 +793,10 @@ float cos(float x) @safe pure nothrow @nogc { return cos(cast(real) x); }
 real sin(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.sin(x); }
 //FIXME
 ///ditto
-double sin(double x) @safe pure nothrow @nogc { return sin(cast(real) x); }
+double sin(double x) @safe pure nothrow @nogc { return core.stdc.math.sin(x); }
 //FIXME
 ///ditto
-float sin(float x) @safe pure nothrow @nogc { return sin(cast(real) x); }
+float sin(float x) @safe pure nothrow @nogc { return core.stdc.math.sinf(x); }
 
 ///
 @safe unittest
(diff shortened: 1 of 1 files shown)


thewilsonator:
I totally agree we need to stop using real unless its asked for. 
I _believe_ the idea was to use steal some of the polynomial coefficients from CEPHES.

ibuclaw:
C functions may violate pure.

thewilsonator:
Also 
```
../../phobos/std/math.d(754): Error: `cos` cannot be interpreted at compile time, because it has no available source code
```

Alexibu:
So I guess from the /* intrinsic */ after cos in core.math, that the compiler does something special with it, which magically makes cos pure.
I don't think people that evaluate cos at compile time care about performance. 
I am currently consuming a significant amount of electricity computing the long double version of cos when I have been asking for the float version knowing that it will save time.
I have my own local overloads now that fix the problem for me.

It looks like it could work to use ;
static if (__ctfe)
call the real one that is intrinsic
else
call the faster one.

I'll see if this passes tests, not sure about purity issues I don't use pure.

ibuclaw:
> So I guess from the /* intrinsic */ after cos in core.math, that the compiler does something special with it, which magically makes cos pure.

They are all x87 asm ops.

Alexibu:
OK, so I guess the reason no one has fixed it is because there is no difference in speed in the dmd reference compiler ?
In gdc, it results in calls to   _sinl and _cosl which are long double versions of sin and cos, and take a long time to execute.

thewilsonator:
Also if you are caring about performance you shouldn't be using DMD.
LDC [does this properly](https://github.com/ldc-developers/phobos/blob/ldc/std/math.d#L778) already.

ibuclaw:
> I believe the idea was to use steal some of the polynomial coefficients from CEPHES.

There is a pending refactor to split std math into asm and generic parts, similar to bigint.

There exists no support for IBM reals, which may be a good enough reason to introduce a pureCMath template, like pureMalloc, etc...

These can be discussed in bug reports.

thewilsonator:
I suspect that DConf will be a much more effective medium ;)

Alexibu:
> Also if you are caring about performance you shouldn't be using DMD.
> LDC [does this properly](https://github.com/ldc-developers/phobos/blob/ldc/std/math.d#L778) already.

I am using GDC, did you mean to say I shouldn't be using GDC ?

thewilsonator:
No, GDC has good codegen as well, just don't use DMD ;) 
I'm just saying this problem for LDC is already fixed.

Alexibu:
So the compilers have their own branches of phobos, that aren't being integrated with dlang/phobos ? That sounds a bit unsustainable.

thewilsonator:
Not really, most of it is the same, and the bits that are different don't change too often.

Alexibu:
OK, I can see the gdc phobos version doesn't have a version(GDC) with compiler specific cos functions.

Alexibu:
OK,  that didn't work. 
I think this is a job for experts, not me trying to fix it through a web interface.
Thanks for your help.

thewilsonator:
Hopefully we'll get this fixed at dconf.

ibuclaw:
> OK, I can see the gdc phobos version doesn't have a version(GDC) with compiler specific cos functions.

That's because it isn't strictly required, core.math functions are all recognised intrinsics.  One proposal would be to add float and double overloads to core.math, then std.math becomes just an alias.

PR #6885, opened 2019-03-01
Add trait getOverload

thaven:
Add a trait to get a specific overload of a method.

Apart from just being handy sometimes, this will help to fix the `self` and `parent` aliases in `std.typecons.AutoImplement` in case of overloaded methods. For more info about that: see #6876.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -20,6 +20,7 @@
  *           $(LREF hasFunctionAttributes)
  *           $(LREF functionLinkage)
  *           $(LREF FunctionTypeOf)
+ *           $(LREF getOverload)
  *           $(LREF isSafe)
  *           $(LREF isUnsafe)
  *           $(LREF isFinal)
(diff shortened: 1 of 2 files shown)


std/traits.d
```diff
@@ -2518,6 +2519,69 @@ private:
...
+{
+    import std.meta : Filter;
+
+    private enum bool isMatch(alias fun) = (functionAttributes!fun == attrs
+        && is(ParamTypes == Parameters!fun));
+
+    private alias matches = Filter!(isMatch, __traits(getOverloads,
+        __traits(parent, overloadSet), __traits(identifier, overloadSet)));
+
+    static assert(matches.length <= 1, "Multiple overloads seem to match...");
```
thewilsonator: please use a more appropriate message

std/traits.d
```diff
@@ -2518,6 +2519,69 @@ private:
...
+
+    static if(matches.length != 0)
+        alias getOverload = matches[0];
+    else
+    {
+        pragma(msg, "No matching overload found for " ~
+            __traits(identifier, overloadSet) ~ ParamTypes.stringof);
```
thewilsonator: Please remove the pragma msg (as an aside the preferred form of pragma message and mixin is the comma separated list form: `pragma(msg, "No matching", __traits(identifier, overloadSet), ParamTypes);`

Basile-z: I'd suggest either `alias getOverload = void;` or `static assert(0, "No matching....");`when there's no match.

std/traits.d
```diff
@@ -2518,6 +2519,69 @@ private:
...
+ *    ParamTypes = Types of the parameters of the overload to be aliased.
+ */
+template getOverload(alias overloadSet, FunctionAttribute attrs, ParamTypes...)
+{
+    import std.meta : Filter;
+
+    private enum bool isMatch(alias fun) = (functionAttributes!fun == attrs
+        && is(ParamTypes == Parameters!fun));
+
+    private alias matches = Filter!(isMatch, __traits(getOverloads,
```
Basile-z: the `getOverloads` compiler trait can take an optional `bool` parameter that indicates if the templates are included in the results. I wonder if, using this parameter, the template could also instantiate an overload, which would make it more versatile.

thaven: Interesting suggestion, but I don't think it is feasible. ˋParamTypesˋ just isn't enough information to instantiate the overload.

ZombineDev:
@thaven There's also https://github.com/dlang/phobos/pull/5818 by @Biotronic. I suggest you join forces with him.

thaven:
Anyone an idea about how to fix the documentation error reported in CircleCI?

`std/traits.d(2532:10)[warn]: Public declaration 'getOverload' has no documented example.`

Actually there is a unittest marked with `///`, but somehow it isn't recognized as an example?

wilzbach:
> Actually there is a unittest marked with ///, but somehow it isn't recognized as an example?

It looks like the problem is the `alias` declaration which is correctly recognized by dmd, but probably creates problems for the dscanner unittest checker.
See also:

https://github.com/dlang-community/D-Scanner/blob/0ec1e9557aba8e867ce932ecbcbadf4b095dc7cc/src/dscanner/analysis/has_public_example.d#L37

I would assume that the problem is that `aliasDeclaration` doesn't appear in `possibleDeclarations`.

thaven:
Okay, wrapped the alias in a template declaration now.

WalterBright:
I'd like to see a more compelling use case than "handy sometimes". Also, overload matching is more complex than just parameter types, there's rvalue vs lvalue, implicit conversions, etc.

thaven:
This is most useful in templates that take an alias to a function, and need to find the function being overridden. (In my specific case: given an alias to ˋC.fooˋ from  ˋclass C : I { int foo(); }ˋ find ˋI.fooˋ to inspect its UDAs). This can be done now, as long as the functions aren't overloaded. Otherwise you'll end up with just an alias to the overload set of ˋI.fooˋ, which isn't very helpful in this case.

About rvalue / lvalue, can't that be sorted out using ˋFunctionAttribute.refˋ?

In general, there certainly are alternative solutions to this specific problem. Most straight to the point would be another ˋ__traitsˋ keyword. E.g.. ˋ__traits(getOverriddenFunctions, BaseType, func)ˋ. All parameters being optional. If none specified, it would resolve to an AliasSeq of all functions overridden by the function where ˋ__traits(getOverriddenFunctions)ˋ appears. Adding the ˋBaseTypeˋ would limit it to the function from the specified ˋBaseTypeˋ. Adding ˋfuncˋ would allow to get functions overridden by func instead of the current function.

wilzbach:
We could always put it in the existing std.experimental.typecons for a bit to see whether it's really appreciated by the community?

thaven:
@atilaneves You are right that in case of just calling a function, it doesn't matter whether it is the overload set or a function. However, when doing reflection, e.g. accessing UDAs, it does matter.

To provide a concrete use case: [Issue 19715](https://issues.dlang.org/show_bug.cgi?id=19715), this bug makes it impossible to reflect on `self` or `parent` if it happens to be overloaded. Since parameter types are known, using `getOverload` for those aliases will solve that problem in most cases.

thaven:
Btw, I don't really understand the buildkite failure... is it even related to this PR?

thewilsonator:
(should be gone now)

RazvanN7:
ping @thaven

thaven:
Sorry for the delay. I have been quite occupied lately.

@atilaneves It's a long time ago this PR was opened, but I think the rationale was that such generic functionality should not be hidden somewhere deep inside the `AutoImplement` implementation.

Anyway, I may get back to working on `AutoImplement` itself within a few months, and I'll do some experiments with overloads then.

RazvanN7:
@thaven The fact that this functionality is required in the innards of `AutoImplement` is not sufficient to make it public.

RazvanN7:
@thaven given @atilaneves 's response I will close this. If you come up with a stronger case, feel free to reopen.

PR #6886, opened 2019-03-02
Backport: Fix unaligned access in std.digest.murmurhash

jpf91:
Backport for GDC/`dmd-cxx` from latest stable to fix  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89177

### DIFF
--- a/std/digest/murmurhash.d
+++ b/std/digest/murmurhash.d
@@ -9,7 +9,7 @@ The older MurmurHash 1 and 2 are currently not supported.
 
 MurmurHash3 comes in three flavors, listed in increasing order of throughput:
 $(UL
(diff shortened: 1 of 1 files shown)


wilzbach:
It would be a lot better to use [`merge_stable`](https://github.com/wilzbach/git-tools/blob/master/merge_stable) or similar tools as this will avoid merge conflicts for the release manager.

jpf91:
Is the `dmd-cxx` branch ever merged back into stable/master? I though this is a GDC-only branch?

wilzbach:
Ah sorry. I read "from latest stable" to fast. Nevermind then.

jpf91:
Maybe I should have added a better description ;-)

ibuclaw:
I usually prefix [dmd-cxx] just in case there's any room for doubt

PR #6887, opened 2019-03-03
[GDC] Cleanup MinGW code

jpf91:
Remove old code: We now simply set `CRuntime_Microsoft`, as MinGW is really MS runtime+ some extra libraries. Because of that, we no longer need MinGW specific code.

Also use generic math implementations as DMD-style IASM isn't available. Part of MinGW support for GDC/GCC9, https://github.com/D-Programming-GDC/gcc/pull/1

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -162,6 +162,12 @@ else version (D_InlineAsm_X86_64)
     version = InlineAsm_X86_Any;
 }
 
+version (CRuntime_Microsoft)
+{
+    version (InlineAsm_X86_Any)
+        version = MSVC_InlineAsm;
+}
+
 version (X86_64) version = StaticallyHaveSSE;
 version (X86) version (OSX) version = StaticallyHaveSSE;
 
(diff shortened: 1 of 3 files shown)


std/math.d
```diff
@@ -4707,7 +4711,24 @@ real hypot(real x, real y) @safe pure nothrow @nogc
...
+            y += 1.0;
+
+        return y;
+    }
+
+    version(GNU)
+    {
+        return genericCeil(x);
+    }
+    else version (Win64_DMD_InlineAsm)
```
ibuclaw: Wouldn't this be a problem with the version definition? Based on this change alone, I don't see a reason why this would be necessary.

ibuclaw:
```
version (CRuntime_Microsoft)
{
    version (D_InlineAsm_X86)
        version = MSVC_InlineAsm;
    else version (...)
        version = MSVC_InlineAsm;
}
```
Or whatever suitable name in place of Cruntime_Microsoft.

jpf91: That's indeed much better.

ibuclaw:
Looks reasonable.

jpf91:
This needs some more testing.

PR #6890, opened 2019-03-04
Disable the style_lint target on CircleCi

wilzbach:
It has been running there for a while.

See also: https://github.com/dlang/ci/pull/259

(CircleCi is intended to be fully retired soon, see e.g. https://github.com/dlang/phobos/pull/6663)

PR #6882, opened 2019-02-28
Update D-scanner version

Basile-z:

PR #6891, opened 2019-03-04
std.range.package.d dip1000

WalterBright:


Basile-z:
Looks good but Same CircleCI failure as in #6882. @wilzbach still no solution ?

Basile-z:
Sorry, ignore my previous comment @wilzbach ,  i've just seen #6890, i suppose this will fix the problems.

Basile-z:
@WalterBright or any with write access , you can rebase this PR on master, this will work this time.

PR #6892, opened 2019-03-05
get std.path to compile with -dip1000

WalterBright:


std/range/package.d
```diff
@@ -1446,9 +1446,10 @@ private struct ChooseResult(R1, R2)
     }
     private bool r1Chosen;
 
-    private static auto ref actOnChosen(alias foo, ExtraArgs ...)(ref ChooseResult r,
-            auto ref ExtraArgs extraArgs)
+    private static auto ref actOnChosen(alias foo, ExtraArgs ...)(return scope ref ChooseResult r,
+            auto ref ExtraArgs extraArgs) @trusted
```
thewilsonator: This will trust `foo`, which may in turn trust the postblit and/or the destructor, which may contain arbitrary `@system` code.

WalterBright:
Due to a lot of entanglement with other Phobos modules, a lot of changes were needed.

WalterBright:
Some of these changes might be useful in the future.

PR #6893, opened 2019-03-05
Remove redundant body of @disable functions. Issue 14854 DMD

tibi77:
As the issue 14854 suggested the body of a disable function is redundant.

### DIFF
--- a/std/container/binaryheap.d
+++ b/std/container/binaryheap.d
@@ -516,11 +516,11 @@ BinaryHeap!(Store, less) heapify(alias less = "a < b", Store)(Store s,
     static struct StructWithoutDup
     {
         int[] a;
-        @disable StructWithoutDup dup()
-        {
-            StructWithoutDup d;
-            return d;
-        }
+        @disable StructWithoutDup dup();
+        //{
+        //    StructWithoutDup d;
+        //    return d;
+        //}
         alias a this;
     }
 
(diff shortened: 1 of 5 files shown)


std/container/binaryheap.d
```diff
@@ -516,11 +516,11 @@ BinaryHeap!(Store, less) heapify(alias less = "a < b", Store)(Store s,
...
-        @disable StructWithoutDup dup()
-        {
-            StructWithoutDup d;
-            return d;
-        }
+        @disable StructWithoutDup dup();
+        //{
+        //    StructWithoutDup d;
+        //    return d;
+        //}
```
quickfur: No need to comment out a useless function body, just delete it.

std/parallelism.d
```diff
@@ -558,10 +558,10 @@ struct Task(alias fun, Args...)
...
     else
     {
-        @disable typeof(this) opAssign(typeof(this) rhs)
-        {
-            assert(0);
-        }
+        @disable typeof(this) opAssign(typeof(this) rhs);
+        //{
+        //    assert(0);
+        //}
```
quickfur: ditto

std/container/binaryheap.d
tibi77: Done. Thank you for the input!

PR #6895, opened 2019-03-07
Fix purity problem in std/path

RazvanN7:
This is currently blocking https://github.com/dlang/dmd/pull/9428

### DIFF
--- a/std/path.d
+++ b/std/path.d
@@ -2738,17 +2738,23 @@ else version (Posix)
     See_Also:
         $(LREF asAbsolutePath) which does not allocate
 */
-string absolutePath(string path, lazy string base = getcwd())
+string absolutePath(string path, string base)
     @safe pure
 {
     import std.array : array;
-    if (path.empty)  return null;
-    if (isAbsolute(path))  return path;
     auto baseVar = base;
     if (!isAbsolute(baseVar)) throw new Exception("Base directory must be absolute");
     return chainPath(baseVar, path).array;
 }
 
+/// ditto
+string absolutePath(string path)
+    @safe
+{
+    if (path.empty) return null;
+    if (isAbsolute(path)) return path;
+    return absolutePath(path, getcwd());
+}
 ///
 @safe unittest
 {


CyberShadow:
Maybe it's better to make it so that `@pure` moves the onus of purity enforcement for default arguments to the caller? That would break less code and seems to make more sense with how default arguments should work.

std/path.d
```diff
@@ -2738,17 +2738,23 @@ else version (Posix)
...
         $(LREF asAbsolutePath) which does not allocate
 */
-string absolutePath(string path, lazy string base = getcwd())
+string absolutePath(string path, string base)
     @safe pure
 {
     import std.array : array;
-    if (path.empty)  return null;
-    if (isAbsolute(path))  return path;
     auto baseVar = base;
```
CyberShadow: `baseVar` is no longer necessary.

std/path.d
```diff
@@ -2738,17 +2738,23 @@ else version (Posix)
     See_Also:
         $(LREF asAbsolutePath) which does not allocate
 */
-string absolutePath(string path, lazy string base = getcwd())
+string absolutePath(string path, string base)
     @safe pure
 {
     import std.array : array;
-    if (path.empty)  return null;
-    if (isAbsolute(path))  return path;
```
CyberShadow: I don't think you can remove these here. It changes the behavior of this overload.

CyberShadow:
@thewilsonator How did you find this patch OK for merging?

RazvanN7:
>Maybe it's better to make it so that @pure moves the onus of purity enforcement for default arguments to the caller? That would break less code and seems to make more sense with how default arguments should work.

Why would you do that? This way every time a function is called with the default argument it has to by typechecked  for purity, even though it will always call an impure function

CyberShadow:
> This way every time a function is called with the default argument it has to be typechecked for purity,

What is the problem with that? Performance?

The way I'm thinking about this is that

```d
void fun(T arg = calcDefault()) {}
fun();
```

should be equivalent to

```d
void fun(T arg) {}
fun(calcDefault());
```

except for scope / visibility.

Doing it any other way makes the default argument expressions be treated in a different way than non-default argument expressions, and that doesn't seem right to me.

RazvanN7:
Hmm... you are right. The problem here is with visibility: you may have functions that look like they are `pure`, but suddenly if you use the default argument value they become `impure`.  I will investigate how to implement what you suggested.

RazvanN7:
And then there's lazy parameters:

```d
int b = 7;
string m() { b = 2; return "asd"; }
pure void g(lazy string i = m() ){ string l = i; }
pure void f()
{
    g(); 
}

void main() pure
{
    import std.stdio : writeln;
    writeln(b);                           // prints 7
}  
```

RazvanN7:
@CyberShadow : The spec says that default parameters are evaluated in the function declaration context [1]. Although it makes more sense to evaluate them on caller site, the compiler has to semantically analyze them in the function declaration to make sure that the type of the default argument matches the type of the parameter declaration. 

[1] https://dlang.org/spec/function.html#function-default-args

CyberShadow:
> The spec says that default parameters are evaluated in the function declaration context [1].

Looks pretty vague. I doubt that the matter we're discussing here was taken into account when that text was written. It probably refers to visibility.

> Although it makes more sense to evaluate them on caller site, the compiler has to semantically analyze them in the function declaration to make sure that the type of the default argument matches the type of the parameter declaration.

Well, if you move checks of other stuff beside visibility to the callee context, it has unpleasant consequences such as, in theory, not being able to call a function with its default argument explicitly specified, i.e. given `R fun(T arg = expr()){}`, `fun();` might work even when `fun(expr())` would not (for reasons other than visibility), which doesn't make sense.

A more drastic approach would be to move typechecking of default argument expressions to caller time. I think that would be fine as we already do this for overloads.

In any case, I'm not the expert on such things, maybe ask Walter / Andrei. It does seem like a messy matter worthy of more in-depth consideration.

PR #6896, opened 2019-03-08
Fix Issue 18728 - std.math.fdim does not handle nan correctly

tibi77:
Added the correct asserts as well.[1]

[1] https://issues.dlang.org/show_bug.cgi?id=18728

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -7359,16 +7359,20 @@ T nextafter(T)(const T x, const T y) @safe pure nothrow @nogc
  *      $(TR $(TD x $(LT)= y) $(TD +0.0))
  *      )
  */
-real fdim(real x, real y) @safe pure nothrow @nogc { return (x > y) ? x - y : +0.0; }
+real fdim(real x, real y) @safe pure nothrow @nogc
+{
+    return (isNaN(x) || isNaN(y)) ? real.nan : ((x > y) ? x - y : +0.0);
+}
 
 ///
 @safe pure nothrow @nogc unittest
 {
     assert(fdim(2.0, 0.0) == 2.0);
     assert(fdim(-2.0, 0.0) == 0.0);
-    assert(fdim(real.infinity, 2.0) == real.infinity);
-    assert(fdim(real.nan, 2.0) == 0.0);
-    assert(fdim(2.0, real.nan) == 0.0);
+    assert(isInfinity(fdim(real.infinity, 2.0)));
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -7359,16 +7359,22 @@ T nextafter(T)(const T x, const T y) @safe pure nothrow @nogc
  *      $(TR $(TD x $(LT)= y) $(TD +0.0))
  *      )
  */
-real fdim(real x, real y) @safe pure nothrow @nogc { return (x > y) ? x - y : +0.0; }
+real fdim(real x, real y) @safe pure nothrow @nogc
+{
+    if (isNaN(x) || isNaN(y))
+        return real.nan;
+    return (x > y) ? x - y : +0.0;
+}
```
thewilsonator: You should be able to make this branchless by converting this to `return (x < y) ? +0.0 : x - y;`

see https://run.dlang.io/is/G0Ka0Z (LDC) DMD does about the same.

tibi77: But return (x < y) ? +0.0 : x - y; is not checking for NaN. Should I do it like return (isNaN(x) || isNan(y)) ? real.nan : ((x < y) ? +0.0 : x - y); ?

thewilsonator: Ah, but it is! 

https://run.dlang.io/is/o3siXK

The reasons is that `NaN ? a : b` is always `b`. by inverting the condition and swapping `a` and `b`, you get for any NaN `x` or `y`,`NaN ? +0.0 : x-y ` ->  `x-y`, and since one of them is NaN so is `x-y`.

aG0aep6G: > `NaN ? a : b` is always `b`

Just to avoid confusion, you mean `x < y` is false when `x` and/or `y` are NaN.

NaN's boolean value is actually `true`, so `NaN ? a : b` is actually `a`.

thewilsonator: >NaN's boolean value is actually `true`

Thanks I forgot about that, how could I forget Sociomantic's $9T bug?

tibi77: Didn't knew that! Very interesting. Thank you. I will change the commit asap.

thewilsonator: > NaN's boolean value is actually true, so NaN ? a : b is actually a.

My bad, this is actually `NaN  > 0.0 ? a : b` and `NaN > 0.0` is always false. So I was right the first time :)

tibi77: Ping @ibuclaw

PR #6898, opened 2019-03-10
String Interpolation 

jash11:
Based on #6339

### DIFF
--- a/std/experimental/interpolate.d
+++ b/std/experimental/interpolate.d
@@ -142,8 +142,6 @@ unittest
     
     static assert(mixin(interp("$()")).length == 0);
     static assert(mixin(interp("$(/* a comment!*/)")).length == 0);
-    // DOESN'T WORK
-    //static assert(mixin(interp("$(// another comment)")).length == 0);
     static assert(mixin(interp("$(/+ yet another comment+/)")).length == 0);
 
     // DOESN'T WORK (see "HANDLE EMPTY EXPRESSIONS")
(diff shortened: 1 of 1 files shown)


std/experimental/interpolate.d
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/**
+Interpolate an expression. `str` starts with an expression to interpolate.
+*/
+private string interpExpression(string str, string file, size_t line) pure @safe
+{
+    if (str[0] == '(')
+    {
+        uint depth = 1;
```
jash11: Would it make more sense to use `size_t` or `ptrdiff_t` instead of `uint` here?

jash11: nvm, I was confusing depth with i

jash11:
Okay, this is ready for review now

thewilsonator:
This needs a changelog for this to go in.

wilzbach:
Also, please mind the style failures on Buildkite.

jash11:
> I'm still highly in favor of a language solution as this is just too much too type for such a basic feature (closing parentheses and opening them is less characters!), the compiler implementation will be faster and I would have a hard time explaining to beginners why we have to use a `mixin` for this rudimentary feature.

@wilzbach I 100% agree with you, and I am also in favor of string interp as a language feature (Although I was getting the impression that I was the only one, because everyone was so strongly against it). Maybe that's not true? Is there still some hope in a language solution to interp 😄.

wilzbach:
Hehe, the DMD PR has the highest voting count ever reached for a DMD PR and a lot of people are in favor of this (except for Andrei, of course).

jash11:
I will close this then, because it is better to fight for that cause than for this.

PR #6899, opened 2019-03-11
Fix for the issue 18806: InputRange for std.algorithm.minIndex

Vasniktel:
Fixes https://issues.dlang.org/show_bug.cgi?id=18806

I'll add a changelog when this is ready to be merged.

### DIFF
--- a/std/algorithm/searching.d
+++ b/std/algorithm/searching.d
@@ -3789,7 +3789,7 @@ irreflexive (`pred(a, a)` is `false`).
 Params:
     pred = The ordering predicate to use to determine the extremum (minimum or
         maximum) element.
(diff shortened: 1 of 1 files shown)


thewilsonator:
Please change the commit message to include "fix issue 18806 - (rest of description)" so that the dlang bot picks up on it. I'm not sure what caused the linux failure, looks random?

thewilsonator:
A changelog will be automatically added because this is a closed issue. Changlelogs are only necessary when users need to know more, such as how to work around issues with the fix.

wilzbach:
What about maxIndex?
Also, `std.internal.test.dummyrange` helps a lot with avoiding the annoying work of defining test ranges.

PR #6901, opened 2019-03-13
byUTF: don't read from range in constructor

WalterBright:
Constructors should not start reading in / consuming the data. Reading should only be done when asked for.

### DIFF
--- a/std/utf.d
+++ b/std/utf.d
@@ -4208,46 +4208,46 @@ if (isSomeChar!C)
         {
             static struct Result
             {
+                enum Empty = uint.max;  // range is empty or just constructed
+
                 this(R val)
                 {
                     r = val;
-                    popFront();
                 }
 
                 @property bool empty()
                 {
-                    return buff == uint.max;
+                    return buff == Empty && r.empty;
                 }
 
-                @property auto front()
+                @property dchar front() scope // 'scope' required by call to decodeFront() below
                 {
-                    assert(!empty, "Attempting to access the front of an empty byUTF");
-                    return cast(dchar) buff;
-                }
+                    if (buff == Empty)
+                    {
+                        auto c = r.front;
 
-                void popFront() scope
-                {
-                    assert(!empty, "Attempting to popFront an empty byUTF");
-                    if (r.empty)
-                    {
-                        buff = uint.max;
-                    }
-                    else
-                    {
                         static if (is(RC == wchar))
                             enum firstMulti = 0xD800; // First high surrogate.
                         else
                             enum firstMulti = 0x80; // First non-ASCII.
-                        if (r.front < firstMulti)
+                        if (c < firstMulti)
                         {
-                            buff = r.front;
                             r.popFront;
+                            buff = cast(dchar) c;
                         }
                         else
                         {
                             buff = () @trusted { return decodeFront!(Yes.useReplacementDchar)(r); }();
                         }
                     }
+                    return cast(dchar) buff;
+                }
+
+                void popFront()
+                {
+                    if (buff == Empty)
+                        front();
+                    buff = Empty;
                 }
 
                 static if (isForwardRange!R)
@@ -4260,8 +4260,10 @@ if (isSomeChar!C)
                     }
                 }
 
-                uint buff;
+            private:
+
                 R r;
+                uint buff = Empty;      // one character lookahead buffer
             }
 
             return Result(r);


WalterBright:
For reference: https://github.com/dlang/phobos/pull/6900

thewilsonator:
```
std/string.d(4286): Error: template `std.string.rightJustifier` cannot deduce function from argument types `!()(Result, ulong, dchar)`, candidates are:
std/string.d(4277):        `std.string.rightJustifier(Range)(Range r, size_t width, dchar fillChar = ' ') if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) && !isConvertibleToString!Range)`
std/string.d(4387):        `std.string.rightJustifier(Range)(auto ref Range r, size_t width, dchar fillChar = ' ') if (isConvertibleToString!Range)`
std/string.d(4248): Error: template instance `std.string.rightJustifier!string` error instantiating
std/xml.d(943):        instantiated from here: `rightJustify!string`
std/string.d(4286): Error: template `std.string.rightJustifier` cannot deduce function from argument types `!()(Result, ulong, dchar)`, candidates are:
std/string.d(4277):        `std.string.rightJustifier(Range)(Range r, size_t width, dchar fillChar = ' ') if (isForwardRange!Range && isSomeChar!(ElementEncodingType!Range) && !isConvertibleToString!Range)`
std/string.d(4387):        `std.string.rightJustifier(Range)(auto ref Range r, size_t width, dchar fillChar = ' ') if (isConvertibleToString!Range)`
std/string.d(4248): Error: template instance `std.string.rightJustifier!string` error instantiating
std/xml.d(943):        instantiated from here: `rightJustify!string`
posix.mak:330: recipe for target 'generated/linux/release/64/libphobos2.so.0.85.0' failed
```

std/utf.d
```diff
@@ -4208,46 +4208,46 @@ if (isSomeChar!C)
         {
             static struct Result
             {
+                enum Empty = uint.max;  // range is empty or just constructed
```
andralex: this should be lowercase

andralex:
Not a fan of this.

Reading from the input is priming the range object. Admittedly that's awkward but now we have several awkward points in the code:

* `popFront` calling `front` is arguably no less clownly than the constructor calling `popFront`.
* `buff == Empty` is evaluated in `empty`, `front`, and `popFront`, i.e. in a typical loop it gets evaluated three times per iteration!

Granted, optimizations and branch prediction will be of great help with that, but without some measurements I am worried - this is performance sensitive code and everybody is complaining (with or without basis) that decoding is slow.

quickfur:
I agree with @andralex, this new implementation is very messy, and does a lot of extra work all for what? Just to avoid reading from the source range once in the ctor?  That sounds like the tail wagging the dog. Why is this so important anyway?

If not reading the wrapped range in the ctor is an absolute must, other, cleaner solutions present themselves, like lazily initializing the range, etc..

WalterBright:
'Priming' the input in the constructor leads to conceptual problems with things like "what does a copy constructor do". I ran into this trying to implement `save()` with a copy constructor. Other problems with are the popFront() throwing exceptions, which significantly complicates all code that calls the constructor. There's nothing nice or fast about adding RAII try-finally blocks in your constructors (the compiler will add them). popFront() can't be pure, either.

Calling front in popFront does not have these issues.

BTW, the old code would call front() and empty() multiple times per loop.

Also, popFront could inline the code from front.

jmdavis:
> 'Priming' the input in the constructor leads to conceptual problems with things like "what does a copy constructor do".

I point to my comment in: https://github.com/dlang/phobos/pull/6900#issuecomment-472508594

Why on earth are copy constructors involved with this at all? Copying a range is undefined semantics. What it does varies wildly depending on the type, meaning that once you've copied a range, you can't iterate the original anymore without risking bugs. Copy constructors are a struct-specific feature and don't apply to classes. `save` is how you create a new "copy" of a range, and it is the _only_ way to do so in generic code. Relying on anything else to get an independent copy is a bug.

quickfur:
@WalterBright Why are you trying to implement `save` with a copy ctor??  The range API specifies that the only reliable way to make a copy of a range is using the `save` method.  It is currently not possible to have correct semantics any other way, because the bulk of range-based code has been written with this assumption -- even if in theory we should have used copy ctors to implement `save` in the first place. Assuming otherwise breaks a lot of current range-based code.

For instance, consider a range implemented as a class. AIUI classes do not have copy ctors, therefore, if your `save` method uses the copy ctor to copy the wrapped range, it *will not work correctly*. And just in case it's not 100% clear, `std.range.interfaces` provides *standard* ways of constructing ranges that are wrapped in classes and interfaces, and these wrappers are necessary when the specific type of the range is unknown at compile-time (i.e., if you have a function that returns one of two different ranges (with compatible element types) depending on a runtime condition), so it's not as if this is a rare corner case; *it happens in real-life code*.

As things stand, implementing `save` with copy ctors simply does not make any sense.  It will not make any sense unless we also do a major overhaul of the range API to get rid of `save`, which will be a very drastic change that will basically break all existing code.

quickfur:
tl;dr: this PR uglifies the code and adds performance hits for no good reason at all, other than an unfounded assumption that `save` should be implemented using copy ctors. Why was this force-merged without any discussion with Phobos maintainers?

andralex:
cc @thewilsonator

thewilsonator:
@WalterBright please revert

jmdavis:
> @WalterBright please revert

https://github.com/dlang/phobos/pull/6900 is related.

PR #6904, opened 2019-03-14
ChooseResult: use of unions need to be @trusted

WalterBright:


std/range/package.d
```diff
@@ -1475,6 +1481,15 @@ private struct ChooseResult(R1, R2)
         }
     }
 
+    void opAssign(return scope ChooseResult r) @trusted
+    {
+        r1Chosen = r.r1Chosen;
+        if (r1Chosen)
+            r1 = r.r1;  // assigning to union members makes it @trusted
```
thewilsonator: Your implication in the comment is wrong. assigning to union members makes it _`@system`_, hence it need to be `@trusted`

std/range/package.d
```diff
@@ -1475,6 +1481,15 @@ private struct ChooseResult(R1, R2)
         }
     }
 
+    void opAssign(return scope ChooseResult r) @trusted
```
dnadlinger: @WalterBright: Ouch.

```d
import std.range;

struct EvilRange {
    enum empty = true;
    int front;
    void popFront() @safe {}
    auto opAssign(const ref EvilRange other){
        *(cast(uint*)0xcafebabe) = 0xdeadbeef;
        return this;
    }
}

void main() @safe {

    auto c1 = choose(true, EvilRange(), EvilRange());
    auto c2 = c1;
    c1 = c2;
}
```

(Filed as https://issues.dlang.org/show_bug.cgi?id=19738.)

dnadlinger: Still unfixed – should we revert this commit? (I don't have time to look into this right now, sorry.)

WalterBright: No, because of the other fixes in it. Just file a new PR.

PR #6905, opened 2019-03-16
improve documentation for .save() property

WalterBright:


std/range/primitives.d
```diff
@@ -966,6 +966,16 @@ are the same as for an input range, with the additional requirement
...
+`save` behaves in many ways like a copy constructor, and its
+implementation typically is done using copy construction.
+
+The existence of a copy constructor, however, does not imply
+the range is a forward range. For example, a range that reads
+from a TTY consumes its input and cannot save its place and
+read it again, and so cannot be a forward range and cannot
+have a `save` function.
+
+
```
thewilsonator: extra whitespace

thewilsonator:
This should be discussed first, cc @jmdavis

jmdavis:
Why would we associate `save` with copy constructors? Sure, it's similar in the sense that it results in an independent copy of the range that can be iterated separately (though how that affects the elements themselves depends on how the range is implemented). But you don't copy ranges in the normal sense. Classes don't have copy constructors, and ranges that are structs usually doesn't implement `save` with any kind of copy construction or postblit. You _can_ implement it that way, but you shouldn't ever rely on how copying ranges works in generic code, because the semantics of copying ranges varies widely based on how they're implemented. If you want to "copy" a range, you use `save`. You don't use copy construction or postblit constructors or anything like that except as an implementation detail of `save`. Honestly, I think that trying to associate `save` with copy constructors is just begging for trouble. They're separate things.

Personally, I really wish that we had never had `save` and that we had required that postblit / copy constructors be used instead (therefore requiring that a class range be wrapped in a struct with a copy constructor / postblit), but unfortunately, that's not what we did, and I don't know how feasible it is to change it now given the code breakage that that would incur. That being the case, I'm strongly against having the documentation talk like `save` is associated with copy constructors or that you even necessarily implement it via a copy constructor. That's an implementation detail and is not applicable to all ranges.

WalterBright:
The issue came up from:

1. I was trying to figure out what the difference between `save` and a copy constructor was.

2. most implementations of `save` in Phobos use assignment semantics, when they should be using initialization semantics. I.e. most of them are wrong and work correctly by accident.

WalterBright:
See also https://github.com/dlang/phobos/pull/6900

thewilsonator:
I realised @RazvanN7 was never cc'd to this.

RazvanN7 (approved):

RazvanN7:
I think that the addition enables the understanding of the `.save` property by making an association with the copy constructor. I don't see the harm in that.

PR #6909, opened 2019-03-18
FormatSpec scope parameters

WalterBright:
FormatSpec should never escape the functions it is used in, so ensure it.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -470,7 +470,7 @@ if (isSomeString!(typeof(fmt)))
 }
 
 /// ditto
-uint formattedWrite(Writer, Char, A...)(auto ref Writer w, in Char[] fmt, A args)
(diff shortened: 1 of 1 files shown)


WalterBright:
What is the dlang/druntime test in buildkite/phobos? The failures are link failures to std.format functions annotated with `scope`. Perhaps whatever it's being linked to needs to be recompiled? @wilzbach ?

thewilsonator:
IIRC it uses old dmd/druntime and is a PITA when dealing with cross git module problems. This, I'm not so sure about.

thewilsonator:
Oooh, that the benchmark running stuff which uses the wrong phobos. Just merge it.

thewilsonator:
`std/format.d(4016:121)[warn]: Line is longer than 120 characters`

thewilsonator:
Once you fix that just force merge it.

PR #6910, opened 2019-03-18
more scope for FormatSpec

WalterBright:


thewilsonator:
`std/bigint.d(928): Error: scope variable `f` may not be copied into allocated memory`

PR #6914, opened 2019-03-18
std.file: -dip1000

WalterBright:


WalterBright:
@thewilsonator is the ci/circleci failure (undefined reference to a symbol) also a symptom of linking to a previous Phobos, like the last buildkite/phobos one was?

thewilsonator:
CircleCI seems to fail due to stale libs.

WalterBright:
So pull anyway?

thewilsonator:
Ye, we can always revert if need be. (CircleCI is required, so I can't do it).

wilzbach:
> CircleCI seems to fail due to stale libs.

Nope no caches there. It's just the typical annoyance of -dip1000 changing the mangling and now some modules are compiled with -dip1000, some aren't and thus it fails when trying to do single module unittest builds. This could possibly vanish if all these dip1000 PRs are in, but then again this entire exercise is pretty pointless as long as we still have the mangling problem which is what the CI was saying very loudly until it was for force-merged.

> Ye, we can always revert if need be. (CircleCI is required, so I can't do it).

__not cool, not cool at all.__

Master is still failing because of this: https://circleci.com/gh/dlang/phobos/9535

PR #6916, opened 2019-03-18
std.concurrency: -dip1000

WalterBright:


ZombineDev:
Toggling auto-merge as the phobos failure on buildkite is transient.

ZombineDev:
Auto-merge toggled on

PR #6919, opened 2019-03-18
std.socket: -dip1000

WalterBright:


thewilsonator:
Fails on windows `std\socket.d(1756): Error: scope variable `addr` may not be returned`.

WalterBright:
> Error: scope variable addr may not be returned

Looks like dip1000 found a latent corruption bug in `std.socket`

std/socket.d
```diff
@@ -1749,16 +1749,19 @@ public:
     /// Any IPv6 host address.
     static @property ref const(ubyte)[16] ADDR_ANY() pure nothrow @nogc
     {
-        const(ubyte)[16]* addr;
         static if (is(typeof(IN6ADDR_ANY)))
         {
-            addr = &IN6ADDR_ANY.s6_addr;
```
CyberShadow: Wow, it was taking the address of an `enum` `struct` field. That shouldn't be allowed DIP1000 or not :)

PR #6921, opened 2019-03-19
std.algorithm.sorting: compile with -dip1000

WalterBright:


wilzbach:
Closing in favor of #6927

PR #6922, opened 2019-03-19
std.container.rbtree: compile with -dip1000

WalterBright:


thewilsonator:
Circle fails to link as before, please merge when others checks are green.

wilzbach:
> Circle fails to link as before, please merge when others checks are green.

Yes that's because master was broken by force-merging and now all PRs fail.

WalterBright:
@wilzbach what do you suggest? I don't know how to distinguish a real circleci failure from one where it failed because it didn't link with a rebuilt library.

wilzbach:
> @wilzbach what do you suggest?

-preview=dip1000 __MUST NOT__ result in different mangling. If there are unavoidable mangling changes in DMD they need to be done also for non dip1000 code.
The error you're seeing here is just the tip of the iceberg that people would be going through if they every gave -dip1000 a shot.

WalterBright:
@wilzbach ok I understand. Let's first get everything compiling with -dip1000, then we'll look at the name mangling issue.

PR #6924, opened 2019-03-19
std.container.util: compile with -dip1000

WalterBright:


WalterBright:
The failures here seem to be ones corrected by https://github.com/dlang/phobos/pull/6922

PR #6929, opened 2019-03-22
A little bit of safety

drug007:
Can't manage to provide reduced case for compiler error (dmd-fe 2.085) sorta `Error: @safe function [snipped].RedBlackTree.toHash cannot call @system function std.container.rbtree.RBRange[snipped].popFront`. This PR fixes the issue.

### DIFF
--- a/std/container/rbtree.d
+++ b/std/container/rbtree.d
@@ -690,7 +690,7 @@ private struct RBRange(N)
      *
      * Complexity: amortized $(BIGOH 1)
      */
-    void popFront()
+    void popFront() @safe
     {
         _begin = _begin.next;
     }


std/container/rbtree.d
```diff
@@ -690,7 +690,7 @@ private struct RBRange(N)
      *
      * Complexity: amortized $(BIGOH 1)
      */
-    void popFront()
+    void popFront() @safe
```
aG0aep6G: `@safe` should be inferred, because this is in a template. If it's not inferred, that's likely a compiler bug.

drug007: That's what I was afraid of... Getting reduced case turns to be non-trivial.

PR #5513, opened 2017-06-26
[PoC] Add version(NoAutoDecode) to allow opt-out of auto-decoding

wilzbach:
This is a Proof of Concept (PoC) idea to allow people to opt-in to use Phobos without auto-decoding.
If this works well, it could be improved (e.g. per module) to allow gradual migration.

It's an adaption of an KISS idea posted on [the NG](http://forum.dlang.org/post/xzizleuozzqnjinbfgoc@forum.dlang.org) by @adamdruppe.

But I guess a code sample shows more than 1000 words:

```d
void main(string[] args)
{
    import std.range;
    string cStr = "abc";
    wstring wStr = "abc";
    pragma(msg, typeof(cStr.front));
    pragma(msg, typeof(wStr.front));
}
```

```
> ~/dlang/dmd/generated/linux/release/64/dmd  -run test2.d
dchar
dchar
```

```
> ~/dlang/dmd/generated/linux/release/64/dmd  -version=NoAutoDecode -run test.d
immutable(char)
immutable(wchar)
```

This obviously needs more work and maybe a real world test.
But before investing more time into this, I am interested: what do you think about this approach? Worth looking into?

PR #6930, opened 2019-03-23
add std.range.primitives.Autodecoding

WalterBright:
Regardless of how we decide how to deal with autodecoding, Phobos' dependencies on autodecoding need to be removed. By manually setting `Autodecoding` to `false`, Phobos can be gone through module by module to remove those dependencies.


WalterBright:
See also previous work https://github.com/dlang/phobos/pull/5513

PR #6932, opened 2019-03-23
std.range.primitive.{Autodecoding -> autodecodeStrings}

dnadlinger:
This follows Phobos naming conventions and is more self-documenting.

The field has just been added by #6930.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -6585,9 +6585,10 @@ package template convertToString(T)
  */
 template isAutodecodableString(T)
 {
-    import std.range.primitives : Autodecoding;
+    import std.range.primitives : autodecodeStrings;
 
-    enum isAutodecodableString = Autodecoding && (is(T : const char[]) || is(T : const wchar[])) && !isStaticArray!T;
+    enum isAutodecodableString = autodecodeStrings &&
+        (is(T : const char[]) || is(T : const wchar[])) && !isStaticArray!T;
 }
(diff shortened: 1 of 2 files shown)


std/traits.d
```diff
@@ -6585,9 +6585,9 @@ package template convertToString(T)
  */
 template isAutodecodableString(T)
 {
-    import std.range.primitives : Autodecoding;
+    import std.range.primitives : autodecodeStrings;
 
-    enum isAutodecodableString = Autodecoding && (is(T : const char[]) || is(T : const wchar[])) && !isStaticArray!T;
+    enum isAutodecodableString = autodecodeStrings && (is(T : const char[]) || is(T : const wchar[])) && !isStaticArray!T;
```
wilzbach: std/traits.d(6590:121)[warn]: Line is longer than 120 characters

WalterBright:
It's a manifest constant, we usually capitalize the first letter.

dnadlinger:
> It's a manifest constant, we usually capitalize the first letter.

Not true.

WalterBright:
It seems dmd does it both ways :-)

Anyhow, it doesn't matter to me.

dnadlinger:
> It seems dmd does it both ways :-)

DMD's coding conventions markedly differ from user-facing code. I don't particularly care either, but let's be consistent.

PR #6933, opened 2019-03-24
Fix Issue 19738 - std.range.choose assignment breaks @safe-ty

thewilsonator:
cc @WalterBright @dnadlinger

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -1434,6 +1434,28 @@ if (isInputRange!(Unqual!R1) && isInputRange!(Unqual!R2) &&
     assert(result.equal(only(6, 7, 8, 9)));
 }
 
+//https://issues.dlang.org/show_bug.cgi?id=19738
+@safe nothrow pure @nogc unittest
+{
+    struct EvilRange
+    {
+        enum empty = true;
+        int front;
+        void popFront() @safe {}
+        auto opAssign(const ref EvilRange other)
+        {
+            *(cast(uint*)0xcafebabe) = 0xdeadbeef;
+            return this;
+        }
+    }
+
+    static assert(!__traits(compiles, () @safe
+    {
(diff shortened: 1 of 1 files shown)


std/range/package.d
```diff
@@ -1434,6 +1434,36 @@ if (isInputRange!(Unqual!R1) && isInputRange!(Unqual!R2) &&
...
+            return this;
+        }
+    }
+    enum isDIP1000 = __traits(compiles, () @safe
+    {
+      int x;
+      int* p;
+      p = &x;
+    });
+    static if (isDIP1000)
```
dnadlinger: Huh? This should never compile; DIP1000 or not.

thewilsonator: Fixed.

thewilsonator:
This is a master regression, so no stable.

thewilsonator:
Stupid question: why does this fail, when the static assert in https://run.dlang.io/is/Vjjcjf doesn't?

WalterBright:
Should remove `@trusted` from the declaration of `opAssign`

thewilsonator:
Whoops! That'd certainly do it!

thewilsonator:
Hmm, this seems to break https://github.com/dlang/phobos/blob/master/std/algorithm/iteration.d#L6181-L6201

on all except 
`"".substitute("","").equal("");` (line 6185)

`""w.substitute(""w,""w).equal(""w);` (line 6190), and 

`""d.substitute(""d,""d).equal(""d);` (line 6195)

thewilsonator:
Seems to come from either https://github.com/dlang/phobos/blob/master/std/algorithm/iteration.d#L5791 or https://github.com/dlang/phobos/blob/master/std/algorithm/iteration.d#L5800

berni44:
Is there any progress on this?

thewilsonator:
Superseded by #7348

PR #6934, opened 2019-03-26
fix issue 19765 - std.algorithm.searching.findAmong doesn't `save` li…

aG0aep6G:
…ke it should


wilzbach:
CircleCi fix is here: https://github.com/dlang/phobos/pull/6936 (should be in master soon)

PR #6935, opened 2019-03-26
make `cycle`, `splitter`, `roundRobin`, and `until` compatible with `RefRange`

aG0aep6G:
Part of a series on [issue 18657](https://issues.dlang.org/show_bug.cgi?id=18657). Previously: #6346.
This concludes std.range and std.algorithm.searching. Everything else still needs to be checked.

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -5225,10 +5225,15 @@ private struct SplitterResult(alias isTerminator, Range)
 
     @property typeof(this) save()
     {
+        import std.algorithm.mutation : move;
         auto ret = this;
-        ret._input = _input.save;
+        auto savedInput = _input.save;
+        move(savedInput, ret._input);
         static if (!fullSlicing)
-            ret._next = _next.save;
+        {
+            auto savedNext = _next.save;
+            move(savedNext, ret._next);
+        }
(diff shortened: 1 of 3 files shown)


thewilsonator:
Apologies CircleCI is having problems and as its required I can't merge it, please ping me if I forget about this after it gets fixed.

wilzbach:
CircleCi fix is here: https://github.com/dlang/phobos/pull/6936 (should be in master soon)

WalterBright:
@aG0aep6G @thewilsonator @wilzbach @andralex  There's something seriously wrong if this wacky code is needed in every `save()` function. I strongly suggest stopping these changes until RefRange is re-evaluated as to why it is in Phobos if it breaks straightforward algorithms.

aG0aep6G:
> There's something seriously wrong if this wacky code is needed in every `save()` function.

It doesn't have to be done with `move`. It can also be done with a constructor like you did in #6900 and #6903. The only thing that matters is that `opAssign` is avoided. I did with `move`, because that seemed the least intrusive to me (doesn't touch the interface at all).

Or if you guys decide that range's shouldn't do `opAssign`, then we can just remove it from `RefRange` and that's it.

PR #6938, opened 2019-03-29
add workaround for slow compilation of large unittest functions

rainers:
This got removed when converting `foreach` over tuple to `static foreach`, but that doesn't make a difference to the slow optimizer.

These changes reduce (single threaded) compile and run time of the phobos unittests from 8min50 to 5min51 for my mobile computer.


rainers:
> Perhaps add that this is a workaround in the comments

You mean replace `avoid` with `workaround`?

thewilsonator:
Yes.

rainers:
Ok, amended.

PR #6939, opened 2019-03-29
use constructors instead of `move`

aG0aep6G:
This is a fix-up of #6346 and #6935.

As far as I understand, @WalterBright favors this style (cf. #6900, #6903).

Most of these are straight forward, but `chain` and `roundRobin` are still a bit awkward, because they involve tuples. We don't have some kind of `map` for tuples in Phobos, do we?

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -1641,6 +1641,12 @@ if (isInputRange!R)
         if (!_input.empty) popFront();
     }
 
+    private this(R input, Tuple!(MutableE, uint) current)
+    {
+        _input = input;
+        _current = current;
+    }
+
     ///
     void popFront()
     {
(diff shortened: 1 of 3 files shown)


aG0aep6G:
> One possibility would be to (re)do this when copy constructors get released

I'm not sure how you'd use copy constructors here. Make every copy save the range? That's fine for new code, but changing the copy behavior of existing types seems like asking for trouble.

aG0aep6G:
I saw CircleCI and Buildkite failing with errors that seemed unrelated, so I thought I'd give them a kick with a force push. But I guess that was a bad idea. It only managed to remove the auto-merge label. Checks are still failing.

Can someone help me out here?

thewilsonator:
If you haven't rebased this recently please do so.

thewilsonator:
Yep, this was forked off from an old master we've since fixed the issues with circle but you need to have branched off after that to see those fixes on this PR, a rebase will fix your problems.

aG0aep6G:
> Yep, this was forked off from an old master we've since fixed the issues with circle but you need to have branched off after that to see those fixes on this PR, a rebase will fix your problems.

I've rebased, but this PR is against stable, not master. Does that make a difference?

thewilsonator:
Ah, circle will still fail, but is not require to pass, sorry I missed that. I'm more concerned about the build kite failure but that might just be an unfortunate feature of the current stable.

thewilsonator:
In short this needs a manual merge, which I will do tomorrow.

PR #6940, opened 2019-03-30
Revert "make `group`, `chain`, and `choose` compatible with `RefRange`"

WalterBright:
Reverts dlang/phobos#6346

The problem lies with RefRange, not with everything else.


MoonlightSentinel:
Is this still needed after #6939 was merged?

thewilsonator:
I'm guessing not.

PR #6941, opened 2019-03-30
Revert "make `cycle`, `splitter`, `roundRobin`, and `until` compatible with `RefRange`"

WalterBright:
Reverts dlang/phobos#6935

RefRange is the problem, not everything else.

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -5225,15 +5225,10 @@ private struct SplitterResult(alias isTerminator, Range)
 
     @property typeof(this) save()
     {
-        import std.algorithm.mutation : move;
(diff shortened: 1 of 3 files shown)


aG0aep6G:
@WalterBright, have you seen my #6939? It changes the code of #6935 and #6346 in the way you did in #6900. If you agree with that, we can skip this revert and its sibling (#6940).

aG0aep6G:
Also, a GitHub mention or a comment on the to-be-reverted PRs would have been nice. I only noticed this because I'm subscribed to Phobos on GitHub.

ZombineDev:
I believe this PR can be closed now that #6939 was merged.

PR #6942, opened 2019-03-31
variant.d: Gated unittests behind version(StdDdoc)

Kriyszig:
Issue in focus: <https://issues.dlang.org/show_bug.cgi?id=18818>

### DIFF
--- a/std/variant.d
+++ b/std/variant.d
@@ -726,7 +726,7 @@ public:
     }
 
     ///
-    version (unittest)
+    version (StdDdoc)
     @system unittest
     {
         Variant a;
@@ -759,7 +759,7 @@ public:
     }
 
     ///
-    version (unittest)
+    version (StdDdoc)
     @system unittest
     {
         Variant a = 5;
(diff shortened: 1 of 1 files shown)


Kriyszig:
Reabesd onto stable. Sorry about that :sweat_smile: 
P.S. Looks like CircleCI test failed as it couldn't download dmd on the VM, Rebuild should definitely pass the test but it's good to confirm.

thewilsonator:
Thanks!

> P.S. Looks like CircleCI test failed as it couldn't download dmd on the VM, Rebuild should definitely pass the test but it's good to confirm

Thats known, fixed in master, evidently wasn't back ported to stable.

wilzbach:
Argh. This should have never been merged without more thought. Now these unittests won't be run __ever__. I hope I don't have to stress that this is _bad_.

I suggest copying them over to below the `Variant` declarations for now, but we should seriously consider reintroducing `StdUnittest`.

PR #6949, opened 2019-04-09
Simplify parameter default fetching code

adamdruppe:
This avoids making a local variable which would cause errors
with `inout` parameters, and, as far as I can tell, causes no
new problems. The old implementation appears to be simply working
around compiler bugs that no longer exist.

let's see what the autotester says.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -1479,27 +1479,7 @@ if (func.length == 1 && isCallable!func)
     {
         template Get(size_t i)
         {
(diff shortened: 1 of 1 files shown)


adamdruppe:
Eh, there's yet another local down there, the whole function is fundamentally broken. (That's my view on most of std.traits, it is trying to turn parameters into variables and D doesn't really support that. Ref is broken this way too.

PR #6953, opened 2019-04-09
Factor out shared code from std.digest.toHexString overloads

benjones:
Pull out the duplicated code from the static and dynamic array overloads

(first PR from me.  LMK if this type of refactoring is useful or if there are other modules worth digging into)

### DIFF
--- a/std/digest/package.d
+++ b/std/digest/package.d
@@ -717,37 +717,10 @@ enum Order : bool
 char[num*2] toHexString(Order order = Order.increasing, size_t num, LetterCase letterCase = LetterCase.upper)
 (in ubyte[num] digest)
 {
-    static if (letterCase == LetterCase.upper)
-    {
-        import std.ascii : hexDigits = hexDigits;
-    }
-    else
-    {
-        import std.ascii : hexDigits = lowerHexDigits;
-    }
-
 
     char[num*2] result;
     size_t i;
-
-    static if (order == Order.increasing)
-    {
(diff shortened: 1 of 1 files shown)


std/digest/package.d
```diff
@@ -801,6 +747,7 @@ string toHexString(LetterCase letterCase, Order order = Order.increasing)(in uby
     return toHexString!(order, letterCase)(digest);
 }
 
+
```
wilzbach: unrelated change

std/net/curl.d
```diff
@@ -203,9 +203,7 @@ version (unittest)
             }
             catch (Throwable e)
             {
-                import core.stdc.stdlib : exit, EXIT_FAILURE;
-                stderr.writeln(e);
-                exit(EXIT_FAILURE); // Bugzilla 7018
+                stderr.writeln(e);  // Bugzilla 7018
```
wilzbach: unrelated?

benjones: Yes, looks like it's a leftover or something from the previous master commit.  What's the easiest way for me to fix these?

carun: @benjones You can amend the commit and force push it.

wilzbach:
@benjones you still have the commit from Ian in your list. Here's what you can do:

```
git reset --soft HEAD~2
git commit
```

This will create a new fresh commit instead of the two existing ones.

PR #6954, opened 2019-04-10
Fix issue 19799: add const toString to Nullable

FeepingCreature:
This avoids miscompilation when trying to format `const(Nullable!string)` due to attempted implicit `.get` conversion to string from T.init causing a compiletime error.

_reaches up, flips down counter: "It's been [0] days since the last bug caused by `alias Nullable.get this`"_

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2781,7 +2781,28 @@ Params:
     }
 
     /// ditto
+    string toString() const
+    {
+        import std.array : appender;
+        auto app = appender!string();
+        auto spec = singleSpec("%s");
+        toString(app, spec);
+        return app.data;
+    }
+
+    /// ditto
     void toString(W)(ref W writer, scope const ref FormatSpec!char fmt)
+    if (isOutputRange!(W, char))
+    {
+        import std.range.primitives : put;
+        if (isNull)
+            put(writer, "Nullable.null");
+        else
+            formatValue(writer, _value.payload, fmt);
+    }
+
+    /// ditto
+    void toString(W)(ref W writer, scope const ref FormatSpec!char fmt) const
     if (isOutputRange!(W, char))
     {
         import std.range.primitives : put;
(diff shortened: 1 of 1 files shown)


std/typecons.d
```diff
@@ -2853,6 +2874,16 @@ Returns:
...
 }
 
+// Issue 19799
+@safe unittest
+{
+    import std.format : format;
+
+    const Nullable!string a = const(Nullable!string)();
+
+    format!"%s"(a);
```
wilzbach: How about asserting on the output here?

Basile-z:
Patch coverage is abnormally low, it should be of 90% at least. Is something not tested ?

ZombineDev:
One potential problem is that `Nullable!T.toString` may not work if `T.toString`  is not `const` as well.

std/typecons.d
FeepingCreature: I don't think this code wants to care what the output is, as long as there is one. I think that'd be misleading considering it's not the point of the bug, and hence not the point of the unittest.

FeepingCreature:
@ZombineDev Weirdly, as far as I can tell it still does. I don't understand why.

MoonlightSentinel:
@FeepingCreature Because `toString() const` is not a template a non-const `T.toString` gets called on a `const(T)`.
Maybe it would be a better approach to test wether T.toString is const and select the corresponding function accordingly.

```
static if(hasConstToString) {
    string toString() const { ... }
else {
    string toString() { ... }
}
```

FeepingCreature:
Well heck!

There is, as far as I can tell, no reliable way to determine if a type allows a `const toString`. If I define a `struct S { string toString() { return null; } }`, I can still call `(const(S)()).to!string`, it'll just use some sort of default conversion.

FeepingCreature:
I figured out why it works.

Both `toString()` and `toString() const` always compile, because they call `formatValue` and `formatValue` works whether or not the type has an appropriate `toString` defined. If it doesn't, it just selects a fallback method.

std/typecons.d
FeepingCreature: Assertions added, purely for coverage's sake.

Basile-z:
Yeah coverage looks correct now. With D sometimes it can be less than 100% because of CTFE and other compile-time stuff but below 90% is a good sign that something is not tested.

PR #6956, opened 2019-04-11
Comment broken test in std/traits.d

RazvanN7:
This currently blocks: https://github.com/dlang/dmd/pull/9585 . The test can be uncommented and fixed after the dmd PR is pulled


wilzbach:
(just in case the DMD PR gets stuck)

RazvanN7:
@wilzbach Good idea, thanks. Done.

PR #6958, opened 2019-04-12
fix buildkite build

aG0aep6G:
Works for me locally. Let's see if Buildkite agrees.


aG0aep6G:
> Any chance you could reduce the example from vibe.d to a test case?

Not right now.

aG0aep6G:
> > Any chance you could reduce the example from vibe.d to a test case?
> 
> Not right now.

There it is: https://issues.dlang.org/show_bug.cgi?id=19807

PR #6963, opened 2019-04-16
bugzilla issue 4391 add curry to std.functional 

benjones:
I wasn't sure the best way to add template constraints to make sure that things without state (e.g. functions) use the alias template version and things with state (callable objects) use the type template version.  LMK how best to go about that.

This addresses bugzilla issue 4391

### DIFF
--- a/std/functional.d
+++ b/std/functional.d
@@ -871,7 +871,8 @@ Params:
 Returns:
     A single parameter callable object
 */
-template curry(alias F) if (isCallable!F && Parameters!F.length)
+template curry(alias F)
+if (isCallable!F && Parameters!F.length)
 {
     //inspired from the implementation from Artur Skawina here:
     //https://forum.dlang.org/post/mailman.1626.1340110492.24740.digitalmars-d@puremagic.com
@@ -927,7 +928,8 @@ pure @safe @nogc nothrow unittest
 }
 
 ///ditto
-auto curry(T)(T t) if (isCallable!T && Parameters!T.length)
+auto curry(T)(T t)
+if (isCallable!T && Parameters!T.length)
 {
     static auto fun(ref T inst, ref Parameters!T args)
     {
--- a/std/functional.d
+++ b/std/functional.d
@@ -36,6 +36,10 @@ $(TR $(TH Function Name) $(TH Description)
         $(TD Creates a function that binds the first argument of a given function
         to a given value.
     ))
+    $(TR $(TD $(LREF curry))
+        $(TD Converts a multi-argument function into a series of single-argument
+        functions.  f(x, y) == curry(f)(x)(y)
+    ))
     $(TR $(TD $(LREF reverseArgs))
         $(TD Predicate that reverses the order of its arguments.
     ))
@@ -856,6 +860,152 @@ template partial(alias fun, alias arg)
     auto result2 = partialDelegate(4)(2);
     assert(result2 == 1.5f);
 }
+
+/**
+Takes a function of (potentially) many arguments, and returns a function taking
+one argument and returns a callable taking the rest.  f(x, y) == curry(f)(x)(y)
+
+Params:
+    F = a function taking at least one argument
+    t = a callable object whose opCall takes at least 1 object
+Returns:
+    A single parameter callable object
+*/
+template curry(alias F)
+{
+    //inspired from the implementation from Artur Skawina here:
+    //https://forum.dlang.org/post/mailman.1626.1340110492.24740.digitalmars-d@puremagic.com
+    //This implementation stores a copy of all filled in arguments with each curried result
+    //this way, the curried functions are independent and don't share any references
+    //eg: auto fc = curry!f;  auto fc1 = fc(1); auto fc2 = fc(2); fc1(3) != fc2(3)
+    import std.traits : isCallable;
+    static if (isCallable!F)
+    {
+        auto curry()
+        {
+            import std.traits : Parameters;
+            alias FParams = Parameters!F;
+
+            struct CurryImpl(size_t N)
+            {
+                FParams[0 .. N] storedArguments;
+                static if (N > 0)
+                {
+                    this(const ref CurryImpl!(N-1) prev, auto ref FParams[N] arg)
+                    {
+                        static foreach (i; 0 .. N-1)
+                        {
+                            storedArguments[i] = prev.storedArguments[i];
+                        }
+                        storedArguments[N-1] = arg;
+                    }
+                }
+
+                auto opCall(auto ref FParams[N] arg) const
+                {
+                    static if (N == FParams.length -1)
+                    {
+                        return F(storedArguments, arg);
+                    }
+                    else
+                    {
+                        return CurryImpl!(N+1)(this, arg);
+                    }
+                }
+            }
+            CurryImpl!0 result;
+            return result;
+        }
+    }
+}
+
+///
+pure @safe nothrow unittest
+{
+    int f(int x, int y, int z)
+    {
+        return x + y + z;
+    }
+    auto cf = curry!f;
+    auto cf1 = cf(1);
+    auto cf2 = cf(2);
+
+    assert(cf1(2)(3) == f(1, 2, 3));
+    assert(cf2(2)(3) == f(2, 2, 3));
+}
+
+///ditto
+auto curry(T)(T t)
+{
+    alias toc = t.opCall;
+    auto fun(Parameters!toc args)
+    {
+        return t(args);
+    }
+    return curry!fun;
+}
+
+///
+pure @safe @nogc nothrow unittest
+{
+    //works with callable structs too
+    struct S
+    {
+        int w;
+        int opCall(int x, int y, int z)
+        {
+            return w + x + y + z;
+        }
+    }
+
+    S s;
+    s.w = 5;
+
+    auto cs = curry(s);
+    auto cs1 = cs(1);
+    auto cs2 = cs(2);
+
+    assert(cs1(2)(3) == s(1, 2, 3));
+    assert(cs1(2)(3) == (1 + 2 + 3 + 5));
+    assert(cs2(2)(3) ==s(2, 2, 3));
+}
+
+
+@safe pure nothrow unittest
+{
+    //currying a single argument function does nothing
+    int pork(int a){ return a*2;}
+    auto curryPork = curry!pork;
+    assert(curryPork(0) == pork(0));
+    assert(curryPork(1) == pork(1));
+    assert(curryPork(-1) == pork(-1));
+    assert(curryPork(1000) == pork(1000));
+
+    //test 2 argument function
+    int mixedVeggies(int a, int b)
+    {
+        return a + b;
+    }
+
+    auto mixedCurry = curry!mixedVeggies;
+    assert(mixedCurry(10)(20) == mixedVeggies(10,20));
+    assert(mixedCurry(100)(200) == mixedVeggies(100,200));
+
+    //struct with opCall
+    struct S
+    {
+        int opCall(int x, int y, int z) const pure nothrow @nogc
+        {
+            return x*y*z;
+        }
+    }
+
+    S s;
+    auto curriedStruct = curry(s);
+    assert(curriedStruct(1)(2)(3) == s(1, 2, 3));
+    assert(curriedStruct(300)(20)(10) == s(300, 20, 10));
+}
+
 
 /**
 Takes multiple functions and adjoins them together.
--- a/std/functional.d
+++ b/std/functional.d
@@ -871,56 +871,48 @@ Params:
 Returns:
     A single parameter callable object
 */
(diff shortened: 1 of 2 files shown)


wilzbach:
Hi. Thanks a lot for your PR!
A few first things:

- contains tabs
- needs to reference the Bugzilla issue in the git commit for auto-close
- new functions require name approval from Andrei

benjones:
Tab issue should be fixed now and I added the BZ number to the commit message.

std/functional.d
```diff
@@ -899,6 +903,89 @@ template partial(alias fun, alias arg)
...
+{
+
+	alias toc = t.opCall;
+	auto fun(Parameters!toc args){
+		return t(args);
+	}
+	return curry!fun;
+}
+
+@safe pure nothrow unittest{
```
wilzbach: All public symbols should have at least one short and simply to understand unittest that briefly explains the function.

std/functional.d
```diff
@@ -899,6 +903,89 @@ template partial(alias fun, alias arg)
...
+    A single parameter callable object
+*/
+template curry(alias F)
+{
+	//adapted from the implementation from Artur Skawina here:
+	//https://forum.dlang.org/post/mailman.1626.1340110492.24740.digitalmars-d@puremagic.com
+	auto curry()
+	{
+		struct CurryImpl {
+			import std.traits;
```
wilzbach: Imports inside of structs should be selective to make the reader aware of the locally used symbols.

std/functional.d
```diff
@@ -899,6 +903,89 @@ template partial(alias fun, alias arg)
...
+					*p = a;
+					static if (N==args.length-1)
+					{
+						return F(args);
+					}
+					else
+						return &opCall1!(N+1);
+				}
+			}
+			alias opCall1!() opCall;
```
wilzbach: This alias style is deprecated. Can we use the newer one with assignment?

std/functional.d
benjones: Fixed

std/functional.d
benjones: Fixed

benjones:
I revamped the implementation so that the returned objects each store a copy of the arguments passed so far, so they're all independent.  I also added some example unittests.

std/functional.d
```diff
@@ -899,6 +903,149 @@ template partial(alias fun, alias arg)
...
+Takes a function of (potentially) many arguments, and returns a function taking
+one argument and returns a callable taking the rest.  f(x, y) == curry(f)(x)(y)
+
+Params:
+    F = a function taking at least one argument
+    t = a callable object whose opCall takes at least 1 object
+Returns:
+    A single parameter callable object
+*/
+template curry(alias F)
```
Basile-z: `if 'isCallable!F` from `std.traits` should work as a constraint.

std/functional.d
```diff
@@ -899,6 +903,149 @@ template partial(alias fun, alias arg)
...
+    //eg: auto fc = curry!f;  auto fc1 = fc(1); auto fc2 = fc(2); fc1(3) != fc2(3)
+    auto curry()
+    {
+        import std.traits : Parameters;
+        alias FParams = Parameters!F;
+
+        struct CurryImpl(size_t N)
+        {
+            FParams[0..N] storedArguments;
+            static if(N > 0)
```
Basile-z: space between `if` and left paren.

std/functional.d
```diff
@@ -899,6 +903,149 @@ template partial(alias fun, alias arg)
...
+            {
+                this(const ref CurryImpl!(N-1) prev, FParams[N] arg)
+                {
+                    static foreach(i; 0..N-1)
```
Basile-z: space between `foreach` and left paren

std/functional.d
```diff
@@ -899,6 +903,149 @@ template partial(alias fun, alias arg)
...
+                    {
+                        storedArguments[i] = prev.storedArguments[i];
+                    }
+                    storedArguments[N-1] = arg;
+                }
+            }
+
+            auto opCall(FParams[N] arg) const
+            {
+                static if(N == FParams.length -1)
```
Basile-z: space...

benjones: Is there an easy way to use tooling to fix these automatically in the future?  Running the whole file through dfmt seemed like it would add lots of unnecessary changes

wilzbach: Unfortunately not for Phobos as it's not formatted consistently (it would create too much churn and unrelated changes to run dfmt on entire Phobos).
There's an open issue for dfmt to run it only on the current diff, but so far this is not supported.

std/functional.d
Basile-z: I've seen only later the other overload (for opCall) so maybe no possible

std/functional.d
Basile-z: And here the brace on its own line, although the style checker didn't detect.

std/functional.d
benjones: I wrapped the first overload in static if(isCallable!F) and my tests still work.  The important distinction is "is there state to be stored or not"

berni44:
ping?

std/functional.d
```diff
@@ -899,6 +903,153 @@ template partial(alias fun, alias arg)
...
+    auto cs1 = cs(1);
+    auto cs2 = cs(2);
+
+    assert(cs1(2)(3) == s(1, 2, 3));
+    assert(cs1(2)(3) == (1 + 2 + 3 + 5));
+    assert(cs2(2)(3) ==s(2, 2, 3));
+}
+
+
+@safe pure nothrow unittest{
```
thewilsonator: Brace on new line

PetarKirov:
@atilaneves do you approve the addition of this function?

std/functional.d
```diff
@@ -899,6 +903,154 @@ template partial(alias fun, alias arg)
...
+                }
+            }
+            CurryImpl!0 result;
+            return result;
+        }
+    }
+}
+
+///
+@safe unittest
```
PetarKirov: It looks like both `curry` overloads should also be usable from `@nogc nothrow pure` functions (not just safe), can you verify this by adding those attributes to the unittest?

std/functional.d
```diff
@@ -899,6 +903,154 @@ template partial(alias fun, alias arg)
...
+    auto cf2 = cf(2);
+
+    assert(cf1(2)(3) == f(1, 2, 3));
+    assert(cf2(2)(3) == f(2, 2, 3));
+}
+
+///ditto
+auto curry(T)(T t)
+{
+
```
PetarKirov: Delete empty line.

std/functional.d
```diff
@@ -899,6 +903,154 @@ template partial(alias fun, alias arg)
...
+    alias toc = t.opCall;
+    auto fun(Parameters!toc args)
+    {
+        return t(args);
+    }
+    return curry!fun;
+}
+
+///
+@safe unittest
```
PetarKirov: Ditto, should woke with all arguments.

std/functional.d
```diff
@@ -899,6 +903,154 @@ template partial(alias fun, alias arg)
...
+
+    assert(cs1(2)(3) == s(1, 2, 3));
+    assert(cs1(2)(3) == (1 + 2 + 3 + 5));
+    assert(cs2(2)(3) ==s(2, 2, 3));
+}
+
+
+@safe pure nothrow unittest
+{
+
```
PetarKirov: Delete empty line

benjones:
Looks like the unittests (the first one at least, not sure about the 2nd yet) can't be @nogc for some reason... Can anyone see what's being allocated?  A closure or something?

MoonlightSentinel:
> Looks like the unittests (the first one at least, not sure about the 2nd yet) can't be @nogc for some reason... Can anyone see what's being allocated? A closure or something?

Exactly because fun is a non-static function referencing its parameter. I think a better apporach to handle opCall would be something like this:
```
auto curry(T)(T t)
{
    static auto fun(ref T op, ref Parameters!T args)
    {
        return op(args);
    }
    return curry!fun()(t);
}
```

std/functional.d
MoonlightSentinel: `isCallable` should definitly be a template constraint to avoid akward template errors, e.g.:

```
curry!1();

std\functional.d(950): Error: function expected before (), not curry!1 of type void
    curry!1();
```

std/functional.d
```diff
@@ -899,6 +903,152 @@ template partial(alias fun, alias arg)
...
+                    {
+                        static foreach (i; 0 .. N-1)
+                        {
+                            storedArguments[i] = prev.storedArguments[i];
+                        }
+                        storedArguments[N-1] = arg;
+                    }
+                }
+
+                auto opCall(FParams[N] arg) const
```
MoonlightSentinel:
```suggestion
                auto opCall(auto ref FParams[N] arg) const
```

std/functional.d
```diff
@@ -899,6 +903,152 @@ template partial(alias fun, alias arg)
...
+        {
+            import std.traits : Parameters;
+            alias FParams = Parameters!F;
+
+            struct CurryImpl(size_t N)
+            {
+                FParams[0 .. N] storedArguments;
+                static if (N > 0)
+                {
+                    this(const ref CurryImpl!(N-1) prev, FParams[N] arg)
```
MoonlightSentinel:
```suggestion
                    this(const ref CurryImpl!(N-1) prev, auto ref FParams[N] arg)
```

MoonlightSentinel:
```suggestion
                    this(ref CurryImpl!(N-1) prev, FParams[N] arg)
```

Does not work with indirections, please add a test like:

```
static class C
{
    int a;
    string b;
}

static C test(int* a, C b, S[] c) {
    return b;
}

auto ct = curry!test();
scope b = new C();
assert(ct(null)(b)(null) is b);
```

std/functional.d
```diff
@@ -899,6 +903,152 @@ template partial(alias fun, alias arg)
...
+    auto cf = curry!f;
+    auto cf1 = cf(1);
+    auto cf2 = cf(2);
+
+    assert(cf1(2)(3) == f(1, 2, 3));
+    assert(cf2(2)(3) == f(2, 2, 3));
+}
+
+///ditto
+auto curry(T)(T t)
```
MoonlightSentinel:
```suggestion
auto curry(T)(auto ref T t)
```

std/functional.d
```diff
@@ -899,6 +903,152 @@ template partial(alias fun, alias arg)
...
+
+    assert(cs1(2)(3) == s(1, 2, 3));
+    assert(cs1(2)(3) == (1 + 2 + 3 + 5));
+    assert(cs2(2)(3) ==s(2, 2, 3));
+}
+
+
+@safe pure nothrow unittest
+{
+    //currying a single argument function does nothing
```
MoonlightSentinel: `curry!pork` could be optimized into `alias curry = pork`

std/functional.d
```diff
@@ -899,6 +903,152 @@ template partial(alias fun, alias arg)
...
+    int pork(int a){ return a*2;}
+    auto curryPork = curry!pork;
+    assert(curryPork(0) == pork(0));
+    assert(curryPork(1) == pork(1));
+    assert(curryPork(-1) == pork(-1));
+    assert(curryPork(1000) == pork(1000));
+
+    //test 2 argument function
+    int mixedVeggies(int a, int b)
```
MoonlightSentinel: Mutliple structurally equivalent tests are useless - use different parameter/return types

std/functional.d
```diff
@@ -899,6 +903,152 @@ template partial(alias fun, alias arg)
...
+    //eg: auto fc = curry!f;  auto fc1 = fc(1); auto fc2 = fc(2); fc1(3) != fc2(3)
+    import std.traits : isCallable;
+    static if (isCallable!F)
+    {
+        auto curry()
+        {
+            import std.traits : Parameters;
+            alias FParams = Parameters!F;
+
+            struct CurryImpl(size_t N)
```
MoonlightSentinel: Could reside outside of `auto curry()...` or at least be static to avoid the uneccessary context pointer

MoonlightSentinel:
Sorry for the WoT above. 

I experimented a bit with your branch to tackle the aforementioned problems and ended up with [this](https://github.com/MoonlightSentinel/phobos/tree/curryV2). It works correctly (as far as I can tell) but `DIP1000` complains about escaping references ...

benjones:
Those look good to me.  What's the best way to incorporate those? For me to merge from your branch into the one for this PR?

MoonlightSentinel:
Guess you could cherry-pick the commit (19ece66023162ec8decbb15dc590f15145163581) from my branch and replace your current one.

Note that i changed my branch a few minutes ago to error gracefully for zero-argument callables

MoonlightSentinel:
@benjones Once the style issue (if constraint on new line) is fixed this should be good.

MoonlightSentinel:
@PetarKirov You might want to give this another look

PR #6964, opened 2019-04-17
moveEmplace: optimise and make it betterC friendly

9il:
originally from https://github.com/dlang/druntime/pull/2565


thewilsonator:
Hmm, I don't understand why this would fail here.

9il:
maybe something define `init`?

9il:
No idea why the first commit does not work. The second commit is wrong (because of `= void` field initialization)

PR #6965, opened 2019-04-18
Fix #19740: Incorrect result of BigInt * BigInt

shove70:
https://issues.dlang.org/show_bug.cgi?id=19740


thewilsonator:
Please rebase and add tests.

PR #6966, opened 2019-04-18
Make the symlink for phobos soname

ErnyTech:
We must have a symlink to the soname because
otherwise it is problematic to link the shared
library.

Before this patch to distribute phobos it was
necessary to manually create the symlink
(e.g. https://github.com/dlang/installer/blob/482840bf5bd8e94ebb12aabaa7ddb0dac319be90/linux/dmd_phobos.sh#L169),
this was a really uncomfortable and confusing solution
for external package maintainers.

This symlink was actually already created in the
phobos build phase but for some reason unknown to
me was not copied into the installation target.

I also noticed that the libphobos2.so symlink is
already created, so I simply removed the redundant
symlink creation with a simple copy from generated/$(OS)/$(BUILD)/$(MODEL)

Signed-off-by: Ernesto Castellotti <erny.castell@gmail.com>

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -451,7 +451,8 @@ install2 : all
 	cp $(LIB) $(INSTALL_DIR)/$(OS)/$(lib_dir)/
 ifeq (1,$(SHARED))
 	cp -P $(LIBSO) $(INSTALL_DIR)/$(OS)/$(lib_dir)/
-	ln -sf $(notdir $(LIBSO)) $(INSTALL_DIR)/$(OS)/$(lib_dir)/libphobos2.so
+	cp -P $(ROOT)/$(SONAME) $(INSTALL_DIR)/$(OS)/$(lib_dir)/
+	cp -P $(ROOT)/libphobos2.so $(INSTALL_DIR)/$(OS)/$(lib_dir)/
 endif
 	mkdir -p $(INSTALL_DIR)/src/phobos/etc
 	mkdir -p $(INSTALL_DIR)/src/phobos/std


CyberShadow:
Sorry, what is a "syslink"? I can't find a definition for that term. Is it a special kind of symlink?

ErnyTech:
> Sorry, what is a "syslink"? I can't find a definition for that term. Is it a special kind of symlink?

common abbreviation of symbolic link

CyberShadow:
It seems neither an abbreviation nor common according to what I found. Perhaps it would be better to stick to established terminology.

ErnyTech:
just a bad writing mistake of symlink

ErnyTech:
I solved the bad mistake that I made

jpf91:
How does dmd call gcc to link to `libphobos2.so`? I think you may be able to pass the full path to a library to specify a certain version, though I never tried that.

But for the common `gcc -lphobos2` way to work, you'll indeed need `libphobos2.so`. `gcc -lphobos2` will search for the `libphobos2.so` file, but the referenced version will be what the lib has in its `soname`: So `-lphobos2` reads `libphobos2.so` which is a link to `libphobos2.so.81` and **has soname `libphobos2.so.81`**. The generated executable will directly reference `libphobos2.so.81`, as that was in the `soname`.  TLDR: You don't need `libphobos2.so` to run programs, but gcc needs it for `-lphobos2` to work (but maybe there's also some way to force a version with gcc which works without the link).

This is also the reason why debian moved `libfoo.so` to the `-dev` packages, these links are only required to build apps, not to run apps. But all in all, this change seems reasonable.

ErnyTech:
libphobos2.so was copied to the installation directory even before, but not libphobos2.so.X which is the soname.

However the build task creates all the links, so I simply modified it to be copied to the installation directory

ErnyTech:
before this path this is the content of the installation directory:
```
libphobos2.a
libphobos2.so -> libphobos2.so.0.X.Y
libphobos2.so.0.X.Y
```

after
```
libphobos2.a
libphobos2.so -> libphobos2.so.0.X.Y
libphobos2.so.0.X -> libphobos2.so.0.X.Y
libphobos2.so.0.X.Y

**Comment by jpf91** [comment]
OK, then this makes even more sense ;-)

**Comment by wilzbach** [comment]
> How does dmd call gcc to link to libphobos2.so? I think you may be able to pass the full path to a library to specify a certain version, though I never tried that.

You can see that with `-v`:

```
cc onlineapp.o -o /tmp/dmd_runhu6rIw -g -m64 -Xlinker --export-dynamic -L/dlang/dmd/linux/bin64/../lib64 -Xlinker -Bstatic -lphobos2 -Xlinker -Bdynamic -lpthread -lm -lrt -ldl 
```

https://run.dlang.io/is/rc7Oz6

I think we should definitely change DMD to require the respective version of Phobos (the code would be here: https://github.com/dlang/dmd/blob/master/src/dmd/link.d), but that's something for another day.

PR #6970, opened 2019-04-20
Fixes #19740: Incorrect result of BigInt * BigInt.

shove70:
#19740


std/bigint.d
```diff
@@ -1900,3 +1900,23 @@ void divMod(const BigInt dividend, const BigInt divisor, out BigInt quotient, ou
...
+    BigInt a = BigInt(
+        "241127122100380210001001124020210001001100000200003101000062221012075223052000021042250111300200000000000" ~
+        "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000");
+    BigInt b = BigInt(
+        "700200000000500418321000401140010110000022007221432000000141020011323301104104060202100200457210001600142" ~
+        "000001012245300100001110215200000000120000000000000000000000000000000000000000000000000000000000000000000" ~
+        "00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000");
+
+    BigInt c = a * b;
+    assert (c == BigInt(
```
thewilsonator:
```
Enforce no space between assert and the opening brace, i.e. assert(
--
  | grep -nrE 'assert +\(' $(find etc std -name '*.d') ; test $? -eq 1
  | std/bigint.d:1916:    assert (c == BigInt(
```

.gitignore
```diff
@@ -25,3 +25,4 @@ obj/
 .dub/
 libphobos_*
 source/
+/.project
```
thewilsonator: Why is this necessary?

thewilsonator:
Otherwise LGTM.

.gitignore
shove70: This is the project engineering document produced by my local IDE. It is not necessary to do so, but it will be more convenient for me to submit it it next time after doing so.

std/bigint.d
shove70: Thanks.

.gitignore
wilzbach: It's okay to add, but please do so in the respective section of the gitignore and in a separate PR. Thanks!

PR #6971, opened 2019-04-20
Fix tests that fail on version (BigEndian)

ibuclaw:
Tested on SystemZ/Linux and SPARC/Solaris.


quickfur:
LGTM

ZombineDev:
I force pushed the branch in order to restart DAutoTest, which was failing due to a temporary network issue (failing to download and 2.079).

PR #6972, opened 2019-04-21
Fix issue 19740 - Incorrect result of BigInt * BigInt.

shove70:
Fix issue 19740.


thewilsonator:
Please make sure the commit message contains "Fix issue 19740" so that its is picked up by the bot.

n8sh:
@thewilsonator the bot didn't pick this up because it looks at commit messages rather than PR titles. At this point what's the the right thing to make the bug get closed and the fix noted in the changelog?

thewilsonator:
Add another PR that has the correct commit message and provide a link to this PR from it, or add a changelog entry directly.

PR #6976, opened 2019-04-24
Fix issue 13471: can't use crc at CTFE

thewilsonator:


std/digest/crc.d
```diff
@@ -279,6 +283,22 @@ if (N == 32 || N == 64)
         }
 }
 
+///
```
n8sh: Why insert this test into the documentation? What does it demonstrate?

PR #6979, opened 2019-04-26
std.datetime.systime: Skip unsupported clocks in unittests

ibuclaw:
For the same reasons why we can't test all clocks in core.time.

### DIFF
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -113,6 +113,16 @@ version (StdUnittest)
     initializeTests();
 }
 
+version (unittest) private bool clockSupported(ClockType c)
+{
+    // Skip unsupported clocks on older linux kernels, assume that only
+    // CLOCK_MONOTONIC and CLOCK_REALTIME exist, as that is the lowest
+    // common denominator supported by all versions of Linux pre-2.6.12.
+    version (Linux_Pre_2639)
+        return c == ClockType.normal || c == ClockType.precise;
+    else
+        return true;
+}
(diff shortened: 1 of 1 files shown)


wilzbach (dismissed):
Let's not complicate the makefiles here, they are already a huge mess. Let's ask for the major/minor versions directly in the code. It's straight-forward:

https://run.dlang.io/is/gHJyqg

```d
import std.stdio, std.range, std.string;
import core.sys.posix.sys.utsname : utsname;
private static extern(C) int uname(utsname* __name) @nogc nothrow;

void main()
{
    utsname uts;
    uname(&uts);
    char* p = uts.release.ptr;
    auto s = p.fromStringz.split(".");
    const major = s.front;
    const minor = s.dropOne.front;
    writeln("Kernel ", major, " ",  minor);
}
```

wilzbach:
For reference, the druntime PR which introduced into the Druntime Makefiles was this one: https://github.com/dlang/druntime/pull/1356

ibuclaw:
Well, there should be a proper configure system in place, rather than assuming that functions and constants (such as random example `utimensat()`) always exists if platform matches a condition.

ibuclaw:
Rebased, removing the makefile change.  GDC will run tests with this version set anyway as the configure/make process builds both druntime and phobos in the same generated file.

ibuclaw:
@thewilsonator missing approval?

thewilsonator (approved):

PR #6984, opened 2019-04-28
fix Issue 19777 - SortedRange.opSlice should infer safety

MartinNowak:
- unittests did fail with DIP1000 due to missing return scope
  inference in chained range, thus leading to a scope violation
  which prevented `@safe` inference
- fixed unittests by adding return scope to chain.Result
- would still fail under dip1000 with other ranges that
  miss return scope on their opSlice
- should not have any effect on non-DIP1000 usage


MartinNowak:
Please ignore the dlang/druntime benchmark link failure in buildkite, this is due to an incorrect setup (the benchmark would import phobos stable but link against phobos head).

PR #6988, opened 2019-04-29
Revert "Decrease template instantiation in std.algorithm.searching : all & any"

n8sh:
Reverts dlang/phobos#6960. Reason: `foreach` does not auto-decode `string` & `wstring`.


wilzbach:
Can we add a test case that fails now, but will be fixed by the revert?

n8sh:
@wilzbach added tests.

ZombineDev:
@n8sh please rebase your branch onto `stable`, as otherwise the upcoming release won't include your fix. As a rule of thumb, if a bug exists in the stable branch, the bug fix should also  be merged there.

n8sh:
@ZombineDev At the time this PR was submitted the bug didn't exist in stable. Time marched on.

PR #6991, opened 2019-05-02
Fix Issue 13882 - Use compile-time length check in std.algorithm.copy

n8sh:
It's unclear to me if this is worth doing. Fixed-length arrays aren't ranges anyway so this case wouldn't be occurring inside generic range templates.


n8sh:
It's unclear to me if this is worth doing. Fixed-length arrays aren't ranges anyway so this case wouldn't be occurring inside generic range templates.

PR #6993, opened 2019-05-06
math.d: disable inline asm tests if unsupported

belka-ew:
All these tests call directly or indirectly `resetIeeeFlags` which throws "Not yet supported" if inline assembly isn't supported.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -2322,7 +2322,7 @@ private T expImpl(T)(T x) @safe pure nothrow @nogc
     return x;
 }
 
-@safe @nogc nothrow unittest
+version (InlineAsm_X86_Any) @safe @nogc nothrow unittest
 {
     FloatingPointControl ctrl;
     if (FloatingPointControl.hasExceptionTraps)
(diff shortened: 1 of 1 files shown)


thewilsonator:
```
std/math.d(5092:6)[warn]: Public declaration 'rint' has no documented example.
--
  | std/math.d(5531:8)[warn]: Public declaration 'IeeeFlags' has no documented example.
  | std/math.d(5768:6)[warn]: Public declaration 'resetIeeeFlags' has no documented example.
  | std/math.d(5787:21)[warn]: Public declaration 'ieeeFlags' has no documented example.
  | std/math.d(5847:8)[warn]: Public declaration 'FloatingPointControl' has no documented example.
  | posix.mak:554: recipe for target 'dscanner' failed
  | make: *** [dscanner] Error 1
```

thewilsonator:
Not sure what the correct course of action is here. Maybe 
```
version (InlineAsm_X86_Any)
    version = doTheTests;
version (StdDoc)
    version = doTheTests;

version (doTheTests) unittest 
{
    //...
}
```
?

ibuclaw:
Should there be a way to add ieeeFlags tests on a per-architecture basis?  There exists at least ARM, AArch64, and RISC-V support for these (not sure about _passing_ the existing tests though).

kinke:
There's already a `version = IeeeFlagsSupport` for supported archs (but currently only guarding the public methods/properties of `struct IeeeFlags`). So guarding the few checks in unittests instead of disabling whole unittest blocks (with important other checks) would sound much better to me.

belka-ew:
> version (doTheTests) unittest 

dscanner doesn't recognize this.

> So guarding the few checks in unittests instead of disabling whole unittest blocks (with important other checks) would sound much better to me.

- Unofficial version checks land in the documentation which isn't nice
- the unittests will do different things dependent on the inline asm support

The actual problem is that `IeeeFlagsSupport` lies about supported platforms.

belka-ew:
Some doc tests already have `version (D_HardFloat)`. Then I'll do the same for these.
https://github.com/dlang-community/D-Scanner/issues/654

kinke:
So no dedicated `exp` tests for non-x86 archs anymore... sigh.

PR #6994, opened 2019-05-06
Issue 19836 followup - also fix on 32 bit machines

n8sh:
Changed the private enhanced seeding method for `rndGen` to something that is fast on both 64 bit and 32 bit machines so can be enabled regardless of architecture. When compiled with LDC it is about 1.35x the speed of public `Mt19937.seed(uint)`.

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -1813,8 +1813,8 @@ A singleton instance of the default random number generator
     {
         static if (isSeedable!(Random, ulong))
             result.seed(unpredictableSeed!ulong); // Avoid unnecessary copy.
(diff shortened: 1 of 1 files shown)


thewilsonator:
Looks fine. Will wait until after dconf for more eyes. Ping me if i forget.

wilzbach:
>  When compiled with LDC it is about 1.35x the speed of public Mt19937.seed(uint).

@n8sh did you measure the difference on 64-bit against the existing version? In other words: will the generalization seeding for 64-bit slower?

n8sh:
@wilzbach with ldc2 -O2 it's also in the neighborhood of 30% faster than #6985 on 64-bit.

thewilsonator:
Awesome.

PR #6995, opened 2019-05-07
Fix Issue 19850 - double nested joiner .back falsely thinks the range…

Spoov:
… is empty


ZombineDev:
I've rebased the branch onto `stable` so the fix can be included in the next point release.

PR #7000, opened 2019-05-11
Fix issues arising from the DMD 9029 fix (types match to alias)

FeepingCreature:
This PR represents two minor upfront fixes to make the DMD 9029 fix pass. (With 9029, primitive types can be passed to alias parameters same as named types.)


thewilsonator:
Tests?

FeepingCreature:
Belong with the DMD pr - this is just fixes.

PR #7001, opened 2019-05-11
std.random/std.datetime.systime: Don't use uint as foreach index

ibuclaw:
Fix for https://github.com/dlang/dmd/pull/9770

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -1350,15 +1350,15 @@ if (isUnsigned!UIntType && !(sa > 0 && sb > 0 && sc > 0))
         static if (UIntType.sizeof == uint.sizeof)
         {
             // Initialization routine from MersenneTwisterEngine.
-            foreach (uint i, ref e; _state)
+            foreach (size_t i, ref e; _state)
             {
-                e = (x0 = (1_812_433_253U * (x0 ^ (x0 >> 30)) + i + 1));
+                e = (x0 = (1_812_433_253U * (x0 ^ (x0 >> 30)) + cast(uint) i + 1));
                 // Xorshift requires merely that not every word of the internal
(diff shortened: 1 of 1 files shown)


std/random.d
```diff
@@ -1350,15 +1350,15 @@ if (isUnsigned!UIntType && !(sa > 0 && sb > 0 && sc > 0))
...
             {
-                e = (x0 = (1_812_433_253U * (x0 ^ (x0 >> 30)) + i + 1));
+                e = (x0 = (1_812_433_253U * (x0 ^ (x0 >> 30)) + cast(uint)i + 1));
                 // Xorshift requires merely that not every word of the internal
                 // array is 0. For historical compatibility the 32-bit word version
                 // has the stronger requirement that not any word of the state
                 // array is 0 after initial seeding.
                 if (e == 0)
-                    e = (i + 1);
+                    e = (cast(uint)i + 1);
```
ZombineDev: Do:

```suggestion
                    e = (cast(uint) i + 1);
```
to fix the style check.

n8sh:
Since the array has a known length that fits in a `uint` why is it an error to index it with a `uint`?

thewilsonator:
BigEndian and truncation is the real problem.

ibuclaw:
See bugzilla issue 16976, which only deprecated _some_ cases of `foreach(int i, val v; ...)`, leaving open other cases that use the `_aApply[wcd]` path, which turns the body into a delegate, and passes the index as a `size_t*`.

ibuclaw:
I'll do a test tweaking the compiler-side test.  Maybe we can get away with some cases to still be permissible, where it's known that no unsafe truncation happens.

ibuclaw:
I think we're good without this after modifying the compiler. :+1:

PR #7003, opened 2019-05-14
Added description to EnumMember documentation to get name of value

Zevenberge:
Explicitly stated that the enum name can be retrieved with __traits. There was a unittest for this, but it was not stated in the documentation.
I've added an example to generate a switch-statement with static foreach. Furthermore, I removed the generate example from the documentation, as (in my eyes) it is basically a duplicate of the square root example.

Suggestions for improvement are appreciated.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -4072,6 +4072,9 @@ Params:
 Returns:
     Static tuple composed of the members of the enumerated type `E`.
     The members are arranged in the same order as declared in `E`.
+    The name of the enum can be found by querying the compiler for the
+    name of the identifier, i.e. `__traits(identifier, EnumMembers!MyEnum[i])`.
+    For enumerations with unique values, $(REF to, std,conv) can also be used.
 
 Note:
     An enum can have multiple members which have the same value. If you want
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -4072,6 +4072,8 @@ Params:
 Returns:
     Static tuple composed of the members of the enumerated type `E`.
     The members are arranged in the same order as declared in `E`.
+    The name of the enum can be found by querying the compiler for the
```
wilzbach: You already have the name, otherwise `case` wouldn't work.

thewilsonator:
```

grep -nr '[[:blank:]]$' etc std ; test $? -eq 1
--
  | Enforce whitespace before opening parenthesis
  | grep -nrE "\<(for\|foreach\|foreach_reverse\|if\|while\|switch\|catch\|version)\(" $(find etc std -name '*.d') ; test $? -eq 1
  | std/traits.d:4203:    s: final switch(var)
  | std/traits.d:4206:        static foreach(i; 0 .. members.length)
  | posix.mak:562: recipe for target 'style_lint' failed
```

std/traits.d
Zevenberge: As far as I know, we don't: https://run.dlang.io/is/FGGmOH
`EnumMembers` returns an AliasSeq of values. E.g. `members[i].stringof` returns `cast(FooEnum)0`. I might be missing something as I'm no thorough expert.

On the other hand, if we can make it such that we can return the name more easily, then that has my preference as well. But I think that will be a breaking change, as members with the same value are no longer unique.

Zevenberge: Apparently std.conv: to!string also works to retrieve the member as a string in this example. (It makes me wonder why there is a unittest for the identifier, but that is out of scope for improving the documentation.)

Zevenberge:
```d
import std.conv;
enum Enum {foo = 0, bar = 1, baz = 0}
static foreach(i, member; EnumMembers!Enum)
{
	pragma(msg, to!string(member) ~ " " ~ __traits(identifier, EnumMembers!Enum[i]));
}
```
prints
```
foo foo
bar bar
foo baz
```

Zevenberge:
Alright, this documentation should fit the implementation (and clarifies the difference between __traits and to!string).

PR #7006, opened 2019-05-15
Fix Issue 19838 - RefCounted fails to instantiate due to pureness of …

marcioapm:
…moveEmplace

### DIFF
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -1375,7 +1375,7 @@ private T moveImpl(T)(ref T source)
  *   source = value to be moved into target
  *   target = uninitialized value to be filled by source
  */
-void moveEmplace(T)(ref T source, ref T target) @system
+void moveEmplace(T)(ref T source, ref T target) pure @system
 {
     import core.stdc.string : memcpy, memset;
     import std.traits : hasAliasing, hasElaborateAssign,


wilzbach:
Please add a test case to avoid future regressions.

marcioapm:
I can't find a reduced test case that reproduces the behavior on our codebase.

Somehow it seems that inferring pureness doesn't work in this particular case, which points to a deeper issue.

This PR will alleviate the situation introduced by manually annotating `RefCountedStore.move()` as `pure` in https://github.com/dlang/phobos/commit/c4324f405c6aea75c9c074c42a21b42a2a361d76#diff-4e008aedb3026d4a84f58323e53bf017

atilaneves:
Shouldn't this be a dmd bug?

marcioapm:
Ping! Any comments on this and if this, or a better fix are going to be merged?

thewilsonator:
Please add the test case for the issue.

marcioapm:
I did try to find a reduced test case but wasn't able to. I suppose once the real bug is found and understood better it will be easier.

Despite not being a real fix, it will at least hopefully revert things to how they used to be before the bug was introduced.

This bug is preventing us from upgrading without a big code refactor so we are stuck in 2.083.1

thewilsonator:
OK.

thewilsonator:
Also please target stable.

marcioapm:
Done - Thanks!

PR #7007, opened 2019-05-15
Add `powmod()` for BigInt operands.

shove70:
Add fast power modulus `powmod()` for BigInt operands.


shove70:
@thewilsonator, I'm sorry for the trouble I've caused you.

PR #7009, opened 2019-05-16
Faster isNaN test

andralex:
To test that at least one bit in the mantissa is set, it's faster to just test for inequality instead of one more masking and an AND.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -6360,41 +6360,55 @@ version (D_HardFloat) @safe unittest // rounding
 bool isNaN(X)(X x) @nogc @trusted pure nothrow
 if (isFloatingPoint!(X))
 {
-    alias F = floatTraits!(X);
-    static if (F.realFormat == RealFormat.ieeeSingle)
+    version (all)
     {
-        const uint p = *cast(uint *)&x;
-        // Sign bit (MSB) is irrelevant so mask it out.
-        // Next 8 bits should be all set.
-        // At least one bit among the least significant 23 bits should be set.
-        return (p & 0x7FFF_FFFF) > 0x7F80_0000;
-    }
-    else static if (F.realFormat == RealFormat.ieeeDouble)
-    {
-        const ulong  p = *cast(ulong *)&x;
-        // Sign bit (MSB) is irrelevant so mask it out.
-        // Next 11 bits should be all set.
-        // At least one bit among the least significant 52 bits should be set.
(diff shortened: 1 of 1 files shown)


thewilsonator:
Numbers please.

andralex:
strictly fewer operations: https://godbolt.org/z/v8LBdv

thewilsonator:
Dunno what circle is doing.

GooberMan:
Just jumping in to say that the code as it reads would faster. But oof. That disassembly. This is exactly where you'd deploy some intrinsics to avoid the integer register shuffle.

But yeah, it's totally legit platform-agnostic code.

thewilsonator:
Presumably it would get inlined at `-O3` and that would no longer be a problem.

GooberMan:
@thewilsonator Negative. The problem I raise isn't an inlining problem, the problem is that the value gets moved out of the floating point registers and in to the general purpose registers. An intrinsic would load the mask directly in to the XMM registers and perform the comparison there. The result would then be pulled in to a general-purpose register if you're going to use it.

Should be faster in a good number of situations. But more importantly, it's also a solution that needs to be tailored to every platform. x86 would need a different approach to the x64 disassembly above, as would ARM/NEON etc.

thewilsonator:
Ah, I see.

andralex:
@GooberMan Hmmm... then I wonder, isn't it faster to just rely on `x != x`? It compiles down to ucomiss, which I assume is done quickly in software. See https://godbolt.org/z/_IN7hc

andralex:
Apparently ucomiss takes 1 cycle and can be run in parallel with other instructions: https://software.intel.com/en-us/forums/intel-isa-extensions/topic/293283. So I guess we should rely on it at least on Intel?

GooberMan:
@andralex Ah ha, yeah, that would probably be best overall.

My brain is telling me that I had weird issues with NaNs on the AMD Jaguar processors, but it's been years since I've had to deal with that and I don't have access to that hardware or the problem code at the moment. Could very well have just been a code gen issue as well with much older compilers. So just out of paranoia my opinion would be "Looks, but if someone gets an error ask them to run the unit tests and give a hardware report."

andralex:
@GooberMan Switched to the simple test, kept existing code disabled and commented.

andralex:
Was approved, will merge.

PR #7010, opened 2019-05-16
Added an enumerated type of ASCII control character.

dukc:


std/ascii.d
```diff
@@ -102,6 +103,70 @@ enum LetterCase : bool
     assert(sha1HMAC == "49f2073c7bf58577e8c9ae59fe8cfd37c9ab94e5");
 }
 
+/++
+    All the control characters in ASCII table
++/
+enum ControlChar : char
+{
+    NUL = '\x00', /// Null
```
jmdavis: The style guide require that enum members be camelCase. So, none of these should start with an uppercase letter, and they should probably all just be all lowercase. Also, it would be more in line with our naming conventions if actual words like newline or linefeed were used rather than abbreviations, though at least some of the names here are well-known in the context of control characters.

std/ascii.d
```diff
@@ -102,6 +103,70 @@ enum LetterCase : bool
...
+    static assert(ControlChar.ACK.sizeof == 1);
+
+    static assert(ControlChar.NUL == '\0');
+    static assert(ControlChar.BEL == '\a');
+    static assert(ControlChar.BS  == '\b');
+    static assert(ControlChar.FF  == '\f');
+    static assert(ControlChar.LF  == '\n');
+    static assert(ControlChar.CR  == '\r');
+    static assert(ControlChar.TAB == '\t');
+    static assert(ControlChar.VT  == '\v');
```
jmdavis: Why aren't you testing all of them?

dukc: I also have a feeling that there's a mention of more escape sequences existing, but I found nothing else here: https://dlang.org/spec/lex.html#characterliteral

Where can I find the rest, if there are?

std/ascii.d
dukc: Yeah, camelCase, I will change that.

But about the names and abbreviations, I'm not so sure. The abbreviations are standartized, I think, but the descriptions are apparently not. Also, it would seen kind of stupid to comment a whole name with an abbreviation.

std/ascii.d
jmdavis: The control characters are mostly all in a row. I'd probably just loop over them and verify that they equal 0 up through the last one in a row and then test any individually that I'd have to. I probably wouldn't even bother testing the escape sequences. I suppose that testing them in addition to the loop helps ensure that you didn't screw up the loop, but a simple loop with `EnumMembers` should make it easy to test almost all of the values.

dukc: Done. I left the escape char literal tests in place, though.

atilaneves:
What's the main use case?

dukc:
> 
> 
> What's the main use case?

See the latter of two unittests added. This markup is benefitical, because:

1: At least I am more likely to remember meaning of `ControlChar.ack` than `'\x06'`
2: If you're using an IDE, you can hit goto definition on ControlChar quicker than you can switch to browser and search for ASCII table. With luck, a tooltip might even directly show you the meaning of control char abbreviation when pointed on.

std/ascii.d
```diff
@@ -102,6 +103,75 @@ enum LetterCase : bool
     assert(sha1HMAC == "49f2073c7bf58577e8c9ae59fe8cfd37c9ab94e5");
 }
 
+/++
+    All the control characters in ASCII table
++/
```
chloekek: Are the names of these enumerations standardized? Where do they come from?

You may want to add a reference to this documentation comment.

dukc: Ref added. as for the abbreviation names, I think they are standartized because they are the same no matter where they are listed, but their descriptions are not. For example, horizonal tab vs. horizonal tabulation, line feed vs. newline.

On the other hand, if we decide to use full names instead, I could just define two names with same values in cases like that latter example.

lesderid: > As for the abbreviation names, I think they are standartized because they are the same no matter where they are listed, but their descriptions are not. For example, horizonal tab vs. horizonal tabulation, line feed vs. newline.

A little old (only about 5 decades), but both the abbreviations and the full names are standardised here: https://tools.ietf.org/html/rfc20#section-4.1.

PR #7012, opened 2019-05-17
[DEBUG CI] [DON'T MERGE] Add a test that should fail to compile on the auto-tester

ZombineDev:
See https://github.com/dlang/phobos/pull/7011.


thewilsonator:
Hmm, it passes.

ZombineDev:
`Darwin_64_32` always passes as we have retired 32-bit support for macOS.

PR #7015, opened 2019-05-19
add isLvalue() to prepare for rvalue ref

WalterBright:
Necessary because with the change an rvalue can be an argument to a ref parameter.


WalterBright:
Uncovered by https://github.com/dlang/dmd/pull/9817

PR #7019, opened 2019-05-20
Slightly improve file iteration for Windows.

look-at-me:
Needs https://github.com/dlang/druntime/pull/2612

Ideal this should also be @nogc but for another time...

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -4562,7 +4562,8 @@ private struct DirIteratorImpl
 
             static auto trustedFindFirstFileW(typeof(searchPattern) pattern, WIN32_FIND_DATAW* findinfo) @trusted
             {
-                return FindFirstFileW(pattern.tempCString!FSChar(), findinfo);
+                return FindFirstFileExW(pattern.tempCString!FSChar(), FINDEX_INFO_LEVELS.FindExInfoBasic, findinfo,
+                    FINDEX_SEARCH_OPS.FindExSearchNameMatch, null, 0);
             }
 
             HANDLE h = trustedFindFirstFileW(searchPattern, &_findinfo);


CyberShadow:
Any benchmarks?

How will this work on Windows versions before 7?

std/file.d
```diff
@@ -4562,7 +4562,8 @@ private struct DirIteratorImpl
 
             static auto trustedFindFirstFileW(typeof(searchPattern) pattern, WIN32_FIND_DATAW* findinfo) @trusted
             {
-                return FindFirstFileW(pattern.tempCString!FSChar(), findinfo);
+                return FindFirstFileExW(pattern.tempCString!FSChar(), FINDEX_INFO_LEVELS.FindExInfoBasic, findinfo
```
CyberShadow: There is a comma missing here.

Perhaps try the patch on your PC first (also to get some benchmark numbers).

look-at-me: copy-paste error, it ran ~25% faster

CyberShadow: 25% is a *LOT*. Was it with a warm/cold cache in both runs?

look-at-me: You can run test for yourself, those are the results I observed.

CyberShadow: I do not have access to a Windows machine at the moment.

std/file.d
```diff
@@ -4562,7 +4562,8 @@ private struct DirIteratorImpl
 
             static auto trustedFindFirstFileW(typeof(searchPattern) pattern, WIN32_FIND_DATAW* findinfo) @trusted
             {
-                return FindFirstFileW(pattern.tempCString!FSChar(), findinfo);
+                return FindFirstFileExW(pattern.tempCString!FSChar(), FINDEX_INFO_LEVELS.FindExInfoBasic, findinfo,
+                    FINDEX_SEARCH_OPS.FindExSearchNameMatch, null, 0);
```
n8sh: To keep working under Windows XP how about:
```d
// FindExInfoBasic is not supported until Windows Server 2008 R2 and Windows 7.
static if (is(typeof(FINDEX_INFO_LEVELS.FindExInfoBasic)))
    return FindFirstFileExW(pattern.tempCString!FSChar(), FINDEX_INFO_LEVELS.FindExInfoBasic, findinfo,
        FINDEX_SEARCH_OPS.FindExSearchNameMatch, null, 0);
else
    return FindFirstFileW(pattern.tempCString!FSChar(), findinfo);
```

wilzbach: We don't support Windows XP Abbie, so that shouldn't be a concern?

CyberShadow: - I think we still care about Vista, so that question stands
- This approach won't work, it will only allow compiling Phobos against Druntime versions that do not have a `FindExInfoBasic` definition. Instead, the function would need to be loaded dynamically using `GetProcAddress` - see `RegDeleteKeyExW` in `std.internal.windows.advapi32` or `getaddrinfo` in `std.socket` for examples within Phobos.

n8sh: [I believe this is the most recent word.](https://forum.dlang.org/post/ktfgps$2ghh$1@digitalmars.com)
>On 8/1/13 10:41 PM, Walter Bright wrote:
>As a practical matter, we do not test on XP, and so it is disingenuous to claim official support for it.
>
>I suggest saying we "officially" support Win7 or later, and unofficially support versions back to XP. Unofficially means there's code there to support it, and we accept pull requests to fix bugs in it, we will try not to break it, but we don't test or guarantee that it'll work.

[At the time that was written Windows XP had an estimated 37% desktop market share. Since then the estimate has fallen:](https://en.wikipedia.org/wiki/Usage_share_of_operating_systems#Desktop_and_laptop_computers)

>Web analysis shows significant variation in different parts of the world. For example, in North America usage of Windows XP has dropped to 2.06%, but in Africa it is still at 11.02% and Asia at 6.63% (even higher in China, but down to 18.21%, tied with Win10 at second) while going down, and has been overtaken by Windows 10.

PR #7020, opened 2019-05-20
fix rvalue test for Final

WalterBright:
Relying on rvalues not converting to ref param is no longer viable.

### DIFF
--- a/std/experimental/typecons.d
+++ b/std/experimental/typecons.d
@@ -978,9 +978,6 @@ pure nothrow @safe unittest
     Final!A a = new A;
     static assert(!__traits(compiles, a = new A));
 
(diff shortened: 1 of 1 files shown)


std/experimental/typecons.d
```diff
@@ -978,9 +978,6 @@ pure nothrow @safe unittest
     Final!A a = new A;
     static assert(!__traits(compiles, a = new A));
 
-    static void foo(ref A a) pure nothrow @safe @nogc {}
-    static assert(!__traits(compiles, foo(a)));
```
WalterBright: with rvalue conversion, this test is no longer valid

WalterBright:
Blocking https://github.com/dlang/dmd/pull/9817

PR #7021, opened 2019-05-22
Explicitly link to pthread

marler8997:
Currently, phobos is using `-defaultlib=` to build itself, but is implicitly relying on its library dependencies to be added by dmd.

This change makes the phobos library dependencies explicit rather than relying on its dependencies to be hardcoded into the compiler, which in turn forces all D programs to link to those libraries as well.

This change is required to allow https://github.com/dlang/dmd/pull/9831 to be merged, which removes these implicit libraries from being linked to ALL D programs that use DMD to link.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -112,6 +112,7 @@ OUTFILEFLAG = -o
 NODEFAULTLIB=-defaultlib= -debuglib=
 ifeq (,$(findstring win,$(OS)))
 	CFLAGS=$(MODEL_FLAG) -fPIC -DHAVE_UNISTD_H
+	NODEFAULTLIB += -L-lpthread
 	ifeq ($(BUILD),debug)
 		CFLAGS += -g
 	else


thewilsonator:
Ping me when this is good.

marler8997:
There may be more libraries that need to be added but I don't know of an easy way to test that without using the CI builders...and I need both this PR and the DMD PR to be tested at the same time.

I know for sure we need `-lpthread`.  We could integrate that one first and see if that's all we need, and make more PRs if we find any other libraries that need to be added.

marler8997:
Closing and re-opening to trigger autotester retry. It looked like it failed on windows because of a race condition between running executables and then removing them afterwards.

marler8997:
This one should be ready, if we end up needing to add more libraries then I'll make another PR.

posix.mak
```diff
@@ -112,6 +112,7 @@ OUTFILEFLAG = -o
 NODEFAULTLIB=-defaultlib= -debuglib=
 ifeq (,$(findstring win,$(OS)))
 	CFLAGS=$(MODEL_FLAG) -fPIC -DHAVE_UNISTD_H
+        NODEFAULTLIB += -L-lpthread
```
wilzbach: Makefile use tabs :man_shrugging:

marler8997: good catch

marler8997:
Been a few days, any objections to merge this now?  Should we apply that label "merge in X days if no objections?"  Do we still have that label?

PR #7022, opened 2019-05-23
Change version(unittest) code to use accessors to wrap the static initializers

schveiguy:
This prevents the compiler from executing the initializers using CTFE during importing, when using unittest.

This works around issue [19874](https://issues.dlang.org/show_bug.cgi?id=19874)

ping @jmdavis

### DIFF
--- a/std/datetime/date.d
+++ b/std/datetime/date.d
@@ -10437,9 +10437,12 @@ if (isSomeString!T)
 }
 
 
+// NOTE: all the non-simple array literals are wrapped in functions, because
+// otherwise importing causes re-evaluation of the static initializers using
+// CTFE with unittests enabled
 version (unittest)
 {
-private:
+private @safe:
     // All of these helper arrays are sorted in ascending order.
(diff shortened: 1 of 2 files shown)


jmdavis:
@thewilsonator Will you please not merge pull requests without giving anyone else a chance to even look at them? Aren't they supposed to be up for something like 48 hours before merging them under normal circumstances? In particular, personally, I'd like the chance to look at PRs for std.datetime and core.time before they get merged. This got merged less than an hour after it was posted. There's nothing obviously wrong with it (though I'm a big fan of `@safe` being used as a label), but if you're merging stuff so quickly without anyone else even getting the chance to look at it, problems are going to make it in sometimes. Didn't we used to have the requirement that two people review a merge and accept it first? I get that you're trying to keep things moving along, but please give other people the chance to review code before merging it. It's one thing to push something in that's been sitting around for a few days with no comment, and you think that it's fine. It's a completely different thing to merge stuff in within hours of it being created.

PR #7024, opened 2019-05-23
Use selective top-level module imports in std.algorithm

wilzbach:
We can finally do this (: 

See also: https://dlang.org/changelog/pending.html#remove_visibility

PR #7025, opened 2019-05-23
Use selective top-level module imports in std.{base64,csv,datetime}

wilzbach:
Doing this in multiple parts.

See also: https://github.com/dlang/phobos/pull/7024

### DIFF
--- a/std/base64.d
+++ b/std/base64.d
@@ -57,9 +57,10 @@
  */
 module std.base64;
 
-import std.exception;  // enforce
-import std.range.primitives;      // isInputRange, isOutputRange, isForwardRange, ElementType, hasLength
-import std.traits;     // isArray
+import std.exception : enforce;
+import std.range.primitives : empty, front, isInputRange, isOutputRange,
+    isForwardRange, ElementType, hasLength, popFront, put, save;
+import std.traits : isArray;
 
 // Make sure module header code examples work correctly.
 @safe unittest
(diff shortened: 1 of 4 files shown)


std/base64.d
```diff
@@ -1764,6 +1765,7 @@ class Base64Exception : Exception
     import std.algorithm.comparison : equal;
     import std.algorithm.sorting : sort;
     import std.conv;
+    import std.exception : assertThrown;
```
thewilsonator: the comment said this was for enforce...

thewilsonator: Oh this is in a unittest. whoops.

wilzbach: Yeah, but the comment is old and wasn't enforced by the compiler.

thewilsonator:
```
std/datetime/timezone.d(1409): Error: template instance `isOutputRange!(W, char)` template `isOutputRange` is not defined
std/datetime/timezone.d(1403): Error: none of the overloads of `toISOString` are callable using argument types `(Appender!string, Duration)`, candidates are:
std/datetime/timezone.d(1398):        `std.datetime.timezone.SimpleTimeZone.toISOString(Duration utcOffset)`
std/datetime/timezone.d(1408):        `std.datetime.timezone.SimpleTimeZone.toISOString(W)(ref W writer, Duration utcOffset) if (isOutputRange!(W, char))`
```

PR #7028, opened 2019-05-23
Use selective top-level module imports in std.{stdio,string}

wilzbach:
Doing this in multiple parts.

See also: #7024

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -13,14 +13,14 @@ Authors:   $(HTTP digitalmars.com, Walter Bright),
  */
 module std.stdio;
 
-import core.stdc.stddef; // wchar_t
+import core.stdc.stddef : wchar_t;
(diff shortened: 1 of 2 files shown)


wilzbach:
IZ fails because it does the following:

```d
static if (!__traits(hasMember, range, "front"))
{
    import std.range.primitives: front, empty, popFront;
}
```

https://github.com/Basile-z/iz/blob/f9cfb7e8ccbe1907b8b07c89b8a2b94704865a2d/import/iz/sugar.d#L1205-L1208

CC @Basile-z

A quick check on run.dlang.io shows that the import holes have been fixed and yield "not found" properly: https://run.dlang.io/is/KkrN1Q :/

Basile-z:
Buildkite fixed, you can merge.

PR #7037, opened 2019-05-24
assert messages for std.getopt

burner:


burner:
@thewilsonator thanks

PR #7038, opened 2019-05-24
std.csv assert messages

burner:


burner:
@thewilsonator thank you

std/csv.d
```diff
@@ -989,7 +989,7 @@ public:
      */
     @property auto front()
     {
-        assert(!empty);
+        assert(!empty, "An empty range has no front element");
```
jondegenhardt: More consistent with the messages in range code (std/range/package.d) would be:
```
"Attempting to fetch the front of an empty CsvReader"
```

std/csv.d
```diff
@@ -1224,7 +1224,7 @@ public:
      */
     @property Contents front() @safe pure
     {
-        assert(!empty);
+        assert(!empty, "An empty range has no front element");
```
jondegenhardt: More consistent with the messages in range code (std/range/package.d) would be:
```
"Attempting to fetch the front of an empty CsvRecord"
```

std/csv.d
```diff
@@ -1425,7 +1425,7 @@ if (isSomeChar!Separator && isInputRange!Range
 
     while (!input.empty)
     {
-        assert(!(quoted && escQuote));
+        assert(!(quoted && escQuote), "Invalid quotation state");
```
jondegenhardt: Consider adding the function to the message in some fashion. This is often helpful when debugging. eg.
```
"Invalid quotation state in csvNextToken"
```

jondegenhardt: Just realized this is part of a much larger effort to add messages to all asserts. (Fantastic!) So, ignore my comment about this specific case and instead be consistent across all cases.

FWIW, I've benefitted when the assert messages include a reference to the code construct that is asserting. This has even been helpful to tracking down bugs in Phobos. But most important is that there is an assert message.

burner:
@jondegenhardt thank you for the review

jondegenhardt:
LGTM. Thanks for making the changes.

burner:
@jondegenhardt that's what review is for

thewilsonator:
@burner please do remind me about these when I forget, (unless you've forgotten too...).

burner:
@thewilsonator sure thing, thanks

PR #7039, opened 2019-05-25
Add messages to some asserts in std/socket.d

Flying-Toast:
Work on project [#2](https://github.com/dlang/phobos/projects/2)

### DIFF
--- a/std/socket.d
+++ b/std/socket.d
@@ -117,7 +117,7 @@ else version (Posix)
 }
 else
 {
-    static assert(0);     // No socket support yet.
+    static assert(0, "No socket support for this platform yet.");
 }
 
 version (unittest)
@@ -268,7 +268,7 @@ bool wouldHaveBlocked() nothrow @nogc
     else version (Posix)
         return _lasterr() == EAGAIN;
     else
-        static assert(0);
+        static assert(0, "No socket support for this platform yet.");
 }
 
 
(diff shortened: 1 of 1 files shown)


Flying-Toast:
I'm not sure what the message should be for this assert:
https://github.com/dlang/phobos/blob/64eac952e385696880cbf6eac26516115fc60eef/std/socket.d#L970

thewilsonator:
`Failed to retrieve AddressInfo`? Not surer why thats an assert rather that an ErrnoException (or something a bit more descriptive). 

cc @CyberShadow `git blame` blames you as author of the assert in question. Any insight?

lesderid:
AFAICT that can never happen. Both POSIX[1] and Windows[2] will always return "one or more" results when `getaddrinfo` returns 0, so that should already be handled by: https://github.com/dlang/phobos/blob/64eac952e385696880cbf6eac26516115fc60eef/std/socket.d#L956

[1] https://pubs.opengroup.org/onlinepubs/9699919799/functions/getaddrinfo.html
[2] https://docs.microsoft.com/en-us/windows/desktop/api/ws2tcpip/nf-ws2tcpip-getaddrinfo

Flying-Toast:
I think that https://github.com/dlang/phobos/blob/64eac952e385696880cbf6eac26516115fc60eef/std/socket.d#L271 could also be removed, because it is already covered by https://github.com/dlang/phobos/blob/64eac952e385696880cbf6eac26516115fc60eef/std/socket.d#L120?

lesderid:
I think it's better to keep that. If the `static assert(0)` is removed, what should it return?

PR #7040, opened 2019-05-25
(s)brk support has been removed from DragonFlyBSD

dkgroot:
brk support has been removed from DragonFlyBSD
sbrk has been declared deprecated
mmap() is suggested as a replacement for sbrk()


thewilsonator:
Should also follow up with a druntime PR to remove them from `core.sys.whatever`.

thewilsonator:
Vibe.d failure is unrelated:

```
Fetching memutils 0.4.13 (getting selected version)...
  | Failed to download package memutils from https://code.dlang.org/packages/memutils/0.4.13.zip
```

dkgroot:
@thewilsonator : Thanks for merging

> Should also follow up with a druntime PR to remove them from `core.sys.whatever`.

I checked druntime, but there is not reference to brk or sbrk in there. It would have been nice if there were a standard interface in druntime, but as brk and sbrk are mostly legacy, there is not need to add them.

thewilsonator:
Hmm, thanks for checking.

PR #7041, opened 2019-05-26
Ternary operator no longer promotes characters to integers

SSoulaimane:
blocks and requires DMD https://github.com/dlang/dmd/pull/9889.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -1274,8 +1274,8 @@ pure @safe nothrow unittest
     auto s3 = "string three".byCodeUnit;
     static assert(is(typeof(s3.front) == immutable char));
     auto r2 = s1.chain(s3);
-    // type is promoted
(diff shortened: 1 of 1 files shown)


std/range/package.d
```diff
@@ -1274,8 +1274,8 @@ pure @safe nothrow unittest
     auto s3 = "string three".byCodeUnit;
     static assert(is(typeof(s3.front) == immutable char));
     auto r2 = s1.chain(s3);
-    // type is promoted
-    static assert(is(typeof(r2.front) == uint));
```
thewilsonator: If there is a two way requirement, I think its better to temporarily comment this out, then fix DMD, then add this back.

SSoulaimane: Ok.

SSoulaimane: Sequel PR #7043.

PR #7043, opened 2019-05-26
Ternary operator no longer promotes characters to integers [Part 2]

SSoulaimane:
Re-enable test from #7041.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -1274,8 +1274,8 @@ pure @safe nothrow unittest
     auto s3 = "string three".byCodeUnit;
     static assert(is(typeof(s3.front) == immutable char));
     auto r2 = s1.chain(s3);
-    //// chaining ranges of mixed character types gives `dchar`
-    //static assert(is(typeof(r2.front) == dchar));
+    // chaining ranges of mixed character types gives `dchar`
+    static assert(is(typeof(r2.front) == dchar));
 
     // use byChar on character ranges to correctly convert them to UTF-8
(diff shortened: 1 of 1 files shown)


std/range/package.d
```diff
@@ -1274,8 +1274,8 @@ pure @safe nothrow unittest
     auto s3 = "string three".byCodeUnit;
     static assert(is(typeof(s3.front) == immutable char));
     auto r2 = s1.chain(s3);
-    //// chaining ranges of mixed character types gives `dchar`
-    //static assert(is(typeof(r2.front) == dchar));
+    // chaining ranges of mixed character types gives `dchar`
+    static assert(is(typeof(r2.front) == dchar));
```
ZombineDev: @SSoulaimane @thewilsonator wasn't there consensus that the common type of any two types `T` and `U` is `T` if `U` is convertible to `T` and `U` if `T` is convertible to `U`? Or in other words, that the common type of `char` and `wchar` is `wchar`?

thewilsonator:
```
byte b;
short s;
auto u = 0 ? b : s;
pragma(msg, typeof(u));
```
prints int.

SSoulaimane: This was made withing the current practice of preemptive promotion even for ternary expressions. If/when that's going to change, it will affect more than just `char`/`wchar` but also the rest of the basic types.

SSoulaimane: Here is a DMD PR for that https://github.com/dlang/dmd/pull/9943/.

PR #7045, opened 2019-05-27
[TRIVIAL] Make link clickable and safe

andralex:
BTW we should replace all of the HTTP we can with HTTPS.


andralex:
Wait a sec, needs two arguments.

thewilsonator:
No need for both commits, using auto-merge-squash.

PR #7047, opened 2019-05-28
Make nullable.get `@trusted`

thewilsonator:
required for dmd#9909

cc @RazvanN7

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2969,7 +2969,11 @@ Returns:
     {
         enum message = "Called `get' on null Nullable!" ~ T.stringof ~ ".";
         assert(!isNull, message);
-        return _value.payload;
+        inout(T)* get2() inout @trusted
+        {
+            return &_value.payload;
+        }
+        return *get2();
     }
 
     /// ditto


RazvanN7:
Looks good to me. Although, a union with a member should be treated as a no op.

thewilsonator:
You mean a union with a single member? Also, how do you not have commit access to phobos?

wilzbach:
How is this @safe in release mode and the assert gone?
`get` was a huge mistake already and encourages so many anti-patterns..

RazvanN7:
Yes, I mean a union with a single member.

> Also, how do you not have commit access to phobos?

Well, I rarely make PRs to phobos nowadays, so maybe it is considered that I don't have sufficient experience :-?

thewilsonator:
> How is this @safe in release mode and the assert gone?

Urgh, its not.  Thanks for catching that, fixed.

> get was a huge mistake already and encourages so many anti-patterns..

True dat!

RazvanN7:
I think that the lambda needs to explicitly have all the attributes that the enclosing function has.

thewilsonator:
> Although, a union with a member should be treated as a no op.

https://issues.dlang.org/show_bug.cgi?id=19908

thewilsonator:
```
../../phobos/std/typecons.d(2972): Error: `inout` on `return` means `inout` must be on a parameter as well for `pure nothrow @nogc @trusted inout(char)()`
```

Huh? does that mean I need to do `return () @trusted inout{ return _value.payload; } ();`?

thewilsonator:
Yay, finally its green. That was a lot harder than it needed to be.

std/typecons.d
```diff
@@ -2969,7 +2969,11 @@ Returns:
     {
         enum message = "Called `get' on null Nullable!" ~ T.stringof ~ ".";
         assert(!isNull, message);
-        return _value.payload;
+        inout(T)* get2() inout @trusted pure nothrow
```
atilaneves: Except for `@trusted`, everything else should be inferred, no?

thewilsonator: inout is necessary I think because I got https://github.com/dlang/phobos/pull/7047#issuecomment-496536998 before with one of the previous tries. the other attributes, not sure.

atilaneves: Even `inout` probably isn't:

```d
auto get2() @trusted {
    return &_value.payload;
}
```

thewilsonator: It is, same error.

Geod24:
I fail to see how this is required for dlang/dmd#9909
The PR you link clearly mentions anonymous unions, while `_value` is [not anonymous](https://github.com/dlang/phobos/blob/49e624fa2fc18a1ce60a07bb3390775aa62c80e8/std/typecons.d#L2646-L2649)

thewilsonator:
Good point.

RazvanN7:
It is not anonymous, but I would argue that the problem highlighted is real. What #9909 does is to eagerly set the overlapped field for any union members. Before that PR this was done somewhere after semantic1 was over. It was a mere collateral effect the compiler did not complain before, so I would argue that this PR still needs to be merged.

Geod24:
@RazvanN7 : Let's continue this discussion on the DMD PR.

PR #7049, opened 2019-05-29
Provide utilities to auto implement opEquals, opCmp and toHash

edi33416:
This still needs a changelog entry and ddoc'ed style unittests.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -5574,6 +5574,685 @@ private
     extern(C) pure nothrow Object typecons_d_toObject(void* p);
 }
 
+// Begin addition
+
+private template ContravariantRhsT(Rhs)
+{
+    static if (is (Rhs == class))
+    {
+        alias ContravariantRhsT = Object;
+    }
+    else static if (is (Rhs == struct))
+    {
+        alias ContravariantRhsT = auto ref const Unqual!Rhs;
+    }
+    else
+    {
+        static assert (0, "Unsupported type " ~ Rhs.stringof ~ ". Rhs must be a `class` or `struct`");
+    }
+}
+
+mixin template ImplementOrdered(M...)
+{
+    //isAggregateType ? (support interface and union?)
+    static assert(is(typeof(this) == struct) || is(typeof(this) == class));
+
+    static if (is(typeof(this) == class))
+    {
+        override
+        int opCmp(ContravariantRhsT!(typeof(this)) rhso)
+        {
+            int compare(U1, U2)(U1 u1, U2 u2)
+            {
+                int r = 0;
+                static if (__traits(compiles, __cmp(u1, u2)))
+                {
+                    r = __cmp(u1, u2);
+                }
+                else static if (__traits(compiles, u1.opCmp(u2)))
+                {
+                    r = u1.opCmp(u2);
+                }
+                else static if (__traits(compiles, u1.opCmp((() @trusted => cast(Unqual!U2) u2)())) &&
+                                is(U2 == class))
+                {
+                    // Required because of `object.opCmp(Object)` doesn't accept qualified types.
+                    r = u1.opCmp((() @trusted => cast(Unqual!U2) u2)());
+                }
+                else static if (__traits(compiles, u1 < u2))
+                {
+                    if (u1 != u2)
+                        r = u1 < u2 ? -1 : 1;
+                }
+                else
+                {
+                    // TODO: fix this legacy bad behavior, see
+                    // https://issues.dlang.org/show_bug.cgi?id=17244
+                    static assert(is(U1 == U2), "Internal error.");
+                    import core.stdc.string : memcmp;
+                    r = (() @trusted => memcmp(&u1, &u2, U1.sizeof))();
+                }
+                return r;
+            }
+
+            alias T = typeof(this);
+            static if (is(T == class))
+            {
+                auto rhs = (() @trusted => cast(T) rhso)();
+                if (rhs is null) return 1;
+            }
+            else
+            {
+                alias rhs = rhso;
+            }
+
+            static if (M.length == 0)
+            {
+                alias U = GetAllFields!T;
+            }
+            else
+            {
+                alias U = M;
+            }
+            enum len = U.length;
+
+            int r = 0;
+            import std.range : iota;
+            //static foreach (i; 0 .. len)
+            static foreach (i; iota(0, len))
+            {{
+
+                static if (is(typeof(U[i]) == string))
+                {
+                    enum fieldName = U[i];
+                    // If assert fails, propagate compiler error
+                    static assert(is(typeof(__traits(getMember, this, fieldName))),
+                                  typeof(__traits(getMember, this, fieldName)));
+
+                    static if (((i + 1) < len) && !is(typeof(U[i + 1]) == string))
+                    {
+                        // If the next element is not a string, then it must be the comparator
+                        alias compareFun = U[i + 1];
+                    }
+                    else
+                    {
+                        alias compareFun = compare;
+                    }
+
+                    alias thisMember = __traits(getMember, this, fieldName);
+                    alias rhsMember = __traits(getMember, rhs, fieldName);
+
+                    // If we can't call the function with the args, throw the compiler error
+                    static assert(is(typeof({compareFun(thisMember, rhsMember);})),
+                            typeof({compareFun(thisMember, rhsMember);}));
+
+                    r = compareFun(__traits(getMember, this, fieldName), __traits(getMember, rhs, fieldName));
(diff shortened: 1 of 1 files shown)


n8sh:
Out of curiosity, what is the vision for how this will relate to [DMD clone.d](https://github.com/dlang/dmd/blob/67e8328f25d7c9733f83ad4c33226d2cb2c3a63b/src/dmd/clone.d)'s `buildXopEquals`/`buildXopCmp`/`buildXtoHash`?

std/typecons.d
```diff
@@ -5394,6 +5394,582 @@ private
...
+
+private template ContravariantRhsT(Rhs)
+{
+    static if (is (Rhs == class))
+    {
+        alias ContravariantRhsT = const Object;
+    }
+    else static if (is (Rhs == struct))
+    {
+        alias ContravariantRhsT = auto ref const Unqual!Rhs;
```
ZombineDev: Unfortunately, I don't think you can use storage classes like `auto ref`, `ref`, `lazy` or `scope` with aliases.

Basile-z: interesting. If these attributes are accepted but "noop" then their presence should be prevented by the compiler.

edi33416: > Unfortunately, I don't think you can use storage classes like `auto ref`, `ref`, `lazy` or `scope` with aliases.

It didn't even occur to me that this might be ignored. I agree with @Basile-z that this should throw a compiler error if it's not supported: silently ignoring can lead to hard to discover bugs.

@ZombineDev do you know what is the reason to ignore the storage classes?

edi33416:
> Out of curiosity, what is the vision for how this will relate to [DMD clone.d](https://github.com/dlang/dmd/blob/67e8328f25d7c9733f83ad4c33226d2cb2c3a63b/src/dmd/clone.d)'s `buildXopEquals`/`buildXopCmp`/`buildXtoHash`?

I might be wrong, but this will not cause trouble for `buildXopEquals`/`buildXopCmp`/`buildXtoHash`.
The mixins in this PR provide the user with greater flexibility, so he can pick and choose what are his significant (to be taken into account) member fields.

The user can even provide the name of a method/property to be used inside the operators. The reason for this is: the user used to have a field `X` that he now changed to a `@property` -> his code will continue to work.

Another benefit, imho, is that the implementation is in the library and it's easier for a contributor to understand.

What do you think?

n8sh:
>What do you think?

I think it could be convenient. Syntax-wise it might be worth considering an approach based around annotating specific fields (either whitelist or blacklist) like in Java's [Project Lombok](https://projectlombok.org/features/EqualsAndHashCode):

```java
import lombok.EqualsAndHashCode;

@EqualsAndHashCode
public class EqualsAndHashCodeExample {
  private transient int transientVar = 10;
  private String name;
  private double score;
  @EqualsAndHashCode.Exclude private Shape shape = new Square(5, 10);
  private String[] tags;
  @EqualsAndHashCode.Exclude private int id;
  
  public String getName() {
    return this.name;
  }
  
  @EqualsAndHashCode(callSuper=true)
  public static class Square extends Shape {
    private final int width, height;
    
    public Square(int width, int height) {
      this.width = width;
      this.height = height;
    }
  }
}
```

edi33416:
> > What do you think?
> 
> I think it could be convenient. Syntax-wise it might be worth considering an approach based around annotating specific fields (either whitelist or blacklist) like in Java's [Project Lombok](https://projectlombok.org/features/EqualsAndHashCode):
> 
> ```java
> import lombok.EqualsAndHashCode;
> 
> @EqualsAndHashCode
> public class EqualsAndHashCodeExample {
>   private transient int transientVar = 10;
>   private String name;
>   private double score;
>   @EqualsAndHashCode.Exclude private Shape shape = new Square(5, 10);
>   private String[] tags;
>   @EqualsAndHashCode.Exclude private int id;
>   
>   public String getName() {
>     return this.name;
>   }
>   
>   @EqualsAndHashCode(callSuper=true)
>   public static class Square extends Shape {
>     private final int width, height;
>     
>     public Square(int width, int height) {
>       this.width = width;
>       this.height = height;
>     }
>   }
> }
> ```

At DConf, @atilaneves also asked me if I have thought about using UDAs to exclude fields.
At the time, I had not. I've given it some thought since and I think we could have UDAs as an enhancement to the current implementation. I'm afraid we could "clash" with some existing UDAs that users have in their codebases.

The reason I'm saying we can have it as an enhancement is because I think we could see how the users use the current implementation and if they feel that UDAs would increase their productivity / make the code nicer.

As a final note, I'm not opposing to using UDAs those are just my thoughts.

atilaneves:
@edi33416:

> I'm afraid we could "clash" with some existing UDAs that users have in their codebases.

This is easily disambiguated by using the fully qualified name when/if necessary.

std/typecons.d
ZombineDev: Because they are not type constructors, but parameter storage classes and `alias` can "point" to only types, symbols and occasionally values of which storage classes are neither.
For example, you can't declare a function local or class/struct member variable with the same storage classes as a parameter.

This is unfortunately by design and will take an exceptional DIP to convince W&A otherwise. Manu has tried to argue for making `ref` a first class citizen (i.e. a type constructor), so it can play nice with `std.meta` metaprogramming for years in the newsgroup, but has not succeeded so far. Many others have tried to independently argue that `scope` should be a transitive (again meaning a type constructor like `const`), but have again failed.

For example, my interpretation of the outcome of this bug report https://issues.dlang.org/show_bug.cgi?id=17764 is that Walter values parameter storage classes not being type constructors more then memory safety.

I wonder what would be @atilaneves take on this.

atilaneves: I don't have an opinion on it yet, but would love to hear arguments for/against/about it.

ZombineDev: In general, the problem with parameter storage classes is that they don't compose well. (Other storage classes like `static` are fine as they are, because you rarely have composition problems there.)

In case of DIP1000 / `scope` the problem is more dramatic as the compiler simply gives up and allows memory-unsafe code in `@safe` functions.

Basile-z: > Manu has tried to argue for making ref a first class citizen

about `ref`, as you should know, it has been partially addressed lately (`ref` for alias function is handled but still not as a type ctor, rather as a special case).

ZombineDev: > ref for alias function is handled but still not as a type ctor, rather as a special case

@Basile-z thanks for point this out! Can you point me to the PR that changed this lately?

Basile-z: https://dlang.org/changelog/2.086.0.html#funcliteralsbyref

ZombineDev: I know this enhancement, but it's unrelated. The core issue is the following:
```D
alias RI = ref int;
pragma (msg, RI);   // prints "int" during compilation

void inc(RI ri) { ri++; }

void main()
{
    int x = 10;
    x.writeln;   // prints "10" at run-time
    x.inc;
    x.writeln;   // prints "10" at run-time
}
```

That doesn't work either:
```D
alias Ref(T) = ref T;
void inc(Ref!int ri) { ri++; }
pragma (msg, typeof(&inc));   // prints "void function(int ri)" at compile-time
```

edi33416:
@atilaneves is UDAs support required as part of this PR or can it come as a future PR?

atilaneves:
@edi33416 I wouldn't say UDA support is required, no.

RazvanN7:
@edi33416 What are your plans with this PR?

RazvanN7:
@edi33416 Is there anything blocking this PR?

edi33416:
@RazvanN7 I don't think there are any blockers for this PR. I've just rebased with master and pushed the updated branch.
Let's see if the checks pass

std/typecons.d
```diff
@@ -5578,6 +5578,779 @@ private
...
+    }
+
+    auto t1 = new Test(1);
+    auto t2 = new Test(2);
+    assert((t1 < t2) == 0);
+    assert((() @trusted => t1 == t2)());
+    assert(t1.toHash() == t2.toHash());
+}
+
+template GetAllFields(T)
```
RazvanN7, line 5910: This is a public symbol, therefore it needs to be documented.

RazvanN7:
Since this PR is adding so many public symbols, it requires a changelog entry.

RazvanN7:
This assert seems to be tripping:

```
core.exception.AssertError@std/typecons.d(5802): unittest failure
```

RazvanN7:
@edi33416 please address @atilaneves 's comments. Also, can you split this up into 3 PRs?

LightBender:
PR closed as stalled. If you wish to resurrect it you are welcome to do so, however it is recommend to split this up in to multiple PR's.

PR #7050, opened 2019-05-31
Explicitly link to math library

marler8997:
Second pass of https://github.com/dlang/phobos/pull/7021

Looks like we also need `-lm`.


marler8997:
@thewilsonator can you trigger a rebuild on `buildkite`?  It looks like there a `Rebuild` button but I don't have permissions.

thewilsonator:
Done.

PR #7052, opened 2019-06-01
Add conf file to tell compiler about the standard library

marler8997:
Instead of hardcoding phobos' dependencies, this change adds a `defaultlib.conf` file which DMD can read to know how to import and link to phobos.  The suppliemental PR in DMD is here: https://github.com/dlang/dmd/pull/9934

This is just an exploratory implementation right now until leadership approves the concept.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -186,6 +186,7 @@ ifeq (,$(findstring win,$(OS)))
 else
 	LIB:=$(ROOT)/phobos.lib
 endif
+CONF:=$(ROOT)/defaultlib.conf
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


marler8997:
Closing per @WalterBright comment from corresponding DMD PR (https://github.com/dlang/dmd/pull/9934)

> No. It's so you can configure the compiler. I don't see the point for enabling multiple ones. They aren't that large - just copy an existing one and edit it for the project's needs. I do it all the time (necessary when mixing and matching for developing the compiler). I've never felt a need for multiple ones at the same time.
>
> This PR needs more justification.

PR #7053, opened 2019-06-01
Remove redundant GC.addRange/removeRange from std.container.array

n8sh:
The RefCounted wrapper will register Array._data for scanning by the GC so we don't need to separately register/deregister the payload.


n8sh:
Closed: incorrect.

PR #7054, opened 2019-06-01
Avoid mulu in std.container.array.Array!T when T.sizeof == 1

n8sh:
LDC 1.15 and GDC with optimizations enabled recognize that `mulu` with a second argument of 1 will never overflow and elide unnecessary operations but DMD and LDC 1.14 do not.

### DIFF
--- a/std/container/array.d
+++ b/std/container/array.d
@@ -310,12 +310,19 @@ if (!is(Unqual!T == bool))
             if (_capacity < newLength)
             {
                 // enlarge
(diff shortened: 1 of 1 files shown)


Basile-z:
However, @WalterBright, would it be possible to detect and optimize mulu in cgelem ?

Basile-z:
Actually wait. Wouldn't it be possible to do the bypass in the backend ?

PR #7055, opened 2019-06-02
Better selection of a common type

SSoulaimane:
For https://github.com/dlang/dmd/pull/9943.

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -2271,13 +2271,13 @@ if (alternatives.length >= 1 &&
     auto c = 2;
     const d = 3;
     auto cd = either!(a => a == 3)(c, d); // use predicate
-    static assert(is(typeof(cd) == int));
+    //static assert(is(typeof(cd) == const(int)));
     assert(cd == d);
 
     auto e = 0;
     const f = 2;
     auto ef = either(e, f);
-    static assert(is(typeof(ef) == int));
+    //static assert(is(typeof(ef) == const(int)));
     assert(ef == f);
 }
 
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -11729,7 +11729,7 @@ private @safe:
 
         immutable ot = otherTZ.utcToTZ(0);
 
-        auto diffs = [0L, lt, ot];
+        auto diffs = [0L, cast()lt, cast()ot];
         auto diffAA = [0L : Rebindable!(immutable TimeZone)(UTC())];
         diffAA[lt] = Rebindable!(immutable TimeZone)(LocalTime());
         diffAA[ot] = Rebindable!(immutable TimeZone)(otherTZ);
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -1274,8 +1274,7 @@ pure @safe nothrow unittest
     auto s3 = "string three".byCodeUnit;
     static assert(is(typeof(s3.front) == immutable char));
     auto r2 = s1.chain(s3);
-    // chaining ranges of mixed character types gives `dchar`
-    static assert(is(typeof(r2.front) == dchar));
+    //static assert(is(typeof(r2.front) == const(dchar)));
 
     // use byChar on character ranges to correctly convert them to UTF-8
     auto r3 = s1.byChar.chain(s3);


std/datetime/systime.d
```diff
@@ -11729,7 +11729,7 @@ private @safe:
 
         immutable ot = otherTZ.utcToTZ(0);
 
-        auto diffs = [0L, lt, ot];
+        auto diffs = [0L, cast()lt, cast()ot];
```
jmdavis: Don't cast these. If it's a problem that they're `immutable`, then just change their declarations to `auto`. However, I don't see why the change should be necessary.

std/algorithm/comparison.d
```diff
@@ -2271,13 +2271,13 @@ if (alternatives.length >= 1 &&
     auto c = 2;
     const d = 3;
     auto cd = either!(a => a == 3)(c, d); // use predicate
-    static assert(is(typeof(cd) == int));
+    //static assert(is(typeof(cd) == const(int)));
```
jmdavis: Why are these all suddenly becoming `const`? And simply commenting out tests is definitely not good. Are you planning to comment them out now and then create a separate PR to uncomment them once the dmd PR is merged? Either way, it sounds like this is a breaking change. Having a bunch of ranges suddenly have const elements is definitely going to break code - as evidenced by the change to std.datetime in this PR. What's the motivation here? Looking at the bugs referenced in the dmd PR, they all refer to widening types, not changing to `const`.

std/datetime/systime.d
SSoulaimane: These are  value types, the variables `lt` and `ot` are never changed still even with the cast. The purpose of the cast is to make the array `diffs` mutable such that it can be sorted in-place.

std/algorithm/comparison.d
SSoulaimane: Yes, the PR was mentioned in the initial comment https://github.com/dlang/dmd/pull/9943.
See also this discussion with Walter prior to this https://github.com/dlang/dmd/pull/9893#issuecomment-496031539.

SSoulaimane: The loss of `const` was linked to the widening.
What do you think the type of `[0, cast(const)1]` should be? `const(int)[]` or `int[]`, likewise with `[cast(const)1]`.

std/datetime/systime.d
jmdavis: Given that the main purpose of these variables is to use them in this array, it's just silly to explicitly cast away `immutable` rather than use `auto` if the compiler isn't going to decide to implicitly convert to remove `immutable`.

std/algorithm/comparison.d
SSoulaimane: This is not clearly defined in the spec yet.

std/datetime/systime.d
SSoulaimane: Still making variables that are not going to change `const` or `immutable` is a good practice.

std/algorithm/comparison.d
jmdavis: If we weren't talking about types which implicitly converted from `const` to mutable, the resulting elements would need to be `const` (and thus have to be now without your changes), but when they implicitly convert to mutable, having them be mutable is far more useful and wouldn't break code the way that the current changes are. If @WalterBright is in favor of making the result `const`, then fine, but it's clearly going to break code, and in general is going to be annoying. `[cast(const)1]` by itself should obviously be `const(int)[]` though, since all of the elements are `const` (and that's the current behavior).
Either way, the change needs to be approved by @WalterBright , and I see no discussion of arrays in the linked PRs.

std/datetime/systime.d
jmdavis: It's dumb if the code is just going to turn around and cast to make them mutable. And any time that `cast()` is used, it's a code smell. There are cases where it makes sense, and it's not actually causing problems here, because the type is a value type, but it's still a code smell any time that it's in code and reason to examine the code more carefully. As such, it's far better to avoid needing such casts.

std/algorithm/comparison.d
SSoulaimane: Arrays were affected because DMD uses a ternary expression (`?:`) internally to find a common type for array literals. As It does it with associative array literals (`[cast(const)"0": 0]` is `int[const(string)]`)

std/datetime/systime.d
SSoulaimane: There is a caveat either way you go. `auto` means the variable is likely to change which is not true in this situation, while a `cast()` means something is not right which is correct in this case, the type deduction doesn't result in the needed type hence the `cast()`.

jmdavis: It's not like this code needs to protect against changing the variables in question, and `cast()` should almost always be avoided if possible. The code only uses `immutable` here, because I default to `immutable` for variables of primitive types if it doesn't matter. Once it does, I use `auto`, and I'm very much against using `cast()`.

std/algorithm/comparison.d
jmdavis: The way that the compiler decides the type of an array literal has never seemed consistent to me, and this change definitely seems to be for the worse with regards to `const`. Regardless, the dmd PR needs to be approved before it makes any sense to merge this one. And if this PR gets merged, another one will need to be created where the tests are not commented out.

atilaneves:
Please don't comment out assertions.

thewilsonator:
That is the standard way of dealing with changes that spread across DMD + druntime/phobos. If you have a better way of dealing with it I'm all ears.

atilaneves:
@thewilsonator Not off the top of my head, no. I'll have to think of a better way, if there even is one.

thewilsonator:
```
Check for trailing whitespace
  | grep -nr '[[:blank:]]$' etc std ; test $? -eq 1
  | Enforce whitespace before opening parenthesis
```

jmdavis:
> That is the standard way of dealing with changes that spread across DMD + druntime/phobos. If you have a better way of dealing with it I'm all ears.

In the past, we've just pushed both the dmd and druntime/phobos changes without requiring that the autotester passed first, but with how much extra CI stuff and rules that we've added, I don't know how easy that is to do now. Regardless, if the dmd changes are approved, and this PR is merged, then there's definitely going to need to be another PR to fix the assertions once the dmd PR has been merged.

RazvanN7:
I'm going to close this as the author has abandoned it. This PR is linked to the dmd one, which, in turn is linked to the associated bugs. If the author or anyone else wants to pick this up the trail can be easily discovered.

PR #7058, opened 2019-06-05
Pass writer as `auto ref` to string formatting functions in datetime

david-eckardt-frequenz:
Currently this raises a compile error:
```d
import std.stdio;
import std.datetime.systime;
Clock.currTime().toISOExtString(stdout.lockingTextWriter());
```
while this works:
```d
import std.stdio;
import std.datetime.systime;
auto w = stdout.lockingTextWriter();
Clock.currTime().toISOExtString(w);
```
The cause is that the argument for `toISOExtString` is `ref` so it accepts only lvalues.  This PR changes all output stream `ref` arguments for string formatting functions in `datetime` to `auto ref` so that rvalues are accepted, too.

### DIFF
--- a/std/datetime/date.d
+++ b/std/datetime/date.d
@@ -2874,7 +2874,7 @@ public:
     }
 
     /// ditto
-    void toISOString(W)(ref W writer) const
+    void toISOString(W)(auto ref W writer) const
     if (isOutputRange!(W, char))
     {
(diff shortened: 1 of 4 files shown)


david-eckardt-frequenz:
All right, I didn’t foresee this problem—it has never happened to me, but fair enough. Closing.

PR #7061, opened 2019-06-06
Variant of an array can be compared with another array

Zevenberge:
Fixes 11061

### DIFF
--- a/std/variant.d
+++ b/std/variant.d
@@ -285,7 +285,14 @@ private:
         static bool tryPutting(A* src, TypeInfo targetType, void* target)
         {
             alias UA = Unqual!A;
-            alias MutaTypes = AliasSeq!(UA, ImplicitConversionTargets!UA);
+            static if (isStaticArray!A && is(typeof(UA.init[0])))
+            {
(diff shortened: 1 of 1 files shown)


std/variant.d
```diff
@@ -3005,3 +3019,33 @@ if (isAlgebraic!VariantType && Handler.length > 0)
     auto b = v | s;
     assert(b == 3);
 }
+
+@system unittest
```
thewilsonator: Can these be `@safe`?

Zevenberge: I am afraid not. The function pointer taking care of the type magic is `@system`, as it might call some `core.stdc.*` functions like memcpy. It can be made `@trusted`, but it might be out of scope for this issue.

std/variant.d
```diff
@@ -285,7 +285,14 @@ private:
         static bool tryPutting(A* src, TypeInfo targetType, void* target)
         {
             alias UA = Unqual!A;
-            alias MutaTypes = AliasSeq!(UA, ImplicitConversionTargets!UA);
+            static if(isStaticArray!A && is(typeof(UA.init[0])))
```
Basile-z: take care to the D style, here a space misses between the keyword and the left paren. `if (`

Zevenberge: Thanks.

PR #7063, opened 2019-06-10
 Make ErrnoException.errno nothrow pure @nogc @safe 

chloekek:
I think the field could be a public const(uint) instead; without a property accessor. However, doing so would be a breaking change.


thewilsonator:
Thanks! Looks like the unittest just below it can be made `@safe` too.

chloekek:
@thewilsonator fixed, also made the constructors @safe (rather than @trusted, so no changelog entry required).

chloekek:
Squashed to keep history clean. No other changes made.

thewilsonator:
This fails to pass because benchmark builds against the master and link against the PR.

PR #7064, opened 2019-06-11
function attributes go on the right of function aliases

thewilsonator:
FuncType is actually `@system` because of this, for some reason it is still `pure` and `nothrow`.


thewilsonator:
Yes, https://github.com/dlang/dmd/pull/10018 they could probably be made to work for function attributes, but things like `scope` for delegates doesn't.

PR #3915, opened 2016-01-10
Add std.range.Option

JakobOvrum:
Modeled after Scala's [scala.Option](http://www.scala-lang.org/api/2.8.1/scala/Option.html) type. It's been discussed on the forums before, I found at least [this thread](http://forum.dlang.org/post/l87ivq$263r$1@digitalmars.com).

~~(I'll move it to std.experimental.typecons once #2945 is merged, as it establishes the module)~~

PR #7065, opened 2019-06-11
Add std.typecons: Optional, a version of `Nullable` without `alias get this`

FeepingCreature:
It's basically a copypaste.

Follow-up to https://github.com/dlang/phobos/pull/7060 .

Changes: `.get` is now `.value` (`.get` was always a weird naming); behavior is well-defined in the absence of a value. Old toString methods have been removed. `nullify` has no equivalent. Everything else is the same, down to the unittests. `apply` has been changed to work with both.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -3909,6 +3909,375 @@ if (is (typeof(nullValue) == T))
     assert(ntts.to!string() == "2.5");
 }
 
+/**
+Defines a type that combines its parameter `T` and a value where that
+parameter is absent. Its distinction to `Nullable` is that an
+`Optional` value may be either present or absent, whereas a `Nullable`
+value is assumed to be non-null. This shows in that `Optional!T`
+does not implicitly convert to `T`.
+
+Practically `Optional!T` stores a `T` and a `bool`.
+ */
+struct Optional(T)
+{
+    private union DontCallDestructorT
+    {
+        T payload;
+    }
+
+    private DontCallDestructorT _value = DontCallDestructorT.init;
+
+    private bool _present = false;
+
+/**
+Constructor initializing `this` with `value`.
+
+Params:
+    value = The value to initialize this `Optional` with.
+ */
+    this(inout T value) inout
+    {
+        _value.payload = value;
+        _present = true;
+    }
+
+    static if (is(T == struct) && hasElaborateDestructor!T)
+    {
+        ~this()
+        {
+            if (_present)
+            {
+                destroy(_value.payload);
+            }
+        }
+    }
+
+    /**
+      If they are both absent, then they are equal. If one is absent and the other
+      is present, then they are not equal. If they are both present, then they are
+      equal if their values are equal.
+      */
+    bool opEquals()(auto ref const(typeof(this)) rhs) const
+    {
+        if (!_present)
+            return !rhs._present;
+        if (!rhs._present)
+            return false;
+        return _value.payload == rhs._value.payload;
+    }
+
+    /// Ditto
+    bool opEquals(U)(auto ref const(U) rhs) const
+    if (is(typeof(this.value == rhs)))
+    {
+        return _present ? rhs == _value.payload : false;
+    }
+
+    ///
+    @safe unittest
+    {
+        Optional!int absent;
+        Optional!int a = 42;
+        Optional!int b = 42;
+        Optional!int c = 27;
+
+        assert(absent == absent);
+        assert(absent == Optional!int.init);
+        assert(absent != a);
+        assert(absent != b);
+        assert(absent != c);
+
+        assert(a == b);
+        assert(a != c);
+
+        assert(absent != 42);
+        assert(a == 42);
+        assert(c != 42);
+    }
+
+    @safe unittest
+    {
+        // Test constness
+        immutable Optional!int a = 42;
+        Optional!int b = 42;
+        immutable Optional!int c = 29;
+        Optional!int d = 29;
+        immutable e = 42;
+        int f = 29;
+        assert(a == a);
+        assert(a == b);
+        assert(a != c);
+        assert(a != d);
+        assert(a == e);
+        assert(a != f);
+
+        // Test rvalue
+        assert(a == const Optional!int(42));
+        assert(a != Optional!int(29));
+    }
+
+    // Issue 17482
+    @system unittest
+    {
+        import std.variant : Variant;
+
+        Optional!Variant a = Variant(12);
+        assert(a == 12);
+        Optional!Variant e;
+        assert(e != 12);
+    }
+
+    size_t toHash() const @safe nothrow
+    {
+        static if (__traits(compiles, .hashOf(_value.payload)))
+            return _present ? .hashOf(_value.payload) : 0;
+        else
+            // Workaround for when .hashOf is not both @safe and nothrow.
+            return _present ? typeid(T).getHash(&_value.payload) : 0;
+    }
+
+    /**
+     * Gives the string `"Optional.absent"` if `present` is `false`. Otherwise, the
+     * result is equivalent to calling $(REF formattedWrite, std,format) on the
+     * underlying value.
+     *
+     * Params:
+     *     writer = A `char` accepting
+     *     $(REF_ALTTEXT output range, isOutputRange, std, range, primitives)
+     *     fmt = A $(REF FormatSpec, std,format) which is used to represent
+     *     the value if this Optional is present
+     * Returns:
+     *     A `string` if `writer` and `fmt` are not set; `void` otherwise.
+     */
+    string toString()
+    {
+        import std.array : appender;
+        auto app = appender!string();
+        auto spec = singleSpec("%s");
+        toString(app, spec);
+        return app.data;
+    }
+
+    /// ditto
+    string toString() const
+    {
+        import std.array : appender;
+        auto app = appender!string();
+        auto spec = singleSpec("%s");
+        toString(app, spec);
+        return app.data;
+    }
+
+    /// ditto
+    void toString(W)(ref W writer, scope const ref FormatSpec!char fmt)
+    if (isOutputRange!(W, char))
+    {
+        import std.range.primitives : put;
+        if (present)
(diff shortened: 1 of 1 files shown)


std/typecons.d
```diff
@@ -3909,6 +3909,335 @@ if (is (typeof(nullValue) == T))
...
+
+    private DontCallDestructorT _value = DontCallDestructorT.init;
+
+    private bool _present = false;
+
+/**
+Constructor initializing `this` with `value`.
+
+Params:
+    value = The value to initialize this `Nullable` with.
```
pineapplemachine: Copypaste error in docs?

FeepingCreature: oops thanks

wilzbach:
> nullify has no equivalent

Could you  please explain why `.optional` isn't possible.

FeepingCreature:
`.optional` is possible, and is the equivalent of `.nullable`. `.nullify`, the method to explicitly empty out an existing `Nullable`, has no equivalent, because it's a weird operation whose purpose I don't understand.

std/typecons.d
```diff
@@ -3909,6 +3909,335 @@ if (is (typeof(nullValue) == T))
...
+    /// ditto
+    void toString(W)(ref W writer, scope const ref FormatSpec!char fmt) const
+    if (isOutputRange!(W, char))
+    {
+        import std.range.primitives : put;
+        if (present)
+            formatValue(writer, _value.payload, fmt);
+        else
+            put(writer, "Optional.absent");
+    }
```
aliak00: Why is a non const and const version needed and not just a const version?

std/typecons.d
```diff
@@ -3909,6 +3909,335 @@ if (is (typeof(nullValue) == T))
...
+throw a `MissingValueException`.
+
+Returns:
+    The value held by this `Optional`.
+ */
+    @property ref inout(T) value() inout @nogc pure @safe
+    {
+        if (!_present)
+        {
+            throw missingValueException_;
```
pineapplemachine: This should not be a recoverable error imo. It's more normal (and I think cleaner) for optional types to be more about asking for permission than asking for forgiveness. I'd suggest either an assertion or an Error as opposed to an Exception.

Here's an example showing how attempting to unwrap an empty optional value in Rust produces a panic (unrecoverable error) and accessing an Option's value is normally done via pattern matching (i.e. checking state first, and then accessing value): http://www.ameyalokare.com/rust/2017/10/23/rust-options.html

FeepingCreature: This is the defining difference between `Nullable` and `Optional.` The entire reason why `get` can be aliased to this in `Nullable` _is_ that accessing a `Nullable` that is null is _undefined_, so from a certain perspective `Nullable!T` to `T` is an operation that "cannot fail" and is thus "safe". It is thus 100% deliberate for `value` to throw an `Exception`, because the state of `Optional but without a value` is a fully defined one and thus one that must result in ordinary control flow in all methods. (That's why `valueUnsafe` is called `value_Unsafe_`.)

std/typecons.d
FeepingCreature: `toString` should still work if `Optional` embeds a data structure with non-const `toString` (as is, sadly, the default).

pineapplemachine:
I agree with @jacob-carlborg about either making Optional a range or make
it very easy to acquire a range over it, as well as adding Some and None. I
don't have a strong opinion about what module the definition goes in.

I do strongly disagree with the opDispatch suggestion, though. This would
make it difficult to change the API later without breaking dependent code,
and I don't believe this is a normal thing for optional types to do in the
standard libraries of other languages which support them.

jacob-carlborg:
> I don't believe this is a normal thing for optional types to do in the standard libraries of other languages which support them.

No, because they have built-in language support. Or the language does not support anything like `opDispatch`.

std/typecons.d
pineapplemachine: @FeepingCreature this is not what normally distinguishes an optional type from a nullable type.

> The core difference between option types and nullable types is that option types support nesting (Maybe (Maybe A) ≠ Maybe A), while nullable types do not (String?? = String?).

https://en.wikipedia.org/wiki/Option_type

Very often, the fundamental motivation for using an optional type is to not involve exceptions. Wrongly attempting to access the value of an optional type is a programmer error, in the same sense attempting to index an out-of-bounds array index is a programmer error. (In which case D produces an unrecoverable RangeError, not an exception.)

pineapplemachine:
@jacob-carlborg I'm not sure what you mean by built-in language support. Of the optional type implementations I've looked at in other languages, you always explicitly `.get` or `.unwrap` or etc. to refer to the value and its properties. I think that using opDispatch this way is bound to cause headaches when expanding the Optional interface with more functionality later on, and bound to cause confusion for people trying to use this type, especially those coming from C++ or Rust.

jacob-carlborg:
> @jacob-carlborg I'm not sure what you mean by built-in language support

I mean that in other language like Swift and C# it's possible to do something like this (here with D syntax):

```d
Object? a;
string b = a?.toString ?? "foobar";
```

In the above example, if `a` does not have a value `"foobar"` will be assigned to `b`. If `a` has a value, whatever `toString` returns will be assigned to `b`.

pineapplemachine:
> I mean that in other language like Swift and C# it's possible to do something like this (here with D syntax):

In your example `Object?` is not an optional Object but a nullable one, and `??` is the null-coalesce operator.

jacob-carlborg:
>  Of the optional type implementations I've looked at in other languages, you always explicitly .get or .unwrap or etc. to refer to the value and its properties

That's not the recommend way to do it in Scala, for example. It's recommended to use algorithms like `map`, `flatMap`, `filter` and so on. Here are some examples [1].

Swift has a different approach. It supports optional chaining and optional nil-coalescing [2], which is basically my `opDispatch` suggestion. It also has a dedicated `if`-syntax:

```swift
if let starPath = imagePaths["star"] {
    print("The star image is at '\(starPath)'")
} else {
    print("Couldn't find the star image")
}
```

[1] https://danielwestheide.com/blog/2012/12/19/the-neophytes-guide-to-scala-part-5-the-option-type.html

[2] https://developer.apple.com/documentation/swift/optional

jacob-carlborg:
> In your example Object? is not an optional Object but a nullable one

It's an optional Object in Swift: https://developer.apple.com/documentation/swift/optional.

aliak00:
I agree with @jacob-carlborg on everything as well and also with @pineapplemachine disagreement about opDispatch. It's iffy at best to have an opDispatch inside a wrapper type because as mentioned, it will hinder the api. What if I put a range in a wrapper that's a range that has an opDispatch and I call .front? Please no.

aliak00:
> That's not the recommend way to do it in Scala, for example. It's recommended to use algorithms like map, flatMap, filter and so on. Here are some examples [1].

Yes, and having optional be a range will enable the scala approach. We can't have Swift's approach without language support. The best I think I at least came up with is [this](https://github.com/aliak00/optional#example-dispatch-usage)

pineapplemachine:
Was just reading about this in Swift after you posted some links. I stand corrected, I wasn't thinking about the safe member access or unwrapping member access operators. I for one would be happy to introduce `?.` and `!.` to D, and then we might use these for accessing attributes of the wrapped value. But please let's not try to contort opDispatch into serving that purpose.

You can at least easily get the same `if` behavior with feep's implementation by adding a `bool opCast(T: bool)` method. (I would recommend this, actually, now seeing that it hasn't got this already.)

As for map and filter and so on, these are not about accessing properties of the enclosed value but are about using a range interface to enumerate the optional's zero-or-one element. This is not a call for opDispatch but an argument for adding a range interface. (Like you mentioned before, and I agree with this.)

Potentially of interest is mach's optional type implementation: https://github.com/pineapplemachine/mach.d/blob/master/mach/types/option.d

aliak00:
I'd much rather have https://github.com/dlang/phobos/pull/3915 revived and worked on instead of another nullable. Plus all the feature that are requested in this thread (except for jacob's opDispatch) are already in the [optional dub package](https://code.dlang.org/packages/optional) - and is already being used and proven. Which I'm more than happy to put in to phobos if desired.

jacob-carlborg:
There should be a method that returns the value if present, otherwise the passed in value. It can be called `or`, `orElse` or `getOrElse`.

FeepingCreature:
It seems I have aesthetic disagreements about "how an Optional type should behave" with pretty much everybody here. Fun!

It seems a lot of people are looking at `Optional` types from functional languages, whereas I see the core examples for it as `Nullable` and `Flag` in the same file.

> I would expect to be able to do something like this: `some(3)` and `none!int`, to indicate that a value is present or not

You are able to do this: `3.present` and `Optional!int.absent`. `absent!int` can be added, I guess, for consistency with things like `Yes` and `No`. `orElse` seems like the sort of thing that should be its own PR. I really hate `optional.front` - what does "front" have to do with accessing a value? - so I'm not sure what to do about that. `opDispatch` seems like it's asking for all kinds of confusing errors.

jacob-carlborg:
> I really hate optional.front - what does "front" have to do with accessing a value?

`front` should not be used directly to access the value. Then you're missing the point. There should be not direct access to the value. Technically `front` will give you direct access to the value but it's only to be used by the algorithms.

> Anyway, these all - aside range support, which I dislike for other reasons - seem like wishlist features, not necessary requirements.

Well, if there are no improvements over `Nullable` I don't see a point in adding this.

FeepingCreature:
The improvement over `Nullable` is the lack of the `alias get this`, basically. Which is a significant improvement, since `alias get this` keeps actively making problems.

aliak00:
> The improvement over Nullable is the lack of the alias get this, basically. Which is a significant improvement, since alias get this keeps causing problems and wasting developer time on bugs undetected until runtime.

I don't think adding an exact copy of a whole type to fix a bug in the original type is the way to go. I think we should press harder on PR #7060 and get that in.

FeepingCreature:
Looks like we're going with improving `Nullable` instead. I'll close this once the deprecation PR is merged.

FeepingCreature:
Deprecation got merged!

PR #7068, opened 2019-06-12
Fix issue 19939 - std.format count width error

shove70:


shove70:
@n8sh @quickfur, Thanks.

thewilsonator:
this should probably target stable.

n8sh:
Changing to target stable.

PR #7070, opened 2019-06-13
Fix issue 19835 - std.algorithm.mutation.copy static array source arg…

shove70:
…ument broken

### DIFF
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -78,7 +78,7 @@ T2=$(TR $(TDNW $(LREF $1)) $(TD $+))
 module std.algorithm.mutation;
 
 import std.range.primitives;
-import std.traits : isArray, isAssignable, isBlitAssignable, isNarrowString,
+import std.traits : isArray, isStaticArray, isAssignable, isBlitAssignable, isNarrowString,
        Unqual, isSomeChar, isMutable;
 import std.meta : allSatisfy;
 import std.typecons : tuple, Tuple;
@@ -365,7 +365,7 @@ Returns:
     The unfilled part of target
  */
 TargetRange copy(SourceRange, TargetRange)(SourceRange source, TargetRange target)
-if (isInputRange!SourceRange && isOutputRange!(TargetRange, ElementType!SourceRange))
+if ((isInputRange!SourceRange || isStaticArray!SourceRange) && isOutputRange!(TargetRange, ElementType!SourceRange))
 {
     static if (areCopyCompatibleArrays!(SourceRange, TargetRange))
     {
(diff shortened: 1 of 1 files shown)


shove70:
@thewilsonator, Thanks!

shove70:
@n8sh, Thank you.  ^_^

std/algorithm/mutation.d
```diff
@@ -365,7 +365,7 @@ Returns:
     The unfilled part of target
  */
 TargetRange copy(SourceRange, TargetRange)(SourceRange source, TargetRange target)
-if (isInputRange!SourceRange && isOutputRange!(TargetRange, ElementType!SourceRange))
+if ((isInputRange!SourceRange || isArray!SourceRange) && isOutputRange!(TargetRange, ElementType!SourceRange))
```
ZombineDev: Most functions in phobos work with ranges, but not with static arrays. Array slices (dynamic arrays) are already ranges and so are covered by the `isInputRange` constraint. If the consensus is that we want to allow `copy` to receive static array arguments, I suggest being explicit and change `isArray` to `isStaticArray`.

Additionally, receiving large objects by value can be a performance issue, so I suggest changing the `SourceRange source` parameter to `auto ref SourceRange source` to avoid unnecessary copying large static arrays.

ZombineDev:
CC @andralex @jmdavis

std/algorithm/mutation.d
shove70: @ZombineDev, Thank you! It has been adjusted to your suggestion: a5b83ca

shove70:
@andralex, @jmdavis, @ZombineDev:

Thank you for your comments, I have benefited a lot. The great Masters'love of technology and their meticulous attitude are highly respected.

In rangebase's function, if copying is given special treatment, it will destroy the style of the overall architecture and possibly cause problems in the future direction of development. I can understand the Masters' concerns. I will close this PR.

The original intention of submitting this PR is that the ability of copy static arrays exists in v2.085.1, and it was suddenly lost in a submission from v2.086.0 onwards. This sudden loss of functionality can cause problems for users.

In v2.085.1:

```
private enum bool areCopyCompatibleArrays(T1, T2) =
    isArray!T1 && isArray!T2 && is(typeof(T2.init[] = T1.init[]));

TargetRange copy(SourceRange, TargetRange)(SourceRange source, TargetRange target)
if (areCopyCompatibleArrays!(SourceRange, TargetRange))
{
    ...
}

TargetRange copy(SourceRange, TargetRange)(SourceRange source, TargetRange target)
if (!areCopyCompatibleArrays!(SourceRange, TargetRange) &&
    isInputRange!SourceRange &&
    isOutputRange!(TargetRange, ElementType!SourceRange))
{
    ...
}

```
In v2.086.0:

```
private enum bool areCopyCompatibleArrays(T1, T2) =
    isArray!T1 && isArray!T2 && is(typeof(T2.init[] = T1.init[]));

TargetRange copy(SourceRange, TargetRange)(SourceRange source, TargetRange target)
if (isInputRange!SourceRange && isOutputRange!(TargetRange, ElementType!SourceRange))
{
    static if (areCopyCompatibleArrays!(SourceRange, TargetRange))
    {
        ...
    }
    else
    {
        ...
    }
}

```

andralex:
Interesting, thanks. The culprit is https://github.com/dlang/phobos/commit/439b9a701b076fe87f6f73c39bc25d7b0edd681b. The previous codebase was "working" for static arrays unintentionally: `copy()` would copy the entire array on the stack prior to copying it to the destination, so two copies would be made instead of one.

shove70:
> Interesting, thanks. The culprit is [439b9a7](https://github.com/dlang/phobos/commit/439b9a701b076fe87f6f73c39bc25d7b0edd681b). The previous codebase was "working" for static arrays unintentionally: `copy()` would copy the entire array on the stack prior to copying it to the destination, so two copies would be made instead of one.

Salute you and thank you for all you have done for us.

PR #7071, opened 2019-06-13
Fix issue 19823 - std.range.dropOne doesn't drop the element when cal…

shove70:
…led after std.algorithm.iteration.filter

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -3426,7 +3426,7 @@ if (isBidirectionalRange!R)
 R dropOne(R)(R range)
 if (isInputRange!R)
 {
-    range.popFront();
+    range.popFrontN(1);
     return range;
 }
 /// ditto
@@ -3459,6 +3459,23 @@ pure @safe nothrow unittest
     auto bd = filterBidirectional!"true"([1, 2, 3]);
     assert(bd.dropOne().equal([2, 3]));
     assert(bd.dropBackOne().equal([1, 2]));
+}
+
+// issue 19823
+@safe unittest
+{
+    import std.format : format;
+
+    auto a = [1, 2, 3, 4];
+    assert(format("%s", a.filter!(a => a != 1).dropOne) == "[3, 4]");
+    assert(format("%s", a.filter!(a => a != 2).dropOne) == "[3, 4]");
+    assert(format("%s", a.filter!(a => a != 3).dropOne) == "[2, 4]");
+    assert(format("%s", a.filter!(a => a != 4).dropOne) == "[2, 3]");
+    assert(format("%s", a.filter!(a => a == 1).dropOne) == "[]");
+    assert(format("%s", a.filter!(a => a == 2).dropOne) == "[]");
+    assert(format("%s", a.filter!(a => a == 3).dropOne) == "[]");
+    assert(format("%s", a.filter!(a => a == 4).dropOne) == "[]");
+    assert(a.filter!(a => a != 1).array.dropOne == [3, 4]);
 }
 
 /**


shove70:
> dropOne should error if there's no element. If this is not wanted, you can use `.drop(1)`

@wilzbach, Thanks.
If used `drop(1)`, the issue still exists. This issue can be solved by using `popFrontN(1)`, which I tested is normal when there are no elements.

jmdavis:
@wilzbach 
> dropOne should error if there's no element.

That's not quite true. If there are no elements, `dropOne` should never be called on the range in the first place, because that means calling `front` on an empty range. Many ranges do assert in `front` that the range is not empty, but it's not required, and that goes away with `-release`. The behavior of calling `front` on an empty range is unspecified. Rather, it's up to the caller to never do that, and it's a bug in the caller if they do. It _might_ then be caught by an assertion, but it might not. Either way, the bug is in the caller.

In any case, this change is clearly wrong. The documentation is quite clear that calling `dropOne`) is equivalent to calling `popFront()` and returning the range, not calling `popFrontN(1)`, which checks `empty` and then only calls `popFront()` if there's an element to pop. If `dropOne` is being called on an empty range, then the bug is definitely in the caller. If it's not being called on an empty range, and calling `dropOne` and `drop(1)` aren't equivalent, then something else is going wrong.

@shove70 Using `drop(1)` in your example makes the code work.

    import std.range: drop;
    import std.stdio: writefln;
    import std.algorithm.iteration: filter;
    import std.array: array;

    void main()
    {
        auto a = [1,2,3,4];
        writefln("%s", a.filter!(a => a != 1).drop(1));
        writefln("%s", a.filter!(a => a != 2).drop(1));
        writefln("%s", a.filter!(a => a != 3).drop(1));
        writefln("%s", a.filter!(a => a != 4).drop(1));
        writefln("%s", a.filter!(a => a == 1).drop(1));
        writefln("%s", a.filter!(a => a == 2).drop(1));
        writefln("%s", a.filter!(a => a == 3).drop(1));
        writefln("%s", a.filter!(a => a == 4).drop(1));
        writefln("%s", a.filter!(a => a != 1).array.drop(1));
    }

prints the correct thing. And if we change it to use assertions and get rid of all of the unnecessary string stuff, `drop(1)` still works. e.g. this example uses assertions but expands them out for no dropping, calling `drop(1)`, and calling `dropOne()`.

    import std.range: drop, dropOne;
    import std.algorithm.iteration: filter;
    import std.array: array;

    void main()
    {
        auto a = [1,2,3,4];

        assert(a.filter!(a => a != 1).equal([2, 3, 4]));
        assert(a.filter!(a => a != 2).equal([1, 3, 4]));
        assert(a.filter!(a => a != 3).equal([1, 2, 4]));
        assert(a.filter!(a => a != 4).equal([1, 2, 3]));
        assert(a.filter!(a => a == 1).equal([1]));
        assert(a.filter!(a => a == 2).equal([2]));
        assert(a.filter!(a => a == 3).equal([3]));
        assert(a.filter!(a => a == 4).equal([4]));
        assert(a.filter!(a => a != 1).equal([2, 3, 4]));

        assert(a.filter!(a => a != 1).drop(1).equal([3, 4]));
        assert(a.filter!(a => a != 2).drop(1).equal([3, 4]));
        assert(a.filter!(a => a != 3).drop(1).equal([2, 4]));
        assert(a.filter!(a => a != 4).drop(1).equal([2, 3]));
        assert(a.filter!(a => a == 1).drop(1).empty);
        assert(a.filter!(a => a == 2).drop(1).empty);
        assert(a.filter!(a => a == 3).drop(1).empty);
        assert(a.filter!(a => a == 4).drop(1).empty);
        assert(a.filter!(a => a != 1).drop(1).equal([3, 4]));

        assert(a.filter!(a => a != 1).dropOne().equal([3, 4]));
        assert(a.filter!(a => a != 2).dropOne().equal([3, 4]));
        assert(a.filter!(a => a != 3).dropOne().equal([2, 4]));
        assert(a.filter!(a => a != 4).dropOne().equal([2, 3]));
        assert(a.filter!(a => a == 1).dropOne().empty);
        assert(a.filter!(a => a == 2).dropOne().empty);
        assert(a.filter!(a => a == 3).dropOne().empty);
        assert(a.filter!(a => a == 4).dropOne().empty);
        assert(a.filter!(a => a != 1).dropOne().equal([3, 4]));
    }

It passes all of the assertions until it hits the first `dropOne` line. So, there is no problem with `drop`, and if you used `drop` in your code, you'd be fine. However, there clearly _is_ a problem with `dropOne`. If I reduce the tests down to

    void main()
    {
        import std.algorithm.iteration : filter;
        import std.range : drop;

        auto arr = [1, 2, 3, 4];
        auto a = [1,2,3,4];

        auto result = a.filter!(a => a != 1)();
        assert(result.save.equal([2, 3, 4]));
        assert(result.drop(1).equal([3, 4]));
    }

that passes, because it uses `drop(1)`. If I then change it to use `dropOne` instead

    void main()
    {
        import std.algorithm.iteration : filter;
        import std.range : dropOne;

        auto arr = [1, 2, 3, 4];
        auto a = [1,2,3,4];

        auto result = a.filter!(a => a != 1)();
        assert(result.save.equal([2, 3, 4]));
        assert(result.dropOne().equal([3, 4]));
    }

then the second assertion fails, still indicating a problem with `dropOne`. However, if I then factor out `dropOne` entirely, and just use `popFront` directly

    void main()
    {
        import std.algorithm.iteration : filter;

        auto arr = [1, 2, 3, 4];
        auto a = [1,2,3,4];

        auto result = a.filter!(a => a != 1)();
        assert(result.save.equal([2, 3, 4]));
        result.popFront();
        assert(result.equal([3, 4]));
    }

the second assertion _still_ fails. This indicates that `dropOne` isn't the problem at all. Rather, it looks like the problem is either with `Filter`'s `popFront`, or the compiler is generating wrong code. I don't have time to dig into this any further at the moment, but if you want to fix this bug, then you'll need to make this reduced example work. Once that's fixed, `dropOne` should then work just fine.

As an aside, even if your fix here were correct, the tests shouldn't be generating strings, because it's not necessary to reproduce the bug. Printing out the results like you did in the bug report is useful for seeing what's happening, but unless converting the ranges to strings is required for reproducing the bug, there's no reason to do all of that string formatting in the tests.

shove70:
@jmdavis, Thank you very much!

The essence of the problem is indeed `filter()` and `struct FilterResult`. Simply fixing the `dropOne()` to solve the issue is indeed too sloppy, sorry.
I will carefully analyze the problem of `struct FilterResult` and fix it.

Also, for your last suggestion, I also feel very helpful to me, thank you.
My intention is to completely cover the usage scenarios in the bug report in tests. The `writefln("%s"...` of the bug report is converted to a string by calling `format()`. Of course this is just my original Thought, I very much agree with your suggestion, I will not perform string conversion operation in the resubmitted PR.

shove70:
@jmdavis, Due to my mistake, I closed this PR by mistake. The new PR is:
https://github.com/dlang/phobos/pull/7072

PR #7072, opened 2019-06-13
Fix issue 19823 - std.range.dropOne doesn't drop the element when cal…

shove70:
…led after std.algorithm.iteration.filter

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -1358,11 +1358,11 @@ private struct FilterResult(alias pred, Range)
 
     void popFront()
     {
+        prime;
         do
         {
             _input.popFront();
         } while (!_input.empty && !pred(_input.front));
-        _primed = true;
     }
 
     @property auto ref front()
@@ -1478,6 +1478,23 @@ private struct FilterResult(alias pred, Range)
     int underX(int a) { return a < x; }
     const(int)[] list = [ 1, 2, 10, 11, 3, 4 ];
     assert(equal(filter!underX(list), [ 1, 2, 3, 4 ]));
+}
+
+// issue 19823
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+    import std.range : dropOne;
+
+    auto a = [1, 2, 3, 4];
+    assert(a.filter!(a => a != 1).dropOne.equal([3, 4]));
+    assert(a.filter!(a => a != 2).dropOne.equal([3, 4]));
+    assert(a.filter!(a => a != 3).dropOne.equal([2, 4]));
+    assert(a.filter!(a => a != 4).dropOne.equal([2, 3]));
+    assert(a.filter!(a => a == 1).dropOne.empty);
+    assert(a.filter!(a => a == 2).dropOne.empty);
+    assert(a.filter!(a => a == 3).dropOne.empty);
+    assert(a.filter!(a => a == 4).dropOne.empty);
 }
 
 /**


shove70:
dropOne should error if there's no element. If this is not wanted, you can use .drop(1)

@wilzbach, Thanks.
If used drop(1), the issue still exists. This issue can be solved by using popFrontN(1), which I tested is normal when there are no elements.

shove70:
Please help me, Masters:

1. In std.range.package.d:
```
...
R drop(R)(R range, size_t n)
if (isInputRange!R)
{
    range.popFrontN(n);
    return range;
}
...
R dropOne(R)(R range)
if (isInputRange!R)
{
    // No 1: drop(1);
    // No 2: range.popFrontN(1);
    return range;
}
...
```

On the surface, these two sentences should be completely equivalent, but in fact they produce different results. It seems a bit confusing. I think this should be the strength and charm of dlang's template.

I don't think the essence of the problem is` drop(), dropOne()` itself, but `struct FilterResult` of std.algorithm.iteration.d:
```
...
private struct FilterResult(alias pred, Range)
{
    alias R = Unqual!Range;
    R _input;
    private bool _primed;

    private void prime()
    {
        if (_primed) return;
        while (!_input.empty && !pred(_input.front))
        {
            _input.popFront();
        }
        _primed = true;
    }

    this(R r)
    {
        _input = r;
    }

    private this(R r, bool primed)
    {
        _input = r;
        _primed = primed;
    }

    auto opSlice() { return this; }

    static if (isInfinite!Range)
    {
        enum bool empty = false;
    }
    else
    {
        @property bool empty() { prime; return _input.empty; }
    }

    //void popFront()
    //{
    //    do
    //    {
    //        _input.popFront();
    //    } while (!_input.empty && !pred(_input.front));
    //    _primed = true;
    //}
    // Modify it to read as follows:
    void popFront()
    {
        prime;
        do
        {
            _input.popFront();
        } while (!_input.empty && !pred(_input.front));
    }

    @property auto ref front()
    {
        prime;
        assert(!empty, "Attempting to fetch the front of an empty filter.");
        return _input.front;
    }

    static if (isForwardRange!R)
    {
        @property auto save()
        {
            return typeof(this)(_input.save, _primed);
        }
    }
}
...
```

Modification here can also solve this issue, I think it is more appropriate to modify here? But the impact here is so great that I dare not modify it easily.

Masters, please give us some advice.

std/range/package.d
```diff
@@ -3461,6 +3461,25 @@ pure @safe nothrow unittest
...
 }
 
+// issue 19823
+@safe unittest
+{
+    import std.format : format;
+    import std.algorithm.iteration : filter;
+
+    auto a = [1, 2, 3, 4];
+    assert(format("%s", a.filter!(a => a != 1).dropOne) == "[3, 4]");
```
Basile-z: Maybe drop `format()` and compare `.array`, like done for the last assertion.

shove70: Hi, Basile-z, Thank you. 

```d
void main() {
    auto a = [1,2,3,4];
    writefln("%s", a.filter!(a => a != 1).dropOne); // [2, 3, 4] WRONG
    writefln("%s", a.filter!(a => a != 2).dropOne); // [3, 4]
    writefln("%s", a.filter!(a => a != 3).dropOne); // [2, 4]
    writefln("%s", a.filter!(a => a != 4).dropOne); // [2, 3]
    writefln("%s", a.filter!(a => a == 1).dropOne); // []
    writefln("%s", a.filter!(a => a == 2).dropOne); // [2] WRONG
    writefln("%s", a.filter!(a => a == 3).dropOne); // [3] WRONG
    writefln("%s", a.filter!(a => a == 4).dropOne); // [4] WRONG
    writefln("%s", a.filter!(a => a != 1).array.dropOne); // [3, 4]
}
```

In this issue, there is no bug in compare `.array`, and the bug will be trigger only when the `dropOne()` operation is performed directly on `filter()` return value `FilterResult`.

shove70:
```d
@safe unittest
{
    import std.algorithm.iteration : filter;

    auto a = [1, 2, 3, 4];
    assert(a.filter!(a => a != 1).dropOne.save.equal([3, 4]));
    assert(a.filter!(a => a != 2).dropOne.save.equal([3, 4]));
    assert(a.filter!(a => a != 3).dropOne.save.equal([2, 4]));
    assert(a.filter!(a => a != 4).dropOne.save.equal([2, 3]));
    assert(a.filter!(a => a == 1).dropOne.save.empty);
    assert(a.filter!(a => a == 2).dropOne.save.empty);
    assert(a.filter!(a => a == 3).dropOne.save.empty);
    assert(a.filter!(a => a == 4).dropOne.dropOne.dropOne.save.empty);
    assert(a.filter!(a => a != 1).array.dropOne == [3, 4]);
}
```

Shall I change it to this way? Thank you!

Basile-z: sorry, you can mark as resolved. I hadn't see that this was the code from the report.

PR #7077, opened 2019-06-17
Fix Issue 19980 - File.byLine skips first line in some cases when used inside map!()

CyberShadow:
Don't read the line upon the construction of `byLine`, as that advances the file pointer and interacts badly with some range constructs involving `byLine` (such as `joiner`).

It's possible to avoid introducing `haveLine` by encoding it into null-ness of `line` (and moving EOF state from null-ness of `line` to null-ness of `buffer`), but that's probably too hacky for Phobos.

This patch also changes `byLine` so that `popFront` doesn't advance the file pointer (only calling `front` after `popFront` does that), but that's not necessary to fix the bug.

The overhead of the boolean checks added in this patch is probably negligible compared to that of the C `FILE*` implementation and the actual cost of doing I/O.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2131,6 +2131,7 @@ Allows to directly use range operations on lines of a file.
             Char[] buffer;
             Terminator terminator;
             KeepTerminator keepTerminator;
+            bool haveLine;
 
         public:
             this(File f, KeepTerminator kt, Terminator terminator)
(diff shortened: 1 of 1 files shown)


thewilsonator:
test case doesn't use map. Is that a reduced test case? Otherwise looks good.

CyberShadow:
Yep, I posted a little explanation in the issue. The `map` range's `front` was evaluated several times, which caused `byLine` to be called more than once.

PR #7080, opened 2019-06-19
Fix issue 19986 - Can't assign large const T to std.Variant.VariantN

shove70:


John-Colvin:
A solution should probably handle this case too:

```D
import std.variant;
void foo()
{
    VariantN!32 v;
    v = const(S).init;
}

struct S
{
    ubyte[33] s;
}
```

shove70:
> A solution should probably handle this case too:
> 
> ```d
> import std.variant;
> void foo()
> {
>     VariantN!32 v;
>     v = const(S).init;
> }
> 
> struct S
> {
>     ubyte[33] s;
> }
> ```

Okay, let me deal with it. Thank you for reminding me.

shove70:
@John-Colvin, Help me see, is it feasible to solve this issue? Thank you!

std/variant.d
```diff
@@ -690,12 +690,29 @@ public:
...
-                    *p = rhs;
+                    static if (!isMutable!T)
+                    {
+                        alias UT = Unqual!T;
+                        auto p = cast(UT*)(new U[n]).ptr;
+                        *p = cast(UT) rhs;
+                    }
+                    else
+                    {
+                        auto p = cast(T*)(new U[n]).ptr;
```
thewilsonator: You can just drop this branch and use `Unqual` everywhere since `Unqual!S == S`

shove70: Thanks!
It has been handled as review required.

PR #7082, opened 2019-06-21
19987 - Reduce memory used by VariantN

Kriyszig:
Issue: [19987](https://issues.dlang.org/show_bug.cgi?id=19987)

Swapping the `fptr` and `union` around will avoid larger alignment.

### DIFF
--- a/std/variant.d
+++ b/std/variant.d
@@ -179,8 +179,6 @@ private:
             apply, postblit, destruct }
 
     // state
-    ptrdiff_t function(OpID selector, ubyte[size]* store, void* data) fptr
-        = &handler!(void);
     union
     {
         align(maxVariantAlignment!(AllowedTypes)) ubyte[size] store;
@@ -188,6 +186,8 @@ private:
         static if (size >= (void*).sizeof)
             void*[size / (void*).sizeof] p;
     }
+    ptrdiff_t function(OpID selector, ubyte[size]* store, void* data) fptr
+        = &handler!(void);
 
     // internals
     // Handler for an uninitialized value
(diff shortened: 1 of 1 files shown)


std/variant.d
```diff
@@ -1234,6 +1234,14 @@ public:
     assert(a[5] == 7);
 }
 
+@system unittest
```
thewilsonator: It seems this should be `@safe`

Kriyszig: Addressed in 6553c2e
Thanks for the review :+1:

thewilsonator: No problems, thanks for the bug fix.

thewilsonator:
Seems to fail on 32bit.

Kriyszig:
Noticed that! I'll try to get my hands on a 32bit system and lot the result of `sizeof`

thewilsonator:
just pass `-m32` to the compiler.

std/variant.d
```diff
@@ -1234,6 +1234,22 @@ public:
     assert(a[5] == 7);
 }
 
+@safe version(X86_64) unittest
+{
+    alias Var = VariantN!(24);
+
+    Var a;
+    assert(a.sizeof == 32);
```
MoonlightSentinel: You could calculate the expected size based on the pointer size:
```suggestion
    assert(a.sizeof == 24 + (void*).sizeof);
```

Kriyszig: Thanks a lot for this suggestion  (^_^)
Changed the unit test in 6da85c4

std/variant.d
```diff
@@ -1234,6 +1234,14 @@ public:
     assert(a[5] == 7);
 }
 
+@safe unittest
+{
+    alias Var = VariantN!(24);
+
+    Var a;
+    assert(a.sizeof == 24 + (void*).sizeof);
```
Geod24:
```d
@safe unittest
{
    static assert(VariantN!(24).sizeof == 24 + (void*).sizeof);
    static assert((VariantN!(24)[4]).sizeof == ...);
}
```

Kriyszig: Addressed in c163aee
I copied the test below it so the alias remained. Sorry for the inconvenience :sweat_smile:

Kriyszig:
Ready for review :octocat:

PR #7084, opened 2019-06-22
Fix issue 19994 - Can't nest self-referential Algebraic types

pbackus:
Previously, self-referential `Algebraic` types using `This` could not be nested, since the outer `Algebraic` instance would replace the inner `Algebraic`'s `This` argument when computing `AllowedTypes`, leading to a recursive template expansion. With these changes, each occurrence of `This` in `Algebraic`'s argument list is replaced only by the inner-most `Algebraic` instance that contains it. This ensures that no occurrence of `This` is ever replaced more than once.

These changes are based on pbackus/sumtype@5a6136855acd9e7f3ab7c421fc2f531055114240, which fixed an analagous issue in the `sumtype` DUB package.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -8708,6 +8708,8 @@ public:
     assert(flags.A && !flags.B && !flags.C);
 }
 
+private enum false_(T) = false;
+
 // ReplaceType
 /**
 Replaces all occurrences of `From` into `To`, in one or more types `T`. For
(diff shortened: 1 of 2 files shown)


Basile-z:
This PR should be split in two.

1. the typecons addition
2. fix for 19994

std/typecons.d
```diff
@@ -8725,28 +8727,50 @@ placeholder type `This` in $(REF Algebraic, std,variant).
 Returns: `ReplaceType` aliases itself to the type(s) that result after
 replacement.
 */
-template ReplaceType(From, To, T...)
+alias ReplaceType(From, To, T...) = ReplaceTypeUnless!(false_, From, To, T);
```
n8sh: Adding new public templates will make the approval process much slower (potentially taking weeks longer). I suggest keeping `ReplaceType` and `ReplaceTypeUnless` internal-use-only in this PR and making them in a future PR if you want.

pbackus: I'd really like `ReplaceTypeUnless` to be public so that I can use it in `sumtype` too without copy+pasting. I'm happy to split this into two PRs if it will make the overall process easier or more efficient, but that's a question of "how" best to make `ReplaceTypeUnless` public, not "if".

atilaneves: Seems like `ReplaceType` would be a candidate for `std.traits`.

pbackus: If I were starting from scratch, I'd put it in `std.meta`, alongside non-recursive `Replace` and `ReplaceAll`. Given that `ReplaceType` is already in `std.typecons`, however, I don't think it makes sense to put `ReplaceTypeUnless` anywhere else.

atilaneves: That makes sense.

pbackus:
Ping @andralex @ZombineDev @MetaLang. The auto-tester issue is a spurious failure of [this `std.datetime` test](https://github.com/pbackus/phobos/blob/fix-19994/std/datetime/stopwatch.d#L442-L462), and all review comments have been addressed.

thewilsonator:
(all the autotester results are out of date, if you push to this, you'll see they will update).

PR #7086, opened 2019-06-24
Tid.hasMessages

zorael:
I want to make the case for adding Tid.hasMessages. Naming is hard.

NG thread: https://forum.dlang.org/post/iuvdrthsugkwhxqjgetl@forum.dlang.org

TL;DR: There is no way to tell if there are concurrency messages waiting in a thread (essentially no way to check for threads' message box .empty) without actually attempting to receive, thereby consuming any message and committing to handling it.

Going by the range analogy, currently it's as if Tid's message box only has `popFront` that also returns a `front` (no separate `front`, no `empty`). The only way to know if the message queue is empty is to try to pop and see if it failed.

Knowing whether it's empty allows you to only receive when there's actually something there. Drawbacks to 0-second timeouts mentioned in the NG thread; I'll be happy to argue it in more detail if requested.

What this is:
* The equivalence of .empty for the concurrency message queue
* Literally a oneliner, plus tests

What this is not:
* Breaking
* An attempt to make concurrency message boxes ranges
* Exposing the private MessageBox member of Tid, subverting send and receive
* Mere syntactic sugar for something already possible with existing library functions

You could make the argument that this should only be possible with `thisTid`. I don't mind such a limitation but I also don't really see the need for it.

I'm not sure what a good name is. Tid.empty rings nice but isn't quite true, since it's (the private) Tid.mbox that's empty. Tid.messagesWaiting. Tid.messageBoxEmpty. Tid.canReceive.

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -334,6 +334,14 @@ public:
         formattedWrite(sink, "Tid(%x)", cast(void*) mbox);
     }
 
+    /**
+     * Returns whether or not this Tid has concurrency messages waiting to be
+     * received.
+     */
+    bool hasMessages() @safe pure nothrow @nogc
+    {
+        return (mbox.m_localMsgs + mbox.m_sharedBox.length) > 0;
+    }
 }
 
 @system unittest
@@ -346,6 +354,12 @@ public:
     assert(text(tid2) != "Tid(0)");
     auto tid3 = tid2;
     assert(text(tid2) == text(tid3));
+
+    assert(!thisTid.hasMessages);
+    thisTid.send(123);
+    assert(thisTid.hasMessages);
+    int i = receiveOnly!int();
+    assert(!thisTid.hasMessages);
 }
 
 /**


thewilsonator:
`hasMessagesPending` maybe?

kubo39:
Needs changelog?

std/concurrency.d
```diff
@@ -334,6 +334,14 @@ public:
         formattedWrite(sink, "Tid(%x)", cast(void*) mbox);
     }
 
+    /**
+     * Returns whether or not this Tid has concurrency messages waiting to be
+     * received.
+     */
+    bool hasMessages() @safe pure nothrow @nogc
+    {
+        return (mbox.m_localMsgs + mbox.m_sharedBox.length) > 0;
```
shove70: Should the access to the private members `m_localMsgs` and `m_sharedBox.length` of the private member m_box be considered Mutex:
`synchronized (m_lock)`?

zorael: `mboxFull` doesn't, but maybe I'm misinterpreting what the comment means by "local".

```d
        // Routines involving local data only, no lock needed.

        bool mboxFull() @safe @nogc pure nothrow
        {
            return m_maxMsgs && m_maxMsgs <= m_localMsgs + m_sharedBox.length;
        }
```

zorael:
Assuming *pty* means priority, it needs to check `m_localPty.empty` and `m_sharedPty.empty` too. I'll push another commit once I figure out why the tests fail when from inside the Phobos test suite but works fine outside of it. Apologies, I should have tested it better first.

It's like there's a ghost `LinkTerminatedException` in `m_localBox` from an earlier unittest that gets ignored during receives yet keeps the list from being empty.

std/concurrency.d
```diff
@@ -334,6 +334,14 @@ public:
         formattedWrite(sink, "Tid(%x)", cast(void*) mbox);
     }
 
+    /**
+     * Returns whether or not this Tid has concurrency messages waiting to be
```
Basile-z: `Returns:`

zorael:
TL;DR: Threads leak `linkDead` control messages so `m_localBox.length` and `m_localMsgs` cannot be trusted. Unsure how to proceed.

Sending creates a message and places it in a message box, which is essentially a linked list[1]. There are four message boxes in play. One is the local box `m_localBox`, which is just local storage. Then there's the priority box `m_localPty`, which is the same except for priority messages. Then there's `m_sharedBox` and `m_sharedPty`, which are the same as the previous two but used with locks in `synchronized` blocks. `m_localMsgs` is just cached `m_localBox.length`[2].

When a message is sent, it is appended to the shared box[3], where it awaits a receive. When a thread receives, it tries to match all messages in both the shared and local boxes with the handler functions/delegates passed to the receive call. If something can be handled, it is, and the message is removed. If something did not match any handler, it is moved to the local box[4].

When a thread dies, it sends a `linkDead` control message back to its owner. If the owner thread had spawned the subthread with `spawnLinked` instead of `spawn`, the control message is converted to a `LinkTerminated` exception, to be caught with a normal receive handler or try-catch[5]. (If the dying thread is the owner thread of the current one, it is converted to an `OwnerTerminated`.) If the owner *did* use `spawn`, the `linkDead` message is ignored and gets moved into `m_localBox`.

These are never removed. It's not apparent since the message box is private to `Tid`, but a thread that spawns 1000 non-linked threads over its lifetime will have its `m_localBox` full of 1000 leaked `linkDead` control messages from long-dead threads. These naturally have to be iterated past to get to any real messages due to how linked lists work.

This illustrates the issue:

```d
unittest
{
    // Clean slate since other unittests will have left stray messages
    thisInfo.ident = Tid(new MessageBox);

    foreach (immutable i; 0..100)
    {
        spawn({});
    }

    thread_joinAll();
    assert(thisTid.mbox.m_sharedBox.length == 100);  // 100 linkDead control messages
    assert(thisTid.mbox.m_localBox.length == 0);

    bool receivedSomething;

    // Trigger move to m_localBox
    receivedSomething = receiveTimeout(0.seconds, (Variant v) {});  // "catch any message"
    //assert(receivedSomething);  // fails
    assert(thisTid.mbox.m_sharedBox.length == 0);  // Empty, all moved to m_localBox
    assert(thisTid.mbox.m_localBox.length == 100);  // Would be 99 if it was a linked thread, but as it is it is never received

    receivedSomething = receiveTimeout(0.seconds, (Variant v) {});
    //assert(receivedSomething);  // fails
    assert(thisTid.mbox.m_sharedBox.length == 0);
    assert(thisTid.mbox.m_localBox.length == 100);  // Would be 98, ditto

    receivedSomething = receiveTimeout(0.seconds, (Variant v) {});
    //assert(receivedSomething);  // fails
    assert(thisTid.mbox.m_sharedBox.length == 0);
    assert(thisTid.mbox.m_localBox.length == 100);  // Would be 97, ditto

    // ...
}
```

Should it outright remove `linkDead` messages that aren't from linked threads[6]? How would that play with nested threads? How can you tell if a message is defunct?

[1]: https://github.com/dlang/phobos/blob/6010a7cc/std/concurrency.d#L2263
[2]: https://github.com/dlang/phobos/blob/6010a7cc/std/concurrency.d#L2222-L2225
[3]: https://github.com/dlang/phobos/blob/6010a7cc/std/concurrency.d#L1931
[4]: https://github.com/dlang/phobos/blob/6010a7cc/std/concurrency.d#L2156
[5]: https://github.com/dlang/phobos/blob/6010a7cc/std/concurrency.d#L2006
[6]: https://github.com/dlang/phobos/blob/6010a7cc/std/concurrency.d#L2016-L2023

PR #7089, opened 2019-06-25
Adds asserts messages

burner:


std/algorithm/searching.d
```diff
@@ -848,7 +848,8 @@ if (isForwardRange!R
     }
 
     //Because of @@@8804@@@: Avoids both "unreachable code" or "no return statement"
-    static if (isInfinite!R) assert(0);
+    static if (isInfinite!R) assert(false, R.stringof ~ "must be an infinite"
```
thewilsonator: must _not_ be.

thewilsonator: also just `static assert(!isInfinite!R, msg)`

std/algorithm/searching.d
```diff
@@ -951,7 +952,8 @@ if (isInputRange!R &&
     }
 
     //Because of @@@8804@@@: Avoids both "unreachable code" or "no return statement"
-    static if (isInfinite!R) assert(0);
+    static if (isInfinite!R) assert(false, R.stringof ~ " must be an inifite"
```
thewilsonator: ditto.

std/algorithm/searching.d
burner: that will no compile, I tried. depending on the type of R that function will not have a return and an assert(false) at the end.

thewilsonator: Oh, I misread that as a static assert. My bad.

burner:
@thewilsonator I was a second to slow

std/algorithm/searching.d
```diff
@@ -848,7 +848,8 @@ if (isForwardRange!R
     }
 
     //Because of @@@8804@@@: Avoids both "unreachable code" or "no return statement"
-    static if (isInfinite!R) assert(0);
+    static if (isInfinite!R) assert(false, R.stringof ~ "must not be an"
```
shove70: static if (isInfinite!R) assert(false, R.stringof ~ " must not be an"
Lost a space. ^_^

burner:
@thewilsonator could you have a look?

thewilsonator:
Done.

burner:
that was quick, how do you do it? thanks

thewilsonator:
[magic](https://media2.giphy.com/media/12NUbkX6p4xOO4/giphy.gif?cid=790b76115d133d727034654d36fb1161&rid=giphy.gif)

burner:
yeah, that is what I expected :-)

thewilsonator:
You know me too well.

burner:
thanks

PR #7091, opened 2019-06-25
Adds assert messages to std.algorithm.sorting

burner:


std/algorithm/sorting.d
```diff
@@ -617,7 +617,8 @@ size_t pivotPartition(alias less = "a < b", Range)
 (Range r, size_t pivot)
 if (isRandomAccessRange!Range && hasLength!Range && hasSlicing!Range && hasAssignableElements!Range)
 {
-    assert(pivot < r.length || r.length == 0 && pivot == 0);
+    assert(pivot < r.length || r.length == 0 && pivot == 0, "pivot must be"
+        ~ " less then the length of r or r must be empty and pivot zero");
```
thewilsonator: > less then

less th_a_n

std/algorithm/sorting.d
```diff
@@ -1682,7 +1683,7 @@ private void shortSort(alias less, Range)(Range r)
             break;
     }
 
-    assert(r.length >= 6);
+    assert(r.length >= 6, "r must be longer than 5");
```
thewilsonator: r must have more than 5 elements

std/algorithm/sorting.d
```diff
@@ -1740,7 +1741,7 @@ Sorts the first 5 elements exactly of range r.
 */
 private void sort5(alias lt, Range)(Range r)
 {
-    assert(r.length >= 5);
+    assert(r.length >= 5, "r must be longer than 4 elements");
```
thewilsonator: ditto

std/algorithm/sorting.d
burner: this always happens to me.

burner: fixed

std/algorithm/sorting.d
burner: fixed

std/algorithm/sorting.d
burner: fixed

std/algorithm/sorting.d
thewilsonator: No it isn't.

std/algorithm/sorting.d
thewilsonator: ditto

std/algorithm/sorting.d
burner: it helps to commit, my bad

burner:
thanks, sorry for the noise

WalterBright:
@burner @thewilsonator please do not use `format` in assert messages. `format` expands to a truly epic blizzard of template instantiations, which causes lots of problems with debug and unittest builds. It also causes problems by making problems hard to isolated, due to depending on most of the rest of Phobos.

`assert` messages should be as lightweight as possible, and should not instantiate templates. `sprintf` can be used for formatting.

PR #7094, opened 2019-06-27
Fix enum type in std.socket to match the receiver type

RazvanN7:
This currently blocks [1]. The enum is declared as an integer, but it is assigned to ushort variables. This worked due to a bug in dmd which is now being fixed.

[1] https://github.com/dlang/dmd/pull/10099

### DIFF
--- a/std/experimental/allocator/building_blocks/stats_collector.d
+++ b/std/experimental/allocator/building_blocks/stats_collector.d
@@ -729,7 +729,7 @@ public:
         private void addPerCall(string f, uint n, names...)(ulong[] values...)
         {
             import std.array : join;
-            enum uint mask = mixin("Options."~[names].join("|Options."));
+            enum ulong mask = mixin("Options."~[names].join("|Options."));
             static if (perCallFlags & mask)
             {
                 // Per allocation info
--- a/std/parallelism.d
+++ b/std/parallelism.d
@@ -480,7 +480,7 @@ struct Task(alias fun, Args...)
         static if (isFunctionPointer!(_args[0]))
         {
             private enum bool isPure =
-            functionAttributes!(Args[0]) & FunctionAttribute.pure_;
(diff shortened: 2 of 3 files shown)


std/experimental/allocator/building_blocks/stats_collector.d
```diff
@@ -729,7 +729,7 @@ public:
         private void addPerCall(string f, uint n, names...)(ulong[] values...)
         {
             import std.array : join;
-            enum uint mask = mixin("Options."~[names].join("|Options."));
+            enum ulong mask = mixin("Options."~[names].join("|Options."));
```
RazvanN7: Assigning ulong to uint.

std/parallelism.d
```diff
@@ -480,7 +480,7 @@ struct Task(alias fun, Args...)
         static if (isFunctionPointer!(_args[0]))
         {
             private enum bool isPure =
-            functionAttributes!(Args[0]) & FunctionAttribute.pure_;
+            (functionAttributes!(Args[0]) & FunctionAttribute.pure_) != 0;
```
RazvanN7: Assigning uint to bool.

RazvanN7:
I had to fix 2 more cases.

thewilsonator:
Wait this should target stable.

RazvanN7:
It is ok to target master. https://github.com/dlang/dmd/pull/10099 cannot target stable.

std/experimental/allocator/building_blocks/stats_collector.d
Geod24: This is known at compile time, isn't it ? So it should not be a problem.

RazvanN7: The idea is that if the user has forced a type upon an enum, it should always be taken into account.

Geod24: That sounds like a very arbitrary limitation without apparent benefit.

PR #7095, opened 2019-06-28
Fix issue 20009 - isForwardRange doesn't work with alias range this o…

shove70:
…r inheritance

### DIFF
--- a/std/range/primitives.d
+++ b/std/range/primitives.d
@@ -975,8 +975,13 @@ object with `save` and using it later.
 See_Also:
     The header of $(MREF std,range) for tutorials on ranges.
  */
-enum bool isForwardRange(R) = isInputRange!R
-    && is(ReturnType!((R r) => r.save) == R);
+template isForwardRange(R)
+{
+    static if (isInputRange!R && is(typeof((R r) => r.save) RT))
+        enum bool isForwardRange = is(R : ReturnType!RT);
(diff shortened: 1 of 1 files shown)


shove70:
Since struct S inherits from struct FR, I think it should have the characteristics of ForwardRange too.
But unfortunately, after I do it according to issue reporter idea, although struct S passed the isForwardRange test, it caused a lot of compilation errors in the entire Phobos, including function code and assertions in unit tests. The reasons for this include at least the following:

1. is(T1 : T2) is somewhat different than what I have understood before:
    is(R!(int) : int[]) will pass the test, obviously this is not what is expected here.
2. Too many templates and APIs assume that the return type of save() is exactly equal to itself, and cannot be equal to the base class. It is equal to the base class that will overturn these assumptions and cause the program to fail or assert the failure. Such as:

    std.std.algorithm.searching.d line 3359:
    Range least = range.save;

    Obviously, the base type cannot be assigned to a derived type, and this assumption has limited the expectations of this issue.

Therefore, if we want isForwardRange to support derived types, then Phobos has too many places to change, and the workload is a bit large.

It might be a good idea to add a new template to the existing one that is specifically designed to test derived types with ForwardRange characteristics:

template isDerivedForwardRange(R)
{
    ...
}

But I don't know if it has practical use.

PR #7096, opened 2019-06-30
Make locally-defined `FiberScheduler` actually usable

Geod24:
So... Do people actually use this module ?

I have a test library which spawns `Thread` to simulate Vibe.d servers. Within each node, I need to replicate a simple, re-entrant event loop. I tried to use a local `scheduler` (thread-local) as I don't want to mess with a global instance (why do we even have one to begin with ? The user can define that on their own if they need it...), but it seems this use case hasn't been tested.

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -1377,7 +1377,7 @@ private:
                  !notified && !period.isNegative;
                  period = limit - MonoTime.currTime)
             {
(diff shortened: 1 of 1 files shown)


Geod24:
This was a case were commits were actually atomic and independent... So why did you use squash ?

PR #7099, opened 2019-07-04
Replace D1 operator overloads with D2 operator overloads - Part 2

JinShil:
Follow up to https://github.com/dlang/phobos/pull/7098  (I missed one)

In support of https://github.com/dlang/dmd/pull/10130


thewilsonator:
Er that seems like an awful lot of build kite breakage.

thewilsonator:
Hmm, seems to be network issues.

JinShil:
All green now.

PR #7100, opened 2019-07-04
Fix Issue 19105 - Bogus recursive template expansion via getSymbolsByUDA

RazvanN7:
Issue 19105 was reported as a compiler bug. The reported code is:

```d
module junk;

import std.traits;

struct A(Args...) {}
struct B {}
alias C = A!( getSymbolsByUDA!(junk, B) );
```

But reading the docs of `getSymbolsByUDA`is seems that it was meant to work only with aggregate declarations (not with modules). I added a template constraint that requires that the symbol passed is an aggregated type.

The code fails because while getting the members of `junk`, the innards of `getSymbolsByUDA` uses aliases for the list of symbols and at some point alias C ends up aliasing to something that aliases itself.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -8353,6 +8353,7 @@ Note:
     nested structs or unions.
  */
 template getSymbolsByUDA(alias symbol, alias attribute)
+if (isAggregateType!symbol)
 {
     alias membersWithUDA = getSymbolsByUDAImpl!(symbol, attribute, __traits(allMembers, symbol));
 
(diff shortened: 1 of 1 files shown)


thewilsonator:
This should have that as a test then.

RazvanN7:
@thewilsonator ok. done.

Biotronic:
This is the wrong fix for 19105. 19105 is in fact a DMD issue, as can be showed by this code:

```
import std.traits;

struct A(Args...) {}

struct S1 {
    // Recursive template expansion
    alias C = A!( getSymbolsByUDA!(S1, A) );
}
struct S2 {
    // precompute the result:
    alias tmp = getSymbolsByUDA!(S2, A);
    // No recursive template expansion
    alias C = A!( getSymbolsByUDA!(S2, A) );
}
```

The problem seems to be with caching in the compiler, and the only thing this PR has changed is it's made valid use cases fail to compile.

Here's a reduced test case for the behavior, which doesn't use `getSymbolsByUDA`:

```
template test(alias T) {
    enum i = __traits(compiles, __traits(getMember, T, "C"));
    alias test = int;
}

struct A(Args...) {}

struct S1 {
    // Recursive template expansion
    alias C = A!(test!S1);
}
struct S2 {
    // No error
    alias a = test!S2;
    alias C = A!(test!S2);
}
```

ZombineDev:
CC @RazvanN7

zorael:
Any chance of getting this reverted in time for 2.087.1? It broke my whole thing.

zorael:
Ping. Still in 2.088 beta

PR #7101, opened 2019-07-04
Fix issue 19834 - File exception for [std.file.copy] on windows shows the target file rather than the source file

shove70:


std/file.d
```diff
@@ -4218,11 +4218,26 @@ private void copyImpl(scope const(char)[] f, scope const(char)[] t,
...
-            if (!t)
-                t = to!(typeof(t))(toz[0 .. wcslen(toz)]);
+            // Reference resources: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-copyfilew
+            immutable uint errno = .GetLastError();
+            const(char)[] name;
+
+            if (!f)
+                f = to!(typeof(f))(fromz[0 .. wcslen(fromz)]);
+
+            if ((errno == ERROR_PATH_NOT_FOUND) || (errno == ERROR_FILE_NOT_FOUND) ||
```
CyberShadow: I think `ERROR_PATH_NOT_FOUND` and `ERROR_ACCESS_DENIED` could be caused by a problem with either the source path or destination path (not existing or not accessible).

Could `ERROR_FILE_NOT_FOUND` be caused when copying over a dangling symlink?

I think it would be simpler to show both filenames in all cases when the OS API call is handling both paths at once.

shove70: Thank you, I have already done it. Indeed, the information returned by the OS is limited, and it is not easy to try to handle it very accurately.

shove70:
@thewilsonator, excuse me, May I bother you? I want to ask you a question:

```d
/// issue 19834
version (Windows) @safe unittest
{
    import std.exception : collectException;
    import std.algorithm.searching : startsWith;
    import std.format : format;

    auto f = deleteme;
    auto t = f ~ "2";
    auto ex = collectException(copy(f, t));
    version (Windows) assert(ex.msg.startsWith(format!"Copy from %s to %s"(f, t)));
}

```
It's a little strange here. Why do me have to write `version (Windows)` again in `unittest` to pass the test? From the information of build log, In Linux OS, the code of `version (Windows) @safe unittest` module is also running.

Thank you!

thewilsonator:
Don't know, it looks as though you shouldn't.

shove70:
Thank you. Before the last push, I didn't add version (Windows) to it, and I couldn't pass the test.
Now it's all green, or delete it and push it again to see how it works.

thewilsonator:
You don't need to keep deleting:

```
git reset HEAD~1
# (make your changes)
git add files # or use whatever method you usually use e.g. GUI
git commit -m "message goes here" # ditto
git push -f
```

shove70:
OK, push again.
ci/circleci tests failed.

```
generated/linux/debug/64/publictests/std_file.d(4353): [unittest] unittest failure
core.exception.AssertError@generated/linux/debug/64/publictests/std_file.d(4353): unittest failure
----------------
src/core/exception.d:447 [0x5585eace0da3]
src/core/exception.d:464 [0x5585eace0e25]
src/core/exception.d:608 [0x5585eacb03ec]
src/core/exception.d:613 [0x5585eacb042d]
src/core/exception.d:603 [0x5585eacb03a5]
generated/linux/debug/64/publictests/std_file.d:4353 [0x5585eac763e6]
??:? [0x5585eacad700]
src/core/runtime.d:629 [0x5585eace13f3]
src/object.d:2335 [0x5585eacaf6e8]
src/rt/minfo.d:777 [0x5585eacb5db8]
src/rt/sections_elf_shared.d:82 [0x5585eacb5e48]
src/rt/minfo.d:770 [0x5585eacb5d31]
src/object.d:2335 [0x5585eacaf6bd]
src/core/runtime.d:617 [0x5585eace11c3]
src/rt/dmain2.d:509 [0x5585eacb2681]
src/rt/dmain2.d:483 [0x5585eacb25fc]
src/rt/dmain2.d:542 [0x5585eacb2562]
…

```

std/file.d
```diff
@@ -4332,6 +4340,19 @@ private void copyImpl(scope const(char)[] f, scope const(char)[] t,
     assert(readText(t) == "a");
 }
 
+/// issue 19834
```
CyberShadow: Here is the problem. The three `///` mark this as a public example. You need to use two `//` to turn it into an internal unit test.

```suggestion
// issue 19834
```

shove70: Oh, thank you very much. I'll try.

CyberShadow:
Probably the same should be done with `rename`, `link`, and other OS APIs which take more than one file path argument.

shove70:
> Probably the same should be done with `rename`, `link`, and other OS APIs which take more than one file path argument.

Ok, I will handle it as a whole :)

PR #7102, opened 2019-07-04
Turn on -preview=markdown

dgileadi:
This PR does two things:

1. Adds the `-preview=markdown` flag to the Phobos Ddoc build.
2. Fixes places in the docs that were unintentionally interpreted as markdown.

The fixes were generally for:

* Lines that start with `-`, `+` or `*`, which in Markdown starts a list item.
* Backslashes that aren't in code blocks, and so are treated as escaping the next character.
* Words in brackets that aren't in code blocks, which Ddoc treats as [reference links](https://dlang.org/spec/ddoc.html#reference_links) to symbols with the same name.

### DIFF
--- a/etc/c/sqlite3.d
+++ b/etc/c/sqlite3.d
@@ -38,7 +38,7 @@ import core.stdc.config : c_ulong;
 extern (C) __gshared nothrow:
 
 /**
-** CAPI3REF: Compile-Time Library Version Numbers
+* CAPI3REF: Compile-Time Library Version Numbers
 */
 enum SQLITE_VERSION = "3.28.0";
 /// Ditto
@@ -47,7 +47,7 @@ enum SQLITE_VERSION_NUMBER = 3028000;
 enum SQLITE_SOURCE_ID = "2019-04-16 19:49:53 884b4b7e502b4e991677b53971277adfaf0a04a284f8e483e2553d0f83156b50";
 
 /**
-** CAPI3REF: Run-Time Library Version Numbers
+* CAPI3REF: Run-Time Library Version Numbers
 */
 extern immutable(char)* sqlite3_version;
 /// Ditto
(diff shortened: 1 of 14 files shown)


etc/c/sqlite3.d
```diff
@@ -1,35 +1,35 @@
...
+* if experience from use "in the wild" suggest such changes are prudent.
+*
+* The official C-language API documentation for SQLite is derived
+* from comments in this file.  This file is the authoritative source
+* on how SQLite interfaces are suppose to operate.
+*
+* The name of this file under configuration management is "sqlite.h.in".
+* The makefile makes some minor changes to this file (such as inserting
+* the version number) and changes its name to "sqlite3.h" as
+* part of the build process.
```
CyberShadow: This is a regular comment, not a doc comment. Why are these changes necessary?

etc/c/sqlite3.d
```diff
@@ -2094,49 +2094,49 @@ enum
...
+* 2010 August 30
+*
+* The author disclaims copyright to this source code.  In place of
+* a legal notice, here is a blessing:
+*
+*    May you do good and not evil.
+*    May you find forgiveness for yourself and forgive others.
+*    May you share freely, never taking more than you give.
+*
+************************************************************************
```
CyberShadow: Ditto (and more in the same file)

std/regex/package.d
```diff
@@ -136,7 +136,7 @@ $(TR $(TD Objects) $(TD
     $(REG_ROW \W, Matches any non-word character.)
     $(REG_ROW \s, Matches whitespace, same as \p{White_Space}.)
     $(REG_ROW \S, Matches any character except those recognized as $(I \s ). )
-    $(REG_ROW \\, Matches \ character. )
+    $(REG_ROW \\\\, Matches \ character. )
```
CyberShadow: It would have been nice to make it so that it would be possible to write DDoc that would render the same with or without `-preview=markdown`, but I guess such changes are unavoidable in practice. Main concern is tools which consume DDoc directly, such as DDox, or editors/IDEs which parse DDoc to display formatted documentation in their UI.

etc/c/sqlite3.d
dgileadi: Unnecessary—fixed.

12345swordy:
ping 3 days are up.

PR #7103, opened 2019-07-06
std.experiment.allocator: use Allocator.expand when available for expandArray

ErnyTech:
Some allocators may have optimized calls for expanding the memory 
block instead of the classic reallocation, so expandArray will use 
Allocator.expand instead of Allocator.reallocate.

### DIFF
--- a/std/experimental/allocator/package.d
+++ b/std/experimental/allocator/package.d
@@ -2133,7 +2133,7 @@ bool expandArray(T, Allocator)(auto ref Allocator alloc, ref T[] array,
     if (array is null) return false;
     immutable oldLength = array.length;
     void[] buf = array;
-    if (!alloc.reallocate(buf, buf.length + T.sizeof * delta)) return false;
+    if (!implExpandArray(alloc, buf, T.sizeof * delta)) return false;
     array = cast(T[]) buf;
     array[oldLength .. $].uninitializedFillDefault;
     return true;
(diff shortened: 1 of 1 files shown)


thewilsonator:
That should be a function, rather than 3 duplicated sections of code.

std/experimental/allocator/package.d
```diff
@@ -2263,6 +2263,17 @@ if (isInputRange!R)
     assert(arr == [1, 2, 3, 1, 2, 3, 4]);
 }
 
+private bool implExpandArray(Allocator)(auto ref Allocator alloc, ref void[] array,
+        size_t delta)
+{
+    static if (__traits(hasMember, Allocator, "expand"))
+    {
+        return alloc.expand(buf, T.sizeof * delta);
+    } else {
```
lesderid: Braces and `else` should be on their own lines (https://dlang.org/dstyle.html#phobos_brackets)

shove70:
@thewilsonator, excuse me, is `unittest` necessary for any fix, enhancement or adjustment? Or when we all think we don't need `unittest`, we don't need to write it? thanks.

thewilsonator:
This change is essentially a fast path optimisation, but one that is provided by the user (or library allocator)in the form of an explicit expand capability (rather than simply reallocate with a larger buffer). The individual components should already be tested and assuming they perform to spec, this change should only result in a performance difference.

So, TL;DR its difficult to test, and a test doesn't provide much value in this case.

shove70:
Thanks, I see  :)

PR #7106, opened 2019-07-12
Fix issue 20043: support Tuple.rename on const tuples

FeepingCreature:


thewilsonator:
Issue number / changelog entry? Also this is pretty trivial you may as well target stable.

FeepingCreature:
Hm, didn't bother. Gonna file one and retrofit/rebase to stable.

FeepingCreature:
Buildkite failure seems unrelated.

Basile-z:
yes, it's related to the fact that some old D1 operators are still present in phobos.

PR #7107, opened 2019-07-13
Get make!T to compile when T is a shared value type

thaven:
Making a shared value object currently fails with a message 'template std.conv.emplaceRef cannot deduce function from argument types...'. This PR fixes that.


thewilsonator:
Is there an issue number for this? If not, please make or or add a changelog entry.

thewilsonator:
DAutoTest and circle fail due to code.dlang.org being down.

PR #7109, opened 2019-07-15
std.math: change tests for rounding

WalterBright:
Changing floating point rounding modes, because it happens off to the side, interferes with optimizing. In particular, common subexpressions may not be common if the rounding mode is changed.

Solutions are:

1. disable CSEs for floating point (what dmd used to do). Not doing CSEs for floating point is kinda sad.
2. make a compiler intrinsic for changing the rounding mode, so the optimizer knows about it, this would be a fair amount of rewrite both for the optimizer and std.math.
3. change the code that mucks with the rounding mode so it won't interfere with optimization

This does option (3).

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -1501,7 +1501,8 @@ if (isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R) &&
             return CloseHandle(hObject);
         }
         static auto trustedSetFileTime(HANDLE hFile, const scope FILETIME *lpCreationTime,
-                                       const scope ref FILETIME lpLastAccessTime, const scope ref FILETIME lpLastWriteTime) @trusted
+                                       const scope ref FILETIME lpLastAccessTime,
+                                       const scope ref FILETIME lpLastWriteTime) @trusted
         {
             return SetFileTime(hFile, lpCreationTime, &lpLastAccessTime, &lpLastWriteTime);
         }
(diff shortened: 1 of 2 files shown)


WalterBright:
Option (2) remains a problem even with intrinsics, because if the intrinsic is wrapped in a function call, the optimizer still won't know it happened.

WalterBright:
This problem was exposed by https://github.com/dlang/dmd/pull/10177 which improves CSE computation for DMD, and is blocking it.

ZombineDev:
Option (2) sounds like the best one in the long-term. Obviously you prefer a short-term solution which doesn't block your work, but I'm not sure about the conferences of (3). What concerns me is that a change in dmd's optimizer necessitates a change in user's code and I am not sure I fully understand the consequences. Does it mean that all code that changes the FP rounding mode inside one function body will be broken with the new optimizations? If not, then why this change is necessary, after all it is changing the optimization of unittest code, which doesn't seem actually important (enough so to block other work).

WalterBright:
> Does it mean that all code that changes the FP rounding mode inside one function body will be broken with the new optimizations?

It may be broken already. Optimizations that involve constant folding will be done in the compiler with round to nearest, which won't do the same thing at runtime if the rounding mode is set differently. Basically, you're going to have to know what you're doing when manipulating the rounding mode. Fortunately, few people ever do it.

WalterBright:
And as I noted, a compiler intrinsic probably won't be effective anyway.

WalterBright:
The buildkite weka error seems unrelated.

PR #7110, opened 2019-07-15
Replace usages of `in` with `const scope`

JinShil:
In support of https://github.com/dlang/dmd/pull/10179

`in` as a parameter storage class is defined as `scope const`.  However `in` has not yet
been properly implemented so its current implementation is equivalent to `const`.  Properly
implementing `in` now will likely break code, so it is recommended to avoid using `in`, and
explicitly use `const` or `scope const` instead, until `in` is properly implemented.

The use of `in` as a parameter storage class is already discouraged in the documentation.  See https://dlang.org/spec/function.html#parameters

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -1500,8 +1500,8 @@ if (isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R) &&
         {
             return CloseHandle(hObject);
         }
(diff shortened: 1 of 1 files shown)


JinShil:
I think the buildkite Mecca failure is unrelated.  See https://github.com/dlang/ci/pull/397

PR #7112, opened 2019-07-19
Fix issue 20064,20069 - format functions make their final width and digits grouping conform to POSIX definition

shove70:
The format functions of FloatingPointType and IntegralType are adjusted and optimized to make their Final width and digits grouping conform to POSIX definition.


shove70:
@thewilsonator, It's all green   ^_^

PR #7114, opened 2019-07-23
Fix issue 20070 - std.conv.toImpl matches two functions when using st…

shove70:
…atic arrays as inout(T)


thewilsonator:
Please target stable.

shove70:
@thewilsonator, thanks, done.

PR #7116, opened 2019-07-28
std.process: add missing check for failed malloc()

kubo39:


std/process.d
```diff
@@ -555,6 +560,7 @@ private Pid spawnProcessImpl(scope const(char[])[] args,
                         close(i);
                     }
                 }
+                free(pfds);
```
CyberShadow: Why? The current process is either getting replaced by `execve`, or killed by exiting with an error.

kubo39: Ah yes, remove this.

std/process.d
```diff
@@ -510,9 +511,9 @@ private Pid spawnProcessImpl(scope const(char[])[] args,
             if (!(config & Config.inheritFDs))
             {
                 // NOTE: malloc() and getrlimit() are not on the POSIX async
-                // signal safe functions list, but practically this should not
-                // be a problem. Tha Java VM and CPython also use malloc() in
-                // its own implementation.
+                // signal safe functions list, but practically this should
+                // not be a problem. Tha Java VM and CPython also use malloc()
```
n8sh: "Tha"

kubo39:
How can I retest buildkite?

wilzbach:
You need to be part of the Buildkite D team (otherwise there's only the old `commit --amend` trick). I just restarted it for you, but if you ping me on Slack with your email, I can add you to Buildkite. Of course, traditional mail instead of Slack works too.

kubo39:
Thanks!

wilzbach:
Restarted it again as Baszile has removed his GitHub account and his packages are no longer available.

WalterBright:
Given that you found this bug in Phobos, their may be more lurking. Please grep and check all instances of `malloc`, `calloc` and `realloc`.

std/process.d
```diff
@@ -98,6 +98,9 @@ version (Windows)
     import std.windows.syserror;
 }
 
+import core.exception : OutOfMemoryError;
+
+import std.exception : enforce;
```
CyberShadow: Sorry, why were the local imports moved to a global import? I think that's opposite to current Phobos policy.

kubo39: I didn't care about, thank you for remind! Fixed at https://github.com/dlang/phobos/pull/7150

PR #7117, opened 2019-07-29
std.range.enumerate should work with a non-mutable start index

Geod24:
```
The deduced type for `Enumerator` could contain type modifier
(`const`, `immutable`), but the type was used directly as
a state variable.
```


thewilsonator:
DIff is trivial, please target stable.

Geod24:
Targeting stable now

PR #7119, opened 2019-07-31
Appender and RefAppender: use opSlice() instead of data()

WalterBright:
When building an array, the natural way to get the array is with opSlice(), not data().

### DIFF
--- a/changelog/array-opslice.dd
+++ b/changelog/array-opslice.dd
@@ -1,0 +1,8 @@
+std.array.Appender and RefAppender: use .opSlice() instead of data()
+
+Previonsly, Appender.data() was used to extract a slice of the Appender's array.
+Now use the [] slice operator instead.
+The same goes for RefAppender.
+
+$(REF Appender, std, array)
+$(REF RefAppender, std, array)
--- a/std/array.d
+++ b/std/array.d
@@ -3271,9 +3271,18 @@ if (isDynamicArray!A)
     }
 
     /**
+     * Use opSlice() from now on.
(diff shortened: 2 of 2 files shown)


changelog/array-opslice.dd
```diff
@@ -0,0 +1,8 @@
+std.array.Appender and RefAppender: use .opSlice() isntead of data()
```
wilzbach:
```suggestion
std.array.Appender and RefAppender: use .opSlice() instead of data()
```

changelog/array-opslice.dd
```diff
@@ -0,0 +1,8 @@
...
+
+Previonsly, Appender.data() was used to extract a slice of the Appender's array.
+Now use the [] slice operator instead.
+The same goes for RefAppender.
+
+$(REF Appender, std, array)
+#(REF RefAppender, std, array)
```
wilzbach:
```suggestion
$(REF RefAppender, std, array)
```

std/array.d
```diff
@@ -3774,12 +3783,20 @@ if (isDynamicArray!A)
         return impl.capacity;
     }
 
-    /**
-     * Returns the managed array.
+    /* Use opSlice() instead.
+     * Returns: the managed array.
      */
     @property inout(ElementEncodingType!A)[] data() inout
```
wilzbach: Deprecation?

WalterBright: I'll leave the deprecation for the future. For starters, I simply un-documented data().

PR #7121, opened 2019-07-31
Fix issues 19696, 19697, 16132 - ReplaceType bugs

pbackus:
ReplaceType no longer gives incorrect results for types that can be
implicitly converted to a templated type, either via inheritance (19697,
16132) or alias this (19696).

An attempt to address this issue was made in PR #6880, but it did not fix the inheritance-related issues, and appears to have been abandoned by its author.

These changes are based on pbackus/sumtype@b764942.


thewilsonator:
Please target stable.

pbackus:
The buildkite failure is due to availability issues with code.dlang.org, so there shouldn't be anything blocking this.

PR #7124, opened 2019-08-05
Allows passing by reference through spawn* if is a shared type

ErnyTech:
It is often useful to pass a mutable by reference type to the spawned
thread, the need to have to use a raw pointer before this patch made
the syntax boring and not in line with the rest of the DLang style
(use pointer as little as possible).

With this patch it will be possible to write code like this:
```
import core.thread : thread_joinAll;

static void f1(ref shared(int) number)
{
  atomicOp!"+="(number, 1);
}

shared(int) number = 10;
spawn(&f1, number);
thread_joinAll();
assert(number == 11);
```

Instead of the boring:
```
import core.thread : thread_joinAll;

static void f1(ref shared(int)* number)
{
  atomicOp!"+="(*number, 1);
}

shared(int) number = 10;
spawn(&f1, &number);
thread_joinAll();
assert(number == 11);
```

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -106,6 +106,30 @@ private
         static assert(!hasLocalAliasing!(SysTime, Container));
     }
 
+    bool hasLocalReferencePassing(F, Types...)()
+    {
+        bool doesIt = false;
+        static foreach (i, T; Types)
+        {
+            static if (!is(T == shared) && !is(T == Tid))
+            {
+                static if (ParameterStorageClassTuple!F[i] == ParameterStorageClass.ref_)
+                {
+                    doesIt = true;
+                }
+            }
+        }
+
+        return doesIt;
+    }
+
+    @safe unittest
+    {
+        static void fun(ref int) {}
+        static assert(hasLocalReferencePassing!(typeof(fun), int));
+        static assert(!hasLocalReferencePassing!(typeof(fun), shared(int)));
+    }
+
     enum MsgType
     {
         standard,
@@ -457,11 +481,13 @@ private template isSpawnable(F, T...)
  *  pointer indirection.  This is necessary for enforcing isolation among
  *  threads.
  */
-Tid spawn(F, T...)(F fn, T args)
+Tid spawn(F, T...)(F fn, auto ref T args)
 if (isSpawnable!(F, T))
 {
+    import std.functional : forward;
     static assert(!hasLocalAliasing!(T), "Aliases to mutable thread-local data not allowed.");
-    return _spawn(false, fn, args);
+    static assert(!hasLocalReferencePassing!(F, T), "Passing by reference thread-local data is not allowed.");
+    return _spawn(false, fn, forward!args);
 }
 
 ///
@@ -513,6 +539,36 @@ if (isSpawnable!(F, T))
     assert(receivedMessage == "Hello World");
 }
 
+@system unittest
+{
+    import core.atomic : atomicOp;
+    import core.thread : thread_joinAll;
+    import std.stdio;
+
+    static void f1(ref shared(int) number)
+    {
+        atomicOp!"+="(number, 1);
+    }
+
+    shared(int) number = 10;
+    spawn(&f1, number);
+    thread_joinAll();
+    assert(number == 11);
+}
+
+@system unittest
+{
+    import core.atomic : atomicOp;
+
+    static void f1(ref shared(int) number)
+    {
+        atomicOp!"+="(number, 1);
+    }
+
+    int number = 10;
+    static assert(!__traits(compiles, spawn(&f1, number)));
+}
+
 /**
  * Starts fn(args) in a logical thread and will receive a LinkTerminated
  * message when the operation terminates.
@@ -532,19 +588,23 @@ if (isSpawnable!(F, T))
  * Returns:
  *  A Tid representing the new thread.
  */
-Tid spawnLinked(F, T...)(F fn, T args)
+Tid spawnLinked(F, T...)(F fn, auto ref T args)
 if (isSpawnable!(F, T))
 {
+    import std.functional : forward;
     static assert(!hasLocalAliasing!(T), "Aliases to mutable thread-local data not allowed.");
-    return _spawn(true, fn, args);
+    static assert(!hasLocalReferencePassing!(F, T), "Passing by reference thread-local data is not allowed.");
+    return _spawn(true, fn, forward!args);
 }
 
 /*
  *
  */
-private Tid _spawn(F, T...)(bool linked, F fn, T args)
+private Tid _spawn(F, T...)(bool linked, F fn, auto ref T args)
 if (isSpawnable!(F, T))
 {
+    import std.functional : forward;
+
     // TODO: MessageList and &exec should be shared.
     auto spawnTid = Tid(new MessageBox);
     auto ownerTid = thisTid;
@@ -553,7 +613,7 @@ if (isSpawnable!(F, T))
     {
         thisInfo.ident = spawnTid;
         thisInfo.owner = ownerTid;
-        fn(args);
+        fn(forward!args);
     }
 
     // TODO: MessageList and &exec should be shared.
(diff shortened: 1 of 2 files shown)


ErnyTech:
no one wants to review this?

ErnyTech:
CC @wilzbach

std/concurrency.d
```diff
@@ -455,13 +479,27 @@ private template isSpawnable(F, T...)
...
  */
+
 Tid spawn(F, T...)(F fn, T args)
 if (isSpawnable!(F, T))
 {
     static assert(!hasLocalAliasing!(T), "Aliases to mutable thread-local data not allowed.");
+    return _spawnnoref(false, fn, args);
+}
+
+Tid spawn(F, T...)(F fn, ref T args)
```
atilaneves: Doesn't this mean that all arguments must be passed by ref? What if one wants to pass some by ref and some not?

ErnyTech: Yes they are all passed as ref, then when the function (fn) is called the arguments without ref will be passed as a value

atilaneves: Why not `auto ref`?

ErnyTech: I had problems using auto ref but I don't remember exactly what, it's been too long
I think that today I reproach with auto ref to check again

ErnyTech: Ok the problem of using auto ref is that when calling spawn (F, T ...) (F fn, ref T args) with a rvalue this is passed as a value and is ok but then when spawn calls _spawn (F, T ...) (bool linked, F fn, auto ref T args) at this point the rvalue is passed as a reference

Here I gave an example of what happens: https://run.dlang.io/is/yMHf8r

Not well I know what could be done to avoid this, perhaps combine _spawn and spawn or manually generate templates to avoid passing as a reference in _spawn

atilaneves: That's because after it's been bound to a name (a variable) it comes an lvalue. When dealing with `auto ref`, you need to use `std.functional.forward` to pass it along to another function.

ErnyTech: Ah nice to know, I didn't know the existence of forward, ok then I'll rewrite the PR
Thank you for teaching me something new

ErnyTech:
@atilaneves Now do you think it can be approved?

std/concurrency.d
```diff
@@ -106,6 +106,30 @@ private
         static assert(!hasLocalAliasing!(SysTime, Container));
     }
 
+    bool hasLocalReferencePassing(F, Types...)()
```
edi33416: Probably this should be made private. Otherwise it needs to be ddoc'ed.

std/concurrency.d
```diff
@@ -530,17 +586,19 @@ if (isSpawnable!(F, T))
...
-    return _spawn(true, fn, args);
+    static assert(!hasLocalReferencePassing!(F, T), "Passing by reference thread-local data is not allowed.");
+    return _spawn(true, fn, forward!args);
 }
 
 /*
  *
  */
-private Tid _spawn(F, T...)(bool linked, F fn, T args)
+private Tid _spawn(F, T...)(bool linked, F fn, auto ref T args)
```
edi33416: Should the `fn(args)` [call](https://github.com/dlang/phobos/pull/7124/files#diff-25bf2af9a45af72e9499b850a56d6f32R612) (L612 in changed) from `exec` also forward ( `fn(forward!args);` )?

std/concurrency.d
ErnyTech: Yes, it's under this private block https://github.com/dlang/phobos/pull/7124/files#diff-25bf2af9a45af72e9499b850a56d6f32R73

std/concurrency.d
ErnyTech: oh true! I'm going to fix

ErnyTech:
circleci has failed for `dmd/src/dmd/dmsc.d:110: undefined reference to `out_config_debug(bool, bool, bool, bool, bool, bool, bool)` is really weird

ZombineDev:
If you rebase to master the error should go away. I fixed it in https://github.com/dlang/phobos/pull/7160.

ErnyTech:
Ok now circleci is green, thanks!

ZombineDev:
I think the pointer-based syntax has the advantage of signifying to the reader that the spawned function will access by reference the arguments.

With concurrency one should be extra careful about the objects' lifetime.
GC-allocated class objects have relatively easier to manage lifetime, so passing them implicitly by reference is not a big problem. However stack allocated objects have much shorter lifetime which makes it especially easy to get into undefined behavior by escaping their address. Of course technically references have no more problems than pointers, but for human readers it is also easier to miss when a piece of code shares the address of an object.

Consider the following:
```D
void main()
{
    int var;
    f();
}

void f()
{
    // run asynchronously, without immediately awaiting it to finish
    shared(int) number = 10;
    spawn(&f1, number); // passed by value before
                        // after: accesses an expired stack frame from another thread
}

static void g(ref shared(int) number)
{
  atomicOp!"+="(number, 1);
}
```

ErnyTech:
I do not believe that a reader has particular difficulties in understanding what is happening when he reads "ref" in the parameters

thewilsonator:
This seems to result in a consistent failure in excel-d. 
```
ut.wrap.wrap.wrapAsync:
--
  | tests/ut/wrap/wrap.d:30 - Expected: 6.400000
  | tests/ut/wrap/wrap.d:30 -      Got: nan

ut.wrap.module_.wrapModuleFunctionStr async double -> double:
--
  | tests/ut/wrap/module_.d:620 - Expected: 6.000000
  | tests/ut/wrap/module_.d:620 -      Got: nan
```

@atilaneves

ErnyTech:
How about making a spawnRef function to make everyone happy and avoid any kind of code breaks?

ErnyTech:
I believe there is a logical error in excel-d, this function (https://github.com/kaleidicassociates/excel-d/blob/91034ca039979f945828820e2eafd817963a7a1e/source/xlld/wrap/wrap.d#L544) contains ref in the parameters but  we know that before this PR could not be passed by ref with spawn, in my opinion this is what causes the test to fail

However, this keeps me thinking that maybe spawnRef is the best solution, at least you don't risk breaking code

ZombineDev:
I'd be okay with having a `spawnRef` that took all of its arguments by `ref`. No need for `auto ref` as now there's `-preview=rvaluerefparam` to account for rvalues.

ErnyTech:
Alternative PR: https://github.com/dlang/phobos/issues/7167

atilaneves:
I think that spawn should work with functions that take their arguments by ref. It's a limitation that doesn't make any sense to me at the moment. I don't think that explicitly adding a `spawnRef` function makes much sense.

ErnyTech:
@atilaneves any advice to solve the problem with excel-d?

ZombineDev:
@atilaneves what about the example I provided in https://github.com/dlang/phobos/pull/7124#issuecomment-526283116? We can't allow such brutally unsafe things, while other languages such as Rust focus on concurrency safety, in addition to memory safety.

atilaneves:
> @atilaneves any advice to solve the problem with excel-d?

@ErnyTech I haven't had time to look into it. I'll try tomorrow but if not then I'm on holiday for 3 weeks.

atilaneves:
> @atilaneves what about the example I provided in [#7124 (comment)](https://github.com/dlang/phobos/pull/7124#issuecomment-526283116)? We can't allow such brutally unsafe things, while other languages such as Rust focus on concurrency safety, in addition to memory safety.

@ZombineDev How is the `ref` version less safe than the pointer one? I understand the argument about how `ref` makes it invisible to the caller that the argument may be modified; it's been made before in the Google C++ style guide and in Rust (I remain unconvinced, possibly because I make everything and its sibling `const` and have never written a bug due to it). That however seems to be an odd argument to make in a language that already has `ref`.

Lifetime problems need to be dealt with the same way regardless if an argument is `ref T` or `T*`; your example is equally buggy when using a pointer. That needs to be fixed (although `spawn` et al are all @system now anyway).

RazvanN7:
Filed: https://issues.dlang.org/show_bug.cgi?id=21848

Closing.

PR #7125, opened 2019-08-05
Clarify that buf can be shortened in readln

mmtrebuchet:
It wasn't clear on inspection if buf could be shortened, or if it was just lengthened. "As necessary" suggested that buf would only be resized if absolutely necessary, which would only be when it needs to be lengthened.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -1688,7 +1688,8 @@ $(REF newline, std,ascii) for portability (unless the file was opened in
 text mode).
 
 Returns:
-0 for end of file, otherwise number of characters read (i.e., buf.length)
+0 for end of file, otherwise number of characters read.
+The return value will always be equal to `buf.length`.
 
 Throws: `StdioException` on I/O error, or `UnicodeException` on Unicode
 conversion error.
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -1682,13 +1682,13 @@ must copy the previous contents if you wish to retain them.
 
 Params:
 buf = Buffer used to store the resulting line data. buf is
-resized as necessary.
+resized to exactly contain the line.
 terminator = Line terminator (by default, `'\n'`). Use
 $(REF newline, std,ascii) for portability (unless the file was opened in
 text mode).
 
 Returns:
-0 for end of file, otherwise number of characters read
+0 for end of file, otherwise number of characters read (i.e., buf.length)
 
 Throws: `StdioException` on I/O error, or `UnicodeException` on Unicode
 conversion error.
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -1682,7 +1682,7 @@ must copy the previous contents if you wish to retain them.
 
 Params:
 buf = Buffer used to store the resulting line data. buf is
-resized to exactly contain the line.
+enlarged if necessary, then set to the slice exactly containing the line.
 terminator = Line terminator (by default, `'\n'`). Use
 $(REF newline, std,ascii) for portability (unless the file was opened in
 text mode).


CyberShadow:
So, if you want to use `readln` in such a way that it allocates only when it sees a line longer than any other line it has seen before (which I would assume is the intended use case for `readln` with a buffer), that makes the included example wrong, right? Passing a `buf` that was shortened by a previous call to `readln` will cause it to reallocate it to avoid clobbering the part of it that it doesn't see (past the array end), so you need to manually check and maintain the "longest buffer" in a separate variable.

mmtrebuchet:
@CyberShadow I think it just adjusts buf.length but not the actual buf.capacity, so that if your lines are of length 20, 10, 15, it'd allocate 20 characters, then just set buf.length to 10 and 15, without ever shortening the underlying storage. (I'm not sure if this is the *intended* behavior, but this PR is just to make the documentation reflect the behavior of the function.)

CyberShadow:
Here is what I mean:

https://run.dlang.io/is/UwJsaG

After shortening `buf` to 10, it is reallocated when reading the third, 15-character line.

mmtrebuchet:
Oh, I didn't notice that. Yeah, that's a bad thing, I think. Definitely *not* how I'd expect that function to behave.
I'm not sure why buf.capacity would ever be shortened, but if you change the code to have 
`	f.readln(buf); writeln(buf.ptr); writeln(buf.capacity); `
(I'm not sure how you got the link you posted; where's the button for that?)
then capacity is shortened after the second line is read. In fact, it's shortened to 0 capacity.  
I'm stumped by (I've not studied stdio.d carefully) `ReadlnAppender.data()`, line 4980 in stdio.d:
```
    @property char[] data() @trusted
    {
        if (safeAppend)
            assumeSafeAppend(buf.ptr[0 .. pos]);
        return buf.ptr[0 .. pos];
    }
```

I've never tried to slice a pointer until today. Is this where the reallocation is occurring?

CyberShadow:
> Yeah, that's a bad thing, I think.

I agree, it looks like the function interface was poorly designed. It should have been returning the slice, and only growing `buf`, never shrinking it.

> (I'm not sure how you got the link you posted; where's the button for that?)

"Shorten" in the upper-left corner.

>  then capacity is shortened after the second line is read. In fact, it's shortened to 0 capacity.

The problem is that it's returning (and setting `buf` to) a slice of the input buffer. Slices cannot have a capacity, as appending to them will always reallocate.

> Is this where the reallocation is occurring?

It happens in the third `readln`. Because the `buf` it receives is a slice, appending to it will reallocate. It wouldn't reallocate if it were to call `assumeSafeAppend`, but it doesn't know that the `buf` the user provided is one where the data past its end is safe to overwrite.

std/stdio.d
```diff
@@ -1682,13 +1682,13 @@ must copy the previous contents if you wish to retain them.
 
 Params:
 buf = Buffer used to store the resulting line data. buf is
-resized as necessary.
+resized to exactly contain the line.
```
CyberShadow:
```suggestion
enlarged if necessary, then set to the slice exactly containing the line.
```

std/stdio.d
```diff
@@ -1682,13 +1682,13 @@ must copy the previous contents if you wish to retain them.
...
 terminator = Line terminator (by default, `'\n'`). Use
 $(REF newline, std,ascii) for portability (unless the file was opened in
 text mode).
 
 Returns:
-0 for end of file, otherwise number of characters read
+0 for end of file, otherwise number of characters read (i.e., buf.length)
```
CyberShadow:
```suggestion
0 for end of file, otherwise number of characters read.
The return value will always be equal to `buf.length`.
```

mmtrebuchet:
Urk. I guess we're stuck with this strange behavior. A shame that getline() hasn't made its way to core.stdc.stdio yet.

PR #7126, opened 2019-08-09
ieeeFlags unittest: the optimizer might remove operations that the te…

rainers:
…st is depending upon
The optimizer removes all traces of `a` because it is const-folded to `nan`. That means no side effects appear.

This test is failing here: https://dev.azure.com/dlanguage/dmd/_build/results?buildId=3982&view=logs&jobId=995050be-48e9-58a0-3553-d9a547fe6c20&taskId=3bcde2a0-396a-5c5e-c7d5-d0568bb0ee4b&lineStart=4275&lineEnd=4276&colStart=1&colEnd=1

Not sure why it doesn't appear on other platforms, but it triggers when using dmd built against the MS C runtime.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -5690,15 +5690,17 @@ public:
         static void func() {
             int a = 10 * 10;
         }
-
+        pragma(inline, false) static void blockopt(ref real x) {}
         real a = 3.5;
         // Set all the flags to zero
         resetIeeeFlags();
         assert(!ieeeFlags.divByZero);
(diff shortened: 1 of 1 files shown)


rainers:
Just noticed that there are two more unittests that almost do the same. Amended similar fixes, but they could also just be deleted (they don't look like useful examples with these workarounds).

rainers:
Ping! Any reason for holdup?

PR #7129, opened 2019-08-13
start making use of isAutodecodableString()

WalterBright:
I don't know why the `!isAggregate()` needs to be there, I suspect it is a mistake, but this is a refactoring not a redesign.

### DIFF
--- a/std/range/primitives.d
+++ b/std/range/primitives.d
@@ -380,7 +380,7 @@ void put(R, E)(ref R r, E e)
     }
     //Optional optimization block for straight up array to array copy.
     else static if (isDynamicArray!R &&
-                    !(autodecodeStrings && isNarrowString!R) &&
+                    !(isAutodecodableString!R && !isAggregateType!R) &&
                     isDynamicArray!E &&
                     is(typeof(r[] = e[])))
     {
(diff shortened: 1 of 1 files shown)


thewilsonator:
It appears to be because https://github.com/dlang/phobos/blob/135f58080ef3126d378ce92de1e2f8c33ec3ab78/std/traits.d#L6681 a struct with an alias this to a string is considered autodecodeable.

WalterBright:
> a struct with an alias this to a string is considered autodecodeable.

Right, but I reproduced this behavior exactly:
`autodecodeStrings && isNarrowString!R)` is identical to `isAutodecodableString!R && !isAggregateType!R`

WalterBright:
Where is this file being generated?
```
generated/linux/debug/64/publictests/std_range_primitives.d(1108): Error: template instance `isAutodecodableString!R` template `isAutodecodableString` is not defined
```
and it should import `std.traits`

WalterBright:
Yeah, but why does circleci fail and the other tests pass?

thewilsonator:
I think those are the doc examples. Maybe it does doc testing?

WalterBright:
Thanks, that's the clue I needed.

PR #7130, opened 2019-08-13
find Phobos dependencies on autodecode

WalterBright:
To get rid of autodecode, first we must make Phobos independent of whether autodecoding is on or off. So turn it off and see what breaks.

Fixes should be in separate PRs labelled with [no autodecode]. They should be one of:

1. use https://dlang.org/phobos/std_utf.html#byChar
2. use https://dlang.org/phobos/std_utf.html#byDchar
3. hardcode decoding

### DIFF
--- a/std/range/primitives.d
+++ b/std/range/primitives.d
@@ -2412,7 +2412,7 @@ if (isAutodecodableString!(T[]) && !isAggregateType!(T[]))
 Autodecoding is enabled if this is set to true.
 */
 
-enum autodecodeStrings = true;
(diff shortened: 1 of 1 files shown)


thewilsonator:
> So turn it off and see what breaks.

This should eventually go under a `version` that we can set defaults on

WalterBright:
Added [no autodecode] label. Please add this to all PRs that fix autodecoding to work with no autodecoding.

LightBender:

PR #7131, opened 2019-08-13
Fix Issue 20129 - AA require with File values gives "Error: `return` …

CyberShadow:
…expression expected" in object.d


CyberShadow:
There's a bunch more of `opAssign` returning `void` in Phobos, may be worth looking at.

PR #7132, opened 2019-08-14
Update x86 assembly for DMD windows x64 ABI

SSoulaimane:
Required by https://github.com/dlang/dmd/pull/10280.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -3228,12 +3228,15 @@ creal expi(real y) @trusted pure nothrow @nogc
     {
         version (Win64)
         {
+            enum offset_ = creal.sizeof / 2;
             asm pure nothrow @nogc
             {
                 naked;
-                fld     real ptr [ECX];
+                fld     real ptr [RDX];
                 fsincos;
-                fxch    ST(1), ST(0);
+                fstp    real ptr [RCX];
+                fstp    real ptr [RCX+offset_];
+                mov     RAX,RCX;
                 ret;
             }
         }


wilzbach:
Looks like dscanner has troubles parsing this assembly :/

SSoulaimane:
It does now.

kinke:
Isn't this a pessimization, enforcing sret for `creal`? IIRC, the Win64 ABI doesn't specify the behavior for x87 `real` (unused by MSVC and outright forbidden in kernel mode), so DMD would be free to choose.

SSoulaimane:
It's a struct greater than one byte. The [spec](https://dlang.org/spec/abi.html#function_calling_conventions) doesn't make any exceptions.

kinke:
That spec is ridiculous:
```
The extern (C) and extern (D) calling convention matches the C calling convention used by the supported C compiler on the host system. Except that the extern (D) calling convention for Windows x86 is described here.
```

So `extern(D)` matching the C ABI on all non-32-bit-Windows platforms? Of course not, parameter reversal being one obvious difference.

The [MS spec](https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=vs-2019) does mention (although in the context of arguments):
> The x87 register stack is unused

So yeah, should be fine.

SSoulaimane:
If you can edit the spec, otherwise I'm planning to change everything to conform to what the spec says.
It doesn't matter if DMD chooses to use the FPU or not as long as calling `extern(D)` functions can be made form C without resorting to hand crafted assembly. I guess this was the motive behind that decision.

kinke:
> change everything to conform to what the spec says

You mean you're going to undo the parameter reversal? Oh, be me guest, I've been wanting that for a long time.

SSoulaimane:
Yes, good then.

RazvanN7:
this version of `expi` has been removed from phobos (it was deprecated at the time of the patch anyway). It is now implemented in `std.complex` and no longer uses inline assembly. Therefore this PR is no longer necessary.

PR #7133, opened 2019-08-15
fix array(String) to work with no autodecode

WalterBright:
With autodecode turned off, the return value wasn't a `dchar[]`.

First fix for https://github.com/dlang/phobos/pull/7130

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -260,7 +260,7 @@ Returns:
     a `dchar[]`, `const(dchar)[]`, or `immutable(dchar)[]` depending on the constness of
     the input.
 */
-ElementType!String[] array(String)(scope String str)
+CopyTypeQualifiers!(ElementType!String,dchar)[] array(String)(scope String str)
 if (isNarrowString!String)
 {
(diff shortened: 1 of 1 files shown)


std/array.d
```diff
@@ -260,15 +260,15 @@ Returns:
...
+CopyTypeQualifiers!(ElementType!String,dchar)[] array(String)(scope String str)
 if (isNarrowString!String)
 {
     import std.utf : toUTF32;
     auto temp = str.toUTF32;
     /* Unsafe cast. Allowed because toUTF32 makes a new array
        and copies all the elements.
      */
-    return () @trusted { return cast(ElementType!String[]) temp; } ();
+    return () @trusted { return cast(CopyTypeQualifiers!(ElementType!String, dchar)[]) temp; } ();
```
thewilsonator: Make one of these auto. That is a horrible return type to spell out twice.

thewilsonator: Gah I meant `typeof(return)`

PR #7134, opened 2019-08-15
fix assocArray() unittests for no autodecode

WalterBright:
The type of the assocArray winds up different.

https://github.com/dlang/phobos/pull/7130

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -535,6 +535,7 @@ if (isInputRange!Values && isInputRange!Keys)
 {
     import std.range : repeat, zip;
     import std.typecons : tuple;
+    import std.range.primitives : autodecodeStrings;
     auto a = assocArray(zip([0, 1, 2], ["a", "b", "c"])); // aka zipMap
     static assert(is(typeof(a) == string[int]));
     assert(a == [0:"a", 1:"b", 2:"c"]);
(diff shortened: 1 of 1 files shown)


WalterBright:
Seems spurious that std.array unittests would affect buildkite?

thewilsonator:
I don't think it should but 
```
core.exception.AssertError@/var/lib/buildkite-agent/builds/ci-agent-06d7ec31-635b-4e0c-b810-f59d64e4393e-7/dlang/phobos/build/distribution/bin/../imports/std/range/package.d(6373): Assertion failure
```

cc @BlackEdder

lesderid:
Just looks like unfortunate random data hitting an assert due to floating point error:

> `source/ggplotd/stat.d:604 void ggplotd.stat.__unittest_L590_C1() [0xd69269]
`

https://github.com/BlackEdder/ggplotd/blob/19ad94ec78a89ffd8e76794c03de38b5faca30a4/source/ggplotd/stat.d#L598-L604

> `/var/lib/buildkite-agent/builds/ci-agent-06d7ec31-635b-4e0c-b810-f59d64e4393e-7/dlang/phobos/build/distribution/bin/../imports/std/range/package.d:6373 pure ref @safe std.range.iota!(double, double, double).iota(double, double, double).Result std.range.iota!(double, double, double).iota(double, double, double).Result.__ctor(double, double, double) [0xb6db7e]
`

 https://github.com/dlang/phobos/blob/5b83b74bf11418b85d0dae7fda1f4ada061ac441/std/range/package.d#L6373

std/array.d
```diff
@@ -587,11 +592,16 @@ if (isInputRange!Values && isInputRange!Keys)
...
 
     auto r = "abcde".enumerate.filter!(a => a.index == 2);
     auto a = assocArray(r.map!(a => a.value), r.map!(a => a.index));
     assert(is(typeof(a) == size_t[dchar]));
-    assert(a == [dchar('c'): size_t(2)]);
+    static if (autodecodeStrings)
+        alias achar = dchar;
+    else
+        alias achar = immutable(char);
+    assert(a == [achar('c'): size_t(2)]);
```
CyberShadow: How about this instead (shorter and less DRY):
```d
assert(a == [ElementType!string('c'): size_t(2)]);
```

WalterBright: It's a good idea for user code, but for a test case I think it's better to see what the actual type is rather than a proxy for it.

PR #7136, opened 2019-08-15
adds assert messages to std.concurrency

burner:


std/concurrency.d
```diff
@@ -186,7 +186,8 @@ private
     {
         foreach (i, t1; T)
         {
-            static assert(isFunctionPointer!t1 || isDelegate!t1);
+            static assert(isFunctionPointer!t1 || isDelegate!t1,
+                    "Not all passed T are function pointer or delegates");
```
thewilsonator: function pointer_s_. Also you have the index of the T that is not a function pointer or delegate, you may as well give that as part of the message.

std/concurrency.d
```diff
@@ -198,7 +199,8 @@ private
 
                 foreach (t2; T[i + 1 .. $])
                 {
-                    static assert(isFunctionPointer!t2 || isDelegate!t2);
+                    static assert(isFunctionPointer!t2 || isDelegate!t2,
+                            "Not all passed T are function pointer or delegates");
```
thewilsonator: this static assert is redundant with the one above, no?

std/concurrency.d
burner: done

std/concurrency.d
burner: done

burner:
@thewilsonator thanks

PR #7138, opened 2019-08-15
std.format: Fix compilation with no autodecode

CyberShadow:
The entire instantiated template hierarchy there actually can't compile or work with a narrow char type, because at some point it wants to decode an escape, which will fit only in a `dchar` anyway.

For https://github.com/dlang/phobos/pull/7130


CyberShadow:
There's more instances of these; need to look at how to fix this with the least changes.

CyberShadow:
There is a runtime unittest failure remaining in std.format, but the lead goes into std.uni, so I suspect it's the same problem as with std.regex / std.net.curl.

PR #7139, opened 2019-08-15
std.file: Fix unittest for no autodecode

CyberShadow:


WalterBright:
part of no autodecode project: https://github.com/dlang/phobos/pull/7130

PR #7140, opened 2019-08-15
std.stdio: Fix failing unittest with no autodecode

CyberShadow:
The code compiled and ran, but behaved differently.

The silent breakage was due to the implicit conversion of `char` to `dchar`. Forbidding such conversions would prevent the silent breakage.


WalterBright:
Part of the https://github.com/dlang/phobos/pull/7130 initiative.

PR #7142, opened 2019-08-15
Rewrite dubious refRange() shortcuts in unittests

kinke:
As preparation for dlang/dmd#10124, which makes elements of array literals rvalues, so this trick for converting a string literal to an lvalue doesn't work anymore.


wilzbach:
Don't you need this on stable then?

kinke:
I'm not sure whether stable is still appropriate (for the DMD PR) - due to breaking changes. I hoped to get it done before 2.088, but now stable has already branched..

PR #7144, opened 2019-08-15
fix Issue 20134 - autodecode should use replacementDchar rather than throwing on invalid

WalterBright:
https://issues.dlang.org/show_bug.cgi?id=20134

### DIFF
--- a/changelog/replacementDchar.dd
+++ b/changelog/replacementDchar.dd
@@ -1,0 +1,31 @@
+# Change autodecode from throwing `UTFException` on invalid encodings to returning `std.utf.replacementDchar` `('\uFFFD')` instead.
+
+When `char` and `wchar` slices are iterated, they are autodecoded meaning the UTF-8 and
+UTF-16 code unit sequences are automatically consumed and decoded into `dchar` code points.
+If an invalid sequence is encountered, a
+$(LINK2 https://dlang.org/phobos/std_utf.html#UTFExceptions,`UTFException`)
+is raised.
+
+This change instead causes
+$(LINK2 https://dlang.org/phobos/std_utf.html#replacementDchar, `std.utf.replacementDchar`)
+to be the decoded result of an invalid sequence.
+
+If the old behavior is desired, the autodecoding iteration can be replaced with
+$(LINK2 https://dlang.org/phobos/std_utf.html#byUTF, `std.utf.byUTF!(dchar, UseReplacementDchar.no)`)
+to detect the invalid sequences and throw the exception, as in the following example:
+
+---
+import std.utf;
+import std.stdio;
+
+void test(string s)
+{
+    s.byDchar!(dchar, UseReplacementDchar.no)().write();
+}
+
+void main()
+{
+    test("hello");
+    test("\xFFbetty");
+}
+---
--- a/std/exception.d
+++ b/std/exception.d
@@ -1908,17 +1908,21 @@ pure @safe unittest
 ///
 pure @safe unittest
 {
-    import std.algorithm.comparison : equal;
-    import std.range : retro;
-    import std.utf : UTFException;
+    version (none) // autodecode no longer throws on invalid UTF sequences
+                   // issues.dlang.org/show_bug.cgi?id=20134
+    {
+        import std.algorithm.comparison : equal;
+        import std.range : retro;
+        import std.utf : UTFException;
 
-    auto str = "hello\xFFworld"; // 0xFF is an invalid UTF-8 code unit
+        auto str = "hello\xFFworld"; // 0xFF is an invalid UTF-8 code unit
 
-    auto handled = str.handle!(UTFException, RangePrimitive.access,
-            (e, r) => ' '); // Replace invalid code points with spaces
+        auto handled = str.handle!(UTFException, RangePrimitive.access,
+                (e, r) => ' '); // Replace invalid code points with spaces
 
-    assert(handled.equal("hello world")); // `front` is handled,
-    assert(handled.retro.equal("dlrow olleh")); // as well as `back`
+        assert(handled.equal("hello world")); // `front` is handled,
+        assert(handled.retro.equal("dlrow olleh")); // as well as `back`
+    }
 }
 
 /** Handle exceptions thrown from range primitives.
@@ -2201,17 +2205,21 @@ pure @safe unittest
 ///
 pure @safe unittest
 {
-    import std.algorithm.comparison : equal;
-    import std.range : retro;
-    import std.utf : UTFException;
+    version (none) // autodecode no longer throws on invalid UTF sequences
+                   // issues.dlang.org/show_bug.cgi?id=20134
+    {
+        import std.algorithm.comparison : equal;
+        import std.range : retro;
+        import std.utf : UTFException;
 
-    auto str = "hello\xFFworld"; // 0xFF is an invalid UTF-8 code unit
+        auto str = "hello\xFFworld"; // 0xFF is an invalid UTF-8 code unit
 
-    auto handled = str.handle!(UTFException, RangePrimitive.access,
-            (e, r) => ' '); // Replace invalid code points with spaces
+        auto handled = str.handle!(UTFException, RangePrimitive.access,
+                (e, r) => ' '); // Replace invalid code points with spaces
 
-    assert(handled.equal("hello world")); // `front` is handled,
-    assert(handled.retro.equal("dlrow olleh")); // as well as `back`
+        assert(handled.equal("hello world")); // `front` is handled,
+        assert(handled.retro.equal("dlrow olleh")); // as well as `back`
+    }
 }
 
 pure nothrow @safe unittest
--- a/std/range/primitives.d
+++ b/std/range/primitives.d
@@ -2498,10 +2498,10 @@ if (!isAutodecodableString!(T[]) && !is(T[] == void[]))
 @property dchar front(T)(scope const(T)[] a) @safe pure
 if (isAutodecodableString!(T[]))
 {
-    import std.utf : decode;
+    import std.utf;
     assert(a.length, "Attempting to fetch the front of an empty array of " ~ T.stringof);
     size_t i = 0;
-    return decode(a, i);
+    return decode!(UseReplacementDchar.yes)(a, i);
 }
 
 /**
@@ -2542,10 +2542,10 @@ if (!isAutodecodableString!(T[]) && !is(T[] == void[]))
 @property dchar back(T)(scope const(T)[] a) @safe pure
 if (isAutodecodableString!(T[]))
 {
-    import std.utf : decode, strideBack;
+    import std.utf;
     assert(a.length, "Attempting to fetch the back of an empty array of " ~ T.stringof);
     size_t i = a.length - strideBack(a, a.length);
-    return decode(a, i);
+    return decode!(UseReplacementDchar.yes)(a, i);
 }
 
 /*


WalterBright:
> Your PR doesn't reference any Bugzilla issue.

Yes it does.

CyberShadow:
> Yes it does.

It needs to be in the commit message. GitHub pull request titles don't end up in git history and remain only on GitHub.

WalterBright:
> It needs to be in the commit message.

I amended it. Still no joy.

CyberShadow:
Here is the discussion:
https://issues.dlang.org/show_bug.cgi?id=14519

CyberShadow:
Looks like it needed a reopen to see the new commit message.

CyberShadow:
I think the correct fix here is to throw an Error, not an Exception. Allowing strings (which should have been sanitized at the inputs) to get all the way to any autodecoding machinery is a program bug.

WalterBright:
Hmm, I had thought my PR to fix foreach() had gone through.

WalterBright:
Curiously, the foreach() throws a UnicodeException while std.utf throws UTFException. Right hand, meet left hand.

CyberShadow:
> Curiously, the foreach() throws a UnicodeException while std.utf throws UTFException. Right hand, meet left hand.

Looks like `UnicodeException` is in Druntime (as the `foreach` decoding would be) and `UTFException` is in `std.utf`.

WalterBright:
While your arguments that this change could result in data loss where some was relying on invalid UTF data and read then wrote it is pedantically true, I am doubtful it matters in practice. Consider:

1. toUTF uses replacementDchar and nobody notices
2. the same thing happens with invalid floating point values - they just get converted to NaN
3. text editors often change the file read and then written, even if nothing else changes. It'll, for example, reset how line endings are done, it'll truncate anything past a 0 or a ^Z in the file, etc. Nobody cares.
4. Normal practice in the industry is replacementDchar, not rejecting the entire string
5. even Errors are expensive in D
6. the fact that nobody has noticed UnicodeException and UTFException are both used for the same thing suggests that nobody depends on catching these

And lastly, we cannot remove autodecode and yet keep these exceptions.

CyberShadow:
> toUTF uses replacementDchar and nobody notices

Yep, but probably because no one's using D for anything too serious in that area, of they are they're not using Phobos.

>     3\. text editors often change the file read and then written, even if nothing else changes

I don't see how that's related! D programs can be processing all kinds of data using strings, and it may not be necessarily "text files" that you would open in a text editor. (Also decent text editors don't do that any more)

> Normal practice in the industry is replacementDchar, not rejecting the entire string

Is that actually true? Pretty sure that's not what Rust does, for example. I think the advice only applies when displaying strings, not when converting them in general.

> the fact that nobody has noticed UnicodeException and UTFException are both used for the same thing suggests that nobody depends on catching these

Maybe they don't need to be caught! The program crashing with UnicodeException / UTFException is a good signal that there is something wrong with the input and it needs to be fixed.

> even Errors are expensive in D

That's bad but I don't know how to fix it. For something as low-level as implicit autodecoding maybe we can do something like `onOutOfMemory` (which isn't too different from an `abort()`).

CyberShadow:
BTW, a problem with the current design of `byUTF` / `decode` / etc: When it returns `replacementDchar`, how can you tell if it was because of bad UTF in its input, or because the input string literally contained a `'\uFFFD'` character? A better return value to signal bad UTF would be some value that's unencodable in a valid UTF stream, such as `cast(char)0xFF`...

wilzbach:
I'm in favor of this change in general as it unblocks a common problem: string algorithms are hard to do in nothrow and @nogc (dip1008 is still not a thing). However, we at least need to provide a range that allows opt-in validation (i.e. expose the current front/back etc.) to give users a choice.

CyberShadow:
We can throw errors in nothrow and @nogc. `onOutofMemory` works from anywhere.

WalterBright:
> A better return value to signal bad UTF would be some value that's unencodable in a valid UTF stream, such as cast(char)0xFF...

It's best to stick with convention, which is simply substitute in the replacement char.

CyberShadow:
It's your language, boss.

Don't forget to create a changelog entry if you insist on this one.

WalterBright:
> Don't forget to create a changelog entry

done

> if you insist on this one.

I don't think there's another path that leads to consistency and sanity in how character strings are handled in Phobos.

changelog/replacementDchar.dd
```diff
@@ -0,0 +1,42 @@
...
+void test(string s)
+{
+    s.byDchar()
+     .filter!((c)
+              {
+                  if (c == replacementDchar)
+                      throw new UTFException("invalid UTF-8 encoding");
+                  return true;
+              }) ()
+     .write();
```
wilzbach: This is too complicated for a breaking change.
It needs to be simpler for an easy migration path:

```d
private struct ValidationRange(R)
{
    R r;
    this(R r)
    {
        this.r = r;
    }

    @property dchar front() @safe pure
    {
        import std.utf;

        assert(r.length, "Attempting to fetch the front of an empty array of " ~ R.stringof);
        size_t i = 0;
        return decode!(UseReplacementDchar.no)(r, i);
    }

    @property dchar back(T)(scope const(T)[] a) @safe pure
    {
        import std.utf;

        assert(r.length, "Attempting to fetch the back of an empty array of " ~ T.stringof);
        size_t i = r.length - strideBack(a, r.length);
        return decode!(UseReplacementDchar.no)(r, i);
    }

    alias r this;

}

auto validationRange(T)(scope const(T)[] a)
if (isAutodecodableString!(T[]) && !isAggregateType!(T[]))
{
    return ValidationRange!(typeof(a))(a);
}
```

https://run.dlang.io/is/MgKafm

WalterBright: A simpler scheme is to just encapsulate the filter into its own range, validate(), i.e.:
```
s.byDchar().validate().write();
```
which would work the same whether or not autodecode is present, which is where we want to get to anyway.

wilzbach: Sure, though that will be a bit slower (more comparisons + more ranges).
Anyhow, my point was that if we do such a breaking change the solution to revert back to the current behavior must be simple. When we removed `delete`, we had `__delete` as an alternative. Whatever it does, a solution here needs to be similarly easy to reach.

WalterBright: Something like this:
```
import std.utf;
import std.algorithm;
import std.stdio;

void test(string s)
{
    s.byDchar()
     .validateDchar()
     .write();
}

void main()
{
    test("hello");
    test("\xFFbetty");
}

auto validateDchar(R)(R r)
{
     return filter!((c)
              {
                  if (c == replacementDchar)
                      throw new UTFException("invalid UTF-8 encoding");
                  return true;
              })(r);
}
```

WalterBright:
buildkite is failing at the link step - suggesting it is trying to link with another library that needs to be recompiled as well.

changelog/replacementDchar.dd
```diff
@@ -0,0 +1,41 @@
...
+
+This change instead causes
+$(LINK2 https://dlang.org/phobos/std_utf.html#replacementDchar, `std.utf.replacementDchar`)
+to be the decoded result of an invalid sequence.
+
+If the old behavior is desired, the autodecoding iteration can be replaced with
+$(LINK2 https://dlang.org/phobos/std_utf.html#byDchar, `std.utf.byDchar()`)
+to decode into `dchar`s followed by
+$(LINK2 https://dlang.org/phobos/std_algorithm_iteration.html#.filter, `std.algorithm.filter()`)
+to detect the `replacementDchar` and throw the exception, as in the following example:
```
CyberShadow: Please add some text here how this approach will not work with any valid UTF streams that already contain the replacement character, and they will be indistinguishable from invalid streams.

changelog/replacementDchar.dd
```diff
@@ -0,0 +1,41 @@
...
+import std.algorithm;
+import std.stdio;
+
+void test(string s)
+{
+    s.byDchar()
+     .filter!((c)
+              {
+                  if (c == replacementDchar)
+                      throw new UTFException("invalid UTF-8 encoding");
```
CyberShadow:
```suggestion
                      throw new UTFException("invalid UTF-8 encoding or replacement character in input");
```

CyberShadow:
Should we make `std.utf.validate` throw if it detects the replacement character in input too? It's not an uncommon character... but, if we don't, then we open D programs up for a situation where they validate the input, assume it's valid, and if this approach is used somewhere else to revalidate it then it crashes.

CyberShadow:
Maybe we can change the *value* of `replacementDchar`? It would work except if users have been hard-coding its value, or using it to intentionally put it in streams as the Unicode replacement character.

WalterBright:
> Maybe we can change the value of replacementDchar?

It's a standard:

https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character

https://hsivonen.fi/broken-utf-8/

I.e. the correct thing to do is whatever everyone else does.

changelog/replacementDchar.dd
WalterBright: done

changelog/replacementDchar.dd
WalterBright: done

WalterBright:
@CyberShadow a better solution for what existing autodecoding users can do: https://github.com/dlang/phobos/pull/7146

CyberShadow:
> I.e. the correct thing to do is whatever everyone else does.

I don't think it's right to say that in that context for three reasons:

1. `replacementDchar` is defined as "Inserted in place of invalid UTF sequences." This is the contract we have bound ourselves to. If we change its value then we will not have broken that contract per se.

2. If `replacementDchar` is redefined to an unencodable value (e.g. `0xFFFFFFFF`), then any attempts to encode it will throw. As such we would not have deviated from common practice 

3. I don't think we established that this is actually common practice. What other languages silently convert bad characters to the replacement character by default?

WalterBright:
> As such we would not have deviated from common practice which is throwing on bad UTF.

"An initial reaction was to design decoders to throw exceptions on errors.[17] This can turn what would otherwise be harmless errors (producing a message such as "no such file") into a denial of service. Early versions of Python 3.0 would exit immediately if the command line or environment variables contained invalid UTF-8,[18] making it impossible to handle such errors. The inability to deal with UTF-8 without first confirming it was valid actually greatly impeded adoption of Unicode.
Modern practice is to replace errors with a replacement character, and to ensure that systems do not interpret these replacement characters in any dangerous way[citation needed]. The errors can be detected later when it is convenient to report an error, or display as blocks when the string is drawn for the user."

https://en.wikipedia.org/wiki/UTF-8#Invalid_byte_sequences

CyberShadow:
Convenient of you to leave out all the "citation needed" annotations :)

Since you brought up Python:

https://docs.python.org/3/library/stdtypes.html#bytes.decode

The default behavior is "strict", i.e. throwing.

CyberShadow:
@WalterBright  Why did you leave this part out?

> ''The Unicode Standard'' requires decoders to "...treat any ill-formed code unit sequence as an error condition. This guarantees that it will neither interpret nor emit an ill-formed code unit sequence."

WalterBright:
> Why did you leave this part out?

Because that's not common practice.

> The default behavior is "strict", i.e. throwing.

And users found it unworkable.

> Convenient of you to leave out all the "citation needed" annotations :)

I removed one for brevity and to keep the focus as it was a long quote, and would have removed the other if I'd noticed it. I didn't quote the entire section, either. That's what the link I provided is for, if you want to see the context in situ and are suspicious of my quote.

> The default behavior is "strict", i.e. throwing.

Not surprising since Python is older than D and that was once thought to be the best solution (as I did for D), and Python probably didn't want to change the behavior. The cite shows that Python users worked around the default solution to get a better solution.

CyberShadow:
> Because that's not common practice.

Why are you so sure? You said this many times, but what is backing this statement? I looked at a few programming languages and what I see is the complete opposite.

WalterBright:
> You said this many times, but what is backing this statement?

**Modern practice is to replace errors with a replacement character**

https://en.wikipedia.org/wiki/UTF-8#Invalid_byte_sequences

CyberShadow:
You realize you're 

1. quoting Wikipedia
2. the part you quoted has a [citation needed] tag
3. the part you quoted is no longer in the article.

WalterBright:
> the part you quoted is no longer in the article.

Because you deleted it. Sheesh.

lesderid:
Maybe we simply ought to refer to the Unicode Consortium's [own guidelines](https://www.unicode.org/versions/Unicode12.0.0/UnicodeStandard-12.0.pdf)?

5.3 Unknown and Missing Characters:
> This section briefly discusses how users or implementers might deal with characters that are not supported [...] An implementation should not attempt to interpret such code points. However, in practice, applications must deal with unassigned code points or private-use characters. [...] An implementation should not blindly delete such characters, nor should it unintentionally transform them into something else.

5.22 U+FFFD Substitution in Conversion:

> When converting text from one character encoding to another, a conversion algorithm may encounter unconvertible code units. This is most commonly caused by some sort of corruption of the source data, so that it does not correctly follow the specification for that character encoding. [...] When a conversion algorithm encounters such unconvertible data, the usual practice is either to throw an exception or to use a defined substitution character to represent the unconvertible data. In the case of conversion to one of the encoding forms of the Unicode Standard, the substitution character is defined as U+FFFD replacement character.

> [...]

> For conversion between different encoding forms of the Unicode Standard, “U+FFFD Substitution of Maximal Subparts” in Section 3.9, Unicode Encoding Forms defines a practice for the use of U+FFFD which is consistent with the W3C standard for encoding. It is useful to apply the same practice to the conversion from non-Unicode encodings to an encoding form of the Unicode Standard.

> This practice is more secure because it does not result in the conversion consuming parts of valid sequences as though they were invalid. It also guarantees at least one replacement character will occur for each instance of an invalid sequence in the original text. Furthermore, this practice can be defined consistently for better interoperability between different implementations of conversion.

Unless I'm misinterpreting something, this suggests that unless we're explicitly converting from another encoding to Unicode, the recommended practice is *not* to transform characters into something else (e.g. the replacement character).

CyberShadow:
Just in case it's not clear, I did not delete that text from the WP article to maliciously manipulate this discussion. The deleted text was unencyclopedic, and in violation of all three of Wikipedia's core content policies:

- **Neutral Point of View** - the text was quite biased in favor of using replacement characters for some reason.
- **Verifiability** - there were no sources to back up the usage of replacement characters as a favored, common, or recommended practice.
- **No Original Research** - aside from the stuff about replacement characters, there was a non-trivial number of uncited statements regarding how invalid UTF is handled or should be handled.

I was not the first person to notice the problems with the text, as could be seen from all the annotations in it. However, none of this is really relevant because (as students today can tell you) **Wikipedia is not meant to be cited**. At best, it is a hivemind's (*verifiable*) summary of other resources, and at worst it is just someone's unchecked ramblings.

WalterBright:
> Unless I'm misinterpreting something, this suggests that unless we're explicitly converting from another encoding to Unicode, the recommended practice is not to transform characters into something else (e.g. the replacement character).

Autodecoding **is** converting.

D users now have a choice with https://github.com/dlang/phobos/pull/7146

lesderid:
Actually, I did misinterpret the text in my last comment. It's talking about invalid _code points_, not invalid _code unit sequences_.

Nevertheless, while not a recommendation the Standard (3.9 Unicode Encoding Forms, p128) does mention this:

> An increasing number of implementations are adopting the handling of ill-formed subsequences as specified in the W3C standard for encoding to achieve consistent U+FFFD replacements.
> See: http://www.w3.org/TR/encoding/

lesderid:
OT:

> Just in case it's not clear, I did not delete that text from the WP article to maliciously manipulate this discussion.

While I agree with you that text without citations obviously shouldn't be the norm, it's not common practice to delete parts of an article that are unsourced but aren't "patently ridiculous". I.e. don't forget about [WP:AGF](https://en.wikipedia.org/wiki/WP:AGF) and try to find a source for the text instead. In case you find sources that are in contradiction with the original text, replace it and cite these sources. If you do delete text, it's [recommended](https://en.wikipedia.org/wiki/Wikipedia:Verifiability/Removal_of_Uncited_Material#How_to_remove_material) to at least add a section to the article's talk page. Simply deleting a paragraph is not an improvement.

CyberShadow:
Everywhere I look, I see that (unless the particular situation calls to do otherwise) the default or accepted practice is to signal an error on invalid UTF. Other programming languages do it. The standards recommend or demand it. It is very easy to find sources for this. I did not want to outright change the article to state the opposite of what it previously was implying. And, in this particular case, I disagree that removing that text did not improve the article.

WalterBright:
> Everywhere I look, I see that (unless the particular situation calls to do otherwise) the default or accepted practice is to signal an error on invalid UTF.

Except for @lesderid 's quote from the Unicode consortium.

As for being default practice, that could very well be legacy, which is outdated now.

Besides, D is hardly consistent with this. byUTF has been there for years, and did not throw. Nobody complained. Autodecoding does not work for ranges of char, which is why byUTF came about.

CyberShadow:
> Except for @lesderid 's quote from the Unicode consortium.

We must be reading it differently. None of the text he quoted recommends one practice over the other.

> As for being default practice, that could very well be legacy, which is outdated now.

This is a bold assumption. Do you have anything to back it up?

Rust is newer than D and raises errors on bad UTF. You are saying that the convention predates Rust?

> Besides, D is hardly consistent with this.

I agree. However, we are fixing the inconsistency it in the wrong direction.

> byUTF has been there for years, and did not throw. Nobody complained.

I have been pretty vocal that this is the wrong practice for five years now.

Let me turn that around. `foreach` throws when decoding UTF. Where are the complaints?

CyberShadow:
> Where are the complaints?

I looked. `UnicodeException` is what's thrown when `foreach` encounters bad UTF, right?

Here are the top three search results I get on the forum for `UnicodeException`:

1. https://forum.dlang.org/post/vkhugmslgbmlfupnvwma@forum.dlang.org

   The user is trying to load a CSV that is not (entirely) UTF-8. Without the exception, the mis-encoded data would have been corrupted and the errors would have propagated to the output.

2. https://forum.dlang.org/thread/cxlmmnjnqnaqlkilnned@forum.dlang.org

   The problem is a mis-use of Windows APIs. Without the exception, the operation would have silently succeeded and resulted in garbage.

3. https://forum.dlang.org/post/mvkbemmgatrfbxnmjada@forum.dlang.org

    The user explicitly wants to work with a "badly encoded" string. Never mind that `string`s should not contain bad UTF, or that `std.string.tr` should not autodecode. Though we don't know if the user cares about preserving non-UTF parts of the string in this case, decoding happens as part of a different operation, and the non-UTF parts would be replaced implicitly, without the user's consent.

Why is this still even an argument?

WalterBright:
> None of the text he quoted recommends one practice over the other.

It says both are acceptable.

> Do you have anything to back it up?

Yes, but you deleted it.

> Rust is newer than D and raises errors on bad UTF.

It offers both forms, not a default as far as I can tell.

> Where are the complaints?

They come in the form of poor performance, not being nothrow, and using the gc. These are general problems with using exceptions for reporting errors. Having them at the root of all string processing just makes for problems. It's not the string processing itself that is made slow, it's everything that calls it because they have to build exception unwinding sections for the RAII objects. Exception unwinding code turns off data flow optimizations, register allocation, etc.

I expect this is why Rust uses Option to report errors, and Herb Sutter has been working on a similar thing for C++.

The 1990's sales job of "zero cost exceptions" has turned out to be a snow job.

I anticipate that exceptions will retreat into legacy territory, and that Option types will become the preferred solution. We should get in front of this change before it drives over us. I've done a lot of work to remove gc dependencies from Phobos, and we should be working on ways to remove throwing exceptions, too.

WalterBright:
1. one wonders why the user was decoding the line in the first place. Unicode is not symmetric and doing input<=>output conversions is going to result in differences, and not just from replacement characters. (Normalization, etc.)
2. that was a buffer overflow. Not really fair, as the overflow bytes could have just as easily been valid ASCII
3. string.tr shouldn't be decoding strings, it's a bug in it.

CyberShadow:
> Yes, but you deleted it.

If your argument is standing solely on an uncited sentence on Wikipedia, well ...

> It offers both forms, not a default as far as I can tell.

Yes. But, to be fair, the functions are `from_utf8` and `from_utf8_lossy`, and not `from_utf8` and `from_utf8_non_lossy`. Due to the naming, the code explicitly needs to request the lossy version. We can't do that in D because autodecoding usually happens a few layers below.

> They come in the form of poor performance, not being nothrow, and using the gc.

All these are problems with D's model of handling exceptions, not with how auto-decoding treats bad UTF.

Why not handle it in the same way as an out-of-memory error, then? You don't need to be able to catch the exception, because you can achieve the same effect by requesting decoding through a means that signals errors in other ways.

CyberShadow:
>     1. Unicode is not symmetric and doing input<=>output conversions is going to result in differences, and not just from replacement characters. (Normalization, etc.)

What? No, it won't! There is exactly one valid encoding for any code point in UTF-8, UTF-16, and UTF-32. Normalization is completely unrelated to this.

> string.tr shouldn't be decoding strings, it's a bug in it.

That's completely beside the point. It could very well be any function that does need to auto-decode. Could be the same `tr` but with `dchar` arguments and `char[]` string, then it would be forced to decode.

WalterBright:
https://issues.dlang.org/show_bug.cgi?id=20140

Note how this change completely avoids the problem with invalid UTF sequences, pushing the decision up to the user to make an explicit choice.

WalterBright:
https://unicode.org/reports/tr15/

While I think normalization is a bug in Unicode's design, it certainly is a thing.

CyberShadow:
> Why not handle it in the same way as an out-of-memory error, then?

Here is a better analogy. Why not handle it in the same way that integer division by zero is handled? This situation is actually quite similar to a proposal to make integer division by zero result in zero:

- Would be more consistent with floating-point division, by never raising an error.
- Allows making arithmetic code actually-really-nothrow
- Some programming languages already do exactly this! Though, [not to great public acclaim](https://www.reddit.com/r/programming/comments/7al9s2/pony_a_programming_language_that_allows_dividing/).

> This leaves open what to do if `from` and `to` contain invalid UTF sequences if the conversion of them is necessary. The most pragmatic solution is to reject `from` and `to` arguments to `tr` that are not of the same UTF encoding as `str`.

I agree with you 100% here: **this is the correct solution**. However, most unfortunately, we cannot do this in D everywhere, because of how deeply ingrained auto-decoding is.

> While I think normalization is a bug in Unicode's design, it certainly is a thing.

Normalization is completely unrelated here because it happens on a different layer. Decoding, encoding, autodecoding work with and transform code units. Normalization transfroms code points.

WalterBright:
> I agree with you 100% here: this is the correct solution. However, most unfortunately, we cannot do this in D everywhere, because of how deeply ingrained auto-decoding is.

I'm not so sure about that. I managed to fix quite a bit of std.string so it did not do autodecoding, with no effect on the user (I don't remember why I didn't do `tr()`). I got rid of the autodecoding in std.path, too. We need to get rid of autodecoding one way or another, or D will get stuck in the past.

My evolved opinion is that an app should decide which encoding to use, and stick with it everywhere. Any decoding/encoding should be restricted to input/output.

My further evolved opinion is that D should not have been agnostic about character types, and just been UTF-8 everywhere. UTF-16 and UTF-32 should be treated as library types only.

WalterBright:
BTW https://issues.dlang.org/show_bug.cgi?id=20140 is a fine example of what I'm talking about with regard to defining an error condition out of existence so Phobos can be made nothrow and @nogc.

CyberShadow:
Agreed 100% there, that's great. Still, I disagree with the change in this pull request, because it allows mis-encoded data to be silently corrupted into unrecoverable errors and propagated before they're noticed, and because it is a change that will affect existing D programs. Please consider one of the alternative approaches discussed above.

WalterBright:
Ok I'll think about it. In the meantime, I could use help with the two failing tests - I have no idea what's wrong from reading the logs. Though the buildkite one might be caused by linking with an older library.

CyberShadow:
DAutoTest is failing due to a network error. Just rebase and force-push.

CyberShadow:
I don't know what's wrong with buildkite, looks like it died before it even started. Probably a rebase will fix it. CC @wilzbach

WalterBright:
just rebased

wilzbach:
It's very likely unrelated, but I would be careful with this PR here.
BTW with Buildkite you can also login and just restart the failing job in question, that might save time.

schveiguy:
Jumping in because of the `no-autodecode` tag, I'm starting an effort to remove autodecoding completely from Phobos as an option. See #7595 

FWIW I *did* run into an issue with `wstring.front` not being `@nogc`. But I worked around it by fixing `std.algorithm.comparison.equals`.

I like @CyberShadow's approach here -- make it throw an Error. But most of this is moot if we just get rid of autodecoding altogether, which is the approach I'm taking.

RazvanN7:
I took the liberty to rebase this. @WalterBright @CyberShadow should we merge this?

CyberShadow:
No, this is a breaking change that may result in silent data corruption. We've discussed this thoroughly. This change is also going against what modern languages are doing.

@RazvanN7 Sorry to waste your work, but I'm going to go ahead and close this :)

RazvanN7:
@CyberShadow No worries. I just wanted to get some resolution for this PR. Thanks!

CyberShadow:
Yeah, I hope we can put this behind this, since I know @WalterBright feels very strongly about this (because it simplifies things a lot on the implementation side). But, it really is a really bad compromise, and it would put us in a bad situation that we can't get out of easily.

PR #7145, opened 2019-08-16
UTFException should derive from UnicodeException

WalterBright:
UTFException should actually be replaced with UnicodeException, but there are some slight differences in behavior, sigh.


thewilsonator:
'std/utf.d(63): Error: module `std.exception` import `UnicodeException` not found'

PR #7146, opened 2019-08-18
std.utf.byUTF: add option to throw on invalid UTF

WalterBright:
Makes `byUTF()` more of a drop-in replacement for autodecoding.

### DIFF
--- a/std/utf.d
+++ b/std/utf.d
@@ -4177,15 +4177,24 @@ private int impureVariable;
  * Iterate an $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
  * of characters by char type `C` by encoding the elements of the range.
  *
- * UTF sequences that cannot be converted to the specified encoding are
+ * UTF sequences that cannot be converted to the specified encoding are either
  * replaced by U+FFFD per "5.22 Best Practice for U+FFFD Substitution"
- * of the Unicode Standard 6.2. Hence byUTF is not symmetric.
+ * of the Unicode Standard 6.2 or result in a thrown UTFException.
+ *  Hence byUTF is not symmetric.
  * This algorithm is lazy, and does not allocate memory.
  * `@nogc`, `pure`-ity, `nothrow`, and `@safe`-ty are inferred from the
  * `r` parameter.
  *
  * Params:
  *      C = `char`, `wchar`, or `dchar`
+ *      useReplacementDchar = UseReplacementDchar.yes means replace invalid UTF with `replacementDchar`,
+ *                            UseReplacementDchar.no means throw `UTFException` for invalid UTF
+ *
+ * Throws:
+ *      `UTFException` if invalid UTF sequence and `useReplacementDchar` is set to `UseReplacementDchar.yes`
+ *
+ * GC:
+ *      Does not use GC if `useReplacementDchar` is set to `UseReplacementDchar.no`
  *
  * Returns:
  *      A forward range if `R` is a range and not auto-decodable, as defined by
@@ -4198,7 +4207,7 @@ private int impureVariable;
  *
  *      Otherwise, an input range of characters.
  */
-template byUTF(C)
+template byUTF(C, UseReplacementDchar useReplacementDchar = Yes.useReplacementDchar)
 if (isSomeChar!C)
 {
     static if (!is(Unqual!C == C))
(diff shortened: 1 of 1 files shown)


std/utf.d
```diff
@@ -4385,4 +4386,12 @@ if (isSomeChar!C)
     "𐐷".byUTF!char().equal([0xF0, 0x90, 0x90, 0xB7]);
     "𐐷".byUTF!wchar().equal([0xD801, 0xDC37]);
     "𐐷".byUTF!dchar().equal([0x00010437]);
+
+    try
+    {
+        "hello0xFFbetty".byUTF!(dchar, UseReplacementDchar.yes).equal("hello");
```
thewilsonator: this should be an `assert[Not]Thrown`. That whole unittest looks like it does nothing, I assume that the above (not including this one) are supposed to no throw? 
This one explicit uses `ReplacementDchar` so it would seem to me that it should not be throwing, why the catch?

WalterBright: > That whole unittest looks like it does nothing

If it doesn't throw, the `equal()` will assert.

> why the catch?

So the unittest run won't exit.

thewilsonator: > the equal() will assert.

Really??? That is `std.algorithm.comparison.equal` we are talking about here, right? Why on earth would that assert? 

For reference, every other unittest in this file e.g. https://github.com/dlang/phobos/blob/61200ae059ea3d5a7fd068aef9d1a425e0f7e7f9/std/utf.d#L2934 is of the form `rangeOrString.toUTFBlah.equal(["expected output"])` except for this one. Why this is different I have no idea.

Regardless, If this is expected to throw then it should use, `assertThrown`, if not it should use `assertNotThrown`.

wilzbach: > equal() will assert.

Not without an `assert`.

> the unittest run won't exit.

Isn't this the point of a unittest?

WalterBright: Ah, you're right.

std/utf.d
```diff
@@ -4386,3 +4387,17 @@ if (isSomeChar!C)
     "𐐷".byUTF!wchar().equal([0xD801, 0xDC37]);
     "𐐷".byUTF!dchar().equal([0x00010437]);
```
wilzbach: `assert` is missing here in the existing code too :/

std/utf.d
wilzbach: > rangeOrString.toUTFBlah.equal(["expected output"]) except for this one. Why this is different I have no idea.

They need to use `assert` too.

thewilsonator: No, they already do that, I forgot to include that, whoops! Its only the one just above this that doesn't.

std/utf.d
```diff
@@ -4386,3 +4387,12 @@ if (isSomeChar!C)
...
 }
+
+///
+unittest
+{
+    import std.algorithm.comparison : equal;
+    import std.exception : assertThrown;
+
+    assertThrown!UTFException("hello0xF0betty".byUTF!(dchar, UseReplacementDchar.no).equal("hellobetty"));
```
thewilsonator: don't you need to do `"hello\xF0betty"`?

WalterBright: No, because it won't complete.

What I'd like to know is why the exception isn't being thrown.

thewilsonator: ... because `0`, `x` and `f` are all valid unicode characters and not e.g. random garbage that can't be converted from utf8 to 32?

WalterBright: Dang it, you're right again. I should just give up and go to bed.

std/utf.d
```diff
@@ -4186,6 +4186,7 @@ private int impureVariable;
  *
  * Params:
  *      C = `char`, `wchar`, or `dchar`
+ *      useReplacementDchar = if invalid UTF, return replacementDchar rather than throwing
```
CyberShadow: Invert this (`If No.useReplacementDchar gives, throw rather than return replacementDchar on invalid UTF`), or update the documentation block above. Currently it's worded as an additional behavior to the behavior described in the second paragraph.

std/utf.d
```diff
@@ -4386,3 +4387,13 @@ if (isSomeChar!C)
...
 }
+
+///
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+    import std.exception : assertThrown;
+
+    assert("hello\xF0betty".byChar.byUTF!(dchar, UseReplacementDchar.yes).equal("hello\uFFFDetty"));
+    assertThrown!UTFException("hello\xF0betty".byChar.byUTF!(dchar, UseReplacementDchar.no).equal("hello\xF0betty"));
```
CyberShadow: This test is flawed, the `assectThrown` might be catching an exception thrown from `equal` (as it will be autodecoding its second parameter too).

CyberShadow: Perhaps use `each` or `walkLength` to expend the range instead.

PR #7147, opened 2019-08-18
prepare for fix of issue 20138 ("is expression not evaluating correct…

aG0aep6G:
…ly?")

With issue 20138 fixed, `const U` and `shared U` do match `const shared T`,
and they leave the other qualifier intact. So `const shared U` must be
attempted first if we want to strip both qualifiers.

This is blocking <https://github.com/dlang/dmd/pull/10317>.


TurkeyMan:
This is very strange... I don't understand, but I take your word for it :)

thewilsonator:
Please make sure this effect is highlighted in a changelog for dlang/dmd#10317

PR #7149, opened 2019-08-20
std.net.curl: Fix example in HTTP documentation

BoQsc:
https://dlang.org/phobos/std_net_curl.html#.HTTP
The Example seems to not work without these fixes.

* Fixed missing http variable
* Fixed deprecated concatenation 
```
Error: Implicit string concatenation is deprecated, use "http://upload.wikimedia.org/wikipedia/commons/" ~ "5/53/Wikipedia-logo-en-big.png" instead
```

### DIFF
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -2332,7 +2332,8 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   * // Track progress
   * auto http = HTTP();
   * http.method = HTTP.Method.get;
-  * http.url = "http://upload.wikimedia.org/wikipedia/commons/5/53/Wikipedia-logo-en-big.png";
+  * http.url = "http://upload.wikimedia.org/wikipedia/commons/" ~
+               "5/53/Wikipedia-logo-en-big.png";
   * http.onReceive = (ubyte[] data) { return data.length; };
   * http.onProgress = (size_t dltotal, size_t dlnow,
   *                    size_t ultotal, size_t ulnow)
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -2325,6 +2325,7 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   * ---
   * import std.net.curl, std.stdio;
   *
+  * auto http = HTTP();
   * auto msg = "Hello world";
   * http.contentLength = msg.length;
   * http.onSend = (void[] data)
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -2330,8 +2330,9 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   * http.perform();
   *
   * // Track progress
+  * auto http = HTTP();
   * http.method = HTTP.Method.get;
-  * http.url = "http://upload.wikimedia.org/wikipedia/commons/"
+  * http.url = "http://upload.wikimedia.org/wikipedia/commons/~" ~
   *            "5/53/Wikipedia-logo-en-big.png";
   * http.onReceive = (ubyte[] data) { return data.length; };
   * http.onProgress = (size_t dltotal, size_t dlnow,
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -2328,7 +2328,7 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   *     return len;
   * };
   * http.perform();
-  *
+  * ---
   * // Track progress
   * auto http = HTTP();
   * http.method = HTTP.Method.get;
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -2349,7 +2349,7 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   * auto http = HTTP();
   * http.method = HTTP.Method.get;
   * http.url = "http://upload.wikimedia.org/wikipedia/commons/" ~
-               "5/53/Wikipedia-logo-en-big.png";
+  *            "5/53/Wikipedia-logo-en-big.png";
   * http.onReceive = (ubyte[] data) { return data.length; };
   * http.onProgress = (size_t dltotal, size_t dlnow,
   *                    size_t ultotal, size_t ulnow)
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -2314,8 +2314,17 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   *     (in char[] key, in char[] value) { writeln(key ~ ": " ~ value); };
   * http.onReceive = (ubyte[] data) { /+ drop +/ return data.length; };
   * http.perform();
+  * ---
   *
-  * // Put with data senders
+  */
+
+/**
+  * HTTP Put with data senders
+  *
+  * Example:
+  * ---
+  * import std.net.curl, std.stdio;
+  *
   * auto msg = "Hello world";
   * http.contentLength = msg.length;
   * http.onSend = (void[] data)
@@ -2328,8 +2337,17 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   *     return len;
   * };
   * http.perform();
+  *
   * ---
-  * // Track progress
+  */
+
+/**
+  * HTTP Track progress
+  *
+  * Example:
+  * ---
+  * import std.net.curl, std.stdio;
+  *
   * auto http = HTTP();
   * http.method = HTTP.Method.get;
   * http.url = "http://upload.wikimedia.org/wikipedia/commons/" ~
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -2323,7 +2323,7 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   * ---
   * import std.net.curl, std.stdio;
   *
-  * auto http = HTTP();
+  * auto http = HTTP("dlang.org");
   * auto msg = "Hello world";
   * http.contentLength = msg.length;
   * http.onSend = (void[] data)
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -2308,7 +2308,7 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   * import std.net.curl, std.stdio;
   *
   * // Get with custom data receivers
-  * auto http = HTTP("dlang.org");
+  * auto http = HTTP("https://dlang.org");
   * http.onReceiveHeader =
   *     (in char[] key, in char[] value) { writeln(key ~ ": " ~ value); };
   * http.onReceive = (ubyte[] data) { /+ drop +/ return data.length; };
@@ -2323,7 +2323,7 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   * ---
   * import std.net.curl, std.stdio;
   *
-  * auto http = HTTP("dlang.org");
+  * auto http = HTTP("https://dlang.org");
   * auto msg = "Hello world";
   * http.contentLength = msg.length;
   * http.onSend = (void[] data)
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -2303,11 +2303,11 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
 
 /**
   * HTTP client functionality.
+  * Get with custom data receivers
   *
   * ---
   * import std.net.curl, std.stdio;
   *
-  * // Get with custom data receivers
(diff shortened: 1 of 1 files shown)


std/net/curl.d
```diff
@@ -2330,9 +2330,9 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   * http.perform();
   *
   * // Track progress
+  * auto http = HTTP();
   * http.method = HTTP.Method.get;
-  * http.url = "http://upload.wikimedia.org/wikipedia/commons/"
-  *            "5/53/Wikipedia-logo-en-big.png";
+  * http.url = "http://upload.wikimedia.org/wikipedia/commons/5/53/Wikipedia-logo-en-big.png";
```
BoQsc: With concatination:
newfigy

Without
figy

std/net/curl.d
```diff
@@ -2330,9 +2330,9 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   * http.perform();
   *
   * // Track progress
+  * auto http = HTTP();
```
CyberShadow: The parts of this example above do not redeclare `http`.

std/net/curl.d
CyberShadow: Don't do that, just concatenate the strings using `~`.

CyberShadow: I'm not sure what you're trying to show here. The generated code should be identical, and even if it weren't, the concatenation happens only once and will take an unmeasurable amount of time. Any perceived difference in speed will be due to fluctuations in your network connection.

BoQsc: Well, I'm trying to say that new concatination (`~`) completely breaks the functionality, so I removed concatination completely and made the statement in a single line. 


### Original version:
```
http.url = "http://upload.wikimedia.org/wikipedia/commons/"
                 "5/53/Wikipedia-logo-en-big.png";
```
**Shows Error:** 
`Error: Implicit string concatenation is deprecated, use "http://upload.wikimedia.org/wikipedia/commons/" ~ "5/53/Wikipedia-logo-en-big.png" instead`

---


**Rewritten Original version:** uses new concatination syntax 
```
http.url = "http://upload.wikimedia.org/wikipedia/commons/~" ~
                 "5/53/Wikipedia-logo-en-big.png";
```
**Does not work:** the output file is corrupt, download progress does not match what is downloaded.

---

**Rewritten Original version:** removed/do not use concatination 
```
http.url = "http://upload.wikimedia.org/wikipedia/commons/5/53/Wikipedia-logo-en-big.png";
```
**Works Perfectly,** downloads the file, shows correct information about file progress.

CyberShadow: Oh, I see.

You accidentally typed an extra `~` in your "Rewritten Original Version". Remove the `~` from before the `"` and it should work.

BoQsc: Trust me, I tried that. When you concatinate - nothing is being downloaded.

CyberShadow: Sorry, but what you're saying is not possible. Please check for errors on your side.

If you are confident in that statement, then you need to find the core reason for this. Do not submit a change if you don't fully understand it.

std/net/curl.d
BoQsc: I though it did redeclared. 
I'm sorry then.

std/net/curl.d
BoQsc: Yep, I completely messed up this one.

BoQsc: Shall I Close this Pull Request, or try to fix it?

std/net/curl.d
CyberShadow: OK, then either this should be removed again, or this example split up into several examples using `---` delimiters for each block.

std/net/curl.d
CyberShadow: Looks OK now. Just the `http` redeclaration issue remains.

std/net/curl.d
BoQsc: I think it would be better to spilt up into two examples. We shall do that.

CyberShadow: - You need to add a terminating `---` in addition to the starting `---`.
- You should move the comments from the start of each example to a DDoc paragraph.
- The import statements at the top of the first example should be duplicated in every example.
- There should probably be two examples, not three, one per each request.

BoQsc: Okay, I spilt it into three examples. I'm keep on hoping I didn't make more mistakes/errors.

BoQsc: Whoops, still failing.

CyberShadow: Please test that the separate examples work. For example I see the second one is missing a `http` declaration.

std/net/curl.d
```diff
@@ -2328,11 +2338,21 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
...
   * };
   * http.perform();
+  * ---
+  *
+  */
+
+/**
+  * HTTP Track progress
+  *
+  * Example:
```
CyberShadow: Probably remove the `Example:` lines as they don't look great in DDox.

std/net/curl.d
```diff
@@ -2328,11 +2338,21 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   *     return len;
   * };
   * http.perform();
+  * ---
+  *
+  */
+
+/**
+  * HTTP Track progress
```
CyberShadow: Punctuation, capitalization. Perhaps: `Tracking progress:`

std/net/curl.d
```diff
@@ -2314,8 +2314,18 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   *     (in char[] key, in char[] value) { writeln(key ~ ": " ~ value); };
   * http.onReceive = (ubyte[] data) { /+ drop +/ return data.length; };
   * http.perform();
+  * ---
+  *
+  */
+
+/**
+  * HTTP Put with data senders
```
CyberShadow: Punctuation, capitalization. `HTTP` is redundant. Perhaps: `Put with data senders:`

std/net/curl.d
BoQsc: All examples work except this one:

I tested this one Manually, and it seems that there might be problems with **curl**, it is throwing exceptions.

https://github.com/dlang/phobos/blob/e775bd9973c8135c0e2881ec914d0d98fc9a4838/std/net/curl.d#L2320-L2341

```
std.net.curl.CurlException@std\net\curl.d(4364): URL using bad/illegal format or missing URL on handle 842560
----------------
0x004159E1
0x00402362
0x00403E27
0x00403DA1
0x00403C3C
0x004024DF
0x770B6359 in BaseThreadInitThunk
0x77977A94 in RtlGetAppContainerNamedObjectPath
0x77977A64 in RtlGetAppContainerNamedObjectPath
```

CyberShadow: > missing URL on handle

You did not specify an URL for the request.

Note how the first example specifies an URL as the parameter to the `HTTP` constructor.

std/net/curl.d
```diff
@@ -2314,8 +2313,17 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
...
+  */
+
+/**
+  * Put with data senders:
+  *
+  * ---
+  * import std.net.curl, std.stdio;
   *
-  * // Put with data senders
+  * auto http = HTTP();
```
BoQsc: **Probably requires a working host.**
```
auto http = HTTP("dlang.org");
```
This examples probably never worked, since it requires a working host that provides response. But I'm not sure.

std/net/curl.d
BoQsc: Yep, I noticed that. But it still throws error. The host is probably responding, otherwise it would throw exception "`Couldn't resolve host name on handle`".

```
std.net.curl.CurlException@std\net\curl.d(4364): Send failed since rewinding of the data stream failed on handle 9A2560
----------------
0x00414DD9
0x0040236A
0x00403E2F
0x00403DA9
0x00403C44
0x004024E7
0x770B6359 in BaseThreadInitThunk
0x77977A94 in RtlGetAppContainerNamedObjectPath
0x77977A64 in RtlGetAppContainerNamedObjectPath
```

CyberShadow:
On 20/08/2019 16.10, Feldwor wrote:
> This examples probably never worked, since it requires a working host that provides response. But I'm not sure.

The only thing that doesn't work for me is that a request to "dlang.org" 
now redirects to https (because the dlang.org server configuration 
changed to redirect to HTTPS). Fixing that, the example as it is now in 
std.net.curl works for me.

-- 
Best regards,
  Vladimir

std/net/curl.d
CyberShadow: I don't know why you get that error. Perhaps you made a typo? Can you post the entire program you're trying to run?

std/net/curl.d
BoQsc: I removed all three, I'm not sure if that's what you meant.

CyberShadow: I think we should keep the first one.

Have a look at the generated docs in the DAutoTest CI status when it finishes, you can see the rendered old version and the new version with your changes there.

std/net/curl.d
BoQsc:
```
import std.net.curl, std.stdio;

void main(){
 auto http = HTTP("dlang.org");
 auto msg = "Hello world";
 http.contentLength = msg.length;
 http.onSend = (void[] data)
 {
     auto m = cast(void[]) msg;
     size_t len = m.length > data.length ? data.length : m.length;
     if (len == 0) return len;
     data[0 .. len] = m[0 .. len];
     msg = msg[len..$];
     return len;
 };
 http.perform();
}
```

std/net/curl.d
```diff
@@ -2323,7 +2323,7 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   * ---
   * import std.net.curl, std.stdio;
   *
-  * auto http = HTTP();
+  * auto http = HTTP("dlang.org");
```
CyberShadow: Might as well fix it to `"https://dlang.org"` so the example actually works.

std/net/curl.d
CyberShadow: Strange that you get "couldn't resolve host name". Perhaps it is a peculiarity of curl and SSL requests on WIndows. I get a different error message (`Send failed since rewinding of the data stream failed on handle ...`), as mentioned above fixed by changing the URL to use HTTPS.

BoQsc: That was my mistake, I should get proper text editor. 
```
std.net.curl.CurlException@std\net\curl.d(4364): Send failed since rewinding of the data stream failed on handle C32560
```

BoQsc: OK. Works perfectly.

std/net/curl.d
```diff
@@ -2328,11 +2336,20 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
...
+  * ---
+  * import std.net.curl, std.stdio;
   *
-  * // Track progress
+  * auto http = HTTP();
   * http.method = HTTP.Method.get;
-  * http.url = "http://upload.wikimedia.org/wikipedia/commons/"
-  *            "5/53/Wikipedia-logo-en-big.png";
+  * http.url = "http://upload.wikimedia.org/wikipedia/commons/" ~
+               "5/53/Wikipedia-logo-en-big.png";
```
CyberShadow: All lines begin with an asterisk except this one. For consistency it would be good to preserve the comment style.
```suggestion
  *            "5/53/Wikipedia-logo-en-big.png";
```

CyberShadow:
Current rendered version:

DDoc: http://dtest.dlang.io/artifact/website-72f8ad8cb8eb0fccaf4f5bbb5abff6a77efead2c-584936282081e09597e23ea51763f7d6/web/phobos-prerelease/std_net_curl.html#.HTTP

DDox: http://dtest.dlang.io/artifact/website-72f8ad8cb8eb0fccaf4f5bbb5abff6a77efead2c-584936282081e09597e23ea51763f7d6/web/library-prerelease/std/net/curl/http.html

Looks pretty good.

std/net/curl.d
```diff
@@ -2304,18 +2304,26 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
 /**
   * HTTP client functionality.
   *
-  * Example:
   * ---
   * import std.net.curl, std.stdio;
   *
   * // Get with custom data receivers
```
CyberShadow: For consistency this comment should be moved out of the code block too. Then the first example will look like the two examples below.

std/net/curl.d
```diff
@@ -2305,17 +2305,28 @@ private bool decodeLineInto(Terminator, Char = char)(ref const(ubyte)[] basesrc,
   * HTTP client functionality.
   *
   * Example:
+  *
+  * Get with custom data receivers
```
CyberShadow:
```suggestion
  * Get with custom data receivers:
```

CyberShadow:
For some reason GitHub did not allow me to commit my own suggestion, even though it allowed me to edit the file manually.

PR #7150, opened 2019-08-20
std.process: use local imports

kubo39:
In current phobos policy, local imports should be preferred over global
imports.


thewilsonator:
@CyberShadow DAutoTest timed out.

CyberShadow:
Not a timeout, code.dlang.org was flaky once again. @kubo39 Kindly rebase and force-push.

PR #7151, opened 2019-08-20
std.bigint.BigInt: allow casting to floating point numbers

n8sh:
Limitation: ~~casting to & comparison with quadruple precision floating point numbers are not supported.~~ The path for 80 bit (EDIT: and greater) extended floats use std.math.scalbn which currently is impure ~~and non-CTFE-able~~.

### DIFF
--- a/std/bigint.d
+++ b/std/bigint.d
@@ -826,13 +826,82 @@ public:
         assertThrown!ConvOverflowException(BigInt("9223372036854775808").to!long);
     }
 
-    // Cast to float, discarding any portion of the value
-    // beyond the precision of the floating point type.
-    private T toFloatTruncating(T)() @safe nothrow @nogc const
-    if (__traits(isFloating, T))
+    /**
+        Implements casting to floating point types.
+     */
+    T opCast(T)() @safe nothrow @nogc const if (isFloatingPoint!T)
+    {
+        return toFloat!(T, "nearest");
+    }
+
+    ///
+    @system unittest
+    {
+        assert(0x1.abcd_e8p+124f        == cast(float)  BigInt("0x1abc_de80_0000_0000_0000_0000_0000_0000"));
+        assert(0x1.abcd_ef12_3456p+124  == cast(double) BigInt("0x1abc_def1_2345_6000_0000_0000_0000_0000"));
+        assert(0x1.abcd_ef12_3456p+124L == cast(real)   BigInt("0x1abc_def1_2345_6000_0000_0000_0000_0000"));
+
+        assert(-0x1.3456_78p+108f        == cast(float)  BigInt("-0x1345_6780_0000_0000_0000_0000_0000"));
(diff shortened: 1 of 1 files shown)


Biotronic:
Is `scalbn` a better solution than `r * 2.0L^^N`? At the very least, that solution should be a good enough stopgap for CTFE.

n8sh:
@Biotronic: let's use that to make `scalbn` CTFE-able in general (https://github.com/dlang/phobos/pull/7153).

std/bigint.d
```diff
@@ -776,6 +809,216 @@ public:
...
+            const ulong w2 = ulongLength < 2 ? 0 : data.peekUlong(ulongLength - 2);
+            const uint w1BitCount = bsr(w1) + 1;
+            ulong sansExponent = (w1 << (64 - w1BitCount)) | (w2 >>> (w1BitCount));
+            size_t exponent = (ulongLength - 1) * 64 + w1BitCount + 1;
+            static if (performRounding)
+            {
+                // If we wanted to round ties away from 0 we could just add
+                // roundUpInc to sansExponent. The more complicated logic is
+                // so ties are rounded to even, which matches the default
+                // rounding behavior when casting int/long to float/double.
```
n8sh: Since this seemingly isn't the default rounding mode on the Win32 & FreeBSD_32 test machines, my inclination is to abandon this rounding mode and round ties away from zero for float/double.

berni44:
Is this still WIP?

n8sh:
@berni44 the WIP tag is because the rounding behavior [might astonish](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) someone who used it without reading the documentation.
```
// BigInts that fall exactly between two non-infinite floats/doubles
// are rounded away from zero when cast to float/double. (Note that
// in most environments this is NOT the same rounding rule rule used
// when casting int/long to float/double.)
```
Any way of handling rounding had disadvantages. I'm not convinced that this choice is the best.

n8sh:
Also, this isn't great:
> When cast to float or double or 64-bit real the rounding is strictly defined. When cast to extended-precision real the rounding rules vary by environment.

berni44:
> @berni44 the WIP tag is because the rounding behavior [might astonish](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) someone who used it without reading the documentation.

I had exactly the same problems with the functions that print floating point numbers in std.format. But there is a slight difference: Printing is IMHO mainly done for humans and there is IMHO "ties away from zero" the expected choice. Here, the numbers might be used to do some more calculating and as far as I know "ties to even" works better for that, because it can straighten out the accumulating errors.

What I currently do in std.format is to read out the rounding mode of the processor and use this mode (round down, round up, round to zero, round to nearest with ties to even). That's what printf does and the reason why printing floats isn't pure.

PR #7152, opened 2019-08-22
Update ThreadInfo.cleanup() to unregister itself instead of the local thread's

look-at-me:


thewilsonator:
Does this have a bugzilla number?

Geod24:
@look-at-me : Now that you found the case, could you craft a test case for it, and create a bugzilla issue so that it shows up in the changelog ?
@thewilsonator : Not yet. It is one of the bounties that we are sponsoring. I didn't have a small test case to reproduce it, but it looks like  @look-at-me found the root cause (see https://github.com/Geod24/localrest/pull/5 )

look-at-me:
Not sure if it's worth adding a unittest over. Not exactly an issue that will likely be repeated.

AndrejMitrovic:
> Not sure if it's worth adding a unittest over. Not exactly an issue that will likely be repeated.

Someone in the future might tag along and think "huh, why is this code here, let's see what happens if I delete it? Hey look, the tests still pass. Great."

look-at-me:
There's tests that check the functionality. Just not the edge case that caused this error. This edge case isn't worth testing as it is unlikely to happen again in the future.

PR #7153, opened 2019-08-23
Fix Issue 20159 - Make std.math.scalbn work in CTFE

n8sh:


WalterBright:
Getting:
```
std\math.d(4566): Error: static assert:  `nanL == infL` is false
```
now in Azure pipelines (Windows_VisualD win32-ldc)

WalterBright:
@n8sh please fix.

n8sh:
That's the same problem I was seeing with FreeBSD. It seemed like CTFE didn't have 80 bit reals even though they existed at runtime. I'll try enabling the slow path for all platforms.

n8sh:
@WalterBright PR #7155

PR #7155, opened 2019-08-24
Issue 20159 Followup - fix ldc-win32 test failure

n8sh:
https://github.com/dlang/phobos/pull/7153#issuecomment-524536212


WalterBright:
@n8sh thanks for the quick action!

PR #7157, opened 2019-08-27
llroundl is available on windows, so why not use it?

John-Colvin:
I suspect this PR will fall foul of some platforms/runtimes... let's see :)

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -5414,26 +5414,18 @@ version (Posix)
  *
  * If the fractional part of x is exactly 0.5, the return value is rounded
  * away from zero.
- *
- * $(BLUE This function is Posix-Only.)
  */
 long lround(real x) @trusted nothrow @nogc
 {
-    version (Posix)
-        return core.stdc.math.llroundl(x);
-    else
-        assert(0, "lround not implemented");
+    return core.stdc.math.llroundl(x);
 }
 
 ///
 @safe nothrow @nogc unittest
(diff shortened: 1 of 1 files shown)


John-Colvin:
So... Windows 32 has llroundl declared but not actually present in the C runtime by the looks of things.

Is the autotester Win_32 working with DigitalMars or Microsoft C runtime?

thewilsonator:
Digital Mars.

John-Colvin:
What system tests with the 32 bit Microsoft runtime?

ZombineDev:
> What system tests with the 32 bit Microsoft runtime?

Look for the the `Win_32` rows on the `auto-tester`.

ZombineDev:
If you want to test with `-m32mscoff`, the best would be to setup Azure Pipelines, similarly to how it's done for dmd:

* https://github.com/dlang/dmd/blob/master/azure-pipelines.yml
* https://github.com/dlang/dmd/blob/f5069353042e50ea566f8ed28f6eb5d19a78cc9b/.azure-pipelines/windows.sh#L67

RazvanN7:
`std/math.d` no longer exists; it has been refactored into multiple files. Right now the functions touched by this PR no longer call `core.stdc.math.llroundl` only for Posix. It has been refactored to assert if `version=CRuntime_DigitalMars`, which is correct.

Closing this.

PR #7160, opened 2019-08-29
Fix CircleCI build

ZombineDev:
*Edit (see GH message history if needed):*

* Upgrade `HOST_DMD_VER` from `2.078.1` to `2.079.1`, to match the dmd CircleCI build
* Use `../dmd/scr/build.d`, instead of `../dmd/src/posix.mak`, because `BUILD=debug` is broken since dlang/dmd#10212 (comment)
* Merge `gdb` download step into `install_deps`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,9 +14,6 @@ jobs:
           command: ./.circleci/run.sh setup-repos
           name: Clone DMD & DRuntime
       - run:
-          command: sudo apt-get update && sudo apt-get install -y gdb
-          name: Install gdb
-      - run:
           command: ./.circleci/run.sh publictests
           name: Run all public unittests
       - run:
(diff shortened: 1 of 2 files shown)


ZombineDev:
Nope, it looks like there's more changes. Since the latest changes around `build.d`, we're building dmd on CircleCI like this:
* `/home/circleci/dlang/dmd-2.079.1/linux/bin64/rdmd ./src/build.d MODEL=64 HOST_DMD=dmd BUILD=debug ENABLE_WARNINGS=1 PIC=1 all`
While for phobos we build it like this:
* `make -j2 -C ../dmd/src -f posix.mak MODEL=64 HOST_DMD=dmd BUILD=debug all`

ZombineDev:
Found the root cause: https://github.com/dlang/dmd/pull/10212#issuecomment-526181232

ZombineDev:
CircleCI is now green -> `auto-merge`-ing.

PR #7164, opened 2019-09-01
Fix Issue 20186 - File size of "Hello, world" executable increased by 185KB

CyberShadow:
We can delay the std.format import until the relevant templates are instantiated.

This brings the filesize from 1114048 to 908544 bytes.

CC @JackStouffer

### DIFF
--- a/.dscanner.ini
+++ b/.dscanner.ini
@@ -290,6 +290,7 @@ object_const_check="-std.algorithm.searching,-std.array,-std.bitmanip,-std.concu
 ; Checks that opEquals and toHash are both defined or neither are defined
 opequals_tohash_check="-std.complex,-std.container.array,-std.container.dlist,-std.container.rbtree,-std.container.slist,-std.datetime,-std.datetime.date,-std.experimental.checkedint,-std.internal.test.dummyrange,-std.json,-std.numeric,-std.random,-std.socket,-std.typecons,-std.uni"
 ; Check for properly documented public functions (Returns, Params)
+; Note: DScanner doesn't understand documenting parameters of IFTI/eponymous templates.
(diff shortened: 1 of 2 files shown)


CyberShadow:
```
std/array.d(3561:12)[warn]: Documented parameter w isn't a function parameter.
std/array.d(3561:12)[warn]: Documented parameter fmt isn't a function parameter.
std/array.d(3584:14)[warn]: Template parameters Writer isn't documented in the `Params` section.
```

How to make Dscanner happy with eponymous templates? Any choice but to document the outer template and inner function separately?

wilzbach:
Yeah, the alternative would be fixing that dscanner check :/

CyberShadow:
The generated docs look OK to me:
http://dtest.dlang.io/artifact/website-8f8f5bb3ff6c01cf962adcd8e6c4cc0cfe77994a-755d84188d74ab3c2a342152bd4d0b7f/web/phobos-prerelease/std_array.html#.Appender.toString

Making DScanner happy would mean making the HTML docs worse.

Is it possible to silence or ignore the DScanner warning?

wilzbach:
You could add an ignore rule to .dscanner.ini (bottom part).

CyberShadow:
OK, done. Surprising that nothing else in Phobos was tripped by that buggy check. Or was everything changed to not trigger it when Dscanner was added to Phobos? Hopefully it didn't affect readability of the documentation too much in that case.

CyberShadow:
It worked! https://perf.dlang.io/#program-hello-binarysize;;1566904042.0561454;1567421184

John-Colvin:
It also shaved about 30% off the diet template compile-time!

PR #7165, opened 2019-09-01
Add Unconst to std.traits

thaven:
Existing ˋstd.traits.Unqualˋ removes all qualifiers from a type. Quite often that is a bit too aggressive, because it would also strip off 'shared'. ˋUnconstˋ (which already exists in ˋcore.internal.traitsˋ) is helpful in such cases. This PR adds an alias for ˋUnconstˋ to ˋstd.traitsˋ, very similar to what is already there for ˋUnqualˋ.

### DIFF
--- a/changelog/unconst.dd
+++ b/changelog/unconst.dd
@@ -1,0 +1,5 @@
+Added std.traits.Unconst
+
+$(REF Unconst, std, traits) works like $(REF Unqual, std, traits), but removes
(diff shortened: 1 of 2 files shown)


n8sh:
Forgot about the policy for new public symbols. Applying `72h no objection -> merge` was inappropriate.

thewilsonator:
cc @atilaneves

PR #7167, opened 2019-09-01
Added spawnRef and spawnLinkedRef for passing by reference through spawn

ErnyTech:
It is often useful to pass a mutable by reference type to the spawned
thread, the need to have to use a raw pointer before this patch made
the syntax boring and not in line with the rest of the DLang style
(use pointer as little as possible).

With this patch it will be possible to write code like this:
```
import std.concurrency : spawnRef;
import core.thread : thread_joinAll;

static void f1(ref shared(int) number)
{
  atomicOp!"+="(number, 1);
}

shared(int) number = 10;
spawnRef(&f1, number);
thread_joinAll();
assert(number == 11);
```

Instead of the boring:
```
import core.thread : thread_joinAll;

static void f1(ref shared(int)* number)
{
  atomicOp!"+="(*number, 1);
}

shared(int) number = 10;
spawn(&f1, &number);
thread_joinAll();
assert(number == 11);
```

This PR is an alternative of https://github.com/dlang/phobos/pull/7124

### DIFF
--- a/changelog/spawn-by-ref.dd
+++ b/changelog/spawn-by-ref.dd
@@ -1,0 +1,20 @@
+Added spawnRef and spawnLinkedRef for passing by reference through spawn
+
+Now it will be possible to pass variables to the thread created with spawnRef/spawnLinkedRef by reference,
+to do this you need the shared type and add ref to the attributes of the arguments of the function to be executed.
+
+---
+import std.concurrency : spawnRef;
+import core.atomic : atomicOp;
+import core.thread : thread_joinAll;
+
+static void f1(ref shared(int) number)
+{
+  atomicOp!"+="(number, 1);
+}
+
+shared(int) number = 10;
+spawnRef(&f1, number);
+thread_joinAll();
+assert(number == 11);
+---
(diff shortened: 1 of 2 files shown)


atilaneves:
As mentioned in the other PR, I'm not sure of this at all.

ErnyTech:
I don't know, everyone does not like the previous PR and currently does not pass all the tests yet

atilaneves:
> I don't know, everyone does not like the previous PR and currently does not pass all the tests yet

I'd been previously miffed by not being able to use spawn with ref functions, so I like the premise of it. A separate function that uses `ref` is not something I think is useful and, at this point, unlikely to convince me to add to Phobos.

RazvanN7:
@ErnyTech @atilaneves How should we move forward here? Should we close this?

atilaneves:
Is there a reason why it has to have a different name?

ErnyTech:
Sorry but currently I can no longer take care of this, I think it is better to close this.

If someone else intends to continue they can open a new PR

RazvanN7:
Filed: https://issues.dlang.org/show_bug.cgi?id=21848

Closing.

PR #7169, opened 2019-09-04
Fix Issue 17705 - std.math.isFinite cannot run at compile-time

n8sh:


thewilsonator:
Hmm, 
```
Failed to download package openssl from https://code.dlang.org/packages/openssl/1.1.6+1.0.1g.zip
posix.mak:774: recipe for target 'dpl-docs' failed
```
give it another kick?

PR #7172, opened 2019-09-06
Fix Issue 20147 - Enable comparison (==, >, >=, <=, <) between std.bigint.BigInt and floating point numbers

n8sh:
Split off as its own PR separate from #7151 because there is a single mathematically correct way to compare numbers without needing to consider rounding mode.


thewilsonator:
Good to go?

n8sh:
Yep

PR #7173, opened 2019-09-07
Fix Issue 15881 - approxEqual Ignores maxAbsDiff

crocopaw:
I implemented the algorithm given in [1] and added and sorted the unittest for approxEqual. As the new algorithm does not always give the same result as the old one did, lots of unittests in phobos broke. I repaired them all, mostly by just giving a few more digits (making the tests stronger) or by adding new difference values (not changing much).

There was one place, where approxEqual was used inside of a function, namely in secantMethod in std/numerical.d. As my mathematical background is not good enough, I don't exactly understand, what that function is doing. Probably the new version will lead to more correct digits in the result, but will also be a little bit slower, because the loop is executed more often.

One more problem are floats of type float. The default value of maxRelDiff is optimized for doubles. For floats it should better be 1e-5. I don't know, if it's possible to give a different default value to the template for floats. For tests, where floats are involved, I added the constant manually.

The change will also break 4 tests in dmd/test/compilable/ctfe_math.d. They should now be as follows:

    static assert(approxEqual(sin(2.0L), 0.9092974268L));
    static assert(approxEqual(cos(2.0), -0.4161468365));
    static assert(approxEqual(tan(2.0f), -2.185040f, 1e-5));
    static assert(approxEqual(sqrt(2.0L), 1.414213562L));

Although I don't think so, maybe more tests in dmd will break. 

And last but not least, I added (and commented out) one unittest, that still breaks: approxEqual([],[]) will not work.

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -970,7 +970,7 @@ template equal(alias pred = "a == b")
 @safe unittest
 {
     import std.algorithm.comparison : equal;
-    import std.math : approxEqual2;
+    import std.math : approxEqual;
 
     int[] a = [ 1, 2, 4, 3 ];
     assert(!equal(a, a[1..$]));
@@ -984,7 +984,7 @@ template equal(alias pred = "a == b")
 
     // predicated: ensure that two vectors are approximately equal
     double[] c = [ 1.0000000005, 2, 4, 3];
-    assert(equal!approxEqual2(b, c));
+    assert(equal!approxEqual(b, c));
 }
 
 /++
@@ -1008,7 +1008,7 @@ range of range (of range...) comparisons.
     import std.algorithm.iteration : map;
     import std.internal.test.dummyrange : ReferenceForwardRange,
         ReferenceInputRange, ReferenceInfiniteForwardRange;
-    import std.math : approxEqual2;
+    import std.math : approxEqual;
 
     // various strings
     assert(equal("æøå", "æøå")); //UTF8 vs UTF8
@@ -1043,10 +1043,10 @@ range of range (of range...) comparisons.
     assert(equal([2, 4, 8, 6], map!"a*2"(a)));
     double[] b = [ 1.0, 2, 4, 3];
     double[] c = [ 1.0000000005, 2, 4, 3];
-    assert(equal!approxEqual2(map!"a*2"(b), map!"a*2"(c)));
+    assert(equal!approxEqual(map!"a*2"(b), map!"a*2"(c)));
     assert(!equal([2, 4, 1, 3], map!"a*2"(a)));
     assert(!equal([2, 4, 1], map!"a*2"(a)));
-    assert(!equal!approxEqual2(map!"a*3"(b), map!"a*2"(c)));
+    assert(!equal!approxEqual(map!"a*3"(b), map!"a*2"(c)));
 
     //Tests with some fancy reference ranges.
     ReferenceInputRange!int cir = new ReferenceInputRange!int([1, 2, 4, 3]);
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -970,7 +970,7 @@ template equal(alias pred = "a == b")
 @safe unittest
 {
     import std.algorithm.comparison : equal;
-    import std.math : approxEqual;
+    import std.math : approxEqual2;
 
     int[] a = [ 1, 2, 4, 3 ];
     assert(!equal(a, a[1..$]));
@@ -983,8 +983,8 @@ template equal(alias pred = "a == b")
     assert(equal(a, b));
 
     // predicated: ensure that two vectors are approximately equal
-    double[] c = [ 1.005, 2, 4, 3];
-    assert(equal!approxEqual(b, c));
+    double[] c = [ 1.0000000005, 2, 4, 3];
+    assert(equal!approxEqual2(b, c));
 }
 
 /++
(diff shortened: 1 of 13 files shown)


n8sh:
Changing the behavior could cause problems for anyone who is using the function right now. I think it would be better to document the current behavior of `approxEquals` so it's clear that the difference must be less than EITHER `maxRelDiff` OR `maxAbsDiff` rather than less than BOTH, and if you think there's a better / more generally useful algorithm implement that as a new function.

jondegenhardt:
Regarding the tsv-utils test that is failing (buildkite) - I didn't try to debug, but I suspect that the issue is that the `maxRelDiff` default value was changed. In the failing test, the calculations are done in `float`, then converted to `double`, the `approxEqual` called. In the current function, `maxRelDiff` was 1e-5 for both `float` and `double`. In the PR it is 1e-9 for `double`. The exact same unit test succeeds when the calculation is done as double. (This happens earlier in the unit tests.)

It might be worth changing `maxRelDiff` to the values in the current function for both double and float and see how far that goes towards fixing the failures in the buildkite tests.

crocopaw:
> Changing the behavior could cause problems for anyone who is using the function right now. I think it would be better to document the current behavior of `approxEquals` so it's clear that the difference must be less than EITHER `maxRelDiff` OR `maxAbsDiff` rather than less than BOTH, and if you think there's a better / more generally useful algorithm implement that as a new function.

You're right. That's a better approach. I'll file some more PRs soon.

> Regarding the tsv-utils test that is failing (buildkite) - I didn't try to debug, but I suspect that the issue is related to the changes in the `maxRelDiff` and `maxAbsDiff` default values.

It was to expect, that some tests will fail, as the current version is a much weaker test, than the new one. And that's exactly, why I think, a new one should be used. People rely on tests, that actually might accept wrong results too, without noticing, that the tests don't work properly.

> In the PR both are 1e-5 for `float` and 1e-9 for `double`. 

Unfortunately not. That's what I'd like to have (at least for maxRelDiff), but I don't know how to add different default values to different template instantiations. I'll figure that out, probably with the help of the learn-forum, for a new PR. maxAbsDiff is set to 0.0 in the new version. This is primarily used for comparison to zero, but what's close to zero depends heavily on the usecase.

Thanks for your replies.

jondegenhardt:
@crocopaw - Even switching to a new function, it might be worth identifying why the tsv-utils tests failed. I'm rather confident the calculations are correct and don't fall under the cases listed in the bug reports being addressed.

The only legitimate possibility is switching from a tolerance intended for floats to one intended for doubles. Not clear from the PR comments if this is intended or not. If not intended, it suggests a bug in the PR function not covered by unit tests. If intended, it might be worth some design consideration around the goal.

In the case of the failed tsv-utils tests, the values are initially represented as floats, converted to doubles, all the calculations done as doubles, and approxEqual called on the doubles. The same tests passed when the initial representation is as doubles. (The double versions of the tests precede the float versions, so they did pass.) Note that Python does not have scenarios mixing float and double like this, so the tolerances specified in a Python PEP might not be the best default choices in D. Or, the documentation will want to be clear about how callers should specify the tolerances.

crocopaw:
> @crocopaw - Even switching to a new function, it might be worth identifying why the tsv-utils tests failed. I'm rather confident the calculations are correct and don't fall under the cases listed in the bug reports being addressed.

For the same reason I'll have a look at the complex power function, which passed the tests on 64bit machines and failed on 32bit. When I'll find the time, I'll have a look at the tsv-utils too, but don't want to promise that right now.

I'll keep an eye on mixing float and double, when designing the new function. Might take a few days, as I have to investigate on this.

jondegenhardt:
I wrote a small program showing a simplified version of what I think is happening in the failing tsv-utils test. This just has precision loss due only float to double conversion. You can then decide how a new version of `approxEqual` should behave in cases like this, include it in unit tests, etc.
```
void main()
{
    import std.stdio;
    import std.conv : to;

    {
        writeln("--- Float to double conversion ---");
        float f = 31.79f;
        double d = 31.79;
        double f2d = f.to!double;

        assert(d != f2d);

        writefln("  f: %23f", f);
        writefln("  d: %23f", d);
        writefln("f2d: %23f", f2d);
    }
    {
        writeln("--- Double to float to double round trip ---");
        double d = 31.79;
        float f = d.to!float;
        double f2d = f.to!double;

        assert(d != f2d);

        writefln("  d: %23f", d);
        writefln("  f: %23f", f);
        writefln("f2d: %23f", f2d);
    }
}
```
When run this produces:
```
--- Float to double conversion ---
  f:               31.790001
  d:               31.790000
f2d:               31.790001
--- Double to float to double round trip ---
  d:               31.790000
  f:               31.790001
f2d:               31.790001
```

crocopaw:
Thanks for this example. I think it's clear what happens here. With the float to double conversion the function will be able to detect, that one of the operands is a float and will set the parameters accordingly, while with the double to float to double conversation, the function will not know about this itermediate result and use double for selecting the parameters. In this case it's the responsibility of the programmer (who knows that there has been this conversion) to pass correct tolerance parameters. This is the unittest, the new version should pass:

```
unittest
{
    float f = 31.79f;
    double d = 31.79;
    double f2d = f.to!double;

    assert(approxEqualNew(f,f2d));

    double d = 31.79;
    float f = d.to!float;
    double f2d = f.to!double;

    assert(!approxEqualNew(d,f2d));
    assert(approxEqualNew(d,f2d,1e-5));
}
```

PR #7176, opened 2019-09-08
Fix Issue 20198 - Make std.math.nextUp and nextDown and nextafter work in CTFE for float and double

n8sh:


n8sh:
Diagnosing the FreeBSD problems.

n8sh:
I don't foresee working on this in the immediate future so I'm going to submit this without support for extended precision reals.

>Making nextUp/nextDown work for extended precision reals in CTFE is difficult because some D compilers interally use 64 bit precision for CTFE real calculations even though the target has extended precision reals. This specifically happens in the FreeBSD_32 and FreeBSD_64_64 tests in the CI suite.

PR #7180, opened 2019-09-09
Clearer documentation for approxEqual

crocopaw:
In issue 15881 and 15763 people complain about unintuitiv behaviour of approxEqual. I adapted the documentation of the function to address this.

a) Symmetry: I renamed the first two parameters to make clear, that they have different meaning. Also changed the wording in the documentation and added more unittest to clearify.

b) Purpose of the two checks: I made clearer, why two checks are run (basically, one for numbers close to zero and one for all other) and that passing one check is sufficient. Added unittests to document this.

Additionally I added some more unittests. E.g. for testing 3rd and 4th parameters and for some strange results and cornercases.

I did not change the behaviour of the function, let alone, I removed an (almost) superfluous overload - almost, because an alias in dummyrange.d now needs a third parameter.

I refrained from removing "maxAbsDiff != 0 &&" in the last line of the template, which in my eyes is also unnecessary. (If you think, I should remove this too, please tell me.)

I'm not sure, if this can be called a fix for issue 15881.

As already mentioned in PR #7173, I'll provide an additional function in a separate PR, which has similar functionallity, but is better suited for testing for floatingpoint rounding errors.

### DIFF
--- a/std/internal/test/dummyrange.d
+++ b/std/internal/test/dummyrange.d
@@ -354,7 +354,7 @@ if (is(T == double))
         arr = iota().array;
     }
 
-    alias cmp = approxEqual!(double,double);
+    alias cmp = approxEqual!(double,double,double);
 
     enum dummyValue = 1337.0;
     enum dummyValueRslt = 1337.0 * 2.0;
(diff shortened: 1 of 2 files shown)


crocopaw:
ping

PR #4316, opened 2016-05-12
Fix Issue 10448: min and max are not NaN aware

JackStouffer:
min and max now return NaN when ever one of the arguments is NaN.

PR #7181, opened 2019-09-11
Fix Issue 10448 - min and max are not NaN aware

crocopaw:
The discussion of issue 10448 and of the related pull requests show, that there is no consensus on the best result of min and max, when some parameters are NaNs, even experts, namely people who where involved in writing IEEE754, cannot agree on that. But most people agree, that the functions should be fast and extra treatment for NaNs would slow them down. Therefore, I think, the best solution is to document, that the functions will result in undefined behaviour if NaNs are used and give examples on how to cope with NaNs beyond using this functions.

In the documentation for minElement I added two such examples, namely for ignoring NaNs and for returning NaN in case at least one element is NaN. In the documentation for maxElement, min and max I just added a pointer to minElement, to avoid duplication.

I'm aware, that there are more functions, which have this problem. But before doing all the work to identify them and to add a note to everyones documentation, I'd make like to make sure, that this will be accepted. Therefore I'll wait for this PR to be merged, before continuing.


thewilsonator:
So, to clarify,`min` and `max` are currently NaN propagating? or not?

crocopaw:
Neither nor. That's the problem.

min(0, float.nan); //Yields nanF
max(0, float.nan); //Yields 0
min(float.nan, 0); //Yields 0
max(float.nan, 0); //Yields nanF

thewilsonator:
Ow.

AsumFace:
Is it necessary to go as far as undefined behavior? 'Result taking a unspecified value' is IMO preferable since that does not formally require you to check your entire input for NaN.

wilzbach:
Guys, NaN should actually propagate.

crocopaw:
> Is it necessary to go as far as undefined behavior? 'Result taking a unspecified value' is IMO preferable since that does not formally require you to check your entire input for NaN.

I havn't thought about that specific wording. Changed it to "unspecified value".

crocopaw:
> Guys, NaN should actually propagate.

Well, that's what I think too. But one has to accept, that experts disagree on this. And making min etc. propagate NaNs will (as far as I can see) not be possible without a speedloss.

jondegenhardt:
> > Guys, NaN should actually propagate.
> 
> Well, that's what I think too. But one has to accept, that experts disagree on this. And making min etc. propagate NaNs will (as far as I can see) not be possible without a speedloss.

@crocopaw - Sounds like you've done some research into the IEEE standards aspects of this question. Are there some standards or common practices that come into play? e.g. Does IEEE754 take a position? Do major programming languages or libraries tend to agree on behavior, or is it mixed?

crocopaw:
> @crocopaw - Sounds like you've done some research into the IEEE standards aspects of this question.

I did not manage to get an original document, but I did research in the internet (and in old PRs on this and similar issues) and found a few useful quotes. [1] [2] [3]

> Are there some standards or common practices that come into play? e.g. Does IEEE754 take a position?

In [2] is a quote from section 5.3 of IEEE754, which essentially states, that NaNs should be ignored, if any other number is available. But only few programming languages actually comply with this.

> Do major programming languages or libraries tend to agree on behavior, or is it mixed?

No, there is no agreement. [3] gives an overview over the behavior of several programming languages.

In my oppinion, as long, as we do not have any consensus on this issue (and I think this will never happen), the current behaviour should be clearly documented.

[1] https://stackoverflow.com/questions/1565164/what-is-the-rationale-for-all-comparisons-returning-false-for-ieee754-nan-values
[2] Issue #4316 (long debate for a NaN-propagating solution, without result)
[3] https://stackoverflow.com/questions/1565164/what-is-the-rationale-for-all-comparisons-returning-false-for-ieee754-nan-values

crocopaw:
Just I found another strange result of the current implementation:

`
assert(min(2,1,3,real.nan,18,17,19)==17);
`

jondegenhardt:
> I did not manage to get an original document, but I did research in the internet (and in old PRs on this and similar issues) and found a few useful quotes. [1] [2] [3]

Thanks, this is very helpful info. Minor thing - Your first and third links are identical. Was there a different url you meant to include?

crocopaw:
> > I did not manage to get an original document, but I did research in the internet (and in old PRs on this and similar issues) and found a few useful quotes. [1] [2] [3]
> 
> Thanks, this is very helpful info. Minor thing - Your first and third links are identical. Was there a different url you meant to include?

Sorry, was a copy & paste error. Corrected that link above.

Biotronic:
> In my oppinion it's better to have a fast algorithm, that returns nonsens when confronted with NaNs

I have to agree - it's easier to preprocess data to make the unsafe safe than to make the slow fast.

> Better: assert(isNaN(v1)==isNaN(v2) && (isNaN(v1) || isNaN(v2) || v1==v2)); // ;)

Doi. I'll do you one better, though:

    assert((v1 == v2) != (isNaN(v1) && isNaN(v2)));

crocopaw:
> Doi. I'll do you one better, though:

Indeed better! :-)

crocopaw:
ping

crocopaw:
I'd be happy to have a decission on this, because I'd like to delete the account "crocopaw" soon. In my oppinion, accepting this PR doesn't harm, even when you think, that NaNs should propagate or handled differently. For the time until this is decided we would at least have a proper description and the behavior of the function can still be changed later.

See also PR #7214. This is a similar, but simpler case.

berni44:
I had to remove my old account on GitHub. Therefore I created a new version of this PR with my new account. This one can be closed (unfortunately I can't do that anymore).

PR #7182, opened 2019-09-12
Fix issue 20205 - std.math: Wrong result for abs(int.min)

ghost:
Implemented the idea of Dominikus Dittes Scherkl - see issue 20205.


dukc:
Please revert this before release. Consider:

```
import std;
void main()
{
    writeln(50 - abs(-100));
}
```

Before this change, this will print `-50`, but now it'll print `4294967246`, because `abs(-100)` has changed to `unsigned int`. Far more likely to create confusion than the prior situation.

wilzbach:
https://github.com/dlang/phobos/pull/7212

PR #7185, opened 2019-09-17
Fix issue 20218: format: don't try to print the entire range while merely testing the format string

FeepingCreature:
The error-checking benefit is probably achieved by just printing the first element and then stopping.


std/format.d
```diff
@@ -3390,7 +3393,10 @@ if (isInputRange!T)
                         {
                             printed = 0;
                             for (; !val.empty; val.popFront(), ++printed)
+                            {
```
acehreli: Thanks for catching this 'for' statement that was missing the curly braces. ;)

PR #7187, opened 2019-09-18
Add isClose, a replacement for approxEqual

crocopaw:
As suggested in PR #7173, I wrote a new version of approxEqual (called isClose to avoid naming conflicts), which is based on the thoughts given in [1]. The purpose of both functions is the same, namely providing a utility for comparing floating point numbers. This is necessary, because operations on floating point numbers inevitably produce rounding errors and therefore cannot be compared reasonable using ==.

The main difference is, that this function has much stronger default values for the error allowed, based on the capabilities of the used floating point type. This helps in catching errors, especially as approxEqual is mainly used in unittests. This addresses issue #15881. Together with PR #7180 this can be considered a fix for that issue.

The new function is furthermore symmetric, to avoid users being confused or have to remember the order of the operands. This addresses issue #15763.

After this PR has been approved I think, it would be wise to continue with the following roadmap:

1. replace all occurences of approxEqual in Phobos by isClose. With one exception, approxEqual is only used in unittests. The exception is secantMethod in std/numerical.d.
2. deprecate approxEqual in favour of isClose. (I don't know if I'm in the position to deprecate anything, nor to ask for it, but I think, this would be a good idea in the long-run.)
3. After some time, remove approxEqual alltogether.


crocopaw:
Add more unittests for 100% code coverage.

crocopaw:
resolved merge conflict

berni44:
I had to remove the github account with which I created this PR but forgot to close this before. Can anyone with appropriate write access close this PR?

atilaneves:
> I had to remove the github account with which I created this PR but forgot to close this before. Can anyone with appropriate write access close this PR?

Done.

PR #7188, opened 2019-09-19
Fix Issue 14478 - isInputRange should allow ranges of non-copyable elements

RazvanN7:
The condition

```d
is(typeof((return ref R r) => r.front))
```

Creates a lambda that returns a by value element of type R. Because of that, ranges with uncopyable elements will not pass the `isInputRange` test even though `front` might return by ref. To fix this, I simply changed the return type of the lambda by adding `.sizeof`; if the type R has `.front` then the test will pass (it doesn't matter the return type of `front`), otherwise you would have `__error__.front` which obviously is not going to produce a valid type.

The problem is that the original bug report is still not going to compile, because `chain` has code that does copies even though it does not explicitly check for this ability (because `isInputRange` assumed it). I suspect this pattern is prevalent throughout phobos and the fix would be to use `isCopyable!(ElementType!Range)` wherever isInputRange is used.

### DIFF
--- a/std/range/primitives.d
+++ b/std/range/primitives.d
@@ -171,7 +171,8 @@ Returns:
 enum bool isInputRange(R) =
     is(typeof(R.init) == R)
     && is(ReturnType!((R r) => r.empty) == bool)
-    && is(typeof((return ref R r) => r.front))
+    && (is(typeof(ref (return ref R r) => r.front))
+    || is(typeof((return ref R r) => r.front)))
     && !is(ReturnType!((R r) => r.front) == void)
     && is(typeof((R r) => r.popFront));
 
@@ -250,6 +251,26 @@ enum bool isInputRange(R) =
 
     static assert(isInputRange!R);
     assert(r.equal([ 0, 1, 2 ]));
+}
+
+
+@safe unittest
+{
+    // https://issues.dlang.org/show_bug.cgi?id=14478
+    static struct S
+    {
+        @disable this (this);
+    }
+
+    struct Range(T)
+    {
+        T a;
+        bool empty() { return false; }
+        void popFront() {}
+        ref T front() { return a; }
+    }
+
+    static assert(isInputRange!(Range!S));
 }
 
 /+


aG0aep6G:
With this PR, a type is accepted for `front`:

```d
struct R
{
    bool empty;
    alias front = int;
    void popFront() {}
}
pragma(msg, isInputRange!R); /* Is now true. Used to be false. */
```

RazvanN7:
You are right. Needs more refinement.

MoonlightSentinel:
You could try an additional lambda which returns by ref:

`is(typeof((return` ref R r) => r.front)) || is(typeof(ref (return ref R r) => `r.front))`

jmdavis:
Even if the check were correct, I'm still against this, because in the vast majority of range-based code, it's assumed that the elements are copyable, and so if it becomes possible for ranges to contain elements which cannot be copied, anyone who cares about full correctness is suddenly going to have checks all over the place for whether the element types are copyable, which is a whole lot of extra machinery for no benefit except in rare cases.

Also, if the elements are actually directly in the range type rather than using indirections, the range itself becomes uncopyable, and such a range wouldn't even work with `foreach`, because `foreach` copies the range it's given. And ranges in general are passed by value and must be copyable to work. So, a range wrapping non-copyable elements would have to either generate each element on each call to `front`, or it would have to use indirections, and if it has to use indirections, why not just require that `front` return them by indirection as well (be it by pointer or whatever) just like we do now?

I fully expect that in practice, all anyone would get out of this would be a few functions in Phobos that worked with such ranges, whereas the vast majority of code would still work just fine without at. Rather, we'd just end up having various functions in Phobos be less efficient for the cases where `front` calculates anything (e.g. like happens with `map`), because they would be forced to never store the result of `front` and thus call `front` multiple times any time that it needs to access it multiple times.

Allowing this is akin to allowing the return value of `front` to not be valid after a call to `popFront`, which has caused us all kinds of problems. Unfortunately, there's no way to actually prevent such ranges with `isInputRange`, since that's an issue of runtime behavior and not the type system, but ranges such as `ByLine` don't work with all kinds of code (including `std.array.array` last time I checked), because so much code assumes that it can copy `front` and store the copy.

IMHO, this is a fix for a niche case that causes far more trouble than it's worth.

RazvanN7:
> You could try an additional lambda which returns by ref:
is(typeof((return ref R r) => r.front)) || is(typeof(ref (return ref R r) => r.front))

I've thought about this, buy I was under the impression that ref does not work with return types of lambdas. I'm surprised that it does, I just checked. I updated the solution and I am now running the tests; it seems that it works.

RazvanN7:
>IMHO, this is a fix for a niche case that causes far more trouble than it's worth.

The biggest problem is that phobos assumes that if a range is an Input Range then its elements are copyable. Maybe this would work better in phobos v2 where we would redesign most of it to work with ranges with non-copyable types?

PR #7189, opened 2019-09-20
adds assert messages to std.container.package

burner:


thewilsonator:
Given that the types that contain those asserts are intended to not be instantiated, much less used, i.e. Doc only, I think there is not much value in doing this. 

Making those types conditionally defined as `version(StdDoc)` (or whatever it is) is a better idea.

burner:
yeah I had the same feeling. It might only be a bit out of the scope of this project.

RazvanN7 (approved):

RazvanN7:
I see no harm in adding those messages.

burner:
@RazvanN7 thank you

RazvanN7:
@burner You're welcome! Thanks for your patience!

burner:
> @burner You're welcome! Thanks for your patience!

to be honest, I forgot that I had this ;-)

PR #7194, opened 2019-09-25
BitArray fixes

CyberShadow:


std/bitmanip.d
```diff
@@ -1146,11 +1146,37 @@ public:
...
+        ba[0] = 1;
+        ba.length = 0;
+        ba.length = 1;
+        assert(ba[0] == 0); // OK
+
+        ba.length = 2;
+        ba[1] = 1;
+        ba.length = 1;
+        ba.length = 2;
+        assert(ba[1] == 0); // Fail
```
crocopaw: Shouldn't the comment be removed? With the fix it doesn't fail anymore...

PR #7195, opened 2019-09-25
Make task() example compilable.

veelo:
This example seems to not have worked ever.

You cannot create a delegate to an uninstantiated template function, and for the task() template function to be able to deduce a function, all optional parameters to read() need to be given.

The delegate issue is kind of obvious. The default parameter issue I don't know whether is a compiler limitation that could be lifted or not.

These fixes make the example less pretty, especially compared to the example of the other overload of task(), but is it how it is.

Runnable variant: https://run.dlang.io/is/WAns8z (not for publication)

### DIFF
--- a/std/parallelism.d
+++ b/std/parallelism.d
@@ -847,7 +847,7 @@ void main()
 {
     // Create and execute a Task for reading
     // foo.txt.
-    auto file1Task = task(&read, "foo.txt");
+    auto file1Task = task(&read!string, "foo.txt", size_t.max);
     file1Task.executeInNewThread();
 
     // Read bar.txt in parallel.


veelo:
Thanks.

CyberShadow:
> This example seems to not have worked ever.

According to my local dreg result, it does seem to have worked from 2.053 to 2.059 inclusively.

veelo:
> According to my local dreg result, it does seem to have worked from 2.053 to 2.059 inclusively.

Interesting. Out of curiosity, do you get different versions for both fixes separately?

    auto file1Task = task(&read, "foo.txt", size_t.max);

and

    auto file1Task = task(&read!string, "foo.txt");

CyberShadow:
> `auto file1Task = task(&read, "foo.txt", size_t.max);`

Worked 2.053 to 2.068.2 inclusively.

> `auto file1Task = task(&read!string, "foo.txt");`

Never worked.

This is on Linux; for some versions, `read` had a different signature on Windows and POSIX.

PR #7196, opened 2019-09-25
Remove posix only unittest relying on presence of unzip untility.

crocopaw:
Even on posix systems you cannot rely on unzip being installed. In such cases, the unittest will fail, but should not fail.

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -978,6 +978,12 @@ version (Posix) @system unittest
 {
     import std.datetime, std.file, std.format, std.path, std.process, std.stdio;
 
+    if (executeShell("unzip").status != 0)
+    {
+        writeln("Can't run unzip, skipping unzip test");
+        return;
+    }
+
     auto zr = new ZipArchive();
     auto am = new ArchiveMember();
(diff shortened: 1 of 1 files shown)


atilaneves:
At most, this should be not run on systems that can't, but definitely not deleted.

std/zip.d
```diff
@@ -970,29 +970,3 @@ the quick brown fox jumps over the lazy dog\r
     assert(amAfter.expandedData == am.expandedData);
     assert(amAfter.time == am.time);
 }
-
-// Non-Android Posix-only, because we can't rely on the unzip command being
-// available on Android or Windows
-version (Android) {} else
-version (Posix) @system unittest
-{
```
CyberShadow: Maybe add something like 

```d
if (executeShell("unzip") != 0)
{
   writeln("Can't run unzip, skipping unzip test"; 
   return; 
}
```

crocopaw: Good idea! Thanks! :-)

crocopaw:
I still feel unhealthy with this, but I believe you, that it is ok. I added the check proposed by cybershadow though.

PR #7197, opened 2019-09-25
Reference std.format.formatValue from std.conv

Panke:
Documentation should also say how to define toString


std/conv.d
```diff
@@ -410,6 +410,7 @@ template to(T)
  *   $(LI Pointer to string conversions prints the pointer as a `size_t` value.
  *        If pointer is `char*`, treat it as C-style strings.
  *        In that case, this function is `@system`.))
+ * See $(REF formatValue, std,format) on how toString should be defined.
```
crocopaw: LGTM, although the place, where this points to, is a little bit strange. In my oppinion the documentation for overloading toString should be in https://dlang.org/library/object/object.to_string.html But that's not your fault.

PR #7198, opened 2019-09-25
Fix issue 20239 - chameleon zip-file should be rejected by std.zip

crocopaw:
The lines around 640 where wrongly indented. I corrected them, although this is not directly related to the bug fix. Hope, this doesn't disturb.

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -598,9 +598,7 @@ public:
      */
 
     this(void[] buffer)
-    {   uint iend;
-        uint i;
-        int endcommentlength;
+    {
         uint directorySize;
         uint directoryOffset;
 
@@ -609,32 +607,17 @@ public:
         if (data.length > uint.max - 2)
             throw new ZipException("zip files bigger than 4 GB are unsupported");
 
-        // Find 'end record index' by searching backwards for signature
-        iend = (data.length > 66_000 ? to!uint(data.length - 66_000) : 0);
-        for (i = to!uint(data.length) - endOfCentralDirLength; 1; i--)
+        findEndOfCentralDirRecord();
+        uint i = endrecOffset;
+
+        int endcommentlength = getUshort(i + 20);
+        comment = cast(string)(_data[i + 22 .. i + 22 + endcommentlength]);
+
+        uint k = i - zip64EndOfCentralDirLocatorLength;
+        if (k < i && _data[k .. k + 4] == cast(ubyte[])"PK\x06\x07")
         {
-            if (i < iend || i >= data.length)
-                throw new ZipException("no end record");
-
-            if (_data[i .. i + 4] == endOfCentralDirSignature)
-            {
-                endcommentlength = getUshort(i + 20);
-                if (i + endOfCentralDirLength + endcommentlength > data.length
-                        || i + endOfCentralDirLength + endcommentlength < i)
-                    continue;
-                comment = cast(string)(_data[i + endOfCentralDirLength ..
-                                             i + endOfCentralDirLength + endcommentlength]);
-                endrecOffset = i;
-
-                uint k = i - zip64EndOfCentralDirLocatorLength;
-                if (k < i && _data[k .. k + 4] == zip64EndOfCentralDirLocatorSignature)
-                {
-                    _isZip64 = true;
-                    i = k;
-                }
-
-                break;
-            }
+            _isZip64 = true;
+            i = k;
         }
 
         if (isZip64)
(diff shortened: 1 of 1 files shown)


std/zip.d
```diff
@@ -293,6 +293,10 @@ final class ZipArchive
     import std.conv : to;
     import std.datetime.systime : DosFileTime;
 
+    enum endOfCentralDirRecordLength = 22;
```
CyberShadow: It might be simpler declaring this data structure as a D struct. Then this would be its `sizeof`.

std/zip.d
```diff
@@ -293,6 +293,10 @@ final class ZipArchive
...
+
+    enum endOfCentralDirSignature = cast(ubyte[])"PK\x05\x06";
```
Geod24: This is `public` and breaks the type system (it's mutable).
```suggestion
private static immutable ubyte[] endOfCentraliDirSignature = [ 'P', 'K', 5, 6];
```

std/zip.d
```diff
@@ -583,31 +587,17 @@ final class ZipArchive
...
             throw new ZipException("zip files bigger than 4 GB are unsupported");
 
-        // Find 'end record index' by searching backwards for signature
-        iend = (data.length > 66_000 ? to!uint(data.length - 66_000) : 0);
-        for (i = to!uint(data.length) - 22; 1; i--)
-        {
-            if (i < iend || i >= data.length)
-                throw new ZipException("no end record");
+        findEndOfCentralDirRecord();
+        i = endrecOffset;
```
Geod24: Move the declaration here then
```suggestion
        uint i = endrecOffset;
```

Some for the following variables.

std/zip.d
CyberShadow: > This is public and breaks the type system (it's mutable).

It's an `enum`, so I don't think so?

Geod24: It does, because it casts a string literal (which is stored in `.rodata`) to mutable.
```D
class Foo
{
    enum foo = cast(ubyte[])"PK\x05\x06";
}

void main ()
{
    ubyte[] a = Foo.foo;
    a[0] = 5;
}
```

On my machine (OSX):
```console
dmd -run test.d
Error: program killed by signal 10
```

crocopaw:
> Finding and reading a directory from the end seems like an optimization attempt for slow disk devices. But, `std.zip` already has the data in a `void[] buffer`. In this case, wouldn't it be simpler and more correct to simply walk all archive members from the beginning to the end?

That was my first idea too. But when I dived more deeply into the data format of zip files it was soon clear, that this is impossible. The chunks inside a zip file do not need to be alligned (nor sorted). There can be other stuff before (think of self extracting zip files) and between (e.g. removed files from the archive, where only the entry in the end record has been deleted). Even overlapping is not prohibitted by the specs. This gives a lot of possibilities for malware. Avoiding being trapped by malware makes it necessary to start at the end of central directory record.

Unfortunately zip allowes for a comment (of apriori unknown size) after that record, which again gives leeway for malware (see bug report). I think, best one can do is to scan these 16k of bytes. While the old approach stopped when it found one correct entry, the new version looks if there is another correct entry and rejects the whole zip file in this case. Having more than one correct entry is a string indicator of having to do with malware, because in normal zip files, there should never be comments containing x05/x06 bytes. And the chances, that the data before the intended record form a valid second entry is almost zero too, but in rare circumstances possible.

And one more reason to start at the end: Although the whole file is kept in memory at the moment, I think it might be reasonable to improve this later by replacing the 'void[] buffer' with a random-access-range. This will speed things up in some circumstances (think of a large zip-archive, where only few small files need to be retrieved and therefore only part of the archive needs to be read from file).

std/zip.d
crocopaw: Is planed to be done in a later PR. Maybe even a union can be used to avoid those calles of getUshort/getUint/getUlong.

std/zip.d
crocopaw: This code seems to be old and there are lot of possibilities for refactoring left (e.g. endcommentlength could be defined later too). In the longrun I'd like to get rid of variable i alltogether by putting everything in separate methods and/or using unions/structs to access the data. For the time being I tried to keep as much of the constructor in the old state and focused on the new findEndOfCentralDirRecord method. But maybe I should do some refactoring first, before I continue to work on issue #20027.

crocopaw:
I moved the problematic enum back into the code. I'll cope with those magic constants in a separate PR.

crocopaw:
See PR #7201 for a solution of the problem with the enum.

crocopaw:
Resolved merge conflicts with PR #7201.

PR #7199, opened 2019-09-26
std.zip: Update and correct main documentation

crocopaw:
I replaced the link below "zip archive" to the wikipedia entry, because the old link points to the zip/unzip utility webpage which contains information about this utility but not about zip archieves in general.

Further: From the 6 listed bugs, the first four are no bugs, but just limitations of the current implementation. Bug 592 is "won't fix", refering to D1; Bug 2137 is allready fixed. Therefore I removed the two real bugs and replaced the headline by "Limitations".

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -1,17 +1,15 @@
 // Written in the D programming language.
 
 /**
- * Read/write data in the $(LINK2 http://www.info-zip.org, zip archive) format.
+ * Read/write data in the $(LINK2 https://en.wikipedia.org/wiki/Zip_%28file_format%29, zip archive) format.
  * Makes use of the etc.c.zlib compression library.
  *
- * Bugs:
+ * Limitations:
  *      $(UL
  *      $(LI Multi-disk zips not supported.)
  *      $(LI Only Zip version 20 formats are supported.)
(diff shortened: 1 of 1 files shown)


std/zip.d
```diff
@@ -1,17 +1,15 @@
 // Written in the D programming language.
 
 /**
- * Read/write data in the $(LINK2 http://www.info-zip.org, zip archive) format.
+ * Read/write data in the $(LINK2 https://en.wikipedia.org/wiki/Zip_%28file_format%29, zip archive) format.
  * Makes use of the etc.c.zlib compression library.
  *
- * Bugs:
+ * Limitations:
```
CyberShadow: Worth noting that "Bugs" is a "standard" DDoc section but "Limitations" isn't:

https://dlang.org/spec/ddoc.html#standard_sections

DDoc renders the two a bit differently:

http://dtest.dlang.io/diff/website-4b9608e139d86858b56b33df0deeff2d34c822b1-0a5ced4df5df7b15f158d0f21d39a31b/website-4b9608e139d86858b56b33df0deeff2d34c822b1-4412a24a825aa4718eb7f8ead6eec493/web/phobos-prerelease/std_zip.html

The difference is mainly cosmetic though.

PR #7200, opened 2019-09-26
Add type that is always stored in little endian order.

crocopaw:
For my work at std.zip I could use a type that always stores it's value in little endian, regardless of the architecture used. For me, it looks like it's better to place such a type in bitmanip.d to make it available for everyone. Of course, it's companion BigEndian should also be implemented, but before doing so, I'd be happy to know your oppinion on this.

### DIFF
--- a/std/bitmanip.d
+++ b/std/bitmanip.d
@@ -13,6 +13,7 @@ $(TR $(TD Bit constructs) $(TD
 ))
 $(TR $(TD Endianness conversion) $(TD
     $(LREF bigEndianToNative)
+    $(LREF littleEndian)
     $(LREF littleEndianToNative)
     $(LREF nativeToBigEndian)
     $(LREF nativeToLittleEndian)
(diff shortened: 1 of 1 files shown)


crocopaw:
I'm wondering why circleci is not able to build although auto-tester does. Does circleci use a different compiler version? Or different settings? I tried to find out, but my knowledge of makefiles is too limited and I'm lost here.

thewilsonator:
cc @atilaneves

std/bitmanip.d
```diff
@@ -3272,6 +3273,106 @@ private template canSwapEndianness(T)
...
+
+    test.d = 1;
+
+    version (LittleEndian)
+    {
+        assert(test.a[0]==1 && test.a[1]==0 && test.a[2]==0 && test.a[3]==0);
+    }
+    else
+    {
+        assert(test.a[0]==0 && test.a[1]==0 && test.a[2]==0 && test.a[3]==1);
```
dkorpel: Why are you testing for big endian order on big endian architectures? I thought the entire point of the type was to be the same on both.

atilaneves:
I'm trying to understand what the use case is for caring about the order in memory, as opposed to how it's serialised. In nearly every instance I've seen of code that is specific to endianness, [it wasn't needed](https://commandcenter.blogspot.com/2012/04/byte-order-fallacy.html).

Could you please explain in more detail what you're trying to achieve? Thanks.

crocopaw:
> I'm trying to understand what the use case is for caring about the order in memory, as opposed to how it's serialised. In nearly every instance I've seen of code that is specific to endianness, [it wasn't needed](https://commandcenter.blogspot.com/2012/04/byte-order-fallacy.html).
> 
> Could you please explain in more detail what you're trying to achieve? Thanks.

Of course. I'm working on std/zip.d (zip-files use little endian). This file contains code like 

```
        ...
        directorySize = getUint(i + 12);
        directoryOffset = getUint(i + 16);
        ...
                
        if (directoryOffset + directorySize > i) ...
```

I would like to replace this with something, that's not using these magic constants, but a struct, wrapped in a union, namely:

```
    enum endOfCentralDirLength = 22;

    union endOfCentralDirRecord
    {
        ubyte[endOfCentralDirLength] data;
        struct
        {
            align(1):
            ubyte[4] signature;
            LittleEndian!ushort numberOfDisks;
            LittleEndian!ushort numberOfThisDisk;
            LittleEndian!ushort entriesOfThisDisk;
            LittleEndian!ushort numberOfEntries;
            LittleEndian!uint directorySize;
            LittleEndian!uint directoryOffset;
            LittleEndian!ushort commentLength;
        }
    }

    auto r = endOfCentralDirRecord();
    r.data = data[start .. start + endOfCentralDirLength];

    if (directoryOffset + directorySize > i) ...   
```

I can't see, how I'm falling for the byte-order-falacy. But maybe I'm overlooking something. And I'm open for better ideas. :-)

atilaneves:
@crocopaw When you say "zip-files use little endian", I assume you mean on disk? I still don't see how this has to do with a little endian type and not with serialisation.

crocopaw:
> @crocopaw When you say "zip-files use little endian", I assume you mean on disk? I still don't see how this has to do with a little endian type and not with serialisation.

Yes, on disk. At the moment, the constructor of the class ZipArchive takes an array with the whole file. (In the longrun I want to replace this by a (random access) range of ubytes, but that's probably an other topic.) The question is, how to come from the ubytes to, say ints. The formular given in the article you linked to is nice, but this needs to be applied to every single variable. I hoped for something, where I can just define a struct, that matches the records used in a zip file, so I don't have to apply the formula to every variable explicitly. At the moment std.zip uses littleEndianToNative, which uses a swap algorithm, not the formula given in the article.

atilaneves:
> The question is, how to come from the ubytes to, say ints

Serialisation.

> but this needs to be applied to every single variable

Which is tedious in a language without static reflection, but not D.

> I hoped for something, where I can just define a struct, that matches the records used in a zip file, so I don't have to apply the formula to every variable explicitly

The best way to do that in general would be to use a serialisation library that handles little-endian as a compile-time option (or UDAs on variables that are supposed to be serialised as little-endian).

crocopaw:
Thanks for your ideas. I think I got the main picture now. Still, it feels a little bit beyond my current abilities. (I did use serialisation in Java years ago, but never used it in D though.) I will take some time, to think about how this ideas can be implemented in case of std.zip.

PR #7201, opened 2019-09-27
std.zip: Replace magic constants by constants named after specs

crocopaw:
This PR removes three public items: digiSignLength, eocd64LocLength and eocd64Length. They are replaced by private enums with names taken directly from the specs, making all names consistent.

Theoretically they should go through a deprecation process. But it's hard to imagine a usecase, where someone used these constants outside of std.zip. In my oppinion they should have been private from the very beginning.

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -291,6 +291,27 @@ final class ZipArchive
     import std.conv : to;
     import std.datetime.systime : DosFileTime;
 
+private:
+    // names are taken directly from the specification
+    // https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
+    static immutable ubyte[] centralFileHeaderSignature = [ 0x50, 0x4b, 0x01, 0x02 ];
+    static immutable ubyte[] localFileHeaderSignature = [ 0x50, 0x4b, 0x03, 0x04 ];
+    static immutable ubyte[] endOfCentralDirSignature = [ 0x50, 0x4b, 0x05, 0x06 ];
+    static immutable ubyte[] archiveExtraDataSignature = [ 0x50, 0x4b, 0x06, 0x08 ];
+    static immutable ubyte[] digitalSignatureSignature = [ 0x50, 0x4b, 0x05, 0x05 ];
+    static immutable ubyte[] zip64EndOfCentralDirSignature = [ 0x50, 0x4b, 0x06, 0x06 ];
+    static immutable ubyte[] zip64EndOfCentralDirLocatorSignature = [ 0x50, 0x4b, 0x06, 0x07 ];
+
+    enum centralFileHeaderLength = 46;
+    enum localFileHeaderLength = 30;
+    enum endOfCentralDirLength = 22;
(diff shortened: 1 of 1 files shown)


std/zip.d
```diff
@@ -291,6 +291,27 @@ final class ZipArchive
     import std.conv : to;
     import std.datetime.systime : DosFileTime;
 
+private:
+    // names are taken directly from the specification
+    // https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
+    enum centralFileHeaderSignature = [ 0x50, 0x4b, 0x01, 0x02 ];
```
CyberShadow: This is an array of `int`s:

```d
enum centralFileHeaderSignature = [ 0x50, 0x4b, 0x01, 0x02 ];
pragma(msg, typeof(centralFileHeaderSignature[0])); // int
```

Assigning it to an `ubyte[]` will work, but it would probably be better to fix this anyway.

std/zip.d
```diff
@@ -302,9 +323,6 @@ final class ZipArchive
     private uint _totalEntries;
     private bool _isZip64;
     static const ushort zip64ExtractVersion = 45;
-    static const int digiSignLength = 6;
-    static const int eocd64LocLength = 20;
-    static const int eocd64Length = 56;
```
CyberShadow: These constants were public, so I think they would need to go through a deprecation cycle.

std/zip.d
crocopaw: It's a bad idea to use enums for arrays anyway. But as far as I know, the ints would not be a problem, because the compiler notices at the places where this enum is used, that this should be bytes and uses bytes. So there are only ints because you queried them with the pragma. A little bit like the "Uncertainty principle" from physics. Unfortunately I cannot find the quote in the documentation, where I read this.

std/zip.d
crocopaw: Is I wrote in my comment above, I felt a little bit unsure if they should be deprecated, but indeed, it should. ;-)

std/zip.d
```diff
@@ -302,9 +323,9 @@ final class ZipArchive
...
+    deprecated("Use digitalSignatureLength instead") static const int digiSignLength = 6;
+    deprecated("Use zip64EndOfCentralDirLocatorLength instead") static const int eocd64LocLength = 20;
+    deprecated("Use zip64EndOfCentralDirLength instead") static const int eocd64Length = 56;
```
CyberShadow: These could also be aliases to the new definitions, but this is fine too.

I think according to https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1013.md#summary-of-the-steps-in-the-normal-deprecation-process this should have something like "will be removed in 2.098.0" so we know when we can remove them.

crocopaw: 2.093.0 is sufficient as there are two releases a year, isn't it?

CyberShadow: I think we decided to use a fixed number of major versions as we can't really predict when releases will occur.

PR #7202, opened 2019-09-27
std.zip: Add data structure for detecting overlapping chunks.

crocopaw:
This is a preparation step for recognising overlapping data chunks, which eventually will solve the main problem of issue #20027.

Later, each found chunk will be removed from _seg. When overlapping chunks exist, removeSegment will notice, that an allready removed part shall be removed and throw an exception.

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -306,6 +306,8 @@ final class ZipArchive
     static const int eocd64LocLength = 20;
     static const int eocd64Length = 56;
 
+    private Segment[] _segs;
+
     /// Read Only: array representing the entire contents of the archive.
     @property @safe @nogc pure nothrow ubyte[] data() { return _data; }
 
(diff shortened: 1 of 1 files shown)


std/zip.d
```diff
@@ -823,6 +825,78 @@ final class ZipArchive
...
+    // throws zipException if start .. end is not completely available in _segs;
+    void removeSegment(uint start, uint end) pure @safe
+    {
+        if (_segs.length == 0)
+            throw new ZipException("overlapping data detected");
+
+        uint best = 0;
+        size_t bestPos = 0;
+        foreach (i,seg;_segs)
+            if (seg.start <= start && seg.start > best)
```
CyberShadow: `seg.start > best` will fail to find a segment that starts at offset 0 but is not the first segment in `_segs`. Not sure if this will affect how you will use this.

std/zip.d
```diff
@@ -823,6 +825,78 @@ final class ZipArchive
...
+    // defines a segment of the zip file, including start, excluding end
+    struct Segment
+    {
+        uint start;
+        uint end;
+    }
+
+    // removes Segment start .. end from _segs
+    // throws zipException if start .. end is not completely available in _segs;
+    void removeSegment(uint start, uint end) pure @safe
```
CyberShadow: Perhaps add a contract here, e.g. `in { assert(start > end); }`.

std/zip.d
```diff
@@ -823,6 +825,78 @@ final class ZipArchive
...
+
+    /* ============== for detecting overlaps =============== */
+
+private:
+
+    // defines a segment of the zip file, including start, excluding end
+    struct Segment
+    {
+        uint start;
+        uint end;
```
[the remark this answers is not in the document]
crocopaw: The whole file uses uint for indexing the data structure and rejects files larger than uint.max-2. For consistency I used it here too. I'm thinking about switching to ulong to be able to handle longer files too. In the list given in issue #20027 there is stated, that longer files should be rejected. I havn't found out yet, why. Before I'm going to switch to ulong I'll have to investigate on this.

std/zip.d
```diff
@@ -823,6 +825,78 @@ final class ZipArchive
...
+    // removes Segment start .. end from _segs
+    // throws zipException if start .. end is not completely available in _segs;
+    void removeSegment(uint start, uint end) pure @safe
+    {
+        if (_segs.length == 0)
+            throw new ZipException("overlapping data detected");
+
+        uint best = 0;
+        size_t bestPos = 0;
+        foreach (i,seg;_segs)
```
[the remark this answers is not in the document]
crocopaw: No. The segments will represent the parts *not* identified. 

In theory this can indeed lead to an overall performance of O(n^2), when the files are randomly distributed inside the zip file. But in normal (todays) zip archives they are lined up and in this case at most 3 segments are needed (one at the start if it's selfextracting, one between the files and the directory and one at the end of the directory), making this foreach O(1). I think, making this faster could be an improvement, left for later. But for the time being, I'd like to stick to a simple algorithm to avoid bugs.

std/zip.d
crocopaw: This is indeed a bug. I'll fix this in a few hours.

std/zip.d
crocopaw: Good idea. :-)

PR #7205, opened 2019-09-29
Fix Issue 15230 - Inconsistent std.range.SortedRange predicate checks

crocopaw:
This may reopen issue 12661. I checked this and I think it does not reopen. But IMHO someone else should confirm this.

Anyway, I think, the original fix for 12661 did not fix the real problem, but was more of a hack inducing issue 15230.

### DIFF
--- a/std/algorithm/sorting.d
+++ b/std/algorithm/sorting.d
@@ -1954,7 +1954,7 @@ if (((ss == SwapStrategy.unstable && (hasSwappableElements!Range ||
 
     // This should get smaller with time. On occasion it may go larger, but only
     // if there's thorough justification.
-    debug enum uint watermark = 1676280;
+    debug enum uint watermark = 1676220;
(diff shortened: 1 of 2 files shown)


crocopaw:
I'd like to add the WIP label, but could not find out, how to do so... Maybe I've got not the rights to do so...

crocopaw:
While thinking about this issue I had to make up my mind about debug statements inside phobos. In my oppinion they should not appear (aside from code used for debugging phobos itself). It's a feature, that should be available for users to use to debug their own code. If the behaviour of phobos depends on being compiled with or without -debug, debugging usercode is complicated unnecessary. This may also lead to people not using pure, because "debug writeln(...)"  and "writeln(...)" might lead to different results.

thewilsonator:
Is this still WIP?

crocopaw:
No. :-)

PR #7207, opened 2019-10-01
Resolve expi() documentation issues due to deprecation.

veelo:
Fix issue 19733: https://issues.dlang.org/show_bug.cgi?id=19733

Nowadays, the x87 instruction fsincos is slower, not faster, than computing sin and cos separately.
https://stackoverflow.com/questions/12485190/calling-fsincos-instruction-in-llvm-slower-than-calling-libc-sin-cos-functions
Therefore, there is no need for assembly in the expi() implementation, which was removed in commit 029440c.


thewilsonator:
Please amend the _commit_ message to say "Fix issue 19733: ..." so the bot picks it up.

veelo:
> Please amend the _commit_ message to say "Fix issue 19733: ..." so the bot picks it up.

Will try.

kinke:
As stated in bugzilla, I [saw](https://github.com/ldc-developers/ldc/pull/2855#issuecomment-491560143) a significant boost (~5x faster) with the x87 asm version with LDC on my i5-3550. The linked StackOverflow question isn't related to 80-bit `real` at all.

veelo:
@kyllingstad For what reasons was the x87 asm version removed from std.complex.expi? 029440c2f9008ff2fe8803c97900394bfe9b4e4e

berni44:
Just noticed this - I wrote a similar PR yesterday, which is meanwhile merged. Hence, this PR can be closed. ("Having Fix - issue 19733 [...]" in the commit message would have maked the bot adding a link to this PR in bugzilla and I would have known of it earlier.)

PR #7210, opened 2019-10-03
Set up CI with Azure Pipelines

wilzbach:


wilzbach:
Sorry, unfortunately this PR is required when setting up Azure Pipelines.

PR #7211, opened 2019-10-03
add std.conv.hexData()

WalterBright:
This is an alternative to `hexString` based on usage feedback from Dunnhumby.

1. does not use templates so it does not have problems with very large strings nor object file bloat
2. `wstring` and `dstring` are not supported, as it is hard to imagine a compelling use case for them
3. `ubyte[]` is returned instead of `string` as it only makes sense for hex data to initialize `ubyte` arrays, not strings
4. return value is `immutable` so it can be used to initialize global tables with need for casting
5. works with any string - no error cases. Non-hex digits are treated as separators
6. it is intended to be used in CTFE code as a better replacement for the removed `x` string core language feature
7. it has no dependencies on anything else, and so can be cut-pasted into user code that does not rely on Phobos.

It is designed to work with https://github.com/dlang/DIPs/pull/177 although that is not required for `hexData()` to be useful.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -6031,6 +6031,10 @@ Params:
 
 Returns:
     a `string`, a `wstring` or a `dstring`, according to the type of hexData.
+
+References:
+    Prefer using $(LREF hexData) instead, especially if the `hexData` string is long
+    because it does not use templates.
  */
 template hexString(string hexData)
 if (hexData.isHexLiteral)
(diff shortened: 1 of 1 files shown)


std/conv.d
```diff
@@ -6136,6 +6140,88 @@ private auto hexStrLiteral(String)(scope String hexData)
...
+            append(cast(ubyte)((u << 4) + c));
+            digits = 0;
+        }
+        else
+            u = c;
+    }
+    if (digits)
+        append(u);
+
+    return cast(immutable)result[0 .. i];
```
thewilsonator: The convention is to use assumeUnique

std/conv.d
```diff
@@ -6136,6 +6140,88 @@ private auto hexStrLiteral(String)(scope String hexData)
...
+        else
+        {
+            if (digits)
+                append(u);
+            u = 0;
+            digits = 0;
+            continue;
+        }
+
+
```
thewilsonator: One line space will suffice.

std/conv.d
WalterBright: I wanted this function to not have dependencies on anything but the gc.

std/conv.d
WalterBright: I use two lines to separate one function from the next unrelated one.

WalterBright:
The buildkite error is:
```
generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std5regex8internal9kickstart__T7ShiftOrTaZQl6__ctorMFNcNeKSQCiQChQCe2ir__T5RegexTaZQjAkZSQDmQDlQDiQDc__TQCvTaZQDb: error: undefined reference to '_D3std5range__T12assumeSortedVAyaa6_61203c3d2062TAkZQBlFNaNbNiNfQpZSQCoQCn__T11SortedRangeTQBqVQCna6_61203c3d2062ZQBl'
```
which has nothing to do with this PR.

wilzbach:
Yeah the druntime benchmark check uses a wrong version of phobos.

WalterBright:
@wilzbach so what's the fix?

thewilsonator:
Nothing, as Buildkite is not required to pass to pull. You should double-check that the benchmark is the _only_ failure though.

std/conv.d
```diff
@@ -6136,6 +6140,88 @@ private auto hexStrLiteral(String)(scope String hexData)
...
+ * at compile time to, for example, initialize manifest constants
+ * and statically initialize global variables.
+ * Params:
+ *      s = string of data to convert
+ * Returns:
+ *      unique immutable array of ubytes allocated on the GC heap
+ * References:
+ *      $(LREF hexString)
+ */
+immutable(ubyte)[] hexData(const(char)[] s) pure nothrow @trusted
```
CyberShadow: Since the function is `pure`, it doesn't *have* to be immutable, right? Making the return type mutable (`ubyte[]`) still allows initializing immutable data because it's `pure`.

https://run.dlang.io/is/eH2zDC

WalterBright: you're right

std/conv.d
```diff
@@ -6136,6 +6140,88 @@ private auto hexStrLiteral(String)(scope String hexData)
...
+    enum a1 = hexData("AAbb");
+    assert(a1 == [0xAA, 0xBB]);
+    enum a2 = hexData("");
+    assert(a2 == []);
+    enum a3 = hexData(" \t\r\n");
+    assert(a3 == []);
+    enum a4 = hexData("f");
+    assert(a4 == [0xF]);
+    enum a5 = hexData(" f,");
+    assert(a5 == [0xF]);
```
CyberShadow: Why support odd number of digits? This is probably going to almost always be an error or typo.

std/conv.d
```diff
@@ -6136,6 +6140,88 @@ private auto hexStrLiteral(String)(scope String hexData)
...
+    enum a6 = hexData("0BF ");
+    assert(a6 == [0x0B, 0x0F]);
```
CyberShadow: There should probably be some tests to test the behavior with invalid inputs. Speaking of which, they seem to be ignored? Combined with the decision to support one-digit characters, this seems like a bad idea, e.g. in `"54 32 1O"` the last character is the letter O.

ntrel: G-Z should be an error, that's bug prone.

std/conv.d
```diff
@@ -6136,6 +6140,88 @@ private auto hexStrLiteral(String)(scope String hexData)
...
+ * References:
+ *      $(LREF hexString)
+ */
+ubyte[] hexData(const(char)[] s) pure nothrow @safe
+{
+    ubyte[] result;
+    result.length = (s.length + 1) / 2;
+    size_t i;
+
+    void append(ubyte u)
```
andralex: this seems a bit much, it's called from three places

std/conv.d
n8sh: As currently specified there are no invalid inputs to this function.

RazvanN7:
@WalterBright a rebase will most likely fix the buildkite error and then we are good to go.

thewilsonator:
Hex string were added back into the language, there is no need for this.

PR #7216, opened 2019-10-08
Refactor std.zip: Split constructor in several methods.

berni44:


CyberShadow:
Aimless refactorings are generally frowned upon here, so it would be good to have some rationale for what you aim to achieve with these changes and how they fit in in the bigger plan.

The best place to include this rationale would be in the commit message for the change, describing what benefit the change brings.

berni44:
> Aimless refactorings are generally frowned upon here, so it would be good to have some rationale for what you aim to achieve with these changes and how they fit in in the bigger plan.

I'm sorry for confusion. I contributed last weeks with the account crocopaw, but have to delete this account soon. Therefore, and because I want to unify my accounts, I switched to a new one. The bigger plan is to resolve issue 20027, making zip rejecting malware (and maybe later add some more features like support for encryption, multidisks and bzip compression). This task is easier when doing some refectoring first. And I also hope that this enables me to parallelise some PRs.

CyberShadow:
> I'm sorry for confusion. I contributed last weeks with the account crocopaw, but have to delete this account soon. Therefore, and because I want to unify my accounts, I switched to a new one.

Oh, I didn't even notice this PR came from a different account than the previous std.zip ones :)

> The bigger plan is to resolve issue 20027, making zip rejecting malware (and maybe later add some more features like support for encryption, multidisks and bzip compression). This task is easier when doing some refectoring first.

Could you please go into more detail how each change here helps you achieve that end goal, and add that to the commit messages?

berni44:
> Could you please go into more detail how each change here helps you achieve that end goal, and add that to the commit messages?

It's difficult to answer here. It took me by some surprise, that refectorings aren't welcome. From [1] I concluded, that I actually should refactor the code and that I should use separate PRs for that. But when you say it isn't like this, I'll probably have to accept that. I think, that the reason for rejecting refactorings is to avoid new bugs. While I can understand that reasoning, my experience is, that refactoring will also reveal hidden bugs and it prevents new bugs creeping in. In my oppinion, there are several corners in Phobos, that would benefit from some refactoring.

I think, this file is one of them. The code is old. As far as I can see, it's been written long before 2007, when D2 did not exist. So noone to blame for that. At that time this was probably good and clean code, but D has changed since. And Phobos should keep pace.

The oldest version of this constructor found in the git archive had about 95 lines. During time this has grown to 150 lines. I prefer functions that fit at least on one screen, which is about 50 lines. Obviously I could add the checks for overlaps to this function without splitting it. That would increase the size again. But making a function longer, that's allready much too long doesn't feel good to me. I actually would even prefer to split extractEndOfCentralDirRecord once more (separating the parts for Zip64 and for Zip32). 

Now I don't know, how to continue...

[1] https://wiki.dlang.org/Starting_as_a_Contributor#Characteristics_of_a_good_pull_request

PR #7217, opened 2019-10-08
Remove two duplicate unittests from std.bitmanip;

berni44:


std/bitmanip.d
```diff
@@ -309,51 +316,6 @@ one bitfield with an empty name.
...
 
-/**
-Creates a bitfield pack of eight bits, which fit in
-one `ubyte`. The bitfields are allocated starting from the
-least significant bit, i.e. x occupies the two least significant bits
-of the bitfields storage.
-*/
-@safe unittest
-{
-    struct A
```
thewilsonator: Thanks, this one looks good to remove...

std/bitmanip.d
```diff
@@ -309,51 +316,6 @@ one bitfield with an empty name.
...
-    assert(obj.x == 2);
-    assert(obj.y == 0);
-    assert(obj.z == 2);
-    assert(obj.flag == false);
-}
-
-/// Add empty fields for padding to have a total bit length of 8, 16, 32, or 64
-@safe unittest
-{
-    struct A
```
thewilsonator: ...this `A` however is different from the one above. Please leave this one as is.

std/bitmanip.d
```diff
@@ -294,6 +295,12 @@ one bitfield with an empty name.
             uint, "",         6));
     }
 
+    A a;
+    assert(a.flag1 == 0);
+    a.flag1 = 1;
+    assert(a.flag1 == 1);
+    a.flag1 = 0;
+    assert(a.flag1 == 0);
```
thewilsonator: which means this can also be left out.

std/bitmanip.d
berni44: Where's the difference? I can't see any. I even tried the diff utility...

thewilsonator:
```diff
-             uint, "x",    2,		
-             int,  "y",    3,		
-             uint, "z",    2,		
-             bool, "flag", 1));
+            bool, "flag1",    1,		
+             bool, "flag2",    1,		
+             uint, "",         6));
```

berni44: Ah no, this are not the two I compared: See lines 289 and 342 in the original file. They are both this 1, 1, 6 pattern. Just the comment is different.

thewilsonator: Ah, my bad.

PR #7218, opened 2019-10-09
Add more unittests for CustomFloat.

berni44:
Before going to fix issues #20261, #20263, #20281, #20282, #20283, #20284 and #20286 I want to make sure, that I don't break the stuff, that's allready working.

Corrected some small style mistakes too.


atilaneves:
Can any of those asserts be static asserts instead?

berni44:
> Can any of those asserts be static asserts instead?

Yes, I changed them.

PR #7208, opened 2019-10-02
Fix issue #20260 - CustomFloat with 0 precision/exponentWidth

crocopaw:

PR #7219, opened 2019-10-09
Fix issues 20260, 20281, 20284 - Limitations of CustomFloat.

berni44:
Not all combinations of template parameters for CustomFloat make sense. Most of them do not even work and yield strange error messages. I tried to narrow the bounds to catch every combination that doesn't work, with template constraints making for better error messages. This PR superseeds PR #7208.

### DIFF
--- a/std/numeric.d
+++ b/std/numeric.d
@@ -171,8 +171,7 @@ struct CustomFloat(uint             precision,  // fraction bits (23 for float)
                    uint             exponentWidth,  // exponent bits (8 for float)  Exponent width
                    CustomFloatFlags flags,
                    uint             bias)
-if (((flags & flags.signed)  + precision + exponentWidth) % 8 == 0 &&
-    precision + exponentWidth > 0)
+if (isCorrectCustomFloat(precision, exponentWidth, flags))
 {
     import std.bitmanip : bitfields;
     import std.meta : staticIndexOf;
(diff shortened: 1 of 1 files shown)


std/numeric.d
```diff
@@ -171,8 +171,24 @@ struct CustomFloat(uint             precision,  // fraction bits (23 for float)
                    uint             exponentWidth,  // exponent bits (8 for float)  Exponent width
                    CustomFloatFlags flags,
                    uint             bias)
-if (((flags & flags.signed)  + precision + exponentWidth) % 8 == 0 &&
-    precision + exponentWidth > 0)
+    // Restrictions from bitfield
+if ((((flags & flags.signed) + precision + exponentWidth) == 8 ||
+    ((flags & flags.signed) + precision + exponentWidth) == 16 ||
+    ((flags & flags.signed) + precision + exponentWidth) == 32 ||
+    ((flags & flags.signed) + precision + exponentWidth) == 64 ||
```
berni44: The former check only tested if the number is dividable by 8. But bitfield is more restrictive. To catch all cases, that don't work, I added the four numbers, that work, explicitly.

std/numeric.d
```diff
@@ -171,8 +171,24 @@ struct CustomFloat(uint             precision,  // fraction bits (23 for float)
...
+    precision == 64 &&  // due to CustomFloat!80 support hack, see below
+        ((flags & flags.signed) + exponentWidth == 8 ||
+        (flags & flags.signed) + exponentWidth == 16 ||
+        (flags & flags.signed) + exponentWidth == 32 ||
+        (flags & flags.signed) + exponentWidth == 64)) &&
```
berni44: As bitfields does not allow for more than 64 bits, there is a hack in here, which makes the mantissa a ulong variable, using only the other parts in the bitfield. These lines address this the same way, the tests above did in the standard case.

std/numeric.d
```diff
@@ -171,8 +171,24 @@ struct CustomFloat(uint             precision,  // fraction bits (23 for float)
...
+    // mantissa needs to fit into real mantissa
+    (precision <= real.mant_dig - 1 || precision == 64) &&
```
berni44: To be able to use the custom floats, they need to fit into real. Therefore the number of bits in the mantissa needs to be smaller or equal to the number of bits in the mantissa of a real number. (mant_dig counts the not saved leading 1, while precision doesn't therefore we have to reduce mant_dig by one.) Again we have to take care for the special case.

std/numeric.d
```diff
@@ -171,8 +171,24 @@ struct CustomFloat(uint             precision,  // fraction bits (23 for float)
...
+    // exponent needs to fit into real exponent
+    1L << exponentWidth - 1 <= real.max_exp &&
```
berni44: This is the same for the exponent: The number of bits in the exponent may not exceed the number of bits in the exponent of real for the same reason. Because there is no property telling the number of bits in the exponent (e) of a type, I resorted to max_exp which is 2^^(e-1). To avoid complications with rounding errors of floating point calculation, I don't use log2() but <<. The tests before allready made sure, that exponentWidth<=64 and therefore no overflow will happen.

std/numeric.d
```diff
@@ -171,8 +171,24 @@ struct CustomFloat(uint             precision,  // fraction bits (23 for float)
...
+    // mantissa should have at least one bit
+    precision > 0 &&
```
berni44: The current implementation assumes, that the mantissa has at least one bit. Therefore we make sure, that there is at least one.

std/numeric.d
```diff
@@ -171,8 +171,24 @@ struct CustomFloat(uint             precision,  // fraction bits (23 for float)
...
+    // exponent should have at least one bit, in some cases two
+    exponentWidth > ((flags & (flags.allowDenorm | flags.infinity | flags.nan)) != 0))
```
berni44: The current implementation assumes, that the exponent has at least one bit. In some cases, it is wishable to have at least two bits, namely, when denormalised numbers are used or infinity or nan is available. In this case, most operations don't work correctly with only one bit, mainly, because the result of the operation cannot be represented. These types cause much trouble and I think it's worth to disallow them. (I can't think of any usecase for them anyway.)

thewilsonator:
Please make that template constraint a private function as it is very hard to read otherwise.

thewilsonator:
Please refactor that first clause, it doesn't need to be based on a single expression. Otherwise looks great.

berni44:
> Please refactor that first clause, it doesn't need to be based on a single expression. Otherwise looks great.

You're right. Somehow (don't know why) I thought I cannot use local variables... Sorry.

thewilsonator:
One final thing, could you put back the comment about the 80-floact hack with 64-bit precision. Otherwise this looks good to go.

thewilsonator:
Auto-merge toggled on

std/numeric.d
kinke: This broke the unittests for the following now erroneous instantiation, for all targets with a `real` exponent of less than 15 bits, i.e., about all non-x86 and non-AArch64 targets out there + MSVC targets for LDC: https://github.com/dlang/phobos/blob/5eddbcbac2e23c4d72a06d68149d979dd72f9199/std/numeric.d#L623

berni44: > This broke the unittests for the following now erroneous instantiation, for all targets with a `real` exponent of less than 15 bits, i.e., about all non-x86 and non-AArch64 targets out there + MSVC targets for LDC:

I've got some ideas, but I'm too tired to figure this out now. I'll have a closer look tomorrow. Probably this unittest needs to be removed, when real.max_exp < 15.

kinke: Okay thx for taking care of this. - `CustomFloat!(1, 7, CustomFloatFlags.ieee ^ CustomFloatFlags.signed)` seems to work as well.

This fails too with 64-bit `real`: https://github.com/dlang/phobos/blob/5eddbcbac2e23c4d72a06d68149d979dd72f9199/std/numeric.d#L1510

berni44: Funnily, I finally came up with that `CustomFloat!(1, 7, ...)` too. See PR #7328.

The second issue is unrelated to this PR. It's caused by PR #7267. I'll take care of this too.

PR #7221, opened 2019-10-10
Fix issue 20287 - std.zip: Wrong compressed data

berni44:
While working on issue 20027 I noticed this bug. It's there because someone assumed, that the namelength and extralength are the same in the central directory and the local file header. This is not necessarily the case, even in correct files (assuming, that my local zip utility produces correct files).

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -723,7 +723,11 @@ public:
             de.comment = cast(string)(_data[i .. i + commentlen]);
             i += commentlen;
 
-            immutable uint dataOffset = de.offset + localFileHeaderLength + namelen + extralen;
+            auto localFileHeaderNamelen = getUshort(de.offset + 26);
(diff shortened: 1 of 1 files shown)


berni44:
> When would these values be different in the directory and the local header in a valid zip file? 

The name fields should not differ, but can. It's one of the next issues I'll address to reject such files. The extra field contains 3rd party stuff, that's not completely covered in the specs. In this file, there are two entries: Extended time stamp (0x5455) and something, that I did not manage to find out yet (0x7875) which is the same in both places. The two entries differ in the time stamp where I don't know the exact format neither. But I guess it's more accurate at one place (9 bytes), than at the other (5 bytes). Or one is localised, one not, or such stuff. 

> What values does your zip utility produce?

The unittest is directly taken from that utility. You can find the two entries at position 0x22 and 0x75.

PR #7222, opened 2019-10-10
Fix Issue 20288 - std.format with separator and double causes RangeError

burner:
std.format with separator and double causes RangeError

https://issues.dlang.org/show_bug.cgi?id=20288

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -1857,9 +1857,12 @@ private void testFloatingToIntegral(Floating, Integral)()
         foreach (T; AllNumerics)
         {
             T a = 42;
-            assert(to!string(a) == "42");
-            assert(to!wstring(a) == "42"w);
-            assert(to!dstring(a) == "42"d);
+            string s = to!string(a);
+            assert(s == "42", s);
+            wstring ws = to!wstring(a);
+            assert(ws == "42"w, to!string(ws));
+            dstring ds = to!dstring(a);
+            assert(ds == "42"d, to!string(ds));
             // array test
             T[] b = new T[2];
             b[0] = 42;
@@ -3065,7 +3068,7 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
 @safe unittest
 {
     import std.exception;
-    import std.math : isNaN, fabs;
+    import std.math : isNaN, fabs, isInfinity;
 
     // Compare reals with given precision
     bool feq(in real rx, in real ry, in real precision = 0.000001L)
@@ -3123,6 +3126,16 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
     d = to!double("1.79769e+308");
     assert(to!string(d) == to!string(1.79769e+308));
     assert(to!string(d) == to!string(double.max));
+
+    auto z = real.max / 2L;
+    static assert(is(typeof(z) == real));
+    assert(!isNaN(z));
+    assert(!isInfinity(z));
+    string a = to!string(z);
+    real b = to!real(a);
+    string c = to!string(b);
+
+    assert(c == a, "\n" ~ c ~ "\n" ~ a);
 
     assert(to!string(to!real(to!string(real.max / 2L))) == to!string(real.max / 2L));
 
--- a/std/format.d
+++ b/std/format.d
@@ -2571,9 +2571,25 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     import std.algorithm.comparison : min;
     import std.algorithm.searching : find;
     import std.string : indexOf, indexOfAny, indexOfNeither;
+    import std.math : isInfinity, isNaN, signbit;
+    import std.ascii : isUpper;
+
+    string nanInfStr(scope const ref FormatSpec!Char f, const bool nan,
+            const bool inf, const int sb, const bool up) @safe pure nothrow
+    {
+        return nan
+            ? up
+                ? sb ? "-NAN" : f.flPlus ? "+NAN" : "NAN"
+                : sb ? "-nan" : f.flPlus ? "+nan" : "nan"
+            : inf
+                ? up
+                    ? sb ? "-INF" : f.flPlus ? "+INF" : "INF"
+                    : sb ? "-inf" : f.flPlus ? "+inf" : "inf"
+                : "";
+    }
 
     FloatingPointTypeOf!T val = obj;
-    const spec = f.spec;
+    const char spec = f.spec;
 
     if (spec == 'r')
     {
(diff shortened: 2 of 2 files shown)


std/format.d
```diff
@@ -6872,3 +6875,15 @@ char[] sformat(Char, Args...)(return scope char[] buf, scope const(Char)[] fmt,
...
+// Issue 20288
+@safe unittest
+{
+    import std.format;
+    string s = format("%,.2f", double.nan);
+    assert(s == "nan", s);
+
+    import std.format;
+    s = format("%,.2F", double.nan);
+    assert(s == "NAN", s);
```
John-Colvin: how about `-nan`?

burner: posix looks good, windows 32 not so much.

CyberShadow:
Please put the issue title in the commit/PR title, it's impossible to know what "Fix Issue 20288" means without further action.

wilzbach:
> Robert Schadek: I need a win32_64 box

This should help -> https://github.com/dlang/phobos/pull/7226 (Azure 10 parallel boxes and thus drastically faster start times).

std/format.d
burner: works as well, see the tests

berni44:
Shouldn't "-nan" not exist at all? I'm aware, that there are several ways to express nans in the IEEE standard, but I always thought, that D does not distinguish between them and therefore should still output "nan" instead of "-nan" when the sign bit of a nan is set.

I'd like to point to [issue 19878](https://issues.dlang.org/show_bug.cgi?id=19878) which questions the result "-nan"  when using writeln.

Is there anything official about using -nan or not? I was not able to find anything useful.

burner:
@berni44 I looked at the standard. Section 6.3 of 754 2008 states that the standard does not interpret the sign bit of NaNs, but some operation specify it. IMHO phobos should there display the sign.

berni44:
rikki cattermole wrote in the [Forum](https://forum.dlang.org/post/qoipqf$j8c$1@digitalmars.com) in a related thread, that the behaviour of printf is the baseline. According to this -nan is correct and issue 19878 should be marked "invalid".

burner:
printf on which platfrom, win32 win64, glibc, bsd? they all do things differently.

atilaneves:
> printf on which platfrom, win32 win64, glibc, bsd? they all do things differently.

Shouldn't that be covered by the C standard?

burner:
I don't have the latest standard at hand, but in the latest publicly available one
http://www.open-std.org/JTC1/SC22/WG14/www/docs/n1570.pdf

on page 312 it states that 

> Adouble argument representing a NaN is converted in one of the styles
> [-]nan or [-]nan(n-char-sequence) — which style, and the meaning of
> any n-char-sequence, is implementation-defined. 

The phrase implementation-defined appears quite a lot in the paragraph.
real values are not mentioned at all.

More interesting is footnote 277
> When applied to infinite and NaN values, the -, +, and space flag characters have their > usual meaning; the # and 0 flag characters have no effect

I therefore think that the implemented behavior of this PR is "correct" in that it makes all platforms act as linux 64bit.
Which I turn IMHO is the sanest of them all (sign is preserved)

berni44:
Interesting. So printf and the C documentation is not really a reliable source...

Maybe we (= D community) have to define ourselves, what format() should do and clearly document that. Looks like lot of work...

burner:
eventually yes, but not in this PR. This PR fixes a crash and makes a minor thing behave equal across all platforms.

Doing the thing you mentioned, would firstly require phobos to have its own float formatting. Currently we call snprintf.

But again, the this PR does a tiny narrow thing. And shouldn't grow to do all that.

berni44:
Yes, I agree. In my oppinion, this PR can be merged.

Beyond this PR, for me the question is, what's better: Do more of such corrections or provide a complete rewrite of formattedWrite and Co. using regex for parsing the format string, providing our own float formatting and so on. When looking into bugzilla, there are lots of bugs in here and correcting them might be more work than a sound rewrite.

burner:
I think doing our own formatImpl from float/double/real to string would be pretty good already.

berni44:
Having had a closer look at the files, I don't understand how the changes in std.conv are related to this bug.

burner:
windows 32 is doing some really crazy stuff with floats, these changes just add some  sanity checks.

burner:
and std.conv.to!string( some_float ) uses std.format

std/format.d
```diff
@@ -2669,7 +2676,7 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
         "floating point formatting failure");
 
     size_t len = min(n, buf.length-1);
-    if (fs.flSeparator)
+    if (fs.flSeparator && !isNaN(val))
```
berni44: Why this check? I think nan is allready ruled out above, isn't it?

burner: true, I'll fix that

burner: fixed

std/conv.d
```diff
@@ -1857,9 +1857,12 @@ private void testFloatingToIntegral(Floating, Integral)()
...
             T a = 42;
-            assert(to!string(a) == "42");
-            assert(to!wstring(a) == "42"w);
-            assert(to!dstring(a) == "42"d);
+            string s = to!string(a);
+            assert(s == "42", s);
+            wstring ws = to!wstring(a);
+            assert(ws == "42"w, to!string(ws));
+            dstring ds = to!dstring(a);
+            assert(ds == "42"d, to!string(ds));
```
ZombineDev: Is this change necessary?

burner: I would say yes, while working on std.format this broke and I couldn't see the assert inputs. So this will help the next one person how gets gray hair looking at std.format!double.

ZombineDev: Thanks for clarifying, I actually missed that you added the inputs as second argument to the asserts.

burner: no worries, thanks for taking an interest.

MoonlightSentinel: Maybe checkaction=context should be enabled for DRuntime (and other repos) unittest builds

burner: yes, please

dejlek:
We have discussed yesterday performance issue with float --> string conversion. The default implementation in libC is painfully slow. PyPy is 2x faster (!!) because they have their own implementation. Something like this in Phobos would be welcome: https://www.reddit.com/r/ProgrammingLanguages/comments/930pch/ryu_a_new_algorithm_to_quickly_convert_floating/

marler8997:
I ported the ryu algorithm to D in about an hour, source here: https://github.com/dragon-lang/mar/blob/master/src/mar/ryu.d

berni44:
Meanwhile I read through the paper on ryu. I may be wrong, but I think this algorithm is not useful for this task, for two reasons:

a) The goal is slightly different to that of printf. While ryu tries to print as few digits as possible, printf prints as much as the format string defines. If that's larger than what ryu computes, the remaining digits will be wrong.
b) Ryu uses a lookuptable which is IMHO far too large for a standard library. For 128bit reals it's about 300K of data, that needs to be stored.

berni44:
@UplinkCoder I'll have a look at the grisu implementations to see if it can be used for this task. But due to my time schedule this will probably have to wait one or two weeks.

moon-chilled:
@marler8997 I'm working on a more robust (than the original) translation that has everything you need to replace the current implementation `format()` uses.  Already have support for only printing out some digits; main task now is not to print out numbers with small exponents as scientific notation (that is, `123.45` instead of `1.2345e+2`).  Then trailing/leading zeroes should be easy enough.

Benchmarks suggest a not quite 2x bump in speed; maybe 1.8x.

marler8997:
@moon-chilled great.  You'll have to share it.  I only ported the 32-bit version of ryu.  For the 64-bit version I wasn't sure if it should be completely separate like the original did, or if I should try to combine them and re-use common code.  I look forward to seeing your port.

moon-chilled:
[Here](http://sprunge.us/2FasL1?d).  (Lookup tables end at line 380--I really need to ctfe those, because they kill compile times.)

I was planning on mostly re-using code, once I figured out a good architecture for the 64-bit version.

Oh, also - I forgot; it'll probably also get a decent speedboost once I get the 128-bit multiplies in hardware.

burner:
https://issues.dlang.org/show_bug.cgi?id=19124

MoonlightSentinel:
This caused a regression: https://issues.dlang.org/show_bug.cgi?id=20623

PR #7223, opened 2019-10-10
Fix issue 20027 - std.zip susceptible to zip malware attacks

berni44:
This fixes the main problem of 20027, namely rejecting zip bombs. I moved all remaining problems to own issues (20290 - 20294). Therefore this issue can be closed with this PR.

There is one unittest missing, namely checking for overlaps of zip64 central dir records with other stuff. This unittests needs to be postponed due to a bug in the implementation of zip64 (see issue 20289).

I verified, that it indeed rejects the zip bombs zbsm.zip and zblg.zip mentioned in the issues. The zip64 zip bomb zbxl.zip does not work yet due to bugs in the zip64 implementation, probably introduced by me in PR #7198. I'll fix this as soon as possible.

I also checked some normal zip-files to make sure that the size of _segs is always O(1).

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -599,26 +599,34 @@ public:
 
     this(void[] buffer)
     {
-        uint directorySize;
-        uint directoryOffset;
-
         this._data = cast(ubyte[]) buffer;
 
         if (data.length > uint.max - 2)
             throw new ZipException("zip files bigger than 4 GB are unsupported");
 
+        _segs = [Segment(0, cast(uint) data.length)];
+
         findEndOfCentralDirRecord();
         uint i = endrecOffset;
 
-        int endcommentlength = getUshort(i + 20);
-        comment = cast(string)(_data[i + 22 .. i + 22 + endcommentlength]);
+        int endCommentLength = getUshort(i + 20);
+        comment = cast(string)(_data[i + 22 .. i + 22 + endCommentLength]);
+
+        // end of central dir record
+        removeSegment(endrecOffset, endrecOffset + endOfCentralDirLength + endCommentLength);
 
         uint k = i - zip64EndOfCentralDirLocatorLength;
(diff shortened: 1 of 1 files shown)


std/zip.d
```diff
@@ -599,27 +599,35 @@ public:
...
 
+        _segs = [Segment(0, cast(uint) data.length)];
+
         findEndOfCentralDirRecord();
         uint i = endrecOffset;
 
-        int endcommentlength = getUshort(i + 20);
-        comment = cast(string)(_data[i + 22 .. i + 22 + endcommentlength]);
+        int endCommentLength = getUshort(i + 20);
+        comment = cast(string)(_data[i + 22 .. i + 22 + endCommentLength]);
```
CyberShadow: It would have been good to put style fixes in their own commit.

std/zip.d
```diff
@@ -726,6 +741,10 @@ public:
             auto localFileHeaderNamelen = getUshort(de.offset + 26);
             auto localFileHeaderExtralen = getUshort(de.offset + 28);
 
+            // file data
+            removeSegment(de.offset, de.offset + localFileHeaderLength + localFileHeaderNamelen
+                                     + localFileHeaderExtralen + de._compressedSize);
```
CyberShadow: I'm hoping all of these can be later refactored away to avoid the DRY of calculating all these lengths all over the place. The code is pretty difficult to follow as it is. Maybe once the structures are represented as D structs, the `removeSegment` call can be done by whatever function will be doing the reading of the entire structs.

CyberShadow: Here is my code to read a Windows PE file's headers:

https://github.com/CyberShadow/ae/blob/218c29e17f6ecfc49f696b92c879c5c6b9917f2e/sys/windows/pe/pe.d#L31-L51

I'm hoping std.zip could closer to that.

PR #7224, opened 2019-10-10
Fix issue #20295 - std.zip: zip64 with 0xff bytes in end of central dir record do not work

berni44:
While working on this fix I realised, that I missed two magic constant signatures when I merged two older PRs. I adapted them too. Hope, this is ok.


berni44:
This is a regression, where the original PR has not made it to stable yet. Therefore I target master, not stable.

std/zip.d
```diff
@@ -760,16 +760,22 @@ public:
 
             if (numberOfThisDisc < numberOfStartOfCentralDirectory) continue;
 
+            uint k = start - zip64EndOfCentralDirLocatorLength;
+            auto maybeZip64 = k < start && _data[k .. k + 4] == zip64EndOfCentralDirLocatorSignature;
```
CyberShadow: If `zip64EndOfCentralDirLocatorLength` is a non-zero constant, and here `k == start - zip64EndOfCentralDirLocatorLength`, when would `k < start` be false? If it underflows?

berni44: It's basically a copy of line 619. And yes, I think it's to detect underflows.

berni44:
> It would be really helpful and aid in code review if you could provide links to or cite the relevant portions of the specifications that these changes are in accordance with.

Sorry, I was not aware of that. I'm using https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT

You can find the end of central dir record in 4.3.16. What I overlooked was 4.4.1.4.

berni44:
> The code is becoming a bit difficult to follow, so I'm hoping a future refactoring can reduce the amount of arithmetics and comparisons done at every step.

I hope so too. :-)

By the way. As far as I know, today new stable will be created. I'm not sure, but I think, things will be easier, when this PR is merged before this step, isn't it?

PR #7226, opened 2019-10-14
Run Azure Pipelines

wilzbach:
See also: https://github.com/dlang/druntime/pull/2806

FYI: @rainers

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,91 @@
+# Learn more: https://aka.ms/yaml
+
+variables:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,91 @@
...
+    steps:
+      - script: set
+        displayName: environment
+      - script: systeminfo
+        displayName: systeminfo
+      - task: CopyFiles@2
+        inputs:
+          sourceFolder: '$(Build.SourcesDirectory)'
+          contents: '**/*'
+          TargetFolder: '$(Build.SourcesDirectory)/../druntime'
```
rainers: You should not copy the phobos source to ../druntime ;-) Same below.

wilzbach: Ah that was stupid. Thanks!

PR #7228, opened 2019-10-14
std/math.d: Explicitly handle zero and inf in scalbn.

ibuclaw:
Fixes the CTFE test for gdc when x is real.infinity.  It fails as an inf result from ldexp is treated as an overflow in gcc's const fold internals.


wilzbach:
> Fixes the CTFE test for gdc when x is real.infinity. It fails as an inf result from ldexp is treated as an overflow in gcc's const fold internals.

Should we version it with GDC or `__ctfe`?

ibuclaw:
@wilzbach can do `__ctfe`, as `isInfinity` is a built-in anyway...

PR #7229, opened 2019-10-16
Fix issue #20295 - std.zip: zip64 with 0xff bytes in end of central dir record do not work

berni44:
This is PR #7224, but rebased on stable (I probably got something wrong while rebasing, therefore a new PR, sorry), as it's now a real regression...

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -779,16 +779,22 @@ public:
 
             if (numberOfThisDisc < numberOfStartOfCentralDirectory) continue;
 
+            uint k = start - zip64EndOfCentralDirLocatorLength;
+            auto maybeZip64 = k < start && _data[k .. k + 4] == cast(ubyte[])"PK\x06\x07";
+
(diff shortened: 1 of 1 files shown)


berni44:
I think, the failed checks where unrelated problems. I removed all refactoring stuff, which isn't that urgent, though.

Currently findEndOfCentralDirRecord() may fail in case of zip64 format (the new unittest contains such a case). Because in this format some variables may have the value 0xffff instead of the correct one, which might be larger and can therefore not stored using only two bytes.

The fix first finds out, if zip64 is possible at all, using essentially the same routine found in lines 618-619. (This routine can be refactored at both places, but I wanted not to change too much at once.) 

Later, three checks had to be extended to accept 0xffff if zip64 is possible.

In the [specs](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) you can find the relevant stuff in section 4.3.16. What I overlooked when I originally wrote findEndOfCentralDirRecord() was section 4.4.1.4.

PR #7231, opened 2019-10-16
Fix issues 20301 and 20302

ZombineDev:
This PR includes the following commits:

* Fix issue 20302 - Support head const in memoize
  Make `std.functional.memoize` accept functions whose return types or parameters are head const.
* ~[refactor]: Simplify memoize(alias fun) implementation~
  * ~Use `object.require` to save one hash table lookup~
  * ~[style] Consolidate imports~
* Fix Issue 20301 - Support const array of patterns in std.regex.regex

### DIFF
--- a/std/functional.d
+++ b/std/functional.d
@@ -1204,9 +1204,10 @@ template memoize(alias fun, uint maxSize)
  * To _memoize a recursive function, simply insert the memoized call in lieu of the plain recursive call.
  * For example, to transform the exponential-time Fibonacci implementation into a linear-time computation:
  */
-@safe unittest
+@safe nothrow
+unittest
 {
-    ulong fib(ulong n) @safe
+    ulong fib(ulong n) @safe nothrow
     {
         return n < 2 ? n : memoize!fib(n - 2) + memoize!fib(n - 1);
     }
--- a/std/functional.d
+++ b/std/functional.d
@@ -1129,8 +1129,9 @@ template memoize(alias fun)
     {
         alias Args = Parameters!fun;
         import std.typecons : Tuple;
+        import std.traits : Unqual;
 
-        static ReturnType!fun[Tuple!Args] memo;
+        static Unqual!(ReturnType!fun)[Tuple!Args] memo;
         auto t = Tuple!Args(args);
         if (auto p = t in memo)
             return *p;
@@ -1145,9 +1146,10 @@ template memoize(alias fun, uint maxSize)
     // alias Args = Parameters!fun; // Bugzilla 13580
     ReturnType!fun memoize(Parameters!fun args)
     {
-        import std.traits : hasIndirections;
+        import std.meta : staticMap;
+        import std.traits : hasIndirections, Unqual;
         import std.typecons : tuple;
-        static struct Value { Parameters!fun args; ReturnType!fun res; }
+        static struct Value { staticMap!(Unqual, Parameters!fun) args; Unqual!(ReturnType!fun) res; }
         static Value[] memo;
         static size_t[] initialized;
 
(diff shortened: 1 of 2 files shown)


ZombineDev:
I'll drop commit https://github.com/dlang/phobos/commit/df58d9028469bf287b371c46e6127895ea9c0b43 as it breaks users that assume `memoize` is `nothrow` friendly. The reason for the breakage is that require uses a `lazy` parameter, whose attributes can't be inferred (bug [#]()). I'll also push a commit that ensures `memoize!fib` produces a `nothrow` function.

std/regex/package.d
```diff
@@ -356,13 +356,13 @@ public alias StaticRegex = Regex;
...
 +/
-@trusted public auto regex(S)(S[] patterns, const(char)[] flags="")
+@trusted public auto regex(S : C[], C)(const S[] patterns, const(char)[] flags="")
 if (isSomeString!(S))
 {
     import std.array : appender;
     import std.functional : memoize;
     enum cacheSize = 8; //TODO: invent nice interface to control regex caching
-    S pat;
+    const(C)[] pat;
```
Geod24: If the only reason to introduce `C` is this, maybe `ElemType` would be better ?

std/functional.d
```diff
@@ -1376,6 +1379,33 @@ template memoize(alias fun, uint maxSize)
     assert(executed == 1);
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=20302
+@system unittest
+{
+    version (unsupported) // TODO
```
Geod24: `none` ?

std/functional.d
```diff
@@ -1129,8 +1129,9 @@ template memoize(alias fun)
     {
         alias Args = Parameters!fun;
         import std.typecons : Tuple;
+        import std.traits : Unqual;
 
-        static ReturnType!fun[Tuple!Args] memo;
+        static Unqual!(ReturnType!fun)[Tuple!Args] memo;
```
Geod24: The `Unqual` really shouldn't be necessary (in a perfect world), but I guess the compiler can't recognize that.

ZombineDev: Given `const(int) fun(int a, int b);`, `typeof(memoize!fun.memo)` becomes: `ReturnType!fun[Tuple!Args]` -> `const(int)!fun[Tuple!(int, int)]`. Later on, for the line `memo[t] = fun(args)` you get:
```
Error: cannot modify const expression memo[t]
```

Geod24: Yep, but since this is insert-only, it should work.
However, it would take the compiler to recognize insert-only path (e.g. by handling `require` ?) to make this work.

std/regex/package.d
ZombineDev: I don't think so (maybe I'm missing something?), because:
1. `regexImpl` supports only string types (not ranges).
2. `ElemType` (I assume you mean `std.range.primitives.ElementType`) will return `dchar`, given `T = string`, or `T = wstring`. In case we're called with such narrow strings, we would need to re-encode them to dchar[], which would be the type of `pat`.
3. `const`-ness. Changing the parameter from `S[] patterns` to `const S[] patterns` is not strictly necessary here (but it is for `regexImpl`), but by doing this I'm saving on distinct template instances. Before for `S = char[]`, `S = const(char)[]` and `S = const(char[])` (ditto for `immutable`) we would get different template instances. Now there's only one.
So, with this change in mind, when `regex(const S[] patterns, ..)` is called with a `char[][]`, `patterns` is deduced to be `const(char[][])`, while `S` becomes `char[]` and we get the following error: 
    ```
    Error: cannot implicitly convert expression patterns[0] of type const(char[]) to char[]
    ```
    (Because in the line `pat = patterns[0]`, `pat` is `S = char[]` and `patterns[0]` is `const(char[])`.)
     Additionally, I can't define `pat` as `const S` as then I won't be able to assign to it.
    So what I need is the char type, so I can define `pat` as tail-`const`.
4. Using the template specialization syntax `S : C[], C` is much shorter and readable then `ElementEncodingType` (from `std.range.primitives`). It's probably also slightly faster to compile.

std/functional.d
ZombineDev: It's effectively the same thing, but ok.

std/functional.d
ZombineDev: I know, I thought as well and I even tried using `require` and without `Unqual` change, but it has this [line](https://github.com/dlang/druntime/blob/master/src/object.d#L2600) inside, which would fail because `*p` is `const(int)`.

ZombineDev: That's said, I think it's a would be a good enhancement request.

ZombineDev: `Unqual + emplace` should work there, but I think that change is better left for the master branch, while this PR is targeting `stable`.

trikko:
What about https://issues.dlang.org/show_bug.cgi?id=20300 ?

ZombineDev:
@trikko I took a stab at implementing it and the good news is that it's definitely doable - see https://run.dlang.io/gist/ZombineDev/677cea7195209dec2e2b393bcf94ae24?compiler=dmd. However it wouldn't be trivial to turn my code into a proper Phobos pull request for the limited time I have today, so I'll leave it for a future PR.

trikko:
Too bad (:

ZombineDev:
It looks like @dlang-bot is offline...

PR #7232, opened 2019-10-16
Fix issue 19626 - RedBlackTree of an enum fails in unittest mode

berni44:


thewilsonator:
Auto-merge toggled on

PR #7233, opened 2019-10-16
Fix Issue 19283 - [std.mathspecial] documentation for normal distribution doesn't list parameters

berni44:
The two functions use *standard* normal distribution, which implies that mean=0.0, stdev=1.0.

In the bugreport is stated, that reportee was not able to follow the implementation for the case z>=1. Me neither. But as the case z<1 is obviously the standard normal distribution it would be an error, if the case z>=1 would be some other distribution. I compared some computed values to the [table given in the German Wikipedia](https://de.wikipedia.org/wiki/Standardnormalverteilungstabelle) and found no difference.

### DIFF
--- a/std/mathspecial.d
+++ b/std/mathspecial.d
@@ -327,7 +327,7 @@ real erfc(real x)
 }
 
 
-/** Normal distribution function.
+/** Standard normal distribution function.
  *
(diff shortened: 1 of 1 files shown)


thewilsonator:
Auto-merge toggled on

PR #7234, opened 2019-10-17
Fix Issue 20282 - CustomFloat.dig fails at some values.

berni44:
If Flags.storeNormalized is *not* set, we have to subtract one from the precision, because one bit is used for the leading 1.

Due to this bug it was never possible that shiftcnt==64 and a second bug in the next line went unnoticed. I fixed that one too.


thewilsonator:
Auto-merge toggled on

PR #7235, opened 2019-10-17
Fix Issue 20283 - CustomFloat.max_exp not working in some cases

berni44:
There were actually two bugs in here, that cancelled out in most cases.

First, flags.nan was used instead of Flags.nan.

Second, it was tried to mix the substraction of 1 in case of infinity or nan with the addition of 1 that's needed by definition, by using ~flags instead of flags, which does not always work. Although it's probably possible to find a term that mixes that, I separated these two terms. The speedgain would be little and done at compile time anyway.


thewilsonator:
Auto-merge toggled on

PR #7236, opened 2019-10-17
Fix Issue 20263 - Wrong value for CustomFloat.min_exp

berni44:
Just a cast was missing to allow for negative values...


thewilsonator:
Auto-merge toggled on

PR #7238, opened 2019-10-17
Fix Issue 20261 - CustomFloat.epsilon yields infinity

berni44:
I completely rewrote the code, because I think it's difficult to give the bit patterns of epsilon directly (*). There are so many exceptions and quirks around, where you allways have to consider all the flags involved. There are cases, where epsilon is a denormalized number, there are cases, where epsilon does not exist in that type and so on.

My solution here is, not to care about this at all. What I do is to take the value 1, calculate 1+epslion from it by setting the least significant bit of the mantissa to 1 and subtract the value 1 from it. (This is not CTFE though.)

There are obviously cases, where this doesn't work. For example when 1 cannot be represented by that type or 1+epsilon=infinity, but infinity does not exist or where epsilon cannot be represented. As far as I can see, these types are all nonsensical ones. I leave this to PR #7219, which removes most of these cases, if not all.

(*) To give an impresson of the bit patterns,. here's a list of some of them:
- CustomFloat(1,6): `011110 | 0`
- CustomFloat(2,5): `01101 | 00`
- CustomFloat(3,4): `0100 | 000`
- CustomFloat(4,3): `000 | 0100` (denormalised number, if allowDenorm is set, else `000 | 0001`, not 100% sure on this last one)
- CustomFloat(5,2): `00 | 00001` (denormalised number, if allowDenorm is set, else does not exist)
- CustomFloat(6,1): Does not exist

### DIFF
--- a/std/numeric.d
+++ b/std/numeric.d
@@ -410,24 +410,11 @@ public:
     /// Returns: smallest increment to the value 1
     static @property CustomFloat epsilon()
     {
-        CustomFloat value;
-        static if (flags & Flags.signed)
-            value.sign       = 0;
-        T_signed_exp exp = -precision;
-        T_sig        sig = 0;
+        CustomFloat one = CustomFloat(1);
+        CustomFloat onePlusEpsilon = one;
+        onePlusEpsilon.significand = onePlusEpsilon.significand | 1; // |= does not work here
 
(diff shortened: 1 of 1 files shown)


thewilsonator:
Auto-merge toggled on

PR #7240, opened 2019-10-21
Fix Issue 10448 - min and max are not NaN aware

berni44:
This is a replacement for PR #7181, because I had to delete my old account on github. Here my original comment:

The discussion of issue 10448 and of the related pull requests show, that there is no consensus on the best result of min and max, when some parameters are NaNs, even experts, namely people who where involved in writing IEEE754, cannot agree on that. But most people agree, that the functions should be fast and extra treatment for NaNs would slow them down. Therefore, I think, the best solution is to document, that the functions will result in undefined behaviour if NaNs are used and give examples on how to cope with NaNs beyond using this functions.

In the documentation for minElement I added two such examples, namely for ignoring NaNs and for returning NaN in case at least one element is NaN. In the documentation for maxElement, min and max I just added a pointer to minElement, to avoid duplication.

I'm aware, that there are more functions, which have this problem. But before doing all the work to identify them and to add a note to everyones documentation, I'd make like to make sure, that this will be accepted. Therefore I'll wait for this PR to be merged, before continuing.

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -1513,17 +1513,21 @@ if (isConvertibleToString!Range1 || isConvertibleToString!Range2)
     assert(levenshteinDistanceAndPath(S("cat"), "rat")[0] == 1);
 }
 
+
 // max
 /**
-Iterates the passed arguments and return the maximum value.
+Iterates the passed arguments and returns the maximum value.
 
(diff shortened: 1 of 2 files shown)


thewilsonator:
Auto-merge toggled on

PR #7241, opened 2019-10-21
Add isClose, a replacement for approxEqual

berni44:
This is a replacement for PR #7187, because I had to remove my old account on GitHub.

My original comment was:

As suggested in PR #7173, I wrote a new version of approxEqual (called isClose to avoid naming conflicts), which is based on the thoughts given in [1]. The purpose of both functions is the same, namely providing a utility for comparing floating point numbers. This is necessary, because operations on floating point numbers inevitably produce rounding errors and therefore cannot be compared reasonable using ==.

The main difference is, that this function has much stronger default values for the error allowed, based on the capabilities of the used floating point type. This helps in catching errors, especially as approxEqual is mainly used in unittests. This addresses issue #15881. Together with PR #7180 this can be considered a fix for that issue.

The new function is furthermore symmetric, to avoid users being confused or have to remember the order of the operands. This addresses issue #15763.

After this PR has been approved I think, it would be wise to continue with the following roadmap:

1. replace all occurences of approxEqual in Phobos by isClose. With one exception, approxEqual is only used in unittests. The exception is secantMethod in std/numerical.d.
2. deprecate approxEqual in favour of isClose. (I don't know if I'm in the position to deprecate anything, nor to ask for it, but I think, this would be a good idea in the long-run.)
3. After some time, remove approxEqual alltogether.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -40,7 +40,7 @@ $(TR $(TDNW Modulus) $(TD
 ))
 $(TR $(TDNW Floating-point operations) $(TD
     $(MYREF approxEqual) $(MYREF feqrel) $(MYREF fdim) $(MYREF fmax)
-    $(MYREF fmin) $(MYREF fma) $(MYREF nextDown) $(MYREF nextUp)
+    $(MYREF fmin) $(MYREF fma) $(MYREF isClose) $(MYREF nextDown) $(MYREF nextUp)
     $(MYREF nextafter) $(MYREF NaN) $(MYREF getNaNPayload)
     $(MYREF cmp)
 ))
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -9219,6 +9219,254 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
...
+       the result is the logical and-ing of calling `isClose` on
+       each element of the ranged item against the single item. If
+       both items are ranges, then `isClose` returns `true` if
+       and only if the ranges have the same number of elements and if
+       `isClose` evaluates to `true` for each pair of elements.
+
+    See_Also:
+        Use $(LREF feqrel) to get the number of equal bits in the mantissa.
+ */
+bool isClose(T, U)(T lhs, U rhs, real maxRelDiff = CommonDefaultFor!(T,U), real maxAbsDiff = 0.0)
```
thewilsonator: Please make the use of `real` optional by making `maxRelDiff` and `maxAbsDiff` a template type parameter.

std/math.d
```diff
@@ -9219,6 +9219,254 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
...
+
+    assert(isClose(f,f2d));
+    assert(!isClose(d,f2d));
+    assert(isClose(d,f2d,1e-4));
+}
+
+private template CommonDefaultFor(T,U)
+{
+    import std.algorithm.comparison : min;
+    enum CommonDefaultFor = 10.0L ^^ -((min(FloatingPointBaseType!T.dig, FloatingPointBaseType!U.dig) + 1) / 2 + 1);
```
thewilsonator: `typeof(CommonType!(T,U)(10.0)` or `enum CommonType!(T,U) CommonDefaultFor =`

berni44:
Don't worry. It's not that important to me (else I would have pinged) - I thought, it might be a problem, because a new name is introduced. 

Patches on the way.

thewilsonator:
Thanks.

std/math.d
```diff
@@ -9210,6 +9210,258 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
...
+    assert(isClose(num, -real.infinity));
+
+    assert(!isClose(1,real.nan));
+    assert(!isClose(real.nan,real.max));
+    assert(!isClose(real.nan,real.nan));
+}
+
+@safe pure nothrow @nogc unittest
+{
+//    assert(isClose([],[])); //FIXME: does not work yet
```
thewilsonator: You should be able to do 
```
assert(isClose!(real, real,real)([],[]));
```

std/math.d
```diff
@@ -9210,6 +9210,258 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
...
+       the result is the logical and-ing of calling `isClose` on
+       each element of the ranged item against the single item. If
+       both items are ranges, then `isClose` returns `true` if
+       and only if the ranges have the same number of elements and if
+       `isClose` evaluates to `true` for each pair of elements.
+
+    See_Also:
+        Use $(LREF feqrel) to get the number of equal bits in the mantissa.
+ */
+bool isClose(T, U, V)(T lhs, U rhs, V maxRelDiff = CommonDefaultFor!(T,U), V maxAbsDiff = 0.0)
```
thewilsonator: Also `bool isClose(T, U, V = CommonType!(T,U))`

quickfur:
This is a pretty significant change, and IMO notable enough to be included in the release notes.  Can you submit a follow-up PR to add the corresponding file in `changelog/` describing this change? (We're still using that for generating release notes, right?)

berni44:
I'm currently working on replacing all occurences of approxEqual in Phobos with corresponding calls to isClose; see #7340. Essentially I'm through, but from my first try I remember, that some tests in std.complex failed on 32bit computers, because the calculation of log() is less accurate. I have to experiment a little with the autotesters to get that right but maybe I'll manage to still finish that this year. :-)

After that PR has been accepted, I plan to file another, which deprecates approxEqual and adds a changelog-entry (I actually planed to ask if I should do so, but thanks to your post now I know I should :-) ). OK?

std/math.d
```diff
@@ -9210,6 +9210,259 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
...
+       each element of the ranged item against the single item. If
+       both items are ranges, then `isClose` returns `true` if
+       and only if the ranges have the same number of elements and if
+       `isClose` evaluates to `true` for each pair of elements.
+
+    See_Also:
+        Use $(LREF feqrel) to get the number of equal bits in the mantissa.
+ */
+bool isClose(T, U, V = CommonType!(FloatingPointBaseType!T,FloatingPointBaseType!U))
+    (T lhs, U rhs, V maxRelDiff = CommonDefaultFor!(T,U), V maxAbsDiff = 0.0)
```
jondegenhardt, line 9243: `CommonDefaultFor` needs to be public and documented if it used as the default value in a public API. Currently it has visibility in the math package, not public visibility. 

In this case, not having it exposed means that callers cannot provide the default for the `maxRelDiff` parameter if they only want to change the `maxAbsDiff` parameter.

PR #7242, opened 2019-10-21
Remove multidisk support completely from zip.std.

berni44:
I removed the remaining artefacts of multidisk support, with the final goal in mind to make this file conformant to [ISO/IEC 21320-1:2015](https://standards.iso.org/ittf/PubliclyAvailableStandards/c060101_ISO_IEC_21320-1_2015.zip). This standard defines a subset of the original zip file which sounds very reasonable (and is quite similar to that, what's allready supported).

(Multidisk has been usefull in the 80ies, when zip files had to be stored on 640K-disks. Nowadays, there is no sensible usecase anymore.)

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -116,7 +116,6 @@ final class ArchiveMember
     private CompressionMethod _compressionMethod;
     private ushort _madeVersion = 20;
     private ushort _extractVersion = 20;
-    private ushort _diskNumber;
     private uint _externalAttributes;
     private DosFileTime _time;
     // by default, no explicit order goes after explicit order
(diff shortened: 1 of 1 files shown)


berni44:
> When deprecating a functionality, it should still _work_. Now I know it's unlikely someone in the wild uses it, but we can't just break the functionality while deprecating it.

I'm sorry, I should have written that in my comment above: Although it might look like, there is no change in functionality. These two variables where always 0: _diskNumber (in both places) and _diskStartDir are private members that cannot be changed from the outside. The default value of ushort is 0 and zip archives that do not contain a 0 at that places are rejected in findEndOfCentralDirRecord.

berni44:
ping

PR #7243, opened 2019-10-22
More min and max are not NaN aware.

berni44:
This is a followup on PR #7240. Fixing documentation at some more places, where NaNs do not work as expected.


thewilsonator:
Auto-merge toggled on

PR #7244, opened 2019-10-23
Make tests for issue 19797 Windows-only

Geod24:
```
Value '3' is a valid value on Linux (SEEK_DATA, since 3.18).
Hence, in some cases (depending on FS support / Kernel version),
this test can fails.
```

If you're curious, more info [here](https://github.com/alpinelinux/aports/pull/11931#issuecomment-544831142).

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -1098,6 +1098,14 @@ Throws: `Exception` if the file is not opened.
         import std.conv : to, text;
         import std.exception : enforce, errnoEnforce;
 
+        // Some libc sanitize the whence input (e.g. glibc), but some don't,
+        // e.g. Microsoft runtime crashes on an invalid origin,
+        // and Musl additionally accept SEEK_DATA & SEEK_HOLE (Linux extension).
+        // To provide a consistent behavior cross platform, we use the glibc check
+        // See also https://issues.dlang.org/show_bug.cgi?id=19797
+        enforce(origin == SEEK_SET || origin == SEEK_CUR ||  origin == SEEK_END,
+                "Invalid `origin` argument passed to `seek`, must be one of: SEEK_SET, SEEK_CUR, SEEK_END");
+
         enforce(isOpen, "Attempting to seek() in an unopened file");
         version (Windows)
         {
@@ -1105,9 +1113,6 @@ Throws: `Exception` if the file is not opened.
             {
                 alias fseekFun = _fseeki64;
                 alias off_t = long;
-                // Issue 19797
-                enforce(origin >= SEEK_SET && origin <= SEEK_END,
-                        "Could not seek in file `"~_name~"' (Invalid argument)");
             }
             else
             {
@@ -1151,7 +1156,8 @@ Throws: `Exception` if the file is not opened.
         // f.rawWrite("abcdefghijklmnopqrstuvwxyz");
         // f.seek(-3, SEEK_END);
         // assert(f.readln() == "xyz");
-        assertThrown(f.seek(0, 3));
+
+        assertThrown(f.seek(0, ushort.max));
     }
 
 /**


CyberShadow:
Why not `int.max` as per your earlier comment? I think that would make fewer assumptions about the implementation of `fseek`.

CyberShadow:
In `man 3 fseek` I see:

```
ERRORS
       EINVAL The whence argument to fseek() was not SEEK_SET, SEEK_END, or SEEK_CUR.  Or: the resulting file offset would be negative.
```

Though [POSIX doesn't explicitly state this](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fseek.html), one way to interpret this would be that the `fseek` implementation you're using is non-confromant in that it passes the `whence` argument to the kernel without checking it.

Perhaps a better change would be to wrap the check in `version (CRuntime_Musl) {} else {...}`.

Geod24:
> Why not `int.max` as per your earlier comment?

`int.max` is too special of a value, so `short.max` would have been a better recommendation.
I guess that depends on how you view the bug: is it about the lack of throw when an invalid value is passed (my original comment), or what the value are. I'm happy with either, and if you think a higher value is better, that's what I'll go with.

Geod24:
> Though POSIX doesn't explicitly state this, one way to interpret this would be that the fseek implementation you're using is non-confromant in that it passes the whence argument to the kernel without checking it.


Good catch! In my case, `fseek` [forwards to `lseek`](https://github.com/bminor/musl/blob/84d061d5a31c9c773e29e1e2b1ffe8cb9557bc58/src/stdio/fseek.c#L3-L25), which has a slightly different [man](http://man7.org/linux/man-pages/man2/lseek.2.html):

> EINVAL whence is not valid.  Or: the resulting file offset would be negative, or beyond the end of a seekable device.

And `lseek` mention `SEEK_DATA` and `SEEK_HOLE`. To me, it looks like an oversight in the man pages.

CyberShadow:
> To me, it looks like an oversight in the man pages.

Well, an oversight in which way?

`SEEK_DATA` / `SEEK_HOLE` are very apparently features of the Linux kernel. That `SEEK_SET`/`SEEK_CUR`/`SEEK_END` have the same values in the C interface and the Linux kernel seems more like a "happy coincidence" that musl is taking advantage of, and it looks like the code simply did not anticipate that the kernel was going to add new values to the accepted ones for the parameter.

I'm not sure how tied musl is to the Linux kernel. If, hypothetically, a FreeBSD port of musl existed, and the same implementation of `fseek` was used there, then should the FreeBSD kernel similarly add new valid values for the `lseek` (or equivalent) system call for this parameter, which are incompatible with the Linux ones, then that would make C (and D) programs using musl to behave incorrectly.

Would you be interested in reporting this to the musl developers? If they rule this behavior as a bug, then we should disable the test for musl only; if they rule it as a feature (implicit compatibility with however the Linux kernel decides to extend C's `whence` parameter semantics) and within POSIX, then it would make sense to restrict the test to only Windows or use a more obviously invalid value for the test.

Geod24:
Alright, I went through glibc sources and it [indeed checks it explicitly](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/libio/ioseekoff.c#L32-L38). Will report upstream.

Geod24:
@CyberShadow : https://www.openwall.com/lists/musl/2019/10/23/2

CyberShadow:
> @CyberShadow : https://www.openwall.com/lists/musl/2019/10/23/2

Honestly I find that disappointing, as that implies that 1) our issue 19797 is invalid and should have been left unfixed, and 2) it says "there is no meaningful, non-erroneous way to plug arbitrary integers into it", but that is exactly what musl is doing - it is plugging arbitrary unsanitized integers received from the application into the kernel.

In this case, I suggest extending the Windows check to apply to all platforms (throw an exception when `whence` is not one of the standard three values), and leaving the test as-is.

std/stdio.d
```diff
@@ -1098,16 +1098,21 @@ Throws: `Exception` if the file is not opened.
         import std.conv : to, text;
         import std.exception : enforce, errnoEnforce;
 
+        // Some libc sanitize the whence input (e.g. glibc), but some don't,
+        // e.g. Microsoft runtime crashes on an invalid origin,
+        // and Musl additionally accept SEEK_DATA & SEEK_HOLE (Linux extension).
+        // To provide a consistent behavior cross platform, we use the glibc check
+        // See also https://issues.dlang.org/show_bug.cgi?id=19797
+        enforce(origin == SEEK_SET || origin == SEEK_CUR ||  origin == SEEK_END,
```
ZombineDev: Shouldn't this be an `assert` instead, as `origin` is not coming from user input, but from the developer that called this function?

Geod24:
> it is plugging arbitrary unsanitized integers received from the application into the kernel.

The responsibility for validation lies with the kernel (and the syscall [lseek](http://man7.org/linux/man-pages/man2/lseek.2.html)), I don't see why Musl would need to validate those arguments if it doesn't interact with them. But I would like this solved one way or another, and loosing the ability to use `SEEK_DATA` and `SEEK_HOLE` with that function isn't a big deal.

> In this case, I suggest extending the Windows check to apply to all platforms (throw an exception when `whence` is not one of the standard three values), and leaving the test as-is.

Applied that suggestion

std/stdio.d
Geod24: Well the case that sparked this was a programmer's error, so I think it's fair. @CyberShadow do you agree to change `enforce` to `assert` here ?

CyberShadow:
> The responsibility for validation lies with the kernel (and the syscall [lseek](http://man7.org/linux/man-pages/man2/lseek.2.html)), I don't see why Musl would need to validate those arguments if it doesn't interact with them.

No, because the kernel implements the Linux kernel API, but Musl implements the C (POSIX) API. That the two are similar is carelessly exploited by Musl.

> and loosing the ability to use `SEEK_DATA` and `SEEK_HOLE` with that function isn't a big deal.

They "worked" only with Musl, but I understand that a Musl port is not likely to exist due to the Phobos tests not passing, so this is theoretical. In any case, that is certainly functionality better reserved for OS file descriptors and not C buffered `FILE*` objects or D `std.stdio.File` instances.

std/stdio.d
CyberShadow: We can only use `assert` in very particular circumstances in Phobos (namely template code which is *not* instantiated within Phobos) due to practical considerations (the compiler favoring the existing version in phobos.lib, which is not affected by the presence of `-release` on the user's command line), but otherwise I vacuously agree.

thewilsonator:
Auto-merge toggled on

PR #7245, opened 2019-10-23
Replace `in` with `const` in preparation of dmd change

atilaneves:
Changing dmd to implement `in` meaning `scope const` when DIP1000 is
selected causes Phobos to fail to compile. This changes Phobos in
preparation for that change.


burner:
thank you

PR #7247, opened 2019-10-23
File as range

vporton:
The methods `asUbytesRange()` and `asCharsRange()` of `File` object added.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2808,7 +2808,7 @@ is empty, throws an `Exception`. In case of an I/O error throws
         assert(i == witness.length);
     }
 
-    auto toUbytesRange(size_t bufferSize = 4096)
+    auto asUbytesRange(size_t bufferSize = 4096)
     {
         assert(bufferSize > 0);
 
@@ -2816,10 +2816,10 @@ is empty, throws an `Exception`. In case of an I/O error throws
         return byChunk(bufferSize).joiner;
     }
 
-    auto toCharsRange(size_t bufferSize = 4096)
+    auto asCharsRange(size_t bufferSize = 4096)
     {
         import std.algorithm;
-        return map!(b => cast(char) b)(toUbytesRange(bufferSize)); // TODO: efficient?
+        return map!(b => cast(char) b)(asUbytesRange(bufferSize)); // TODO: efficient?
     }
 
     @system unittest
@@ -2841,9 +2841,9 @@ is empty, throws an `Exception`. In case of an I/O error throws
             std.file.remove(deleteme);
         }
 
-        assert(equal(f.toCharsRange(), testString));
+        assert(equal(f.asCharsRange(), testString));
         f.seek(0);
-        assert(equal(f.toCharsRange(2), testString));
+        assert(equal(f.asCharsRange(2), testString));
     }
 
     // Note: This was documented until 2013/08
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2831,9 +2831,9 @@ Example:
 void main()
 {
     // Read standard input
-    foreach (char buffer; stdin.asCharsRange)
+    foreach (char c; stdin.asCharsRange)
     {
-        ... use buffer ...
+        ... use c ...
     }
 }
 ---------
@@ -2846,9 +2846,9 @@ Example:
 void main()
 {
     // Read standard input
-    foreach (char buffer; stdin.asCharsRange(4096))
+    foreach (char c; stdin.asCharsRange(4096))
     {
-        ... use buffer ...
+        ... use c ...
     }
 }
 ---------
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2856,8 +2856,7 @@ void main()
 With the mentioned limitations, `asCharsRange` works with any algorithm
 compatible with input ranges.
 
-Returns: A call to `asCharsRange` returns a range initialized with the `File`
-object.
+Returns: A range initialized with the `File` object.
 
 Asserts: Buffer size is non-zero. In case of an I/O error throws
 `StdioException`.
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2821,7 +2821,7 @@ Like `asCharsRange` but returns a range o `ubyte`s, not `char`s.
 
 /**
 Returns an $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
-set up to read from the file handle a character at a time.
+set up to read from the file handle.
 
 The element type for the range will be `char`. Range primitives
 may throw `StdioException` on I/O error.
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2808,6 +2808,9 @@ is empty, throws an `Exception`. In case of an I/O error throws
         assert(i == witness.length);
     }
 
+/**
+Like `asCharRange` but returns a range o `ubyte`s, not `char`s.
+*/
     auto asUbytesRange(size_t bufferSize = 4096)
     {
         assert(bufferSize > 0);
@@ -2816,6 +2819,49 @@ is empty, throws an `Exception`. In case of an I/O error throws
         return byChunk(bufferSize).joiner;
     }
 
+/**
+Returns an $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
+set up to read from the file handle a character at a time.
+
+The element type for the range will be `char`. Range primitives
+may throw `StdioException` on I/O error.
+
+Example:
+---------
+void main()
+{
+    // Read standard input
+    foreach (ubyte buffer; stdin.asCharsRange)
+    {
+        ... use buffer ...
+    }
+}
+---------
+
+The parameter may be a number (as shown in the example above) recommending the
+size of the buffer.
+
+Example:
+---------
+void main()
+{
+    // Read standard input
+    foreach (char buffer; stdin.asCharsRange(4096))
+    {
+        ... use buffer ...
+    }
+}
+---------
+
+With the mentioned limitations, `asCharsRange` works with any algorithm
+compatible with input ranges.
+
+Returns: A call to `asCharsRange` returns a range initialized with the `File`
+object.
+
+Asserts: Buffer size is non-zero. In case of an I/O error throws
+`StdioException`.
+ */
     auto asCharsRange(size_t bufferSize = 4096)
     {
         import std.algorithm;
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -2809,21 +2809,10 @@ is empty, throws an `Exception`. In case of an I/O error throws
     }
 
 /**
-Like `asCharsRange` but returns a range o `ubyte`s, not `char`s.
-*/
-    auto asUbytesRange(size_t bufferSize = 4096)
-    {
-        assert(bufferSize > 0);
-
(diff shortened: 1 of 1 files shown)


std/stdio.d
```diff
@@ -2808,6 +2808,90 @@ is empty, throws an `Exception`. In case of an I/O error throws
         assert(i == witness.length);
     }
 
+/**
+Like `asCharRange` but returns a range o `ubyte`s, not `char`s.
```
berni44: asChar*s*Range

std/stdio.d
```diff
@@ -2808,6 +2808,90 @@ is empty, throws an `Exception`. In case of an I/O error throws
...
+
+The element type for the range will be `char`. Range primitives
+may throw `StdioException` on I/O error.
+
+Example:
+---------
+void main()
+{
+    // Read standard input
+    foreach (ubyte buffer; stdin.asCharsRange)
```
berni44: "char buffer"?

std/stdio.d
```diff
@@ -2808,6 +2808,90 @@ is empty, throws an `Exception`. In case of an I/O error throws
...
+
+        import std.algorithm.iteration;
+        return byChunk(bufferSize).joiner;
+    }
+
+/**
+Returns an $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
+set up to read from the file handle a character at a time.
+
+The element type for the range will be `char`. Range primitives
```
CyberShadow: Ranges with elements of type `char` are a bit unusual due to autodecoding. Strings are decoded as ranges of `dchar`.

std/stdio.d
```diff
@@ -2808,6 +2808,90 @@ is empty, throws an `Exception`. In case of an I/O error throws
...
+{
+    // Read standard input
+    foreach (char buffer; stdin.asCharsRange)
+    {
+        ... use buffer ...
+    }
+}
+---------
+
+The parameter may be a number (as shown in the example above) recommending the
```
CyberShadow: "recommending" seems to be misleading as it implies that the value may not be followed, but I don't think there are any circumstances where the "recommendation" is ignored (other than there being less that the indicated number of bytes available for reading at that time).

std/stdio.d
```diff
@@ -2808,6 +2808,90 @@ is empty, throws an `Exception`. In case of an I/O error throws
...
+size of the buffer.
+
+Example:
+---------
+void main()
+{
+    // Read standard input
+    foreach (char buffer; stdin.asCharsRange(4096))
```
CyberShadow: Ditto

std/stdio.d
```diff
@@ -2808,6 +2808,90 @@ is empty, throws an `Exception`. In case of an I/O error throws
         assert(i == witness.length);
     }
 
+/**
+Like `asCharsRange` but returns a range o `ubyte`s, not `char`s.
+*/
+    auto asUbytesRange(size_t bufferSize = 4096)
+    {
+        assert(bufferSize > 0);
```
CyberShadow: There is already a check in `byChunk`:

https://github.com/dlang/phobos/blob/master/std/stdio.d#L2669

vporton:
> 1. As ruled previously, one-liners need a very strong justification and proof that they are a very common pattern in D code to pass the entry barrier.

Do you mean my implementation of `asCharsRange()`?

I don't know if it is efficient. I think it is likely not. Please provide a patch to make it efficient (or I may do it myself). However, what if D compiler is already enough clever? I ask the experts.

> 2. `asCharsRange` does nothing to validate the bytes as valid UTF, which a `char[]` is implied to be. This is going to be tricky as UTF sequences may span chunk boundaries.

Validating it UTF would be wrong, because the file may be in non-UTF encoding.

I recommend to use `.asCharsRange.byDchar` to convert from UTF. Representing file as character and simultaneously decoding it fro UTF-8 would be a rude violation of single responsibility.

CyberShadow:
> Validating it UTF would be wrong, because the file may be in non-UTF encoding.

Then you may not present it as a range of `char`s, which in D, are defined to be UTF-8 code units.

CyberShadow:
> Do you mean my implementation of `asCharsRange()`?

Both functions added here.

> I don't know if it is efficient.

The problem is not that they are not efficient; the problem is that they are trivial. Standard library additions should generally either implement non-trivial functionality or be common enough or encapsulate some pattern used frequently enough to warrant their addition. They should bring significant value over sometimes saving a few characters or at most one line of code in D programs.

std/stdio.d
vporton: Without ranges with elements of type `char`, it would be impossible to use functions such as `.byDchar` from `std.utf`. So, it is necessary.

If you want to change the API, first change the API of `std.utf` to use `ubyte` instead of `char`. My design is required by compatibility with `std.utf`.

std/stdio.d
vporton: For forward compatibility, we need to mark it as recommended, because a future implementation may for example reduce the buffer size if the available memory is low.

std/stdio.d
vporton: But should it be an exception or assert? Maybe different in `byChunk` and `asUbytesRange`? Not a very easy question.

std/stdio.d
CyberShadow: There is an ongoing effort to try to make autodecoding `nothrow`. The way this is going to work is that all `char` arrays / ranges are assumed to form valid UTF-8, so `char` arrays / ranges must be validated on creation. A standard library construct which produces `char` arrays / ranges without validation would thus break that contract.

Unvalidated characters, or characters in other encodings, are represented an `ubyte` arrays / ranges.

std/stdio.d
CyberShadow: Sorry, that makes no sense to me. You are overdesigning.

vporton:
> > Validating it UTF would be wrong, because the file may be in non-UTF encoding.
> 
> Then you may not present it as a range of `char`s, which in D, are defined to be UTF-8 code units.

I didn't know this.

Then it seems to impossible to do this easily (because `std.utf.validate` accepts only strings, not ranges) and efficiently (as we would validate in both `asCharsRange` and `std.utf.validate`). Need to change `std.utf` API to solve this problem!)

vporton:
> > Do you mean my implementation of `asCharsRange()`?
> 
> Both functions added here.
> 
> > I don't know if it is efficient.
> 
> The problem is not that they are not efficient; the problem is that they are trivial. Standard library additions should generally either implement non-trivial functionality or be common enough or encapsulate some pattern used frequently enough to warrant their addition. They should bring significant value over sometimes saving a few characters or at most one line of code in D programs.

It would be not a few characters, but a 6-7 or so of complex, error-prone code. So my one-liner is justified.

CyberShadow:
> Need to change `std.utf` API to solve this problem!

A range which lazily consumes `ubyte`s and produces validated `char`s would probably be a worthwhile addition to `std.utf`.

> It would be not a few characters, but a 6-7 or so of complex, error-prone code.

6-7 what?

Before:

```d
	writeln(File("test.d").byChunk(4096).joiner.sum);
```

After:

```d
	writeln(File("test.d").asUbytesRange.sum);
```

I'd say this definitely fits into the definition of "trivial".

> So my one-liner is justified.

Wording it as a fact doesn't change reality...

BTW, there is a bigger problem which I should have mentioned earlier. Because `byChunk` reuses the buffer, `joiner` results in an input range (i.e. no `save`, so slicing / indexing...), which imposes severe limits on what the range can be used for. Even something as basic as splitting the range by lines is not possible.

std/stdio.d
```diff
@@ -2808,6 +2808,89 @@ is empty, throws an `Exception`. In case of an I/O error throws
...
+
+Returns: A range initialized with the `File` object.
+
+Asserts: Buffer size is non-zero. In case of an I/O error throws
+`StdioException`.
+ */
+    auto asCharsRange(size_t bufferSize = 4096)
+    {
+        import std.algorithm;
+        return map!(b => cast(char) b)(asUbytesRange(bufferSize)); // TODO: efficient?
```
CyberShadow: What you probably want here is `byChunk(bufferSize).map!assumeUTF.joiner`, but the implicit non-validating conversion issue stands.

wilzbach:
> I'd say this definitely fits into the definition of "trivial".

Also, at this point new additions to Phobos must have been proven themselves outsides of Phobos.
There are a variety of projects like e.g. `phobos-next` that are such nursing grounds.

Hence, I strongly encourage _everyone_ to only pursue their dream libraries on dub. This is a lot easier than trying to push things into Phobos and everyone still profits from this code via Dub packages.

BTW the future of this I/O in D is here: https://code.dlang.org/packages/io

vporton:
> > Need to change `std.utf` API to solve this problem!
> 
> A range which lazily consumes `ubyte`s and produces validated `char`s would probably be a worthwhile addition to `std.utf`.
> 
> > It would be not a few characters, but a 6-7 or so of complex, error-prone code.
> 
> 6-7 what?
> 
> Before:
> 
> ```d
> 	writeln(File("test.d").byChunk(4096).joiner.sum);
> ```
> 
> After:
> 
> ```d
> 	writeln(File("test.d").asUbytesRange.sum);
> ```
> 
> I'd say this definitely fits into the definition of "trivial".

I misunderstood you: I assumed you meant that my _implementation_ should be rewritten without the one-liner.

`File("test.d").byChunk(4096).joiner` is a confusing, non-obvious code, that one may need to consult the documentation to understand. Choice of this 4096 bytes is not to be a burden on the user of the library, but on the library itself (it may be differently optimized for different OSes and architectures, etc.), what is impossible without my patch.

Moreover, my API gives additional flexibility to the implemented, like:
* reduce the buffer size if the memory is low
* reuse the same buffer (even if specified of a different size).

So in the future using my API may help to improve efficiency.

> BTW, there is a bigger problem which I should have mentioned earlier. Because `byChunk` reuses the buffer, `joiner` results in an input range (i.e. no `save`, so slicing / indexing...), which imposes severe limits on what the range can be used for. Even something as basic as splitting the range by lines is not possible.

Hm, it is not my error: https://dlang.org/library/std/stdio/file.by_chunk.html says "With the mentioned limitations, `byChunk` works with any algorithm compatible with input ranges." It implies that it should work with `joiner`. If it does not, it is a bug in the docs!

vporton:
> 2. `asCharsRange` does nothing to validate the bytes as valid UTF, which a `char[]` is implied to be. This is going to be tricky as UTF sequences may span chunk boundaries.

I've removed the questionable function `asCharsRange` (after merging my other PR that makes this function unnecessary).

std/stdio.d
```diff
@@ -2808,6 +2808,80 @@ is empty, throws an `Exception`. In case of an I/O error throws
...
+
+Asserts: Buffer size is non-zero. In case of an I/O error throws
+`StdioException`.
+ */
+    auto asUbytesRange(size_t bufferSize = 4096)
+    {
+        assert(bufferSize > 0);
+
+        import std.algorithm.iteration;
+        return byChunk(bufferSize).joiner;
```
jmdavis: We generally try not to add "shallow" functions to Phobos. This is just a simple wrapper. It doesn't provide any new functionality, and IMHO, it's just too simple to add.

RazvanN7:
ping @vporton . What is the status of this PR?

PR #7248, opened 2019-10-23
Added forgotten assert keyword

vporton:
Bugfix in unittest code.


vporton:
It seems that checks failed due to a network outage. Please retry.

PR #7249, opened 2019-10-23
Support of byUTF for ubyte[] argument

vporton:
For the reasons outlined in the discussion of [that pull request](https://github.com/dlang/phobos/pull/7247), we concluded that we need to be able to call `byUTF` on the argument of type `ubyte[]`. This PR implements exactly that.

I remind that this is necessary:

1. to support converting a chunk extracted from a file;
1. to eliminate the need to validate a string of `char`s two times: when it is created and when converted by `byUTF` (this simplifies programming and improves efficiency).

### DIFF
--- a/std/utf.d
+++ b/std/utf.d
@@ -4334,13 +4334,13 @@ if (isSomeChar!C)
     import std.algorithm.comparison : equal;
 
     // hellö as a range of `char`s, which are UTF-8
-    "hell\u00F6".byUTF!char().equal(['h', 'e', 'l', 'l', 0xC3, 0xB6]);
+    assert("hell\u00F6".byUTF!char().equal(['h', 'e', 'l', 'l', 0xC3, 0xB6]));
 
     // `wchar`s are able to hold the ö in a single element (UTF-16 code unit)
-    "hell\u00F6".byUTF!wchar().equal(['h', 'e', 'l', 'l', 'ö']);
+    assert("hell\u00F6".byUTF!wchar().equal(['h', 'e', 'l', 'l', 'ö']));
 
     // 𐐷 is four code units in UTF-8, two in UTF-16, and one in UTF-32
-    "𐐷".byUTF!char().equal([0xF0, 0x90, 0x90, 0xB7]);
-    "𐐷".byUTF!wchar().equal([0xD801, 0xDC37]);
-    "𐐷".byUTF!dchar().equal([0x00010437]);
+    assert("𐐷".byUTF!char().equal([0xF0, 0x90, 0x90, 0xB7]));
+    assert("𐐷".byUTF!wchar().equal([0xD801, 0xDC37]));
+    assert("𐐷".byUTF!dchar().equal([0x00010437]));
 }
--- a/std/utf.d
+++ b/std/utf.d
@@ -303,6 +303,8 @@ pure nothrow @safe @nogc unittest
 }
 
 
+private enum bool isCharOrUbyte(R) = isSomeChar!R || is(R == ubyte);
+
 /++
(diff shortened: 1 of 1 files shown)


lesderid:
What happens for a range of `ubyte`s that aren't valid UTF-8? Is this covered by a test?

std/utf.d
```diff
@@ -4354,6 +4354,9 @@ if (isSomeChar!C)
     // hellö as a range of `ubyte`s, which are UTF-8
     assert((cast(ubyte[]) [0x68, 0x65, 0x6c, 0x6c, 0xC3, 0xB6]).byUTF!char().equal(['h', 'e', 'l', 'l', 0xC3, 0xB6]));
 
+    assertThrown((cast(ubyte[]) [0xC3, 0x28]).byUTF!char());
```
lesderid: Needs an `import std.exception : assertThrown;`.

std/utf.d
```diff
@@ -60,7 +60,7 @@ $(TR $(TD Miscellaneous) $(TD
    +/
 module std.utf;
 
-import std.exception;  // basicExceptionCtors
+import std.exception;  // basicExceptionCtors, assertThrown
```
lesderid: This won't work because tests are extracted. Please import it inside the unittest block.

atilaneves:
I read the discussion on the other PR and AIUI it called for something that took `ubyte[]` and lazily produced validaded `char[]`. This doesn't seem to be it. Could you please explain what you're trying to accomplish? Thanks.

std/utf.d
```diff
@@ -4331,16 +4338,26 @@ if (isSomeChar!C)
...
     // 𐐷 is four code units in UTF-8, two in UTF-16, and one in UTF-32
-    "𐐷".byUTF!char().equal([0xF0, 0x90, 0x90, 0xB7]);
-    "𐐷".byUTF!wchar().equal([0xD801, 0xDC37]);
-    "𐐷".byUTF!dchar().equal([0x00010437]);
+    assert("𐐷".byUTF!char().equal([0xF0, 0x90, 0x90, 0xB7]));
+    assert("𐐷".byUTF!wchar().equal([0xD801, 0xDC37]));
+    assert("𐐷".byUTF!dchar().equal([0x00010437]));
+
+    // hellö as a range of `ubyte`s, which are UTF-8
+    assert((cast(ubyte[]) [0x68, 0x65, 0x6c, 0x6c, 0xC3, 0xB6]).byUTF!char().equal(['h', 'e', 'l', 'l', 0xC3, 0xB6]));
```
dukc, line 4356: Good, but I think you should test (directly) converting an `ubyte` range to utf-16 and/or utf-32

dukc:
> I read the discussion on the other PR and AIUI it called for something that took `ubyte[]` and lazily produced validaded `char[]`. This doesn't seem to be it. Could you please explain what you're trying to accomplish? Thanks.

The added unittest explains it:
`assert((cast(ubyte[]) [0x68, 0x65, 0x6c, 0x6c, 0xC3, 0xB6]).byUTF!char().equal(['h', 'e', 'l', 'l', 0xC3, 0xB6]));`

You pass in a range of `ubyte` and get a range of `char`. Handy, as no separate step to deal with autodecoding is needed.

std/utf.d
dukc: And also dealing with invalid UTF-8 in the `ubyte`, both with exceptions and replacement characters.

dukc:
Perhaps this should also accept `ushort` (assumed to be UTF-16) and `uint` (UTF-32)?

Of course, this can be implemented later on just as well.

RazvanN7:
ping @vporton

LightBender:

PR #7250, opened 2019-10-23
handle timeout on Windows the same as on Linux - no other way in curr…

adamdruppe:
…ent thing without dropping down to Win32 functions

### DIFF
--- a/std/socket.d
+++ b/std/socket.d
@@ -273,6 +273,16 @@ bool wouldHaveBlocked() nothrow @nogc
         static assert(0, "No socket support for this platform yet.");
 }
 
+@safe unittest
+{
+    auto sockets = socketPair();
+    auto s = sockets[0];
+    s.setOption(SocketOptionLevel.SOCKET, SocketOption.RCVTIMEO, dur!"msecs"(10));
+    ubyte[] buffer = new ubyte[](16);
+    auto rec = s.receive(buffer);
+    assert(rec == -1 && wouldHaveBlocked());
+}
+
 
 private immutable
 {
--- a/std/socket.d
+++ b/std/socket.d
@@ -259,12 +259,14 @@ class SocketFeatureException: SocketException
 /**
  * Returns:
  * `true` if the last socket operation failed because the socket
- * was in non-blocking mode and the operation would have blocked.
+ * was in non-blocking mode and the operation would have blocked,
+ * or if the socket is in blocking mode and set a SNDTIMEO or RCVTIMEO,
+ * and the operation timed out.
  */
 bool wouldHaveBlocked() nothrow @nogc
 {
     version (Windows)
-        return _lasterr() == WSAEWOULDBLOCK;
+        return _lasterr() == WSAEWOULDBLOCK || _lasterr() == WSAETIMEDOUT;
     else version (Posix)
         return _lasterr() == EAGAIN;
     else


adamdruppe:
I'm also a bit meh on the error message, trying to print the error fails on my test too

```
import std.socket;
import core.time;
import std.conv;
import std.stdio;
import core.sys.windows.windows;

void main() {
        auto socket = new Socket(AddressFamily.INET,  SocketType.STREAM);
        socket.setOption(SocketOptionLevel.SOCKET, SocketOption.RCVTIMEO, dur!
        "seconds"(1));
        socket.connect(new InternetAddress("localhost", 80.to!ushort));

        char[1024] buffer;
        auto got = socket.receive(buffer[]);
        if(got == -1)
                writeln(lastSocketError(), " ", wouldHaveBlocked());
}
```

but meh at least with this wouldHaveBlocked has consistent behavior.

CyberShadow:
Test please!

AIUI the test would pass with no changes on POSIX and fail on Windows without this change, but pass with this change.

adamdruppe:
A test for this would be super fickle since it is based on network stuff... any failure would be more likely caused by transient nonsense on the system rather than a code change causing a legitimate bug.

CyberShadow:
You could set up a server / peer in a separate thread. That's what e.g. the std.net.curl tests do.

CyberShadow:
Also `socketPair` could be helpful.

adamdruppe:
std/socket.d(281): Error: reference to local variable `buffer` assigned to non-scope parameter `buf` calling std.socket.Socket.receive

come on. this is the way i have always used this function - it is basically the only sane way TO use it - and now it is a random error too. but meh one more push i actually used dip1000 here now.

I feel like it is going outside the scope of this tiny fix to change the public API just to make a test *compile*.... so i'll dynamically allocate it too.

but tbh send and receive legit should be `scope`. I have a std.socket derived class and I suspect it'd technically not compile without adding scope there too so that'd be technically a breaking change too.... but like meh this scope stuff is all breaking anyway.

CyberShadow:
Yes, that's really annoying. We need to figure out some way to introduce new `Socket` method overloads with all the fancy attributes, and deprecate then remove the old ones.

CyberShadow:
Also needs some kind of documentation (changelog entry or linked issue).

adamdruppe:
ok i made an issue and edited the commit message, the bot seems to have picked it up in the comments https://issues.dlang.org/show_bug.cgi?id=20313 hopefully that will apply to the changelog too

adamdruppe:
on send and receive with attributes, we might not need to deprecate the old ones.

When inheriting, the subclass is allowed to tighten requirements. I just changed my subclass to scope and it still compiles. So we could potentially announce and document to subclassers to make the change NOW, then make the change in the base class then. It would still be a breaking change but a positive one, easy to fix, and give library people time to adjust.

CyberShadow:
> So we could potentially announce and document to subclassers to make the change NOW

The only reasonable way we can do this is through deprecation. Announcements in changelogs etc. aren't really sufficient - even with the assumption that all users to read the changelog, they may not be able to discover all affected code without the help of some tool.

PetarKirov:
I think the most straightforward way to go is to simply add new `final` member functions with the right signature, following the precedent set by `core.sync.mutex.Mutex`. 

> When inheriting, the subclass is allowed to tighten requirements. I just changed my subclass to scope and it still compiles.

In general yes, but that's not possible here without ugly casts, because to implement the method overrides in the derived class you still need to call the base class and then we're back to square one.

adamdruppe:
well in any case the attributes are a separate issue anyway.

PR #7251, opened 2019-10-24
Fix forgotten replacement of magic constants during PR merge.

berni44:


Geod24:
What happened ?

berni44:
> What happened ?

I had two PRs running on std.zip. The first one was about detection and rejection of chameleon-files and the second one about replacement of the magic constants by named constants. Then the second one was merged and produced merge conflicts in the first one. I resolved the conflict but did not notice, that the first PR still contained some magic constants.

I finally noticed this, when I wrote the fix for a regression. This fix has been merged into stable a few days ago and yesterday stable has been merged back into master. Which means, that now this can be fixed.

thewilsonator:
Auto-merge toggled on

PR #7253, opened 2019-10-24
Fix issue 20314: pass `only` parameters by value

FeepingCreature:
Let's see if there's any actual reason that the arguments to `only` are `auto ref`.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -9883,7 +9883,7 @@ Returns:
 
 See_Also: $(LREF chain) to chain ranges
  */
-auto only(Values...)(return scope auto ref Values values)
+auto only(Values...)(return scope Values values)
 if (!is(CommonType!Values == void) || Values.length == 0)
 {
     return OnlyResult!(CommonType!Values, Values.length)(values);
@@ -9919,6 +9919,16 @@ if (!is(CommonType!Values == void) || Values.length == 0)
 
     static assert(is(typeof(only((const(char)[]).init, string.init)) ==
         typeof(only((const(char)[]).init, (const(char)[]).init))));
+}
+
+// https://issues.dlang.org/show_bug.cgi?id=20314
+@safe unittest
+{
+    import std.algorithm.iteration : joiner;
+
+    const string s = "foo", t = "bar";
+
+    assert([only(s, t), only(t, s)].joiner(only(", ")).join == "foobar, barfoo");
 }
 
 // Tests the zero-element result


Geod24:
> Let's see if there's any actual reason that the arguments to `only` are `auto ref`.

Non copy-able elements ?

jmdavis:
> Non copy-able elements ?

As things stand, it's not legal for a range to have non-copyable elements, and even if it were, they wouldn't work with `only` anyway, because in order for `only` to work, they'd have to be moved into the range object that's returned from `only` instead of copied, whereas normally, the arguments are copied, and making any lvalues that were passed in be moved instead of copied could cause problems in the cases where the objects are copyable. If anything, `copy` is a prime example of why allowing non-copyable elements would be a problem, because `only`'s elements live directly in the object rather than being referred to by reference as occurs with many range types.

Geod24:
> As things stand, it's not legal for a range to have non-copyable elements

Is there any actual exhaustive definition of what a range is supposed / not supposed to do ?
For example, https://issues.dlang.org/show_bug.cgi?id=15413 might be invalid by your definition.

jmdavis:
> Is there any actual exhaustive definition of what a range is supposed / not supposed to do ?
For example, https://issues.dlang.org/show_bug.cgi?id=15413 might be invalid by your definition.

The core definition is `isInputRange`, and it specifically requires that the elements of a range be copyable. It also used to specifically require that the range be assignable, since it used to be

    template isInputRange(R)
    {
        enum bool isInputRange = is(typeof(
        (inout int = 0)
        {
            R r = R.init;     // can define a range object
            if (r.empty) {}   // can test for empty
            r.popFront();     // can invoke popFront()
            auto h = r.front; // can get the front of the range
        }));
    }

The current definition seems to have removed the requirement that the range be assignable, but I suspect that that was an oversight when it was changed to be a series of `&&`-ed conditions instead of one large condition. I'd have to dig through the history though to know for sure whether that change was on purpose.

    enum bool isInputRange(R) =
        is(typeof(R.init) == R)
        && is(typeof((return ref R r) => r.front))
        && is(typeof((R r) => r.popFront));

There is no specific requirement in `isInputRange` that the range be copyable, but since pretty much all range-based code takes arguments by value, and you have to copy a range in order to wrap it, it really doesn't make much sense to have a range that can't be copied. Also, as I just noted on that issue in bugzilla, the way that `foreach` in D works, it always copies any range it iterates over, and changing that would break a _lot_ of code. So, `isInputRange` should probably require that the range be copyable, but it doesn't currently have any checks for that one way or the other.

We could choose to purposefully allow non-copyable elements and non-copyable ranges, but with the way that range-based code works, both elements of ranges and the ranges themselves are copied all over the place (especially when one range wraps another), so neither non-copyable elements nor non-copyable ranges would work with vast majority of range-based code, and in principle, we'd then have to add checks for non-copyable ranges and elements all over the place. So, from what I can see, adding support for non-copyable types is a major complication that yields very little benefit.

Either way, `only` would _not_ play well with non-copyable elements, because the only way for it to allow them would be to move them, which is definitely not the behavior that you'd want in the vast majority of cases, and the range itself would then be non-copyable, which would mean that almost no range-based functions would work with it, because ranges are almost always passed by value (and of course, they're copied whenever they're wrapped).

FeepingCreature:
Uh... ping?

JohanEngelen:
This PR broke code.

The problem is that inside `only` there is a conversion of types if the parameters do not all have the same type but can be converted to a common type. After this PR, this conversion happens on the _local copy_ and then the local copy is thrown away (OnlyResult copies the converted value).

Example of code that broke (added some debug lines that show the pointer values to understand what is going wrong):
```d
import std.range;
import std.stdio;
import std.algorithm;

void main()
{
    // Comment-out one of the two to test the other, both are broken.
    broken_staticarray();
    broken_fixedstring();
}

void broken_staticarray()
{
    char[9] something = "something";
    writeln(__LINE__,":", cast(size_t)something.ptr);
    auto range = only("one", something, "three");
    foreach (ref v; range) {
        writeln(__LINE__,":", cast(size_t)v.ptr, " ,", v.length);
    }

    assert(only("one", something, "three").joiner(" ").equal("one something three"));
}

struct FixedLengthArray(T, size_t capacity_)
{
    enum capacity = capacity_;
 
	private size_t _length;
    private T[capacity] _elems;
    
    this(U)(U val) {
        opAssign(val);
    }
    
    auto ref opAssign(U)(U[] arr) {
        _length = arr.length;
        _elems[0 .. _length] = arr;
    }
    
    @property inout(T)[] slice() inout nothrow @nogc pure @safe {
        return _elems[0 .. _length];
    }

    alias slice this;
}

alias FixedString = FixedLengthArray!(char, 256);

void broken_fixedstring()
{
    FixedString something = "something";
    writeln(__LINE__,":", cast(size_t)something.slice.ptr);
    auto range = only("one", something, "three");
    foreach (ref v; range) {
        writeln(__LINE__,":", cast(size_t)v.ptr, " ,", v.length);
    }

    assert(only("one", something, "three").joiner(" ").equal("one something three"));
}
```

JohanEngelen:
https://issues.dlang.org/show_bug.cgi?id=21129

FeepingCreature:
I see. So the conversion should probably happen inside the `OnlyResult` struct with a tuple as template parameter rather than a `CommonType` static array.

PR #7254, opened 2019-10-25
Fix Issue 9592 - Justified Tuple printing

berni44:
Issue 9592 is about using width (and - flag) in a format parameter. For some types this is used, for others not. More strange: Sometimes it's used for elements (in arrays, ranges, AAs, partly classes and structs, where it shouldn't). More difficult: Elements of arrays and so on are handled different sometimes - e.g. strings are put into quotation marks. I do not address all of this with this PR.

I concentrated on %s only and only fixed some types. Additionally I put a note in the docs, where width is currently used and where not. (Which actually is the fix to issue 9592.)

There is one more, unrelated thing: In line 4541, the style checker hit a false positive. I used a hack to get arround this, but I think, this bug should be reported. Is bugzilla with component Phobos the right place to do so?

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -190,7 +190,7 @@ $(I FormatChar):
     $(BOOKTABLE Flags affect formatting depending on the specifier as
     follows., $(TR $(TH Flag) $(TH Types&nbsp;affected) $(TH Semantics))
 
-    $(TR $(TD $(B '-')) $(TD numeric) $(TD Left justify the result in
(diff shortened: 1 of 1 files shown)


PetarKirov:
> Is bugzilla with component Phobos the right place to do so?

That should be the `tools` component.

PetarKirov:
> I do not address all of this with this PR.

I'll leave the bug open, until we address all cases.

PR #7255, opened 2019-10-25
Rework documentation of std.zip.

berni44:
The format of the table in the main documentation is not yet perfect, but I could not find anything better. Especially I found no local link version, where the displayed value is different to the anchor value. I used REF1_ALTTEXT for this.

I left out the flags and internalAttributes of ArchiveMember, because I don't know exactly their meaning. As I have to investigate on this anyway, I'll fill this in, when I know better.

I would be happy if a native english speaker could correct any language mistakes I made.

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -1,61 +1,104 @@
 // Written in the D programming language.
 
 /**
- * Read/write data in the $(LINK2 https://en.wikipedia.org/wiki/Zip_%28file_format%29, zip archive) format.
- * Makes use of the etc.c.zlib compression library.
- *
- * Limitations:
- *      $(UL
- *      $(LI Multi-disk zips not supported.)
- *      $(LI Only Zip version 20 formats are supported.)
- *      $(LI Only supports compression modes 0 (no compression) and 8 (deflate).)
- *      $(LI Does not support encryption.)
- *      )
- *
- * Example:
- * ---
-// Read existing zip file.
-import std.digest.crc, std.file, std.stdio, std.zip;
+Read and write data in the
+$(LINK2 https://en.wikipedia.org/wiki/Zip_%28file_format%29, zip archive)
+format.
+
+Standards:
+
+The current implementation mostly conforms to
+$(LINK2 https://www.iso.org/standard/60101.html, ISO/IEC 21320-1:2015),
+which means,
+$(UL
+$(LI that files can only be stored uncompressed or using the deflate mechanism,)
+$(LI that encryption features are not used,)
+$(LI that digital signature features are not used,)
+$(LI that patched data features are not used, and)
+$(LI that archives may not span multiple volumes.)
+)
+
+Additionally, archives are checked for malware attacks and rejected if detected.
+This includes
+$(UL
+$(LI $(LINK2 https://news.ycombinator.com/item?id=20352439, zip bombs) which
+     generate gigantic amounts of unpacked data)
+$(LI zip archives that contain overlapping records)
+$(LI chameleon zip archives which generate different unpacked data, depending
+     on the implementation of the unpack algorithm)
+)
+
+The current implementation makes use of the zlib compression library.
+
+Usage:
+
+There are two main ways of usage: Extracting files from a zip archive
+and storing files into a zip archive. These can be mixed though (e.g.
+read an archive, remove some files, add others and write the new
+archive).
+
+Examples:
+
+Example for reading an existing zip archive:
+---
+import std.stdio : writeln, writefln;
+import std.file : read;
+import std.zip;
 
 void main(string[] args)
 {
     // read a zip file into memory
     auto zip = new ZipArchive(read(args[1]));
-    writeln("Archive: ", args[1]);
-    writefln("%-10s  %-8s  Name", "Length", "CRC-32");
+
     // iterate over all zip members
+    writefln("%-10s  %-8s  Name", "Length", "CRC-32");
     foreach (name, am; zip.directory)
     {
         // print some data about each member
         writefln("%10s  %08x  %s", am.expandedSize, am.crc32, name);
         assert(am.expandedData.length == 0);
+
         // decompress the archive member
         zip.expand(am);
         assert(am.expandedData.length == am.expandedSize);
     }
 }
+---
 
-// Create and write new zip file.
+Example for writing files into a zip archive:
+---
 import std.file : write;
 import std.string : representation;
+import std.zip;
 
 void main()
 {
-    char[] data = "Test data.\n".dup;
-    // Create an ArchiveMember for the test file.
-    ArchiveMember am = new ArchiveMember();
-    am.name = "test.txt";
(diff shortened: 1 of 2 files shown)


std/zip.d
```diff
@@ -1,61 +1,149 @@
...
- * ---
-// Read existing zip file.
-import std.digest.crc, std.file, std.stdio, std.zip;
-
-void main(string[] args)
+Read and write data in the $(LINK2 https://en.wikipedia.org/wiki/Zip_%28file_format%29, zip archive) format.
+
+Standards:
+
+The current implementation mostly conforms to $(LINK2 https://www.iso.org/standard/60101.html, ISO/IEC 21320-1:2015),
```
berni44: Mostly, because there might be some nonconformation left due to UTF8 issues or with the flags. I'm not totally in this currently, but have to investigate soon for fixing some further bugs.

std/zip.d
```diff
@@ -1,61 +1,149 @@
...
+Additionally, archives are checked for malware attacks and rejected if detected. This includes
+$(UL
+$(LI $(LINK2 https://news.ycombinator.com/item?id=20352439, zip bombs) which generate gigantic amounts of
+     unpacked data)
+$(LI zip archives that contain overlapping records)
+$(LI chameleon zip archives which generate different unpacked data, depending on the implementation of the
+     unpack algorithm)
+)
+
+The current implementation makes use of the etc.c.zlib compression library.
```
berni44: I'm not sure, if this is still correct. The implementation uses std.zlib, which probably is based on etc.c.zlib. I'm not sure if this comment is of any help, but it was in the original documentation and so I decided to preserve it.

std/zip.d
```diff
@@ -420,22 +546,26 @@ public:
         assert(de._compressedData.length == de._compressedSize, "Archive member compressed failed.");
     }
 
-    /** Delete de from the archive.
+    /**
+     * Delete ArchiveMember de from the archive. Uses the name of the member to detect which element to delete.
```
CyberShadow: Preferably keep lines under 80-100 characters.

std/zip.d
CyberShadow: Neither `etc.c.zlib` nor `std.zlib` are a compression library, they merely wrap one. Probably it would be more correct if this was changed to either say that it uses the zlib library or the `etc.c.zlib` bindings / `std.zlib` wrapper for the zlib library.

std/zip.d
```diff
@@ -156,8 +256,11 @@ final class ArchiveMember
     }
 
     /**
-     * Set the OS specific file attributes, as obtained by
-     * $(REF getAttributes, std,file) or $(REF DirEntry.attributes, std,file), for this archive member.
+     * Set the OS specific file attributes for this archive member.
+     *
+     * Params:
```
CyberShadow: In [.dscanner.ini](https://github.com/dlang/phobos/blob/master/.dscanner.ini) you can enable various checks by removing std.zip from the blacklists. This will help find more violations of the style guidelines and once addressed, ensure that future changes maintain them.

std/zip.d
```diff
@@ -1,61 +1,149 @@
...
+               $(TD Create an array, representing the entire contents of this $(D ZipArchive).))
+          $(TR $(TD $(REF1_ALTTEXT $(D directory), ZipArchive.directory, std_zip))
+               $(TD Associative Array containing all ArchiveMembers.))
+          $(TR $(TD $(REF1_ALTTEXT $(D data), ZipArchive.data, std_zip))
+               $(TD Array, representing the entire contents of this $(D ZipArchive).))
+          $(TR $(TD $(REF1_ALTTEXT $(D comment), ZipArchive.comment, std_zip))
+               $(TD The global comment of the archive file.))
+          $(TR $(TD $(REF1_ALTTEXT $(D isZip64), ZipArchive.comment, std_zip))
+               $(TD True, if $(D ZipArchive) is in zip64 format. Set this to true, if you want to force zip64 format
+                    when using $(D build).))
```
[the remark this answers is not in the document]
CyberShadow: Doesn't look good in DDox, though I don't know if the problem is due to the DDoc here or DDox:

http://dtest.dlang.io/artifact/website-2ad4ace80675bb9b91837246ebe69a2b668ec297-59a3acd7ce221d4dcfb2be853c277543/web/library-prerelease/std/zip.html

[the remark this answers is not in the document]
berni44: I wrote the table, because Jonathan M Davis suggested that in the forum. But if they are generated automatically in DDox, that's better. I didn't know about DDox though. And indeed the tables don't look good there; it's a little bit of a hack anyway and I wasn't quite happy with the result.

Do you know, when the default view of the documentation will be switched to DDox? From first sight, I'd say the sooner the better.

std/zip.d
berni44: I read somewhere in the docs, that they should be less than 120, but 100 or 80 as a limit is fine for me too. :-)

std/zip.d
berni44: Good to know! Thanks for that hint.

std/zip.d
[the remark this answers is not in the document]
CyberShadow: > Do you know, when the default view of the documentation will be switched to DDox? From first sight, I'd say the sooner the better.

Probably no sooner than it getting an active maintainer, which hasn't happened for a few years now.

std/zip.d
CyberShadow: 120 is a hard limit, where even "ugly" methods of breaking up a line should be used. There is nothing ugly about wrapping a paragraph of text at 80 characters instead of 120, though.

std/zip.d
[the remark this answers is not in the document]
wilzbach: Well, at this point it's more realistic that we'll remove Ddox as the transition never happened :/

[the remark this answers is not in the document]
berni44: That's a pity. Ddox looks so much better that ddoc... And it's allready there. Looks like the only thing that's needed is to toggle a switch... What sort of maintainance is needed in addition?

std/zip.d
```diff
@@ -1,61 +1,98 @@
...
+---
 import std.file : write;
 import std.string : representation;
+import std.zip;
+
+// Create an ArchiveMembers for each file.
+ArchiveMember file1 = new ArchiveMember();
+file1.name = "test1.txt";
+file1.expandedData("Test data.\n".dup.representation);
+file2.compressionMethod = CompressionMethod.none; // don't compress
```
n8sh: Need to change `file2` to `file1` to make this runnable. Maybe the example should be made into a documented unit test, although adding cleanup code to delete the generated zip file would make the example slightly worse. (Worse because it includes code that does nothing to make the demonstration clearer but exists solely for the convenience of the unit test framework.)

berni44: Corrected the bug. Thanks for noting.

berni44:
Resolved merge conflict and changed Posix to POSIX.

PR #7256, opened 2019-10-27
fix issue 20328 - deprecation message about Nullable.get in isInputRange

clYd3r:
Please not well that actually this test case **cant** be fixed by using `.get` but at least the deprecation message is prevented when a user module that imports a `Nullable` is unittested, which can happend even indirectly, e.g when using `std.json.parseJSON`.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2886,8 +2886,9 @@ Returns:
     assert(app.data == "1");
 }
 
-// Issue 19799
-@safe unittest
+// Issue 19799, disabled because of issue 20328. Formatting a null Nullable triggers a deprecation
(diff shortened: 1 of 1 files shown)


Geod24:
Why not mark the unittest as deprecated instead ?

clYd3r:
Yeah just do it this way if it's better, I don't care. My commitment is just about the fact that I don't want that the deprecation message gets unfairly displayed until the end of the deprecation. Do it quick on stable branch since next release is about to be dropped.

clYd3r:
move this test at the global scope and attribute it as deprecated, ASAP.

PR #7257, opened 2019-10-29
Fix issue 20328: only check isInputRange when we've already excluded toString overload.

FeepingCreature:
This prevents `isInputRange` from noticing that `Nullable!string` is technically an input range (because string is) and hitting the `.get` deprecation.


thewilsonator:
test?

FeepingCreature:
Untestable.

As the code just reorders a boolean expression, the behavior is the exact same. The only difference is whether a deprecation warning is output, which cannot be unittested.

thewilsonator:
Ok. Stable?

FeepingCreature:
Moved to stable.

thewilsonator:
Auto-merge toggled on

PR #7258, opened 2019-10-29
Issue20330 json toString with OutRange

burner:
std.json.JSONValue.toString and toJSON does not work with output ranges.
This PR fixes it.

https://issues.dlang.org/show_bug.cgi?id=20330

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -761,6 +761,12 @@ struct JSONValue
         return toJSON(this, false, options);
     }
 
+    ///
+    void toString(Out)(Out sink, in JSONOptions options = JSONOptions.none) const
+    {
+        toJSON(sink, this, false, options);
+    }
+
     /***
      * Implicitly calls `toJSON` on this JSONValue, like `toString`, but
      * also passes $(I true) as $(I pretty) argument.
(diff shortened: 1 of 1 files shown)


std/json.d
```diff
@@ -1287,10 +1299,15 @@ If `pretty` is false no whitespaces are generated.
 If `pretty` is true serialized string is formatted to be human-readable.
 Set the $(LREF JSONOptions.specialFloatLiterals) flag is set in `options` to encode NaN/Infinity as strings.
 */
-string toJSON(const ref JSONValue root, in bool pretty = false, in JSONOptions options = JSONOptions.none) @safe
-{
+string toJSON(const ref JSONValue root, in bool pretty = false, in JSONOptions options = JSONOptions.none) @safe {
```
CyberShadow: Style regression.

std/json.d
```diff
@@ -1290,7 +1302,13 @@ Set the $(LREF JSONOptions.specialFloatLiterals) flag is set in `options` to enc
 string toJSON(const ref JSONValue root, in bool pretty = false, in JSONOptions options = JSONOptions.none) @safe
 {
     auto json = appender!string();
+    toJSON(json, root, pretty, options);
+    return json.data;
+}
 
+void toJSON(Out)(auto ref Out json, const ref JSONValue root, in bool pretty = false, in JSONOptions options = JSONOptions.none) @safe
```
PetarKirov:
```suggestion
void toJSON(Out)(
    auto ref Out json,
    const ref JSONValue root,
    in bool pretty = false,
    in JSONOptions options = JSONOptions.none) @safe
```

Otherwise it fails the style check with:
```
std/json.d(1309:112)[warn]: Line is longer than 120 characters
```

burner: thank you, should be fixed now

burner:
@atilaneves @CyberShadow @PetarKirov thank you

PR #7264, opened 2019-11-08
Partial replace call to snprintf for formating floatingpoint numbers.

berni44:
Because the description of this PR is long and contains diagrams, I put everything into [printFloat.pdf](https://github.com/berni44/printFloat/blob/master/printFloat.pdf).

TL;DR: Read only the abstract.

Still TL;DR: This only replaces the 'f' specifier for float and double. The algorithm is correct, in most cases faster than the old one (some optimizations are still WIP), it implements rounding according to IEEE754 (with a different tie than the old one), but no internationalisation, because that causes a lot of trouble and should be done elsewhere.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -2977,6 +2977,13 @@ useSnprintf:
     assert(format!"%a"(nextUp(0.0)) == "0x0.0000000000001p-1022");
 }
 
+// issue 20371
+@safe unittest
+{
+    assert(format!"%.1000a"(1.0).length == 1007);
+    assert(format!"%.600f"(0.1).length == 602);
+}
+
 /*
     Formatting a `creal` is deprecated but still kept around for a while.
  */
--- a/std/format.d
+++ b/std/format.d
@@ -7131,7 +7131,7 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
         case 'e': case 'E':
             return printFloatE(buf, val, f, rm, sgn, exp, mnt, is_upper);
         case 'f': case 'F':
-            return printFloatF(val, f, rm, sgn, exp, mnt, is_upper);
+            return printFloatF(buf, val, f, rm, sgn, exp, mnt, is_upper);
     }
 }
 
@@ -7627,7 +7627,7 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
 
     // special treatment for 0.0
     if (exp == 0 && mnt == 0)
-        return printFloat0(f, sgn, is_upper);
+        return printFloat0(buf, f, sgn, is_upper);
 
     // add leading 1 for normalized values or correct exponent for denormalied values
     if (exp != 0)
@@ -8430,7 +8430,7 @@ printFloat_done:
     assert(printFloat(buf[], -245.666f, f) == "-2.5E+02");
 }
 
-private auto printFloatF(T, Char)(T val, FormatSpec!Char f, RoundingMode rm,
+private auto printFloatF(T, Char)(return char[] buf, T val, FormatSpec!Char f, RoundingMode rm,
                                   string sgn, int exp, ulong mnt, bool is_upper)
 if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.mant_dig))
 {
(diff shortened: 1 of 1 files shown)


berni44:
The failure on eBay/tsv-utils is due to the different tie in rounding. This is intentional (I think the new version is better) but might be discussed.

p0nce:
There are only 2^32 float so you could do an exhaustive check there.

berni44:
> There are only 2^32 float so you could do an exhaustive check there. (I don't mean in a commited unittest)

I did that allready with two different format strings - see pdf.

berni44:
> this has not gotten enough testing. Try double.max (1.0/3.0) inf -inf machine epsilon pi 2pi and a few other well known constants, at the very least.

inf (line 7743/7744) and epsilon (line 7788) are allready in the unittests. I'll add max and some special constants.

berni44:
added more unittests (and sorry @UplinkCoder I overlooked, that the eps-tests where float not double, added the double tests now)

berni44:
While working on `%a` I discovered a bug in the implementation of the rule "Round to nearest, ties away from zero". It was actually "Round to nearest, ties up", which is different for negative numbers. I fixed this (and the wrong unittests) now.

p0nce:
Just to say that this seems like fantastic work if the PDF says the truth. For localization I also think it fixes more undiscovered bugs than it breaks (there are more people unaware of localization using snprintf unknowingly, than people using D and relying on printf localization)

berni44:
> Just to say that this seems like fantastic work if the PDF says the truth.

Thanks for compliment! :-) If you want to check the content of the pdf: I put all needed functions in the [repo](https://github.com/berni44/printFloat/) and would be happy, if someone could verify the results...

berni44:
Changed tie for rounding to the current behaviour. I'll address this later in a separate PR. I also added improvements suggested in PR #7285.

berni44:
Does anyone has an idea, what this 32bit-windows compiler bug is about?

berni44:
Merged everything and run the external tests (for %f and %a): all passed.

I changed bias from long to int, because I guess, this is the reason, why win32 crashed after the last change (had something similar resently in an other PR, as far as I remember). Let's see...

I also unified the tests for large precissions. I think there is no merit in having a string literal, which essentially consists of 1000 zeros in the source code. For checking the bug, it's sufficient to just check the length.

berni44:
Fixed problem with global import of `FloatingPointControl`. See #7346 for more details.

berni44:
Fixed the merge conflicts.

std/format.d
```diff
@@ -8403,3 +8449,696 @@ printFloat_done:
...
+    import std.conv : to;
+    import std.algorithm.comparison : max;
+
+    enum int bias = T.max_exp - 1;
+
+    if (f.precision == f.UNSPECIFIED)
+        f.precision = 6;
+
+    // special treatment for 0.0
+    if (exp == 0 && mnt == 0)
```
thewilsonator: Can you factor this branch out into a separate function and merge the logic with the `0.0` case for `printFloatE`. They look almost identical, modulo some minor handling for the difference in the format specifiers.

berni44: done.

thewilsonator: Thanks!

std/format.d
```diff
@@ -7085,6 +7130,8 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
             return printFloatA(val, f, rm, sgn, exp, mnt, is_upper);
         case 'e': case 'E':
             return printFloatE(val, f, rm, sgn, exp, mnt, is_upper);
+        case 'f': case 'F':
+            return printFloatF(val, f, rm, sgn, exp, mnt, is_upper);
```
thewilsonator: Following the success of the float zero factorisation, I'm beginning to think that the best way to organise the printing is not by format specifier (%a,%f,%g,%e) but by "category" (for want of a better term), i.e. 
* inf/nan (handled above), 
* Algorithm A: For large exponents (exp >= T.mant_dig)
*  Algorithm B: For small exponents (exp < T.mant_dig - 61)
* Algorithm C: For exponents close to 0. (already done)

Since there seems to be much more in common between the different ranges of numbers than there is between the different format specifiers.

rainers:
Rebased and adapted to not allocate as in https://github.com/dlang/phobos/pull/7393

thewilsonator:
Implemented in another PR

rainers:
> Implemented in another PR

I'm not following phobos development closely, but which one would that be?

thewilsonator:
#7393 No?

rainers:
> #7393 No?

#7285 and #7318 implemented "%a" and "%e" formatting, and #7393 removed some introduced allocations. This PR implements "%f" formatting. I only rebased it and modified it to not introduce allocations again.

ljmf00:
Do you anyone can push this forward @rainers or @thewilsonator ? This PR is crucial to have purity on `format` for floating point.

RazvanN7:
Closing this in favor of #7757

PR #7266, opened 2019-11-11
Made SortedRange verifications conditionally compiled.

dukc:
Those verifications broke SortedRange with -betterC even when they were off.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -233,6 +233,7 @@ public import std.range.interfaces;
 public import std.range.primitives;
 public import std.typecons : Flag, Yes, No;
 
+import std.internal.attributes : betterC;
 import std.meta : allSatisfy, staticMap;
 import std.traits : CommonType, isCallable, isFloatingPoint, isIntegral,
     isPointer, isSomeFunction, isStaticArray, Unqual, isInstanceOf;
@@ -10645,6 +10646,7 @@ if (isInputRange!Range && !isInstanceOf!(SortedRange, Range))
     }
 
     // Assertion only.
+    static if (opt == SortedRangeOptions.checkRoughly)
     private void roughlyVerifySorted(Range r)
     {
         if (!__ctfe)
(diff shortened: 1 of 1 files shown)


dukc:
> Please add a `@BetterC` test

How? Haven't done that

MoonlightSentinel:
Create a unittest annotated with `@betterC` (declared in `std.internal.attributes`):

https://github.com/dlang/phobos/blob/073960ecae0a9646981186818484069f81473f23/std/algorithm/comparison.d#L119-L130

dukc:
One more question: How do I test locally only the -@betterC unittests?

MoonlightSentinel:
Have a look at `posix.mak`:

https://github.com/dlang/phobos/blob/073960ecae0a9646981186818484069f81473f23/posix.mak#L633-L651

I think `make -f posix.mak std/range/package.betterc` should work

dukc:
Nah, requires also cloning the dmd repository, which I'm not in a mood to do now. I'll just push and hope it passes the tests in cloud.

PetarKirov:
The Makefile target for `%.betterc` extracts `unittest`s with the `@betterC` to separate files and the runs them with a command like this:
```
../dmd/generated/linux/release/64/dmd  -conf= -I../druntime/import  -w -de -preview=dip1000 -m64 -fPIC -transition=complex -O -release -defaultlib= -debuglib= -L-lpthread -L-lm -betterC -unittest -run generated/linux/release/64/betterctests/std_range_package.d
```

A simplified command-line that also reproduce the problem is as follows:
```
../dmd/generated/linux/release/64/dmd -conf= -I../druntime/import -preview=dip1000 -betterC -unittest generated/linux/release/64/betterctests/std_range_package.d
```


@wilzbach The reason for the failure is that we're building the extracted `@betterC` files with `-unittest`, however this makes dmd also build phobos unit tests part of the imported files, which as expected fail when compiled with `-betterC` (as not all of them are meant to be built with `-betterC`).

I think we should change the test extractor to emit `void test_line1234()`, instead of `unitest`, for each extracted unit test. What do you think?

PetarKirov:
> I think we should change the test extractor to emit void test_line1234(), instead of unitest, for each extracted unit test. What do you think?

-> https://github.com/dlang/tools/pull/386

PetarKirov:
Blocked, waiting for:
* [ ] https://github.com/dlang/tools/pull/386
* [ ] https://github.com/dlang/phobos/pull/7269

PR #7267, opened 2019-11-12
Reactivate unittests for findRoot

berni44:
When trying to do more covering for findRoot I noticed, that all unittests where commented out. I tried to find out, why this happend. It was in 2009 in commit 0c1429... The reason was probably, that an erroneous assert inside of testFindRoot was added and with little time left (the next release was due), the solution was to comment out the unittests.

I fixed the assert (although I think it's not 100% correct now, because one of the numbers could be 0 and the other >=0 which means, both have the same sign and still the call is correct; maybe, the assert should be removed alltogether?).


thewilsonator:
Auto-merge toggled on

PR #7268, opened 2019-11-12
Add closures to warning and mark warning in red.

berni44:
This addresses [issue 9603](https://issues.dlang.org/show_bug.cgi?id=9603). I think, inside of Phobos this cannot be repaired (maybe with a complete rewrite), nor is it possible to add an exception to prevent the crash.

I remember, that I've seen an even more impressive warning somewhere, but could not find it again. I think it would be the right thing to add here.


berni44:
ping

PetarKirov:
Thanks for the reminder.

PR #7270, opened 2019-11-12
Fix Issue 10902 - some phobos unittests take an excessive amount of time

berni44:
Top10 before:

0.190s PASS release64 std.datetime
0.246s PASS release64 std.uni
0.296s PASS release64 std.uri
0.302s PASS release64 std.socket
0.447s PASS release64 std.digest.sha
0.678s PASS release64 std.datetime.systime
2.066s PASS release64 std.random
3.492s PASS release64 std.parallelism
5.003s PASS release64 std.algorithm.setops
5.099s PASS release64 std.process

Top10 after:

0.192s PASS release64 std.datetime
0.248s PASS release64 std.uni
0.251s PASS release64 std.algorithm.setops
0.298s PASS release64 std.uri
0.305s PASS release64 std.socket
0.447s PASS release64 std.digest.sha
0.653s PASS release64 std.datetime.systime
0.875s PASS release64 std.parallelism
0.917s PASS release64 std.process
1.370s PASS release64 std.random

For std.process I checked the time needed on my computer, it was less than 200 microseconds; I put in 200 milliseconds which is 1000 times what's needed here. I think, this should suffice.

For std.random I identified the slowest test and halved the number of samples in that test. It's still the slowest test there and maybe it should be removed for normal phobos tests alltogether?

I also made sure, that coverage did not drop.

### DIFF
--- a/std/algorithm/setops.d
+++ b/std/algorithm/setops.d
@@ -485,7 +485,7 @@ if (!allSatisfy!(isForwardRange, R1, R2, RR) ||
     assert(is(ElementType!(typeof(N3)) == Tuple!(size_t,size_t,size_t)));
 
     assert(canFind(N3, tuple(0, 27, 7)));
-    assert(canFind(N3, tuple(50, 23, 71)));
+    assert(canFind(N3, tuple(50, 23, 11)));
     assert(canFind(N3, tuple(9, 3, 0)));
 }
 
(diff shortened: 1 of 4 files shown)


berni44:
200ms seems to be too few for the win32 machine. I moved it up to 500ms.

thewilsonator:
Nice, it would be great if you could run coverage for this (there is a bot that does it but it seems to take a while).

wilzbach:
> Nice, it would be great if you could run coverage for this (there is a bot that does it but it seems to take a while).

https://codecov.io/gh/dlang/phobos/compare/073960ecae0a9646981186818484069f81473f23...97abb919524017f87fdfcedc4e86f0aaea03cd29

thewilsonator:
Oh, it just seems to randomly comment on really old PRs and I though that was an indication of analysis time.

thewilsonator:
Auto-merge toggled on

wilzbach:
> Oh, it just seems to randomly comment on really old PRs and I though that was an indication of analysis time.

Yes, those don't have a `.codecov.yml` which disables the bot.

https://github.com/dlang/phobos/blob/master/.codecov.yml

The GH notification update should appear always though.

PR #7271, opened 2019-11-13
Fix partially Issue 11809 - unit tests leave behind temporary files

berni44:
On linux there is nothing left now, but I have no means to check that for other OS.


thewilsonator:
Auto-merge toggled on

PR #7273, opened 2019-11-13
Fix partially Issue 9621 - std.conv.parseEscape fails on octals and named

berni44:
This is the simple part of that issue...


thewilsonator:
Auto-merge toggled on

PR #7276, opened 2019-11-14
Fix Issue 11013 - ignoring variable inside the predicate of find

berni44:
Two versions of find matched, when called with two strings. Now only the one found in line 1878 matches.


atilaneves:
Issue 11013 was  about `findSplitBefore`, but that's not what the unit tests is testing.

berni44:
> Issue 11013 was about `findSplitBefore`, but that's not what the unit tests is testing.

Actually findSplitBefore just calls find and that's the place, where the error occurs - see the errormessages in the bug report.

atilaneves:
> > Issue 11013 was about `findSplitBefore`, but that's not what the unit tests is testing.
> 
> Actually findSplitBefore just calls find and that's the place, where the error occurs - see the errormessages in the bug report.

That's an implementation detail. The bug is specifically about findSplitBefore.

PR #7277, opened 2019-11-14
Fix Issue 11007 - std.format.formattedRead with %*s specifier causes AssertError

berni44:
I replaced the assert by a throw.

I also fixed, that %*d and %*u should at least read one digit, else it's no number.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -5501,12 +5501,48 @@ private void skipData(Range, Char)(ref Range input, scope const ref FormatSpec!C
             if (input.front == '+' || input.front == '-') input.popFront();
             goto case 'u';
         case 'u':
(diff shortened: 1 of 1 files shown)


atilaneves:
I don't understand the original issue itself - why an exception instead of an assertion? Does the failure depend on what's passed at runtime?

MoonlightSentinel:
This seems wrong, an invalid format string is usually a programming error (except hypothetical cases in which the string is user input). 

Additionally this makes `formattedRead` always non-`@nogc`

berni44:
You're right. It's correct to be an error. But there are several other issues around here. I added some bug reports and will address these in separate PRs.

Thanks for this fast feedback! :-)

PR #7279, opened 2019-11-16
std.zip: Remove some unnecessary private variables

berni44:
Without multidisk support, numEntries and totalEntries are the same numbers and furthermore they should always be the same as the length of the directory. This is enforced, when reading a directory, but when adding a member to the directory or deleting one this is currently not the case. See [issue 20398](https://issues.dlang.org/show_bug.cgi?id=20398).

This creates a change-feature-and-deprecate case. But I see no merit in first fixing the bug with a PR and then deprecating the feature in the next PR.

I also replaced the private variable endrecOffset by a local variable inside findEndOfCentralDirRecord(). Beside being a return value, this was only used in lines 844 to 851. The checks and the copy there are allready done in the constructor and therefore there is no need to do it once again.

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -439,7 +439,6 @@ public:
     string comment; /// The archive comment. Must be less than 65536 bytes in length.
 
     private ubyte[] _data;
-    private uint endrecOffset;
 
     private bool _isZip64;
     static const ushort zip64ExtractVersion = 45;
@@ -736,7 +735,6 @@ public:
         }
 
         // Write end record
-        endrecOffset = i;
         _data[i .. i + 4] = endOfCentralDirSignature;
         putUshort(i + 4,  cast(ushort) 0);
         putUshort(i + 6,  cast(ushort) 0);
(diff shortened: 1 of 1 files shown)


std/zip.d
```diff
@@ -343,8 +340,9 @@ public:
     @property @safe @nogc pure nothrow uint diskStartDir() const { return 0; }
 
     /// Read Only: number of ArchiveMembers in the directory.
-    @property @safe @nogc pure nothrow uint numEntries() const { return _numEntries; }
-    @property @safe @nogc pure nothrow uint totalEntries() const { return _totalEntries; }    /// ditto
+    deprecated("Use totalEntries instead; will be removed in 2.099.0")
+    @property @safe @nogc pure nothrow uint numEntries() const { return cast(uint) _directory.length; }
+    @property @safe @nogc pure nothrow uint totalEntries() const { return cast(uint) _directory.length; }    /// ditto
```
n8sh: I question the need to mark `numEntries` for removal. Every existing D program that used std.zip and wrote `numEntries` instead of `totalEntries` will need to be manually changed or else fail to compile within the next year, but there will be no improvement in the behavior of those programs in exchange for this labor. This is good for keeping programmers employed but little else.

CyberShadow: One or the other needs to go, otherwise we will forevermore have two getters with the same value.

n8sh: What is wrong with an alias?

berni44: IMHO programs that use `numEntries` instead of `totalEntries` probably have a potential bug. I don't think, that this bug will ever appear, but that's only, because I think, we will never add multi disk support. With multi disk support, the two entries would differ (`numEntries` is the number of entries on this disk, while `totalEntries` is the number of entries on all disks) and it's high likely that `numEntries` is the wrong one.

Having said this, I'm not sure, what's the better option...

berni44:
resolved merge conflict

berni44:
ping

thewilsonator:
Whoops.

PR #7281, opened 2019-11-16
Remove deprecated symbols from std.digest.digest

MoonlightSentinel:
Blocked by:

- [x] CyberShadow/ae#42
- [x] Upgrade [dlang-bot](https://github.com/dlang/dlang-bot) to newer vibe-d

### DIFF
--- a/changelog/digest.dd
+++ b/changelog/digest.dd
@@ -1,0 +1,7 @@
+The deprecated aliases in std.digest.digest were removed
+
+They were deprecated since 2.076.1 and have now been removed.
+Import `std.digest` or its submodules instead.
+
+Additionally the deprecation cycle for `std.digest` was started
+and the module will be removed in 2.101.
--- a/std/digest/crc.d
+++ b/std/digest/crc.d
@@ -135,13 +135,13 @@ alias CRC32 = CRC!(32, 0xEDB88320);
 
 /**
  * Template API CRC64-ECMA implementation.
- * See `std.digest.digest` for differences between template and OOP API.
+ * See `std.digest` for differences between template and OOP API.
  */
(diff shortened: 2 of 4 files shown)


CyberShadow:
Uh-oh, looks like there's some un-announced breakage. The symbols in `std.digest.digest` were deprecated, but the module itself was not. I guess it would need to go through a deprecation cycle too, but maybe we can expedite this one since by itself the import either does nothing, or would have already caused deprecation messages due to used aliases.

MoonlightSentinel:
You're right, this seems to be an oversight as the deprecation messages already indicated that the entire module will be removed.

I would suggest removing the entire file instead of keeping an empty module as it only requires changes in `ae` and `vibe-d:mongo` (which requires 2.077.1 while the deprecation was introduced in 2.076.1).

But I can adapt this PR if an additional deprecation cycle is preferred.

CyberShadow:
> as it only requires changes in `ae` and `vibe-d:mongo`

The projects we have in the project tester are really just canaries to catch some % unexpected breaking changes. We can't know about the effects of changes on code we can't see (private codebases, and also the long tail of smaller open-source codebases), but seeing how there is more than one breakage in the project tester, we can extrapolate that it seems likely that it will affect other projects too.

MoonlightSentinel:
Changed to only remove the deprecated symbols and started the deprecation cycle

MoonlightSentinel:
Added the missing changelog

thewilsonator:
```
../.dub/packages/vibe-d-0.8.4/vibe-d/mongodb/vibe/db/mongo/settings.d(16,8): Error: module `std.digest.digest` import `toHexString` not found
```

MoonlightSentinel:
This should be green now thanks to dlang/dlang-bot#239

PR #7282, opened 2019-11-16
Advance deprecation of benchmarking functionality in std.datetime

MoonlightSentinel:
Undocumented all public examples and extended the deprecation message stating that all deprecated symbols will be removed after 2.094.


std/datetime/package.d
```diff
@@ -182,12 +182,13 @@ import std.typecons : Flag, Yes, No;
 
     Otherwise, if set to `AutoStart.yes`, then the stopwatch is started when
     it is constructed.
-  +/
-deprecated("Use std.datetime.stopwatch.AutoStart.") alias AutoStart = Flag!"autoStart";
+  */
+deprecated("To be removed after 2.940. Use std.datetime.stopwatch.AutoStart instead.")
```
PetarKirov:
```suggestion
deprecated("To be removed after 2.094. Use std.datetime.stopwatch.AutoStart instead.")
```

MoonlightSentinel: Done

MoonlightSentinel:
Style passes now that `std/datetime/package.d` is excluded from `undocumented_declaration_check` (second commit)

The `vibe-d/vibe-core+epoll` failure seems unrelated

MoonlightSentinel:
Ping

MoonlightSentinel:
Ping

MoonlightSentinel:
@wilzbach Removed some sneaky `/// ditto` and `///` leading deprecated unittests

PR #7280, opened 2019-11-16
Remove deprecated symbols sheduled for/after 2.089

MoonlightSentinel:

PR #7283, opened 2019-11-16
Remove deprecated enforceEx

MoonlightSentinel:
Extracted from #7280 and blocked by:

- [x] vibe-d/vibe.d#2388
- [ ] new vibe-d release


n8sh:
This is a fine example of why deprecations of this kind are bad.

[Change Log: 2.080.0](https://dlang.org/changelog/2.080.0.html#std-exception-enforceEx)
>std.exception.enforceEx was deprecated in favor of std.exception.enforce
>With 2.079 std.exception.enforce became a complete super set of std.exception.enforceEx

`enforce` behaves exactly the same as `enforceEx` given the same arguments. The name `std.exception.enforceEx` isn't misleading and there are no plans to reuse it to mean something else. The best approach would have been to write
```
/// Legacy alias for `enforce`.
alias enforceEx = enforce
```
and never speak of it again.

MoonlightSentinel:
I wouldn't mind keeping a legacy alias but it was already scheduled for removal and would probably need some formal approval (akin to adding a new symbol). 

One might argue that introducing this could introduce hidden behaivour changes in highly templated code because `enforce` is a superset of `enforceEx` (allthough this is propably a very hypothetical case)

MoonlightSentinel:
Closing in favour of #7545

PR #7284, opened 2019-11-16
Add automatic number conversion to std.json

GrimMaple:
The JSON standard only defines the "number" type. Therefore it is expected that any whole number would be castable as floating and it doesn't. This little PR changes the behaviour of `floating()` so it will cast any numeric type to double and return it.

### DIFF
--- a/changelog/json_number_conversion.dd
+++ b/changelog/json_number_conversion.dd
@@ -1,0 +1,12 @@
+Added get!(T) getter to std.json
+
+This getter will try to return underlying json type as T if possible.
+It is convenient for automatic integer conversion like this:
+-------
+import std.json;
+string s = `{ "a": 123 }`;
+auto json = parseJSON(s);
+
+// This will throw with json["a"].floating
+assert(json["a"].get!double == 123.0);
+-------
--- a/std/json.d
+++ b/std/json.d
@@ -359,6 +359,90 @@ struct JSONValue
         return type == JSONType.null_;
     }
 
+    /***
+     * Generic type value getter
+     * A convenience getter that returns this `JSONValue` as the specified D type.
+     * Note: only numeric, `bool`, `string`, `JSONValue[string]` and `JSONValue[]` types are accepted
+     * Throws: `JSONException` if `T` cannot hold the contents of this `JSONValue`
+     */
+    @property inout(T) get(T)() inout const pure @safe
+    {
+        static if (is(Unqual!T == string))
+        {
+            return str;
+        }
+        else static if (is(Unqual!T == bool))
+        {
+            return boolean;
+        }
+        else static if (isFloatingPoint!T)
+        {
+            switch (type)
+            {
+            case JSONType.float_:
+                return cast(T) floating;
+            case JSONType.uinteger:
+                return cast(T) uinteger;
+            case JSONType.integer:
+                return cast(T) integer;
+            default:
+                throw new JSONException("JSONValue is not a number type");
+            }
+        }
+        else static if (__traits(isUnsigned, T))
+        {
+            return cast(T) uinteger;
+        }
+        else static if (isSigned!T)
+        {
+            return cast(T) integer;
+        }
+        else
+        {
+            static assert(false, "Unsupported type");
+        }
+    }
+    // This specialization is needed because arrayNoRef requires inout
+    @property inout(T) get(T : JSONValue[])() inout pure @trusted /// ditto
+    {
+        return arrayNoRef;
+    }
+    /// ditto
+    @property inout(T) get(T : JSONValue[string])() inout pure @trusted
+    {
+        return object;
+    }
+    ///
+    @safe unittest
+    {
+        import std.exception;
+        string s =
+        `{
+            "a": 123,
+            "b": 3.1415,
+            "c": "text",
+            "d": true,
+            "e": [1, 2, 3],
+            "f": { "a": 1 }
+         }`;
+
+        struct a { }
+
+        immutable json = parseJSON(s);
+        assert(json["a"].get!double == 123.0);
+        assert(json["a"].get!int == 123);
+        assert(json["b"].get!double == 3.1415);
+        assertThrown(json["b"].get!int);
+        assert(json["c"].get!string == "text");
+        assert(json["d"].get!bool == true);
+        assertNotThrown(json["e"].get!(JSONValue[]));
+        assertNotThrown(json["f"].get!(JSONValue[string]));
+        static assert(!__traits(compiles, json["a"].get!a));
+        assertThrown(json["e"].get!float);
+        assertThrown(json["d"].get!(JSONValue[string]));
+        assertThrown(json["f"].get!(JSONValue[]));
+    }
+
     private void assign(T)(T arg) @safe
     {
         static if (is(T : typeof(null)))


CyberShadow:
Not sure about this for two reasons:

1. Not all `long`s / `ulong`s fit losslessly in a `double`, which means that this opens up the possibility of silent data  corruption / precision loss

2. There are many similar conversions which we could do, and it becomes a slippery slope quickly:
  - If the `type` is `JSONType.float_` but the value is actually an integer, should we allow accessing it via the `.integer` property?
  - If the `type` is `JSONType.null_`, should `.string` return `null` (which is distinct from `""` in D)?
  - If the `type` is `JSONType.string` but the value contains a valid integer, should we convert and return it in `.integer`?

I think any discussion of conversions would need to draw a well-defined line and provide strong rationale for why the line is drawn there and not anywhere else.

std/json.d
```diff
@@ -213,13 +213,15 @@ struct JSONValue
...
      * Throws: `JSONException` for read access if `type` is not
-     * `JSONType.float_`.
+     * `JSONType.float_`, `JSONType.integer` or `JSONType.uinteger`.
      */
     @property double floating() const pure @safe
     {
-        enforce!JSONException(type == JSONType.float_,
-                                "JSONValue is not a floating type");
-        return store.floating;
+        if(type == JSONType.float_) return store.floating;
```
StianGulpen: use  `switch (type) { ... }` here

std/json.d
```diff
@@ -213,13 +213,15 @@ struct JSONValue
...
+        if(type == JSONType.uinteger) return cast(double)store.uinteger;
+        if(type == JSONType.integer) return cast(double)store.integer;
+
+        throw new JSONException("JSONValue is not a floating type");
```
StianGulpen: This is not covered by the tests. You can use `assertThrown` from `std.exception` for that purpose.

n8sh:
@CyberShadow
>Not all `long`s / `ulong`s fit losslessly in a `double`, which means that this opens up the possibility of silent data corruption / precision loss

I consider this acceptable for two reasons:
* The D language and the general family of C/C++/Java/C# allow implicit promotion from integers to floating point numbers even though precision can be lost. Someone writing in one of those languages is more likely to be surprised by a library function that internally forbids integer to floating point conversion than by one that performs it.
* If someone is specifically asking for a floating point number, that implies that the precision of a floating point number is sufficient for his purpose.

That said, the PR could be altered to address your concern. What do you think of this?

```d
if (type == JSONType.uinteger) {
    const result = cast(double) store.uinteger;
    if (cast(ulong) result != store.uinteger) // It would be faster to check `store.uinteger > (1UL << 53)` but this is clearer.
    	throw new JSONException("JSONValue is an integer that cannot be cast to floating point without precision loss");
    return result;
}
if (type == JSONType.integer) {
    const result = cast(double) store.integer;
    if (cast(long) result != store.integer)
    	throw new JSONException("JSONValue is an integer that cannot be cast to floating point without precision loss");
    return result;
}
throw new JSONException("JSONValue is not a floating type");
```

GrimMaple:
@CyberShadow the reason I initially made this PR is the inconvenience that the original behaviour had. In D `123` is `int`, while `123.0` is double. In [JSON (Page 2 Paragraph 5)](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf), both `123` and `123.0` have the same type - `number`, and therefore are expected to be treated as a single type. The `uintiger` and `integer` parts of `std.json` are more of convenience functions rather than having anything to do with JSON standard. If you try to [beautify](https://jsonformatter.org/) a json like this `{ "a": 123.0 }` the beautifier will amend `.0` and just leave you with `123`. Even though you know that it was a float. This leads to the programmer having to introduce wrapping functions to check for all three possible types introduced by D. So to answer 1. `double` is simply one of the closest types that represents JSON `number` (`real` may be an alternative).
To answer 2,

- The value can't be "actually integer", it can only be "number". Therefore `integer` can do whatever it wants as it's not standard for JSON. I think that we shouldn't allow converting `123.0` to `integer`, but we can make it happen if we want.
- `null` is a distinctive type in JSON and therefore is not convertable to `string`. `""` and `null` are different types by JSON standard
- `string` and `number` are different JSON types and therefore are not interchangeable.

std/json.d
GrimMaple: Good idea! Thank you!

std/json.d
GrimMaple: I will add a test case for this exception, thank you!

CyberShadow:
> The D language and the general family of C/C++/Java/C# allow implicit promotion from integers to floating point numbers even though precision can be lost. Someone writing in one of those languages is more likely to be surprised by a library function that internally forbids integer to floating point conversion than by one that performs it.

OK, agreed

> If someone is specifically asking for a floating point number, that implies that the precision of a floating point number is sufficient for his purpose.

No, because they might not have checked the underlying JSON type beforehand, and might have assumed that the type was already a float. An exception would then be the desirable behavior.



> the reason I initially made this PR is the inconvenience that the original behaviour had. In D `123` is `int`, while `123.0` is double. In [JSON (Page 2 Paragraph 5)](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf), both `123` and `123.0` have the same type - `number`, and therefore are expected to be treated as a single type. The `uintiger` and `integer` parts of `std.json` are more of convenience functions rather than having anything to do with JSON standard. If you try to [beautify](https://jsonformatter.org/) a json like this `{ "a": 123.0 }` the beautifier will amend `.0` and just leave you with `123`. Even though you know that it was a float. This leads to the programmer having to introduce wrapping functions to check for all three possible types introduced by D.

Agreed

> So to answer 1. `double` is simply one of the closest types that represents JSON `number` (`real` may be an alternative).

No, the closest type would be a string which returns the number as it appears in the JSON. Then, the user could convert it to a binary number of the desired precision, incl. `BigInt` / `CustomFloat`.

What do you think about introducing a `.number` property which returns a string as described above, and making `.integer` etc. work on top of that?

GrimMaple:
As far as I understand how json works in D, the current implementation is far from ready for such a change. It would possibly require some discussion first as it would change the way parsing is done, and it would change the way JSON stores data. As of now, during the parse state, the value stored in a union. What you're suggesting would change the underlying storage and parsing, and that would take hundreds of lines.

I also have storage / speed concerns. Is it really a good idea to always store an extra string when we can store a double instead? Just allowing to convert integers to double seems like the most convenient way without redoing half of the code.

CyberShadow:
> As far as I understand how json works in D, the current implementation is far from ready for such a change. It would possibly require some discussion first as it would change the way parsing is done, and it would change the way JSON stores data. As of now, during the parse state, the value stored in a union. What you're suggesting would change the underlying storage and parsing, and that would take hundreds of lines.

Sure.

> I also have storage / speed concerns. Is it really a good idea to always store an extra string when we can store a double instead?

It could be a net win, as it would allow deferring parsing until the field is accessed, and would avoid re-stringification when editing only some fields in a big JSON structure.

> Just allowing to convert integers to double seems like the most convenient way without redoing half of the code.

Here's something I noticed: there is a templated setter - the `assign` method - which sets the `type` according to the type of the argument, but there is no equivalent getter. I think there should be a `get!T` method which performs the inverse of `assign` (not `opCast` or `to` as that would conflict with the existing `toString`). That would then be the place to perform any harmless implicit conversions.

std/json.d
```diff
@@ -213,13 +213,21 @@ struct JSONValue
...
     {
-        enforce!JSONException(type == JSONType.float_,
-                                "JSONValue is not a floating type");
-        return store.floating;
+        switch(type)
+        {
+        case JSONType.float_:
+            return store.floating;
+        case JSONType.uinteger:
+            return cast(double)store.uinteger;
```
CyberShadow: I suggest removing this cast to emphasize that this is an implicit conversion according to D rules, per your argument above.

std/json.d
```diff
@@ -213,13 +213,21 @@ struct JSONValue
      * Value getter/setter for `JSONType.float_`. Note that despite
      * the name, this is a $(B 64)-bit `double`, not a 32-bit `float`.
      * Throws: `JSONException` for read access if `type` is not
-     * `JSONType.float_`.
+     * `JSONType.float_`, `JSONType.integer` or `JSONType.uinteger`.
```
CyberShadow: I think this should be more explicit.

```suggestion
     * `JSONType.float_`.
     * For convenience, if `type` is `JSONType.integer` or `JSONType.uinteger`,
     * the integer value is silently converted to a `double` and returned.
```

std/json.d
```diff
@@ -213,13 +213,21 @@ struct JSONValue
...
+        switch(type)
+        {
+        case JSONType.float_:
+            return store.floating;
+        case JSONType.uinteger:
+            return cast(double)store.uinteger;
+        case JSONType.integer:
+            return cast(double)store.integer;
+        default:
+            throw new JSONException("JSONValue is not a floating type");
```
CyberShadow:
```suggestion
            throw new JSONException("JSONValue is not a numeric type");
```

CyberShadow:
Please add a changelog entry describing the change, or file a bug and mention its number in the commit message.

GrimMaple:
Just a quick reply before I start changing anything.

> Here's something I noticed: there is a templated setter - the assign method - which sets the type according to the type of the argument, but there is no equivalent getter. I think there should be a get!T method which performs the inverse of assign (not opCast or to as that would conflict with the existing toString). That would then be the place to perform any harmless implicit conversions.

I think that it's a good idea, though I'm not sure if it's a good idea to implement this change as of now. First of all it's not really straight forward and requries some thinking about, eg shall we allow classes and structs for `T`, what kinds of conversions are legal etc. I could potentially rewrite my PR by reverting the change in `floating()` and adding a `get(T)()` with a simple switch where I will put my code for floating, but I'll wait for you to decide if I shall keep the current code or change it.

CyberShadow:
> First of all it's not really straight forward and requries some thinking about, eg shall we allow classes and structs for `T`, what kinds of conversions are legal etc.

Since unsupported types will result in an error, it would be safe to start from the minimal set of supported `T` / `JSONType` combinations. Later it can be expanded if we decide that it makes sense to do so.

> I could potentially rewrite my PR by reverting the change in `floating()` and adding a `get(T)()` with a simple switch where I will put my code for floating, but I'll wait for you to decide if I shall keep the current code or change it.

In absence of other arguments, I would suggest this approach.

CyberShadow:
CC @atilaneves for approval of addition of new symbol.

std/json.d
```diff
@@ -359,6 +359,80 @@ struct JSONValue
...
+            "d": true,
+            "e": [1, 2, 3],
+            "f": { "a": 1 }
+         }`;
+
+        immutable auto json = parseJSON(s);
+        assert(json["a"].get!double == 123.0);
+        assert(json["a"].get!int == 123);
+        assert(json["b"].get!double == 3.1415);
+        assert(json["b"].get!int == 3);
```
CyberShadow: No, this should throw

std/json.d
```diff
@@ -359,6 +359,80 @@ struct JSONValue
...
+        assert(json["c"].get!string == "text");
+        assert(json["d"].get!bool == true);
+        assertNotThrown(json["e"].get!(JSONValue[]));
+        assertNotThrown(json["f"].get!(JSONValue[string]));
+        assertThrown(json["a"].get!JSONException);
```
CyberShadow: Why `get!JSONException`?

std/json.d
```diff
@@ -359,6 +359,80 @@ struct JSONValue
...
+        assertThrown(json["e"].get!string);
+        assertThrown(json["d"].get!double);
```
CyberShadow: Ironically `double d = true` works in D, but probably we should avoid perpetuating that mistake.

std/json.d
```diff
@@ -359,6 +359,80 @@ struct JSONValue
         return type == JSONType.null_;
     }
 
+    /***
+     * Generic type value getter
+     * A convenience getter that will try to convert the underlaying data to the given type.
```
atilaneves: The name `get` doesn't suggest any conversion taking place. I'd suggest renaming it.

GrimMaple: Would `getAs!T` be more intuitive?

n8sh: With a name like `getAs` and verbiage like `* A convenience getter that will try to convert the underlaying data to the given type`, I would expect it to be trying to do things like converting strings to numbers or numbers to booleans. It's a bit misleading if the function is merely a unified way to read a number but does not allow any kind of conversion between different JSON types.

CyberShadow: We'd need a good rationale for why *conversion* between types should be in `std.json` while we have a dedicated facility for this purpose (`std.conv`).

I suggest keeping the name `get` and changing the description to e.g. "Returns the `JSONValue` contents as the specified D type `T`".

CyberShadow: We'd need a good rationale for why *conversion* between types should be in `std.json` while we have a dedicated facility for this purpose (`std.conv`).

I suggest keeping the name `get` and changing the description to e.g. "Returns the `JSONValue` contents as the specified D type `T`".

atilaneves: After thinking about it some more, `get` should be ok.

GrimMaple: I will change the comment as CyberShadow suggested nontheless

std/json.d
GrimMaple: I initially thought to show that it will not work for custom types, but I'll create a temporary `struct` just for that

std/json.d
GrimMaple: OK, I will change it to throw

std/json.d
GrimMaple: OK, I will change that. Thank you!

CyberShadow: Sorry for being vague. I think no change is required here.

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
+     * A convenience getter that returns `JSONValue` as the specified D type.
+     * Note: only arithmetic, `bool`, `string`, `JSONValue` and `JSONValue[]` types are accepted
+     * Throws: `JSONException` if `T` is not a valid type or conversion is impossible
+     */
+    @property inout(T) get(T)() inout const pure @safe
+    {
+        static if (is(T : string))
+        {
+            return str;
+        }
```
CyberShadow: Nit, single-statement blocks waste vertical space.

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
         return type == JSONType.null_;
     }
 
+    /***
+     * Generic type value getter
+     * A convenience getter that returns `JSONValue` as the specified D type.
```
CyberShadow:
```suggestion
     * A convenience getter that returns this `JSONValue` as the specified D type.
```

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
+     * Note: only arithmetic, `bool`, `string`, `JSONValue` and `JSONValue[]` types are accepted
+     * Throws: `JSONException` if `T` is not a valid type or conversion is impossible
```
CyberShadow: More avoiding using the "conversion" term: (Also, in some cases, conversion is "possible", we just refuse to do it implicitly)
```suggestion
     * Throws: `JSONException` if `T` cannot hold the contents of this `JSONValue`
```

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
+        {
+            return cast(T)integer;
+        }
+        else
+        {
+            throw new JSONException("Unsupported type");
+        }
+    }
+    // This specialization is needed because arrayNoRef requires inout
+    @property inout(T) get(T : JSONValue[])() inout pure @trusted
```
CyberShadow:
```suggestion
    @property inout(T) get(T : JSONValue[])() inout pure @trusted /// ditto
```

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
+    {
+        return arrayNoRef;
+    }
+    // ditto
```
CyberShadow:
```suggestion
    /// ditto
```

CyberShadow:
Also `changelog/json_number_conversion.dd` is currently an empty file.

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
+            "b": 3.1415,
+            "c": "text",
+            "d": true,
+            "e": [1, 2, 3],
+            "f": { "a": 1 }
+         }`;
+
+        struct a { }
+
+        immutable auto json = parseJSON(s);
```
PetarKirov:
```suggestion
        immutable json = parseJSON(s);
```

In D you don't need to use `auto` for type inference - any storage class / type qualifier will work.

GrimMaple: Good idea, thank you!

std/json.d
GrimMaple: If you don't mind, I will leave this as is for style consistency across this file. Many functions (eg `assign`, `opBinary`, `opAssign`, `toValue`) have multiple one-line ifs implemented in similar way. I'm okay with changing this if you insist though.

GrimMaple:
I made a few styling changes to hopefully pass style checks

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
+                return cast(T) integer;
+            default:
+                throw new JSONException("JSONValue is not a number type");
+            }
+        }
+        else static if (__traits(isUnsigned, T))
+        {
+            return cast(T) uinteger;
+        }
+        else static if (__traits(isArithmetic, T))
```
PetarKirov:
```suggestion
        else static if (isSigned!T)
```

Use `std.traits.isSigned` as `__traits(isArithmetic, T)` would include pointer types, which we shouldn't allow here.

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
         return type == JSONType.null_;
     }
 
+    /***
+     * Generic type value getter
+     * A convenience getter that returns this `JSONValue` as the specified D type.
+     * Note: only arithmetic, `bool`, `string`, `JSONValue` and `JSONValue[]` types are accepted
```
PetarKirov:
```suggestion
     * Note: only numeric, `bool`, `string`, `JSONValue` and `JSONValue[]` types are accepted
```

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
+    {
+        static if (is(T : string))
+        {
+            return str;
+        }
+        else static if (is(T : bool))
+        {
+            return boolean;
+        }
+        else static if (isFloatingPoint!T)
```
PetarKirov: Off topic for this PR: I wonder what would be the best way to support user-defined types, such as bigint, decimal or checkedint, without whitelisting them one by one here.

changelog/json_number_conversion.dd
```diff
@@ -0,0 +1,12 @@
+Add get!(T) getter to std.json
```
MoonlightSentinel:
```suggestion
Added get!(T) getter to std.json
```

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
 
+    /***
+     * Generic type value getter
+     * A convenience getter that returns this `JSONValue` as the specified D type.
+     * Note: only arithmetic, `bool`, `string`, `JSONValue` and `JSONValue[]` types are accepted
+     * Throws: `JSONException` if `T` cannot hold the contents of this `JSONValue`
+     */
+    @property inout(T) get(T)() inout const pure @safe
+    {
+        static if (is(T : string))
```
MoonlightSentinel: This `is`-expression requires that T is convertible to a string, not that the stores string can be (implicitly) coverted into a string. This would e.g. match the following `S`:

```
struct S {
	string toString() { return ""; }
	alias toString this;
}
```

This should either require an exact match `is(T == string)` or `is(string : T)` to support user defined types

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
+        {
+            return str;
+        }
+        else static if (is(T : bool))
```
MoonlightSentinel: Ditto, but needs to be reordered after integers and floats
```suggestion
        else static if (is(bool : T))
```

std/json.d
GrimMaple: Good catch! Thank you!

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
+        {
+            return cast(T) uinteger;
+        }
+        else static if (__traits(isArithmetic, T))
+        {
+            return cast(T) integer;
+        }
+        else
+        {
+            throw new JSONException("Unsupported type");
```
MoonlightSentinel: This is known at comile time
```suggestion
            static assert(false, "Unsupported type");
```

std/json.d
GrimMaple: If you ask me, it seems possible to make something like `return T(floating);` but I'm not entirely sure if it's a good solution both in style and performance. It would properly initialize the T with value if that is possible though.

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
+     * Generic type value getter
+     * A convenience getter that returns this `JSONValue` as the specified D type.
+     * Note: only arithmetic, `bool`, `string`, `JSONValue` and `JSONValue[]` types are accepted
+     * Throws: `JSONException` if `T` cannot hold the contents of this `JSONValue`
+     */
+    @property inout(T) get(T)() inout const pure @safe
+    {
+        static if (is(T : string))
+        {
+            return str;
```
MoonlightSentinel: If we want to implicitly support user defined types:

```suggestion
            return cast(T) str;
```

GrimMaple: As far as I understand everyone in this trhead we don't want to make tricky conversions. CC @CyberShadow for clearance.

MoonlightSentinel: Fine by me, just a suggestion based on @PetarKirov's comment

> Off topic for this PR: I wonder what would be the best way to support user-defined types, such as bigint, decimal or checkedint, without whitelisting them one by one here.

GrimMaple: Gotcha

CyberShadow: Reading a JSON string, parsing it into a fixed-size number (`double` or `long`), then converting it to a `BigInt` is completely meaningless, and providing such a conversion path would be disingenious. We would have to store and convert the string to `BigInt` directly as per the discussion above. @PetarKirov

GrimMaple:
@CyberShadow ping. It seems that everything is addressed. Anything else needing work?

std/json.d
```diff
@@ -359,6 +359,90 @@ struct JSONValue
...
+        immutable json = parseJSON(s);
+        assert(json["a"].get!double == 123.0);
+        assert(json["a"].get!int == 123);
+        assert(json["b"].get!double == 3.1415);
+        assertThrown(json["b"].get!int);
+        assert(json["c"].get!string == "text");
+        assert(json["d"].get!bool == true);
+        assertNotThrown(json["e"].get!(JSONValue[]));
+        assertNotThrown(json["f"].get!(JSONValue[string]));
+        // assertThrown(json["a"].get!a); // This will trigger static assert
```
CyberShadow:
```suggestion
        static assert(!__traits(compiles, json["a"].get!a));
```

berni44:
Is this a fix of [issue 16432](https://issues.dlang.org/show_bug.cgi?id=16432)?

GrimMaple:
@berni44 no, this fixes completely different issue :)

GrimMaple:
@CyberShadow ping. Is there something else I need to do?

CyberShadow:
I already approved this pull request.

PR #7285, opened 2019-11-17
Partial replace call to snprintf for '%a' and float or double.

berni44:
This is a followup on #7264. This time it implements %a and %A specifiers. Part of this is identical to #7264 to make this PR independed.

I run essentially the same tests as in #7264. 

Concerning correctness: All tests passed. There where two exceptions:
a) There is a bug for denormalized floats in snprintf. The new version fixes this.
b) Ties for roundToNearest are different, like in #7264.

Concerning speed: dmd is a little bit slower with the new version, while ldc2 is slightly faster. See [pdf-file](https://github.com/berni44/printFloat/blob/master/printHex.pdf) for some diagrams.


std/format.d
```diff
@@ -7017,3 +7063,538 @@ char[] sformat(Char, Args...)(return scope char[] buf, scope const(Char)[] fmt,
...
+            // hack to work around issue 20363
+            ival ^= rm;
+            ival ^= rm;
+        }
+
+        ulong mnt = ival & ((1L << 52) - 1);
+        int exp = (ival >> 52) & ((1L << 11) - 1);
+        int maxexp = 2047;
+        string sgn = (ival >> (52 + 11)) & 1 ? "-" : "";
+        enum long bias = 1023;
```
thewilsonator: You should be able to deduplicate this branch (modulo the DigitalMars hack) by using the floating point's mantissa and exponent properties, something like
```
ulong mnt = ival & ((1L << T.mant_dig) - 1);
int exp = (ival >> T.mant_dig) & ((1L << (T.sizeof * 8) - 1 -T.mant_dig) - 1);
enum maxexp = T.max_exp;
string sgn = (ival >> ((T.sizeof * 8) - 1)) & 1 ? "-" : "";
enum long bias = (maxexp+1)>>1;
```

std/format.d
```diff
@@ -7017,3 +7063,538 @@ char[] sformat(Char, Args...)(return scope char[] buf, scope const(Char)[] fmt,
...
+            result[offset + 0] = is_upper?'I':'i';
+            result[offset + 1] = is_upper?'N':'n';
+            result[offset + 2] = is_upper?'F':'f';
+        }
+        else
+        {
+            result[offset + 0] = is_upper?'N':'n';
+            result[offset + 1] = is_upper?'A':'a';
+            result[offset + 2] = is_upper?'N':'n';
+        }
```
thewilsonator: please write this as 
```
result[offset .. offset +3] = (mnt == 0) ? ( is_upper ? "INF" : "inf" )
                                                                  : ( is_upper ? "NAN" : "nan" )
```

std/format.d
```diff
@@ -2565,7 +2566,7 @@ private enum ctfpMessage = "Cannot format floating point types at compile-time";
 /*
     Floating-point values are formatted like $(REF printf, core, stdc, stdio)
  */
-private void formatValueImpl(Writer, T, Char)(auto ref Writer w, T obj, scope const ref FormatSpec!Char f)
+private void formatValueImpl(Writer, T, Char)(auto ref Writer w, T obj, scope const ref FormatSpec!Char f) @trusted
```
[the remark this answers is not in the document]
berni44: I actually don't know, what triggers the need for `@trusted` here. `printFormat` itself is `@safe` (I forgot to add the keyword, but it is; btw it's also pure), `FloatingPointControl.rounding` is `@trusted` and the rest is just a copy of code that worked without `@trusted`. When all specifiers are implemented (im close to finish the work on %e) the duplicated code is not needed any more and I hope, that the need for `@trusted` will vanish too or at least it will make it easier to hunt for the place, which triggers the need.

std/format.d
```diff
@@ -2628,55 +2632,118 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     const nan = isNaN(tval);
     const inf = isInfinity(tval);
 
-    if (nan || inf)
+    static if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.mant_dig))
     {
```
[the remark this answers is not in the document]
berni44: I first thought to do this, but it's a little bit difficult, because of a mix of `static if` and `if`. I would probably need to put everything in a separate function. But as this is only temporarily needed (until all specifiers are implemented) I think it's better to keep it simple here.

std/format.d
berni44: I'll try, to do so. :-)

std/format.d
berni44: When I remember right, it were speed considerations. But I don't know, if printing NaNs need to be fast...

thewilsonator: Indeed. Which compiler did you use? I could imagine DMD doing a terrible job optimising that but LDC and GDC should have no problem.

berni44:
As I mentioned above, I'm almost finished with the work on '%e' - it works allready for all floats, but the code needs some refactorisation, I did not check doubles and probably I have to do something for speedup; I think I can do the final tests next weekend.

Now I'm not sure if we want to do all four specifiers (%g should be easy once %f and %e works) separate or all in one PR. What do you prefere?

std/format.d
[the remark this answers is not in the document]
thewilsonator: You should be able to figure out where the problem is pretty quick by removing it and seeing where the compiler complains.

berni44:
One more thing: %f currently does not pass one test, which is due to a different rounding tie. I meanwhile thought, that it might be easier to first use the old tie (so all tests pass) and then address this in a separate PR. In this case I should change it here too for consistency. What do you think?

thewilsonator:
> and then address this in a separate PR.

Yes please.

> In this case I should change it here too for consistency

I'm not sure I follow (I have a cold that the moment so its probably that), but do whatever you think is best.

std/format.d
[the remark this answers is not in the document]
thewilsonator: > keep it simple here.

There is a difference between simple to write and simple to read. In this case I am trying to figure out what the difference between the two branches are

For this case 
```
size_t len;
char[] buf;
if (fs.spec=='a' || fs.spec=='A')
{
    static if(...)
    {
        buf = printFloat(val, fs, FloatingPointControl.rounding).dup;
        len = buf.length;
    }
    else
        goto notFloating;
}
else
{
    notFloating:
    ...
}
```
will work.

std/format.d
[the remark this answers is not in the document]
berni44: All I get is:

```
std/conv.d(1262): Error: @safe function std.conv.__unittest_L1259_C7 cannot call @system function std.conv.to!string.to!(double[2]).to
std/conv.d(226):        std.conv.to!string.to!(double[2]).to is declared here
```

And a dozen of similar messages, but nothing usefull.

std/format.d
```diff
@@ -2628,55 +2632,118 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
...
+                                // negative precision is same as no precision specified
+                                fs.precision == fs.UNSPECIFIED ? -1 : fs.precision,
+                                tval);
+            }();
+
+            enforceFmt(n >= 0,
+                       "floating point formatting failure");
+
+            len = min(n, buf_tmp.length-1);
+            buf = buf_tmp;
```
thewilsonator: Not that it seems used later but assigning to a stack variable that is about to go out of scope is a bad idea. Also this is probably the reason for your `@safe`ty problem.

std/format.d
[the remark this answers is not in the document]
thewilsonator: If https://github.com/dlang/phobos/pull/7285#pullrequestreview-326656921 is not the cause then mark `formatValueImpl` `@safe` explicitly then the compiler should complain where in `formatValueImpl` is the problem.

std/format.d
berni44: It's too long ago, I do not remember. I did a lot of checks with dmd and ldc2 with and without optimization. From different tests it seems, that `result[a..b] = something` is generally slower than expected, even with optimization.

std/format.d
thewilsonator: Thanks, its much easier knowing what the magic numbers mean.

berni44:
> I'm not sure I follow (I have a cold that the moment so its probably that), but do whatever you think is best.

Oh, my best wishes for a speedy recovery! :-)

std/format.d
berni44: It's a workaround until snprintf is removed completely. After that this is not needed anymore. I'll see if I find a way to do this better.

std/format.d
thewilsonator: Hmm, I would have thought that the LDC simplify druntime pass would have dealt with that fine. maybe the conditionals throw it off? 
Still, I don't think that too many people's programs will be slowed down by writing NaNs, and if for some reason they are they can use LTO with druntime which should definitely work.

std/format.d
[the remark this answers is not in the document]
berni44: This indeed points to the `buf = buftmp` stuff. I added `.dup` now and it works. But I still cannot declare `formatValueImpl` `@safe`. It tells me, that a `@safe`-function cannot call `std.range.primitives.put`. (I remember, that I got this some weeks ago too and it was the reason why I put that `@trusted` there, because it was in code, that I even did not touch.) 

Seems to be a little bit weird, like if there is some other (undocumented or somehow implicit) state than `@safe`, `@trusted` and `@system`. Could this be a compiler bug? Or is it something I don't know - maybe if one template is `@safe`, all other versions are assumed `@safe` too or something like this?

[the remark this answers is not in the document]
thewilsonator: IDK, I'm not up to figuring that out now. Might be a good idea to ask on the forum or he slack channel.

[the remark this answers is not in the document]
berni44: Well, here I think we are done. With `.dup` I can remove `@trusted`. All else is an other matter that does not belong here. When I'll find the time I'll ask about this in the forum.

std/format.d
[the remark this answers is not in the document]
berni44: Nice trick with that goto. I tried it and it works.

std/format.d
berni44: I meanwhile ran all variants though the compilers and there is little difference.

thewilsonator: Good to see.

std/format.d
berni44: Changed this too.

berni44:
I changed everything you requested locally. Before uploading it, I'll rerun the external tests to make sure I didn't break anything. This will take some time.

thewilsonator:
Ping me when you're good to go.

berni44:
@thewilsonator All external tests passed. :-)

wilzbach:
> Concerning speed: dmd is a little bit slower with the new version, while ldc2 is slightly faster. See pdf-file for some diagrams.

FYI: You NEVER need to worry about DMD when doing perf analysis for druntime or phobos. DMD is absolutely slow and years ago it was decided that it's a massive waste of time to bother about it ;-)
Any serious application will use LDC anyhow (less bugs, more architectures, better perf, ...).

std/format.d
```diff
@@ -62,6 +62,7 @@ import std.exception;
 import std.meta;
 import std.range.primitives;
 import std.traits;
+import std.math : FloatingPointControl;
```
CyberShadow: Perhaps this top-level import is the cause.

Would it be possible to move it to a local import, by splitting the definition of the `printFloat` function into an eponymous template?

berni44: I'll take care. 

Probably I'll use a local enum for rounding modes instead of the solution with eponymous templates. That has the advantage, that I can add the fifth mode that IEEE allows for printing floats, which `FloatingPointControl` does not know about. (I plan a PR to make formatting floating point numbers pure by using this fifth mode for format and add a separate (nonpure) function, probably formatIEEE for rounding according to the processor state. I will need this enum there anyway.)

Shall I base this PR on `stable`, as it's some sort of regression?

n8sh: [Reading the floating point rounding mode is explicitly allowed in `pure` code:](https://dlang.org/spec/function#pure-functions)

>As a concession to practicality, a pure function can also:
>* read and write the floating point exception flags
>* read and write the floating point mode flags, as long as those flags are restored to their initial state upon function entry
>* perform impure operations in statements that are in a _ConditionalStatement_ controlled by a _DebugCondition_.

PR #7287, opened 2019-11-18
Replace throw exception by enforce and add unittests. 

berni44:
I went through all exception handling, added a unittest where possible, removed throw statements that could never fire and replaced the remaining throws by calling enforce.

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -69,6 +69,8 @@ void main()
  */
 module std.zip;
 
+import std.exception : enforce;
+
 //debug=print;
 
 /** Thrown on error.
@@ -247,8 +249,7 @@ final class ArchiveMember
     {
         if (cm == _compressionMethod) return;
 
-        if (_compressedSize > 0)
-            throw new ZipException("Can't change compression method for a compressed element");
+        enforce!ZipException(_compressedSize == 0, "Can't change compression method for a compressed element");
 
         _compressionMethod = cm;
     }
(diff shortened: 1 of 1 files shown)


std/zip.d
```diff
@@ -862,12 +1070,55 @@ public:
                 import std.zlib : uncompress;
                 de._expandedData = cast(ubyte[]) uncompress(cast(void[]) de.compressedData, de.expandedSize, -15);
                 return de.expandedData;
-
-            default:
-                throw new ZipException("unsupported compression method");
```
CyberShadow: I think this will now throw an `Error` instead of `Exception`, so this change is wrong.

Also not sure about the other similar change for similar reasons.

berni44: Good point. I'll add a unittest here. When I remember right (I started preparing this about a month ago), I did not manage to cover the other place (but I'll try again) and therefore changed to the final switch. Here I just thought, it's the same, but that is not true.

std/zip.d
```diff
@@ -672,14 +718,8 @@ public:
             _numEntries = getUshort(i + 8);
             _totalEntries = getUshort(i + 10);
 
-            if (numEntries != totalEntries)
```
MoonlightSentinel: Why is this check (and multiple other checks thorughout the file) removed?

berni44: That are checks, that can never happen, because they are allready filtered by findEndOfCentralDirRecord or by the calls to removeSegment.

MoonlightSentinel: Okay

PR #7288, opened 2019-11-18
Clarify the behaviour of rename on posix systems.

berni44:
This is due to issue [11631](https://issues.dlang.org/show_bug.cgi?id=11631). I checked, that the issue still is valid.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -886,11 +886,20 @@ version (Windows) private void writeImpl(scope const(char)[] name, scope const(F
 }
 
 /***************************************************
- * Rename file `from` _to `to`.
+ * Rename file `from` _to `to`, moving it between directories if required.
  * If the target file exists, it is overwritten.
+ *
+ * It is not possible to rename a file across different mount points
+ * or drives. On POSIX, the operation is atomic. That means, if `to`
+ * already exists there will be no time period during the operation
+ * where `to` is missing. See
+ * $(HTTP man7.org/linux/man-pages/man2/rename.2.html, manpage for rename)
+ * for more details.
+ *
  * Params:
  *    from = string or range of characters representing the existing file name
  *    to = string or range of characters representing the target file name
+ *
  * Throws: $(LREF FileException) on error.
  */
 void rename(RF, RT)(RF from, RT to)


CyberShadow:
The bug is invalid.

CyberShadow:
Feel free to convert this to a clarification that this does a non-copying rename only.

std/file.d
```diff
@@ -886,11 +886,20 @@ version (Windows) private void writeImpl(scope const(char)[] name, scope const(F
 }
 
 /***************************************************
- * Rename file `from` _to `to`.
+ * Rename file `from` _to `to`, moving it between directories if required.
  * If the target file exists, it is overwritten.
+ *
+ * On Posix, the operation is atomic, which means, that it is not possible to
+ * rename a file across different mount points. If `to` already exists, it
```
CyberShadow: AFAIK it's not possible to rename a file across volumes on Windows either. Generally this means across drives, but with the introduction of reparse points, any two paths may conceivably fail, as on POSIX.

std/file.d
```diff
@@ -886,11 +886,20 @@ version (Windows) private void writeImpl(scope const(char)[] name, scope const(F
 }
 
 /***************************************************
- * Rename file `from` _to `to`.
+ * Rename file `from` _to `to`, moving it between directories if required.
  * If the target file exists, it is overwritten.
+ *
+ * On Posix, the operation is atomic, which means, that it is not possible to
```
CyberShadow: POSIX is "POSIX", not "Posix". `Posix` is just what the `version` label is called for conformity with C.

std/file.d
CyberShadow: `rename` is part of C, and is specified to be atomic in the standard, so probably none of this is actually specific to POSIX other than the "mount points" term.

std/file.d
berni44: Most of the time, it's called "Posix" in the docs of this file.

CyberShadow: That should be fixed too.

std/file.d
berni44: I have little experience with Windows. Changed this now.

std/file.d
```diff
@@ -886,11 +886,20 @@ version (Windows) private void writeImpl(scope const(char)[] name, scope const(F
...
- * Rename file `from` _to `to`.
+ * Rename file `from` _to `to`, moving it between directories if required.
  * If the target file exists, it is overwritten.
+ *
+ * The operation is atomic, which means, that it is not possible to
+ * rename a file across different mount points or drives. If `to` already exists, it
+ * will be replaced atomically, which means, that during the operation, there
+ * is no time period, when `to` is missing. See
+ * $(HTTP man7.org/linux/man-pages/man2/rename.2.html, manpage for rename)
+ * for more details.
```
CyberShadow: English is my third language but it seems to me that there are too many commas here.

Specifically the ones in "which means, that" and "time period, when" sound redundant to me.

CyberShadow: Nit: s/Clearify/Clarify/ in commit message

std/file.d
berni44: I'll do this in a sperate PR.

std/file.d
berni44: Commas are not my strength... :-)

CyberShadow:
Oof, just noticed that the Windows implementation uses `MoveFileEx` and not `rename`, which is *not* atomic. So please change the wording to reflect that the function is atomic only on POSIX. `MoveFileEx` (in the way we use it) still does not allow moving across volumes though.

CyberShadow:
Thanks! :)

PR #7293, opened 2019-11-21
Fix Issue 20411 - wrong default datetime when saving files in zip file

berni44:
This is a prerequisit for fixing issue 20292, because else some unittests will break (the created zip files are rejected when read in again...)

### DIFF
--- a/std/zip.d
+++ b/std/zip.d
@@ -135,7 +135,13 @@ enum CompressionMethod : ushort
 final class ArchiveMember
 {
     import std.conv : to, octal;
-    import std.datetime.systime : DosFileTime, SysTime, SysTimeToDosFileTime;
+    import std.datetime.systime : DosFileTime, SysTime, SysTimeToDosFileTime, Clock;
+
+    /// Create a new ArchiveMember
+    this() @safe
+    {
+        _time = SysTimeToDosFileTime(Clock.currTime());
+    }
 
     /**
      * The name of the archive member; it is used to index the
@@ -389,7 +395,7 @@ final class ArchiveMember
     }
 }
 
-@safe pure unittest
+@safe unittest
 {
     import std.exception : assertThrown, assertNotThrown;
 
@@ -1532,6 +1538,9 @@ debug(print)
         assert(member.index == i, "member " ~ name ~ " had index " ~
                 member.index.to!string ~ " but we expected index " ~ i.to!string ~
                 ". The input array was " ~ names.to!string);
+
+        // issue 20411
+        assert(cast(uint) member._time != 0);
     }
 }
 


CyberShadow:
Not sure about this one:

1. Hidden information leak
2. Purity regression
3. Programs which expected created ZIP files to be reproducible may regress.

berni44:
>     1. Hidden information leak

What do you mean with this? I don't understand.

>     2. Purity regression

Yeah, but I think, that cannot be helped. Getting system time cannot be pure. The alternative would be to make ZipArchive.build() nonpure and remove some more attributes there. Looked more problematic to me.

>     3. Programs which expected created ZIP files to be reproducible may regress.

Can one really expect zip files to be always identical?

```
$> zip test1.zip a
  adding: a (stored 0%)
$> zip test2.zip a
  adding: a (stored 0%)
$> md5sum *.zip
e3d4e5807755856fdd3027798b385d0b  test1.zip
572fc6cd340391facbec00edb01136a0  test2.zip
```
(they differ in these mysterious extra fields)

But anyway I think that fixing a bug is more important.

CyberShadow:
> What do you mean with this? I don't understand.

Applications may be written with the expectation that the time of the archive creation is not included in the archive itself. Including the time of archive creation leaks information, and could be used to e.g. greatly improve the likelihood of guessing the RNG seed used to initialize `rndGen`, and then predict the random values that the program generated (such as security tokens).

> Can one really expect zip files to be always identical?

The error lies within the expectation that a library for creating ZIP files programmatically behaves in a similar way as a tool primarily meant to be used by humans.

Generally speaking, yes, reproducible outputs should be the default, as the growing trend for pure programming languages / functional-defined systems / reproducible builds shows.

CyberShadow:
> Getting system time cannot be pure.

Ideally there would be a function or constructor for adding an archive member, which could then take an argument for the member date. `Clock.currTime` could the default value for this parameter (if we don't consider backwards compatibility or the information leak). Then, specifying a constant value as the argument would make it pure.

IMHO it is fortunate that 0 is an invalid value, as then it can be unambiguously interpreted to mean "no date specified", which is exactly what happens here, so no changes in `std.zip` are needed.

berni44:
> IMHO it is fortunate that 0 is an invalid value, as then it can be unambiguously interpreted to mean "no date specified", which is exactly what happens here, so no changes in `std.zip` are needed.

Convinced. I havn't found any specs that forbid 0 as a value. It's just not a valid date. I need to be a little more coutious when implementing the validation of date/time issues though. With that I'll also update the docs to make clear, that no time is saved, when no one is specified.

PR #7294, opened 2019-11-23
prevent cases of illegal implicit convertions in CaseStatement

StianGulpen:
DMD allows some illegal expressions in the CaseStatments , which might be fixed by https://github.com/dlang/dmd/pull/10603. This PR fix the cases detected in Phobos.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -7753,12 +7753,20 @@ real fma(real x, real y, real z) @safe pure nothrow @nogc { return (x * y) + z; 
 Unqual!F pow(F, G)(F x, G n) @nogc @trusted pure nothrow
 if (isFloatingPoint!(F) && isIntegral!(G))
 {
-    import std.traits : Unsigned;
+    import std.traits : Unsigned, isSigned;
     real p = 1.0, v = void;
     Unsigned!(Unqual!G) m = n;
+
+    static if (G.sizeof < 4 && !isSigned!G)
+    {
+        import std.conv : to;
+        const int n2 = n.to!int;
+    }
+    else alias n2 = n;
+
     if (n < 0)
     {
-        switch (n)
+        switch (n2)
         {
         case -1:
             return 1 / x;


std/math.d
```diff
@@ -7774,11 +7774,11 @@ if (isFloatingPoint!(F) && isIntegral!(G))
     {
         switch (n)
         {
-        case 0:
+        case cast(G) 0:
             return 1.0;
-        case 1:
+        case cast(G) 1:
             return x;
-        case 2:
```
thewilsonator: Are these needed?

std/math.d
```diff
@@ -7760,9 +7760,9 @@ if (isFloatingPoint!(F) && isIntegral!(G))
     {
         switch (n)
         {
-        case -1:
+        case cast(G) -1:
             return 1 / x;
-        case -2:
+        case cast(G) -2:
             return 1 / (x * x);
```
thewilsonator: I think it would be better to write this as 
```
 if (isSigned!G && n < 0)
 {
         switch (cast(SignedOf!G)n)
```

That way the meaning of the label remain uncluttered.

std/math.d
StianGulpen: indeed they don't. I have used a wrong version of DMD to test. Thanks for pointing this to me.

std/math.d
StianGulpen: That SignedOf thing does not appear to exist or be public to me

thewilsonator: I thought there would be something like that in `std.traits`

StianGulpen:
The cases that are fixed here should compile as-is actually. This is not so different from

```
enum short i8 = -1; 
enum byte i16 = -1; 
```

which compiles. Maybe there's another bug in the switch/case semantic.

StianGulpen:
Oh it's a problem of sign...

std/math.d
StianGulpen: I've used another approach which is to promote the switch condition and delegate the task to std.conv.to

PR #7296, opened 2019-11-25
Make std.bigint `@safe`

dkorpel:
A start of making std.bigint `@safe`.
It's WIP because it will certainly fail on targets with `version (D_InlineAsm_X86)` active since I haven't touched that.
If anyone knows how to test that on a 64-bit machine, please tell.

### DIFF
--- a/std/bigint.d
+++ b/std/bigint.d
@@ -907,7 +907,7 @@ public:
         Warning: Casting to/from `const` or `immutable` may break type
         system guarantees. Use with care.
      */
-    T opCast(T)() pure nothrow @nogc @safe const
+    T opCast(T)() pure nothrow @nogc const
     if (is(Unqual!T == BigInt))
     {
         return this;
--- a/std/bigint.d
+++ b/std/bigint.d
@@ -1131,6 +1131,7 @@ public:
     /// ditto
     void toString(Writer)(scope Writer sink, scope const ref FormatSpec!char f) const
     {
+        import std.range.primitives : put;
         const spec = f.spec;
         immutable hex = (spec == 'x' || spec == 'X');
         if (!(spec == 's' || spec == 'd' || spec =='o' || hex))
@@ -1177,23 +1178,23 @@ public:
 
         if (!f.flDash && !f.flZero)
             foreach (i; 0 .. difw)
-                sink(" ");
+                put(sink, " ");
 
         if (signChar)
         {
             scope char[1] buf = signChar;
-            sink(buf[]);
+            put(sink, buf[]);
         }
 
         if (!f.flDash && f.flZero)
             foreach (i; 0 .. difw)
-                sink("0");
+                put(sink, "0");
 
-        sink(buff);
+        put(sink, buff);
 
         if (f.flDash)
             foreach (i; 0 .. difw)
-                sink(" ");
+                put(sink, " ");
     }
 
     /**
--- a/std/bigint.d
+++ b/std/bigint.d
@@ -123,7 +123,7 @@ public:
         this(s.byCodeUnit);
     }
 
-    @system unittest
+    @safe unittest
     {
         // system because of the dummy ranges eventually call std.array!string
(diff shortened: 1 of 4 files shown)


std/bigint.d
```diff
@@ -1124,15 +1121,15 @@ public:
      * $(TR $(TD null) $(TD Default formatting (same as "d") ))
      * )
      */
-    void toString(scope void delegate(const (char)[]) sink, string formatString) const
+    void toString(Writer)(scope Writer sink, string formatString) const
```
dkorpel: This is because a `@safe` function cannot call a `@system` sink and making the delegate `@safe` would be a breaking change.

std/internal/math/biguintcore.d
```diff
@@ -652,14 +652,14 @@ public:
...
-            bool wantSub, ref bool sign) pure nothrow if (is(Tulong == ulong))
+            bool wantSub, ref bool sign) pure nothrow @safe if (is(Tulong == ulong))
     {
         BigUint r;
         if (wantSub)
         {   // perform a subtraction
             if (x.data.length > 2)
             {
-                r.data = cast(immutable) subInt(x.data, y);
+                r.data = (() @trusted => cast(immutable) subInt(x.data, y))();
```
dkorpel: since subInt returns an array allocated with `new`, this can be safely cast to immutable.

PetarKirov:
> If anyone knows how to test that on a 64-bit machine, please tell.

You need to build both druntime and phobos like this:
```
make -f posix.mak MODEL=32
```

After that you should be able to run the unit tests like this:
```
make -f posix.mak unittest MODEL=32
```

PetarKirov:
@dkorpel Thanks for working on this! It looks like you have a few more asm blocks to Mark as `@trusted`: 
https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3851883&isPull=true

pbackus:
Delegates are considered output ranges:

```d
import std.range.primitives;
static assert(isOutputRange!(void delegate(const(char)[]), const(char)[]));
```

The correct way to work with output ranges is to use [`std.range.primitives.put`][1], which will call `sink(...)` if the range is a delegate and `sink.put(...)` if it's an object with a `.put` method.

[1]: https://dlang.org/library/std/range/primitives/put.html

dkorpel:
Should be ready for review now.
I don't know why buildkite failed, but it looks unrelated.

thewilsonator:
Rebase and it should go away.

std/internal/math/biguintcore.d
thewilsonator: Might want to put a note of that there.

thewilsonator: Ditto throughout

std/bigint.d
```diff
@@ -910,14 +907,14 @@ public:
         Warning: Casting to/from `const` or `immutable` may break type
         system guarantees. Use with care.
      */
-    T opCast(T)() pure nothrow @nogc const
+    T opCast(T)() pure nothrow @nogc @safe const
```
thewilsonator: Should this really be `@safe` then?

dkorpel: I don't actually know what "type system guarantees" that comment refers to, to me it seems just as `@safe` as casting `immutable int` or `const int` to `int`: a simple copy is made that can be mutated.
Also there's not `@trusted` code involved, so if it compiles it should be good.

atilaneves: Why is a template function being marked expliictily as `@safe`?

dkorpel: So that when it gets turned into something not `@safe` the error message points out why instead of simply saying " the `@safe` unittest can't call the `@system` template function`.
I mostly did it to be consistent with existing annotations on template functions in std.bigint, but I can do without if that's better.

dkorpel:
I removed the `@safe` annotation on the `opCast`. 
Is there anything I can do to help advance this PR?

PR #7297, opened 2019-12-01
Deal with x^^3, just like PowExp in expressionsem

kubo39:


atilaneves:
Tests? What's the motivation?

ibuclaw:
> Tests? What's the motivation?

Not sure, however it might not be a bad idea to move the constfold/optimize handling of `^^` from the compiler to phobos.

berni44:
I compared all floats on correctness:

For +3 8.9% of the results differed, but it where only rounding differences and I guess non of the two algorithms does the rounding accurately, but I havn't checked that.

For -3 14.7% of the results differed, some where far of, for example

pow(6984345059328.0f,-3) == .000000000000000000000000000000000000002935100238...

old: 0.000000000000000000000000000000000000002935100908...
new: 0.0

The reason is, that x*x*x with x=6984345059328.0f is allready inf for floats; casting x to real before multiplying might fix this.

It seems, that the old version (for -3) produces better results (and even the current implementation of -2 is questionable).

berni44:
@RazvanN7 With PR #7783 merged, IMHO this one can be closed.

RazvanN7:
Thanks @berni44 . @kubo39 Do you mind closing this?

PR #7298, opened 2019-12-01
Fix issue 20425: Remove Proxy.opCmp constraints

aliak00:
They fail with types that have an opCmp defined and are value
types.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -6696,7 +6696,6 @@ mixin template Proxy(alias a)
         }
 
         auto ref opCmp(this X, B)(auto ref B b)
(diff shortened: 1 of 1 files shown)


thewilsonator:
Issue number?

aliak00:
There’s isn’t one I could find. Just went off this:
https://forum.dlang.org/thread/lnaiyxvfxodckurrfysi@forum.dlang.org

thewilsonator:
Please either make one or add a changelog entry so this is recorded.

Thanks!

aliak00:
@thewilsonator made and added to title, is there a way to trigger the dlang bot to pick it up?

thewilsonator:
You need it to be part of the commit message. The title isn't picked up at all.

PR #7299, opened 2019-12-03
Fix Issue 13853 - Assert fail in std.random unittests

berni44:
I'm not sure, if we want to have unittests, that only work high likely but not certain. If yes, the bug report can be closed as invalid. If not, this PR fixes it by commenting out the test.

Anyway, I reduced the amount of cycles run in this tests in PR #7270, because the test did run too long. It's still one of the slowest of all tests of phobos.

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -3986,6 +3986,7 @@ if (isInputRange!Range && hasLength!Range && isUniformRNG!UniformRNG)
             }
         }
 
+/+
         /* Test behaviour if .popFront() is called before sample is read.
          * This is a rough-and-ready check that the statistical properties
          * are in the ballpark -- not a proper validation of statistical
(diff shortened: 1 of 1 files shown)


atilaneves:
> I'm not sure, if we want to have unittests, that only work high likely but not certain. 

No, we do not. Unit tests need to be 100% reliable.

> If yes, the bug report can be closed as invalid. 

It's valid if it fails every now and again.

> If not, this PR fixes it by commenting out the test.

Please don't do that. At most loop through the test N amount of times so that it'll nearly always pass.

berni44:
> > I'm not sure, if we want to have unittests, that only work high likely but not certain.
> 
> No, we do not. Unit tests need to be 100% reliable.
> 
> [...]
> 
> > If not, this PR fixes it by commenting out the test.
> 
> Please don't do that. At most loop through the test N amount of times so that it'll nearly always pass.

I don't understand. IMHO these two sentences are a contradiction. The first one states a unittest should be "100% reliable" and the second one askes me to make it "nearly always pass" which I translate to <100% reliable. What did I get wrong?

atilaneves:
You're right, they're contradictory. But looping should be enough to guarantee 0% flakiness. If it doesn't, we need to rethink how to do this.

berni44:
> You're right, they're contradictory. But looping should be enough to guarantee 0% flakiness. If it doesn't, we need to rethink how to do this.

I think, that's exactly the problem. It's allready a loop of 50.000 steps (has been even 100.000 when the bug was reported and the original author confesses that this is actually to few for good estimates). I would not mind to increase this value, but the test is very slow, the slowest test in whole Phobos. So we should really rethink, what we want to test here exactly...

PR #7300, opened 2019-12-03
Fix Issue 16223 - BigUint: undefined shift for small instantiation type

berni44:
Additionally I removed a wrong comment.

### DIFF
--- a/std/internal/math/biguintcore.d
+++ b/std/internal/math/biguintcore.d
@@ -724,12 +724,14 @@ public:
 
 
     //  return x*y.
-    //  y must not be zero.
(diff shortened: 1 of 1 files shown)


thewilsonator:
```
std/internal/math/biguintcore.d(1141): Error: `@safe` function `std.internal.math.biguintcore.__unittest_L1137_C20` cannot call `@system` function `std.internal.math.biguintcore.BigUint.mulInt!int.mulInt`
std/internal/math/biguintcore.d(727):        `std.internal.math.biguintcore.BigUint.mulInt!int.mulInt` is declared here
```
Otherwise looks good.

PR #7301, opened 2019-12-04
Fix Issue 13663 - Comparison of Tuples with floating point fields

berni44:
Thanks to Simen Kjaeraas for pointing out this solution.


std/typecons.d
```diff
@@ -804,11 +804,15 @@ if (distinctFieldNames!(Specs))
          *   $(LI A positive integer if the expression `v1 > v2` is true.)
          *   $(LI 0 if the expression `v1 == v2` is true.))
          */
-        int opCmp(R)(R rhs)
+        float opCmp(R)(R rhs)
         if (areCompatibleTuples!(typeof(this), R, "<"))
         {
             static foreach (i; 0 .. Types.length)
             {
+                if (field[i] != field[i] || rhs.field[i] != rhs.field[i])
```
MoonlightSentinel: This is quite a pessimization, especially for user defined types. I would suggest:

1. Apply this check only for floating point types
2. Move it into the following `if`-block as `field[i] != rhs.field[i]` should be false if either operand is `NaN` (unless my FPN math is wrong)

std/typecons.d
```diff
@@ -804,11 +804,15 @@ if (distinctFieldNames!(Specs))
          *   $(LI A positive integer if the expression `v1 > v2` is true.)
          *   $(LI 0 if the expression `v1 == v2` is true.))
          */
-        int opCmp(R)(R rhs)
+        float opCmp(R)(R rhs)
```
MoonlightSentinel:
```suggestion
        auto opCmp(R)(R rhs)
```

std/typecons.d
```diff
@@ -818,11 +822,15 @@ if (distinctFieldNames!(Specs))
         }
 
         /// ditto
-        int opCmp(R)(R rhs) const
+        float opCmp(R)(R rhs) const
```
MoonlightSentinel: ditto

std/typecons.d
```diff
@@ -818,11 +822,15 @@ if (distinctFieldNames!(Specs))
...
         if (areCompatibleTuples!(typeof(this), R, "<"))
         {
             static foreach (i; 0 .. Types.length)
             {
+                if (field[i] != field[i] || rhs.field[i] != rhs.field[i])
```
MoonlightSentinel: ditto

berni44:
Two of the tests in buildkite produce this error:

std/typecons.d(814,32): Error: template instance
`isFloatingPoint!"key"` does not match template declaration
`isFloatingPoint(T)`

It's clear, that this comes from the tuple having named elements. But it seems like I need a more thorough understanding of how Tuple works for getting this right.

std/typecons.d
```diff
@@ -804,27 +804,33 @@ if (distinctFieldNames!(Specs))
...
          */
-        int opCmp(R)(R rhs)
+        auto opCmp(R)(R rhs)
         if (areCompatibleTuples!(typeof(this), R, "<"))
         {
             static foreach (i; 0 .. Types.length)
             {
                 if (field[i] != rhs.field[i])
                 {
+                    static if (isFloatingPoint!(Specs[i]) || isFloatingPoint!(typeof(rhs.field[i])))
```
MoonlightSentinel: `Specs` includes the optional field names but `Types` should work.
```suggestion
                    static if (isFloatingPoint!(Types[i]) || isFloatingPoint!(typeof(rhs.field[i])))
```

MoonlightSentinel: I might also be a good idea to split this up, e.g. when comparing `Tuple!(int)` and `Tuple!(double)`

berni44:
I switched to math:isNaN, because now we know, that we compare floats.

And: It feels a little bit strange to me, that the two functions differ only by a "const"...

berni44:
I just noticed, that the docs of opCmp where wrong too and obviously did not account for NaNs. I fixed that too.

std/typecons.d
```diff
@@ -796,35 +796,71 @@ if (distinctFieldNames!(Specs))
...
-        int opCmp(R)(R rhs)
+        auto opCmp(R)(R rhs)
         if (areCompatibleTuples!(typeof(this), R, "<"))
         {
             static foreach (i; 0 .. Types.length)
             {
                 if (field[i] != rhs.field[i])
                 {
+                    import std.math : isNaN;
+                    static if (isFloatingPoint!(Types[i]) && isFloatingPoint!(typeof(rhs.field[i])))
```
MoonlightSentinel: This is kinda redundant as both checks will be generated individually below.
(=> Smaller change)

std/typecons.d
```diff
@@ -796,35 +796,61 @@ if (distinctFieldNames!(Specs))
...
+         * 0 if the expression `v1 == v2` is true for all members, or else for
+         * the first member, where this is not true:
          *
          * $(UL
-         *   $(LI A negative integer if the expression `v1 < v2` is true.)
-         *   $(LI A positive integer if the expression `v1 > v2` is true.)
-         *   $(LI 0 if the expression `v1 == v2` is true.))
+         *   $(LI NaN, in case one of the operands is a NaN.)
+         *   $(LI A negative number if the expression `v1 < v2` is true.)
+         *   $(LI A positive number if the expression `v1 > v2` is true.))
```
Biotronic: It would appear the `v1 == v2` case is no longer covered here.

std/typecons.d
```diff
@@ -796,35 +796,61 @@ if (distinctFieldNames!(Specs))
...
+                    {
+                        if (isNaN(field[i]))
+                            return float.nan;
+                    }
+                    static if (isFloatingPoint!(typeof(rhs.field[i])))
+                    {
+                        if (isNaN(rhs.field[i]))
+                            return float.nan;
+                    }
+
```
Biotronic: Suggest adding this for user-defined types with float `opCmp`s:
```
static if (isFloatingPoint!(typeof(field[i].opCmp(field[i]))))
{
    if (field[i] != field[i])
        return float.nan;
}
static if (isFloatingPoint!(typeof(rhs.field[i].opCmp(rhs.field[i]))))
{
    if (rhs.field[i] != rhs.field[i])
        return float.nan;
}
```

Biotronic: Bah. The above fails since of course `float` doesn't have an `opCmp` member. Something like this is required:
`static if (is(typeof({static assert(isFloatingPoint!(typeof(field[i].opCmp(field[i]))));})))`
For the record: that's ugly.

std/typecons.d
```diff
@@ -1531,6 +1557,15 @@ if (distinctFieldNames!(Specs))
...
 
+// issue 13663
+@safe unittest
+{
+    auto t = tuple(real.nan);
+    assert(!(t > t));
+    assert(!(t < t));
+    assert(!(t == t));
+}
+
```
Biotronic: Would also need a test for user-defined types with float `opCmp`s:
```
@safe unittest
{
    struct S
    {
        float opCmp(S s) { return float.nan; }
        bool opEquals(S s) { return false; }
    }
    
    auto t = tuple(S());
    assert(!(t > t));
    assert(!(t < t));
    assert(!(t == t));
}
```

std/typecons.d
berni44: That's covered in line 801.

Biotronic: In my defense, I'm an idiot. :p

std/typecons.d
berni44: What about that:

```
                    auto tmp1 = field[i] < rhs.field[i];
                    auto tmp2 = field[i] > rhs.field[i];
                    if (!tmp1 && !tmp2)
                        return float.nan;
```

Biotronic: That works, but brings us back to the performance discussion from earlier.

Biotronic: Also, kinda obsoletes the `isFloatingPoint` test above.

berni44: OK. So we need something else here...

berni44:
Next try. ;-)

std/typecons.d
Biotronic: This looks better. One very minor nitpick is it's possible to have `opCmp` return a type with an overloaded `opCmp` that returns float (and actually any number of levels to this `opCmp` tower). I'm not sure why anyone would do that, but as a serial offender on operator overload abuse, I feel it's my duty to inform people of its dangers. Should we want to actually support this, I would suggest adding this to std.traits:
``` 
template hasNanComparison(T1, T2 = T1)
{
    static if (isFloatingPoint!T1)
        enum hasNanComparison = true;
    else static if (!isAggregateType!T1)
        enum hasNanComparison = false;
    else static if (!is(typeof((T1 a, T2 b) => a.opCmp(b))))
        enum hasNanComparison = false;
    else
        enum hasNanComparison = hasNanComparison!(ReturnType!((T1 a, T2 b) => a.opCmp(b)));
}

unittest {
    assert( hasNanComparison!float);
    assert( hasNanComparison!real);
    assert(!hasNanComparison!int);
    assert(!hasNanComparison!string);
    assert(!hasNanComparison!(int*));
    assert(!hasNanComparison!(int[]));
    
    struct S
    {
        float opCmp(S s) { return float.nan; }
    }
    assert( hasNanComparison!S);
    
    struct S2
    {
        S opCmp(S2 s) { return S(); }
    }
    assert( hasNanComparison!S2);
}
```
I would argue, however, that this addition is not necessary.

berni44: Maybe this could be done in a separate PR if anyone needs it?!? I would be happy to finish with this PR, though.

Biotronic: [Agreed.](https://github.com/dlang/phobos/pull/7342)

RazvanN7:
Closing this in favor of: #7748

PR #7302, opened 2019-12-04
Fix Issue 15891 - Compiler error when std.algorithm.cache after 2 layers of std.algorithm.map

berni44:
Let's make the compiler happy...

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -207,6 +207,12 @@ if (isBidirectionalRange!Range)
     assert(counter == iota(-4, 5).length);
 }
 
+// issue 15891
+@safe pure unittest
+{
+    assert([1].map!(x=>[x].map!(y=>y)).cache.front.front == 1);
+}
+
 /++
 Tip: `cache` is eager when evaluating elements. If calling front on the
 underlying range has a side effect, it will be observable before calling
@@ -343,9 +349,17 @@ private struct _Cache(R, bool bidir)
         source = range;
         if (!range.empty)
         {
-             caches[0] = source.front;
-             static if (bidir)
-                 caches[1] = source.back;
+            caches[0] = source.front;
+            static if (bidir)
+                caches[1] = source.back;
+        }
+        else
+        {
+            // needed, because the compiler cannot deduce, that 'caches' is initialized
+            // see issue 15891
+            caches[0] = UE.init;
+            static if (bidir)
+                caches[1] = UE.init;
         }
     }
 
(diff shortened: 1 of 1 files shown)


std/algorithm/iteration.d
```diff
@@ -380,7 +392,7 @@ private struct _Cache(R, bool bidir)
         if (!source.empty)
             caches[0] = source.front;
         else
-            caches = CacheTypes.init;
+            caches[0] = UE.init;
```
berni44: When running the unittests, the compiler complained here about a missing frame pointer. I don't know, how this is related to the original issue, but I think, it's a bug here anyway, because setting `caches` to `CacheTypes.init` overwrites the back element in a bidirectional range.

berni44: I was wrong on this. Of course if the range is empty, front and back should both be set to init.

But meanwhile I found out, that `CacheTypes` is `MapResult!(__lambda2, int[])` for a cached map and `CacheTypes.init` is `MapResult(null, null)`, which doesn't look like what want's to have here. I replaced this by explizit initializers now.

std/algorithm/iteration.d
```diff
@@ -343,9 +349,15 @@ private struct _Cache(R, bool bidir)
...
         if (!range.empty)
         {
-             caches[0] = source.front;
-             static if (bidir)
-                 caches[1] = source.back;
+            caches[0] = source.front;
+            static if (bidir)
+                caches[1] = source.back;
+        }
+        else
```
atilaneves: I'd add a comment here explaining why it's needed.

std/algorithm/iteration.d
```diff
@@ -343,9 +349,15 @@ private struct _Cache(R, bool bidir)
...
         if (!range.empty)
         {
-             caches[0] = source.front;
-             static if (bidir)
-                 caches[1] = source.back;
+            caches[0] = source.front;
+            static if (bidir)
+                caches[1] = source.back;
+        }
+        else // needed, because the compiler cannot deduce, that 'caches' is initialized
```
atilaneves: Thank you for adding the comment. Please include a link to the issue that this is fixing so someone can look it up if necessary.

std/algorithm/iteration.d
```diff
@@ -380,7 +392,11 @@ private struct _Cache(R, bool bidir)
         if (!source.empty)
             caches[0] = source.front;
         else
```
atilaneves: Comment here as well please.

std/algorithm/iteration.d
```diff
@@ -389,7 +405,10 @@ private struct _Cache(R, bool bidir)
         if (!source.empty)
             caches[1] = source.back;
         else
```
atilaneves: Ditto.

std/algorithm/iteration.d
berni44: I added the number of the issue. I hope this is what you meant with "link", because weblinks to bugreports are quite uncommon in Phobos - I couldn't find a single one...

berni44:
ping?

PR #7303, opened 2019-12-04
Bring back "std.range.dropBack" code examples

tastyminerals:
For some reason `std.range.dropBack` **Example** field contains examples from `std.range.drop` in the [docs](https://dlang.org/library/std/range/drop_back.html). I think removing them from here should bring back the actual ones.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -3313,16 +3313,6 @@ if (isBidirectionalRange!R)
 {
     import std.algorithm.comparison : equal;
 
-    assert([0, 2, 1, 5, 0, 3].drop(3) == [5, 0, 3]);
-    assert("hello world".drop(6) == "world");
-    assert("hello world".drop(50).empty);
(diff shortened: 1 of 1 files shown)


std/range/package.d
```diff
@@ -3319,6 +3318,7 @@ if (isBidirectionalRange!R)
     assert("hello world".take(6).drop(3).equal("lo "));
 }
 
+///
```
MoonlightSentinel: This is fine. It ensures that the following unittest is an example for `dropBack`

tastyminerals: Is there anything else I should do?

tastyminerals:
Is there anything else I should do?

MoonlightSentinel:
Looks fine to me.

http://dtest.dlang.io/artifact/website-d393e0b0617db4094bf1ceb9d783dbc4ba278904-a1038298f98269122d41740b2381d730/web/phobos-prerelease/std_range.html#drop

PR #7306, opened 2019-12-07
Beware using std.file.copy it incorrect changes creation time!

AuoroP:
See auorop's answer to Beware using std.file.copy it incorrect changes creation time which:
http://stackoverflow.com/questions/39290690/watching-a-directory-tree-whitout-inotify/39301466#39301466


thewilsonator:
Please rebase on stable, not retarget, thanks!

rikkimax:
When ``Yes.preserveAttributes`` is passed to ``copy`` on platforms that are not Windows, it should preserve these attributes. If it does not it is a bug.

Creation time (at least on Linux) is file system dependent. In the VM I tested this in, it stores it specific to the file system and cannot be accessed via e.g. stat. But in this specific case not even the file explorer of the desktop environment supported getting it. At which point if Phobos does not support it, it's not a bug.

ibuclaw:
@thewilsonator this looks like a nonsense/spam pull request to me.

Tentatively closing.

PR #7307, opened 2019-12-07
Fix Issue 9588 - format prints context pointer for struct

berni44:
Thanks to Simen Kjaeraas, who pointed out this solution - seems to be the most straight forward to me.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -4391,23 +4391,22 @@ if ((is(T == struct) || is(T == union)) && (hasToString!(T, Char) || !is(Builtin
         put(w, left);
         foreach (i, e; val.tupleof)
         {
-            static if (0 < i && val.tupleof[i-1].offsetof == val.tupleof[i].offsetof)
+            static if (__traits(identifier, val.tupleof[i]) == "this")
+                continue;
+            else static if (0 < i && val.tupleof[i-1].offsetof == val.tupleof[i].offsetof)
             {
                 static if (i == val.tupleof.length - 1 || val.tupleof[i].offsetof != val.tupleof[i+1].offsetof)
                     put(w, separator~val.tupleof[i].stringof[4..$]~"}");
                 else
                     put(w, separator~val.tupleof[i].stringof[4..$]);
             }
+            else static if (i+1 < val.tupleof.length && val.tupleof[i].offsetof == val.tupleof[i+1].offsetof)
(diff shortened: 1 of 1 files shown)


std/format.d
```diff
@@ -4391,22 +4391,27 @@ if ((is(T == struct) || is(T == union)) && (hasToString!(T, Char) || !is(Builtin
...
-                else
-                    put(w, separator~val.tupleof[i].stringof[4..$]);
-            }
+            static if (__traits(identifier, val.tupleof[i]) == "this")
+                continue;
             else
             {
-                static if (i+1 < val.tupleof.length && val.tupleof[i].offsetof == val.tupleof[i+1].offsetof)
-                    put(w, (i > 0 ? separator : "")~"#{overlap "~val.tupleof[i].stringof[4..$]);
+                static if (0 < i && val.tupleof[i-1].offsetof == val.tupleof[i].offsetof)
```
thewilsonator: Please make this an `else static if` rather than introduce an extra pair of braces.

std/format.d
```diff
@@ -4391,22 +4391,27 @@ if ((is(T == struct) || is(T == union)) && (hasToString!(T, Char) || !is(Builtin
...
+                        put(w, separator~val.tupleof[i].stringof[4..$]~"}");
+                    else
+                        put(w, separator~val.tupleof[i].stringof[4..$]);
+                }
                 else
                 {
-                    static if (i > 0)
-                        put(w, separator);
-                    formatElement(w, e, f);
+                    static if (i+1 < val.tupleof.length && val.tupleof[i].offsetof == val.tupleof[i+1].offsetof)
```
thewilsonator: Ditto

std/format.d
```diff
@@ -4391,22 +4391,27 @@ if ((is(T == struct) || is(T == union)) && (hasToString!(T, Char) || !is(Builtin
...
+                        put(w, (i > 0 ? separator : "")~"#{overlap "~val.tupleof[i].stringof[4..$]);
+                    else
+                    {
+                        static if (i > 0)
```
thewilsonator: ditto

berni44: I think, this cannot be replaced, because of the following `formatElement` call.

thewilsonator: Oh I didn't see that

PR #7308, opened 2019-12-07
Win32test

berni44:
This is meant for hunting down a bug on win32. As I have no local access to such a machine I use these PR. (I hope, this is fine.)


berni44:
Closing this, cannot reproduce...

PR #7312, opened 2019-12-09
Fix Issue 20440 - Associative arrays of BigInt does not support require or get function

GoaLitiuM:


Biotronic:
While this PR does fix the specific issue as described in the bug report, the underlying problem is in `require()`, as it will fail to compile for any type with an overloaded `opAssign` that doesn't return ref. D's own [operator overloading page](https://dlang.org/spec/operatoroverloading.html) uses `void` return type, so we obviously cannot expect `opAssign` to return ref.

In short: the proposed changes to `object.require()` also need to be done.

GoaLitiuM:
I'll make the changes, but the new unittest is causing invariant to fail in `BigUint`.

n8sh:
I agree with @Biotronic:
>The fix in PR 7312 only fixes this for BigInt specifically. If we are not going to require that opAssign return ref for every type, this should be fixed in druntime instead.
>In other words: this is a bug in require(), not in BigInt.

As such I'm proposing as an alternative https://github.com/dlang/druntime/pull/2869

GoaLitiuM:
Closing for dlang/druntime#2869

PR #4342, opened 2016-05-21
fix issue 15940 - ImplicitConversionTargets and class alias in struct

BBasile:

PR #7313, opened 2019-12-10
Fix Issue 15940 - ImplicitConversionTargets and class alias in struct

berni44:
Based on PR #4342. I don't really completely understand the implications of this change, but all tests pass and as the original PR has been more or less approved by JackStouffer and I think, that PR was only closed, because the author of the PR left, I resubmit it.


FeepingCreature:
As a side effect, this introduced a "bug" - `class A; ImplicitConversionTargets!(const A)` is now wrongly `(Object)` instead of, as previous, wrongly `()`. Correct would be `(const Object)`.

PR #7317, opened 2019-12-12
Fix Issue 18446 - Wrong curl onProgress examples

berni44:
The other examples in std.net.curl should IMHO also be checked...

Is it possible to move them to unittests instead? Or does the internet access needed lead to a problem? And if yes, what is best practice here?

### DIFF
--- a/std/net/curl.d
+++ b/std/net/curl.d
@@ -2187,10 +2187,11 @@ private mixin template Protocol()
       * ----
       * import std.net.curl, std.stdio;
       * auto client = HTTP("dlang.org");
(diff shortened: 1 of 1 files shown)


CyberShadow:
Thank you!

> Is it possible to move them to unittests instead? Or does the internet access needed lead to a problem?

Yeah.

> And if yes, what is best practice here?

It would be nice if they were, at least, syntax-checked.

I remember @wilzbach wrote some code to extract examples or unittests in documentation somewhere (in connection with the "try it yourself" feature on dlang.org), but I don't remember more specific details.

thewilsonator:
Do mind the (unrelated) buildkite failure.

PR #7318, opened 2019-12-14
Partial replace call to snprintf for '%e' and float or double.

berni44:
This is a followup on #7264 (not yet merged) and #7285 (merged). This time it implements %e and %E specifiers. Part of this is identical to #7264 to make this PR independed.

I run essentially the same tests as in #7264.

Concerning correctness: All tests passed. No exceptions this time.

Concerning speed: For floats the new version is faster, but for doubles numbers with large or small exponent are somewhat problematic. snprintf manages here to get much better results. I've got some ideas on how it might be possible to improve this, but I'd like to postpone this for an extra PR. See [pdf-file](https://github.com/berni44/printFloat/blob/master/printE.pdf) for some diagrams.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -2654,7 +2654,7 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
 
     size_t len;
     char[] buf;
-    if (fs.spec=='a' || fs.spec=='A')
+    if (fs.spec=='a' || fs.spec=='A' || fs.spec=='e' || fs.spec=='E')
     {
         static if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.mant_dig))
         {
@@ -7065,12 +7065,6 @@ private auto printFloat(T, Char)(T val, FormatSpec!Char f,
                                  RoundingMode rm = RoundingMode.toNearestTiesToEven)
 if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.mant_dig))
 {
-    import std.conv : to;
-    import std.algorithm.comparison : max;
-
-    enum byte[16] alpha = ['0','1','2','3','4','5','6','7','8','9','a','b','c','d','e','f'];
-    enum byte[16] Alpha = ['0','1','2','3','4','5','6','7','8','9','A','B','C','D','E','F'];
-
     union FloatBits
     {
         T floatValue;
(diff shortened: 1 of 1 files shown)


std/format.d
```diff
@@ -7538,3 +7553,782 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
...
+
+        if (f.flDash)
+        {
+            if (sgn!="")
+                result[0] = sgn[0];
+
+            int dot_pos = cast(int) (sgn.length + 1);
+            if (f.precision > 0 || f.flHash)
+                result[dot_pos] = '.';
+            if (dot_pos + ((f.precision > 0 || f.flHash) ? 1 : 0) + f.precision + 4 < result.length)
```
thewilsonator: Please give a name to the expression `dot_pos + ((f.precision > 0 || f.flHash) ? 1 : 0)`

std/format.d
```diff
@@ -7538,3 +7553,782 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
...
+        int upper = 60 - (-exp - 1) % 60;
+        if (upper < T.mant_dig)
+        {
+            mybig[0] = (mnt & ((1L << (T.mant_dig - upper)) - 1)) << 60 - (T.mant_dig - upper);
+            mybig[1] = mnt >> (T.mant_dig - upper);
+        }
+        else
+            mybig[0] = mnt << (upper - T.mant_dig);
+
+        int lsu = 0; // Least significant ulong; when it get's zero, we can ignore it furtheron
```
thewilsonator: space in furtheron?

std/format.d
```diff
@@ -7538,3 +7553,782 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
...
+    size_t start = max_left;
+    size_t left = max_left;
+    size_t right = max_left;
+
+    int final_exp = 0;
+
+    enum roundType { ZERO, LOWER, FIVE, UPPER }
+    roundType next;
+
+    if (exp >= T.mant_dig)
```
thewilsonator: There seems to be a lot of similarity between this branch and the branch below, can you facts it a bit more. its rather hard to follow.

berni44: Not sure if you want me to add more comments or to explain it here?

berni44:
I added a description of the three main algorithms in the comments.

For the sake of reviewing I recommend to do %f (PR #7264) first, because the algorithm there is more straight forward (and therefore easier to understand) and the algorithm here builds on that algorithm.

berni44:
Fixed problem with global import of `FloatingPointControl`. See #7346 for more details.

berni44:
> That is a monster of a function. I've a good look over it and in the interests of moving things forward lets merge this.

First of all: Thanks for doing this.

Well, I havn't found a way to split this into smaller PRs. Maybe it could have been split in one function generating only the digits of the number and an other one taking care of the flags (not sure if that would have a negative impact on speed). But as there are so many of these flags and corner cases I feard, that this would imply a lot of overload handing information from one function to the other.

Having said this: Now I'm quite unsure if you did this because of my posting in the forum or if it was just a coincidence. I believe the former. And I don't know, if you are aware of the similar %f-PR or not. Again I believe the former...

PR #7321, opened 2019-12-15
Fix Issue 5628 - std.math unittest disabled - roundoff error in pow()

berni44:
The original tests compare apples and oranges, that is, left side double/float, right side real. The values on the left side are correct in the precision available but real has better precision and therefore the tests failed. Casting to the lower precision fixes this.

As a side note: I dislike these tests, because the right side could still produce wrong results, depending on the implementation of * and / and rounding errors. I'd prefere to compare to manually calculated literals.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -7892,17 +7892,8 @@ if (isFloatingPoint!(F) && isIntegral!(G))
 
     assert(pow(x, neg1) == 1 / x);
 
-    // Test disabled on most targets.
-    // See https://issues.dlang.org/show_bug.cgi?id=5628
-    version (X86_64)   enum BUG5628 = false;
-    else version (ARM) enum BUG5628 = false;
-    else               enum BUG5628 = true;
-
-    static if (BUG5628)
-    {
-        assert(pow(xd, neg2) == 1 / (x * x));
-        assert(pow(xf, neg8) == 1 / ((x * x) * (x * x) * (x * x) * (x * x)));
-    }
+    assert(approxEqual(pow(xd, neg2), cast(double) (1 / (x * x)), 1e-25, 0.0));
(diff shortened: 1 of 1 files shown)


berni44:
WIP: The unittest failed on FreeBSD. i added some writelns to see, what happens...

wilzbach:
> WIP: The unittest failed on FreeBSD. i added some writelns to see, what happens...

FYI: We only semi-officially support FreeBSD. For example, std.net.curl isn't supported at all anymore and with the upcoming FreeBSD 12 isn't supported either. So if there's no immediate fix/solution, I would suggest not to bother about FreeBSD (i.e. disable it for FreeBSD) and waste time into that sink, but into more interesting and impactful things ;-)

berni44:
Unfortunately win32 has the same bug.

berni44:
Not WIP anymore. I found a workaround with `approxEqual`. The constants used there are so small, that it's actually comparing for equality. 

IMHO it's a bug in DMD and I filed a [bugreport](https://github.com/dlang/phobos/pull/7264). We are discussing it there currently.

kinke:
I don't think the compiler is to blame here at all. The problem/surprising fact is that this templated `pow` overload takes a `F` floating-point input type, but *always* uses `real p = 1.0, v = void;` - that's what's causing the excess/surprising intermediate precision.

berni44:
> I don't think the compiler is to blame here at all. The problem/surprising fact is that this templated `pow` overload takes a `F` floating-point input type, but _always_ uses `real p = 1.0, v = void;` - that's what's causing the excess/surprising intermediate precision.

From the discussions in the bugreport I meanwhile know, that "intermediate results" is stretched far more on 32bit machines than I thought. The result type of `pow` is essentially `F`, but the compiler returns a `real`. If that would be cast to `F` before returning, as I think it should, it would work.

thewilsonator:
This is green, is it good to go?

berni44:
IMHO yes

PR #7322, opened 2019-12-16
Fix Issue 7006 - std.math.pow (integral, integral) crashes on negative exponents

berni44:
I improved the docs too, including the public unittest.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -7903,19 +7903,43 @@ if (isFloatingPoint!(F) && isIntegral!(G))
     assert(equalsDigit(pow(2.0L, 10.0L), 1024, 19));
 }
 
-/** Compute the value of an integer x, raised to the power of a positive
- * integer n.
+/**
+ * Compute the power of two integral numbers.
  *
- *  If both x and n are 0, the result is 1.
- *  If n is negative, an integer divide error will occur at runtime,
- * regardless of the value of x.
+ * Params:
+ *     x = base
+ *     n = exponent
+ *
+ * Returns:
+ *     x raised to the power of n. If n is negative the result is 1 / pow(x, -n),
+ *     which is calculated as integer division with remainder. This may result in
+ *     a division by zero error.
+ *
+ *     If both x and n are 0, the result is 1.
+ *
+ * Throws:
+ *     If x is 0 and n is negative, the result is the same as the result of a
+ *     division by zero.
  */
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -7954,10 +7969,32 @@ if (isIntegral!(F) && isIntegral!(G))
         }
         break;
     }
+
```
thewilsonator: Unneeded.

std/math.d
```diff
@@ -7912,20 +7912,35 @@ if (isFloatingPoint!(F) && isIntegral!(G))
...
+ *     division by zero.
  */
 typeof(Unqual!(F).init * Unqual!(G).init) pow(F, G)(F x, G n) @nogc @trusted pure nothrow
 if (isIntegral!(F) && isIntegral!(G))
 {
-    if (n<0) return x/0; // Only support positive powers
     typeof(return) p, v = void;
     Unqual!G m = n;
 
+    if (x == -1) return cast(typeof(return)) (m % 2 == 0 ? 1 : -1);
```
ibuclaw: `m & 1` is more efficient.

berni44: Isn't that a job for the optimizer? ;-) Anyway, I changed it.

berni44:
I do not understand, why this always fails on ci/circleci. First I thought it is unrelated, but meanwhile the error was reproduced with each run: std/range/package.test fails. When I run that test locally it succeeds (but needs about 6 minutes). What does ci/circleci actually do?

std/math.d
```diff
@@ -7912,20 +7912,35 @@ if (isFloatingPoint!(F) && isIntegral!(G))
...
+ * Compute the power of two integral numbers.
+ *
+ * Params:
+ *     x = base
+ *     n = exponent
+ *
+ * Returns:
+ *     x raised to the power of n. If n is negative the result is 1 / pow(x, -n),
+ *     which is calculated as integer division with remainder. This may result in
+ *     a division by zero error.
```
ahmetsait: I think `1 / pow(x, -n)` supposed to be `1 / pow(x, n)`

thewilsonator:
Circle is green now, but `mir/algorithm` still fails consistently.

berni44:
> Circle is green now, but `mir/algorithm` still fails consistently.

Found (and fixed) the problem: `m <= -1` is not a good idea, when `m` is unsigned. I added also a unittest (`assert(pow(0,uint.max) == 0);`) to guard for that problem. (There are two more expressions I put into `static if`, but that is mere for extra security; IMHO they could not break and therefore I could not add a guarding unittest for them.)

PR #7323, opened 2019-12-16
std.math: Split tan into internal.math.mathnoasm and mathx86

ibuclaw:
Something that I think has been long overdue, splitting out the architecture-specific parts of std.math into a separate module.  Beginning with tan() - others will follow suit if this is fine.

@kinke - Just followed in the example of biguint.  Is this similar to what you had in mind, or were you more for splitting modules on a per-thing basis (math.elemental, math.fp_operations, etc...)

### DIFF
--- a/std/internal/math/mathnoasm.d
+++ b/std/internal/math/mathnoasm.d
@@ -20,9 +20,9 @@ import std.math : floatTraits, RealFormat, PI, PI_4,
 
 /////////////////////////////////////////////////////////////////////////////
 
-T tan(T)(T x) @safe pure nothrow @nogc
+// Coefficients for tan(x) and PI/4 split into three parts.
+private template tanCoefficents(T)
 {
-    // Coefficients for tan(x) and PI/4 split into three parts.
     enum realFormat = floatTraits!T.realFormat;
     static if (realFormat == RealFormat.ieeeQuadruple)
     {
@@ -88,12 +88,17 @@ T tan(T)(T x) @safe pure nothrow @nogc
     }
     else
         static assert(0, "no coefficients for tan()");
+}
 
+T tan(T)(T x) @safe pure nothrow @nogc
+{
     // Special cases.
     if (x == cast(T) 0.0 || isNaN(x))
         return x;
     if (isInfinity(x))
         return T.nan;
+
+    mixin tanCoefficents!T;
 
     // Make argument positive but save the sign.
     bool sign = false;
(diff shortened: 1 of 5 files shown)


std/internal/math/mathnoasm.d
```diff
@@ -0,0 +1,154 @@
...
+import std.math : floatTraits, RealFormat, PI, PI_4,
+       isNaN, isInfinity, signbit, floor, poly;
+
+/////////////////////////////////////////////////////////////////////////////
+
+T tan(T)(T x) @safe pure nothrow @nogc
+{
+    // Coefficients for tan(x) and PI/4 split into three parts.
+    enum realFormat = floatTraits!T.realFormat;
+    static if (realFormat == RealFormat.ieeeQuadruple)
```
thewilsonator: Since we're in the process of making this nicer, can we move these out onto a (private) mixin template?

ibuclaw: Like this?

thewilsonator: Yep.

kinke:
I'd prefer per-thing ;) modules - potentially based on the current grouping in the docs (https://dlang.org/phobos/std_math.html), e.g., `std.math.{trig,rounding,exp,log,mod,introspection}`. The asm implementations in those submodules could be further extracted.

ibuclaw:
> I'd prefer per-thing ;) modules - potentially based on the current grouping in the docs (https://dlang.org/phobos/std_math.html), e.g., std.math.{trig,rounding,exp,log,mod,introspection}. The asm implementations in those submodules could be further extracted.

Couldn't that be done independently of where the internal implementation is?

e.g:
```
// module std.math;
public import std.math.trig;
```
```
// module std.math.trig;
real tan(real x) { return std.internal.math.mathcore.tan(x); }
```

kinke:
It could, but I definitely prefer the generic implementations + unittests to reside in those submodules directly instead of a similar src tree in std.internal.math.

thewilsonator:
+1 for  `import std.math.trig;`

ibuclaw:
> +1 for import std.math.trig;

I'll need to remove `package` access from some functions, or maybe I stumbled upon a discrepancy for importing `std/math.d` vs. `std/math/package.d` that may require resolving first.

ibuclaw:
> > +1 for import std.math.trig;
> 
> I'll need to remove `package` access from some functions, or maybe I stumbled upon a discrepancy for importing `std/math.d` vs. `std/math/package.d` that may require resolving first.

Yep, there's potentially a compiler bug. :-(

```
// stdx/a.d
module stdx.a;
import stdx.b;

void main()
{
    foo();
}
```
```
// stdx/b.d  -> OK
// stdx/b/package.d -> Error: undefined identifier foo
module stdx.b;

package void foo()
{
}
```

PR #7324, opened 2019-12-16
Remove useless warnings

drug007:
to get rid of annoying deprecation warnings because in case the rhs is an other Nullable instance the constraint `is(typeof(this.get == rhs))` triggers the deprecation warning because the rhs implicitly calls deprecated `get_` member. This guard constraint prevents the rhs from being an other Nullable instance. No resolution logic change. Just getting rid of useless warning.


thewilsonator:
Restarting buildkite due to network failures.

PR #7325, opened 2019-12-17
Fix Issue 15405 - FormatSpec.writeUpToNextSpec() not documented

berni44:
The style checker gives the following error, that I don't understand, because the normal unittests can compile this. Can anyone help me here?

```
generated/linux/release/64/publictests/std_format.d(1242): Error: template std.format.FormatSpec cannot deduce function from argument types !()(string), candidates are:
std/format.d(1011):        FormatSpec(Char)
  with Char = char
  must satisfy the following constraint:
       !is(Unqual!Char == Char)
std/format.d(1021):        FormatSpec(Char)
```

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -1192,6 +1192,22 @@ if (is(Unqual!Char == Char))
         trailing = fmt;
     }
 
+    /**
+       Write the format string to an output range until the next format
+       specifier is found and parse that format specifier.
+
+       See $(LREF FormatSpec) for an example, how to use `writeUpToNextSpec`.
+
+       Params:
+           writer = the $(REF_ALTTEXT output range, isOutputRange, std, range, primitives)
+
+       Returns:
+           True, when a format specifier is found.
+
+       Throws:
+           A $(LREF FormatException) when the found format specifier
+           could not be parsed.
+     */
     bool writeUpToNextSpec(OutputRange)(ref OutputRange writer) scope
     {
         if (trailing.empty)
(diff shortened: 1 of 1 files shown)


berni44:
Meanwhile I found out, that I had to add the template parameter `dchar` explicitly in the example. But I still don't understand, why I have to do so in a public example, while the other unittests work without...

berni44:
There are still linker errors when I run the style checker - similar to buildkite/phobos: style. I have no idea, what goes wrong here...

berni44:
I noticed, that the public unittest already existed outside of the template. So I just added a link in the docs and changed that unittest a little bit to show how the return value works. (And I replaced %2.4f by %6.4f, because 2.4 is a typical fallacy when using `format`.)

MoonlightSentinel:
> Meanwhile I found out, that I had to add the template parameter `dchar` explicitly in the example. But I still don't understand, why I have to do so in a public example, while the other unittests work without...

Probably because there is no template interference for constructors IIRC (albeit it should probably result in a better error message)

schveiguy:
> Probably because there is no template interference for constructors

It's because of the documented unit test being inside the template, in that case `FormatSpec` is the instantiated struct.

But when it gets copied outside of the template to make sure the public example works, it needs the instantiating parameter.

schveiguy:
The test seemed to have failed for an environmental issue. I'm re-running it.

quickfur:
Question: *should* `writeUpToNextSpec` even be a public function in the first place?  It looks more like implementation detail to me than a public function that must be documented, even though apparently it has never been marked `private`.

schveiguy:
It was in the original docs for FormatSpec (in fact, this PR barely changes any unittests at all, just verifies the value is correct). I'm not 100% familiar with the type, but I'm guessing if you wanted to use FormatSpec outside of std.format, you would need this function, right?

berni44:
> Question: _should_ `writeUpToNextSpec` even be a public function in the first place? It looks more like implementation detail to me than a public function that must be documented, even though apparently it has never been marked `private`.

That was my first thought too. But there might be usecases for this struct (and than you need these function as it is somewhat the heart of this struct): Think of an export function, where the user can specify the order and the appearence of the exported items. This function could be used to check the userinput by parsing it.

schveiguy:
I'm also guessing that if DIP 1027 is approved, this will become an important struct for 3rd party libraries.

PR #7329, opened 2019-12-23
Fix unittest, which assumes real > double.

berni44:
See https://github.com/dlang/phobos/pull/7219#discussion_r360729014 for the motivation.

When `real==double`, the function calculates for `x==-double.max` the value `-inf + inf + ...` which produces a `nan` and therefore the overall result is a `nan` which breaks the test in `testFindRoot`.

### DIFF
--- a/std/numeric.d
+++ b/std/numeric.d
@@ -1498,8 +1498,8 @@ T findRoot(T, R)(scope R delegate(T) f, in T a, in T b,
         //++numCalls;
         if (x>float.max)
             x = float.max;
-        if (x<-double.max)
-            x = -double.max;
+        if (x<-float.max)
+            x = -float.max;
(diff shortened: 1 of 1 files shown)


kinke:
Thx; should target stable though.

berni44:
> Thx; should target stable though.

The problematic change hasn't made it into stable yet. Therefore I decided to target it at master.

kinke:
It is in stable; LDC only uses stable.

berni44:
Oops, you're right. When I checked this this morning, I probably used an outdated version locally.

Now I'm lost a little bit, what I can do now... I assume it's better to rebase those two commits on stable than to add another PR, isn't it? But in that case, I fear, I cannot do anything about it...

PR #7330, opened 2019-12-23
Provide a complete description of dirName output

bachmeil:
The current documentation is not complete. The user may be surprised when this function returns ".", as it may lead to problems if you are using the output of a call to dirName elsewhere.


thewilsonator:
`Error - file 'std/path.d' contains trailing whitespace at line 537`

PR #7331, opened 2019-12-26
Merge pull request #7328 from berni44/reg_numeric_customfloat

berni44:
Fix unittest, which assumed real exponents having at least 15 bit.
merged-on-behalf-of: Nicholas Wilson <thewilsonator@users.noreply.github.com>

It's the first time I'm using cherry-pick, so I hope I got everything right. The original PR #7328 was merged on master, but should have been on stable. This PR is intended to fix this...


berni44:
The problems in buildkite/phobos seem not to be related. Is it possible to restart this test?

Anyway, as this is a regression fix, which has allready been merged into master, I think, this should be merged rather soon to avoid a final stable release without it.

berni44:
ping?

PR #7333, opened 2019-12-27
Fix Issue 7446 - [TDPL] Trivial asynchronous file copy example crashes

berni44:
The original issue (docs telling, that the runtime catches the exception) seems not to exist anymore, but anyway, the functions should document, that they might throw. i also cleaned up the docs of the three functions a little bit.

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -669,6 +669,12 @@ private void _send(T...)(MsgType type, Tid tid, T vals)
  * matched by an earlier delegate.  If more than one argument is sent,
  * the `Variant` will contain a $(REF Tuple, std,typecons) of all values
  * sent.
+ *
+ * Params:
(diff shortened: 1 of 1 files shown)


atilaneves:
The "docs" in issue 7446 is [TDPL](https://www.amazon.com/D-Programming-Language-Andrei-Alexandrescu/dp/0321635361), meaning that someone who read the book would get behaviour different from described there.

berni44:
> The "docs" in issue 7446 is [TDPL](https://www.amazon.com/D-Programming-Language-Andrei-Alexandrescu/dp/0321635361), meaning that someone who read the book would get behaviour different from described there.

Ah, I thought TDPL was just an old way to reference the D docs.

The changes i suggest (including closing the issue) make sense anyway, don't they?

berni44:
ping?

PR #7337, opened 2019-12-29
Make assumeUTF nothrow

berni44:
I subscribe to the view of the author of the bug (and the people in the corresponding forum thread) that `debug` should not be used inside Phobos.

I also sligthly changed the example unittest as the current output `writeln(a) // c` is somewhat confusing.

### DIFF
--- a/std/string.d
+++ b/std/string.d
@@ -7394,16 +7394,27 @@ Params:
 Returns:
     arr retyped as an array of chars, wchars, or dchars
 
+Throws:
+    In debug mode `AssertError`, when the result is not a well-formed UTF string.
+
 See_Also: $(LREF representation)
 */
-auto assumeUTF(T)(T[] arr) pure
+auto assumeUTF(T)(T[] arr)
 if (staticIndexOf!(Unqual!T, ubyte, ushort, uint) != -1)
 {
     import std.traits : ModifyTypePreservingTQ;
+    import std.exception : collectException;
     import std.utf : validate;
+
     alias ToUTFType(U) = AliasSeq!(char, wchar, dchar)[U.sizeof / 2];
-    auto asUTF = cast(ModifyTypePreservingTQ!(ToUTFType, T)[])arr;
-    debug validate(asUTF);
+    auto asUTF = cast(ModifyTypePreservingTQ!(ToUTFType, T)[]) arr;
+
+    debug
+    {
+        scope ex = collectException(validate(asUTF));
+        assert(!ex, ex.msg);
+    }
+
     return asUTF;
 }
 
@@ -7414,7 +7425,7 @@ if (staticIndexOf!(Unqual!T, ubyte, ushort, uint) != -1)
     immutable(ubyte)[] b = a.representation;
     string c = b.assumeUTF;
 
-    assert(a == c);
+    assert(c == "Hölo World");
 }
 
 pure @system unittest
@@ -7451,4 +7462,17 @@ pure @system unittest
         assert(equal(jt, htc));
         assert(equal(jt, hti));
     }}
+}
+
+pure @system unittest
+{
+    import core.exception : AssertError;
+    import std.exception : assertThrown, assertNotThrown;
+
+    immutable(ubyte)[] a = [ 0xC0 ];
+
+    debug
+        assertThrown!AssertError( () nothrow @nogc @safe {cast(void) a.assumeUTF;} () );
+    else
+        assertNotThrown!AssertError( () nothrow @nogc @safe {cast(void) a.assumeUTF;} () );
 }


berni44:
But, the behaviour in debug mode would still differ from the behaviour in normal mode, which, at least in my oppinion, is still a bug.

IMHO it should be the user who decides, what should be debugged and what not, not the designer of a library. If the user wants to make sure, that the return value from assertUTF is valid, he can add a `debug validate(result)`. If this is done automatically by the library, the user cannot decide anymore and has no better workaround than writing his own assertUTF function.

MoonlightSentinel:
Another reason why this should be checked as an `in(isValidUTF(...))` unless `in`/`out`/`assert`/... does not belong into phobos as well.

If thats not desired then he can use a simple cast ....

MoonlightSentinel:
Anyway, the real culprit here is a dmd bug that infers not-`nothrow` from an `debug` statement which violates the [spec](https://dlang.org/spec/version.html#DebugStatement)

berni44:
Not sure, how to continue. Can we decide on debug in Phobos or in/out or not here, or should we ask that in that forum? IMHO it's a broader question... What do you think?

MoonlightSentinel:
Not entirely sure about this either but I think it might be favourable to split this problem in its independent parts - `debug` in Phobos and `UTFException` instead of `assert`.

`assumeUTF` should be consistent with the other `assumeX` methods in Phobos which validate that their assumptions hold. IMHO these methods should only be used if you are 100% sure their input satisfies `X`, otherwise its a programming error (which might be hard to detect when dealing with malformed unicode). 
Hence the current `validate(...)` (which throws an `UTFException`) should be replaced with something like `assert(isValidUTF(...))` because the programmer assumed something which does not hold.

`debug` on the other hand is a more complicated issue. Assuming we switch to `assert` it wouldn't matter anymore because the observable behaviour wouldn't change (IIRC an `assert` violation is allowed to result in undefined behaviour as it would usually terminate the programm ).
One argument in favour of `debug` is that the validation of arbitrary strings can incur a lot of overhead which might not be desired for releaase builds with enabled assertions.

I wouldn't exptect the forum post to reach a consenus but considering other views on this topic won't be detrimental either.

PS: Removing these checks also makes it questionable whether `assumeUTF` even belongs in Phobos because it would be equivalent to `(() @trusted => cast(string) bytes)()` (except automatically selecting the appropriate character type)

berni44:
I changed the PR to throw an error in case of an invalid UTF provided. This still uses `validate` and thus is still not `@nogc`. Changing that looks like a lot of work and I do not know, when I'll find the time to do so, especially as I'm currently not familiar with `std.utf`.

I'm aware, that this change is the opposite of the desired one: Now it's never `@nogc` instead of always, which probably was hoped for by the bug reporter.

MoonlightSentinel:
> I changed the PR to throw an error in case of an invalid UTF provided. This still uses `validate` and thus is still not `@nogc`. Changing that looks like a lot of work and I do not know, when I'll find the time to do so, especially as I'm currently not familiar with `std.utf`.

There is already #4012 to introduce `isValidUTF` which shared most of its implementation with `validate`. Sadly its was blocked because the original author was expected to fix all the cruft of the original implementation as well.

> I'm aware, that this change is the opposite of the desired one: Now it's never `@nogc` instead of always, which probably was hoped for by the bug reporter.

Then lets keep it behind `debug`  for now to be consistent with the other assumeX methods (whether we should have `debug` statements inside of phobos is an entirely different issue).

Otherwise we could resort to `std.traits`  to fake `@nogc`

std/string.d
```diff
@@ -7385,25 +7385,27 @@ return it typed as a UTF string.
...
 {
     import std.traits : ModifyTypePreservingTQ;
+    import std.exception : assumeWontThrow;
     import std.utf : validate;
+
     alias ToUTFType(U) = AliasSeq!(char, wchar, dchar)[U.sizeof / 2];
-    auto asUTF = cast(ModifyTypePreservingTQ!(ToUTFType, T)[])arr;
-    debug validate(asUTF);
+    auto asUTF = cast(ModifyTypePreservingTQ!(ToUTFType, T)[]) arr;
+    assumeWontThrow(validate(asUTF));
```
MoonlightSentinel: `assumeWontThrow` produces the following message which hides the real issue:
```
assumeWontThrow failed: Expression did throw
```

The assert should probably contain the exceptions message:


```suggestion
    scope ex = collectException!UTFException(validate(asUTF));
    assert(!ex, ex.msg);
```

berni44:
> Then lets keep it behind `debug` for now

Not sure, if I got that right: You want me to *not* add `debug`?

MoonlightSentinel:
> Not sure, if I got that right: You want me to _not_ add `debug`?

No, keep the call to `validate` in a `debug` block.

berni44:
> No, keep the call to `validate` in a `debug` block.

I started to implement this this morning, but realised, that I do not commit to this PR anymore. It's neither closing an issue nor removing a `debug` with this change; just a small improvement on a public example... Is it OK for you, when I start the debate about `debug` in the forum first?

MoonlightSentinel:
> I started to implement this this morning, but realised, that I do not commit to this PR anymore. It's neither closing an issue nor removing a debug with this change

This changes fixes the original issue that `assumeUTF` isn't always `nothrow` (`@nogc` is already inferred because of the `debug` condition). Whether we should have `debug` conditions in phobos is independent of the current issue, hence i propose to keep it for now.

> Is it OK for you, when I start the debate about `debug` in the forum first?

Do however  you like,

berni44:
> This changes fixes the original issue that `assumeUTF` isn't always `nothrow` (`@nogc` is already inferred because of the `debug` condition).

No, that's not the original issue. It complains about `@nogc` and different behaviour with and without `debug`. Throwing is not mentioned there.

Anyway I changed the PR now to address the throwing stuff.

MoonlightSentinel:
The issue title refers to attributes in general, the description just mentions `@nogc` in general.

But lets not waste time bikeshedding, I think this PR is a good change for the better.

std/string.d
```diff
@@ -7394,16 +7394,27 @@ Params:
 Returns:
     arr retyped as an array of chars, wchars, or dchars
 
+Throws:
+    In debug mode `AssertError`, when the result is not a well-formed UTF string.
+
 See_Also: $(LREF representation)
 */
-auto assumeUTF(T)(T[] arr) pure
+auto assumeUTF(T)(T[] arr) pure nothrow @nogc
```
atilaneves: Unless there's a compiler error, none of the attributes should be required. The way to make sure the attributes are inferred correctly is to add them to a unittest that calls assumeUTF.

berni44: OK. Moved the attributes in the unittest.

std/string.d
```diff
@@ -7394,16 +7394,27 @@ Params:
...
     import std.utf : validate;
+
     alias ToUTFType(U) = AliasSeq!(char, wchar, dchar)[U.sizeof / 2];
-    auto asUTF = cast(ModifyTypePreservingTQ!(ToUTFType, T)[])arr;
-    debug validate(asUTF);
+    auto asUTF = cast(ModifyTypePreservingTQ!(ToUTFType, T)[]) arr;
+
+    debug
+    {
+        scope ex = collectException(validate(asUTF));
```
Geod24: On a side note, it would be great to have a version of `validate` that calls a user-provided routine on invalid UTF, instead of throwing. It'd be useful here, and it'd be useful in places where performance is important (e.g. Vibe.d HTTP handling).

PR #7339, opened 2019-12-30
Improved error message when calling sum without a default value

Zevenberge:
If sum() is called, it tries to find a default value by constructing it with 0 (e.g. float(0)). This works for numbers, but produces obscure errors if the value cannot be constructed this way (e.g. issue #19525). The error message now states that the problem lies in the default value and suggests to supply one instead.

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -6398,6 +6398,9 @@ if (isInputRange!R && !isInfinite!R && is(typeof(r.front + r.front)))
         alias Seed = typeof(E.init  + 0.0); //biggest of double/real
     else
         alias Seed = typeof(r.front + r.front);
+    static assert(is(typeof(Unqual!Seed(0))),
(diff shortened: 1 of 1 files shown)


thewilsonator:
Please amend the commit message to explicitly say `Fix issue 19525` if this fixes that issue.

thewilsonator:
Ah sorry, that should have been the commit title.

Zevenberge:
I don't think it really fixes the issue: the provided code example `[1.minutes].sum()` still doesn't work out of the box. This PR is really just a clarification of the error message for user defined value types (like Duration).

MoonlightSentinel:
Maybe `sum` could fall back to `Seed.init` allthough that would rely on `.init` being a logical valid state (which might be problematic for `struct`s with explicit factory functions)

atilaneves:
@MoonlightSentinel I've been thinking about the T.init idea and I'm not sure yet.

PR #7340, opened 2019-12-30
Replace all uses of approxEqual in Phobos with isClose

berni44:
Still a draft...


std/complex.d
```diff
@@ -464,16 +464,16 @@ if (isFloatingPoint!T)
...
-    assert(approxEqual(abs(cdc), abs(c1)/abs(c2), EPS));
-    assert(approxEqual(arg(cdc), arg(c1)-arg(c2), EPS));
+    assert(isClose(abs(cdc), abs(c1)/abs(c2), EPS));
+    assert(isClose(arg(cdc), arg(c1)-arg(c2), EPS));
 
     auto cec = c1^^c2;
-    assert(approxEqual(cec.re, 0.11524131979943839881, EPS));
-    assert(approxEqual(cec.im, 0.21870790452746026696, EPS));
+    assert(isClose(cec.re, 0.1152413197994, 1e-12));
+    assert(isClose(cec.im, 0.2187079045274, 1e-12));
```
berni44: This might look like a reduction in accuracy, but the opposite is true. The original tests are essentialy `approxEqual(cec.re, 0.11, 0.0)` and `approxEqual(cec.im, 0.21, 0.0)`, the remaining digits where not used for comparison.

std/complex.d
```diff
@@ -525,26 +525,26 @@ if (isFloatingPoint!T)
...
-    assert(approxEqual(arg(rec2b), PI_2, EPS));
+    assert(isClose(abs(rec2b), std.math.exp(-2 * PI), EPS));
+    assert(isClose(arg(rec2b), PI_2, EPS));
 
     auto rec3a = 0.79 ^^ complex(6.8, 5.7);
     auto rec3b = complex(0.79, 0.0) ^^ complex(6.8, 5.7);
-    assert(approxEqual(rec3a.re, rec3b.re, EPS));
-    assert(approxEqual(rec3a.im, rec3b.im, EPS));
+    assert(isClose(rec3a.re, rec3b.re, 1e-14));
+    assert(isClose(rec3a.im, rec3b.im, 1e-14));
```
berni44: Here, and in several more places, the absolute difference check with accuracy of 1e-2 was used, instead of the intended relative difference check with accuracy of `EPS`. With `isClose` this test is disabled until you explicitly ask for. The given accuracy of the relative difference check was too much for 32bit computers, so I reduced it a little bit. Might still be too much on computers where `real==double`.

std/complex.d
```diff
@@ -557,13 +557,13 @@ if (isFloatingPoint!T)
...
     assert(isIdentical(rer2.re, rcheck));
-    assert(approxEqual(rer2.im, 0.0, EPS));
+    assert(isClose(rer2.im, 0.0, 0.0, 1e-10));
 
     auto rer3 = (-a) ^^ complex(-2.0, 0.0);
     rcheck = (-a) ^^ (-2.0);
     assert(feqrel(rer3.re, rcheck) == double.mant_dig);
     assert(isIdentical(rer3.re, rcheck));
-    assert(approxEqual(rer3.im, 0.0, EPS));
+    assert(isClose(rer3.im, 0.0, 0.0, EPS));
```
berni44: Whenever a comparison to zero is applied, the releative difference check does not work anymore. I disabled it with a default of `0.0` and instead used `EPS` for the absolute difference check, which is appropriate here. I did this at several other places too.

std/internal/math/gammafunction.d
```diff
@@ -1829,12 +1829,12 @@ real logmdigammaInverse(real y)
...
-    assert(approxEqual(logmdigammaInverse(logmdigamma(real.min_normal)), real.min_normal, 1e-15, 0));
-    assert(approxEqual(logmdigammaInverse(logmdigamma(real.max/2)), real.max/2, 1e-15, 0));
+        assert(isClose(logmdigammaInverse(test[0]), test[1], 2e-15));
+
+    assert(isClose(logmdigamma(logmdigammaInverse(1)), 1, 1e-15));
+    assert(isClose(logmdigamma(logmdigammaInverse(real.min_normal)), real.min_normal, 1e-15));
+    assert(isClose(logmdigamma(logmdigammaInverse(real.max/2)), real.max/2, 1e-15));
+    assert(isClose(logmdigammaInverse(logmdigamma(1)), 1, 1e-15));
+    assert(isClose(logmdigammaInverse(logmdigamma(real.min_normal)), real.min_normal, 1e-15));
+    assert(isClose(logmdigammaInverse(logmdigamma(real.max/2)), real.max/2, 1e-15));
```
berni44: 0 is the default for isClose, so it is not needed here.

std/math.d
```diff
@@ -1210,13 +1210,13 @@ private T tanImpl(T)(T x) @safe pure nothrow @nogc
             T t = tan(x);
 
             //printf("tan(%Lg) = %Lg, should be %Lg\n", cast(real) x, cast(real) t, cast(real) r);
-            assert(approxEqual(r, t));
+            assert(isClose(r, t, CommonDefaultFor!(T,T), CommonDefaultFor!(T,T)));
```
berni44: This is a bit tricky, because the test appears inside of two `static foreach`. For most of this tests `isClose(r, t)` would be enough, but there are two tests included which compare to `0.0`. For these tests, the absolute test needs to be enabled with the fourth parameter. As these are templated tests, the floating point numbers are of different types. `CommonDefaultFor!(T,T)` calculates the best values for these tests.

Alternatively one could check if `r==0` and apply a different test in that case - this would avoid the use of the private function `CommonDefaultFor`. See the tests for `atan2` where the original tests contained such a solution. I will change that here (and at some other places too) if you think this would be better.

std/numeric.d
```diff
@@ -1914,9 +1914,9 @@ do
         }
         {
             auto ret = findLocalMin!T((T x) => -fabs(x), -1, 1, T.min_normal, 2*T.epsilon);
-            assert(ret.x.fabs.approxEqual(T(1)));
-            assert(ret.y.fabs.approxEqual(T(1)));
-            assert(ret.error.approxEqual(T(0)));
+            assert(ret.x.fabs.isClose(T(1)));
+            assert(ret.y.fabs.isClose(T(1)));
+            assert(ret.error.isClose(T(0), 0.0, 100*T.epsilon));
```
berni44: `T.epsilon` was too small here, so I increased this value by two magnitudes. Still better than `1e-2` which was used by `approxEqual`. But maybe this is a hint, that the implementation of `findLocalMin` could be improved.

std/numeric.d
```diff
@@ -2345,8 +2345,8 @@ if (isInputRange!(Range1) && isInputRange!(Range2))
     assert(kullbackLeiblerDivergence(p, p1) == 2);
     assert(kullbackLeiblerDivergence(p1, p) == double.infinity);
     double[] p2 = [ 0.2, 0.2, 0.2, 0.4 ];
-    assert(approxEqual(kullbackLeiblerDivergence(p1, p2), 0.0719281));
-    assert(approxEqual(kullbackLeiblerDivergence(p2, p1), 0.0780719));
+    assert(isClose(kullbackLeiblerDivergence(p1, p2), 0.0719281, 1e-5));
+    assert(isClose(kullbackLeiblerDivergence(p2, p1), 0.0780719, 1e-5));
```
berni44: As I have no knowledge of `kullbackLeiblerDivergence` I could not calculate more digits for comparison (4 more digits would be needed, but I do not want to use the function to get these digits, because the function might return wrong digits.). So I reduced the accuracy instead (still better than `1e-2`).

std/numeric.d
```diff
@@ -2420,17 +2420,17 @@ if (isInputRange!Range1 && isInputRange!Range2 &&
...
     assert(jensenShannonDivergence(p1, p1) == 0);
-    assert(approxEqual(jensenShannonDivergence(p1, p), 0.548795));
+    assert(isClose(jensenShannonDivergence(p1, p), 0.548795, 1e-5));
     double[] p2 = [ 0.2, 0.2, 0.2, 0.4 ];
-    assert(approxEqual(jensenShannonDivergence(p1, p2), 0.0186218));
-    assert(approxEqual(jensenShannonDivergence(p2, p1), 0.0186218));
-    assert(approxEqual(jensenShannonDivergence(p2, p1, 0.005), 0.00602366));
+    assert(isClose(jensenShannonDivergence(p1, p2), 0.0186218, 1e-5));
+    assert(isClose(jensenShannonDivergence(p2, p1), 0.0186218, 1e-5));
+    assert(isClose(jensenShannonDivergence(p2, p1, 0.005), 0.00602366, 1e-5));
```
berni44: dito

std/numeric.d
```diff
@@ -3599,37 +3599,37 @@ void inverseFft(Ret, R)(R range, Ret buf)
...
     auto arr = [1,2,3,4,5,6,7,8];
     auto fft1 = fft(arr);
-    assert(approxEqual(map!"a.re"(fft1),
-        [36.0, -4, -4, -4, -4, -4, -4, -4]));
-    assert(approxEqual(map!"a.im"(fft1),
-        [0, 9.6568, 4, 1.6568, 0, -1.6568, -4, -9.6568]));
+    assert(isClose(map!"a.re"(fft1),
+        [36.0, -4, -4, -4, -4, -4, -4, -4], 1e-4));
+    assert(isClose(map!"a.im"(fft1),
+        [0, 9.6568, 4, 1.6568, 0, -1.6568, -4, -9.6568], 1e-4));
```
berni44: dito

std/parallelism.d
```diff
@@ -4553,8 +4553,12 @@ version (unittest)
 // tons of stuff and should not be run every time make unittest is run.
 version (parallelismStressTest)
```
berni44: I ran this stress test and found out, that it was not run for ages (lot's of bugs). So I tried to fix it as good as possible.

std/parallelism.d
```diff
@@ -4734,10 +4744,11 @@ version (parallelismStressTest)
                                )
                            );
 
-            assert(approxEqual(sumSqrt, 4.437e8));
+            assert(isClose(sumSqrt, 4.437e8, 1e-2));
             stderr.writeln("Done sum of square roots.");
 
             // Test whether tasks work with function pointers.
+            /+ // This part is buggy and needs to be fixed...
```
berni44: This test is completely broken, I wasn't able to fix it, so I commented it out instead.

std/algorithm/comparison.d
```diff
@@ -983,8 +983,8 @@ template equal(alias pred = "a == b")
     assert(equal(a, b));
 
     // predicated: ensure that two vectors are approximately equal
-    double[] c = [ 1.005, 2, 4, 3];
-    assert(equal!approxEqual(b, c));
+    double[] c = [ 1.0000000005, 2, 4, 3];
```
atilaneves: This worries me a little bit because I'd expect 1.005 to be close to 1.0 with the defaults.

berni44: That's true for `approxEqual` but not for `isClose`. Actually, this is the reason, why I want to replace `approxEqual`: The default (`1e-2`) is much too weak in most cases (a little bit like a security belt, which doesn't work when needed). I replaced it with a default which depends on the used type, which is calculated in `CommonDefaultFor` (in `std.math`): `1e-4` for `float`, `1e-9` for `double` and `1e-10` for x87 `real`.

RazvanN7:
Closing as this has been superseded by #7751

PR #7341, opened 2019-12-31
Disable `make install` for the Darwin_64_32 target on the auto-tester

SSoulaimane:
OSX 32 bits on the auto-tester isn't supposed to be busy.

It isn't causing any trouble being run here in phobos but in dmd `make install` sometimes hangs and it has to be disabled in phobos first before it can be disabled in dmd.
See: https://github.com/dlang/dmd/pull/10705.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -292,9 +292,14 @@ else
 all : lib
 endif
 
+ifneq (,$(findstring Darwin_64_32, $(PWD)))
+install:
+	echo "Darwin_64_32_disabled"
+else
 install :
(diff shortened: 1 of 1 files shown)


berni44:
> It isn't causing any trouble being run here in phobos

It does. I noticed this several times in the last two weeks. See [this test](https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3883370&isPull=true) for example.

SSoulaimane:
That one also fails in the "build dmd" step which apparently can block phobos pull requests too?

PR #7343, opened 2020-01-04
Mark CTFE bitmanip function for generating longs to a string as @safe and pure

schveiguy:
Helps with proposed DIP1028 (https://github.com/dlang/dmd/pull/10709, https://github.com/dlang/DIPs/blob/1b705f8d4faa095d6d9e3a1b81d6cfa6d688554b/DIPs/DIP1028.md)

Not sure if there will be more. It's difficult to test this, but this seems to be the most common problem.

cc @WebFreak001

### DIFF
--- a/std/bitmanip.d
+++ b/std/bitmanip.d
@@ -56,13 +56,22 @@ import std.range.primitives;
 public import std.system : Endian;
 import std.traits;
 
-private string myToString(ulong n)
+private string myToString(ulong n) pure @safe
 {
     import core.internal.string : UnsignedStringBuf, unsignedToTempString;
     UnsignedStringBuf buf;
     auto s = unsignedToTempString(n, buf);
-    return cast(string) s ~ (n > uint.max ? "UL" : "U");
+    // pure allows implicit cast to string
+    return s ~ (n > uint.max ? "UL" : "U");
 }
+
+@safe pure unittest
+{
+    assert(myToString(5) == "5U");
+    assert(myToString(uint.max) == "4294967295U");
+    assert(myToString(uint.max + 1UL) == "4294967296UL");
+}
+
 
 private template createAccessors(
     string store, T, string name, size_t len, size_t offset)


WebFreak001:
ok fixed the bitfields issues in dub packages and has improved the quota so far a bit. What I see now is left when randomly sampling error logs:

- **`std.regex : ctRegex` doesn't compile**
- **`std.functional : toDelegate` broken** (actually nearly fixed by this PR aswell!)
- delegate arguments are still `@system` by default (toString with delegate sink)
- lots of libraries using `__gshared`
- usage of `.ptr`
- usage of `void*`
- cast from `void[]` to `ubyte[]` is not allowed
- calls of `core.stdc.stdlib` not `@safe`
- multiple overloads 

See https://wfr.moe/safe-test/phobos-7343.html (currently building)

But this PR is good to fix the bitfields issue.

schveiguy:
OK good. Those sound more like real issues to be looked at.

I'll see if I can pick some more low hanging fruit...

schveiguy:
> * delegate arguments are still `@system` by default (toString with delegate sink)

This should be fixed in the DIP PR, I would think. I can't imagine why they wouldn't be safe by default.

WebFreak001:
@schveiguy I made a comment about it what it breaks

WebFreak001:
there are also issues with `assumeUnique` and `assumeSafeAppend` being `@system`.

I can see why they are `@system` but one could argue they should be `@trusted` because they are simply just wrappers to make something unsafe safe

aG0aep6G:
> there are also issues with `assumeUnique` and `assumeSafeAppend` being `@system`.
> 
> I can see why they are `@system` but one could argue they should be `@trusted` because they are simply just wrappers to make something unsafe safe

No. `assumeUnique` and `assumeSafeAppend` don't make anything safe.

WebFreak001:
ok dub packages are all tested. We went from
```
compiles: 316, broken: 902, n/a: 499
works: 25.9%
```
to
```
compiles: 448, broken: 770, n/a: 499
works: 36.8%
```
+9.9% improvement in all of dub!

std/bitmanip.d
```diff
@@ -56,14 +56,26 @@ import std.range.primitives;
...
 import std.traits;
 
-private string myToString(ulong n)
+private string myToString(ulong n) pure @safe
 {
-    import core.internal.string : UnsignedStringBuf, unsignedToTempString;
-    UnsignedStringBuf buf;
-    auto s = unsignedToTempString(n, buf);
-    return cast(string) s ~ (n > uint.max ? "UL" : "U");
+    static char[] generateStr(ulong n) pure
```
aG0aep6G: Do you really need the second function? This seems to work for me:

```d
string myToString(ulong n) pure @safe
{
    import core.internal.string : UnsignedStringBuf, unsignedToTempString;
    UnsignedStringBuf buf;
    auto s = unsignedToTempString(n, buf);
    return s ~ (n > uint.max ? "UL" : "U");
}
```

schveiguy: first thing I tried was returning char[] and then I thought I don’t need to for this case and I didn’t try your idea. But not really any reason except I thought that wouldn’t work. I’ll switch to your idea

schveiguy: OK, I updated to just remove the cast, and add pure/safe.

std/bitmanip.d
```diff
@@ -56,14 +56,22 @@ import std.range.primitives;
...
 import std.traits;
 
-private string myToString(ulong n)
+private string myToString(ulong n) pure @safe
 {
     import core.internal.string : UnsignedStringBuf, unsignedToTempString;
     UnsignedStringBuf buf;
     auto s = unsignedToTempString(n, buf);
-    return cast(string) s ~ (n > uint.max ? "UL" : "U");
+    return s ~ (n > uint.max ? "UL" : "U");
```
WebFreak001: I think with the comment "use pure to implicitly cast to string" this was much better

schveiguy: Fixed.

schveiguy:
Thanks. Is there a reason why auto merge label didn't work here? Is it the buildkite failure? @wilzbach

PR #7345, opened 2020-01-05
Some version (unittest) blocks removed

alexandrumc:
Please see https://github.com/dlang/druntime/pull/2888

### DIFF
--- a/std/uni.d
+++ b/std/uni.d
@@ -715,33 +715,6 @@ debug(std_uni) import std.stdio; // writefln, writeln
 
 private:
 
-version (unittest)
-{
-private:
-    struct TestAliasedString
-    {
-        string get() @safe @nogc pure nothrow { return _s; }
-        alias get this;
-        @disable this(this);
-        string _s;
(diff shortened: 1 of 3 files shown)


std/uni.d
```diff
@@ -9255,8 +9225,32 @@ if (isConvertibleToString!Range)
 
 @safe unittest
 {
+    struct TestAliasedString
```
atilaneves: static struct

std/uni.d
```diff
@@ -9255,8 +9225,32 @@ if (isConvertibleToString!Range)
...
+    {
+        string get() @safe @nogc pure nothrow { return _s; }
+        alias get this;
+        @disable this(this);
+        string _s;
+    }
+
+    bool testAliasedString(alias func, Args...)(string s, Args args)
```
atilaneves: static?

std/uni.d
alexandrumc: sure, that makes sense.

PR #7346, opened 2020-01-07
Fix performance regression introduced by PR #7285.

berni44:
The global import of `std.math : FloatingPointControl` caused a performance regression. This PR fixes this by making the import local.

See [discussion](https://github.com/dlang/phobos/pull/7285#pullrequestreview-338238228) for more details.

PS: I checked, using valgrind, that this import indeed caused the regression.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -62,7 +62,6 @@ import std.exception;
 import std.meta;
 import std.range.primitives;
 import std.traits;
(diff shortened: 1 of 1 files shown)


berni44:
Wow, never had a PR with 3 approvals. ;-) Unfortunately the tests got stuck, both just not returning the positive result, as far as I can see...

PR #7347, opened 2020-01-10
Fix issue 20495 (choose copies unused union member, which is unsafe)

schveiguy:
ping @aG0aep6G I think this looks much better.

After this, I probably will try and fix [19738](https://issues.dlang.org/show_bug.cgi?id=19738) as the [PR](https://github.com/dlang/phobos/pull/6933) seems stalled (found it while looking to see if this issue already existed).

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -1546,37 +1546,17 @@ private struct ChooseResult(R1, R2)
     }
 
     static if (isForwardRange!R1 && isForwardRange!R2)
+    @property auto save() return scope
     {
-        private auto systemSave() return scope
+        if (r1Chosen)
         {
-            return r1Chosen
-                ? ChooseResult(r1Chosen, r1.save, r2)
-                : ChooseResult(r1Chosen, r1, r2.save);
+            ref R1 getR1() @trusted { return r1; }
+            return ChooseResult(r1Chosen, getR1.save, R2.init);
         }
-        private auto trustedSave()() @trusted return scope
+        else
         {
-            /*
-            Unsafe operations in this function:
-            - copying r1/r2 (postblit or copy constructor),
-            - r1.save/r2.save,
-            - accessing the union of r1 and r2.
-            The first two cannot be trusted, because they involve calling user
-            code. So they're only called via @safe wrappers.
-            The last one can be trusted. We're not returning a reference into
-            the union.
-            */
-            R safeSave(R)(ref R r) @safe { return r.save; }
-            R safeCopy(R)(ref R r) @safe { return r; }
-            return r1Chosen
-                ? ChooseResult(r1Chosen, safeSave(r1), safeCopy(r2))
-                : ChooseResult(r1Chosen, safeCopy(r1), safeSave(r2));
-        }
-        @property auto save() return scope
-        {
-            static if (__traits(compiles, trustedSave()))
-                return trustedSave();
-            else
-                return systemSave();
+            ref R2 getR2() @trusted { return r2; }
+            return ChooseResult(r1Chosen, R1.init, getR2.save);
         }
     }
 
@@ -1743,6 +1723,29 @@ pure @safe unittest // issue 18657
     static assert(!__traits(compiles, choose(true, R(), R()).save));
     static assert(!__traits(compiles, choose(true, [0], R()).save));
     static assert(!__traits(compiles, choose(true, R(), [0]).save));
+}
+
+@safe unittest // issue 20495
+{
+    static struct KillableRange
+    {
+        int *item;
+        ref int front() { return *item; }
+        bool empty() { return *item > 10; }
+        void popFront() { ++(*item); }
+        this(this)
+        {
+            assert(item is null || cast(size_t) item > 1000);
+            item = new int(*item);
+        }
+        KillableRange save() { return this; }
+    }
+
+    auto kr = KillableRange(new int(1));
+    int[] x = [1,2,3,4,5]; // length is first
+
+    auto chosen = choose(true, x, kr);
+    auto chosen2 = chosen.save;
 }
 
 /**


aG0aep6G:
@thewilsonator, please give people time to respond before merging. I've been pinged on this and before I even became aware of it, it was already in. Maybe the auto-merge thingy should have a timer.

@schveiguy, I maintain that the `@trusted` helpers are strictly speaking invalid. I don't see this PR as an obvious improvement. (But the `.init` thing is fine.)

schveiguy:
I agree, this probably should not have been merged immediately, I wasn't expecting that.

Regarding the trusted helpers, they are defined within context where only the trusted helpers are valid. That is, they are defined *inside* the if statement that verifies they are valid.

The trusted helpers could actually have asserts in them, but having something that looks like:

```D
if(condition)
{
    assert(condition);
    ...
}
```

Seems noisy and unnecessary. The point of trusted functions is that they are valid for all cases they can be called with. Since they are only accessible in this case when they are valid, they should be safe.

The only thing that would make these more solid is to use trusted lambda escapes, so they could only be used within the expressions they were defined, but the function is short enough that it's difficult to see how abuse could creep in. The only way to have this blow up is for someone to change the boolean inside the if statement, or take the address of one of the functions and call it elsewhere.

aG0aep6G:
> The point of trusted functions is that they are valid for all cases they can be called with. Since they are only accessible in this case when they are valid, they should be safe.

The point of `@trusted` functions is that they're as safe as `@safe` functions. If you have a function that cannot be called in certain ways or at certain times, that's `@system`.

> The only way to have this blow up is for someone to change the boolean inside the if statement, or take the address of one of the functions and call it elsewhere.

In other words: There are ways that this can blow up. With correctly applied `@trusted`s, it could not blow up. You could accidentally negate the bool, or leak a delegate to a helper, or mix up the up order of the ranges, and you would still be guaranteed that no memory corruption or other undefined behavior happens.

I know I'm being pedantic. But I think it matters. If we can't follow the rules of the language, we should either try harder, or we should change the rules.

schveiguy:
The trusted functions could have an assert, I thought of doing this:

```D
    @property auto save() return scope
    {
        if (r1Chosen)
        {
            ref R1 getR1() @trusted { assert(r1Chosen); return r1; }
            return ChooseResult(r1Chosen, getR1.save, R2.init);
        }
        else
        {
            ref R2 getR2() @trusted { assert(!r1Chosen); return r2; }
            return ChooseResult(r1Chosen, R1.init, getR2.save);
        }
    }
```

But the functions can only be called when those asserts will be true. So I left them out. But with that, you agree that the trusted functions are safe? I just felt the `if(x) assert(x)` pattern was trivially unnecessary.

The previous way it was, the trusted parts were not any more safe than now, so I don't see a problem.

> In other words: There are ways that this can blow up

Let's not forget that any function with trusted pieces essentially needs full checking. And can always blow up. Even with your function the way it was, you could blow it up by making changes to it in the trusted or safe parts.

The review needs to be on the entire function, and whether THAT code is safe for all calls (that is, as long as the provided type has safe calls for `save`). And from what I can tell, it is.

The only way to make this unassailable is to store both items and not use a union. Then the whole thing can be inferred safe, and no problems can occur. We are using trusted escapes to gain some performance.

aG0aep6G:
> But the functions can only be called when those asserts will be true. So I left them out. But with that, you agree that the trusted functions are safe?

No. That doesn't address the other ways that `getR1` can be misused. And (non-static) asserts can't be used to ensure safety anyway. Run time asserts are entirely optional, even to a safety-compliant compiler.

> The previous way it was, the trusted parts were not any more safe than now, so I don't see a problem.

I'm not convinced that this is true. It's very possible that my `trustedSafe` wasn't as robust as I hoped, but `@safe` and `@trusted` interact differently than in your code.

In your version, the (inferred) `@safe` part has `@trusted` access to the union. You mentioned yourself how that can be used to break safety.

In my version, the `@trusted` part calls the `@safe` part with a reference to the union. That may be problematic, but it's less clear to me how to exploit it. I thought `safeSave` could probably leak the reference to a global, but it turns that doesn't compile, because `ref` implies `scope`.

If you see a way to break my version by editing `safeSave`, then I agree that my code had essentially the same problem as yours has now.

> The review needs to be on the entire function, and whether THAT code is safe for all calls (that is, as long as the provided type has safe calls for `save`).

This is the key to our disagreeing. I don't agree with that sentence. And I don't think that your idea of `@trusted` is backed by the spec, or that it's intended to work that way. Instead, my understanding is that every `@trusted` function (including nested ones, tiny ones, private ones, etc.) must be 100% as safe as any `@safe` function. If an `@trusted` function relies on its inputs (parameters, surroundings) for safety, then it's just plain wrong and should be rejected in review.

As mentioned in the forum, I could get on board with your view for `@trusted` blocks, if we'd be getting them. But for `@trusted` functions I'm not.

> The only way to make this unassailable is to store both items and not use a union. Then the whole thing can be inferred safe, and no problems can occur. We are using trusted escapes to gain some performance.

Just because you and I can't make it work, doesn't mean it's impossible to write a version of this that's properly `@trusted`.

Anyway, I'm not calling for a revert of this or anything like that. I can live fine with two more wrong `@trusted`s in Phobos.

schveiguy:
> No. That doesn't address the other ways that `getR1` can be misused. And (non-static) asserts can't be used to ensure safety anyway. Run time asserts are entirely optional, even to a safety-compliant compiler.

This is a little too pedantic. I could also do `if(!r1Chosen) assert(0);` and it would always compile in all cases.  A trusted function should be trusted within its context. what other ways can it get misused? The idea that an inner function can get misused without someone taking the context of where it is defined into account is not cause for alarm. I trust that function in that context, because that's the only context it's accessible from. Yes, someone can come along and make new ways to access it by doing bizarre things. In fact, I realized you can't screw this up without changing the boolean, which should never happen in this method. Even taking the address of the function and putting it outside the context won't be a problem because you can't get to the function to take its address without the boolean having been correct. With the above assert, the function is totally safe, even if you do stupid things.

> If you see a way to break my version by editing `safeSave`, then I agree that my code had essentially the same problem as yours has now.

 ```D
R safeSave(R)(ref R r) @safe { r1Chosen = !r1Chosen; return r.save; }
```

Not unsafe at this point, but violates safety on the next call. As I said, we have to accept at some point that anyone can break the code by doing unlikely things.

**Comment by aG0aep6G** [comment]
> This is a little too pedantic. I could also do `if(!r1Chosen) assert(0);` and it would always compile in all cases.

That's fair. But you need to write it like that. The form `assert(r1Chosen);` is not equivalent.

> A trusted function should be trusted within its context. what other ways can it get misused? The idea that an inner function can get misused without someone taking the context of where it is defined into account is not cause for alarm. I trust that function in that context, because that's the only context it's accessible from.

No. An `@trusted` function must be safe for all valid inputs. Context is just a kind of input. But I think we're getting stuck in a loop here. You say: "`@trusted` means this". I say: "No, `@trusted` means that`. Repeat.

> Even taking the address of the function and putting it outside the context won't be a problem because you can't get to the function to take its address without the boolean having been correct. With the above assert, the function is totally safe, even if you do stupid things.

I don't think that's true. The bool can change after you've extracted the function (but before you call it):

```d
alias D = ref R1 delegate() @safe;
D g;

struct ChooseResult
{
    /* ... fields and constructor ... */
    @property auto save() return scope
    {
        if (r1Chosen)
        {
            ref R1 getR1() @trusted { return r1; }
            auto dg = &getR1;
            g = dg; /* Aside: This might be exploiting a loophole in the implementation of DIP 1000. */
            return ChooseResult(r1Chosen, getR1.save, R2.init);
        }
        else { /* ... */ }
    }
}

void main() @safe
{
    import std.stdio;
    auto cr = ChooseResult(true, R1(), R2());
    cr.save(); /* `g` now returns a reference to `cr.r1`. */
    cr = ChooseResult(false, R1(), R2()); /* Invalidating `cr.r1`. */
    writeln(g()); /* Shows values of `r2`, or crashes, or whatever - undefined behavior. */
}
```

> ```d
> R safeSave(R)(ref R r) @safe { r1Chosen = !r1Chosen; return r.save; }
> ```
> 
> Not unsafe at this point, but violates safety on the next call. As I said, we have to accept at some point that anyone can break the code by doing unlikely things.

Right. That's a problem that both versions share. Or rather, it's a problem that exists independently of any `save` method. And I'm not willing to accept that problem as inevitable, either. It's a use case for an [`@system` variable](https://github.com/dlang/DIPs/pull/179), which would shield `r1Chosen` against such accidents.

Taking a step back, I know that my nagging might seem excessive, because it's really unlikely that someone will make a mistake that actually triggers UB here. But my point is: I don't subscribe to the idea that "unlikely to break" is good enough when it comes to `@safe`.

D's promise is that it's impossible to write/edit `@safe` code so that it exhibits undefined behavior. And the only exception from that is when you write a bad `@trusted` function. That means, if you can manage to trigger UB in `@safe` code, you've got a bad `@trusted` function somewhere (or a compiler bug, or a hole in the spec).

As far as I understand, your take on that is: "We can't actually keep that promise. We need to accept that some `@safe` code is actually `@trusted` in disguise." I don't agree with that. If we can't keep the promise, we need to change the language so that we can (e.g., by adding `@system` variables). And we also just have to try harder when writing `@trusted` code.

aG0aep6G:
PS: I feel like my comments are getting longer without really saying anything new. I.e., I'm ranting. Maybe we should agree to disagree?

schveiguy:
D's promise for `@trusted` code to be callable by any code exists at a function level, not inside a function. Trusted parts inside a function are not necessarily marked that way because they are "watertight" (they can be, and in that case, no review of the surrounding code is necessary).

In reality, the function itself needs review, but we don't have a mechanism to say "this function has trusted parts". What you need to do is, if you see any trusted parts to a function, and those trusted parts' safety require certain invariants are true in the context, you need to review the ENTIRE function (and maybe even the entire type's functions). There's no escaping that.

But there are very legitimate reasons to only mark parts of a function trusted. Using inference to check the normal code in the function, while trusting certain things that are not provable by the compiler (as is the case here). Another is to ensure certain parts are safe, knowing that other parts are semantically safe. None of these cases mean that you can always draw a bright line between the safe and trusted parts. It makes things less obvious, but does not break the guarantees that these attributes provide.

> Maybe we should agree to disagree?

Maybe we are actually closer to agreement than you think. We just lack the syntax to say what we want to say. I understand your point of view, but practically one wants to be able to use the compiler's safety checking and inference where possible.

> The only way to make this unassailable is to store both items and not use a union.

I may actually be wrong on this, to a certain degree. If we used a proper tagged-union type that existed outside this type, that type itself can be the trusted part, and we could write the rest of the code using safety inference based on the guarantees afforded by the smart union type. In other words, take the boolean out of the equation, and I think it's a lot easier to trust that you can get what is valid. It may not perform as well, but it would be easier to review.

aG0aep6G:
> But there are very legitimate reasons to only mark parts of a function trusted.

I agree. And I wouldn't mind `@trusted` blocks that work like you want. But I don't think you can do it with the `@trusted` function attribute and still write valid code.

> None of these cases mean that you can always draw a bright line between the safe and trusted parts. It makes things less obvious, but does not break the guarantees that these attributes provide.

You're saying that your use cases are incompatible with drawing a bright line between `@safe` and `@trusted`, right? I agree. But the bright line is part of the language. The bright line is synonymous with the guarantees that the attributes provide. If you step over the line, you're breaking the guarantees. Maybe you're breaking them only locally, always contained within a function. But you do break them.

> I understand your point of view, but practically one wants to be able to use the compiler's safety checking and inference where possible.

Sure. But then we need to put that in the spec. If we have a rule in the language which we routinely ignore because respecting it would be too hard or impossible, then that's a sign that the language needs improving.

And instead of trying to formalize the current abuse of `@trusted`, I would prefer proper `@trusted { ... }` blocks (not nested functions) that taint the `@safe` functions they're in. Or even better: `@system { ... }` blocks in `@trusted` functions, like you mentioned in the forum. But I'm not sure if there can be a reasonable migration path for that.

>  If we used a proper tagged-union type that existed outside this type, that type itself can be the trusted part, and we could write the rest of the code using safety inference based on the guarantees afforded by the smart union type.

I'm not saying that it wouldn't improve the situation, but you're just compartmentalizing the `@trusted` violations. That tagged union type would still have a safety-critical tag field. And it could still be set incorrectly by an `@safe` method of that type. Compartmentalization is fine, but if we could make the field inaccessible from `@safe` code (by making it an `@system` variable), that would be better.

PR #7348, opened 2020-01-10
Fix issue 19738 -- std.range.choose assignment breaks @safe-ty

schveiguy:
This was more tricky than I originally thought, mostly due to the existing requirement that `chooseAmong` works with strings during CTFE.

Essentially, we only trust the fetching of the pointer to the chosen range, everything else should be inferred.

I'm not sure why I had to remove `return scope`, but that doesn't seem to work, I let the attributes get inferred. The error messages are totally unrelated to this code, it's just not correctly inferring safety with those attributes. But I don't really understand what the `return scope` did in the first place.

Should obsolete #6933 

@thewilsonator @dnadlinger

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -1500,13 +1500,25 @@ private struct ChooseResult(R1, R2)
         }
     }
 
-    void opAssign(return scope ChooseResult r) @trusted
+    void opAssign(ChooseResult r)
     {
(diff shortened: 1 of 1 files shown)


thewilsonator:
Doc tester timed out please force push to restart.

schveiguy:
I don't have much confidence that the macair is going to succeed on a third run. So I'll force push. Meanwhile, @braddr you have any ideas? It seems to time out quite a bit.

PR #7349, opened 2020-01-11
Fix issue 20493 - incorrect result of BigInt * BigInt

dkorpel:
The problem comes down to a buffer overflow.
How could a buffer overflow happen in `@safe` code?
Well, it's basically a logical buffer overflow, while technically no out of bounds memory is accessed.

Two buffers are created in one allocation of size a + b (presumably as an optimization).
`buffer[a..$]` is given to `partial`, while `scratchbuf` remains the whole `buffer[0..$]` instead of `buffer[0..a]`.
That means that if more than `a` elements are written to `scratchbuf`, it messes up `partial`, which is what happens in the bug report.

The buffer size for Karatsuba multiplication of `x * y` is calculated as `2 * x.length` (x and y are a `BigDigit[]` which is a `uint[]`).
Most calls to `mulKaratsuba` claim `2 * half(x.length)` of `scratchbuf` and pass the rest to a recursive call of half the instance size. I think the underlying assumption is that the geometric series 1 + 0.5 + 0.25 ... <= 2 so the bound is sufficient.

It fails to account for two things however:
- The problem size is integer, and when it is halved, it is rounded up. So when x and y have length 641, the next problem size is 321, then 161, 81 etc. Every time a little more than half of the available space is claimed.
- When the multiplication of the recursive case is unbalanced (i.e. `x.length` is more than `y.length * sqrt(2)`), another half of the buffer is claimed.

So why wasn't this caught earlier? Well, there is a `KARATSUBALIMIT` currently set to `10`, and a `BigDigit[]` with `length < KARATSUBALIMIT` gets a small size optimization so no Karatsuba multiplication is done. This limit gave some constant leeway to the buffer. When numbers get sufficiently large, this asymptotic growth of the required buffer size catches up with the constant leeway of the actual buffer size with lower asymptotic growth however.

The smallest case of this bug with the current limit is when `(x.length, y.length) = (116, 99)`, so this bug only manifests once numbers reach a size of `2^(32*116) ~= 10^1117` which is probably why it got undetected for so long.

Finding the correct minimum buffer size is hard, since it depends on `KARATSUBALIMIT` (for base cases), log2(x.length) (for rounding errors), and how often the recursive case is unbalanced. 
With [some brute force](https://gist.github.com/dkorpel/d5dc56eb27205aa011aaf74e5933d365) testing I found that a little over factor 2 seems sufficient for at least 10 000 big digits even with a smaller KARATSUBALIMIT of 2, so I settled with `2.25` for now.

I also truncated the `scratchbuff` so that at least instead of silently giving an incorrect answer, it raises a RangeError on failure.

### DIFF
--- a/std/internal/math/biguintcore.d
+++ b/std/internal/math/biguintcore.d
@@ -1601,6 +1601,7 @@ void mulInternal(BigDigit[] result, const(BigDigit)[] x, const(BigDigit)[] y)
             // (1.414 ~ 1.707)/2:1 is balanced.
             BigDigit [] scratchbuff = new BigDigit[karatsubaRequiredBuffSize(y.length) + y.length];
             BigDigit [] partial = scratchbuff[$ - y.length .. $];
+            scratchbuff = scratchbuff[0 .. $ - y.length];
             mulKaratsuba(result[0 .. half + y.length], y, x[0 .. half], scratchbuff);
             partial[] = result[half .. half + y.length];
             mulKaratsuba(result[half .. $], y, x[half .. $], scratchbuff);
@@ -1638,6 +1639,7 @@ void mulInternal(BigDigit[] result, const(BigDigit)[] x, const(BigDigit)[] y)
             // We make the buffer a bit bigger so we have space for the partial sums.
             BigDigit [] scratchbuff = new BigDigit[karatsubaRequiredBuffSize(maxchunk) + y.length];
             BigDigit [] partial = scratchbuff[$ - y.length .. $];
+            scratchbuff = scratchbuff[0 .. $ - y.length];
             size_t done; // how much of X have we done so far?
             if (paddingY)
             {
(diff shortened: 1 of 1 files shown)


std/internal/math/biguintcore.d
```diff
@@ -2224,10 +2251,12 @@ bool inplaceSub(BigDigit[] result, const(BigDigit)[] x, const(BigDigit)[] y)
 
 /* Determine how much space is required for the temporaries
  * when performing a Karatsuba multiplication.
+ * TODO: determining a tight bound is non-trivial and depends on KARATSUBALIMIT, see:
+ * https://issues.dlang.org/show_bug.cgi?id=20493
```
thewilsonator: Ping me when you're happy with this `TODO`

dkorpel: I was hoping to leave that for a future PR. The comment is there to explain that the formula is a bit arbitrary and not the result of rigorous mathematical analysis or benchmarking, but it should be good enough.

PR #7350, opened 2020-01-13
add some more auth types to etc.c.curl

John-Colvin:
I'm not sure this is the right thing to do, because I'm not updating the whole header ¯\\_(ツ)_/¯

### DIFF
--- a/etc/c/curl.d
+++ b/etc/c/curl.d
@@ -549,7 +549,7 @@ alias curl_proxytype = int;
 
 ///
 enum CurlAuth : ulong {
-  none =         0ul,        /** None */
+  none =         0UL,        /** None */
   basic =        1UL << 0,   /** Basic (default) */
   digest =       1UL << 1,   /** Digest */
   negotiate =    1UL << 2,   /** Negotiate (SPNEGO) */
--- a/etc/c/curl.d
+++ b/etc/c/curl.d
@@ -548,29 +548,36 @@ enum CurlProxy {
 alias curl_proxytype = int;
 
 ///
-enum CurlAuth : long {
-  none =         0,
-  basic =        1,  /** Basic (default) */
-  digest =       2,  /** Digest */
-  gssnegotiate = 4,  /** GSS-Negotiate */
-  ntlm =         8,  /** NTLM */
-  digest_ie =    16, /** Digest with IE flavour */
-  only =         2_147_483_648, /** used together with a single other
-                                type to force no auth or just that
-                                single type */
-  any = -17,     /* (~CURLAUTH_DIGEST_IE) */  /** all fine types set */
-  anysafe = -18  /* (~(CURLAUTH_BASIC|CURLAUTH_DIGEST_IE)) */ ///
+enum CurlAuth : ulong {
+  none =         0ul,        /** None */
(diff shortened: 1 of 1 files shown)


etc/c/curl.d
```diff
@@ -548,29 +548,36 @@ enum CurlProxy {
...
-  default_  = -1 // CURLSSH_AUTH_ANY;
+  publickey = 1 << 0, /** public/private key files */
+  password  = 1 << 1, /** password */
+  host      = 1 << 2, /** host key files */
+  keyboard  = 1 << 3, /** keyboard interactive */
+  agent     = 1 << 4, /** agent (ssh-agent, pageant...) */
+  gssapi    = 1 << 5, /** gssapi (kerberos, ...) */
+  
+  default_  = any // CURLSSH_AUTH_ANY;
 }
```
MoonlightSentinel: Maybe also add the CURLGSSAPI_DELEGATION_* enum as well since it's missing from the current bindings AFAICT

```suggestion
 }
 enum CurlGssApiDelegation {
     none         = 0          /** no delegation (default) */,
     policyFlag = (1<<0) /** if permitted by policy */,
     flag           = (1<<1) /** delegate always */
 }
```

etc/c/curl.d
```diff
@@ -548,29 +548,36 @@ enum CurlProxy {
...
+enum CurlAuth : ulong {
+  none =         0ul,        /** None */
+  basic =        1UL << 0,   /** Basic (default) */
+  digest =       1UL << 1,   /** Digest */
+  negotiate =    1UL << 2,   /** Negotiate (SPNEGO) */
+  gssnegotiate = negotiate,  /** GSS-Negotiate */
+  gssapi =       negotiate,  /** GSS-Negoatiate */
+  ntlm =         1UL << 3,   /** NTLM */
+  digest_ie =    1UL << 4,   /** Digest with IE flavour */
+  ntlm_WB =      1UL << 5,   /** NTML delegated to winbind helper */
```
n8sh: Should it be `ntlm_wb` for consistency?

etc/c/curl.d
```diff
@@ -548,29 +548,36 @@ enum CurlProxy {
...
-  gssnegotiate = 4,  /** GSS-Negotiate */
-  ntlm =         8,  /** NTLM */
-  digest_ie =    16, /** Digest with IE flavour */
-  only =         2_147_483_648, /** used together with a single other
-                                type to force no auth or just that
-                                single type */
-  any = -17,     /* (~CURLAUTH_DIGEST_IE) */  /** all fine types set */
-  anysafe = -18  /* (~(CURLAUTH_BASIC|CURLAUTH_DIGEST_IE)) */ ///
+enum CurlAuth : ulong {
+  none =         0ul,        /** None */
```
RazvanN7, line 552:
```
etc/c/curl.d(552): Error: lower case integer suffix 'l' is not allowed. Please use 'L' instead
```

RazvanN7:
@John-Colvin fix the error on line 552 then we can get this in.

RazvanN7:
@John-Colvin I fixed it for you. Let's see if we can merge this.

PR #7352, opened 2020-01-13
Fix Issue 20502: Converting std.typecons.RefCounted!T to a string gives T's storage location instead of T's fields when T is a struct without an explicit toString

n8sh:
Example of what this PR fixes:
```d
void main()
{
    import std.typecons : refCounted;
    import std.stdio : writeln;

    writeln(123); // "123"
    writeln(refCounted(123)); // "123"
    
    struct A { string toString() { return "a"; } }
    writeln(A.init); // "a"
    writeln(refCounted(A.init)); // "a"

    // This one doesn't match:
    struct B { int x; }
    writeln(B(123)); // "B(123)"
    writeln(refCounted(B(123))); // "RefCounted!(B, cast(RefCountedAutoInitialize)0)(RefCountedStore(558C7FFBC660))"
}
```

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -6640,6 +6640,15 @@ refCountedStore.ensureInitialized). Otherwise, just issues $(D
 assert(refCountedStore.isInitialized)).
  */
     alias refCountedPayload this;
+
+    static if (is(T == struct) && !is(typeof((ref T t) => t.toString())))
+    {
+        string toString(this This)()
+        {
+            import std.conv : to;
+            return to!string(refCountedPayload);
+        }
+    }
 }
 
 ///
@@ -6767,6 +6776,24 @@ pure @system unittest
     }
     auto rc = RefCounted!(NoDefaultCtor, RefCountedAutoInitialize.no)(5);
     assert(rc.x == 5);
+}
+
+// https://issues.dlang.org/show_bug.cgi?id=20502
+@system unittest
+{
+    import std.conv : to;
+    // Check that string conversion is transparent for refcounted
+    // structs that do not have either toString or alias this.
+    struct A { Object a; }
+    auto a  = A(new Object());
+    auto r = refCounted(a);
+    assert(to!string(r) == to!string(a));
+    assert(to!string(cast(const) r) == to!string(cast(const) a));
+    // Check that string conversion is still transparent for refcounted
+    // structs that have alias this.
+    struct B { int b; alias b this; }
+    struct C { B b; alias b this; }
+    assert(to!string(refCounted(C(B(123)))) == to!string(C(B(123))));
 }
 
 /**


MoonlightSentinel:
Compiling with DMD master outputs:
```
123
123
a
a
B(123)
B(123)
```

Is there a mistake in your example?

n8sh:
@MoonlightSentinel I've verified this is still happening with freshly-compiled DMD master with that example verbatim. Pasting the example into https://run.dlang.io/ and running it also gives that result.

MoonlightSentinel:
Strange, are you on Mac/Posix?

std/typecons.d
```diff
@@ -6434,6 +6434,21 @@ refCountedStore.ensureInitialized). Otherwise, just issues $(D
...
+        {
+            import std.conv : to;
+            return to!string(refCountedPayload);
+        }
+
+        string toString()() const
+        {
+            import std.conv : to;
+            return to!string(refCountedPayload);
+        }
```
atilaneves: You only need one. Please either use `inout` or a templated `this`.

std/typecons.d
```diff
@@ -6544,6 +6559,23 @@ pure @system unittest
     RefCounted!(int*) c;
 }
 
+@system unittest // Issue 20502
```
atilaneves: Is there a reason for this to be `@system` other than the cast?

std/typecons.d
```diff
@@ -6544,6 +6559,23 @@ pure @system unittest
...
+{
+    import std.conv : to;
+    // Check that string conversion is transparent for refcounted
+    // structs that do not have either toString or alias this.
+    struct A { Object a; }
+    auto a  = A(new Object());
+    auto r = RefCounted!A(a);
```
atilaneves: Why not `auto r = refCounted(a)`?

std/typecons.d
n8sh: RefCounted in  general is `@system`.

std/typecons.d
n8sh: `inout` doesn't work here (the new unittest will fail) but templated `this` is exactly what I'm looking for.

n8sh:
@MoonlightSentinel I've tested this on Windows 10 using dmd v2.090.0. Maybe you had this branch checked out and dmd used it when you ran the example so the fix was already applied.

MoonlightSentinel:
> @MoonlightSentinel I've tested this on Windows 10 using dmd v2.090.0. Maybe you had this branch checked out and dmd used it when you ran the example so the fix was already applied.

I somehow missed that the above behaviour was *fixed*  and thought it was still present. Sorry for the noise

std/typecons.d
```diff
@@ -6544,6 +6553,23 @@ pure @system unittest
...
+    struct A { Object a; }
+    auto a  = A(new Object());
+    auto r = refCounted(a);
+    assert(to!string(r) == to!string(a));
+    assert(to!string(cast(const) r) == to!string(cast(const) a));
+    // Check that string conversion is still transparent for refcounted
+    // structs that have alias this.
+    struct B { int b; alias b this; }
+    struct C { B b; alias b this; }
+    assert(to!string(refCounted(C(B(123)))) == to!string(C(B(123))));
```
MoonlightSentinel: Maybe add tests for `const/immutable RefCounted` (and maybe other storage classes as well)

berni44:
@RazvanN7 Browsing through old issues I came across this PR. Seems to be one of the forgotten PRs, that could be merged (after fixing the trivial merge conflict). In my opinion the request of @MoonlightSentinel for `const/immutable RefCounted` is an enhancement, not a must have (don't let the best be the enemy of the better).

RazvanN7 (approved):
LGTM. Please fix the merge conflict and we're good to go.

std/typecons.d
```diff
@@ -6434,6 +6434,15 @@ refCountedStore.ensureInitialized). Otherwise, just issues $(D
 assert(refCountedStore.isInitialized)).
  */
     alias refCountedPayload this;
+
+    static if (is(T == struct) && !is(typeof((T t) => t.toString())))
```
n8sh, line 6438:
```suggestion
    static if (is(T == struct) && !is(typeof((ref T t) => t.toString())))
```

PR #7353, opened 2020-01-15
Replaced version (unittest) with version (StdUnittest) to avoid unnecessary overhead when compiling with -unittest

alexandrumc:
Related to https://github.com/dlang/druntime/pull/2888

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -150,7 +150,7 @@ ifdef ENABLE_COVERAGE
 override DFLAGS  += -cov
 endif
 
-UDFLAGS=-unittest
+UDFLAGS=-unittest -version=StdUnittest -version=CoreUnittest
 
 # Set DOTOBJ and DOTEXE
 ifeq (,$(findstring win,$(OS)))
--- a/std/conv.d
+++ b/std/conv.d
@@ -1700,77 +1700,74 @@ if (!isImplicitlyConvertible!(S, T) && isAssociativeArray!S &&
     assert(a.to!(const(int)[int]).byPair.equal(a.byPair));
 }
 
-version (unittest)
-private void testIntegralToFloating(Integral, Floating)()
-{
-    Integral a = 42;
-    auto b = to!Floating(a);
-    assert(a == b);
-    assert(a == to!Integral(b));
-}
-
-version (unittest)
-private void testFloatingToIntegral(Floating, Integral)()
-{
-    bool convFails(Source, Target, E)(Source src)
-    {
-        try
-            cast(void) to!Target(src);
-        catch (E)
-            return true;
-        return false;
-    }
-
-    // convert some value
-    Floating a = 4.2e1;
-    auto b = to!Integral(a);
-    assert(is(typeof(b) == Integral) && b == 42);
-    // convert some negative value (if applicable)
-    a = -4.2e1;
-    static if (Integral.min < 0)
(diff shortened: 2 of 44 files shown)


CyberShadow:
Won't this break compiling user programs with `-unittest`? Meaning, due to `-unittest`, DMD will try to parse Phobos unittests too, which will now fail due to required definitions being versioned out?

alexandrumc:
> Won't this break compiling user programs with `-unittest`? Meaning, due to `-unittest`, DMD will try to parse Phobos unittests too, which will now fail due to required definitions being versioned out?

No, because parsing doesn't have anything to do with valid definitions. Semantic analysis checks if the referred symbols are defined or not, but it is not done on the `unittest {}` blocks placed in libraries.

CyberShadow:
> No, because parsing doesn't have anything to do with valid definitions.

Sorry for using the wrong term :)

> Semantic analysis checks if the referred symbols are defined or not, but it is not done on the `unittest {}` blocks placed in libraries.

Meaning, DMD will not look inside `unittest` blocks for files that are not on its command line, even with `-unittest`. I see.

Then, the question is, why not make `version(unittest)` behave the same way as `unittest` blocks? I.e. make the compiler ignore them if the file is not on its command line.

alexandrumc:
> Then, the question is, why not make `version(unittest)` behave the same way as `unittest` blocks? I.e. make the compiler ignore them if the file is not on its command line.

Sure, a design decision must be made. However, we have to take into account the fact that `version {}` and `unittest {}` are different mechanisms, and ignoring version(unittest) might be only an unwanted particular case.

alexandrumc:
@atilaneves

posix.mak
```diff
@@ -150,7 +150,7 @@ ifdef ENABLE_COVERAGE
 override DFLAGS  += -cov
 endif
 
-UDFLAGS=-unittest
+UDFLAGS=-unittest -version=StdUnittest -version=CoreUnittest
```
MoonlightSentinel: Why include druntime's unittests with `CoreUnittest`?

alexandrumc: Phobos uses https://github.com/dlang/druntime/blob/9baf9735c923a9bbb72b8e4ee05f513ea0499bbc/src/core/time.d#L120-L123 to run its unittests. But that `version (unittest)` will change into `version (CoreUnittest)` (https://github.com/dlang/druntime/pull/2902), so the flag has to be there.

MoonlightSentinel: Right, -`unittest` cascades into druntime. Thanks for the explaination.

std/conv.d
```diff
@@ -5687,10 +5688,6 @@ pure nothrow @safe /* @nogc */ unittest
 
 @system unittest
 {
-    int var = 6;
```
atilaneves: Why are these lines deleted?

alexandrumc: Because I moved the content of a `version (unittest)` inside a `unittest {}`, so I had to move those lines too. They are now at line 5009.

posix.mak
PetarKirov: Should now be not necessary with my changes to https://github.com/dlang/druntime/pull/2902.

PetarKirov: Addressed in https://github.com/dlang/phobos/pull/7356.

PR #7354, opened 2020-01-16
Fix issue 20511: make toJSON determine safeness based on output range safeness.

FeepingCreature:
Works around DMD being unable to infer the recursive toJson call as @safe.


FeepingCreature:
Pretty sure the auto-tester failure is sporadic.

PR #7355, opened 2020-01-19
Add initial support for iOS, tvOS and watchOS

jacob-carlborg:
I've only tested this on a 64 bit iPhone running iOS 12. 32 bit is obsolete and I don't have access to the other platforms.

* `std.net.curl` is disabled, libcurl is not shipped on these platforms. This should possible to fix by static linking libcurl and shipping libcurl together with the D compilers. But for that, some extra work are required due to `dlopen` not being allowed.

* Most of the functionality in `std.process` is disabled. Creating new processes are not allowed.

* Some of the unit tests in `std.experimental.allocator` fail. Perhaps the https://github.com/dlang/phobos/commit/537cef3c792f797f68bf75db12709d9ca87387d2 commit should be removed since it doesn't fix the unit test anyway
* Some math related unit tests fail
* One test in `std.socket` fail, but that is allowed to fail

The rest of the unit tests pass.

I recommend reviewing the commits one by one. The individual commit messages contain information of why something was changed. The diff for `std.process` is very big, but the only thing I did was to move `environment`, `thisThreadID` and `thisProcessID` to the top of the module. This allows to easily version out the rest.

druntime PR: https://github.com/dlang/druntime/pull/2910.
LDC PR: https://github.com/ldc-developers/ldc/pull/3288.

### DIFF
--- a/std/experimental/allocator/building_blocks/allocator_list.d
+++ b/std/experimental/allocator/building_blocks/allocator_list.d
@@ -4,6 +4,8 @@ Source: $(PHOBOSSRC std/experimental/allocator/building_blocks/allocator_list.d)
 */
 module std.experimental.allocator.building_blocks.allocator_list;
 
+import core.memory : minimumPageSize;
+
 import std.experimental.allocator.building_blocks.null_allocator;
 import std.experimental.allocator.common;
 import std.experimental.allocator.gc_allocator;
@@ -823,7 +825,7 @@ version (Posix) @system unittest
     import std.algorithm.comparison : max;
     import std.typecons : Ternary;
 
-    enum pageSize = 4096;
+    enum pageSize = minimumPageSize;
 
     static void testrw(void[] b)
     {
@@ -883,7 +885,7 @@ version (Posix) @system unittest
     import std.algorithm.comparison : max;
     import std.typecons : Ternary;
 
-    enum pageSize = 4096;
+    enum pageSize = minimumPageSize;
 
     static void testrw(void[] b)
     {
@@ -950,7 +952,7 @@ version (Posix) @system unittest
     import std.algorithm.comparison : max;
     import std.typecons : Ternary;
 
-    enum pageSize = 4096;
+    enum pageSize = minimumPageSize;
 
     static void testrw(void[] b)
     {
@@ -966,15 +968,15 @@ version (Posix) @system unittest
     AllocatorList!((n) => AscendingPageAllocator(max(n, numPages * pageSize)), NullAllocator) a;
     auto b = a.alignedAllocate(1, pageSize * 2);
     assert(b.length == 1);
-    assert(a.expand(b, 4095));
-    assert(b.ptr.alignedAt(2 * 4096));
-    assert(b.length == 4096);
+    assert(a.expand(b, pageSize - 1));
+    assert(b.ptr.alignedAt(2 * pageSize));
+    assert(b.length == pageSize);
 
-    b = a.allocate(4096);
-    assert(b.length == 4096);
+    b = a.allocate(pageSize);
+    assert(b.length == pageSize);
     assert(a.allocators.length == 1);
 
-    assert(a.allocate(4096 * 5).length == 4096 * 5);
+    assert(a.allocate(pageSize * 5).length == pageSize * 5);
     assert(a.allocators.length == 2);
 
     assert(a.deallocateAll());
--- a/std/parallelism.d
+++ b/std/parallelism.d
@@ -40,6 +40,15 @@ License:    $(HTTP boost.org/LICENSE_1_0.txt, Boost License 1.0)
 */
 module std.parallelism;
 
+version (OSX)
+    version = Darwin;
+else version (iOS)
+    version = Darwin;
+else version (TVOS)
+    version = Darwin;
+else version (WatchOS)
+    version = Darwin;
+
 ///
 @system unittest
 {
@@ -86,7 +95,7 @@ import std.meta;
 import std.range.primitives;
 import std.traits;
 
-version (OSX)
+version (Darwin)
 {
     version = useSysctlbyname;
 }
@@ -986,9 +995,9 @@ uint totalCPUsImpl() @nogc nothrow @trusted
     }
     else version (useSysctlbyname)
     {
-        version (OSX)
+        version (Darwin)
         {
-            auto nameStr = "machdep.cpu.core_count\0".ptr;
+            enum nameStr = "hw.physicalcpu";
         }
         else version (FreeBSD)
         {
--- a/std/socket.d
+++ b/std/socket.d
@@ -2461,12 +2461,21 @@ public:
 
 @safe unittest
 {
-    softUnittest({
+    version (iOSDerived)
+    {
+        enum PAIRS = 256;
+        enum LIMIT = 1024;
+    }
+    else
+    {
         enum PAIRS = 768;
+        enum LIMIT = 2048;
+    }
+
+    softUnittest({
         version (Posix)
         () @trusted
         {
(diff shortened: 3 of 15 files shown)


jacob-carlborg:
Anyone wants to have a look at this PR?

std/parallelism.d
```diff
@@ -986,9 +995,9 @@ uint totalCPUsImpl() @nogc nothrow @trusted
     }
     else version (useSysctlbyname)
     {
-        version (OSX)
+        version (Darwin)
         {
-            auto nameStr = "machdep.cpu.core_count\0".ptr;
+            enum nameStr = "hw.physicalcpu";
```
CyberShadow: Wouldn't `hw.logicalcpu` (and, previously, `machdep.cpu.thread_count`) be more consistent with the other values? The goal of `totalCPUs` is to return a good number for how many threads to create, and the number of logical CPUs is likely a better fit than the number of physical CPUs.

std/socket.d
```diff
@@ -2041,14 +2041,23 @@ static if (is(sockaddr_un))
     @safe unittest
     {
         import core.stdc.stdio : remove;
-        import std.file : deleteme;
+
+        static string deleteme()
```
CyberShadow: A short comment explaining the need for this local definition would be nice.

CyberShadow:
```
std/file.d(139): Error: `std.process.thisProcessID` called with argument types `()` matches both:
std/process.d(195):     `std.process.thisProcessID()`
and:
std/process.d(678):     `std.process.thisProcessID()`
```

std/parallelism.d
jacob-carlborg: I didn't want to change the semantics.

jacob-carlborg:
> Are these security limitations non-negotiable, e.g. would they still hold on a jailbroken device? If not, perhaps it would make sense to leave the code but only disable the tests.

I don't know. I don't plan to jailbreak my phone to find out. The APIs are available on iOS but not on watchOS or tvOS.

> The std.process diff is unfortunately not very readable. I'm guessing things were moved in order to section off a part of definitions with version(...): or equivalent.

Yes, it's in the PR description.

std/socket.d
jacob-carlborg: Done. It's in the commit message as well.

jacob-carlborg:
Why does this fail: https://circleci.com/gh/dlang/phobos/11297?

std/experimental/allocator/building_blocks/allocator_list.d
```diff
@@ -823,7 +830,10 @@ version (Posix) @system unittest
     import std.algorithm.comparison : max;
     import std.typecons : Ternary;
 
-    enum pageSize = 4096;
+    version (AppleARM)
+        enum pageSize = 16384;
+    else
+        enum pageSize = 4096;
```
wilzbach: Maybe define this at the top of the module for to avoid repetition?

wilzbach:
I can't see the CircleCi failure at the moment, but buildkite fails because the public tests must be runnable outside of the module.

jacob-carlborg:
> I can't see the CircleCi failure at the moment, but buildkite fails because the public tests must be runnable outside of the module.

Why? Isn't half of the point of having the unit tests inside the same module as the implementation that it can access non-public symbols?

wilzbach:
> Why? Isn't half of the point of having the unit tests inside the same module as the implementation that it can access non-public symbols?

Public unit tests are displayed in the documentation. Users must be able to execute the examples.

jacob-carlborg:
> Public unit tests are displayed in the documentation

What do you suggest then? Using `__traits` to bypass the protection or duplicate the unit test for iOS and and remove `///`? None of those alternatives seem ideal.

std/experimental/allocator/building_blocks/allocator_list.d
jacob-carlborg: Done.

wilzbach:
What about making the symbol public?

jacob-carlborg:
> What about making the symbol public?

That's possible, but it doesn't seem ideal either.

Perhaps come up with a test that doesn't depend on these platform specific details. But on the other hand, `AscendingPageAllocator` seems to be very platform specific.

jacob-carlborg:
`pageSize` and `minimumPageSize` have been made available in druntime. Using those now.

etcimon:
This commit fixes the socket tests on iOS

https://github.com/etcimon/phobos/commit/984dbd18e75eebeccb7fb369861f554f98279d92

jacob-carlborg:
> This commit fixes the socket tests on iOS

Awesome, thanks 👍. 

> The reason for failing was that the number of sockets monitored by select exceeded the FD_MAXSET limit on iOS of 1024

Hmm, ok. How is that possible? What other sockets were open? Something in Phobos that is not closed ?

etcimon:
> Hmm, ok. How is that possible? What other sockets were open? Something in Phobos that is not closed ?

The unittest tries to open `PAIRS*2` file descriptors to run a `.select` on them, which basically waits for one of those sockets to trigger an event. `PAIRS*2` was `1533` or so, but on iOS, there's a limit of 1024 open file descriptors, thus the `INVALID_ARGUMENT` error that was returned.

std/zip.d
```diff
@@ -1700,10 +1709,8 @@ the quick brown fox jumps over the lazy dog\r
     assert(za.directory.length == 0);
 }
 
-// Non-Android POSIX-only, because we can't rely on the unzip command being
-// available on Android or Windows
-version (Android) {} else
-version (Posix) @system unittest
+version(HasUnzip)
```
Geod24:
```suggestion
version (HasUnzip)
```

The linter complains about this

jacob-carlborg:
dlang.phobos fails with:

```
/d/a/1/dmd/.azure-pipelines/lib.sh: line 37: HOST_DMD_VERSION: unbound variable
```

std/zip.d
jacob-carlborg: Done.

Geod24:
@jacob-carlborg : Should be fixed now, just rebase on master

jacob-carlborg:
`dlang.phobos Windows_VisualD win32-ldc` is still failing.

Geod24:
You didn't rebase (or didn't push, at least).

jacob-carlborg:
> You didn't rebase (or didn't push, at least).

Ah, I missed the second part of your message.

Done. Let's see what happens now.

Geod24:
```
dlp 0.1.0: building configuration "dlp-test-application"...
--
  | Linking...
  | Running ./dlp-test-application
  | vendor/dmd/src/dmd/gluelayer.d(47): [unittest] Assertion failure
  | core.exception.AssertError@vendor/dmd/src/dmd/gluelayer.d(47): Assertion failure
```

That failure looks legit. Everything else is green or should be.

jacob-carlborg:
> That failure looks legit. Everything else is green or should be.

I don't know why it would start to fail now. But I updated to a later version of the DMD frontend (which doesn't have this assert 😃). It passes now.

But now vibe.d was failing. Might be due to out of memory. I'm retrying that test now.

jacob-carlborg:
Everything is green now. Let's get this merged sooner rather than later. The changes in `std.process` are quite prone to cause merge conflicts.

jacob-carlborg:
@Geod24 ping.

Geod24:
I have no power here. Let's summon @thewilsonator .

thewilsonator:
Well I can't be arsed reviewing that, fortunately it seems @burner already has.

Geod24:
Congrats @jacob-carlborg !

jacob-carlborg:
> Congrats

Thanks 😃.

PR #7357, opened 2020-01-21
Fix Issue 20521 - Checking for and getting empty variables on Wine / …

CyberShadow:
…XP fails

Wine-Bug: https://bugs.winehq.org/show_bug.cgi?id=48471


CyberShadow:
There is already a test for this. Of course, we don't run tests under Wine or XP, so this hasn't been caught.

PR #7358, opened 2020-01-22
WIP: Add @nogc to relevant std.algorithm.iteration.map unittests

atilaneves:
Only exists to trigger the autotester.

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -524,14 +524,13 @@ if (fun.length >= 1)
 }
 
 ///
-@safe unittest
+@safe @nogc unittest
 {
     import std.algorithm.comparison : equal;
-    import std.range : chain;
-    int[] arr1 = [ 1, 2, 3, 4 ];
-    int[] arr2 = [ 5, 6 ];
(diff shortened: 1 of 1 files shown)


wilzbach:
This was the first visible public example of the function. It is now vastly less legible for beginners (i.e. the kind of people who would read this example in the first place).

atilaneves:
It was WIP and not meant to be merged, I was going to add @nogc to more tests.

atilaneves:
> This was the first visible public example of the function. It is now vastly less legible for beginners (i.e. the kind of people who would read this example in the first place).

Because of the use of `only`?

atilaneves:
I used `staticArray` in a different PR and hit an assertion in the backend for Windows. I'm not sure that was the cause, so I did a very small change here with `only` to see if it passed CI. I didn't mean for it to get merged.

Would this have looked better?

```d
auto rng1 = only(1, 2, 3, 4);
auto rng2 = only(5, 6);
auto squares = map!(a => a* a)(chain(rng1, rng2));
```

It's less of a change. I'm not sure how much harder it is for beginners given that the original example had `chain` instead of `~`.

I'm open to suggestions.

PR #7359, opened 2020-01-23
fix Issue 20500 - running examples on the home page only displays '1 …

rainers:
…unittests passed'

move unittests out of templated struct FormatSpec so they are not generated when using the template

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -1234,40 +1234,6 @@ if (is(Unqual!Char == Char))
         return false;
     }
 
-    @safe unittest
-    {
-        import std.array;
-        auto w = appender!(char[])();
-        auto f = FormatSpec("abc%sdef%sghi");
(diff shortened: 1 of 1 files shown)


thewilsonator:
I'm not sure why this is mark major, but it is, so should it target stable?

rainers:
> I'm not sure why this is mark major, but it is, so should it target stable?

Do you mean it should target master? I think it's pretty bad that the examples don't work properly on the website, but simple user code as
```
import std.stdio;
void main()
{
	writefln("The answer is %d", 42);
}
```
doesn't run when compiled with -unittest. This doesn't work as advertised (run main if there are no unittests).

thewilsonator:
Derp, ignore that. I can't read.

wilzbach:
The Buildkite failures are unrelated and due to new agents without a full config being recently added.

PR #7360, opened 2020-01-23
Fix Issue 9298 - some std.math functions not implemented for Win64

e10s:
To be precise, for -m32mscoff / -m64.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -4596,33 +4596,23 @@ if (is(typeof(rfunc(F.init)) : F) && isFloatingPoint!F)
  *
  * Unlike the rint functions, nearbyint does not raise the
  * FE_INEXACT exception.
- *
- * Note:
- *     Not implemented for Microsoft C Runtime
  */
 real nearbyint(real x) @safe pure nothrow @nogc
 {
-    version (CRuntime_Microsoft)
-        assert(0, "nearbyintl not implemented in Microsoft C library");
-    else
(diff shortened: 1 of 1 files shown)


PetarKirov:
CC @kinke @rainers

kinke:
Please don't auto-merge - DMD still ships with VS 2010 import libs, which doesn't provide these functions.

e10s:
Hmm, I was under an illusion. I have to test with `-mscrtlib=msvcr100` switch...

This depends on dlang/installer#377?

kinke:
Heh yeah, great that you've found it yourself. :) - That, in combination with a switch to the VS 2015/UCRT libs, should work; but that requires testing and some further adaptations wrt. linker symbol aliases (which would already be in LDC's Phobos and just need to be moved here upstream).

thewilsonator:
Is this good to go?

dukc:
@kinke is this good to go?

kinke:
AFAIK, DMD now defaults to the MinGW-based libs for the VS2013 runtime (which is already a big improvement over the previous VS2010 ones). So someone needs to check whether these functions are available in VS2013.

thewilsonator:
cc @rainers do you know?

rainers:
Finally tested now: std.math unittests in this PR link and run fine with -mscrtlib=msvcrt120 (but not earlier), so should be good to go.

PR #7361, opened 2020-01-23
Fix issue 20527. Set the correct union member for integers so CTFE

schveiguy:
doesn't get confused.


schveiguy:
I can't figure out what buildkite is doing or how to knock it loose.

PR #7363, opened 2020-01-24
Fix Issue 20508 - std.math.pow(-infinity, y) does not return NaN for imaginary or complex results

berni44:
In addition to the fix I changed the public unittests. The current one mostly showed that corner cases work but not the usual usage (which already heavily varies, depending on the signs and integrity of the operands). I removed the corner cases completely from public view, because they are allready listed in a table above the examples and IMHO this doublication isn't useful. (The tests itself are preserved though.)

There are several additional issues burried in this function (including powers of `-infinity`). I started to identify them and I'll eventually report and address them as separate issues.


berni44:
One of the public unittests did not work on windows, I replaced it with an other test (for the public tests, I want to use `==` instead of `isClose`/`approxEqual`, as the conversion to `writeln` makes the result a little bit ugly).

berni44:
+WIP (win64 is good now, but I need to find something, that works on win32 too)

berni44:
looks like this is issue 20451 again...

berni44:
I commented the first unittest out for win32 and freebsd. This works, but now the style checker does not recognize anymore, that there is a public example... For me, good examples are an important thing that I do not want to give up, just because the implementation for win32 is overdoing in accuracy. I could reverse the two public examples. That would work, but again, the documentation would suffer.

So I commented out the style check for public unittests for std.math until either the stylechecker gets smarter or issue 20451 is fixed (or someone has a better idea ;-) ).

If the tests work now, it's not anymore WIP.

std/math.d
```diff
@@ -8281,27 +8292,80 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
     return impl(x, y);
 }
 
+// Due to issue 20451 this doesn't work on Win32 and FreeBSD. It would work
+// using isClose, but the main purpose of this test is to provide a good example,
+// not to test the pow function, and this would lead to a somewhat ugly example.
```
tsbockman: The standard library documentation should not mislead users into thinking that floating-point operations are perfectly precise when they're not. Using `isClose`, `approxEqual`, or `feqrel` is the honest and correct option, even if it is less aesthetically pleasing.

berni44: I'm not sure if you are aware of the transformations on the unittests, that are done by ddoc. Every `assert` is replaced by a `writeln`. But the output of `writeln` isn't compared anymore by a computer (but by the human reading the example), so the problem does not exist there. The strange comparisons can only be seen, when the user looks into the source code. And there they are next to this comment explaining what's going on.

I actually changed the comment and the versioning a little bit, because there might be other OSs which fail. Now this unittest is only used by ddoc. I hope this is sort of a compromise for you.

std/math.d
```diff
@@ -8314,27 +8325,89 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
     return impl(x, y);
 }
 
+/*
+Of course, floating point numbers should be compared using isClose,
+but the main purpose of this test is to provide a good example,
+not to test the pow function, and this would lead to ugly examples like:
```
tsbockman: A good example is one that is appropriate to imitate, not just look at. Users should not be comparing the results of this `pow` function using `==`, so the examples shouldn't either.

As noted in the author's **TODO**, the `exp2(y * log2(x))` method currently used is very imprecise in some parts of its domain, so this isn't just quibbling about the occasional single-bit rounding error. Using `isClose`, `feqrel`, or `approxEqual` serves the practical purpose of warning the user about this, in addition to enabling automatic verification of the example.

std/math.d
```diff
@@ -8314,27 +8325,89 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
...
+
+writeln(isClose(pow(1.5, 10.0), 57.6650390625)); // true
```
wilzbach: It wouldn't add the `true`.

std/math.d
```diff
@@ -8314,27 +8325,89 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
...
+
+instead of
+
+writeln(pow(1.5, 10.0)); // 57.6650390625
+
+So I versioned this unittest out, because on some 32bit computers the tests
+would fail with ==.
```
wilzbach: That's exactly what the dscanner tests tries to prevent people from doing. This will lead to user confusion if they copy this code to user machines and it fails there. Either:
- use is close
- use `version (X86_64)`  inside the test, so that it's obvious to the user what's going on.

wilzbach:
> So I commented out the style check for public unittests for std.math until either the stylechecker gets smarter 

No, this behavior is on purpose and by design as examples shouldn't be written with `version (Ddoc)`. There's no way for us to ensure they actually work.

tsbockman:
Here's an example of a particularly imprecise result from `pow`:
```D
void main()
{
    import std.stdio, std.math;
    
    // Computed in arbitrary precision via WolframAlpha:
    const(real) expected = 1.50354040799963854183e1854L;
    writefln("expected: %.20g", expected);
    
    const(real) got = pow(0.9832834778410573217L, -253259.351392875174952L);
    writefln("got: %.20g", got);
    
    const(int) incorrectBits = real.mant_dig - feqrel(got, expected); // 16
    writeln("number of incorrect bits: ", incorrectBits);
}
```
As a user, if I see official examples for a floating-point math function using `==`, I expect a maximum error of 1 or 2 bits. The above example's 16 bits of error is bad enough to leave 80-bit `real` with worse precision than I would naively expect from 64-bit `double`.

Now floating-point `pow` turns out to be a very difficult function to implement precisely (I've tried), so I'm not saying this to attack the implementation. But, the documentation should reflect the reality that this `pow` is only an approximation.

RazvanN7:
Closing this in favor of: #7747

PR #7364, opened 2020-01-24
Remove 'could not locate' warning from std unittests

berni44:
In debug mode, the standard unittests produces a series of (useless) warnings. With the new feature of -version=StdUnittest, this warnings can be removed without removing them in other uses too.

```
Could not locate matching function for: FuncInfo!("foo", void())
Could not locate matching function for: FuncInfo!("bar", void())
Could not locate matching function for: FuncInfo!("foo", void())
Could not locate matching function for: FuncInfo!("refresh", int())
Could not locate matching function for: FuncInfo!("refresh", int())
Could not locate matching function for: FuncInfo!("refresh", int())
```

### DIFF
--- a/std/experimental/typecons.d
+++ b/std/experimental/typecons.d
@@ -111,12 +111,15 @@ if (Targets.length >= 1 && allSatisfy!(isMutable, Targets))
             else
             {
                 enum foundFunc = findCovariantFunction!(TargetMembers[i], Source, SourceMembers);
-                debug
+
+                version (StdUnittest) {}
+                else debug
                 {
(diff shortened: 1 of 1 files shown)


std/experimental/typecons.d
```diff
@@ -114,8 +114,12 @@ if (Targets.length >= 1 && allSatisfy!(isMutable, Targets))
                 debug
```
MoonlightSentinel: There's no need to evaluate the `static if` for an empty block. You could move the version condition e.g. here.
```suggestion
                version (StdUnittest) {}
                else debug
```

PR #7365, opened 2020-01-24
Fix Issue 20523 - std.zip: Invalid data yields RangeError in ZipArchive constructor

berni44:
This PR does two things: Moving from `uint` to `ulong` for support of files larger then 4GB. And adding access of the records via structs. It was easier to add both at once, because else I would have had to add hundreds of casts just to remove them again in the next PR.

This PR was allready planed for a while and I recognized that it fixes issue 20523 (and several similar ones) for free, because the calls to `removeSegment` are now done before accessing the data and therefore prevent the access of not existing data.


std/zip.d
```diff
@@ -1137,59 +1131,117 @@ public:
...
         }
 
-        enforce!ZipException(endrecOffset != to!uint(data.length),
+        enforce!ZipException(endrecOffset != data.length,
                              "found no valid 'end of central dir record'");
 
         return endrecOffset;
     }
 
+    private void checkForZip64(ulong offset)
```
thewilsonator: This is only used once so please make it a `private bool checkForZip64(ulong offset) const` and assign to `_isZip64` at the use site (assuming it at false by default.

std/zip.d
```diff
@@ -1137,59 +1131,117 @@ public:
...
+
+        de.name = cast(string) getSegment(next + CentralFileHeader.sizeof, namelen);
+        de.extra = getSegment(next + CentralFileHeader.sizeof + namelen, extralen);
+        de.comment = cast(string) getSegment(next + CentralFileHeader.sizeof + namelen + extralen, commentlen);
+
+        extractAndVerifyArchiveMemberData(de);
+
+        _directory[de.name] = de;
+
+        return next + CentralFileHeader.sizeof + namelen + extralen + commentlen;
```
thewilsonator: Please do 
```
next += CentralFileHeader.sizeof;
de.name = ...;
next += namelen;
de.extra =  ...;
next += extralen;
...
return next;
```

std/zip.d
```diff
@@ -1137,59 +1131,117 @@ public:
...
+    }
+
+    void extractAndVerifyArchiveMemberData(ref ArchiveMember de)
```
thewilsonator: `private`

std/zip.d
```diff
@@ -1700,6 +1768,17 @@ the quick brown fox jumps over the lazy dog\r
     assert(za.directory.length == 0);
 }
 
+@system unittest
```
thewilsonator: whats the reason this can't be `@safe`

std/zip.d
```diff
@@ -810,141 +885,60 @@ public:
...
-                                 && directorySizeUlong + directoryOffsetUlong <= i,
+            enforce!ZipException(size <= z64eocdrOffset && next <= z64eocdrOffset && end <= z64eocdrOffset,
                                  "corrupted directory");
 
-            directoryCount = to!uint(totalEntriesUlong);
-            directorySize = to!uint(directorySizeUlong);
-            directoryOffset = to!uint(directoryOffsetUlong);
+            // The extensible data segment is not yet used, but we need to query it to make sure
+            // it doesn't overlap anything else. Might be provided to the user in the future.
+            auto extensibleData = getSegment(fromLE!ulong(z64eocdlr.offset) + Zip64EndOfCentralDirRecord.sizeof,
```
thewilsonator: this doesn't seem used.

std/zip.d
```diff
@@ -810,141 +885,60 @@ public:
...
+                                             fromLE!ulong(z64eocdr.thisSize) - Zip64EndOfCentralDirRecord.sizeof + 12);
```
thewilsonator: what is `12` in this context?

std/zip.d
thewilsonator: Also you could make `next` `ref` and make this `const` and `return de;` to be assigned to `_directory[de.name];` at the call site.

schveiguy:
I can't say I understand much of this. But if I recall correctly, isn't the size limitation inherent in the zlib library? Or is that only for certain interfaces? Been a long time since I did anything with zlib.

berni44:
+WIP - the change to `ulong` did not work well on 32bit machines. I have to rethink, if we should not add support for >4GB files, or only for 64bit machines or find another solution for 32bit machines... Might take a while...

thewilsonator:
You could just template it on the size and default to `size_t`.

std/zip.d
```diff
@@ -435,6 +437,79 @@ private:
     enum zip64EndOfCentralDirLocatorLength = 20;
     enum dataDescriptorLength = 12;
 
+    align (1) // just to make sure there are no gaps
```
MoonlightSentinel: `align` must be declared inside the struct to affect it's members (according to https://dlang.org/spec/attribute.html#align)

std/zip.d
berni44: The `cast` to `void`.

std/zip.d
berni44: Yes, that's why I wrote the comment above - the `getSegment` needs to be called to check for invalid input. @CyberShadow once suggested to make this data available to the user, but I'd like to address this in a separate PR, as it will introduce a new public symbol.

std/zip.d
berni44: For whatever reasons, the designers of the zip format decided to reduce the size of this chunk by 12. From the specs:

```
4.3.14.1 The value stored into the "size of zip64 end of central
directory record" SHOULD be the size of the remaining
record and SHOULD NOT include the leading 12 bytes.
  
Size = SizeOfFixedFields + SizeOfVariableData - 12.
```

thewilsonator: Thanks, could you put that in a comment.

std/zip.d
thewilsonator: Hmm, could you have that as part of the initialiser of `file`?

thewilsonator: Oh I see all the other tests a re like that too, never mind.

berni44:
I added some casts and asserts to address the 32bit problem, but I haven't made up my mind yet if some of the asserts should not actually be enforces. I uploaded it anyway to see if the tests on 32-bit machines work now.

berni44:
-WIP

I added `enforce`s at all places where I had to insert casts. To avoid, that these `enforce`s are evaluated on 64bit machines, I put them into `static if`s. IMHO everything is working now and I addressed all your concernes.

std/zip.d
```diff
@@ -242,7 +242,7 @@ final class ArchiveMember
     @property @safe void expandedData(ubyte[] ed)
     {
         _expandedData = ed;
-        _expandedSize  = to!uint(_expandedData.length);
+        _expandedSize  = to!ulong(_expandedData.length);
```
n8sh: Future-proofing for 128 bit size_t?

std/zip.d
```diff
@@ -365,8 +365,8 @@ final class ArchiveMember
      *
      * Returns: The index.
      */
-    @property uint index(uint value) @safe pure nothrow @nogc { return _index = value; }
-    @property uint index() const @safe pure nothrow @nogc { return _index; } /// ditto
+    @property ulong index(uint value) @safe pure nothrow @nogc { return _index = value; }
```
n8sh:
```suggestion
    @property ulong index(ulong value) @safe pure nothrow @nogc { return _index = value; }
```

std/zip.d
```diff
@@ -637,39 +703,48 @@ public:
...
-        uint dataSize = archiveSize + directorySize + endOfCentralDirLength + cast(uint) comment.length;
+        ulong dataSize = archiveSize + directorySize + EndOfCentralDirRecord.sizeof + cast(uint) comment.length;
         if (isZip64)
-            dataSize += zip64EndOfCentralDirLocatorLength + zip64EndOfCentralDirLength;
+            dataSize += Zip64EndOfCentralDirLocatorRecord.sizeof + Zip64EndOfCentralDirRecord.sizeof;
 
-        _data = uninitializedArray!(ubyte[])(dataSize);
+        static if (size_t.sizeof < ulong.sizeof)
+        {
+            enforce(dataSize <= size_t.max, "cannot handle zip archives larger than 4 GB on 32bit computers");
```
n8sh:
```suggestion
            enforce!ZipException(dataSize <= size_t.max, "cannot handle zip archives larger than 4 GB on 32bit computers");
```

std/zip.d
```diff
@@ -684,17 +759,28 @@ public:
...
 
             _data[i .. i + de.name.length] = (de.name.representation)[];
             i += de.name.length;
             _data[i .. i + de.extra.length] = (cast(ubyte[]) de.extra)[];
             i += de.extra.length;
-            _data[i .. i + de.compressedSize] = de.compressedData[];
+
+            static if (size_t.sizeof < ulong.sizeof)
+            {
+                enforce(i + de.compressedSize <= size_t.max,
```
n8sh:
```suggestion
                enforce!ZipException(i + de.compressedSize <= size_t.max,
```

std/zip.d
```diff
@@ -1210,19 +1302,22 @@ public:
         uint namelen;
         uint extralen;
 
-        enforce!ZipException(_data[de.offset .. de.offset + 4] == localFileHeaderSignature,
+        enforce(de.offset + 30 <= size_t.max,
```
n8sh:
```suggestion
        enforce!ZipException(de.offset + 30 <= size_t.max,
```

std/zip.d
```diff
@@ -1430,6 +1533,35 @@ private:
...
         }
     }
+
+    auto getSegment(ulong start, ulong length) pure @safe
+    {
+        removeSegment(start, start + length);
+
+        static if (size_t.sizeof < ulong.sizeof)
+        {
+            enforce(start + length <= size_t.max,
```
n8sh:
```suggestion
            enforce!ZipException(start + length <= size_t.max,
```

std/zip.d
```diff
@@ -1245,7 +1340,19 @@ public:
                 // It has the effect of not requiring the 2 byte header
                 // and 4 byte trailer.
                 import std.zlib : uncompress;
-                de._expandedData = cast(ubyte[]) uncompress(cast(void[]) de.compressedData, de.expandedSize, -15);
+
+                static if (size_t.sizeof < ulong.sizeof)
+                {
+                    enforce(de.expandedSize <= size_t.max,
```
n8sh:
```suggestion
                    enforce!ZipException(de.expandedSize <= size_t.max,
```

std/zip.d
UplinkCoder: I can only assume this is a joke

schveiguy: It was probably part of a global search/replace.

n8sh: Whatever I was thinking when I wrote the comment, it can be dismissed. I am leaving this conversation marked "resolved" rather than deleting the original comment outright because I think it would be poor form to delete a comment that had been replied to.

schveiguy: I think it's valid to question this, because technically the `to` call does nothing except return its input. My point was simply that it's likely @berni44 just did a global search and replace, and this one got caught. I think you can remove the `to` call.

berni44:
This is an old PR of mine, which I'm not going to handle in the near future, so I'm closing it - it's still accessible from the issue and I dislike having stalled PRs hanging around... Maybe there will be a time, when I'm coming back on `std.zip`.

PR #7362, opened 2020-01-23
std.math: Hide IEEE flags and FloatingPointControl for soft-float targets

denizzzka:
Cherry-pick of this PR: https://github.com/ldc-developers/phobos/pull/66

(Will be squashed before merge.)

PR #7366, opened 2020-01-24
std.math: Follow-up to #7362

kinke:
Get rid of redundant `version` conditions and don't disable whole unittest blocks (like the extensive `exp()` tests) just because IEEE flags are tested too.


std/math.d
```diff
@@ -2449,17 +2461,17 @@ version (IeeeFlagsSupport) @safe @nogc nothrow unittest
...
+        version (IeeeFlagsSupport) assert(f.flags == 0);
 
-        resetIeeeFlags();
+        version (IeeeFlagsSupport) resetIeeeFlags();
         x = exp(-T.nan);
-        f = ieeeFlags;
+        version (IeeeFlagsSupport) f = ieeeFlags;
         assert(isIdentical(abs(x), T.nan));
-        assert(f.flags == 0);
+        version (IeeeFlagsSupport) assert(f.flags == 0);
```
thewilsonator: Just have two version block. I don't think its very nice to read with every other statement versioned.

PR #7367, opened 2020-01-24
Deprecate std.xml

berni44:
See [forum](https://forum.dlang.org/thread/giykzxblfcgaavvndpbn@forum.dlang.org).


berni44:
I added a release note.

12345swordy:
@thewilsonator is there any way you can contact @andralex or @atilaneves on this?

PR #7368, opened 2020-01-24
std.algorithm.sorting: Enable binary transform function for schwartzSort

kinke:
Receiving the element's index in the original range as 2nd parameter. I have a use case for this.


kinke:
Done.

PR #7370, opened 2020-01-27
adding assert messages to rbtree

burner:


burner:
thank you @berni44 @n8sh

PR #7372, opened 2020-01-27
adding assert messages to std.conv

burner:


std/conv.d
```diff
@@ -1469,7 +1469,8 @@ if (!isImplicitlyConvertible!(S, T) &&
         }
         else
         {
-            static assert(tSmallest < 0);
+            static assert(tSmallest < 0,
+                "minimum value of T most be smaller than 0");
```
berni44: most => must

burner: thank you

std/conv.d
```diff
@@ -6328,7 +6330,7 @@ if ((radix == 2 || radix == 8 || radix == 10 || radix == 16) &&
         else static if (radix == 16)
             enum SHIFT = 4;
         else
-            static assert(0);
+            static assert(false, "radix must be 2, 8, or 16");
```
berni44: could also be 10...

burner: not in that branch, I believe.

berni44: ah ok, I see

std/conv.d
```diff
@@ -5813,14 +5823,6 @@ if (isIntegral!T)
...
 }
 
-auto unsigned(T)(T x)
-if (isSomeChar!T)
-{
-    // All characters are unsigned
-    static assert(T.min == 0);
-    return cast(Unqual!T) x;
-}
-
```
n8sh: Why is this being removed?

burner: its not it was moved for the /// Ditto

std/conv.d
n8sh:
```suggestion
            static assert(false, "radix must be 2, 8, 10, or 16");
```

The radix 10 case is earlier in this function.

burner:
thank you @berni44 @n8sh

PR #7373, opened 2020-01-27
adding assert messages to std.datetime.package

burner:


burner:
thank you

burner:
ldc buildkite failed with: virtual memory exhausted: Cannot allocate memory

berni44:
> ldc buildkite failed with: virtual memory exhausted: Cannot allocate memory

There are a lot of such false positives around currently...

burner:
@berni44 @thewilsonator  thank you

PR #7377, opened 2020-01-27
Fix std.format FP printing for soft-float targets

kinke:
After #7362.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -2658,26 +2658,30 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     {
         static if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.mant_dig))
         {
-            import std.math : FloatingPointControl;
+            import std.math;
 
             auto mode = RoundingMode.toNearestTiesToEven;
-            switch (FloatingPointControl.rounding)
-            {
-            case FloatingPointControl.roundUp:
-                mode = RoundingMode.up;
-                break;
-            case FloatingPointControl.roundDown:
-                mode = RoundingMode.down;
-                break;
-            case FloatingPointControl.roundToZero:
-                mode = RoundingMode.toZero;
-                break;
-            case FloatingPointControl.roundToNearest:
-                mode = RoundingMode.toNearestTiesToEven;
-                break;
-            default: assert(false);
-            }
 
+            // std.math's FloatingPointControl isn't available on all target platforms
+            static if (is(FloatingPointControl))
+            {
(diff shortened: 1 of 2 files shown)


berni44:
Instead of importing `std.math` completely, wouldn't it be better to have something like `enum hasFloatingPointControlSupport` in `std.math` import this and in case of it being `true` doing `import std.math : FloatingPointControl`?

kinke:
The local import only affects the little `then` branch of the `if`...

berni44:
That "little `then` branch" will grow with PRs #7264 and #7318 and the followups and eventually end in being the whole template. Anyway, it's still local of course and I personally would not mind. But as far as I know, we are currently trying to reduce the imports as much as possible, don't we? I think, some modules are even checked for whole imports by the style checker and would be rejected...

kinke:
AFAIK, the only difference for a whole import here is the namespace pollution, but as long as that's restricted to a little `then` (or in the future, possibly its own little `getRoundingMode()`), I see no problem with it, and would find a
```D
import std.math : isFloatingPointControlAvailable;
static if (isFloatingPointControlAvailable)
{
    import std.math : FloatingPointControl;
    ...
}
```
much uglier.

berni44:
For me both is fine.

std/format.d
```diff
@@ -2658,27 +2658,31 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
...
         static if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.mant_dig))
         {
-            import std.math : FloatingPointControl;
+            import std.math;
 
             auto mode = RoundingMode.toNearestTiesToEven;
-            switch (FloatingPointControl.rounding)
+
+            // std.math's FloatingPointControl isn't available on all target platforms
+            static if (is(typeof(FloatingPointControl)))
```
n8sh: This needs to be `is(FloatingPointControl)` not `is(typeof(FloatingPointControl))`.

kinke: Thx, fixed.

n8sh:
>std/format.d(2661:20)[warn]: Local imports should specify the symbols being imported to avoid hiding local symbols.

Yeah maybe this wasn't a good rule since it defeats this kind of compile time introspection.

wilzbach:
> Yeah maybe this wasn't a good rule since it defeats this kind of compile time introspection.

The rules can be selectively disabled in .dscanner.ini

kinke:
Does the bot ignore the 72h... label?

wilzbach:
Yes, the label is just for humans.

PR #7380, opened 2020-01-28
Fix Issue 20542 - std.math.nextafter(NaN, y) and nextafter(x, NaN) sh…

e10s:
…ould return NaN

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -7731,8 +7731,14 @@ T nextafter(T)(const T x, const T y) @safe pure nothrow @nogc
     enum real c = 3;
     static assert(is(typeof(nextafter(c, c)) == real));
     //static assert(nextafter(c, c.infinity) > c);
-    static assert(isNaN(nextafter(c, c.nan)));
-    static assert(isNaN(nextafter(c.nan, c)));
+
+    enum real negZero = nextafter(+0.0L, -0.0L); // specially CTFEable
+    static assert(negZero == -0.0L);
+    static assert(signbit(negZero));
+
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -7703,14 +7719,20 @@ T nextafter(T)(const T x, const T y) @safe pure nothrow @nogc
...
     enum double b = 2;
     static assert(is(typeof(nextafter(b, b)) == double));
     static assert(nextafter(b, b.infinity) > b);
+    static assert(isNaN(nextafter(b, b.nan)));
+    static assert(isNaN(nextafter(b.nan, b)));
 
-    //enum real c = 3;
-    //static assert(is(typeof(nextafter(c, c)) == real));
+    enum real c = 3;
+    static assert(is(typeof(nextafter(c, c)) == real));
```
berni44: When I read #7176 correctly, this unittest was commented out on purpose by @n8sh, as it might not work on every real implementation yet.

e10s: It doesn't matter in these special cases as long as `realX == realY` and `isNaN(realVal)` work properly at compile time. #7176 just mentions `nextUp` / `nextDown` for `real`, called if both of the arguments are different and not NaN.

thewilsonator:
Sorry there has been some network problems, could you please force-push to restart the tests, thanks.

std/math.d
n8sh: I didn't need to comment out those two lines since `typeof` does not require the contained expression to run in CTFE.

PR #5710, opened 2017-08-28
Fix Issue 8388 - std.traits.MemberFunctionsTuple doesn't work with constructors or destructors

RazvanN7:
The template MemberFunctionsTuple calls getVirtualFunctions in order to avoid static functions, but it seems that the constructor and destructor are not grabbed (probably because those are not virtual?). In order to fix the issue I added a static if which tests if the name of the function called for is `__ctor` or `__dtor`; if that is the case, then the getOverloads function is called which correctly grabs all the ctor and dtor overloads.

PR #7385, opened 2020-01-29
Fix Issue 8388 - std.traits.MemberFunctionsTuple doesn't work with constructors or destructors

berni44:
It's essentially a copy of #5710 from @RazvanN7 but with the problematic `static this` commented out. I tried to hunt down the problem of this `static this` with dustmite (took two weeks to run) and finally ended up with issue 20529, but as I forgot to include the dependency cycle in the test for dustmite I ended up with a different cycle. Not sure if a fix of 20529 will fix this too or if there is more to discover. 

Anyway, IMHO this `static this` is only a minor point and should not hinder fixing the original bug.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -4585,8 +4585,11 @@ if (is(C == class) || is(C == interface))
                 alias CollectOverloads = AliasSeq!(); // no overloads in this hierarchy
         }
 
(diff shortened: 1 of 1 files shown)


Geod24:
This broke Vibe.d: https://github.com/vibe-d/vibe.d/issues/2438#issuecomment-622462944

PR #7387, opened 2020-01-31
Fix Issue 20548 - Use bit vector instead of bool[] in RandomCover when choices cannot be packed in a single word

n8sh:


std/random.d
```diff
@@ -3023,17 +3032,17 @@ private struct RandomCoverChoices
...
-        c[2] = true;
-        assert(c[0]);
-        assert(!c[1]);
-        assert(c[2]);
-        c[0] = false;
-        c[1] = true;
-        c[2] = false;
-        assert(!c[0]);
-        assert(c[1]);
-        assert(!c[2]);
```
thewilsonator: It is necessary to change this unittest?

n8sh: Removed change.

n8sh:
Taking off auto-merge to use core.bitop bit testing intrinsics.

thewilsonator:
OK, merge it when you're happy.

PR #7392, opened 2020-02-05
Added a chmod example in the description...

mark-summerfield:
I added this because if you search the D docs for `chmod` no results appear which is confusing for Unix/Linux users. This should help with the search and also provide a useful example.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -2204,7 +2204,7 @@ if (isConvertibleToString!R)
 
 /++
     Set the _attributes of the given file.
-    
+
     For example, a programmatic equivalent of Unix's `chmod +x name` to make a file executable is
     `name.setAttributes(name.getAttributes | octal!700)`.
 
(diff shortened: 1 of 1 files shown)


std/file.d
```diff
@@ -2204,6 +2204,9 @@ if (isConvertibleToString!R)
 
 /++
     Set the _attributes of the given file.
+
```
Geod24: You have trailing whitespace here

std/file.d
```diff
@@ -2205,6 +2205,9 @@ if (isConvertibleToString!R)
 /++
     Set the _attributes of the given file.
 
+    For example, a programmatic equivalent of Unix's `chmod +x name` to make a file executable is
```
CyberShadow: Nit, lines should be wrapped to 80 columns when possible.

kubo39:
Why not merged?

PR #7394, opened 2020-02-09
Include filename when CreateProcess fails with file not found

marler8997:
Print a more helpful error message when `spawnProcess` fails on windows.  Instead of:
```
Failed to spawn new process (The system cannot find the file specified.)
```
You now get this:
```
Failed to spawn process 'SOMEPROGRAM' (The system cannot find the file specified.)
```

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -290,7 +290,7 @@ Pid spawnProcess(scope const(char[])[] args,
                  scope const char[] workDir = null)
     @trusted // TODO: Should be @safe
 {
-    version (Windows)    auto  args2 = escapeShellArguments(args);
+    version (Windows)    auto  args2 = WindowsSpawnArgs(escapeShellArguments(args), args[0]);
     else version (Posix) alias args2 = args;
     return spawnProcessImpl(args2, stdin, stdout, stderr, env, config, workDir);
 }
@@ -682,6 +682,14 @@ private Pid spawnProcessImpl(scope const(char[])[] args,
     }
 }
 
+// enables extra arguments for the windows version of spawnProcessImpl
+version (Windows)
+private struct WindowsSpawnArgs
+{
+    const(char)[] commandLine;
+    const(char)[] arg0;
+}
+
 /*
 Implementation of spawnProcess() for Windows.
 
@@ -692,7 +700,7 @@ envz must be a pointer to a block of UTF-16 characters on the form
 "var1=value1\0var2=value2\0...varN=valueN\0\0".
 */
 version (Windows)
-private Pid spawnProcessImpl(scope const(char)[] commandLine,
+private Pid spawnProcessImpl(scope const WindowsSpawnArgs args,
                              File stdin,
                              File stdout,
                              File stderr,
@@ -703,7 +711,7 @@ private Pid spawnProcessImpl(scope const(char)[] commandLine,
 {
     import core.exception : RangeError;
 
-    if (commandLine.empty) throw new RangeError("Command line is empty");
+    if (args.commandLine.empty) throw new RangeError("Command line is empty");
 
     // Prepare environment.
     auto envz = createEnv(env, !(config & Config.newEnv));
@@ -760,10 +768,14 @@ private Pid spawnProcessImpl(scope const(char)[] commandLine,
         CREATE_UNICODE_ENVIRONMENT |
         ((config & Config.suppressConsole) ? CREATE_NO_WINDOW : 0);
     auto pworkDir = workDir.tempCStringW();     // workaround until Bugzilla 14696 is fixed
-    if (!CreateProcessW(null, commandLine.tempCStringW().buffPtr,
+    if (!CreateProcessW(null, args.commandLine.tempCStringW().buffPtr,
                         null, null, true, dwCreationFlags,
                         envz, workDir.length ? pworkDir : null, &startinfo, &pi))
+    {
+        if (args.arg0)
+            throw ProcessException.newFromLastError("Failed to spawn process '" ~ cast(string) args.arg0 ~ "'");
         throw ProcessException.newFromLastError("Failed to spawn new process");
+    }
 
     // figure out if we should close any of the streams
     if (!(config & Config.retainStdin ) && stdinFD  > STDERR_FILENO
@@ -1309,8 +1321,8 @@ Pid spawnShell(scope const(char)[] command,
         // It does not use CommandLineToArgvW.
         // Instead, it treats the first and last quote specially.
         // See CMD.EXE /? for details.
-        auto args = escapeShellFileName(shellPath)
-                    ~ ` ` ~ shellSwitch ~ ` "` ~ command ~ `"`;
+        auto args = WindowsSpawnArgs(escapeShellFileName(shellPath)
+                    ~ ` ` ~ shellSwitch ~ ` "` ~ command ~ `"`);
     }
     else version (Posix)
     {


marler8997:
unittests for the `peelApplicationName` function or for the CreateProcess part as well?

thewilsonator:
Both.

std/process.d
```diff
@@ -2642,15 +2642,45 @@ class ProcessException : Exception
...
+    if (commandLine.length == 0) return commandLine;
+    char endChar = ' ';
+    if (commandLine[0] == '"')
+    {
+        commandLine = commandLine[1..$];
+        endChar = '"';
+    }
+    const end = commandLine.indexOf(endChar);
+    return commandLine[0 .. (end == -1) ? $ : end];
+}
```
CyberShadow: Why is this function at all necessary? `std.process` functions already take the command as an array of arguments, and its first element will be the executable name. This function isn't even correct, it would have to call or precisely replicate the logic of `CommandLineToArgvW`, which it doesn't.

marler8997: `spawnProcessImpl` actually takes a single string, since that's what `CreateProcessW` takes.  And I think the logic is correct here.  Take a look at the documentation for `CreateProcess`.  It finds the application name based on the first whitespace characters, unless it starts with a `"` character.  And windows filenames cannot contain quotes so you don't need to handle escapes in that instance.

https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw

marler8997: Here's the applicable documentation on it:

> The lpApplicationName parameter can be NULL. In that case, the module name must be the first white space–delimited token in the lpCommandLine string. If you are using a long file name that contains a space, use quoted strings to indicate where the file name ends and the arguments begin....

marler8997:
> More detailed error messages are nice but I think this is not a good way to do it. We don't know if the "not found" error was generated due to the path given to CreateProcess not existing - it could be due to some DLL not being found or the system not finding some other resource. I suggest to instead reword the message to something like "Failed to spawn process " ~ exePathHere ~ ": " ~ e.msg

I'm not sure if that's correct.  If you are missing a DLL you might get a different error code.  I'll do some experiments...

CyberShadow:
>  I'll do some experiments...

Don't waste your time. You cannot predict the exact behavior of all past and future implementations of the API. You also cannot know for sure the total set of conditions that create that error, unless you look at the implementation, which you 1) can't, and even if you could, would 2) still apply to only the version you looked at.

marler8997:
> Don't waste your time. You cannot predict the exact behavior of all past and future implementations of the API. You also cannot know for sure the total set of conditions that create that error, unless you look at the implementation, which you 1) can't, and even if you could, would 2) still apply to only the version you looked at.

True, however I think if we can rule out the case of a missing dll, then it's pretty safe to assume this error code will only apply when the application name is missing.  If we are wrong, then we can fix the error message later to be more ambiguous.

CyberShadow:
The error message I proposed is not any less unclear and does not make unwarranted assumptions. Why do you insist on keeping your wording?

marler8997:
> Why do you insist on keeping your wording?

I'm not insisting on it.  I'm suggesting we try to find out whether or not we can make this assumption.  If we can't, then you're wording is better.

CyberShadow:
We cannot make such assumptions anywhere in Phobos for the reasons I already mentioned above.

std/process.d
CyberShadow: > `spawnProcessImpl` actually takes a single string

This is a very bad reason to 1) undo the work that's done in the same module, and 2) duplicate the work of the operating system.

>  And I think the logic is correct here.

No. Sorry, it's not even close. For one thing, Windows ignores whitespace at the end of filenames. For another, tabs can also be used to separate arguments. There are likely many other subtle differences in behavior which would be pointless to try to match.

Sorry, the way you are approaching this is not good at all. Please rework this submission.

marler8997:
@CyberShadow could you describe what way you'd like me to implement pulling the application name from the commandLine string?

CyberShadow:
> @CyberShadow could you describe what way you'd like me to implement pulling the application name from the commandLine string?

I suggest changing the first argument of the Windows version of `spawnProcessImpl` to a `struct` type or tuple which contains both the command-line string (for `CreateProcess`) and a text description of the action being performed (to be displayed in case of error). At the `spawnProcess` call site, make it reference `args[0]`; at the `spawnShell` call site, make it explicitly say something like `"Failed to run shell command " ~ command`.

marler8997:
Making it print the full shell command is a bit more than I was planning on doing and I'm not sure if we want to print the full shell command...

CyberShadow:
Anything works as long as the error is not misleadingly inaccurate.

std/process.d
```diff
@@ -2642,13 +2654,30 @@ class ProcessException : Exception
...
     }
+    // ditto
+    version (Windows)
+    static ProcessException newCreateProcessError(SpawnContext context,
+                                                  string file = __FILE__,
+                                                  size_t line = __LINE__)
+    {
+        const customMsg = "Failed to spawn new process";
+        const lastError = GetLastError();
+        if (lastError == ERROR_FILE_NOT_FOUND)
```
CyberShadow: There is no need for this condition.

std/process.d
```diff
@@ -682,6 +685,12 @@ private Pid spawnProcessImpl(scope const(char[])[] args,
     }
 }
 
+private struct SpawnContext
```
CyberShadow: There is no need to add a struct type if you're going to add a new argument. The reason I suggested using a struct was because a newly added argument would be unused in the POSIX version. The POSIX version doesn't need this informaiton, so there is no need to change its signature at all; but, if you're going to add a new argument, there is no need to introduce a struct.

std/process.d
```diff
@@ -2649,6 +2661,18 @@ class ProcessException : Exception
...
+    // ditto
+    version (Windows)
+    static ProcessException newCreateProcessError(SpawnContext context,
+                                                  string file = __FILE__,
+                                                  size_t line = __LINE__)
+    {
+        if (context.arg0)
+            return newFromLastError("Failed to spawn process '" ~ cast(string)context.arg0 ~ "'");
+        if (context.shell)
+            return newFromLastError("Failed to run shell command '" ~ cast(string)context.shell ~ "'");
```
CyberShadow: This logic should be at the call site, not here. Either preformat the string there, or pass a `scope` delegate which does so on demand (which is less wasteful).

std/process.d
marler8997: If the signatures between windows and posix don't match then we'll need to have 2 different calls to these functions inside version statements.  Is that what you want?

CyberShadow: This is not true. They already don't match in `master`.

marler8997: didn't compile when they had different signatures, I've pushed the change with the versions.  Not sure what you want to do here.

std/process.d
marler8997: done

std/process.d
marler8997: yeah I saw that...fixed already

std/process.d
CyberShadow: > didn't compile when they had different signatures

Please take a moment to have a closer look at the signatures of the Windows and POSIX versions of `spawnProcessImpl`, as they are in master right now (before your changes).

>  Not sure what you want to do here.

Would actually putting in a little effort be too much to ask? :(

marler8997: I'm reading alot of hostility coming from you right now and I have no idea why?  Why are you being so aggressive and insulting me?

I'm looking at master and they have the same signatures.

https://github.com/dlang/phobos/blob/master/std/process.d#L355
https://github.com/dlang/phobos/blob/master/std/process.d#L695

CyberShadow: > I'm looking at master and they have the same signatures.

Please, look a little closer...

One of them takes an array, the other takes a string.

My suggestion was to replace the array in the Windows version with a `struct`. Then, no changes to the POSIX signature or additional `version` blocks would be required.

> I'm reading alot of hostility coming from you right now and I have no idea why? Why are you being so aggressive and insulting me?

Sorry, I feel like you are abusing my time by not spending a little more of your time to research this issue and submit a good contribution.

If you like, let's close this PR and I will submit one which addresses this issue in a manner I consider to be optimal.

BTW you still need to file an issue in Bugzilla for the changelog.

marler8997: I don't care much how this is implemented.  I can see that you care very much, so feel free to let me know what you want.  There's no need for saying things like "would actually putting in a little effort be too much to ask".  I'm willing to do what you want, and I don't think I've given you any reason to think otherwise or any reason to think that I'm not willing to put in work.  I've already completely re-written this thing 3 times in the last 40 minutes based on your suggestions but I can't read your mind to know what you want.

CyberShadow: I don't see why you think reading my mind is required. I wrote:

> I suggest changing the first argument of the Windows version of `spawnProcessImpl` to a `struct` type or tuple

I don't see how that's unclear.

marler8997:
Now I see what you wanted.  Sorry I didn't notice that extra set of square brackets on the first argument of `spawnProcessImpl`.

CyberShadow:
Still needs Bugzilla issue + mention of its ID in commit message.

CyberShadow:
>  Sorry I didn't notice that extra set of square brackets on the first argument of `spawnProcessImpl`.

It couldn't be any other way because the POSIX program execution primitive (`exec`) takes an array of strings (`argc`/`argv`).

marler8997:
> It couldn't be any other way because the POSIX program execution primitive (exec) takes an array of strings (argc/argv).

Yes I'm keenly aware of this, I just didn't put 2 and 2 together.  I didn't get what you were wanting and I'm sorry for the frustration.

marler8997:
Thanks for taking the time @CyberShadow, I like your solution much better :)

marler8997:
Looking back at your comments I don't know why I wasn't getting it.  I think my brain turned off a little when I started sensing the hostility.  I've heard that people's brains can stop working when they feel attacked, I think that may be what happened here.

CyberShadow:
I guess I could have been less antagonizing. Sorry about that. Thanks for dealing with me and the contribution :)

marler8997:
Cleaning up old PRs that aren't going anywhere.

CyberShadow:
Oops, what happened here? I thought this was ready for merging.

marler8997:
Yeah it was ready, just no one merged it...so it gets the axe as part of my PR spring cleaning :)

PR #7396, opened 2020-02-14
[docs] update deprecation schedule

drug007:
@jmdavis could you take a look at the subject?


drug007:
Ah, I see that the comment is not outdated itself. So since October 2019 has passed what is the current date for removing deprecated symbols now?

MoonlightSentinel:
The deprecation was delayed/forgotten. #7282 made the documentation private in december and hence the removal needs to be delayed as well.

drug007:
@MoonlightSentinel Thanks for info. I updated deprecation schedule.

PR #7397, opened 2020-02-17
Fix Issue 20585 - std.stdio.File.open() failure leaves File in invalid

kubo39:
state

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -501,6 +501,20 @@ Throws: `ErrnoException` in case of error.
         resetFile(name, stdioOpenmode, false);
     }
 
+    @system unittest // bugzilla 20585
+    {
+        File f;
+        try
+            f.open("doesn't exist");
+        catch (Exception _e)
+        {
+        }
+
+        assert(!f.isOpen);
+
+        f.close();  // to check not crash here
+    }
+
     private void resetFile(string name, scope const(char)[] stdioOpenmode, bool isPopened) @trusted
     {
         import core.stdc.stdlib : malloc;
@@ -513,7 +527,6 @@ Throws: `ErrnoException` in case of error.
             detach();
         }
 
-        _p = cast(Impl*) enforce(malloc(Impl.sizeof), "Out of memory");
         FILE* handle;
         version (Posix)
         {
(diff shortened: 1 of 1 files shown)


std/stdio.d
```diff
@@ -514,7 +530,10 @@ Throws: `ErrnoException` in case of error.
         }
 
         _p = cast(Impl*) enforce(malloc(Impl.sizeof), "Out of memory");
+        memset(_p, 0x0, Impl.sizeof);
```
CyberShadow: Why not calloc?

std/stdio.d
```diff
@@ -514,7 +530,10 @@ Throws: `ErrnoException` in case of error.
...
         FILE* handle;
+
+        scope (failure) _p.handle = null;
```
CyberShadow: Wouldn't this leak memory?

std/stdio.d
```diff
@@ -501,9 +501,25 @@ Throws: `ErrnoException` in case of error.
...
+        File f;
+        try
+            f.open("doesn't exist");
+        catch (Exception _e)
+        {
+        }
+
+        assert(!f.isOpen);
+
+        if (f.isOpen)
```
CyberShadow: This can never be true due to the assert above.

std/stdio.d
kubo39: Yes, I missed.. thanks for pointing out.

kubo39:
CI failed, couldn't reproduce in my machine. I'll add `_p = null` for checking.

std/stdio.d
```diff
@@ -513,8 +527,14 @@ Throws: `ErrnoException` in case of error.
             detach();
         }
 
-        _p = cast(Impl*) enforce(malloc(Impl.sizeof), "Out of memory");
+        _p = cast(Impl*) enforce(calloc(1, Impl.sizeof), "Out of memory");
```
CyberShadow: Any reason this problem couldn't be fixed simply by moving this line to after the `_fopen` call? As far as I can see, then the other changes will become unnecessary.

kubo39: This looks simpler solution.

std/stdio.d
```diff
@@ -536,6 +549,7 @@ Throws: `ErrnoException` in case of error.
                          text("Cannot open file `", name, "' in mode `",
                               stdioOpenmode, "'"));
         }
+        _p = cast(Impl*) enforce(calloc(1, Impl.sizeof), "Out of memory");
```
CyberShadow: `calloc` is now unnecessary because `initImpl` initializes all fields of `_p` and it doesn't look like anything can fail in-between which would lead to a partial initialization.

PR #7398, opened 2020-02-20
Fix issue 20589 - typeof may give wrong result in case of classes defining `opCall` operator

drug007:
See bugzilla [here](https://issues.dlang.org/show_bug.cgi?id=20589)

### DIFF
--- a/std/container/array.d
+++ b/std/container/array.d
@@ -1585,6 +1585,26 @@ if (!is(Unqual!T == bool))
     ai.insertBack(arr);
 }
 
+/**
+ * issue 20589 - typeof may give wrong result in case of classes defining `opCall` operator
+ * https://issues.dlang.org/show_bug.cgi?id=20589
+ *
+ * destructor std.container.array.Array!(MyClass).Array.~this is @system
+ * so the unittest is @system too
+ */
+@system unittest
+{
+    class MyClass
+    {
+        T opCall(T)(T p)
+        {
+            return p;
+        }
+    }
+
+    Array!MyClass arr;
+}
+
 
 ////////////////////////////////////////////////////////////////////////////////
 // Array!bool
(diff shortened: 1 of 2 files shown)


drug007:
@wilzbach could you point out where can I add a test for this issue?

thewilsonator:
Please change the commit message to say `Fix issue 20589`.

Geod24:
> Please change the commit message to say `Fix issue 20589`.

`Fix issue 20589 - <Issue title>`

People watching the core repositories don't want to open every single notification they get, just those that seems relevant to their area of expertise. Likewise, when going through old PR / commits (trying to find where a bug was fixed, regression introduced, or feature added), it is extremely painful to see commit without correct title / description.

Geod24:
Thanks for changing the title, but the bot picks up the commit, so you also have to go and edit it.

std/container/array.d
```diff
@@ -1585,6 +1585,19 @@ if (!is(Unqual!T == bool))
     ai.insertBack(arr);
 }
 
+@system unittest
```
atilaneves: Is there a reason this can't be `@safe`?

drug007: destructor std.container.array.Array!(MyClass).Array.~this is @system

atilaneves: Could you please add a comment explaining that then? Thanks!

drug007: Done.

std/container/array.d
```diff
@@ -1585,6 +1585,21 @@ if (!is(Unqual!T == bool))
     ai.insertBack(arr);
 }
 
+// destructor std.container.array.Array!(MyClass).Array.~this is @system
+// so the unittest is @system too
```
atilaneves: Sorry to have missed this the last time around, I think this just needs a reference (with a link) to the issue on bugzilla and this should be good to go.

drug007: Sorry that I didn't do it myself. I promise the next time I'll be a good boy!

PR #7403, opened 2020-02-23
more printf format fixes

WalterBright:
more endless buggy printf formats. Blocking https://github.com/dlang/dmd/pull/10812


adamdruppe:
The website used to suggest that `%*s` thing for D strings, so I'm not surprised to see it in a lot of places.

I always suspected it was a bad idea and indeed now we can see that for sure!

PR #7407, opened 2020-02-26
Allow floatTraits to work with inout

John-Colvin:
which in turn means `isInfinity` and others can too.


ibuclaw:
If the tests actually worked... :-)

John-Colvin:
> If the tests actually worked... :-)

sorry, should be fixed now

PR #7408, opened 2020-02-27
Fix Issue 20618 - std.random.unpredictableSeed can work in betterC

n8sh:
The diff is easier to review with whitespace changes hidden.

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -104,6 +104,7 @@ Distributed under the Boost Software License, Version 1.0.
 module std.random;
 
 
+import std.internal.attributes : betterC;
 import std.range.primitives;
 import std.traits;
 
@@ -1656,7 +1657,7 @@ version (AnyARC4Random)
 }
 else
 {
-    private ulong bootstrapSeed() @nogc nothrow
+    private ulong bootstrapSeed()() @nogc nothrow
     {
         // Issue 19580 - previously used `ulong result = void` to start with
         // an arbitary value but using an uninitialized variable's value is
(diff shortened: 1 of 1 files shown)


n8sh:
At least on my machine the `@betterC` test extractor doesn't seem to be running any tests (it always succeeds). I manually tested this locally but this probably shouldn't be merged unless or until automatic tests are working properly.

std/random.d
```diff
@@ -1656,7 +1657,7 @@ version (AnyARC4Random)
 }
 else
 {
-    private ulong bootstrapSeed() @nogc nothrow
+    private ulong bootstrapSeed()() @nogc nothrow
```
n8sh: On my machine changing this function to a template was necessary to avoid a symbol not found error with `betterC`. This seems like a bug.

wilzbach:
> At least on my machine the `@betterC` test extractor doesn't seem to be running any tests (it always succeeds). I manually tested this locally but this probably shouldn't be merged unless or until automatic tests are working properly.

Good catch! -> https://github.com/dlang/phobos/pull/7409

std/random.d
kinke: Why would that be a bug? Everything that works in betterC must be templated, as druntime and Phobos aren't linked. So of course we could needlessly templatize most of druntime, but what would be the point? Static functionality that can be precompiled once for many compilation cycles does provide benefits.

n8sh: I misunderstood how betterC treats Phobos then. I thought it treated it like any library.

n8sh:
https://github.com/dlang/phobos/pull/7408#discussion_r391957141
>So of course we could needlessly templatize most of druntime, but what would be the point? Static functionality that can be precompiled once for many compilation cycles does provide benefits.

Closing because I now have doubts about whether this PR is a net improvement.

PR #7409, opened 2020-02-27
Fix @betterC testsuite

wilzbach:
Found in https://github.com/dlang/phobos/pull/7408

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -10924,6 +10924,7 @@ if (isInputRange!Range && !isInstanceOf!(SortedRange, Range))
     }
 
     ///
+    static if (is(Range : int[]))
     @safe unittest
     {
         import std.algorithm.comparison : equal;
@@ -10966,6 +10967,7 @@ policies are allowed, and `SearchPolicy.binarySearch` is the default.
     }
 
     ///
+    static if (is(Range : int[]))
     @safe unittest
     {
         import std.algorithm.comparison : equal;
@@ -11028,6 +11030,7 @@ policies are allowed, and `SearchPolicy.binarySearch` is the default.
     }
 
     ///
+    static if (is(Range : int[]))
     @safe unittest
     {
         import std.algorithm.comparison : equal;
@@ -11088,6 +11091,7 @@ equalRange). Completes the entire search in $(BIGOH log(n)) time.
     }
 
     ///
+    static if (is(Range : int[]))
     @safe unittest
     {
         import std.algorithm.comparison : equal;
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -1192,9 +1192,16 @@ if (distinctFieldNames!(Specs))
                 static if (__traits(compiles, h = .hashOf(field[i])))
                     const k = .hashOf(field[i]);
                 else
+                {
                     // Workaround for when .hashOf is not both @safe and nothrow.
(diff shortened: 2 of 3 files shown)


std/range/package.d
```diff
@@ -10924,6 +10924,7 @@ if (isInputRange!Range && !isInstanceOf!(SortedRange, Range))
     }
 
     ///
+    static if (is(Range : int[]))
```
n8sh: Did this get included by accident?

wilzbach: No, these tests are now also run and included. We can either limit the effect or move the tests out of the struct. This is exactly why a `version(StdUnittest)` would be nice.

std/typecons.d
```diff
@@ -1184,7 +1184,7 @@ if (distinctFieldNames!(Specs))
             Returns:
                 A `size_t` representing the hash of this `Tuple`.
          */
-        size_t toHash() const nothrow @safe
+        size_t toHash()() const nothrow @safe
```
n8sh:
```suggestion
        size_t toHash() const nothrow @safe
```

And below replace
```d
static if (__traits(compiles, h = .hashOf(field[i])))
    const k = .hashOf(field[i]);
else
    // Workaround for when .hashOf is not both @safe and nothrow.
    // BUG: Improperly casts away `shared`!
    const k = typeid(T).getHash((() @trusted => cast(const void*) &field[i])());
```

with something like

```d
static if (__traits(compiles, h = .hashOf(field[i])))
    const k = .hashOf(field[i]);
else static if (is(T : shared U, U) && __traits(compiles, h = .hashOf((() @trusted => cast(U) field[i])())))
    // BUG: Improperly casts away `shared`!
    const k = .hashOf((() @trusted => cast(U) field[i])());
else
    // Workaround for when .hashOf is not both @safe and nothrow.
    // BUG: Improperly casts away `shared`!
    const k = typeid(T).getHash((() @trusted => cast(const void*) &field[i])());
```

This change is reasonable because we also have to cast away shared to call `typeid(T).getHash`. So if removing shared would also let us call `.hashOf` there is no reason to involve `typeid`. This is enough to make the current tests compile and pass both with and without -betterC.

n8sh:
Pushing a change to make the tests pass.

std/typecons.d
```diff
@@ -1192,9 +1192,16 @@ if (distinctFieldNames!(Specs))
...
                     const k = .hashOf(field[i]);
                 else
+                {
                     // Workaround for when .hashOf is not both @safe and nothrow.
-                    // BUG: Improperly casts away `shared`!
-                    const k = typeid(T).getHash((() @trusted => cast(const void*) &field[i])());
+                    static if (is(T : shared U, U) && __traits(compiles, (U* a) nothrow @safe => .hashOf(*a))
+                            && !__traits(hasMember, T, "toHash"))
+                        // BUG: Improperly casts away `shared`!
+                        const k = .hashOf(*(() @trusted => cast(U*) &field[i])());
```
atilaneves: Why is this needed?

n8sh: So it compiles in betterC. When not in betterC it isn't expected to have different result than using typeid.

PetarKirov:
This is wrong, see e857ce1127b5ff0a3b2b54159d86ff31c00dc0dd

PR #7410, opened 2020-03-02
Typo fix in Systime describtion

burner:
ntended -> intended


burner:
@thewilsonator thank you

PR #7413, opened 2020-03-03
Add tests for issue 20623 - std.conv.to!string error when argument is an inout float

MoonlightSentinel:
This fixes the regression for the upcoming release but it the should probably reconsidered how `inout` behaves in heavy templated code

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -488,6 +488,35 @@ template to(T)
     assertThrown!ConvException(to!AA(" [1:1]"));
 }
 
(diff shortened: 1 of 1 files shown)


MoonlightSentinel:
This will be fixed as a side effect of #7407. We should probably cherry-pick/merge that commit to stable and only add the test case.

MoonlightSentinel:
Rebased to master since this didn't make the cut for 2.091.1

PR #7418, opened 2020-03-10
mark access to `value.store.str` as @trusted

aG0aep6G:
When issue 20655 ("attribute inference accepts unsafe union access as
@safe") is fixed, DMD correctly complains about the operation being unsafe.

Needed for <https://github.com/dlang/dmd/pull/10884>.

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -1205,8 +1205,7 @@ if (isInputRange!T && !isInfinite!T && isSomeChar!(ElementEncodingType!T))
                     break;
                 }
 
-                value.type_tag = JSONType.string;
-                () @trusted { value.store.str = str; } ();
+                value.assign(str);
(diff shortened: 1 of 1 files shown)


std/json.d
```diff
@@ -1206,7 +1206,7 @@ if (isInputRange!T && !isInfinite!T && isSomeChar!(ElementEncodingType!T))
                 }
 
                 value.type_tag = JSONType.string;
-                value.store.str = str;
+                () @trusted { value.store.str = str; } ();
```
CyberShadow: I think it would be simpler to replace both lines with `assign(str)`, which already has a `() @trusted { store.str = t; }();`.

aG0aep6G: done

PR #7425, opened 2020-03-18
Fix Issue 13595 - added chunkByAny

dukc:
This is a variant on `std.algorithm.iteration.chunkBy` that does not place restrictions on the binary predicate.

The idea isn't new - #2654 did essentially the same (the function was called `groupBy` back then), but it was reverted because of worries about the algortihm iterating the underlying range twice. I do not think those issues apply anymore -nowadays `chunkBy`, if used linearly, will only iterate once, and I included an unittest that verifies that `chunkByAny` does the same.

I also documented this issue: https://forum.dlang.org/thread/ztgtmumenampiobbuiwd@forum.dlang.org to `chunkBy`. It was the added `chunkByAny` documentation unittest that discovered the issue, and I felt the extra `save` calls I had to add mandated the documentation change.

### DIFF
--- a/changelog/add-bychunkany
+++ b/changelog/add-bychunkany
@@ -1,0 +1,16 @@
+As a fix to issue #13595, added a variation of 'std.algorithm.iteration.chunkBy'
+called 'chunkByAny'. Unlike the original algorithm, this variation is designed
+to work with predicates that are neither reflexive, symmetric nor transitive:
+
+```
+auto source = [4, 3, 2, 11, 0, -3, -3, 5, 3, 0];
+auto result = source.chunkByAny!((a,b) => a > b);
+assert(result.equal!equal([
+	[4, 3, 2],
+	[11, 0, -3],
+	[-3],
+	[5, 3, 0]
+]));
+```
+
+However, it requires forward ranges to work.
--- a/changelog/add-bychunkany.dd
+++ b/changelog/add-bychunkany.dd
@@ -1,3 +1,5 @@
+Added 'std.algorithm.iteration.chunkByAny'
+
 As a fix to issue #13595, added a variation of 'std.algorithm.iteration.chunkBy'
 called 'chunkByAny'. Unlike the original algorithm, this variation is designed
 to work with predicates that are neither reflexive, symmetric nor transitive:
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -1908,9 +1908,14 @@ private template ChunkByImplIsUnary(alias pred, Range)
 }
 
 // Implementation of chunkBy for non-forward input ranges.
-private struct ChunkByImpl(alias pred, Range)
+private struct ChunkByImpl(alias pred, Range, bool predEquivelanceQuaranteed)
 if (isInputRange!Range && !isForwardRange!Range)
 {
+	static assert
+	(	
+		predEquivelanceQuaranteed,
+		"implementing chunkBy with non-forward ranges without predicate equivalence quarantees is impossible"
+	);
     enum bool isUnary = ChunkByImplIsUnary!(pred, Range);
 
     static if (isUnary)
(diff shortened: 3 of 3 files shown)


changelog/add-bychunkany
```diff
@@ -0,0 +1,16 @@
+As a fix to issue #13595, added a variation of 'std.algorithm.iteration.chunkBy'
```
Hackerpilot: One of the steps of the CI failure is caused by this file's name:

> ERROR: /dev/shm/dtest/work/repo/phobos/changelog/add-bychunkany needs to have .dd or .md as extension

jondegenhardt:
I agree that having a less restricted predicate would be useful. I've wanted it myself. However, I have couple concerns that I think are at least worth discussion:

(A) Requiring the source range to be a forward range - Nearly all range algorithms take non-forward input ranges as arguments. And it's common to be working with non-forward input ranges in real code.

Having this restriction will add to user confusion. The distinction between forward and non-forward input ranges is at least an intermediate-level D topic. Users will naturally expect to be able to use `chunkBy` and `chunkByAny` in the same places, but won't be able to. Unless compiler errors are really good, users will have difficulty understanding compilation errors and their options for address the restriction. This will be particular problematic when `chunkByAny` is used in pipelines with other range algorithms like `roundRobin` and `merge`.

Given the forward range restriction, the best practices guideline might be to prefer `chunkBy` except when the less restricted predicate is needed. Only then should `chunkByAny` be used. This feels suboptimal if indeed it is the case.

(B) DIP-1000 and `@safe` code complexity - `chunkBy` is already a complicate piece of code,  one that does not currently satisfy DIP-1000 safety rules. That's why many of the unit tests are `@system`. A significant part of this due to the use of `RefCounted` (unsafe). However, when I've looked at this code I felt that the iteration of a main range by its sub-chunks gave rise to lifetime questions under DIP-1000 that need to be resolved. I could be wrong about this, I'm not a DIP-1000 expert.

These concerns make me wonder if it might be better to resolve DIP-1000 safety issues before making enhancements to the `chunkBy` implementation. (This PR modifies the underlying `chunkBy` implementation.)

dukc:
> 
> (A) Requiring the source range to be a forward range - Nearly all range algorithms take non-forward input ranges as arguments. And it's common to be working with non-forward input ranges in real code.

I was originally going to support forward ranges, but then I realized I can't. Unlike original `chunkBy`, `chunkByAny` needs to always know what the last element was to answer whether it's `empty`. With non-forward ranges, if the user has called `popFront` on the source range, there's no way to know the last element.

> 
> Given the forward range restriction, the best practices guideline might be to prefer `chunkBy` except when the less restricted predicate is needed. Only then should `chunkByAny` be used. This feels suboptimal if indeed it is the case.

Yes, that might well be a good guideline. And for natural reasons: If you can assure the compiler that the predicate has equivalence behaviour, it can make some assumptions it can't with `chunkByAny`. So not suboptimal at all.

> 
> (B) DIP-1000 and `@safe` code complexity - `chunkBy` is already a complicate piece of code, one that does not currently satisfy DIP-1000 safety rules. That's why many of the unit tests are `@system`. A significant part of this due to the use of `RefCounted` (unsafe). However, when I've looked at this code I felt that the iteration of a main range by its sub-chunks gave rise to lifetime questions under DIP-1000 that need to be resolved. I could be wrong about this, I'm not a DIP-1000 expert.
> 
> This concern make me wonder if it might be better to resolve DIP-1000 safety issues before making enhancements to the `chunkBy` implementation. (This PR modifies the underlying `chunkBy` implementation.)

I felt that I might push the "focus pull request on one issue" too much if I started fixing that. Judging by the already existing unittests `@safe`ty was already broken even before this change, and the implementation was also reference-counted reqardless of my changes. But I do agree that something should be done on the `@safe`ty, just as a separate PR.

dukc:
Reboot here: #7794

PR #7427, opened 2020-03-22
Fix Issue 15615: Creating a Variant with a const primitive type doesn't compile

apz28:


std/variant.d
```diff
@@ -3097,3 +3101,26 @@ if (isAlgebraic!VariantType && Handler.length > 0)
...
+    static struct Wrapper
+    {
+        long* ptr;
+    }
+
+    alias VariantWrapper = Algebraic!(Wrapper, double);
+
+    long l = 2;
+    const cw = Wrapper(&l);
+    static assert(!__traits(compiles, VariantWrapper(cw)), "Stripping `const` from type with indirection!");
```
dukc, line 3125: If this triggered, I'm worried the error message would mislead. If I were to get a message like this, I'd think that the unit test is accidently trying to strip `const` and `Algebraic` is trying to prevent it. Would "std.variant.Variant is wrongly allowing stripping `const` from type with indirection!" be better?

RazvanN7:
ping @apz28 . What is the state of this PR?

PR #7435, opened 2020-03-31
[dmd-cxx] Backport IeeeFlagsSupport and FloatingPointControlSupport

ibuclaw:
The FP unit-tests are a little better in upstream, but still only work when not optimizing.


thewilsonator:
I'll leave merging to you.

PR #7436, opened 2020-04-07
std.random.unpredictableSeed: on x86/x86-64 try using RDRAND when there is no arc4random

n8sh:
This enhancement is now possible (without being DMD-only) because LDC version 1.20.0 added support for data directives in DMD-style inline asm.

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -116,6 +116,9 @@ else version (TVOS)
 else version (WatchOS)
     version = Darwin;
 
+version (D_InlineAsm_X86) version = InlineAsm_X86_Any;
+version (D_InlineAsm_X86_64) version = InlineAsm_X86_Any;
+
 ///
 @safe unittest
 {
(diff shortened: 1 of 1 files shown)


std/random.d
```diff
@@ -1688,6 +1691,46 @@ else
         result = (result ^ (result >>> 47)) * m;
         return result ^ (result >>> 47);
     }
+
+    // If we don't have arc4random and we don't have RDSEED fall back to this.
+    private ulong fallbackSeed() @nogc nothrow
+    {
```
n8sh, line 1697: This code was moved out of `unpredictableSeed` into its own function.

n8sh:
It might make sense to use RDRAND instead since it suffers less from contention. RDRAND is similar to macOS arc4random: it's an AES-based CPRNG that periodically reseeds itself.

PR #5695, opened 2017-08-17
Fix Issue 9183 - Add a Nullable.get(x) overload

RazvanN7:

PR #7437, opened 2020-04-07
Add get(default) same-type overload in Nullable

FeepingCreature:
Works around (I suspect) https://issues.dlang.org/show_bug.cgi?id=20670 , which may
cause `Nullable` with arrays of `immutable struct`s to incorrectly strip
away the `immutable`, creating impossible types that can't be reused.

@RazvanN7 Since you added `get(default)`, is there any reason why it would have to infer a different type than the `Nullable`'s?

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2955,6 +2955,9 @@ Gets the value if not null. If `this` is in the null state, and the optional
 parameter `fallback` was provided, it will be returned. Without `fallback`,
 calling `get` with a null state is invalid.
 
+When the fallback type is different from the Nullable type, `get(T)` returns
+the common type.
+
 Params:
     fallback = the value to return in case the `Nullable` is null.
 
@@ -2969,7 +2972,13 @@ Returns:
     }
 
     /// ditto
-    @property get(U)(inout(U) fallback) inout @safe pure nothrow
+    @property inout(T) get()(inout(T) fallback) inout @safe pure nothrow
+    {
+        return isNull ? fallback : _value.payload;
+    }
+
+    /// ditto
+    @property auto get(U)(inout(U) fallback) inout @safe pure nothrow
     {
         return isNull ? fallback : _value.payload;
     }
@@ -4007,6 +4016,25 @@ nothrow pure @nogc @safe unittest
     result = sample.apply!greaterThree;
     assert(!sample.isNull && !result.isNull);
     assert(result.get == 4);
+}
+
+// test that Nullable.get(default) can merge types
+@safe @nogc nothrow pure
+unittest
+{
+    Nullable!ubyte sample = Nullable!ubyte();
+
+    // Test that get(U) returns the common type of the Nullable type and the parameter type.
+    assert(sample.get(1000) == 1000);
+}
+
+// Workaround for https://issues.dlang.org/show_bug.cgi?id=20670
+@safe @nogc nothrow pure
+unittest
+{
+    immutable struct S { }
+
+    S[] array = Nullable!(S[])().get(S[].init);
 }
 
 /**


thewilsonator:
Implicit conversions?

FeepingCreature:
Should be done by explicitly specifying the parameter type...

FeepingCreature:
ping (@RazvanN7 )

RazvanN7:
My code was originally used the same type as the `Nullable`'s, however @schveiguy asked for this change: "Can you change the T to U as it masks the overarching template type from the Nullable struct?"

But I guess this was only as a precautionary measure for the case where T would be used further in the function, so if this causes trouble, I have nothing against this PR, but maybe @schveiguy has a different perspective.

schveiguy:
Wouldn’t this be a breaking change?  I.e. Nullable!ubyte.get(1000) would stop working?

I don’t remember the original discussion. On my phone right now I will look tomorrow.

FeepingCreature:
Yeah, I think you're right. Damn. When the default value is a more generic type than the stored value, this would break.

Geod24:
@FeepingCreature : Sure
Original discussion here: https://github.com/dlang/phobos/pull/5695#pullrequestreview-57003239

> Wouldn’t this be a breaking change? I.e. Nullable!ubyte.get(1000) would stop working?

True, but that also means it is currently kinda broken.
The problem you mentioned in the link I posted above was that, by using a ternary, you'll get the more generic type. Since number literals are `int`, currently the following code errors out:
```D
import std.typecons;

void main ()
{
    Nullable!ubyte nl = 42;
    ubyte v1 = nl.get();
    ubyte v2 = nl.get(0); // Error: cannot implicitly convert expression nl.get(0) of type int to ubyte
    ubyte v3 = nl.get(1000); // Error: cannot implicitly convert expression nl.get(1000) of type int to ubyte
}
```

The error on `v3` is understandable, but `v2` ?
You'll have the same problem if you pass two completely different `class` types, you'll get an `Object` out of it. I think `get` returning a more generic type than what it stores is a bigger issue.

schveiguy:
So my review of @RazvanN7's PR that includes this statement:

> Can you change the T to U as it masks the overarching template type from the Nullable struct?

Was not the same as what is being proposed here. In the original, the code was:

```D
T get(T)(T datum) {...}
```
And my point was simply that the `T` in that template is going to override the `T` in `Nullable(T)` type,  and look confusing in documentation.

In other words, I just wanted a name change, not functionality change.

> The error on `v3` is understandable, but `v2` ?

Can't we fix this with an overload?

i.e.:

```D
T get(T defaultVal) { ... } // should be preferred when literal can fit
auto get(U)(U defaultVal) {...}
```

The point of get(default) is to replace `someVeryLongSymbolOrExpression.isNull ? defaultVal : someVeryLongSymbolOrExpression.get` with `someVeryLongSymbolOrExpression(defaultVal)`. Preventing defaultVal from being compatible with the ternary expression seems like a step backwards.

schveiguy:
Note @FeepingCreature the proposed overload solution fixes your issue as well. I think that's the best solution (for now, of course, fixing 20670 would be good too).

FeepingCreature:
Looks good, I'll change it tomorrow.

std/typecons.d
```diff
@@ -4009,6 +4015,24 @@ nothrow pure @nogc @safe unittest
...
+// test that Nullable.get(default) can merge types
+@safe @nogc nothrow pure
+unittest
+{
+    Nullable!ubyte sample = Nullable!ubyte();
+
+    assert(sample.get(1000) == 1000);
+}
+
+// regression test for workaround for https://issues.dlang.org/show_bug.cgi?id=20670
```
Geod24, line 4027:
```suggestion
// Workaround for https://issues.dlang.org/show_bug.cgi?id=20670
```

Technically all unittests are regression tests (but not all of them are tests for regressions).

std/typecons.d
```diff
@@ -4009,6 +4015,24 @@ nothrow pure @nogc @safe unittest
     assert(result.get == 4);
 }
 
+// test that Nullable.get(default) can merge types
+@safe @nogc nothrow pure
+unittest
+{
+    Nullable!ubyte sample = Nullable!ubyte();
+
+    assert(sample.get(1000) == 1000);
```
Geod24, line 4024:
```suggestion
    assert(sample.get(1000) == 1000);
    // Test that `get(U)` return the common type of `T` and `U`
    static assert(is(typeof(sample.get(1000)) == int));
```

std/typecons.d
```diff
@@ -2969,7 +2969,13 @@ Returns:
     }
 
     /// ditto
-    @property get(U)(inout(U) fallback) inout @safe pure nothrow
+    @property inout(T) get()(inout(T) fallback) inout @safe pure nothrow
+    {
+        return isNull ? fallback : _value.payload;
+    }
+
+    /// ditto
```
Geod24, line 2977: You might want to document its special (and IMHO, surprising) semantic.

std/typecons.d
FeepingCreature: If it's not `int`, the `get(1000)` call can't return `1000`.

std/typecons.d
FeepingCreature: documented?

std/typecons.d
FeepingCreature: done

std/typecons.d
Geod24: I'm sorry I don't understand this comment ?

FeepingCreature: The type is checked implicitly by comparing with a value that's too big for `ubyte`. We don't really care about whether it returns `ubyte` or `int` except in that if we pass in a value that's larger than the data range, then when the Nullable is null, we get the same value back out without truncation.

schveiguy: @FeepingCreature is correct, the test is implicitly checking that it's `int`, because `ubyte` can't be 1000. It also captures the semantic expectation -- if it's `null` I want to use 1000, whatever type that means.

Geod24: Fine by me

schveiguy:
This didn't close any issue, so it won't show up in the changelog. Should an issue be made for this, or should a changelog entry be added?

FeepingCreature:
There isn't really a separate issue, and the root issue is hard to fix. I think this feature is new enough that most people haven't noticed it exists yet, and immutable structs are sadly kind of rare, so it'll probably be fine - people will notice the change just by virtue of trying to use .get and it working.

PR #7438, opened 2020-04-09
Fix incorrect use of sprintf format

Luhrel:


std/math.d
```diff
@@ -238,13 +238,19 @@ version (StdUnittest) private
...
+        {
             alias real_t = real;
+            string fmt = "Lg";
+        }
 
         () @trusted {
-            ix = sprintf(bufx.ptr, "%.*Lg", ndigits, cast(real_t) x);
-            iy = sprintf(bufy.ptr, "%.*Lg", ndigits, cast(real_t) y);
+            ix = sprintf(bufx.ptr, "%.*" ~ fmt, ndigits, cast(real_t) x);
+            iy = sprintf(bufy.ptr, "%.*" ~ fmt, ndigits, cast(real_t) y);
```
thewilsonator, line 253: `fmt` being a runtime variable means this is not guaranteed to be nul terminated. Also its not a literal so cannot be implicitly cast to `const(char)*`.

PR #7447, opened 2020-04-16
std.range.primitives: Fix static condition for X32 targets

ibuclaw:
`version (X86_64)` is not a sure way to test whether length sizes are 64-bits.


ibuclaw:
> stable ?

I don't use stable, because the merge-master commits always breaks too many stuff.

PR #7453, opened 2020-04-21
Fix/issue 20756 implicitconversiontargets interfaces

FeepingCreature:
Repro:

```
interface A { }
interface B : A { }

static assert(is(ImplicitConversionTargets!B == AliasSeq!(A)));
```

Expected: `AliasSeq!(A)`

Got: `AliasSeq!()`

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -4560,6 +4560,8 @@ template TransitiveBaseTypeTuple(T)
 {
     static if (is(T == Object))
         alias TransitiveBaseTypeTuple = AliasSeq!();
+    else static if (is(T == interface))
+        alias TransitiveBaseTypeTuple = InterfacesTuple!T;
     else
         alias TransitiveBaseTypeTuple =
             AliasSeq!(BaseClassesTuple!T, InterfacesTuple!T);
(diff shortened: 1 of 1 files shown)


FeepingCreature:
I'm a bit worried that the docs explicitly call out "a class derived from `Object`". Does supporting interfaces break anything?

thewilsonator:
If you're changing the docs, you should clarify if it works with `extern(C++)`/`extern(Objective-C)` classes or not.

FeepingCreature:
I cannot think of any reason why it wouldn't, but I don't know anything about C++ classes/interfaces and their D binding. Who could I ping about that?

thewilsonator:
Sorry I missed this. Probably @jacob-carlborg and/or @Geod24

jacob-carlborg:
I guess the above example should work for for C++ and Objective-C interfaces. But keep in mind they don't inherit from Object.

There are also COM interfaces, which I don't know how they behave. 

Ideally there should be tests for all kinds. Although Objective-C interfaces are deprecated.

RazvanN7:
ping @FeepingCreature . What is the status of this?

FeepingCreature:
Rebased, but note that I don't remember what I needed this for.

FeepingCreature:
Actually, can't this definitely be closed? Looking at the changelog, we have `AllImplicitConversionTargets` now, and that handles interfaces just fine.

PR #7455, opened 2020-04-25
Fix Issue 20765 - Can't run processes with relative paths when specif…

CyberShadow:
…ying a working directory

Remove the dubious check which attempted to duplicate the operating
system's behavior.

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -878,12 +878,7 @@ private Pid spawnProcessImpl(scope const(char[])[] args,
 
     if (args.empty) throw new RangeError();
     const(char)[] name = args[0];
-    if (any!isDirSeparator(name))
-    {
-        if (!isExecutable(name))
-            throw new ProcessException(text("Not an executable file: ", name));
-    }
-    else
+    if (!any!isDirSeparator(name))
     {
         name = searchPathFor(name);
         if (name is null)
@@ -1779,6 +1774,22 @@ version (Windows)
 
     auto data = readText(fn);
     assert(data == "AAAAAAAAAABBBBB\r\n", data);
+}
+
+// https://issues.dlang.org/show_bug.cgi?id=20765
+// Test that running processes with relative path works in conjunction
+// with indicating a workDir.
+version (Posix) @system unittest
+{
+    import std.file : mkdir, write, setAttributes;
+    import std.conv : octal;
+
+    auto dir = uniqueTempPath();
+    mkdir(dir);
+    write(dir ~ "/program", "#!/bin/sh\necho Hello");
+    setAttributes(dir ~ "/program", octal!700);
+
+    assert(execute(["./program"], null, Config.none, size_t.max, dir).output == "Hello\n");
 }
 
 /**


n8sh:
>Remove the dubious check which attempted to duplicate the operating system's behavior.

Maybe the motivation was producing a better error message, or someone saw value in bailing out early when it can be determined early on that the process will fail (which is optimizing for a case that I assume is uncommon). The check could be modified to be:

```d
    if (any!isDirSeparator(name))
    {	
        if (workDir.length == 0 && !isExecutable(name))	
            throw new ProcessException(text("Not an executable file: ", name));	
    }
```

Or it could be removed entirely as you did.

CyberShadow:
It is dangerous to make assumptions such as those made by the removed code, because things such as ACLs, virtual filesystems, and not-quite-POSIX systems may cause the result to differ from what is expected. Generally in all such cases it is more correct to try the action and handle the outcome, rather than attempt to predict it.

CyberShadow:
It looks like the test for this fix uncovered an additional problem on FreeBSD.

CyberShadow:
I can't figure out why it's failing on FreeBSD.

I tried the following C program:

```c
#include <stdio.h>
#include <fcntl.h>
#include <unistd.h>

int main()
{
	int fd = open("dir", O_RDONLY);
	fchdir(fd);
	static const char *argv[] = {"./test.sh", 0};
	execve(argv[0], argv, NULL);
}
```

But, it works as expected.

@jmdavis perhaps might know?

kyllingstad:
> > Remove the dubious check which attempted to duplicate the operating system's behavior.
> 
> Maybe the motivation was producing a better error message, or someone saw value in bailing out early when it can be determined early on that the process will fail (which is optimizing for a case that I assume is uncommon).

It was so the error could be handled in the parent, before the `fork()`. It predates the current pipe-based error transporting mechanism, which wasn't in there from the start. It can safely be removed now, and I agree that it should.

CyberShadow:
Fixed the FreeBSD failure. The cause was embarrassingly mundane - I used `#!/bin/bash` in the test script, and `bash` isn't part of FreeBSD.

std/process.d
```diff
@@ -878,12 +878,7 @@ private Pid spawnProcessImpl(scope const(char[])[] args,
 
     if (args.empty) throw new RangeError();
     const(char)[] name = args[0];
-    if (any!isDirSeparator(name))
-    {
-        if (!isExecutable(name))
-            throw new ProcessException(text("Not an executable file: ", name));
-    }
-    else
+    if (!any!isDirSeparator(name))
```
schveiguy, line 881: With this new check, won't path be prepended to all checks? So if you try to run "/bin/sh", it would try "/usr/bin//bin/sh", etc.?

Perhaps the better check is, does it start with a directory separator?

schveiguy: Hm, nevermind. If there is NO path separator, then it searches. I was misreading the check.

CyberShadow: Yes, same as what `execvp` etc. do. "The execlp(), execvp(), and execvpe() functions duplicate the actions of the shell in searching for an executable file **if the specified filename does not contain a slash (/) character**."

PR #7456, opened 2020-04-25
std.complex: Add complex exp(), log(), log10(), and pow()

ibuclaw:
Pushing out initial implementation. Unittest incoming...

### DIFF
--- a/std/complex.d
+++ b/std/complex.d
@@ -1047,3 +1047,104 @@ Complex!T sqrt(T)(Complex!T z)  @safe pure nothrow @nogc
     // Test ease of use (vanilla toString() should be supported)
     assert(complex(1.2, 3.4).toString() == "1.2+3.4i");
 }
+
+/**
+ * Calculates e$(SUPERSCRIPT x).
+ * Params:
+ *      x = A complex number
+ * Returns:
+ *      The complex base e exponential of `x`
+ */
+Complex!T exp(T)(Complex!T x) @trusted pure nothrow @nogc // TODO: @safe
+{
+    static import std.math;
+
+    return fromPolar!(T, T)(std.math.exp(x.re), x.im);
+}
+
+/**
+ * Calculate the natural logarithm of x.
+ * The branch cut is along the negative axis.
+ * Params:
+ *      x = A complex number
+ * Returns:
+ *      The complex natural logarithm of `x`
+ */
+Complex!T log(T)(Complex!T x) @safe pure nothrow @nogc
+{
+    static import std.math;
+
+    return Complex!T(std.math.log(abs(x)), arg(x));
+}
+
+/**
+ * Calculate the base-10 logarithm of x.
+ * Params:
+ *      x = A complex number
+ * Returns:
+ *      The complex base 10 logarithm of `x`
+ */
+Complex!T log10(T)(Complex!T x) @safe pure nothrow @nogc
+{
+    static import std.math;
+
+    return log(x) / Complex!T(std.math.log(10.0));
+}
+
+/**
+ * Calculates x$(SUPERSCRIPT n).
+ * The branch cut is on the negative axis.
+ * Params:
+ *      x = base
+ *      n = exponent
+ * Returns:
+ *      `x` raised to the power of `n`
+ */
+Complex!T pow(T, Int)(Complex!T x, const Int n) @safe pure nothrow @nogc
+if (isIntegral!Int)
+{
+    alias UInt = Unsigned!(Unqual!Int);
+
+    UInt m = (n < 0) ? -cast(UInt) n : n;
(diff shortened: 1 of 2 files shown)


std/complex.d
```diff
@@ -1047,3 +1047,104 @@ Complex!T sqrt(T)(Complex!T z)  @safe pure nothrow @nogc
...
+/**
+ * Calculates x$(SUPERSCRIPT n).
+ * The branch cut is on the negative axis.
+ * Params:
+ *      x = base
+ *      n = exponent
+ * Returns:
+ *      `x` raised to the power of `n`
+ */
+Complex!T pow(T, Int)(Complex!T x, Int n) @safe pure nothrow @nogc
```
n8sh, line 1103:
```suggestion
Complex!T pow(T, Int)(Complex!T x, const Int n) @safe pure nothrow @nogc
```
For IFTI.

ibuclaw:
Looks like tests are failing because std.complex doesn't have any template constraints for all it's functions.  Apparently indirect constraints aren't enough (`Complex!T foo(T)() { ... }` doesn't inherit constraints from it's return type `struct Complex(T) if (isFloatingPoint!T)`)

ibuclaw:
Actually, I can't reproduce with what I thought would be a reduced test, so must be a bug in partial then.

std/complex.d
```diff
@@ -1047,3 +1047,294 @@ Complex!T sqrt(T)(Complex!T z)  @safe pure nothrow @nogc
...
+
+    return fromPolar!(T, T)(std.math.exp(x.re), x.im);
+}
+
+///
+@safe pure nothrow @nogc unittest
+{
+    import std.math : PI;
+
+    assert(exp(complex(0.0, 1.0) * PI) == complex(1.0));
```
thewilsonator, line 1111: Ummm...

std/complex.d
```diff
@@ -1047,3 +1047,294 @@ Complex!T sqrt(T)(Complex!T z)  @safe pure nothrow @nogc
...
+            return Complex!T(T.nan, T.nan);
+        if (x.im == 0.0)
+            return x;
+    }
+    if (x.re == 0.0)
+    {
+        if (std.math.isNaN(x.im) || std.math.isInfinity(x.im))
+            return Complex!T(T.nan, T.nan);
+        else
+            return Complex!T(1.0, 0.0);
```
thewilsonator, line 1100: there is still a phase you need to worry about here.

ibuclaw: Forgot the `if (x.im == 0.0)` bit. :-)

thewilsonator: Right, and fixing this should now cause https://github.com/dlang/phobos/pull/7456#discussion_r418970871 to fail.

std/complex.d
ibuclaw: Yep, complex(-1) ... and needs approxEqual too. :-\

thewilsonator: That is a pity that we need approxEqual for such a fundamental equality.

ibuclaw:
Still WIP are:
- log() special cases.
- Put both exp() and log() special cases in Ddoc table.
- Changelog entry?

Best for another PR...
- tan is missing, as are also:
- asin, acos, atan
- sinh, cosh, tanh
- norm (squared magnitude)
- proj (projection onto the Riemann sphere)

std/complex.d
```diff
@@ -1047,3 +1056,536 @@ Complex!T sqrt(T)(Complex!T z)  @safe pure nothrow @nogc
...
+@safe pure nothrow @nogc unittest
+{
+    import std.math : LN10, PI, sqrt;
+
+    auto a = complex(2.0, 1.0);
+    assert(log10(a) == log(a) / log(complex(10.0)));
+
+    auto b = complex(0.0, 1.0);
+    auto c = complex(sqrt(2.0) / 2, sqrt(2.0) / 2);
+    assert(log10(b) * 2.0 == log10(c) * 4.0);
```
ibuclaw, line 1381: Still a test that fails on 32bits for "reasons"

ibuclaw:
@thewilsonator - green after changing a few more tests from `==` to `approxEqual`.

These are a little tedious to write out, would it make sense to add an `approxEqual` helper for `Complex!T`?

PetarKirov:
> would it make sense to add an approxEqual helper for Complex!T?

Or better yet, to `isClose` (https://github.com/dlang/phobos/pull/7384).

ibuclaw:
Excellent, I better update all my tests then. :-)

ibuclaw:
> Excellent, I better update all my tests then. :-)

That can wait for a follow-up PR anyhow, as most tests are isClose, but some only pass with approxEqual.

ibuclaw:
@thewilsonator - anything left here?

PR #7458, opened 2020-04-26
Fix Issue 20426 - doesPointTo with void[N] fails

MoonlightSentinel:
Accept a false positive/negative for may/doesPointTo

### DIFF
--- a/std/exception.d
+++ b/std/exception.d
@@ -1111,7 +1111,7 @@ If `source` is a class, then it will be handled as a pointer.
 If `target` is a pointer, a dynamic array or a class, then these functions will only
 check if `source` points to `target`, $(I not) what `target` references.
 
-If `source` is or contains a union, then there may be either false positives or
+If `source` is or contains a union or `void[n]`, then there may be either false positives or
 false negatives:
 
 `doesPointTo` will return `true` if it is absolutely certain
@@ -1149,8 +1149,11 @@ if (__traits(isRef, source) || isDynamicArray!S ||
     }
     else static if (isStaticArray!S)
     {
-        foreach (ref s; source)
-            if (doesPointTo(s, target)) return true;
+        static if (!is(S == void[n], size_t n))
+        {
+            foreach (ref s; source)
+                if (doesPointTo(s, target)) return true;
+        }
         return false;
     }
     else static if (isDynamicArray!S)
@@ -1191,8 +1194,33 @@ if (__traits(isRef, source) || isDynamicArray!S ||
     }
     else static if (isStaticArray!S)
     {
-        foreach (size_t i; 0 .. S.length)
-            if (mayPointTo(source[i], target)) return true;
+        static if (is(S == void[n], size_t n))
+        {
+            static if (n >= (void*).sizeof)
+            {
+                // Reinterpreting cast is impossible during ctfe
+                if (__ctfe)
+                    return true;
+
+                // Only check for properly aligned pointers
+                enum al = (void*).alignof - 1;
+                const base = cast(size_t) &source;
+                const alBase = (base + al) & ~al;
+                const elems = (n - (alBase - base)) / (void*).sizeof;
+
+                foreach (const s; (cast(void**) alBase)[0 .. elems])
+                {
+                    if (mayPointTo(s, target))
+                        return true;
+                }
+            }
+        }
+        else
+        {
+            foreach (size_t i; 0 .. S.length)
+                if (mayPointTo(source[i], target)) return true;
+        }
+
         return false;
     }
     else static if (isDynamicArray!S)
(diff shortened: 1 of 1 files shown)


std/exception.d
```diff
@@ -1191,9 +1194,16 @@ if (__traits(isRef, source) || isDynamicArray!S ||
     }
     else static if (isStaticArray!S)
     {
-        foreach (size_t i; 0 .. S.length)
-            if (mayPointTo(source[i], target)) return true;
-        return false;
+        static if (is(S == void[n], size_t n))
+        {
+            return true;
+        }
```
n8sh, line 1200: We can do more than this in `mayPointTo`. Minimally `return S.sizeof >= typeof(void*).sizeof` as we know that a data chunk smaller than a pointer cannot contain a pointer. Maximally we could interpret every properly-aligned pointer-sized chunk of the static array as if it were a pointer, which is safe since we never actually dereference the pointer but merely compare it to other pointers.

MoonlightSentinel: Implemented the latter

n8sh: Branch for CTFE?

std/exception.d
```diff
@@ -1440,6 +1464,33 @@ version (StdUnittest)
     assert( doesPointTo(ss, a));  //The array contains a struct that points to a
     assert(!doesPointTo(ss, b));  //The array doesn't contains a struct that points to b
     assert(!doesPointTo(ss, ss)); //The array doesn't point itself.
+
+    // https://issues.dlang.org/show_bug.cgi?id=20426
+    void[32] voidArr = void;
```
n8sh, line 1469:
```suggestion
    align((void*).alignof) void[32] voidArr = void;
```

std/exception.d
MoonlightSentinel: Added a `__ctfe` branch which pessimistically returns `true` because I didn't find a way to implement the actual logic without using a reinterpreting cast.

std/exception.d
```diff
@@ -1440,6 +1468,39 @@ version (StdUnittest)
     assert( doesPointTo(ss, a));  //The array contains a struct that points to a
     assert(!doesPointTo(ss, b));  //The array doesn't contains a struct that points to b
     assert(!doesPointTo(ss, ss)); //The array doesn't point itself.
+
+    // https://issues.dlang.org/show_bug.cgi?id=20426
+    align((void*).alignof) void[32] voidArr = void;
+    (cast(void*[]) voidArr[])[] = null; // Ensure no false pointers
+
+    *cast(void**) &voidArr[16] = &a; // Pointers should be found
+    *cast(void**) &voidArr[3] = &b;  // But only if they are properly aligned
```
n8sh, line 1477: I'm not sure if misaligned writes are guaranteed to work with LDC. I guess if some supported platform actually ignores the lowest few bits of a destination address that it "knows" is aligned on a word boundary, the automated test suite will tell us.

MoonlightSentinel: Would writing individual bytes using slice assignment work for all platforms? E.g.:

```
void*[1] tmp = [&b];
(cast(ubyte[]) voidArr[3 .. 3 + (void*).sizeof])[] = cast(ubyte[]) tmp[];
```

n8sh: Definitely.

std/exception.d
```diff
@@ -1191,8 +1194,33 @@ if (__traits(isRef, source) || isDynamicArray!S ||
...
     {
-        foreach (size_t i; 0 .. S.length)
-            if (mayPointTo(source[i], target)) return true;
+        static if (is(S == void[n], size_t n))
+        {
+            static if (n >= (void*).sizeof)
+            {
+                // Reinterpreting cast is impossible during ctfe
+                if (__ctfe)
+                    return true;
```
schveiguy, line 1203: I tried to make this work, you should be able to cast to some form of `ubyte[]`, but the compiler refuses to see anything but a `void[N]`. I tried slicing, slicing a pointer, and using a helper function.

n8sh:
Unfortunately there's a problem with the `mayPointTo(void[n], T)` check that didn't occur to me in time. Aside from containing pointers, the mystery memory region could also contain the language's built-in array slices that are represented as (length, pointer) pairs. Since we also don't know the element size, any (length, pointer) pair where the pointer is non-null and the length is non-zero potentially refers to the entire memory range from the pointer to the pointer plus the greatest multiple of the length that can be added to it without overflow.

IDK if it's worth checking this. It might be better to change `mayPointTo(void[n], T)` back to always returning true like in the first version of this PR.

schveiguy:
Hm... I think it's worth revising, the use case that prompted this is TaggedAlgebraic, which uses a `void[N]` for storage. That could easily contain an array. Given that `mayPointTo` returning false is supposed to be "100% sure", I'd say we should opt for the cautious version.

schveiguy:
We could still implement checks like if the `N` is 3, it can't be a pointer, or if it's 5, it can't be an array (on 32-bit), but I'd say that the easiest thing is to return `true`. It can be enhanced later if a need arises.

PR #7462, opened 2020-04-27
Followup to #7458 - Always return true for mayPointTo if the void[N] …

schveiguy:
…can contain a slice.

Note, there are very few checks we can do for actual slices. We could rule out an entirely zero'd array, or one where we can prove the length is always zero. But not knowing the "true" type of the array stored, we don't know what a non-zero length means.

You may want to view with whitespace off. I had to refactor the `return false` because otherwise dmd complains of unreachable statements.

### DIFF
--- a/std/exception.d
+++ b/std/exception.d
@@ -1196,7 +1196,15 @@ if (__traits(isRef, source) || isDynamicArray!S ||
     {
         static if (is(S == void[n], size_t n))
         {
-            static if (n >= (void*).sizeof)
+            static if (n >= (void[]).sizeof)
+            {
+                // could contain a slice, which could point at anything.
+                // But a void[N] that is all 0 cannot point anywhere
+                import std.algorithm.searching : any;
+                if (__ctfe || any(cast(ubyte[]) source[]))
+                    return true;
+            }
+            else static if (n >= (void*).sizeof)
             {
                 // Reinterpreting cast is impossible during ctfe
                 if (__ctfe)
@@ -1206,13 +1214,10 @@ if (__traits(isRef, source) || isDynamicArray!S ||
                 enum al = (void*).alignof - 1;
                 const base = cast(size_t) &source;
                 const alBase = (base + al) & ~al;
-                const elems = (n - (alBase - base)) / (void*).sizeof;
 
-                foreach (const s; (cast(void**) alBase)[0 .. elems])
-                {
-                    if (mayPointTo(s, target))
-                        return true;
-                }
+                if ((n - (alBase - base)) >= (void*).sizeof &&
+                    mayPointTo(*(cast(void**) alBase), target))
+                    return true;
             }
         }
         else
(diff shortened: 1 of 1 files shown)


schveiguy:
ping @n8sh @MoonlightSentinel

std/exception.d
```diff
@@ -1196,32 +1196,40 @@ if (__traits(isRef, source) || isDynamicArray!S ||
...
+                    // Only check for properly aligned pointers
+                    enum al = (void*).alignof - 1;
+                    const base = cast(size_t) &source;
+                    const alBase = (base + al) & ~al;
+                    const elems = (n - (alBase - base)) / (void*).sizeof;
+
+                    foreach (const s; (cast(void**) alBase)[0 .. elems])
+                    {
+                        if (mayPointTo(s, target))
+                            return true;
```
MoonlightSentinel, line 1221: This path could be simplified because the chunk can contain at most one aligned pointer, e.g.:

```
if ((n - (alBase - base)) >= (void*).sizeof && mayPointTo(*alBase, target))
    return true;
```

schveiguy:
Updated. I decided also to implement a zero check for larger `void[N]`, since one that is completely zero can't point anywhere. A benefit of this is that I can move the return false back to the bottom.

std/exception.d
```diff
@@ -1196,7 +1196,18 @@ if (__traits(isRef, source) || isDynamicArray!S ||
...
-            static if (n >= (void*).sizeof)
+            static if (n >= (void[]).sizeof)
+            {
+                // could contain a slice, which could point at anything.
+                // But a void[N[ that is all 0 cannot point anywhere
+                if (__ctfe)
+                    return true;
+                import std.algorithm.comparison : equal;
+                import std.range : repeat;
+                if (!(cast(ubyte[])source[]).equal(repeat(ubyte(0), n)))
```
MoonlightSentinel, line 1207:
```suggestion
                import std.algorithm : any;
                if (!any(cast(ubyte[]) source[]))
```

std/exception.d
```diff
@@ -1206,13 +1217,10 @@ if (__traits(isRef, source) || isDynamicArray!S ||
...
                 const alBase = (base + al) & ~al;
-                const elems = (n - (alBase - base)) / (void*).sizeof;
 
-                foreach (const s; (cast(void**) alBase)[0 .. elems])
-                {
-                    if (mayPointTo(s, target))
-                        return true;
-                }
+                if ((n - (alBase - base)) >= (void*).sizeof &&
+                    mayPointTo(*(cast(void**)alBase), target))
```
MoonlightSentinel, line 1222: DStyle enforces a space after `cast(...)`
```suggestion
                    mayPointTo(*(cast(void**) alBase), target))
```

schveiguy:
@MoonlightSentinel updated, thanks.

std/exception.d
```diff
@@ -1196,7 +1196,17 @@ if (__traits(isRef, source) || isDynamicArray!S ||
     {
         static if (is(S == void[n], size_t n))
         {
-            static if (n >= (void*).sizeof)
+            static if (n >= (void[]).sizeof)
+            {
+                // could contain a slice, which could point at anything.
+                // But a void[N[ that is all 0 cannot point anywhere
```
n8sh, line 1202:
```suggestion
                // But a void[N] that is all 0 cannot point anywhere
```

std/exception.d
```diff
@@ -1196,7 +1196,17 @@ if (__traits(isRef, source) || isDynamicArray!S ||
...
+                if (__ctfe)
+                    return true;
+                import std.algorithm.searching : any;
+                if (any(cast(ubyte[]) source[]))
```
MoonlightSentinel, line 1206: You could join both `if`s
```suggestion
                import std.algorithm.searching : any;
                if (__ctfe || any(cast(ubyte[]) source[]))
```

std/exception.d
schveiguy: 🤦

schveiguy:
Updated

schveiguy:
> You could add a CTFE test for the second branch

I can do this, hang on.

schveiguy:
Anyone? I can pull this, but I don't like to merge my own requests if I can help it.

PR #7463, opened 2020-04-28
Mostly fix 18559: std.math.* should stop using real overloads by default

thewilsonator:
Changes from old PR:`LN2` are are now cast to the appropriate precision, `0.5* _` now have the appropriate precision suffix.

Still to go, functions with `real` only implementations:
* remainder
* remquo
* NaN (creating one with a given payload)
* getNaNPayload
* logarithmic functions

Blocked on:
* https://github.com/Netflix/vectorflow/pull/33 merged, waiting on new tag.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -376,12 +376,10 @@ if ((is(immutable Num == immutable short) || is(immutable Num == immutable byte)
  */
 
 real cos(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.cos(x); }
-//FIXME
 ///ditto
-double cos(double x) @safe pure nothrow @nogc { return cos(cast(real) x); }
-//FIXME
+double cos(double x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.cos(x); }
 ///ditto
-float cos(float x) @safe pure nothrow @nogc { return cos(cast(real) x); }
+float cos(float x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.cos(x); }
 
 ///
 @safe unittest
@@ -418,12 +416,10 @@ float cos(float x) @safe pure nothrow @nogc { return cos(cast(real) x); }
  */
 
 real sin(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.sin(x); }
-//FIXME
 ///ditto
-double sin(double x) @safe pure nothrow @nogc { return sin(cast(real) x); }
-//FIXME
+double sin(double x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.sin(x); }
 ///ditto
-float sin(float x) @safe pure nothrow @nogc { return sin(cast(real) x); }
+float sin(float x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.sin(x); }
 
 ///
 @safe unittest
@@ -1563,35 +1559,6 @@ long rndtol(float x) @safe pure nothrow @nogc { return rndtol(cast(real) x); }
 {
     long function(real) prndtol = &rndtol;
     assert(prndtol != null);
-}
-
-/**
-$(RED Deprecated. Please use $(LREF round) instead.)
-
-Returns `x` rounded to a `long` value using the `FE_TONEAREST` rounding mode.
-If the integer value of `x` is greater than `long.max`, the result is
-indeterminate.
-
-Only works with the Digital Mars C Runtime.
-
-Params:
-    x = the number to round
-Returns:
-    `x` rounded to an integer value
- */
-deprecated("rndtonl is to be removed by 2.089. Please use round instead")
-extern (C) real rndtonl(real x);
-
-///
-deprecated @system unittest
-{
-    version (CRuntime_DigitalMars)
-    {
-        assert(rndtonl(1.0) is -real.nan);
-        assert(rndtonl(1.2) is -real.nan);
-        assert(rndtonl(1.7) is -real.nan);
-        assert(rndtonl(1.0001) is -real.nan);
-    }
 }
 
 /***************************************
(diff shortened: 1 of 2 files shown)


thewilsonator:
Supersedes https://github.com/dlang/phobos/pull/6257

thewilsonator:
@Geod24 note that this doesn't actually fix the entire issue as noted in the OP. NaN and getNaNPayload can probably be safely ignored, I doubt any one uses those. remainder, remquo, is a little more serious. But the most glaring one is the log functions.

Geod24:
@thewilsonator : Thanks for the precision.
For context, I did the renamed because I am subscribed to all dlang repository, and need to quickly sort through my notifications to see what's relevant to my use cases and knowledge, so titles like "Fix18559 1" tend to annoy me :)

std/math.d
```diff
@@ -1764,14 +1772,22 @@ real cosh(real x) @safe pure nothrow @nogc
...
+    const double y = exp(x);
+    return (y + 1.0/y) * 0.5;
+}
 
 /// ditto
-float cosh(float x) @safe pure nothrow @nogc  { return cosh(cast(real) x); }
+float cosh(float x) @safe pure nothrow @nogc
+{
+    const float y = exp(x);
+    return (y + 1.0/y) * 0.5f;
```
kinke, line 1789: Division, addition and multiplication all in double precision because of the `1.0`.

std/math.d
```diff
@@ -1849,10 +1883,28 @@ real tanh(real x) @safe pure nothrow @nogc
...
+float tanh(float x) @safe pure nothrow @nogc
+{
+    if (fabs(x) > float.mant_dig * cast(float)LN2)
+    {
+        return copysign(1, x);
+    }
+    
+    const float y = expm1(2*x);
+    return y / (y + 2);
+}
```
kinke, line 1907: I think a private impl template would be much nicer than duplicating the body 3 times.

std/math.d
thewilsonator: Fixed.

std/math.d
```diff
@@ -1764,14 +1772,22 @@ real cosh(real x) @safe pure nothrow @nogc
     //  cosh = (exp(x)+exp(-x))/2.
     // The naive implementation works correctly.
     const real y = exp(x);
-    return (y + 1.0/y) * 0.5;
+    return (y + 1.0/y) * 0.5L;
```
n8sh, line 1775:
```suggestion
    return (y + 1.0/y) * 0.5;
```
No code change was necessary here.

std/math.d
```diff
@@ -1802,18 +1818,36 @@ real sinh(real x) @safe pure nothrow @nogc
     // is x = 0.5 * (real.mant_dig) * LN2. // = 22.1807 for real80.
     if (fabs(x) > real.mant_dig * LN2)
     {
-        return copysign(0.5 * exp(fabs(x)), x);
+        return copysign(0.5L * exp(fabs(x)), x);
```
n8sh, line 1821: Code change appears unnecessary.

std/math.d
```diff
@@ -1802,18 +1818,36 @@ real sinh(real x) @safe pure nothrow @nogc
...
     }
 
     const real y = expm1(x);
-    return 0.5 * y / (y+1) * (y+2);
+    return 0.5L * y / (y+1) * (y+2);
```
n8sh, line 1825: Code change appears unnecessary.

std/math.d
```diff
@@ -2011,14 +2085,20 @@ float asinh(float x) @safe pure nothrow @nogc { return asinh(cast(real) x); }
 real atanh(real x) @safe pure nothrow @nogc
 {
     // log( (1+x)/(1-x) ) == log ( 1 + (2*x)/(1-x) )
-    return  0.5 * log1p( 2 * x / (1 - x) );
+    return  0.5L * log1p( 2 * x / (1 - x) );
```
n8sh, line 2088: Code change appears unnecessary.

std/math.d
```diff
@@ -8131,10 +8261,10 @@ if (isIntegral!(F) && isIntegral!(G))
 }
 
 /**Computes integer to floating point powers.*/
-real pow(I, F)(I x, F y) @nogc @trusted pure nothrow
+Unqual!F pow(I, F)(I x, F y) @nogc @trusted pure nothrow
 if (isIntegral!I && isFloatingPoint!F)
 {
-    return pow(cast(real) x, cast(Unqual!F) y);
+    return pow(cast(Unqual!F) x, cast(Unqual!F) y);
```
n8sh, line 8267: I would suggest changing the signature to `F pow(I, F)(const I, const F)` and also dispensing with the `Unqual!F` in the body.

Geod24:
> Netflix/vectorflow#33 merged, waiting on new tag.

That's a pretty obvious breaking change though :/

kinke:
> That's a pretty obvious breaking change though :/

Yeah, the adaptation is ugly and probably shouldn't be necessary.
```D
float foo(float a, float b);
double foo(double a, double b);
real foo(real a, real b);

alias T = double;
T bar(T x)
{
    pragma(msg, typeof(x * short(1)));
    return foo(x, short(1));
}
```

```
double
onlineapp.d(9): Error: `onlineapp.foo` called with argument types `(double, short)` matches both:
onlineapp.d(1):     `onlineapp.foo(float a, float b)`
and:
onlineapp.d(3):     `onlineapp.foo(real a, real b)`
```

Note that `foo(double, double)` is not even in the list (apparently depends on lexical order...).

I don't know whether that can/should be handled based on the types alone (a short to double conversion is guaranteed to be loss-less, so ``foo(double, double)` should be the obvious overload to be taken), or whether literals (e.g., common `0`, `1`, `0.5`...) can/should be treated specially.

andralex (dismissed):
please rebase

std/math.d
```diff
@@ -4646,13 +4613,20 @@ real nearbyint(real x) @safe pure nothrow @nogc
  * $(LREF nearbyint) performs the same operation, but does
  * not set the FE_INEXACT exception.
  */
-real rint(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.rint(x); }
-//FIXME
+real rint(real x) @safe pure nothrow @nogc
+{
+    pragma(inline, true); return core.math.rint(x);
```
kinke, line 4618: Please keep in mind that these `pragma(inline, true)` *statements* inside non-templates are ignored by LDC when imported by another module (function body needs semantic...).

thewilsonator: Ahh, noted. Although I suspect that LLVM's inline will be smart enough to do that with the hint.

kinke: That hint really needs to be outside the function, otherwise there's no LLVM hint ;) - our std.math is full of additions like this: https://github.com/ldc-developers/phobos/blob/0d04cf72770a356033e8e5fca657b9b7316bc957/std/math.d#L1614

PR #7465, opened 2020-04-29
Fix Issue 20783 - std.string.lastIndexOf doesn't work correctly in CTFE

wolframw:
Previously, the subsequence comparison was broken and returned a value > -1 if the first character matched (even if the rest of the sequence did not match).


burner:
@wolframw thank you

PR #7466, opened 2020-04-29
std.file.setTimes: on macOS use setattrlist to avoid truncating timestamp precision to microseconds

n8sh:
This was brought to my attention by [a unit test in CyberShadow/ae](https://github.com/CyberShadow/ae/blob/master/sys/persistence/core.d) that was failing on macOS because it was storing a result of `timeLastModified` then using the stored value as an argument to `setTimes` and expecting that the next call to `timeLastModified` would yield the same result as the stored value. The problem is that core.sys.posix.sys.stat does not have `utimensat` for macOS, which is because it only became available in macOS starting from version 10.13 which was released in 2017. As an alternative `setattrlist` can be used with any version of macOS to set timestamps with the same precision as `utimensat`.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -1483,6 +1483,20 @@ version (Windows) @system unittest
     }
 }
 
+version (Darwin)
+private
+{
+    import core.stdc.config : c_ulong;
+    enum ATTR_CMN_MODTIME  = 0x00000400, ATTR_CMN_ACCTIME  = 0x00001000;
+    alias attrgroup_t = uint;
+    static struct attrlist
+    {
+        ushort bitmapcount, reserved;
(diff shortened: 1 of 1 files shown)


CyberShadow:
Thank you!

What do you think about adding the `extern(C)` declarations to Druntime first, then using them here?

n8sh:
Created https://github.com/dlang/druntime/pull/3076. Let's see where it goes.

n8sh:
Changed this PR to target `stable`. Please consider moving forward with this bug fix without waiting for https://github.com/dlang/druntime/pull/3076. If/when that PR is accepted the code here can be updated to use it.

PR #7467, opened 2020-05-02
Only use 'from' if a packet was actually received

Ace17:
This fixes a systematic assertion error when trying to receiveFrom a
non-blocking UDP socket. In this case, 'from.addressFamily' is equal to UNSPEC,
eventually failing the comparison and crashing the program.

### DIFF
--- a/std/socket.d
+++ b/std/socket.d
@@ -3572,7 +3572,7 @@ class UdpSocket: Socket
     }
 }
 
-unittest
+@safe unittest
 {
     byte[] buf;
     buf.length = 1;
--- a/std/socket.d
+++ b/std/socket.d
@@ -3119,25 +3119,17 @@ public:
             from = createAddress();
         socklen_t nameLen = from.nameLen;
         version (Windows)
-        {
             auto read = .recvfrom(sock, buf.ptr, capToInt(buf.length), cast(int) flags, from.name, &nameLen);
-            if (read >= 0)
-            {
-                from.setNameLen(nameLen);
-                assert(from.addressFamily == _family);
-            }
-            return read;
-        }
+
         else
-        {
             auto read = .recvfrom(sock, buf.ptr, buf.length, cast(int) flags, from.name, &nameLen);
-            if (read >= 0)
-            {
-                from.setNameLen(nameLen);
-                assert(from.addressFamily == _family);
-            }
-            return read;
+
+        if (read >= 0)
+        {
+            from.setNameLen(nameLen);
+            assert(from.addressFamily == _family);
         }
+        return read;
     }
 
 
--- a/std/socket.d
+++ b/std/socket.d
@@ -3121,17 +3121,21 @@ public:
         version (Windows)
         {
             auto read = .recvfrom(sock, buf.ptr, capToInt(buf.length), cast(int) flags, from.name, &nameLen);
-            from.setNameLen(nameLen);
-            assert(from.addressFamily == _family);
-            // if (!read) //connection closed
+            if (read >= 0)
+            {
+                from.setNameLen(nameLen);
+                assert(from.addressFamily == _family);
+            }
             return read;
         }
         else
         {
             auto read = .recvfrom(sock, buf.ptr, buf.length, cast(int) flags, from.name, &nameLen);
-            from.setNameLen(nameLen);
-            assert(from.addressFamily == _family);
-            // if (!read) //connection closed
+            if (read >= 0)
+            {
+                from.setNameLen(nameLen);
+                assert(from.addressFamily == _family);
+            }
             return read;
         }
     }
@@ -3574,6 +3578,17 @@ class UdpSocket: Socket
     {
         this(AddressFamily.INET);
     }
+}
+
+unittest
+{
+    byte[] buf;
+    buf.length = 1;
+    Address addr;
+    auto s = new UdpSocket;
+    s.blocking = false;
+    s.bind(new InternetAddress(InternetAddress.PORT_ANY));
+    s.receiveFrom(buf, addr);
 }
 
 // https://issues.dlang.org/show_bug.cgi?id=16514


thewilsonator:
Does this have a bugzilla issue number? If not please create one and title the commit message `Fix issue ####: Only use 'from' if a packet was actually received`. Please also add a test case.

std/socket.d
```diff
@@ -3150,6 +3153,11 @@ public:
             from.setNameLen(nameLen);
             assert(from.addressFamily == _family);
             // if (!read) //connection closed
```
CyberShadow, line 3155: I think you mean to delete these lines, so they're replaced with the lines added below.

Ace17: Done!

CyberShadow:
Could you please add a `unittest` block which fails without this change, and succeeds after this change?

Ace17:
I have a few concerns regarding the relevance of a unittest here.

Here's what such a test would look like:
```d
unittest
{
  byte[] buf;
  buf.length = 256;
  Address addr;
  auto s = new UdpSocket;
  s.blocking = false;
  s.bind(new InternetAddress(10999)); // arbitrary, must be an available UDP port receiving no data
  s.receiveFrom(buf, addr);
}
```

Is it okay to establish network communication from a unittest? 
Basically, to reproduce the crash you need to create a nonblocking UDP socket and call receiveFrom while there's no data. This is neither isolated nor deterministic.
Moreover, it's not self-validating either: this is a smoke test at best, i.e there's no assertion in the unittest body.
If you remove the line "buf.length = 256", the test will always pass, and can't detect the bug anymore.

This is why I'm asking for your confirmation: do you really want to add this to Phobos' unittests?

CyberShadow:
> ```d
> // arbitrary, must be an available UDP port receiving no data
> ```

Can you use `InternetAddress.PORT_ANY` here?
And, maybe an address on the loopback interface, e.g. `127.12.34.56`?

> Is it okay to establish network communication from a unittest?

Yes, have a look at the existing tests.

> This is neither isolated nor deterministic.

Assuming you mean that the extent of this is that it will break if something is already listening on that port, or something just happens to send something to that port during the test's execution, then I think that's entirely acceptable.

> Moreover, it's not self-validating either: this is a smoke test at best, i.e there's no assertion in the unittest body.

This is fine too. Checking for successful execution of some code is a valid unittest.

CyberShadow:
> Here's what such a test would look like (I'm able to reproduce the crash with the following code):

This code runs successfully for me.

Ace17:
> Can you use InternetAddress.PORT_ANY here?

Yes, I'm still getting the crash with PORT_ANY. However, please note that listening to data on *any* UDP port actually increases the likelihood that some real packet is received on some port on the test machine. It might also increase the likelihood of interfering with other sockets of other processes.

> Checking for successful execution of some code is a valid unittest.

Provided that the code under test actually gets executed, which my unittest can't guarantee ( it would consider as "valid" a dummy socket implementation that does nothing ).

Here's how I'm reproducing the issue (I understand this isn't the exact way I'm supposed to run the tests). I can't explain how using the latest compiler might prevent the crash on your side, though. Are assertions enabled in stock prebuilt dmd packages ?
```
phobos % git diff
```
```
diff --git a/std/socket.d b/std/socket.d
index 49e4087..100e0f9 100644
--- a/std/socket.d
+++ b/std/socket.d
@@ -993,6 +993,16 @@ private AddressInfo[] getAddressInfoImpl(scope const(char)[] node, scope const(c
         "on this system.");
 }
 
+unittest
+{
+  byte[] buf;
+  buf.length = 256;
+  Address addr;
+  auto s = new UdpSocket;
+  s.blocking = false;
+  s.bind(new InternetAddress(10999)); // arbitrary, must be an available UDP port receiving no data
+  s.receiveFrom(buf, addr);
+}
 
 @safe unittest
 {
```
```
phobos % dmd | head -n2
```
```
DMD64 D Compiler v2.091.1
Copyright (C) 1999-2020 by The D Language Foundation, All Rights Reserved written by Walter Bright
```
```
phobos % rdmd -version=StdUnittest -main -unittest std/socket.d
```
```
std/socket.d(3161): [unittest] Assertion failure
core.exception.AssertError@std/socket.d(3161): Assertion failure
----------------
??:? _d_assertp [0x5578aef94cf5]
??:? @trusted long std.socket.Socket.receiveFrom(void[], std.socket.SocketFlags, ref std.socket.Address) [0x5578aef79698]
??:? @safe long std.socket.Socket.receiveFrom(void[], ref std.socket.Address) [0x5578aef796d1]
??:? @safe void std.socket.__unittest_L996_C1() [0x5578aef7530c]
??:? void std.socket.__modtest() [0x5578aef92089]
??:? int core.runtime.runModuleUnitTests().__foreachbody2(object.ModuleInfo*) [0x5578aefb74cc]
??:? int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)).__lambda2(immutable(object.ModuleInfo*)) [0x5578aef940d7]
??:? int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*))).__foreachbody2(ref rt.sections_elf_shared.DSO) [0x5578aef9c99f]
??:? int rt.sections_elf_shared.DSO.opApply(scope int delegate(ref rt.sections_elf_shared.DSO)) [0x5578aef9d07c]
??:? int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*))) [0x5578aef9c92d]
??:? int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)) [0x5578aef940a9]
??:? runModuleUnitTests [0x5578aefb7309]
??:? void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).runAll() [0x5578aef9795c]
??:? void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x5578aef978e8]
??:? _d_run_main2 [0x5578aef97850]
??:? _d_run_main [0x5578aef975cd]
??:? main [0x5578aef920e3]
??:? __libc_start_main [0x7f676422ce0a]
1/1 unittests FAILED
phobos %                                   
```

CyberShadow:
> I can't explain how using the latest compiler might prevent the crash on your side, though. Are assertions enabled in stock prebuilt dmd packages ?

Oh, I see. So, this crash happens only when Phobos is built with assertions enabled.

> Yes, I'm still getting the crash with PORT_ANY. However, please note that listening to data on _any_ UDP port actually increases the likelihood that some real packet is received on some port on the test machine.

I don't see why that would be. It's not like it's likely that the tests will be run on an Internet-facing machine, and even so, *how* a port is chosen shouldn't affect the probability that a random packet will find its way there.

Note that `PORT_ANY` doesn't actually allocate an absolutely arbitrary port. There is a range of ports dedicated specifically for `PORT_ANY` allocations.

Ace17:
> Oh, I see. So, this crash happens only when Phobos is built with assertions enabled.

Indeed (I initially got it from gdc on Debian).

I forced-pushed the commit including fix+unittest.

std/socket.d
```diff
@@ -3594,6 +3598,17 @@ class UdpSocket: Socket
...
 
+unittest
+{
+  byte[] buf;
+  buf.length = 1;
+  Address addr;
+  auto s = new UdpSocket;
+  s.blocking = false;
+  s.bind(new InternetAddress(InternetAddress.PORT_ANY));
+  s.receiveFrom(buf, addr);
```
CyberShadow, line 3609: Four spaces for indentation

std/socket.d
```diff
@@ -3139,17 +3139,21 @@ public:
...
-            from.setNameLen(nameLen);
-            assert(from.addressFamily == _family);
-            // if (!read) //connection closed
+            if (read >= 0)
+            {
+                from.setNameLen(nameLen);
+                assert(from.addressFamily == _family);
+            }
             return read;
         }
```
CyberShadow, line 3158: After the `recvfrom` call, both branches are identical, so there's an opportunity to DRY

CyberShadow:
> Does this have a bugzilla issue number? If not please create one and title the commit message `Fix issue ####: Only use 'from' if a packet was actually received`.

This still needs to be done.

RazvanN7:
ping @Ace17 . This needs a rebase and a bugzilla issue and we are good to go.

RazvanN7:
@CyberShadow I rebased and applied the feedback. Is this good to go?

CyberShadow (approved):

CyberShadow:
Thank you @RazvanN7 and @Ace17 :)

PR #7468, opened 2020-05-05
Make std.uni a package

thewilsonator:
Its way too big to comprehend upgrading to a new standard as is. More PRs will follow. Best reviewed by commit.

This series of PRs will need https://github.com/dlang/dmd/pull/11100 at some point


thewilsonator:
Hmm
```
grep -nrE '[[:alnum:]][.][.][[:alnum:]]|[[:alnum:]] [.][.][[:alnum:]]|[[:alnum:]][.][.] [[:alnum:]]' $(find etc std -name '*.d' | grep -vE 'std/string.d|std/uni.d') ; test $? -eq 1
std/uni/package.d:2427:        assert(unicode.ASCII.to!string == "[0..128$(RPAREN)");
std/uni/package.d:2472:            "[1024..1157) [1159..1320) [7467..7468) [7544..7545) [11744..11776) [42560..42648) [42655..42656)");
std/uni/package.d:2478:            "[0x400..0x485) [0x487..0x528) [0x1d2b..0x1d2c) [0x1d78..0x1d79) [0x2de0..0x2e00) "
std/uni/package.d:2479:            ~"[0xa640..0xa698) [0xa69f..0xa6a0)");
std/uni/package.d:2482:            "[0X400..0X485) [0X487..0X528) [0X1D2B..0X1D2C) [0X1D78..0X1D79) [0X2DE0..0X2E00) "
std/uni/package.d:2483:            ~"[0XA640..0XA698) [0XA69F..0XA6A0)");
std/uni/package.d:3118:    assert(unicode.ASCII.to!string() == "[0..128)");
```
What dscanner setting were used to originally suppress these?

PetarKirov:
@thewilsonator According to: https://github.com/dlang/phobos/blob/75553a70eb10dcfb4e33d2dac997a56f88f603e8/posix.mak#L595-597 I think that the idea was to enforce code style for array slicing, however since the regex is not powerful enough there can be false positives when `..` appears inside a string.

```
@echo "Enforce space between a .. b"
	grep -nrE '[[:alnum:]][.][.][[:alnum:]]|[[:alnum:]] [.][.][[:alnum:]]|[[:alnum:]][.][.] [[:alnum:]]' $$(find etc std -name '*.d' | grep -vE 'std/string.d|std/uni.d') ; test $$? -eq 1
```

thewilsonator:
> I think that the idea was to enforce code style for array slicing,

yes

>  however since the regex is not powerful enough there can be false positives when .. appears inside a string.

Then how was it passing before?

PetarKirov:
The simple solution is to change the `grep -vE 'std/string.d|std/uni.d'` part is `posix.mak`. The better one is to re-implement these checks using DScanner.

PetarKirov:
> Then how was it passing before?

It was passing before, because `std.string` and `std.uni` we're explicitly removed from these checks. Now that you changed the file structure, you need to update the grep whitelist.

thewilsonator:
Thanks, done.

DmitryOlshansky:
LGTM

PR #7470, opened 2020-05-06
Fix Issue 20799 - schwartzSort does not pin transformation results wi…

CyberShadow:
…th indirections, leading to memory corruption


thewilsonator:
Memory corruptions sound like it should be targeting stable.

CyberShadow:
-> https://github.com/dlang/phobos/pull/7471

PR #7471, opened 2020-05-06
Fix Issue 20799 - schwartzSort does not pin transformation results wi…

CyberShadow:
…th indirections, leading to memory corruption


MoonlightSentinel:
`auto-merge` got stuck because of the unrelated BuildKite failure.

PR #7481, opened 2020-05-13
std.math: use 'radians' for trig functions instead of 'x'

marler8997:
Once named parameters are enabled, the parameter name `x` doesn't convey any useful information to the caller, i.e.
```D
sin(x: 90);
```
However, the name `radians` would be useful for the client to include as it makes the unit type part of the API:
```D
sin(radians: 90);
```
Now it's immediately apparent that the value is in 'radians', and we see that the caller likely meant `PI/2` instead of `90`.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -645,24 +645,24 @@ if ((is(Unqual!Num == short) || is(Unqual!Num == byte)) ||
 }
 
 /***********************************
- * Returns cosine of x. x is in radians.
+ * Returns cosine of `radians`.
  *
  *      $(TABLE_SV
- *      $(TR $(TH x)                 $(TH cos(x)) $(TH invalid?))
- *      $(TR $(TD $(NAN))            $(TD $(NAN)) $(TD yes)     )
- *      $(TR $(TD $(PLUSMN)$(INFIN)) $(TD $(NAN)) $(TD yes)     )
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -1068,16 +1074,16 @@ private T tanImpl(T)(T x) @safe pure nothrow @nogc
  *      $(TR $(TD $(NAN))    $(TD $(NAN))  $(TD yes))
  *  )
  */
-real acos(real x) @safe pure nothrow @nogc
+real acos(real radians) @safe pure nothrow @nogc
```
dnadlinger, line 1077: `radians` doesn't make sense for the inverse functions (it applies to the return value).

marler8997: good catch that was a silly mistake

dnadlinger:
Unless the GitHub UI is deceiving me, looks like something went wrong when amending the commit – now no longer has the sin/cos changes, but still contains a bunch of spurious formatting changes.

marler8997:
> Unless the GitHub UI is deceiving me, looks like something went wrong when amending the commit – now no longer has the sin/cos changes, but still contains a bunch of spurious formatting changes.

Should be fixed now.

marler8997:
Closing due to Atilas comment: https://github.com/dlang/phobos/pull/7480#issuecomment-629609406

dnadlinger:
I still think it would have been useful here and there (with named parameters, of course).

PR #7473, opened 2020-05-09
std.math: Remove deprecated creal functions

ibuclaw:
First deprecation was in Jan 2018.

Unittests have been copied across to std.complex where appropirate.

PR #7483, opened 2020-05-15
Restore accidentally removed import from std.math

Geod24:
It was removed in PR7473: https://github.com/dlang/phobos/pull/7473#issuecomment-628176305
It caused a two-days-long outage of Buildkite: https://github.com/dlang/ci/pull/418#issuecomment-629223669

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -133,6 +133,9 @@ static import core.stdc.fenv;
 import std.traits :  CommonType, isFloatingPoint, isIntegral, isNumeric,
     isSigned, isUnsigned, Largest, Unqual;
 
+// Note: Exposed accidentally, should be deprecated / removed
+public import std.meta : AliasSeq;
+
 version (DigitalMars)
 {
     version = INLINE_YL2X;        // x87 has opcodes for these


MoonlightSentinel:
Wait, that is a private import ... that shouldn't be visible when importing `std.math`

Geod24:
It's a bug, selective imports create an alias. Some say it's a feature...

Geod24:
My bad, the bug have been fixed, but it was [public anyway](https://github.com/ibuclaw/phobos/blob/c71bd193507fe8d1836985393f10de68e5c0b7dd/std/math.d#L543).

MoonlightSentinel:
Thanks, did not see that `public:`.

Should this already deprecate the `import`?

Geod24:
> Should this already deprecate the `import`?

I think it should be deprecated, but I have no interest in deprecating that import *myself*.

I'm only raising this PR because the original reviewer 1) ignored the CI and 2) didn't act when the error started to creep up everywhere.  If this import is to be deprecated, it should be by someone that has interest in spending the time to do it, not the person fixing someone else's careless mistake.

And IMO this reasoning stands regardless of the amount of work required. Deprecating an import might be just a few minutes of time, but it's a few minutes more than I was willing to spend on Phobos when I started to work on my own stuff. This is also why I prefer to outright revert stuff than attempt to fix the mistake, as you've seen recently.

[We could call this the bug asymmetry principle, if you will](https://en.wikipedia.org/wiki/Brandolini%27s_law).

MoonlightSentinel:
Fair enough

dnadlinger:
Apparently the DAutoTest result upload failed… Is there a better way these days than just to empty-amend the branch?

wilzbach:
Nope unfortunately not, but in this case the force-merge option can be used to save some time.

ibuclaw:
@Geod24 - we're blocked by you. https://github.com/dlang/dmd/pull/10534

PR #7485, opened 2020-05-16
fix Issue 20836 - std.math: reorder declarations from most visible to least

ibuclaw:


ibuclaw:
Git(hub) has not done a very good job at creating the diff.  [This](https://pastebin.com/g4aKMVNy) is what the change really looks like.

PR #7487, opened 2020-05-18
std.math: Begin deprecation of accidentally leaked AliasSeq

ibuclaw:
~No, there isn't a better way of doing this.  Complements to @schveiguy for an initial version of this PR.~

Rebased and updated following dmd fix.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -133,7 +133,10 @@ static import core.stdc.fenv;
 import std.traits :  CommonType, isFloatingPoint, isIntegral, isNumeric,
     isSigned, isUnsigned, Largest, Unqual;
 
+// @@@DEPRECATED_2.102@@@
 // Note: Exposed accidentally, should be deprecated / removed
+deprecated("std.meta.AliasSeq was unintentionally available from std.math "
+           ~ "and will be removed after 2.102. Please import std.meta instead")
 public import std.meta : AliasSeq;
 
 version (DigitalMars)


MoonlightSentinel:
That workaround is really ugly...
 
The weird part is that this bug doesn't apply to deprecated function imports.

dnadlinger:
Maybe `std.meta.AliasSeq was unintentionally available from std.math and will be removed after …. Please import std.meta instead.` or something that makes clear where the symbol came from? If I saw this as an user not too familiar with metaprogramming in D, I would try to look up docs for std.math.AliasSeq to figure out what was going on.

schveiguy:
The worst part about this is that the deprecation message will be printed for the mixin line, not the usage. However, one possible saving grace is that the full instantiation is printed, so you might be able to deduce where it's coming from based on that.

std/math.d
```diff
@@ -133,8 +133,15 @@ static import core.stdc.fenv;
 import std.traits :  CommonType, isFloatingPoint, isIntegral, isNumeric,
     isSigned, isUnsigned, Largest, Unqual;
 
+// @@@DEPRECATED_2.102@@@
 // Note: Exposed accidentally, should be deprecated / removed
-public import std.meta : AliasSeq;
+private template DeprecatedAliasSeq(V...)
```
schveiguy, line 138: I think you need to make this public, no?

ibuclaw: No, otherwise we'd have to deprecated the DeprecatedAliasSeq too.

schveiguy: Can you publicly alias a private symbol? I didn't think that was possible.

ibuclaw: It works for me.

schveiguy: OK, you are right, I tested locally and it works.

MoonlightSentinel:
Please don't merge this yet,  I think i have a fix for the missing deprecation.

std/math.d
```diff
@@ -133,8 +133,15 @@ static import core.stdc.fenv;
...
+{
+    deprecated("std.meta.AliasSeq was unintentionally available from std.math and will be removed after 2.102. Please import std.meta instead")
```
schveiguy, line 140: You need to break this line in half, style check complains.

schveiguy:
Should we close this now that https://github.com/dlang/dmd/pull/11158 is merged?

MoonlightSentinel:
Not really, just add a plain deprecation message?

schveiguy:
Right, maybe adjust this to properly deprecate the import.

12345swordy:
@ibuclaw ping

ibuclaw:
Rebased.

PR #7488, opened 2020-05-20
std.typecons: Fix inconsistent indentation in Nullable implementation

CyberShadow:
It was previously difficult to see where the implementation of
Nullable ends, because the indentation did not represent the nesting
level.

This also fixes the different DDoc styles used to document Nullable
entries to a single style.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2678,12 +2678,12 @@ struct Nullable(T)
 
     private bool _isNull = true;
 
-/**
-Constructor initializing `this` with `value`.
-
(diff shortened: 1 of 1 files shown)


CyberShadow:
I think the unit tests should actually be moved outside of `Nullable`, because as things are, they are instantiated once for each `T` in every `Nullable!T`. If they were to test the behaviour of `Nullable` with that `T`, then their current position would make sense, but that doesn't seem to be the case. But that's orthogonal to this change.

PR #7489, opened 2020-05-20
Fix issue 20850 - Can't assign enum of Tuple

Biotronic:
`areCompatibleTuples` has to check `OriginalType` to account for enums. Also, swap is confused, so we're not using that for these assignments. Casting to `OriginalType` would work, but breaks `@safe`.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -545,7 +545,7 @@ if (distinctFieldNames!(Specs))
         }
     }
 
-    enum areCompatibleTuples(Tup1, Tup2, string op) = isTuple!Tup2 && is(typeof(
+    enum areCompatibleTuples(Tup1, Tup2, string op) = isTuple!(OriginalType!Tup2) && is(typeof(
     (ref Tup1 tup1, ref Tup2 tup2)
     {
         static assert(tup1.field.length == tup2.field.length);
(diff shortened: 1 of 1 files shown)


schveiguy:
I actually just envisioned that assignment to the same type would be supported. For instance, a type that alias this' itself to a `Tuple` should be assignable to that type.

Biotronic:
If you mean this should work, it does:
```d
import std.typecons;

struct S {
    alias t this;
    Tuple!int t;
}

unittest {
    Tuple!int a;
    S s;
    a = s;
}
```
Assigning `s = a;` doesn't work, but I'm pretty sure that can't be fixed in `Tuple`.

Strangely, `a = S();` also doesn't work. That seems like a DMD issue though, since the above code works.

I don't think there's a way to allow this to work:

```
import std.typecons;

enum T1 : Tuple!(int*) { a = T1(null) }
enum T2 : Tuple!(int*) { a = T2(null) }

unittest {
    T1 t1;
    T2 t2;
    // Allow assignment from T1 to T1:
    t1 = T1.a;
    // But not T2 to T1:
    static assert(!__traits(compiles, t1 = T2.a));
}
```

schveiguy:
Well, that's not what I expected. That seems like an odd difference between enum and alias this.

schveiguy:
https://issues.dlang.org/show_bug.cgi?id=20854

PR #7490, opened 2020-05-21
speed up instantiation of Filter and staticMap templates

WalterBright:
by simply unrolling the template instantiations


UplinkCoder:
I cannot believe this.

PR #7491, opened 2020-05-21
speed up instantation of Reverse and Filter templates

WalterBright:
by simply unrolling the template instantiations


WalterBright:
@Geod24 then please check out https://github.com/dlang/phobos/pull/7490 !

PR #7495, opened 2020-05-25
Remove all Replace definitions except for alias

andralex:
Unittests pass so I'm putting this up for review. If there was a reason for all, we'll see soon enough.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -543,24 +543,6 @@ template NoDuplicates(TList...)
  * Returns an `AliasSeq` created from TList with the first occurrence
  * of type T, if found, replaced with type U.
  */
-template Replace(T, U, TList...)
-{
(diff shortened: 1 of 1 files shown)


andralex:
> Best get rid of GenericReplace and move that into Replace. See also ReplaceAll

Right, that's the grand plan. Per Chesterton's fence rule, I'm waiting for comments from the folks who introduced the code to make sure I'm not tripping any wires.

andralex:
> See also ReplaceAll

See also a third of this module :)

thewilsonator:
Touché.

andralex:
I'll be back with a more substantive rework.

dnadlinger:
This would have been because basic types used to be incompatible with alias parameters.

PR #7497, opened 2020-05-26
staticMap: explain design

WalterBright:
It isn't obvious why the implementation strategy is used, so added an explanation.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -858,11 +858,11 @@ template staticMap(alias F, T...)
     else
     {
         /* While:
-         *   alias staticMap = aliasSeq!(F!T[0], staticMap!(F, T[0 .. $]));
+         *   alias staticMap = AliasSeq!(F!T[0], staticMap!(F, T[1 .. $]));
          * does fewer template instantiations, the compiler implements
          * recursive template instantiations with recursion, and long
          * sequences overflow the compiler's stack.
-         * The divide-and-conquer approach uses the square root as much stack.
+         * The divide-and-conquer approach uses log_2(n) stack frames.
          */
         alias staticMap =
             AliasSeq!(
--- a/std/meta.d
+++ b/std/meta.d
@@ -857,6 +857,13 @@ template staticMap(alias F, T...)
 
     else
     {
+        /* While:
+         *   alias staticMap = aliasSeq!(F!T[0], staticMap!(F, T[0 .. $]));
+         * does fewer template instantiations, the compiler implements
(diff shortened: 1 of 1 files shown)


UplinkCoder:
:+1:

dnadlinger:
Hmm, wasn't the original design intent rather to reduce the total amount of mangled name bytes?

adamdruppe:
I'd also probably argue there should be another threshold on that too. idk where, what N should be, but the simple recursion might actually be better if T.length < N, else try the divide and conquer.

Though the < 8 cases special-cased by hand might already handle this.

Or we could do a string mixin that works for all N reasonably well.

Better to comment the existing behavior than nothing though so I'd merge this regardless if i had the pwoer.

dnadlinger:
> Hmm, wasn't the original design intent rather to reduce the total amount of mangled name bytes?

Misremembered that, this must have been some other template.

Is it square root, though, or log?

WalterBright:
Maybe it's log2(n)

FeepingCreature:
It's log_2. Each depth halves the set size, so given a depth n the total number of steps is 2^n.

std/meta.d
```diff
@@ -857,6 +857,13 @@ template staticMap(alias F, T...)
 
     else
     {
+        /* While:
+         *   alias staticMap = aliasSeq!(F!T[0], staticMap!(F, T[0 .. $]));
```
schveiguy, line 861: Even though it's an uncompiled comment, The second parameter should be `T[1 .. $]`

schveiguy: And `AliasSeq` should be capitalized.

adamdruppe:
> Hmm, wasn't the original design intent rather to reduce the total amount of mangled name bytes?

staticMap should never mangle to anything at all nor affect anything else, since it just yields an alias of an existing thing.

And as far as I can tell, it indeed doesn't in the generated binary. Not sure about inside dmd itself though but it probably shouldn't there either.

schveiguy:
> Not sure about inside dmd itself though but it probably shouldn't there either.

It gets into the symbol table. This is why it consumes DMD memory. One of the benefits of possibly having the `...` improvement, or using type functions.

dnadlinger:
> It's log_2. Each depth halves the set size, so given a depth n the total number of steps is 2^n.

That's what I said. Clearly I'm becoming too British for my own good. :P

schveiguy:
This is trivial, I just made the changes myself.

PR #7498, opened 2020-05-26
use auto return instead of MaxType and MinType

WalterBright:
By using `auto` we avoid expensive instantiations of `MaxType` and `MinType`.


WalterBright:
This isn't going to work

PR #7499, opened 2020-05-26
specialize max/min for common case

WalterBright:
because the general case instantiates a large number of templates.

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -1681,7 +1681,7 @@ if (is(T == U) && is(typeof(a < b)))
    /* Handle the common case without all the template expansions
     * of the general case
     */
-    return a < b ? a : b;
+    return b < a ? b : a;
 }
 
 
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -1516,7 +1516,7 @@ Iterates the passed arguments and returns the maximum value.
 
 Params:
     args = The values to select the maximum from. At least two arguments must
-    be passed, and they must be comparable with `>`.
+    be passed, and they must be comparable with `<`.
 
 Returns:
     The maximum of the passed-in values. The type of the returned value is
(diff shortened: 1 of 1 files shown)


std/algorithm/comparison.d
```diff
@@ -1568,6 +1568,15 @@ if (T.length >= 2)
     return cast(typeof(return)) (chooseB ? b : a);
 }
 
+/*****
+ * Handle the common case without all the template expansions
+ * of the general case
+ */
+T max(T, U)(T a, U b) if (is(T == U) && is(typeof(a < b)))
```
Geod24, line 1575: If `is(T == U)` why not simply:

```suggestion
T max(T)(T a, T b) if (is(typeof(a < b)))
```

?

WalterBright: I tried that. It doesn't work:
```
T max(T)(T a, T b) { return 0; }

bool max(T...)(T t) { return true; }

void test()
{
    int a = -10;
    uint f = 10;
    pragma(msg, typeof(max(f, a)));
}
```
prints `uint`

UplinkCoder:
If it does cut down on instances I am all for it.
Can't believe it's used that often tough?
Perhaps 1024 different instances at most?

std/algorithm/comparison.d
```diff
@@ -1568,6 +1568,16 @@ if (T.length >= 2)
     return cast(typeof(return)) (chooseB ? b : a);
 }
 
+///
```
n8sh, line 1571:
```suggestion
/// ditto
```

std/algorithm/comparison.d
```diff
@@ -1688,6 +1698,17 @@ if (T.length >= 2)
     return cast(typeof(return)) (chooseA ? a : b);
 }
 
+///
```
n8sh, line 1701:
```suggestion
/// ditto
```

RazvanN7:
@WalterBright this fails the style checker:

```
std/algorithm/comparison.d(1544:11)[warn]: Public declaration 'max' has no documented example.
std/algorithm/comparison.d(1674:11)[warn]: Public declaration 'min' has no documented example.
```

RazvanN7 (approved):

PR #7500, opened 2020-05-26
alias FSChar = char for a wider range of platforms

denizzzka:
Especially for non-windows && non-posix

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -107,16 +107,11 @@ version (Windows)
 {
     private alias FSChar = wchar;
 }
-else version (Posix)
-{
-    private alias FSChar = char;
-}
-else version (GENERIC_IO)
-{
-    private alias FSChar = char;
-}
 else
-    static assert(0);
+{
+    private alias FSChar = char;
+}
+
 
 version (Windows)
 {
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -111,6 +111,10 @@ else version (Posix)
 {
     private alias FSChar = char;
 }
+else version (GENERIC_IO)
+{
+    private alias FSChar = char;
+}
 else
     static assert(0);
 


std/stdio.d
```diff
@@ -107,7 +107,11 @@ version (Windows)
 {
     private alias FSChar = wchar;
 }
-else version (Posix)
+else version (linux)
+{
+    private alias FSChar = char;
+}
+else version (GENERIC_IO)
```
atilaneves, line 114: What's GENERIC_IO?

schveiguy: It's set on a number of platforms. Look at the file. I don't really know how to review, because I don't know all those platforms.

denizzzka: And if some platform isn't covered static assert will catch it

schveiguy: Looking at the different platforms, the following set `GENERIC_IO`, and I don't know if there are any that aren't `version(Posix)`:

```
CRuntime_Bionic,
CRuntime_Musl,
CRuntime_UClibc,
OSX,
iOS,
TVOS,
WatchOS,
FreeBSD,
NetBSD,
DragonFlyBSD,
Solaris
```

So I'm not sure which platforms this is targeting. Or is this intended as a "just in case"? If so, I'd suspect that once that platform is added, it would need to be addressed when that platform is added.

denizzzka: I use it for ARM semihosting output on non-posix bare metal device.

schveiguy: > `"dwarfeh(364) fatal error"` is printed from druntime

and doesn't with this change?

schveiguy: So what OS version is defined for that configuration?

denizzzka: > and doesn't with this change?

yes

> Oh wait, it works, because of this change, and properly prints that. I got it.

It isn't compiles without this change, because here is not Windows and also not Posix version is set.

> So what OS version is defined for that configuration? And what version is defined that adds GENERIC_IO. In other words, what is the version configuration that applies to this case.

I use ldc2 with `--mtriple=thumbv7m-unknown-none-eabi` option. Looks like no OS version provided.

But here is available druntime and not posix-compilant libc.

(Also it is impossible to pass Posix version into compiler - it forbids passing of built-in versions values.)

schveiguy: when you run ldc2 -v what does it list for `predefs`?

denizzzka: DruntimeAbstractRt CRuntime_Abstract AsmExternal GENERIC_IO CoreUnittest LDC all D_Version2 unittest assert D_NoBoundsChecks D_ModuleInfo D_Exceptions D_TypeInfo ARM Thumb ARM_Thumb ARM_SoftFP D_HardFloat LittleEndian LDC_LLVM_1000

std/stdio.d
```diff
@@ -107,7 +107,11 @@ version (Windows)
 {
     private alias FSChar = wchar;
 }
-else version (Posix)
+else version (linux)
```
n8sh, line 110: Why not just keep as `else version (Posix)`?

denizzzka: https://github.com/dlang/phobos/pull/7500#discussion_r431447218

denizzzka: So more clearly, for my opinion

std/stdio.d
schveiguy: Wow, so `GENERIC_IO` is added by ldc2? that surprises me. And I think I have to agree with this change then.

std/stdio.d
schveiguy: Yeah, this needs to be switched back. Just add the case for your new situation, and don't change the existing situations.

std/stdio.d
denizzzka: > Wow, so GENERIC_IO is added by ldc2?

No, I added it from cmd line while compiling

std/stdio.d
denizzzka: done

std/stdio.d
schveiguy: OK, that makes sense. Yeah, I understand now. Honestly, with no defined OS, it should make sense to be able to configure whichever way you want. I'm OK with the changes then.

kinke:
`FSChar` has absolutely nothing to do with the `*_IO` versions, see https://github.com/dlang/phobos/blob/5ebf458b509963725e1143b733fba1b22f22ed3f/std/stdio.d#L129-L257.

I think the simplest solution is to define it as `wchar` for Windows, and `char` for *everything* else. If another OS pops up with non-char, fine, let's add the special case then instead of needlessly harassing users of exotic or no OS.

schveiguy:
@kinke yeah, doing a search for FSChar seems to show that it's literally only used to specify file names, and has nothing to do with the IO structures. @denizzzka can you simply change the code for the `else version(Posix)` to `else`?

n8sh:
@kinke's and @schveiguy's feedback has been addressed so merging.

PR #7501, opened 2020-05-28
add ascii to functional

heavyk:
ran into this when trying to install dlangide

fixes https://github.com/buggins/dlangide/issues/398

### DIFF
--- a/std/functional.d
+++ b/std/functional.d
@@ -68,6 +68,7 @@ module std.functional;
 
 import std.meta : AliasSeq, Reverse;
 import std.traits : isCallable, Parameters;
+import std.ascii;
 
 
 private template needOpCallAlias(alias fun)


std/functional.d
```diff
@@ -68,6 +68,7 @@ module std.functional;
 
 import std.meta : AliasSeq, Reverse;
 import std.traits : isCallable, Parameters;
+import std.ascii;
```
atilaneves, line 71: Surely this can't be the way forward. Why exactly can't this be local?

thewilsonator: See the other two comments.

atilaneves: I did. Still not 100% sure.

heavyk:
ok, just let me know what you want me to change when it's decided :)

heavyk:
@NilsLankila yes that seems quite appropriate, actually. nevermind then :)

PR #7502, opened 2020-05-28
Fix Issue 20869 - `move` is overly trusting of `opPostMove`

MoonlightSentinel:
Remove the manual check whether move is `@safe` and instead let the compiler do the attribute interference by adding appropriate `@trusted` blocks.

Note: I could've extend the current checks `trustedMoveImpl` but that could easily miss other corner cases.

TLDR: Only use `@trusted` when necessary

### DIFF
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -1059,11 +1059,7 @@ Params:
 */
 void move(T)(ref T source, ref T target)
 {
-    // test @safe destructible
-    static if (__traits(compiles, (T t) @safe {}))
-        trustedMoveImpl(source, target);
-    else
-        moveImpl(source, target);
+    moveImpl(source, target);
 }
 
 /// For non-struct types, `move` just performs `target = source`:
@@ -1178,11 +1174,7 @@ pure nothrow @safe @nogc unittest
 /// Ditto
 T move(T)(return scope ref T source)
 {
-    // test @safe destructible
-    static if (__traits(compiles, (T t) @safe {}))
-        return trustedMoveImpl(source);
-    else
-        return moveImpl(source);
+    return moveImpl(source);
 }
 
 /// Non-copyable structs can still be moved:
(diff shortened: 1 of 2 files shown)


thewilsonator:
Please target stable.

MoonlightSentinel:
> Please target stable.

Is stable broken?

```
std/math.d(4423): Error: `core.math.yl2x` called with argument types `(real, int)` matches both:
../druntime/import/core/math.d(167):     `core.math.yl2x(float x, float y)`
and:
../druntime/import/core/math.d(169):     `core.math.yl2x(real x, real y)`
posix.mak:326: recipe for target 'generated/linux/debug/64/libphobos2.a' failed
```

std/algorithm/mutation.d
```diff
@@ -1219,9 +1211,22 @@ pure nothrow @safe @nogc unittest
     assert(s2.a == 42);
 }
 
-private void trustedMoveImpl(T)(ref T source, ref T target) @trusted
+// https://issues.dlang.org/show_bug.cgi?id=20869
+// `move` should propagate the attributes of `opPostMove`
+unittest
```
Geod24, line 1216: Needs to be explicitly attributed no ?

MoonlightSentinel: Don't think that attributes matter for this `unittest` as it doesn't call any function.

n8sh: Doesn't matter but the style checker demands it.

MoonlightSentinel: Added `@system`

MoonlightSentinel:
Rebased to master because this might break code relying on the invalid `@trusted` (and stable seems to be broken ATM).

n8sh:
> Is stable broken?

The broken compilation is because it's using `stable` phobos with `master` druntime instead of `stable` with `stable`.

PR #7505, opened 2020-05-29
Size(). Make description more clear and add example

welkam:
what title says

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -3326,7 +3326,21 @@ void main()
         assert(dcharsOut == "foo");
     }
 
-/// Get the size of the file, ulong.max if file is not searchable, but still throws if an actual error occurs.
+/** Get the size of the file in bytes, ulong.max if file is not searchable, but still throws if an actual error occurs.
+Example:
+---
+import std.stdio, std.file;
+
+void main()
+{
+	auto file_handle = File(deleteme, "w");
+	file_handle.write("abc"); //create temporary file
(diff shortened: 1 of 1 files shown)


std/stdio.d
```diff
@@ -3326,7 +3326,21 @@ void main()
         assert(dcharsOut == "foo");
     }
 
-/// Get the size of the file, ulong.max if file is not searchable, but still throws if an actual error occurs.
+/** Get the size of the file in bytes, ulong.max if file is not searchable, but still throws if an actual error occurs.
```
thewilsonator, line 3329: Get_s_ the size of the file in bytes, ulong.max if file is not searchable, ~~but still~~ _or_  throws if an ~~actual error occurs~~ the operation fails.

std/stdio.d
```diff
@@ -3326,7 +3326,21 @@ void main()
...
+Example:
+---
+import std.stdio, std.file;
+
+void main()
+{
+    auto file_handle = File(deleteme, "w");
```
n8sh, line 3336: `deleteme` probably shouldn't be appearing in documented examples!

```d
// Purposefully not documented. Use at your own risk
@property string deleteme() @safe
```

welkam: It is used in many places so I just tried to be consistent. I agree with you that it probably should not be used in examples but that means other examples need to be changed. If this pull request get merged I will change other examples to be consistent with this change

std/stdio.d
welkam: Incorporated your suggestion and then some

PR #7506, opened 2020-05-29
Fix  issue 20665 - std.concurrency.spawn should document not working with delegates

Biotronic:
Improved compiler error message and amended documentation to document the limitation.

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -511,9 +511,8 @@ private template isSpawnable(F, T...)
  *  pointer indirection.  This is necessary for enforcing isolation among
  *  threads.
  *
- * Similarly, `fn` cannot be a `delegate`, as the type system can not
- * guarantee that a delegate does not have unshared aliasing.
- */
+ * Similarly, if `fn` is a delegate, it must not have unshared aliases, meaning
+ * `fn` must be either `shared` or `immutable`. */
 Tid spawn(F, T...)(F fn, T args)
 if (isSpawnable!(F, T))
 {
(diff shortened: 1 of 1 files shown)


thewilsonator:
```
std/concurrency.d(595): Error: static assert:  `__traits(compiles, spawn(F, T...)(F fn, T args) if (isSpawnable!(F, T))(dg1, 1))` is false
```

std/concurrency.d
```diff
@@ -498,6 +501,12 @@ if (isSpawnable!(F, T))
     assert(receiveOnly!string == "This is so great!");
 }
 
+void spawn(F, T...)(F fn, T args)
```
kubo39, line 504: Could you explain why use `void` instead of `Tid` ?

Biotronic: Because it's going to fail anyway, I guess. I didn't really think about it. Mayhap `Tid` would be better.

RazvanN7:
@thewilsonator I have modified this PR to reject only delegates that are not shared or immutable and documented that behavior. Since I deviated from the original fix, I think someone else shoud merge this.

thewilsonator:
```
std/concurrency.d(520): Error: static assert:  "Cannot spawn a delegate that is not `shared` or `immutable`"
generated/linux/debug/64/publictests/std_concurrency.d(107):        instantiated from here: `spawn!(void function(Tid) @system, Tid)`
```

RazvanN7:
Apparently, `is(void delegate(int) shared == shared)` yields false due to: https://issues.dlang.org/show_bug.cgi?id=1983 . Damn it!

RazvanN7:
I gave up on improving the error message. I updated the documentation. Delegates that mutate unshared data are still going to fail with a non-perfect error mesage, but at least the thing is documented now.

@thewilsonator I think that we can merge this now.

PR #7507, opened 2020-05-29
Fix #10380 - Memoize should handle lambdas

### LINKED ISSUE #10380: Memoize should handle lambdas
### Description

````markdown
Currently, std.functional.memoize uses std.traits.ReturnType and std.traits.Parameters to get the return type and parameters of the memoized function. This fails for stuff like memoize!(a => a.field). Suggested new implementation:

template memoize(alias fun) {
    auto memoize(Args...)(Args args) if (is(typeof(fun(args)))) {
        import std.typecons : Tuple;

        static if (__traits(isTemplate, fun)) {
            import std.traits : isDelegate;
            static assert(!isDelegate!(fun!Args), fun.stringof~" is a delegate, and thus has context memoize can't access.");
        }

        static typeof(fun(args))[Tuple!Args] memo;
        auto t = Tuple!Args(args);
        if (auto p = t in memo)
            return *p;
        return memo[t] = fun(args);
    }
}
````

### PR DESCRIPTION
**Pull request by Biotronic**

### CODE REVIEW

File: std/functional.d [hunk 1]
```diff
@@ -1260,17 +1260,12 @@ Note:
 */
 template memoize(alias fun)
 {
-    import std.traits : ReturnType;
-     // https://issues.dlang.org/show_bug.cgi?id=13580
-    // alias Args = Parameters!fun;
-
-    ReturnType!fun memoize(Parameters!fun args)
+    auto memoize(Args...)(Args args) if (is(typeof(fun(args))))
```
n8sh, line 1263: This could cause unexpected breakage. This code works right now but won't work after this PR:
``` d
import std.functional : memoize;
static Object objNum(int a) { return new Object(); }
assert(memoize!objNum(0) is memoize!objNum(0U)); 
```

std/functional.d
```diff
@@ -1400,6 +1393,16 @@ unittest
     assert(fact(10) == 3628800);
 }
 
+// Issue 20099
+@system unittest // not @safe due to memoize
+{
+    int i = 3;
+    alias a = memoize!((n) => i + n);
```
kinke, line 1400: I think having to add the param types for lambdas is acceptable to work around this: `memoize!((int n) => i + n)`, and that the danger illustrated by n8sh above isn't worth the trouble.

kinke: For user-friendliness, we could catch that case though, e.g.:
```D
static if (fun.stringof.length > 8 && fun.stringof[0 .. 8] == "__lambda" && !isCallable!fun)
    static assert(false, "memoized lambdas require explicit parameter types");
```
=>
```
onlineapp.d(6): Error: static assert:  "memoized lambdas require explicit parameter types"
onlineapp.d(29):        instantiated from here: `memoize!((n) => i + n)`
```

std/functional.d
```diff
@@ -1260,17 +1260,12 @@ Note:
...
     {
-        alias Args = Parameters!fun;
         import std.typecons : Tuple;
         import std.traits : Unqual;
 
-        static Unqual!(ReturnType!fun)[Tuple!Args] memo;
+        static Unqual!(typeof(fun(args)))[Tuple!Args] memo;
         auto t = Tuple!Args(args);
         if (auto p = t in memo)
             return *p;
```
kinke, line 1271: Lookup and optional insertion could be handled at once via `return memo.require(Tuple!Args(args), fun(args));`.

Biotronic:
That should take care of the issues that have been pointed out, and also makes `memoize` handle overloads, which confused it before.

kinke:
I don't think it does. Try
```D
alias foo = memoize!(n => new Object());
assert(foo(0) is memoize(0U);
```

Biotronic:
I don't see the problem there - you have a templated function and you expect it to be the same with different template arguments? First, that's a completely different issue from the above. Second, it's one that cannot possibly be handled sensibly in the language today. Third, it's code that didn't compile before, so it doesn't break any existing code.

kinke:
You open the door for a single `memoize!Func` instantiation to break down into a multitude of actually memoized functions, depending on the used args, just for the sake of saving to type down the explicit parameter types for lambdas.

I think a memoize!F instance should stick binding to a single function. The user can always declare some `alias memoizedFun(T) = memoize!((T p) => make(p))` in case a templated wrapper is actually desired.

Biotronic:
I do see that, but I think the issue is much smaller than you make it out to be. Note also that you have the same issue for explicit template functions:

```d
int foo(T)(T t) {
    return T.stringof.length;
}
unittest {
    assert(memoize!foo(0) != memoize!foo(0U));
}
```

I hope the above both compiles and passes the assert.

We could in fact test for lamba-osity by static asserting on `fun.stringof[0..8] == "__lambda"`, if we find that this is a big issue.

n8sh:
> I hope the above both compiles and passes the assert.

It turns out that prior to this PR that code doesn't compile because `Parameters!foo` doesn't work for an uninstantiated template:

```
Error: template instance std.traits.Parameters!(foo) does not match template declaration
 Parameters(func...)
  with func = (foo(T)(T t))
  must satisfy the following constraint:
       isCallable!func
```

kinke:
Yes, it currently doesn't support any templates and always binds to *one* function (thus the 'confusion' for overloads), and the short-hand lambda syntax is one such case, but not so obvious by the produced error msges.

std/functional.d
```diff
@@ -1400,6 +1451,37 @@ unittest
...
+    alias a = memoize!((n) => i + n);
+    alias b = memoize!((n) => i + n, 3);
+
+    assert(a(3) == 6);
+    assert(b(3) == 6);
+}
+
+@system unittest // not @safe due to memoize
+{
+    static Object objNum(int a) { return new Object(); }
```
n8sh, line 1467: If you annotate `objNum` with `@safe` the test can be `@safe`.

kinke:
> I do see that, but I think the issue is much smaller than you make it out to be.

You're probably right. I suppose the memoized wrapper now cleanly wraps a symbol, incl. overloads and templates, and doesn't expand the domain of argument types.

std/functional.d
Biotronic: Nope, since `memoize`'s second overload isn't `@safe`.

std/functional.d
Biotronic: Sadly, `require` isn't `nothrow`, so that's a `no-go`.

std/functional.d
Biotronic: That case is now handled, except for the case discussed below, which does not introduce breakage.

RazvanN7:
@Biotronic can you please rebase so we can get this in?

std/functional.d
```diff
@@ -1225,6 +1225,15 @@ alias pipe(fun...) = compose!(Reverse!(fun));
     assert(compose!(`a + 0.5`, `to!(int)(a) + 1`, foo)(1) == 2.5);
 }
 
+private template getOverloads(alias fun)
+{
+    import std.meta : AliasSeq;
+    static if (__traits(compiles, __traits(getOverloads, __traits(parent, fun), __traits(identifier, fun))))
+        alias getOverloads = __traits(getOverloads, __traits(parent, fun), __traits(identifier, fun));
```
pbackus, line 1232: Need to pass `true` as an additional argument to `__traits(getOverloads)` here in order to include template overloads. As-is, the check for `anySatisfy!(isTemplate, overloads)` below will always evaluate to `false`.

RazvanN7:
ping @Biotronic

n8sh:
Rebased & expanded commit message.

thewilsonator (approved):

0xEAB:
This patch is currently breaking DMD’s `build.d`.

PR #7510, opened 2020-05-31
std.path: Fix tilde expansion for HOME=/

kinke:
That's the case for CirrusCI/FreeBSD - the user is root, the `USER` env var unset and `HOME` set to `/`, and the tilde expansion for `~` in that case led to an empty string and according failures in https://github.com/dlang/phobos/blob/95e4fb49de41359b6c1f48003601243be691ff5e/std/path.d#L4158-L4163

### DIFF
--- a/std/path.d
+++ b/std/path.d
@@ -3972,17 +3972,27 @@ string expandTilde(string inputPath) @safe nothrow
             // Search end of C string
             size_t end = strlen(c_path);
 
-            // Remove trailing path separator, if any
-            if (end && isDirSeparator(c_path[end - 1]))
-                end--;
+            const cPathEndsWithDirSep = end && isDirSeparator(c_path[end - 1]);
 
             string cp;
             if (char_pos < path.length)
+            {
+                // Remove trailing path separator, if any (with special care for root /)
+                if (cPathEndsWithDirSep && (end > 1 || isDirSeparator(path[char_pos])))
+                    end--;
+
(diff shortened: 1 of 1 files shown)


Geod24:
Looks like a no-brainer. Don't you want it in stable ? Also I'm not sure copying bash's behavior for `~/` is the most sensible, as it will be the only place where `//` is produced, will it not ?

kinke:
stable would be somewhat easier for me/LDC, but as that bug has probably been there for ages, I thought it'd be safer to target master. No strong opinion though.

As to 'copying bash's behavior for ~/' - I took care of that first, but then realized that the failing test would then fail again, because it unconditionally expects an added slash compared to ~ alone. While the test can of course be adapted for this (or that whole test only exected if USER is set - although it revealed this bug), I've only later seen that bash doesn't try to get rid of duplicate dir separators at all, unlike Phobos here, so `HOME=/` is no special case for bash. Open for suggestions wrt. how to proceed.

kinke:
> D:\a\1\dmd\test>run.exe all   || exit /B 6 
>
> core.exception.AssertError@D:\a\1\dmd\dmd2\windows\bin\..\..\src\phobos\std\algorithm\mutation.d(2790): Swap: lhs internal pointer.

This also happens consistently when running the `run` testdriver for LDC's AArch64 CI. Definitely needs some investigation.

PR #7511, opened 2020-06-01
Support construction of std.bigint.BigInt from a sign and a byte-array magnitude

n8sh:
For consistency with other std.bigint functions allows arbitrary ranges.

### DIFF
--- a/std/bigint.d
+++ b/std/bigint.d
@@ -145,6 +145,39 @@ public:
         assertThrown!ConvException(BigInt(r4));
     }
 
+    /**
+     * Construct a `BigInt` from a sign and a magnitude.
+     *
+     * The magnitude is an $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
+     * of unsigned integers that satisfies either $(REF hasLength, std,range,primitives)
+     * or $(REF isForwardRange, std,range,primitives). The first (leftmost)
+     * element of the magnitude is considered the most significant.
+     *
+     * Params:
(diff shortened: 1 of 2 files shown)


n8sh:
Needed to re-push because of an oversight regarding `isNegative`.

std/bigint.d
```diff
@@ -145,6 +145,36 @@ public:
...
     }
 
+    /**
+     * Construct a `BigInt` from a sign and a magnitude. The magnitude is a
+     * $(REF_ALTTEXT bidirectional range, isBidirectionalRange, std,range,primitives)
+     * of unsigned integers. The first (leftmost) element of the magnitude is
+     * considered the most significant.
+     *
+     * Params:
+     *     isNegative = true for native, false for non-negative
```
MoonlightSentinel, line 155:
```suggestion
     *     isNegative = true for negative, false for non-negative
```

n8sh:
Fixed the spelling error in the documentation and changed the requirement from `isBidirectionalRange` to `hasLength || isForwardRange`.

std/bigint.d
```diff
@@ -145,6 +145,38 @@ public:
         assertThrown!ConvException(BigInt(r4));
     }
 
+    /**
+     * Construct a `BigInt` from a sign and a magnitude. The magnitude is an
```
dkorpel, line 149: Add to newlines after the first sentence or else the whole paragraph is interpreted as a big title by ddoc.

12345swordy:
More than 72 hours has passed.

PR #7513, opened 2020-06-04
[std.meta] Remove workarounds for an alias parameter not accepting basic types

ntrel:
Alias parameters accept basic types so various overloads are unnecessary now. E.g.:
```d
template Replace(T, U, TList...);
template Replace(alias T, U, TList...);
template Replace(T, alias U, TList...);
template Replace(alias T, alias U, TList...);
```
Only the last one is needed.

This also cleans up e.g. `GenericReplace` which the overloads forward to, and for those I've renamed `TList` to `Seq`. I've also replaced some usage of `OldAlias` with `Alias`.


ntrel:
> generated/linux/debug/64/publictests/std_traits.d(5111): Error: static assert:  `is((Object, A, B) == (const(Object), const(A), const(B)))` is false

This failure is due to:
>         alias ImplicitConversionTargets = staticMap!(ApplyLeft!(CopyConstness, T), TransitiveBaseTypeTuple!(T));

`ApplyLeft!(CopyConstness, T)` is failing, as `Alias!(const C)` now gives `C` (not `const C`), due to:
https://issues.dlang.org/show_bug.cgi?id=20863

std/meta.d
```diff
@@ -267,12 +267,23 @@ if (!isAggregateType!T || is(Unqual!T == T))
...
+template staticIndexOf(alias T, Seq...)
 {
-    enum staticIndexOf = genericIndexOf!(T, TList);
+    static foreach (idx, arg; Seq)
+    {
+        static if (is(typeof(staticIndexOf) == void) && // not yet defined
+                   isSame!(T, arg))
+        {
+            enum staticIndexOf = idx;
+        }
```
MoonlightSentinel, line 281: Simply `break` out of the loop?

```suggestion
    Loop:
    static foreach (idx, arg; Seq)
    {
        static if (isSame!(T, arg))
        {
            enum staticIndexOf = idx;
            break Loop;
        }
```

ntrel: @MoonlightSentinel:
From https://dlang.org/spec/version.html#break-continue:
> Note that static foreach itself cannot be broken nor continued even if it is explicitly labeled

(I haven't actually tried it recently though). Also I didn't write the code, the diff just makes it look like that. I'm just renaming some identifiers ;-)

MoonlightSentinel: You're right, I misremembered the interaction between `static foreach` and `break`.

MoonlightSentinel: Im curious how this implementation compares to something like this:

```
template staticIndexOf(alias T, Seq...)
{
    enum staticIndexOf = ()
    {
        foreach(idx, arg; Seq)
        {
            if (isSame!(T, arg))
                return idx;
        }
        
        return -1;
    }();
}
```

ntrel: Interesting solution. I don't know, but for large sequence length I think static foreach is v slow, at least ATM:
http://dpldocs.info/this-week-in-d/Blog.Posted_2020_05_11.html#dmd-and-static-foreach

ntrel: That instantiates `isSame` for every element of `Seq`, which could be expensive if there's a match near the start.

andralex (approved):
love the red

RazvanN7:
@ntrel What is the status of this PR?

ntrel:
@RazvanN7 this depends on https://github.com/dlang/dmd/pull/11320/

LightBender:
@jmdavis does this still apply to Phobos 3 traits?

PR #7515, opened 2020-06-06
Make std.math.isIdentical work in CTFE

n8sh:
>Works because we are in CTFE and there is no way in CTFE to set more bits of NaN payload than can fit in a double, and since 2.087 changed real.init to be non-signaling I *think* there is no way in CTFE for a real to be a signaling NaN unless real and double have the same representation so real's bits can be manipulated directly.

If the above is not true this should not be approved.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -6418,6 +6418,27 @@ if (isFloatingPoint!(X))
  */
 bool isIdentical(real x, real y) @trusted pure nothrow @nogc
 {
+    if (__ctfe)
+    {
+        if (x !is y) return false;
+        if (x == x) return true; // If not NaN `is` implies identical representation.
+        static if (double.mant_dig != real.mant_dig)
+        {
+            // Works because we are in CTFE and there is no way in CTFE to set more
+            // bits of NaN payload than can fit in a double, and since 2.087
+            // changed real.init to be non-signaling I *think* there is no way in
+            // CTFE for a real to be a signaling NaN unless real and double have
+            // the same representation so real's bits can be manipulated directly.
+            double d1 = x, d2 = y;
+        }
+        else
+        {
+            // Alias to avoid converting signaling to quiet.
+            alias d1 = x;
+            alias d2 = y;
+        }
+        return *cast(long*) &d1 == *cast(long*) &d2;
+    }
     // We're doing a bitwise comparison so the endianness is irrelevant.
     long*   pxs = cast(long *)&x;
     long*   pys = cast(long *)&y;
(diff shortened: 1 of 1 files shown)


kinke:
In general, it seems the only reason for that function is that DMD includes the padding bytes for `real is real`. LDC handles this, so this would work:
```D
bool isIdentical(real a, real b) {
    return a is b;
}

static assert(isIdentical(2, 0x1p1));
static assert(!isIdentical(0.0L, -0.0L));
```

The comment is misleading as well, as equal but different representations (think 0x2p0, 0x1p1) aren't (bitwise) identical, so it's not just about sign for zero and NaN payloads.

n8sh:
>In general, it seems the only reason for that function is that DMD includes the padding bytes for `real is real`.

I'd give another reason: `is` considers all NaNs equivalent regardless of representation.

kinke:
If you look at the asm, LDC and GDC definitely don't - they use memcmp.

n8sh:
>If you look at the asm, LDC and GDC definitely don't - they use memcmp.

Really? Because `assert(real.nan is -real.nan)` runs fine with LDC on run.dlang.io and on my computer.

kinke:
Yes, really, as easily shown in IR or ASM output. When looking at the AST output, you'll see that the frontend 'optimizes' your assert to an `assert(true)`.

kinke:
The culprit almost certainly being https://github.com/dlang/dmd/blob/2efa5242753e75f88c2f559c5be8adb4584d5e5c/src/dmd/expression.d#L566-L569.

std/math.d
```diff
@@ -6382,6 +6382,27 @@ if (isFloatingPoint!(X))
  */
 bool isIdentical(real x, real y) @trusted pure nothrow @nogc
 {
+    if (__ctfe)
+    {
+        if (x !is y) return false;
+        if (x == x) return true;
```
berni44, line 6388: Shouldn't that be `x == y`?

n8sh: Yes

n8sh: Ah, now I remember. It doesn't matter: the only circumstance where `x is y` but not `x == y` is if both x and y are NaN. So checking `x == x` is the same as checking `x == y` (it's a NaN check).

n8sh:
[By the current specification](https://dlang.org/spec/expression.html#identity_expressions)
>For struct objects and floating point values, identity is defined as the bits in the operands being identical

so this whole function could in principle be rewritten as `return x is y`. However as of DMD 2.095 that still isn't true for floating point numbers during CTFE. Possibly related open PR: https://github.com/dlang/dmd/pull/7568

LightBender:
Sending this to @thewilsonator to see if he can Salvage it.

0xEAB:
<https://github.com/dlang/dmd/pull/13780> (“modify float is float to do a bitwise compare”) has been merged.

0xEAB:
Anyway, this had been adapted and merged already: #10576

PR #7516, opened 2020-06-06
Fix documentation of std.math.isIdentical for NaN: compares all bits not just payload

n8sh:
Do documentation fixes need Bugzilla issue numbers? In this case `isIdentical`'s documented unit test already included `assert(!isIdentical(real.nan, -real.nan))` showing the actual behavior.


kinke:
It's still not correct... ;)

PR #7517, opened 2020-06-06
scalbn() doesn't inline

WalterBright:
Blocking https://github.com/dlang/dmd/pull/11236

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -4041,8 +4041,6 @@ real modf(real x, ref real i) @trusted nothrow @nogc
  */
 real scalbn(real x, int n) @safe pure nothrow @nogc
 {
-    pragma(inline, true);
-
     if (__ctfe)
     {
         // Handle special cases.


n8sh:
```d
real scalbn(real x, int n) @safe pure nothrow @nogc
{
    if (__ctfe)
    {
        // Handle special cases.
        if (x == 0.0 || isInfinity(x))
            return x;
    }
    return core.math.ldexp(x, n);
}
```
What is it about this code that prevents the DMD frontend from being able to inline it?

WalterBright:
DMD's inliner doesn't recognize all the patterns it could potentially inline. Adding that pattern could be done, but is beyond the scope of the PR it is blocking.

n8sh:
I see. Guessing this is [Bugzilla 7625](https://issues.dlang.org/show_bug.cgi?id=7625).

PR #7518, opened 2020-06-06
Faster staticMap and Filter

John-Colvin:
### staticMap
There are a few separate optimisations here:
1. make `staticMapImpl` , meaning that the template you're instantiating a lot of doesn't have to take the `F` every time. Improvement: only ~1-2% but very stable, I suspect this approach might be more rewarding elsewhere.
2. expand out many small cases (150). Any more than 3 was slower if done inline with `static if` chains, but combined with the following you can really crank it
3. generate the small cases as an `AliasSeq` of strings, so no static if chains, just an O(1) lookup (at least I hope so). This is the big one, enabling the expansion to work cheaply.
4. in the binary split branch immediately do what work you can. Combined with the rest this is an extra ~20% speedup and 10% memory reduction

Overall, this was ~2.5x faster and ~4x less RAM in my tests.
My tests were all biased towards very long `Args` and simple `F`s.

### Filter
Armed with this nice basic tool, Filter can be attacked, I saw ~4x speedup and ~3x memory reduction

I'm sure a combination of either implementing in terms of this faster `staticMap` or using same techniques could be done across a lot of `std.meta`

Todo:
* [ ] Proper benchmarks across different scales. I heavily tested with large synthetic `AliasSeq`s and then did some tweaks for a large codebase mostly using small ones.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -909,51 +909,21 @@ template anySatisfy(alias F, T...)
     static assert( anySatisfy!(isIntegral, int, double));
 }
 
-
+private alias Nothing = AliasSeq!(); // yes, this really does speed up compilation! By 10% !! Even with DMD master!
 /**
  * Filters an `AliasSeq` using a template predicate. Returns an
  * `AliasSeq` of the elements which satisfy the predicate.
  */
-template Filter(alias pred, TList...)
+template Filter(alias Pred, Args ...)
 {
-    static if (TList.length == 0)
+    template MaybeNothing(alias Q)
     {
-        alias Filter = AliasSeq!();
-    }
-    else static if (TList.length == 1)
-    {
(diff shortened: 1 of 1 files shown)


John-Colvin:
Tests on a large internal codebase suggest in practice low single-digit % reductions in overall compile times and compiler memory usage.

n8sh:
FYI I've been holding off on reviewing this under the impression that you still intend to perform more benchmarking.

John-Colvin:
> FYI I've been holding off on reviewing this under the impression that you still intend to perform more benchmarking.

I'm just not going to have time for this right now, sorry.

12345swordy:
More than 72 hours has passed. What is the hold up?

PR #7522, opened 2020-06-09
Reuse new builtin trait isCopyable in std.traits.isCopyable

nordlow:
By looking at the build errors in the calls to `std.conv.to` I reckon I didn't manage to exactly mimic the current definition of `std.traits.isCopyable` in dmd's definition of `__traits(isCopyable)`. Any clues on how to detect these discrepancies?

Add `static assert` inside the current definition `std.traits.isCopyable` that checks that `__traits(isCopyable)` equals `std.traits.isCopyable` is a way forward. I'll try that.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -512,7 +512,11 @@ template to(T)
         void fun() inout
         {
             static foreach (const idx; 0 .. this.tupleof.length)
-                this.tupleof[idx].to!string();
+            {
+                {
+                    const _ = this.tupleof[idx].to!string();
+                }
+            }
         }
     }
 }
--- a/std/traits.d
+++ b/std/traits.d
@@ -8904,9 +8904,7 @@ if (X.length == 1)
  + Returns:
  +  `true` if `S` can be copied. `false` otherwise.
  + ++/
-enum isCopyable(S) = is(typeof(
-    { S foo = S.init; S copy = foo; }
-));
+enum isCopyable(S) = __traits(isCopyable, S);
 
 ///
 @safe unittest
--- a/std/traits.d
+++ b/std/traits.d
@@ -8904,15 +8904,7 @@ if (X.length == 1)
  + Returns:
  +  `true` if `S` can be copied. `false` otherwise.
  + ++/
-template isCopyable(S)
-{
-    enum isCopyable_old = is(typeof({ S foo = S.init; S copy = foo; }));
-    enum isCopyable = __traits(isCopyable, S);
-    static if (isCopyable_old !is isCopyable)
-    {
-        pragma(msg, "New builtin trait isCopyable differs from old for type " ~ S.stringof);
-    }
-}
+enum isCopyable(S) = __traits(isCopyable, S);
 
 ///
 @safe unittest
--- a/std/traits.d
+++ b/std/traits.d
@@ -8908,7 +8908,10 @@ template isCopyable(S)
 {
     enum isCopyable_old = is(typeof({ S foo = S.init; S copy = foo; }));
     enum isCopyable = __traits(isCopyable, S);
-    static assert(isCopyable_old is isCopyable, "New builtin trait isCopyable differs from old for type " ~ S.stringof);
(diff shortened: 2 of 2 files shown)


PetarKirov:
This change fails to built because the warning are treated as errors and the following warning is now emitted:

> std/conv.d(515): Warning: calling std.conv.to!string.to!(inout(int)).to without side effects discards return value of type string, prepend a cast(void) if intentional

nordlow:
Added a static assert to check when new definition doesn't give same answer as old.

nordlow:
So it seems the new implementation doesn't match the old for `inout(char)[]`.

The following example highlights the difference. 

import std.traits : isCopyable;

inout(char)[] f(inout(char)[] x)
{
    auto y = x;
    static assert(__traits(isCopyable, typeof(x))); // passes
    static assert(isCopyable!(typeof(x))); // fails
    return x;
}

AFAICT, the current implementation is incorrect in this case, right?

tgehr:
@nordlow @andralex Yes, the current implementation is incorrect (for multiple reasons). It should have been:
```d
enum isCopyable(S) = __traits(compiles,(S orig){ S copy = orig; });
```

andralex:
> @nordlow @andralex Yes, the current implementation is incorrect (for multiple reasons). It should have been:
> 
> ```d
> enum isCopyable(S) = __traits(compiles,(S orig){ S copy = orig; });
> ```

Thank you! Why do we make simple things complicated? `S.init` and all that nonsense. (I think I wrote that code initially...)

nordlow:
ci/circleci: build fails

```
New builtin trait isCopyable differs from old for type inout(bool)

New builtin trait isCopyable differs from old for type inout(int)

std/conv.d(515): Warning: calling std.conv.to!string.to!(inout(int)).to without side effects discards return value of type string, prepend a cast(void) if intentional

New builtin trait isCopyable differs from old for type inout(float)

New builtin trait isCopyable differs from old for type inout(int)[]

New builtin trait isCopyable differs from old for type inout(int[int])

New builtin trait isCopyable differs from old for type inout(S)*

New builtin trait isCopyable differs from old for type inout(uint[4])

New builtin trait isCopyable differs from old for type inout(uint)[]

1 modules passed unittests

    ../dmd/generated/linux/debug/64/dmd -od$T  -conf= -I../druntime/import  -w -de -preview=dip1000 -m64 -fPIC -transition=complex -g -debug -main -unittest -version=StdUnittest generated/linux/debug/64/libphobos2.a -defaultlib= -debuglib= -L-lpthread -L-lm -L-ldl -cov -run std/complex.d ;     \

  )

posix.mak:407: recipe for target 'std/conv.test' failed

make: *** [std/conv.test] Error 1

1 modules passed unittests


Exited with code exit status 2

CircleCI received exit code 2
```

. Is this related to this commit?

std/conv.d
```diff
@@ -512,7 +512,11 @@ template to(T)
         void fun() inout
         {
             static foreach (const idx; 0 .. this.tupleof.length)
-                this.tupleof[idx].to!string();
+            {
+                {
+                    const _ = this.tupleof[idx].to!string();
```
nordlow, line 517: Is this ok or should I use

```D
cast(void)this.tupleof[idx].to!string();
```

instead?

nordlow:
Other than that I'm satisfied now.

n8sh:
It would be ideal if it didn't take 5 commits for 9 lines of changes but that's not a deal-breaker. Aside from that everything looks dandy.

nordlow:
> It would be ideal if it didn't take 5 commits for 9 lines of changes but that's not a deal-breaker. Aside from that everything looks dandy.

What's preferred solution to this nowadays. A rebase and squash into a single commit followed by a forced push?

Geod24:
> Can Github handle such situations now?

Yeah all those issues have been resolved. Github even shows notifications for force-push, so you can clearly see the diff applied as a result of the review.

PR #7524, opened 2020-06-10
Fix issue - sort should move elements instead of copy...

MoonlightSentinel:
... because it's reordering elements. Calling `opAssign` on unitialized members might also  cause errors if the implementation requires a valid instance (see `TimSortImpl`)

### DIFF
--- a/std/algorithm/sorting.d
+++ b/std/algorithm/sorting.d
@@ -1698,17 +1698,30 @@ private void shortSort(alias less, Range)(Range r)
                 auto t = r[0]; if (pred(t, r[0])) r[0] = r[0];
             }))) // Can we afford to temporarily invalidate the array?
         {
+            import core.lifetime : move;
+
             size_t j = i + 1;
-            auto temp = r[i];
+            static if (hasLvalueElements!Range)
+                auto temp = move(r[i]);
+            else
+                auto temp = r[i];
+
             if (pred(r[j], temp))
             {
                 do
                 {
-                    r[j - 1] = r[j];
+                    static if (hasLvalueElements!Range)
+                        trustedMoveEmplace(r[j], r[j - 1]);
+                    else
+                        r[j - 1] = r[j];
                     ++j;
                 }
                 while (j < r.length && pred(r[j], temp));
-                r[j - 1] = temp;
+
+                static if (hasLvalueElements!Range)
+                    trustedMoveEmplace(temp, r[j - 1]);
+                else
+                    r[j - 1] = move(temp);
             }
         }
         else
@@ -1723,6 +1736,13 @@ private void shortSort(alias less, Range)(Range r)
         }
         if (i == 0) break;
     }
+}
+
+/// @trusted wrapper for moveEmplace
+private void trustedMoveEmplace(T)(ref T source, ref T target) @trusted
+{
+    import core.lifetime : moveEmplace;
+    moveEmplace(source, target);
 }
 
 @safe unittest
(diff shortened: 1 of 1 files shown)


std/algorithm/sorting.d
```diff
@@ -1698,17 +1698,26 @@ private void shortSort(alias less, Range)(Range r)
                 auto t = r[0]; if (pred(t, r[0])) r[0] = r[0];
             }))) // Can we afford to temporarily invalidate the array?
         {
+            import std.algorithm.mutation : move;
+
             size_t j = i + 1;
             auto temp = r[i];
```
n8sh, line 1704: Is `move` also needed here?

MoonlightSentinel: Think so. Fixed

radcapricorn: It does look like it needs to be a `move`. And, since this yanks `r[i]` and moves it at the back, the following loop + fixup can all be `moveEmplace`s, to avoid unnecessary destructor calls. I'd almost vote for `moveEmplaceAndDontInitSource`, but we don't have that :)

std/algorithm/sorting.d
```diff
@@ -1698,17 +1698,30 @@ private void shortSort(alias less, Range)(Range r)
                 auto t = r[0]; if (pred(t, r[0])) r[0] = r[0];
             }))) // Can we afford to temporarily invalidate the array?
         {
+            import core.lifetime : move;
+
             size_t j = i + 1;
-            auto temp = r[i];
+            static if (hasLvalueElements!Range)
+                auto temp = move(r[i]);
```
radcapricorn, line 1705: So, for a range with lvalue elements, all the `move`s inside the following `if` can all now be `moveEmplace`s, to avoid unnecessary dtor calls ;)

MoonlightSentinel: Fixed

12345swordy:
It's all green.

std/algorithm/sorting.d
```diff
@@ -1725,6 +1738,13 @@ private void shortSort(alias less, Range)(Range r)
     }
 }
 
+/// @trusted wrapper for emplaceEmplace
```
radcapricorn, line 1741: for `moveEmplace`

std/algorithm/sorting.d
```diff
@@ -2806,6 +2835,22 @@ private template TimSortImpl(alias pred, R)
     alias gallopForwardUpper = gallopSearch!(false,  true);
     alias gallopReverseLower = gallopSearch!( true, false);
     alias gallopReverseUpper = gallopSearch!( true,  true);
+
+    /// Helper method that moves from[fIdx] into to[tIdx] if both are lvalues and
+    /// uses a plain assignment if not (necessary for backwards compatibility)
+    void moveEntry(X, Y)(ref X from, const size_t fIdx, ref Y to, const size_t tIdx)
+    do
```
radcapricorn, line 2842: No contracts?

std/algorithm/sorting.d
```diff
@@ -2918,6 +2963,40 @@ private template TimSortImpl(alias pred, R)
     sort!("a < b", SwapStrategy.stable)(arr);
 }
 
+@safe unittest
+{
+    static struct Postblit
```
radcapricorn, line 2968: Since this uses a copy ctor and not postblit, maybe it shouldn't be called `Postblit`? ;)

MoonlightSentinel: Fixed

std/algorithm/sorting.d
MoonlightSentinel: Fixed

std/algorithm/sorting.d
MoonlightSentinel: Forgot to remove `do` as well. Slices and decent range implementations will check the index anyway.

std/algorithm/sorting.d
```diff
@@ -2274,9 +2294,9 @@ private template TimSortImpl(alias pred, R)
     alias T = ElementType!R;
 
     alias less = binaryFun!pred;
-    alias greater = (a, b) => less(b, a);
-    alias greaterEqual = (a, b) => !less(a, b);
-    alias lessEqual = (a, b) => !less(b, a);
+    bool greater()(auto ref T a, auto ref T b) { return less(b, a); }
+    bool greaterEqual()(auto ref T a, auto ref T b) { return !less(a, b); }
+    bool lessEqual()(auto ref T a, auto ref T b) { return !less(b, a); }
```
andralex, line 2299: Not introduced here, but it seems these are superfluous (everything can be elegantly expressed with `less`). `greater` is not even used.

radcapricorn: I would agree for `greaterEqual(a, b)` vs `!less(a,b)`. However, not for `lessEqual(a, b)` vs `!less(b, a)`. More state for brain to track.

PR #7525, opened 2020-06-11
Remove an unnecessary allocation from BigInt powmod

n8sh:


thewilsonator:
Please force push to clear the doc test fail.

edi33416:
Is this still necessary if #7544 get's in?

n8sh:
@edi33416 it would still be faster (although the remark about allocation would no longer be true) unless the compiler performs aggressive cross-module function specialization.

std/bigint.d
```diff
@@ -2312,12 +2312,13 @@ BigInt powmod(BigInt base, BigInt exponent, BigInt modulus) pure nothrow @safe
...
     {
-        if (exponent & 1)
+        if (exponent.data.peekUint(0) & 1)
         {
             result = (result * base) % modulus;
         }
 
-        base = ((base % modulus) * (base % modulus)) % modulus;
+        auto tmp = base % modulus;
+        base = (tmp * tmp) % modulus;
```
n8sh, line 2321: Added a manual temporary value because `ldc2 -O3 --flto=full` (LDC version 1.22.0) doesn't perform common subexpression elimination here. Neither does `gdc -O4` (GDC version 10.2).

PR #7528, opened 2020-06-12
Issue 20257 - Add wait(Pid, Duration) for windows

MoonlightSentinel:
`wait(Pid, Duration)` allows to wait until the process terminates
or the timeout expires.

It's currently windows-only because a decent posix implementation
is quite difficult and this allows us to sort out the API beforehand.

CC @andralex as this introduces a new symbol

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -2254,7 +2254,7 @@ private:
     }
     else version (Windows)
     {
-        int performWait(bool block) @trusted
+        int performWait(const bool block, const DWORD timeout = INFINITE) @trusted
         {
             import std.exception : enforce;
             enforce!ProcessException(owned, "Can't wait on a detached process");
@@ -2262,9 +2262,15 @@ private:
             assert(_handle != INVALID_HANDLE_VALUE);
             if (block)
             {
-                auto result = WaitForSingleObject(_handle, INFINITE);
+                auto result = WaitForSingleObject(_handle, timeout);
                 if (result != WAIT_OBJECT_0)
+                {
+                    // Wait time exceeded `timeout` milliseconds?
+                    if (result == WAIT_TIMEOUT && timeout != INFINITE)
+                        return 0;
+
                     throw ProcessException.newFromLastError("Wait failed.");
+                }
             }
             if (!GetExitCodeProcess(_handle, cast(LPDWORD)&_exitCode))
                 throw ProcessException.newFromLastError();
@@ -2273,6 +2279,20 @@ private:
             _handle = INVALID_HANDLE_VALUE;
             _processID = terminated;
             return _exitCode;
+        }
+
+        int performWait(Duration timeout) @safe
+        {
+            import std.exception : enforce;
+            const msecs = timeout.total!"msecs";
+
+            // Limit this implementation the maximum wait time offered by
+            // WaitForSingleObject. One could theoretically break up larger
+            // durations into multiple waits but (DWORD.max - 1).msecs
+            // (> 7 weeks, 17 hours) should be enough for the usual case.
+            // DWORD.max is reserved for INFINITE
+            enforce!ProcessException(msecs < DWORD.max, "Timeout exceeds maximum wait time!");
+            return performWait(true, cast(DWORD) msecs);
         }
 
         ~this()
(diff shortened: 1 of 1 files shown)


std/process.d
```diff
@@ -2219,6 +2239,52 @@ int wait(Pid pid) @safe
...
+    return Tuple!(bool, "terminated", int, "status")(pid._processID == Pid.terminated, code);
+}
+
+version (Windows)
+@system unittest // Pid and wait()
+{
+    import std.typecons : tuple;
+    TestScript prog = "timeout /T %~1";
+    auto pid = spawnProcess([prog.path, "0"]);
+    assert(waitTimeout(pid, 1.msecs) == tuple(true, 0));
```
CyberShadow, line 2282: `timeout` docs say the argument takes seconds:
https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/timeout_1

Why are the test durations in milliseconds?

std/process.d
```diff
@@ -2219,6 +2239,52 @@ int wait(Pid pid) @safe
...
+
+Example:
+See the $(LREF spawnProcess) documentation.
+
+See_also:
+$(LREF wait), for a blocking function without timeout.
+$(LREF tryWait), for a non-blocking function without timeout.
+*/
+version (Windows)
+auto waitTimeout(Pid pid, Duration timeout) @safe
```
CyberShadow, line 2268: Docs should be in `version (StdDdoc)` so that they're generated on all platforms.

std/process.d
```diff
@@ -2219,6 +2239,52 @@ int wait(Pid pid) @safe
...
+Waits until either the process associated with `pid` terminates or the
+elapsed time exceeds the given timeout.
+
+If the process terminates within the given duration it behaves exactly like
+`wait`, except that it returns a tuple `(true, exit code)`.
+
+If the process does not terminate within the given duration it will stop
+waiting and return `(false, 0).`
+
+The timeout may not exceed `(uint.max - 1).msecs` (~ 7 weeks, 17 hours).
```
CyberShadow, line 2252: `$(BLUE This function is Windows-Only.)`

std/process.d
MoonlightSentinel: Fixed

std/process.d
MoonlightSentinel: Fixed

std/process.d
MoonlightSentinel: Fixed

MoonlightSentinel:
> Good addition though unsure about more platform-specific functionality to Phobos.

Agreed, implementations for other platforms should be added in the future.

> Would be good to ensure that it's possible to implement such things on top of Phobos functionality (i.e. using `Pid.osHandle`).

You mean a generic implementation for other platforms that doesn't depend on os-specific functions?

CyberShadow:
> You mean a generic implementation for other platforms that doesn't depend on os-specific functions?

Not necessarily. I recall that there was some friction with doing some things which are partially doable with `std.process`. E.g., `kill(Pid)` would work perfectly fine for processes which are actually not children of the D program, however, there is no way to construct a `Pid` for an arbitrary process. I think we should try to avoid such situations when possible, and strive for interoperability between the D wrappers and the low-level C/OS API. In this case, that would be ensuring that this feature is implementable on top of `std.process`, which I think is the case thanks to `Pid` exposing `osHandle`.

std/process.d
```diff
@@ -2219,6 +2239,58 @@ int wait(Pid pid) @safe
...
+@system unittest // Pid and wait()
+{
+    import std.typecons : tuple;
+    TestScript prog = "timeout /T %~1";
+    auto pid = spawnProcess([prog.path, "0"]);
+    assert(waitTimeout(pid, 2.seconds) == tuple(true, 0));
+
+    pid = spawnProcess([prog.path, "2"]);
+    assert(waitTimeout(pid, 1.seconds) == tuple(false, 0));
+    assert(waitTimeout(pid, 2.seconds) == tuple(true, 0)); // 2ms to avoid a flaky test
```
CyberShadow, line 2292: The comment still says `2ms`.

schveiguy:
There's not a portable option for Linux AFAIK. You need to use signals with another thread which can send the signal. I don't know if this is something we would want to implement in Phobos. I doubt there are going to be any advancements in Linux/Unix that allow for a wait with a timeout.

So if we add this, it will likely be permanently Windows-only. And I think that's fine, we should expose the OS functionality available for the OS you are using. For example, we have config options which are OS specific (though those are just ignored on other OSes).

schveiguy:
Hm... I take it back (a bit), SIGALRM can be used without threads to implement this. So perhaps it is possible.

CyberShadow:
> Hm... I take it back (a bit), SIGALRM can be used without threads to implement this. So perhaps it is possible.

Internet says `SIGALRM` is a bad idea due to race conditions.

BTW, this task is achievable in a cross-platform manner without any changes in Phobos by creating a pipe and passing one end to the process, and then using `select` or equivalent with a timeout on the other end to wait until the pipe is closed. (The pipe handle will be propagated to further child processes, which may or may not be what is needed.)

schveiguy:
It highly depends on the child process. The process could close the pipe without exiting because it doesn't plan to do any output.

So I don't think it's portable for all situations.

schveiguy:
> Internet says `SIGALRM` is a bad idea due to race conditions.

Yeah, the more I look at it, the less nice it looks.

CyberShadow:
> The process could close the pipe without exiting because it doesn't plan to do any output.

It doesn't have to be `STDOUT_FILENO`, can be an arbitrary handle. But, your point does still stand, the process technically could iterate over open fds and close them - I imagine some programs do that as part of daemonification.

MoonlightSentinel:
Could also consider `sigtimedwait`, e.g. as proposed [here](https://www.linuxprogrammingblog.com/code-examples/signal-waiting-sigtimedwait)

NilsLankila:
How about a control loop with `Thread.sleep()` ? This is actually what I occasionaly use when I have to do a non blocking wait on several Pid.

MoonlightSentinel:
Busy waiting is inefficient because the parent keeps consuming resources and also imprecise because you wait way too long depending on your sleep duration.

I could kinda see it as an "_if all else fails_" solution but i don't think it's appropriate for phobos.

schveiguy:
sigtimedwait probably isn't usable, that example doesn't seem to account for the possibility of waiting for a specific child, or in a thread.

CyberShadow:
I think the idea is that after every SIGCHLD, you check (as in `tryWait`) to see if it is the waited-on process which exited, or something else.

schveiguy:
I would expect the SIGCHLD to immediately re-fire if you don't wait for the child that caused the signal. Is that not true?

CyberShadow:
Sorry, I don't know. You may be right.

MoonlightSentinel:
Anybody got an idea how to fix LDC-Win32 on Azure?
```
.\unittest.exe
ERROR: Input redirection is not supported, exiting the process immediately.
```

RazvanN7:
@MoonlightSentinel Maybe it was a spurious failure? Maybe you can rebase this and see what the autotester has to say?

MoonlightSentinel:
Don't think so, the failure persistent across several runs. I'll try a different test case (`wait` is problematic because Microsoft apparently changed the CLI interface between some versions ...)

RazvanN7 (approved):
Looks good to me.

PR #7529, opened 2020-06-14
Eliminate the need for shared static this() on Posix

andralex:
It would be great if we could eliminate all static ctors/dtors from Phobos and use lazy initialization instead. This PR is a simple step in that direction - there's really no need to initialize those pointers to function (or test them) in Posix.

### DIFF
--- a/std/socket.d
+++ b/std/socket.d
@@ -290,30 +290,34 @@ bool wouldHaveBlocked() nothrow @nogc
 }
 
 
-private immutable
+version (Windows)
 {
+    private: immutable:
     typeof(&getnameinfo) getnameinfoPointer;
     typeof(&getaddrinfo) getaddrinfoPointer;
     typeof(&freeaddrinfo) freeaddrinfoPointer;
 }
-
-shared static this() @system
+else version (Posix)
 {
-    version (Windows)
+    enum getnameinfoPointer = &getnameinfo;
+    enum getaddrinfoPointer = &getaddrinfo;
+    enum freeaddrinfoPointer = &freeaddrinfo;
+}
+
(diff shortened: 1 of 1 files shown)


std/socket.d
```diff
@@ -290,30 +290,34 @@ bool wouldHaveBlocked() nothrow @nogc
...
+    private: immutable:
     typeof(&getnameinfo) getnameinfoPointer;
     typeof(&getaddrinfo) getaddrinfoPointer;
     typeof(&freeaddrinfo) freeaddrinfoPointer;
 }
+else version (Posix)
+{
+    enum getnameinfoPointer = &getnameinfo;
+    enum getaddrinfoPointer = &getaddrinfo;
+    enum freeaddrinfoPointer = &freeaddrinfo;
```
looked-at-me, line 304: Remove these for `Posix`, add `version(Windows) { ... } else version(Posix) { ... } else static assert(0);` where applicable to avoid un-necessary indirections.

andralex: Thanks, I'll add the assert. To get rid of the enums I'd need to wait for the updated Windows version.

andralex: Update: I see there's a `static assert` on line 126, so no need to duplicate that.

andralex: Just checked, only dmd generates actual indirections :). https://godbolt.org/z/RmDYYV

andralex: Even better: I consolidated those to the top version blocks.

andralex:
I had flashbacks that I worked on that before... and indeed I did: https://github.com/dlang/phobos/pull/5813. Let's continue there.

PR #7530, opened 2020-06-14
Fix issue 20929 - std.experimental.allocator.expandArray's range-based overload fails to instantiate for char and wchar arrays.

Just-Harry:
https://issues.dlang.org/show_bug.cgi?id=20929

The instantiation failure was caused by trying to get the address of an lvalue
for `emplace`, which was caused by `front`'s auto-decoding of the slice used to
refer to the uninitialised elements of the array.

The fix was to forgo the range functions for the slice, and just index it.

The changes to `ForcedInputRange` may seem unrelated to the fix, but I wanted
to test the input-range code-path, and the only usage of `ForcedInputRange`
was to test this overload of `expandArray`, so it seemed more appropriate to
reuse it, rather than define another input-range.

### DIFF
--- a/std/experimental/allocator/package.d
+++ b/std/experimental/allocator/package.d
@@ -2082,12 +2082,12 @@ if (isInputRange!R && !isInfinite!R)
 
 version (StdUnittest)
 {
-    private struct ForcedInputRange
+    private struct ForcedInputRange(T)
     {
-        int[]* array;
+        T[]* array;
         pure nothrow @safe @nogc:
         bool empty() { return !array || (*array).empty; }
-        ref int front() { return (*array)[0]; }
+        ref T front() { return (*array)[0]; }
         void popFront() { *array = (*array)[1 .. $]; }
     }
 }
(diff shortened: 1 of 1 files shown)


std/experimental/allocator/package.d
```diff
@@ -2221,13 +2221,13 @@ if (isInputRange!R)
             toFill.uninitializedFillDefault;
         }
 
-        for (; !range.empty; range.popFront, toFill.popFront)
+        for (; !range.empty; range.popFront, toFill = toFill[1 .. $])
```
andralex, line 2224: `toFill.popFront` should work, no?

std/experimental/allocator/package.d
```diff
@@ -2082,12 +2082,12 @@ if (isInputRange!R && !isInfinite!R)
 
 version (StdUnittest)
 {
-    private struct ForcedInputRange
+    private struct ForcedInputRange (T)
```
andralex, line 2085:
```suggestion
    private struct ForcedInputRange(T)
```

Just-Harry: Whoops, fixed

std/experimental/allocator/package.d
Just-Harry: [`popFront` does some decoding of its own](https://github.com/dlang/phobos/blob/a394c1348b4b73765bd0bbf4d08b0a8548c5778b/std/range/primitives.d#L2257), which could cause some uninitialised chars and wchars to be skipped

PR #7531, opened 2020-06-15
Fix issue 20924 - std.numeric.gcd cannot be used with const BigInt

Biotronic:


std/numeric.d
```diff
@@ -2987,13 +2987,16 @@ if (isIntegral!T)
...
         is(typeof(T.init % T.init)) &&
         is(typeof(T.init == 0 || T.init > 0)))
 {
     import std.algorithm.mutation : swap;
 
+    Unqual!T a = _a;
+    Unqual!T b = _b;
+
     enum canUseBinaryGcd = is(typeof(() {
         T t, u;
```
kinke, line 3001: This check will still fail for const/immutable T.

std/numeric.d
```diff
@@ -2987,15 +2987,18 @@ if (isIntegral!T)
...
+auto gcd(T)(T _a, T _b)
 if (!isIntegral!T &&
         is(typeof(T.init % T.init)) &&
         is(typeof(T.init == 0 || T.init > 0)))
 {
     import std.algorithm.mutation : swap;
 
+    Unqual!T a = _a;
+    Unqual!T b = _b;
+
```
benjones, line 2999: Can you `static if` this so you don't make copies if you don't have to?  If T is not const, just `alias a = _a`?

kinke: Other option: use `const T` for the params and then always make a mutable copy. This would prevent all the 'redundant' const/immutable instantiations, another (common) problem of such public templates. Or just forward to the mutable version as integer gcd above: https://github.com/dlang/phobos/blob/a394c1348b4b73765bd0bbf4d08b0a8548c5778b/std/numeric.d#L2936-L2939

Biotronic: The main problem with the `const T` idea over the `alias` is it pessimizes all code: the copy will always happen. There's also a theoretical issue of a type `T` such that `const(T)` can't be converted to `T`.

std/numeric.d
```diff
@@ -2987,64 +2987,70 @@ if (isIntegral!T)
...
-        }
-
-        if ((a & 1) == 0) swap(a, b);
+        return gcd!(Unqual!T)(a, b);
+    }
+    else
+    {
+        import std.algorithm.mutation : swap;
+        enum canUseBinaryGcd = is(typeof(() {
+            Unqual!T t, u;
```
n8sh, line 3003: In this branch you've already established `is(T == Unqual!T)`.

std/numeric.d
kinke: So you've changed your mind?

std/numeric.d
```diff
@@ -2987,64 +2987,70 @@ if (isIntegral!T)
 // This overload is for non-builtin numerical types like BigInt or
 // user-defined types.
 /// ditto
-T gcd(T)(T a, T b)
+auto gcd(T)(T a, T b)
```
kinke, line 2990: I dislike `auto` as return type - it requires the user to look at the source to infer it; for the compiler, it means it always has to analyze the function to infer it too (but for templates it doesn't matter that much).

PR #7533, opened 2020-06-16
[std.meta] Improve AliasSeq Map example & other tweaks

ntrel:
* Improve `AliasSeq` Map example: inline simple functions, move `testRef` next to where it's used. Fix capitalization of `Func`.
* Don't write *type* when alias is meant.
* Avoid instantiating AliasSeq with a sequence.
* Update private `Pack` template to use sequence terminology.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -183,8 +183,8 @@ alias Alias(T) = T;
 {
     // Without Alias this would fail if Args[0] was e.g. a value and
     // some logic would be needed to detect when to use enum instead
-    alias Head(Args ...) = Alias!(Args[0]);
-    alias Tail(Args ...) = Args[1 .. $];
+    alias Head(Args...) = Alias!(Args[0]);
+    alias Tail(Args...) = Args[1 .. $];
 
     alias Blah = AliasSeq!(3, int, "hello");
     static assert(Head!Blah == 3);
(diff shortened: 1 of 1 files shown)


ntrel:
CircleCI:
```
/usr/bin/ld: final link failed: No space left on device
```

std/meta.d
```diff
@@ -269,8 +249,8 @@ if (!isAggregateType!T || is(Unqual!T == T))
 }
 
 /**
- * Returns the index of the first occurrence of type T in the
- * sequence of zero or more types TList.
+ * Returns the index of the first occurrence of T in the
```
atilaneves, line 252: Why remove the word "type" here? I also don't know what these documentation changes have to do with the changes relating to the unittest above.

ntrel: Because staticIndexOf can find symbols or values, not just types.
The unittest is part of the modules documentation too, but I can split it out into another pull if you prefer.

atilaneves: But `T` there is a type, it's not an alias template parameter.

ntrel: For that overload, but T is also an alias parameter for the other overload:
https://github.com/dlang/phobos/blob/8308cef30bdd843d67ef22f5feb64aadbe0ca35f/std/meta.d#L261

atilaneves: Ah, I see.

PR #7534, opened 2020-06-17
increase the speed of SHA256 and SHA512 in non-optimized builds and bring optimized DMD builds closer to the speed of LDC

n8sh:
The below tables show that the new code compiled by `dmd` with no optimizations is faster than the old code compiled with `dmd -O -inline`. Compiled with `ldc2 -O3` the old code and new code have the same speed.

Comparing the time to run the benchmark when compiled with `dmd` with no optimizations vs when compiled with `ldc2 -O3`, the new code at worst takes 2.04x as much time while the old code takes at worst 26.13x as much time if 32-bit and at worst 10.89x if 64-bit. Built with `dmd -O -inline` the new code takes at worst 1.56x as much time as with `ldc2 -O3` while the old code takes at worst 2.13x as much time if 32-bit and at worst 3.06x if 64-bit.

**64-bit Windows, times normalized in terms of `ldc2 -O3`**

| Hash        | Data len | dmd -m64 | dmd -m64 -O | dmd -m64 -O -inline | ldc2 -m64 -O3 |
|-------------|----------|----------|-------------|---------------------|---------------|
| Old SHA-256 | 2^^4     |  8.78x   | 5.44x       |  2.02x              | 1x            |
| Old SHA-256 | 2^^20    |  9.97x   | 6.05x       |  1.96x              | 1x            |
| Old SHA-512 | 2^^4     |  9.37x   | 7.03x       |  3.06x              | 1x            |
| Old SHA-512 | 2^^20    | 10.89x   | 6.86x       |  2.77x              | 1x            |
| New SHA-256 | 2^^4     |  1.51x   | 1.21x       |  1.14x              | 1x            |
| New SHA-256 | 2^^20    |  1.48x   | 1.15x       |  1.16x              | 1x            |
| New SHA-512 | 2^^4     |  2.02x   | 1.57x       |  1.52x              | 1x            |
| New SHA-512 | 2^^20    |  2.04x   | 1.55x       |  1.56x              | 1x            |

**32-bit Windows, times normalized in terms of `ldc2 -O3`**

| Hash        | Data len | dmd -m32 | dmd -m32 -O | dmd -m32 -O -inline | ldc2 -m32 -O3 |
|-------------|----------|----------|-------------|---------------------|---------------|
| Old SHA-256 | 2^^4     | 21.08x   | 5.63x       | 2.00x               | 1x            |
| Old SHA-256 | 2^^20    | 26.13x   | 6.74x       | 2.13x               | 1x            |
| Old SHA-512 | 2^^4     |  5.60x   | 2.79x       | 1.72x               | 1x            |
| Old SHA-512 | 2^^20    |  5.64x   | 2.79x       | 1.69x               | 1x            |
| New SHA-256 | 2^^4     |  1.50x   | 1.15x       | 1.09x               | 1x            |
| New SHA-256 | 2^^20    |  1.37x   | 1.06x       | 1.08x               | 1x            |
| New SHA-512 | 2^^4     |  1.37x   | 1.41x       | 1.37x               | 1x            |
| New SHA-512 | 2^^20    |  1.02x   | 1.37x       | 1.35x               | 1x            |

Oddly the new version of SHA-512 when compiled by DMD in 32-bit mode is slower with the `-O` flag than without it but it is still a significant improvement.

### DIFF
--- a/std/digest/sha.d
+++ b/std/digest/sha.d
@@ -115,45 +115,16 @@ version (LittleEndian) import core.bitop : bswap;
 
 public import std.digest;
 
-/*
- * Helper methods for encoding the buffer.
- * Can be removed if the optimizer can inline the methods from std.bitmanip.
- */
-private ubyte[8] nativeToBigEndian(ulong val) @trusted pure nothrow @nogc
+version (LittleEndian)
 {
-    version (LittleEndian)
-        immutable ulong res = (cast(ulong)  bswap(cast(uint) val)) << 32 | bswap(cast(uint) (val >> 32));
-    else
-        immutable ulong res = val;
-    return *cast(ubyte[8]*) &res;
+    private alias bigEndianToNative = bswap;
+    private alias nativeToBigEndian = bswap;
 }
-
-private ubyte[4] nativeToBigEndian(uint val) @trusted pure nothrow @nogc
+else
 {
-    version (LittleEndian)
-        immutable uint res = bswap(val);
-    else
-        immutable uint res = val;
-    return *cast(ubyte[4]*) &res;
-}
-
-private ulong bigEndianToNative(ubyte[8] val) @trusted pure nothrow @nogc
-{
-    version (LittleEndian)
-    {
-        import std.bitmanip : bigEndianToNative;
-        return bigEndianToNative!ulong(val);
-    }
-    else
-        return *cast(ulong*) &val;
-}
-
-private uint bigEndianToNative(ubyte[4] val) @trusted pure nothrow @nogc
-{
-    version (LittleEndian)
-        return bswap(*cast(uint*) &val);
-    else
-        return *cast(uint*) &val;
+    private uint bigEndianToNative(uint val) @safe pure nothrow @nogc { return val; }
+    private ulong bigEndianToNative(ulong val) @safe pure nothrow @nogc { return val; }
+    private alias nativeToBigEndian = bigEndianToNative;
 }
 
 //rotateLeft rotates x left n bits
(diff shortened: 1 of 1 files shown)


kinke:
Very nice numbers and overview. :+1:

I obviously dislike code legibility regressions to compensate for DMD's weak codegen though. Phobos should IMO be one of the D showcases and attract people when they haven't seen much D yet.

dnadlinger:
Agreed – just as for the recent PRs where template compile times were improved by manually copy-paste-unrolling a gazillion different cases, to me this looks like a step backwards. Wouldn't it be much preferable to improve ? (If I needed SHA256 performance, I'd just use LDC anyway, or if I _really_ needed it, an optimised version that makes use of SHA intrinsics, possibly in assembly.)

wilzbach:
> I obviously dislike code legibility regressions to compensate for DMD's weak codegen though.

Agreed. We all gave up on DMD a long time ago. It's really just there, because Walter doesn't want to give up on it. At the very least we shouldn't regress code to cater for legacy DMD.

dnadlinger:
Also, can `pragma(inline, true)` be used to the same effect without manually expanding the rotates, or does DMD fail to inline them?

WalterBright:
> can pragma(inline, true) be used to the same effect without manually expanding the rotates, or does DMD fail to inline them?

Until https://github.com/dlang/dmd/pull/11236 is merged, only the `-inline` switch will have an effect.

WalterBright:
@n8sh thank you. It looks like the main cause of the slowdown is the loops aren't being unrolled. I've already filed that as https://issues.dlang.org/show_bug.cgi?id=20991 and know how to fix it. There also may be an inlining issue, I'll check that.

std/digest/sha.d
```diff
@@ -115,45 +115,16 @@ version (LittleEndian) import core.bitop : bswap;
...
 
-/*
- * Helper methods for encoding the buffer.
- * Can be removed if the optimizer can inline the methods from std.bitmanip.
- */
-private ubyte[8] nativeToBigEndian(ulong val) @trusted pure nothrow @nogc
+version (LittleEndian)
 {
-    version (LittleEndian)
-        immutable ulong res = (cast(ulong)  bswap(cast(uint) val)) << 32 | bswap(cast(uint) (val >> 32));
```
WalterBright, line 125: This is reinventing `core.bitmap.bswap()`

n8sh: Understandable because [this code was written](https://github.com/dlang/phobos/commit/9dc61f673446fe299c1064f39ab8f244099802ae) before [64-bit bswap was added to core.bitop](https://github.com/dlang/druntime/commit/d07e682f21b0043354fc4b9ca6d5d2b176a63539).

WalterBright: https://github.com/dlang/phobos/pull/7534

WalterBright:
The current dmd, phobos, and druntime, along with https://github.com/dlang/dmd/pull/11423 successfully inlines everything in `transform()`, and there are no branches at all in the result. It successfully uses the `ror`, `rol`, and `bswap` instructions.

The inlining implementation was not changed, it works fine for this code. You just have to add `-inline` to get inlining.

But there is one new feature - you can select which functions to inline even without the `-inline` switch by annotating with `pragma(inline,true)`. This is much preferable to the mixin implementation here.

@n8sh I would appreciate it if you could update this PR by replacing the mixins with `pragma(inline,true)` and see where we stand with the benchmark results. I want to see what impact the other improvements in this PR have on the benchmark results.

WalterBright:
https://github.com/dlang/dmd/pull/11423 has been merged now.

n8sh:
@WalterBright Redid benchmarks using DMD master 4190a3e. The version using `mixin` and the version using `pragma(inline, true)` should be the same but they're not.

**DMD master 4190a3e, 64-bit Windows, times normalized in terms of `ldc2 -O3`**

| Hash           | Data len | dmd -m64 | dmd -m64 -O | dmd -m64 -O -inline | ldc2 -m64 -O3 |
|----------------|----------|----------|-------------|---------------------|---------------|
| Old SHA-256    | 2^^4     |  5.59x   |  5.69x      |  1.96x              |  1.0x         |
| Old SHA-256    | 2^^20    |  5.97x   |  6.12x      |  1.89x              |  1.0x         |
| Old SHA-512    | 2^^4     |  5.78x   |  6.01x      |  2.07x              |  1.0x         |
| Old SHA-512    | 2^^20    |  9.03x   |  9.27x      |  2.86x              |  1.0x         |
| Mixin SHA-256  | 2^^4     |  1.49x   |  1.21x      |  1.14x              |  1.0x         |
| Mixin SHA-256  | 2^^20    |  1.50x   |  1.16x      |  1.16x              |  1.0x         |
| Mixin SHA-512  | 2^^4     |  1.63x   |  1.23x      |  1.16x              |  1.0x         |
| Mixin SHA-512  | 2^^20    |  1.63x   |  1.18x      |  1.18x              |  1.0x         |
| Inline SHA-256 | 2^^4     |  3.58x   |  1.87x      |  1.80x              |  1.0x         |
| Inline SHA-256 | 2^^20    |  3.82x   |  1.90x      |  1.91x              |  1.0x         |
| Inline SHA-512 | 2^^4     |  2.67x   |  1.40x      |  1.35x              |  1.0x         |
| Inline SHA-512 | 2^^20    |  2.87x   |  1.82x      |  1.84x              |  1.0x         |

WalterBright:
Thank you.

> should be the same but they're not

The times with/without `-inline` show that it is inlining as the times are the same. The baffling thing, however, is why the mixin results are better.

WalterBright:
@n8sh is it possible that your `pragma(inline,true)` version does not include the other modifications, like replacing `ubyte[8]` with `ulong` ?

n8sh:
I went back and verified that's not what was going on.

WalterBright:
Are the functions not being inlined even with `-inline` ?

looked-at-me:
`inline` won't remove everything, it still needs to construct a scope and such to have the same behavior (even though it is inlined). A `mixin` won't have any additional overhead as it is literally in the same scope that it is mixed into. Why have a feature if you aren't going to utilise it? 

This PR LGTM.

quickfur:
I agree with @wilzbach, why aren't we fixing DMD's optimizer instead of introducing these kinds of hacks into Phobos code?

wilzbach:
> I agree with @wilzbach, why aren't we fixing DMD's optimizer instead of introducing these kinds of hacks into Phobos code?  

Well, because everyone except for Walter has given up on fixing DMD's optimizer.
There was even a recent internal discussion about fully dropping it.
Why would you invest _massive_ amounts of time into it when you have two world-class optimizing backends available?

ibuclaw:
The bswap64 change looks fine, however, so maybe break that out of this PR?

WalterBright:
> Why would you invest massive amounts of time into it

Actually, I've fixed most of the shortcomings this PR exposes. It was simple stuff, like bswap64 hadn't been implemented, and the rotate 64 wasn't either. A large chunk was simply that DMD requires a separate `-inline` switch rather than it being implicitly on with `-O`. Nothing massive.

If people find other issues, please tag them with the "performance" keyword in bugzilla. Thanks!

Here's the current list of tagged issues:
https://issues.dlang.org/buglist.cgi?bug_status=NEW&bug_status=ASSIGNED&bug_status=REOPENED&keywords=performance&keywords_type=allwords&list_id=232386&query_format=advanced

n8sh:
I see no further need to keep this open.

PR #7535, opened 2020-06-18
fix issue 20949 - std.range.popFront is unsafe in release mode

aG0aep6G:


aG0aep6G:
I don't know where to add a test for this. The issue only manifests when the `assert` in `popFront` is omitted. Apparently, `-unittest` overrides `-release` and keeps the `assert`. So a `unittest` doesn't cut it. Do we have a place in Phobos for testing a complete program?

PR #7536, opened 2020-06-19
Issue 17712 - Add tests & Revert std.conv.toChars workaround for dmd link error

ntrel:
Draft for now. ~~Once the test cases in https://github.com/dlang/dmd/pull/11300 are merged~~ This pull now contains the tests, which use Phobos. I had to edit `test/run.sh` to run them.

The dmd bug for the `b17712_c13.d` test case has been fixed in dlang/dmd#9636.
The original test case is still unfixed in dmd. The workaround in `std.conv.text`, added in #6659 causes unnecessary allocations when there are multiple arguments and >0 integer arguments.

Undefined reference to `std.conv.toChars!(10, char, 1, uint).toChars(uint)`:
https://issues.dlang.org/show_bug.cgi?id=17712

### DIFF
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -124,6 +124,21 @@ publictests()
     # run -betterC tests
     make -f posix.mak test_extractor # build in single-threaded mode
     make -f posix.mak -j$N betterc
+
+    # independent tests
+    test_other
+}
+
+# FIXME: tests should run in parallel
+test_other()
+{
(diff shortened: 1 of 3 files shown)


ntrel:
Closing this, as the original test case actually fails, even without reverting the workaround (as shown by the test failures). That workaround now seems to be not working and should probably be reverted.

PR #7537, opened 2020-06-21
[stable] std.math: Remove accidental public AliasSeq import

kinke:
This wasn't in 2.092.x, so better get rid of this ASAP.


kinke:
Oops: https://github.com/dlang/phobos/pull/7483

PR #7539, opened 2020-06-21
Allow std.getopt to accept receivers by reference

benjones:
This allows users to pass "receivers" by reference to getopt, instead of by pointer.  Basically, this allows the caller to omit the `&` that they've been putting in front of the variables they pass to std.getopt.  See changelog entry/unittest changes for examples.

### DIFF
--- a/changelog/getoptAcceptReferences.dd
+++ b/changelog/getoptAcceptReferences.dd
@@ -1,0 +1,13 @@
+Allow std.getopt to accept out pameters by reference
+
+Previously get opt takes the "receivers" for options as pointers,
+which will be assigned with `*receiver = parsedValue`.  This change allows
+the caller to pass those parameters by reference instead.  Basically, this
+allows code to delete the `&` before variable names when calling getopt
+-------
+bool opt;
+string[] args = ["program", "-a"];
+//previously required passing &opt
+getopt(args, config.passThrough, 'a', opt);
+assert(opt);
+------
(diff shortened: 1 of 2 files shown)


benjones:
I'm not confident about how option strings that could be considered pass by reference and string parameters are going to interact... It may require some more thought, but let's see what the CIs say

benjones:
So as far as I can tell, there's no good way to distinguish between an optional char or string description parameter that happens to be passed as a reference, and a receiver for a char or string that is passed by reference.  My current solution is to disallow passing a char/string receiver as a reference, and requiring a pointer for those cases.  It makes the API a little bit nonumiform, and it's less obvious that the convenience you get for bool/numeric parameters is worth special casing the API.  Any thoughts?

jondegenhardt:
Since nobody else is offering an opinion - Passing by reference is nicer, but I'm not sure that having two separate ways to do it is an improvement. Mostly from a learning, consistency, and documentation perspective. Sometimes it's better to have a single way to do something.

I have a slight concern with increasing the complexity of the templates. I've had several problems with LDC optimization of programs having with long option lists. This appears due to the recursive nature of the template expansions. It doesn't look like the templates were changed all that much, so likely won't be an issue.

So, I'm lukewarm on this. Passing by reference is nicer, but passing by pointer wasn't causing much trouble.

atilaneves:
I'm also unsure of this. I agree that references are nicer, but at this point it might be better to mark the pointers as `scope` so that `getopt` can be `@safe` instead.

benjones:
OK, thanks for the feedback, I'll close this.

PR #7540, opened 2020-06-23
add comment doc: FieldNameTuple and Fields are always in-sync.

mw66:
@radcapricorn


mingwugmail:
The point of doc is that the user don't have to dive into the code to learn its usage.

atilaneves:
Personally I think these comments are more confusing than their absence. Maybe it's just me, but why wouldn't they be in the same order and have the same length? It's strictly dependent on the order of field declaration.

mingwugmail:
Then given a tuple of field names how do the user know their corresponding types? e.g.

https://github.com/mingwugmail/jdiutil/blob/master/source/jdiutil.d#L105

how do I know types[i] is correctly corresponding to fields[i]?

You are an language expert, so you know. But to a normal user, esp new-comer to the D lang, how would s/he know by reading the doc (without dig into phobos source code)?

https://forum.dlang.org/thread/xipkvmafkradzyygcgdg@forum.dlang.org

n8sh:
>Personally I think these comments are more confusing than their absence. Maybe it's just me, but why wouldn't they be in the same order and have the same length? It's strictly dependent on the order of field declaration.

Then perhaps a better remark is "Fields appear in order of declaration" instead of a note about being in sync with some other thing. I agree that the remark about number of fields is superfluous.

std/experimental/checkedint.d
```diff
@@ -1083,6 +1083,25 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
...
+}
+
+// toString
+@safe unittest
+{
+    Checked!(int, void) x;
+    x = 42;
+    assert(x.toString == "42");
+    x = -42;
+    assert(x.toString == "-42");
```
atilaneves, line 1104: How is this related to the PR title?

mingwugmail:
sorry branch messed up.

std/experimental/checkedint.d
mingwugmail: sorry branch messed up.

RazvanN7:
@mingwugmail What is the status of this PR? It seems like it needs a rebase.

PR #7543, opened 2020-06-25
Prevent calling of modules shared dtors while run std.stdio unittest

denizzzka:
Otherwise it calls `__run_exit_handlers` what can contain call to `runModuleDestructors` (at least for LDC)

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -1482,14 +1482,14 @@ Removes the lock over the specified file segment.
         // the same process. fork() is used to create a second process.
         static void runForked(void delegate() code)
         {
-            import core.stdc.stdlib : exit;
+            import core.stdc.stdlib : _Exit;
             import core.sys.posix.sys.wait : waitpid;
             import core.sys.posix.unistd : fork;
             int child, status;
             if ((child = fork()) == 0)
             {
                 code();
-                exit(0);
+                _Exit(0);
             }
             else
             {


n8sh:
>Otherwise it calls `__run_exit_handlers` what can contain call to `runModuleDestructors` (at least for LDC)

Why is that a problem if run in a different process?

denizzzka:
@n8sh after thread is done (exit(0) is called) it goes to dtors and frees shared memory what used by still running modules.

I don't know why  DMD is not affected by this.

denizzzka:
@n8sh ahh, I understand your point - fork() creates process, not thread.

I have no idea why this problem affects only LDC.
Maybe new created process "inherits" already opened shared libraries and tries to destruct them?

denizzzka:
So, this is wrong solution

RazvanN7:
A PR needs to be accompanied by a bug report that highlights the issue that it is fixing. Also, every addition (apart from documentation PRs) needs to implement a test that highlights the issue that the PR is fixing. Currently, this PR does not do any of the above, therefore it is hard to assess the usefulness of this additon.

denizzzka:
https://github.com/dlang/phobos/pull/7543

denizzzka:
What it looked like during the crash:
https://github.com/ldc-developers/druntime/pull/188#issuecomment-649602627

denizzzka:
What it looked like during the crash:
https://github.com/ldc-developers/druntime/pull/188#issuecomment-649602627

Since this problem does not appear anywhere then I close this issue. In case of this problem appears again anyone will google it and it will be easier to understand how to fix it

PR #7544, opened 2020-06-26
std.bigint.BigInt: special case x & non-negative int to avoid unnecessary allocation

n8sh:
`x & 1` is a natural way to check the low bit of `std.bigint.BigInt x` but that allocates unlike the less natural `x.getDigit!uint(0) & 1`. Examples of this appearing in Phobos  are `std.bigint.powmod(BigInt, BigInt, BigInt)` ([PR #7525 fixes it](https://github.com/dlang/phobos/pull/7525)) and `std.numeric.gcd!T(T, T)` when instantiated for `BigInt`. Adding special handling for `x & 1` will aside from helping naive uses also improve performance of generic templated code that operates on arbitrary number-like types.

### DIFF
--- a/std/bigint.d
+++ b/std/bigint.d
@@ -29,6 +29,7 @@ import std.conv : ConvException;
 
 import std.format : FormatSpec, FormatException;
 import std.internal.math.biguintcore;
+import std.internal.math.biguintnoasm : BigDigit;
 import std.range.primitives;
 import std.traits;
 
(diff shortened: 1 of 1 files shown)


n8sh:
Generalized for non-negative built-in integers.

thewilsonator (approved):

thewilsonator:
@n8sh please rebase it seems CI is stuck

PR #7547, opened 2020-06-28
Make CowArray.freeThisReference @safe

nordlow:
This is needed for my build to pass when dmd flag `-checkaction=context` is used in my dub project.

Currently, the line

dflags "-checkaction=context"

in `dub.sdl` makes build error as

```
/home/per/.local/dlang/linux/bin64/src/phobos/std/uni/package.d(3376,17): Error: `this.data.ptr` cannot be used in `@safe` code, use `&this.data[0]` instead
/home/per/.local/dlang/linux/bin64/src/phobos/std/uni/package.d(2128,29): Error: template instance `std.uni.CowArray!(GcPolicy).CowArray.__ctor!(uint[])` error instantiating
/home/per/.local/dlang/linux/bin64/src/phobos/std/uni/package.d(2137,32):        instantiated from here: `__ctor!()`
/home/per/.local/dlang/linux/bin64/src/phobos/std/uni/package.d(1956,29):        instantiated from here: `InversionList!(GcPolicy)`
```

I can replace more of the `.ptr` statements if someone feels the need to.

### DIFF
--- a/std/uni/package.d
+++ b/std/uni/package.d
@@ -3373,7 +3373,7 @@ private:
         }
         else
             SP.destroy(data);
-        assert(!data.ptr);
+        assert(!&data[0]);
     }
 
     void dupThisReference(uint count)


MoonlightSentinel:
This seems to be a dmd/druntime issue if `-checkaction=context` makes a difference.

nordlow:
> This seems to be a dmd/druntime issue if `-checkaction=context` makes a difference.

So, how to move forward?

std/uni/package.d
```diff
@@ -3373,7 +3373,7 @@ private:
         }
         else
             SP.destroy(data);
-        assert(!data.ptr);
+        assert(!&data[0]);
```
MoonlightSentinel, line 3376: This cause a range violation.
```suggestion
        assert((() => !data.ptr)());
```

MoonlightSentinel:
> So, how to move forward?

Improving the `@safety` is good but we should determine why `-checkaction=context` makes a difference. I assume other assertion modes are missing safety checks because the error message seems valid. 

BTW: Fix for the CircleCI failure: #7548

std/uni/package.d
Geod24: Wait what ? Something is off. `!data.ptr` should be safe.

MoonlightSentinel: Isn't acessing `.ptr` always un-`@safe` by convention? (Allthough it cannot really do unsafe stuff)

MoonlightSentinel: E.g. `assert(arr.ptr)` is rejected.

MoonlightSentinel:
`-checkaction=context` fails because the expression is lowered as:
```
assert(!arr.ptr, _d_assert_fail!"!"(arr.ptr));
```

The lowering strips the `!` to format the actual value instead of `true`/`false`

std/uni/package.d
nordlow: I guess ideally the compiler should allow `@safe` `.ptr` access for boolean testing right like in this case?

nordlow:
> `-checkaction=context` fails because the expression is lowered as:
> 
> ```
> assert(!arr.ptr, _d_assert_fail!"!"(arr.ptr));
> ```
> 
> The lowering strips the `!` to format the actual value instead of `true`/`false`

So how to progress with this assert statement? 

Should we change this commit to 

```
assert(!cast(bool)data.ptr);
```

for now or should we fix the implementation or lowering of _d_assert_fail!?

std/uni/package.d
Geod24: Accessing the content *pointer to* by ptr is unsafe. Checking that the pointer itself is not null is completely `@safe`.
E.g. the following compiles fine for me:
```D
void main () @safe
{
    int[] arr;
    assert(!arr.ptr);
}
```

But if you try to assign it to a variable you'll get an error.

nordlow: > Accessing the content _pointer to_ by ptr is unsafe. Checking that the pointer itself is not null is completely `@safe`.
> E.g. the following compiles fine for me:

As describe below the problem is the call _d_assert_fail!"!"(arr.ptr).

nordlow:
> `-checkaction=context` fails because the expression is lowered as:
> 
> ```
> assert(!arr.ptr, _d_assert_fail!"!"(arr.ptr));
> ```
> 
> The lowering strips the `!` to format the actual value instead of `true`/`false`

Ideally the compiler would need to introspect the definition  of `_d_assert_fail` for this specific overload (template params `"!"`) to determine that the access to arr.ptr in this case is safe. Provided printing the expression `arr.ptr` is `@safe`.

std/uni/package.d
MoonlightSentinel: > Accessing the content _pointer to_ by ptr is unsafe. Checking that the pointer itself is not null is completely `@safe`.

Agreed but I wasn't sure because the following doesn't compile:
```d
void main () @safe
{
    int[] arr;
    assert(arr.ptr); // no !
}
```

MoonlightSentinel:
> So how to progress with this assert statement?

I guess the implementation for `checkaction=context` could detect `XXX.ptr` and then `@trust` the read.

> If we were to redefine `_d_assert_fail` as a template mixin we could help the compiler understand that this use case is `@safe`.

That wouldn't work because `_d_assert_fail` calls other functions to format it's argument(s).

nordlow:
> That wouldn't work because `_d_assert_fail` calls other functions to format it's argument(s).

But what wrapping that case in a @trusted lambda?

MoonlightSentinel:
> What about wrapping that case in a `@trusted` lambda?

Maybe but the compiler should only trust the read, not the entire `assert`. Otherwise it would create `@safe` holes, e.g.

```d
void main() @safe
{
     int[] arr = ....;
     assert(arr.ptr is bad());
}

int* bad() @system { .... }
```

MoonlightSentinel:
Fix: dlang/dmd#11337

nordlow:
What about sticking to

Assert(!!

for now.

PR #7549, opened 2020-07-01
fix issue 8904 - revert workaround added to prevent a linker error

NilsLankila:


NilsLankila:
this was just to verify if the iussue can be closed as is

PR #7554, opened 2020-07-10
Move stableRemove alias to correct place for documentation

schveiguy:
stableRemove was listed under the linearRemove documentation for some reason, even though it's aliased to remove.

### DIFF
--- a/std/container/dlist.d
+++ b/std/container/dlist.d
@@ -642,6 +642,9 @@ Complexity: $(BIGOH 1)
         return remove(r);
     }
 
+    /// ditto
+    alias stableRemove = remove;
+
 /**
 Removes first element of `r`, wich must be a range obtained originally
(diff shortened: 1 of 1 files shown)


n8sh:
Both definitions of `linearRemove` should also be documented together instead of one appearing under the definition of `remove`.

schveiguy:
> Both definitions of `linearRemove` should also be documented together instead of one appearing under the definition of `remove`.

I saw that, and wasn't sure what to do. The truth is that the first definition of `linearRemove` is a simple call to `remove`. So in essence, that form is not linear. I don't know why it's that way (perhaps `linearRemove` means "at worst linear" performance?)

I left it alone, because it's technically under the right docs.

Perhaps the correct solution is to put all definitions for `remove` and `linearRemove` together, and document them as one.

But this change is a simple thing that stuck out at me when I was reading the docs.

PR #7559, opened 2020-07-15
std.digest.sha: don't reinvent rotate functions

WalterBright:


Geod24:
Might as well eliminate the function and use `rol` and `ror` directly, no ?

WalterBright:
I'll leave that for a later refactoring PR.

PR #7560, opened 2020-07-21
change rotateLeft to rol, rotateRight to ror

WalterBright:
No longer any purpose to the wrappers.


std/digest/sha.d
```diff
@@ -113,6 +113,8 @@ else version (D_InlineAsm_X86_64)
 
 version (LittleEndian) import core.bitop : bswap;
 
+import core.bitop;
```
NilsLankila, line 116: suggestion

```d
import core.bitop : rotateRight = ror;
import core.bitop : rotateLeft = rol;
```

and then no need to use to touch the code. Just the alias and delete the wrappers.

NilsLankila: @thewilsonator, about auto-merge, check my review, this PR can change much less lines

WalterBright: I considered that, then decided that pointless wrappers should be removed.

PR #7561, opened 2020-07-21
fix fabs() to not use x87 for float/double

WalterBright:
This is a partial fix for https://issues.dlang.org/show_bug.cgi?id=19663

A complete fix would be to make it a compiler intrinsic.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -4115,19 +4115,38 @@ real cbrt(real x) @trusted nothrow @nogc
  *      )
  */
 real fabs(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.fabs(x); }
-//FIXME
+
 ///ditto
-double fabs(double x) @safe pure nothrow @nogc { return fabs(cast(real) x); }
-//FIXME
+pragma(inline, true)
+double fabs(double d) @trusted pure nothrow @nogc
+{
+    ulong tmp = *cast(ulong*)&d & 0x7FFF_FFFF_FFFF_FFFF;
+    return *cast(double*)&tmp;
+}
+
 ///ditto
-float fabs(float x) @safe pure nothrow @nogc { return fabs(cast(real) x); }
+pragma(inline, true)
+float fabs(float f) @trusted pure nothrow @nogc
+{
+    uint tmp = *cast(uint*)&f & 0x7FFF_FFFF;
+    return *cast(float*)&tmp;
+}
 
 ///
 @safe unittest
 {
+
+    assert(isIdentical(fabs(0.0f), 0.0f));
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -4115,19 +4115,38 @@ real cbrt(real x) @trusted nothrow @nogc
...
-//FIXME
+
 ///ditto
-double fabs(double x) @safe pure nothrow @nogc { return fabs(cast(real) x); }
-//FIXME
+pragma(inline, true)
+double fabs(double d) @trusted pure nothrow @nogc
+{
+    ulong tmp = *cast(ulong*)&d & 0x7FFF_FFFF_FFFF_FFFF;
+    return *cast(double*)&tmp;
```
thewilsonator, line 4124: is this Endian safe?

std/math.d
```diff
@@ -4115,19 +4115,38 @@ real cbrt(real x) @trusted nothrow @nogc
...
+}
+
 ///ditto
-float fabs(float x) @safe pure nothrow @nogc { return fabs(cast(real) x); }
+pragma(inline, true)
+float fabs(float f) @trusted pure nothrow @nogc
+{
+    uint tmp = *cast(uint*)&f & 0x7FFF_FFFF;
+    return *cast(float*)&tmp;
```
thewilsonator, line 4132: ditto

std/math.d
WalterBright: It should be fine. I added unittests to verify.

WalterWaldron:
Since it's still round tripping through the stack ```core.bitop.btr``` is an option, but it's not a big improvement.
```
    import core.bitop:btr;
    version (BigEndian) btr(cast(ulong*)&d,7);
    else                btr(cast(ulong*)&d,63);
    return d;
```
Link to compare functions & compilers: https://d.godbolt.org/z/bvd1Ev

std/math.d
ibuclaw: The CI wouldn't really catch that though, as it's all LittleEndian.

However... FYI @thewilsonator, all targets I'm aware use _same_ endianness for both float and integer words, regardless whether the it's a BigEndian or LittleEndian architecture.  The one target that is the exception to the rule is the PDP-11, and I doubt Phobos will ever support that. :-)

ibuclaw:
> Since it's still round tripping through the stack `core.bitop.btr` is an option, but it's not a big improvement.
> 
> Link to compare functions & compilers: https://d.godbolt.org/z/bvd1Ev

To be fair, the compiler can recognize the pattern and emit btr all by itself. https://d.godbolt.org/z/r5nWM4

PR #7562, opened 2020-07-21
sha: inline critical functions

WalterBright:
This causes critical functions identified in https://github.com/dlang/phobos/pull/7534 to be always inlined.

### DIFF
--- a/std/digest/sha.d
+++ b/std/digest/sha.d
@@ -121,6 +121,7 @@ public import std.digest;
  * Helper methods for encoding the buffer.
  * Can be removed if the optimizer can inline the methods from std.bitmanip.
  */
+pragma(inline, true)
 private ubyte[8] nativeToBigEndian(ulong val) @trusted pure nothrow @nogc
 {
     version (LittleEndian)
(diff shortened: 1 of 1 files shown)


rainers:
It seems this PR has made the phobos build a lot slower, e.g. for the Linux64 build from about 20 seconds to more than 2 minutes. See https://auto-tester.puremagic.com/platform-history.ghtml?projectid=1&os=Linux_64_64

Inlining with dmd is not a great tool because it makes compilation incredibly slow.

RazvanN7:
I think that such PRs should also be accompanied with some performance tests. Is there any provable runtime benefit for this? If not, I find the performance regression unacceptable. On my machine it takes 5 minutes to compile phobos now, while before it took around 20 seconds.

n8sh:
> I think that such PRs should also be accompanied with some performance tests. Is there any provable runtime benefit for this? If not, I find the performance regression unacceptable.

[There are benchmarks](https://github.com/dlang/phobos/pull/7534#issuecomment-659708190) but considering that the distributed Phobos is pre-built with LDC anyone using the predefined SHA aliases will not see a speed difference.

>On my machine it takes 5 minutes to compile phobos now, while before it took around 20 seconds.

I am very surprised. I noticed no such slowdown when experimenting with adding `pragma(inline, true)` elsewhere.

PR #7563, opened 2020-07-22
refact,digest, for core.bitops, use aliases instead of FQNs

NilsLankila:
Partially revert ddca869335a85d, because this is what should have been done initially.

### DIFF
--- a/std/digest/sha.d
+++ b/std/digest/sha.d
@@ -113,7 +113,8 @@ else version (D_InlineAsm_X86_64)
 
 version (LittleEndian) import core.bitop : bswap;
 
-import core.bitop;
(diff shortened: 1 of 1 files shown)


NilsLankila:
TF, @WalterBright, this is what i suggested earlier today. What's going on ?

NilsLankila:
`toAlias()` is not called somewhere in DMD...

NilsLankila:
just a guess, intrinsic detection is defeated by aliases.

Geod24:
Now that the diff has been put in, is there any use in reverting ?

NilsLankila:
Duplicate of #

PR #7565, opened 2020-07-25
fabs(): make it an intrinsic

WalterBright:
Leaving the function body there means if it gets inlined, it is no longer recognized by the glue code as an intrinsic. Since the bodies serve no purpose anymore, they are removed.

Improved the documentation, too.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -4106,30 +4106,23 @@ real cbrt(real x) @trusted nothrow @nogc
 }
 
 /*******************************
- * Returns |x|
- *
+ * Compute the absolute value.
  *      $(TABLE_SV
  *      $(TR $(TH x)                 $(TH fabs(x)))
  *      $(TR $(TD $(PLUSMN)0.0)      $(TD +0.0) )
  *      $(TR $(TD $(PLUSMN)$(INFIN)) $(TD +$(INFIN)) )
  *      )
+ * It is implemented as a compiler intrinsic.
+ * Params:
+ *      x = floating point value
+ * Returns: |x|
+ * References: equivalent to `core.math.fabs`
  */
-real fabs(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.fabs(x); }
-
-///ditto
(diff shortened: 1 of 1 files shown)


WalterBright:
Blocked by https://github.com/dlang/dmd/pull/11452

ibuclaw:
Why can't core.math be the intrinsics, and std.math just forwards to them?

WalterBright:
The error is:
```
generated/freebsd/release/32/unittest/libphobos2-ut.so: undefined reference to `_D3std4math4fabsFNaNbNiNfeZe'
```
which suggests the fabs is not being recognized as an intrinsic. But it my tests, it always is.

Geod24:
> which suggests the fabs is not being recognized as an intrinsic. But it my tests, it always is.

DMD was fixed to only recognize builtin from `core` modules a few months ago, so as @ibuclaw suggested, you need to forward to `core.math.fabs`:
https://github.com/dlang/dmd/blob/20059778f26aa119658c153692df1f1b0a41f2bd/src/dmd/builtin.d#L91-L93

std/math.d
```diff
@@ -4106,30 +4106,23 @@ real cbrt(real x) @trusted nothrow @nogc
...
-///ditto
-pragma(inline, true)
-float fabs(float f) @trusted pure nothrow @nogc
+@safe pure nothrow @nogc
 {
-    uint tmp = *cast(uint*)&f & 0x7FFF_FFFF;
-    return *cast(float*)&tmp;
+    float  fabs(float  x);
+    double fabs(double x); /// ditto
+    real   fabs(real   x); /// ditto
```
n8sh, line 4125:
```suggestion
    version (StdDdoc)
    {
        float  fabs(float  x);
        double fabs(double x); /// ditto
        real   fabs(real   x); /// ditto
    }
    else
    {
        alias fabs = core.math.fabs;
    }
```

RazvanN7:
ping @WalterBright

PR #7570, opened 2020-07-29
std.traits: Add support for 'in' storage class

Geod24:
In support of the DMD PR dlang/dmd#11474 , which will be red until this is pulled.

Commit message follows:
```
For a long time, the 'in' storage class was only a second class citizen,
merely an alias for 'const', which is a type constructor.
While it was also documented for a long time as being 'scope',
this was removed when 'scope' actually started to have an effect
(when DIP1000 started to be implemented).
Currently, a switch (-preview=in) allows to get back this 'scope'.

However, the 'in' storage class does not really exists,
it gets lowered to 'const [scope]' at an early stage by the compiler,
which means that we expose what is essentially an implementation
detail to the user.

There is a PR in DMD (dlang/dmd#11474) that aims to give 'in' an actual
identity, instead of it being lowered to 'const [scope]'.
The underlying motivation is to allow for extending 'in''s functionality,
giving it the ability to pass by 'ref' when necessary, and accept rvalues.

However, regardless of the second goal, having proper support for 'in'
would lead to less confusing messages, better code generation,
and less confusion w.r.t. the behavior of `std.traits.ParameterStorageClass`.
```

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -795,14 +795,17 @@ private template fqnType(T,
 
     string storageClassesString(uint psc)() @property
     {
+        import std.conv : text;
+
         alias PSC = ParameterStorageClass;
 
-        return format("%s%s%s%s%s",
+        return text(
             psc & PSC.scope_ ? "scope " : "",
             psc & PSC.return_ ? "return " : "",
+            psc & PSC.in_ ? "in " : "",
             psc & PSC.out_ ? "out " : "",
             psc & PSC.ref_ ? "ref " : "",
-            psc & PSC.lazy_ ? "lazy " : ""
+            psc & PSC.lazy_ ? "lazy " : "",
         );
     }
 
@@ -1213,12 +1216,13 @@ enum ParameterStorageClass : uint
      * These flags can be bitwise OR-ed together to represent complex storage
      * class.
      */
-    none    = 0,
-    scope_  = 1,    /// ditto
-    out_    = 2,    /// ditto
-    ref_    = 4,    /// ditto
-    lazy_   = 8,    /// ditto
-    return_ = 0x10, /// ditto
+    none    = 0x00,
+    in_     = 0x01, /// ditto
+    ref_    = 0x02, /// ditto
+    out_    = 0x04, /// ditto
+    lazy_   = 0x08, /// ditto
+    scope_  = 0x10, /// ditto
+    return_ = 0x20, /// ditto
 }
 
 /// ditto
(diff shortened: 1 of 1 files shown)


Geod24:
Can we get this merged ? Without the DMD PR it's harmless, so we can always revert it.

std/traits.d
```diff
@@ -797,12 +797,13 @@ private template fqnType(T,
     {
         alias PSC = ParameterStorageClass;
 
-        return format("%s%s%s%s%s",
+        return format("%s%s%s%s%s%s",
```
wilzbach, line 800: Format is an overkill for pure concatenation and has pretty bad CTFE performance.
(I'm aware it was like this)

Geod24: This whole template is a bit useless... It could (should) just call the trait directly, since this already returns a tuple of strings...

PetarKirov: I'd say just change this to `std.conv.text` or pure string concatenation and let's get this PR merged.

Geod24: Done

std/traits.d
```diff
@@ -1213,12 +1216,13 @@ enum ParameterStorageClass : uint
...
-    ref_    = 4,    /// ditto
-    lazy_   = 8,    /// ditto
-    return_ = 0x10, /// ditto
+    none    = 0x00,
+    in_     = 0x01, /// ditto
+    ref_    = 0x02, /// ditto
+    out_    = 0x04, /// ditto
+    lazy_   = 0x08, /// ditto
+    scope_  = 0x10, /// ditto
+    return_ = 0x20, /// ditto
```
JohanEngelen, line 1225: @Geod24 @wilzbach @thewilsonator 
Please next time consider that changing the values of enum members can have a large impact on end users, for no clear benefit. The value changes were not needed, just adding `in_ = 0x20` would have sufficed, with less user pain. As a practical example, the reassigning of these values means that at Weka we will have to recalculate ABI hashes for many functions.

JohanEngelen:
@Geod24 @wilzbach @thewilsonator
Can we please revert the change of the values in the `ParameterStorageClass` enum? I.e. make the values correspond to:
```d
    none    = 0,
    scope_  = 1,    /// ditto
    out_    = 2,    /// ditto
    ref_    = 4,    /// ditto
    lazy_   = 8,    /// ditto
    return_ = 0x10, /// ditto
    in_     = 0x20, /// ditto
```

Geod24:
@JohanEngelen : We could change the order to that if that works better for you, I personally don't have a horse in this race, but wouldn't that potentially cause the same problems for other users now that this has been released for two releases ?
The point about the re-ordering is taken, the change should have been proposed independently and not tied to the addition of `in`.

JohanEngelen:
I guess one can also argue that users shouldn't rely on the values of the enum, and only use the value identifiers... Afaik dlang gives no guarantees about ABI compatibility between versions (i.e. shared stdlibs will not work as one might hope).
I think we'll solve it at Weka, turning the magic numbers (phobos) into stable Weka numbers using a simple conversion function.

PR #7574, opened 2020-08-02
Fix Issue 21103 - isDynamicArray instantiates unecessary templates

MoonlightSentinel:
Inspired by the discussion in dlang/dmd#9014

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -6849,7 +6849,19 @@ enum bool isStaticArray(T) = __traits(isStaticArray, T);
 /**
  * Detect whether type `T` is a dynamic array.
  */
-enum bool isDynamicArray(T) = is(DynamicArrayTypeOf!T) && !isAggregateType!T;
+template isDynamicArray(T)
+{
+    static if (is(T == U[], U))
+        enum bool isDynamicArray = true;
+    else static if (is(T U == enum))
+        // BUG: isDynamicArray / isStaticArray considers enums
+        // with appropriate base types as dynamic/static arrays
+        // Retain old behaviour for now, see
+        // https://github.com/dlang/phobos/pull/7574
+        enum bool isDynamicArray = isDynamicArray!U;
+    else
+        enum bool isDynamicArray = false;
+}
 
 ///
 @safe unittest
@@ -6873,6 +6885,24 @@ enum bool isDynamicArray(T) = is(DynamicArrayTypeOf!T) && !isAggregateType!T;
             static assert(!isDynamicArray!( SubTypeOf!(Q!T) ));
         }
     }
+
+    static assert(!isDynamicArray!(int[5]));
+
+    static struct AliasThis
+    {
+        int[] values;
+        alias values this;
+    }
+
+    static assert(!isDynamicArray!AliasThis);
+
+    // https://github.com/dlang/phobos/pull/7574/files#r464115492
+    enum E : string
+    {
+        a = "a",
+        b = "b",
+    }
+    static assert( isDynamicArray!E);
 }
 
 /**


std/traits.d
```diff
@@ -6840,7 +6840,7 @@ enum bool isStaticArray(T) = __traits(isStaticArray, T);
 /**
  * Detect whether type `T` is a dynamic array.
  */
-enum bool isDynamicArray(T) = is(DynamicArrayTypeOf!T) && !isAggregateType!T;
+enum bool isDynamicArray(T) = is(T == U[], U);
```
n8sh, line 6843:
```suggestion
enum bool isDynamicArray(T) = is(T == U[], U) && __traits(getAliasThis, T).length == 0;
```
Unless you mean to change this to allows structs etc.

MoonlightSentinel: Note the `==`, it already excludes static arrays and structs with `alias this`.

MoonlightSentinel: Added some tests

n8sh: Ah, I see then.

nordlow: Good. This at least avoids any further successive template insts.

n8sh: This currently passes but fails after this PR:

```d
import std.traits;
enum E : string
{
    a = "a",
    b = "b",
}
static assert(isDynamicArray!E);
```

MoonlightSentinel: `E` is a unique type which is convertible to `string`, so `isDynamicArray` should return `false`?

n8sh: Aside from being a behavior change it also breaks symmetry with `isStaticArray`:
```d
import std.traits;
enum E : char[1]
{
    a = ['a'],
    b = ['b'],
}
static assert(isStaticArray!E);
```

MoonlightSentinel: Interesting ... I would consider this a bug in `isStaticArray` given the wording in the spec. But they should indeed be symmetric.

andralex: Eh, enums are weird. Try this?
```suggestion
template isDynamicArray(T) {
    static if (is(T == U[], U))
    	enum bool isDynamicArray = true;
    else static if (is(T U == enum))
    	enum bool isDynamicArray = isDynamicArray!U;
    else
    	enum bool isDynamicArray = false;
}
```

andralex: fwiw this is the shortest form I could muster:
```d
enum bool isDynamicArray(T) = is(T == U[], U)
    || { static if (is(T U == enum)) return isDynamicArray!U; else return false; }();
```

n8sh (dismissed):
Marking this so it isn't mistakenly merged before the discussed issues are resolved.

std/traits.d
MoonlightSentinel: Thanks for your suggestion, but I think we should determine the "correct" behaviour first.

FWIW, @nordlow's proposed trait yields the same result:

```
import std.traits : isDynamicArray;

enum E : string
{
    a = "a",
    b = "b",
}

pragma(msg, "isStaticArray!E = ", isDynamicArray!E);
pragma(msg, "is(E == T[], T) = ", is(E == T[], T));
pragma(msg, "__traits(isStaticArray, E) = ", __traits(isStaticArray, E));

```

```
.\generated\windows\debug\64\dmd.exe -c -o- .\array.d
isStaticArray!E = true
is(E == T[], T) = false
__traits(isStaticArray, E) = false
```

MoonlightSentinel: Ping. Still convinced that `isDynamicArray!E` should yield false because `E` is a distinct type (which complies with the spec).

Note that this doesn't affect any project on BuildKite, so iI think this change is fine (with an appropriate changelog entry).

nordlow:
How about getting this merged?

PR #7576, opened 2020-08-03
Replace Unqual-based idiom with immutable-based idiom

andralex:
Aplying `immutable` to `T` makes it `immutable T` regardless of whatever qualifiers `T` has. This allows for a simple idiom for type comparison, per this PR.


thewilsonator:
`std/variant.d(500:120)[warn]: Line is longer than 120 characters`

andralex:
@thewilsonator fixed line length, will tag for merge since you approved. Thanks!

PR #7577, opened 2020-08-04
Work around issue 21111: Temporarily disable inlining of T_SHA2_16_79

andralex:
...because it increases build speed by 10x.

See also https://forum.dlang.org/thread/lcgitgilixahbnzwpwfp@forum.dlang.org

### DIFF
--- a/std/digest/sha.d
+++ b/std/digest/sha.d
@@ -494,7 +494,8 @@ struct SHA(uint hashBlockSize, uint digestSize)
             H = T1 + T2;
         }
 
-        pragma(inline, true)
+        // Temporarily disable inlining because it increases build speed by 10x.
+        // pragma(inline, true)
         static void T_SHA2_16_79(Word)(int i, ref Word[16] W,
             Word A, Word B, Word C, ref Word D, Word E, Word F, Word G, ref Word H, Word K)
             pure nothrow @nogc


andralex:
https://issues.dlang.org/show_bug.cgi?id=21111

schveiguy:
@andralex thanks!

wilzbach:
> This needs a referenced bug report, so it's not forgotten. Other than that, I approve this.

Issue needs to be referenced in the commit message, not PR title (otherwise it won't be found by the bots).

schveiguy:
@wilzbach not a problem here -- the bug is not fixed, and a buggy release didn't happen. I just wanted to make sure the workaround didn't prompt everyone to forget about the real issue.

rainers:
The issue is not new, from a cursory search:
https://issues.dlang.org/show_bug.cgi?id=17135
https://issues.dlang.org/show_bug.cgi?id=18221
https://issues.dlang.org/show_bug.cgi?id=7157
https://issues.dlang.org/show_bug.cgi?id=6401

schveiguy:
Right, that makes sense. But still, I didn't want the real problem to be swept under the rug when this was merged.

Perhaps an aggregate bug report is in order.

schveiguy:
https://issues.dlang.org/show_bug.cgi?id=21121

PR #7578, opened 2020-08-04
Fix issue 21113: Use sysctl for thisExePath on BSD.

euantorano:
Fixes [issue 21113](https://issues.dlang.org/show_bug.cgi?id=21113).

I'm quite new to D, and wasn't sure if there's a way to do a condition inside a version, like the following:

```d
version (FreeBSD || NetBSD || DragonFlyBSD) {}
```

Which would make some of this code slightly cleaner in my opinion.

Sysctl numbers are taken from relevant headers:

- [FreeBSD](https://github.com/freebsd/freebsd/blob/8b261beea6bd825db7c891eeea6aaf8621b30268/sys/sys/sysctl.h#L999)
- [NetBSD](https://github.com/NetBSD/src/blob/4455df1e707aa79dd1e5aa5cab05aca949996023/sys/sys/sysctl.h#L587)
- [DragonFlyBSD](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/4c898ae594e26f4901a584d0cfe471cb018bea11/sys/sys/sysctl.h#L610)

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -3493,9 +3493,25 @@ version (Darwin)
 else version (FreeBSD)
     private extern (C) int sysctl (const int* name, uint namelen, void* oldp,
         size_t* oldlenp, const void* newp, size_t newlen);
+else version (DragonFlyBSD)
+    private extern (C) int sysctl (const int* name, uint namelen, void* oldp,
+        size_t* oldlenp, const void* newp, size_t newlen);
 else version (NetBSD)
     private extern (C) int sysctl (const int* name, uint namelen, void* oldp,
         size_t* oldlenp, const void* newp, size_t newlen);
+
+version (FreeBSD)
+{
+    version = thisExePathHasSysCtlProcName;
+}
+else version (NetBSD)
+{
+    version = thisExePathHasSysCtlProcName;
+}
+else version (DragonFlyBSD)
+{
+    version = thisExePathHasSysCtlProcName;
+}
 
 /**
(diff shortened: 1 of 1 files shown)


thewilsonator:
> `version (FreeBSD || NetBSD || DragonFlyBSD) {}`

No, the way you have done it is the intended way. Forcing giving a name to the version gives documentation and forces you to think about the versioning logic, i.e. not making arbitrary combinations.

euantorano:
> > `version (FreeBSD || NetBSD || DragonFlyBSD) {}`
> 
> No, the way you have done it is the intended way. Forcing giving a name to the version gives documentation and forces you to think about the versioning logic, i.e. not making arbitrary combinations.

Ah, ok. That makes sense, thanks!

thewilsonator:
The linked netBSD has `#define KERN_PROC_PATHNAME 5` not 9

euantorano:
> The linked netBSD has `#define KERN_PROC_PATHNAME 5` not 9

```d
        version (NetBSD)
        {
            enum
            {
                KERN_PROC_ARGS = 48,
                KERN_PROC_PATHNAME = 5
            }

            int[4] mib = [CTL_KERN, KERN_PROC_ARGS, -1, KERN_PROC_PATHNAME];
        }
```

I think you might be misreading the `else` which applies to DragonFlyBSD. It's a little confusing because of the nesting of versions.

thewilsonator:
Oh right, I stand corrected.

PR #7581, opened 2020-08-05
Fix Issue 20984 - Heisenbug: FreeBSD64 Phobos random test suite failu…

CyberShadow:
Replace the arbitrary wait duration with a poll, which should only succeed when the result is known to be ready.

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -1609,32 +1609,39 @@ version (Posix) @system unittest
     version (Windows) TestScript prog =
        "set /p INPUT=
         echo %INPUT% output %~1
-        echo %INPUT% error %~2 1>&2";
+        echo %INPUT% error %~2 1>&2
+        echo done > %3";
     else version (Posix) TestScript prog =
        "read INPUT
         echo $INPUT output $1
-        echo $INPUT error $2 >&2";
+        echo $INPUT error $2 >&2
+        echo done > \"$3\"";
 
     // Pipes
     void testPipes(Config config)
     {
+        import std.file, std.uuid, core.thread, std.exception;
         auto pipei = pipe();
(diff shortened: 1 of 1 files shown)


CyberShadow:
Oh the irony:

```
/home/braddr/sandbox/at-client/pull-4135049-FreeBSD_64_64/dmd/generated/freebsd/release/64/dmd -conf= -m64 -fPIC -g -lowmem -i -Itools -version=NoMain -unittest -run tools/d_do_test.d
gmake[1]: *** [Makefile:217: generated/d_do_test] Segmentation fault (core dumped)
```
Is that another test suite Heisenbug? :)

MoonlightSentinel:
Looks like a recent regression, it started a few days ago and is probably related to/the same issue as the `unit` segfault.

std/process.d
```diff
@@ -1609,50 +1609,58 @@ version (Posix) @system unittest
...
+            wait(pid);
         assert(readText(patho).chomp() == "INPUT output bar");
         assert(readText(pathe).chomp().stripRight() == "INPUT error baz");
-        remove(pathi);
-        remove(patho);
-        remove(pathe);
+        while (remove(pathi).collectException) Thread.sleep(10.msecs);
+        while (remove(patho).collectException) Thread.sleep(10.msecs);
+        while (remove(pathe).collectException) Thread.sleep(10.msecs);
+        while (remove(done).collectException) Thread.sleep(10.msecs);
```
n8sh, line 1663: I'd suggest not including this change unless it is solving a known problem.

CyberShadow: Yes, they're solving a problem. Without this change there was another random failure. The files were still open by the shell even though the content was written to it.

PR #7582, opened 2020-08-06
Try to make `std.variant.maxSize` non-recursive

nordlow:
Initial discussion at https://forum.dlang.org/thread/wbpnncxepehgcswhuazl@forum.dlang.org?page=1

### DIFF
--- a/std/variant.d
+++ b/std/variant.d
@@ -77,28 +77,53 @@ import std.meta, std.traits, std.typecons;
 
 /++
     Gives the `sizeof` the largest type given.
+
+    See_Also: https://forum.dlang.org/thread/wbpnncxepehgcswhuazl@forum.dlang.org?page=1
   +/
-template maxSize(T...)
+template maxSize(Ts...)
 {
-    static if (T.length == 1)
+    align(1) union Impl
     {
-        enum size_t maxSize = T[0].sizeof;
+        static foreach (i, T; Ts)
+        {
+            static if (!is(T == void))
+                mixin("T _field_", i, ";");
+        }
     }
-    else
-    {
-        import std.algorithm.comparison : max;
-        enum size_t maxSize = max(T[0].sizeof, maxSize!(T[1 .. $]));
-    }
+    enum maxSize = Impl.sizeof;
 }
 
 ///
 @safe unittest
 {
+    struct Cat { int a, b, c; }
+
+    align(1) struct S
+    {
+        long l;
+        ubyte b;
+    }
+
+    align(1) struct T
+    {
+        ubyte b;
+        long l;
+    }
+
     static assert(maxSize!(int, long) == 8);
     static assert(maxSize!(bool, byte) == 1);
-
-    struct Cat { int a, b, c; }
     static assert(maxSize!(bool, Cat) == 12);
+    static assert(maxSize!(char) == 1);
+    static assert(maxSize!(char, short, ubyte) == 2);
+    static assert(maxSize!(char, long, ubyte) == 8);
+    import std.algorithm.comparison : max;
+    static assert(maxSize!(long, S) == max(long.sizeof, S.sizeof));
+    static assert(maxSize!(S, T) == max(S.sizeof, T.sizeof));
+    static assert(maxSize!(int, ubyte[7]) == 7);
+    static assert(maxSize!(int, ubyte[3]) == 4);
+    static assert(maxSize!(int, int, ubyte[3]) == 4);
+    static assert(maxSize!(void, int, ubyte[3]) == 4);
+    static assert(maxSize!(void) == 1);
 }
 
 struct This;


std/variant.d
```diff
@@ -75,17 +75,21 @@ import std.meta, std.traits, std.typecons;
...
-    {
-        import std.algorithm.comparison : max;
-        enum size_t maxSize = max(T[0].sizeof, maxSize!(T[1 .. $]));
+        // See_Also: https://forum.dlang.org/thread/wbpnncxepehgcswhuazl@forum.dlang.org?page=1
+        static foreach (T; Ts)
+        {
+            static if (!is(T == void))
+            {
+                pragma(msg, T.stringof ~ " _store" ~ T.mangleof ~ ";");
+                mixin(T.stringof ~ " _store" ~ T.mangleof ~ ";");
```
WebFreak001, line 88: you can make this
```suggestion
                mixin("T _store" ~ T.mangleof ~ ";");
```

quickfur: Note that `T.mangleof` does not guarantee uniqueness, e.g., `maxSizeOf!(int,int,float)`.

quickfur:
OK, I've found a way to support void:
````
template UniqName(T...) {
	struct Impl {}
	enum UniqName = Impl.mangleof;
}

template maxSizeOf(T...)
{
	align(1) union Impl {
		static foreach (i, U; T) {
			static if (is(U == void))
				mixin("ubyte[1] " ~ UniqName!(i,U) ~ ";");
			else
				mixin("U " ~ UniqName!(i,U) ~ ";");
		}
	}
	enum maxSizeOf = Impl.sizeof;
}

unittest
{
	align(1) struct S {
		long l;
		ubyte b;
	}

	align(1) struct T {
		ubyte b;
		long l;
	}

	static assert(maxSizeOf!(char) == 1);
	static assert(maxSizeOf!(char, short, ubyte) == 2);
	static assert(maxSizeOf!(char, long, ubyte) == 8);
	static assert(maxSizeOf!(long, S) == 9);
	static assert(maxSizeOf!(S, T) == 16);
	static assert(maxSizeOf!(int, ubyte[7]) == 7);
	static assert(maxSizeOf!(int, ubyte[3]) == 4);
	static assert(maxSizeOf!(int, int, ubyte[3]) == 4);
	static assert(maxSizeOf!(void, int, ubyte[3]) == 4);
	static assert(maxSizeOf!(void) == 1);
}
````
It's ugly, though, since it will instantiate a template per type passed to `maxSizeOf`. At this point, using CTFE to sum up `.sizeof` might actually be the better option. :-/

File: std/variant.d (hunk 1 above)
Comment (nordlow) [reply std/variant.d]: Done.

**Comment by nordlow** [comment]
I'll await other ideas until I choose to go the CTFE way.

**Comment by WebFreak001** [comment]
here is a simpler unique name:
```
template maxSizeOf(Ts...)
{
	align(1) union Impl {
		static foreach (i, T; Ts) {
			static if (!is(T == void))
                mixin("T _field_" ~ i.stringof ~ ";");
		}
	}
	enum maxSizeOf = Impl.sizeof;
}
```

while .stringof isn't really defined for this afaik, it always worked and a simple unittest here will make sure it keeps working :p

There is also no need to insert ubyte[1] for void because empty union Impl.sizeof == 1 (like void.sizeof)

**Comment by nordlow** [comment]
Should correct the name from `maxSize` to `maxSizeOf` while we're at it?

**Comment by wilzbach** [comment]
Could you please post a benchmark of the current results with LDC?

(We don't care about DMD here)

**Comment by PetarKirov** [comment]
@wilzbach  LDC vs DMD doesn't matter as this happens purely in the frontend

**Comment by wilzbach** [comment]
It still matters because for Linux and OSX the official DMD is still compiled with DMD ;-)

File: std/variant.d [hunk 2]
```diff
@@ -74,28 +74,48 @@ import std.meta, std.traits, std.typecons;
...
-        enum size_t maxSize = T[0].sizeof;
-    }
-    else
-    {
-        import std.algorithm.comparison : max;
-        enum size_t maxSize = max(T[0].sizeof, maxSize!(T[1 .. $]));
+    align(1) union Impl {
+        static foreach (i, T; Ts) {
+            static if (!is(T == void))
+                mixin("T _field_" ~ i.stringof ~ ";");
```
NilsLankila, line 85: mixin can implictly concatenate and convert i : `mixin("T _field_", i, ";");`

nordlow: Thanks

NilsLankila:
I'm afraid that this will have no impact on the memory used as this template is mostly CTFE.

```d
int memBefore(int variantMemberCount)
{
    enum TemplateInstanceSize = 371;
    return (TemplateInstanceSize * variantMemberCount) ;
}

int memAfter(int variantMemberCount)
{
    enum UnionDeclSize  = 590;
    enum VarDeclSize    = 297;
    enum TypeSize       = 72;
    return UnionDeclSize + (VarDeclSize + TypeSize) * variantMemberCount;
}

void main()
{
    foreach (i; 1..100)
        writeln(memBefore(i), ' ', memAfter(i));
}
```
Note that this does no take syntax copy in account (for the before calculation) but that is friendly with the Type size (for the after calculation).

Maybe that the full CTFE approach is better, e.g static foreach ?

std/variant.d
```diff
@@ -74,28 +74,48 @@ import std.meta, std.traits, std.typecons;
...
     static assert(maxSize!(int, long) == 8);
     static assert(maxSize!(bool, byte) == 1);
-
-    struct Cat { int a, b, c; }
     static assert(maxSize!(bool, Cat) == 12);
+    static assert(maxSize!(char) == 1);
+    static assert(maxSize!(char, short, ubyte) == 2);
+    static assert(maxSize!(char, long, ubyte) == 8);
+    static assert(maxSize!(long, S) == 9);
+    static assert(maxSize!(S, T) == 16);
```
quickfur, line 113: This line is causing a failure on systems with different alignment requirements where only `T.sizeof` is 12 rather than 16.

nordlow: So should we remove the line or case on some version?

quickfur: I was thinking more along the lines of using an inequality instead of a strict equality. But you'd have to think about it a bit more carefully to formulate it in a way that's meaningful.

pbackus: Would something like this work?

```d
static assert(maxSize!(S, T) == max(S.sizeof, T.sizeof));
```

RazvanN7:
@nordlow What is the status of this PR? Apart from buildkite failing from style violations everything passes.

nordlow:
> @nordlow What is the status of this PR? Apart from buildkite failing from style violations everything passes.

It's ready. I removed the WIP tag from the title. Put braces on separate lines to see if that pleases CI.

RazvanN7 (approved):

std/variant.d
RazvanN7: @nordlow Can you please replace this line with the suggested one from @pbackus ?

nordlow: Done

nordlow:
Failing buildkite tests seem to be network related. Can somebody try restarting?

RazvanN7:
> Failing buildkite jobs seem to be network related. Can somebody try restarting them?

Done

thewilsonator:
```
generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T10FormatSpecTaZQp__T17writeUpToNextSpecTSQCdQCc8NoOpSinkZQBlMFNaNlNfKQBdZb: error: undefined reference to '_D3std6format__T10FormatSpecTaZQp6fillUpMFNaNlNfZv'
--
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T14formatUnsignedTSQBgQBf8NoOpSinkTmTaZQBmFNaNbNiNfKQBimMKxSQCvQCu__T10FormatSpecTaZQpkbZv: error: undefined reference to '_D3std6format__T10FormatSpecTaZQp6flZeroMxFNaNbNdNiNfZb'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T14formatUnsignedTSQBgQBf8NoOpSinkTmTaZQBmFNaNbNiNfKQBimMKxSQCvQCu__T10FormatSpecTaZQpkbZv: error: undefined reference to '_D3std6format__T10FormatSpecTaZQp6flPlusMxFNaNbNdNiNfZb'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T14formatUnsignedTSQBgQBf8NoOpSinkTmTaZQBmFNaNbNiNfKQBimMKxSQCvQCu__T10FormatSpecTaZQpkbZv: error: undefined reference to '_D3std6format__T10FormatSpecTaZQp7flSpaceMxFNaNbNdNiNfZb'
```
maybe try rebasing instead.

nordlow:
> ```
> generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T10FormatSpecTaZQp__T17writeUpToNextSpecTSQCdQCc8NoOpSinkZQBlMFNaNlNfKQBdZb: error: undefined reference to '_D3std6format__T10FormatSpecTaZQp6fillUpMFNaNlNfZv'
> --
>   | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T14formatUnsignedTSQBgQBf8NoOpSinkTmTaZQBmFNaNbNiNfKQBimMKxSQCvQCu__T10FormatSpecTaZQpkbZv: error: undefined reference to '_D3std6format__T10FormatSpecTaZQp6flZeroMxFNaNbNdNiNfZb'
>   | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T14formatUnsignedTSQBgQBf8NoOpSinkTmTaZQBmFNaNbNiNfKQBimMKxSQCvQCu__T10FormatSpecTaZQpkbZv: error: undefined reference to '_D3std6format__T10FormatSpecTaZQp6flPlusMxFNaNbNdNiNfZb'
>   | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T14formatUnsignedTSQBgQBf8NoOpSinkTmTaZQBmFNaNbNiNfKQBimMKxSQCvQCu__T10FormatSpecTaZQpkbZv: error: undefined reference to '_D3std6format__T10FormatSpecTaZQp7flSpaceMxFNaNbNdNiNfZb'
> ```
> 
> maybe try rebasing instead.

Done

nordlow:
Only failure is broken TLS connection during building of vibe. Ok to merge?

thewilsonator:
build kite is required unfortunately, the issue is known

nordlow:
> build kite is required unfortunately, the issue is known

Ok. So, what to do?

thewilsonator:
yes

nordlow:
> yes

Auto-merge toggled upon pass?

RazvanN7:
> Auto-merge toggled upon pass?

I have my eyes on this one. :eye:

UplinkCoder:
this pr will most likely slow things down rather than speed them up because of quadratic behaviour of static foreach

nordlow:
My guess is that it's gonna be faster for small number of arguments but slower otherwise. I'll see if it get the time to do benchmark.

PR #7585, opened 2020-08-07
Make the style checks run first

schveiguy:
They are way more likely to fail than the documented unittests (which don't change much), and putting them last makes the push/test/debug cycle much longer for stuff like a missing space on an if statement.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -555,17 +555,17 @@ $(DSCANNER_DIR)/dsc: | $(DSCANNER_DIR) $(DMD) $(LIB)
 	mv $(DSCANNER_DIR)/dscanner_makefile_tmp $(DSCANNER_DIR)/makefile
 	DC=$(abspath $(DMD)) DFLAGS="$(DFLAGS) -defaultlib=$(LIB)" $(MAKE) -C $(DSCANNER_DIR) githash debug
 
-style: publictests style_lint
+style: style_lint publictests
 
 # runs static code analysis with Dscanner
-dscanner:
+dscanner: $(LIB)
 	@# The dscanner target is without dependencies to avoid constant rebuilds of Phobos (`make` always rebuilds order-only dependencies)
 	@# However, we still need to ensure that the DScanner binary is built once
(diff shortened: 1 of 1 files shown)


schveiguy:
Don't merge until the buildkite style check passes. I think there's an issue with dmd not being built.

wilzbach:
Looks like this is missing:

```
######################## Build DMD if non-existent ##############################

$(DMD):
	$(MAKE) -C $(DMD_DIR)/src -f posix.mak BUILD=$(BUILD) OS=$(OS) MODEL=$(MODEL)
```

Alternatively, one could depend on a druntime library as druntime has such a rule.

schveiguy:
@wilzbach I think I got it. I was testing on my local box with an already-built dmd, the dscanner needed the lib built (which builds dmd too). I also moved all the grep/cat stuff first, so it should be super-quick if there are simple spacing errors.

schveiguy:
OK, I made dscanner buildable on its own with the correct dependencies. I think this should be good now.

schveiguy:
So for me, who cannot follow the style guideline because my brain is wired to code a certain way, it is painful to make changes  and find out only 20 minutes later that my change is rejected because I didn't put in a space. I'd rather know that immediately up front so I can make that change and push again. It's helpful for all the other CI tasks too, because they are going to have to rerun whatever they are working on.

Because buildkite seems to be so taxed by all the other stuff, maybe it is a good idea to move it to a less used CI. But this change is good regardless of where the make command is run.

posix.mak
```diff
@@ -555,17 +555,17 @@ $(DSCANNER_DIR)/dsc: | $(DSCANNER_DIR) $(DMD) $(LIB)
 	mv $(DSCANNER_DIR)/dscanner_makefile_tmp $(DSCANNER_DIR)/makefile
 	DC=$(abspath $(DMD)) DFLAGS="$(DFLAGS) -defaultlib=$(LIB)" $(MAKE) -C $(DSCANNER_DIR) githash debug
 
-style: publictests style_lint
+style: style_lint publictests
 
 # runs static code analysis with Dscanner
-dscanner:
+dscanner: $(LIB)
```
MoonlightSentinel, line 561: The execution should not depend on LIB, it’s a build dependency for dsc...

PR #7586, opened 2020-08-08
Add makefile target for testing with no autodecode.

schveiguy:
This is so CI can run a valid target without failing. It currently does nothing as no autodecode unittests run.

step 2 is to alter a CI (probably buildkite) to run the autodecode tests.
step 3 is to start fixing issues with removing autodecode, and adding no autodecode modules to the list of tests to run.
step 4 (eventually) is to run all of Phobos with autodecode switched off in addition to normal tests.
step 5 is to deprecate autodecoding.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -150,6 +150,10 @@ ifdef ENABLE_COVERAGE
 override DFLAGS  += -cov
 endif
 
+ifdef NO_AUTODECODE
+override DFLAGS += -version=NoAutodecodeStrings
+endif
+
 UDFLAGS=-unittest -version=StdUnittest
 
 # Set DOTOBJ and DOTEXE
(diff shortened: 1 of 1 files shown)


schveiguy:
Note: the mechanism I'm going to use is to add modules to the list, and then it will be like you typed `make std/mod1.test std/mod2.test ...`

This is crude, and not scalable (eventually I want to test all the no-autodecode modules in one build). But it will work for now, and we can work on the test design later once there's a target . My makefile-fu is weak, so I'm very much interested in better ideas. But first, let's get this in so I can start submitting no-autodecode PRs!

wilzbach:
> so I'm very much interested in better ideas

Start freshly over with a D3 standard library that could even be shipped in tandem with Phobos.

schveiguy:
> Start freshly over with a D3 standard library that could even be shipped in tandem with Phobos.

Heh, I just meant better ideas for CI testing of no-autodecode.

schveiguy:
Can someone review this? I need something in place so I can start submitting my no-autodecode PRs. I don't want to merge things that will break later because they are not tested.

schveiguy:
I have tested adding the modules to the makefile that I have made work with noautodecode locally, and adjusting Phobos to obey the `NoAudotecodeStrings` version. I just need a CI to pass with this make target, so when I submit my PRs to Phobos they will get tested properly.

Thanks!

PR #7588, opened 2020-08-10
Update random.d

bachmeil:
Demonstrate basic usage of uniform01. The existing example is confusing because it mixes in feqrel without any explanation, as you can see from [this question on the forum](https://forum.dlang.org/post/utjsiocsbearbenpaiut@forum.dlang.org).

### DIFF
--- a/std/random.d
+++ b/std/random.d
@@ -2603,9 +2603,16 @@ do
     import std.math : feqrel;
 
     auto rnd = MinstdRand0(42);
-
-    assert(rnd.uniform01.feqrel(0.000328707) > 20);
-    assert(rnd.uniform01!float.feqrel(0.524587) > 20);
+    
+    // Generate random numbers in the range in the range [0, 1)
+    auto u1 = uniform01(rnd);
+    auto u2 = rnd.uniform01!float;
+    assert(u1 >= 0 && u1 < 1);
+    assert(u2 >= 0 && u2 < 1);
+    
+    // Confirm that their values are 0.000328707 and 0.524587
+    assert(u1.feqrel(0.000328707) > 20);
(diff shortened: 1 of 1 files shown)


std/random.d
```diff
@@ -2603,9 +2603,16 @@ do
...
     auto rnd = MinstdRand0(42);
-
-    assert(rnd.uniform01.feqrel(0.000328707) > 20);
-    assert(rnd.uniform01!float.feqrel(0.524587) > 20);
+    
+    // Generate random numbers in the range in the range [0, 1)
+    auto u1 = uniform01(rnd);
+    auto u2 = rnd.uniform01!float;
+    assert(u1 >= 0 && u1 < 1);
+    assert(u2 >= 0 && u2 < 1);
```
wilzbach, line 2611: Why not move the assert closer to the declaration?

```suggestion
    auto u1 = uniform01(rnd);
    assert(u1 >= 0 && u1 < 1);

    auto u2 = rnd.uniform01!float;
    assert(u2 >= 0 && u2 < 1);
```

std/random.d
```diff
@@ -2603,9 +2603,16 @@ do
...
+    
+    // Confirm that their values are 0.000328707 and 0.524587
```
wilzbach, line 2613:
```suggestion
    // Confirm that the random values with the seed 42 are 0.000328707 and 0.524587
```

wilzbach:
> Error - file 'std/random.d' contains trailing whitespace at line 2606

bachmeil:
I'm not sure what's failing. I fixed the whitespace issue.

MoonlightSentinel:
That's an unrelated issue, restarted the check for you.

PR #7590, opened 2020-08-13
Fix issue 7033: File.rawWrite is slow on Windows

canopyofstars:
To fix this issue, the newly added `openModeFlags` (and the enum `OpenMode` from which it's derived) is used to store the current open mode of a file. Thus, `rawWrite` can check if the file was already opened in binary mode to obviate temporarily changing the mode and thus executing two flushes.
A new method, `parseOpenModeString` is used to parse an open mode string, and it stores the desired open mode in `openModeFlags`. It's called in every method that changes a file's open mode. (See the documentation of this function for more information.)

I'd like your input on this, @schveiguy, since you asked to be notified about this.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -1077,6 +1077,12 @@ Throws: `ErrnoException` if the file is not opened or if the call to `fwrite` fa
         {
             immutable fd = ._fileno(_p.handle);
             immutable oldMode = ._setmode(fd, _O_BINARY);
+
+            version (DIGITAL_MARS_STDIO)
+            {
+                immutable info = __fhnd_info[fd];
+            }
+
             
             if (oldMode != _O_BINARY)
             {
@@ -1089,8 +1095,7 @@ Throws: `ErrnoException` if the file is not opened or if the call to `fwrite` fa
                 {
                     import core.atomic : atomicOp;
 
-                    // https://issues.dlang.org/show_bug.cgi?id=4243
-                    immutable info = __fhnd_info[fd];
+                    // https://issues.dlang.org/show_bug.cgi?id=4243 
                     atomicOp!"&="(__fhnd_info[fd], ~FHND_TEXT); 
                 }
             } 
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -417,8 +417,8 @@ struct File
     }
 
     /**
-        Parses an open mode string where each mode is stored as a flag in
-        _p.openModeFlags.
+        Parses an open mode string where each specified mode is stored as a
+        flag in _p.openModeFlags.
 
         This function doesn't check if the string is invalid (e.g. nonsensical
         combinations); it merely returns if an unknown character is found.
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -400,10 +400,10 @@ struct File
     package this(FILE* handle, string name, uint refs = 1, bool isPopened = false) @trusted
     {
         import core.stdc.stdlib : malloc;
-        import std.exception : enforce; 
+        import std.exception : enforce;
         assert(!_p);
         _p = cast(Impl*) enforce(malloc(Impl.sizeof), "Out of memory");
-        initImpl(handle, name, refs, isPopened); 
+        initImpl(handle, name, refs, isPopened);
     }
 
     private void initImpl(FILE* handle, string name, uint refs = 1, bool isPopened = false)
@@ -476,12 +476,12 @@ Throws: `ErrnoException` if the file could not be opened.
     this(string name, scope const(char)[] stdioOpenmode = "rb") @safe
     {
         import std.conv : text;
-        import std.exception : errnoEnforce; 
+        import std.exception : errnoEnforce;
 
         this(errnoEnforce(_fopen(name, stdioOpenmode),
                         text("Cannot open file `", name, "' in mode `",
                                 stdioOpenmode, "'")),
-                name); 
+                name);
 
         // MSVCRT workaround (issue 14422)
         version (MICROSOFT_STDIO)
@@ -520,7 +520,7 @@ Throws: `ErrnoException` if the file could not be opened.
     }
 
     ~this() @safe
-    { 
+    {
         detach();
     }
 
@@ -561,7 +561,7 @@ Throws: `ErrnoException` in case of error.
  */
     void open(string name, scope const(char)[] stdioOpenmode = "rb") @trusted
     {
-        resetFile(name, stdioOpenmode, false); 
+        resetFile(name, stdioOpenmode, false);
     }
 
     // https://issues.dlang.org/show_bug.cgi?id=20585
@@ -778,7 +778,7 @@ Params:
  */
     void fdopen(int fd, scope const(char)[] stdioOpenmode = "rb") @safe
     {
-        fdopen(fd, stdioOpenmode, null); 
+        fdopen(fd, stdioOpenmode, null);
     }
 
     package void fdopen(int fd, scope const(char)[] stdioOpenmode, string name) @trusted
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -1077,7 +1077,7 @@ Throws: `ErrnoException` if the file is not opened or if the call to `fwrite` fa
         {
             immutable fd = ._fileno(_p.handle);
             immutable oldMode = ._setmode(fd, _O_BINARY);
-            
+
             if (oldMode != _O_BINARY)
             {
                 // need to flush the data that was written with the original mode
@@ -1090,8 +1090,8 @@ Throws: `ErrnoException` if the file is not opened or if the call to `fwrite` fa
             {
                 import core.atomic : atomicOp;
 
-                immutable info = __fhnd_info[fd];
                 // https://issues.dlang.org/show_bug.cgi?id=4243
+                immutable info = __fhnd_info[fd];
                 atomicOp!"&="(__fhnd_info[fd], ~FHND_TEXT);
                 scope (exit) __fhnd_info[fd] = info;
             }
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -375,10 +375,24 @@ struct File
 
     private struct Impl
     {
+        import std.typecons : BitFlags;
+
         FILE * handle = null; // Is null iff this Impl is closed by another File
         shared uint refs = uint.max / 2;
         bool isPopened; // true iff the stream has been created by popen()
-        Orientation orientation;
+        Orientation orientation; 
+
+        enum OpenMode
+        {
+            r = 1 << 0,
+            w = 1 << 1,
+            a = 1 << 2,
+            b = 1 << 3,
+            // +
+            p = 1 << 4
+        }
+
+        BitFlags!OpenMode openModeFlags;
     }
     private Impl* _p;
     private string _name;
@@ -386,11 +400,10 @@ struct File
     package this(FILE* handle, string name, uint refs = 1, bool isPopened = false) @trusted
     {
         import core.stdc.stdlib : malloc;
-        import std.exception : enforce;
-
+        import std.exception : enforce; 
         assert(!_p);
         _p = cast(Impl*) enforce(malloc(Impl.sizeof), "Out of memory");
-        initImpl(handle, name, refs, isPopened);
+        initImpl(handle, name, refs, isPopened); 
     }
 
     private void initImpl(FILE* handle, string name, uint refs = 1, bool isPopened = false)
@@ -402,6 +415,45 @@ struct File
         _p.orientation = Orientation.unknown;
         _name = name;
     }
+
+    /**
+        Parses an open mode string where each mode is stored as a flag in
+        _p.openModeFlags.
+
+        This function doesn't check if the string is invalid (e.g. nonsensical
+        combinations); it merely returns if an unknown character is found.
+        Thus, this function should be called after checking errno.
+
+        Params:
+            modeStr = The open mode string.
+    */
+    private void parseOpenModeString(scope const(char)[] modeStr) pure nothrow
+        @safe @nogc
+    { 
+        foreach (ch; modeStr)
+        {
+            switch (ch) with (Impl.OpenMode)
+            {
+                case 'r':
+                    _p.openModeFlags |= r;
+                    break;
+                case 'w': 
+                    _p.openModeFlags |= w;
+                    break;
+                case 'a':
+                    _p.openModeFlags |= a;
+                    break;
+                case 'b':
+                    _p.openModeFlags |= b;
+                    break;
+                case '+':
+                    _p.openModeFlags |= p;
+                    break;
+                default: 
+                    return; 
+            }
+        }
+    } 
 
 /**
 Constructor taking the name of the file to open and the open mode.
@@ -424,18 +476,20 @@ Throws: `ErrnoException` if the file could not be opened.
     this(string name, scope const(char)[] stdioOpenmode = "rb") @safe
     {
         import std.conv : text;
-        import std.exception : errnoEnforce;
+        import std.exception : errnoEnforce; 
 
         this(errnoEnforce(_fopen(name, stdioOpenmode),
                         text("Cannot open file `", name, "' in mode `",
                                 stdioOpenmode, "'")),
-                name);
+                name); 
 
         // MSVCRT workaround (issue 14422)
         version (MICROSOFT_STDIO)
         {
             setAppendWin(stdioOpenmode);
         }
+
+        parseOpenModeString(stdioOpenmode);
     }
 
     /// ditto
@@ -466,7 +520,7 @@ Throws: `ErrnoException` if the file could not be opened.
     }
 
     ~this() @safe
-    {
+    { 
         detach();
     }
 
@@ -507,7 +561,7 @@ Throws: `ErrnoException` in case of error.
  */
     void open(string name, scope const(char)[] stdioOpenmode = "rb") @trusted
     {
-        resetFile(name, stdioOpenmode, false);
+        resetFile(name, stdioOpenmode, false); 
     }
 
     // https://issues.dlang.org/show_bug.cgi?id=20585
@@ -565,6 +619,8 @@ Throws: `ErrnoException` in case of error.
         {
             setAppendWin(stdioOpenmode);
         }
+
+        parseOpenModeString(stdioOpenmode);
     }
 
     private void closeHandles() @trusted
@@ -648,6 +704,8 @@ Throws: `ErrnoException` in case of error.
 
         if (name !is null)
             _name = name;
+
+        parseOpenModeString(stdioOpenmode);
     }
 
     @system unittest // Test changing filename
@@ -702,6 +760,7 @@ Throws: `ErrnoException` in case of error.
     version (Posix) void popen(string command, scope const(char)[] stdioOpenmode = "r") @safe
     {
         resetFile(command, stdioOpenmode ,true);
+        parseOpenModeString(stdioOpenmode);
     }
 
 /**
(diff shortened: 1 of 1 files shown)


std/stdio.d
```diff
@@ -375,10 +375,24 @@ struct File
...
+        {
+            r = 1 << 0,
+            w = 1 << 1,
+            a = 1 << 2,
+            b = 1 << 3,
+            // +
+            p = 1 << 4
+        }
+
+        BitFlags!OpenMode openModeFlags;
```
MoonlightSentinel, line 395: - Make it `version(Windows)` (it's only used on windows after all)
- `BitFlags` is redundant given your enum definition
- `openModeFlags` should be placed after `isPopened` for better packing.

std/stdio.d
```diff
@@ -375,10 +375,24 @@ struct File
...
     {
+        import std.typecons : BitFlags;
+
         FILE * handle = null; // Is null iff this Impl is closed by another File
         shared uint refs = uint.max / 2;
         bool isPopened; // true iff the stream has been created by popen()
-        Orientation orientation;
+        Orientation orientation; 
+
+        enum OpenMode
```
MoonlightSentinel, line 385: Short DDOC comments are appreciated

std/stdio.d
```diff
@@ -1074,21 +1134,25 @@ Throws: `ErrnoException` if the file is not opened or if the call to `fwrite` fa
...
-            {
-                import core.atomic : atomicOp;
+        { 
+            // Change mode iff current mode isn't binary
+            if (!(_p.openModeFlags & _p.OpenMode.b))
+            { 
+                flush(); // before changing translation mode
+                immutable fd = ._fileno(_p.handle);
+                immutable mode = ._setmode(fd, _O_BINARY);
+                scope(exit) ._setmode(fd, mode);
```
MoonlightSentinel, line 1144: This now happens before the actual write because the `if` introduces a scope relevant for `scope (exit)`.

CyberShadow:
Is this actually faster? Are there any benchmarks?

I think the proper way to do bulk writes is to use `lockingBinaryWriter`.

schveiguy:
> I think the proper way to do bulk writes is to use `lockingBinaryWriter`.

I think you are right -- in the general case. However, when you open a file for writing binary mode (as the bug report indicates), the fact that `rawWrite` is going to flush before and after every write is insane. So we still should try and solve that problem.

However, my view on the approach has changed, now that I've looked at this more closely. I originally thought you could cache the last mode (binary or not), but there are other modes besides BINARY which the FILE could be. Plus there's the valid concern that @MoonlightSentinel brings up that someone can alter the mode outside of the `File` operations.

Here is my recommendation for implementation now (sorry for not giving more time to it earlier):

1. Create a member `bool _likelyBinary`, which is set to true when we expect the FILE * is in binary mode. You do not need to parse the input string, just use `auto curMode = _setmode(_O_BINARY)` to read the current mode (and then set it back if necessary).
2. Inside `rawWrite`, if `_likelyBinary` is true, try setting the mode before flushing, paying attention to the previous mode. If the mode is not binary, then clear the flag, set it back to the original mode, and run the current code (flush 2x, etc). If it is binary, no need to set the mode at the end, and no need to flush either time. Conversely, if it was binary, set the flag to avoid any further flushing.

So rough pseudocode:
```D
   private void initializeLikelyBinary() { // called from all ctors
      auto oldMode = _setMode(fd, _O_BINARY);
      _setMode(fd, oldMode);
      _likelyBinary = oldMode == _O_BINARY;
   }

   ...

   rawWrite(T[] arr) {
      int oldMode;
      if(_likelyBinary) {
         oldMode = _setMode(fd, _O_BINARY);
         if(oldMode != _O_BINARY) {
              // oops, it wasn't binary!
              _likelyBinary = false;
              _setMode(fd, oldMode);
         }
         else
              goto skipFlush;
      }
      flush();
      oldMode = _setMode(fd, _O_BINARY);
      if(oldMode == _O_BINARY)
         _likelyBinary = true; // don't do it next time
skipFlush:
      ... // normal rawWrite
      if(!_likelyBinary) {
          flush();
          _setMode(fd, oldMode);
      }
   }
```

All this should be versioned with Windows. This should solve the bug, AND satisfy @MoonlightSentinel's concerns.

CyberShadow:
That sounds like it's in the right direction. Though, if `_setmode` is O(1), then there is no need for `_likelyBinary`. We can look at the implementation for DMC, at least.

I think it would be good to introduce a `version` indicating whether text mode and binary mode are effectively the same, just so it's clear which particular difference between Windows and POSIX is addressed by using conditional compilation here.

schveiguy:
It's likely O(1), but still an opaque call at least, and at worst may have to lock something, which might make it more expensive? I'd expect that the likelihood that the binary mode changes between calls to `rawWrite` is pretty low.

For sure, we can do it without storing an extra boolean in the heap-allocated struct, but the cost of storing that extra boolean is pretty small.

CyberShadow:
> For sure, we can do it without storing an extra boolean in the heap-allocated struct, but the cost of storing that extra boolean is pretty small.

I was thinking more about the complexity and the additional state (e.g. to a casual observer it wouldn't be clear what the interaction is between `likelyBinary` and `orientation`).

schveiguy:
The single point of the boolean is to do the minimum `_setmode` calls. We have to do at least one to see what the mode is. Certainly it's an optimization and isn't required to make the code faster than the current situation.

And thinking about it more, if you use `_setmode` and find out it *wasn't* binary, the cost of calling the extra `_setmode` is likely miniscule compared to the premature flushing that is about to happen. So sure, let's do it without the boolean at first. It makes the ctor simpler too. I'll update my proposal above.

canopyofstars:
Implemented the proposal.

By the way, this is DMC's `setmode` [implementation](https://github.com/DigitalMars/dmc/blob/9a774f3f2b3227fd416ec3a83cb9eb8f8751425f/src/core/SETMODE.C); rather, this is it locking streams in a loop.
```c
int setmode(int fd, int mode)
{
    // ...
    for (fp = &_iob[0]; fp < &_iob[_NFILE]; fp++)
    {
        __fp_lock(fp);
        // ...
    }
    // ...
}
```

std/stdio.d
canopyofstars: Not that it matters anymore, but isn't the whole point of `BitFlags` to provide a safer interface than the raw enum value? If that's the case, I don't think it's superfluous.

schveiguy:
> By the way, this is DMC's `setmode` [implementation](https://github.com/DigitalMars/dmc/blob/9a774f3f2b3227fd416ec3a83cb9eb8f8751425f/src/core/SETMODE.C); rather, this is it locking streams in a loop.

Yikes! Note that DMC has a limit of 64 open file descriptors anyway. But yeah, that's painful. However, it's clear from that code that the binary mode is not set on the file descriptor but rather the FILE * or `_iob`. Perhaps locking that and checking the `flag` member instead of using setmode to determine the current mode is an option. There's already precedent for this inside std.stdio.

schveiguy:
Trailing whitespace on lines 1086, 1098, 1099, 1101

std/stdio.d
```diff
@@ -1075,27 +1075,53 @@ Throws: `ErrnoException` if the file is not opened or if the call to `fwrite` fa
...
             version (DIGITAL_MARS_STDIO)
             {
-                import core.atomic : atomicOp;
-
-                // https://issues.dlang.org/show_bug.cgi?id=4243
                 immutable info = __fhnd_info[fd];
-                atomicOp!"&="(__fhnd_info[fd], ~FHND_TEXT);
-                scope(exit) __fhnd_info[fd] = info;
             }
-            scope(exit) flush(); // before restoring translation mode
```
schveiguy, line 1091: You can still use scope guards throughout here. Just prefix them with the proper `if`. e.g.:

```D
scope(exit) if (oldMode != _O_BINARY) flush();
```

Same with the setting of the mode, and the storing of the info back.

also restore the whole `version (DIGITAL_MARS_STDIO)` block as it was, and put it after the if statement below.

canopyofstars: Am I to assume to that the atomic operation performed in that version block doesn't have a relatively discernible performance penalty?

schveiguy: It's just 1. easier to follow if you put everything into one version block, and 2. easier to prove that it does the same thing as before (your ordering of calls was different than the original).

canopyofstars:
> Yikes! Note that DMC has a limit of 64 open file descriptors anyway. But yeah, that's painful. However, it's clear from that code that the binary mode is not set on the file descriptor but rather the FILE * or `_iob`. Perhaps locking that and checking the `flag` member instead of using setmode to determine the current mode is an option. There's already precedent for this inside std.stdio.

Unless I misunderstood, it seems that there is no _flag field for MS' [implementation](https://github.com/dlang/druntime/blob/master/src/core/stdc/stdio.d).
```c
else version (CRuntime_Microsoft)
{
    // ...

    ///
    struct _iobuf
    {
        void* undefined;
    }

    // ...
}
```

std/stdio.d
```diff
@@ -1075,21 +1075,37 @@ Throws: `ErrnoException` if the file is not opened or if the call to `fwrite` fa
...
+                .setmode(fd, _O_BINARY);
+            }
+
             version (DIGITAL_MARS_STDIO)
             {
                 import core.atomic : atomicOp;
 
-                // https://issues.dlang.org/show_bug.cgi?id=4243
                 immutable info = __fhnd_info[fd];
+                // https://issues.dlang.org/show_bug.cgi?id=4243
```
schveiguy, line 1094: Probably an artifact from refactoring? Can you move the comment back?

std/stdio.d
```diff
@@ -1075,21 +1075,37 @@ Throws: `ErrnoException` if the file is not opened or if the call to `fwrite` fa
 
         version (Windows)
         {
-            flush(); // before changing translation mode
             immutable fd = ._fileno(_p.handle);
-            immutable mode = ._setmode(fd, _O_BINARY);
-            scope(exit) ._setmode(fd, mode);
+            immutable oldMode = ._setmode(fd, _O_BINARY);
+
```
schveiguy, line 1080: trailing whitespace line 1080 here (according to autotester)

schveiguy:
> Unless I misunderstood, it seems that there is no _flag field for MS

Yep, you did not misunderstand. You would have to do something separate for digital mars stdio. It's not critical to do now I don't think, let's just get this in, and we can worry about squeezing more performance later. I still think the speed is going to be possibly comparable, and I also don't imagine many people are using DMC runtime over MSVC runtime.

It might be a good gut-check to run against the DMC runtime with the original bug report (but set the open mode to non-binary) and see how it does before and after.

std/stdio.d
canopyofstars: Should be good, assuming my editor did its job since I'm incapable of discerning needless whitespace.

std/stdio.d
canopyofstars: Done.

canopyofstars:
> Yep, you did not misunderstand. You would have to do something separate for digital mars stdio. It's not critical to do now I don't think, let's just get this in, and we can worry about squeezing more performance later. I still think the speed is going to be possibly comparable, and I also don't imagine many people are using DMC runtime over MSVC runtime.

Wouldn't it be something to this effect?

```d
version (Windows)
{
	immutable fd = ._fileno(_p.handle);

	version (MICROSOFT_STDIO)
	{
		// Do changes made in this commit.
	}

	version (DIGITAL_MARS_STDIO)
	{
		immutable info = __fhnd_info[fd];

		if (info & FHND_TEXT)
		{
			flush();
			atomicOp!"&="(__fhnd_info[fd], ~FHND_TEXT);
		}
		
		scope (exit) __fhnd_info[fd] = info;
	}
}
```

schveiguy:
> Wouldn't it be something to this effect?

No, you have to duplicate the code in DMC and just read the flags (though you don't have to search, you know the `FILE *`). The `__fhnd_info` seems to be a separate piece related to that old bug.

canopyofstars:
> No, you have to duplicate the code in DMC runtime and just read the flags (though you don't have to search, you know the `FILE *`). The `__fhnd_info` seems to be a separate piece related to that old bug.

Understood. But as you mentioned, the prevalence of DMC's runtime makes me question whether it's worth it. Regardless, thank you for posting this issue on the NG. I didn't do all that much but bungle your proposal's implementation, but it was still a learning experience.

schveiguy:
> I didn't do all that much but bungle your proposal's implementation, but it was still a learning experience.

You did great! Thanks for making this fix.

schveiguy:
@CyberShadow any more comments on this?

CyberShadow:
Well, in lieu of tests, some benchmark numbers would be nice, just so we're sure we're solving the right performance problem. (I see in principle why this would obviously be faster, but also, programs like `cat` don't buffer at all, and that hasn't caused problems, because the OS also does buffering unless the file was opened in direct mode.) That would also be an opportunity to check the effect of minimizing `_setmode` calls.

schveiguy:
> programs like `cat` don't buffer at all

Sure it does: https://github.com/coreutils/coreutils/blob/5b8161ee4dc95308c9ef89a5268aa87a7f32d4bf/src/cat.c#L513-L517

buffering is faster because you spend less system calls. Yes, the OS does buffering, but the cost of getting at the buffer is much higher than writing to memory.

In the case of `cat`, it can decide how much data to fetch at a time (because the whole file is available). The use case here provides data a tiny bit at a time (i.e. the bug report example). flushing after every tiny bit is always going to perform worse than flushing after a buffer worth of data 100% of the time.

> in lieu of tests

I hate to set up CI tests that depend on performance, because we can't control the system being used to test.

> some benchmark numbers would be nice

I agree. I can see about running this PR on a Windows box. @canopyofstars if you have some benchmarks you can run, it would be useful to post here.

CyberShadow:
> Sure it does: https://github.com/coreutils/coreutils/blob/5b8161ee4dc95308c9ef89a5268aa87a7f32d4bf/src/cat.c#L513-L517
> 
> buffering is faster because you spend less system calls. Yes, the OS does buffering, but the cost of getting at the buffer is much higher than writing to memory.

No, `cat` does not buffer in the sense that C `FILE` does. `cat` continuously waits until data arrives, and as soon as it does, it writes out however much it received. You can check this by doing e.g. `for n in 1 2 3 4 5 ; do printf %s $n ; sleep 1 ; done | cat > a.txt`.

> In the case of `cat`, it can decide how much data to fetch at a time (because the whole file is available).

Pretty sure that's not how it works.

schveiguy:
The use case is completely different. Yes, `cat` reads as much as it has buffer space for, and immediately writes it out, but it does read it into a buffer. What it doesn't do is hold back data to write until it has a full buffer's worth of data, even if the read source is slow. In this sense, `cat` performs as well as the underlying source performs, and that's fine for that use case.

There is no way to for `cat` to duplicate the problem we are experiencing here, because what would need to happen is to have the data available, but only 4 bytes provided per read call. The kernel doesn't work that way.

My usual way of showing that buffering is worth it is to use `dd`:

```sh
$ time dd if=/dev/zero of=zeroes bs=4 count=1000000
1000000+0 records in
1000000+0 records out
4000000 bytes (4.0 MB, 3.8 MiB) copied, 1.91583 s, 2.1 MB/s

real	0m1.918s
user	0m0.681s
sys	0m1.237s
$ time dd if=/dev/zero of=zeroes bs=4000 count=1000
1000+0 records in
1000+0 records out
4000000 bytes (4.0 MB, 3.8 MiB) copied, 0.0069534 s, 575 MB/s

real	0m0.011s
user	0m0.004s
sys	0m0.007s
```

Both write 4MB, the first does it 4 bytes at a time, the second does it 4000 bytes at a time.

CyberShadow:
> The use case is completely different. Yes, `cat` reads as much as it has buffer space for, and immediately writes it out, but it does read it into a buffer. What it doesn't do is hold back data to write until it has a full buffer's worth of data, even if the read source is slow. In this sense, `cat` performs as well as the underlying source performs, and that's fine for that use case.

Yes, that's exactly what I meant. What I meant to say in addition to that is that, when `cat`'s output is a file, there is no additional buffering going on.

> There is no way to for `cat` to duplicate the problem we are experiencing here, because what would need to happen is to have the data available, but only 4 bytes provided per read call. The kernel doesn't work that way.

The kernel does the buffering in the case of process pipes. I'm not sure if there's a guarantee there, though. I see there's a way to create a datagram-like pipe, but I didn't get it to work.

> My usual way of showing that buffering is worth it is to use `dd`:

Good illustration, thanks.

canopyofstars:
I've benchmarked two programs, where one uses the current rawWrite (`current.exe`) and the other uses the version from this PR (`new.exe`). Each program is built from the D example used in the original [issue](https://issues.dlang.org/show_bug.cgi?id=7033#c0). `bytes_test.dat` is written to an SSD (a SSD?).

ldc2 was used to compile both executables.
| Program   | Command                                      |
| ------------|------------------------------------|
| `current.exe` | `ldc2 --O2 --of current.exe test.d` |
| `new.exe`       | `ldc2 --O2 --of new.exe test.d` |

_Please note that `new.exe` was built in the presence of this PR's `std/stdio.d`._

To benchmark them, I created a brief [PowerShell script](https://gist.github.com/canopyofstars/cac3446002eef3b210a1fc779898c458). This script will print out the mean execution time (in milliseconds) over a number of repetitions (e.g. 1000).

`.\benchmark.ps1 -prog .\current.exe -reps 1000`
`.\benchmark.ps1 -prog .\new.exe -reps 1000`

The results:
| Program   | Time (ms) |
|------------|-----------|
| `current.exe` | `3216.4089` |
| `new.exe` | `76.3247` |

As you can see, `current.exe` is __~42x__ slower.

schveiguy:
Thanks @canopyofstars. If possible, do you think you can run dmd instead of ldc, and make sure you use -m32 to ensure we are using the DMC runtime? I still think the improvement will be just as stark.

CyberShadow:
> As you can see, `current.exe` is, on average, **~42x** slower on my machine.

Fantastic. Thank you very much. It's good to have numbers like this on record to justify the change.

canopyofstars:
> Thanks @canopyofstars. If possible, do you think you can run dmd instead of ldc, and make sure you use -m32 to ensure we are using the DMC runtime? I still think the improvement will be just as stark.

| Program       | Command                                |
|---------------|----------------------------------------|
| `current.exe` | `dmd -m32 -O -of="current.exe" test.d` |
| `new.exe`     | `dmd -m32 -O -of="new.exe" test.d`     |

`.\benchmark.ps1 -prog .\current.exe -reps 100`
`.\benchmark.ps1 -prog .\new.exe -reps 100`

_I reduced the number of iterations in the interest of time. After a few tests with `new.exe`, I don't believe this change will have a significant impact on the results._

| Program       | Time (ms)   |
|---------------|-------------|
| `current.exe` | `3361.1693`  |
| `new.exe`     | `119.3520` |

In this instance, `current` is __~28x__ slower.

> Fantastic. Thank you very much. It's good to have numbers like this on record to justify the change.

Indeed. I should've done so from the beginning with my first commit.

wilzbach:
> Thanks @canopyofstars. If possible, do you think you can run dmd instead of ldc, and make sure you use -m32 to ensure we are using the DMC runtime? I still think the improvement will be just as stark.

FWIW (and future discussions) we no longer cater to DMD outdated optimizer or DMC's runtime here. It's a waste of time.
So the PR would have been good to go already ;-)

schveiguy:
> we no longer cater to DMD outdated optimizer or DMC's runtime here.

understood. I just wanted to make sure it's not *worse*. I didn't see how it could be, but we do have another option in terms of handling DMC's runtime without the slow loop that currently is `setmode`. But the question has been answered, so we are all set.

PR #7595, opened 2020-08-16
Make it so std.utf can be tested with no autodecoding.

schveiguy:
First module to pass with no autodecoding!

Note that I changed a few unittests in std.array. These aren't strictly necessary to get this one to pass, but they are in the right direction (I was trying to do std.array first, but I found it had too many dependencies on std.utf, so I tackled that one first). std.array is next, but still isn't passing all unittests.

The goal is to make it so eventually, one can simply disable autodecoding and all unittests pass in all dlang repositories.

DO NOT MERGE THIS unless the buildkite job that tests autodecoding passes.

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -887,11 +887,22 @@ template equal(alias pred = "a == b")
 
     enum hasFixedLength(T) = hasLength!T || isNarrowString!T;
 
+    // use code points when comparing two ranges of UTF code units that aren't
+    // the same type. This is for backwards compatibility with autodecode
+    // strings.
+    enum useCodePoint(R1, R2) =
+        isSomeChar!(ElementEncodingType!R1) && isSomeChar!(ElementEncodingType!R2) &&
+        (ElementEncodingType!R1).sizeof != (ElementEncodingType!R2).sizeof;
+
     /++
     Compares two ranges for equality. The ranges may have
     different element types, as long as `pred(r1.front, r2.front)`
     evaluates to `bool`.
     Performs $(BIGOH min(r1.length, r2.length)) evaluations of `pred`.
+
+    If the two ranges are different kinds of UTF code unit (`char`, `wchar`, or
+    `dchar`), then the arrays are compared using UTF decoding to avoid
+    accidentally integer-promoting units.
 
     Params:
         r1 = The first range to be compared.
@@ -902,7 +913,8 @@ template equal(alias pred = "a == b")
         for element, according to binary predicate `pred`.
     +/
     bool equal(Range1, Range2)(Range1 r1, Range2 r2)
-    if (isInputRange!Range1 && isInputRange!Range2 &&
+    if (!useCodePoint!(Range1, Range2) &&
+        isInputRange!Range1 && isInputRange!Range2 &&
         is(typeof(binaryFun!pred(r1.front, r2.front))))
     {
         static assert(!(isInfinite!Range1 && isInfinite!Range2),
@@ -928,7 +940,7 @@ template equal(alias pred = "a == b")
         // can be avoided if they have the same ElementEncodingType
         else static if (is(typeof(pred) == string) && pred == "a == b" &&
             isAutodecodableString!Range1 != isAutodecodableString!Range2 &&
-            is(ElementEncodingType!Range1 == ElementEncodingType!Range2))
+            is(immutable ElementEncodingType!Range1 == immutable ElementEncodingType!Range2))
         {
             import std.utf : byCodeUnit;
 
@@ -967,6 +979,14 @@ template equal(alias pred = "a == b")
             static if (!isInfinite!Range1)
                 return r2.empty;
         }
+    }
+
+    /// ditto
+    bool equal(Range1, Range2)(Range1 r1, Range2 r2)
+    if (useCodePoint!(Range1, Range2))
+    {
+        import std.utf : byDchar;
+        return equal(r1.byDchar, r2.byDchar);
     }
 }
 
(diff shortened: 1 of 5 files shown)


schveiguy:
AAAAAAH why is Azure pipeline failing in xtest46_gc? Wasn't this fixed already?

MoonlightSentinel:
I think you are referring to the `runnable/testptrref_gc.d` failure which happened for `dlang/dmd`.
Let's try the same "fix":#7596

schveiguy:
Ah, OK, this is a different project. Yes, let's try that.

schveiguy:
OK, that seems to have fixed it, thanks @MoonlightSentinel

nordlow:
That went fast!

std/algorithm/comparison.d
```diff
@@ -887,12 +887,23 @@ template equal(alias pred = "a == b")
 
     enum hasFixedLength(T) = hasLength!T || isNarrowString!T;
 
+    // use code points when comparing two ranges of UTF code units that aren't
+    // the same type. This is for backwards compatibility with autodecode
+    // strings.
+    enum useCodePoint(R1, R2) =
```
andralex, line 893: `private`

std/algorithm/comparison.d
```diff
@@ -902,7 +913,8 @@ template equal(alias pred = "a == b")
         for element, according to binary predicate `pred`.
     +/
     bool equal(Range1, Range2)(Range1 r1, Range2 r2)
-    if (isInputRange!Range1 && isInputRange!Range2 &&
+    if (!useCodePoint!(Range1, Range2) &&
```
andralex, line 916: Now we have an undocumented artifact `useCodePoint` used in the interface.

std/array.d
```diff
@@ -2175,21 +2196,12 @@ if (isInputRange!RoR &&
...
-                assert(jarr == arr2);
-            }}
-        }
-        else
-        {
-            // Turning off autodecode means the join() won't
-            // just convert arr[] to dchar, so mixing char
-            // types fails to compile.
-        }
+        static foreach (S; AliasSeq!(char,wchar,dchar))
```
andralex, line 2199: space after comma

std/range/primitives.d
```diff
@@ -2406,10 +2406,19 @@ if (isAutodecodableString!(T[]) && !isAggregateType!(T[]))
...
 
 /**
-Autodecoding is enabled if this is set to true.
+EXPERIMENTAL: to try out removing autodecoding, set the version
+`NoAutodecodeStrings`. Most things are expected to fail with this version
+currently.
 */
-
-enum autodecodeStrings = true;
+version (NoAutodecodeStrings)
```
andralex, line 2413: this is a disaster

std/range/primitives.d
```diff
@@ -2406,10 +2406,19 @@ if (isAutodecodableString!(T[]) && !isAggregateType!(T[]))
...
+{
+    enum autodecodeStrings = false;
```
andralex, line 2415: this is a disaster

std/range/primitives.d
```diff
@@ -2406,10 +2406,19 @@ if (isAutodecodableString!(T[]) && !isAggregateType!(T[]))
...
+}
+else
+{
+    ///
+    enum autodecodeStrings = true;
```
andralex, line 2420: this is a disaster

std/algorithm/comparison.d
schveiguy: It's inside an  eponymous template. Can't access it.

std/algorithm/comparison.d
schveiguy: I can spell out the `useCodePoint` thing, or make it more public. I have a feeling it's going to be needed elsewhere anyway.

std/array.d
schveiguy: This is copy-pasted from above, I didn't change anything.

CyberShadow:
> This line of attack with versioning the entire library on autodecoding is a massive disaster in the making.
> 
> PLEASE DO NOT DO THIS.

Calm down, this is just for CI and evaluation of impact. We can't make this switch accessible to D users anyhow because `phobos.lib` still needs to be compiled with it either on or off.

As I understand, probably the way forward would be:
1. Fix Phobos to work with autodecoding off, as Steven is doing right now
2. Try disabling autodecoding globally in a pull request (this is mentioned in the PR description)
3. Using the project tester (BuildKite), assess breakages in the greater ecosystem
4. Fix stuff in Phobos where applicable to let things work with autodecoding off
5. Fix community projects so they work with autodecoding either on or off
6. Turn autodecoding off
7. Rejoice
8. Delete the old `NoAutodecodeStrings` blocks.

Users who want to try out no-autodecoding Phobos early can do so with Digger (e.g. `digger run master+phobos#7595 -- dub build --compiler=dmd`).

It remains to be seen at around step 5 if this approach is feasible. @schveiguy can correct me if I got anything wrong :)

schveiguy:
@CyberShadow thanks for the analysis, and I pretty much agree with everything you wrote.

At this time, I have had a phone call with both Andrei and Walter, and I am working on an experiment to see how feasible it is to have a v202x package inside Phobos which allows people to use an "updated" version of Phobos without having to change Phobos. I'm not sure yet what it's going to look like, but the idea is to see if I can wrap Phobos but remove the autodecoding portions without writing any significant code aside from plumbing. I'm not sure how pretty or automated this can be done, but I also don't want to work on something that is sure to be removed/rejected later.

So I will post my findings when I'm done there. My plan is to have this exact PR done, but built as a wrapper on top of Phobos instead of using a version identifier.

wilzbach:
> this is just for CI and evaluation of impact.

FWIW it's already merged and affects the documentation and public examples:

https://dlang.org/phobos-prerelease/std_array.html#.array.2

PR #7598, opened 2020-08-17
Reduce the number of isAggregateType!T checks

n8sh:
Some are redundant. Others can be replaced by cheaper traits checks.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -5638,7 +5638,6 @@ Note: Trying to use returned value will result in a
 //::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::://
 
 private template AliasThisTypeOf(T)
-if (isAggregateType!T)
 {
     alias members = __traits(getAliasThis, T);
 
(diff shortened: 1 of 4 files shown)


std/traits.d
```diff
@@ -6967,7 +6966,7 @@ enum bool isSIMDVector(T) = is(T : __vector(V[N]), V, size_t N);
 /**
  * Detect whether type `T` is a pointer.
  */
-enum bool isPointer(T) = is(T == U*, U) && !isAggregateType!T;
+enum bool isPointer(T) = is(T == U*, U) && __traits(isScalar, T);
```
n8sh, line 6969: It's not clear to me that `is(T == U*, U)` could ever be true for an aggregate type `T` but I kept a check just in case there's some nuance I'm overlooking.

MoonlightSentinel: Seems like a workaround for an old bug:

```
struct S
{
    int* ptr;
    alias ptr this;
}

pragma(msg, is(S == T*, T));
```
```
2.063   to 2.083.1: Success with output: true
Since      2.084.1: Success with output: false
```

PR #7600, opened 2020-08-19
Fix Issue 14224 - std.range chunks of variable chunk sizes

AsterMiha:


std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+If the source range is shorter than the sum of the sizes, only the sizes that
+cover the range will be used. In case the last available elements of the source
+range don't reach the specified size, the chunk will only contain those elements.
+
+Params:
+    source = range from which the chunks will be selected
+    chunkSizes = range specifying the size of each chunk
+*/
+struct VarChunks(Source, Sizes)
+if (isInputRange!Source && isInputRange!Sizes && isIntegral!(ElementType!Sizes))
```
MoonlightSentinel, line 8622: Why limit this to plain integrals?
```suggestion
if (isInputRange!Source && isInputRange!Sizes && is(ElementType!Sizes : size_t))
```

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+        {
+
+            auto opIndex(size_t index)
+            {
+                size_t start = 0;
+                size_t size;
+                Sizes sizesCopy = _chunkSizes.save();
+                while (index)
+                {
+                    start += sizesCopy.front();
```
MoonlightSentinel, line 8694: `sizeCopy` might be empty before `index` is `0`

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+                return length;
+            }
+
+        static if (hasSlicing!Source && hasSlicing!Sizes)
+        {
+
+            auto opIndex(size_t index)
+            {
+                size_t start = 0;
+                size_t size;
```
MoonlightSentinel, line 8690: Declare variables on first use. Ditto below

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+                    sizesCopy.popFront();
+                    index--;
+                }
+                size = sizesCopy.front();
+
+                static if (isInfinite!Source)
+                    return _source[start..(start+size)];
+                else
+                {
+                    size_t len = _source.length;
```
MoonlightSentinel, line 8704: Use `const` wherever possible

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+
+If the source range is shorter than the sum of the sizes, only the sizes that
+cover the range will be used. In case the last available elements of the source
+range don't reach the specified size, the chunk will only contain those elements.
+
+Params:
+    source = range from which the chunks will be selected
+    chunkSizes = range specifying the size of each chunk
+*/
+struct VarChunks(Source, Sizes)
```
MoonlightSentinel, line 8621:
```suggestion
private struct VarChunks(Source, Sizes)
```

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+
+                _source = _source[0..(start-lastSize)];
+            }
+        }
+
+    private:
+        Source _source;
+        Sizes _chunkSizes;
+    }
+    else static if (isIntegral!(ElementType!Sizes))
```
MoonlightSentinel, line 8788: Redundant to the template constraint

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+
+        private this(Source r, Sizes s)
+        {
+            impl = RefCounted!Impl(r, s, r.empty ? 0 : (s.empty ? 0 : s.front));
+        }
+
+        @property bool empty() { return impl.r.empty || impl.s.empty; }
+        @property Chunk front() return { return Chunk(impl); }
+
+        void popFront()
```
MoonlightSentinel, line 8824: This `popFront` implementation invalidates the last returned `front` while the previous version probably keeps the last element valid. That is quite dangerous and should at least be documented.

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+            if (!impl.r.empty && !impl.s.empty)
+            {
+                impl.s.popFront();
+                impl.curSizeLeft = impl.s.front;
+            }
+            else
+                impl.curSizeLeft = 0;
+        }
+
+        static assert(isInputRange!(typeof(this)));
```
MoonlightSentinel, line 8836: This is always true

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+    }
+}
+
+VarChunks!(Source, Sizes) varChunks(Source, Sizes)(Source source, Sizes chunkSizes)
```
MoonlightSentinel, line 8840: The name is bad, I think it should simply become another overload of `chunks`

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+if (isInputRange!Source && isInputRange!Sizes && isIntegral!(ElementType!Sizes))
+{
+    return typeof(return)(source, chunkSizes);
+}
+
+@safe unittest
```
MoonlightSentinel, line 8846: Documentation examples require a leading `///`
```suggestion
///
@safe unittest
```

std/range/package.d
MoonlightSentinel: Also needs a detailed DDoc comment

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+                else
+                    return _source.empty;
+            }
+            else
+                enum empty = false;
+
+        @property auto front()
+        {
+            assert(!_source.empty, "Attempting to fetch the front of an empty VarChunks");
+            assert(!_chunkSizes.empty, "Not enough chunkSizes to create more VarChunks");
```
MoonlightSentinel, line 8647: These messages are pretty implementation-specific, maybe condense it for users.
```suggestion
            assert(!empty, "Attempting to fetch the front of an empty range");
```

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+    assert(chunks.empty == false);
+    assert(array(chunks[1 .. 3]) == [[2, 3], [4, 5, 6]]);
+    assert(equal(retro(array(chunks)), array(retro(chunks))));
+    static assert(isRandomAccessRange!(typeof(chunks)));
+}
+
+@safe unittest
+{
+    import std.algorithm.comparison : equal;
+    auto source = [1, 2, 3, 4, 5, 6, 7, 8, 9];
```
MoonlightSentinel, line 8868: Needs a test for excess elements in `_source`

MoonlightSentinel (commented):
Rough first review

std/range/package.d
```diff
@@ -8606,6 +8606,302 @@ if (isForwardRange!Source && hasLength!Source)
...
+            _source = source;
+            _chunkSizes = chunkSizes;
+        }
+
+        // Input range primitives
+        static if (!isInfinite!(Source))
+            @property bool empty()
+            {
+                static if (!isInfinite!(Sizes))
+                    return _source.empty || _chunkSizes.empty;
```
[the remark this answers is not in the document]
AsterMiha: I was thinking that if not enough sizes are provided then the rest of the elements are not needed. But it might be a better idea to just make a chunk out of all the remaining elements so that those are not simply lost.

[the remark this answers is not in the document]
MoonlightSentinel: That sounds like a more sensible approach and it also matches the existing `chunks` behaviour (which will return a smaller chunk if there are not enough remaining elements)

AsterMiha:
Should chunks of size 0 be allowed? At the moment, they can be added but when popBack removes a non empty chunk next to an empty one that one dissappears as well.

wilzbach (commented):
Hi, thanks a lot for your PR!
I added a few smaller comments, here are a few high-level points
- [x] changelog entry
- [x] range operations should be in O(1)
- [x] the documentation is hard to read and I would recommend to use a better name than "sizes range"
- [x] new functions should use Voldemort types
- [ ] new functions should be fully covered (see e.g. https://codecov.io/gh/dlang/phobos/compare/bf258e56f2dbc28cf810e9329a0696afbe114907...fc330a87850596023983b0627486772a793b1e3f/diff)
- [x] ddoc build fails (DAuotTest), because the new public `chunks` overload is undocumented. This should be fixed when you move the documentation to `chunks` and set `Chunks` to `private`.

std/range/package.d
```diff
@@ -8606,6 +8606,380 @@ if (isForwardRange!Source && hasLength!Source)
...
+            }
+
+        static if (hasSlicing!Source && (isInfinite!Source || hasLength!(Source)))
+        {
+            auto opIndex(size_t index)
+            {
+                size_t start = 0;
+                Sizes sizesCopy = _chunkSizes.save();
+
+                while (index && !sizesCopy.empty)
```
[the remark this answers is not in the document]
AsterMiha: For opIndex and opSlice all of the first index lengths are needed to get the right slice from the source range so I can't see how this could become O(1). However I can change the behaviour a bit so length, back and popBack become faster: when source doesn't have enough elements, instead of ignoring the last given lengths an empty chunk will be formed. This way the range will have the same length as chunkSizes so instead of changing the source at popBack I can just remove the last length. Would that be alright? Or would it be better to just remove these operations and keep the current behaviour?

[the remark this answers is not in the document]
wilzbach: I would (personally) remove these operations as it can have undesirable side-effects as it's easy to loose the overhead cost in generic algorithms.
However, as there are two sides to this argument and I would recommend to open a discussion in the general NG (aka Forum) and ask for opinions on this issue.
(Maybe there are other historic precedents, but IIRC so far everything in std.range does O(1) operations.)

wilzbach:
> Should chunks of size 0 be allowed? At the moment, they can be added but when popBack removes a non empty chunk next to an empty one that one dissappears as well.

I don't see a valid reason to use them. So I think it's better to be conservative and disallow as it's easier to enable a feature when asked for them disable it. The documentation should mention it though.

schveiguy:
Regarding the opIndex controversy: O(1) is expected, but I think anything sub-linear should be acceptable. However, in this case, this is not sub-linear. I'm not sure that the addition of opIndex is worth it (one can use popFrontN for lower-performance indexing), but it's not a clear cut decision either.

Consider that something like `chain(a[0 .. 4], a[4 .. $])` is going to allow opIndex, and will perform EXACTLY as well as `a.chunks(only(4, a.length-4))` does when indexing. The complexity (technically) is O(x) where x is the number of ranges in the chain, or the number of elements in the chunk size list. But with chain, of course, the number of chained ranges is a constant. So does that mean we can get away with it?

Another possibility here is to take the indexes that end each chunk instead of the sizes of each chunk, which would provide O(1) indexing. In other words, someRange.chunked(only(1, 2, 3)) becomes someRange.chunked(only(1, 3, 6)). One could still use the original concept via:

```D
auto r = somerange.chunked(sizes.cumulativeFold!("a + b")); // no opIndex
auto r = somerange.chunked(sizes.cumulativeFold!("a + b").array); // opIndex support, but allocates
```

You would have to assert that the "index" range is strictly increasing. Also, obviously, ranges that don't support random access cannot support opIndex when chunked this way.

AndrejMitrovic:
Why does the API take a range of indices, instead of a range of lengths? `chunks` itself takes a length, it seems odd to add a variable-sized overload but now the user needs to manually calculate the indices. It makes for an awkward API.

I would expect something like this to work:

```D
int[] arr = [1, 2, 3, 4, 5, 6, 7, 8];
assert(array(chunks(source, 4) == [[1, 2, 3, 4], [5, 6, 7, 8]]);
assert(array(chunks(source, [1, 2, 3])) == [[1], [2, 3], [4, 5, 6]]);
assert(array(chunks(source, 2.repeat(3))) == [1, 2], [3, 4], [5, 6]));
```

See how well it composes with other range functions? This would be very hard to replicate if we're forced to use indices instead.

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+
+            /// Ditto
+            @property auto front()
+            {
+                assert(!empty, "Attempting to fetch the front of an empty range");
+
+                if (_endIndexes.empty && !_source.empty)
+                    return _source.save;
+
+                assert(_endIndexes.front != 0, "Attempting to create a chunk of length 0");
```
schveiguy, line 8668: This assert doesn't take into account the offset. In any case, I don't actually see a problem with having 0 length chunks when the sizes are variable. Really, you should assert that `_endIndexes.front >= _offset`

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+                assert(!empty, "Attempting to popFront an empty range");
+
+                // source range is longer than index
+                if (_endIndexes.empty && !_source.empty)
+                {
+                    _isEmpty = true;
+                    return;
+                }
+
+                _source.popFrontN(_endIndexes.front()-_offset);
```
schveiguy, line 8685: protect this line with `assert(_endIndexes.front >= _offset)`. Also, remove the parentheses for front

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+            @property auto front()
+            {
+                assert(!empty, "Attempting to fetch the front of an empty range");
+
+                if (_endIndexes.empty && !_source.empty)
+                    return _source.save;
+
+                assert(_endIndexes.front != 0, "Attempting to create a chunk of length 0");
+
+                return _source.save.take(_endIndexes.front()-_offset);
```
schveiguy, line 8670: remove parentheses from `front` call. A range can have a non-function front member.

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+
+                // source range is longer than index
+                if (_endIndexes.empty && !_source.empty)
+                {
+                    _isEmpty = true;
+                    return;
+                }
+
+                _source.popFrontN(_endIndexes.front()-_offset);
+                _offset = _endIndexes.front();
```
schveiguy, line 8686: Remove front parantheses.

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+            {
+                return typeof(this)(_source, _endIndexes, _offset);
+            }
+
+            static if (isInfinite!Source && hasLength!Indexes)
+            /// Length is only provided for infinite sources, when hasLength!Indexes is true
+                @property size_t length()
+                {
+                    return _endIndexes.length + 1;
+                }
```
schveiguy, line 8711: length should be provided for `hasLength!Indexes && hasLength!Source && isBidirectionalRange!Indexes`

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+                {
+                    assert(index<=_endIndexes.length, "Chunks index out of bounds");
+
+                    const size_t startIndex = (index == 0 ? 0 : _endIndexes[index-1] - _offset);
+
+                    static if (hasLength!Source)
+                        assert(startIndex<=_source.length, "Chunks index out of bounds");
+
+                    if (index == _endIndexes.length)
+                        return _source[startIndex .. $];
```
schveiguy, line 8729: This is tricky, because if your indexes end exactly at the end of the source, then opIndex is providing an empty range at the end of the range, when normal iteration would not.

I'm wondering if a private `_hasTail` property, which is true when there are extra elements past the last index are present, could be helpful for these corner cases.

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+
+                    static if (hasLength!Source)
```
schveiguy, line 8731: Took me a bit to realize that this can only be false if Source is infinite. Maybe add a comment.

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+                    else
+                        const size_t endIndex = _endIndexes[index] - _offset;
+
+                    return _source[startIndex .. endIndex];
+                }
+
+                /// Ditto
+                auto opSlice(size_t lower, size_t upper)
+                {
+                    assert(lower <= upper, "Start index must be lower than end index");
```
schveiguy, line 8745: "lower than or equal to"

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+                    assert(lower <= _endIndexes.length+1 && upper <= _endIndexes.length+1,
+                        "Chunks index out of bounds");
```
schveiguy, line 8747: You should have a `length` member always in this set of circumstances (I think), and check that instead. See trickiness I cited above.

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+
+                    if (lower == upper || lower == _endIndexes.length+1)
+                        return Chunks(_source[$ .. $], _endIndexes[$ .. $]);
```
schveiguy, line 8750: I think remove this part. The rest should take care of the empty case on its own.

schveiguy:
Because using lengths means indexing features are not available. You can always use sizes instead, by making a range that builds the indexes instead (see my comment above).

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+                        const size_t endIndex = min(_endIndexes[upper-1] - _offset, source.length);
+                    }
+                    else
+                        const size_t endIndex = _endIndexes[upper-1] - _offset;
+
+                    return Chunks(_source[startIndex .. endIndex], _endIndexes[lower .. upper-1], startIndex);
+                }
+            }
+
+            // Used when slicing to avoid taking extra elements
```
schveiguy, line 8770: I'm not understanding why this is needed. Can you give an example?

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+
+            void popFront()
+            {
+                if (impl.endIndexes.empty)
+                {
+                    tookLastChunk = true;
+                    return;
+                }
+
+                impl.curSizeLeft -= impl.source.popFrontN(impl.curSizeLeft) - offset;
```
schveiguy, line 8847: I think the subtraction of offset is not correct here. Doesn't `curSizeLeft` denote the amount left in the chunk?

std/range/package.d
```diff
@@ -8606,6 +8606,385 @@ if (isForwardRange!Source && hasLength!Source)
...
+                if (!impl.source.empty && !impl.endIndexes.empty)
+                {
+                    offset = impl.endIndexes.front;
```
schveiguy, line 8850: This is the only place offset is set. It's only used (IMO) erroneously above, and just below. I think it probably doesn't need to be a member of the struct.

schveiguy (changes requested):
Much nicer I think with indexes vs. lengths. More possibilities.

std/range/package.d
schveiguy: Actually, I can't see any reason to reassign to impl.curSizeLeft. If popFrontN succeeds in popping off N elements, it will be set to 0. If it doesn't pop that many off, then the source is empty, and curSizeLeft is irrelevant.

std/range/package.d
AsterMiha: I was thinking of this case: when the source range is longer than the last index provided and the requested slice should end just before the chunk containing the extra elements:
source = [1, 2, 3, 4, 5]
indexes = [1, 2, 3]
chunks[1 .. 3] should be [[2], [3]]
But since the whole range is supposed to also include one last chunk with the extra items, when returning the new range without specifying the needed length chunks[1 .. 3] would turn into [[2], [3], [4, 5]]. This should only be a problem when hasLength!Source and isInfinite!Source are both false since in this case we can't check if the end index is still in range. Actually it might be safer to only allow indexing in these two cases.

schveiguy: Ah, ok, I get it. I think all you need is a boolean saying whether the tail should be returned.

schveiguy: Please include this example in the unittests (if not already).

AndrejMitrovic:
@schveiguy 

> Because using sizes means indexing features are not available. You can always use sizes instead, by making a range that builds the indexes from your sizes (see my comment above).

Alright, that works. But could we make it an overload (perhaps with a different name, not an overload) to take sizes? It's not going to be obvious at the call site what `cumulativeFold` does.

schveiguy:
> But could we make it an overload

Yeah, I'm actually thinking a compile-time parameter flag that indicates the parameters are sizes or indexes (defaulted to sizes, because that is what most will expect).

RazvanN7:
ping @AsterMiha

RazvanN7:
@AsterMiha Any chance on taking this to the finish line?

AsterMiha:
> @AsterMiha Any chance on taking this to the finish line?

I'll work on it next week. Thanks for reminding me!

std/range/package.d
```diff
@@ -8578,6 +8578,503 @@ if (isForwardRange!Source && hasLength!Source)
...
+        }
+    }
+
+    return Chunks(source, partitioners);
+}
+
+/// The Partitioners range contains the desired lengths of the chunks
+struct ByLength;
+/// The Partitioners range contains the end indexes of each chunk
+struct ByIndex;
```
RazvanN7, line 8846: Maybe these should documented enum values? (currently the style tester sees them as undocumented/untested public symbols)

std/range/package.d
```diff
@@ -8578,6 +8578,517 @@ if (isForwardRange!Source && hasLength!Source)
...
+                }
+            }
+        }
+    }
+
+    return Chunks(source, partitioners);
+}
+
+/// The mode of partitioning a given range
+enum Partitioning {
```
RazvanN7, line 8846: I think that this enum should be put above the chunks declarations. That will probably get rid of the style checker error:

```
std/range/package.d(8607:6)[warn]: Public declaration 'chunks' has no documented example.
```

RazvanN7:
cc @atilaneves as this is a new addition to phobos.

std/range/package.d
AsterMiha: If I just move the enum the error will simply move.
I also added a short example to avoid that.

RazvanN7:
@atilaneves I have fixed the changelog. Is there anything else blocking this PR?

atilaneves:
> @atilaneves I have fixed the changelog. Is there anything else blocking this PR?

Seems to me there are many unaddressed comments.

RazvanN7:
Which ones are you referring to? I haven't seen any. As far as I can see the code comments have been resolved.

schveiguy:
Hm... I'd have to take another look. This has been a while, I have to try and remember what exactly is happening here. Can't at the moment, but later this week. Send me a ping if you don't hear from me.

RazvanN7:
ping @schveiguy

std/range/package.d
schveiguy: This seems moot at this point.

std/range/package.d
schveiguy: Moot.

std/range/package.d
schveiguy: Still not addressed. Would add that the mode has to be `ByIndex`.

std/range/package.d
```diff
@@ -8578,6 +8578,529 @@ if (isForwardRange!Source && hasLength!Source)
...
+                {
+                    _isEmpty = true;
+                    return;
+                }
+
+                assert(_partitioners.front > _offset);
+                _source.popFrontN(_partitioners.front - _offset);
+                static if (partitionType == Partitioning.ByIndex)
+                    _offset = _partitioners.front;
+                _partitioners.popFront();
```
schveiguy, line 8686: It's completely unclear to me how this takes into account the ByLength and ByIndex difference. Can you explain how this works? Does `_offset` stay 0 for the entire time if ByLength is used? If so, this is terrible design for saving a few static ifs (plus the compiler isn't going to be able to optimize out the subtraction).

I also don't understand why we seem to be forbidding zero-sized chunks.

schveiguy (commented):
Sorry to say I'm extremely confused about the semantics on this. I almost think this should be abandoned and rewritten.

Originally, the code was by index only, with a documented mechanism on how to use a range of lengths to generate the desired indexes.

In addition, the handling of running out of source or partitioning elements leads to lots of confusion.

Thinking of practical uses, you probably want to have a partitioning list that you may want to apply to various lengths of elements. This requires fuzzy borders, and what to do at the end isn't necessarily clear. Maybe the following set of rules:

1. zero length chunks should be allowed. (disallowing this is an unnecessary special case)
2. Indexes are all end points, with an assumed beginning index of 0 for the first chunk.
3. Indexes beyond the length will return the rest of the range. Indexes past that endpoint will not generate chunks.
3. Trailing data is not ever returned if the partitioning range ends early. One can always do `chain(endpoints, only(size_t.max))`
4. Length mode should be implemented as I described by wrapping with a cumulative_fold, and the semantics that follow from that.

If I can find some time, I might reimplement. But very busy at the moment.

PR #7604, opened 2020-08-22
Provide float and double versions of fmin and fmax. Improve docs.

andralex:
Per https://forum.dlang.org/post/ukdlpmrnwfddeofcgbot@forum.dlang.org.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -7228,43 +7228,63 @@ real fdim(real x, real y) @safe pure nothrow @nogc
 }
 
 /**
- * Returns the larger of x and y.
+ * Returns the larger of `x` and `y`.
  *
- * If one of the arguments is a NaN, the other is returned.
+ * If one of the arguments is a `NaN`, the other is returned.
+ *
+ * See_Also: $(REF max, std,algorithm,comparison) is faster because it does not perform the `isNaN` test.
  */
-real fmax(real x, real y) @safe pure nothrow @nogc
+F fmax(F)(const F x, const F y) @safe pure nothrow @nogc
+if (isFloatingPoint!F)
 {
-    return (y > x || isNaN(x)) ? y : x;
+    // Do the more predictable test first. Generates 0 branches with ldc and 1 branch with gdc.
+    // See https://godbolt.org/z/erxrW9
+    if (isNaN(y)) return x;
+    return x >= y ? x : y;
 }
 
 ///
 @safe pure nothrow @nogc unittest
 {
-    assert(fmax(0.0, 2.0) == 2.0);
-    assert(fmax(-2.0, 0.0) == 0.0);
-    assert(fmax(real.infinity, 2.0) == real.infinity);
-    assert(fmax(real.nan, 2.0) == 2.0);
-    assert(fmax(2.0, real.nan) == 2.0);
+    import std.meta : AliasSeq;
+    static foreach (F; AliasSeq!(float, double, real))
+    {
+        assert(fmax(F(0.0), F(2.0)) == 2.0);
+        assert(fmax(F(-2.0), 0.0) == F(0.0));
+        assert(fmax(F.infinity, F(2.0)) == F.infinity);
+        assert(fmax(F.nan, F(2.0)) == F(2.0));
+        assert(fmax(F(2.0), F.nan) == F(2.0));
+    }
 }
 
 /**
- * Returns the smaller of x and y.
+ * Returns the smaller of `x` and `y`.
  *
- * If one of the arguments is a NaN, the other is returned.
+ * If one of the arguments is a `NaN`, the other is returned.
+ *
+ * See_Also: $(REF min, std,algorithm,comparison) is faster because it does not perform the `isNaN` test.
  */
-real fmin(real x, real y) @safe pure nothrow @nogc
+F fmin(F)(const F x, const F y) @safe pure nothrow @nogc
+if (isFloatingPoint!F)
 {
-    return (y < x || isNaN(x)) ? y : x;
+    // Do the more predictable test first. Generates 0 branches with ldc and 1 branch with gdc.
+    // See https://godbolt.org/z/erxrW9
+    if (isNaN(y)) return x;
+    return x <= y ? x : y;
 }
 
 ///
 @safe pure nothrow @nogc unittest
 {
-    assert(fmin(0.0, 2.0) == 0.0);
-    assert(fmin(-2.0, 0.0) == -2.0);
-    assert(fmin(real.infinity, 2.0) == 2.0);
-    assert(fmin(real.nan, 2.0) == 2.0);
-    assert(fmin(2.0, real.nan) == 2.0);
+    import std.meta : AliasSeq;
+    static foreach (F; AliasSeq!(float, double, real))
+    {
+        assert(fmin(F(0.0), F(2.0)) == 0.0);
+        assert(fmin(F(-2.0), F(0.0)) == -2.0);
+        assert(fmin(F.infinity, F(2.0)) == 2.0);
+        assert(fmin(F.nan, F(2.0)) == 2.0);
+        assert(fmin(F(2.0), F.nan) == 2.0);
+    }
 }
 
 /**************************************
--- a/std/math.d
+++ b/std/math.d
@@ -7234,13 +7234,18 @@ real fdim(real x, real y) @safe pure nothrow @nogc
  *
  * See_Also: $(REF max, std,algorithm,comparison) is faster because it does not perform the `isNaN` test.
  */
-F fmax(F)(const F x, const F y) @safe pure nothrow @nogc
-if (isFloatingPoint!F)
+auto fmax(N1, N2)(const N1 x, const N2 y) @safe pure nothrow @nogc
+if ((__traits(isFloating, N1) || __traits(isFloating, N2)) && is(typeof(x < y)))
 {
+    import std.algorithm.comparison : max;
+    alias R = typeof(max(x, y));
     // Do the more predictable test first. Generates 0 branches with ldc and 1 branch with gdc.
     // See https://godbolt.org/z/erxrW9
-    if (isNaN(y)) return x;
-    return x >= y ? x : y;
+    static if (__traits(isFloating, N1))
+    {
+        if (isNaN(x)) return R(y);
+    }
+    return max(x, y);
 }
 
 ///
@@ -7264,13 +7269,18 @@ if (isFloatingPoint!F)
  *
  * See_Also: $(REF min, std,algorithm,comparison) is faster because it does not perform the `isNaN` test.
  */
-F fmin(F)(const F x, const F y) @safe pure nothrow @nogc
-if (isFloatingPoint!F)
+auto fmin(N1, N2)(const N1 x, const N2 y) @safe pure nothrow @nogc
+if ((__traits(isFloating, N1) || __traits(isFloating, N2)) && is(typeof(x < y)))
 {
+    import std.algorithm.comparison : min;
+    alias R = typeof(min(x, y));
     // Do the more predictable test first. Generates 0 branches with ldc and 1 branch with gdc.
     // See https://godbolt.org/z/erxrW9
-    if (isNaN(y)) return x;
-    return x <= y ? x : y;
+    static if (__traits(isFloating, N1))
+    {
+        if (isNaN(x)) return R(y);
+    }
+    return min(x, y);
 }
 
 ///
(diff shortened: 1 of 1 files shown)


kinke:
I very much prefer your first version, without `std.algorithm.comparison.max` - that's IMO unnecessary template bloat, see the codegen AST: https://run.dlang.io/is/6tIO0f - 53 lines vs. 226 lines.

PetarKirov:
Ping @andralex

andralex:
Well I kinda like it with using min and max because whatever improvements we bring to those would be reflected in fmin and fmax. One counterargument is there aren't many improvements that could possibly be brought to floating point comparison. One other counterargument was made about template bloat - I very much believe we should resist subscribing to a programming-by-fear-of-bloat methodology.

I'll make the change because it does at a level feel excessive to reuse a one-liner.

kinke:
> I very much believe we should resist subscribing to a programming-by-fear-of-bloat methodology.

Well I'd say the fear is clearly justified in this case, and min/max (and probably a lot more Phobos stuff) definitely need some work to account for the probably vast majority of trivial cases (2 operands of the same arithmetic type), the bloat is IMO embarrassing and doesn't exactly favor compiler performance (possibly something for @n8sh?). And as all of that would now have been instantiated with templated f{min,max} as opposed to the previous 'import' of a precompiled function in Phobos, this wouldn't have set a nice precedent for more templatizing of std.math.

Thanks, lgtm now.

kinke (approved):

andralex:
> > I very much believe we should resist subscribing to a programming-by-fear-of-bloat methodology.
> 
> Well I'd say the fear is clearly justified in this case, and min/max (and probably a lot more Phobos stuff) definitely need some work to account for the probably vast majority of trivial cases (2 operands of the same arithmetic type), the bloat is IMO embarrassing and doesn't exactly favor compiler performance (possibly something for @n8sh?). And as all of that would now have been instantiated with templated f{min,max} as opposed to the previous 'import' of a precompiled function in Phobos, this wouldn't have set a nice precedent for more templatizing of std.math.
> 
> Thanks, lgtm now.

There'd be two counterarguments to this.

Per measurements done my myself and others, the real pain in templates most often comes from machine-generated recursive instantiations, not the meager instantiations created by user code. (`ctRegex` would be a sort of an exception as even one little use will instantiate a bunch of stuff). So this meager thing with a handful of instances creates no trouble one way or another, no two ways about it.

Second, it's just a defeatist way to go about things, a bad case of premature optimization. I remember people were worried about template bloat in C++ since like 1994. And some folks wrote and promoted libraries built with fear of template bloat (I remember a bunch of container libraries with inheritance and virtuals, oh boy) and - well they're in the dustbin of history. Nobody gives a rat's tail. And then there were folks who used templates everywhere and went from strength to strength - the standard library/STL, Boost, Eigen, you name it - and now pretty much anybody in C++ can't write a dozen lines of code without using some template. And guess what, some people kept on saying that template bloat will end up being the undoing of C++, and coders and compilers just keep on getting better and everybody's exchanging high fives.

So I'm serious: don't code in fear of template bloat. It's a losing proposition.

andralex:
What's the matter with the SumType failure?

thewilsonator:
Unrelated, fails on multiple PRs.

andralex:
Thanks, Nick. I won't make the mistake of force merging  again though :o).

wilzbach:
> Thanks, Nick. I won't make the mistake of force merging  again though :o).

Well, someone else did it this time - https://github.com/dlang/phobos/pull/7611 ;-)
We should really remove the ability to do that...
Anyhow, Buildkite should be green soon.

kinke:
> Per measurements done my myself and others, ...

Some hard numbers with LDC v1.23 on Win64 for `ldc2 -c -betterC [-d-version=Simple]` and
```D
F fmin(F)(const F x, const F y) @safe pure nothrow @nogc
if (__traits(isFloating, F))
{
    import std.math : isNaN;
    if (isNaN(x)) return y;
    version (Simple)
    {
        return y < x ? y : x;
    }
    else
    {
        return min(x, y);
    }
}

alias T = float;
T foo(T a, T b) { return fmin(a, b); }
```

~80 ms (`Simple`) vs. ~100 ms. When putting `isNaN()` in the same module instead of importing `std.math`, the same ~20ms difference (~60ms vs. ~80ms).

andralex:
@kinke thanks for continuing what I think is an important discussion. So important, in fact, that I sat down and reproduced the measurements to get and share more insight into the matter. Hopefully we will be able to debunk a couple of myths together.

(To be clear, the discussion in this PR is settled; I agree it's excessive to import a module just for a one-liner.)

The numbers you posted consolidate two distinct overheads that would be good to distinguish. They are:

* the overhead of importing a module; and
* the overhead of instantiating a template from that module.

My baseline for the simple version is 39 ms (as `time` measures) for a warm build (repeat the compilation a few times, then measure). The version with the import clocks at 49 ms, i.e. clearly measurable.

It turns out, as one would probably expect, that in this case virtually the entire overhead is the sheer import of the module. To check that hypothesis, I reworked the code as follows:

```
F fmin(F)(const F x, const F y) @safe pure nothrow @nogc
if (__traits(isFloating, F))
{
    import std.math : isNaN;
    if (isNaN(x)) return y;
    version (Simple)
    {
        return y < x ? y : x;
    }
    else
    {
        return y < x ? y : x;
    }
}

alias T = float;
T foo(T a, T b) { return fmin(a, b); }
```

i.e. the non-simple version does the import but uses nothing from it. Time: 49 ms, virtually indistinguishable from the version that instantiates templates.

So importing a module has overhead. But it is understood and scales well, and that's why we don't go around and tell people to avoid imports and copypaste code instead. Within the context of actually using `std.math`, I didn't bother to measure but I think it's pretty obvious you won't be able to measure a percentage difference whether you import `std.algorithm.comparison` or not. There are already 5 top-level imports and a lot more local imports (among which `std.algorithm.comparison` itself). So, not a scaling problem.

I should add numbers for memory consumption (as `/bin/time` measures). Here they are, lowest of 5 consecutive runs:

* simplest: 62852 KB
* import but do not instantiate: 69216 KB
* import and instantiate: 69544 KB

So the overhead of instantiating straightforward templates is hardly a blip on the radar. Of course we know that scales poorly for recursive templates. This all seems to support the basic point: we should not tell people to not use templates, just the same as we should not tell people not to import modules.

kinke:
You're right, and I was aware of the obvious import factor and maybe should have emphasized that it's not just about the template bloat, but the potential extra import (when using `fmin` alone, it is indeed an extra import and causes ~20 ms overhead on my box, after adapting fmin in actual `std.math`). The bloat alone seems hardly measurable in this case, too much measuring noise.

> This all seems to support the basic point: we should not tell people to not use templates, just the same as we should not tell people not to import modules.

Just to be clear, I've never suggested avoiding the min/max template per se, but tried to show that it's not optimized for the trivial case of 2 operands with the same arithmetic type, instantiating a lot of other extra templates, that's the bloat I was getting at.

andralex:
@kinke yes, thanks very much for these insights!

Independently of all (and ironically): I looked through this module and it looks like a lot of the functions (`sin`, `cos` etc) follow the three-overloads pattern, not the templated functio pattern. Should we go with this instead of the current code for uniformity?

```d
static foreach (F; AliasSeq!(float, double, real))
    F fmin(F x) { ... }
static foreach (F; AliasSeq!(float, double, real))
    F fmax(F x) { ... }
```

instead of what we have right now? That would be more in keeping with the other functions, and we can use this pattern in the entire module. Cons:

- more work
- not sure what documentation will look like
- unclear whether there's a real advantage

Pros:

- we can use this instead of the current troikas everywhere in the module
- there is a nice minimalism about the approach

kinke:
The biggest problem with the 3 explicit overloads I've previously preferred is the overload resolution wrt. integers (some common literals like `0`, `1`, `-1` etc.) mentioned earlier on (but only applying to non-unary functions); working around that is IMO the biggest advantage of the single template, besides less (verbose) code. I think this template approach here sets a nice precedent, and functions currently still only accepting `real` should be modified accordingly, and later also the existing troikas, at least for small functions. The only catch is probably the template constraint, as `__traits(isFloatingPoint)` includes floating-point vectors too, so something like `is(F == float) || is(F == double) || is(F == real)` might be better suited in other instances.

n8sh (approved):

andralex:
Is the failure in `SumType` related? There are warnings about unreachable code in here:

```d
@safe unittest {
	import std.algorithm.mutation: move;

	static struct NoCopy
	{
		@disable this(this);
	}

	alias MySum = SumType!NoCopy;

	NoCopy lval = NoCopy();

	MySum x = NoCopy();
	MySum y = NoCopy();

	assert(__traits(compiles, SumType!NoCopy(NoCopy())));
	assert(!__traits(compiles, SumType!NoCopy(lval)));

	assert(__traits(compiles, y = NoCopy()));
	assert(__traits(compiles, y = move(x))); // UNREACHABLE
	assert(!__traits(compiles, y = lval)); // UNREACHABLE
	assert(!__traits(compiles, y = x)); // UNREACHABLE

	assert(__traits(compiles, x == y)); // UNREACHABLE
}
```

andralex:
CC @pbackus

pbackus:
Can't reproduce locally and buildkite is green now, so I guess there's no problem?

Geod24:
The failure was due to a regression introduced in https://github.com/dlang/phobos/pull/7609
Well technically it was an unintentional bugfix which uncovered an issue in SumType, which @pbackus [fixed](https://github.com/pbackus/sumtype/commit/940d9052eb6bc14ea37d1196de5e9aab30eece59).

UplinkCoder:
Caused a regression in our code.
@andralex you changed the API.

andralex:
Yah, the breakage is in this code:

```d
	static foreach(f;AliasSeq!( 	fabs,		sqrt,		sin,		cos,		tan,		asin,
					acos,		atan,		sinh,		cosh,		tanh,		asinh,
					acosh,		atanh, 		/*log*/		log2,		log10,		logb,
					log1p,		exp,		exp2,		expm1,		ceil,		floor,
					round,		lround,		trunc,		rint,		lrint,		nearbyint,
					rndtol,		gamma,		logGamma,	sgnGamma,	digamma,	logmdigamma,
					logmdigammaInverse,		erf,		erfc,		normalDistribution,
					normalDistributionInverse,	atan2,		fmod,		remainder,	isIdentical,
					fmax,		fmin,		gammaIncomplete, gammaIncompleteCompl,		gammaIncompleteComplInverse,
	))
		handlers.registerHandler!(doubleOverloadOf!f);
```

The code attempts to pick up the address of each function, including `fmin` and `fmax`, which causes compilation to fail. If we introduce three overloads as per https://github.com/dlang/phobos/pull/7604#issuecomment-683472216, I just tested and the build works. However, @kinke mentioned (https://github.com/dlang/phobos/pull/7604#issuecomment-683493900) that the three overloads introduce other breakages wrt literals.

So that leaves us between a rock and hard place. Insights welcome.

Somewhat independently of that, we can fix the problem on Symmetry's side. The code is very fragile at the moment and makes assumptions it oughtn't:

```d
// this assumes that exactly one double overload actually exists and is the first arg...
// safe for std.math but not in general.
private template doubleOverloadOfHelper(alias what) {
	static foreach(overload; __traits(getOverloads, __traits(parent, what), __traits(identifier, what))) {
		static if(is(Parameters!(overload)[0] == double)) {
			//pragma(msg, "****************", __traits(identifier, what));
			alias doubleOverloadOfHelper = overload;
		}
	}
}
// so this extra wrapper will fallback to old behavior if there isn't a double overload
private template doubleOverloadOf(alias what) {
	static if(!is(typeof(doubleOverloadOfHelper!what) ==  void))
		alias doubleOverloadOf = doubleOverloadOfHelper!what;
	else
		alias doubleOverloadOf = what; // best we can do for now
}
```

wilzbach:
> So that leaves us between a rock and hard place. Insights welcome.

Considering that all the projects on Buildkite passed (it shows that there likely not going to be many projects affected) and you have to fix the Symmetry code anyhow, I would just do that and move on.

UplinkCoder:
@wilzbach yes, that seems okay.
though I'd still like to get our SIL project on the buildkite soonish.

wilzbach:
> @wilzbach yes, that seems okay.
> though I'd still like to get our SIL project on the buildkite soonish.

Nobody is preventing you from doing that. Though as I mentioned, you don't want it to be main Buildkite project, because logs would be public.

PR #7608, opened 2020-08-28
std.typecons : AutoImplement: Handle 'in' storage class

Geod24:
This became a problem with the change to make `in` its own storage class, but luckily no release has been made yet.


thewilsonator (approved):

thewilsonator:
Tests?

Geod24:
> Tests?

Well it requires `-preview=in` to be noticeable, otherwise it'll just work, and we don't have any `-preview=in` tests here (but working on one).

PR #7609, opened 2020-08-28
std.traits: Adapt unittest to pass with `-preview=in`

Geod24:
I'm using https://github.com/dlang/dmd/pull/11632 as a way to adapt projects.
With this PR, it should pass for Phobos and the documentation tester.
I should probably add it to Buildkite like what Steven did for auto-decoding 🤔

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -5219,11 +5219,26 @@ private enum isLvalueAssignable(Lhs, Rhs = Lhs) = __traits(compiles, { lvalueOf!
     static assert( isAssignable!(S4, immutable int));
 
     struct S5 { @disable this(); @disable this(this); }
-    struct S6 { void opAssign(in ref S5); }
-    static assert(!isAssignable!(S6, S5));
-    static assert(!isRvalueAssignable!(S6, S5));
-    static assert( isLvalueAssignable!(S6, S5));
-    static assert( isLvalueAssignable!(S6, immutable S5));
+
+    // `-preview=in` is enabled
+    static if (!is(typeof(mixin(q{(in ref int a) => a}))))
+    {
+        struct S6 { void opAssign(in S5); }
+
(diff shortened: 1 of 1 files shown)


Geod24:
CC @wilzbach , this is what motivated dlang/dmd#11635 . As you can see the workaround isn't terrible either.

std/traits.d
```diff
@@ -5219,11 +5219,26 @@ private enum isLvalueAssignable(Lhs, Rhs = Lhs) = __traits(compiles, lvalueOf!Lh
...
+        struct S6 { void opAssign(in S5); }
+
+        static assert(isRvalueAssignable!(S6, S5));
+        static assert(isLvalueAssignable!(S6, S5));
+        static assert(isAssignable!(S6, S5));
+        static assert(isAssignable!(S6, immutable S5));
+    }
+    else
+    {
+        mixin(q{ struct S6 { void opAssign(in ref S5); } });
```
MoonlightSentinel, line 5235: Why is this a mixin?

Geod24: The error with `-preview=in` is in the parser (where "redundant storage class" checks are).
And yes, I hate it too and I'm having second thoughts, but it was the most consistent place to have it.

MoonlightSentinel:
> I should probably add it to Buildkite like what Steven did for auto-decoding 🤔

Good idea, except implementing it for `stable` first ^^

wilzbach:
> this is what motivated dlang/dmd#11635 .

Yeah, I'm well aware of the problem. We unfortunately did similar workarounds in the past, e.g.

https://github.com/dlang/phobos/blob/6194b172955571ea3eb87bc8380d413dd7694598/std/typecons.d#L7356-L7362

thewilsonator (approved):

PR #7611, opened 2020-08-30
Revert "std.traits: Adapt unittest to pass with `-preview=in`"

wilzbach:
Reverts dlang/phobos#7609

Checking whether this PR broke the Buildkite sumtype build.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -5180,10 +5180,10 @@ enum isAssignable(Lhs, Rhs = Lhs) = isRvalueAssignable!(Lhs, Rhs) && isLvalueAss
 }
 
 // ditto
-private enum isRvalueAssignable(Lhs, Rhs = Lhs) = __traits(compiles, { lvalueOf!Lhs = rvalueOf!Rhs; });
+private enum isRvalueAssignable(Lhs, Rhs = Lhs) = __traits(compiles, lvalueOf!Lhs = rvalueOf!Rhs);
 
 // ditto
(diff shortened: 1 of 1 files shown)


wilzbach:
This fixes the Buildkite failure.

thewilsonator (approved):

Geod24:
Ergh, sorry about this. Looks like my first commit was not as innocent as I thought, and the bug I encounter might not be `in`-specific. Which is a relief, in one way.
Thanks for the revert @wilzbach , and I'll investigate what went wrong.

Geod24:
Narrowed it down to https://issues.dlang.org/show_bug.cgi?id=21210
CC @pbackus , could you fix `sumtype` so we can fix this bug ?

pbackus:
@Geod24 Would it be possible to make `isRvalueAssignable` public so that I can fix this without duplicating Phobos code?

wilzbach:
Force-merged as the two Windows machines on the auto-tester seem to be missing in action :/

Geod24:
@pbackus : That's not going to be of much help, since your code wouldn't compile with older DMD / LDC versions.
In fact, it wouldn't compile at all for the time being, until a new version is released.
The duplication should only be a few LoC though.

pbackus:
@Geod24 I understand that. I'm okay with adding my own copy of `isRvalueAssignable` as a temporary workaround; I just want to be sure that I'll be able to delete it in a future `sumtype` release.

pbackus:
@Geod24 Fixed the issue in `sumtype` and tagged a new release. Tests pass locally with phobos commit aca1e1208fc7f914b44e44f07bee836c7eeb941a.

Geod24:
Follow up #7612

PR #7612, opened 2020-08-30
Fix 21210:  std.traits : isAssignable false positive on disabled copy struct

Geod24:
```
`isAssignable` would previously return `true` for non-copyable types,
even though code that tried to use an lvalue would not compile.
This behavior was originally found when implementing `-preview=in`.

With the new -preview=in check, the const-folding seemed to be a bit
too aggressive when an rvalue is passed, meaning that the check might
fail (probably due to the code that initialize the temporary).
```

The failure with SumType has been fixed.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -5180,10 +5180,10 @@ enum isAssignable(Lhs, Rhs = Lhs) = isRvalueAssignable!(Lhs, Rhs) && isLvalueAss
 }
 
 // ditto
-private enum isRvalueAssignable(Lhs, Rhs = Lhs) = __traits(compiles, lvalueOf!Lhs = rvalueOf!Rhs);
+private enum isRvalueAssignable(Lhs, Rhs = Lhs) = __traits(compiles, { lvalueOf!Lhs = rvalueOf!Rhs; });
 
 // ditto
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -5180,10 +5180,10 @@ enum isAssignable(Lhs, Rhs = Lhs) = isRvalueAssignable!(Lhs, Rhs) && isLvalueAss
 }
 
 // ditto
-private enum isRvalueAssignable(Lhs, Rhs = Lhs) = __traits(compiles, lvalueOf!Lhs = rvalueOf!Rhs);
+private enum isRvalueAssignable(Lhs, Rhs = Lhs) = __traits(compiles, { lvalueOf!Lhs = rvalueOf!Rhs; });
```
MoonlightSentinel, line 5183: Isn't this a more of a workaround for a compiler issue?

Geod24: I'm conflicted about this, because I'm not exactly sure of what code path is taken within the compiler.
However, I can see why this behavior (CTFE-ing the expression) would make sense. But it's also a fatal flaw with `__traits(compiles, ...)`, which is not shared by `is(typeof(...))`. On a related not, I never understood why we added the former if we had the latter.

wilzbach (approved):
Fine by me (it's not the first "workaround" in Phobos and won't be the last)

thewilsonator (approved):

PR #7613, opened 2020-08-30
Disable Windows builds on auto-tester

wilzbach:
See https://github.com/dlang/dmd/pull/11653 for details.

### DIFF
--- a/win32.mak
+++ b/win32.mak
@@ -535,11 +535,13 @@ install: phobos.zip
 	+rd/s/q $(DIR)\src\phobos
 	unzip -o phobos.zip -d $(DIR)\src\phobos
 
-auto-tester-build: targets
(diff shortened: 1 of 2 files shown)


MoonlightSentinel:
Shouldn't these jobs be transferred to Azure as well?

wilzbach:
> Shouldn't these jobs be transferred to Azure as well?

There are already Azure jobs here which run the build and unittests.
However, we don't need to add two additional bootstrap jobs here as Phobos can only ever be compiled with current master (same for druntime).

MoonlightSentinel (approved):
Right, forgot that there are no bootstrapping guarantees for druntime/phobos

thewilsonator (approved):

PR #7614, opened 2020-08-31
std.traits: Adapt unittest to pass with `-preview=in`

Geod24:
Redo of #7609 after it was reverted because a library relied on a bug that has since been fixed (revert: #7611 , bugfix: #7612 ).

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -5219,14 +5219,28 @@ private enum isLvalueAssignable(Lhs, Rhs = Lhs) = __traits(compiles, { lvalueOf!
     static assert( isAssignable!(S4, immutable int));
 
     struct S5 { @disable this(); @disable this(this); }
-    struct S6 { void opAssign(in ref S5); }
-    static assert(!isAssignable!(S6, S5));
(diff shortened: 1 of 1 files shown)


wilzbach (approved):
Let's wait for Buildkite this time :)

Geod24:
The commit that broke Buildkite was in #7612 . But yeah, no problem. Also, we should really fix the bug, especially since we're moving stuff away from the auto-tester... And/or make some more checks required ?

wilzbach:
> Also, we should really fix the bug, especially since we're moving stuff away from the auto-tester... 

It wasn't a bug here. It was merged manually. The bug only applies to GitHub actions as they aren't returned by the traditional CI APi.

> And/or make some more checks required ?

Yes, that's the one. Yesterday I set more checks as required on DMD and I think we just need to do the same here. Considering our history IMHO all CIs should be set as required.

Geod24:
> It wasn't a bug here. It was merged manually. The bug only applies to GitHub actions as they aren't returned by the traditional CI APi.

Also Azure (https://github.com/dlang/dlang-bot/issues/250).

PR #7615, opened 2020-09-04
Enable -preview=dtorfields for phobos

MoonlightSentinel:
Let's see what breaks.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -139,7 +139,7 @@ endif
 
 # Set DFLAGS
 DFLAGS=
-override DFLAGS+=-conf= -I$(DRUNTIME_PATH)/import $(DMDEXTRAFLAGS) -w -de -preview=dip1000 $(MODEL_FLAG) $(PIC) -transition=complex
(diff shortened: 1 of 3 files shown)


wilzbach (approved):

Geod24:
@MoonlightSentinel : So there's no risk of ABI mismatch if the client doesn't use this switch ?

MoonlightSentinel:
Don't think so, `-preview=dtorfields` just adds additional code in the constructor to destroy the instance which was previously  ignored when throwing an exception.
Unless I'm missing something(?)

UplinkCoder:
This appears to have broken something ?
https://issues.dlang.org/show_bug.cgi?id=21250

UplinkCoder:
@MoonlightSentinel please revert.
on stable.
@MartinNowak We need a patch release.

UplinkCoder:
This is a timebomb.
Regardless of whether you fix that instance in phobos.
We don't know what else it breaks.
Defer to next release!

wilzbach:
It's already released.

UplinkCoder:
Yes indeed.
We need a patch release asap.

PR #7616, opened 2020-09-05
Remove deprecated -preview=markdown flag

MoonlightSentinel:
The preview is now enabled by default.

CC @wilzbach


wilzbach (approved):

wilzbach:
If you're trying to deprecate the flag, IIRC it's also used by dlang.org

PR #7617, opened 2020-09-05
Fix issue 21228 - Templated toString silently ignored when it cant be…

Biotronic:
… instantiated

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -4423,6 +4423,24 @@ if ((is(T == struct) || is(T == union)) && (hasToString!(T, Char) || !is(Builtin
     static if (__traits(hasMember, T, "toString") && isSomeFunction!(val.toString))
         static assert(!__traits(isDisabled, T.toString), T.stringof ~
             " cannot be formatted because its `toString` is marked with `@disable`");
+    static if (__traits(hasMember, T, "toString") && !hasToString!(T, Char))
+    {
+        pragma(msg, T.stringof ~
+            " cannot be formatted because its `toString` cannot be instantiated.");
+        pragma(msg, T.stringof ~ ".toString is defined here:");
+        static foreach (ol; __traits(getOverloads, T, "toString", true))
+        {
+            static if (__traits(compiles, __traits(getLocation, ol)))
+            {
+                pragma(msg, __traits(getLocation, ol)[0], "(", __traits(getLocation, ol)[1], "): ",
+                    AliasSeq!(ol).stringof[6..$-1]);
+            }
+            else
+            {
+                pragma(msg, "(unknown location): ", AliasSeq!(ol).stringof[6..$-1]);
+            }
+        }
+    }
 
     enforceValidFormatSpec!(T, Char)(f);
     static if (hasToString!(T, Char))


wilzbach:
`hasToString` already works with templates partially, e.g. the first test case is:

```d
static struct A
{
    void toString(Writer)(ref Writer w)
    if (isOutputRange!(Writer, string))
    {}
}
```

Hence, I argue that we should be able to improve the trait to instantiate argument-less templates.

wilzbach (approved):

std/format.d
```diff
@@ -4423,6 +4423,9 @@ if ((is(T == struct) || is(T == union)) && (hasToString!(T, Char) || !is(Builtin
     static if (__traits(hasMember, T, "toString") && isSomeFunction!(val.toString))
         static assert(!__traits(isDisabled, T.toString), T.stringof ~
             " cannot be formatted because its `toString` is marked with `@disable`");
+    static if (__traits(hasMember, T, "toString") && !hasToString!(T, Char))
+        pragma(msg, T.stringof ~
+            " cannot be formatted because its `toString` cannot be instantiated.");
```
WebDrake, line 4428: Is there any possibility of getting a more informative error message here -- i.e. one that explains _why_ `toString` cannot be instantiated?  It would be sufficient to pass on the error message that would be produced if the problematic code were being compiled directly.

std/format.d
```diff
@@ -4423,6 +4423,9 @@ if ((is(T == struct) || is(T == union)) && (hasToString!(T, Char) || !is(Builtin
     static if (__traits(hasMember, T, "toString") && isSomeFunction!(val.toString))
         static assert(!__traits(isDisabled, T.toString), T.stringof ~
             " cannot be formatted because its `toString` is marked with `@disable`");
+    static if (__traits(hasMember, T, "toString") && !hasToString!(T, Char))
```
WebDrake, line 4426: Will this identify the bug in the case where `T` has an additional, valid, `toString` implementation?  Example:

```D
struct S
{
    string toString()
    {
        // this is valid
    }

    void toString (Output) (ref Output output)
        if (isOutputRange!(Output, char))
    {
        // fails to instantiate because we didn't import `isOutputRange`
    }
}
```

In this case we'd still like the bug in the invalid second implementation to be flagged by the compiler.  Will it be?

WebDrake (commented):
Thanks for the quick fix.  I have a query on whether the current version will catch all the potential problem cases -- example below.

WebDrake:
Oh, and a minor note: there's a typo in the commit title: should be `can't`, not `cant` (it was a typo in my original issue title:-)

std/format.d
WebDrake: > Will this identify the bug in the case where T has an additional, valid, toString implementation?

Part of the reason I'm concerned about this is about predictability of behaviour: if one writes an output-range-based `toString` to allow formatting without generating a new `string` instance, but the `string toString()` is silently used under the hood instead, this may create hard-to-track-down problems for apps that care about avoiding garbage generation.

std/format.d
Biotronic: Afraid not - there's [so many different ways to call `toString`](https://github.com/dlang/phobos/blob/master/std/format.d#L4024), and very hard (possibly no way) to tell which overloads are implemented and failed to instantiate. This also means we can't call `toString` to get the correct error messages.

std/format.d
Biotronic: It will not. We could special-case the existence of a non-templated overload with templated overloads that fail to instantiate, but that will complicate the code significantly, and I'm not sure there are no cases where that'd be the wrong thing to do.

Biotronic:
Had to remove the test, since 1) it wasn't testing the right thing (oops), and 2) this is just a warning message, which can't properly be tested for in unittest blocks.

RazvanN7 (approved):

berni44:
@RazvanN7 The autotester seems to be stalled here. Is there anything, that could be done?

RazvanN7:
I closed and reopened the PR. That seems to have restarted the auto-tester.

Geod24:
I don't think this is the right fix. It leaves the functionality in a weird place:
1) `format` will *still* compile;
2) `format` will print an un-silence-able pragma, possible multiple-line longs;

If we believe that this is a user error, then we should just check for the presence of `toString`, and error-out if it can't be used. If we believe that it's reasonable to ignore a non-instantiating template, then we should silently ignore it (and mark the issue as WONTFIX).

Personally, I think that when a `toString` is present, it should be picked up by the standard library. We saw times and times again that silent fallbacks because the template doesn't instantiate is not what the user expect.

RazvanN7:
ping @Biotronic

PR #7620, opened 2020-09-08
A fix for chunkBy so that it can take predicate which has references to the frame of the calling function.

jamesragray:
Moved nested structs outside of ChunkByImpl so that it can take a predicate which has references to the frame of the calling function.


thewilsonator:
Please reference or create a bugzilla issue.

jamesragray:
Opened new one because I had some problems with git. I added tests to the new one.

PR #7622, opened 2020-09-08
Enable preview=fieldwise & fix blocking bug in VariantN

MoonlightSentinel:
The first commit removes unreachable code in VaraintN for struct without members. This bug was exposed by `-preview=fieldwise` wich enables early const-folding for `EmptyStruct == EmptyStruct` (always `true`).

The second commit enables the flag for phobos s.t. it stays compatible in the future.

### DIFF
--- a/std/variant.d
+++ b/std/variant.d
@@ -253,7 +253,13 @@ private:
                 {
                     return 0;
                 }
-                static if (is(typeof(*zis < *rhsPA)))
+                static if (is(A == struct) && A.init.tupleof.length == 0)
+                {
+                    // The check above will always succeed if A is an empty struct.
+                    // Don't generate unreachable code as seen in
+                    // https://issues.dlang.org/show_bug.cgi?id=21231
+                }
+                else static if (is(typeof(*zis < *rhsPA)))
                 {
                     // Many types (such as any using the default Object opCmp)
                     // will throw on an invalid opCmp, so do it only
@@ -1561,6 +1567,25 @@ pure nothrow @nogc
     auto other = a.get!S;
     assert(msg.array[0] == 3);
     assert(other.array[0] == 3);
+}
+
+// https://issues.dlang.org/show_bug.cgi?id=21231
+// Compatibility with -preview=fieldwise
+@system unittest
+{
+    static struct Empty
+    {
+        bool opCmp(const scope ref Empty) const
+        { return false; }
+    }
+
+    Empty a, b;
+    assert(a == b);
+    assert(!(a < b));
+
+    VariantN!(4, Empty) v = a;
+    assert(v == b);
+    assert(!(v < b));
 }
 
 /**
--- a/posix.mak
+++ b/posix.mak
@@ -139,7 +139,7 @@ endif
 
 # Set DFLAGS
 DFLAGS=
-override DFLAGS+=-conf= -I$(DRUNTIME_PATH)/import $(DMDEXTRAFLAGS) -w -de -preview=dip1000 -preview=dtorfields $(MODEL_FLAG) $(PIC) -transition=complex
+override DFLAGS+=-conf= -I$(DRUNTIME_PATH)/import $(DMDEXTRAFLAGS) -w -de -preview=dip1000 -preview=dtorfields  -preview=fieldwise $(MODEL_FLAG) $(PIC) -transition=complex
 ifeq ($(BUILD),debug)
 override DFLAGS += -g -debug
 else
--- a/win32.mak
+++ b/win32.mak
@@ -41,13 +41,13 @@ DRUNTIMELIB=$(DRUNTIME)/lib/druntime.lib
 
 ## Flags for dmd D compiler
 
-DFLAGS=-conf= -O -release -w -de -preview=dip1000 -preview=dtorfields -transition=complex -I$(DRUNTIME)\import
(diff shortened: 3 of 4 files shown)


MoonlightSentinel:
Blocked because it won't compile without `-preview=fieldwise` after the bugfix

std/variant.d
```diff
@@ -253,7 +253,13 @@ private:
                 {
                     return 0;
                 }
-                static if (is(typeof(*zis < *rhsPA)))
+                static if (is(A == struct) && A.init.tupleof.length == 0)
+                {
+                    // The check above will always succeed if A is an empty struct.
```
nordlow, line 258: Why not `return 0;` here?

nordlow:
> Blocked because it won't compile without `-preview=fieldwise` after the bugfix.

How does this error?

std/variant.d
MoonlightSentinel: Because it would be unreachable code - diagnosed with `-preview=fieldwise`.

MoonlightSentinel:
> How does this error?

Missing `return..`, `assert(0)`, ... at the end of the function.

std/variant.d
nordlow: What about moving

```D
if (*rhsPA == *zis) { return 0; }
```

as two separate code duplications into the `else static if` and `else` branches? I guess we can afford some code duplication here.

That should remove the warning for the

```D
static if (is(A == struct) && A.init.tupleof.length == 0)
```

case right?

nordlow:
> Missing `return..`, `assert(0)`, ... at the end of the function.

I think this is solvable. See my code comment above.

std/variant.d
nordlow: This compiles for me without the flag:

```d
        static ptrdiff_t compare(A* rhsPA, A* zis, OpID selector)
        {
            static if (is(typeof(*rhsPA == *zis)))
            {
                static if (!(is(A == struct) && A.init.tupleof.length == 0))
                {
                    if (*rhsPA == *zis)
                        return 0;
                }
                static if (is(A == struct) && A.init.tupleof.length == 0)
                {
                    return 0;
                }
                else static if (is(typeof(*zis < *rhsPA)))
                {
                    // Many types (such as any using the default Object opCmp)
                    // will throw on an invalid opCmp, so do it only
                    // if the caller requests it.
                    if (selector == OpID.compare)
                        return *zis < *rhsPA ? -1 : 1;
                    else
                        return ptrdiff_t.min;
                }
                else
                {
                    // Not equal, and type does not support ordering
                    // comparisons.
                    return ptrdiff_t.min;
                }
            }
            else
            {
                // Type does not support comparisons at all.
                return ptrdiff_t.min;
            }
        }
```

MoonlightSentinel: That would also need to check for user-defined `opEquals` inside of an empty struct.

MoonlightSentinel:
> I believe I have a solution. See code above.

Nice.

But i really think the compiler should ignore "unused code" for such scenarios. Trying to find an akward workaround wastes a lot of a developers time (and probably increases compilation time) for very little benefit.

nordlow:
Are you putting this in another PR?

std/variant.d
nordlow: Can you elaborate on why you need to check for `opEquals`?

MoonlightSentinel: The user-defined `opEquals` may have different semantics than the compiler generated one (e.g. `return false`
). Your proposal skips it and simply returns "equal".

Additionally, the warning disappears when the empty struct defines an `opEquals` (see https://run.dlang.io/is/rvo4Yt), which suggests that it breaks the early const-folding. That would trigger the `missing return`,... again

nordlow:
I have an idea. What about instead of disabling unreachable code analysis making its diagnostics better in conveing to the developer how he should change the code. For instance

    struct S {} // line 1

    void f(S s, S y) // line 3
    {
        if (x == y) // line 5
            return true; // line 6
        return false; // line 7
    }

could instead say something like

    line 7: statement/line is unreachable because
    line 5: expression `x == y` is always true resulting in
    line 6: unconditional return here

, right?

std/variant.d
nordlow: > The user-defined `opEquals` may have different semantics than the compiler generated one (e.g. `return false`
> ). Your proposal skips it and simply returns "equal".

What about statically checking for presence of user-defined `opEquals` and avoid special handling then?

Have you asked @WalterBright how to proceed with these problems?

nordlow:
I hope it's ok if I [reopen this](https://github.com/dlang/phobos/pull/7624/files).

MoonlightSentinel:
Well, better diagnostics for non-templated code would be nice indeed. But that doesn't  help code where "unreachable" is dependent on template parameters (and somewhat on compiler flags). 

Also note that the compiler should only suppress "unreachable code" warnings when there is a conditionally compiled `return`, `throw`, ..., not in general.

std/variant.d
MoonlightSentinel: > What about statically checking for presence of user-defined `opEquals` and avoid special handling then?

That could work
 
> Have you asked @WalterBright how to proceed with these problems?

No and I don't see a reason to do so. He rarely contributes to phobos and there's more imporant work only he can do (e.g. review backend pull requests)

nordlow: Ok, thanks.

wilzbach: Yup, the bright side of the lack of leadership is that we have more decision freedom here than at DMD ;-)

PR #7624, opened 2020-09-10
Make Phobos compilable with preview=fieldwise

nordlow:
Reopened to see if things build.

### DIFF
--- a/std/variant.d
+++ b/std/variant.d
@@ -249,31 +249,35 @@ private:
         {
             static if (is(typeof(*rhsPA == *zis)))
             {
-                if (*rhsPA == *zis)
-                {
-                    return 0;
-                }
-                static if (is(A == struct) && A.init.tupleof.length == 0)
+                enum isEmptyStructWithoutOpEquals = is(A == struct) && A.tupleof.length == 0 &&
+                                                    !__traits(hasMember, A, "opEquals");
+                static if (isEmptyStructWithoutOpEquals)
                 {
                     // The check above will always succeed if A is an empty struct.
                     // Don't generate unreachable code as seen in
                     // https://issues.dlang.org/show_bug.cgi?id=21231
-                }
-                else static if (is(typeof(*zis < *rhsPA)))
-                {
-                    // Many types (such as any using the default Object opCmp)
-                    // will throw on an invalid opCmp, so do it only
-                    // if the caller requests it.
-                    if (selector == OpID.compare)
-                        return *zis < *rhsPA ? -1 : 1;
-                    else
-                        return ptrdiff_t.min;
+                    return 0;
                 }
                 else
                 {
-                    // Not equal, and type does not support ordering
-                    // comparisons.
-                    return ptrdiff_t.min;
+                    if (*rhsPA == *zis)
+                        return 0;
+                    static if (is(typeof(*zis < *rhsPA)))
+                    {
+                        // Many types (such as any using the default Object opCmp)
+                        // will throw on an invalid opCmp, so do it only
+                        // if the caller requests it.
+                        if (selector == OpID.compare)
+                            return *zis < *rhsPA ? -1 : 1;
+                        else
+                            return ptrdiff_t.min;
+                    }
+                    else
+                    {
+                        // Not equal, and type does not support ordering
+                        // comparisons.
+                        return ptrdiff_t.min;
+                    }
                 }
             }
             else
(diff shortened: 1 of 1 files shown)


nordlow:
Alternatively, is it better to statically check whether an `==`-expression constant folds or not?

nordlow:
No correctly handles opEquals aswell, @MoonlightSentinel.

thewilsonator (approved):

MoonlightSentinel:
Nice, that seems to work with and without the flag. But I'm not sure if we should enable it by default for all CI machines given that code could break when not compiling with `-preview=fieldwise`.

Let's merge the bugfix first and deal with the transition in the dmd pr when it becomes the default.

nordlow:
Do you mean that this PR is the bugfix or you should I remove anything from this PR? I'm satisfied.

MoonlightSentinel:
I meant the changes to the Makefiles, we want to ensure Phobos is compilable without `-preview=fieldwise`.

std/variant.d
```diff
@@ -249,11 +249,21 @@ private:
...
             {
-                if (*rhsPA == *zis)
+                enum isEmptyStructWithoutOpEquals = (is(A == struct) && A.tupleof.length == 0 &&
+                                                     !__traits(hasMember, A, "opEquals"));
+                static if (!isEmptyStructWithoutOpEquals)
+                {
+                    if (*rhsPA == *zis)
+                        return 0;
+                }
+                static if (isEmptyStructWithoutOpEquals)
```
andralex, line 259:
```suggestion
                else
```

std/variant.d
```diff
@@ -249,11 +249,21 @@ private:
...
                 {
                     return 0;
+                    // The check above will always succeed if A is an empty struct.
```
andralex, line 262: Would the comment usually go before the code?

std/variant.d
```diff
@@ -249,11 +249,21 @@ private:
         {
             static if (is(typeof(*rhsPA == *zis)))
             {
-                if (*rhsPA == *zis)
+                enum isEmptyStructWithoutOpEquals = (is(A == struct) && A.tupleof.length == 0 &&
+                                                     !__traits(hasMember, A, "opEquals"));
```
andralex, line 253:
```suggestion
                enum isEmptyStructWithoutOpEquals = is(A == struct) && A.tupleof.length == 0 &&
                                                     !__traits(hasMember, A, "opEquals");
```

nordlow:
> Nice, that seems to work with and without the flag. But I'm not sure if we should enable it by default for all CI machines given that code could break when not compiling with `-preview=fieldwise`.

Shall we revert the additions of `-preview=dtorfields` aswell?

std/variant.d
nordlow: Fixed.

wilzbach:
> Shall we revert the additions of `-preview=dtorfields` aswell?

No that's fine and very unlikely to break something. Furthermore, that one is more of a bug fix.

std/variant.d
nordlow: This doen't work because there's an `else static if` after this `static if`.

std/variant.d
nordlow: Done.

nordlow:
Corrected according to comments.

std/variant.d
andralex: Not a big deal, but just seeing a condition followed by its negation seemed odd. The refactoring that would take care of that would be:

```d
                static if (isEmptyStructWithoutOpEquals)
                {
                    // The check above will always succeed if A is an empty struct.
                    // Don't generate unreachable code as seen in
                    // https://issues.dlang.org/show_bug.cgi?id=21231
                    return 0;
                }
                else
                {
                    if (*rhsPA == *zis)
                        return 0;
                    static if (is(typeof(*zis < *rhsPA)))
                    {
                        // Many types (such as any using the default Object opCmp)
                        // will throw on an invalid opCmp, so do it only
                        // if the caller requests it.
                        if (selector == OpID.compare)
                            return *zis < *rhsPA ? -1 : 1;
                        else
                            return ptrdiff_t.min;
                    }
                    else
                    {
                        // Not equal, and type does not support ordering
                        // comparisons.
                        return ptrdiff_t.min;
                    }
            }
```

This better clarifies the flow depending on compile-time conditions.

nordlow: done.

nordlow:
Ping. Ready to roll?

nordlow:
AFAICT, [opPostMove](https://github.com/dlang/DIPs/pull/109) will be possible if we `-preview=fieldwise` becomes the default.

nordlow:
Reverted changes to `win32.mak` and `win64.mak`. Ready for merge?

PR #7628, opened 2020-09-11
Reduce trait template instantiation in std.utf

n8sh:


atilaneves:
Skimmed it but didn't understand what the PR was trying to do. Could you please describe it? Thanks.

thewilsonator (approved):

n8sh:
>Skimmed it but didn't understand what the PR was trying to do. Could you please describe it? Thanks.

Overall goal was to slightly decrease compilation complexity without having an effect on the produced code. In some cases templates were redundant and could be removed altogether. In others template-based idioms were replaced by `is` statements for a smaller win. In the case of `toUTFz`/`toUTFzImpl` some template constraints that were duplicated for each `toUTFzImpl` were moved up a level to `toUTFz`.

PR #7629, opened 2020-09-12
On POSIX std.file.copy was copying file times at second precision

n8sh:


CyberShadow:
It would be nice if this could use the same code as `statTimeToStdTime`, `setTimes` etc. to avoid duplication and support more platforms.

std/datetime/systime.d
```diff
@@ -2523,7 +2523,7 @@ public:
 
     version (StdDdoc)
     {
-        private struct timespec {}
+        version (Windows) private struct timespec {}
```
n8sh, line 2526: This is to avoid
>/home/dtest/DAutoTest/work/tmp/ddoc_preprocessor_062038e10dce4e6eb0a29d181365ced6/std/file.d(1557): Error: cannot implicitly convert expression `accessTime.toTimeSpec()` of type `std.datetime.systime.SysTime.timespec` to `core.sys.posix.signal.timespec`

n8sh:
>std/file.d(1512:6)[warn]: Public declaration 'setTimes' has no documented example.

Moved `setTimesImpl` after the documented tests to work around a Dscanner limitation but it makes the diff uglier.

CyberShadow (approved):

PR #7632, opened 2020-09-13
Simplify min and max

andralex:
No need to compute the type separately - just use `auto` and compute it incrementally.

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -1569,10 +1569,13 @@ if (T.length >= 2 && !is(CommonType!T == void))
         " and " ~ T1.stringof ~ " for ordering.");
 
     // Compute the returned type.
-    static if (!is(typeof(T0.min)))
-        alias Result = CommonType!(T0, T1);
-    else
+    static if (is(typeof(mostNegative!T0 < mostNegative!T1)))
+        // Both are numeric (or character or Boolean), so we choose the one with the highest maximum.
+        // (We use mostNegative for num/bool/char testing purposes even if it's not used otherwise.)
         alias Result = Select!(T1.max > T0.max, T1, T0);
+    else
+        // At least one is non-numeric, so just go with the common type.
+        alias Result = CommonType!(T0, T1);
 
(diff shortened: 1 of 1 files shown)


std/algorithm/comparison.d
```diff
@@ -1690,31 +1644,39 @@ Returns:
...
+        auto b = min(args[($ + 1) / 2 .. $]);
     alias T1 = typeof(b);
 
     static assert(is(typeof(a < b)),
         "Invalid arguments: Cannot compare types " ~ T0.stringof ~
-        " and " ~ T1.stringof ~ ".");
+        " and " ~ T1.stringof ~ " for ordering.");
+
+    // Compute the returned type.
+    static if (is(typeof(mostNegative!T0)) && is(typeof(mostNegative!T1)))
```
n8sh, line 1669:
```suggestion
    static if (is(typeof(mostNegative!T0 < mostNegative!T1)))
```

n8sh:
The type determination logic is not identical. The below static asserts currently pass but this PR will make them fail:
```d
void main()
{
    static struct Box(T) if (__traits(isArithmetic, T))
    {
        import std.traits : mostNegative;
        enum T max = T.max, min = mostNegative!T;
        T value;
        alias value this;
    }

    auto a = Box!int(10);
    auto b = Box!long(5);
    static assert(is(typeof(max(a, b)) == Box!long));
    static assert(is(typeof(max(b, a)) == Box!long));
}
```

andralex:
@n8sh correct, and the change in behavior intentional. It was never in the charter of `min` or `max` to establish an ad-hoc contract for types that define `min` and `max`. It's not in the documentation and the entire logic is predicated on the semantics of built-in numerics. Consider:

```d
struct Interval {
    double min, max; // interval bounds
    ...
}
```

Now if `Interval` defines some functions for conversion, comparison etc., it's liable to have trouble with `min` and `max`.

So the charter of `min` and `max` is to work well for numeric types, and to defer to the default implementation for other types.

atilaneves (approved):

PR #7635, opened 2020-09-15
A few more minor improvements to std.algorithm.comparison: pushed imp…

andralex:
…lementation constraint down, undocument superfluous flag name, allow comparison of lengths involving infinite ranges, simplify implementation of isSameLength

Was taking a look at the module and found a few more things to improve that don't fit a particular category or issue.

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -891,9 +891,6 @@ Compares two ranges for equality, as defined by predicate `pred`
 */
 template equal(alias pred = "a == b")
 {
-    enum isEmptyRange(R) =
-        isInputRange!R && __traits(compiles, {static assert(R.empty, "");});
-
     enum hasFixedLength(T) = hasLength!T || isNarrowString!T;
 
     // use code points when comparing two ranges of UTF code units that aren't
@@ -931,11 +928,7 @@ template equal(alias pred = "a == b")
         static if (!useCodePoint!(Range1, Range2))
         {
             // No pred calls necessary.
-            static if (isEmptyRange!Range1 || isEmptyRange!Range2)
-            {
-                return r1.empty && r2.empty;
-            }
-            else static if (isInfinite!Range1 || isInfinite!Range2)
+            static if (isInfinite!Range1 || isInfinite!Range2)
             {
                 // No finite range can be ever equal to an infinite range.
                 return false;
@@ -983,8 +976,7 @@ template equal(alias pred = "a == b")
                 //Generic case, we have to walk both ranges making sure neither is empty
                 for (; !r1.empty; r1.popFront(), r2.popFront())
                 {
-                    if (r2.empty) return false;
-                    if (!binaryFun!(pred)(r1.front, r2.front)) return false;
+                    if (r2.empty || !binaryFun!(pred)(r1.front, r2.front)) return false;
                 }
                 return r2.empty;
             }
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -889,20 +889,14 @@ Compares two ranges for equality, as defined by predicate `pred`
 */
 template equal(alias pred = "a == b")
 {
-    // use code points when comparing two ranges of UTF code units that aren't
-    // the same type. This is for backwards compatibility with autodecode
-    // strings.
-    enum useCodePoint(R1, R2) =
-        isSomeChar!(ElementEncodingType!R1) && isSomeChar!(ElementEncodingType!R2) &&
-        (ElementEncodingType!R1).sizeof != (ElementEncodingType!R2).sizeof;
-
     /++
     Compares two ranges for equality. The ranges may have
     different element types, as long as `pred(r1.front, r2.front)`
     evaluates to `bool`.
     Performs $(BIGOH min(r1.length, r2.length)) evaluations of `pred`.
 
-    At least one of the ranges must be finite. If one range involved is infinite, the result is `false`.
+    At least one of the ranges must be finite. If one range involved is infinite, the result is
+    (statically known to be) `false`.
 
     If the two ranges are different kinds of UTF code unit (`char`, `wchar`, or
     `dchar`), then the arrays are compared using UTF decoding to avoid
@@ -917,32 +911,37 @@ template equal(alias pred = "a == b")
         for element, according to binary predicate `pred`.
     +/
     bool equal(Range1, Range2)(Range1 r1, Range2 r2)
-    if (useCodePoint!(Range1, Range2) ||
-        isInputRange!Range1 && isInputRange!Range2 &&
+    if (isInputRange!Range1 && isInputRange!Range2 &&
         !(isInfinite!Range1 && isInfinite!Range2) &&
         is(typeof(binaryFun!pred(r1.front, r2.front))))
     {
-        static if (useCodePoint!(Range1, Range2))
+        // Use code points when comparing two ranges of UTF code units that aren't
+        // the same type. This is for backwards compatibility with autodecode
+        // strings.
+        enum useCodePoint =
+            isSomeChar!(ElementEncodingType!Range1) && isSomeChar!(ElementEncodingType!Range2) &&
+            ElementEncodingType!Range1.sizeof != ElementEncodingType!Range2.sizeof;
+
+        static if (useCodePoint)
         {
             import std.utf : byDchar;
             return equal(r1.byDchar, r2.byDchar);
         }
         else
         {
-            // No pred calls necessary.
             static if (isInfinite!Range1 || isInfinite!Range2)
             {
                 // No finite range can be ever equal to an infinite range.
                 return false;
             }
-            //Detect default pred and compatible dynamic array
+            // Detect default pred and compatible dynamic arrays.
             else static if (is(typeof(pred) == string) && pred == "a == b" &&
                 isArray!Range1 && isArray!Range2 && is(typeof(r1 == r2)))
             {
                 return r1 == r2;
             }
-            // if one of the arguments is a string and the other isn't, then auto-decoding
-            // can be avoided if they have the same ElementEncodingType
+            // If one of the arguments is a string and the other isn't, then auto-decoding
+            // can be avoided if they have the same ElementEncodingType.
             else static if (is(typeof(pred) == string) && pred == "a == b" &&
                 isAutodecodableString!Range1 != isAutodecodableString!Range2 &&
                 is(immutable ElementEncodingType!Range1 == immutable ElementEncodingType!Range2))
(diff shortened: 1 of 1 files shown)


std/algorithm/comparison.d
```diff
@@ -913,81 +913,80 @@ template equal(alias pred = "a == b")
...
             {
-                if (!binaryFun!(pred)(r1.front, r2.front)) return false;
+                //Generic case, we have to walk both ranges making sure neither is empty
+                for (; !r1.empty; r1.popFront(), r2.popFront())
+                {
+                    if (r2.empty) return false;
+                    if (!binaryFun!(pred)(r1.front, r2.front)) return false;
+                }
+                static if (!isInfinite!Range1)
+                    return r2.empty;
```
andralex, line 981: The hunk above looks a lot of a bigger change than it is... I just merge the two overloads into one with a static if inside.

std/algorithm/comparison.d
```diff
@@ -913,81 +912,72 @@ template equal(alias pred = "a == b")
...
-        }
-        else static if ((isInfinite!Range1 && hasFixedLength!Range2) ||
-            (hasFixedLength!Range1 && isInfinite!Range2))
-        {
-            return false;
-        }
-        //Detect default pred and compatible dynamic array
-        else static if (is(typeof(pred) == string) && pred == "a == b" &&
-            isArray!Range1 && isArray!Range2 && is(typeof(r1 == r2)))
+        static if (!useCodePoint!(Range1, Range2))
```
schveiguy, line 919: By combining the functions, we don't need `useCodePoint` outside this function (it was only used to distinguish the overload). However, there is a slight difference here, as the constraint has checked the predicate against the range type, but in the case of two differing character widths, the code is going to reroute it to comparing dchars instead. Given that the vast vast majority (possibly unanimous) usage of comparing ranges of characters is going to rely on the default comparison of a == b, it's probably not an issue.

std/algorithm/comparison.d
```diff
@@ -900,6 +897,8 @@ template equal(alias pred = "a == b")
     evaluates to `bool`.
     Performs $(BIGOH min(r1.length, r2.length)) evaluations of `pred`.
 
+    At least one of the ranges must be finite. If one range involved is infinite, the result is `false`.
```
MetaLang, line 900: I know that conceptually, ∞ != ∞, but it doesn't make much sense that `repeat(3) != repeat(3)`.

andralex: Two infinite ranges won't be accepted at all. The result `false` is returned if exactly one is infinite.

std/algorithm/comparison.d
andralex: Oh wait that wasn't intended. Can you explain a bit more? So the old code had these overloads:

```d
bool equal(Range1, Range2)(Range1 r1, Range2 r2)
if (!useCodePoint!(Range1, Range2) &&
    isInputRange!Range1 && isInputRange!Range2 &&
    is(typeof(binaryFun!pred(r1.front, r2.front))));
bool equal(Range1, Range2)(Range1 r1, Range2 r2)	
    if (useCodePoint!(Range1, Range2));
```

My intent was simply to merge these into one while preserving the semantics:

```d
bool equal(Range1, Range2)(Range1 r1, Range2 r2)
if (isInputRange!Range1 && isInputRange!Range2 &&
    is(typeof(binaryFun!pred(r1.front, r2.front))));
```

and use `static if` inside to discriminate on `useCodePoint!(Range1, Range2`. Where did I go wrong?

schveiguy: What I mean is effectively prior to this change, you had this constraint on the second form:

```D
if (useCodePoint!(Range1, Range2))
```

And now the constraint on the second form is:

```D
if(useCodePoint!(Range1, Range2) && // this first part is embedded inside the function
    isInputRange!Range1 && isInputRange!Range2 &&
    is(typeof(binaryFun!pred(r1.front, r2.front))));
```

So suppose you pass in a range of `char`, and a range of `wchar`. This is going to use the "decode-to-dchar" mechanism, but the constraint is checking if `binaryFun!pred(char, wchar)` compiles even though that is never actually called.

Again, I think it's a nonsensical case -- who is going to use the function called `equal` to check for something other than equality? Most people who want a transformation would use map to first transform the elements, and then use `equal` on that. But there is that slight difference.

If we wanted it to be "correct", I would say leave the secondary form separate. And in fact, the secondary form technically should check that the predicate works on `(dchar, dchar)` in that case. I left it out because it was a simple wrapper, and I knew it would be checked in the forwarded call properly.

quickfur: In theory, `pred` does not need to be strict equality; it can implement an equivalence relation.  For example, one might want to compare two strings in a case-insensitive manner (e.g., 'A' and 'a' are in the same equivalence class, and so forth).

schveiguy: Right, but the question is, will evaluating whether the predicate can be called with the individual char types should affect being able to call it.

I mean, I can't imagine a case where you can call the predicate using dchar, but not using char or wchar.

std/algorithm/comparison.d
MetaLang: My mistake, I misread the template constraints and thought it allowed two infinite ranges.

std/algorithm/comparison.d
andralex: Got it, will fix just to preserve existing behavior. Thanks @schveiguy !

andralex: I woke up with the feeling we have a bad solution. First off, the documentation will show the private symbol `useCodePoint`. That could be fixed with some awkwardness, but then there's the simple truth that `useCodePoint` is a sheer implementation choice that the user does not want or need to know about.

I'll work out something else - perhaps move the predicate validation to a `static assert`.

andralex: I opted for pushing `useCodePoint` all the way in and leaving the predicate in the condition. The conditions under that wouldn't compile are vanishingly rare.

std/algorithm/comparison.d
```diff
@@ -1964,60 +1950,31 @@ Returns:
...
-if (isInputRange!Range1 &&
-    isInputRange!Range2 &&
-    !isInfinite!Range1 &&
-    !isInfinite!Range2)
+if (isInputRange!Range1 && isInputRange!Range2)
 {
-    static if (hasLength!(Range1) && hasLength!(Range2))
+    static if (isInfinite!Range1 || isInfinite!Range2)
+    {
+        return isInfinite!Range1 && isInfinite!Range2;
```
wilzbach, line 1957: Cantor disagrees:
- https://en.wikipedia.org/wiki/Cantor%27s_diagonal_argument
- https://en.wikipedia.org/wiki/Cantor%27s_theorem

(tl;dr: infinite sets can have different length)

andralex: Heh, crossed my mind. Both are countable so they actually do have "the same" length.

std/algorithm/comparison.d
schveiguy: Yeah, I agree. Thanks!

std/algorithm/comparison.d
```diff
@@ -530,6 +529,7 @@ Returns:
 
 */
 auto clamp(T1, T2, T3)(T1 val, T2 lower, T3 upper)
+if (is(typeof(max(lower, min(upper, val)))))
```
schveiguy, line 532: eh. This is kind of annoying. You are repeating the entire function in the constraints, which means that you will get a matching error with the provided code instead of a more specific error. I would recommend against this.

In this case, it's simple, but in the general case, I don't like repeating the entire function in the constraints, unless there is a need to do this (e.g. for overloading). Later, if you change one but not the other, really confusing things happen on corner cases.

schveiguy (changes requested):
I don't think the added complete-implementation-as-constraint is an improvement. Other than that, everything else looks good.

std/algorithm/comparison.d
schveiguy: The error message looks OK, and this example is somewhat saved by the fact that the current error message looks laughably confusing:

```D
import std.algorithm;
auto x = clamp("2", 1, 3);
```

```
/dlang/dmd/linux/bin64/../../src/phobos/std/algorithm/comparison.d(1634): Error: no property max for type string, perhaps import std.algorithm; is needed?
/dlang/dmd/linux/bin64/../../src/phobos/std/algorithm/comparison.d(1664): Error: template instance std.algorithm.comparison.MinType!(int, string) error instantiating
/dlang/dmd/linux/bin64/../../src/phobos/std/algorithm/comparison.d(540):        instantiated from here: min!(int, string)
onlineapp.d(11):        instantiated from here: clamp!(string, int, int)
```

But this is actually a problem with `MinType`, which makes some incorrect assumptions (if T[0] has a min property, then all T must have min/max properties). A better error message down in `min` would be useful not just for this case.

My preference is to leave the constraint off, and fix the error message from min/max to be useful. I don't think even for one-liners we should be repeating the entire code in the constraint.

An exception would be for cases where constraints are needed for overloading.

andralex: MinType will be eliminated by another PR. Not sure where this would leave us.

Phobos functions have constraints that prevent them from accepting arbitrary arguments, to then fail in their own code in arbitrary ways. `clamp` was an outlier, and this PR fixes that.

schveiguy: Yeah, actually, the error getting pushed down doesn't help. Because min or max has a similar error problem of gagging actual errors that might be useful.

schveiguy:
Needs a rebase.

wilzbach (commented):
Not happy with these self-merges that still have a lot of problems with them and have zero net benefit.

std/algorithm/comparison.d
```diff
@@ -1973,60 +1958,31 @@ Returns:
...
-            r2.popFront;
-
-            if (++length > r1.length)
-            {
-                return false;
-            }
-        }
-
-        return !(length < r1.length);
+        return r2.walkLength(r1.length + 1) == r1.length;
```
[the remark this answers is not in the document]
schveiguy: why is this a performance regression? Doesn't walkLength do the same thing as the original code?

std/algorithm/comparison.d
```diff
@@ -1973,60 +1958,31 @@ Returns:
...
-            }
-        }
-
-        return !(length < r2.length);
+        return r1.walkLength(r2.length + 1) == r2.length;
     }
     else
     {
-        while (!r1.empty)
+        for (; !r1.empty; r1.popFront, r2.popFront)
```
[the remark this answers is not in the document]
schveiguy: It does. The third clause is not run for the first loop.

std/algorithm/comparison.d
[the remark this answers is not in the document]
schveiguy: In fact, I'd say the walkLength call might be faster, as it stores r1.length into a local instead of calling r1.length in every loop.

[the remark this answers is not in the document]
wilzbach: Previous code stopped when either length mismatched. Now it needs to walk through the full length.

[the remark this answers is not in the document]
schveiguy: No, you misread the call:

```
The `upTo` parameter is useful to "cut the losses" in case
the interest is in seeing whether the range has at least some number
of elements. If the parameter `upTo` is specified, stops if $(D
upTo) steps have been taken and returns `upTo`.
```

PR #7637, opened 2020-09-17
Fix issue 21253 - Can't compile Variant.visit!(...) with generic func…

Biotronic:
…tion

### DIFF
--- a/std/variant.d
+++ b/std/variant.d
@@ -2430,11 +2430,11 @@ if (Handlers.length > 0)
 
     Algebraic!(int, string) maybenumber = 2;
     // ok, x ~ "a" valid for string, x + 1 valid for int, only 1 generic
-    static assert( __traits(compiles, number.visit!((string x) => x ~ "a", x => x + 1)));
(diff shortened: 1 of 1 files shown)


Biotronic:
Turns out, the tests on lines 2433-2437 probably should have referred to `maybenumber` all along, and were testing for the wrong thing.

pbackus:
The `nonmatch` check is a breaking change, because `visit` currently ignores handlers that don't match any types. Here's a unit test that passes on `master` but breaks with this PR:

```d
@system unittest
{
    Algebraic!(int, string) x = 123;
    assert(__traits(compiles,
        x.visit!(
            (int n) => "got int",
            (string s) => "got string",
            (char c) => "got char"
        )
    ));
}
```

myCrack:
> breaking change, because visit currently ignores handlers that don't match any types.

https://dlang.org/phobos/std_variant.html#.visit:
> It is statically ensured that all held types of variant are handled across **_all_** handlers.

So, behavior from the master branch must be treated as a bug.

pbackus:
I'm pretty sure that just means that you have to handle every type (which, indeed, is statically ensured on `master`), but the wording is certainly not as clear as it could be.

thewilsonator (approved):

PR #7638, opened 2020-09-18
Refactor std.algorithm.iteration

andralex:
The refactorings involve simplifying code that was working around compiler bugs, pushing symbols from top level to where they are needed, and a few more simplifications.

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -363,16 +363,6 @@ private struct _Cache(R, bool bidir)
         }
     }
 
-    static if (isInfinite!R)
-        enum empty = false;
-    else
-        bool empty() @property
-        {
-            return source.empty;
-        }
-
-    mixin ImplementLength!source;
-
     E front() @property
     {
         version (assert) if (empty) throw new RangeError();
@@ -412,42 +402,9 @@ private struct _Cache(R, bool bidir)
         }
     }
 
-    static if (hasSlicing!R)
-    {
-        enum hasEndSlicing = is(typeof(source[size_t.max .. $]));
-
-        static if (hasEndSlicing)
-        {
-            private static struct DollarToken{}
-            enum opDollar = DollarToken.init;
-
-            auto opSlice(size_t low, DollarToken)
-            {
-                return typeof(this)(source[low .. $]);
-            }
-        }
-
-        static if (!isInfinite!R)
-        {
-            typeof(this) opSlice(size_t low, size_t high)
-            {
-                return typeof(this)(source[low .. high]);
-            }
-        }
-        else static if (hasEndSlicing)
-        {
-            auto opSlice(size_t low, size_t high)
-            in
-            {
-                assert(low <= high, "Bounds error when slicing cache.");
-            }
-            do
-            {
-                import std.range : takeExactly;
-                return this[low .. $].takeExactly(high - low);
-            }
-        }
-    }
+    mixin ImplementEmpty!source;
+    mixin ImplementLength!source;
+    mixin ImplementSlicing!source;
 }
 
 /**
@@ -542,37 +499,6 @@ if (fun.length >= 1)
             alias opDollar = length;
         }
 
-        static if (hasSlicing!R)
-        {
-            static if (is(typeof(_input[ulong.max .. ulong.max])))
-                private alias opSlice_t = ulong;
-            else
-                private alias opSlice_t = uint;
-
-            static if (hasLength!R)
-            {
-                auto opSlice(opSlice_t low, opSlice_t high)
-                {
-                    return typeof(this)(_input[low .. high]);
-                }
-            }
-            else static if (is(typeof(_input[opSlice_t.max .. $])))
-            {
-                struct DollarToken{}
-                enum opDollar = DollarToken.init;
-                auto opSlice(opSlice_t low, DollarToken)
-                {
-                    return typeof(this)(_input[low .. $]);
-                }
-
-                auto opSlice(opSlice_t low, opSlice_t high)
-                {
-                    import std.range : takeExactly;
-                    return this[low .. $].takeExactly(high - low);
-                }
-            }
-        }
-
         static if (isForwardRange!R)
         {
             @property auto save()
@@ -580,6 +506,8 @@ if (fun.length >= 1)
                 return typeof(this)(_input.save);
             }
         }
+
+        mixin ImplementSlicing!_input;
     }
 
     /**
@@ -875,7 +803,7 @@ stopping.
 See_Also: $(REF tee, std,range)
  */
 Flag!"each" each(alias fun = "a", Range)(auto ref Range range)
-if (isInputRange!Range || isStaticArray!Range || hasMember!(Range, "opApply"))
+if (isInputRange!Range || isArray!Range || hasMember!(Range, "opApply"))
 {
     import std.meta : AliasSeq;
     import std.traits : Parameters;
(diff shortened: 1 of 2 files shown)


std/algorithm/iteration.d
```diff
@@ -482,6 +469,122 @@ See_Also:
...
+        alias _funs = staticMap!(unaryFun, fun);
+        alias _fun = adjoin!_funs;
+    }
+    else
+    {
+        alias _fun = unaryFun!fun;
+        alias _funs = AliasSeq!(_fun);
+    }
+
+    struct Result(Range)
```
andralex, line 486: moved from top level

std/algorithm/iteration.d
```diff
@@ -955,11 +920,19 @@ private:
          isForeachUnaryWithIndexIterable!R);
 
 public:
+    Flag!"each" each(Range)(auto ref Range r)
+    if (isForeachIterable!Range ||
```
andralex, line 924: Largely unsuccessful attempt to make the constraint nicer. Will get back to this.

std/algorithm/iteration.d
```diff
@@ -1333,7 +1306,71 @@ if (is(typeof(unaryFun!predicate)))
      */
     auto filter(Range)(Range range) if (isInputRange!(Unqual!Range))
     {
-        return FilterResult!(unaryFun!predicate, Range)(range);
+        return Result!Range(range);
+    }
+
+    private struct Result(Range)
```
andralex, line 1312: moved from top level

std/algorithm/iteration.d
```diff
@@ -1570,7 +1545,56 @@ template filterBidirectional(alias pred)
      */
     auto filterBidirectional(Range)(Range r) if (isBidirectionalRange!(Unqual!Range))
     {
-        return FilterBidiResult!(unaryFun!pred, Range)(r);
+        return Result!Range(r);
+    }
+
+    private struct Result(Range)
```
andralex, line 1551: moved from top level

std/algorithm/iteration.d
```diff
@@ -1886,64 +1862,10 @@ pure @safe unittest
...
-        enum ChunkByImplIsUnary = true;
-    else
-        static assert(0, "chunkBy expects either a binary predicate or "~
-                         "a unary predicate on range elements of type: "~
-                         ElementType!Range.stringof);
-}
-
 // Implementation of chunkBy for non-forward input ranges.
-private struct ChunkByImpl(alias pred, Range)
+private struct ChunkByImpl(alias pred, alias eq, bool isUnary, Range)
```
andralex, line 1866: Could move this inside `chunkBy` as well, but it's fairly bulky.

std/algorithm/iteration.d
```diff
@@ -1973,48 +1894,69 @@ if (isInputRange!Range && !isForwardRange!Range)
     @property auto front()
     {
         assert(!empty, "Attempting to fetch the front of an empty chunkBy.");
+
+        static struct Result
```
andralex, line 1898: this was small enough to plop here

std/algorithm/iteration.d
```diff
@@ -1973,48 +1894,69 @@ if (isInputRange!Range && !isForwardRange!Range)
...
+            return Result(r, _prev);
         }
     }
 
     void popFront()
     {
         assert(!empty, "Attempting to popFront an empty chunkBy.");
         openChunk = false;
-        while (!r.empty)
+        for (; !r.empty; r.popFront)
```
andralex, line 1943: my dream: that more loops are written like this

std/algorithm/iteration.d
```diff
@@ -1973,48 +1894,69 @@ if (isInputRange!Range && !isForwardRange!Range)
...
                 break;
             }
-            r.popFront();
         }
     }
 }
 
 // Single-pass implementation of chunkBy for forward ranges.
-private struct ChunkByImpl(alias pred, Range)
+private struct ChunkByImpl(alias pred, alias eq, bool isUnary, Range)
```
andralex, line 1955: this is also too bulky to move inside `chunkBy`

std/algorithm/iteration.d
```diff
@@ -2223,7 +2167,23 @@ if (isForwardRange!Range)
...
+        enum bool isUnary = true;
+    else
+        static assert(0, "chunkBy expects either a binary predicate or "~
+                        "a unary predicate on range elements of type: "~
+                        ElementType!Range.stringof);
+
+    static if (isUnary)
+        alias eq = binaryFun!((a, b) => unaryFun!pred(a) == unaryFun!pred(b));
+    else
+        alias eq = binaryFun!pred;
```
andralex, line 2184: this code was duplicated across the two implementations of `ChunkByImpl`.

std/algorithm/iteration.d
```diff
@@ -918,187 +883,106 @@ template each(alias fun = "a")
...
-    /**
-    Params:
-        r = range or iterable over which each iterates
-     */
-    Flag!"each" each(Range)(Range r)
-    if (!isForeachIterable!Range && (
-        isRangeIterable!Range ||
-        __traits(compiles, typeof(r.front).length)))
+    public Flag!"each" each(Range)(auto ref Range range)
+    if (isInputRange!Range || isStaticArray!Range || hasMember!(Range, "opApply"))
```
andralex, line 889: This is a simpler condition (albeit incomplete because it doesn't verify the function). Also the entire implementation is a lot shorter.

Checked the speed and memory consumed - this version is about as fast and eats 0.1% less memory.

std/algorithm/iteration.d
```diff
@@ -905,182 +870,91 @@ iteration).
...
+    range = range or iterable over which `each` iterates
 
 Returns: `Yes.each` if the entire range was iterated, `No.each` in case of early
 stopping.
 
 See_Also: $(REF tee, std,range)
  */
-template each(alias fun = "a")
+public Flag!"each" each(alias fun = "a", Range)(auto ref Range range)
+if (isInputRange!Range || isStaticArray!Range || hasMember!(Range, "opApply"))
```
andralex, line 881: whew, nice, it can be reduced to a single function

looked-at-me: I don't think this is something that should be celebrated.

```D
        static if (is(typeof(f1!()))) return f1();
        else static if (is(typeof(f2!()))) return f2();
        else static if (is(typeof(f3!()))) return f3();
        else static if (is(typeof(f4!()))) return f4();
        else static if (is(typeof(f5!()))) return f5();
        else static if (is(typeof(f6!()))) return f6();
        else static if (is(typeof(f7!()))) return f7();
        else static assert(0, "Cannot iterate " ~ Range.stringof ~ " with each().");
```

andralex: The "celebration" was in regard to moving `each` from a member in a template to a top-level function.

I agree that the "try each pattern and see what works" strategy does raise an eyebrow, and am always on lookout for better solutions - please advise. I took inspiration from @pbackus's amazing [sumtype library](https://code.dlang.org/packages/sumtype) which works almost miraculously well at scale. There, you get to pass one  or more template lambdas to a sum type and the library simply iterates all possibilities and picks the first that compiles. Experience with sumtype suggests that's a really solid idiom for D.

There are certain advantages to this strategy as implemented here compared to the existing code:

* It's easy to understand and maintain - each pattern is in the clear. This is a very important aspect! It's very difficult to make heads and tails of the code otherwise. Once you see what patterns the library is looking for, it's very easy to get into it.
* It's much more terse than the baseline - this PR deletes 166 lines, many of which I think are attributable to this change.
* There's good precedent with sumtype

dukc:
Good work - just a little nicer to read all over the module. Modulo getting the test suite to pass, LGTM.

andralex:
whoa blast from the past

RazvanN7:
@andralex what are your plans with regards to this PR?

LightBender:

PR #7640, opened 2020-09-19
EASY: factor implementations of range length into a mixin template

andralex:
Arguably this is closer to correctness because some of these ranges don't define `opDollar`.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -349,15 +349,7 @@ if (isBidirectionalRange!(Unqual!Range))
                     }
             }
 
-            static if (hasLength!R)
-            {
-                @property auto length()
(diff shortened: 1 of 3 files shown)


thewilsonator (approved):

std/range/package.d
```diff
@@ -11786,12 +11734,10 @@ public:
         {
             return (*_range).length;
         }
-
```
wilzbach, line 11789: Unrelated!

andralex: cool

andralex: ah I noticed it was merged... guess it'll stay, yay

PR #7641, opened 2020-09-20
EASY: obey the rule of silence.

andralex:
Messages such as "please ignore this message" or warnings that you need to remember are benign have no business in a build log.

### DIFF
--- a/std/experimental/checkedint.d
+++ b/std/experimental/checkedint.d
@@ -1641,24 +1641,34 @@ static:
 
 @system unittest
 {
-    // Temporarily disable output to stderr so as to not spam the build log.
-    import std.stdio : stderr;
-    auto t = stderr;
-    version (Posix)
-        stderr.open("/dev/null", "w");
-    else
-        stderr.open("nul", "w");
-    scope(exit) stderr = t;
-
     auto a = checked!Warn(int.min);
     auto b = checked!Warn(-1);
-    assert(a / b == a * b);
-
-    import std.exception : assertThrown;
-    import core.exception : AssertError;
     auto x = checked!Abort(int.min);
     auto y = checked!Abort(-1);
-    assertThrown!AssertError(x / y);
+
+    // Temporarily redirect output to stderr to make sure we get the right output.
+    import std.process : uniqueTempPath;
+    import std.stdio : stderr;
+    auto tmpname = uniqueTempPath;
+    auto t = stderr;
+    stderr.open(tmpname, "w");
+    // Open a new scope to minimize code ran with stderr redirected.
+    {
+        scope(exit) stderr = t;
+        assert(a / b == a * b);
+        import std.exception : assertThrown;
+        import core.exception : AssertError;
+        assertThrown!AssertError(x / y);
+    }
+    import std.file : readText;
+    auto witness = readText(tmpname);
+    auto expected =
+"Overflow on binary operator: int(-2147483648) / const(int)(-1)\r\n" ~
+"Overflow on binary operator: int(-2147483648) * const(int)(-1)\r\n" ~
+"Overflow on binary operator: int(-2147483648) / const(int)(-1)\r\n";
+    import std.string : replace;
+    version(Posix) expected = expected.replace("\n\r", "\n");
+    assert(witness == expected, "'" ~ witness ~ "'");
 }
 
 // ProperCompare
--- a/std/experimental/checkedint.d
+++ b/std/experimental/checkedint.d
@@ -1661,13 +1661,12 @@ static:
         assertThrown!AssertError(x / y);
     }
     import std.file : readText;
+    import std.ascii : newline;
     auto witness = readText(tmpname);
     auto expected =
-"Overflow on binary operator: int(-2147483648) / const(int)(-1)\r\n" ~
-"Overflow on binary operator: int(-2147483648) * const(int)(-1)\r\n" ~
-"Overflow on binary operator: int(-2147483648) / const(int)(-1)\r\n";
-    import std.string : replace;
-    version(Posix) expected = expected.replace("\n\r", "\n");
+"Overflow on binary operator: int(-2147483648) / const(int)(-1)" ~ newline ~
+"Overflow on binary operator: int(-2147483648) * const(int)(-1)" ~ newline ~
+"Overflow on binary operator: int(-2147483648) / const(int)(-1)" ~ newline;
     assert(witness == expected, "'" ~ witness ~ "'");
 }
 
--- a/std/experimental/checkedint.d
+++ b/std/experimental/checkedint.d
@@ -1642,29 +1642,23 @@ static:
 @system unittest
 {
     // Temporarily disable output to stderr so as to not spam the build log.
-    import std.stdio;
+    import std.stdio : stderr;
     auto t = stderr;
-    stderr.open("/dev/null", "w");
+    version (Posix)
+        stderr.open("/dev/null", "w");
+    else
+        stderr.open("nul", "w");
     scope(exit) stderr = t;
+
     auto a = checked!Warn(int.min);
     auto b = checked!Warn(-1);
     assert(a / b == a * b);
-}
-
-@system unittest
-{
-    // Temporarily disable output to stderr so as to not spam the build log.
-    import std.stdio;
-    auto t = stderr;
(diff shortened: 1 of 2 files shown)


std/experimental/checkedint.d
```diff
@@ -1641,13 +1641,24 @@ static:
 
 @system unittest
 {
+    // Temporarily disable output to stderr so as to not spam the build log.
+    import std.stdio;
+    auto t = stderr;
+    stderr.open("/dev/null", "w");
```
wilzbach, line 1647: This won't work on Windows.

wilzbach (changes requested):
/dev/null doesn't exist on Windows.
Also in theory tests might be run in parallel.

std/process.d
```diff
@@ -3122,7 +3122,7 @@ private auto executeImpl(alias pipeFunc, Cmd, ExtraPipeFuncArgs...)(
     auto r3 = executeShell("exit 123");
     assert(r3.status == 123);
     assert(r3.output.empty);
-    auto r4 = executeShell("echo stderr test, please ignore 1>&2",
+    auto r4 = executeShell("echo -n '\r' 1>&2",
```
CyberShadow, line 3125: This won't work on Windows.

Even on POSIX, I don't think this is portable. It is better to use `printf`. (Which still won't work on Windows. See the test above.)

CyberShadow (changes requested):

std/experimental/checkedint.d
andralex: The name on windows is `nul`, correct?

std/process.d
andralex: I'll try the same trick as above - reopen stderr with the null handle.

std/experimental/checkedint.d
wilzbach: I don't know much about Windows, sorry.
However, it might be better to do a `version (StdUnittest)` at the beginning to define a different stderr temp file as this will avoid all the shenanigans with rewriting the global output. However, it does come with its own set of problems...

wilzbach (commented):
.

andralex:
Looks like that works on osx, let's see windows.

std/experimental/checkedint.d
schveiguy: I thought it was `NUL:`, but wikipedia seems to agree with you: https://en.wikipedia.org/wiki/Null_device

andralex: I think the most robust solution is to write to a temp file and make sure it contains what it's supposed to. Will be back.

andralex:
Not so easy anymore but arguably improve the state of affairs: now the unittests verify that the text output to stderr is as expected.

std/experimental/checkedint.d
```diff
@@ -1643,17 +1643,32 @@ static:
...
+        assertThrown!AssertError(x / y);
+    }
+    import std.file : readText;
+    auto witness = readText(tmpname);
+    auto expected =
+"Overflow on binary operator: int(-2147483648) / const(int)(-1)\r\n" ~
+"Overflow on binary operator: int(-2147483648) * const(int)(-1)\r\n" ~
+"Overflow on binary operator: int(-2147483648) / const(int)(-1)\r\n";
+    import std.string : replace;
+    version(Posix) expected = expected.replace("\n\r", "\n");
```
wilzbach, line 1670: Why not use `newline` from std.ascii directly (above)?

andralex: noice

std/experimental/checkedint.d
```diff
@@ -1643,17 +1643,31 @@ static:
...
+    // Open a new scope to minimize code ran with stderr redirected.
+    {
+        scope(exit) stderr = t;
+        assert(a / b == a * b);
+        import std.exception : assertThrown;
+        import core.exception : AssertError;
+        assertThrown!AssertError(x / y);
+    }
+    import std.file : readText;
+    import std.ascii : newline;
```
andralex, line 1664: thanks @wilzbach for the idea

std/process.d
```diff
@@ -3122,10 +3122,28 @@ private auto executeImpl(alias pipeFunc, Cmd, ExtraPipeFuncArgs...)(
...
-    assert(r4.output.empty);
+}
+
+unittest
+{
+    // Temporarily disable output to stderr so as to not spam the build log.
+    import std.stdio : stderr;
+    import std.typecons : Tuple;
+    import std.file : readText;
+    Tuple!(int, "status", string, "output") r;
```
CyberShadow, line 3133: Maybe
```suggestion
    ReturnType!executeShell r;
```

std/process.d
```diff
@@ -3266,8 +3284,7 @@ private struct TestScript
     string path;
 }
 
-version (StdUnittest)
-private string uniqueTempPath() @safe
+package(std) string uniqueTempPath() @safe
```
CyberShadow, line 3287: There is also `std.file.deleteme`

std/process.d
```diff
@@ -3122,10 +3122,28 @@ private auto executeImpl(alias pipeFunc, Cmd, ExtraPipeFuncArgs...)(
     auto r3 = executeShell("exit 123");
     assert(r3.status == 123);
     assert(r3.output.empty);
-    auto r4 = executeShell("echo stderr test, please ignore 1>&2",
-        null, Config.stderrPassThrough);
-    assert(r4.status == 0);
-    assert(r4.output.empty);
+}
+
+unittest
```
CyberShadow, line 3127:
```suggestion
// Test Config.stderrPassThrough
@system unittest
```

std/process.d
```diff
@@ -3122,10 +3122,28 @@ private auto executeImpl(alias pipeFunc, Cmd, ExtraPipeFuncArgs...)(
...
+    {
+        stderr.open(tmpname, "w");
+        scope(exit) stderr = t;
+        r = executeShell("echo D rox. 1>&2", null, Config.stderrPassThrough);
+    }
+    assert(r.status == 0);
+    assert(r.output.empty);
+    auto witness = readText(tmpname);
+    import std.ascii : newline;
+    assert(witness == "D rox." ~ newline, "'" ~ witness ~ "'");
```
CyberShadow, line 3146: It looks like this is still failing.

`text([str])` or `format("%(%s%)", str.only)` is good for stringifying strings with special characters, as they will show up quoted and with escapes.

The problem might be the space before the `1>&2`, which I think will be literal on Windows.

andralex:
Trying to stick the text to the `>` sign in echo. It only fails on windows32/ldc as far as I can tell.

std/process.d
andralex: Looks like it worked with `echo D rox>&2`.

std/process.d
andralex: Good to know, thanks. Not sure we need both - will leave this decision to the future.

std/process.d
andralex: done

andralex:
@wilzbach one more look please

std/process.d
```diff
@@ -3122,10 +3122,30 @@ private auto executeImpl(alias pipeFunc, Cmd, ExtraPipeFuncArgs...)(
...
+    import std.file : readText;
+    import std.traits : ReturnType;
+
+    ReturnType!executeShell r;
+    auto tmpname = uniqueTempPath;
+    auto t = stderr;
+    // Open a new scope to minimize code ran with stderr redirected.
+    {
+        stderr.open(tmpname, "w");
+        scope(exit) stderr = t;
```
wilzbach, line 3141: `scope (exit)` should probably be the first line (not that it matters much).

wilzbach (approved):
LGTM.

std/process.d
andralex: There is a minor difference - there would be an unnecessary self-assignment if `open` throws (I _hope_/_trust_ it was implemented in a transactional manner).

schveiguy:
@andralex please use the bot to merge. Just add the label auto-merge. If everything has passed, then it will merge immediately.

The bot makes sure that things are not merged that don't pass the CI. In this case, no harm. But it also makes sure that 2 PRs don't get merged simultaneously that break the build if merged at the same time.

andralex:
Oh forgot about that, sorry, it was mentioned previously. Thanks.

PR #7642, opened 2020-09-21
Fix Issue 21233 - std.conv.parse doesn't report the number of charact…

adelavais:
…ers consumed

### DIFF
--- a/changelog/conv_parse_count.dd
+++ b/changelog/conv_parse_count.dd
@@ -1,0 +1,15 @@
+Return the number of characters consumed by `std.conv.parse`
+
+By setting the flag `doCount` to `Yes.doCount`, the function returns a named tuple.
+The tuple contains the fields `data` and `count`.
+The field `data` is of type `Target` and it contains the result of the original function.
+The field `count` is of type `size_t` and it contains the number of characters consumed.
+The function is keeping its original behaviour otherwise.
+Example:
+-------
+import std.typecons : Flag, Yes, No;
+
+string s1 = "123";
+auto a1 = parse!(int, string, Yes.doCount)(s1);
+assert(a1.data == 123 && a1.count == 3);
+-------
--- a/std/conv.d
+++ b/std/conv.d
@@ -53,6 +53,7 @@ public import std.ascii : LetterCase;
 import std.meta;
 import std.range.primitives;
 import std.traits;
+import std.typecons : Flag, Yes, No, tuple;
 
 // Same as std.string.format, but "self-importing".
 // Helps reduce code and imports, particularly in static asserts.
@@ -2173,9 +2174,12 @@ This overload converts a character input range to a `bool`.
 Params:
     Target = the type to convert to
     source = the lvalue of an $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
+    doCount = the flag for deciding to report the number of consumed characters
 
 Returns:
-    A `bool`
+$(UL
+    $(LI A `bool` if `doCount` is set to `No.doCount`)
+    $(LI A `tuple` containing a `bool` and a `size_t` if `doCount` is set to `Yes.doCount`))
 
 Throws:
     A $(LREF ConvException) if the range does not represent a `bool`.
@@ -2184,7 +2188,7 @@ Note:
     All character input range conversions using $(LREF to) are forwarded
     to `parse` and do not require lvalues.
 */
-Target parse(Target, Source)(ref Source source)
+auto parse(Target, Source, Flag!"doCount" doCount = No.doCount)(ref Source source)
 if (isInputRange!Source &&
     isSomeChar!(ElementType!Source) &&
     is(immutable Target == immutable bool))
@@ -2218,7 +2222,16 @@ if (isInputRange!Source &&
             static if (isNarrowString!Source)
                 source = cast(Source) s;
 
-            return result;
+            static if (doCount)
+            {
+                if (result)
+                    return tuple!("data", "count")(result, 4);
+                return tuple!("data", "count")(result, 5);
+            }
+            else
+            {
+                return result;
+            }
         }
     }
 Lerr:
@@ -2228,9 +2241,19 @@ Lerr:
 ///
 @safe unittest
 {
+    import std.typecons : Flag, Yes, No;
     auto s = "true";
     bool b = parse!bool(s);
     assert(b);
+    auto s2 = "true";
+    bool b2 = parse!(bool, string, No.doCount)(s2);
+    assert(b2);
+    auto s3 = "true";
+    auto b3 = parse!(bool, string, Yes.doCount)(s3);
+    assert(b3.data && b3.count == 4);
+    auto s4 = "falSE";
+    auto b4 = parse!(bool, string, Yes.doCount)(s4);
+    assert(!b4.data && b4.count == 5);
 }
 
 @safe unittest
@@ -2273,25 +2296,37 @@ to an integral value.
 Params:
     Target = the integral type to convert to
     s = the lvalue of an input range
+    doCount = the flag for deciding to report the number of consumed characters
 
 Returns:
-    A number of type `Target`
+$(UL
+    $(LI A number of type `Target` if `doCount` is set to `No.doCount`)
+    $(LI A `tuple` containing a number of type `Target` and a `size_t` if `doCount` is set to `Yes.doCount`))
 
 Throws:
     A $(LREF ConvException) If an overflow occurred during conversion or
     if no character of the input was meaningfully converted.
 */
-Target parse(Target, Source)(ref Source s)
+auto parse(Target, Source, Flag!"doCount" doCount = No.doCount)(ref Source s)
 if (isSomeChar!(ElementType!Source) &&
     isIntegral!Target && !is(Target == enum))
 {
     static if (Target.sizeof < int.sizeof)
     {
         // smaller types are handled like integers
-        auto v = .parse!(Select!(Target.min < 0, int, uint))(s);
-        auto result = ()@trusted{ return cast(Target) v; }();
-        if (result == v)
-            return result;
+        auto v = .parse!(Select!(Target.min < 0, int, uint), Source, Yes.doCount)(s);
+        auto result = (() @trusted => cast (Target) v.data)();
+        if (result == v.data)
+        {
+            static if (doCount)
+            {
+                return tuple!("data", "count")(result, v.count);
+            }
+            else
+            {
+                return result;
+            }
+        }
         throw new ConvOverflowException("Overflow in integral conversion");
     }
     else
@@ -2316,6 +2351,8 @@ if (isSomeChar!(ElementType!Source) &&
             alias source = s;
         }
 
+        size_t count = 0;
+
         if (source.empty)
             goto Lerr;
 
@@ -2329,6 +2366,7 @@ if (isSomeChar!(ElementType!Source) &&
                     sign = true;
                     goto case '+';
                 case '+':
+                    ++count;
                     source.popFront();
 
                     if (source.empty)
@@ -2347,6 +2385,7 @@ if (isSomeChar!(ElementType!Source) &&
         {
             Target v = cast(Target) c;
 
+            ++count;
             source.popFront();
 
             while (!source.empty)
@@ -2362,7 +2401,7 @@ if (isSomeChar!(ElementType!Source) &&
                     // Note: `v` can become negative here in case of parsing
                     // the most negative value:
                     v = cast(Target) (v * 10 + c);
-
+                    ++count;
                     source.popFront();
                 }
                 else
@@ -2375,7 +2414,14 @@ if (isSomeChar!(ElementType!Source) &&
             static if (isNarrowString!Source)
                 s = cast(Source) source;
 
-            return v;
+            static if (doCount)
+            {
+                return tuple!("data", "count")(v, count);
+            }
+            else
+            {
+                return v;
+            }
         }
 Lerr:
         static if (isNarrowString!Source)
@@ -2388,9 +2434,14 @@ Lerr:
 ///
 @safe pure unittest
 {
+    import std.typecons : Flag, Yes, No;
     string s = "123";
     auto a = parse!int(s);
     assert(a == 123);
+
+    string s1 = "123";
+    auto a1 = parse!(int, string, Yes.doCount)(s1);
+    assert(a1.data == 123 && a1.count == 3);
 
     // parse only accepts lvalues
     static assert(!__traits(compiles, parse!int("123")));
@@ -2400,6 +2451,7 @@ Lerr:
 @safe pure unittest
 {
     import std.string : tr;
+    import std.typecons : Flag, Yes, No;
     string test = "123 \t  76.14";
     auto a = parse!uint(test);
     assert(a == 123);
@@ -2409,6 +2461,17 @@ Lerr:
     auto b = parse!double(test);
     assert(b == 76.14);
     assert(test == "");
+
+    string test2 = "123 \t  76.14";
+    auto a2 = parse!(uint, string, Yes.doCount)(test2);
+    assert(a2.data == 123 && a2.count == 3);
+    assert(test2 == " \t  76.14");// parse bumps string
+    test2 = tr(test2, " \t\n\r", "", "d"); // skip ws
+    assert(test2 == "76.14");
+    auto b2 = parse!(double, string, Yes.doCount)(test2);
+    assert(b2.data == 76.14 && b2.count == 5);
+    assert(test2 == "");
+
 }
 
 @safe pure unittest
@@ -2617,6 +2680,13 @@ Lerr:
     assertCTFEable!({ string s =  "1234abc"; assert(parse! int(s) ==  1234 && s == "abc"); });
     assertCTFEable!({ string s = "-1234abc"; assert(parse! int(s) == -1234 && s == "abc"); });
     assertCTFEable!({ string s =  "1234abc"; assert(parse!uint(s) ==  1234 && s == "abc"); });
+
+    assertCTFEable!({ string s =  "1234abc"; assert(parse!( int, string, Yes.doCount)(s) ==
+        tuple( 1234, 4) && s == "abc"); });
+    assertCTFEable!({ string s = "-1234abc"; assert(parse!( int, string, Yes.doCount)(s) ==
+        tuple(-1234, 5) && s == "abc"); });
+    assertCTFEable!({ string s =  "1234abc"; assert(parse!(uint, string, Yes.doCount)(s) ==
+        tuple( 1234 ,4) && s == "abc"); });
 }
 
 // https://issues.dlang.org/show_bug.cgi?id=13931
@@ -2642,6 +2712,9 @@ Lerr:
     }
     auto input = StrInputRange("777");
     assert(parse!int(input) == 777);
+
+    auto input2 = StrInputRange("777");
+    assert(parse!(int, StrInputRange, Yes.doCount)(input2) == tuple(777, 3));
 }
 
 // https://issues.dlang.org/show_bug.cgi?id=9621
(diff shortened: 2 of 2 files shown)


std/conv.d
```diff
@@ -2665,7 +2755,16 @@ do
...
+    {
+        static if (doCount)
+        {
+            return parse!(Target, Source, doCount)(source);
+        }
+        else
+        {
+            return parse!Target(source);
+        }
+    }
```
thewilsonator, line 2767: Just do `return parse!(Target, Source, doCount)(source);`

std/conv.d
```diff
@@ -2918,65 +3066,134 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
 
     enforce(!p.empty, bailOut());
 
+    static if (doCount)
+    {
+        size_t count = 0;
+    }
```
thewilsonator, line 3072: I think I probably better to unconditionally declare this...

std/conv.d
```diff
@@ -2918,65 +3066,134 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
...
+
     bool sign = false;
     switch (p.front)
     {
     case '-':
         sign = true;
+        static if (doCount)
+        {
+            ++count;
```
thewilsonator, line 3081: ... and unconditionally increment it...

std/conv.d
```diff
@@ -2918,65 +3066,134 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
...
+        }
         p.popFront();
         static if (isNarrowString!Source)
             source = cast(Source) p;
-        return sign ? -Target.infinity : Target.infinity;
+        static if (doCount)
+        {
+            return tuple(sign ? -Target.infinity : Target.infinity, count);
```
thewilsonator, line 3122: ... and then conditionally return the tuple (or not, if doCount is false), rather than have `static if (doCount)` all over the place. Ditto throughout.

adelavais: Thank you for the feedback! I have a question, are these last 3 comments regarding only this overload, or all the overloads?

thewilsonator: All of the overloads, sorry if that wasn't clear from `Ditto throughout`.

adelavais: Ok, thank you for the clarification!

std/conv.d
```diff
@@ -2184,7 +2187,7 @@ Note:
     All character input range conversions using $(LREF to) are forwarded
     to `parse` and do not require lvalues.
 */
-Target parse(Target, Source)(ref Source source)
+auto parse(Target, Source, Flag!"doCount" doCount = No.doCount)(ref Source source)
```
thewilsonator, line 2190: can you try 

```d
auto parse(Target, Flag!"doCount" doCount = No.doCount, Source)(ref Source source)
```
instead of 
```d
auto parse(Target, Source, Flag!"doCount" doCount = No.doCount)(ref Source source)
```
so that you only need to do 
```d
T t = parse!(T,Yes.doCount)(str);
```
rather than 
```d
T t = parse!(T, string, Yes.doCount)(str);
```
and update any explicit instantiation unit tests. I want to see if any external actually use explicit instantiation.

std/conv.d
```diff
@@ -2273,25 +2295,36 @@ to an integral value.
...
+        auto result = ()@trusted{ return cast(Target) v[0]; }();
+        if (result == v[0])
+        {
+            static if (doCount)
+            {
+                return tuple(result, v[1]);
+            }
+            else
+            {
+                return result;
```
thewilsonator, line 2325: `result == v[0]` already so this is equivalent to `return v;` 
You might want to rename `result` to `truncate` and write this as 
```d
auto truncate = ()@trusted{ return cast(Target) v[0]; }();
if (v[0] == truncate)
    return v;
```

thewilsonator (commented):
A few more things, looking good so far.

std/conv.d
adelavais: My understanding is that the difference between result and v[0] is their type, which is why I was explicitly returning result. I tried the modification and it breaks both pre-existing and newly-added tests (I was returning v and v[0] respectively).

std/conv.d
adelavais: That was a great suggestion, I addressed it.

std/conv.d
```diff
@@ -2228,9 +2240,19 @@ Lerr:
 ///
 @safe unittest
 {
+    import std.typecons : Flag, Yes, No;
```
RazvanN7, line 2243: This import duplicates the one from the global scope.

std/conv.d
```diff
@@ -2388,10 +2432,15 @@ Lerr:
 ///
 @safe pure unittest
 {
+    import std.typecons : Flag, Yes, No;
```
RazvanN7, line 2435: ditto

std/conv.d
```diff
@@ -2400,6 +2449,7 @@ Lerr:
 @safe pure unittest
 {
     import std.string : tr;
+    import std.typecons : Flag, Yes, No;
```
RazvanN7, line 2452: ditto

std/conv.d
MoonlightSentinel: The import is necessary because this unittest is  runnable example on the website. The CI compiles those outside of this module to ensure that users can run / copy&paste examples.

adelavais: Yes, also when I tested the code locally some tests were failing because I did not write these imports.

std/conv.d
adelavais: So, it looks like there are errors. How do I continue from here? Do I go back and redo the order of the parameters, or do I try to modify the code that makes it fail? If the latter, can you give me a push in the right direction? (I never used dub).

std/conv.d
RazvanN7: Makes sense, sorry for the spam

std/conv.d
thewilsonator: Hmm. Maybe 
```d
if (v[0] == result)
{
    static if (doCount) return v;
    else return result;
}
```

std/conv.d
edi33416: > can you try

> `auto parse(Target, Flag!"doCount" doCount = No.doCount, Source)(ref Source source)`

@thewilsonator this won't work as this will be a breaking change for code that was previously explicit about it's `Target` and `Source`.

I don't have an example at hand, but I'm sure there is code out there that does `parse!(Target, Source)(..)`

edi33416: > That was a great suggestion, I addressed it.

@adelavais I think you should revert the code to the original order of the template arguments.

adelavais: @edi33416 of course that instead of changing the commit I just re-pushed the same one. Fixed.

thewilsonator: > but I'm sure there is code out there that does parse!(Target, Source)(..)

I highly doubt it. `Source` is inferred by the first argument, which must always be supplied.

thewilsonator: @edi33416 the only use I can think of for specifying the source is if for some reason you want to have the source be implicitly cast to or constructed as a different source. Given that the buildkite passed before, it suggests that nobody explicitly instantiates `Source`.

adelavais: @thewilsonator but some tests did fail at the last commit, with the signature
  auto parse(Target, Flag!"doCount" doCount = No.doCount, Source)(ref Source source)
Have a look at https://buildkite.com/dlang/phobos/builds/3945#47569863-a57d-45b5-a0f2-f052ba417631

WebFreak001: > > but I'm sure there is code out there that does parse!(Target, Source)(..)
> 
> I highly doubt it. `Source` is inferred by the first argument, which must always be supplied.

@thewilsonator I think your "doubt" of being used here isn't a very good reason to introduce breaking changes. I can certainly think of reasons to do this like aliasing `alias parseInt = parse!(int, string);` (or similar) and other generic template code. If not with `parse` I have certainly done this with `to` before. I don't have any code where this is actually done but I think (especially with `parse`) this might be done more often than never.

std/conv.d
adelavais: Sorry for the radio silence, I tried this and it's still failing. It fails because the expected return value is the same type as result, not v[0].

12345swordy:
What the status of this PR?

std/conv.d
```diff
@@ -2273,25 +2295,36 @@ to an integral value.
...
 {
     static if (Target.sizeof < int.sizeof)
     {
         // smaller types are handled like integers
-        auto v = .parse!(Select!(Target.min < 0, int, uint))(s);
-        auto result = ()@trusted{ return cast(Target) v; }();
-        if (result == v)
-            return result;
+        auto v = .parse!(Select!(Target.min < 0, int, uint), Source, Yes.doCount)(s);
+        auto result = ()@trusted{ return cast(Target) v[0]; }();
```
edi33416, line 2316: Please the lambda format as `() @trusted { ... }();

edi33416: You could also write it as
`auto result = (() @trusted => cast (Target) v)();`

I think using the `() =>` syntax is prefered to the curly braces `{ ... }`, as you have a oneliner

adelavais: I used the second suggestion. Thank you for the feedback, @edi33416 !

std/conv.d
```diff
@@ -2173,9 +2174,11 @@ This overload converts a character input range to a `bool`.
 Params:
     Target = the type to convert to
     source = the lvalue of an $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
+    doCount = the flag for deciding to report the number of consumed characters
 
 Returns:
-    A `bool`
+    A `bool` if doCount is set to No.doCount
```
edi33416, line 2180: Please wrap function arguments (`doCount`, `No.doCount`) within backticks ( **\`** ), as you do with `bool`

std/conv.d
```diff
@@ -2173,9 +2174,11 @@ This overload converts a character input range to a `bool`.
...
+    A `tuple` containing a `bool` and a `size_t` if doCount is set to Yes.doCount
```
edi33416, line 2181: ditto

std/conv.d
edi33416: I'm marking this as resolved

std/conv.d
```diff
@@ -2273,25 +2295,36 @@ to an integral value.
 Params:
     Target = the integral type to convert to
     s = the lvalue of an input range
+    doCount = the flag for deciding to report the number of consumed characters
 
 Returns:
-    A number of type `Target`
+    A number of type `Target` if doCount is set to No.doCount
```
edi33416, line 2301: wrap arguments in backticks and convert it to an `$(UL)`

std/conv.d
```diff
@@ -2273,25 +2295,36 @@ to an integral value.
...
+    A `tuple` containing a number of type `Target` and a `size_t` if doCount is set to Yes.doCount
```
edi33416, line 2302: ditto

std/conv.d
edi33416: Please make this an itemized unordered list.
The syntax is:
```
/** 
  * $(UL
  *   $(LI Item 1)
  *   $(LI Item 2))
  */
```

std/conv.d
edi33416: @thewilsonator I'm guessing it can't be done as you're suggesting as it seems that the cast is important. With your suggestion the types will differ.

std/conv.d
```diff
@@ -2786,15 +2884,17 @@ do
  * Params:
  *     Target = the `enum` type to convert to
  *     s = the lvalue of the range to _parse
+ *     doCount = the flag for deciding to report the number of consumed characters
  *
  * Returns:
- *     An `enum` of type `Target`
+ *     An `enum` of type `Target` if doCount is set to No.doCount
```
edi33416, line 2890: Please add backticks and convert it to an `$(UL)`

std/conv.d
```diff
@@ -2876,15 +2993,18 @@ if (isSomeString!Source && !is(Source == enum) &&
  * Params:
  *     Target = a floating point type
  *     source = the lvalue of the range to _parse
+ *     doCount = the flag for deciding to report the number of consumed characters
  *
  * Returns:
- *     A floating point number of type `Target`
+ *     A floating point number of type `Target` if doCount is set to No.doCount
```
edi33416, line 2999: ditto

std/conv.d
```diff
@@ -3159,16 +3319,47 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source) && !is(Source == enum
...
     static if (isNarrowString!Source)
         source = cast(Source) p;
-    return sign ? -ldval : ldval;
+    static if (doCount)
+    {
+        return tuple(cast (Target) (sign ? -ldval : ldval), count);
+    }
+    else
+    {
+        return cast (Target)  (sign ? -ldval : ldval);
```
edi33416, line 3328: Looks like you have an extra space between `(Target)` and `(sign...`

std/conv.d
```diff
@@ -3461,14 +3678,16 @@ Parsing one character off a range returns the first element and calls `popFront`
 Params:
     Target = the type to convert to
     s = the lvalue of an $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
+    doCount = the flag for deciding to report the number of consumed characters
 
 Returns:
-    A character of type `Target`
+    A character of type `Target` if doCount is set to No.doCount
```
edi33416, line 3684: Please add backticks and `$(UL)`

std/conv.d
```diff
@@ -3563,14 +3819,16 @@ spells `"null"`. This function is case insensitive.
 Params:
     Target = the type to convert to
     s = the lvalue of an $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
+    doCount = the flag for deciding to report the number of consumed characters
 
 Returns:
-    `null`
+    `null` if doCount is set to No.doCount
```
edi33416, line 3825: ditto

std/conv.d
```diff
@@ -3642,11 +3934,14 @@ package void skipWS(R)(ref R r)
  *     lbracket = the character that starts the array
  *     rbracket = the character that ends the array
  *     comma = the character that separates the elements of the array
+ *     doCount = the flag for deciding to report the number of consumed characters
  *
  * Returns:
- *     An array of type `Target`
+ *     An array of type `Target` if doCount is set to No.doCount
```
edi33416, line 3940: ditto

std/conv.d
```diff
@@ -3860,11 +4228,14 @@ Lfewerr:
  *     rbracket = the character that ends the associative array
  *     keyval = the character that associates the key with the value
  *     comma = the character that separates the elements of the associative array
+ *     doCount = the flag for deciding to report the number of consumed characters
  *
  * Returns:
- *     An associative array of type `Target`
+ *     An associative array of type `Target` if doCount is set to No.doCount
```
edi33416, line 4234: ditto

std/conv.d
```diff
@@ -2218,7 +2221,16 @@ if (isInputRange!Source &&
             static if (isNarrowString!Source)
                 source = cast(Source) s;
 
-            return result;
+            static if (doCount)
+            {
+                if (result)
+                    return tuple(result, 4);
+                return tuple(result, 5);
```
edi33416, line 2228: Can you please change the code to return a named tuple?
```
return tuple!("data", "count")(result, 5);
```
Thus, the result can be accessed as `result.data` and `result.count`, making the code easier to read.
After looking at the unittests, I think this change would be a nice addition, as it will make it easier to read, rather than using `result[0]` (for data) and `result[1]` (for count). It's also less error prone; it's easy to use the wrong index, but the named version will be impossible to mistake.

edi33416 (changes requested):
I've requested some minor changes.

Please also add a [changelog](https://github.com/dlang/phobos/tree/master/changelog) entry.

adelavais:
> I've requested some minor changes.
> 
> Please also add a [changelog](https://github.com/dlang/phobos/tree/master/changelog) entry.

Thank you for the feedback, I made the modifications.

std/conv.d
```diff
@@ -4015,7 +4472,14 @@ if (isInputRange!Source && isSomeChar!(ElementType!Source))
 
     s.popFront();
 
-    return result;
+    static if (doCount)
+    {
+        return tuple(cast (dchar) result, ++count);
```
edi33416, line 4477: Forgot the named fields

edi33416 (changes requested):
Thank you for your work! Just two small nits

changelog/conv_parse_count.dd
```diff
@@ -0,0 +1,12 @@
+Return the number of characters consumed by `std.conv.parse`
+
+By setting the flag `doCount` to `Yes.doCount`, the function returns a named tuple. The tuple contains the fields `data`, of type `Target`, which contains the result of the original function, and the field `count` of type `size_t`, which contains the number of characters consumed.
```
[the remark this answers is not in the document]
adelavais: I did mention the field, but re-reading this I do realize it is phrased weirdly. I addressed this.

adelavais:
I don't know how to interpret the errors in the failing test, can someone please explain them to me?

pbackus:
Buildkite failure is due to your branch missing the commits from #7678. Rebasing should fix it.

adelavais:
> Buildkite failure is due to your branch missing the commits from #7678. Rebasing should fix it.

Thank you for the response! I made a rebase.

changelog/conv_parse_count.dd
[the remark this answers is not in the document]
adelavais: @edi33416 can you please verify if this can be marked as resolved or it still needs modifications?

std/conv.d
adelavais: @edi33416 I modified this, thank you for catching the mistake. Can this be marked as resolved?

changelog/conv_parse_count.dd
```diff
@@ -0,0 +1,12 @@
+Return the number of characters consumed by `std.conv.parse`
+
+By setting the flag `doCount` to `Yes.doCount`, the function returns a named tuple. The tuple contains the fields `data` and `count`. The field `data` is of type `Target` and it contains the result of the original function. The field `count` is of type `size_t` and it contains the number of characters consumed.
```
edi33416, line 3: Sorry for nagging, but can you please keep the number of characters per line to a maximum of 120 chars / line?

Thanks

edi33416 (changes requested):

changelog/conv_parse_count.dd
adelavais: Done! Thank you for the feedback.

edi33416 (approved):

edi33416:
@adelavais thank you for your contribution!

I've added the `72h no objection -> merge` label

edi33416:
Seems the `Darwin_64_64` test is failing with a time out
`../generated/build OS="osx" BUILD="release" MODEL="64" HOST_DMD="/Users/braddr/sandbox/at-client/release-build/dmd-2.079.0/osx/bin/dmd" CXX="c++" AUTO_BOOTSTRAP="" DOCDIR="" STDDOC="" DOC_OUTPUT_DIR="" MAKE="/Applications/Xcode.app/Contents/Developer/usr/bin/make" VERBOSE="" ENABLE_RELEASE="" ENABLE_DEBUG="" ENABLE_ASSERTS="" ENABLE_UNITTEST="" ENABLE_PROFILE="" ENABLE_COVERAGE="" DFLAGS="" dmd
timed out after 1800 seconds, step failed`

adelavais:
> @adelavais thank you for your contribution!
> 
> I've added the `72h no objection -> merge` label

Thank you!

adelavais:
> Seems the `Darwin_64_64` test is failing with a time out
> `../generated/build OS="osx" BUILD="release" MODEL="64" HOST_DMD="/Users/braddr/sandbox/at-client/release-build/dmd-2.079.0/osx/bin/dmd" CXX="c++" AUTO_BOOTSTRAP="" DOCDIR="" STDDOC="" DOC_OUTPUT_DIR="" MAKE="/Applications/Xcode.app/Contents/Developer/usr/bin/make" VERBOSE="" ENABLE_RELEASE="" ENABLE_DEBUG="" ENABLE_ASSERTS="" ENABLE_UNITTEST="" ENABLE_PROFILE="" ENABLE_COVERAGE="" DFLAGS="" dmd timed out after 1800 seconds, step failed`

I really don't know what is happening the time out errors, now all the tests pass again.

PetarKirov:
@adelavais congrats on getting this merged, thanks for the useful addition!

adelavais:
> @adelavais congrats on getting this merged, thanks for the useful addition!

Thank you! I'm glad this got merged.

PR #7643, opened 2020-09-22
Fix Issue 21267: Make std.complex work with -betterC

n8sh:


std/math.d
```diff
@@ -8886,7 +8886,16 @@ private template CommonDefaultFor(T,U)
...
+    version (D_BetterC)
+    {
+        import core.internal.string : unsignedToTempString;
+        enum CommonType!(baseT, baseU) CommonDefaultFor = mixin("10.0e-" ~
+            unsignedToTempString((min(baseT.dig, baseU.dig) + 1) / 2 + 1) ~ "L");
+    }
+    else
+    {
+        enum CommonType!(baseT, baseU) CommonDefaultFor = 10.0L ^^ -((min(baseT.dig, baseU.dig) + 1) / 2 + 1);
+    }
```
n8sh, line 8898: Without this change `std.math.isClose(real, real)` can't be used with -betterC on my machine due to needing `std.math.exp` at link time.

n8sh:
>If [compiled with -betterC](https://dlang.org/spec/betterc.html) (i.e. if using math functions from the C standard library instead of `std.math`) some operations on this type may be impure

This makes the documentation less useful since the function can't have an explicit `pure` attribute.

>and on some platforms may be calculated in `double` precision even when `T` is an extended-precision `real`.

This might be a showstopper. The two culprits here are `version (CRuntime_Microsoft)` and `version (CRuntime_UClibc)` whose `atan2`, `coshl`, `expl`, `hypotl`, `logl`, `powl`, and `sinhl` are effectively just aliases of the double-precision versions of those functions.

RazvanN7:
What is the status of this draft?

n8sh:
@RazvanN7 given the problems mentioned above it's not evident to me that this can/should be salvaged.

PR #7644, opened 2020-09-27
RedBlackTree.empty: const, .front/.back: inout

SimonN:
In `RedBlackTree`, add all possible qualifiers `const pure nothrow @safe @nogc` to `empty()` and `length()`. The class is final, thus I assume that this agrees with any user code.

Add the single qualifier `inout` to `front()` and `back()`.

### DIFF
--- a/std/container/rbtree.d
+++ b/std/container/rbtree.d
@@ -964,7 +964,7 @@ if (is(typeof(binaryFun!less(T.init, T.init))))
      * Check if any elements exist in the container.  Returns `false` if at least
      * one element exists.
      */
-    @property bool empty()
+    @property bool empty() const // pure, nothrow, @safe, @nogc: are inferred
     {
         return _end.left is null;
     }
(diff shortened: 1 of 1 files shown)


Geod24:
The class is also templated so some changes are superfluous. The `const` and `inout` additions are good though.

SimonN:
Thanks! Right, my user code will need only `const` and `inout`.

Tonight, I'll add a unittest to this PR that calls into `empty` and `length` from a `const pure nothrow @safe @nogc` context, and remove all qualifiers from `empty` and `length` that are auto-inferred -- which I think will be all four of `pure nothrow @safe @nogc`.

SimonN:
Now, I annotate only with `const` and `inout`. I test explicitly that `empty()` and `length()` still qualify as `const pure nothrow @safe @nogc`.

Already before this PR, `length()` had the single necessary qualifier `const`. Therefore this PR modifies only `empty()`, `front()`, and `back()`.

SimonN:
The unittest for all attributes is now part of the existing constness unittest.

I test all three of `empty`, `length`, and `opBinaryRight"in"` for `const pure nothrow @safe @nogc` and separately for `immutable`.

Reviewers: Requiring all qualifiers to be inferred for `opBinaryRight!"in"` is potentially up for debate. It's a search in a sorted tree for at most a single element, though, and thus not expected to allocate.

thewilsonator (approved):

SimonN:
Thanks for the quick review and merge!

PR #7645, opened 2020-09-30
Remove void initialization for member with destructor

andralex:
Having a member with a destructor initialized to `void` is clearly problematic. I think it's so certifiably problematic, the compiler could actually issue an error about it.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -4855,7 +4855,7 @@ struct DirIterator
 {
 @safe:
 private:
-    RefCounted!(DirIteratorImpl, RefCountedAutoInitialize.no) impl = void;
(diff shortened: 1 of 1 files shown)


andralex:
This may or may not be a fix for https://issues.dlang.org/show_bug.cgi?id=21250 - couldn't tell because I was unable to reproduce locally.

UplinkCoder:
@andralex 
reproduces:  https://issues.dlang.org/show_bug.cgi?id=21250 5 out 5 for me.
build ~master on everything.

andralex:
Not sure what I did wrong, did `git fetch --all` and switched to this branch. Yet github said I should create a new PR. Anyhow, closing this and opening https://github.com/dlang/phobos/pull/7646

PR #7646, opened 2020-09-30
Issue 21250 - dirEntries on non-existent directory causes assert error

andralex:
Reproduced the issue on a Linux box, problem was with throwing constructors during `RefCounted` initialization.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -5156,6 +5156,12 @@ auto dirEntries(string path, string pattern, SpanMode mode,
     assert(equal(files, result));
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=21250
+@system unittest
+{
+    import std.exception : assertThrown;
+    assertThrown!Exception(dirEntries("237f5babd6de21f40915826699582e36", "*.bin", SpanMode.depth));
+}
 
 /**
  * Reads a file line by line and parses the line into a single value or a
--- a/std/file.d
+++ b/std/file.d
@@ -4855,7 +4855,7 @@ struct DirIterator
 {
 @safe:
 private:
-    RefCounted!(DirIteratorImpl, RefCountedAutoInitialize.no) impl = void;
+    RefCounted!(DirIteratorImpl, RefCountedAutoInitialize.no) impl;
     this(string pathname, SpanMode mode, bool followSymlink) @trusted
     {
         impl = typeof(impl)(pathname, mode, followSymlink);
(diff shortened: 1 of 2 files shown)


andralex:
Hmm, so it breaks some betterC code. What is usually done in such circumstances? Should I version things around?

wilzbach:
Shouldn't this include the regression test?

andralex:
> Shouldn't this include the regression test?

done

andralex:
It occurred to me this could target stable because it fixes a regression. cc @MartinNowak

MartinNowak:
> It occurred to me this could target stable because it fixes a regression. cc @MartinNowak 

See https://github.com/dlang/phobos/pull/7646#issuecomment-701608229 on how to change the target branch.

andralex:
@MartinNowak did so, please let me know if I took the right steps. Thanks!

MartinNowak (approved):

UplinkCoder (approved):
LGTM

andralex:
I reckon no need to record 4 commits for this, so I'll change label to auto-merge from `72h no objection -> merge` because there's no `72h no objection -> merge-squash`

UplinkCoder:
@andralex you did not fix the typo ... ah well.

CyberShadow:
> I reckon no need to record 4 commits for this, so I'll change label to auto-merge from `72h no objection -> merge` because there's no `72h no objection -> merge-squash`

As I understand there are two fixes here (in `std.typecons` and in `std.file`), so it would have been better to group those into two individual commits.

andralex:
@CyberShadow I'd say it's one fix spanning two files. For full disclosure, the change in `file.d` was very related but technically not required, as it currently breaks no tests. However, it was a disaster waiting to happen.

andralex:
@CyberShadow @MartinNowak so currently the change is in stable but not in master. When will it be migrated to master? Thanks.

Geod24:
Usually Martin raises a PR when he tags something, so in this case it would be a couple weeks from now, when the first beta for v2.094.1 is made. If someone needs such a merge sooner (it happened in the past), usually that someone raises the PR with the rationale.

MartinNowak:
We could make the point release a bit sooner with this.

andralex:
Thanks folks. FWIW making this a bit sooner would help us here at Symmetry.

MoonlightSentinel:
Thanks for the swift fix @andralex

PR #7648, opened 2020-10-01
std.file: WinAPI post-conditions. tempDir: TLS in doc, check isDir.

DoctorNoobingstoneIPresume:
We check that some WinAPI post-conditions hold (e.g.: ReadFile has not read more bytes than requested).

If such a post-condition were broken, it would affect everything (e.g.: buffer overflow).

The checking is done with "(condition) || assert (0);", which is a stronger form of "assert (condition);" -- it is stronger because it is always evaluated, even in release builds (as per 2.3.4.1 in Andrei's book).

This checking clarifies that we have considered various execution paths. I (humbly) believe this can serve as a model for a safer style, including when dealing with API's which are not as "reliable" as the Windows API.

For tempDir:
    - the documentation now clarifies that the procedures and the caching are performed per-thread;
    - the POSIX version now checks not only the "exists(dir)" condition, but the stronger "isDir(dir)" condition.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -437,7 +437,10 @@ version (Windows) private void[] readImpl(scope const(char)[] name, scope const(
             const uint chunkSize = min(nNumberOfBytesToRead - totalNumRead, 0xffff_0000);
             DWORD numRead = void;
             const result = ReadFile(hFile, lpBuffer + totalNumRead, chunkSize, &numRead, null);
-            if (result == 0 || numRead != chunkSize)
+            if (result == 0)
+                return false;
+            numRead <= chunkSize || assert(0);
+            if (numRead != chunkSize)
                 return false;
             totalNumRead += chunkSize;
         }
@@ -888,6 +891,12 @@ version (Windows) private void writeImpl(scope const(char)[] name, scope const(F
     {
         cnt = (size - sum < 2^^30) ? (size - sum) : 2^^30;
         WriteFile(h, buffer.ptr + sum, cast(uint) cnt, &numwritten, null);
+
+        // https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-writefile
+        // "WriteFile sets this value [numWritten] to zero before doing any work or error checking.",
+        // therefore we no longer have to start by checking the BOOL returned by WriteFile.
+        numWritten <= cnt || assert(0);
+
         if (numwritten != cnt)
             break;
         sum += numwritten;
@@ -3534,6 +3543,7 @@ else version (DragonFlyBSD)
         while (true)
         {
             auto len = GetModuleFileNameW(null, buffer.ptr, cast(DWORD) buffer.length);
+            len <= buffer.length || assert(0);
             enforce(len, sysErrorString(GetLastError()));
             if (len != buffer.length)
                 return to!(string)(buffer[0 .. len]);
@@ -4286,7 +4296,7 @@ private void copyImpl(scope const(char)[] f, scope const(char)[] t,
 
             /++
             Reference resources: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-copyfilew
-            Because OS copyfilew handles both source and destination paths,
+            Because OS CopyFileW handles both source and destination paths,
             the GetLastError does not accurately locate whether the error is for the source or destination.
             +/
             if (!f)
@@ -5248,6 +5258,9 @@ subsequent runs will return the same string, regardless of whether
 environment variables and directory structures have changed in the
 meantime.
 
+The cache is in thread-local storage.
+Each thread will perform the procedures once and cache its own result.
+
 The POSIX `tempDir` algorithm is inspired by Python's
 $(LINK2 http://docs.python.org/library/tempfile.html#tempfile.tempdir, `tempfile.tempdir`).
 
@@ -5277,9 +5290,22 @@ string tempDir() @trusted
         version (Windows)
         {
             import std.conv : to;
-            // http://msdn.microsoft.com/en-us/library/windows/desktop/aa364992(v=vs.85).aspx
+
+            // https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppathw
+            // Around 2005, a new phrase was added to the documentation:
+            // "The maximum possible return value is MAX_PATH+1 (261).".
+            // If not for this, we might have needed to call GetTempPathW a second time.
+            // (This comment should prevent new developers from rushing to implement that.)
             wchar[MAX_PATH + 2] buf;
+
             DWORD len = GetTempPathW(buf.length, buf.ptr);
+
+            // The doc is unclear regarding the "len == buf.length" case.
+            // We can deduce from other phrases that it signals the need for one more character.
+            // But the length of buf is already greater than the maximum possible return value.
+            // Therefore, we use "<" instead of "<=" in the following comparison.
+            len < buf.length || assert(0);
+
             if (len) cache = buf[0 .. len].to!string;
         }
         else version (Posix)
@@ -5290,7 +5316,7 @@ string tempDir() @trusted
             static string findExistingDir(T...)(lazy T alternatives)
             {
                 foreach (dir; alternatives)
-                    if (!dir.empty && exists(dir)) return dir;
+                    if (!dir.empty && isDir(dir)) return dir;
                 return null;
             }
 


CyberShadow:
OK, so currently we have:

- `assert(condition)`, which is used to check for logic bugs, i.e. situations which indicate that the program reached an invalid state due to a bug within the program
- Exceptions (`throw new FooException` / `enforce`), which indicate invalid input.

I understand that you are proposing a third item here, for external APIs violating their documented contract, because it doesn't quite fit in either of these two.

The questions/observations I have with this are:

1. I think discussing the addition of a new pattern to be used in D should probably happen somewhere other than a Phobos pull request.
2. What is the practical utility of this? Can it guide compiler optimizations? Is it quantifiable?
3. The external API can technically do whatever the heck it wants, including randomly overwriting D-managed memory. Where do you draw the line of what to verify and what not?
4. `assert(0)` will produce poor error messages, worse than regular assertion failures in release builds (where it compiles to an invalid instruction). If the condition is wrong in some situations, said situations may be undebuggable for users of D software.

I think this also needs a more extensive discussion on whether the LOC/complexity cost is justified.

DoctorNoobingstoneIPresume:
Hello CyberShadow, and thank you for careful and insightful review.

The short answer to both (3) and (5) is:
I believe it is useful to add such checks when they can clarify the correctness of client code.



Example of previous code in trustedReadFile:

    if (result == 0 || numRead != chunkSize)
        return false;

This makes the reader think "Oh, so if actual-read-size is less-than or greater-than wanted-read-size, we return false... and this ends up in a FileException being thrown by the caller (readImpl). Oh, so the code is prepared to deal with the greater-than case.".

But, in truth, the code is not prepared to deal with the greater-than case (and we should probably not even try to prepare it for such a case). And the reaction for the greater-than case (indeed, as you have pointed out, if the program "survives" such a violation) is not a (catchable) Exception, but a (crashing => debugger-firing-or-core-file-dumping) Error.



We could change the previous code in trustedReadFile to use <:

    if (result == 0 || numRead < chunkSize)
        return false;

But now the reader thinks: "Oh, so we are checking equal-to and less-than, but what about greater-than ?".

In order to understand why greater-than need not be checked, the reader has to carefully review more code. In these cases, it is just the previous call and the doc for ReadFile. But:

(a) I believe there are more complex cases, as I am going to exemplify soon.
(b) Even in simple cases, it is (I believe) more clear to explicitly show (even redundantly) that we are considering all cases.

This can help us discover the (improbable) failed post-conditions of API functions.
But also (as I am going to exemplify soon) internal errors in our algorithms, not just one line above.

DoctorNoobingstoneIPresume:
The short answer to (1) is that you are of course correct.

But I am planning to use this module as the Beach Head of Normandy. If I manage to trick you into merging this, then the two of us can start patching some more modules with this new style. Others are going to find it useful and safer (I hope).

By the time the Community figures out the plan, it is going to be too late to revert all the commits.

Bwahahahahhahahahahahahahahaha ! (Fade out in evil laughter.)

DoctorNoobingstoneIPresume:
Regarding (2):

The benefit I am considering is clarification (even redundant clarification, in some cases) that the client code is correct.

Instead of:
- adding a comment ("the actual size cannot be greater than the wanted size")
or
- letting the reader figure it out (without even a comment),

we are going to:
- write the comment and replace it with an always-enabled assertion,
- with the added benefit that it crashes earlier than otherwise (with a backtrace pointing closer to the detection of the bug) in the rare-but-possible case when an API post-condition fails and in the maybe-not-so-rare case when we have an internal error in our logic.

DoctorNoobingstoneIPresume:
Regarding (5), currently I am considering this useful for client code, to clarify that all execution paths have been considered, even if this is redundant to a post-condition (contract of the callee, or of a newly-introduced callee-wrapper).

(Because the reader/reviewer no longer has to check previous code-and-doc to see that the programmer has considered all possibilities, e.g. "equal, less-than and greater-than".)

DoctorNoobingstoneIPresume:
Regarding "I understand that you are proposing a third item here, for external APIs violating their documented contract, because it doesn't quite fit in either of these two.":

The examples in the pull request (and the commit message) are incomplete, as they only check post-conditions of API functions (and the check almost immediately follows the call).

I hope that a new example of a check, not of an API function post-condition, but of the internal logic of several lines of code, might clarify the usefulness (as I see it).

DoctorNoobingstoneIPresume:
Here is an example, in readImpl for Posix.
I suggest we add the lines marked with "+".
They are meant to detect bugs in a block of code which is several lines long (not just one line long, i.e. an API function).

             if (actual == 0) break;
             size += actual;
     +        size <= upTo || assert(0, "readImpl: internal error.");
             if (size >= upTo) break;
    +        size <= result.length || assert(0, "readImpl: internal error.");
             if (size < result.length) continue;

And, again, they clarify (I think) that all cases are considered ("equal, less-than, greater-than").

DoctorNoobingstoneIPresume:
Regarding (4), perhaps we can change

    numRead <= chunkSize || assert(0);

to

    numRead <= chunkSize || assert(0, "ReadFile: failed post-condition.");

DoctorNoobingstoneIPresume:
I am sorry for so many comments and for flooding you.
But I am also happy. (-:

Thank you !
(DoctorNoobingstoneIPresume at gmail dot com)

CyberShadow:
OK, this is a bit too much for me to decide on my own. @schveiguy @andralex @n8sh What do you think about this?

> Regarding (4), perhaps we can change

IIRC, `assert(0, message)` still compiles to the same code as `assert(0)` in release builds - the message is discarded.

schveiguy:
I agree with some of the changes. But not others. For example, checking that `ReadFile` or `WriteFile` doesn't return more than the number of bytes requested is not useful. Windows will never change their API to give you back more bytes than requested. And the way it's done, `assert(0)` is never removed from the code, even on release builds. Essentially, that `assert(0)` will never be triggered, under any circumstances, it is completely wasted code space.

On the other hand, I'm OK with the code that checks that `GetTempPathW` returns less than the number of wchar requested, because the API DOES note that returning exactly the requested amount is a condition that the real value could be larger, and we are not prepared to deal with that situation. Windows could legitimately increase the max number of wchar, and our code will not handle that. Likewise for ensuring that `$TMPDIR` is not a regular file on Linux -- that could happen.

CyberShadow:
> Essentially, that `assert(0)` will never be triggered, under any circumstances, it is completely wasted code space.

I think that the arguments presented here are describing why it may be useful despite that it will never be triggered in all practical circumstances. It explicitly clarifies that it is a case that should and will never occur in practice (a bit like a `final switch` case for a `case` that can never occur).

andralex:
s/@andralex/@atilaneves/g

My opinion for what it's worth: it's not our job to debug the operating system, so it seems excessive to babysit it to the level we verify that read/write operations don't return more bytes than possible. This is more likely to throw maintainers into a loop ("wait a second, you mean that could happen? let me take a close look at the docs..." which is exactly what I've done). Also a slippery slope - where to we stop verifying a variety of other OS-provided invariants?

schveiguy:
> It explicitly clarifies that it is a case that should and will never occur in practice

This can be accomplished with a comment. But I think even that is not necessary. Perhaps one questions whether `ReadFile` can return greater than the number of elements requested, and looks it up, finds that it doesn't, and now has more knowledge about the Windows API.

andralex:
Also, where do we stop? On Linux, do we verify that the read and write primitives also return no more than requested?

This seems the kind of precedent that can be easily taken to an extreme by slow accretion from different contributors who see existing code and go like, "wait, then we should do the same on posix" etc.

I think something of this nature happened to that safe/trusted idiom. By the time we defined a robust policy for it, quite a bit of code in Phobos looked bizarre.

CyberShadow:
> Also, where do we stop?

(Asked and answered in the first and second comment)

> On Linux, do we verify that the read and write primitives also return no more than requested?

Presumably yes, see https://github.com/dlang/phobos/pull/7648#issuecomment-703052419

DoctorNoobingstoneIPresume:
> I agree with some of the changes. But not others. For example, checking that `ReadFile` or `WriteFile` doesn't return more than the number of bytes requested is not useful. Windows will never change their API to give you back more bytes than requested. And the way it's done, `assert(0)` is never removed from the code, even on release builds. Essentially, that `assert(0)` will never be triggered, under any circumstances, it is completely wasted code space.

If the main concern is efficiency of the generated code, would it be more acceptable to replace

    (actualSize <= wantedSize) || assert(0);

with

    assert(actualSize <= wantedSize);

?

On a more important subject, I wish to point out that I see the check/assert/comment useful primarily as a pre-condition for the correctness of *our* code.

(And only secondarily as a post-condition of the OS API. Although I wish to point out that the API's declared intention will not change, I agree, but bugs might appear even in *that* OS.)

(My commit message has mistakenly bragged primarily about "API post-conditions".)

> On the other hand, I'm OK with the code that checks that `GetTempPathW` returns less than the number of wchar requested, because the API DOES note that returning exactly the requested amount is a condition that the real value could be larger, and we are not prepared to deal with that situation. Windows could legitimately increase the max number of wchar, and our code will not handle that. Likewise for ensuring that `$TMPDIR` is not a regular file on Linux -- that could happen.

To clarify whether my understanding is correct: GetTempPathW might legitimately change its documented behaviour of only returning MAX_PATH+1 WCHARs (e.g. in Windows 12) ?

(I obviously agree with $TMPDIR.)

We have gathered four examples here:

- ReadFile (impossible to change spec, unlikely to have bugs);
- GetTempPathW (debatable whether possible to change spec (IIRC), unlikely to have bugs);
- $TMPDIR being a regular file instead of a directory (probably not recommended by the OS, but possible depending on user/admin);
- our internal logic (described inhttps://github.com/dlang/phobos/pull/7648#issuecomment-703052419) being affected by a bug.

However, what is common (and in my opinion useful) is not the post-conditions they check, but the clarification they bring to code that follows (the pre-conditions that they assure). For example, they might be useful for someone who reads only/mostly code that follows.

Less important: It just happens that in all these examples, the checks seemed to me cheap enough to enable them in release mode.

Perhaps we could first agree/discuss whether a comment/assert/check is useful (for clarification) ?

DoctorNoobingstoneIPresume:
> > It explicitly clarifies that it is a case that should and will never occur in practice
> 
> This can be accomplished with a comment. But I think even that is not necessary. Perhaps one questions whether `ReadFile` can return greater than the number of elements requested, and looks it up, finds that it doesn't, and now has more knowledge about the Windows API.

Less important:
ReadFile does not do that by design, but maybe by a bug. Indeed, not our job to babysit the OS API. But if it (primarily) serves our own purpose to clarify our code, then this is just extra benefit. (And not all API specs are as clear as ReadFile. Perhaps we can detect misreading of other API specs.)

More important:
I have mistakenly introduced examples of mostly checks that immediately follow API calls.

Even worse, they were well-known API's (known by all since first learning about fread in <stdio.h>).

This has created confusion.

I hope to be able to add another example soon.

andralex:
> > Also, where do we stop?
> 
> (Asked and answered in the first and second comment)

You mean https://github.com/dlang/phobos/pull/7648#issuecomment-702371638 and https://github.com/dlang/phobos/pull/7648#issuecomment-703050673? Those answers do little to put any concern at ease. What reader other than a complete newbie reasons like this? No, a reader goes: "Wait, what's with this assert in here? Do they seriously check for `read` to not return more bytes than requested? That's a definite career first. Could that happen? Let me read the docs. What could be going on? ..." It's nonsense.

> > On Linux, do we verify that the read and write primitives also return no more than requested?
> 
> Presumably yes, see [#7648 (comment)](https://github.com/dlang/phobos/pull/7648#issuecomment-703052419)

Oh God, no. Let's not do this at all. Please please please close this pull request.

DoctorNoobingstoneIPresume:
> [...] which is exactly what I've done [...].

I was hoping on the contrary to save reviewers time:

"This branch (greater-than) has been considered and it should never happen [as per API spec] -- hence the "assert" keyword (whether optimized out in release builds or not). The not-equal-to branch really means less-than, not less-than-or-greater-than.".

(The comparison words refer to obtainedSize vs requestedSize.)

And in case of API's less obvious than ReadFile, perhaps we catch not a bug on their side, but a mis-understanding (of the spec) on our side.
 
> [...] Also a slippery slope - where do we stop verifying a variety of other OS-provided invariants?

Where it ceases to bring clarity to our own, client code.

Though I admit that the benefit of clarity is highly subjective, especially on cases where the check/assert/comment (almost) immediately follows the call.

Let me see if I can come up with a more detached example "soon".

andralex:
For the sake of academic integrity, I do remember we do that kind of stuff on occasion. A search for `assert(0` found a number of interesting related instances.

* We do `assert(0)` when mmap and munmap fail unexpectedly, see https://github.com/dlang/phobos/blob/master/std/experimental/allocator/building_blocks/ascending_page_allocator.d#L62 (that and a few other lines).

* In https://github.com/dlang/phobos/blob/master/std/experimental/allocator/building_blocks/region.d#L794 we `assert(0)` if `pthread_mutex_lock` or `pthread_mutex_unlock` fail unexpectedly. A variety of application (not OS) memory corruption could cause that.

* https://github.com/dlang/phobos/blob/master/std/experimental/allocator/mallocator.d#L263 tests for an unexpected code from `posix_memalign`.

* A few `assert(0)` are inserted at the end of functions that couldn't get there if the OS does its job. I think those are necessary to get the code to compile.

andralex:
> Where it ceases to bring clarity to our own, client code.

Well speaking for myself, I have literally never seen this or even thought of it, ever. Granted, this may say a number of negative things about my lack of competence and/or overspecialization. At any rate, I can tell that when I saw this my mind went in all directions - asynchronous stuff, files being changed surreptitiously while read, network errors, etc. etc. etc. So I went over the Windows docs and pored over them.

So no, at least to me this insertion absolutely brought the opposite of clarity. 

And I can tell you surprising things like this DO happen, and THOSE are the important and interesting cases that could use more clarity. Take a look at https://github.com/dlang/phobos/blob/master/std/file.d#L353. Read an entire file. Easy... not! Initially we had the simple code that fetches the size and then reads the file in one shot. Then we got bug reports from people that their file are not being read. Turns out certain (some? many? most? all?) Unix special files report size 0, yet when you read from them they do give you data. So we had to do some heuristics, a loop, etc.

THAT's where my mind went when I read this PR. THOSE are the difficult cases. A few code dribblings when everything is easy do not add clarity, they subtract because they seed doubt in the mind of the reader.

DoctorNoobingstoneIPresume:
> > > Also, where do we stop?
> > 
> > 
> [...] What reader other than a complete newbie reasons like this? No, a reader goes: "Wait, what's with this assert in here? Do they seriously check for `read` to not return more bytes than requested? [...]

Perhaps a reader who has not opened the doc for GetTempPath(A|W) after 2004 and has therefore not remembered that the maximum returned string is MAX_PATH characters (not counting the terminator) ?

Because without that phrase in the spec, our code may seem incorrect: "Wait, why are we not checking that our buffer has been large enough ? Cause otherwise we may want to retry with a larger buffer.".
 
> > > On Linux, do we verify that the read and write primitives also return no more than requested?
> > 
> > 
> > Presumably yes, see [#7648 (comment)](https://github.com/dlang/phobos/pull/7648#issuecomment-703052419)
> 
> Oh God, no. Let's not do this at all. Please please please close this pull request.

The purpose was to verify that, across a loop spawning several lines of code, we:

- have correctly added the size of the newly-obtained chunk to the correct variable (which tracks the total obtained size), on all branches;

- have always properly prepared the size of the newly requested size (e.g. in a computation such as "min (a, max (b, c))", we have not mixed up anything);

- have not exceeded the user-specified limit ("upTo").

Yes, the pull request is not perfect and we should not hurry to merge it.

But I would like to ask whether, in an implementation of std::sort, we could agree to call std::is_sorted at the end, at least in non-optimized builds, and call std::abort on the "false" branch.

DoctorNoobingstoneIPresume:
> Perhaps a reader who has not opened the doc for GetTempPath(A|W) after 2004 and has therefore not remembered that the maximum returned string is MAX_PATH characters (not counting the terminator) ?
> 
> Because without that phrase in the spec, our code may seem incorrect: "Wait, why are we not checking that our buffer has been large enough ? Cause otherwise we may want to retry with a larger buffer.".

Example: Please see @schveiguy's first comment above (https://github.com/dlang/phobos/pull/7648#issuecomment-703128965).

(With apologies if I have mis-interpreted that comment.)

andralex:
>Perhaps a reader who has not opened the doc for GetTempPath(A|W) after 2004 and has therefore not remembered that the maximum returned string is MAX_PATH characters (not counting the terminator) ?

A good point. I recall we have a comment in there about that, which is entirely (and all that is) appropriate.

 >But I would like to ask whether, in an implementation of std::sort, we could agree to call std::is_sorted at the end, at least in non-optimized builds, and call std::abort on the "false" branch.

We have something to that extent in `assumeSorted`. Probably we return that at the end of `sort`. That's a very different matter because `sort` can - and historically did :o) - have bugs, user-defined predicate may be buggy, etc. That is part of Phobos' business. Babysitting the OS is not.

DoctorNoobingstoneIPresume:
> Yes, the pull request is not perfect [...].

Among other mistakes, it is obvious (even to me) that I have mixed up several issues which could benefit from being evaluated separately:

- benefit for clarification (subjective -- to consider: distance, in lines of code etc., from code which could affect the checked condition);

- enabling the check in release builds (or even in non-release builds--perhaps a comment is better sometimes, or even letting the reader do her homework and learn something for all life);

- documentation for thread-local storage being used for caching of tempDir;

- perhaps others. :'(

DoctorNoobingstoneIPresume:
> Well speaking for myself, I have literally never seen this or even thought of it, ever. Granted, this may say a number of negative things about my lack of competence and/or overspecialization.

Certainly not that. At most, it tells about my bad examples.

I wish not to focus on the code-above:

    ReadFile(hFile, lpBuffer + totalNumRead, chunkSize, &numRead, null);
    azzert(numRead <= chunkSize);

but on the code-below:

    azzert(totalNumRead <= totalSize);
    if (totalNumRead != totalSize) throw new FileException();
        // Not a good throw for the > case.
        // The > case should never happen, but we have to look at more code-above to see this.

(where "azzert" is either "assert" or the always-enabled-even-in-release-builds version -- not important here).

Please note that in the second example, I have changed "numRead" and "chunkSize" to "totalNumRead" and "totalSize", in an effort to draw attention that the code-above might not be as simple as a single call to "ReadFile" (but a loop).

But, indeed, this has not been the case in this pull request.

> At any rate, I can tell that when I saw this my mind went in all directions - asynchronous stuff, files being changed surreptitiously while read, network errors, etc. etc. etc. So I went over the Windows docs and pored over them.
> 
> So no, at least to me this insertion absolutely brought the opposite of clarity.

Thank you--and it is useful for me to see the reaction of other people. This is my first ever pull request and almost my first contact with open source.

> And I can tell you surprising things like this DO happen, and THOSE are the important and interesting cases that could use more clarity. Take a look at https://github.com/dlang/phobos/blob/master/std/file.d#L353. Read an entire file. Easy... not! Initially we had the simple code that fetches the size and then reads the file in one shot. Then we got bug reports from people that their file are not being read. Turns out certain (some? many? most? all?) Unix special files report size 0, yet when you read from them they do give you data. So we had to do some heuristics, a loop, etc.

Last night, I realized that the Windows version does not currently work with named pipes (because those "file" handles do not support GetFileSize).

> THAT's where my mind went when I read this PR. THOSE are the difficult cases. A few code dribblings when everything is easy do not add clarity, they subtract because they seed doubt in the mind of the reader.

I cannot argue.

DoctorNoobingstoneIPresume:
> 
> 
> > Perhaps a reader who has not opened the doc for GetTempPath(A|W) after 2004 and has therefore not remembered that the maximum returned string is MAX_PATH characters (not counting the terminator) ?
> 
> A good point. I recall we have a comment in there about that, which is entirely (and all that is) appropriate.

The current comment points to the entire documentation for GetTempPath(A|W).

I am trying to point out that the important phrase is "The maximum possible return value is MAX_PATH+1 (261).".

We can do that via a [more focused] comment (one line of text) or via an assert (one line of text) or via an always-enabled-assert (one line of text). Of course, as more experienced people see fit.

Either way, clarification is the purpose.

schveiguy:
> To clarify whether my understanding is correct: GetTempPathW might legitimately change its documented behaviour of only returning (MAX_PATH+1) WCHARs (e.g. in Windows 12) ?

`MAX_PATH` is defined in druntime as a copy of what we saw from Windows at the time it was copied. It's entirely possible for Windows 12 to change that to something bigger, and our code hasn't been updated. That's why I'd be OK with that check, as it's legitimately something that we aren't actually getting from Windows.

I am not concerned _at all_ with a reader being skeptical that the conditions have been met. He will read the code, look up the docs, prove to himself that, "yes, this is ok", and move on. I'm only concerned with conditions that might need to be checked because the API semantics can change.

I'd recommend -- remove the asserts that will never trigger, even in any future version of the OS, and leave the other stuff.

andralex:
> This is my first ever pull request and almost my first contact with open source.

Ugh. Sorry. I now feel like I pulled a Linus on you without good reason. Thanks for being gracious about it all.

DoctorNoobingstoneIPresume:
> > But I would like to ask whether, in an implementation of std::sort, we could agree to call std::is_sorted at the end, at least in non-optimized builds, and call std::abort on the "false" branch.
> 
> We have something to that extent in `assumeSorted`. Probably we return that at the end of `sort`. That's a very different matter because `sort` can - and historically did :o) - have bugs, user-defined predicate may be buggy, etc. That is part of Phobos' business. Babysitting the OS is not.

Then may I ask for opinion on the following example: https://github.com/dlang/phobos/pull/7649/files ?

andralex:
> Then may I ask for opinion on the following example: https://github.com/dlang/phobos/pull/7649/files ?

Since that includes our logic I'd cautiously say "looks good".

DoctorNoobingstoneIPresume:
[...]

> > And I can tell you surprising things like this DO happen, and THOSE are the important and interesting cases that could use more clarity. Take a look at https://github.com/dlang/phobos/blob/master/std/file.d#L353. Read an entire file. Easy... not! Initially we had the simple code that fetches the size and then reads the file in one shot. Then we got bug reports from people that their file are not being read. Turns out certain (some? many? most? all?) Unix special files report size 0, yet when you read from them they do give you data. So we had to do some heuristics, a loop, etc.
> 
> Last night, I realized that the Windows version does not currently work with named pipes (because those "file" handles do not support GetFileSize).

Here is an attempt to provide a similar fix for Windows, useful because GetFileSize is not supported by all "files":

https://github.com/dlang/phobos/pull/7650

DoctorNoobingstoneIPresume:
> `MAX_PATH` is defined in druntime as a copy of what we saw from Windows at the time it was copied. It's entirely possible for Windows 12 to change that to something bigger, and our code hasn't been updated. That's why I'd be OK with that check, as it's legitimately something that we aren't actually getting from Windows.

My personal opinion is that changing MAX_PATH might break binary compatibility (or at least make older programs not work in all scenarios in newer versions of the OS) and is not probable.

In a way, it seems (to me) more probable that ReadFile returns a larger size.

Third party device drivers expose their functionality via CreateFile(A|W), CloseHandle, ReadFile, WriteFile. If none of the layers of code (in the OS, runtime library, standard library etc.) provides a sanity check, then a faulty device driver might return such abominations.

In contrast, MAX_PATH is entirely under the control of Microsoft.

Either of us could be right/wrong. But the fact remains that our code is not prepared for such a situation. GetTempPath's returning (MAX_PATH+1 or fewer) characters is not so much a *post-condition* for the *precedent* call to OS/driver code, but (I believe) mainly a *pre-condition* for the correctness of the *subsequent* code.

But, as much as I personally wanted this check to be merged, even if for different (and maybe wrong) reasoning:

As I was writing this, I have just realized that "len <= buf.length" is tested (at least in some builds) by operator[]'s checking the bounds in the next line:

    if (len) cache = buf[0 .. len].to!string;

:'(

> I am not concerned _at all_ with a reader being skeptical that the conditions have been met. He will read the code, look up the docs, prove to himself that, "yes, this is ok", and move on. I'm only concerned with conditions that might need to be checked because the API semantics can change.

Personally, I believe that clarity for the reader of the code is important for correctness and robustness (including robustness in the face of changes).
 
> I'd recommend -- remove the asserts that will never trigger, even in any future version of the OS, and leave the other stuff.

Thank you ! I am going to try to split this in smaller commits which can easily be rejected or accepted for whatever seems useful in any of them.

DoctorNoobingstoneIPresume:
> 
> 
> > Then may I ask for opinion on the following example: https://github.com/dlang/phobos/pull/7649/files ?
> 
> Since that includes our logic I'd cautiously say "looks good".

Caution is very appropriate, as I am (obviously) attempting to build up a trap with this example. (-: And I hope to prepare another example, but I cannot right now.

In the meanwhile:

I also think the example assert looks good, but for different reasons.

Not (as much) because it is a check of the logic of the lines that have come *before*:

        immutable actual = core.sys.posix.unistd.read(fd, result.ptr + size.get,
                (min(result.length, upTo) - size).get);
        cenforce(actual != -1, name, namez);
        if (actual == 0) break;
        size += actual;
        if (size >= upTo) break;
        assert(size <= result.length); // Ensure there is no bug in the previous 5-6 lines.

but (mainly) because it allows the reader/reviewer/apprentice to understand the logic of the lines that come *after*, even/especially in isolation from the lines that have come before:

        assert(size <= result.length);
        if (size < result.length) continue;
        immutable newAlloc = size + sizeIncrement;
        result = GC.realloc(result.ptr, newAlloc.get, GC.BlkAttr.NO_SCAN)[0 .. newAlloc.get];

"Aha," says the user, "so reallocation takes place when we have reached the length -- not when we have reached-or-exceeded the length, because we cannot exceed the length !"

Indeed, if I had a choice of inserting an empty line (to add vertical space), I would (personally) place it *before* the assert, in order for the assert-line to be glued next to the *subsequent* lines of code. I also note that the paranthesized conditions in the assert-line and in the subsequent-line are very similar.

It is my feeling (please correct me) that the more "previous" logic it tests (e.g. for the sort algorithm, there was much more "previous" logic to test), the higher the value of the check in your eyes.

And that whatever clarification of subsequent code is brought in by the comment/assert/check is somehow less important.

Just to be sure: I am not in any position to judge or say who is right and where the emphasis should be placed. I am trying to clarify for myself what is appropriate, and maybe to learn, at the expense of other programmers' time and patience. With thanks though. (-:

RazvanN7:
@DoctorNoobingstoneIPresume Please reopen if you want to pick this up.

PR #7649, opened 2020-10-03
std.file: Posix readImpl: An internal check of correctness.

DoctorNoobingstoneIPresume:
The last two lines of the loop:

    immutable newAlloc = size + sizeIncrement;
    result = GC.realloc(result.ptr, newAlloc.get, GC.BlkAttr.NO_SCAN)[0 .. newAlloc.get];

should only be executed when "size == result.length", not generally when "size >= length", though the "if" condition immediately above may indicate the general case.

The "size > result.length" case can never happen. If it does, we have an internal bug (or there is a bug in the implementation of the POSIX read function).

However, this is not obvious without looking attentively at the code (this is subjective--it might simply mean that I am slow).

I wish to suggest that we clarify by a comment/assertion/always-enabled-assertion. Currently, by an assertion.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -390,6 +390,7 @@ version (Posix) private void[] readImpl(scope const(char)[] name, scope const(FS
         if (actual == 0) break;
         size += actual;
         if (size >= upTo) break;
+        assert(size <= result.length);
         if (size < result.length) continue;
         immutable newAlloc = size + sizeIncrement;
         result = GC.realloc(result.ptr, newAlloc.get, GC.BlkAttr.NO_SCAN)[0 .. newAlloc.get];


std/file.d
```diff
@@ -390,6 +390,7 @@ version (Posix) private void[] readImpl(scope const(char)[] name, scope const(FS
         if (actual == 0) break;
         size += actual;
         if (size >= upTo) break;
+        assert(size <= result.length);
         if (size < result.length) continue;
```
RazvanN7, line 394: I agree that it is not obvious that after this line `size` needs to be equal to `result.length`, however the assert above the if makes things even more confusing. If you want to clarify things, the assert should be immediately after this if and should test for equality: `assert(size == result.length)`.

RazvanN7:

RazvanN7:
I'm going to close this as there is not activity and the impact of the change is minimal.

PR #7650, opened 2020-10-04
std.file: Windows readImpl: Some file handles do not support GetFileSize.

DoctorNoobingstoneIPresume:
The documentation for GetFileSize:

https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfilesize

states that:

"You cannot use the GetFileSize function with a handle of a nonseeking device such as a pipe or a communications device. To determine the file type for hFile, use the GetFileType function."

The user specifies the file by name, not by handle. Therefore, this fix might be useful for named pipes.

Another fix: trustedGetFileSize correctly distinguishes between obtaining a file size of 0xFFFFFFFF and failing to obtain the file size.

Thank you for considering this ! >:D<

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -401,11 +401,18 @@ version (Posix) private void[] readImpl(scope const(char)[] name, scope const(FS
 }
 
 
+// Hack to support unittesting "read" for file handles which do not support GetFileSize:
+version (Windows) private bool readImplTest = false;
+
 version (Windows) private void[] readImpl(scope const(char)[] name, scope const(FSChar)* namez,
-                                          size_t upTo = size_t.max) @safe
+                                          size_t upTo = size_t.max) @trusted
 {
     import core.memory : GC;
     import std.algorithm.comparison : min;
+    import std.conv : to;
+    import std.experimental.checkedint : checked;
+    //import std.stdio;
+
     static trustedCreateFileW(scope const(wchar)* namez, DWORD dwDesiredAccess, DWORD dwShareMode,
                               SECURITY_ATTRIBUTES *lpSecurityAttributes, DWORD dwCreationDisposition,
                               DWORD dwFlagsAndAttributes, HANDLE hTemplateFile) @trusted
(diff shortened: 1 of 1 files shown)


std/file.d
```diff
@@ -401,11 +401,18 @@ version (Posix) private void[] readImpl(scope const(char)[] name, scope const(FS
 }
 
 
+// Hack to support unittesting "read" for file handles which do not support GetFileSize:
+version (Windows) private bool readImplTest = false;
```
CyberShadow, line 405: This won't do. Can the test create a pipe instead?

std/file.d
```diff
@@ -401,11 +401,18 @@ version (Posix) private void[] readImpl(scope const(char)[] name, scope const(FS
...
+
 version (Windows) private void[] readImpl(scope const(char)[] name, scope const(FSChar)* namez,
-                                          size_t upTo = size_t.max) @safe
+                                          size_t upTo = size_t.max) @trusted
 {
     import core.memory : GC;
     import std.algorithm.comparison : min;
+    import std.conv : to;
+    import std.experimental.checkedint : checked;
+    //import std.stdio;
```
CyberShadow, line 414: Be sure to remove temporary debug code before submitting.

std/file.d
```diff
@@ -450,22 +459,112 @@ version (Windows) private void[] readImpl(scope const(char)[] name, scope const(
...
     cenforce(h != INVALID_HANDLE_VALUE, name, namez);
     scope(exit) cenforce(trustedCloseHandle(h), name, namez);
-    ulong fileSize = void;
-    cenforce(trustedGetFileSize(h, fileSize), name, namez);
-    size_t size = min(upTo, fileSize);
-    auto buf = () @trusted { return GC.malloc(size, GC.BlkAttr.NO_SCAN)[0 .. size]; } ();
+
+              ulong  fileSize     = void;
+    immutable bool   fileSizeOkay = trustedGetFileSize(h, fileSize);
+    immutable size_t maxSize      = fileSizeOkay ? min(fileSize, upTo) : upTo;
```
CyberShadow, line 467: I don't think vertical alignment like here is conformant to Phobos style. (Leading alignment is also at least indistinguishable from a mis-indented line.)

std/file.d
```diff
@@ -450,22 +459,112 @@ version (Windows) private void[] readImpl(scope const(char)[] name, scope const(
...
+
+    //writef("readFile %s: fileSize %u.\n", name, fileSize);
+
+    // Shamelessly reproducing the rest of the function from Posix version, with adaptations kept to a minimum:
```
CyberShadow, line 471: Perhaps instead of copying the code, the copy loop could be refactored into a platform-agnostic template function, with the primitive read APIs passed as small wrappers.

std/file.d
```diff
@@ -450,22 +459,112 @@ version (Windows) private void[] readImpl(scope const(char)[] name, scope const(
...
+    }
+
+    return buf[0 .. size.get];
+}
+
+@safe unittest
+{
+    import std.stdio;
+
+    enum size_t numIdenticalChars = 3989; // Dummy uncomfortable value (not a multiple of any beautiful number).
```
CyberShadow, line 537: So... a prime number?

CyberShadow (changes requested):

std/file.d
DoctorNoobingstoneIPresume: I am going to try the pipe too !

std/file.d
DoctorNoobingstoneIPresume: The writef's ! Okay, I am going to remove them !

Regarding the highlighted lines in your comment, does "temporary debug code" [also] refer to checkedint, please ?

std/file.d
DoctorNoobingstoneIPresume: Okay. It was merely a suggestion, because maybe other developers find it useful (if so, maybe the "style" is relaxed). But the lines can easily be re-indented.

std/file.d
DoctorNoobingstoneIPresume: I would love to try that ! Thank you for suggestion !

std/file.d
DoctorNoobingstoneIPresume: Indeed, the largest prime number less than 4096.

The various magical numbers which appear in the new code might not be the best possible choices. My (rather vague) plan has been to avoid working with buffers of 4096, both in the "write" code and in the "read" code. It felt like too round a number to use everywhere.

By the way:
Would it make sense to allow the caller of "std.file.read" to optionally specify parameters (e.g. chunk size, maximum slack size etc.) -- e.g. in an optional "struct" ?

DoctorNoobingstoneIPresume:
Thank you again for the careful review, @CyberShadow, and for the observations. I hope I can get back with more adjustments "soon". Have great days in the meanwhile !

std/file.d
CyberShadow: >  Would it make sense to allow the caller of "std.file.read" to optionally specify parameters (e.g. chunk size, maximum slack size etc.) -- e.g. in an optional "struct" ?

I don't think so. Reading from a non-seekable file is by itself uncommon, and the requirement to be able to tweak such settings would be even more so. For uncommon use cases, a custom implementation satisfying the particular needs is usually more appropriate. It's also rather simple to implement using e.g. `File.byChunk`.

std/file.d
CyberShadow: No, though I understand that's a candidate to be refactored away into a common read loop function.

RazvanN7:
@DoctorNoobingstoneIPresume What is the status on this?

RazvanN7:

RazvanN7:
Closing this as the author seems to have abandoned it. @DoctorNoobingstoneIPresume please reopen if you revisit this.

PR #7651, opened 2020-10-07
fix real precision use for inverse and hyperbolic trig, scalbn, fdim, fma, pow

thewilsonator:
part of the fix for https://issues.dlang.org/show_bug.cgi?id=18559
Previous PR: https://github.com/dlang/phobos/pull/7463

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -4035,6 +4035,8 @@ float scalbn(float x, int n) @safe pure nothrow @nogc { return _scalbn(x,n); }
 {
     assert(scalbn(0x1.2345678abcdefp0L, 999) == 0x1.2345678abcdefp999L);
     assert(scalbn(-real.infinity, 5) == -real.infinity);
+    assert(scalbn(2.0,10) == 2048.0);
+    assert(scalbn(2048.0f,-10) == 2.0f);
 }
 
 private F _scalbn(F)(F x, int n)
--- a/std/math.d
+++ b/std/math.d
@@ -1420,19 +1420,13 @@ private F _tanh(F)(F x)
  *    $(SV  +$(INFIN),+$(INFIN))
  *  )
  */
-real acosh(real x) @safe pure nothrow @nogc
-{
-    if (x > 1/real.epsilon)
-        return LN2 + log(x);
-    else
-        return log(x + sqrt(x*x - 1));
-}
+real acosh(real x) @safe pure nothrow @nogc { return _acosh(x); }
 
 /// ditto
-double acosh(double x) @safe pure nothrow @nogc { return acosh(cast(real) x); }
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -1322,39 +1322,45 @@ float cosh(float x) @safe pure nothrow @nogc  { return cosh(cast(real) x); }
...
+    import std.meta : AliasSeq;
+    static foreach (F; AliasSeq!(float, double, real))
+    {
+        assert(isIdentical(sinh(F(0.0)), F(0.0)));
+        assert(sinh(F(1.0)).approxEqual(F(sinh1)));
+    }
 }
 
-@safe @nogc nothrow unittest
+private F _sinh(F)(F x)
```
ibuclaw, line 1345: Similar functions have just gone with the naming convention `sinhImpl`, but not bothered either way.

ibuclaw (approved):

std/math.d
```diff
@@ -4010,35 +4022,47 @@ real modf(real x, ref real i) @trusted nothrow @nogc
  *      $(TR $(TD $(PLUSMN)0.0)      $(TD $(PLUSMN)0.0) )
  *      )
  */
-real scalbn(real x, int n) @safe pure nothrow @nogc
+real scalbn(real x, int n) @safe pure nothrow @nogc { return _scalbn(x,n); }
+
+///
+double scalbn(double x, int n) @safe pure nothrow @nogc { return _scalbn(x,n); }
```
thewilsonator, line 4028:
```
std/math.d(4025:6)[warn]: Public declaration 'scalbn' has no documented example.
std/math.d(4028:8)[warn]: Public declaration 'scalbn' has no documented example.
```
Huh?

std/math.d
```diff
@@ -4010,35 +4022,47 @@ real modf(real x, ref real i) @trusted nothrow @nogc
...
+///
+float scalbn(float x, int n) @safe pure nothrow @nogc { return _scalbn(x,n); }
+
+///
+@safe pure nothrow @nogc unittest
+{
+    assert(scalbn(0x1.2345678abcdefp0L, 999) == 0x1.2345678abcdefp999L);
+    assert(scalbn(-real.infinity, 5) == -real.infinity);
+    assert(scalbn(2.0,10) == 2048.0);
+    assert(scalbn(2048.0f,-10) == 2.0f);
```
thewilsonator, line 4039: Why do these not count?

std/math.d
wilzbach: You need to use `ditto` to group functions.

wilzbach (commented):
-

std/math.d
thewilsonator: Ahh, thanks!

PR #7653, opened 2020-10-08
[stable] allocator_list unittests: Use core.memory.pageSize instead of minimumPageSize

kinke:
Nothing else in druntime/Phobos uses the problematic `minimumPageSize` symbol. Problematic because the page size changes from 4 to 16 KB with macOS 11.0. See https://github.com/dlang/druntime/blob/c27346a7f471b1e31194c71e962c44a0d28a2c85/src/core/memory.d#L168-L179.


p0nce:
I can confirm on the DTK this change is making the `allocator_list.d` unittest pass in x86_64 and arm64 arch. Where it would crash o, both previously.

thewilsonator (approved):

PR #7654, opened 2020-10-09
[LDC] std.math move pragma(inline, true) out of function bodies

thewilsonator:
cc @kinke


Geod24:
No diff ?

thewilsonator:
Oh, guess someone else must have all the diff in another PR.

PR #7655, opened 2020-10-09
Fix issue 13930, 19345 - Fix `receiveOnly` for non-assignable types 

tom-tan:
This request fixes the following two issues:
- Fix [issue 13930](https://issues.dlang.org/show_bug.cgi?id=13930) - std.concurrency can't send immutable AA to another thread
- Fix [issue 19345](https://issues.dlang.org/show_bug.cgi?id=19345) - std.concurrency does not work with structs of const value type

I used a different issue title for this request from the above two issues because the source of the problem is about `receiveOnly` rather than `send` function or `std.concurrency` itself.

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -802,7 +802,17 @@ do
 
     thisInfo.ident.mbox.get((T val) {
         static if (T.length)
-            ret.field = val;
+        {
+            static if (isAssignable!T)
+            {
+                ret.field = val;
+            }
+            else
+            {
+                import core.lifetime : emplace;
+                emplace(&ret, val);
+            }
+        }
     },
     (LinkTerminated e) { throw e; },
     (OwnerTerminated e) { throw e; },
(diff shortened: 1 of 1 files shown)


thewilsonator:
please rewrite the commit message to say `fixes issues 13930, 19345` so the bot picks them up. (note the cross next to auto-close).

tom-tan:
@thewilsonator Thank you for your advice. I updated the commit message.

thewilsonator (approved):

tom-tan:
Thanks!

Geod24:
This fails on Win32:
```
"D:\a\1\dmd\generated\Windows\RelWithAsserts\Win32\dmd.exe" -conf= -g -m32mscoff -O -w -preview=dip1000 -transition=complex -I../druntime\import -unittest -version=StdUnittest -version=CoreUnittest -c  -ofunittest3c.obj std\signals.d  std\meta.d  std\typetuple.d  std\traits.d  std\encoding.d  std\xml.d  std\random.d  std\exception.d  std\compiler.d  std\system.d  std\concurrency.d
std\variant.d(703): Error: cannot modify struct instance `*p` of type `Aggregate` because it contains `const` or `immutable` members
std\variant.d(604): Error: template instance `std.variant.VariantN!20u.VariantN.opAssign!(const(Aggregate))` error instantiating
std\concurrency.d(126):        instantiated from here: `__ctor!(const(Aggregate))`
std\concurrency.d(660):        instantiated from here: `__ctor!(const(Aggregate))`
std\concurrency.d(650):        instantiated from here: `_send!(const(Aggregate))`
std\concurrency.d(627):        instantiated from here: `_send!(const(Aggregate))`
std\concurrency.d(2756):        instantiated from here: `send!(const(Aggregate))`
```

I have no idea why it doesn't fail on other platform, but this needs to be reverted. I'm also going to set Azure as required so that the bot doesn't merge failing PRs. Sorry @tom-tan , could you re-submit a PR after the revert ?

tom-tan:
@Geod24 Done! See #7661.

PR #7661, opened 2020-10-13
Fix issue 13930, 19345 - Fix `receiveOnly` for non-assignable types

tom-tan:
This is a re-submit version of #7655.

### DIFF
--- a/std/variant.d
+++ b/std/variant.d
@@ -700,7 +700,15 @@ public:
                 {
                     alias UT = Unqual!T;
                     auto p = new UT;
-                    *p = rhs;
+                    static if (isAssignable!UT)
+                    {
+                        *p = rhs;
+                    }
+                    else
+                    {
+                        import core.lifetime : emplace;
+                        emplace(p, rhs);
+                    }
                 }
                 memcpy(&store, &p, p.sizeof);
             }
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -802,7 +802,17 @@ do
 
     thisInfo.ident.mbox.get((T val) {
         static if (T.length)
-            ret.field = val;
+        {
+            static if (isAssignable!T)
+            {
+                ret.field = val;
+            }
+            else
+            {
+                import core.lifetime : emplace;
+                emplace(&ret, val);
+            }
+        }
     },
     (LinkTerminated e) { throw e; },
     (OwnerTerminated e) { throw e; },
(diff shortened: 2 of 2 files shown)


thewilsonator (approved):

thewilsonator:
Looks all green this time, thanks!

kinke:
This might cause random segfaults: https://github.com/dlang/dmd/pull/11868#issuecomment-709309172

From a quick glance, the `emplace` usage seems invalid, as that's for *uninitialized* destination memory only.

tom-tan:
This request uses `emplace` in two places. In the both cases, I used `emplace` as a workaround to initialize `Variant` with a struct with `const` members. Both cases internally need an assignment of the struct with `const` members

- The first is in [std.concurrency.receiveOnly](https://github.com/dlang/phobos/blob/6224cb03a5e272ffc6f0b95b2a8fff498925cc36/std/concurrency.d#L813) to initialize the variable with the received value in the `MessageBox`.
  It internally initializes the value of `Tuple!T ret` after the variable declaration. I used `emplace` because `ret` cannot be initialized directly due to its variable scope.

  I have no idea for better solutions for it :-(

- The second is in [std.variant.Variant.opAssign](https://github.com/dlang/phobos/blob/6224cb03a5e272ffc6f0b95b2a8fff498925cc36/std/variant.d#L710) called in `Variant.this`. It internally allocates the memory for the struct (in the case of large size of struct) using `new` and uses `emplace` to initialize its value.

  Using `core.memory.malloc` instead of `new` might be a better alternative.

Do you have any ideas or thoughts?

std/variant.d
```diff
@@ -700,7 +700,15 @@ public:
...
                     auto p = new UT;
-                    *p = rhs;
+                    static if (isAssignable!UT)
+                    {
+                        *p = rhs;
+                    }
+                    else
+                    {
+                        import core.lifetime : emplace;
+                        emplace(p, rhs);
```
kinke, line 710: AFAIK, this requires the type `UT` to feature a constructor taking a `T rhs` lvalue argument, and should thus be equivalent to `auto p = new UT(rhs)`. - I think that's not suited in general and should probably be replaced by a manual copy-emplace as used for the `T.sizeof <= size` case above. It cheats around immutability by faking a copy-construct, performing a bitcopy and invoking the postblit if necessary. By the looks of it, it currently seems to ignore copy ctors...

kinke: => https://github.com/dlang/druntime/pull/3239

pbackus:
This PR introduced a regression:

https://issues.dlang.org/show_bug.cgi?id=21663

PR #7663, opened 2020-10-14
Fix CommonType template to no longer assume typeof(type) == _error_

UplinkCoder:
This enables us to introduce a type of types in the future

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -4975,6 +4975,10 @@ template CommonType(T...)
             alias CommonType = T[0];
         }
     }
+    else static if (T.length == 2 && is(typeof(true ? T[0].init : T[1].init) U))
+    {
+        alias CommonType = U;
+    }
     else static if (is(typeof(true ? T[0].init : T[1].init) U))
     {
         alias CommonType = CommonType!(U, T[2 .. $]);


UplinkCoder:
@andralex you might be interested in this one.

UplinkCoder:
The build seems to fail because of template emission bugs ...

thewilsonator:
A fix should come with a test case, I cannot tell from the context why the clause that follows does not already cover the case added.

atilaneves:
What @thewilsonator said.

UplinkCoder:
There can't be a testcase unless typefunctions are in a preview switch

atilaneves:
Then there's no reason for this to exist right now.

UplinkCoder:
It'd be nicer to fix it for the case before.
Because that makes it more general

Geod24:
> There can't be a testcase unless typefunctions are in a preview switch

Then raise a PR with a functioning preview switch for it.
It's impossible for someone to review a change based on a bug they haven't see, or a feature they don't know about.
On the other hand, it's very easy to envision that things will work one way, and realize at the end that they didn't.

If you raise small and intermediate PRs, that's only useful if they have intrinsic values, or if they implement a small part of a much larger feature that is already visible to the reviewer.

PR #7665, opened 2020-10-18
math: remove unused code inside of unittest block

ljmf00:
This code affects code coverage and shouldn't be here at all.

### DIFF
--- a/std/internal/math/biguintcore.d
+++ b/std/internal/math/biguintcore.d
@@ -2871,18 +2871,21 @@ pure nothrow @safe
 
 @system unittest
 {
-    import core.stdc.stdio;
-
-    void printBiguint(const uint [] data)
+    version (none)
     {
-        char [] buff = biguintToHex(new char[data.length*9], data, '_');
-        printf("%.*s\n", cast(int) buff.length, buff.ptr);
-    }
+        import core.stdc.stdio;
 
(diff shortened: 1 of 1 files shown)


thewilsonator:
I presume that  is there to aid debugging, could you pease wrap that in `version(none)` instead.

ljmf00:
> I presume that is there to aid debugging, could you pease wrap that in `version(none)` instead.

Force pushed :smile: . Can you check now?

thewilsonator (approved):

std/internal/math/biguintcore.d
```diff
@@ -2871,18 +2871,21 @@ pure nothrow @safe
 
 @system unittest
 {
-    import core.stdc.stdio;
-
-    void printBiguint(const uint [] data)
+    version(none)
```
thewilsonator, line 2874:
```
Enforce whitespace before opening parenthesis
  | grep -nrE "\<(for\|foreach\|foreach_reverse\|if\|while\|switch\|catch\|version)\(" $(find etc std -name '*.d') ; test $? -eq 1
  | std/internal/math/biguintcore.d:2874:    version(none)
 ```

This needs to be `version (none)`

ljmf00: Amended. Sorry for the inconvenience.

PR #7667, opened 2020-10-19
[LDC] move `pragma(inline, true);` out of function bodies

thewilsonator:
cc @kinke

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -374,12 +374,14 @@ if ((is(immutable Num == immutable short) || is(immutable Num == immutable byte)
  * Bugs:
  *      Results are undefined if |x| >= $(POWER 2,64).
  */
-
-real cos(real x) @safe pure nothrow @nogc { pragma(inline, true); return core.math.cos(x); }
+pragma(inline, true)
+real cos(real x) @safe pure nothrow @nogc { return core.math.cos(x); }
 ///ditto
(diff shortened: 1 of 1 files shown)


Geod24:
Can you expand on the rationale ?

thewilsonator:
see https://github.com/dlang/phobos/pull/7463#discussion_r502157278 and https://github.com/dlang/phobos/pull/7463#discussion_r502180674

Geod24:
Thanks. A mention in the commit message would have been great, too, so people can get this information offline / without having to go through 2 indirections :)

std/math.d
```diff
@@ -4038,11 +4048,9 @@ float scalbn(float x, int n) @safe pure nothrow @nogc { return _scalbn(x,n); }
     assert(scalbn(2.0,10) == 2048.0);
     assert(scalbn(2048.0f,-10) == 2.0f);
 }
-
+pragma(inline, true)
```
Geod24, line 4051: Please re-add the newline

Geod24 (approved):
LGTM. Just a minor formatting comment. Feel free to self-merge once fixed.

Geod24:
@WalterBright : The LDC developers raise an important point. To allow for cross-module inlining, `pragma(inline)` should be outside of the function body, otherwise the compiler will have to analyze body of functions for non-root modules.
Should we consider deprecating usage of `pragma(inline)` inside of function body, and restrict it to be on the function's prototype ?

PR #7669, opened 2020-10-20
std.complex: Add inverse, hyperbolic, and inverse hyperbolic trigonometric functions.

ibuclaw:
Implements complex versions of the trig functions:
- asin, sinh, asinh
- acos, cosh, acosh
- atan, tanh, atanh

This finishes off the last of what's missing in `std.complex` for a relatively complete replacement of `creal` and friends.


ibuclaw:
As I imagine @kinke will no doubt be relaxing some of the assertions in the future, I'll run it through gdc on arm as well.

thewilsonator (approved):

std/complex.d
```diff
@@ -973,6 +973,179 @@ Complex!T tan(T)(Complex!T z) @safe pure nothrow @nogc
...
+    const T x = 1.0 - re2 - z.im * z.im;
+
+    T num = z.im + 1.0;
+    T den = z.im - 1.0;
+
+    num = re2 + num * num;
+    den = re2 + den * den;
+
+    return Complex!T(0.5 * std.math.atan2(2.0 * z.re, x),
+                     0.25 * std.math.log(num / den));
```
kinke, line 1026: It's probably better to use `T(0.5)` etc. for the literals, to prevent temporary double-precision for `T == float`.

ibuclaw: Unless the compiler has learnt a new trick, I don't think that really makes a difference, as internally it's all `real` anyway.

ibuclaw:
> I'll run it through gdc on arm as well.

Looks like on aarch64 there's a 128-bit real bug with `std.math.log`.  Something isn't quite right with the log coefficients.

std/complex.d
kinke: I don't know what you mean by 'internally' in this context. This doesn't even compile for `float`, due to ambiguous `atan2` call:
```
Error: std.math.atan2 called with argument types (double, const(float)) matches both:
/dlang/dmd/linux/bin64/../../src/phobos/std/math.d(1106):     std.math.atan2(real y, real x)
and:
/dlang/dmd/linux/bin64/../../src/phobos/std/math.d(1123):     std.math.atan2(float y, float x)
```

ibuclaw:
> > I'll run it through gdc on arm as well.
> 
> Looks like on aarch64 there's a 128-bit real bug with `std.math.log`. Something isn't quite right with the log coefficients.

Found it, fix in #7671.

std/complex.d
ibuclaw: Right, I didn't think of that.  I wonder how many other functions here might have issues due to lack of testing `Complex!float`.

thewilsonator: > Unless the compiler has learnt a new trick,

well `std.math.atan2` uses float now

std/complex.d
```diff
@@ -973,6 +973,187 @@ Complex!T tan(T)(Complex!T z) @safe pure nothrow @nogc
...
+    assert(isClose(acos(complex(0.5L)), PI / 3));
+    assert(isClose(acos(complex(0.5f)), float(PI) / 3));
+}
+
+/// ditto
+Complex!T atan(T)(Complex!T z) @safe pure nothrow @nogc
+{
+    static import std.math;
+    const T re2 = z.re * z.re;
+    const T x = 1.0 - re2 - z.im * z.im;
```
kinke, line 1019: See https://d.godbolt.org/z/Efrqf6 as to why `1.0` is still bad here.

ibuclaw: Using integers where possible it is then.

std/complex.d
```diff
@@ -973,6 +973,186 @@ Complex!T tan(T)(Complex!T z) @safe pure nothrow @nogc
...
+
+/**
+    Inverse hyperbolic trigonometric functions on complex numbers.
+
+    Params: z = A complex number.
+    Returns: The hyperbolic arcsine, arccosine and arctangent of `z`, respectively.
+*/
+Complex!T asinh(T)(Complex!T z)  @safe pure nothrow @nogc
+{
+    auto t = Complex!T((z.re - z.im) * (z.re + z.im) + 1.0, 2.0 * z.re * z.im);
```
kinke, line 1102: Still 2 double literals.

std/complex.d
```diff
@@ -973,6 +973,186 @@ Complex!T tan(T)(Complex!T z) @safe pure nothrow @nogc
...
+    assert(isClose(asin(complex(0.5L, 0)), PI / 6));
+    assert(isClose(asin(complex(0.5f, 0)), float(PI) / 6));
+}
+
+/// ditto
+Complex!T acos(T)(Complex!T z)  @safe pure nothrow @nogc
+{
+    static import std.math;
+    auto as = asin(z);
+    return Complex!T(std.math.PI_2 - as.re, as.im);
```
kinke, line 1002: `T(std.math.PI_2)`

ibuclaw:
DMD produces inaccurate code-gen on x86 by the look of the testsuite results (not reproducible with gdc using -m32/-mx32/-m64).

I'm not sure whether it was @kinke's suggestion of removing double literals that did it.

12345swordy:
What reviewer is this PR waiting on, for this to get merge?

quickfur:
It's failing the autotester. Generally it won't be merged unless it passes all checks.

ibuclaw:
> What reviewer is this PR waiting on, for this to get merge?

I'm not going to bother fixing dmd to emit more precise math instructions, so will just have to relax the tests to accommodate for dmd.

ibuclaw:
FYI, the reduced test has been reported in this issue https://issues.dlang.org/show_bug.cgi?id=21376

12345swordy:
@WalterBright can assist @ibuclaw on the code generation issues so that this can be merge.

12345swordy:
@ibuclaw can you rebase this?

ibuclaw:
Still blocked by dmd bugs, so I'll have more relaxed tests in `version (DigitalMars)` blocks.

ibuclaw:
I've started seeing a couple 32-bit pipelines passing, so I've squashed all intermediary commits.

Geod24 (approved):

PR #7670, opened 2020-10-20
Help compiler deduce `pure` for moveEmplaceImpl

JohanEngelen:
This fixes https://issues.dlang.org/show_bug.cgi?id=21202 .

This puts no extra requirements on `moveEmplaceImpl`. `moveEmplaceImpl` already has to be `pure` because `moveEmplaceImpl` is called by `moveEmplace` which is explicitly marked `pure`.

### DIFF
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -1534,7 +1534,7 @@ unittest
 
     Strukt a;
     Strukt b;
-    moveEmplace(a, b);    
+    moveEmplace(a, b);
 }
 
 // moveAll
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -1400,7 +1400,7 @@ private T trustedMoveImpl(T)(ref T source) @trusted
     move(x, x);
 }
 
-private void moveEmplaceImpl(T)(ref T source, ref T target)
+private void moveEmplaceImpl(T)(ref T source, ref T target) pure
 {
     import core.stdc.string : memcpy, memset;
     import std.traits : hasAliasing, hasElaborateAssign,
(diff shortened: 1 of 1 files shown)


MoonlightSentinel (changes requested):
> This puts no extra requirements on moveEmplaceImpl. moveEmplaceImpl already has to be pure because moveEmplaceImpl is called by moveEmplace which is explicitly marked pure.

But breaks `move` which doesn't have this requirement and calls user-defined `opAssign`/`opPostMove`

kinke:
I'm shocked to see this stuff is still in `std.algorithm.mutation` - hasn't this all been moved to `core.lifetime`?!

wilzbach:
Yes, but we couldn't remove this as there was one compile-time check that pulled in the world (about 2k of additional templates to druntime)

wilzbach:
Details: https://github.com/dlang/phobos/pull/6848

std/algorithm/mutation.d
```diff
@@ -1517,6 +1517,26 @@ pure nothrow @nogc @system unittest
     f(move(ncarray));
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=21202
+unittest
```
edi33416, line 1521: Can you please add attributes for the unittest block?

edi33416:
What stops the compiler from inferring the `pure`ness of `moveEmplaceImpl`?

edi33416:
> But breaks `move` which doesn't have this requirement and calls user-defined `opAssign`/`opPostMove`

What if you put the function body in a `pure` lambda? Then the compiler should be able to infer the purity and the change won’t break `move`

RazvanN7:
@JohanEngelen Why is this necessary? The compiler should be able to deduce the purity of that function.

JohanEngelen:
> @JohanEngelen Why is this necessary? The compiler should be able to deduce the purity of that function.

See https://issues.dlang.org/show_bug.cgi?id=21202

JohanEngelen:
(based on some of the comments, I am no longer convinced this is the correct fix, but the issue still stands)

RazvanN7:
@JohanEngelen It is bad practice to force an attribute on a template (especially one that potentially calls user-defined functions). This is more likely a compiler bug.

I'm going to close this PR and investigate this to get to the bottom of it.

I will close this PR as this is most likely not fixing the root cause. Please reopen if you have anything else to add.

JohanEngelen:
@RazvanN7 Thanks for looking into fixing the compiler. This workaround is still needed at Weka.

RazvanN7:
@JohanEngelen I have looked into it and it looks like a compiler bug: https://issues.dlang.org/show_bug.cgi?id=21850 .

PR #7671, opened 2020-10-20
std.math: Fix log coefficiants for 128-bit implementation

ibuclaw:
- logCoeffsR is upside down.
- logCoeffsS has a `,` bug (and subsequently the wrong size).

Tested on aarch64-linux-gnu, no std.math tests are failing after applying this.


ibuclaw:
FYI @kinke

kinke:
[LDC uses LLVM intrinsics for the log family IIRC, most likely just calling C `logl`.]

dnadlinger (approved):

dnadlinger:
Huh, I thought I tested these on AArch64 a few years ago, but clearly that can't be the case – perhaps we had a bug in the unit test runner back then making this invisible… I haven't checked this on LDC, but merging anyway as I trust your GDC results.

ibuclaw:
Auto-merge toggled on

ibuclaw:
Auto-merge toggled off

ibuclaw:
Restarting auto-testers, as the network on one of the boxes is being flaky.

PR #7672, opened 2020-10-20
std.complex: Relax some more unittests for floating point precision on ARM64

ibuclaw:
One occurs due to 64-bit vs. 128-bit mismatch, the rest are most likely accounting for precision differences in FP instructions.

All tests that use `isClose` would have being previously using `approxEqual`, which is a lot more forgiving in amount of difference which is acceptable.

### DIFF
--- a/std/complex.d
+++ b/std/complex.d
@@ -1515,7 +1515,7 @@ if (isIntegral!Int)
     assert(isClose(pow(a, -3), 1.0 / (a * a * a)));
 
     auto b = complex(2.0);
-    assert(isClose(pow(b, 3), exp(3 * log(b))));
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

ibuclaw:
Azure didn't get the CI hook from GitHub, forcing it again.

ibuclaw:
Now it's @CyberShadow's CI pipeline that appears to be stuck.

CyberShadow:
> Now it's @CyberShadow's CI pipeline that appears to be stuck.

If we have a regular problem with webhooks not being delivered, I believe GitHub provides an API which allows polling for them, if that's something worth pursuing.

PR #7674, opened 2020-10-22
Fix Issue 21337: Join can iterate ranges multiple times

rmanthorpe:


wilzbach (approved):
LGTM

schveiguy:
Great catch. Part of me cringes at the bludgeon used here (changing anything that's not the target array type into the target array type), but that's a different fix.

schveiguy (approved):

PR #7675, opened 2020-10-22
Fix Issue 18157 - std.file.rmdirRecurse should be usable in @safe

ljmf00:
`rmdirRecurse` should be @safe as the cast(string) is safe in this context and
dirEntries, even though @system, it uses a RefCounted iterator which inside
will always make the reference deleted as the reference will never be passed
outside the function scope.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -4436,15 +4436,15 @@ version (Windows) @safe unittest
         $(LREF FileException) if there is an error (including if the given
         file is not a directory).
  +/
-void rmdirRecurse(scope const(char)[] pathname)
+void rmdirRecurse(scope const(char)[] pathname) @safe
 {
     //No references to pathname will be kept after rmdirRecurse,
     //so the cast is safe
-    rmdirRecurse(DirEntry(cast(string) pathname));
+    rmdirRecurse(DirEntry((() @trusted => cast(string) pathname)()));
 }
 
 /// ditto
-void rmdirRecurse(ref DirEntry de)
+void rmdirRecurse(ref DirEntry de) @safe
 {
     if (!de.isDir)
         throw new FileException(de.name, "Not a directory");
@@ -4458,11 +4458,16 @@ void rmdirRecurse(ref DirEntry de)
     }
     else
     {
-        // all children, recursively depth-first
-        foreach (DirEntry e; dirEntries(de.name, SpanMode.depth, false))
-        {
-            attrIsDir(e.linkAttributes) ? rmdir(e.name) : remove(e.name);
-        }
+        // dirEntries is @system because it uses a DirIterator with a
+        // RefCounted variable, but here, no references to the payload is
+        // escaped to the outside, so this should be @trusted
+        () @trusted {
+            // all children, recursively depth-first
+            foreach (DirEntry e; dirEntries(de.name, SpanMode.depth, false))
+            {
+                attrIsDir(e.linkAttributes) ? rmdir(e.name) : remove(e.name);
+            }
+        }();
 
         // the dir itself
         rmdir(de.name);
@@ -4474,7 +4479,7 @@ void rmdirRecurse(ref DirEntry de)
 //"rmdirRecurse(in char[] pathname)" implementation. That is needlessly
 //expensive.
 //A DirEntry is a bit big (72B), so keeping the "by ref" signature is desirable.
-void rmdirRecurse(DirEntry de)
+void rmdirRecurse(DirEntry de) @safe
 {
     rmdirRecurse(de);
 }


ljmf00:
Please tell if I'm thinking in a wrong way and there's logic that need to be added in order to make this properly @safe .

std/file.d
```diff
@@ -4444,6 +4445,7 @@ void rmdirRecurse(scope const(char)[] pathname)
 }
```
thewilsonator, line 4445: `rmdirRecurse(DirEntry(() @trusted {return cast(string) pathname}()));`

std/file.d
```diff
@@ -4444,6 +4445,7 @@ void rmdirRecurse(scope const(char)[] pathname)
...
 
 /// ditto
+@trusted
```
thewilsonator, line 4448: What needs to be `@trusted` in this function?

thewilsonator (commented):
The general idea is to have `@trusted` as narrow in scope as possible

std/file.d
ljmf00: `dirEntries` is @system, so that's why it needs @trusted attribute. In this context, I consider this being safe because `dirEntries` returns a RefCounted wrapped variable, that, inside this will always be eliminated, so it doesn't make much sense to be unsafe here, IMHO.

The unsafe part is here:
```d
foreach (DirEntry e; dirEntries(de.name, SpanMode.depth, false))
{
    attrIsDir(e.linkAttributes) ? rmdir(e.name) : remove(e.name);
}
```

thewilsonator:
Does this have an issue number?

std/file.d
thewilsonator: Please address this and mark `void rmdirRecurse(scope const(char)[] pathname)` `@safe`

std/file.d
thewilsonator: is it the call to dirEntries, or the iteration or both? What about `rmdir`/`remove`?

wilzbach: Can't we use a local @trusted lambda + a comment to make sure nothing unsafe gets added in the future?

wilzbach (commented):
.

std/file.d
ljmf00: > is it the call to dirEntries, or the iteration or both? What about `rmdir`/`remove`?

both `rmdir` and `remove` uses a @trusted `trustedRmdir` and `removeImpl` respectively, so, when called it's both @safe calls. 

> Can't we use a local @trusted lambda + a comment to make sure nothing unsafe gets added in the future?

Sure, I'll do it.

Also, if you follow this example here, https://github.com/dlang/phobos/blob/453faadf5bc25d2f4f4ecd81133f1a9317564bde/std/datetime/timezone.d#L2451 , they are also using `dirEntries` in a @trusted function, so maybe it should be a good idea to wrap it too. I'll do it in another PR.

std/file.d
ljmf00: Done

ljmf00:
> Does this have an issue number?

Yes, https://issues.dlang.org/show_bug.cgi?id=18157 . I'll change the PR title.

Also, I don't think the dependency there to https://issues.dlang.org/show_bug.cgi?id=18155 is necessary, because, since `dirEntries` are using a RefConted object to increase performance and to not rely on GC.

ljmf00:
My fault, I referenced the wrong issue. Already updated the title again.

std/file.d
```diff
@@ -4474,6 +4479,7 @@ void rmdirRecurse(ref DirEntry de)
 //"rmdirRecurse(in char[] pathname)" implementation. That is needlessly
 //expensive.
 //A DirEntry is a bit big (72B), so keeping the "by ref" signature is desirable.
+@safe
```
edi33416, line 4482: Can you put the function attribute inline with the function declaration, as you previously did?

edi33416 (changes requested):
Can it also be marked as `pure`?

std/file.d
ljmf00: Sure

ljmf00:
> Can it also be marked as `pure`?

No. This function can't be pure in any way because it calls impure functions and this perform I/O.

std/file.d
ljmf00: Done

edi33416 (approved):
lgtm

std/file.d
```diff
@@ -4458,11 +4458,16 @@ void rmdirRecurse(ref DirEntry de)
...
     else
     {
-        // all children, recursively depth-first
-        foreach (DirEntry e; dirEntries(de.name, SpanMode.depth, false))
-        {
-            attrIsDir(e.linkAttributes) ? rmdir(e.name) : remove(e.name);
-        }
+        // dirEntries is @system because it uses a DirIterator with a
+        // RefCounted variable, but here, no references to the payload is
+        // escaped to the outside, so this should be @trusted
```
atilaneves, line 4463: Can't this be verified by the compiler by making the `ref DirEntry de` `scope`?

ljmf00: No because `.dirEntries` is @system and `scope` will do nothing about safety here. RefCounted is @system because the implementation uses malloc/free.

ljmf00: See this interesting discussion about @safe ref counting on dlang forum: https://forum.dlang.org/thread/r8ilpu$2mk7$1@digitalmars.com

std/file.d
```diff
@@ -4458,11 +4458,16 @@ void rmdirRecurse(ref DirEntry de)
...
+        () @trusted {
+            // all children, recursively depth-first
+            foreach (DirEntry e; dirEntries(de.name, SpanMode.depth, false))
+            {
+                attrIsDir(e.linkAttributes) ? rmdir(e.name) : remove(e.name);
+            }
+        }();
```
thewilsonator, line 4470: can this be done as 
```d
// all children, recursively depth-first
foreach (DirEntry e; () @trusted { return dirEntries(de.name, SpanMode.depth, false); } ())
{
    attrIsDir(e.linkAttributes) ? rmdir(e.name) : remove(e.name);
}
```
?

ljmf00: No. That's not @safe . The `DirIterator` dtor which is @system is called outside the @trusted zone, when the reference is destroyed. And it will discard the whole purpose of this because the payload reference is escaped from the wrapped @trusted zone anyway, so it's preferable to wrap the whole foreach.

thewilsonator: OK.

thewilsonator (approved):

ljmf00:
@wilzbach the bot didn't detected the issue Fix even though merged with Fix Issue xxx https://issues.dlang.org/show_bug.cgi?id=18157. Can this be closed?

wilzbach:
The bot only parses the commit messages as they are used to assemble the changelog. That's why it's important to check whether the issue displays in its commit message. We can't do much more than closing the issue manually and likely it will end up in the changelog because of the merge commit title.

PR #7676, opened 2020-10-23
std.stdio: move std.process import inside unittest

JohanEngelen:
This moves the std.process import inside the unittest. The import is not used elsewhere, and so is unnecessary for non-unittest builds. This greatly speeds up importing std.stdio. (In a simple testcase, more than 90% of the import time of std.stdio was spent on std.process)

I've tried to keep whitespace changes to a minimum, while keeping the code readable.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -1493,8 +1493,9 @@ Removes the lock over the specified file segment.
     }
 
     version (Posix)
-    static if (__traits(compiles, { import std.process : spawnProcess; }))
     @system unittest
+    {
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

JohanEngelen:
Before:
image

After:
image

Geod24:
> I've tried to keep whitespace changes to a minimum, while keeping the code readable.

Git and Github supports ignoring whitespace, so no need for such tricks.

PR #7677, opened 2020-10-23
Add some conversions to JSONValue.get!T

CromFr:
Currently `JSONValue.get` allows to convert the stored `integer`, `uinteger` or `float_` to a floating point type, but throws a JSONException("JSONValue is not a xxx") when trying to convert a stored `uinteger` to any signed int, or a stored `integer` to any unsigned int.

This can be quite annoying as the parser sets the type to either ulong or long depending on how high is the parsed value.

This patch adds the following conversions:
- `integer` -> any signed or unsigned int
- `uinteger` -> any signed or unsigned int

Note:
- The value will be truncated if the target type is narrower than the stored type, but this seems to be expected as it is the current behaviour when converting for example a `ulong` to `ubyte`
- `JSONException`s are raised if there is a conversion overflow
- Should I also add conversions from `float_` to integer types?

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -368,11 +368,10 @@ struct JSONValue
      * A convenience getter that returns this `JSONValue` as the specified D type.
      * Note: only numeric, `bool`, `string`, `JSONValue[string]` and `JSONValue[]` types are accepted
      * Throws: `JSONException` if `T` cannot hold the contents of this `JSONValue`
-     *         `ConvException` if there is a type overflow issue
+     *         `ConvException` in case of integer overflow when converting to `T`
      */
     @property inout(T) get(T)() inout const pure @safe
     {
-        import std.conv : to;
         static if (is(immutable T == immutable string))
         {
             return str;
@@ -404,7 +403,7 @@ struct JSONValue
             case JSONType.integer:
                 return integer.to!T;
             default:
-                throw new JSONException("JSONValue is not a number type");
+                throw new JSONException("JSONValue is not a an integral type");
             }
         }
         else
--- a/std/json.d
+++ b/std/json.d
@@ -395,11 +395,31 @@ struct JSONValue
         }
         else static if (__traits(isUnsigned, T))
         {
-            return cast(T) uinteger;
+            switch (type)
+            {
+            case JSONType.uinteger:
+                return cast(T) uinteger;
+            case JSONType.integer:
+                auto v = integer;
+                enforce!JSONException(v >= 0, "JSONValue is a negative number");
+                return cast(T) v;
+            default:
+                throw new JSONException("JSONValue is not a number type");
+            }
         }
         else static if (isSigned!T)
         {
-            return cast(T) integer;
+            switch (type)
+            {
+            case JSONType.uinteger:
+                auto v = integer;
+                enforce!JSONException(v > long.max, "JSONValue positive overflow");
+                return cast(T) v;
+            case JSONType.integer:
+                return cast(T) integer;
+            default:
+                throw new JSONException("JSONValue is not a number type");
+            }
         }
         else
         {
@@ -427,7 +447,9 @@ struct JSONValue
             "c": "text",
             "d": true,
             "e": [1, 2, 3],
-            "f": { "a": 1 }
+            "f": { "a": 1 },
+            "g": -45,
+            "h": ` ~ ulong.max.to!string ~ `,
          }`;
 
         struct a { }
@@ -435,6 +457,7 @@ struct JSONValue
         immutable json = parseJSON(s);
         assert(json["a"].get!double == 123.0);
         assert(json["a"].get!int == 123);
+        assert(json["a"].get!uint == 123);
         assert(json["b"].get!double == 3.1415);
         assertThrown(json["b"].get!int);
         assert(json["c"].get!string == "text");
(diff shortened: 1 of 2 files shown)


CyberShadow:
> The value will be truncated if the target type is narrower than the stored type, but this seems to be expected as it is the current behaviour when converting for example a `ulong` to `ubyte`

I think this is a bug. It should probably use `std.conv.to`, which does a checked conversion, instead of doing an unchecked cast.

CromFr:
In this case, maybe it would be better to allow any kind of conversions that to!T accepts ?

Something like this:
```dlang
@property inout(T) get(T)() inout const pure @safe
    {
        final switch (type)
        {
        case JSONType.null_:
            throw new JSONException("JSONValue is null");
        case JSONType.string:
            static if(__traits(compiles, str.to!T))
                return str.to!T;
            else
                throw new JSONException("Cannot convert JSONValue str to " ~ T.stringof);
        case JSONType.integer:
            static if(__traits(compiles, integer.to!T))
                return integer.to!T;
            else
                throw new JSONException("Cannot convert JSONValue integer to " ~ T.stringof);
        // ...
```

CyberShadow:
Not sure about that either, unfortunately `to` is selective in what kind of lossy conversions it allows. E.g. `200.to!byte` fails but `4.3.to!int` happily returns 4.

CromFr:
I think it makes more sense to convert types only when there is an ambiguity in what the JSONParser is reading, forbid lossy conversions as you suggested.

The following conversions feels reasonable:
- `string` => string types only, using to!T
- `integer` => any integral or floating point, using to!T for overflow checking
- `uinteger` => any integral or floating point, using to!T for overflow checking
- `float_` => floating point types only, using to!T (I believe to!T will only do a cast(T))
- `array` => JSONArray[] only
- `object` => JSONArray[string] only
- `true_`, `false_` => bool only

CyberShadow:
> `integer` => any integral or floating point, using to!T for overflow checking

When converting integers to floating point, loss of precision can occur, which is also not checked by `to`.

I suggest not attempting to write something general and handle only the explicit case relevant here, which I understand is `integer`/`uinteger`.

The previous discussion is worth reviewing: https://github.com/dlang/phobos/pull/7284

CromFr:
Sorry for that, I'll try stay focused on the PR subject :/

I updated the PR to only contain changes to the integer/uinteger conversions with to!T.

For reference, the current behaviour is:
- `string` => `string` only
- `integer` => signed integers or floating point types, using cast(T)
- `uinteger` => unsigned integers or floating point types, using cast(T)
- `float_` => floating point types only, cast(T)
- `array` => JSONArray[] only
- `object` => JSONArray[string] only
- `true_`, `false_` => `bool` only

There's already a conversion loss with `get!double`:
```d
else static if (isFloatingPoint!T)
{
    switch (type)
    {
    case JSONType.float_:
        return cast(T) floating;
    case JSONType.uinteger:
        return cast(T) uinteger;
    case JSONType.integer:
        return cast(T) integer;
    default:
        throw new JSONException("JSONValue is not a number type");
    }
}
```

I suggest changing these casts with `to!T`, so there's at least some checks for overflow (e.g. `cast(ubyte)500.0` gives 244, but `500.0.to!ubyte` throws a positive overflow exception), either in this PR or another one.

std/json.d
```diff
@@ -393,13 +395,17 @@ struct JSONValue
...
-        {
-            return cast(T) integer;
+            switch (type)
+            {
+            case JSONType.uinteger:
+                return uinteger.to!T;
+            case JSONType.integer:
+                return integer.to!T;
+            default:
+                throw new JSONException("JSONValue is not a number type");
```
CyberShadow, line 407:
```suggestion
                throw new JSONException("JSONValue is not an integral type");
```

std/json.d
```diff
@@ -368,9 +368,11 @@ struct JSONValue
      * A convenience getter that returns this `JSONValue` as the specified D type.
      * Note: only numeric, `bool`, `string`, `JSONValue[string]` and `JSONValue[]` types are accepted
      * Throws: `JSONException` if `T` cannot hold the contents of this `JSONValue`
+     *         `ConvException` if there is a type overflow issue
      */
     @property inout(T) get(T)() inout const pure @safe
     {
+        import std.conv : to;
```
CyberShadow, line 375: There is a top-level `std.conv` import, so this looks redundant.

std/json.d
```diff
@@ -368,9 +368,11 @@ struct JSONValue
      * A convenience getter that returns this `JSONValue` as the specified D type.
      * Note: only numeric, `bool`, `string`, `JSONValue[string]` and `JSONValue[]` types are accepted
      * Throws: `JSONException` if `T` cannot hold the contents of this `JSONValue`
+     *         `ConvException` if there is a type overflow issue
```
CyberShadow, line 371:
```suggestion
     *         `ConvException` in case of integer overflow when converting to `T`
```

CyberShadow (commented):
Also needs a changelog entry or linked issue.

CromFr:
Thanks for your comments :)

I applied the changes you suggested, and added a changelog entry.

thewilsonator (approved):

thewilsonator:
Please rebase or forcepush to restart the CI

thewilsonator:
Could you please force push the CI seems stuck.

CromFr:
Sorry for the delay, I just rebased on master and pushed ;)

PetarKirov (approved):

PR #7678, opened 2020-10-26
regex: partially make regex pure

ljmf00:
Use pure/enforced version of `malloc` and `free` functions and fake pure the custom memoized `defaultFactory` and `matchOnce`.

### DIFF
--- a/std/regex/internal/ir.d
+++ b/std/regex/internal/ir.d
@@ -468,7 +468,8 @@ interface MatcherFactory(Char)
 // Only memory management, no compile-time vs run-time specialities
 abstract class GenericFactory(alias EngineType, Char) : MatcherFactory!Char
 {
-    import core.stdc.stdlib : malloc, free;
+    import core.memory : pureFree;
+    import std.internal.memory : enforceMalloc;
     import core.memory : GC;
     // round up to next multiple of size_t for alignment purposes
     enum classSize = (__traits(classInstanceSize, EngineType!Char) + size_t.sizeof - 1) & ~(size_t.sizeof - 1);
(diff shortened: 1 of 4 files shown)


DmitryOlshansky:
LGTM

DmitryOlshansky (approved):

jondegenhardt:
I'm seeing a small regex performance degradation and a material increase in memory use as reported by `time`. This is in LDC 1.25, so DMD 2.095. This PR looks like the main modification to std.regex in the release.

I looked through the code changes and didn't see anything real obvious. But, are there known reason why this PR could have caused these behavior changes? Is `pureMalloc` invoking a different version of malloc? My code does repeatedly call matchOnce with the same regex pattern, so it hits the cache implementation that was modified. Could the changes to the caching have resulted in a change?

PetarKirov:
@jondegenhardt `pureMalloc` = `save errno` + `malloc` + `reset errno` (ditto for `pureFree`), so it can't be the root cause. At first glance, the suspect could be the two new delegates `defaultFactoryImpl` and `matchOnceImpl`. They may or may not cause GC allocations. If they are replaced with static nested functions, they would be guaranteed not to cause GC allocations.

jondegenhardt:
> At first glance, the suspect could be the two new delegates defaultFactoryImpl and matchOnceImpl. They may or may not cause GC allocations. If they are replaced with static nested functions, they would be guaranteed not to cause GC allocations.

@PetarKirov - Thanks for looking. Sorry it took me so long to reply. There are indeed GC allocations in 2.095 that do not occur in 2.094. I can't be 100% sure yet that this is the cause, but seems likely. I'll see if I can nail it down to this change.

@ljmf00 fyi on this.

jondegenhardt:
Filed issue [21716](https://issues.dlang.org/show_bug.cgi?id=21716).

jondegenhardt:
I expect to submit a PR fixing this issue tomorrow.

std/regex/internal/ir.d
```diff
@@ -543,25 +547,34 @@ class CtfeFactory(alias EngineType, Char, alias func) : GenericFactory!(EngineTy
...
+            else
+            {
+                if (thompsonFactory is null)
+                    thompsonFactory = new RuntimeFactory!(ThompsonMatcher, Char);
+                return thompsonFactory;
+            }
+        };
+
+        // this should be faked as pure because the static mutable variables are
+        // used to cache the created instance, like memoize
```
n8sh, line 573: In what sense is it valid or at least useful to mark as pure a function that accesses mutable static data? memoize isn't pure.

PR #7680, opened 2020-10-26
std.file: DirEntry: Avoid repeated stat on failure-and-retry.

DoctorNoobingstoneIPresume:
For DirEntry, if stat (or lstat) fails, but the user [eventually] catches the exception and [eventually] retries, do we want to retry performing stat (or lstat) ?

I feel that, in some cases, this might lead to repeated "Failed to stat file ..." exceptions.

And, generally, it is difficult for me to think of a scenario where stat (or lstat) fails on the first attempt(s) and later succeeds (for the same file).

Perhaps we can clarify together.

Thank you.

(And I apologize for not having worked lately on my other pull requests. Thank you for the patience.)

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -4038,10 +4038,10 @@ else version (Posix)
             if (_didStat)
                 return;
 
+            scope(exit) _didStat = true;
+
             enforce(stat(_name.tempCString(), &_statBuf) == 0,
                     "Failed to stat file `" ~ _name ~ "'");
-
-            _didStat = true;
         }
 
         /++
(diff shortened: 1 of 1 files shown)


DoctorNoobingstoneIPresume:
Oh, no ! My logic is broken.

Even if we do not retry stat, I should not set the "[successfully] did stat" flag.

There should perhaps be an extra flag.

I also have to check the unit test...

DoctorNoobingstoneIPresume:
Plus the DirEntry might be re-used after having created the file.
Sorry !!!

PR #7681, opened 2020-10-27
stdio: optimize readln algorithm for non-char type

ljmf00:
Instead of using buf.length = 0, use buf = buf[0..0] slice technique to clean
the array and reuse, if possible, the buffer to reduce GC allocations.

---
Benchmark: https://gist.github.com/run-dlang/7d0a5264d186ac8db487447ec5352366

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -1846,14 +1846,19 @@ is recommended if you want to process a complete file.
         }
         else
         {
-            // TODO: optimize this
             string s = readln(terminator);
-            buf.length = 0;
-            if (!s.length) return 0;
-            foreach (C c; s)
+            if (!s.length)
             {
-                buf ~= c;
+                buf = buf[0 .. 0];
+                return 0;
             }
+
+            import std.utf : codeLength;
+            buf.length = codeLength!C(s);
+            size_t idx;
+            foreach (C c; s)
+                buf[idx++] = c;
+
             return buf.length;
         }
     }


thewilsonator (approved):

ljmf00:
Force pushed due to codestyle issues.

ljmf00:
Btw @wilzbach I think its better to discard approval states on active repositories when force push or new commits to avoid merge unwanted and unapproved changes.

wilzbach:
I appreciate the concern, but I think it's fine here as there isn't so much traffic. We already have the bot automatically remove auto-merge from PRs so that it's not possible for an untrusted contributor to sneak code in.

schveiguy:
> Instead of using buf.length = 0, use buf = buf[0..0] slice technique to clean
> the array and use an appender to do less GC allocations.

buf.length = 0 is *identical* to buf = buf[0 .. 0].

Not only that, but you are only doing that from within the case where s.length is 0. So you are not trying to reuse the buffer at all, right?

The original did not reuse the buffer (even though it claims to in the docs). Any speedups here are purely from using appender vs. the slower runtime append. Reusing the buffer should add a pretty significant improvement.

ljmf00:
> > Instead of using buf.length = 0, use buf = buf[0..0] slice technique to clean
> > the array and use an appender to do less GC allocations.
> 
> buf.length = 0 is _identical_ to buf = buf[0 .. 0].
> 
> Not only that, but you are only doing that from within the case where s.length is 0. So you are not trying to reuse the buffer at all, right?

```d
buf.length = 0
```
uses `core.internal.array.capacity._d_arraysetlengthTImpl` runtime function template, which internally will basically do:

```d
if (newlength <= (*p).length)
{
    *p = (*p)[0 .. newlength];
    void* newdata = (*p).ptr;
    return newdata[0 .. newlength];
}
```

which for `0` is basically `buf = buf[0..0]`. For the compiler this may be hard to optimize and at the end its the same.

> The original did not reuse the buffer (even though it claims to in the docs). Any speedups here are purely from using appender vs. the slower runtime append. Reusing the buffer should add a pretty significant improvement.

Yes, you are right, I can reuse the existing buffer space instead of reallocate a new one. Done the changes, please review.

ljmf00:
> I appreciate the concern, but I think it's fine here as there isn't so much traffic. We already have the bot automatically remove auto-merge from PRs so that it's not possible for an untrusted contributor to sneak code in.

Ok, fair :+1:

schveiguy:
> uses `core.internal.array.capacity._d_arraysetlengthTImpl` runtime function template

ugh, you are right. When did this change? I thought setting array length to 0 was optimized, I don't think it used to call a runtime function at all.

If I use AST on run.dlang.io to give me the difference between setting length to 0 and slicing, it's... disturbing.

std/stdio.d
```diff
@@ -1846,14 +1846,16 @@ is recommended if you want to process a complete file.
...
-            {
-                buf ~= c;
+            if (!s.length) {
+                buf = buf[0..0];
+                return 0;
             }
+
+            buf.length = s.length;
+            foreach (i, C c; s)
+                buf[i] = c;
```
schveiguy, line 1857: You need to use appending. This is the trans-coding path, which means you are reading `dchar` or `wchar` instead of char.

Using an appender might work, or you can use:
```d
buf = buf[0 .. 0];
buf.assumeSafeAppend;
// append same as before
```

schveiguy (changes requested):

std/stdio.d
ljmf00: Didn't knew the `.assumeSafeAppend`. Interesting, but isn't this unsafe? Also, I need to use
```d
import std.utf : codeLength;
buf.length = codeLength!C(s);
```
To determinate the size.

Also, with the following code, the performance is worst than using an appender:
```d
string s = readln(terminator);
if (!s.length)
{
    buf = buf[0..0];
    return 0;
}
import std.utf : codeLength;
buf.length = codeLength!C(s);
buf = buf[0..0];
foreach (i, C c; s)
    buf.assumeSafeAppend() ~= c;
return buf.length;
```

image
Code: https://gist.github.com/run-dlang/8e6e06e357cf4761cc5000d24fdfa4e6

schveiguy: No just call `assumeSafeAppend` once. And it's probably going to be worse than appender, because appender doesn't need to go through opaque calls to figure out whether a block has space.

But yes, you could use codeLength to make it work, and just use an incrementing index.

schveiguy: It's not unsafe, as long as you are just writing characters, and you aren't overwriting immutable data.

ljmf00: > No just call `assumeSafeAppend` once. And it's probably going to be worse than appender, because appender doesn't need to go through opaque calls to figure out whether a block has space.

Yes, it got better but worse than using an appender.

> But yes, you could use codeLength to make it work, and just use an incrementing index.

I thought that index provided by `_aApplycw2` was incremental but isn't suitable for this case because it does `decode(buf, i)` instead.

My conclusion is that using incrementing index and resizing with codeLength is the best way to go.

Order:
1. Old implementation
2. Appender
3. assumeSafeAppend + codeLength
4. incremental idx + codeLength
image

schveiguy (approved):
Looks good, and simpler. Nice!

ljmf00:
> ugh, you are right. When did this change? I thought setting array length to 0 was optimized, I don't think it used to call a runtime function at all.
> 
> If I use AST on run.dlang.io to give me the difference between setting length to 0 and slicing, it's... disturbing.

I made a PR to the compiler frontend to fix that https://github.com/dlang/dmd/pull/11912 .

PR #7687, opened 2020-10-31
Minor fixes for -preview=in and targets preferring by-ref for `real`

kinke:
Code compiled with `-preview=in` is currently supposed to link successfully against druntime/Phobos compiled without `-preview=in`.

There were still a few places where `-preview=in` makes a difference for existing `in` params; convert them to `const [scope]`, analogous to PR #7593.


thewilsonator (approved):

Geod24:
@kinke : Does that mean that LDC v1.24 is not compatible with `-preview=in` for those functions ?
As I assume it contains dlang/dmd#11828 but not this.

kinke:
I guess so, just like DMD 2.094[.1] isn't for targets with by-ref `real` (Win64, 32-bit x86 IIRC).

PR #7698, opened 2020-11-19
Make std.math.NaN and std.math.getNaNPayload work in CTFE

n8sh:


thewilsonator (approved):

ibuclaw:
Did you test this on Aarch64 or SPARC?

n8sh:
I did not test it on Aarch64 or SPARC if they're not in the CI suite. The unit tests should fail on those targets if this doesn't work on them.

PR #7699, opened 2020-11-19
Make std.math.nextUp and nextDown and nextafter work in CTFE for extended-precision real

n8sh:
PR #7176 didn't include extended-precision reals due to problems with FreeBSD in the CI suite. Opening this as a draft to see if I can get tests to pass.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -6883,6 +6883,43 @@ debug(UnitTest)
 real nextUp(real x) @trusted pure nothrow @nogc
 {
     alias F = floatTraits!(real);
+    static if (F.realFormat != RealFormat.ieeeDouble)
+    {
+        if (__ctfe)
+        {
+            if (x == -real.infinity)
+                return -real.max;
+            if (!(x < real.infinity)) // Infinity or NaN.
+                return x;
+            real delta;
+            // Start with a decent estimate of delta.
+            if (x <= 0x1.ffffffffffffep+1023 && x >= -double.max)
+            {
+                const double d = cast(double) x;
+                delta = (cast(real) nextUp(d) - cast(real) d) * 0x1p-11L;
+                while (x + (delta * 0x1p-100L) > x)
+                    delta *= 0x1p-100L;
+            }
+            else
+            {
+                delta = 0x1p960L;
+                while (!(x + delta > x) && delta < real.max * 0x1p-100L)
+                    delta *= 0x1p100L;
+            }
+            if (x + delta > x)
+            {
+                while (x + (delta / 2) > x)
+                    delta /= 2;
+            }
+            else
+            {
+                do { delta += delta; } while (!(x + delta > x));
+            }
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -7082,15 +7120,21 @@ float nextDown(float x) @safe pure nothrow @nogc
 
 @safe pure nothrow @nogc unittest
 {
-    static if (floatTraits!(real).realFormat == RealFormat.ieeeExtended)
+    static if (floatTraits!(real).realFormat == RealFormat.ieeeExtended ||
+               floatTraits!(real).realFormat == RealFormat.ieeeDouble ||
+               floatTraits!(real).realFormat == RealFormat.ieeeExtended53 ||
+               floatTraits!(real).realFormat == RealFormat.ieeeQuadruple)
```
n8sh, line 7126: Let's see if we can enable the tests for all formats except `RealFormat.ibmExtended`.

thewilsonator:
Is this good to go?

n8sh:
Think so.

std/math.d
ibuclaw: @n8sh do you need a freebsd/x86 or aarch64 machine to test?

ibuclaw:
Codecov is making this difficult to review in the UI.

std/math.d
```diff
@@ -6883,6 +6883,44 @@ debug(UnitTest)
 real nextUp(real x) @trusted pure nothrow @nogc
 {
     alias F = floatTraits!(real);
+    static if (F.realFormat != RealFormat.ieeeDouble)
+    if (__ctfe)
```
ibuclaw, line 6887: Braces/indentation?

std/math.d
n8sh: @ibuclaw that would be useful

n8sh:
>Codecov is making this difficult to review in the UI.

I don't disagree, on top of which the Codecov report is certainly wrong.

std/math.d
ibuclaw: Both pass ctfe tests without issue.

std/math.d
```diff
@@ -6883,6 +6883,44 @@ debug(UnitTest)
...
+            return x;
+        real delta;
+        // Start with a decent estimate of delta.
+        if (x <= 0x1.ffffffffffffep+1023 && x >= -double.max)
+        {
+            const double d = cast(double) x;
+            static if (real.mant_dig >= 64)
+                delta = (cast(real) nextUp(d) - cast(real) d) * 0x1p-11L;
+            else // e.g. RealFormat.ieeeExtended53
+                delta = cast(real) nextUp(d) - cast(real) d;
```
ibuclaw, line 6901: Doesn't look like this is necessary for 80-bit reals with 53-bit precision, you can just use the first branch unconditionally.

ibuclaw (approved):

PR #7702, opened 2020-11-21
Add sumtype to Phobos

pbackus:
Based on version 1.0.3 of the `sumtype` package on [code.dlang.org](https://code.dlang.org). See individual commit messages for a summary of changes.


PetarKirov:
@pbackus thank you for your work on [sumtype](https://github.com/pbackus/sumtype)! It's an excellent high-quality module that I've used successfully in various projects, as I'm sure many other D developers have. I believe that adding it to phobos (and later druntime), would be an excellent addition to D's standard distribution!

@atilaneves What's the DLF's latest position on adding new modules to phobos? Is there a checklist from your side about about what an pull request like this one should fulfill?

I see several options:
1. Review and merge this under `std.sumtype` in master after the 2.095.0-beta.1 cut-off point (expected soon™). We would have close to two months of time until the until dmd 2.096.0-beta.1 ([scheduled for 2021-01-15](https://dlang.org/changelog/release-schedule.html)) to make any large API-level changes (obviously non-breaking changes are always possible).
2. If adding a new top-level module is deemed to risky (I don't think so), we can merge it under `std.experimental.sumtype` after a short review and set a deadline for "stabilizing" after which it will become an "official" phobos module under `std.sumtype` (e.g. after 2-3 releases).
3. Announce a 2-3 weeks long RFC process on the forums, during which we ask as many members of the community to review, after we vote on a hard yes/no decision and we either merge under `std.sumtype`, or we restart the RFC process.

pbackus:
Re: stabilization, my view is that stabilization can and should happen on code.dlang.org. `std.experimental` seems in practice to have become a kind of purgatory where modules are considered too stable to break, but also not stable enough that users feel confident depending on them. I would not consider it a successful outcome if sumtype were to end up there.

atilaneves:
> @atilaneves What's the DLF's latest position on adding new modules to phobos? Is there a checklist from your side about about what a pull request like this one should fulfill?

Good question. Given that this is the first time it's happened since I became the deputy leader, I'm going to have to think about it.

Bolpat:
My latest news about adding modules to Phobos is that they need to bring great merit since removing them later is a breaking change.

@pbackus Can you outline why adding this to Phobos is a big win? Especially, why DUB isn't enough?

pbackus:
@Bolpat I'd be glad to. Here's something I wrote about this on the D forums [2] that I think summarizes the argument pretty well:

> In the C++ community, certain types are regarded as "vocabulary types"--types that "provide a single lingua franca, a common language, for dealing with [their] domain." [1] The prototypical example is std::string. What's important about these types is not just that they're available, but that they're standardized. std::string isn't just one possible choice of string class, it's the *obvious* choice.
>
> Which types benefit the most from this kind of standardization? In broad terms, types that are used for communication between different libraries or modules. These tend to be general-purpose, domain-agnostic types like strings, dynamic arrays, associative arrays, tuples, and--that's right--sum types.
>
> I don't particularly care if SumType or some other implementation is accepted into Phobos (though of course, I prefer SumType myself), but I think it's important for there to be a sum type in D that (1) meets the quality standards of today's D programmers, and (2) is the obvious choice. Currently, SumType satisfies (1), but not (2); and Algebraic satisfies (2), but not (1).
>
> Personally, I don't think it's possible to make a dub package "more obvious" than a standard library module, no matter how many shiny stickers you put on it. So, given that Algebraic is already in Phobos, and likely to remain there for a long time, I think the most sensible path forward is for SumType, or something like it, to be added alongside it.

I would add to this that the reasons to prefer a new module over an in-place upgrade of `Algebraic` are that (1) we don't have to worry about breaking existing code, and (2) it gives us an opportunity to leave behind the shortcomings of `Algebraic`'s API, such as the default "uninitialized" state, the use of exceptions for error handling, and of course the name `Algebraic` (I think someone once remarked that naming a discriminated union implementation `Algebraic` is like naming a hash table implementation `Container`).

Please feel free to ask if you have any more questions.

[1] https://www.oreilly.com/library/view/mastering-the-c17/9781787126824/c6737dd3-47fa-4719-a459-72afac90857c.xhtml
[2] https://forum.dlang.org/post/kuwarcdbclbhedgxutbw@forum.dlang.org

RazvanN7:
@atilaneves Have you come to a conclusion regarding this addition?

It is going to require some work to review this.  @andralex @wilzbach @jmdavis @CyberShadow @MartinNowak @thewilsonator and everyone else who has the time. Maybe we can coordinate to review parts of this addition?

atilaneves:
@RazvanN7 No, not yet, sorry. I know I should do that soon.

.dscanner.ini
```diff
@@ -207,11 +207,12 @@ assert_without_msg="-etc.c.SQL_,\
 -std.uuid,\
 -std.variant,\
 -std.windows.registry,\
--std.xml"
+-std.xml,\
+-std.sumtype"
```
berni44, line 211: I'd prefer to have std.sumtype in the correct place (after std.string) to keep the alphabetical order. This applies to all other changes in this file too.

pbackus: Good catch. Fixed.

pbackus:
@atilaneves Can you give any kind of estimate for when we can expect some progress on this? If not, what would it take to generate such an estimate? I'm happy to help in any way I can, but complete radio silence from you/the DLF means my opportunities to do so are severely limited.

std/variant.d
```diff
@@ -1588,6 +1591,8 @@ useful when it is desirable to restrict what a discriminated type
 could hold to the end of defining simpler and more efficient
 manipulation.
 
+$(RED Warning: $(LREF Algebraic) is obsolete and not recommended for use in new
+code. Instead, use $(REF SumType, std,sumtype).)
```
pbackus, line 1595: See above.

pbackus (commented):
Kicking things off here with a self-review. Any and all feedback is appreciated, but I'm especially hoping to hear from experienced Phobos developers regarding the build-system changes.

atilaneves:
> @atilaneves Can you give any kind of estimate for when we can expect some progress on this? If not, what would it take to generate such an estimate? I'm happy to help in any way I can, but complete radio silence from you/the DLF means my opportunities to do so are severely limited.

I'm giving myself until next Wed. Keep me to it.

std/variant.d
```diff
@@ -23,6 +23,9 @@ type constructor. Unlike `Variant`, `Algebraic` only allows a finite set of
 types, which are specified in the instantiation (e.g. $(D Algebraic!(int,
 string)) may only hold an `int` or a `string`).
 
+$(RED Warning: $(LREF Algebraic) is obsolete and not recommended for use in new
+code. Instead, use $(REF SumType, std,sumtype).)
```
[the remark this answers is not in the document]
maxhaton: I think the term obsolete should be reserved for modules that have been formally removed from Phobos. Out-dated is a good term in this case as it still works but is no longer the optimal solution.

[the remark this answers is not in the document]
pbackus: Changed to "outdated."

pbackus:
@atilaneves Ping. What's the next step here?

dukc:
Ping @atilaneves 

Decision still needed

andralex:
> Ping @atilaneves
> 
> Decision still needed

Atila is off for a while for personal reasons. @WalterBright should be able to help in the interim. What is the decision to be taken?

maxhaton:
> > Ping @atilaneves
> > Decision still needed
> 
> Atila is off for a while for personal reasons. @WalterBright should be able to help in the interim. What is the decision to be taken?

The decision is whether to merge into Phobos or not. 

I think sumtype is of Phobos-quality but the decision is two-fold: 

* If we merge should it be in Phobos or core (Not everyone uses Phobos, they now have to copy and paste one module in the worst case) - the library is nogc and betterC so I would vote for core. Putting it druntime also encourages standardisation around it.

* Should we be thinking bigger than this - as good a job as Paul has done here, algebraic data types as-a-library are just not very ergonomic, fast (at least one call to staticMap for every pattern match), and are harder to type check. The lack of ADTs is, in my mind, one of not many things that D really lacks.

pbackus:
> The decision is whether to merge into Phobos or not.

As I read it, there are three questions that need to be answered here:

1. What are the requirements for adding a new module to Phobos?
2. Does `sumtype` meet those requirements?
3. Assuming that `sumtype` meets those requirements, do we *want* to include it in Phobos?

Both 1 and 2 require input from Walter and/or Atila, so progress on those fronts is blocked until one of them is available

pbackus:
>  If we merge should it be in Phobos or core (Not everyone uses Phobos, they now have to copy and paste one module in the worst case) - the library is nogc and betterC so I would vote for core. Putting it druntime also encourages standardisation around it.

I'm agnostic on this, but I think it's important to mention that `sumtype` currently depends on Phobos for its `toString` method. It's possible we'll want to move string formatting into `core` too, especially if DIP 1036 ends up being accepted, so that's not necessarily an insurmountable obstacle, but it's not quite as simple as just copy+pasting one module.

> Should we be thinking bigger than this - as good a job as Paul has done here, algebraic data types as-a-library are just not very ergonomic, fast (at least one call to staticMap for every pattern match), and are harder to type check. The lack of ADTs is, in my mind, one of not many things that D really lacks.

I am not sure it would be wise to give up on useful, working code today in favor of a hypothetical future language feature that hasn't even had a DIP written for it yet, much less an implementation. D's history is littered with ambitious projects that looked promising early on, generated a bunch of excitement, and eventually ended up abandoned. By all means, we should consider our options, but let's be realistic about what those options actually are.

pbackus:
Auto-tester is currently red because one of the Darwin runners timed out while attempting to build DMD. It should fix itself the next time a new run is triggered.

std/sumtype.d
```diff
@@ -0,0 +1,2488 @@
...
+private:
+
+    enum bool canHoldTag(T) = Types.length <= T.max;
+    alias unsignedInts = AliasSeq!(ubyte, ushort, uint, ulong);
+
+    alias Tag = Filter!(canHoldTag, unsignedInts)[0];
+
+    union Storage
+    {
+        // Workaround for dlang issue 20068
```
atilaneves, line 297: Please use a link to bugzilla instead.

std/sumtype.d
```diff
@@ -0,0 +1,2488 @@
...
+
+/// Placeholder used to refer to the enclosing [SumType].
+struct This {}
+
+// Converts an unsigned integer to a compile-time string constant.
+private enum toCtString(ulong n) = n.stringof[0 .. $ - "LU".length];
+
+@safe unittest
+{
+    assert(toCtString!0 == "0");
```
atilaneves, line 245: static assert?

std/sumtype.d
```diff
@@ -0,0 +1,2488 @@
...
+        }
+    }
+}
+
+// Construction
+@safe unittest
+{
+    alias MySum = SumType!(int, float);
+
+    assert(__traits(compiles, MySum(42)));
```
atilaneves, line 798: Wouldn't it be better to just write the code instead of asserting it compiles? If it ever fails to, the compiler will spit out an error message.

std/sumtype.d
```diff
@@ -0,0 +1,2488 @@
...
+
+    assert(__traits(compiles, y = x));
+}
+
+// Equality
+@safe unittest
+{
+    alias MySum = SumType!(int, float);
+
+    MySum x = MySum(123);
```
atilaneves, line 828: Why bother with the variables instead of:

```d
assert(MySum(123) == MySum(123));
assert(MySum(123) != MySum(456));
// ...
```

std/sumtype.d
```diff
@@ -0,0 +1,2488 @@
...
+
+    assert(x == y);
+    assert(x != z);
+    assert(x != w);
+    assert(x != v);
+
+}
+
+// Equality of differently-qualified SumTypes
+version (D_BetterC) {} else
```
atilaneves, line 842: I'd add a comment explaining why this can't be tested in this version.

std/sumtype.d
```diff
@@ -0,0 +1,2488 @@
...
+            {
+        alias MySum = SumType!(const(int[]), immutable(float[]));
+    }));
+}
+
+// Recursive types
+@safe unittest
+{
+    alias MySum = SumType!(This*);
+    assert(is(MySum.Types[0] == MySum*));
```
atilaneves, line 881: static assert

std/sumtype.d
pbackus: Done.

std/sumtype.d
pbackus: I don't remember why I wrote it this way. Variables are gone now.

std/sumtype.d
pbackus: Done.

std/sumtype.d
pbackus: The `assert` communicates intent better (IMO) and avoids masking other compile-time errors in case of a failed `static assert` (e.g., in `match`).

std/sumtype.d
pbackus: A failed `static assert` causes compilation to fail immediately, which can potentially hide compile-time errors further on in the file. I use runtime `assert` throughout to avoid this.

std/sumtype.d
pbackus: See previous comment.

std/sumtype.d
andralex: Yah but you still want to make sure `toCtString` works during compilation.

std/sumtype.d
andralex: I don't think it makes much of a difference to the maintainer - if this fails, it needs to be fixed pronto and that's the end of it. FWIW too many errors can be move overwhelming than informative. I forgot how long building vs. running the unittests takes, if the static assert causes the process to fail a lot sooner it has an advantage.

andralex:
Did I mention how excited I am about this addition. `SumType` is hardly distinguishable from magic.

std/sumtype.d
pbackus: It's a template. If it works at all, it works during compilation (and it will quickly fail elsewhere if it doesn't).

The purpose of this test is to guard against potential changes to the format of `.stringof`—since, technically, it's an implementation detail of the compiler that the spec makes no guarantees about. I'll add a comment to clarify.

std/sumtype.d
pbackus: `sumtype`'s test suite builds quickly enough (~2s on my machine) that I don't think we need to worry too much about this kind of optimization. I suppose if someone is building the entire Phobos test suite sequentially it could help, but I don't expect that's a very common operation, and the error-masking issue is a real annoyance during normal maintenance work (I speak from experience).

std/sumtype.d
```diff
@@ -0,0 +1,2504 @@
...
+    /**
+     * Compares two `SumType`s for equality.
+     *
+     * Two `SumType`s are equal if they are the same kind of `SumType`, they
+     * contain values of the same type, and those values are equal.
+     */
+    bool opEquals(this This, Rhs)(auto ref Rhs rhs)
+    if (is(CommonType!(This, Rhs)))
+    {
+        static if (is(This == Rhs))
```
atilaneves, line 652: Is this not a problem if they differ in constness?

atilaneves:
I think that all the tests that have `assert(__traits(compiles, /* ... */)));` should just have whatever should compile due to compiler errors being printed if they don't - otherwise it'll be hard to diagnose why something is failing.

Other than that (minor anyway) and my question just now, should be good to go.

std/sumtype.d
pbackus: That's what the `else` branch is for. The `unittest` labelled "Equality of differently-qualified SumTypes" covers it.

pbackus:
@atilaneves Should be ready to merge now.

RazvanN7 (approved):

RazvanN7:
Thanks to @pbackus for this awesome addition and to everyone involved in the review process.

PR #7706, opened 2020-11-25
Nullable isNotNull

burner:
I find myself typing !var.isNull so often.

So much so that I think var.isNotNull would make my code look so much better.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2870,6 +2870,27 @@ struct Nullable(T)
         assert(!ni.isNull);
     }
 
+    /**
+     * Check if `this` is not in the null state.
+     *
+     * Returns:
+     *     true $(B iff) `this` is not in the null state, otherwise false.
+     */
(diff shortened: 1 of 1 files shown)


atilaneves:
I'm not sure this is worth adding.

Geod24 (changes requested):
I don't see the point of this. `!var.isNull()` is shorter, easy to understand, and already well established.

burner:
fair enough, I just found it so annoying 

because my brain is not thinking in an order similar to "not is some condition" but rather "is not condition" must be a native German speaking thing.

didn't take long to write this PR, I hope I didn't waste to much of your guys time.

PetarKirov:
What do you guys think about:

```D
bool opCast(T : bool)() { return !isNull; }
```

So that one could do:
```D
if (var) { }
```

instead?

burner:
Idk, I'm not really a fan.

PR #7707, opened 2020-11-26
Change private function BigInt.isNegative to an alias

n8sh:


std/bigint.d
```diff
@@ -1459,10 +1459,7 @@ private:
     {
         return data.isZero();
     }
-    bool isNegative() pure const nothrow @nogc @safe
-    {
-        return sign;
-    }
+    alias isNegative = sign;
```
thewilsonator, line 1462: Does this make sign writable? If so, is this desirable?

atilaneves: Both `sign` and `isNegative` are private so it should be fine.

thewilsonator (approved):

PR #7709, opened 2020-11-27
[stable] std.datetime.timezone: Fix regression for Android

kinke:


Geod24:
What's the regression ?

kinke:
The caller, `getInstalledTZNames()`, is now `@safe`.

Geod24 (approved):

PR #7711, opened 2020-11-28
Reduce LDC diff a bit

kinke:


ibuclaw:
Seems reasonable, but tests are failing.  The title could be better.

PetarKirov:
Can you try splitting the `std.math` changes a bit (and documenting them better as @ibuclaw  proposed)? That can help narrow down the CI failure.

kinke:
What fails is `assert(fmax(F(-2.0), 0.0) == F(0.0));` with enabled optimizations. The only thing that has changed there is that fmax is force-inlined now. So this is most likely yet another DMD optimizer bug.

kinke:
@ibuclaw: I've added some inline asm for other archs (derived from https://github.com/ldc-developers/phobos/commit/48c2e9cd48da803c119e4013548f4620f040d840) which might be interesting for GDC too.

kinke:
Argh, D-Scanner apparently chokes on GDC-style inline asm...

ibuclaw:
> Argh, D-Scanner apparently chokes on GDC-style inline asm...

Yeah, I guess choices are fix DScanner, tell DScanner to ignore (I don't suppose dscanner supports `// lint: nofoo` style comments?), or follow what RISCV does and use a mixin.

kinke:
Nope, mixin hacks are out of the question for me. As the asm style check is already disabled for `std.math` (https://github.com/dlang/phobos/blob/b6a9cd8c649b0c13a39674f74b341655b5964514/.dscanner.ini#L121), I guess these are parser issues and cannot be simply worked around.

I can extract the last commit as separate PR though and hope that someone takes care of the D-Scanner issue.

MoonlightSentinel:
FYI, I've added support for GDC-Style assembly in libdparse (which is also included in a more recent DScanner version):
Try this:
```patch
index 0df2b8fc7..7a958af6b 100644
--- a/posix.mak
+++ b/posix.mak
@@ -78,7 +78,7 @@ ROOT_OF_THEM_ALL = generated
 ROOT = $(ROOT_OF_THEM_ALL)/$(OS)/$(BUILD)/$(MODEL)
 DUB=dub
 TOOLS_DIR=../tools
-DSCANNER_HASH=9364d6f15f4a610fda49a693dbc18608bfc701bb
+DSCANNER_HASH=497a900f210a0a0831110f60162922adc5fc5283
 DSCANNER_DIR=$(ROOT_OF_THEM_ALL)/dscanner-$(DSCANNER_HASH)

 # Set DRUNTIME name and full path
```

kinke:
Oh wow, thanks Florian. The commit you've mentioned didn't upgrade the libdparse submodule, so I'm trying https://github.com/dlang-community/D-Scanner/commit/a40492bc92e86bad6441fc3aba776fe2c6a5090b.

std/math.d
```diff
@@ -5768,67 +5826,159 @@ private:
...
+        version (X86)
         {
-            short cont;
-            asm pure nothrow @nogc
+            version (DigitalMars)
+            {
+                asm pure nothrow @nogc
+                {
+                    xor EAX, EAX;
+                    fstcw cont;
```
kinke, line 5878: I have no idea why EAX/RAX is set to 0 before `fstcw`. That instruction doesn't seem to depend on it (https://www.felixcloutier.com/x86/fstcw:fnstcw).

std/math.d
```diff
@@ -5160,52 +5194,76 @@ private:
...
+                {
+                    asm nothrow @nogc { stmxcsr mxcsr; }
+                    mxcsr &= ~EXCEPTIONS_MASK;
+                    asm nothrow @nogc { ldmxcsr mxcsr; }
+                }
+                else
+                {
+                    asm nothrow @nogc { "stmxcsr %0" : "=m" (mxcsr); }
+                    mxcsr &= ~EXCEPTIONS_MASK;
+                    asm nothrow @nogc { "ldmxcsr %0" : : "m" (mxcsr) : "flags"; }
```
kinke, line 5230: Unsure about the correct clobber for these x87 (`flags`?) and SSE flags (`fpsw`?) - see https://github.com/llvm-mirror/clang/blob/aa231e4be75ac4759c236b755c57876f76e3cf05/lib/Parse/ParseStmtAsm.cpp#L636-L640...

MoonlightSentinel:
> The commit you've mentioned didn't upgrade the libdparse submodule, so I'm trying [dlang-community/D-Scanner@a40492b](https://github.com/dlang-community/D-Scanner/commit/a40492bc92e86bad6441fc3aba776fe2c6a5090b) (**edit**: yeah, and hitting the `UTFException` now...).

Damn, guess I've tested some revisions with a stale submodule...
Currently investigating the `UTFException`.

std/math.d
ibuclaw: glibc uses these clobber flags, so I trust them.

ibuclaw: e.g:
```
unsigned int xnew_exc;
/* Get the current control word.  */
__asm__ ("stmxcsr %0" : "=m" (*&xnew_exc));
xnew_exc &= ~(excepts << 7);
__asm__ ("ldmxcsr %0" : : "m" (*&xnew_exc));
```

kinke:
Thanks! - I've removed the asm commits and will try to follow up another time.

kinke:
Btw, every time I see `feqrel(a, b) >= T.mant_dig - N` I think there's hardly any use for a `feqrel`, and some `fdiffrel` or so instead would be much nicer. Anyone else thinking this way? Name suggestions welcome. :)

RazvanN7 (approved):

PR #7712, opened 2020-11-29
Make `std.concurrency.Tid.toString` @safe-able

tom-tan:
Currently `std.concurrency.Tid` cannot be converted to a string safely by `std.conv` family, for example.
This request fixes this issue.

- It fixes the signature of `Tid.toString` to work with safe code (See also [std.format.formatValue](https://dlang.org/phobos/std_format.html#.formatValue)), and
- encloses an unsafe cast from `MessageBox` to `void*` with a trusted lambda function.

### DIFF
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -387,17 +387,17 @@ public:
      * that a Tid executed in the future will have the same toString() output
      * as another Tid that has already terminated.
      */
-    void toString(scope void delegate(const(char)[]) sink) const
+    void toString(W)(ref W w) const
     {
-        import std.format.write : formattedWrite;
-        formattedWrite(sink, "Tid(%x)", cast(void*) mbox);
+        import std.format : formattedWrite;
+        auto p = () @trusted { return cast(void*) mbox; }();
+        formattedWrite(w, "Tid(%x)", p);
     }
 
 }
 
-@system unittest
+@safe unittest
 {
-    // text!Tid is @system
     import std.conv : text;
     Tid tid;
     assert(text(tid) == "Tid(0)");
--- a/std/concurrency.d
+++ b/std/concurrency.d
@@ -389,7 +389,7 @@ public:
      */
     void toString(W)(ref W w) const
     {
-        import std.format : formattedWrite;
+        import std.format.write : formattedWrite;
         auto p = () @trusted { return cast(void*) mbox; }();
         formattedWrite(w, "Tid(%x)", p);
     }


std/concurrency.d
```diff
@@ -342,17 +342,17 @@ public:
      * that a Tid executed in the future will have the same toString() output
      * as another Tid that has already terminated.
      */
-    void toString(scope void delegate(const(char)[]) sink)
+    void toString(W)(ref W w) const
```
RazvanN7, line 345: What is the point of templating this function if you are going to put the @system line in a trusted block? You can simply leave the function signature as is and simply add @safe then.

tom-tan: > What is the point of templating this function if you are going to put the @System line in a trusted block?

Using template functions is required in [std.format.formatValue](https://dlang.org/phobos/std_format.html#.formatValue) that is used in `formattedWrite`.
The current signature is still available but not recommended because it prevents from inferring attributes such as safe.

In the case of this request, I first tried to solve it by adding safe attribute to the current signature as you suggested.
However, the compiler said that "safe function `toSTring` cannot call a system function `formattedWrite`" and I finally found that the problem is the signature of `toString` rather than `formattedWrite` after the long journey in Phobos.

RazvanN7: Hmm, I see, depending on `sink`, the call to `formattedWrite` might be @safe, right?

RazvanN7: Maybe a better signature would be: `void toString()(scope void delegate(const(char)[]) sink)`. Templated with an empty template paramter list. That way, it is clear what type of argument the function expects.

tom-tan: > Hmm, I see, depending on sink, the call to formattedWrite might be @safe, right?

I guess so.

> Maybe a better signature would be: `void toString()(scope void delegate(const(char)[]) sink)`.

Using different signature from recommended in std.format makes users confused and changing the signatures for `formatValue` is out of scope of this request.

RazvanN7: I'm sorry but I don't get your point. Without this PR, we have a non-templated function with the signature `void toString(scope void delegate(const(char)[]) sink)`. This function is `system` but it specifically accepts only delegate arguments, whereas with this PR it can accept any type of argument. What I am proposing is to simply template `toString` with an empty template parameter list. That way, `@safe` can be inferred but the signature remains mostly the same.

tom-tan: I'm sorry, I used a wrong signature in the comment.
Please replace
> `void toString()(scope void delegate(const(char)[]) sink) @system`

with `void toString()(scope void delegate(const(char)[]) @system sink)`.
That is, even if `toString` is templated, the argument `sink` only accepts system delegates.

In addition to that, as I linked in the first comment, the [official document of std.format.formatValue](https://dlang.org/phobos/std_format.html#.formatValue) says:

> `toString` should have one of the following signatures:
> ```dlang
> void toString(W)(ref W w, scope const ref FormatSpec fmt)
> void toString(W)(ref W w)
> string toString();
> ```

Therefore using `void toString(W)(ref W w)` is reasonable.

RazvanN7 (approved):

Geod24:
And that's a breaking change. For example, Ocean uses a formatter that does not recognize the templated overload, but recognize the delegate one.

tom-tan:
CI for macOS 10.15 fails in runnable/test13117.d and runnable/test13117b.d but I am not sure why it fails.
These tests seem to be unrelated to `std.concurrency`  because they only import `std.file` and `std.stdio`.

RazvanN7:
Maybe try a rebase @tom-tan ?

tom-tan:
Done!

RazvanN7:
> And that's a breaking change. For example, Ocean uses a formatter that does not recognize the templated overload, but recognize the delegate one.

But buildkite tests Ocean and all tests are passing over there.

tom-tan:
Rebased!

tom-tan:
Rebased but it still fails. I'm not sure why it does...
[test13117.d](https://github.com/dlang/dmd/blob/master/test/runnable/test13117.d) and [test13117b.d](https://github.com/dlang/dmd/blob/master/test/runnable/test13117b.d) does not directly import `std.concurrency`.

> >>> TARGET FAILED: runnable/test13117.d
> >>> TARGET FAILED: runnable/test13117b.d

RazvanN7:
The failure is not spurious. It seems that this modification affects the size of the binary and on macOS this size exceeds the limit that is tested for.  The code in ` runnable/test13117.d` is:

```d
// PERMUTE_ARGS: -O -release -g                                                                                                               
import std.file, std.stdio;

int main()
{
    auto size = thisExePath.getSize();
    writeln(size);     // it is  2271408 on macOS
    version (D_LP64)
        enum limit = 1_906_432;
    else
        enum limit = 1_900_000;
    return size > limit * 11 / 10;    // 1906432*11/10 = 2097075 (smaller than 2271408) 
}
```

This is just a shot in the dark, bt since `toString` is becoming a template now it's likely that it gets instantiated more and somehow the code ends up in the executable of the test. Not sure how the limit value was chosen though. cc @MartinNowak @WalterBright

RazvanN7:
@tom-tan Any ideas on how to fix this?

tom-tan:
Sorry, currently I have no idea for that...

To investigate the source of the problem, we have to dive into the compiler and to know what is tested with test13117.d and test13117b.d, I guess.

RazvanN7:
@tom-tan I have provided some info in a previous comment. The test checks that the binary size is within a limit. It might be possible that templating the `toString` method leads to an increase in binary size.

tom-tan:
As I mentioned [before](https://github.com/dlang/phobos/pull/7712#discussion_r559149904), we cannot make non-templated function `toString(void delegate(const(char)[]) sink)` `@safe`-able because they only accept system delegates.

I guess there are three options for this issue:
- Use `string toString()` rather than `toString(void delegate(const(char)[]) sink)` and make it `@safe`. Of course it breaks existing codes.
- Fix the compiler to make the resulted binary size small. IMO it is ideal solution for this issue. However, it is hard for me because I am not familiar with the code of dmd.
- Give up with this issue :-(

Do you have any other options for it or is it acceptable to use the first option?

PetarKirov:
@RazvanN7 I suggest increasing the limit (as we have done several times in the past, IIRC) so we can unblock this PR and pursue ways to optimize the binary size later on, independent of the work here.

RazvanN7:
https://github.com/dlang/dmd/pull/12949

tom-tan:
Thanks!

PR #7713, opened 2020-11-30
Add module std.digest.sha3

dd86k:
I have read the contributing guidelines so I should be fine.

# Introduction

A successor of the SHA-2 algorithm released in 2001, SHA-3 is a hashing algorithm first published in 2015.

# PR Description

This PR adds the SHA-3/SHAKE digest module as `std.digest.sha3` using the Keccak[1600] function. It is compatible with `std.digest` as its main inspiration came from `std.digest.sha`.

Reasons why it is its own module is because:
- I wanted to keep the original copyright of sha intact;
- And I wanted to keep the module a clean implementation.

I already use a variation of this module in my DDH project, which uses it to hash files: https://github.com/dd86k/ddh

Further testing was done with `openssl dgst`.

Bugzilla: https://issues.dlang.org/show_bug.cgi?id=21440

### DIFF
--- a/std/digest/sha3.d
+++ b/std/digest/sha3.d
@@ -9,6 +9,8 @@ module std.digest.sha3;
 
 public import std.digest;
 
+private import core.bitop : rol, bswap;
+
 private immutable ulong[24] K_RC = [
     0x0000000000000001, 0x0000000000008082, 0x800000000000808a, 0x8000000080008000,
     0x000000000000808b, 0x0000000080000001, 0x8000000080008081, 0x8000000000008009,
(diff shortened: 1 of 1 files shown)


std/digest/sha3.d
```diff
@@ -0,0 +1,368 @@
...
+        return (((x) << (y)) | ((x) >> (64 - (y))));
+    }
+
+    version (BigEndian)
+    pragma(inline, true)
+    void swap()
+    {
+        
+        for (size_t i; i < 25; ++i)
+            st64[i] = sit.bitop.bswap(st64[i]);
```
n8sh, line 131:
```suggestion
            st64[i] = core.bitop.bswap(st64[i]);
```

n8sh:
I support of the addition of this to Phobos. I have a mild preference for adding it to `std.digest.sha` to be consistent with the decision to put SHA-1 and SHA-2 in the same module.

dd86k:
> I have a mild preference for adding it to std.digest.sha

I did think of that, it is a sound argument, but while SHA-1 and SHA-2 are similar, SHA-3 is rather different in its core, despite being in the same family as initially thought. Still, I could move it over if e.g. Walter wants to.

Thanks for the correction earlier. That did completely flew over my head.

dd86k:
Currently seeing speeds of 130 MiB/s for phobos' sha256 implementation (matching sha256sum(1)) and around 27 MiB/s for my sha3 implementation (using `pv /dev/urandom | ...`). I'll check if I can improve it.

dd86k:
This improvement is actually usable and improves the throughput by about almost four times (27 MiB/s against 105 MiB/s tested in a VM using a 64 KiB buffer).

n8sh:
@dd86k if you were to change this to allow others to push changes the style complaints could be fixed quickly.

dd86k:
I now just enabled `Allow edits by maintainers`. Was this the one you were talking about? @n8sh

n8sh:
@dd86k Yes, that's it

n8sh:
Note: copied public unit tests from `std.digest.sha`.

std/digest/sha3.d
```diff
@@ -0,0 +1,510 @@
...
+            "digest size must be 224, 256, 384, or 512 bits for SHA-3");
+
+    @safe @nogc pure nothrow:
+
+    enum
+    {
+        blockSize = digestSize, /// digest size in bits
+        dgst_sz_bytes = blockSize >> 3, /// digest size in bytes
+        delim = shake ? 0x1f : 0x06, /// delimiter when finishing
+        rate = 200 - (blockSize >> 2), /// sponge rate
```
n8sh, line 55: Elsewhere in std.digest `blockSize` has the meaning you have given to `rate` here.

```suggestion
        blockSize = 200 - (digestSize >> 2), /// sponge rate in bytes
        dgst_sz_bytes = digestSize >> 3, /// digest size in bytes
        delim = shake ? 0x1f : 0x06, /// delimiter when finishing
        private alias rate = blockSize;
```

n8sh:
druntime test failure appears unrelated.

>generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D8runbench6runCmdFAyabIAaZQi: error: undefined reference to '_D3std7process12executeShellFNeMAxaxHAyaAyaEQBqQBp6ConfigmMQBbQzZSQCm8typecons__T5TupleTiVQCba6_737461747573TQCuVQCya6_6f7574707574ZQBz'
>collect2: error: ld returned 1 exit status
>Error: linker exited with status 1
>posix.mak:345: recipe for target 'generated/linux/release/64/benchmark' failed

@atilaneves can you give an opinion on whether this should be added?

std/digest/sha3.d
dd86k: Pushed ahead a mix of this suggestion, marking the enum as `private`, and changed the field name `dgst_sz_bytes` to `digestSizeBytes` to better fit the D Style.

dd86k:
Note: I noticed the sponge function could use some work. Threw in some quick (broken-on-purpose) loop XOR'ing 4/8 bytes at a time could give about 6/12 MiB/s respectively of an additional throughput (LDC2).

n8sh:
>Note: I noticed the sponge function could use some work. Threw in some quick (broken-on-purpose) loop XOR'ing 4/8 bytes at a time could give about 6/12 MiB/s respectively of an additional throughput (LDC2).

Here's a fully working version I tested on my machine. It also uses .ptr to bypass array bounds checks. For processing a totally-in-memory buffer it improved throughput from 220 MB/s to 276 MB/s with ldc2 -O3.

```d
union
{
    private ubyte[200] st;  /// state (8bit)
    private ulong[25] st64; /// state (64bit)
    private size_t[200 / size_t.sizeof] stz; // <-- Added.
}
```
...
```d
void put(scope const(ubyte)[] input...) @nogc nothrow pure @trusted
{
    size_t j = pt;
    // Process wordwise if properly aligned.
    if ((j | cast(size_t) input.ptr) % size_t.alignof == 0)
    {
        static assert(rate % size_t.sizeof == 0);
        foreach (const word; (cast(size_t*) input.ptr)[0 .. input.length / size_t.sizeof])
        {
            stz.ptr[j / size_t.sizeof] ^= word;
            j += size_t.sizeof;
            if (j >= rate)
            {
                transform;
                j = 0;
            }
        }
        input = input.ptr[input.length - (input.length % size_t.sizeof) .. input.length];
    }
    // Process reminder bytewise.
    foreach (const b; input)
    {
        st.ptr[j++] ^= b;
        if (j >= rate)
        {
            transform;
            j = 0;
        }
    }
    pt = j;
}
```

dd86k:
That's very nice! Just tested it on my laptop machine and got from 153 MiB/s to 180 MiB/s.

Feel free to add the suggestion as-is. Personally I'd trim off `@nogc nothrow pure` since all structure members are already marked as such (under `private size_t pt`).

atilaneves:
I know little to nothing about this. Could you please explain why you want to add this to Phobos? Thanks!

dd86k:
> Could you please explain why you want to add this to Phobos? 

Currently, the only secure hashing algorithm in `std.digest` is the SHA-2 family introduced in 2001. SHA-1 has been practically broken, but still used as a hardened variant (see git and sha1dc as an example). Phobos does not provide a collision detection feature for SHA-1 (unless, I guess, someone adds this in as a separate structure for this whole thing).

Adding SHA-3 adds a step ahead before it's SHA-2's turn. SHA-3, introduced in 2015, already gained some amount of traction (example: sqlite3 officially uses SHA-3 for verifying the data integrity of downloaded packages).

I didn't make the module a DUB package since SHA-1 and SHA-2 are already in Phobos and thus amounts to this PR.

n8sh (commented):
BTW to appear in the release notes one of the commit messages should begin with "Fix issue 21440" (perhaps after squashing the PR into a single commit).

n8sh:
Rebased on master branch.

n8sh (approved):
Approving this to indicate that I have reviewed the code although a decision on whether to add this to Phobos is pending from @atilaneves.

dd86k:
Fun: Tested the implementation on my PinePhone (aarch64) using LDC and through urandom I get a bandwidth of 44 MiB/s for SHA-3-256. Compared to Phobos' SHA-2-256 I only get 6.66 MiB/s.. Using GDC I get about 6.52 MiB/s for SHA-2-256 and 8 MiB/s for SHA-3-256.

atilaneves:
Wouldn't it perhaps be better if this started life as a dub package so it can be battle tested before making it into Phobos?

dd86k:
Half-agreed. I think for the moment being it'd be best for me to add additional unittests (e.g., against empty data/files, against big-endian, etc.) and I'll go and prepare a DUB package for the interested.

Feel free to close the PR if need be. If left open I'll continue to contribute as time goes on in parallel of the future repo I'll setup.

RazvanN7:
Since this PR already has a bugzilla entry [1], I will close for now. Feel free to reopen once this is ready for integration in phobos.

[1] https://issues.dlang.org/show_bug.cgi?id=21440

PR #7714, opened 2020-12-01
std.stdio: Deprecate extern(C) getline

ibuclaw:
The (accidentally?) publicly available `extern(C)` binding for `getline` in `std.stdio` has been deprecated. Any code that still needs it can import the symbol from `core.sys.posix.stdio` in druntime instead.

Within phobos itself, this function is unused.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -264,7 +264,10 @@ else
 version (HAS_GETDELIM) extern(C) nothrow @nogc
 {
     ptrdiff_t getdelim(char**, size_t*, int, FILE*);
+
+    // @@@DEPRECATED_2.104@@@
     // getline() always comes together with getdelim()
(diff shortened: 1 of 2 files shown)


std/stdio.d
```diff
@@ -264,7 +264,10 @@ else
 version (HAS_GETDELIM) extern(C) nothrow @nogc
 {
     ptrdiff_t getdelim(char**, size_t*, int, FILE*);
```
thewilsonator, line 266: What about this?

thewilsonator (approved):

std/stdio.d
ibuclaw: First dealing with the version logic in #7715, then replacing `version (HAS_GETDELIM)` with `import core.sys.posix.stdio;" and checking the existence using `__traits(compiles)`.

PR #7717, opened 2020-12-01
std.stdio: Deprecate extern(C) getdelim

ibuclaw:
The publicly available `extern(C)` binding for `getdelim` in `std.stdio` has been deprecated. Any code that still needs it can import the symbol from `core.sys.posix.stdio` in druntime instead.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -256,6 +256,8 @@ static if (__traits(compiles, core.sys.posix.stdio.getdelim))
 {
     extern(C) nothrow @nogc
     {
+        // @@@DEPRECATED_2.104@@@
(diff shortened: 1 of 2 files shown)


thewilsonator (approved):

std/stdio.d
```diff
@@ -256,6 +256,8 @@ static if (__traits(compiles, core.sys.posix.stdio.getdelim))
 {
     extern(C) nothrow @nogc
     {
+        // @@@DEPRECATED_2.104@@@
+        deprecated("To be removed after 2.104. Use core.sys.posix.stdio.getline instead.")
```
ibuclaw, line 260: @thewilsonator Oops! You failed my test. ;-)

PR #7719, opened 2020-12-01
Remove some unnecessary undocumented explicit ctors in std.range

n8sh:


thewilsonator (approved):

CyberShadow:
This seems to have introduced a regression:

https://github.com/dlang/phobos/pull/7727#pullrequestreview-555934863

CyberShadow:
Unminimized test case:

```d
module std.processtest; // For package(std)

import core.sys.posix.unistd;

import std.algorithm.iteration;
import std.conv;
import std.path;
import std.range.primitives;
import std.traits;
import std.internal.cstring;

private string searchPathFor(scope const(char)[] executable)
    @safe
{
    import std.algorithm.iteration : splitter;
    import std.conv : text;
    import std.path : chainPath;

    string result;

	void fun(scope const(char)[] path) @safe
	{
		if (!path)
			return;

		foreach (dir; splitter(path, ":"))
		{
			auto execPath = chainPath(dir, executable);
			if (isExecutable(execPath))
			{
				result = text(execPath);
				return;
			}
		}
	}
	fun("test");

    return result;
}

private bool isExecutable(R)(R path) @trusted nothrow @nogc
if (isInputRange!R && isSomeChar!(ElementEncodingType!R))
{
    return (access(path.tempCString(), X_OK) == 0);
}
```

PR #7724, opened 2020-12-12
std.regex: improve documentation of character classes.

Panke:


std/regex/package.d
```diff
@@ -209,7 +209,8 @@ $(TR $(TD Objects) $(TD
   $(REG_START Character classes )
   $(REG_TABLE
     $(REG_TITLE Pattern element, Semantics )
-    $(REG_ROW Any atom, Has the same meaning as outside of a character class.)
+    $(REG_ROW Any atom, Has the same meaning as outside of a character class. 
+      Except for ] which has to be written as \])
```
thewilsonator, line 213: > Except for `]` which has to be written as `\]`)

How does this render? I expect that it would be a good idea to encase `]` and `\]` in backticks. the backlash may need additional escaping.

Also `which has to be written as` should be `which must be written as` and that row of text should be one sentence: `... character class, except for...`

Panke: Thank you for your comments!

Is there an easy way to view the docs from the CI build?

thewilsonator:
Error - file 'std/regex/package.d' contains trailing whitespace at line 212

std/regex/package.d
thewilsonator: click on the `details` of the CyberShadow/DAutoTest, then click base result : view, then click phobos and navigate to the page of interest.

Panke: It renders ok now.

thewilsonator:
I have no idea why the azure OMF build fails. nut otherwise looks good.

thewilsonator (approved):

PR #7726, opened 2020-12-14
[stable] std.process: Include program name in exception msg if it cannot be found

kinke:
On Windows, the message for this common error case wasn't as informative as it should be:
```
Failed to spawn new process (The system cannot find the file specified.)
```

Handle this just like on POSIX via a message like:
```
Executable file not found: 7z.exe
```

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -811,9 +811,18 @@ Pid spawnProcess(scope const(char[])[] args,
                  scope const char[] workDir = null)
     @trusted // TODO: Should be @safe
 {
-    version (Windows)    auto  args2 = escapeShellArguments(args);
-    else version (Posix) alias args2 = args;
-    return spawnProcessImpl(args2, stdin, stdout, stderr, env, config, workDir);
+    version (Windows)
+    {
+        const commandLine = escapeShellArguments(args);
+        const program = args.length ? args[0] : null;
+        return spawnProcessImpl(commandLine, program, stdin, stdout, stderr, env, config, workDir);
+    }
+    else version (Posix)
+    {
+        return spawnProcessImpl(args, stdin, stdout, stderr, env, config, workDir);
+    }
+    else
+        static assert(0);
 }
 
 /// ditto
@@ -1160,7 +1169,7 @@ private Pid spawnProcessImpl(scope const(char[])[] args,
                     errorMsg = "getrlimit failed";
                     break;
                 case InternalError.exec:
-                    errorMsg = "Failed to execute program";
+                    errorMsg = "Failed to execute '" ~ cast(string) name ~ "'";
                     break;
                 case InternalError.doubleFork:
                     // Can happen only when starting detached process
@@ -1209,6 +1218,7 @@ envz must be a pointer to a block of UTF-16 characters on the form
 */
 version (Windows)
 private Pid spawnProcessImpl(scope const(char)[] commandLine,
+                             scope const(char)[] program,
                              File stdin,
                              File stdout,
                              File stderr,
@@ -1218,6 +1228,7 @@ private Pid spawnProcessImpl(scope const(char)[] commandLine,
     @trusted
 {
     import core.exception : RangeError;
+    import std.conv : text;
 
     if (commandLine.empty) throw new RangeError("Command line is empty");
 
@@ -1280,7 +1291,12 @@ private Pid spawnProcessImpl(scope const(char)[] commandLine,
     if (!CreateProcessW(null, commandLine.tempCStringW().buffPtr,
                         null, null, true, dwCreationFlags,
                         envz, workDir.length ? pworkDir : null, &startinfo, &pi))
-        throw ProcessException.newFromLastError("Failed to spawn new process");
+    {
+        if (GetLastError() == ERROR_FILE_NOT_FOUND)
+            throw new ProcessException(text("Executable file not found: ", program));
+        else
+            throw ProcessException.newFromLastError("Failed to spawn new process");
+    }
 
     // figure out if we should close any of the streams
     if (!(config & Config.retainStdin ) && stdinFD  > STDERR_FILENO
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

std/process.d
```diff
@@ -1280,7 +1291,12 @@ private Pid spawnProcessImpl(scope const(char)[] commandLine,
     if (!CreateProcessW(null, commandLine.tempCStringW().buffPtr,
                         null, null, true, dwCreationFlags,
                         envz, workDir.length ? pworkDir : null, &startinfo, &pi))
-        throw ProcessException.newFromLastError("Failed to spawn new process");
+    {
+        if (GetLastError() == ERROR_FILE_NOT_FOUND)
+            throw new ProcessException(text("Executable file not found: ", program));
```
CyberShadow, line 1296: What happens if a DLL or other resource is missing?

kinke: For a missing DLL, it doesn't seem to result in CreateProcessW failing (and instead make the child process exit with != 0). Tested with a `pythonw.exe` and taking away its `python37.dll`.

CyberShadow:
I think https://github.com/dlang/phobos/pull/7394 aimed for the same goal, but at the end was a smaller diff and made fewer assumptions. The author lost interest, perhaps it should be continued instead?

kinke:
I don't have a preference wrt. extra struct vs. extra param + some more versioning lines - I only care about the msg, as pretty much all ProcessExceptions I've ever seen on Windows were due to some missing program or incomplete PATH, and all I ever wanted to know was what it was trying to invoke. Using the program as additional context for all other CreateProcess errors might be worth it; that was actually my first version, before I checked what happens on Posix, after which I went for a simpler msg consistent with the Posix one.

CyberShadow:
If we can make things both useful and consistent across platforms, I think that would be a win :)

kinke:
On Posix, that msg is used when the program is only a name (no dir separator). With a path, it's `Not an executable file: ./iDontExist`. At least according to run.dlang.io - I'll need to debug why the testcase fails on Posix.

thewilsonator:
please ping me when this is good to go.

kinke:
Good now IMO, improving the msg on Posix too if the program contains a slash.

std/process.d
```diff
@@ -1160,7 +1169,7 @@ private Pid spawnProcessImpl(scope const(char[])[] args,
                     errorMsg = "getrlimit failed";
                     break;
                 case InternalError.exec:
-                    errorMsg = "Failed to execute program";
+                    errorMsg = "Failed to execute '" ~ cast(string) name ~ "'";
```
CyberShadow, line 1172: Because POSIX file names may contain characters such as newlines or quote characters, may I suggest
```suggestion
                    errorMsg = "Failed to execute " ~ escapeShellFileName(name);
```

std/process.d
CyberShadow, line 1296: I think the other PR was still a little better in this regard for a few reasons:

1. The error is "**file** not found", not "**executable** not found"
2. We forego error message localization provided by the OS for this error code only
3. It seems inconsistent to include the program name only for this error code, and not others.

CyberShadow (approved):

std/process.d
kinke: Well, the other PR was abandoned, resulting in no progress at all. May I suggest you follow up on your own, given that you have a strong opinion about this and seem to care about the improbable error cases and Posix filenames as well? I'm thankful for concerns regarding the DLLs, those haven't crossed my mind at all, but seem to be of no concern.

CyberShadow: No problem. I don't really have a strong opinion on this, just pointing out a few opportunities for improvement while we're on the subject.

CyberShadow:
BTW, why did this target stable?

std/process.d
CyberShadow, line 1296: Just noticed that this error message has the exact same wording as for when `searchPathFor` failed. We should probably avoid having errors that share the message string but indicate different error conditions.

kinke:
> BTW, why did this target stable?

Because I hate waiting for months to finally see a trivial improvement such as this (targeting master now means it shows up in 2.096).

std/process.d
kinke: Yeah, that's what I meant by 'consistent with the Posix msg' (`searchPathFor` is Posix only).

CyberShadow:
I understand, but there are a lot of other PRs that this would apply to, and if that were to happen, I think that would considerably diminish the value of the stable branch.

There is also the matter that follow-up PRs need to target stable too (see #7727).

Perhaps consider using Digger to easily use the current DMD master? :)

std/process.d
CyberShadow: OK. Sorry, at the time I did not know what exactly you meant by that. I understand that the two can refer to different error conditions (a file name was not found in `$PATH` vs. an absolute file name does not exist on disk).

kinke: Can be relative too. FWIW, I think Posix resulting in 2 different msges for the same error (program not found) isn't ideal; to me it doesn't matter whether the program contains a slash or not.

CyberShadow: > to me it doesn't matter whether the program contains a slash or not.

What do you mean? Whether the path contains a path separator or not is what decides if the `PATH` variable is consulted. The `Posix` logic mirrors the Windows API logic (quote "If the file name does not contain a directory path"), to make the Phobos API consistent across both platforms.

kinke:
> There is also the matter that follow-up PRs need to target stable too

Yes, or wait until stable is merged into master again.

> Perhaps consider using Digger to easily use the current DMD master? :)

This was bothering me in production (countless errors due to missing programs and me having no idea what it tried to invoke), where using DMD master is out of the question.

CyberShadow:
> This was bothering me in production (countless errors due to missing programs and me having no idea what it tried to invoke),

I see. I would use Process Monitor for this case, but I'm guessing that wasn't an option.

> where using DMD master is out of the question.

FWIW, Digger also allows cherry-picking changes, e.g. `stable+phobos#7726`.

CyberShadow:
Another reason to not put things in stable unless necessary: https://github.com/dlang/phobos/pull/7727#discussion_r546227580

@thewilsonator 👆

PR #7727, opened 2020-12-17
std.process touchups

CyberShadow:
Follow-up to #7726.

This is on stable because it's based on #7726, which targeted stable.

Edit: I will open a separate PR for the changes unrelated to #7726, to avoid putting a lot of stuff in stable.

Edit 2: That didn't work.

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -322,7 +322,7 @@ static:
     multi-threaded programs. See e.g.
     $(LINK2 https://www.gnu.org/software/libc/manual/html_node/Environment-Access.html#Environment-Access, glibc).
     */
-    void remove(scope const(char)[] name) @trusted nothrow @nogc // TODO: @safe
+    void remove(scope const(char)[] name) @trusted nothrow @nogc
     {
         version (Windows)    SetEnvironmentVariableW(name.tempCStringW(), null);
         else version (Posix) core.sys.posix.stdlib.unsetenv(name.tempCString());
--- a/std/process.d
+++ b/std/process.d
@@ -1291,12 +1291,7 @@ private Pid spawnProcessWin(scope const(char)[] commandLine,
     if (!CreateProcessW(null, commandLine.tempCStringW().buffPtr,
                         null, null, true, dwCreationFlags,
                         envz, workDir.length ? pworkDir : null, &startinfo, &pi))
-    {
-        if (GetLastError() == ERROR_FILE_NOT_FOUND)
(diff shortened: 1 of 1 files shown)


CyberShadow:
I am not comfortable about putting all of this on stable.

Maybe we should revert #7726 on stable and put it back on master.

thewilsonator (approved):

std/process.d
```diff
@@ -1434,28 +1429,39 @@ version (Windows) @system unittest
...
-    if (pathz == null)  return null;
+    string result;
 
-    foreach (dir; splitter(to!string(pathz), ':'))
-    {
-        auto execPath = buildPath(dir, executable);
-        if (isExecutable(execPath))  return execPath;
-    }
+    environment.getImpl("PATH",
+        (scope const(char)[] path)
```
MartinNowak, line 1441: The callback here is not `@safe` and apparently fails to compile with ldc-1.24.0 on Linux-32.
```
cannot pass argument `__lambda2` of type `void delegate(scope const(char)[] path) @system` to parameter `scope void delegate(const(char)[]) @safe sink`
```

MartinNowak: After making it `@safe`
`std/process.d(1448): Error: scope variable dir assigned to execPath with longer lifetime`

CyberShadow: If I add `@safe` explicitly, it compiles for me.

How can I reproduce this error?

What error do you get if you add an explicit `@safe` annotation?

CyberShadow: > How can I reproduce this error?

Reproduced with `digger run master -- dmd -o- -preview=dip1000 process.d`

CyberShadow: Looks like something changed in `dip1000` on master. Not yet sure if this is a regression. Bisecting.

CyberShadow: BTW, feel free to revert this if it's blocking anything.

CyberShadow: Wow, it was actually caused by a regression in Phobos.

Introduced in #7719.

I suggest reverting that pull request, since it was intended as a regression.

CyberShadow: https://github.com/dlang/phobos/pull/7730

MartinNowak: See https://github.com/dlang/phobos/pull/7731 @CyberShadow

CyberShadow: I think we should prefer #7730 as that addresses a regression visible from outside Phobos.

PR #7730, opened 2020-12-19
Revert "Remove some unnecessary undocumented explicit ctors in std.range"

CyberShadow:
Reverts dlang/phobos#7719

https://github.com/dlang/phobos/pull/7719#issuecomment-748463165

Blocks merging stable into master after https://github.com/dlang/phobos/pull/7727.

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -928,6 +928,14 @@ if (Ranges.length > 0 &&
             // TODO: use a vtable (or more) instead of linear iteration
 
         public:
+            this(R input)
+            {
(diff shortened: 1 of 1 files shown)


MartinNowak (approved):
Thanks, makes sense that ctors could mess with scope inference in more complicated cases.

MartinNowak:
[jacob-carlborg/dlp failure](https://buildkite.com/dlang/phobos/builds/4233#07771254-a506-448c-9805-4aea5c899495) seeems unrelated.

PR #7731, opened 2020-12-19
workaround scope inference problem leading ldc-1.24.0 compile error

MartinNowak:
- fixed by using explicit for loop

Unfortunately didn't have enough time to reduce this to a proper bug report.


MartinNowak:
Of course Vladimir was faster, closing for #7730

PR #7732, opened 2020-12-20
fix issue 21493 - Documentation broken hyperlink std.stdio

aG0aep6G:


PetarKirov (approved):
Thanks for fix. The OP originally reported this on the @dlang-tour/core issue tracker, but I wasn't sure to which part he was specifically referring to.

PR #7734, opened 2020-12-25
Add return scope annotation to TempCStringBuffer & fix std.process

Geod24:
Currently stable does not compile because of some scope mismatch.
Using `tempCString()` works around the problem while still avoiding allocations.
`isExecutable` was removed since it was only used in one place.


CyberShadow:
Have you seen #7730 / #7731? Seems related.

CyberShadow:
I think cherry-picking or merging #7730 to stable would be a more correct fix?

Geod24:
Thanks for the pointer, I had missed it. Opened https://github.com/dlang/phobos/pull/7735

PR #7740, opened 2021-01-04
Microsoft Windows std.stdio.File.lock(), tryLock(),  unlock(): do not allocate memory for error messages when they are not needed

n8sh:


CyberShadow:
Why not just make the `wenforce` argument `lazy`?

n8sh:
Because I wanted to be able to add `pragma(inline, false)` and have it only affect the cold path. In your opinion was the other way better?

CyberShadow:
From a code readability standpoint, `enforce` variants are better than explicit `if` constructs. `if` being a control flow structure should be used for important aspects of the function's algorithm, but when it's used for trivial error handling, it becomes a distraction.

`lazy` would put the code constructing the error message string into its own function (implicit lambda), so that should achieve something similar, right?

In any case, optimizing without benchmarking is usually not practical, and optimizing CPU for functions doing I/O is usually not very impactful.

n8sh:
I have made the suggested change.

CyberShadow (approved):

CyberShadow:
Thanks!

PR #7743, opened 2021-01-10
Make std.range.enumerate @nogc. 

maxhaton:
std.conv is not nogc but it's not really neccessary here as far as I can see (a manual > check passes all existing tests).

### DIFF
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -10470,7 +10470,7 @@ pure @safe unittest
 }
 @nogc @safe unittest
 {
-	const val = iota(1, 100).enumerate(1);
+   const val = iota(1, 100).enumerate(1);
 }
 // https://issues.dlang.org/show_bug.cgi?id=10939
 version (none)
--- a/std/range/package.d
+++ b/std/range/package.d
@@ -10197,16 +10197,17 @@ in
         {
             alias signed_t = Largest!(Enumerator, Signed!LengthType);
             signed_t signedLength;
+            //This is to trick the compiler because if length is enum
+            //the compiler complains about unreachable code.
+            auto getLength()
+            {
+                return range.length;
+            }
+            import std.stdio;
+            debug writeln(getLength() < signed_t.max);
             //Can length fit in the signed type
-            overflow = range.length > signed_t.max;
-            /*
-            //This isn't nogc(but should be...)
-            try signedLength = to!(typeof(signedLength))(range.length);
-            catch (ConvException)
-                overflow = true;
-            catch (Exception)
-                assert(false);
-            */
+            assert(getLength() < signed_t.max, "a signed length type is required but the range's length() is too great");
+            signedLength = range.length;
             auto result = adds(start, signedLength, overflow);
         }
         else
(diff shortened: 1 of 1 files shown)


maxhaton:
@wilzbach @andralex @JackStouffer CI's (was?) broken, tests pass for the ones that work

atilaneves:
> tests pass for the ones that work

?

std/range/package.d
```diff
@@ -10195,13 +10195,18 @@ in
...
         else static if (isSigned!Enumerator)
         {
-            Largest!(Enumerator, Signed!LengthType) signedLength;
+            alias signed_t = Largest!(Enumerator, Signed!LengthType);
+            signed_t signedLength;
+            //Can length fit in the signed type
+            overflow = range.length > signed_t.max;
+            /*
+            //This isn't nogc(but should be...)
             try signedLength = to!(typeof(signedLength))(range.length);
```
RazvanN7, line 10204: This line also assigns `signedLength` to `range.length`, however this is omitted in the proposed change. Is this not going to affect the code?

std/range/package.d
```diff
@@ -10463,7 +10468,10 @@ pure @safe unittest
         }
     }}
 }
-
+@nogc @safe unittest
```
RazvanN7, line 10471: could the other tests be marked as @nogc as well?

RazvanN7 (approved):

maxhaton:
> > tests pass for the ones that work
> 
> ?
@atilaneves 
The CI was broken due some cert issue (not my area of expertise)

std/range/package.d
maxhaton: I don't think it effects the actual behaviour of anything apart from the check in the static if. 

I'll fix it - there's no test case for this so I'll try and come up with one.

std/range/package.d
```diff
@@ -10195,13 +10195,18 @@ in
             auto result = adds(start, range.length, overflow);
         else static if (isSigned!Enumerator)
         {
-            Largest!(Enumerator, Signed!LengthType) signedLength;
+            alias signed_t = Largest!(Enumerator, Signed!LengthType);
```
atilaneves, line 10198: I'm not sure it's worth giving this type a name. But if it is, then it's worth replacing the `typeof(signedLength)` below with it.

std/range/package.d
```diff
@@ -10195,13 +10195,18 @@ in
...
+            signed_t signedLength;
+            //Can length fit in the signed type
+            overflow = range.length > signed_t.max;
+            /*
```
atilaneves, line 10202: The diff is commenting out code?

maxhaton: I left it in to show what it's supposed to be replacing, I can get rid if you want. 

I'll do it this evening but I think the original code might have been broken in the first place because of the exception causing overflow to be set to true but then immediately trampling over the value anyway in the checkedint operation. I need to fix it and write a test case that actually overflows.

atilaneves: The history of the codebase is preserved in version control, please don't add commented out code.

RazvanN7 (approved):

PR #7744, opened 2021-01-14
Temporarily comment one test

RazvanN7:
This is currently blocking this PR [1] in dmd.

The commented test looks like this:

```d
static struct S2
{    
    int i;
    @disable this(this);
    this(ref S2){}
}    
S2 s2 = void;
static assert(!__traits(compiles, emplace(&s2, 1)));       // this line is commented                                                                             
emplace(&s2, S2.init);
```

Emplace is not able to run this test because the copy constructor cannot be called with an rvalue (1) and its definition has masked the default constructor.

With the fix in [1], the copy constructor no longer disables default construction and therefore emplace is able to work in this scenario.

I commented the code because it is currently blocking [1], however, after [1] is merged I will uncomment it and make it a runnable test.

[1] https://github.com/dlang/dmd/pull/12132

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -5707,7 +5707,7 @@ if (!is(T == class))
         this(ref S2){}
     }
     S2 s2 = void;
-    static assert(!__traits(compiles, emplace(&s2, 1)));
+    //static assert(!__traits(compiles, emplace(&s2, 1)));
     emplace(&s2, S2.init);
 
     static struct SS1


12345swordy (approved):
LGTM

Geod24 (approved):

pbackus:
Can we backport this to `stable`? This test is causing all PRs to `stable` (e.g. #7813) to fail now that dlang/dmd#12132 has been merged.

PR #7745, opened 2021-01-17
Get rid of std.conv.emplace[Ref](), use core[.internal].lifetime

kinke:
The `emplace()` stuff was moved to druntime; for some reason, it's still in Phobos.

I've diffed the two versions, and they are still almost identical (incl. unittests); the druntime version appears to have seen some improvements (e.g., forwarding r/lvalueness of the arguments) in the meantime.


kinke:
The Phobos tests pass on Linux with the druntime PR above.

std/conv.d
```diff
@@ -4984,1040 +4984,8 @@ private bool isOctalLiteral(const string num)
...
-        {
-            int i = 0;
-            static S3 opCall(ref S3){assert(0);}
-        }
-        S3 s = void;
-        emplace(&s, S3.init);
-    }
-}
+// emplace() used to be here but was moved to druntime
+public import core.lifetime : emplace;
```
kinke, line 4988: Open for better suggestions...

RazvanN7: I think it would be best to issue a deprecation that suggests using `import core.lifetime`/`core.inernal.lifetime` instead of `std.conv` when using `emplace`/`emplaceRef`. That would also require adding a changelog entry.

RazvanN7 (changes requested):
Please issue a deprecation uses of std.conv.emplace and emplaceRef

kinke:
I've checked; it was moved with v2.085 (https://github.com/dlang/druntime/commit/5ceab47dd2c24bbccb049897c2784403366f7449). Is that old enough to justify a deprecation? I'd prefer not bugging codebases which want to keep support with older compiler versions about using a harmless (& undocumented) alias.

RazvanN7 (approved):
Fair enough.

RazvanN7:
Let's wait a few more days, maybe someone else has a better alternative. If not, we'll merge this as is.

PR #7747, opened 2021-01-17
Fix Issue 20508 - std.math.pow(-infinity, y) does not return NaN for imaginary or complex results

berni44:
This is a copy of #7363. I'm the original author, but as I deleted my old account, I can't do any changes there.

I rebased everything and added the requested changes to the documentation unittest.


RazvanN7 (approved):
LGTM. cc @wilzbach

PR #7748, opened 2021-01-17
Fix Issue 13663 - Comparison of Tuples with floating point fields

berni44:
This is a copy of #7301. I'm the original author, but as I deleted my old account, I can't do any changes there.

I rebased and replaced "issue 13663" by a link to bugzilla.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -799,21 +799,42 @@ if (distinctFieldNames!(Specs))
          *           for comparison between `Tuple`s.
          *
          * Returns:
-         * For any values `v1` on the right-hand side and `v2` on the
-         * left-hand side:
+         * For any values `v1` contained by the left-hand side tuple and any
+         * values `v2` contained by the right-hand side:
+         *
+         * 0 if `v1 == v2` for all members or the following value for the
+         * first position were the mentioned criteria is not satisfied:
          *
          * $(UL
-         *   $(LI A negative integer if the expression `v1 < v2` is true.)
-         *   $(LI A positive integer if the expression `v1 > v2` is true.)
-         *   $(LI 0 if the expression `v1 == v2` is true.))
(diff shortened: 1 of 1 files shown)


std/typecons.d
```diff
@@ -799,35 +799,73 @@ if (distinctFieldNames!(Specs))
          *           for comparison between `Tuple`s.
          *
          * Returns:
-         * For any values `v1` on the right-hand side and `v2` on the
-         * left-hand side:
+         * For values `v1` on the left-hand side and `v2` on the right-hand side:
+         *
+         * 0 if the expression `v1 == v2` is true for all members, or else for
+         * the first member, where this is not true:
```
RazvanN7, line 805: I would rephrase this to:

"For any values `v1` contained by the left-hand side tuple and any values `v2` contained by the right-hand side:

0 if `v1 == v2` for all members or the following value for the first position were the mentioned criteria is not satisfied:"

RazvanN7 (approved):

std/typecons.d
berni44: done - thanks

std/typecons.d
```diff
@@ -799,35 +799,74 @@ if (distinctFieldNames!(Specs))
          *           for comparison between `Tuple`s.
          *
          * Returns:
-         * For any values `v1` on the right-hand side and `v2` on the
-         * left-hand side:
+         * For any values v1 contained by the left-hand side tuple and any
```
RazvanN7, line 802: Sorry for nit picking, but can you use use quotes (`) for expressions: v1, v2, v1 == v2  etc. ?

berni44: Oh, no nit picking, that's fine. I need to find my way into the D conventions again... Will fix this soon.

PR #7749, opened 2021-01-17
Fix Issue 20539 - std.conv.to: internal overload conflict for enums with base types that have a catch-all opEquals overload (?)

berni44:
This is a copy of #7378. I'm the original author, but as I deleted my old account, I can't do any changes there.

I rebased and replaced "issue 20539" by a link to bugzilla.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -1937,7 +1937,9 @@ $(UL
 */
 private T toImpl(T, S)(S value)
 if (isInputRange!S && isSomeChar!(ElementEncodingType!S) &&
-    !isExactSomeString!T && is(typeof(parse!T(value))))
+    !isExactSomeString!T && is(typeof(parse!T(value))) &&
+    // issue 20539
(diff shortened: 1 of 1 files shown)


std/conv.d
```diff
@@ -2080,6 +2082,49 @@ if (is(T == enum) && !is(S == enum)
     assert(m1 == [[En8143.A, En8143.C], [En8143.C, En8143.A]]);
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=20539
+@safe pure unittest
+{
+    // needs to be static, because else a frame pointer issue makes
+    // it work unintentionally; see next unittest for the same test
+    // with the struct being not static
```
RazvanN7, line 2090: This comment needs to be reworded a bit. How does this sound: "To test that the bug is fixed it is required that the struct is static, otherwise, the frame pointer makes the test pass even if the bug is not fixed"?

Also, there is no point in having two separate unittests. Please merge them and add an additional comment to the second test specifying that this is a guard for the future.

RazvanN7 (approved):
Thanks for this! Just fix the nit below.

PR #7751, opened 2021-01-18
Replace approxEqual with isClose

berni44:
This is a copy of #7340. I rebased everything on master, removed conflicts and fixed one new place where approxEqual sneaked in in the last year.

### DIFF
--- a/std/algorithm/comparison.d
+++ b/std/algorithm/comparison.d
@@ -985,7 +985,7 @@ template equal(alias pred = "a == b")
 @safe @nogc unittest
 {
     import std.algorithm.comparison : equal;
-    import std.math : approxEqual;
+    import std.math : isClose;
 
     int[4] a = [ 1, 2, 4, 3 ];
     assert(!equal(a[], a[1..$]));
@@ -998,8 +998,8 @@ template equal(alias pred = "a == b")
     assert(equal(a[], b[]));
 
     // predicated: ensure that two vectors are approximately equal
-    double[4] c = [ 1.005, 2, 4, 3];
-    assert(equal!approxEqual(b[], c[]));
+    double[4] c = [ 1.0000000005, 2, 4, 3];
+    assert(equal!isClose(b[], c[]));
 }
 
 /++
@@ -1023,7 +1023,7 @@ range of range (of range...) comparisons.
     import std.algorithm.iteration : map;
     import std.internal.test.dummyrange : ReferenceForwardRange,
         ReferenceInputRange, ReferenceInfiniteForwardRange;
-    import std.math : approxEqual;
+    import std.math : isClose;
 
     // various strings
     assert(equal("æøå", "æøå")); //UTF8 vs UTF8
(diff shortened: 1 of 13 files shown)


ibuclaw:
If I haven't already said on the previous PR, it would be good to have testing done on ARM64 at the very least to confirm that the new limits also suit 128-bit real platforms, and possibly a 64-bit real target as well.  `approxEqual` has already been widely tested, so the fewer number of times we have to fiddle with these unit-tests the better.

berni44:
> If I haven't already said on the previous PR, it would be good to have testing done on ARM64 at the very least to confirm that the new limits also suit 128-bit real platforms, and possibly a 64-bit real target as well.

Any ideas, how we could do that? (And yes, I support this, but I have no means to do so.)

> `approxEqual` has already been widely tested, so the fewer number of times we have to fiddle with these unit-tests the better.

I disagree here. `approxEqual` has a large flaw, because it pretends to guard against mistakes, but doesn't in all cases. It performs (with defaults) only a very weak check. In case, a unittest fails with `isClose`, we should be happy, because this high likely points to a bug, that was left undetected before.

ibuclaw:
> > `approxEqual` has already been widely tested, so the fewer number of times we have to fiddle with these unit-tests the better.
> 
> I disagree here. `approxEqual` has a large flaw, because it pretends to guard against mistakes, but doesn't in all cases. It performs (with defaults) only a very weak check. In case, a unittest fails with `isClose`, we should be happy, because this high likely points to a bug, that was left undetected before.

Let me clarify.  All unit-tests pass on Linux/arm32, Linux/arm64, Solaris/sparc, FreeBSD/i386, and a number of others with the current `approxEqual` checks.

berni44:
> > > `approxEqual` has already been widely tested, so the fewer number of times we have to fiddle with these unit-tests the better.
> > 
> > 
> > I disagree here. `approxEqual` has a large flaw, because it pretends to guard against mistakes, but doesn't in all cases. It performs (with defaults) only a very weak check. In case, a unittest fails with `isClose`, we should be happy, because this high likely points to a bug, that was left undetected before.
> 
> Let me clarify. All unit-tests pass on Linux/arm32, Linux/arm64, Solaris/sparc, FreeBSD/i386, and a number of others with the current `approxEqual` checks.

Yes, that's clear. My point is, that the strength of unittests using `approxEqual` is sometimes very weak. An even weaker unittest `assert(true);` would also pass on all these platforms. But of course it won't check much.

std/algorithm/comparison.d
```diff
@@ -998,8 +998,8 @@ template equal(alias pred = "a == b")
     assert(equal(a[], b[]));
 
     // predicated: ensure that two vectors are approximately equal
-    double[4] c = [ 1.005, 2, 4, 3];
-    assert(equal!approxEqual(b[], c[]));
+    double[4] c = [ 1.0000000005, 2, 4, 3];
```
RazvanN7, line 1001: Why is this test changed?

std/parallelism.d
```diff
@@ -4565,8 +4565,12 @@ version (StdUnittest)
 // tons of stuff and should not be run every time make unittest is run.
 version (parallelismStressTest)
 {
-    @safe unittest
+    @system unittest
```
RazvanN7, line 4568: why is this unittest becoming `@system` ?

std/parallelism.d
```diff
@@ -4565,8 +4565,12 @@ version (StdUnittest)
...
     {
+        import std.stdio : stderr, writeln, readln;
+        import std.range : iota;
+        import std.algorithm.iteration : filter, reduce;
```
RazvanN7, line 4572: Why is this PR adding these imports? I see that these functions were used before adding these imports here.

std/parallelism.d
```diff
@@ -4648,8 +4652,14 @@ version (parallelismStressTest)
 
     // These unittests are intended more for actual testing and not so much
     // as examples.
-    @safe unittest
+    @system unittest
```
RazvanN7, line 4655: Same comments as for the previous unittest.

std/parallelism.d
```diff
@@ -4776,6 +4787,7 @@ version (parallelismStressTest)
                     uselessTask.workForce();
                 }
             }
+             +/
```
RazvanN7, line 4790: unrelated change. please undo.

RazvanN7 (approved):
Looks good. See my comments.

RazvanN7:
@ibuclaw Our testing pipeline seems fine. How can we move forward here? How can we test this in a satisfactory way?

PR #7754, opened 2021-01-18
container: array: move memory reserve assert to the correct scope

ljmf00:
Verification for faulty reserve() should only occur if a reserve() actually
occur.


ljmf00:
Maybe I'm wrong about this but not sure. Is this to make sure if there's space without a `reserve()`? If yes, this should be closed.

RazvanN7:
The check is left outside because you might end up with a bugish case where `capacity < length`. If the check is moved inside the `if` scope, then it will not be catched.

Closing.

PR #7756, opened 2021-01-20
Rewrite std.meta.staticMap using AliasAssign

WalterBright:
Try out https://github.com/dlang/dmd/pull/12029

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -787,29 +787,68 @@ template DerivedToFront(TList...)
     static assert(is(TL == AliasSeq!(C, B, A)));
 }
 
-private enum staticMapExpandFactor = 150;
-private string generateCases()
+version = UseAliasAssign; // use new alias assign feature to speed up recursive templates
+
+version (UseAliasAssign)
 {
-    string[staticMapExpandFactor] chunks;
-    chunks[0] = q{};
-    static foreach (enum i; 0 .. staticMapExpandFactor - 1)
-        chunks[i + 1] = chunks[i] ~ `F!(Args[` ~ i.stringof ~ `]),`;
-    string ret = `AliasSeq!(`;
-    foreach (chunk; chunks)
-        ret ~= `q{alias staticMap = AliasSeq!(` ~ chunk ~ `);},`;
-    return ret ~ `)`;
 }
-private alias staticMapBasicCases = AliasSeq!(mixin(generateCases()));
+else
+{
+    private enum staticMapExpandFactor = 150;
+    private string generateCases()
+    {
+        string[staticMapExpandFactor] chunks;
+        chunks[0] = q{};
+        static foreach (enum i; 0 .. staticMapExpandFactor - 1)
+            chunks[i + 1] = chunks[i] ~ `F!(Args[` ~ i.stringof ~ `]),`;
+        string ret = `AliasSeq!(`;
+        foreach (chunk; chunks)
+            ret ~= `q{alias staticMap = AliasSeq!(` ~ chunk ~ `);},`;
+        return ret ~ `)`;
+    }
+    private alias staticMapBasicCases = AliasSeq!(mixin(generateCases()));
+}
 
 /**
 Evaluates to $(D AliasSeq!(F!(T[0]), F!(T[1]), ..., F!(T[$ - 1]))).
  */
 template staticMap(alias F, Args ...)
 {
-    static if (Args.length < staticMapExpandFactor)
-        mixin(staticMapBasicCases[Args.length]);
+    version (UseAliasAssign)
+    {
+        alias A = AliasSeq!();
+
+        static foreach (size_t i; 0 .. Args.length / 8)
+        {
+            A = AliasSeq!(A, F!(Args[i * 8    ]), F!(Args[i * 8 + 1]), F!(Args[i * 8 + 2]), F!(Args[i * 8 + 3]),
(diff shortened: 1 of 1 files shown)


WalterBright:
The error is:
```
std/conv.d(5710): Error: static assert:  `!__traits(compiles, emplace(& s2, 1))` is false
```
but I'm not seeing where `staticMap` is being called?

WalterBright:
@9il can you please have a look at the buildkite failures in Mir?

WalterBright:
dscanner fails with the new syntax (buildkite)

WalterBright:
The buildkite druntime error is:
```
generated/linux/release/64/unittest/test_runner core.sys.windows.dde
--| generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D8runbench6runCmdFAyabIAaZQi: error: undefined reference to '_D3std7process12executeShellFNeMAxaxHAyaAyaEQBqQBp6ConfigmMQBbQzZSQCm8typecons__T5TupleTiVQCba6_737461747573TQCuVQCya6_6f7574707574ZQBz'
```

WalterBright:
@s-ludwig In buildkite taggedalgebraic I'm getting an error:
```
source/taggedalgebraic/taggedunion.d(449,13): Error: function `taggedalgebraic.taggedunion.TaggedUnion!(E).TaggedUnion.set!E.text.set()` is not callable using argument types `(string)`
```

9il:
@WalterBright

Looks like the following `mir-core` unittest is failed because of a syntax error (regresssion), that is probably is unrelated to this MR. The CI compiler can't handle `static F[3][] vals(F) =   ...` expression anymore.

```
@safe @nogc nothrow version(mir_core_test) unittest
{
    import std.meta: AliasSeq;
    static F[3][] vals(F) =    // value,exp,ldexp
    [
    [    0,    0,    0],
    [    1,    0,    1],
    [    -1,    0,    -1],
    [    1,    1,    2],
    [    123,    10,    125952],
    [    F.max,    int.max,    F.infinity],
    [    F.max,    -int.max,    0],
    [    F.min_normal,    -int.max,    0],
    ];
    static foreach(F; AliasSeq!(double, real))
    {{
        int i;

        for (i = 0; i < vals!F.length; i++)
        {
            F x = vals!F[i][0];
            int exp = cast(int) vals!F[i][1];
            F z = vals!F[i][2];
            F l = ldexp(x, exp);
            assert(feqrel(z, l) >= 23);
        }
    }}
}
```

`mir-algorithm` is failed because of another thing, I did a new release to so it will print at least what UDA keys are reflected with the new `staticMap`.

s-ludwig:
>     source/taggedalgebraic/taggedunion.d(449,13): Error: function `taggedalgebraic.taggedunion.TaggedUnion!(E).TaggedUnion.set!E.text.set()` is not callable using argument types `(string)`

It looks like there is a dependency cycle between `staticMap!(TypeOf, EnumMembers!FieldEnum)` (which is indexed to declare the parameter type for that `set` method) and the template `TypeOf`. This works right now, but seems fishy anyway. I'll have to take a closer look tomorrow.

WalterBright:
@s-ludwig thanks! Though this should behave identically to the old implementation of staticMap, warts and all.

maxhaton:
[out.pdf](https://github.com/dlang/phobos/files/5845875/out.pdf)

Plot of dumb timings of staticMap!(Alias, int, int, int [up to 8192 times]). Starts off better, scales worse (I assume due to memory usage, even with lowmem the heap sizes are enormous)

I'll do some heavier operations later.

WalterBright:
@maxhaton Good information. I bet the result is because the old is using a binary method to build long alias sequences, while the new is using the sequential one. Changing the implementation of staticMap to a binary method should fix that.

maxhaton:
@WalterBright more perf plots 

Same N values for fullyQualifiedName (slower, more memory usage) [fullyQualifiedNameData.pdf](https://github.com/dlang/phobos/files/5845950/fullyQualifiedNameData.pdf)

First few 150 plotted in more detail [fullyQualifiedNameData_linear.pdf](https://github.com/dlang/phobos/files/5845952/fullyQualifiedNameData_linear.pdf)

[aliasplots_linear.pdf](https://github.com/dlang/phobos/files/5845954/aliasplots_linear.pdf) Very clear division in the log-log plot

WalterBright:
The old staticMap had changed since I last worked with it. Instead of calling itself, it used a function to generate a string that enumerated all the cases, and did one big (up to 150) case AliasSeq. Unfortunately, that implementation traded off the recursive call for a rather large set of compile-time allocations in generating the string.

WalterBright:
@9il @s-ludwig thank you for fixing buildkite!

9il:
> @9il @s-ludwig thank you for fixing buildkite!

Something has been fixed in DMD / Phobos. I didn't really fix anything.

9il:
I have accidentally edited @WalterBright's comment and was surprised that I am able to do so. That is a bit weird.

WalterBright:
@9il that's because you have privileged access. It's a superpower, I only used it once when the comment writer asked me to modify it.

s-ludwig:
Same here, might be because all of the sequences in those `taggedalgebraic` unit tests are <= 8 entries and the issue doesn't manifest for the first assignment.

RazvanN7:
@WalterBright please rebase? Also, what are the plans with AliasAssign and phobos?

RazvanN7:
I'm going to close this as #8039 seems to be a superior implementation.

PR #7757, opened 2021-01-20
Partial replace call to snprintf for formating floatingpoint numbers for %f and %F

berni44:
This is a copy of #7264 and a followup on #7318 and #7285.

I rebased it, added some minor fixes and a test that can be run to compare the result of snprintf and the result of printFloat. 

To run the test uncomment the line `// version = printFloatTest;` near the end of the file and run a normal unittest. This takes 30 minutes and compares random input. Some known bugs of `snprintf` are written at the top of the test section. They will, of course, be reported as differences.

For more information see the original PR. I was able to restore the PDF document mentioned there, thanks to @skoppe and @m3m0ry who saved it.

@thewilsonator asked there one question I didn't answer yet:
> Following the success of the float zero factorisation, I'm beginning to think that the best way to organise the printing is not by format specifier (%a,%f,%g,%e) but by "category" (for want of a better term), i.e.
> 
> * inf/nan (handled above),
> * Algorithm A: For large exponents (exp >= T.mant_dig)
> * Algorithm B: For small exponents (exp < T.mant_dig - 61)
> * Algorithm C: For exponents close to 0. (already done)
> 
> Since there seems to be much more in common between the different ranges of numbers than there is between the different format specifiers.

Yes, you're right. I allready thought about this too (and `snprintf` does this too). But I'm a little bit reluctant do so immediately for two reasons: First, I fear, that the result would be slower because several `if` switches might be needed. Second and more important, I have some ideas how cases A and B can be speed up. Implementing these ideas might be much more difficult, when I have to code for all specifiers simultaneously.


berni44:
I don't know why the bot thinks issue 20636 is touched here. This is no enhancement!

RazvanN7:
Commit message of 686c5e5 contains the following text:

"Bug **20636** has meanwhile been **fixed**, so no hack needed anymore
Remove two missleading comments (there is no more cast needed)
Adapt new convention for naming **issues**
Add missing spaces arround operators"

The bolded words have triggered the bot. Maybe you can rephrase the commit message to avoid those words.

berni44:
The number was wrong anyway: 20636 instead of the correct 20363... Now, at least, he added the bug fix label, but the enhancement remains... Maybe, someone with write access could just delete the enhancement label?

schveiguy:
I mean, it *is* an enhancement, no? Or does this not remove the requirement of snprintf?

berni44:
> I mean, it _is_ an enhancement, no? Or does this not remove the requirement of snprintf?

Well, it replaces `snprintf` (partially). But is that already an enhancement? Is it an enhancement replacing quicksort by timsort internally? I don't know. Might depend on the definition of enhancement...

PetarKirov:
Making Druntime and Phobos more independent of libc (even partially) is an important improvement in my book.

RazvanN7 (approved):
Approving the idea, however, this needs a proper review.

std/format.d
```diff
@@ -8442,35 +8433,1043 @@ printFloat_done:
...
+    //   This time, we know, that the integral part and the fractional part each fit into a
+    //   ulong. The mantissa might be partially in both parts or completely in the fractional
+    //   part.
+    //
+    //   We first calculate the integral part by consecutive division by 10. Then we calculate
+    //   the fractional part by consecutive multiplication by 10. Again only until we have enough
+    //   digits. Finally, we decide the rounding type, mainly by looking at the next digit.
+
+    ulong[18] bigbuf;
+    if (exp >= T.mant_dig)
```
ibuclaw, line 8543: You mention "does not fit into a ulong" below.  Are you assuming that `mant_dig` is always <= 64?  What happens when the mant_dig is 106 or 113?

berni44: This PR targets only floats, doubles and 64bit reals. See the template constraints of this function. 
The "does not fit" comment refers to the unrolled number, where all the zeros (counted in `exp`) are attached either left or right to the mantissa; these are really large numbers, taking up to a little bit more than 1000 bits.

berni44:
I wrote some explanations accompanied by source code snippets and illustrations:
https://github.com/berni44/printFloat/blob/master/walkthrough.pdf
I hope, this will make the review easier and tear down the fears...

berni44:
Resolved merge conflicts.

With PR #7762 being merged, printing floats with %f will automatically be CTFEable too. I added unittests and corrected the error message.

I also fixed a small bug in an other unittest - FloatingPointControl is not available on targets with soft float.

std/format.d
```diff
@@ -8427,42 +8430,1050 @@ printFloat_done:
...
+            foreach (i;lsu .. count)
+            {
+                mybig[i] = mybig[i] * 10 + over;
+                over = mybig[i] >> 60;
+                mybig[i] &= (1L << 60) - 1;
+            }
+            if (mybig[lsu] == 0)
+                ++lsu;
+
+            buffer[right++] = cast(byte) ('0'+over);
```
thewilsonator, line 8622: This PR has a number of inconsistent (missing) whitespace between binary operands, but otherwise looks fine.

thewilsonator (approved):
See comments

jondegenhardt:
This PR ended up fixing a case of incorrect rounding occurring on Windows in the tsv-utils test suite. 👍

PR #7759, opened 2021-01-21
Fix Issue 20552 - Deprecated Nullable.get warning with Appenders

RazvanN7:
`Nullable!T` has the following assignment operator: `opAssign()(T value)`. This `opAssign` can be called for both `Nullable!T` and `T` because of the inner `alias get this` of Nullable. However, when a Nullable is assigned to another Nullable
a bit copy is preferred rather then an alias this conversion. Despite this, `hasElaborateAssign` specifically calls opAssign thus triggering the `alias this`. To bypass this, I used the solution proposed by @FeepingCreature and added an opAssign overload that receives a `Nullable!T`. This does get rid of the deprecation, however, I don't know how to test this because when compiled with -de, the deprecation does not appear (and phobos is compiled that way). Suggestions?

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -2942,6 +2942,22 @@ struct Nullable(T)
     }
 
     /**
+     * If value is null, sets this to null, otherwise assigns
+     * `value.get` to the internally-held state. If the assignment
+     * succeeds, `this` becomes non-null.
+     *
+     * Params:
+     *     value = A value of type `Nullable!T` to assign to this `Nullable`.
+     */
+    void opAssign()(Nullable!T value)
+    {
+        if (value._isNull)
+            nullify();
+        else
+            opAssign(value.get());
+    }
+
+    /**
      * If this `Nullable` wraps a type that already has a null value
      * (such as a pointer), then assigning the null value to this
      * `Nullable` is no different than assigning any other value of


FeepingCreature:
Isn't it time to remove `.get` yet?

Ah, "after 2.096". Soon... sooon...

PetarKirov:
> Isn't it time to remove `alias get this` yet?

Yes, since according to [DIP1013](https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1013.md):

> [..] a comment of the format @@@DEPRECATED_[version]@@@ should be added to the top of the code which is to be removed or disabled, where the version number is the version where the feature will be fully removed from the language or standard library. [..]

```
//@@@DEPRECATED_2.096@@@
```
=> `alias get this` should not be part of dmd 2.096.0, so the time to remove it is now.

> impatiently hovers finger over `$ git push upstream fix/remove-alias-get-this _`

 `Enter`! 😛

Geod24:
> _impatiently hovers finger over `$ git push upstream fix/remove-alias-get-this _`_

Please don't push to upstream :D

> => `alias get this` should not be part of dmd 2.096.0, so the time to remove it is now.

Looking at the release notes, it was deprecated in [v2.088.0](https://dlang.org/changelog/2.088.0.html#remove-nullable-alias-get-this). So it should go with v2.098.0. However, I have a feeling most (if not all) stakeholders will be happy when this is gone. And non-stakeholders will also be very happy to see an unrelated deprecation message gone.

However, the solution brought by @RazvanN7 is simple, can be put into stable (next point release is nearing), so I think this should be merged into stable regardless.

Geod24 (approved):
Please target stable, and feel free to self-merge.

RazvanN7:
`std/conv.d(5710): Error: static assert:  `!__traits(compiles, emplace(& s2, 1))` is false`

I've seen this failure on other PRs. I think it is unrelated to this change.

Geod24:
Since it triggers on all Cirrus workflow, I don't think we should disregard it.

schveiguy:
Just a note, the deprecation message says "removed *after* 2.096", so I would say we should target 2.097. Pushing to 2.098 is not going to make a difference, I would do 2.097, that's what everyone's been reading (over and over and over), and by now if you haven't fixed your Nullable deprecations, you probably won't.

UplinkCoder:
This PR caused a regression. in the following code:
```
import std.typecons;
struct T {
        Nullable!S s;
}
struct S {}
S f(S s)
{
  return s;
}
T f2(const(T) t) {
        T ret;
        t.s.apply!((v) {
          return ret.s = nullable(f(v));
        });
        return ret;
}
```

UplinkCoder:
I strongly suspect that your since opAssign overload is a template the opAssign inside the template body does not go through proper overload-resolution and therefore you are trying to call the function created by the template instance.

UplinkCoder:
@RazvanN7

FeepingCreature:
Afaics, `Nullable.opAssign` just returns void, and you can't return void as a value. Not sure how this would ever have worked.

FeepingCreature:
This PR causes another regression.

For struct types with invariants that contain Nullables, the invariant is now checked, because the fact that Nullable has an opAssign with its own type triggers recursive opAssign generation, which checks the invariants in every containing type.

Now that `get` is gone, is this still needed?

PR #7760, opened 2021-01-21
Make isAutoDecodableString independent of issue 21570

MoonlightSentinel:
The current implementation relies on issue 21570 to reject enums with
static arrays as their base type.

Use another is-expression instead of `isStaticArray` to detect types
that are (convertible to) static arrays.

See https://issues.dlang.org/show_bug.cgi?id=21570

---

Blocking dlang/dmd#12142

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -6770,7 +6770,7 @@ template isAutodecodableString(T)
     import std.range.primitives : autodecodeStrings;
 
     enum isAutodecodableString = autodecodeStrings &&
-        (is(T : const char[]) || is(T : const wchar[])) && !isStaticArray!T;
+        (is(T : const char[]) || is(T : const wchar[])) && !is(T : U[n], U, size_t n);
 }
 
 ///


std/traits.d
```diff
@@ -6770,7 +6770,8 @@ template isAutodecodableString(T)
     import std.range.primitives : autodecodeStrings;
 
     enum isAutodecodableString = autodecodeStrings &&
-        (is(T : const char[]) || is(T : const wchar[])) && !isStaticArray!T;
+        ((is(T : const char[]) && !is(T : const char[n], size_t n)) ||
+        (is(T : const wchar[]) && !is(T : const char[n], size_t n)));
```
PetarKirov, line 6774: I think this should work:

```D
    (is(T : const char[]) || is(T : const wchar[])) && !is(T : U[n], size_t n);
```

PetarKirov:
After some testing, I think this is the cleanest solution:

```D
template isAutodecodableString(T)
{
    import std.range.primitives : autodecodeStrings;

    enum isAutodecodableString = autodecodeStrings &&
        is(immutable T == immutable char[]) ||
        is(immutable T == immutable wchar[]);
}

unittest
{
    enum E1 : string { s = "asd" }
    enum E2 : const(char[3]) { s = "asd" }
    enum E3 : wstring { s = "asd" }

    static struct S(T)
    {
        T get() { return T.init; }
        alias get this;
    }
    
    import std.meta : AliasSeq;
    static foreach (i, T; AliasSeq!(
        string[true], wstring[true], dstring[false],
        char[2][false], const char[2][false], const(char)[2][false], wchar[2][false], dchar[2][false],
        E1[false], E2[false], E3[false],
    ))
    {{
        static if (is(T == U[expected], U, size_t expected))
        {{
            pragma(msg, U, i > 3 && i < 8 ? "\t= " : "\t\t= ", isAutodecodableString!U, "\t| isAutodecodableString!(S!U) = ", isAutodecodableString!(S!U));
            static assert(isAutodecodableString!U == expected);
            static assert(!isAutodecodableString!(S!U));
        }}
        else
            static assert(0);
    }}
}
```

https://run.dlang.io/is/7kZkPR

MoonlightSentinel:
That implementation changes the behaviour, e.g. `isAutodecodableString!E1` currently yields `true`.

PetarKirov (approved):

PetarKirov:
> That implementation changes the behaviour, e.g. isAutodecodableString!E1 currently yields true.

I didn't read the unittest right below the current implementation 🤦 and I thought that we want to exclude `enum` and `alias this` types, sorry for the confusion.

PR #7761, opened 2021-01-22
std.typecons.Nullable: Remove deprecated 'alias get this'.

FeepingCreature:
As per https://github.com/dlang/phobos/pull/7759

This is a bit early. There's three possible interpretations of the deprecation plan: the comments say "after 2.096", the deprecation notice says "in 2.096", and strictly speaking it should be removed in 2.098 because it was deprecated two major versions later than I'd originally intended.

But given all the issues we've had with the damn deprecation warning, at this stage I think people will mostly be glad just to be rid of it - especially considering this does not remove functionality, just convenience.

edit: added changelog entry

### DIFF
--- a/changelog/nullable-remove-alias-get-this.dd
+++ b/changelog/nullable-remove-alias-get-this.dd
@@ -1,0 +1,7 @@
+`std.typecons.Nullable`: Remove deprecated `alias get this`.
+
+`Nullable` no longer implicitly converts to its member.
+This feature was problematic because a simple use of a value could
+invisibly cause an assertion due to type conversion.
+To restore the previous behavior, replace uses of a `Nullable` value
+`n` with `n.get`.
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -3071,54 +3071,6 @@ struct Nullable(T)
     {
         return isNull ? fallback : _value.payload;
     }
-
-    //@@@DEPRECATED_2.096@@@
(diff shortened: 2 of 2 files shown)


Geod24:
#7759 should rid us of the message at no cost though.

FeepingCreature:
Sure, that specific one.

RazvanN7:
Looks like some buildkite projects need to be updated.

FeepingCreature:
Repushed to rerun buildkite.

RazvanN7:
I don't know what to make of this phobos failure on buildkite and CirrusCI.

schveiguy:
I think it should be 2.097. The message says after 2.096, so I don't think it's fair if anyone was waiting for 2.097 to worry about it to spring it early. I don't think we need to do it on 2.098, but I'm fine with that also.

RazvanN7:
@FeepingCreature You can try force pushing now? I just did it on my other PR and the tests passed.

FeepingCreature:
I've waited the better part of a year for this. I can wait another four months. (Though I don't have to like it! :) )

FeepingCreature:
Rebase. Is it time...?

RazvanN7:
Next beta release is 2.097 [1], so we can merge this now.

[1] https://dlang.org/changelog/release-schedule.html

RazvanN7 (approved):

PR #7762, opened 2021-01-22
Fix partially Issue 21573 - formatting floating points should be CTFE

berni44:
I just noticed, that CTFE can already be implemented independent of #7757 and followups. Of course, it will be limited to the part, where `snprintf is already replaced`, currently %e and %a applied to float and double.

Thanks to Petar Kirov and Paul Backus for help.

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -2581,7 +2581,8 @@ private void formatUnsigned(Writer, T, Char)
     assert(t2 == "[  -123] [-123  ]");
 }
 
-private enum ctfpMessage = "Cannot format floating point types at compile-time";
+private enum ctfpMessage = "Cannot format all floating point types at compile-time "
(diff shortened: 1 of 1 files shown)


std/format.d
```diff
@@ -2948,6 +2953,18 @@ useSnprintf:
            ~"0000000000000000000000000000000000000000000000000000000000000000000p+0");
 }
 
+@safe unittest
```
atilaneves, line 2956: I guess this can't be `pure` because of the floats?

berni44: Surprisingly (for my) it can be pure, while the quite similar test directly above cannot. The `static` seems to make a difference here.

RazvanN7 (approved):

berni44:
removed merge conflict with #7764

PR #7764, opened 2021-01-23
Add unittest for covering rounding mode selection.

berni44:


berni44:
I changed the test from %f to %e due to issues 20320 and 9889 (%f not working correctly on windows). For testing the rounding mode selection it doesn't matter which one to use.

RazvanN7 (approved):

PR #7765, opened 2021-01-23
Split std.format into submodules

berni44:
std.format is one of the largest files in Phobos and there is more to come.

What do you think, is it a good idea to split this into some submodules?


PetarKirov:
@berni44 Thanks for working on this. In addition to the improved code maintenance, one of the improvements that I hope we can achieve (that also affect the end users) is improved compile times, based on making the code more modular, as the compiler will need to parse and analyze less code (parsing is not really a bottleneck, it's mostly semantic analysis and also code-gen for unused functions).

In terms of organizing the work, I think it would be best if we focus on reviewing and merging this PR first (before the other PRs that affect std.format), as otherwise we're going to end of with a lot of merge conflicts.

berni44:
> In terms of organizing the work, I think it would be best if we focus on reviewing and merging this PR first (before the other PRs that affect std.format), as otherwise we're going to end of with a lot of merge conflicts.

We'll have merge conflicts either way or the other. Maybe you're right and having this first reduces the amount. Anyway I will take care of the merge conflicts and update all PRs with conflicts once an other PR was merged.

berni44:
Should I add a release note here?

RazvanN7 (approved):

RazvanN7:
I think that a changelog entry would be necessary so that people can use more granular imports.

PetarKirov:
@berni44 I forgot to mention that I didn't have time for a thorough review, but from a quick glance I think we need to separate better the code:
* FormatSpec should be in its own module `std.format.spec`
* The exception type should be in `std.format.exception`
* `format` can be in either `package.d`, or `std.format.format`.
* Try to avoid circular imports (e.g. importing `std.format` from `std.format.*`.

I think it's import to do all such changes before the next release, since once we release a new Phobos version with the new module structure, we would need to stick with it.

berni44:
> @berni44 I forgot to mention that I didn't have time for a thorough review, but from a quick glance I think we need to separate better the code:
> 
> * FormatSpec should be in its own module `std.format.spec`
> * The exception type should be in `std.format.exception`
> * `format` can be in either `package.d`, or `std.format.format`.

@PetarKirov That is probably something, that needs a more thorough debate - else some people will complain sooner or later. In my opinion we would end up with a lot of very small moduls, several of them containing only one item. That would increase the number of imports needed and might be somewhat confusing.
Having said this, I was quite unsure where to put `format` and `sformat`. I rejected `std.format.format` because this caused some name clash problems (which probably can be overcome). Anyway the repetition of `format` sounds a little bit silly: Imagine: `import std.format.format : format`. Maybe `package.d` is a better place though.

> * Try to avoid circular imports (e.g. importing `std.format` from `std.format.*`.

I think I removed all of them, but in public unittests. They are needed, because these tests are meant to be runnable standalone.

> I think it's import to do all such changes before the next release, since once we release a new Phobos version with the new module structure, we would need to stick with it.

I'm in no hurry with this PR. We could also postpone merging until march, when the next release is out. Then we've got two more months for discussing the final structure. And we've got also time for repairing, if needed.

ljmf00:
> What do you think, is it a good idea to split this into some submodules?

@berni44 I guess this should be a thing for every module of Phobos. Take the example of @ziglang https://github.com/ziglang/zig/tree/master/lib/std/math . Maybe I'll open something in the forum to discuss this.

berni44:
> avoid painful merge conflicts. Perhaps it would be best if you wait until we have a consensus on the splitting and then start from the beginning to avoid missing on any changes that were merged in the meantime

Yes, I'll do that. Anyway I'll try to keep up merging everything. Then I know the goal and I've got something to compare to later.

> @berni44 I guess this should be a thing for every module of Phobos. Take the example of @ziglang https://github.com/ziglang/zig/tree/master/lib/std/math . Maybe I'll open something in the forum to discuss this.

Yeah, I think, this discussion is beyond the scope of this PR. It should be held in a larger group, maybe even as a DIP.  And maybe even DLF should finally decide upon some sort of guideline. - I've got not the energy to start this. Nor do I have the insights into all the structures here, so I can't do that. So I would be happy, if you could start this process...

berni44:
Leaving aside the debate on how small modules should be in Phobus, I prefere a more pragmatic approach for the time being.

After the discussion above I suggest the following split:
```
package.d
read.d
write.d
spec.d
floats.d
```

In detail:
Module | Visibility | Object | Comment
--|--|--|--
package.d | public | format 
|||sformat
|||FormatException
||package|enforceFmt
|||NoOpSink
|||ctfpMessage|I would like to keep `floats.d` a private module, so this can't go there
|||formatTest|let's postpone refectoring unittests for a later PR
|||formatReflectTest|ditto
||package(std)|checkFormatException|is used in std.stdio
||private|guessLength|only used in `format()`
read.d | public | formattedRead
|||unformatValue
||undocumented public|unformatElement|not used outside std/format.d, should be private
||private|skipData
|||acceptedSpecs
|||unformatValueImpl
|||rawRead
|||unformatRange
write.d | public | formattedWrite
|||formatValue
||undocumented public | formatElement|used in std/typecons.d, should be package(std)
||private|formatValueImpl
|||formatInteger
|||formatUnsigned
|||formatRange
|||formatChar
|||hasToStringResult
|||hasToString
|||formatObject
|||getNthInt
|||getNth
|||needToSwapEndianess
|||writeAligned
spec.d | public | FormatSpec
|||singleSpec
||undocumented public|enforceValidFormatSpec|is used in two tests in dmd repo; maybe this should be changed somehow
floats.d | package | RoundingMode
|||printFloat
||private|printFloatA
|||printFloatE

There is more, that could be splitted out from package.d, but I think, this can be left for later.

IMHO best would be, to do as little as possible aside from splitting. That is: After the split: Make it run, make docs run and don't touch anything else. Of course, there is more refactoring needed (reduce global imports, sort unittests, move and improve docs), but this can be done in separate PRs, because this PR is somewhat critical: Between opening the PR and merging it, there should not be any other PR merged, that touches std.format, to avoid loosing that other PR accidentally (git doesn't manage that and the conflicts are large, so small changes got overlooked easily).

I suggest the following plan: We discuss this until 1st of March (when 2.096.0 is branched off). Then I'll do the split according to the result of our decision and make it ready for review. From 1st of March until this PR is merged, all other PRs touching std.format will be blocked. After this is merged, I'll continue with refactoring PRs that only touch one module (or two if something needs to be moved to a different module) with the goal in mind, that this is finished before 2.097.0.

PR #7766, opened 2021-01-24
Fix issue 21575 - Child processes spawned by std.process.spawnProcess accidentally inherit signal masks in parent process

tom-tan:
It simply unblocks all the signals in a forked process spawned by `spawnProcessPosix`.

Note that the unittest for it uses `std.concurrency.receiveTimeout` and waits 3 seconds in the worst case. It may slows down the speed of unittests.
I will be happy if you have ideas to improve the test for it.

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -808,8 +808,7 @@ Pid spawnProcess(scope const(char[])[] args,
                  File stderr = std.stdio.stderr,
                  const string[string] env = null,
                  Config config = Config.none,
-                 scope const char[] workDir = null,
-                 AdvancedConfig ac = AdvancedConfig.none)
+                 scope const char[] workDir = null)
     @safe
 {
     version (Windows)
@@ -820,7 +819,7 @@ Pid spawnProcess(scope const(char[])[] args,
     }
     else version (Posix)
     {
-        return spawnProcessPosix(args, stdin, stdout, stderr, env, config, workDir, ac);
+        return spawnProcessPosix(args, stdin, stdout, stderr, env, config, workDir);
     }
     else
         static assert(0);
@@ -890,8 +889,7 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
                               File stderr,
                               scope const string[string] env,
                               Config config,
-                              scope const(char)[] workDir,
-                              AdvancedConfig ac = AdvancedConfig.none)
+                              scope const(char)[] workDir)
     @trusted // TODO: Should be @safe
 {
     import core.exception : RangeError;
@@ -916,7 +914,7 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
     argz[$-1] = null;
 
     // Prepare environment.
-    auto envz = createEnv(env, !(config & Config.newEnv));
+    auto envz = createEnv(env, !(config.flags & Config.Flags.newEnv));
 
     // Open the working directory.
     // We use open in the parent and fchdir in the child
@@ -963,13 +961,13 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
     since the first and the second forks will run in parallel.
     */
     int[2] pidPipe;
-    if (config & Config.detached)
+    if (config.flags & Config.Flags.detached)
     {
         if (core.sys.posix.unistd.pipe(pidPipe) != 0)
             throw ProcessException.newFromErrno("Could not create pipe to get process pid");
         setCLOEXEC(pidPipe[1], true);
     }
-    scope(exit) if (config & Config.detached) close(pidPipe[0]);
+    scope(exit) if (config.flags & Config.Flags.detached) close(pidPipe[0]);
 
     static void abortOnError(int forkPipeOut, InternalError errorType, int error) nothrow
     {
@@ -988,7 +986,7 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
     void closePipeWriteEnds()
     {
         close(forkPipe[1]);
-        if (config & Config.detached)
+        if (config.flags & Config.Flags.detached)
             close(pidPipe[1]);
     }
 
@@ -1006,7 +1004,7 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
         // Child process
 
         // no need for the read end of pipe on child side
-        if (config & Config.detached)
+        if (config.flags & Config.Flags.detached)
             close(pidPipe[0]);
         close(forkPipe[0]);
         immutable forkPipeOut = forkPipe[1];
@@ -1041,7 +1039,7 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
             setCLOEXEC(STDOUT_FILENO, false);
             setCLOEXEC(STDERR_FILENO, false);
 
-            if (!(config & Config.inheritFDs))
+            if (!(config.flags & Config.Flags.inheritFDs))
             {
                 // NOTE: malloc() and getrlimit() are not on the POSIX async
                 // signal safe functions list, but practically this should
@@ -1103,7 +1101,7 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
                 if (stderrFD > STDERR_FILENO)  close(stderrFD);
             }
 
-            ac.preExecFunction(&abortOnErrorInPreExec);
+            config.preExecFunction(&abortOnErrorInPreExec);
 
             // Execute program.
             core.sys.posix.unistd.execve(argz[0], argz.ptr, envz);
@@ -1112,7 +1110,7 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
             abortOnError(forkPipeOut, InternalError.exec, .errno);
         }
 
-        if (config & Config.detached)
+        if (config.flags & Config.Flags.detached)
         {
             auto secondFork = core.sys.posix.unistd.fork();
             if (secondFork == 0)
@@ -1153,7 +1151,7 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
         // Save error number just in case if subsequent "waitpid" fails and overrides errno
         immutable lastError = .errno;
 
-        if (config & Config.detached)
+        if (config.flags & Config.Flags.detached)
         {
             // Forked child exits right after creating second fork. So it should be safe to wait here.
             import core.sys.posix.sys.wait : waitpid;
@@ -1183,7 +1181,7 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
                     break;
                 case InternalError.doubleFork:
                     // Can happen only when starting detached process
-                    assert(config & Config.detached);
+                    assert(config.flags & Config.Flags.detached);
                     errorMsg = "Failed to fork twice";
                     break;
                 case InternalError.malloc:
@@ -1199,7 +1197,7 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
                 throw ProcessException.newFromErrno(error, errorMsg);
             throw new ProcessException(errorMsg);
         }
-        else if (config & Config.detached)
+        else if (config.flags & Config.Flags.detached)
         {
             owned = false;
             if (read(pidPipe[0], &id, id.sizeof) != id.sizeof)
@@ -1207,13 +1205,13 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
         }
 
         // Parent process:  Close streams and return.
-        if (!(config & Config.retainStdin ) && stdinFD  > STDERR_FILENO
+        if (!(config.flags & Config.Flags.retainStdin ) && stdinFD  > STDERR_FILENO
                                             && stdinFD  != getFD(std.stdio.stdin ))
             stdin.close();
-        if (!(config & Config.retainStdout) && stdoutFD > STDERR_FILENO
+        if (!(config.flags & Config.Flags.retainStdout) && stdoutFD > STDERR_FILENO
                                             && stdoutFD != getFD(std.stdio.stdout))
             stdout.close();
-        if (!(config & Config.retainStderr) && stderrFD > STDERR_FILENO
+        if (!(config.flags & Config.Flags.retainStderr) && stderrFD > STDERR_FILENO
                                             && stderrFD != getFD(std.stdio.stderr))
             stderr.close();
         return new Pid(id, owned);
@@ -1231,8 +1229,8 @@ version (Posix)
     sigaddset(&ss, SIGINT);
     pthread_sigmask(SIG_BLOCK, &ss, null);
 
-    AdvancedConfig ac = {
-        preExecFunction: (AdvancedConfig.AbortOnError aoe) @trusted @nogc nothrow {
+    Config config = {
+        preExecFunction: (Config.AbortOnError aoe) @trusted @nogc nothrow {
             // Reset signal handlers
             sigset_t ss;
             if (sigfillset(&ss) != 0)
@@ -1251,9 +1249,8 @@ version (Posix)
                             std.stdio.stdout,
                             std.stdio.stderr,
                             null,
-                            Config.none,
-                            null,
-                            ac);
+                            config,
+                            null);
     scope(failure)
     {
         kill(pid, SIGKILL);
@@ -1300,7 +1297,7 @@ private Pid spawnProcessWin(scope const(char)[] commandLine,
     if (commandLine.empty) throw new RangeError("Command line is empty");
 
     // Prepare environment.
-    auto envz = createEnv(env, !(config & Config.newEnv));
+    auto envz = createEnv(env, !(config.flags & Config.Flags.newEnv));
 
     // Startup info for CreateProcessW().
     STARTUPINFO_W startinfo;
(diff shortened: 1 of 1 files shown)


std/process.d
```diff
@@ -1095,6 +1095,17 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
                 if (stderrFD > STDERR_FILENO)  close(stderrFD);
             }
 
+            // Reset signal handlers
+            sigset_t ss;
+            if (sigfillset(&ss) != 0)
+            {
+                abortOnError(forkPipeOut, InternalError.exec, .errno);
```
kubo39, line 1102:
```suggestion
                abortOnError(forkPipeOut, InternalError.sigfillset, .errno);
```

std/process.d
```diff
@@ -1095,6 +1095,17 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
...
+            }
+            if (sigprocmask(SIG_UNBLOCK, &ss, null) != 0)
+            {
+                abortOnError(forkPipeOut, InternalError.exec, .errno);
```
kubo39, line 1106:
```suggestion
                abortOnError(forkPipeOut, InternalError.sigprocmask, .errno);
```

kubo39 (commented):
Please update InternalError.

std/process.d
tom-tan: Done!

RazvanN7:
Is this issue valid? From [1]: "A child created via fork(2) inherits a copy of its parent's signal mask; the signal mask is preserved across execve(2)". This is the exact behavior presented in Issue 21575. On what grounds is this PR changing that?

https://man7.org/linux/man-pages/man7/signal.7.html

tom-tan:
> On what grounds is this PR changing that?

Let us consider the case of implementing an interpreter such as bash and zsh with dlang.
It executes other external commands such as `ls` and `ps`.
When a UNIX signal such as `SIGINT` is sent to the interpreter, it sends the same signal to other executing commands to terminate them but the interpreter itself should not be terminated.

One idea to implement the interpreter is to use the following three types of threads or processes:

1. a main thread that executes other commands using `spawnProcess`
2. a thread that waits UNIX signals using `sigwait` and notifies it to 1 in some ways such as using `std.concurrency`
3. external commands generated by the main thread

The problem I want to solve is that the current `spawnProcess` cannot support this use case.

To guarantee that only the thread 2 waits the signals, the thread 1 should block signals such as `SIGINT`.
However, the process 2 cannot be terminated by `SIGINT` because it inherits the signal masks from the main thread.
The only way to implement it is to give up using `spawnProcess` and to use old-fashioned `fork` and `exec` directly and re-implement the other stuffs that are already done in `spawnProcess`.

It would be nice if `spawnProcess` unblocks the signals as proposed in this request or it provides a way to control the inheritance of signal masks.

RazvanN7:
@tom-tan I understand the usefulness of this, however, the I don't know whether the initial implementation did this on purpose or it was an accident. One might assume that this was done on purpose to mimic the fork/exec paradigm on linux and folks using `spawnProcess` on linux might rely on this behavior.

@PetarKirov @wilzbach @schveiguy Do you have any opinions on this?

CyberShadow:
> Let us consider the case of implementing an interpreter such as bash and zsh with dlang.

This seems like a use case that completely warrants using the low-level platform-specific APIs.

If it's something that really needs to be in `std.process`, I don't see why it couldn't be enabled by a `Config` option. Messing with signal settings by default doesn't seem right.

BTW, won't this also unblock singals which were not blocked by the application, but inherited? That would be a surprising behavior and probably a regression.

tom-tan:
> This seems like a use case that completely warrants using the low-level platform-specific APIs.

I understand that low-level APIs are better for performance for such applications but high-level APIs are still useful for PoC implementations.
IMO, one big advantage of D is that we can easily start with high-level APIs for PoC and then gradually improve its performance with low-level APIs without changing the language :-)

> If it's something that really needs to be in std.process, I don't see why it couldn't be enabled by a Config option. Messing with signal settings by default doesn't seem right.

I agree with that. I will fix this request to add a new config `resetSignalMasks` instead of unblocking signals by default.

> BTW, won't this also unblock singals which were not blocked by the application, but inherited? That would be a surprising behavior and probably a regression.

I have no ideas for reasonable solutions for it. `spawnProcess` cannot know the signal masks that were inherited from the parent of caller of `spawnProcess`.

Do you have any ideas for it?

CyberShadow:
>  IMO, one big advantage of D is that we can easily start with high-level APIs for PoC and then gradually improve its performance with low-level APIs without changing the language :-)

Actually performance has nothing to do with it, I mainly meant cross-platformness. std.process tries to be a cross platform module, providing a unified interface across D-supported platforms, with the biggest bridged divide being between Windows and POSIX.

The pertinent question is whether the task that the application is trying to accomplish makes sense on both Windows and POSIX. Windows doesn't have the concept of signals, and implementing an interactive shell is going to be a very different task on Windows compared to doing this on POSIX. Therefore, it makes little sense to try doing a POSIX-only task using std.process - you are going to be better off using the POSIX APIs directly.

Process management in general is very different on the two platforms, and there are still many extremely-platform-specific things that are not available through std.process. I don't think it makes sense to try to add them all, as doing so would provide little benefit to applications over just letting them use the native API. On the other hand, if a certain feature is available on all supported platforms, then implementing it in std.process is highly valuable, as it enables writing platform-agnostic D code.

So, unless this can be generalized in some way into an option that does the same conceptual thing on both Windows and POSIX, I don't think this makes a valuable addition...

tom-tan:
> Actually performance has nothing to do with it, I mainly meant cross-platformness. std.process tries to be a cross platform module, providing a unified interface across D-supported platforms, with the biggest bridged divide being between Windows and POSIX.

Thank you for the explanation. I get your point.
However, there are already many exceptions in `std.process` such as [`Config.suppressConsole`](https://dlang.org/phobos/std_process.html#.Config.suppressConsole) and [`inheritFDs`](https://dlang.org/phobos/std_process.html#.Config.inheritFDs) (meaningful only in POSIX), [`environment.opIndexAssign`](https://dlang.org/phobos/std_process.html#.environment.opIndexAssign) and [`remove`](https://dlang.org/phobos/std_process.html#.environment.remove) (have limitations in some platforms).

I understand that we have to be careful to add a new config that is meaningful only in the specific platforms, but it is still worth adding it, IMO.

> you are going to be better off using the POSIX APIs directly.

In the case of implementing a shell interpreter, the following features are needed:
- executing specified command
- input/output redirection
- control of environment variables
- control of working directory

`spawnProcess` has a necessary and sufficient API to achive these features.
As a result of that, we have to re-implement `spawnProcess` except for the signal masks.

Controlling child processes is a typical use case. It would be nice if `spawnProcess` supports this use case to allow users not to re-implement the whole things that are already in `spawnProcess`.

CyberShadow:
>  However, there are already many exceptions in `std.process` such as [`Config.suppressConsole`](https://dlang.org/phobos/std_process.html#.Config.suppressConsole) and [`inheritFDs`](https://dlang.org/phobos/std_process.html#.Config.inheritFDs) (meaningful only in POSIX), [`environment.opIndexAssign`](https://dlang.org/phobos/std_process.html#.environment.opIndexAssign) and [`remove`](https://dlang.org/phobos/std_process.html#.environment.remove) (have limitations in some platforms).

Ironically all of these are examples of things that we want less of, and that shouldn't have happened had we had better foresight.

`Config.suppressConsole` is a hack that gives up on trying to do anything about Windows' distinct treatment of console / GUI processes. More useful would have been a flag that attempts to reconcile this from the point of view of cross-platform applications (GUI or console), for example, `Config.requireConsole` with the following semantics:

- On POSIX:
  - If this flag is unset:
    - Run the process as usual.
  - If this flag is set:
    - If `stdin` is a TTY:
      - Run the process as usual.
    - If `stdin` is not a TTY:
      - Throw an exception.
- On Windows:
  - If this flag is unset:
    - Run the process with `CREATE_NO_WINDOW`.
  - If this flag is set:
    - Run the process as usual.

`inheritFDs` attempts to work around a mistake in `std.socket` and other places, specifically that file descriptors are by default created as inheritable. It is almost never useful to create processes which inherit *all* file descriptors, so they should have been created with `SOCK_CLOEXEC` / `O_CLOEXEC` by default. (Arguably D inherits this issue from POSIX, but it still had an opportunity to solve this problem here, and it didn't.)

Environment access through `std.process` should just be fixed to be synchronized on POSIX. It is already internally synchronized on Windows. It's hard to think of a case when environment access is performance sensitive, so we should just fix this.

CyberShadow:
One idea that might be worth discussing is allowing users to specify a delegate that is called from within the forked process, before the `exec`. This would allow them to do arbitrary setup things such as configuring signal masks, closing all but a certain set of file descriptors, or anything else that we can't predict. The delegate would need to conform to the relevant restrictions of what is / isn't allowed to do in that context, so it would be a very low level feature, but it would provide the flexibility for solving this entire class of problems.

tom-tan:
> One idea that might be worth discussing is allowing users to specify a delegate that is called from within the forked process, before the `exec`.

That is a great idea!
I agree with using a delegate rather than introducing new configs or an ad-hook fix that is initially proposed in this request.

RazvanN7:
> That is a great idea!
I agree with using a delegate rather than introducing new configs or an ad-hook fix that is initially proposed in this request.

@tom-tan Are you planning on implementing this in this PR?

tom-tan:
@RazvanN7 If possible, yes, but I find using a delegate will introduce other issues that include breaking changes of behavior in addition to the signature of `spawnProcess`.
I have no ideas for better solutions and I am not sure which breaking changes and platform-dependent behaviors are acceptable or not.

Here is a list of issues I found. To simplify the discussion, I assume a signature of new `spawnProcess` as follows: `spawnProcess(string program, DelegateType dg)` in which `DelegateType` is a type of a delegate that is called before `exec` in the spawned process.

- `spawnProcess` internally uses `fork` and `exec` on POSIX and `CreateProcessW` on Windows. That is, a new parameter `dg` is only meaningful on POSIX.
  - Once we decide `dg` is called before `CreateProcessW` on Windows, it introduces a different semantics between platforms. For example, `dg` does not affect anything in the parent process on POSIX because it is called by a spawned process. On the other hand, it affects the parent process on Windows because it is called before spawning a process.

- It is possible to replace `inheritFDs` (meaningful only on POSIX) with a delegate that closes all unused file descriptors. However, it means that users have to do it in `dg` and will introduce a breaking change of the default behavior only on POSIX because all the file descriptors are inherited by default.
  - We have a workaround to introduce a default parameter of `dg` to close the file descriptors. However, it means that users have to call it in their custom delegates in the most cases. IMO, it is not user friendly API because it is easier to miss to call it.

Do you have ideas for these issues?

CyberShadow:
>     * `spawnProcess` internally uses `fork` and `exec` on POSIX and `CreateProcessW` on Windows. That is, a new parameter `dg` is only meaningful on POSIX.

Yes, this is a POSIX-only feature, and should not be available (same code shouldn't compile) on Windows.

A Windows equivalent could perhaps accept the `STARTUPINFO` struct by ref or otherwise allow customizing Windows-specific CreateProcess parameters, but that would be a different improvement.

One approach to avoid diverging the std.process function signatures too much is to introduce a `struct`, containing advanced platform-specific configuration, which would then have members that are only present on their relative platform. Perhaps the `Config` enum could be converted to a struct for this purpose?

> It is possible to replace inheritFDs

Undortunately all existing functionality must remain, as ugly as it is. If we want to put some of it through a deprecation cycle and remove it, that would be a separate discussion.

RazvanN7:
ping @tom-tan

tom-tan:
Sorry for late reply.

I re-done this request as suggested!

Here is a list of changes:
- `AdvancedConfig` sturct is introduced for user defined hooks.
  - Currently `preExecFunction` for POSIX is the only member for it. Adding an extra field for Windows is out of scope of this request.
  - `AdvancedConfig.preExecFunction` has an argument `AbortOnError` that enables notifications from `preExecFunction` to its parent process and aborts forked process.
  - I guess `Config.inheritFDs` can be replaced with an appropriate `preExecFunction` but it is out of scope of this request.

Several notes to be discussed:
- I need better names for `AdvancedConfig`, `AbortOnError` and `preExecFunction`.
- I made this request not to change the signatures of existing internal functions (such that `abortOnError` and `forkChild`) but the signature of `preExecFunction` seems to be too restrictive.

std/process.d
```diff
@@ -2046,6 +2113,31 @@ enum Config
     stderrPassThrough = 128,
 }
 
+/**
+A sturct that controls the behavior of process creation functions in this module.
```
RazvanN7, line 2117:
```suggestion
A struct that controls the behavior of process creation functions in this module.
```

std/process.d
```diff
@@ -2046,6 +2113,31 @@ enum Config
...
+    A function type that notifies error in forked process.
+    */
+    alias AbortOnError = void delegate() nothrow @nogc @safe;
+
+    /**
+    A function that is called before `exec` in $(LREF spawnProcess).
+    $(LREF AbortOnError) can be called to notify errors in `preExecFunction` and
+    to abort forked process.  On Windows, this member has no effect.
+    */
+    void function(AbortOnError) nothrow @nogc @safe preExecFunction;
```
RazvanN7, line 2131: You can use `version($OS)` to include or exclude this function, depending on the platform. In this case, you can have something like:

```d
version(Windows) {}
else { /* define function here */}
```

std/process.d
```diff
@@ -2046,6 +2113,31 @@ enum Config
...
 
+/**
+A sturct that controls the behavior of process creation functions in this module.
+*/
+struct AdvancedConfig
+{
+    /**
+    A function type that notifies error in forked process.
+    */
+    alias AbortOnError = void delegate() nothrow @nogc @safe;
```
RazvanN7, line 2124: We are forcing the user to provide a pre-exec function that doesn't throw, does not allocate and is safe. This might be too restrictive.

RazvanN7: The alternative would be to simply declare the delegate without the attributes. That way the user can provide both @safe and non-@safe delegates.

RazvanN7:
cc @CyberShadow

std/process.d
tom-tan: I agree with it is too restrictive.

I will remove attributes except safe.
Accepting unsafe delegates breaks memory safety of `spawnProcess` because it is currently marked as trusted and thus the unsafe delegates are accidentally treated as safe.

CyberShadow: > We are forcing the user to provide a pre-exec function that doesn't throw, does not allocate and is safe. This might be too restrictive.

I just realized that I was talking about the restrictions that would be implicit with using `vfork` or something similar, which we do not actually use right now.

I see that an attempt has already been made to use `vfork`, which did not go through: https://github.com/dlang/phobos/pull/3476

However, I'm not sure if we should completely resign ourselves to using `fork` for `std.process` either.

std/process.d
```diff
@@ -1095,6 +1103,8 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
                 if (stderrFD > STDERR_FILENO)  close(stderrFD);
             }
 
+            ac.preExecFunction(&abortOnErrorInPreExec);
```
CyberShadow: Don't you need a `null` check here?

CyberShadow:
> One approach to avoid diverging the std.process function signatures too much is to introduce a new struct, containing advanced platform-specific configuration, which would then have members that are only present on their respective platform. Perhaps the `Config` enum could be converted to a struct for this purpose?

I pushed a commit which does this, what do you think?

std/process.d
CyberShadow: And, you can use `version (StdDdoc)` to write definitions of symbols that will be used only for the documentation, so that the rendered documentation is the same no matter what OS it is rendered on.

std/process.d
```diff
@@ -1095,14 +1101,16 @@ private Pid spawnProcessPosix(scope const(char[])[] args,
                 if (stderrFD > STDERR_FILENO)  close(stderrFD);
             }
 
+            config.preExecFunction(&abortOnErrorInPreExec);
```
CyberShadow, line 1104: Why not simply make `preExecFunction` return a `bool`?

std/process.d
tom-tan: Thank you for the suggestion.
I move to the definitions to `version (Posix)` and `version (StdDdoc)` blocks.

tom-tan:
LGTM.

std/process.d
tom-tan: Yes. I added a check for it.

std/process.d
tom-tan: It is better if `preExecFunction` is nothrow.
I fixed a signature and removed unused `abortOnErrorInPreExec`.

RazvanN7:
@tom-tan This seems to be failing the testsuite.

RazvanN7:
The failure is caused by the fact that `enum Config` is public and therefore is used in the wild. Changing how it is used breaks code out there. I'm wondering if that change is really necessary. Letting `Config` unchanged will fix the current failures.

CyberShadow:
> The failure is caused by the fact that `enum Config` is public and therefore is used in the wild. Changing how it is used breaks code out there. I'm wondering if that change is really necessary. Letting `Config` unchanged will fix the current failures.

I looked into why that error was occurring, because it made no sense to me. It turned out to be a compiler bug:
https://issues.dlang.org/show_bug.cgi?id=21828

We could go back to adding yet another argument to `std.process` functions, but that seems suboptimal to me. Perhaps the compiler bug should just be fixed instead.

RazvanN7:
PR: https://github.com/dlang/dmd/pull/12437

RazvanN7:
The bug was fixed, however, this still fails.

tom-tan:
It still fails but it seems to be unrelated to this change.

RazvanN7:
@tom-tan I think that a rebase will fix that issue.

tom-tan:
@RazvanN7 Thank you for your advice. I rebased the commits.

RazvanN7 (approved):

RazvanN7:
Thanks @tom-tan for your patience and willingness to apply all the feedback and congratulations on this merge!

tom-tan:
Thanks!

PR #7767, opened 2021-01-24
Minimize use of GC for doubles in printFloat.

berni44:
A fixup for #7393. Now it's guaranteed, that printing float and double with %a or %e will not use gc as long as the output is restricted to 500 characters. (Without this change, this guarantee did only hold for floats but not for doubles.)

### DIFF
--- a/std/format.d
+++ b/std/format.d
@@ -7754,7 +7754,7 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
     //   digits. If still necessary, we decide the rounding type, mainly by looking at the
     //   next digit.
 
-    ulong[4] bigbuf;
+    ulong[18] bigbuf;
     if (exp >= T.mant_dig)
     {
         // large number without fractional digits
@@ -7764,7 +7764,7 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
         int count = exp / 60 + 1;
 
         // saved in big endian format
-        ulong[] mybig = count <= bigbuf.length ? bigbuf[0 .. count] : new ulong[count];
+        ulong[] mybig = bigbuf[0 .. count];
 
         // only the first or the first two ulongs contain the mantiassa. The rest are zeros.
         int lower = 60 - (exp - T.mant_dig + 1) % 60;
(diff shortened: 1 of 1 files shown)


std/format.d
```diff
@@ -8448,7 +8448,7 @@ printFloat_done:
     import core.memory;
     auto stats = GC.stats;
 
-    char[256] buf;
+    char[512] buf;
```
berni44, line 8451: I changed this, because the call in `formatValueImpl` also provides a buffer with 512 chars.

std/format.d
```diff
@@ -7754,7 +7754,7 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
     //   digits. If still necessary, we decide the rounding type, mainly by looking at the
     //   next digit.
 
-    ulong[4] bigbuf;
+    ulong[18] bigbuf;
```
RazvanN7, line 7757: N00b question: why 18?

std/format.d
```diff
@@ -7764,7 +7764,7 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
         int count = exp / 60 + 1;
```
RazvanN7, line 7764: I assume that the value of count cannot ever be greater than 18?

RazvanN7 (commented):
Do you plan on making printFloatE completely `@nogc` ?

std/format.d
berni44: Indeed. For float it cannot be greater than 3 and for double it cannot be greater than 18. (exp is maximum 127/1023).

RazvanN7 (approved):
It would be nice if in the future we could get rid of allocations for the entire function and maybe introduce a `@nogc` unittest.

berni44:
> Do you plan on making printFloatE completely `@nogc` ?

This will be hard to achieve this goal and might reduce the speed considerably. The reason is, that the output could be arbitrarily large (in theory). To make this `@nogc` it would be necessary to process the output from left to right, which makes for a much more complicated algorithm. One of the hard things is to calculating the exact length of the useful characters in advance for all combinations of flags, precision, width and rounding mode.

Processing from left to right involves a BigInt multiplication with a large and a very large number. When not using `std.BigInt` but a specialized algorithm to do that, this might improve the speed for `%e` (bit will slow down `%f`). I plan to implement this (for %e only), but this would only be the first step toward `@nogc` on a long road.

If you wonder where's the difference between `%e` and `%f` here, compare `format("%e",double.max)` (`1.797693e+308`) vs. `format("%f",double.max)` (`179769313486231570814527423731704356798070567525844996598917476803157260780028538760589558632766878171540458953514382464234321326889464182768467546703537516986049910576551282076245490090389328944075868508455133942304583236903222948165808559332123348274797826204144723168738177180919299881250404026184124858368.000000`). The current algorithm calculates the whole 330 digits for `%e` and `%f`. It's difficult to imagine any improvement for `%f`, while it's obvious, that for `%e` it would suffice to calculate only the first view digits; but due to rounding issues, you cannot predict in advance the number of needed digits...

PR #7770, opened 2021-01-26
We should not be promoting using GC.free casually.

schveiguy:
See note on the forums: https://forum.dlang.org/post/idpkrwpdhvckvvgdgnom@forum.dlang.org


quickfur (approved):
LGTM

PetarKirov (approved):

PR #7774, opened 2021-01-28
Remove documentation from std.xml due to deprecation in midterm.

berni44:
According to [DIP1013](https://github.com/dlang/DIPs/blob/d8f2e769c3a8c711e7886ccecc93eac9795dae9c/DIPs/accepted/DIP1013.md) all public symbols should loose their public comments after 5 non-patch releases. I hope I didn't remove too much...


pbackus:
Rather than deleting the documentation entirely, it might be better to simply change the DDoc comments into non-DDoc comments. E.g., change the starting delimiter from `/**` to `/*`.

berni44:
Good idea. Thanks. I'll do that.

RazvanN7 (approved):

PR #7775, opened 2021-01-28
Deprecate approxEqual.

berni44:
Meanwhile all occurences of approxEqual in phobos, druntime and dmd are removed. 
In [changelog for release 2.091](https://dlang.org/changelog/2.091.0.html#isClose) it has already been announced that approxEqual will eventually be deprecated.

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -8659,6 +8659,10 @@ private real polyImpl(real x, in real[] A) @trusted pure nothrow @nogc
    Computes whether a values is approximately equal to a reference value,
    admitting a maximum relative difference, and a maximum absolute difference.
 
+   Warning:
+        This template is considered out-dated. It will be removed from
+        Phobos in 2.106.0. Please use $(LREF isClose) instead.
+
    Params:
         value = Value to compare.
         reference = Reference value.
@@ -8683,6 +8687,7 @@ private real polyImpl(real x, in real[] A) @trusted pure nothrow @nogc
     See_Also:
         Use $(LREF feqrel) to get the number of equal bits in the mantissa.
  */
+deprecated("approxEqual will be removed in 2.106.0. Please use isClose instead.")
 bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDiff = 1e-5)
 {
     import std.range.primitives : empty, front, isInputRange, popFront;
@@ -8751,8 +8756,7 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
     }
 }
 
-///
-@safe pure nothrow unittest
+deprecated @safe pure nothrow unittest
 {
     assert(approxEqual(1.0, 1.0099));
     assert(!approxEqual(1.0, 1.011));
@@ -8767,8 +8771,7 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
     assert(approxEqual(arr1, arr2));
 }
 
-///
-@safe pure nothrow unittest
+deprecated @safe pure nothrow unittest
 {
     // relative comparison depends on reference, make sure proper
     // side is used when comparing range to single value. Based on
@@ -8781,8 +8784,7 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
     assert(!b.approxEqual(a));
 }
 
-///
-@safe pure nothrow @nogc unittest
+deprecated @safe pure nothrow @nogc unittest
 {
     assert(!approxEqual(0.0,1e-15,1e-9,0.0));
     assert(approxEqual(0.0,1e-15,1e-9,1e-9));
@@ -8792,8 +8794,7 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
     assert(!approxEqual(1.0000000011,1.0,1e-9,0.0));
 }
 
-///
-@safe pure nothrow @nogc unittest
+deprecated @safe pure nothrow @nogc unittest
 {
     // maybe unintuitive behavior
     assert(approxEqual(1000.0,1010.0));
@@ -8803,7 +8804,7 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
     assert(!approxEqual(-1e-30,1e-30,1e-2,0.0));
 }
 
-@safe pure nothrow @nogc unittest
+deprecated @safe pure nothrow @nogc unittest
 {
     int a = 10;
     assert(approxEqual(10, a));
@@ -8818,7 +8819,7 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
     assert(approxEqual(0.0f,0.0));
 }
 
-@safe pure nothrow @nogc unittest
+deprecated @safe pure nothrow @nogc unittest
 {
     real num = real.infinity;
     assert(num == real.infinity);
@@ -8832,7 +8833,7 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
     assert(!approxEqual(real.nan,real.nan));
 }
 
-@safe pure nothrow unittest
+deprecated @safe pure nothrow unittest
 {
     assert(!approxEqual([1.0,2.0,3.0],[1.0,2.0]));
     assert(!approxEqual([1.0,2.0],[1.0,2.0,3.0]));
--- a/changelog/deprecate-approxEqual.dd
+++ b/changelog/deprecate-approxEqual.dd
@@ -1,0 +1,11 @@
+Deprecate `std.math : approxEqual`
+
+The template `approxEqual` for comparing floating point numbers has
+been deprecated.
+
+Please use the template `isClose` instead, which has better default
+values and is symmetric in its arguments.
+
+To (almost) keep the current behaviour of `approxEqual(a, b)` use
+`isClose(a, b, 1e-2, 1e-2)`, but we recommend to adjust the code to
+make it work with `isClose(a, b)`.


maxhaton (changes requested):
Semicolon after the deprecation attribute

RazvanN7:
It seems like there are still uses of `approxEqual` in public documentation tests.

berni44:
> Semicolon after the deprecation attribute

Thanks for detecting this. I should have tested, that the deprecation message really fires. I had to remove all the unittests now too (but I guess this is intended, isn't it), because they produce errors with that message. Or is there a better way?

std/math.d
```diff
@@ -8751,97 +8756,6 @@ bool approxEqual(T, U, V)(T value, U reference, V maxRelDiff = 1e-2, V maxAbsDif
     }
 }
 
-///
-@safe pure nothrow unittest
```
Geod24, line 8755: Mark the unittests as deprecated instead of removing them

berni44: Ah, better! I'll do that soon, but have to leave now.

berni44: done. But the style checker complains now... Any ideas?

berni44: I could make the public unittests not public. What is the usual workflow here?

Geod24: I think you can just remove the `///` here.

std/math.d
```diff
@@ -8659,6 +8659,10 @@ private real polyImpl(real x, in real[] A) @trusted pure nothrow @nogc
    Computes whether a values is approximately equal to a reference value,
    admitting a maximum relative difference, and a maximum absolute difference.
 
+   Warning:
+        This template is considered out-dated. It will be removed from
```
dkorpel, line 8663: Do we really want to remove `approxEqual`?
It seems like a commonly used and low-maintenance function, I think we can keep it around undocumented for backwards compatibility.

berni44: It's the last step of a long road. It started with [issue15881](https://issues.dlang.org/show_bug.cgi?id=15881), where someone complained that `approxEqual` was not doing, what he expected to do. I had the same experience some months before I read that issue and did some research on this issue and came up with one of my first PRs. 

The reason for replacing `approxEqual` is mainly, that it's like a security belt that doesn't hold, when it is needed. Think of this: You wrote a function, added a unittest, and because with `approxEqual` and felt safe. But later, someone else changed something and your function didn't work anymore as expected. Still, the unittest didn't notice. And why? Because the test with `approxEqual` was too weak.

The main idea of this function is to overcome the problems which arise, when comparing floatingpoint numbers with ==. This doesn't work out well due to rounding errors. So `approxEqual` should accept two numbers as equal, when they only differ by some rounding errors. But acctually `approxEqual` accepts also numbers as equal, that are far off compared to rounding errors.

Having said all this: IMHO it's time to remove `approxEqual` completely, because else people feel safe, where they aren't, probably without knowing.

dkorpel: > Think of this: You wrote a function, added a unittest, and because with approxEqual and felt safe.

That's a fair point, but I'm also thinking of this:
You're learning D, you find a code snippet using approxEqual, (e.g. [this](http://rosettacode.org/wiki/Temperature_conversion#D)), it doesn't compile because of `approxEqual`, and you find no trace of what happened to it in the documentation, and you get confused.

Maybe we can replace the function body of `approxEqual` with a `static assert(0, "use isClose instead")` once the time comes to remove it.

RazvanN7:
cc @atilaneves

RazvanN7 (approved):

std/math.d
RazvanN7: Ideally, until the deprecation period we will update any (external) docs that use `approxEqual`.

jondegenhardt:
This depreciation should be better documented. The behavior difference between `isClose` and `approxEqual` is not simply:

> better default values and is symmetric in its arguments

as described in the release notes. The change I am noticing is that behavior near zero has changed. By default, `0.0` does not satisfy `isClose` for any number, where as for `approxEqual` it did. Whether this is more correct or not is not material. When functions are replaced, users need to know the more important ramifications without having to research it.

Also problematic is that the documentation for `approxEqual` is no longer available, so it is not easy to compare the behavior of the two functions without looking at the code. For this reason, it would make sense to include a note about migrating from `approxEqual` to `isClose` in the `isClose` documentation. It's not easy to find release notes unless one already knows where to look.

berni44:
> The change I am noticing is that behavior near zero has changed. By default, `0.0` does not satisfy `isClose` for any number, where as for `approxEqual` it did.

You are right. The reason for this is, that it strongly depends on the use case, which value is best here, so the users should always specifiy this value, if they want to compare to 0.0. (I took that from pythons implementation of `isClose`.) I'll file a PR to add this information.

> Also problematic is that the documentation for `approxEqual` is no longer available, so it is not easy to compare the behavior of the two functions without looking at the code.

It is: https://dlang.org/phobos-prerelease/std_math_operations.html#.approxEqual

n8sh:
!!!

As mentioned by @dkorpel this is an *extremely* commonly used function in existing D programs. A message cautioning people not to use `approxEqual` in new code is good. Deprecation annotations are good. The Phobos policy that as a deprecated function `approxEqual` will be outright removed after X months is *catastrophic*. See: [Dear Google Cloud: Your Deprecation Policy is Killing You](https://steve-yegge.medium.com/dear-google-cloud-your-deprecation-policy-is-killing-you-ee7525dc05dc).

berni44:
> The Phobos policy that as a deprecated function `approxEqual` will be outright removed after X months is _catastrophic_. See: [Dear Google Cloud: Your Deprecation Policy is Killing You](https://steve-yegge.medium.com/dear-google-cloud-your-deprecation-policy-is-killing-you-ee7525dc05dc).

Nice piece to read, indeed. I think, here is not the right place to discuss this - this applies to much more than just `approxEqual`. And I'm not sure if I'm willing to do this discussion, but I will accept, whatever it will lead to.

ibuclaw:
> !!!
> 
> As mentioned by @dkorpel this is an *extremely* commonly used function in existing D programs. A message cautioning people not to use `approxEqual` in new code is good. Deprecation annotations are good. The Phobos policy that as a deprecated function `approxEqual` will be outright removed after X months is *catastrophic*.

https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1013.md

jondegenhardt:
Another non-backward compatible case: float to double conversion:

text.d
```
import std.stdio;
import std.math: approxEqual, isClose;

void main()
{
    double q = 31.79;
    float rf = 31.79f;
    double r = rf;

    writefln("q: %g; r: %g", q, r);
    writeln("approxEqual(q, r): ", approxEqual(q, r));
    writeln("isClose(q, r)    : ", isClose(q, r));
}
```

Run:
```
$ dmd test.d
$ ./test
q: 31.79; r: 31.79
approxEqual(q, r): true
isClose(q, r)    : false
```

This is obviously the type of case that's a nuisance because it requires code maintainers to debug underlying causes. One could argue that the comments in the release notes describing how to change the default arguments to be closer to `approxEqual` should suffice. But that leaves existing code authors with quandary. The introduction of `isClose` is advertised largely as a bug fix for poor behavior in `approxEqual`. Does making the change suggested in the release notes reintroduce the prior poor behavior? There is not enough info in the release notes, one has to do research to decide what to do.

As an aside, I'm reluctant to be overly critical of work that addresses a long standing issue ([Issue 15881](https://issues.dlang.org/show_bug.cgi?id=15881) was reported 5 years ago.). And I'll admit that I wouldn't have bothered looking into this if depreciation notices didn't start appearing during compiles.

But still, I had to spend time debugging the above case. And, I still need to research how to approach it. If I'm having to do this, it's likely other developers will need to do this also.

Personally, I think float to double conversion should fall into the default `isClose` definition of being close. This maybe a case where D and Python should have slightly different behavior. But regardless, this is another case that is useful to include in the `isClose` documentation.

berni44:
> Another non-backward compatible case: float to double conversion:

There are two things, that catch my eye in your example:

1. `writefln("q: %g; r: %g", q, r);` uses default precision of `%g` which in my opinion is already a bad choice for floats, but even worse for doubles. See #8015 were we discussed this briefly. Replace it with `writefln("q: %.16g; r: %.16g", q, r);` and you'll get immediately the source of the problem: `q: 31.79; r: 31.79000091552734`. (If you count them, there are 257_698_037 double values between these two, while when using floats, they are identical.) 
2. As you wrote below: The release note says to get almost identical behavior, you should replace `approxEqual(q, r)` with `isClose(q, r, 1e-2, 1e-2)`. In that case you'll get `true` in both cases. `isClose` with defaults just shows you, that there is something odd that you should investigate.

> This is the type of case that's a nuisance because it requires code maintainers to debug underlying causes. One could argue that the comments in the release notes describing how to change the default arguments to be closer to `approxEqual` should suffice. But that leaves existing code authors with a quandary. The introduction of `isClose` is advertised largely as a bug fix for poor behavior in `approxEqual`. Does making the change suggested in the release notes reintroduce the prior poor behavior?

Yes it does; in my opinion, this is obvious. (But it still removes the asymmetry, which is a minor thing.) In this case, I'd say, your fallacy is to think in terms of precision of floats while you use doubles (the same trap, that the `%g` defaults trapped into).

> There is not enough info in the release notes to tell, one has to do research to decide what to do.

What additionally should have been written there? I think, we cannot list any fallacy, people do with floating point numbers. That's beyond such a release note.

jondegenhardt:
> > Another non-backward compatible case: float to double conversion:
> 
> There are two things, that catch my eye in your example:
> 
> 1. `writefln("q: %g; r: %g", q, r);` uses default precision of `%g` which in my opinion is already a bad choice for floats, but even worse for doubles. See #8015 were we discussed this briefly. Replace it with `writefln("q: %.16g; r: %.16g", q, r);` and you'll get immediately the source of the problem: `q: 31.79; r: 31.79000091552734`. (If you count them, there are 257_698_037 double values between these two, while when using floats, they are identical.)

The `%g` printed format is not material. It's just something I tossed in. I didn't expect it to cause confusion, but in retrospect it I should have removed it before publishing the example. Unfortunately, it appears to have caused a misinterpretation of my comments. Apologies for that.

> > This is the type of case that's a nuisance because it requires code maintainers to debug underlying causes. One could argue that the comments in the release notes describing how to change the default arguments to be closer to `approxEqual` should suffice. But that leaves existing code authors with a quandary. The introduction of `isClose` is advertised largely as a bug fix for poor behavior in `approxEqual`. Does making the change suggested in the release notes reintroduce the prior poor behavior?
> 
> Yes it does; in my opinion, this is obvious. (But it still removes the asymmetry, which is a minor thing.) In this case, I'd say, your fallacy is to think in terms of precision of floats while you use doubles (the same trap, that the `%g` defaults trapped into).

It's obvious to me and certainly quite a few other people, but why is this obvious to the casual reader? I don't think it will be. (Note: I'm not asking for explanations for myself. I'm trying to describe the situation as I expect it to be perceived by many casual readers.)

> 
> > There is not enough info in the release notes to tell, one has to do research to decide what to do.
> 
> What additionally should have been written there? I think, we cannot list any fallacy, people do with floating point numbers. That's beyond such a release note.

As to what else to write - Describing the ramifications of using the described parameters to achieve the prior behavior (`isClose(a, b, 1e-2, 1e-5)`). In particular, that this will re-introduce the undesirable behavior identified in [PR 15881](https://issues.dlang.org/show_bug.cgi?id=15881). Also the expected behavior changes apart from the type of case identified in PR 15881. (More on this below.)

The bigger issue is with using the `isClose` introduction as a rationale for depreciating `approxEqual`. Unfortunately, the default behavior of `isClose` is not backward compatible with `approxEqual` in a number of basic cases where there are legitimate alternate behaviors available.

An important aspect is that there is no such thing as the one correct way to compare floating point numbers. That
why there are some many flavors of these functions around. There is nothing wrong with the default `isClose` behavior, it is a perfectly reasonable method. However, it treats certain cases that would in many contexts be considered reasonable equality cases differently than `approxEqual`. That makes it non-backward compatible and therefore not a justification for depreciating `approxEqual`. Two such behaviors I've identified are:

* Comparison between zero and numbers close to zero
* Double->float->double round trip.

Both of these are legitimately nearly-equal in  many circumstances. Not all of course. There are plenty of times when the current `isClose` default behavior is preferred.

However, backward compatibility for cases like this matters when considering depreciations. Backward compatibility via extra arguments is not desirable, especially if it re-introduces the [PR 15881](https://issues.dlang.org/show_bug.cgi?id=15881). To put this more strongly, backward compatibility of the defaults matters more than having the best default.

As to what to do about it - Two possibilities:

* Change `isClose` so that its default behavior is the same as `approxEqual` in cases where there are legitimate alternate choices about the desired behavior.
* Leave `isClose` as is but don't depreciate `approxEqual`.

The overriding thing here is to avoid having users need to engage in unnecessary coding due to a depreciation unless it is really necessary or has extremely high value. That's the sort of thing people don't appreciate.

PR #7778, opened 2021-02-01
Remove spurious __EOF__ hiding existing declarations/tests

MoonlightSentinel:
... and fix broken (public) unittests.

### DIFF
--- a/std/experimental/allocator/package.d
+++ b/std/experimental/allocator/package.d
@@ -3323,8 +3323,6 @@ nothrow:
     theAllocator.dispose(arr);
 }
 
-__EOF__
-
 /**
 
 Stores an allocator object in thread-local storage (i.e. non-`shared` D
@@ -3364,6 +3362,7 @@ struct ThreadLocal(A)
 }
 
 ///
+@system
 unittest
 {
     import std.experimental.allocator.building_blocks.free_list : FreeList;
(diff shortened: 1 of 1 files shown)


maxhaton:
```

std/experimental/allocator/package.d(3365:1)[warn]: A unittest should be annotated with at least @safe or @system
--
  | std/experimental/allocator/package.d(3373:54)[warn]: Prefer the new "'alias' identifier '=' type ';'" syntax to the  old "'alias' type identifier ';'" syntax.
  | std/experimental/allocator/package.d(3531:16)[warn]: Local imports should specify the symbols being imported to avoid hiding local symbols.
  | std/experimental/allocator/package.d(3538:16)[warn]: Local imports should specify the symbols being imported to avoid hiding local symbols.
  | std/experimental/allocator/package.d(3569:1)[warn]: A unittest should be annotated with at least @safe or @system
  | std/experimental/allocator/package.d(3714:1)[warn]: A unittest should be annotated with at least @safe or @system
  | std/experimental/allocator/package.d(3752:1)[warn]: A unittest should be annotated with at least @safe or @system
  | std/experimental/allocator/package.d(3821:1)[warn]: A unittest should be annotated with at least @safe or @system
  | std/experimental/allocator/package.d(3846:1)[warn]: A unittest should be annotated with at least @safe or @system


```
Buildkite-style moaning

std/experimental/allocator/package.d
```diff
@@ -3323,8 +3323,6 @@ nothrow:
     theAllocator.dispose(arr);
 }
 
-__EOF__
```
RazvanN7, line 3326: LOL. What was the point of it? Was this an accident?

MoonlightSentinel: Probably, it was introduced at the end of the file just before a `version(none)` declaration.

RazvanN7 (approved):

std/experimental/allocator/package.d
MoonlightSentinel: https://github.com/dlang/phobos/blob/8b249a6240fe9a923da69c7122fc27059640ff59/std/allocator.d#L3987

schveiguy: I just was about to create a PR to remove this.

I found it in my attempts to refactor phobos for phobos v2, my refactoring didn't do well with this. @MoonlightSentinel was on the beerconf call when I found it, thanks for fixing it!

> Was this an accident?

I actually emailed Andrei, and he said he remembered some compiler bug making him put something like that in.

PR #7779, opened 2021-02-01
Implement missing OpenBSD ports in phobos

ibuclaw:
`thisExePath` is yet to be ported.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -3590,6 +3590,44 @@ else version (DragonFlyBSD)
 
         return buffer.assumeUnique;
     }
+    else version (OpenBSD)
+    {
+        import core.sys.openbsd.sys.sysctl : sysctl, CTL_KERN, KERN_PROC_ARGS, KERN_PROC_ARGV;
+        import core.sys.posix.unistd : getpid;
+        import std.conv : to;
+        import std.exception : enforce, errnoEnforce;
+        import std.process : searchPathFor;
+
+        int[4] mib = [CTL_KERN, KERN_PROC_ARGS, getpid(), KERN_PROC_ARGV];
+        size_t len;
+
+        auto result = sysctl(mib.ptr, mib.length, null, &len, null, 0);
+        errnoEnforce(result == 0);
+
+        auto argv = new char*[len - 1];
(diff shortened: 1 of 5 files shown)


RazvanN7 (approved):

ibuclaw:
Added `thisExePath`, which wasn't a straightforward implementation. :-)

RazvanN7:
Sorry, I thought you intended to do a separate PR for it, that's why I slapped `auto-merge` before.

std/file.d
```diff
@@ -3590,6 +3593,63 @@ else version (DragonFlyBSD)
...
+    {
+        import core.sys.posix.unistd : getpid;
+        import std.conv : to;
+        import std.exception : enforce, errnoEnforce;
+        enum
+        {
+            CTL_KERN = 1,
+            KERN_PROC_ARGS = 55,
+            KERN_PROC_ARGV = 1,
+        }
```
RazvanN7, line 3606: why redefine when they could be reused from `core.sys.openbsd.sys.sysctl`?

ibuclaw: This function is not very well separated since someone added `thisExePathHasSysCtlProcName`, I'd rather tackle all platforms at once in a follow-up PR.

maxhaton (approved):

std/file.d
```diff
@@ -3590,6 +3590,44 @@ else version (DragonFlyBSD)
...
+
+        auto result = sysctl(mib.ptr, mib.length, null, &len, null, 0);
+        errnoEnforce(result == 0);
+
+        auto argv = new char*[len - 1];
+        result = sysctl(mib.ptr, mib.length, argv.ptr, &len, null, 0);
+        errnoEnforce(result == 0);
+
+        auto argv0 = argv[0];
+        if (*argv0 == '/' || *argv0 == '.')
```
PetarKirov, line 3612: If the path starts with `/`, it is already absolute, no?

ibuclaw: Are symlinks expected to be resolved?

RazvanN7 (approved):

PR #7785, opened 2021-02-03
Prepare to split up std.format into submodules.

berni44:
The idea is, to make very small steps here. One of the main problem of splitting modules is, that ideally no other PR should be merged during the process, because it's easy to overlook some of these changes while rebasing a splitting PR. So I just moved here everything to std/format/package.d (with only a single slight change, that was needed). This can be repeated, whenever another PR is merged, without fear of any loss.

Ideally PRs #7617 and #5797 are merged before this one, so no rebasing is needed there anymore (both are already labeled with the merge label, but the tests are for unknown reasons stalled since days).

See also PR #7765 for the main idea behind this.


berni44:
Meanwhile I think, we should wait with splitting until #7757 and #7617 are merged. I'm closing this temporarily.

PR #7786, opened 2021-02-04
Fix issue 21609 - LinearCongruentialEngine fails for m = 0

Herringway:


std/random.d
```diff
@@ -603,7 +603,14 @@ Always `false` (random generators are infinite ranges).
  */
     enum bool empty = false;
 
-    private UIntType _x = m ? (a + c) % m : (a + c);
+    static if (m)
+    {
+        private UIntType _x = a + c % m;
```
berni44, line 608: There is a change from `(a + c) % m` to `a + c % m`. Is this intentional?

Herringway: it was not. good catch

berni44 (approved):
Looks like this is a hack around a bug in dmd. I filed a bug report [Issue 21610](https://issues.dlang.org/show_bug.cgi?id=21610). Maybe we should add a comment next to the `static if`, that this is a hack around this bug and could be reverted, when that bug is fixed.

If you like, you could also add a unittest, that guards against the typo in your first version, e.g. this:

```
@safe unittest
{
    alias LGE = LinearCongruentialEngine!(uint, 10000, 19682, 19683);

    LGE rnd;
    assert(rnd.front == 9999);
}
```

Herringway:
Sounds good.

thewilsonator (approved):

PR #7787, opened 2021-02-11
Fix Issue 21628 - The padding bits of bitfields could be calculated automatically

berni44:
I always wondered, why the mixin could not create the padding bits automatically. The solution was much simpler, than I thought... :-)

### DIFF
--- a/std/bitmanip.d
+++ b/std/bitmanip.d
@@ -161,17 +161,18 @@ private template createStorageAndFields(Ts...)
 {
     enum Name = createStoreName!Ts;
     enum Size = sizeOfBitField!Ts;
-    static if (Size == ubyte.sizeof * 8)
+    static if (Size <= ubyte.sizeof * 8)
         alias StoreType = ubyte;
-    else static if (Size == ushort.sizeof * 8)
+    else static if (Size <= ushort.sizeof * 8)
         alias StoreType = ushort;
-    else static if (Size == uint.sizeof * 8)
+    else static if (Size <= uint.sizeof * 8)
         alias StoreType = uint;
-    else static if (Size == ulong.sizeof * 8)
+    else static if (Size <= ulong.sizeof * 8)
         alias StoreType = ulong;
     else
     {
-        static assert(false, "Field widths must sum to 8, 16, 32, or 64");
+        import std.conv : to;
+        static assert(false, "Sum of field widths must be <= 64, is " ~ to!string(Size));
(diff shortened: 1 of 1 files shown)


RazvanN7:
Maybe the limitation was put in place to make sure that the user realizes that the bitfield size is a multiple of byte? cc @andralex

std/bitmanip.d
```diff
@@ -291,18 +292,17 @@ of the bitfields storage.
 }
 
 /**
-The sum of all bit lengths in one $(D_PARAM bitfield) instantiation
-must be exactly 8, 16, 32, or 64. If padding is needed, just allocate
-one bitfield with an empty name.
+In former versions of Phobos the sum of all bit lengths in one
```
RazvanN7, line 295: There is no point in highlighting what older versions did in the documentation. 
```suggestion
The sum of all bit lengths in one $(D_PARAM_bitfield) instantiation must be smaller than 64.
```

berni44:
> Maybe the limitation was put in place to make sure that the user realizes that the bitfield size is a power of 2?

Maybe this should be explained in the docs?

andralex:
This was by design. The layout is important and it is good to be explicit. This is in contrast with C's cavalier approach that is loosely defined and left to platforms, which makes people avoid it in portable code.

andralex:
BTW, one good thing to do with bitfields would be refactoring the implementation, which is quite convoluted and uses ancient technology. E.g. even the definition (template with an alias inside) could be replaced with a templated alias.

berni44:
I added a detailed documentation. @andralex Do you think this makes things explicit enough?

berni44:
> BTW, one good thing to do with bitfields would be refactoring the implementation, which is quite convoluted and uses ancient technology. E.g. even the definition (template with an alias inside) could be replaced with a templated alias.

I'd like to leave this for an other PR. I also thought of adding more error checking (number of args is dividable by 3 etc.) and some more unittests and public examples.

berni44:
Changed the second unittest, to show how a larger storage type can be enforced.

andralex:
>I added a detailed documentation. @andralex Do you think this makes things explicit enough?

Of course more docs are better than fewer, but I disagree with the feature in the first place. But reasonable people may disagree.

> Changed the second unittest, to show how a larger storage type can be enforced.

This is evidence that the proposed feature adds more trouble than it removes.

berni44:
OK. Got it.

PR #7788, opened 2021-02-11
Fixes #21629: std.csv report one record on empty input

burner:
https://issues.dlang.org/show_bug.cgi?id=21629

### DIFF
--- a/std/csv.d
+++ b/std/csv.d
@@ -890,6 +890,12 @@ public:
         _separator = delimiter;
         _quote = quote;
 
+        if (_input.range.empty)
+        {
+            _empty = true;
+            return;
+        }
+
         prime();
     }
 
@@ -919,6 +925,12 @@ public:
         _input = new Input!(Range, ErrorLevel)(input);
         _separator = delimiter;
         _quote = quote;
+
+        if (_input.range.empty)
+        {
+            _empty = true;
+            return;
+        }
 
         size_t[string] colToIndex;
         foreach (h; colHeaders)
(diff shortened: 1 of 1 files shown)


std/csv.d
```diff
@@ -1750,3 +1769,19 @@ if (isSomeChar!Separator && isInputRange!Range
...
+    struct Reccord
+    {
+        string a;
+        string b;
+    }
+
+    string input = "";
+    foreach (it; csvReader!Reccord(input))
+    {
+        assert(false, "We shouldn't make it here");
```
deadalnix, line 1785: Maybe you could count loop iterration and assert the count is zero so that the linter can be happy.

burner: done

deadalnix:
LGTM. I don't know what the next step is but as far as I'm concerned, this is good to go.

Maybe adding a test case with a CSV header then no data after and and checking that this also doesn't return any row, but this also could be done in another PR if that is hard, there is no reason to stop this due to scope creep.

burner:
> Maybe adding a test case with a CSV header then no data after and and checking that this also doesn't return any row, but this also could be done in another PR if that is hard, there is no reason to stop this due to scope creep.

That is actually tested.

deadalnix:
The current test checks csv("", ["a", "b"]) and verify it is empty

But afaik, this is not tested:
csv("a,b\n", ["a", "b"])

This also should return an empty range.

burner:
ah, I miss read. Looking at the other tests this should be handled already.

RazvanN7 (approved):

burner:
thank you @RazvanN7 @Geod24 @deadalnix

PR #7789, opened 2021-02-12
Fix Issue 21627 - macOS: std.stdio.File.sync does not guarantee to be…

kubo39:
… written to disk

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -58,18 +58,22 @@ else version (CRuntime_UClibc)
 else version (OSX)
 {
     version = GENERIC_IO;
+    version = Darwin;
 }
 else version (iOS)
 {
     version = GENERIC_IO;
+    version = Darwin;
 }
 else version (TVOS)
 {
     version = GENERIC_IO;
+    version = Darwin;
 }
 else version (WatchOS)
 {
     version = GENERIC_IO;
+    version = Darwin;
 }
 else version (FreeBSD)
 {
@@ -982,7 +986,9 @@ Call $(LREF flush) before calling this function to flush the C `FILE` buffers fi
 
 This function calls
 $(HTTP msdn.microsoft.com/en-us/library/windows/desktop/aa364439%28v=vs.85%29.aspx,
-`FlushFileBuffers`) on Windows and
+`FlushFileBuffers`) on Windows,
+$(HTTP developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fcntl.2.html,
+`F_FULLFSYNC fcntl`) on Darwin and
 $(HTTP pubs.opengroup.org/onlinepubs/7908799/xsh/fsync.html,
 `fsync`) on POSIX for the file handle.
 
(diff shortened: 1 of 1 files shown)


CyberShadow:
Looking around a bit, it looks like `F_FULLFSYNC` actually does something else than `fsync` on other POSIX platforms:

https://www.postgresql.org/message-id/a06010200be3da9564694@%5B17.202.21.231%5D

Important bit:

> The behavior
> of fsync() on MacOS X is the same as it is on every other version of
> Unix since the dawn of time (well, since the introduction of fsync
> anyway :-).

and

https://news.ycombinator.com/item?id=6792669

> I'm actually surprised Apple is even willing to make the guarantees above about F_FULLFSYNC; I'd read it as, at best, only applying to Apple hardware, as if a third-party controller is doing something silly they can't really do much about that.

So, it looks like there is no equivalent to `F_FULLFSYNC` for other platforms.

Whatever we choose to do, we should document that as the intent of the function, so that it can be updated consistently with new APIs / platforms. One of two things:

- Do we want the function to (vaguely) submit the data to disk? In which case we should keep using `fsync`.
- Do we want the function to provide a best-effort maximum guarantee (as far as the OS API allows) that the data has been permanently saved? In which case this is a change in that direction but the documentation ought to be updated to reflect that.

Finally, there is the question if there are circumstances where D programs wouldn't want to use `F_FULLFSYNC`, in which case this change would be a performance regression for them.

std/stdio.d
```diff
@@ -999,6 +1005,12 @@ Throws: `Exception` if the file is not opened or if the OS call fails.
             import core.sys.windows.winbase : FlushFileBuffers;
             wenforce(FlushFileBuffers(windowsHandle), "FlushFileBuffers failed");
         }
+        version (Darwin)
+        {
+            import core.sys.darwin.fcntl : fcntl, F_FULLFSYNC;
```
CyberShadow, line 1010: std/stdio.d(1010): Error: module `fcntl` is in file 'core/sys/darwin/fcntl.d' which cannot be read

CyberShadow: https://github.com/dlang/druntime/pull/3372 I see

n8sh:
>there is the question if there are circumstances where D programs wouldn't want to use `F_FULLFSYNC`, in which case this change would be a performance regression for them.

The one circumstance I can think of is if power loss to the storage device is believed to be impossible. Are there others?

std/stdio.d
```diff
@@ -982,7 +986,9 @@ Call $(LREF flush) before calling this function to flush the C `FILE` buffers fi
 
 This function calls
 $(HTTP msdn.microsoft.com/en-us/library/windows/desktop/aa364439%28v=vs.85%29.aspx,
-`FlushFileBuffers`) on Windows and
+`FlushFileBuffers`) on Windows,
+$(HTTP developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fcntl.2.html,
```
RazvanN7, line 990: This link does not exist

CyberShadow: Works here?

https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fcntl.2.html

RazvanN7: Hmm, it works now. I must have had a connection issue.

RazvanN7:
At the minimum, I think that we should specify that `fcntl` is called using `FULLSYNC`. @kubo39 What is your stance on this?

CyberShadow:
> At the minimum, I think that we should specify that `fcntl` is called using `FULLSYNC`.

Not sure if that's how you meant it, but just in case I'll clarify that it already does (the text says "..., `F_FULLFSYNC fcntl` on Darwin, and ...".

RazvanN7 (approved):

PR #7790, opened 2021-02-12
Improve documentation and errormessage of bitfields.

berni44:
A leftover of PR #7787.

I'm not sure what exactly is meant with the word 'bitfield': A single entry or the whole thing? (According to [Wikipedia](https://en.wikipedia.org/wiki/Bit_field) it is still something else, more or less a BitArray). I used `bitfield` for the whole thing, but after I finished got in doubt and are somewhat confused now. If you think it should be different, I'll reword it.

I also changed some ddoc style. IMHO structs and classes aren't params - it seems to me, that this was some sort of abuse of these macros. Having said this, I couldn't find a definitive description which macros should be used for Phobos docs. For me, it seems, every module uses it's own convention... If you've got a pointer to something useful, I'll be happy.

### DIFF
--- a/std/bitmanip.d
+++ b/std/bitmanip.d
@@ -171,7 +171,8 @@ private template createStorageAndFields(Ts...)
         alias StoreType = ulong;
     else
     {
-        static assert(false, "Field widths must sum to 8, 16, 32, or 64");
+        import std.conv : to;
+        static assert(false, "Field widths must sum to 8, 16, 32, or 64, not " ~ to!string(Size));
         alias StoreType = ulong; // just to avoid another error msg
     }
     enum result
@@ -247,12 +248,37 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
 }
 
 /**
-Allows creating bit fields inside $(D_PARAM struct)s and $(D_PARAM
-class)es.
+Allows creating `bitfields` inside `structs`, `classes` and `unions`.
 
-The type of a bit field can be any integral type or enumerated
-type. The most efficient type to store in bitfields is $(D_PARAM
-bool), followed by unsigned types, followed by signed types.
+A `bitfield` consists of one or more entries with a fixed number of
+bits reserved for each of the entries. The types of the entries can
+be `bool`s, integral types or enumerated types, arbitrarily mixed.
+The most efficient type to store in `bitfields` is `bool`, followed
+by unsigned types, followed by signed types.
+
+Each non-`bool` entry of the `bitfield` will be represented by the
+number of bits specified by the user. The minimum and the maximum
+numbers that represent this domain can be queried by using the name
+of the variable followed by `_min` or `_max`.
+
+Limitation: The number of bits in a `bitfield` is limited to 8, 16,
+32 or 64. If padding is needed, an entry should be explicitly
+allocated with an empty name.
+
+Implementation_details: `Bitfields` are internally stored in an
+`ubyte`, `ushort`, `uint` or `ulong` depending on the number of bits
+used. The bits are filled in the order given by the parameters,
+starting with the lowest significant bit. The name of the (private)
+variable used for saving the `bitfield` is created by concatenating
+all of the variable names, each preceded by an underscore.
+
+Params: T = A list of template parameters divided into chunks of 3
+            items. Each chunk consists (in this order) of a type, a
+            name and a number. Together they define an entry
+            of the `bitfield`: a variable of the given type and name,
+            which can hold as many bits as the number denotes.
+
+Returns: A string that can be used in a `mixin` to add the `bitfield`.
 
 See_Also: $(REF BitFlags, std,typecons)
 */
@@ -263,10 +289,10 @@ template bitfields(T...)
 }
 
 /**
-Create a bitfield pack of eight bits, which fit in
-one $(D_PARAM ubyte). The bitfields are allocated starting from the
-least significant bit, i.e. x occupies the two least significant bits
-of the bitfields storage.
+Create a `bitfield` pack of eight bits, which fit in
+one `ubyte`. The `bitfields` are allocated starting from the
+least significant bit, i.e. `x` occupies the two least significant bits
+of the `bitfields` storage.
 */
 @safe unittest
 {
@@ -291,7 +317,7 @@ of the bitfields storage.
 }
 
 /**
-The sum of all bit lengths in one $(D_PARAM bitfield) instantiation
+The sum of all bit lengths in one `bitfield` instantiation
 must be exactly 8, 16, 32, or 64. If padding is needed, just allocate
 one bitfield with an empty name.
 */


std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
 /**
-Allows creating bit fields inside $(D_PARAM struct)s and $(D_PARAM
-class)es.
+Allows creating `bitfields` inside `structs`, `classes` and `unions`.
 
-The type of a bit field can be any integral type or enumerated
-type. The most efficient type to store in bitfields is $(D_PARAM
-bool), followed by unsigned types, followed by signed types.
+A `bitfield` consists of one or more entries with a fixed number of
+bits reserved for each. The types of the enties can be `bool`s,
```
RazvanN7, line 254: of the **entries**

RazvanN7: I would still highlight that the most efficient type to store is a bool.

std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+integral types or enumerated types, arbitrarily mixed.
+
+Integral and Enumerated types are accompanied by two `enums`: `min`
+and `max`, giving the minimum and the maximum number, that can be
```
RazvanN7, line 258:
```suggestion
and `max`. The numbers that can be represented by the entry of the `bitfield` are in the interval [`min`, `max`].
```

std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+represented by this entry. The variable names are preceded by the
```
RazvanN7, line 259:
```suggestion
For a given bitfield entry `x`, these limits are represented by the name of the entry followed by an underscore and the limit name, e.g. `x_min` or `x_max`. 
```

std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+name of the variable followed by an underscore, so variable `x` comes
+along with `x_min` and `x_max`.
```
RazvanN7, line 261: Aren't `min` and `max` optional? If so, it should be specified. Also, what are the default values?

std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+
+The most efficient type to store in `bitfields` is `bool`, followed by
+unsigned types, followed by signed types.
```
RazvanN7, line 264: I think that this can be put in the first paragraph.

std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+
+Limitation: The number of bits in a `bitfield` is limited to 8, 16,
+32 or 64. If padding is needed, just allocate an entry with an empty
```
RazvanN7, line 267:
```suggestion
32 or 64. If padding is needed, an entry should be explicitly allocated with an empty name.
```

std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+name.
+
+Implementation_details: `Bitfields` are internally stored in an
+`ubyte`, `ushort`, `uint` or `ulong` depending on the number of bits
+used. The bits are filled in in order given by the parameters,
```
RazvanN7, line 272: I think that the information given from this sentence to the end of the paragraph is too much. Specifying the internal storage should be enough.

std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+starting with the lowest significant bit. The name of the (private)
+variable used for saving the `bitfield` is created by sticking
+together all names of the variables, each preceded by an underscore.
+
+Params: T = A list of template parameters divided into chunks of 3
+            items. Each chunk consists (in this order) of a type a
```
RazvanN7, line 278:
```suggestion
            items. Each chunk consists (in this order) of a type, a
```

std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+            variable name and a number. Together they define an entry
+            of the `bitfield`: A variable of the given type and name,
```
RazvanN7, line 280:
```suggestion
            of the `bitfield`: a variable of the given type and name,
```

std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+            which can hold as many bits, as the number denotes.
```
RazvanN7, line 281:
```suggestion
            which can hold as many bits as the number denotes.
```

RazvanN7 (changes requested):
See comments.

std/bitmanip.d
berni44: The mixin always adds x_min (etc.) and x_max (etc.) for integeres and enums, but not for bools. I'm not sure, if this could be clearified better.

std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
-Allows creating bit fields inside $(D_PARAM struct)s and $(D_PARAM
-class)es.
-
-The type of a bit field can be any integral type or enumerated
-type. The most efficient type to store in bitfields is $(D_PARAM
-bool), followed by unsigned types, followed by signed types.
+Allows creating `bitfields` inside `structs`, `classes` and `unions`.
+
+A `bitfield` consists of one or more entries with a fixed number of
+bits reserved for each of the entried. The types of the enties can be
```
RazvanN7, line 254:
```suggestion
bits reserved for each of the entries. The types of the entries can be
```

std/bitmanip.d
berni44: I disagree here. For example, have a look at FloatRep further down this file. If there is no guarantee on the order of the entries, a user cannot reliably do such things.

The name of the private variable, well yes, probably unnecessary. I thought it might prevent a name clash, to know about it. But I must admit, that this is very unlikely. And I checked the error message, it directly points to the mixin as the cause, so people might infere the reason without having to peek into the source code.

std/bitmanip.d
```diff
@@ -247,12 +248,39 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+`bool`s, integral types or enumerated types, arbitrarily mixed. The
+most efficient type to store in `bitfields` is `bool`, followed by
+unsigned types, followed by signed types.
+
+Integral and Enumerated types are accompanied by two `enums`: `min`
+and `max`. The numbers that can be represented by the entry of the
+`bitfield` are in the interval [`min`, `max`]. For a given bitfield
+entry `x`, these limits are represented by the name of the entry
+followed by an underscore and the limit name, e.g. `x_min` or
+`x_max`.
```
RazvanN7, line 264: I see, so there are generated behind the scenes and the user can access them by using _min/_max.
I think that this paragraph should be rephrased to make this clear.

Something like: "Each non-bool entry of the bitfield will be represented by an arbitrary number of bits. The minimum and the maxim numbers that represent this domain can be queried by using the name of the variable followed by _min or _max"

RazvanN7: Actually, the number of bits is not arbitrary, it is user-specified.

std/bitmanip.d
RazvanN7: Yeah, that makes sense. The please update the double in.
```suggestion
used. The bits are filled in the order given by the parameters,
```

std/bitmanip.d
```diff
@@ -247,12 +248,37 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+Implementation_details: `Bitfields` are internally stored in an
+`ubyte`, `ushort`, `uint` or `ulong` depending on the number of bits
+used. The bits are filled in the order given by the parameters,
+starting with the lowest significant bit. The name of the (private)
+variable used for saving the `bitfield` is created by sticking
+together all names of the variables, each preceded by an underscore.
+
+Params: T = A list of template parameters divided into chunks of 3
+            items. Each chunk consists (in this order) of a type, a
+            variable name and a number. Together they define an entry
```
RazvanN7, line 277:
```suggestion
            name and a number. Together they define an entry
```

std/bitmanip.d
```diff
@@ -247,12 +248,37 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+
+Limitation: The number of bits in a `bitfield` is limited to 8, 16,
+32 or 64. If padding is needed, an entry should be explicitly
+allocated with an empty name.
+
+Implementation_details: `Bitfields` are internally stored in an
+`ubyte`, `ushort`, `uint` or `ulong` depending on the number of bits
+used. The bits are filled in the order given by the parameters,
+starting with the lowest significant bit. The name of the (private)
+variable used for saving the `bitfield` is created by sticking
```
RazvanN7, line 272:
```suggestion
variable used for saving the `bitfield` is created by concatenating
```

std/bitmanip.d
```diff
@@ -247,12 +248,37 @@ private template createTaggedReference(T, ulong a, string name, Ts...)
...
+together all names of the variables, each preceded by an underscore.
```
RazvanN7, line 273:
```suggestion
all of the variable names, each preceded by an underscore.
```

RazvanN7 (commented):
Sorry for being so rigorous, but this one seems to be the last round.

berni44:
> Sorry for being so rigorous, but this one seems to be the last round.

Thank for being so rigorous. :-) I think it's a good thing to do so.

RazvanN7 (approved):

RazvanN7:
Thank you! I will let another 3 days pass until merging so that other people get a chance to take a look.

PR #7575, opened 2020-08-03
Fix issue 16288 - Add splitter functionality that allows to keep the sentinels

Robert-Aron293:
Currently the splitter can't be configured to keep the separators.
By using the flag Yes.keepSeparators, the splitter will keep the separators in the bucket.

Examples:
 "a|bc|def".splitter!("a == b", Yes.keepSeparators)('|')
> [ "a", "|", "bc", "|", "def" ]

"|ab|".splitter!("a == b", Yes.keepSeparators)('|')
> [ "", "|", "ab", "|", "" ]

"a=>bc=>def".splitter!("a == b", Yes.keepSeparators)("=>")
> [ "a", "=>", "bc", "=>", "def" ]

"abXcdxef".splitter!("a.toLower == b", Yes.keepSeparators)('x')
> [ "ab", "X", "cd", "x", "ef" ]

"a|bc|def".splitter!("a == b", Yes.keepSeparators)('|').retro
> [ "def", "|", "bc", "|", "a" ]

PR #7794, opened 2021-02-13
 Fix Issue 13595 - added splitWhen

dukc:
Reboot of #7425

EDIT: No longer just a reboot, has been refactored to a bit different api


changelog/chunkbyany.dd
```diff
@@ -0,0 +1,18 @@
+$(REF chunkByAny, std, algortihm, iteration) added to std.algorithm.iteration
+
+chunkByAny is a variant of existing chunkBy that does not except it's predicate
+to have an equivelance relation, letting you use the function more creatively:
```
pbackus, line 4:
```suggestion
chunkByAny is a variant of the existing chunkBy function that does not require
its predicate to be an equivelance relation, allowing it to be used in ways that
chunkBy cannot. For example:
```

changelog/chunkbyany.dd
```diff
@@ -0,0 +1,18 @@
...
+auto r3 = [1, 3, 2, 5, 4, 9, 10].chunkByAny!((a, b) => abs(a-b) < 3);
+assert(r3.equal!equal([
+    [1, 3, 2],
+    [5, 4],
+    [9, 10]
+]));
+-------
+
+This is illegal to do with chunkBy, because it requires that if pred(a,b) and
+pred(b,c) return true, pred(a,c) must return true too.
```
pbackus, line 18:
```suggestion
This is illegal with chunkBy, because it requires that if pred(a,b) and
pred(b,c) return true, pred(a,c) must also return true.
```

std/algorithm/iteration.d
```diff
@@ -2695,12 +2740,110 @@ version (none) // this example requires support for non-equivalence relations
...
+{
+    [""].map!((s) => s).chunkBy!((x, y) => true);
+}
+
+/**
+Like $(REF chunkBy), but does not require reflexivity, symmetricity nor
+transitivity of the binary predicate. Unlike `çhunkBy`, requires the
+source to be a forward range.
+With binary predicates, the element closer to front is always passed
+first to the predicate.
```
pbackus, line 2756:
```suggestion
With binary predicates, the element closer to front is always passed
as the first argument to the predicate.
```

std/algorithm/iteration.d
```diff
@@ -2695,12 +2740,110 @@ version (none) // this example requires support for non-equivalence relations
...
+
+// https://issues.dlang.org/show_bug.cgi?id=13805
 @system unittest
+{
+    [""].map!((s) => s).chunkBy!((x, y) => true);
+}
+
+/**
+Like $(REF chunkBy), but does not require reflexivity, symmetricity nor
+transitivity of the binary predicate. Unlike `çhunkBy`, requires the
```
pbackus, line 2753:
```suggestion
transitivity of the binary predicate. Unlike `chunkBy`, requires the
 ```

File: std/algorithm/iteration.d [hunk 5]
```diff
@@ -2695,12 +2740,110 @@ version (none) // this example requires support for non-equivalence relations
...
+source to be a forward range.
+With binary predicates, the element closer to front is always passed
+first to the predicate.
+When called with unary predicate, the comparisons are still assumed
+to be reflexive, symmetric and transitive. However, the source range is
+nonetheless required to be a forward range.
```
pbackus, line 2759: Is this limitation really necessary?

pbackus (changes requested):

pbackus:
I'm not sure `chunkByAny` is the best name for this—it invites the question, "chunk by any *what?*" It took me a minute of reading to understand that "any" referred to the type of predicate allowed, and I expect any newcomers to Phobos would be similarly confused.

Maybe a better approach would be to name it `splitBy` and invert the predicate, so that instead of
```d
[1, 3, 2, 5, 4, 9, 10].chunkByAny!((a, b) => abs(a-b) < 3)
```
...one would write:
```d
[1, 3, 2, 5, 4, 9, 10].splitBy!((a, b) => abs(a-b) >= 3);
```

dukc:
> Maybe a better approach would be to name it `splitBy` and invert the predicate

At least quickly thinking, this sounds a good idea. I'll rewrite.

std/algorithm/iteration.d
dukc: Which one - that the source range needs to be a forward range, or that we require the potential `==` overload of the element type to be reflexive, symmetric and transitive?

std/algorithm/iteration.d
dukc: I quess I'm going to give another pass to the documentation anyway when I've written it to `splitBy` you suggested.

std/algorithm/iteration.d
pbackus: That the source needs to be a forward range. Unless I'm missing something, this function has the same behavior as regular `chunkBy` in the unary-predicate case, so it's strange for it to have different requirements.

dukc: The logic goes like this: If the user has used `chunkByAny` instead of `chunkBy`, she probably wants to make sure that if the predicate is changed to a binary one, it's reflexivity, symmetricity and transitivity are not assumed. So we check that the code would compile even if we could not assume the usual things about the predicate.

That the equivalency is still assumed for unary predicates is just a kind of optimization - we assume the user is not ridiculous enough to define a `==` overload that would not satisfy `chunkBy` requirements.

pbackus: If the idea is that `chunkByAny` (or `splitBy`) should only be used when the programmer intends to use a binary predicate (either now or in the future), then maybe the right thing to do is to *only* accept binary predicates, and leave the unary predicate case to regular `chunkBy`.

dukc: I might otherwise argue with this, but as we're taking some distance from `chunkBy` anyway with the refactoring to `splitBy`, I can just as well do that.

dukc:
Rewrite done

std/algorithm/iteration.d
```diff
@@ -2695,12 +2743,110 @@ version (none) // this example requires support for non-equivalence relations
...
+/**
+Splits a forward range into subranges in places determined by a binary
+predicate.
+
+When iterating, one element of `r` is compared with `pred` to the next
+element. If `pred` return true, a new subrange is started for the next element.
+Otherwise, they are part of the same subrange.
+
+If the elements are compared with an unequivalency (!=) operator, it's recommended
+to consider $(LREF chunkBy) instead, as it's likely to take less time to execute.
```
pbackus, line 2763:
```suggestion
If the elements are compared with an inequality (`!=`) operator, consider
$(LREF chunkBy) instead, as it's likely to take less time to execute.
```

std/algorithm/iteration.d
```diff
@@ -2228,7 +2286,8 @@ if (isForwardRange!Range)
  * all elements in a given subrange are equivalent under the given predicate.
  * With a unary predicate, a range of tuples is returned, with the tuple
  * consisting of the result of the unary predicate for each subrange, and the
- * subrange itself.
+ * subrange itself. The range is currently a reference range, but this may
+ * change in the future.
```
pbackus, line 2290: "Reference range" could be read as meaning either that the range's elements are references, or that the range itself is a reference. Probably better to be more specific; something like, "copies of the range currently have reference semantics".

std/algorithm/iteration.d
```diff
@@ -2695,12 +2743,110 @@ version (none) // this example requires support for non-equivalence relations
...
+to consider $(LREF chunkBy) instead, as it's likely to take less time to execute.
+
+Params:
+pred = Predicate for determining where to split. The earlier element in the
+source range is always given as the first argument.
+r = A $(REF_ALTTEXT forward range, isForwardRange, std,range,primitives) to be split.
+Returns: a range of ranges in which
+all elements in a given subrange are `false` when compared to the next element
+with the predicate. The range is currently a reference range, but this may
+change in the future.
```
pbackus, line 2772: This one's tough. Here's my best attempt:
```suggestion
Returns: a range of subranges of `r`, split such that within a given subrange,
calling `pred` with any pair of adjacent elements as arguments returns `false`.
Copies of the range currently have reference semantics, but this may change in
the future.
```

std/algorithm/iteration.d
```diff
@@ -4672,8 +4825,9 @@ Returns:
 
 See_Also:
  $(REF _splitter, std,regex) for a version that splits using a regular
-expression defined separator and
- $(REF _split, std,array) for a version that splits eagerly.
+expression defined separator, 
+ $(REF _split, std,array) for a version that splits eagerly and
+ $(LREF splitBy), useful for splitting without dropping any source range elements.
```
pbackus, line 4830: `splitter` is going to get this feature in #7575, so you may want to reword this.

dukc:
Suggestions applied almost as-is

pbackus:
LGTM. Hopefully someone with commit access can approve this soon.

thewilsonator:
Looks like some network issues, can you please rebase or force push to restart CI.

dukc:
> Looks like some network issues, can you please rebase or force push to restart CI.

It wasn't only network issues, it was a trailing white space I was too lazy to delete earlier

dukc:
Adding a new symbol so I think I need to ping @atilaneves

RazvanN7:
Please add a commit that includes  "Fix Issue 13595" in the commit message so that the bot automatically links this PR to the associated bugzilla entry.

std/algorithm/iteration.d
```diff
@@ -2695,12 +2743,109 @@ version (none) // this example requires support for non-equivalence relations
...
+    import std.algorithm.comparison : equal;
+    import std.math : abs;
+
+    struct SomeRange
+    {
+        int[] elements;
+        static int popfrontsSoFar;
+
+        auto front(){return elements[0];}
+        auto popFront()
```
RazvanN7, line 2821: For some reason, dscanner wants this to be spelled out as `void`.

```bash
std/algorithm/iteration.d(2821:14)[warn]: Auto function without return statement, prefer an explicit void
```

dukc:
I changed the a test range return type of `popFront` from `auto` to `void` and suddently it's comparing it's `.splitBy` lost `nothrow`. WAT?

Gotta investigate a bit.

dukc:
It appears that attribute inference works for non-template member functions if the return type is `auto` but not if it's void. I wonder if that's intended.

RazvanN7:
> It appears that attribute inference works for non-template member functions if the return type is auto but not if it's void.

Inference works for any kind of `auto` function because those are templated functions behind the scenes. That is intended behavior.

RazvanN7:
The buildkite error:

```d

_D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoThTaZQCaFNaNfKQBrhMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T19needToSwapEndianessTaZQyFNaNbNiNfMKxSQCbQCa__T10FormatSpecTaZQpZb'
--
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoThTaZQCaFNaNfKQBrhMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T14formatIntegralTSQBg5array__T8AppenderTAyaZQoTmTaZQBzFNaNfKQBrxmMKxSQDfQDe__T10FormatSpecTaZQpkmZv'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoTkTaZQCaFNaNfKQBrkMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T19needToSwapEndianessTaZQyFNaNbNiNfMKxSQCbQCa__T10FormatSpecTaZQpZb'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoTkTaZQCaFNaNfKQBrkMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T14formatIntegralTSQBg5array__T8AppenderTAyaZQoTmTaZQBzFNaNfKQBrxmMKxSQDfQDe__T10FormatSpecTaZQpkmZv'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTyAaZQoThTaZQCaFNaNfKQBrhMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T19needToSwapEndianessTaZQyFNaNbNiNfMKxSQCbQCa__T10FormatSpecTaZQpZb'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTyAaZQoTkTaZQCaFNaNfKQBrkMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T19needToSwapEndianessTaZQyFNaNbNiNfMKxSQCbQCa__T10FormatSpecTaZQpZb'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoTmTaZQCaFNaNfKQBrmMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T14formatIntegralTSQBg5array__T8AppenderTAyaZQoTmTaZQBzFNaNfKQBrxmMKxSQDfQDe__T10FormatSpecTaZQpkmZv'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoTxhTaZQCbFNaNfKQBsxhMKxSQDhQDg__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T14formatIntegralTSQBg5array__T8AppenderTAyaZQoTmTaZQBzFNaNfKQBrxmMKxSQDfQDe__T10FormatSpecTaZQpkmZv'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std4conv__T7emplaceTCQw5regex8internal12backtracking__T19BacktrackingMatcherTaTSQDdQCiQCf2ir__T5InputTaZQjZQCaTxSQEkQDpQDmQBh__T5RegexTaZQjTQCjTAvZQFgFNaQjKxQBtQDdQtZQFq: error: undefined reference to '_D3std4conv16testEmplaceChunkFNaNbNiNfAvmmZv'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std4conv__T7emplaceTCQw5regex8internal8thompson__T15ThompsonMatcherTaTSQCuQBzQBw2ir__T5InputTaZQjZQBwTxSQEbQDgQDdQBh__T5RegexTaZQjTQCjTAvZQExFNaQjKxQBtQDdQtZQFh: error: undefined reference to '_D3std4conv16testEmplaceChunkFNaNbNiNfAvmmZv'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std4conv__T7emplaceTCQw5regex8internal12backtracking__T19BacktrackingMatcherTaTSQDdQCiQCf2ir__T5InputTaZQjZQCaTxSQEkQDpQDmQBh__T5RegexTaZQjTPFNaNeQEwZbTQCvTAvZQFsFNaQjKxQCfQBgQDsQwZQGf: error: undefined reference to '_D3std4conv16testEmplaceChunkFNaNbNiNfAvmmZv'
  | generated/linux/release/64/unittest/test_runner core.sys.solaris.link
  | generated/linux/release/64/unittest/test_runner core.sys.solaris.stdlib
  | generated/linux/release/64/unittest/test_runner core.sys.solaris.time
  | generated/linux/release/64/unittest/test_runner core.sys.solaris.sys.elf
  | posix.mak:462: recipe for target 'generated/linux/release/64/benchmark' failed
  | make: *** [generated/linux/release/64/benchmark] Error 1
```

I've seen this on #7801 also. Doesn't look spurios.

RazvanN7:
Maybe a rebase might fix the buildkite failure.

dukc:
> Maybe a rebase might fix the buildkite failure.

I'll try that at some point.

dukc:
There's the question if I should reimplement the function to return a value (non-ref) range. Now, it's a reference range because it shares implementation with `chunkBy` that happens to be a reference range.

I did document that the copy semantics may change in the future, but I wonder if it's enough? the consensus at https://forum.dlang.org/thread/ztgtmumenampiobbuiwd@forum.dlang.org?page=1 seems to be that the user should not rely on undefined copy semantics of a Phobos range, but accidents are going to happen.

RazvanN7 (commented):
Can you also add a test of `splitBy` with an empty range?

changelog/splitby.dd
```diff
@@ -0,0 +1,20 @@
+splitBy added to std.algorithm.iteration
+
+$(REF splitBy, std, algortihm, iteration) is a variant of the existing $(REF
+chunkBy, std, algortihm, iteration) function that does not require its predicate
+to be an equivelance relation, allowing it to be used in ways that chunkBy
```
atilaneves, line 5: equivalence

changelog/splitby.dd
```diff
@@ -0,0 +1,20 @@
...
+import std.algorithm;
+auto r3 = [1, 3, 2, 5, 4, 9, 10].splitBy!((a, b) => abs(a-b) >= 3);
+assert(r3.equal!equal([
+    [1, 3, 2],
+    [5, 4],
+    [9, 10]
+]));
+-------
+
+This is illegal with chunkBy, because it requires that if pred(a,b) and
```
atilaneves, line 19: I'd add what "illegal" means here, notably that it throws an AssertError.

std/algorithm/iteration.d
```diff
@@ -2284,20 +2346,6 @@ if (isInputRange!Range)
     ]));
 }
 
-version (none) // this example requires support for non-equivalence relations
```
atilaneves, line 2287: It'd be better instead to keep this and assert it fails.

std/algorithm/iteration.d
```diff
@@ -2695,12 +2743,109 @@ version (none) // this example requires support for non-equivalence relations
...
+relations.
+*/
+
+auto splitBy(alias pred, Range)(Range r)
+if (isForwardRange!Range)
+{   import std.functional : not;
+    return ChunkByImpl!(not!pred, not!pred, GroupingOpType.binaryAny, Range)(r);
+}
+
+//FIXME: these should be @safe
```
atilaneves, line 2784: Why aren't they `@safe`?

dukc: Because `splitBy` shares implementation with `chunkBy`, which is not currently `@safe` either.

changelog/splitby.dd
dukc: Oh true, it isn't illegal - it just causes `chunkBy to malfunction. I'll change.

std/algorithm/iteration.d
dukc: You mean duplicate it? I didn't remove it, just moved it under `splitBy` and changed to use it instead.

dukc:
> Can you also add a test of `splitBy` with an empty range?

Okay

std/algorithm/iteration.d
dukc: Tried to keep the test-fail case, but it failed by triggering an implementation-specific assert:

```
    this(RefCounted!(OuterRange) origin)
    {
        groupNum = origin.groupNum;
        current = origin.current.save;
        assert(!current.empty, "Passed range 'r' must not be empty");
        static if (eqEquivalenceAssured)
        {
            start = origin.current.save;

            // Check for reflexivity.
            assert(eq(start.front, current.front), // <------ this failed
                "predicate is not reflexive");
        }

        mothership = origin;
    }
```

Testing for that would be quite fragile in my opinion.

std/algorithm/iteration.d
```diff
@@ -2709,10 +2852,25 @@ version (none) // This requires support for non-equivalence relations
...
+        [5, 4],
+        [9, 10]
+    ]));
+}
+
+// empty range splitBy
+@nogc nothrow pure @system unittest
+{
+    int[1] sliceable;
+    auto result = sliceable[0..0].splitBy!((a,b) => a+b > 10);
```
RazvanN7, line 2872: The style checker fails on buildkite because it should be `sliceable[0 .. 0]`.

RazvanN7 (approved):

dukc:
Wait, Atila didn't approve yet. This is a new phobos symbol so can't merge before that.

RazvanN7:
I have interpreted @atilaneves 's  willingness to offer feedback on this PR as a sign that he agrees with this addition. Maybe I was wrong. @atilaneves what is your stance on this?

atilaneves:
I'm slightly unsure of the name, although the documentation does tell the user what the difference is.

RazvanN7:
I don't have any better ideas, but `splitBy` sounds intuitive to me.

dukc:
Let's leverage the forum! https://forum.dlang.org/post/izlcfnaqwzpjmfoyohur@forum.dlang.org

RazvanN7:
`splitWhen` sounds very good and is actually more descriptive than `splitBy`

atilaneves (approved):

RazvanN7:
@dukc It seems like @atilaneves has approved this PR. Judging from the likes of my previous post it seems that the name for this function should be `splitWhen`.

dukc:
`splitWhen` it is then. Wait until I have renamed, then you can do a quick recheck and turn auto-merge on

pbackus:
The CI failures should be fixed by a rebase.

PR #7795, opened 2021-02-15
std.system: adds enum OS.unknown

denizzzka:
It is not supported something like "Non-Posix" version in the compilers so I propose this enum here.

Related:
https://github.com/denizzzka/d_c_arm_test/issues/18#issuecomment-766745608


rikkimax:
Unknown would be a better name. It is not tying itself to another platform.

denizzzka:
> Unknown would be a better name.

unknownNonPosix ?

n8sh (approved):

PR #7796, opened 2021-02-15
Prepare to split up std.format into submodules.

berni44:
Now, that no more conflicting PRs are out there I wanted to reopen #7785, but probably got the order of steps wrong. Github now declines me doing so...

The idea is, to make very small steps here. One of the main problem of splitting modules is, that ideally no other PR should be merged during the process, because it's easy to overlook some of these changes while rebasing a splitting PR. So this PR just moves everything to std/format/package.d (with only a single slight change, that was needed).

See also PR #7765 for the main idea behind this.

Concerning the timing: In my opinion we can already do some preparatory steps before 2.096.0 is out: Moving to `package.d` (this PR) as well as extracting the internals. For me this would be convenient, because I could continue working on formatting floats earlier. But if you think, we should wait until March, I'll be happy to wait.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -207,13 +207,13 @@ P2MODULES=$(foreach P,$1,$(addprefix $P/,$(PACKAGE_$(subst /,_,$P))))
 STD_PACKAGES = std $(addprefix std/,\
   algorithm container datetime digest experimental/allocator \
   experimental/allocator/building_blocks experimental/logger \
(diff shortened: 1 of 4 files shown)


PetarKirov (approved):

PetarKirov:
Nice work! This is indeed small enough but useful step in the right direction.

PR #7798, opened 2021-02-16
Splitting std.format: Move floating point algorithms into internal submodule

berni44:
Followup on #7796. See #7765 for the main idea.


thewilsonator (approved):

PetarKirov (approved):
I compared locally the moved code and the change looks good to me.

PR #7800, opened 2021-02-16
ascii: rewrite function to naturally generate more optimized code

ljmf00:
A lot of functions inside std.ascii and std.uni are fast only optimized by the
compiler optimizer. Some functions can be written to naturally generate more
optimized code.

---

Take the example of isDigit implementation:

Here are some performance tests comparing std and new std implementations. Running with 100 000 000 randomized iterations without running the optimizer

```
DMD64 v2.095.0
2 ms, 253 μs, and 9 hnsecs
2 ms, 242 μs, and 6 hnsecs

LDC 1.24.0
2 ms, 148 μs, and 2 hnsecs
2 ms, 105 μs, and 6 hnsecs

GDC 10.2.0
2 ms, 266 μs, and 5 hnsecs
2 ms, 251 μs, and 5 hnsecs
```

NOTE: These times are for 100 000 runs because the benchmark was made with 1000 different random seeds.

Generated Assembly for both implementations without the optimizer: https://godbolt.org/z/sav5qe
Benchmark source code: https://gist.github.com/run-dlang/8198a98f4fb4e99120f6351bd764408a

### DIFF
--- a/std/ascii.d
+++ b/std/ascii.d
@@ -187,7 +187,7 @@ else
   +/
 bool isAlphaNum(dchar c) @safe pure nothrow @nogc
 {
-    return c <= 'z' && c >= '0' && (c <= '9' || c >= 'a' || (c >= 'A' && c <= 'Z'));
+    return isAlpha(c) || isDigit(c);
 }
 
 ///
@@ -219,7 +219,7 @@ bool isAlphaNum(dchar c) @safe pure nothrow @nogc
 bool isAlpha(dchar c) @safe pure nothrow @nogc
 {
     // Optimizer can turn this into a bitmask operation on 64 bit code
-    return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z');
+    return (c | 32) - 'a' < 26;
 }
 
 ///
@@ -250,7 +250,7 @@ bool isAlpha(dchar c) @safe pure nothrow @nogc
   +/
 bool isLower(dchar c) @safe pure nothrow @nogc
 {
-    return c >= 'a' && c <= 'z';
+    return c - 'a' < 26;
 }
 
 ///
@@ -282,7 +282,7 @@ bool isLower(dchar c) @safe pure nothrow @nogc
   +/
 bool isUpper(dchar c) @safe pure nothrow @nogc
 {
-    return c <= 'Z' && 'A' <= c;
+    return c - 'A' < 26;
 }
 
 ///
@@ -314,7 +314,7 @@ bool isUpper(dchar c) @safe pure nothrow @nogc
   +/
 bool isDigit(dchar c) @safe pure nothrow @nogc
 {
-    return '0' <= c && c <= '9';
+    return c - '0' < 10;
 }
 
 ///
@@ -347,7 +347,7 @@ bool isDigit(dchar c) @safe pure nothrow @nogc
   +/
 bool isOctalDigit(dchar c) @safe pure nothrow @nogc
 {
-    return c >= '0' && c <= '7';
+    return c - '0' < 8;
 }
 
 ///
@@ -377,7 +377,7 @@ bool isOctalDigit(dchar c) @safe pure nothrow @nogc
   +/
 bool isHexDigit(dchar c) @safe pure nothrow @nogc
 {
-    return c <= 'f' && c >= '0' && (c <= '9' || c >= 'a' || (c >= 'A' && c <= 'F'));
+    return isDigit(c) || (c | 32) - 'a' < 6;
 }
 
 ///
@@ -410,7 +410,7 @@ bool isHexDigit(dchar c) @safe pure nothrow @nogc
   +/
 bool isWhite(dchar c) @safe pure nothrow @nogc
 {
-    return c == ' ' || (c >= 0x09 && c <= 0x0D);
+    return c == ' ' || c - '\t' < 5;
 }
 
 ///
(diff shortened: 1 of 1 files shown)


maxhaton:
https://godbolt.org/z/MeT148 

Look at the code generated by dmd with no optimizations - it looks longer, right? Well now consider that your processor can do *billions* of instructions per second - the only part of this code that actually makes any difference to performance in a modern processor is the branch (`cmp`) which is unavoidable in the general case. 

Also, if we plot your benchmarking code with the call to isDigitX *removed*, we will see that the tiny optimization you report is only a fraction of a standard deviation away from  the mean. 
image

The black lines are at +- (253-242)

n8sh:
The benchmark doesn't make a convincing case for these changes but I find some of the code more beautiful.

std/ascii.d
```diff
@@ -250,7 +250,7 @@ bool isAlpha(dchar c) @safe pure nothrow @nogc
   +/
 bool isLower(dchar c) @safe pure nothrow @nogc
 {
-    return c >= 'a' && c <= 'z';
+    return c - 'a' < 26;
```
n8sh, line 253: For instance this is equally clear but it looks tidier.

berni44: The disadvantage of the new version is, that one has to look up the definition of `c` to find out if only the letters are meant or all ascii symbols up to 'z'.

berni44:
We had a similar discussion at PR #7773. When I opened this PR I was convinced, that having faster code would be a good idea. After the comment from @kinke I changed my mind: Meanwhile I think, we should restrain from such optimizations if the compilers manage to do the optimizations on their own. There is little speed gain (because when one is interested in speed one will use -O and the change comes down to nothing), but the code get's worse to read and understand.

std/ascii.d
ljmf00: > The disadvantage of the new version is, that one has to look up the definition of `c` to find out if only the letters are meant or all ascii symbols up to 'z'.

Commenting would be a solution for more unreadable logic.

ljmf00:
> [godbolt.org/z/MeT148](https://godbolt.org/z/MeT148)

Just a note, `dmd-nightly` is as old as `dmd-nighlty` in `dlang-core` from https://run.dlang.org/ . It's preferable to use the latest stable version rather than using that version. Somehow is stuck in the time. Maybe it's the same issue of DMD used in dub.

> Look at the code generated by dmd with no optimizations - it looks longer, right? Well now consider that your processor can do _billions_ of instructions per second - the only part of this code that actually makes any difference to performance in a modern processor is the branch (`cmp`) which is unavoidable in the general case.

Sure. I understand the point that optimizers will perfectly do their job, although, when we are talking about those micro-optimizations, jumps are costly, and generating code with a good CPU pipelining flow is preferable. If we take the example of a more "jumpy" generated code, `isAlpha`, we see a tremendous improvement by getting rid of 3 jumps and writing only one time to the CPU registers. And we shouldn't always focus on x86, if we look at ARM-based generated code, it's even worse!

https://godbolt.org/z/E7P4Yx

> Also, if we plot your benchmarking code with the call to isDigitX _removed_, we will see that the optimization you report is only a fraction of a standard deviation away from the mean.

Where did you get those plots btw? Looks fancy :sweat_smile:

kinke:
> and generating code with a good CPU pipelining flow is preferable

If the optimizers do that and the assumption holds that people don't care about saving some nanoseconds for debug builds, it's IMO preferable to keep the code nice, simple & self-explanatory instead of having to add a comment explaining what the magic trick in the next line does.

std/ascii.d
```diff
@@ -187,7 +187,7 @@ else
   +/
 bool isAlphaNum(dchar c) @safe pure nothrow @nogc
 {
-    return c <= 'z' && c >= '0' && (c <= '9' || c >= 'a' || (c >= 'A' && c <= 'Z'));
+    return isAlpha(c) || isDigit(c);
```
kinke, line 190: E.g., this does look like an improvement in readability.

ljmf00: Sure, I can do a PR changing only those things then. I also tried to force inlining for those functions, but DMD is acting weirdly on coverage reports.

maxhaton: Forcing inlining is a fairly pointless microptimization unless there is a bug in the optimizer. 

These types of functions are literally like shooting fish in a barrel with a machine gun for a compiler.

kinke: [Without `pragma(inline, true)`, you don't get cross-module inlining with LDC. So if you don't use LTO (incl. linking the Phobos LTO lib) and don't want to experiment with `-enable-cross-module-inlining`, selective usage of `pragma(inline, true)` for hot little functions can pay off.]

maxhaton:
I've managed to edit your comment by accident sorry. Anyway the script is a python script I use for physics stuff with the styling removed. 

This is still small fry for a function that is typically not called very often - it will be inlined

std/ascii.d
maxhaton: https://d.godbolt.org/z/9Pbof7 

I wasn't aware of this limitation, so in that case I am in favour - the only thing worth considering is potentially profiling the optimized dmd build since I have seen some comments suggesting performance issues with inlining (I personally don't really care about the performance of dmd release builds but I guess someone might)

maxhaton:
@ljmf00 Unfortunately I can't guesstimate the throughput of the arm code by eye and I don't have a box to test it on, but worrying about these things is really not helpful for code built with optimizations turned off - the compiler is barely doing register allocation. 

If you profile this kind of (verbose, let's say) generated code in vTune or similar, you can see that - on a microarchitectural level - the behaviour is so inconsistent that (although there is a trend) an energy spent optimizing things in this manner is rather fruitless in the long run. One problem that you may have bumped into is that doing DIY benchmarking at the level of microarchitecture can be a complete pain because even the act of extracting the data you recorded can affect the measurement.

PR #7801, opened 2021-02-17
fix issue 21592 - two stack traces if high surrogate is printed

aG0aep6G:


aG0aep6G:
An alternative fix would be to reset `highSurrogate` already when throwing the exception originally. But that would alter the behavior of `LockingTextWriter` more than is strictly necessary. The following is currently possible but wouldn't work anymore with the alternative fix:

```d
import std.stdio;
import std.utf : UTFException;
void main()
{
    auto w = stdout.lockingTextWriter();
    w.put("\U00010400"w[0]);
    try w.put('\n');
    catch (UTFException e) w.put("\U00010400"w[1]);
}
```

aG0aep6G:
This doesn't work like I thought it would. I'll have to take another look.

aG0aep6G:
> This doesn't work like I thought it would. I'll have to take another look.

Fixed. I'm not happy with throwing on incomplete UTF-16 while silently ignoring incomplete UTF-8 sequences. But sorting that out is beyond the scope of this PR.

Good to go from my side.

PetarKirov:
Perhaps I'm missing something obvious, but it seems that this shorter implementation doesn't have this bug (or at least the unittest does not trigger it):

```D
void write(S...)(S args)
{
    import std.format : formattedWrite;
    import std.stdio : stdout;
    import std.typecons : tuple;
    auto w = lockingTextWriter();
    formattedWrite(w, "%(%s%|%)", tuple(args));
}
```

Full example:
https://run.dlang.io/gist/run-dlang/a8d99f54cbbe57748faa10e41d847723?compiler=dmd

```D
import std.stdio : File;

@safe unittest // https://issues.dlang.org/show_bug.cgi?id=21592
{
    import std.exception : collectException;
    import std.utf : UTFException;
    static import std.file;
    auto deleteme = "deleteme.txt";
    scope(exit) std.file.remove(deleteme);
    auto f = File(deleteme, "w");
    auto e = collectException!UTFException(f.myWriteln(wchar(0xD801)));
    assert(e.next is null);
}

void myWriteln(S...)(ref File f, S args)
{
    myWrite(f, args, '\n');
}

void myWrite(S...)(File f, S args)
{
    import std.format : formattedWrite;
    import std.stdio : stdout;
    import std.typecons : tuple;
    auto w = f.lockingTextWriter();
    formattedWrite(w, "%(%s%|%)", tuple(args));
}
```

https://run.dlang.io/is/99SIV9

If the `myWrite` implementation above is correct, what do you think about replacing the current code with it? 
If the current implementation of `write(S...)(S s)` handled the common cases as an optimization, I see no reason why such optimizations can't be done instead in `formattedWrite`. CC @berni44 who recently did many improvements to `std.format`.

berni44:
> If the current implementation of `write(S...)(S s)` handled the common cases as an optimization, I see no reason why such optimizations can't be done instead in `formattedWrite`. CC @berni44 who recently did many improvements to `std.format`.

I didn't touch string handling yet. I will have to dive deeper into unicode before doing so. So I fear I cannot say anything qualified here. The only thing I know: There are some known bugs in the implementation of string handling in `formattedWrite`, but when I remember right, they do not touch this subject here.

aG0aep6G:
> If the `myWrite` implementation above is correct, what do you think about replacing the current code with it?
> If the current implementation of `write(S...)(S s)` handled the common cases as an optimization, I see no reason why such optimizations can't be done instead in `formattedWrite`. CC @berni44 who recently did many improvements to `std.format`.
> 
> Edit: One disadvantage is that the stack trace is longer:
> 
> ```
> std.utf.UTFException@/dlang/dmd/linux/bin64/../../src/phobos/std/utf.d(1715): surrogate UTF-16 high value past end of string
> ```

Note that the exception message is different. Your `myWrite` throws more eagerly. The existing `write` can handle a sequence that is split over two strings:

```d
void main()
{
    import std.stdio: stdout;
    stdout.write("\U00010400"w[0], "\U00010400"w[1], "\n"); /* prints "𐐀" */
    stdout.myWrite("\U00010400"w[0], "\U00010400"w[1], "\n"); /* throws */
}
```

PetarKirov:
> The existing write can handle a sequence that is split over two strings

IMO, this is not worth supporting, unless it was already guaranteed by the documentation and unit tests.

aG0aep6G:
> IMO, this is not worth supporting, unless it was already guaranteed by the documentation and unit tests.

That's alright, but I just want to fix a bug. I'd rather not rewrite the whole function, changing its behavior in subtle ways.

PetarKirov:
Okay, fair enough.

I restarted the druntime job on BuildKite as it was failing due to running out of memory. I also restarted FreeBSD_32 job on the auto-tester.

RazvanN7:
This seems to consistently fail druntime on buildkite. The error message:

```

_D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoThTaZQCaFNaNfKQBrhMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T19needToSwapEndianessTaZQyFNaNbNiNfMKxSQCbQCa__T10FormatSpecTaZQpZb'
--
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoThTaZQCaFNaNfKQBrhMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T14formatIntegralTSQBg5array__T8AppenderTAyaZQoTmTaZQBzFNaNfKQBrxmMKxSQDfQDe__T10FormatSpecTaZQpkmZv'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoTkTaZQCaFNaNfKQBrkMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T19needToSwapEndianessTaZQyFNaNbNiNfMKxSQCbQCa__T10FormatSpecTaZQpZb'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoTkTaZQCaFNaNfKQBrkMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T14formatIntegralTSQBg5array__T8AppenderTAyaZQoTmTaZQBzFNaNfKQBrxmMKxSQDfQDe__T10FormatSpecTaZQpkmZv'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTyAaZQoThTaZQCaFNaNfKQBrhMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T19needToSwapEndianessTaZQyFNaNbNiNfMKxSQCbQCa__T10FormatSpecTaZQpZb'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTyAaZQoTkTaZQCaFNaNfKQBrkMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T19needToSwapEndianessTaZQyFNaNbNiNfMKxSQCbQCa__T10FormatSpecTaZQpZb'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoTmTaZQCaFNaNfKQBrmMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T14formatIntegralTSQBg5array__T8AppenderTAyaZQoTmTaZQBzFNaNfKQBrxmMKxSQDfQDe__T10FormatSpecTaZQpkmZv'
  | generated/linux/release/64/benchmark.o:benchmark/runbench.d:function _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoTxhTaZQCbFNaNfKQBsxhMKxSQDhQDg__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T14formatIntegralTSQBg5array__T8AppenderTAyaZQoTmTaZQBzFNaNfKQBrxmMKxSQDfQDe__T10FormatSpecTaZQpkmZv'
  | generated/linux/release/64/unittest/test_runner core.sys.solaris.libelf
  | posix.mak:462: recipe for target 'generated/linux/release/64/benchmark' failed
  | make: *** [generated/linux/release/64/benchmark] Error 1
  | make: *** Waiting for unfinished jobs....
```

aG0aep6G:
> This seems to consistently fail druntime on buildkite. The error message:
> 
> ```
> 
> _D3std6format__T15formatValueImplTSQBh5array__T8AppenderTAyaZQoThTaZQCaFNaNfKQBrhMKxSQDfQDe__T10FormatSpecTaZQpZv: error: undefined reference to '_D3std6format__T19needToSwapEndianessTaZQyFNaNbNiNfMKxSQCbQCa__T10FormatSpecTaZQpZb'
> ```

I don't see how the changes here could cause undefined references.

berni44:
Maybe this is due to the split of `std.format`. I don't understand the details yet, but rebasing this PR to master might help. 

Anyway, it would be nice to know why this fails. Might be something like "private members of std/format/package.d are treated differently to private members of std/format.d" or something...

RazvanN7:
Yeah, rebasing might fix it.

berni44:
Strange things are happening here: I was able to reproduce the error locally by running `make -f posix.mak buildkite-test` in `druntime`. Unfortunately, now this error appears even, when I checkout `master` for `dmd`, `druntime` and `phobos`... Probably `make -f posix.mak clean` doesn't clean up enough...

aG0aep6G:
rebased

RazvanN7 (approved):

PR #7802, opened 2021-02-17
std.format: Move internal write stuff into internal submodule.

berni44:
Followup on #7798. See #7765 for the main idea.

There are more unittests scattered around the file, that should finally end up in `std.format.internal.write`. I left them where they are, for the time being, because I wanted to make this PR as simple as possible.

I didn't move `formatElement` yet. It's undocumented public and used by `typetuple`. I don't want to move public symbols, before 2.096.0 is out.


PetarKirov (approved):
Nice work!

The diff is a slightly harder to follow since it's not all contiguous, but everything seems good.

PR #7805, opened 2021-02-19
std.format: Refactor internal/read.d

berni44:
I hope, I have not been overdoing here...


std/format/internal/read.d
```diff
@@ -12,15 +12,20 @@
...
-import std.format;
+import std.traits : isAggregateType, isArray, isAssociativeArray,
+    isDynamicArray, isFloatingPoint, isIntegral, isSomeChar, isSomeString,
+    isStaticArray, StringTypeOf;
 
-package(std.format) void skipData(Range, Char)(ref Range input, scope const ref FormatSpec!Char spec)
+import std.format : FormatSpec;
+
+package(std.format) void skipData(Range, Char)(ref Range input,
+                                               scope const ref FormatSpec!Char spec)
```
pbackus, line 24: When you find yourself starting a line in column 48 (!), that's a sure sign something has gone wrong, and you need to re-think the decisions that lead you to this outcome.

If you want to break this line, I think a better place to do so is at the start of the function arguments:
```suggestion
package(std.format) void skipData(Range, Char)(
    ref Range input, scope const ref FormatSpec!Char spec)
```
Or if you prefer each argument on its own line:
```suggestion
package(std.format) void skipData(Range, Char)(
    ref Range input,
    scope const ref FormatSpec!Char spec)
```

std/format/internal/read.d
```diff
@@ -33,49 +38,58 @@ package(std.format) void skipData(Range, Char)(ref Range input, scope const ref
...
+        enum acceptedSpecs = "seEfgG";
+    else static if (isSomeChar!T)
+        enum acceptedSpecs = "bcdosuxX";    // integral + 'c'
+    else
+        enum acceptedSpecs = "";
 }
 
-package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input, scope const ref FormatSpec!Char spec)
+package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input,
+                                                        scope const ref FormatSpec!Char spec)
```
pbackus, line 58:
```suggestion
package(std.format) T unformatValueImpl(T, Range, Char)(
    ref Range input, scope const ref FormatSpec!Char spec)
```

std/format/internal/read.d
```diff
@@ -33,49 +38,58 @@ package(std.format) void skipData(Range, Char)(ref Range input, scope const ref
...
     enforceFmt(find(acceptedSpecs!long, spec.spec).length,
-            text("Wrong unformat specifier '%", spec.spec , "' for ", T.stringof));
+               text("Wrong unformat specifier '%", spec.spec , "' for ", T.stringof));
 
     return unformatValue!long(input, spec) != 0;
 }
 
-package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input, scope const ref FormatSpec!Char spec)
+package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input,
+                                                        scope const ref FormatSpec!Char spec)
```
pbackus, line 74:
```suggestion
package(std.format) T unformatValueImpl(T, Range, Char)(
    ref Range input, scope const ref FormatSpec!Char spec)
```

std/format/internal/read.d
```diff
@@ -33,49 +38,58 @@ package(std.format) void skipData(Range, Char)(ref Range input, scope const ref
...
-            text("Wrong unformat specifier '%", spec.spec , "' for ", T.stringof));
+               text("Wrong unformat specifier '%", spec.spec , "' for ", T.stringof));
 
     return parse!T(input);
 }
 
-/// ditto
-package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input, scope const ref FormatSpec!Char spec)
+package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input,
+                                                        scope const ref FormatSpec!Char spec)
```
pbackus, line 87:
```suggestion
package(std.format) T unformatValueImpl(T, Range, Char)(
    ref Range input, scope const ref FormatSpec!Char spec)
```

std/format/internal/read.d
```diff
@@ -105,13 +119,14 @@ if (isInputRange!Range && isIntegral!T && !is(T == enum) && isSomeChar!(ElementT
 
 }
 
-/// ditto
-package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input, scope const ref FormatSpec!Char spec)
+package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input,
+                                                        scope const ref FormatSpec!Char spec)
```
pbackus, line 123:
```suggestion
package(std.format) T unformatValueImpl(T, Range, Char)(
    ref Range input, scope const ref FormatSpec!Char spec)
```

std/format/internal/read.d
```diff
@@ -126,25 +141,29 @@ if (isFloatingPoint!T && !is(T == enum) && isInputRange!Range
...
-            text("Wrong unformat specifier '%", spec.spec , "' for ", T.stringof));
+               text("Wrong unformat specifier '%", spec.spec , "' for ", T.stringof));
 
     return parse!T(input);
 }
 
-/// ditto
-package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input, scope const ref FormatSpec!Char spec)
+package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input,
+                                                        scope const ref FormatSpec!Char spec)
```
pbackus, line 150:
```suggestion
package(std.format) T unformatValueImpl(T, Range, Char)(
    ref Range input, scope const ref FormatSpec!Char spec)
```

std/format/internal/read.d
```diff
@@ -154,22 +173,24 @@ if (isInputRange!Range && isSomeChar!T && !is(T == enum) && isSomeChar!(ElementT
...
     else
         static assert(false, T.stringof ~ ".sizeof must be 1, 2, or 4 not " ~
-                to!string(T.sizeof));
+                      to!string(T.sizeof));
 }
 
-/// ditto
-package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input, scope const ref FormatSpec!Char fmt)
+package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input,
+                                                        scope const ref FormatSpec!Char fmt)
```
pbackus, line 180:
```suggestion
package(std.format) T unformatValueImpl(T, Range, Char)(
    ref Range input, scope const ref FormatSpec!Char spec)
```

std/format/internal/read.d
```diff
@@ -211,39 +232,46 @@ if (isInputRange!Range && is(StringTypeOf!T) && !isAggregateType!T && !is(T == e
         return app.data;
 }
 
-/// ditto
-package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input, scope const ref FormatSpec!Char fmt)
-if (isInputRange!Range && isArray!T && !is(StringTypeOf!T) && !isAggregateType!T && !is(T == enum))
+package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input,
+                                                        scope const ref FormatSpec!Char fmt)
```
pbackus, line 236:
```suggestion
package(std.format) T unformatValueImpl(T, Range, Char)(
    ref Range input, scope const ref FormatSpec!Char spec)
```

std/format/internal/read.d
```diff
@@ -211,39 +232,46 @@ if (isInputRange!Range && is(StringTypeOf!T) && !isAggregateType!T && !is(T == e
...
-            text("Wrong unformat specifier '%", spec , "' for ", T.stringof));
+               text("Wrong unformat specifier '%", spec , "' for ", T.stringof));
 
     return parse!T(input);
 }
 
-/// ditto
-package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input, scope const ref FormatSpec!Char fmt)
+package(std.format) T unformatValueImpl(T, Range, Char)(ref Range input,
+                                                        scope const ref FormatSpec!Char fmt)
```
pbackus, line 256:
```suggestion
package(std.format) T unformatValueImpl(T, Range, Char)(
    ref Range input, scope const ref FormatSpec!Char spec)
```

std/format/internal/read.d
```diff
@@ -275,17 +305,12 @@ if (is(immutable ElementEncodingType!Range == immutable char)
     return x.typed;
 }
 
-//debug = unformatRange;
-
-private T unformatRange(T, Range, Char)(ref Range input, scope const ref FormatSpec!Char spec)
-in
+private T unformatRange(T, Range, Char)(ref Range input,
+                                        scope const ref FormatSpec!Char spec)
```
pbackus, line 309:
```suggestion
private T unformatRange(T, Range, Char)(
    ref Range input, scope const ref FormatSpec!Char spec)
```

pbackus (changes requested):

berni44:
> When you find yourself starting a line in column 48 (!), that's a sure sign something has gone wrong, and you need to re-think the decisions that lead you to this outcome.

For me (and as a side note my editor too) it's the most natural place to break the line. Anyway, breaking after the parenthesis seems to be usual in Phobos...

RazvanN7:
Refactoring PRs should come with a justification on why the refactoring is needed. I, personally, do not see any benefit in the current refactoring. The bulk of the change simply puts parameters on newlines. What is the benefit in doing so? I admit, that I am biased, and I prefer the leading parameter to be on the same line with the definition, but irrespective of that, I don't see what this change buys us.

berni44:
> Refactoring PRs should come with a justification on why the refactoring is needed. I, personally, do not see any benefit in the current refactoring. The bulk of the change simply puts parameters on newlines. What is the benefit in doing so?

I mainly followed the [style guidelines](https://dlang.org/dstyle.html) - see the [commits](https://github.com/dlang/phobos/pull/7805/commits) of this PR for the roadmap. As far as I know, `std.format` is a rather old module from a time, when the style guidelines did not exist (or at least were not written down) and therefore several of the recommendations in the guidelines still do not hold (please notice, that there are lots of exceptions concerning `std.format` in `.dscanner.ini`). In my opinion, it is good timing to use the split of a module to correct such style stuff too. It helps to find bugs and "filthy" code and makes it easier to continue with productive work.

Concerning the line breaks, that you mention: I'm not happy with the result and maybe we could find something better here. The problem is, that the `unformatValueImpl` lines (and the `formatValueImpl` in `write`) are very long, that is, they get close to 120 characters. (In `write` two of them exceed it already and I had to break them while splitting.) On the other hand, these lines are important ones - they mark the beginning of a new section of the file.

I think, it's important to make them shorter, not only because of the limit, but also because longer lines tend to be confusing, because they are outside of the editor window or in a second line. (And even if they are visible as one line in a wide editor window: Design studies tell us, that the eye has got problems to switch from line to line, when they get too long, so reading and working on files with long lines consumes more energy, even if directly unnoticed.)

Maybe we hit here a more general problem of D: With properties, template parameters, and normal parameters (with properties of their own), template declarations tend to get too long to fit in one line. Maybe we should (in general) find a D style to cope with this. Probably this discussion is beyond the scope of this PR.

std/format/internal/read.d
```diff
@@ -83,7 +83,6 @@ if (isInputRange!Range && is(T == typeof(null)))
     return parse!T(input);
 }
 
-/// ditto
```
RazvanN7, line 72: Why remove these?

berni44: Public comment for a non public function. And it does not refer to anything (ditto what?). In my opinion it's just a leftover from former days.

RazvanN7 (approved):

PR #7806, opened 2021-02-20
fix issue 15136 - If we want toStringz to be fully correct, it needs …

aG0aep6G:
…to stop checking for '\0'

### DIFF
--- a/std/string.d
+++ b/std/string.d
@@ -379,7 +379,7 @@ pure nothrow @system unittest
     p = toStringz(test2);
     assert(*p == 0);
 
-    assert(toStringz([]) is "".ptr);
+    assert(toStringz([]) is toStringz(""));
 }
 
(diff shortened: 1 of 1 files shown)


schveiguy (approved):
LGTM. One thing to think about, if there is already a null character on a string value (within the bounds), then reallocation isn't going to do anything.

PetarKirov:
Some related discussion regarding this optimization:
https://forum.dlang.org/post/ildzweoczmwvflodfadk@forum.dlang.org

std/string.d
```diff
@@ -395,6 +378,27 @@ pure nothrow @system unittest
     const string test2 = "";
     p = toStringz(test2);
     assert(*p == 0);
+
+    assert(toStringz([]) is "".ptr);
```
RazvanN7, line 382: This seems to be failing on CircleCi

RazvanN7 (approved):

PR #7807, opened 2021-02-22
Remove a few uses of fully qualified names that rely on a DMD bug

BorisCarvajal:
These fully-qualified names are not really exposed in the current scope but are accessible due to bug dlang/dmd#12215.

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -786,9 +786,9 @@ private struct MapResult(alias fun, Range)
     string  s1 = "hello world!";
     dstring s2 = "日本語";
     dstring s3 = "hello world!"d;
-    auto ms1 = map!(std.ascii.toUpper)(s1);
-    auto ms2 = map!(std.ascii.toUpper)(s2);
-    auto ms3 = map!(std.ascii.toUpper)(s3);
+    auto ms1 = map!(toUpper)(s1);
+    auto ms2 = map!(toUpper)(s2);
+    auto ms3 = map!(toUpper)(s3);
     static assert(!is(ms1[0])); //narrow strings can't be indexed
     assert(ms2[0] == '日');
     assert(ms3[0] == 'H');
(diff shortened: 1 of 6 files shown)


std/array.d
```diff
@@ -3636,13 +3636,15 @@ if (isDynamicArray!A)
...
     }
 
+    import std.format : FormatSpec;
+
     /// ditto
     template toString(Writer)
     if (isOutputRange!(Writer, char))
     {
-        import std.format : FormatSpec;
+        //import std.format : FormatSpec; this doesn't work, see issue 13737
```
BorisCarvajal, line 3645: I need some advise here.
Should I comment out the line or just delete it?

RazvanN7 (commented):
Except for the `abs` one, I don't understand why all the others are bugs. You are allowed to use the fully qualified name for an imported symbol. E.g:

```d
import a : name;
func(a.name);
```
is and should be perfectly legal.

std/algorithm/iteration.d
```diff
@@ -786,9 +786,9 @@ private struct MapResult(alias fun, Range)
     string  s1 = "hello world!";
     dstring s2 = "日本語";
     dstring s3 = "hello world!"d;
-    auto ms1 = map!(std.ascii.toUpper)(s1);
-    auto ms2 = map!(std.ascii.toUpper)(s2);
-    auto ms3 = map!(std.ascii.toUpper)(s3);
+    auto ms1 = map!(toUpper)(s1);
+    auto ms2 = map!(toUpper)(s2);
+    auto ms3 = map!(toUpper)(s3);
```
RazvanN7, line 791: `std.ascii.toUpper` is perfectly legal here.

std/algorithm/iteration.d
```diff
@@ -5459,7 +5459,7 @@ private struct SplitterResult(alias isTerminator, Range)
         ["là", "dove", "terminava", "quella", "valle"]
     ));
     assert(equal(
-        splitter!(std.uni.isWhite)("là dove terminava quella valle"),
+        splitter!(isWhite)("là dove terminava quella valle"),
```
RazvanN7, line 5462: ditto

BorisCarvajal:
> Except for the `abs` one, I don't understand why all the others are bugs. You are allowed to use the fully qualified name for an imported symbol.

Not when it's a selective `import`.
https://dlang.org/spec/module.html#selective_imports

Check yourself: https://run.dlang.io/is/Lh649f

RazvanN7:
Wow! I'm learning something new every day. This will definitely require a deprecation cycle.

std/array.d
RazvanN7: Just delete it.

RazvanN7 (approved):

RazvanN7:
Delete the commented line and this is good to go.

RazvanN7 (approved):

PR #7812, opened 2021-02-25
Make imports of std.format.internal.write local and selective.

berni44:
I kept the two imports `assertCTFEable` and `formatTest` global, because they are used a lot and I have to cope with unittests later anyway (when the non internal modules are split, which is postponed until 2.096.0 is out).

### DIFF
--- a/std/format/internal/write.d
+++ b/std/format/internal/write.d
@@ -12,12 +12,15 @@
  */
 module std.format.internal.write;
 
-import std.exception;
-import std.meta;
-import std.range.primitives;
+import std.format : FormatSpec;
+import std.range.primitives : isInputRange;
 import std.traits;
 
-import std.format;
+version (StdUnittest)
+{
+    import std.exception : assertCTFEable;
+    import std.format : formatTest, format;
+}
 
 /*
     `bool`s are formatted as `"true"` or `"false"` with `%s` and as `1` or
@@ -98,6 +101,8 @@ if (is(BooleanTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
 package(std.format) void formatValueImpl(Writer, T, Char)(auto ref Writer w, T obj, scope const ref FormatSpec!Char f)
 if (is(immutable T == immutable typeof(null)) && !is(T == enum) && !hasToString!(T, Char))
 {
+    import std.format : enforceFmt;
+
     const spec = f.spec;
     enforceFmt(spec == 's', "null literal cannot match %" ~ spec);
 
@@ -106,6 +111,10 @@ if (is(immutable T == immutable typeof(null)) && !is(T == enum) && !hasToString!
 
 @safe pure unittest
 {
+    import std.exception : collectExceptionMsg;
+    import std.format : FormatException;
+    import std.range.primitives : back;
+
     assert(collectExceptionMsg!FormatException(format("%p", null)).back == 'p');
 
     assertCTFEable!(
@@ -126,6 +135,8 @@ if (is(immutable T == immutable typeof(null)) && !is(T == enum) && !hasToString!
 package(std.format) void formatValueImpl(Writer, T, Char)(auto ref Writer w, T obj, scope const ref FormatSpec!Char f)
 if (is(IntegralTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
 {
+    import std.range.primitives : put;
+
     alias U = IntegralTypeOf!T;
     U val = obj;    // Extracting alias this may be impure/system/may-throw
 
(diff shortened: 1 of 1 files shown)


RazvanN7 (commented):
I'm not convinced that this buys us anything. I mean, yeah, selective imports are nice, but it really doesn't make any sense to do selective imports when you use a lot of symbols from that package (see the case of `std.traits` in this PR). Also, you have littered the whole file with imports of `std.format : format`. I don't have a clear metric for this, but adding 185 lines of imports does not seem like an improvement.

berni44:
> I'm not convinced that this buys us anything. I mean, yeah, selective imports are nice, but it really doesn't make any sense to do selective imports when you use a lot of symbols from that package (see the case of `std.traits` in this PR). Also, you have littered the whole file with imports of `std.format : format`. I don't have a clear metric for this, but adding 185 lines of imports does not seem like an improvement.

Well, first of all the [style guidelines](https://dlang.org/dstyle.html) ask us to do so. I'm not completely aware of all the consequences, though. I think the main reason is to reduce the pollution of namespace. I also know, that it makes reducing Phobos with dustmite faster. In one case, the size of the executables was reduced. There might be more reasons.

Concerning `std.format : format`: This is only used in unittests. I could put the import in the version block at the beginning of the file to avoid the large number of import statements. Anyway, I think, they point us to a design flaw: These unittests do not directly test the function they test. Instead, the tests are done indirectly using `format` (or `formatTest`, `formatElement` or `formattedWrite`). If one of these functions is going to be changed in the future, they might not anymore test, what they pretend to test. (I feel somewhat guilty here, because several of these unittests where introduced by myself...)

So I suggest to replace the unittests by versions, that directly test the functions they test. (Maybe with the help of something similar to `formatTest`, to avoid too much boilerplate code.)

Concerning `std.traits`: I don't know what is best here. The guidelines suggest to make them selective, hence I did so. But having said this, I personally don't care much here. If you think I should keep `std.traits` as a non-selective import, I'll do so.

RazvanN7:
> Well, first of all the style guidelines ask us to do so. I'm not completely aware of all the consequences, though. I think the main reason is to reduce the pollution of namespace. I also know, that it makes reducing Phobos with dustmite faster. In one case, the size of the executables was reduced. There might be more reasons.

I don't think that there is any evidence that using selective imports has a better performance then using global imports and the pollution is not an actual problem as dmd is going to prefer local symbols to imported one. Selective imports help with modularity (once a function or unittest is deleted, you don't have to worry about unused global imports). However, I think that common sense should be used when tackling this: if you use a lot of symbols from one imported module, it is best to simply import the entire module (the case of std.traits).

> Concerning std.format : format: This is only used in unittests. I could put the import in the version block at the beginning of the file to avoid the large number of import statements. Anyway, I think, they point us to a design flaw: These unittests do not directly test the function they test. Instead, the tests are done indirectly using format (or formatTest, formatElement or formattedWrite). If one of these functions is going to be changed in the future, they might not anymore test, what they pretend to test. (I feel somewhat guilty here, because several of these unittests where introduced by myself...)

That is fine. There is no such thing as overtesting. If `format` is going to get changed then there will be a ton of unittests that are going to fail pinpointing that the change might negatively impact user code. It's not a design flaw, it is actually a good thing if the uses of `std.format` are valid use cases.

> So I suggest to replace the unittests by versions, that directly test the functions they test. (Maybe with the help of something similar to formatTest, to avoid too much boilerplate code.)

I think that this is overkill.

> Concerning std.traits: I don't know what is best here. The guidelines suggest to make them selective, hence I did so. But having said this, I personally don't care much here. If you think I should keep std.traits as a non-selective import, I'll do so.

I think that leaving it as a non-selective import is fine.

berni44:
> However, I think that common sense should be used when tackling this: if you use a lot of symbols from one imported module, it is best to simply import the entire module (the case of std.traits).

:+1: - changed this.

berni44:
> That is fine. There is no such thing as overtesting. If `format` is going to get changed then there will be a ton of unittests that are going to fail pinpointing that the change might negatively impact user code. It's not a design flaw, it is actually a good thing if the uses of `std.format` are valid use cases.

This still leaves the question, what to do with the unittests: Essentially we are testing the same things at different places: `format`, `sformat`, `formattedWrite`, `formatValueImpl`, `formatValue`, and `formatElement` have a lot in common and the tests at one place actually test another of these functions and vice versa. Currently lots of unittests are scattered around in `std/format/package.d`.

I'd prefer to have a more clear cut idea about the responsibility of the various unittests and would like to sort them. But this is obviously not part of this PR. I'll come back on this, when the public parts of `std.format` are split out.

std/format/internal/write.d
```diff
@@ -569,8 +592,8 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     {
         static if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.mant_dig))
         {
-            import std.math;
-            import std.format.internal.floats : RoundingMode, printFloat;
+            import std.format.internal.floats : printFloat, RoundingMode;
+            import std.math; // cannot be selectiv, because FloatingPointControl might not be defined
```
RazvanN7, line 596:
```suggestion
            import std.math; // cannot be selective, because FloatingPointControl might not be defined
```
This goes for the rest of the comments also.

RazvanN7 (approved):

PR #7813, opened 2021-02-25
Fix issue 21663 - std.concurrency.receiveOnly doesn't work with tuples

pbackus:


pbackus:
CI is failing because this targets `stable` and therefore doesn't have the changes from #7744.

RazvanN7 (approved):

RazvanN7:
This is weird. The latest version of stable does not include this test. It seems like the CI is using a prior version of stable :-?

pbackus:
Looks like I just missed the latest update to stable (2a2a0b4e0). I've rebased again, which should fix it.

PR #7815, opened 2021-03-01
std.format: Add unittests for %g (step 2).

berni44:
Followup on #7810.


thewilsonator:
Why are so many tests comments out?

berni44:
> Why are so many tests comments out?

They don't work yet. %g isn't just selecting between %e and %f. There are other changes needed too: Mainly, precision has a different meaning and trailing zeros are removed. So the unittests are the roadmap to follow. The idea is to make them work step by step.

RazvanN7 (approved):

PR #7816, opened 2021-03-01
std.format: Add %g (step 3): Add template parameter to printFloatE/F.

berni44:
Another followup on #7810. Can be merged independently of #7815.

Precision is already adjusted in `printFloatG` so it is not necessary to repeat this in `printFloatE/F`.


RazvanN7:
I think it would be useful to include in the description of these PRs the initial (closed) PR so that other reviewers have a change to understand the broader picture. I will merge this for now, but anyone else reviewing this will not understand what the point is.

RazvanN7 (approved):

berni44:
> I think it would be useful to include in the description of these PRs the initial (closed) PR so that other reviewers have a change to understand the broader picture. I will merge this for now, but anyone else reviewing this will not understand what the point is.

Added the link to the initial comment. Thanks for pointing this out.

PR #7817, opened 2021-03-01
[stable] Fix little std.math regressions

kinke:


RazvanN7:
Are these regressions filed anywhere?

kinke:
Nope, and doing so would be utter nonsense IMO. The `version` things don't apply to DMD (because it's always using x87 `real`s) and the unittest regression for 64-bit `real` is also another frequent one not observable with DMD.

thewilsonator (approved):

PR #7819, opened 2021-03-01
Fix Issue 21606 - pow(NaN,0) gives 1 not NaN

berni44:
Detecting NaNs early speeds calculation up considerably when NaNs are involved (and removes a bug).


berni44:
Ooops. I just now realized, that the `pow(float, float)` explicitly makes an exception for `NaN^^0`. Internet research brought up this: https://stackoverflow.com/questions/17863619/why-does-nan0-1

Seems to be a weird convention...

PR #7822, opened 2021-03-01
improve documentation of std.algorithm.sorting

dkorpel:
I once was looking whether Phobos had Quickselect, and it took me really long before I found it's actually called `topN` there.
I think it's useful to mention this in the documentation. 

Some other things:
- topN is missing a Returns: section
- most functions have no blank line separating the summary and description paragraphs
- the docs often mention `swap` referring to `std.algorithm.mutation: swap`, but that is not obvious.

### DIFF
--- a/std/algorithm/sorting.d
+++ b/std/algorithm/sorting.d
@@ -57,7 +57,7 @@ $(T2 schwartzSort,
 $(T2 sort,
         Sorts.)
 $(T2 topN,
-        Separates the top elements in a range.)
+        Separates the top elements in a range, akin to $(LINK2 https://en.wikipedia.org/wiki/Quickselect, Quickselect).)
 $(T2 topNCopy,
         Copies out the top elements of a range.)
 $(T2 topNIndex,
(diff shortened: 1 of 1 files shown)


std/algorithm/sorting.d
```diff
@@ -294,7 +296,9 @@ if (isForwardRange!Range)
 
 /**
 Like `isSorted`, returns `true` if the given `values` are ordered
-according to the comparison operation `less`. Unlike `isSorted`, takes values
+according to the comparison operation `less`.
+
+Unlike `isSorted`, takes values
 directly instead of structured in a range.
```
n8sh, line 302: This is the one change I don't consider an improvement. The rest of the changes look good to me.

RazvanN7 (approved):

std/algorithm/sorting.d
RazvanN7: Yes, I agree that it is fine to have both on the same paragraph.

dkorpel: I changed it back

PR #7823, opened 2021-03-01
Fix Issue 16200 - Faster pow implementation for integral exponents

berni44:
A few weeks ago I checked the alternative algorithms for `pow` given in issue 16200. Although not the way @andralex suggests there (see attachment of the issue for why I did not implement the suggested algorithm and more details), this finally lead to some other ideas on how to improve speed for this function. The main idea is to bail out early, when the result is a fixpoint (NaN, +/-inf and 0.0).

IMHO, the result speaks for itself (averaged over random bit patterns for the floating point value and iterating the exponent from -500 to 500 on x-axis, run on 2 billion tests for each FP type):

image

For "normal" operations, that is calls, where the result is not one of the fix points, there is no theoretical advantage in this algorithm. But luckily, there seems to be no disadvantage in practice. Seems, we get the speedup here (almost) for free due to some pipelining techniques or similar in the processor.

Averages on my computer in 1/100 ns:

Type | Result | current | proposed
-|-|-|-
float | NaN | 168,000 | 1
|| +/-inf | 16,253 | 140
|| 0.0 | 11,555 | 160
|| other | 668 | 668
|| avg | 14,302 | 158
double | NaN | 166,192 | 3
|| +/-inf | 52,007 | 18
|| 0.0 | 14,980 | 21
|| other | 194 | 195
|| avg | 33,069 | 22
real | NaN | 162,821 | 4,990
|| +/-inf | 124,132 | 8
|| 0.0 | 12.271 | 8
|| other | 21 | 24
|| avg | 67,210 | 8

A strange casual result is the long duration `isNaN` takes for reals...

Proof of the two conclusions used in the code:

```
   m * floor(log2(abs(v))) >= F.max_exp  
=>        m * log2(abs(v)) >= F.max_exp
=>         log2(abs(v)^^m) >= F.max_exp = log2(F.max+1)
=>               abs(v)^^m >= F.max+1
=>               abs(v)^^m >  F.max
```

For the second one we have to note, that `nextUp(0.0) == 2 ^^ (-bias - F.mant_dig + 2)` and `floor(log2(abs(v))) == ex - bias`.

```
              m * (bias - ex - 1) >= bias + F.mant_dig - 1
=>            m * (bias - ex - 1) >= bias + F.mant_dig - 1
=>           -m * (ex - bias + 1) >= bias + F.mant_dig - 1
=> -m * (floor(log2(abs(v))) + 1) >= bias + F.mant_dig - 1
=>        -m * ceil(log2(abs(v))) >= bias + F.mant_dig - 1          // note: log2(abs(v)) < 0
=>              -m * log2(abs(v)) >= bias + F.mant_dig - 1
=>              -m * log2(abs(v)) >  bias + F.mant_dig - 2
=>               m * log2(abs(v)) <  -(bias + F.mant_dig - 2)
=>                log2(abs(v)^^m) <  -(bias + F.mant_dig - 2)
=>                      abs(v)^^m <  2^^(-bias - F.mant_dig + 2)
```

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -7449,6 +7449,9 @@ real fma(real x, real y, real z) @safe pure nothrow @nogc { return (x * y) + z; 
 Unqual!F pow(F, G)(F x, G n) @nogc @trusted pure nothrow
 if (isFloatingPoint!(F) && isIntegral!(G))
 {
+    // NaN ^^ 0 is an exception defined by IEEE (yields 1 instead of NaN)
+    if (isNaN(x)) return n ? x : 1.0;
+
     import std.traits : Unsigned;
     real p = 1.0, v = void;
     Unsigned!(Unqual!G) m = n;
@@ -7468,13 +7471,76 @@ if (isFloatingPoint!(F) && isIntegral!(G))
             return 1.0;
         case 1:
             return x;
-        case 2:
-            return x * x;
         default:
         }
 
         v = x;
     }
+
+    // Bail out early, if we can estimate that the result is infinity or 0.0:
+    //
+    // We use the following two conclusions:
+    //
+    //    m * floor(log2(abs(v))) >= F.max_exp
+    // =>             abs(v) ^^ m >  F.max == nextDown(F.infinity)
+    //
+    //    m * (bias - ex - 1) >= bias + F.mant_dig - 1
+    // =>         abs(v) ^^ m <  2 ^^ (-bias - F.mant_dig + 2) == nextUp(0.0)
+    //
+    // floor(log2(abs(v))) == ex - bias can be directly taken from the
+    // exponent of the floating point represantation, to avoid long
+    // calculations here.
+
+    enum uint bias = F.max_exp - 1;
+
+    static if (is(F == float))
+    {
+        float f = cast(float) v;
+        uint ival = () @trusted { return *cast(uint*) &f; }();
+        ulong ex = (ival >> 23) & 255;
+    }
+    else static if (is(F == double) || (is(T == real) && T.mant_dig == double.mant_dig))
+    {
+        double d = cast(double) v;
+        ulong ival = () @trusted { return *cast(ulong*) &d; }();
+        ulong ex = (ival >> 52) & 2047;
+    }
+    else static if (is (F == real) && real.mant_dig == 64)
+    {
+        ulong ex = void;
+        if (__ctfe)
+        {
+            // in CTFE we cannot access the bit patterns and have therefore to
+            // fall back to the (slower) general case
+            // skipping subnormals by setting ex = bias
+            ex = abs(v) == F.infinity ? 2 * bias + 1 :
+                (abs(v) < F.min_normal ? bias : cast(ulong) (floor(log2(abs(v))) + bias));
+        }
+        else
+        {
+            ulong[2] ival = () @trusted { return *cast(ulong[2]*) &v; }();
+            ex = ival[1] & 32767;
+        }
+    }
+    else
+    {
+        // ToDo: Add special treatment for other reals too.
+
+        // In the general case we have to fall back to log2, which is slower, but still
+        // a certain speed gain compared to not bailing out early.
+            // skipping subnormals by setting ex = bias
+        ulong ex = abs(v) == F.infinity ? 2 * bias + 1 :
+            (abs(v) < F.min_normal ? bias : cast(ulong) (floor(log2(abs(v))) + bias));
+    }
+
+    // m * (...) can exceed ulong.max, we therefore first check m >= (...).
+    // This is sufficient to know that the result will be infinity or 0.0
+    // and at the same time it guards against an overflow.
+    if (ex > bias && (m >= F.max_exp || m * (ex - bias) >= F.max_exp))
+        return (m % 2 == 0 || v > 0) ? F.infinity : -F.infinity;
+    else if (ex < bias - 1
+             && (m >= bias + F.mant_dig - 1 || m * (bias - ex - 1) >= bias + F.mant_dig - 1))
+        return 0.0;
 
     while (1)
     {


std/math.d
```diff
@@ -7449,6 +7449,9 @@ real fma(real x, real y, real z) @safe pure nothrow @nogc { return (x * y) + z;
 Unqual!F pow(F, G)(F x, G n) @nogc @trusted pure nothrow
 if (isFloatingPoint!(F) && isIntegral!(G))
 {
+    // NaN ^^ 0 is an exception defined by IEEE
```
andralex, line 7452:
```suggestion
    // NaN ^^ 0 is an exception defined by IEEE (yields 1 instead of NaN)
```

std/math.d
```diff
@@ -7449,6 +7449,9 @@ real fma(real x, real y, real z) @safe pure nothrow @nogc { return (x * y) + z;
...
+    if (isNaN(x) && n != 0) return x;
```
andralex, line 7453:
```suggestion
    if (isNaN(x)) return n ? x : 1.0;
```

should help with the slowness of nan

std/math.d
```diff
@@ -7476,6 +7479,69 @@ if (isFloatingPoint!(F) && isIntegral!(G))
...
+    // =>             abs(v) ^^ m >  F.max == nextDown(F.infinity)
+    //
+    //    m * (bias - ex - 1) >= bias + F.mant_dig - 1
+    // =>         abs(v) ^^ m <  2 ^^ (-bias - F.mant_dig + 2) == nextUp(0.0)
+    //
+    // floor(log2(abs(v))) == ex - bias can be directly taken from the
+    // exponent of the floating point represantation, to avoid long
+    // calculations here.
+
+    enum int bias = F.max_exp - 1;
```
andralex, line 7496:
```suggestion
    enum uint bias = F.max_exp - 1;
```

(all calculations involving bias are in unsigned IIUC)

std/math.d
```diff
@@ -7476,6 +7479,69 @@ if (isFloatingPoint!(F) && isIntegral!(G))
...
+    {
+        // ToDo: Add special treatment for other reals too.
+
+        // In the general case we have to fall back to log2, which is slower, but still
+        // a certain speed gain compared to not bailing out early.
+        ulong ex = abs(v) == F.infinity ? 2 * bias + 1 :
+            (abs(v) < F.epsilon ? bias : cast(ulong) (floor(log2(abs(v))) + bias));
+    }
+
+    // m * (...) can excede ulong.max, we therefore first check m >= (...).
```
andralex, line 7536:
```suggestion
    // m * (...) can exceed ulong.max, we therefore first check m >= (...).
```

std/math.d
```diff
@@ -7476,6 +7479,69 @@ if (isFloatingPoint!(F) && isIntegral!(G))
...
+    // This is sufficient to know, that the result will be infinity / 0.0
```
andralex, line 7537:
```suggestion
    // This is sufficient to know that the result will be infinity or 0.0
```

std/math.d
```diff
@@ -7476,6 +7479,69 @@ if (isFloatingPoint!(F) && isIntegral!(G))
         v = x;
     }
 
+    // Bail out early, if we can estimate, that the result is infinity or 0.0:
```
andralex, line 7482:
```suggestion
    // Bail out early, if we can estimate that the result is infinity or 0.0:
```

andralex (approved):
cool beans!

maxhaton:
What compiler and flags was used to generate this data?

berni44:
> What compiler and flags was used to generate this data?

ldc 1.25.0 with -O3 -release -boundscheck=off -flto=full -defaultlib=phobos2-ldc-lto,druntime-ldc-lto -mcpu=native on a Intel(R) Core(TM) i3-7100T CPU @ 3.40GHz

std/math.d
berni44: At least on my box this had no effect...

berni44:
Added all requested changed of @andralex 

In addition to that, I removed the special case of exponent 2. Yesterday I was wondering about the little dark blue peek next to 0 in the diagram and tonight I realized it's caused by this special case: inf*inf and denormalized*denormalized are very slow compared to normal multiplications...

RazvanN7 (approved):

thewilsonator (approved):

berni44:
Just for reference, the same table with gdc-4.8.5 -O3 -fno-bounds-check -frelease:

Type | Result | current | proposed
-|-|-|-
float | NaN | 135,427 | 1
|| +/-inf | 15,818 | 137
|| 0.0 | 10,773 | 157
|| other | 668 | 700
|| avg | 13,578 | 157
double | NaN | 134,838 | 0
|| +/-inf | 42,263 | 18
|| 0.0 | 12,446 | 22
|| other | 194 | 216
|| avg | 27,008 | 23
real | NaN | 130,019 | 0
|| +/-inf | 97,052 | 3
|| 0.0 | 11,172 | 3
|| other | 22 | 179
|| avg | 53,326 | 5

Here is a clear loss visible for the non fix point tasks. But the peculiarity of `isNaN` with reals vanished.

berni44:
Shouldn't the bot remove the auto-merge label whenever I push something? I found a bug in this code (I confused real.epsilon with real.min_normal). Probably, the code works anyway, but I wanted to have a closer look before this is merged and I'm too tired now, to work this out. Will do this tomorrow...

berni44:
Due to technical problems I had no internet for several days, but I used the time to get this PR right: First of all, the former version would have worked too, but would have been slower for CTFE and big reals.

Second, I added some unittests. Obviously they work with the current version too, but I missed them anyway: Checks with results close to infinity and close to 0.0, once at runtime and once during CTFE.

While I created this unittests, I noticed a [CTFE bug](https://issues.dlang.org/show_bug.cgi?id=21688). Hence some of the tests are commented out.

berni44:
Removed the unittests. I'll add them in a separate PR...

ibuclaw:
> Just for reference, the same table with gdc-4.8.5 -O3 -fno-bounds-check -frelease:

I am amazed anyone is using a version older than 9.x 😱

berni44:
> > Just for reference, the same table with gdc-4.8.5 -O3 -fno-bounds-check -frelease:
> 
> I am amazed anyone is using a version older than 9.x

@ibuclaw Funnily I didn't realize, that this one is older, than the one installed from debian stable (8.3.0). I stuggled a lot to get the latest version installed on my box, but didn't manage to do so. Then I found out, that the dlang install script also features gdb and used it to install it, because I thought, that this should be the latest version. I didn't check though and I was not aware of which numbers are the current numbers...

ibuclaw:
> > > Just for reference, the same table with gdc-4.8.5 -O3 -fno-bounds-check -frelease:
> > 
> > I am amazed anyone is using a version older than 9.x
> 
> @ibuclaw Funnily I didn't realize, that this one is older, than the one installed from debian stable (8.3.0). I stuggled a lot to get the latest version installed on my box, but didn't manage to do so. Then I found out, that the dlang install script also features gdb and used it to install it, because I thought, that this should be the latest version. I didn't check though and I was not aware of which numbers are the current numbers...
> 

Yeah, that pulls the binary off my server. It's a $new_project to start making new download binaries again, preferably with some cadence.

std/math.d
```diff
@@ -7468,14 +7471,77 @@ if (isFloatingPoint!(F) && isIntegral!(G))
...
+
+    enum uint bias = F.max_exp - 1;
+
+    static if (is(F == float))
+    {
+        float f = cast(float) v;
+        uint ival = () @trusted { return *cast(uint*) &f; }();
+        ulong ex = (ival >> 23) & 255;
+    }
+    else static if (is(F == double) || (is(T == real) && T.mant_dig == double.mant_dig))
```
kinke, line 7502: There's no `T` here. Even when changing that to `F`, I'm seeing a unittest regression with LDC on Windows, with 64-bit `real`.

berni44: IMHO it wouldn't have done any harm. These `static ifs` only provide shortcut for cases, where you do know ahead the result. If they aren't compiled in, the slower calculation at the bottom of the function is used. With the typo this essentially is `else static if (is(F == double) || false)`. And that's exactly the fix I suggest: Just use `else static if (is(F == double)`, which just means: Don't use the speed up for those 64bit-reals.

Due to personal reasons I'll not going to write this PR. If anyone feels inclined, please go on.

PR #7825, opened 2021-03-02
Throughout Phobos use core.math intrinsics instead of std.math wrappers

n8sh:
Followup to PR #7821.

### DIFF
--- a/std/bigint.d
+++ b/std/bigint.d
@@ -706,11 +706,8 @@ public:
     }
 
     /// ditto
-    bool opEquals(T)(const T y) const nothrow @nogc if (isFloatingPoint!T)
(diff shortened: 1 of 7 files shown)


std/bigint.d
```diff
@@ -706,11 +706,8 @@ public:
     }
 
     /// ditto
-    bool opEquals(T)(const T y) const nothrow @nogc if (isFloatingPoint!T)
+    bool opEquals(T)(const T y) const pure nothrow @nogc if (isFloatingPoint!T)
     {
-        // This is a separate function from the isIntegral!T case
-        // due to the impurity of std.math.scalbn which is used
-        // for 80 bit floats.
```
n8sh, line 713: `std.math.scalbn` became pure a while ago but this comment was not updated.

thewilsonator (approved):

PR #7831, opened 2021-03-08
std.format: Add %g (step 6): %f like formatting

berni44:
Most of this PR is about getting the different meaning of `precision` right. Only at the end (of `printFloatF`), there is a part about removing trailing zeros.

### DIFF
--- a/std/format/internal/floats.d
+++ b/std/format/internal/floats.d
@@ -1429,6 +1429,7 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
 {
     import std.conv : to;
     import std.algorithm.comparison : max;
+    import std.math : log10, abs, floor, ceil;
 
     enum int bias = T.max_exp - 1;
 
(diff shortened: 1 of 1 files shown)


std/format/internal/floats.d
```diff
@@ -1454,19 +1455,31 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
     // this estimate
 
     // Default for the right side is the number of digits given by f.precision plus one for the dot.
-    auto max_right = f.precision + 1;
+    static if (g)
+        auto max_right = max(0, f.precision - cast(int) floor(log10(abs(val))));
```
RazvanN7, line 1459: UFCS can be used here to make it more readable: `val.abs.log10.floor`

std/format/internal/floats.d
```diff
@@ -1454,19 +1455,31 @@ if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.
...
+        auto max_right = f.precision + 1;
 
     // If the exponent is <= 0 there is only the sign and one digit left of the dot, else
     // we have to estimate the number of digits. The factor between exp, which is the number of
     // digits in binary system and the searched number is log_2(10). We round this down to 3.32 to
     // get a conservative estimate. We need to add 3, because of the sign, the fact, that the
     // logarithm is one to small and because we need to round up instead of down, which to!int does.
-    auto max_left = exp > 0 ? to!int(exp / 3.32) + 3 : 2;
+    static if (g)
+        auto max_left = max(2, 2 + cast(int) ceil(log10(abs(val))));
```
RazvanN7, line 1469: ditto

RazvanN7 (approved):

PR #7833, opened 2021-03-08
Miscellaenous improvements to std.array unit tests

dukc:


std/array.d
```diff
@@ -247,12 +245,20 @@ version (StdUnittest)
     assert(nAlive == 3);
 
     // No good way to ensure the GC frees this, just call the lifetime function
-    // directly. If delete wasn't deprecated, this is what delete would do.
-    _d_delarray_t(cast(void[]*)&arr, typeid(S));
+    // directly.
+    __delete(arr);
```
atilaneves, line 249: Why not `destroy`?

std/array.d
```diff
@@ -421,14 +427,6 @@ if (isAutodecodableString!String)
     });
 }
 
-@safe unittest
```
atilaneves, line 424: Why was this moved?

RazvanN7 (approved):

std/array.d
dukc: It was under wrong public function

std/array.d
dukc: It failed the test when I tried. I think `destory` finalizes only the array itself, not it's contents.

RazvanN7:
@dukc The documentation tester seems to have a spurious failure. Please push again your branch to restart it.

std/array.d
n8sh: [Yeah it falls into this case:](https://github.com/dlang/druntime/blob/8d80dec1f6fc2a69a3f26fdfbc704079f8a0d661/src/object.d#L4298-L4304)
```d
/// ditto
void destroy(bool initialize = true, T)(ref T obj)
    if (!is(T == struct) && !is(T == interface) && !is(T == class) && !__traits(isStaticArray, T))
{
    static if (initialize)
        obj = T.init;
}
```

atilaneves (approved):

PR #7834, opened 2021-03-08
Typified an always throwing method as `noreturn`

dukc:


dukc:
autotester errors: `std.exception.bailOut!(OutOfMemoryError).bailOut` without side effects discards return value of type `noreturn`; prepend a `cast(void)` if intentional.

Isn't discarding `noreturn` supposed to be ok?

RazvanN7:
This looks like a bug to me, however, the implementation of `noreturn` is incomplete at the moment so I would wait before actually using it. cc @WalterBright

n8sh:
https://github.com/dlang/dmd/pull/12430 fixes this.

RazvanN7:
dlang/dmd#12430 was merged, however this is still failing.

dukc:
Error message is still the same as last time. Perhaps the test just have not been redone since the DMD pr. I try an empty commit,

dukc:
Still failing, but seems to be something unrelated this time.

dukc:
Rebased. Let's see what happens this time.

thewilsonator (approved):

Geod24:
This pull request introduced a (minor) regression:
https://issues.dlang.org/show_bug.cgi?id=22235

dukc:
Aww, I wonder if I should wait a few releases before applying a new feature in Phobos next time? It could have been something more serious.

PR #7837, opened 2021-03-10
Add extreme unittests for pow()

berni44:
This are the unittests promised in #7823. I had to loosen the constraints somewhat. This is due to the way D handles floating point numbers (using real with unknown size internally), especially in CTFE, where there are `floats`, that are larger then `float.max` but not `float.infinity`...

### DIFF
--- a/std/math.d
+++ b/std/math.d
@@ -7616,6 +7616,66 @@ if (isFloatingPoint!(F) && isIntegral!(G))
     }
 }
 
+@safe @nogc nothrow unittest
+{
+    static float f1 = 19100.0f;
+    static float f2 = 0.000012f;
+
+    assert(isClose(pow(f1,9), 3.3829868e+38f));
+    assert(isInfinity(pow(f1,10)));
+    assert(pow(f2,9) > 0.0f);
+    assert(isClose(pow(f2,10), 0.0f, 0.0, float.min_normal));
+
+    static double d1 = 21800.0;
+    static double d2 = 0.000012;
+
+    assert(isClose(pow(d1,71), 1.0725339442974e+308));
+    assert(isInfinity(pow(d1,72)));
+    assert(pow(d2,65) > 0.0f);
+    assert(isClose(pow(d2,66), 0.0, 0.0, double.min_normal));
+
+    static if (real.mant_dig == 64) // x87
+    {
+        static real r1 = 21950.0L;
+        static real r2 = 0.000011L;
+
+        assert(isClose(pow(r1,1136), 7.4066175654969242752260330529e+4931L));
+        assert(isInfinity(pow(r1,1137)));
+        assert(pow(r2,998) > 0.0L);
+        assert(isClose(pow(r2,999), 0.0L, 0.0, real.min_normal));
+    }
+}
+
+@safe @nogc nothrow pure unittest
+{
+    enum f1 = 19100.0f;
+    enum f2 = 0.000012f;
+
+    static assert(isClose(pow(f1,9), 3.3829868e+38f));
+    static assert(pow(f1,10) > float.max);
+    static assert(pow(f2,9) > 0.0f);
+    static assert(isClose(pow(f2,10), 0.0f, 0.0, float.min_normal));
+
+    enum d1 = 21800.0;
+    enum d2 = 0.000012;
+
+    static assert(isClose(pow(d1,71), 1.0725339442974e+308));
+    static assert(pow(d1,72) > double.max);
+    static assert(pow(d2,65) > 0.0f);
+    static assert(isClose(pow(d2,66), 0.0, 0.0, double.min_normal));
+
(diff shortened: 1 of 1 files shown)


std/math.d
```diff
@@ -7616,6 +7616,71 @@ if (isFloatingPoint!(F) && isIntegral!(G))
     }
 }
 
+@safe @nogc nothrow pure unittest
+{
+    float f1 = 19100.0f;
+    float f2 = 0.000012f;
```
ibuclaw, line 7622: This could still be evaluated at compile time so best make it static or __gshared.

berni44: OK, let's see, if that works...

berni44:
The switch to `static` doesn't change anything (at least, when it comes to the question of failure). It's the same test, that fails on the same two computers (Windows_VisualD x86-mscoff and FreeBSD_32):

```
static double d1 = 21800.0;
assert(pow(d1,72) == double.infinity);
```

What probably happens here is, that this code is transformed under the hood into

```
static double d1 = 21800.0;
assert(pow(cast(real) d1,72) == cast(real) double.infinity);
```

which makes the comparison happen with `reals` instead of `doubles`. The left side is (~ 2.33812e+312), which is larger than `double.max` (~ 1.79769e+308). While on the right side `double.infinity` is cast to `real.infinity`.

Strange as is, the similar test with `floats` passes...

How to proceed here? Remove that test and add a note, that the implementation of `doubles` is buggy on some computers? Or testing '>double.max', what I used for the CTFE-tests, and in my opinion is a little bit awkward? Or anything else?


Not related, but also strange: I forgot to remove the `pure` keyword, when switching to `static`, which made most of the computers fail, as expected, but on Windows x86-OMF it succeeded. I don't know exactly, what this computer is testing, but I guess, it should have failed too...

std/math.d
ibuclaw: It wasn't a suggestion to make something work, it was a suggestion to defeat gdc/ldc's ability to constfold the entire test away. :-)

berni44: Yeah, of course. But it might also have been the reason for the breaking. (But wasn't obviously.)

ibuclaw:
> Not related, but also strange: I forgot to remove the pure keyword, when switching to static, which made most of the computers fail, as expected, but on Windows x86-OMF it succeeded. I don't know exactly, what this computer is testing, but I guess, it should have failed too...

Some tests only build dmd/phobos and skip running the testsuite due to platforms being too old/unstable.

ibuclaw:
> which makes the comparison happen with reals instead of doubles. The left side is (~ 2.33812e+312), which is larger than double.max (~ 1.79769e+308). While on the right side double.infinity is cast to real.infinity.

My guess is that there's two reasons:

Firstly, the algorithm is using `real` as the intermediate store for the result (kept in `p`).
https://github.com/dlang/phobos/blob/22a1e42cb06d5a70df3862977239cb76dedcb839/std/math.d#L7449-L7457
...
https://github.com/dlang/phobos/blob/22a1e42cb06d5a70df3862977239cb76dedcb839/std/math.d#L7554

Secondly, floats on x86 are returned in x87 registers without any rounding taking place.

https://dlang.org/spec/abi.html#return_value

Which is related to this issue https://issues.dlang.org/show_bug.cgi?id=21376#c3

berni44:
> Secondly, floats on x86 are returned in x87 registers without any rounding taking place.

OK. But then, the question remains, what exactly is compared by `==`? Two `reals`, or two `doubles`? I guess reals. And in case of reals, what real is used for `double.infinity`?

ibuclaw:
> > Secondly, floats on x86 are returned in x87 registers without any rounding taking place.
> 
> OK. But then, the question remains, what exactly is compared by `==`? Two `reals`, or two `doubles`? I guess reals. And in case of reals, what real is used for `double.infinity`?
> 

It may be x87 registers where the comparison happens, yeah.

berni44:
> It may be x87 registers where the comparison happens, yeah.

And that causes in my opinion the problem: Comparing at higher precision leads to confusing results, like

```
static if (compare with x87)
    assert(double.max + 1 < double.infinity);
else
    assert(double.max + 1 == double.infinity);
```

or

```
static if (compare with x87)
    assert(nextUp(0.0) / 2 > 0.0);
else
    assert(nextUp(0.0) / 2 == 0.0);
```

Both, `double.infinity` and 0.0 are attracting fixpoints for `pow` and therefore they should be safely usable with `==`, whenever we make sure, that the real-life math results are close enough to them. Comparing in x87 registers breaks this assumption.

So I think, it would be a good idea, if D could define, that comparisons of floating points are always done with the precision of the operands (the precision of the larger, if they differ).

berni44:
I just learned about `isInfinity`. Let's see, if that works...

RazvanN7 (approved):

PR #7839, opened 2021-03-11
std.format: Split off std/format/spec.d

berni44:
As far as I understand the release process, it's meanwhile safe (concerning the release of 2.096.0) to start splitting off the public modules.

See #7765 for the whole picture.


changelog/splitting_format.dd
```diff
@@ -0,0 +1,15 @@
+Module `std.format` splitted into smaller modules
+
+The module `std.format` has been splitted into smaller modules:
+
+$(UL
+    $(LI `std.format.spec`: Stuff concerning the format string, mainly
+         the `struct FormatSpec` and the `template singleSpec`)
+    $(LI `std.format.internals`: Several modules containing non-public
+         stuff)
```
thewilsonator, line 9: `submodules containing private symbols`, although I'm not sure theres much need to draw attention to internal submodules.

changelog/splitting_format.dd
```diff
@@ -0,0 +1,15 @@
...
+)
+
+All public symbols can still be accessed using std.format as usual.
+
+$(I Please note: The splitting process is still underway, more modules
```
thewilsonator, line 14: I thought we had an NB macro, if we do please use that.

changelog/splitting_format.dd
```diff
@@ -0,0 +1,15 @@
...
+The module `std.format` has been splitted into smaller modules:
+
+$(UL
+    $(LI `std.format.spec`: Stuff concerning the format string, mainly
+         the `struct FormatSpec` and the `template singleSpec`)
+    $(LI `std.format.internals`: Several modules containing non-public
+         stuff)
+)
+
+All public symbols can still be accessed using std.format as usual.
```
thewilsonator, line 12: `can still be accessed` -> `are still accessible`

changelog/splitting_format.dd
```diff
@@ -0,0 +1,15 @@
+Module `std.format` splitted into smaller modules
+
+The module `std.format` has been splitted into smaller modules:
+
+$(UL
+    $(LI `std.format.spec`: Stuff concerning the format string, mainly
```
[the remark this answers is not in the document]
berni44: For me, as a non native, it's difficult to find the right words. Maybe some of the native speakers who read this, could help here?

changelog/splitting_format.dd
berni44: I added them for completeness, but without any details.

changelog/splitting_format.dd
berni44: I'll have a look, if I can find something better. Anyway, I hope, that the splitting process is over (and hence this line completely removed) until the next release.

berni44: Couldn't find anything. Maybe you thought of the `Note:` section of public symbols? (Anyway, as I wrote: I hope, this line will never make it in the final changelog, because the split is over prior to that.)

RazvanN7: I wonder: is a changelog necessary for this? I mean, does the user have to know all of this? It's just private symbols being shuffled around.

berni44: > I wonder: is a changelog necessary for this? I mean, does the user have to know all of this? It's just private symbols being shuffled around.

No, not anymore. I finished meanwhile with the private stuff, which is mainly located in `internals`. Here `FormatSpec` and `singleSpec` are public symbols. I plan to split off the write parts (public: `formattedWrite` and `formatValue`) and the read parts (public: `formattedRead` and `unformatValue`) too in the next PRs - see #7765. I just started the changelog already, because in case we will not finish in time (what I do not hope for).

RazvanN7: I suggest we should add a changelog along with the first addition that actually requires one. Right now, a changelog is not needed. All PRs should be stand-alone.

berni44: > I suggest we add a changelog entry along with the first addition that actually requires one. Right now, a changelog is not needed.

I will do what you say, but actually I think, this is "the first that requires one": From [Contributing to Phobos in the wiki](https://wiki.dlang.org/Contributing_to_Phobos#Documentation_checks) : "All user impacting changes should have a corresponding changelog entry."

Here moving the two public symbols from `FormatSpec` and `singleSpec` from `std.format` to `std.format.spec` is IMHO such a change, because users who use one of the two (probably not many users, but...) can now change the import path to get a smaller import. I actually thought, it might even need DLF acceptance, because the new public symbol `std.format.spec` is introduced.

Or turning it the other way round: If this PR does not need a changelog, I can't see, which splitting PR should need one. The next two I plan (adding `std.format.read` and `std.format.write`), are similar to this one, and then the splitting is finished; the next steps after that will be cleaning up, like we did in the internals part.

> All PRs should be stand-alone.

Sounds a little bit opposite to "small steps"?!? :-S Of course, we can postpone adding a changelog until the splitting is completed. But in case it will never finish (I'm dying and no one else cares), people will never be informed about the change.

RazvanN7: I'm sorry, I missed the fact that public symbols are moved to a different file. I was under the impression that only private symbols were moved. In this case, I agree that a changelog is necessary, however, mentioning the `internals` part should be omitted in my opinion.

berni44: > however, mentioning the `internals` part should be omitted in my opinion.

OK. I removed that line.

RazvanN7 (approved):

PR #7842, opened 2021-03-11
Fix Issue 18780 - Inconsistent behavior with Variant holding int converting to unsigned types

berni44:
The docs of `ImplicitConversionTargets` state: "The possible targets are computed more conservatively than the language allows, eliminating all dangerous conversions."

I can't see, why a conversion from `int` to `uint` should be more dangerous than a conversion from `int` to `ulong` and the code is too old, to find any hint, why one was originally allowed and the other not.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -5036,7 +5036,7 @@ template ImplicitConversionTargets(T)
                        float, double, real, char, wchar, dchar);
     else static if (is(T == byte))
         alias ImplicitConversionTargets =
-            AliasSeq!(short, ushort, int, uint, long, ulong, CentTypeList,
+            AliasSeq!(short, int, long, SignedCentTypeList,
                        float, double, real, char, wchar, dchar);
     else static if (is(T == ubyte))
         alias ImplicitConversionTargets =
@@ -5044,13 +5044,13 @@ template ImplicitConversionTargets(T)
                        float, double, real, char, wchar, dchar);
     else static if (is(T == short))
         alias ImplicitConversionTargets =
-            AliasSeq!(int, uint, long, ulong, CentTypeList, float, double, real);
+            AliasSeq!(int, long, SignedCentTypeList, float, double, real);
     else static if (is(T == ushort))
         alias ImplicitConversionTargets =
             AliasSeq!(int, uint, long, ulong, CentTypeList, float, double, real);
     else static if (is(T == int))
         alias ImplicitConversionTargets =
-            AliasSeq!(long, ulong, CentTypeList, float, double, real);
+            AliasSeq!(long, SignedCentTypeList, float, double, real);
     else static if (is(T == uint))
         alias ImplicitConversionTargets =
             AliasSeq!(long, ulong, CentTypeList, float, double, real);
@@ -5102,7 +5102,7 @@ template ImplicitConversionTargets(T)
     import std.meta : AliasSeq;
 
     static assert(is(ImplicitConversionTargets!(ulong) == AliasSeq!(float, double, real)));
-    static assert(is(ImplicitConversionTargets!(int) == AliasSeq!(long, ulong, float, double, real)));
+    static assert(is(ImplicitConversionTargets!(int) == AliasSeq!(long, float, double, real)));
     static assert(is(ImplicitConversionTargets!(float) == AliasSeq!(double, real)));
     static assert(is(ImplicitConversionTargets!(double) == AliasSeq!(real)));
 
--- a/std/variant.d
+++ b/std/variant.d
@@ -1912,7 +1912,6 @@ static class VariantException : Exception
     assert( v.convertsTo!(long) );
     assert( v.get!(int) == 42 );
     assert( v.get!(long) == 42L );
-    assert( v.get!(ulong) == 42uL );
 
     v = "Hello, World!";
     assert( v.peek!(string) );


RazvanN7:
cc @atilaneves @andralex

RazvanN7:
If we are to merge this, it requires a changelog entry and a deprecation process.

std/traits.d
```diff
@@ -5036,21 +5036,21 @@ template ImplicitConversionTargets(T)
                        float, double, real, char, wchar, dchar);
     else static if (is(T == byte))
         alias ImplicitConversionTargets =
-            AliasSeq!(short, ushort, int, uint, long, ulong, CentTypeList,
```
atilaneves, line 5039: This makes sense, but ends up lying, since this compiles:

```d
int i = -5;
ulong oops = i;
```

atilaneves (commented):
I'm not even sure the original issue is valid.

std/traits.d
pbackus: `ImplicitConversionTargets` already "lies" like this in several cases, by design. The issue is that it lies about *some* signed → unsigned conversions (e.g. `int` → `uint`) but not about others (e.g. `int` → `ulong`).

berni44:
> If we are to merge this, it requires a changelog entry and a deprecation process.

The implementation differs from what the documentation claims and the implementation is inconsistent with itself, therefore I consider this a bug fix. And for bug fixes, normally there is no need to add a changelog entry (because it is automatically included in the fixed-issues-list), let alone a deprecation process. - There might be exceptions to this, but I think, this one isn't important enough to warrant one.

(An other solution would be to adapt the documentation, but I'm a little bit reluctant to write something like "The possible targets are computed more conservatively than the language allows, eliminating ~~all~~ some dangerous conversions. For example, `ImplicitConversionTarget!int` does not include `uint`, but it includes `ulong`.")

Additionally, I fear, a deprecation process will probably get on the users nerves, because every use of `Variant` will issue a deprecation warning, although most (if not all) uses are not affected by this change. (I haven't checked that though; perhaps a deprecation cycle could be implemented in a way, that it doesn't nerve.)

> I'm not even sure the original issue is valid.

I think, it's valid. See remark of @pbackus above.

RazvanN7:
> And for bug fixes, normally there is no need to add a changelog entry (because it is automatically included in the fixed-issues-list), let alone a deprecation process.

The problem is that there might be code out there that relies on the behavior prior to this PR. You yourself had to delete a test to make this green. At this point, we don't know whether people rely on this or not, but if they do we are breaking their code with this change. That is why a deprecation is needed.

> Additionally, I fear, a deprecation process will probably get on the users nerves, because every use of Variant will issue a deprecation warning, although most (if not all) uses are not affected by this change. (I haven't checked that though; perhaps a deprecation cycle could be implemented in a way, that it doesn't nerve.)

The deprecation should only be shown in cases where the code is no longer accepted. I see that this might be a bit complicated, however, last resort we can deprecate `ImplicitConversionTargets` entirely and create a newer improved one, but I think we can do better than that (maybe a deprecation can be issued for specific instantiation types of `get`).

berni44:
> > And for bug fixes, normally there is no need to add a changelog entry (because it is automatically included in the fixed-issues-list), let alone a deprecation process.
> 
> The problem is that there might be code out there that relies on the behavior prior to this PR. You yourself had to delete a test to make this green. At this point, we don't know whether people rely on this or not, but if they do we are breaking their code with this change. That is why a deprecation is needed.

Being a little bit pedantic: That's true for all bugfixes too. Anyway, I see that the docs are not very clear about which conversions are accepted and which not, so people will likely have peeked into the source code or tried it out and believed that result to be correct.

> > Additionally, I fear, a deprecation process will probably get on the users nerves, because every use of Variant will issue a deprecation warning, although most (if not all) uses are not affected by this change. (I haven't checked that though; perhaps a deprecation cycle could be implemented in a way, that it doesn't nerve.)
> 
> The deprecation should only be shown in cases where the code is no longer accepted. I see that this might be a bit complicated, however, last resort we can deprecate `ImplicitConversionTargets` entirely and create a newer improved one, but I think we can do better than that (maybe a deprecation can be issued for specific instantiation types of `get`).

I'll see, what I manage to do... Might take a while, because other stuff is more important to me.

std/traits.d
atilaneves: That makes sense, but I'm not sure we should be *adding* to the list of lies.

berni44:
I found the following in [The Deprecation Process](https://github.com/dlang/DIPs/blob/d8f2e769c3a8c711e7886ccecc93eac9795dae9c/DIPs/accepted/DIP1013.md): 

> Whether an "accepts-invalid" bug should include a deprecation period depends on the nature of the bug and the impact of the change, but in general, they should be fixed immediately without a deprecation period.

Isn't this the case here? I ask, because I yet was not successful in finding a satisfying deprecation process. In my eyes

```
Variant v;
v = 42;
v.get!(long);
```

should not trigger the deprecation warning, while

```
Variant v;
v = 42;
v.get!(ulong);
```

should.

std/traits.d
berni44: > That makes sense, but I'm not sure we should be _adding_ to the list of lies.

The alternative would be, to alter the documentation and accept `int` → `uint` too. Might be easier with regard to the deprecation discussion. I've got no opinion on my own here. My motivation is to handle an issue "somehow". (I used a random generator, which selected this issue.)

berni44: An other alternative would be to add `SaveConversionTargets` and `AllConversionTargets` and deprecate this function altogether...

RazvanN7:
@berni44 Now that #7881 has been merged, what should we do with this PR?

PR #6889, opened 2019-03-04
Move std.experimental.all to std

wilzbach:
`std.experimental.all` has been added more than a year ago (see https://github.com/dlang/phobos/pull/5916).

When we added it, we agreed to use `experimental` first to see whether there are any flaws with such a convenience package.
As none of these have been discovered and on the contrary more work has been started (and merged in some cases) which makes the compiler (or at least Phobos) compile faster.

Also, as the D community [started to like this package](https://forum.dlang.org/post/wbybwpjnntewswvzxrgl@forum.dlang.org), I think this was successful experiment and we should stabilize it.

BTW any interested reader can find _a lot_ of examples of `std.experimental.all` in action in D forums posts.
A short exerpt:
- https://forum.dlang.org/post/vntsidmvowboxtfeduxp@forum.dlang.org
- https://forum.dlang.org/post/dqptwzycafzfvwwvyzni@forum.dlang.org
- https://forum.dlang.org/post/syguaofawaxbqfqqyxsw@forum.dlang.org
- https://forum.dlang.org/post/qlvexnrvjriyapolkepk@forum.dlang.org
- https://forum.dlang.org/post/bfvzlyzrzlcotisxqiur@forum.dlang.org
- https://forum.dlang.org/post/rfagavflrtfaisiicufz@forum.dlang.org
- https://forum.dlang.org/post/jkiymjpdqhaealytbqgo@forum.dlang.org
...

Also, once this is stabilized, I can finally enable it by default for `run.dlang.io`.

PR #7843, opened 2021-03-11
Add forgotten modules to std.package.

berni44:
Maybe `typetuple` was intentionally left out?


berni44:
> std.outbuffer, std.typetuple and std.xml were intentionally removed by #6889 (std.typetuple was commented out already, however).

Then it's only about adding sumtype. :-)

RazvanN7 (approved):

PR #7847, opened 2021-03-12
Avoid quadratic template expansion in constraints of multiple search term versions of std.algorithm.searching.startsWith & endsWith

n8sh:
See https://issues.dlang.org/show_bug.cgi?id=21473. There is other stuff going on but fixing this makes that code compilable. Because in certain situations this is causing the compiler to crash I think it is reasonable to consider this a bug fix rather than an enhancement.


thewilsonator (approved):
Ahh, my favourite template constraint to give horrible untraceable error messages

PR #7852, opened 2021-03-13
Fix Issue 21705 - Nullable!T.opEquals fails for T with non-const opEq…

pbackus:
…uals overload


Herringway:
This seems like it would break const(Nullable!T).opEquals.

pbackus:
There's a `unittest` that tests `opEquals` with both `const(Nullable)` and `immutable(Nullable)`.

RazvanN7 (approved):

PR #7859, opened 2021-03-13
Add debug symbols for tests to posix.mak

berni44:
I realized, that I'm repeatedly adding and removing `-g` when using `.test`, because I'd like to know line numbers of runtime exceptions.

I don't know, if there is a better way to do so, because I'm not very much familiar with make files. And I also do not know, if there is a reason, why this isn't added...


pbackus:
You can do this by setting `BUILD=debug` on the command line.

berni44:
> You can do this by setting `BUILD=debug` on the command line.

Ah, thanks! :-)

PR #7861, opened 2021-03-14
Revert "Fix Issue 21707 - std.base64: Faulty input creates range erro…

berni44:
…r instead of Base64Exception"

This reverts commit cf41baba2123e1c4680baf06016b8a93bbaf57ce.

Meanwhile I found out, why the removed piece of code was added. Removing it introduced a regression (see below), so I decided to revert this. I'll have to find an other way to fix that bug, but that seems more difficult...

The following code now fails, but has worked before:

```
ubyte[][] input = [['e', 'y'], ['0', '=']];
Base64.decoder(input);
```

### DIFF
--- a/std/base64.d
+++ b/std/base64.d
@@ -1463,6 +1463,14 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
                     data ~= cast(const(char)[])range_.front;
                 }
             }
+            else
+            {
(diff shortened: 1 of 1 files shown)


Geod24:
Can you add a test for the regression ?

Geod24 (approved):

PR #7862, opened 2021-03-14
std.format: Add x87-reals (step 1) - extract bit patterns

berni44:
Extracting the bits of x87-reals in CTFE is a little bit more difficult than just casting. 

There is one line, where I feel unsure, caused by a remark of @kinke in [issue 21688](https://issues.dlang.org/show_bug.cgi?id=21688#c3): 
```
if (isNaN(val) || val == real.infinity || val == -real.infinity)
``` 
Can we rely on these comparisons to find out, if the value is infinity?

Although it can currently never be executed, I had to change a line inside `printFloatA`, to make the code compile. There is a static assert next to that line, that would fail else. This change would have happened anyway in one of the next PRs.

A side effect of this PR is, that NaNs and infinity can already be formatted, so I added the corresponding unittests.

### DIFF
--- a/std/format/internal/floats.d
+++ b/std/format/internal/floats.d
@@ -22,24 +22,81 @@ package(std.format) enum RoundingMode { up, down, toZero, toNearestTiesToEven, t
 
 package(std.format) auto printFloat(T, Char)(return char[] buf, T val, FormatSpec!Char f,
                                              RoundingMode rm = RoundingMode.toNearestTiesToEven)
-if (is(T == float) || is(T == double) || (is(T == real) && T.mant_dig == double.mant_dig))
+if (is(T == float) || is(T == double)
+    || (is(T == real) && (T.mant_dig == double.mant_dig || T.mant_dig == 64)))
 {
-    static if (is(T == float))
+    static if (is(T == real) && T.mant_dig == 64)
     {
-        ulong ival = () @trusted { return *cast(uint*) &val; }();
+        ulong mnt = void;
+        int exp = void;
+        string sgn = "";
+
+        if (__ctfe)
+        {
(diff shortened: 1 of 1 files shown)


kinke:
Wrt. CTFE math: see https://github.com/dlang/dmd/blob/master/src/dmd/builtin.d and https://github.com/dlang/dmd/blob/master/src/dmd/root/ctfloat.d. E.g., `std.math.isInfinity` is mapped to `CTFloat.isInfinity`.

berni44:
> Wrt. CTFE math: see https://github.com/dlang/dmd/blob/master/src/dmd/builtin.d and https://github.com/dlang/dmd/blob/master/src/dmd/root/ctfloat.d. E.g., `std.math.isInfinity` is mapped to `CTFloat.isInfinity`.

When I understand you right, I can just user `std.math.isInfinity` and be sure, that it works in CTFE as intended? - Changed the PR to use `std.math.isInfinity`.

berni44:
@RazvanN7 Is there anything I can do here?

RazvanN7:
I was hoping that @kinke would make another pass on this.

RazvanN7:
It looks like there are a lot of lines that are not covered by tests. Is that an error in the coverage reporter because of templates?

berni44:
> It looks like there are a lot of lines that are not covered by tests. Is that an error in the coverage reporter because of templates?

That's because of the `assert(false)` in line 133. Currently only `NaN` and `Inf` is implemented. This will change with the next PR which (probably) adds support for `%f` (but I need this one first and didn't want to put too much into one PR.)

RazvanN7 (approved):

PR #7863, opened 2021-03-15
std.format: Split off std/format/read.d

berni44:
Continuing the split off process... See #7765 for the whole picture.


PetarKirov (approved):
I checked the diff locally with the `diff.colormoved` git option and everything looks good to me. Nice work!

PR #7873, opened 2021-03-17
Fix Issue 21725 - Specifying null as bitfields variable name now fails

CyberShadow:


dejlek (approved):
Makes sense

thewilsonator (approved):

dukc:
Thanks for cleaning my mess up! (I approved the regression pr).

PR #7874, opened 2021-03-17
std.format: Refactor package (step 3): Add assert messages.

berni44:


std/format/package.d
```diff
@@ -792,21 +792,25 @@ private void formatReflectTest(T)(ref T val, string fmt, string formatted, strin
         {
             alias aa1 = val;
             alias aa2 = val2;
-            assert(aa1 == aa2);
+            assert(aa1 == aa2, "Identical associative arrays differ.");
```
RazvanN7, line 795: This error message could be improved: "Associative arrays differ, but should be identical"

std/format/package.d
```diff
@@ -792,21 +792,25 @@ private void formatReflectTest(T)(ref T val, string fmt, string formatted, strin
...
 
-            assert(aa1.length == aa2.length);
+            assert(aa1.length == aa2.length, "Length of identical associative arrays differs.");
```
RazvanN7, line 797:
```suggestion
            assert(aa1.length == aa2.length, "Lengths of associative arrays differ, but should be equal .");
```

std/format/package.d
```diff
@@ -792,21 +792,25 @@ private void formatReflectTest(T)(ref T val, string fmt, string formatted, strin
...
 
-            assert(aa1.keys == aa2.keys);
+            assert(aa1.keys == aa2.keys, "Keys of identical associative arrays differ.");
```
RazvanN7, line 799:
```suggestion
            assert(aa1.keys == aa2.keys, "Keys of associative arrays differ, but should be identical");
```

std/format/package.d
```diff
@@ -792,21 +792,25 @@ private void formatReflectTest(T)(ref T val, string fmt, string formatted, strin
...
 
-            assert(aa1.values == aa2.values);
-            assert(aa1.values.length == aa2.values.length);
+            assert(aa1.values == aa2.values, "Values of identical associative arrays differ.");
```
RazvanN7, line 801:
```suggestion
            assert(aa1.values == aa2.values, "Values associative arrays differ, but should be identical");
```

std/format/package.d
```diff
@@ -792,21 +792,25 @@ private void formatReflectTest(T)(ref T val, string fmt, string formatted, strin
...
+            assert(aa1.values.length == aa2.values.length,
+                "Length of values of identical associative arrays differ.");
```
RazvanN7, line 803: ditto

std/format/package.d
```diff
@@ -792,21 +792,25 @@ private void formatReflectTest(T)(ref T val, string fmt, string formatted, strin
...
             foreach (i; 0 .. aa1.values.length)
-                assert(aa1.values[i] == aa2.values[i]);
+                assert(aa1.values[i] == aa2.values[i],
+                    "Element of identical associative arrays differs.");
 
             foreach (i, key; aa1.keys)
-                assert(aa1.values[i] == aa1[key]);
+                assert(aa1.values[i] == aa1[key],
+                    "Value of associative array cannot accessed via values.");
```
RazvanN7, line 810:
```suggestion
                    "Value of associative array cannot be accessed via values.");
```

std/format/package.d
```diff
@@ -792,21 +792,25 @@ private void formatReflectTest(T)(ref T val, string fmt, string formatted, strin
...
             foreach (i, key; aa2.keys)
-                assert(aa2.values[i] == aa2[key]);
+                assert(aa2.values[i] == aa2[key],
+                    "Value of associative array cannot accessed via values.");
```
RazvanN7, line 813:
```suggestion
                    "Value of associative array cannot be accessed via values.");
```

RazvanN7 (changes requested):
Please apply the changes to the other messages that I have not commented on as well.

PetarKirov:
To be honest, I think that only asserts inside user-facing functions should have messages. Adding assert messages inside unit tests is pointless as only library developers are ever going to see them and building the code with `-checkaction=context` should be more than sufficient to provide enough information. Actually having assert messages may be even worse, as (IIRC) even if the  code is compiled with this switch, the special formatting won't be applied and only the manually written message may be displayed, if the assert fails.

I understand that enabling the corresponding check in DScanner is motivating in itself, but I'd argue that many of the DScanner rules need additional tuning to make more practical and useful in general.

PetarKirov:
That said, I wouldn't want to throw away work, so by all means feel free to continue. Just keep this in mind for future module.
To clarify, in case there's any confusion: I'm not against assert messages in general, just those inside unit tests, where the code is self-explanatory without them as well. Every rule has an exception, and if an assert message can provide insightful information it has its place even in unit tests. But my general observation is that it's rarely the case, unless it is verifying higher-level features and conditions that are not obvious.

berni44:
> That said, I wouldn't want to throw away work, so by all means feel free to continue.

That's fine. :-) (Isn't there even a buddhist proverb saying, it's most fun to throw away what you just achieved?)

PR #7875, opened 2021-03-17
Deprecate public access of formatElement

berni44:
`formatElement` is currently `public`, but undocumented. A comment next to it says "undocumented because of deprecation", but I think this is wrong. Looking into the history of this, I found out, that it was made public, because it was useful in `std.typecons`. I presume, that at that time `package(std)` was not available.

I see two ways out of this:
1. Make it really public, that is add documentation and useful public unittests.
2. Deprecate it's public view.

I've chosen the second way, because in my opinion formatting elements and should not be done differently to formatting these elements as single items (i.e. "%(%s%)" should not format an element different than "%s" does). While this PR doesn't really change this behavior, it makes it easier to do so later with an other deprecation cycle.

To deprecate the public use, but not the private use, I moved them to `std.format.internal.write` and replaced the original calls by wrappers with the `deprecated` attribute.

Due to the use in `std.typecons` I had to make the moved functions `package(std)`. Unfortunately this doesn't work well with the package statement `package(std.format):` at the top of the file, so I had to remove this and add it to all symbols that where not private separately.

### DIFF
--- a/changelog/deprecate_formatElement.dd
+++ b/changelog/deprecate_formatElement.dd
@@ -1,0 +1,12 @@
+Deprecate std.format : formatElement
+
+`formatElement` from `std.format` has been accidentally made public and
+will be removed from public view in 2.107.0.
+
+Please use instead of `formatElement(sink, value, fmt)`:
+
+```
+import std.range : only;
+
+sink.put(format!("%("~fmt~"%)")(only(value)));
+```
--- a/std/format/internal/write.d
+++ b/std/format/internal/write.d
@@ -1514,7 +1514,7 @@ private void formatRange(Writer, T, Char)(ref Writer w, ref T val, scope const r
 if (isInputRange!T)
 {
     import std.conv : text;
-    import std.format : formatElement, FormatException, formatValue, NoOpSink;
+    import std.format : FormatException, formatValue, NoOpSink;
     import std.range.primitives : ElementType, empty, front, hasLength,
         walkLength, isForwardRange, isInfinite, popFront, put;
 
@@ -1752,7 +1752,7 @@ void formatChar(Writer)(ref Writer w, in dchar c, in char quote)
 void formatValueImpl(Writer, T, Char)(auto ref Writer w, T obj, scope const ref FormatSpec!Char f)
 if (is(AssocArrayTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
 {
-    import std.format : enforceFmt, formatElement, formatValue;
+    import std.format : enforceFmt, formatValue;
     import std.range.primitives : put;
 
     AssocArrayTypeOf!T val = obj;
@@ -1804,7 +1804,7 @@ if (is(AssocArrayTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
 @safe unittest
 {
     import std.array : appender;
-    import std.format : formatElement, singleSpec;
+    import std.format : singleSpec;
 
     // Bug #17269. Behavior similar to `struct A { Nullable!string B; }`
     struct StringAliasThis
@@ -2529,7 +2529,7 @@ void formatValueImpl(Writer, T, Char)(auto ref Writer w, auto ref T val,
 if ((is(T == struct) || is(T == union)) && (hasToString!(T, Char) || !is(BuiltinTypeOf!T))
     && !is(T == enum))
 {
-    import std.format : enforceValidFormatSpec, formatElement;
+    import std.format : enforceValidFormatSpec;
     import std.range.primitives : put;
 
     static if (__traits(hasMember, T, "toString") && isSomeFunction!(val.toString))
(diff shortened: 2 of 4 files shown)


CyberShadow:
I see a lot of uses here:
https://github.com/search?l=D&q=formatElement&type=Code

PetarKirov (approved):

PetarKirov:
I missed @CyberShadow's comment. I am quite surprised to see so many uses of this in the wild.

CyberShadow:
If `formatElement` is superfluous, why is its usage necessary in `std.typecons`?

berni44:
> If `formatElement` is superfluous, why is its usage necessary in `std.typecons`?

I can't answer this question, because I do not know, if it is superfluous (though my perception was, that it is, at least outside of `std.format`). My motivation here is to handle `formatElement` somehow, because a public function without docs is a bad idea and my current mission is to improve `std.format` in various ways.

Having said this, the purpose of this function inside `std.format` is to format elements of certain ranges, structs and associative arrays. The main idea is, to wrap strings and characters into double or single quotes like D source code literals. (It is a larger debate, if this is a good or a bad idea, but IMHO not the topic here.) `std.typecons` uses this function in a similar way for formatting the elements of tuples.

berni44:
> I am quite surprised to see so many uses of this in the wild.

Well, there aren't. When looking closer into this, it's almost all copies of Phobos...

CyberShadow:
> `std.typecons` uses this function in a similar way for formatting the elements of tuples.

As a rule, it should be possible to reimplement any part of the standard library in third-party code (whether from scratch or by copy-pasting code from Phobos). For this reason, we generally should try to make all the necessary building blocks available. The biggest reason for why we would want to not do that is if we wanted to avoid making a guarantee about the symbols' future availability for now, thus giving us flexibility to adjust the implementation. I can't say to what extent this would apply to `formatElement`.

If I understand correctly that `formatElement!fmt(value)` is exactly equivalent to `format!("%("~fmt~"%)")(only(value))`, then it does seem like a rather small loss. On the other hand looking at the GitHub results it does seem like a useful building block for stringifying sequences of items. So, I can't advise one way or the other.

> The main idea is, to wrap strings and characters into double or single quotes like D source code literals (and escape unprintable characters). (It is a larger debate, if this is a good or a bad idea, but IMHO not the topic here.)

It's definitely useful for debug printing user types using `writeln` and such, but that's a few layers away.

berni44:
> The biggest reason for why we would want to not do that is if we wanted to avoid making a guarantee about the symbols' future availability for now, thus giving us flexibility to adjust the implementation. I can't say to what extent this would apply to `formatElement`.

I've got plans to get rid of the inconsistency between "%(%s%)" and "%s". The idea is, to introduce a new format character %S, which formats the value as source code literal. This is more or less, what "%(%s%)" does with it's elements. So users would need to replace "%(%s%)" with "%(%S%)" to keep the current behavior and "%(%s%)" would produce, what currently is "%-(%s%)". This would free the abused dash to mean left justification again and to add a total width.

It's difficult to do so with a deprecation cycle (a lot of not deprecated uses of `std.format` would get deprecation messages). But, as far as I know, there are rather [concrete plans to go for Phobos 2.0](https://forum.dlang.org/thread/rv7hq6$3aq$1@digitalmars.com). With that it would be possible to do this switch, and in this case `formatElement` would not be needed anymore.

> If I understand correctly that `formatElement!fmt(value)` is exactly equivalent to `format!("%("~fmt~"%)")(only(value))`, then it does seem like a rather small loss.

Unfortunately it is not. What would work though is

```
FormatSpec!Char f = FormatSpec!Char("%(%s%)");
f.seqBefore = "";
f.seqAfter = "";
f.writeUpToNextSpec(sink);
formatValue(sink, [value], f);
```

instead of 

```
FormatSpec!Char f;
formatElement(sink, value, f)
```

Unfortunately `seqBefore` and `seqAfter` are `enums` and can't be changed right now. So it would be needed to make them variables (or template params). But I think, this change could be useful for other things to.

berni44:
I now made `seqBefore`, `seqAfter` (and `keySeparator` and `seqSeparator` for completeness) variables to provide an alternative to `formatElement`. Users might wish this anyway to have more flexibility when using `formatValue` and the docs tell about default values, which makes no sense with `enums`.

I also replaced the only appearance of `formatElement` in Phobos, namely in `std.typecons` with the suggested replacement.

CyberShadow:
> Unfortunately it is not.

What is an example where it will not?

berni44:
> > Unfortunately it is not.
> 
> What is an example where it will not?

Sorry, I overlooked `only`. I didn't know that function and thought you wrote sort of a pseudo code. Seems to work indeed. I'll change the PR accordingly. Many thanks for that idea.

RazvanN7 (approved):

PR #7877, opened 2021-03-18
casting std.BigInts to built-in floating point types not allowed in `@safe` code without -preview=dip1000

n8sh:


thewilsonator:
which lines specifically are unsafe without dip1000? `@trusted` on a large function is not a good idea.

n8sh:
Changed to use `@trusted` lambda pattern.

thewilsonator (approved):

PR #7879, opened 2021-03-18
std.algorithm.mutation.copy fails on overlapping arrays if the source array's pointer is less than the destination array's pointer

n8sh:


thewilsonator (approved):

std/algorithm/mutation.d
```diff
@@ -374,14 +374,22 @@ if (isInputRange!SourceRange && isOutputRange!(TargetRange, ElementType!SourceRa
...
-        immutable overlaps = __ctfe || () @trusted {
+        immutable overlaps = () @trusted {
             return source.ptr < target.ptr + tlen &&
                 target.ptr < source.ptr + slen; }();
 
         if (overlaps)
         {
-            foreach (idx; 0 .. slen)
-                target[idx] = source[idx];
+            if ((() @trusted => source.ptr < target.ptr)())
```
n8sh, line 383: DMD infers this as safe without a `@trusted` lambda.

PR #7880, opened 2021-03-18
std.format: Move description of format string to package.

berni44:
In my opinion, the format string is central to `std.format` and therefore should be described at package level, not at `formattedWrite` (even the examples do not use `formattedWrite`). I refrained myself from doing more changes, although there are a lot of them needed, because I think, that it is much easier to be reviewed, when large chunks are just copied and not modified.


thewilsonator (approved):

std/format/write.d
```diff
@@ -62,318 +62,6 @@ import std.traits : isSomeString;
    Throws: Mismatched arguments and formats result in a $(D
    FormatException) being thrown.
 
-   Format_String: <a name="format-string">$(I Format strings)</a>
```
TungstenHeart, line 65: I suggest to add a link to the new symbol that has for ddoc the old ddoc

TungstenHeart (changes requested):
IN addition to my other comment, I think that the link to the format spec [here](https://dlang.org/phobos/std_format.html#.format) would be broken.

TungstenHeart:
The link I mention is not froken but doeds not lead to what is advertised. It should be changed to a link to the package.

TungstenHeart:
ping @thewilsonator for auto-merge cancel

std/format/write.d
berni44: Sorry, I do not understand, what you want to say with that. What is the "new symbol"? And what does "has for ddoc the old ddoc" mean?

TungstenHeart: a link to the what the new location of the description.

berni44:
> The link I mention is not froken but does not lead to what is advertised. It should be changed to a link to the package.

I just removed it, because I want to go through the whole documentation anyway in the next days and add links to the description of the format string at all places.

TungstenHeart:
ah yeah I see you mentioned it's not complete

> I refrained myself from doing more changes,

but why wait ? It's not something that may have to be reverted.

berni44:
> but why wait ? It's not something that may have to be reverted.

No, not reverted, of course. But it's really a lot, that needs to be done and I want to do small steps here. You know, there are dozens of such small changes, that I could add here. I'd prefer to do them step by step, so I don't overlook anything I don't produce rubbish in between.

For example the change you request above would take me a while to find out how to link to `<a name="...">` on package scope correctly. With a thorough documentation of ddoc it would be much work, but usually I have to go through several other modules of phobos to find a place, where someone did something similar... Of course I could do it quick and dirty (with `$REF_ALTTEXT`) to fulfill your request, but I do not like to do so.

TungstenHeart:
Ok considering that to change the other links is not something that's been forgotten you can dismiss my objections.

berni44:
> Ok considering that to change the other links is not something that's been forgotten you can dismiss my objections.

Oh, thanks. (I actually searched for broken links, but overlooked the one from `format`, because it was not really broken.)

RazvanN7 (approved):

PR #7881, opened 2021-03-18
Add AllImplicitConversionTargets

berni44:
This is meant as an alternative to #7842: The idea is to add two new functions to make clear, if the list contains unsafe conversions, or not.

If accepted, I will use these in a separate PR (because it needs also some changes in `std.variant`) to deprecate `ImplicitConversionTargets`.

I left everything, that is not about builtin types untouched. (I don't understand this part in every detail yet.)

Some points, that might be debatable in `SafeImplicitConversionTargets`:
- I removed `byte` from the list of `char` and likewise `short` from `wchar` and `int` from `dchar`. For me, these casts feel unsafe. What do you think?
- I treated `real` like `double`. For x87-reals this makes no difference, but longer reals should be able to represent all values of `long` and `ulong`. As there is no clear list of which reals D does support, I can't check this.

### DIFF
--- a/changelog/add_conversion_target.dd
+++ b/changelog/add_conversion_target.dd
@@ -1,0 +1,6 @@
+New function: `AllImplicitConversionTargets`
+
+The function `ImplicitConversionTargets` in module `std.traits` has a
+design flaw: The list of conversion targets contains some, but not all
+unsafe conversion targets. To overcome this, a new functions has been
+added.
--- a/std/traits.d
+++ b/std/traits.d
@@ -59,6 +59,7 @@
  * ))
  * $(TR $(TD Type Conversion) $(TD
  *           $(LREF CommonType)
+ *           $(LREF AllImplicitConversionTargets)
  *           $(LREF ImplicitConversionTargets)
  *           $(LREF CopyTypeQualifiers)
  *           $(LREF CopyConstness)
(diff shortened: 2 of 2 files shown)


RazvanN7:
cc @atilaneves

PetarKirov:
I am in favor of adding a separate symbol. This also gives the possibility to make the conversion targets even stricter:
* no signed -> unsigned conversions
* unsigned -> signed conversion only to larger types: `ubyte` -> `short`, `int`, `long`, `cent` (but not to `byte`)
* integer -> floating-point conversions only the value is accurately representable:
  * `[u]{byte,short}` -> `float`
  * `[u]{byte,short,int}` -> `double`
  * `[u]{byte,short,int,long}` -> `real`, when `real.sizeof > double.sizeof` 

Though, if we decide to go this way, the name should probably be `SafeConversionTargets` to avoid confusion.

n8sh:
Is there anything other than `std.variant.Variant` that would use this? If not then just one new template is needed that does whatever you want for `Variant`, not both of `AllImplicitConversionTargets` and `SafeImplicitConversionTargets`.

berni44:
> Is there anything other than `std.variant.Variant` that would use this? 

Not in Phobos. And if I should guess, it is also not used somewhere else, because the selection is at will and therefore hardly usable.

> If not then just one new template is needed that does whatever you want for `Variant`, not both of `AllImplicitConversionTargets` and `SafeImplicitConversionTargets`.

But which one? If we go for `SafeImplicitConversionTargets` we could just use #7842. There it was argued that we need a deprecation cycle - see the discussion there. On the other hand, if we go for `AllImplicitConversionTargets`, we would just relax the result and would have to change the docs.

berni44:
How to continue here? Any opinions?

RazvanN7:
We can go with `AllImplicitConversionTargets` and deprecate `ImplicitConversionTargets` and if the need arises implement `StrictConversionTargets` in later PRs.

berni44:
> We can go with `AllImplicitConversionTargets` and deprecate `ImplicitConversionTargets` and if the need arises implement `StrictConversionTargets` in later PRs.

OK. I removed `SaveImplicitConversionTargets` from this PR. I'll deprecate `ImplicitConversionTargets` in a separate PR, because this needs some prior work in `std.variants`.

RazvanN7 (approved):

PR #5005, opened 2016-12-30
Fix issue 16432 - JSON incorrectly parses to string

wilzbach:
The entire `json` module seems to require a bit of love - though I am still hoping for [std.data.json](https://github.com/s-ludwig/std_data_json)

PR #7885, opened 2021-03-18
Fix Issue 16432 - JSON incorrectly parses to string

berni44:
Inspired by #5005.

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -1680,12 +1680,17 @@ if (isOutputRange!(Out,char))
                 }
                 else
                 {
-                    import std.format : format;
+                    import std.algorithm.searching : canFind;
+                    import std.format : sformat;
                     // The correct formula for the number of decimal digits needed for lossless round
                     // trips is actually:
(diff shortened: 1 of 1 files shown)


std/json.d
```diff
@@ -1680,12 +1680,16 @@ if (isOutputRange!(Out,char))
...
                 {
+                    import std.algorithm.searching : canFind;
                     import std.format : format;
                     // The correct formula for the number of decimal digits needed for lossless round
                     // trips is actually:
                     //     ceil(log(pow(2.0, double.mant_dig - 1)) / log(10.0) + 1) == (double.dig + 2)
                     // Anything less will round off (1 + double.epsilon)
-                    json.put("%.18g".format(val));
+                    auto result = format!"%.18g"(val);
+                    json.put(result);
```
Geod24, line 1690: Use `formattedWrite` to avoid needless intermediate allocation

berni44: In the next line I need this intermediate result. I think, I could use `sformat` though to avoid GC allocation.

RazvanN7 (approved):

PR #7886, opened 2021-03-19
Fix Issue 21731 - SumType should provide convenient access to the typ…

pbackus:
…e index

See pbackus/sumtype#57.

### DIFF
--- a/std/sumtype.d
+++ b/std/sumtype.d
@@ -795,8 +795,8 @@ public:
      * Returns the index of the type of the `SumType`'s current value in the
      * `SumType`'s `Types`.
      *
-     * If the `SumType` is qualified, then its qualifiers are applied to
-     * `Types` before determining the index.
+     * If the `SumType` is qualified, returns the index of the type in `Types`
+     * whose qualified version matches the `SumType`'s current value.
      */
     size_t typeIndex() const
     {
--- a/std/sumtype.d
+++ b/std/sumtype.d
@@ -790,6 +790,18 @@ public:
             return this.match!hashOf;
         }
     }
+
+    /**
+     * Returns the index of the type of the `SumType`'s current value in the
+     * `SumType`'s `Types`.
+     *
+     * If the `SumType` is qualified, then its qualifiers are applied to
+     * `Types` before determining the index.
+     */
+    size_t typeIndex() const
+    {
+        return tag;
+    }
 }
 
 // Construction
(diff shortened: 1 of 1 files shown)


std/sumtype.d
```diff
@@ -1466,6 +1485,30 @@ version (D_BetterC) {} else
...
+}
+
+// Type index for qualified SumTypes
+// Disabled in BetterC due to use of dynamic arrays
+version (D_BetterC) {} else
+@safe unittest
+{
+    alias MySum = SumType!(const(int[]), int[]);
+
+    int[] ma = [1, 2, 3];
```
Geod24, line 1504:
```suggestion
    int[3] ma = [1, 2, 3];
```

Wouldn't that be enough to make it work in `betterC` ?

pbackus: In theory, yes. In practice, if I use a stack-allocated static array and construct the `SumType` with a slice (i.e., `MySum(ma[])`), the tests will fail even with `-preview=dip1000`, because the workaround for [issue 21229][1] used in `SumType`'s constructors does not correctly preserve lifetime information about the constructor's argument.

[1]: https://issues.dlang.org/show_bug.cgi?id=21229

schveiguy:
Why isn't this just `return tag`? Doesn't that do the same thing?

pbackus:
In some rare cases (see unittests) it is possible for the same type to have two different tag values. The current implementation always returns a unique value for a given type, but conversation at pbackus/sumtype#57 has convinced me that this may not be desirable.

schveiguy:
Hm... so in a case where you have SumType!(int, const int), assigned to a const int, the tag will be 1 with a standard instance, and 0 with a const instance?

I'll point out that returning a value of 1 is still going to be valid in the list of (const int, const int). I would recommend just returning the tag, it's the most expected result, even on a const SumType.

pbackus:
Yes, I've come the same conclusion myself.

pbackus:
Buildkite failure appears to be spurious:

```
fork/exec /usr/bin/buildkite-agent: cannot allocate memory
```

thewilsonator (approved):

std/sumtype.d
```diff
@@ -790,6 +790,18 @@ public:
             return this.match!hashOf;
         }
     }
+
+    /**
+     * Returns the index of the type of the `SumType`'s current value in the
+     * `SumType`'s `Types`.
+     *
+     * If the `SumType` is qualified, then its qualifiers are applied to
+     * `Types` before determining the index.
```
schveiguy, line 799: This second paragraph should be removed.

schveiguy (changes requested):
Looks good. Trivial change, looks like some docs were left over from the previous incarnation.

std/sumtype.d
pbackus: Should it? In the case of something like `const(SumType!(A, B))(const(A)())`, the type of the value is `const(A)`, which is not present in `Types`. The first paragraph alone is insufficient to determine the correct behavior in this case (although I'll grant that it is not difficult to guess).

schveiguy: Well, you aren't applying the qualifier. You are just returning the tag.

The paragraph suggests something different happens when a qualifier is used.

Perhaps a rewrite is needed. Maybe:

```
If the SumType is qualified, the index indicates the Type if it were unqualified.
```

pbackus: The specific wording you suggest is incorrect, because the type in `Types` *might* already be the qualified version, but I agree that a rewrite is the best solution here.

schveiguy (approved):

PR #7887, opened 2021-03-19
std.math: Split off std/math/trig.d

berni44:


thewilsonator (approved):

std/math/trig.d
```diff
@@ -0,0 +1,1281 @@
...
+            //printf("tan(%Lg) = %Lg, should be %Lg\n", cast(real) x, cast(real) t, cast(real) r);
+            assert(isClose(r, t, CommonDefaultFor!(T,T), CommonDefaultFor!(T,T)));
+        }
+    }
+
+    import std.meta : AliasSeq;
+    foreach (T; AliasSeq!(real, double, float))
+        testTan!T();
+
+    assert(equalsDigit(tan(PI / 3), sqrt(3.0L), useDigits));
```
berni44, line 482: CircleCI is running into problems here, the linker does not know about `equalsDigit`. It's defined in `std.math.package.d` inside of a `version (StdUnittest) package` block with visibility `package`. Should be available for the linker in my eyes. I could need help here, because I don't see, what could be wrong.

thewilsonator:
You'll also have to contend with https://github.com/dlang/dmd/blob/32313b9f696520fae4910b96d5e4e6bd9b28e4a8/src/dmd/builtin.d#L80 I think.

std/math/trig.d
thewilsonator: try `package(std)` instead. That should leave the behaviour unchanged.

berni44:
> You'll also have to contend with https://github.com/dlang/dmd/blob/32313b9f696520fae4910b96d5e4e6bd9b28e4a8/src/dmd/builtin.d#L80 I think.

Huh, well, I'm not familiar with the internals of dmd. I guess, I need someone to help me here too.

std/math/trig.d
berni44: Doesn't work... I think it's got to do with this `version` stuff. I guess, that `std.math` is compiled without `version(StdUnittest)` while the check for coverage in `std.math.trig` also checks the unittests. I'll try `public`. If that still produces the same error messages, we know, it's not the visibility...

thewilsonator:
You'll need to change [these lines](https://github.com/dlang/dmd/blob/32313b9f696520fae4910b96d5e4e6bd9b28e4a8/src/dmd/builtin.d#L95-L100) to account for the fact that e.g. `std.math.trig` has two packages above it.

ibuclaw:
Originally, the idea was to do `std.math.trig`, though if I were to restart this transition now, I'd probably go for the more verbose `std.math.trigonometry`.

There should be no intrinsics in `std.math` anymore - they are all just aliases or wrappers around `core.math`.  Some functions however have a special built-in status at CTFE only.  While it would be simple to bundle these into a single module, I don't think there's any logic, order, or sense about what math functions are intrinsically CTFE-able.

berni44:
Closing this, until a thorough plan exists.

PR #7889, opened 2021-03-19
std.algorithm docs: tweak `each` examples

ntrel:
Make first example `@safe`.
Add separate example for `each` with index 
Split `opApply` example into separate block.
Remove doc-comment for example that wasn't as clear.

### DIFF
--- a/std/algorithm/iteration.d
+++ b/std/algorithm/iteration.d
@@ -880,7 +880,7 @@ private struct MapResult(alias fun, Range)
 
 // each
 /**
-Eagerly iterates over `r` and calls `fun` over _each element.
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

std/algorithm/iteration.d
```diff
@@ -1113,20 +1115,30 @@ public:
...
+}
+
+/// `each` can pass an index variable for iterable objects which support this
+@safe unittest
+{
+    auto arr = new int[4];
 
-    // Indexes are also available for in-place mutations
-    arr[] = 0;
     arr.each!"a=i"();
```
ntrel, line 1132: This is breaking the test suite, arr was long before. Will fix a bit later today.

PR #7890, opened 2021-03-19
New overload for std.base64.Base64Impl.decoder 

dukc:
It takes char arrays, strings included, without needing to manually call `.byCodeUnit` or `.represenation`

### DIFF
--- a/std/base64.d
+++ b/std/base64.d
@@ -1645,10 +1645,11 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
      *
      * Params:
      *  range = An $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
-     *      over the data to be decoded.
+     *      over the data to be decoded, or a `char` array. Will not accept
+     *      `wchar[]` nor `dchar[]`.
      *
      * Returns:
-     *  If $(D_PARAM range) is a range of characters, a `Decoder` that
+     *  If $(D_PARAM range) is a range or array of `char`, a `Decoder` that
      *  iterates over the bytes of the corresponding Base64 decoding.
      *
      *  If $(D_PARAM range) is a range of ranges of characters, a `Decoder`
@@ -1656,7 +1657,7 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
      *  the range.
      *
      *  In both cases, the returned `Decoder` will be a
-     * $(REF_ALTTEXT forward range, isForwardRange, std,range,primitives) if the
+     *  $(REF_ALTTEXT forward range, isForwardRange, std,range,primitives) if the
      *  given `range` is at least a forward range, otherwise it will be only
      *  an input range.
      *
@@ -1673,7 +1674,6 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
      * }
      * -----
      *
-     * Example:
      * This example shows decoding one byte at a time.
      * -----
      * auto encoded = Base64.encoder(cast(ubyte[])"0123456789");
@@ -1686,6 +1686,24 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
     Decoder!(Range) decoder(Range)(Range range) if (isInputRange!Range)
     {
         return typeof(return)(range);
+    }
+
+    /// ditto
+    Decoder!(const(ubyte)[]) decoder()(const(char)[] range)
+    {
+        import std.string : representation;
+        return typeof(return)(range.representation);
+    }
+
+    ///
+    @safe pure unittest
+    {
+        import std.algorithm.comparison : equal;
+        string encoded =
+            "VGhvdSBzaGFsdCBuZXZlciBjb250aW51ZSBhZnRlciBhc3NlcnRpbmcgbnVsbA==";
+
+        assert(Base64.decoder(encoded)
+            .equal("Thou shalt never continue after asserting null"));
     }
 
 


std/base64.d
```diff
@@ -1683,10 +1683,28 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
      * }
      * -----
      */
+    auto decoder()(const(char)[] range)
```
berni44, line 1686: Is there a reason, why the return type is `auto` and not `Decoder!(const(ubyte)[])`?

berni44 (commented):
Have a look at the [documentation](http://dtest.dlang.io/artifact/website-897ea29f92707d10b66a724326e039ed18f94b09-03e767b27dde5c62f37fbf7a9cef4780/web/phobos-prerelease/std_base64.html#.Base64Impl.decoder):

```
Parameters:
const(char)[] range	An input range over the data to be decoded.
```

That's a mix of the two overloads, which isn't correct anymore. I think, there are two ways out of it:
- Add separate documentation for the overload
- Add a template parameter `Range` and use template constraints

The examples look also strange (i.e. the headers: Example - Example - Examples). That's not your fault - the original examples should have been unittests. Anyway, with this PR it gets a mix and should be fixed.

Finally: When adding this, it would be nice to add it for `encoder` too, to keep the symmetry.

std/base64.d
dukc: The reason is that I initially used `.byCodeUnit`, and there was no good way to mark the return type. But now when it's `.representation` instead there's no good reason anymore. Will change.

dukc:
Okay, attempt made to improve on the documentation.

std/base64.d
```diff
@@ -1645,49 +1645,75 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
      *
      * Params:
      *  range = An $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
-     *      over the data to be decoded.
+     *      over the data to be decoded, or a character array. Will not accept
+     *      `wchar[]` nor `dchar[]`
```
thewilsonator, line 1649:
```
over the data to be decoded, or a `char` array
```

std/base64.d
```diff
@@ -1645,49 +1645,75 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
...
      *
      * Returns:
-     *  If $(D_PARAM range) is a range of characters, a `Decoder` that
+     *  If $(D_PARAM range) is a range or array of characters, a `Decoder` that
```
thewilsonator, line 1652:
```
or array of `char`
```

std/base64.d
```diff
@@ -1645,49 +1645,75 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
...
+    /// ditto
+    Decoder!(const(ubyte)[]) decoder()(const(char)[] range)
+    {
+        import std.string : representation;
+        return typeof(return)(range.representation);
+    }
+
+    //todo: convert these to real unittests
+    ///
+    version (StdDdoc) unittest
```
thewilsonator, line 1682:
```
std/base64.d(1682:23)[warn]: A unittest should be annotated with at least @safe or @system
--
  | std/base64.d(1694:23)[warn]: A unittest should be annotated with at least @safe or @system
```

std/base64.d
```diff
@@ -1645,49 +1645,75 @@ template Base64Impl(char Map62th, char Map63th, char Padding = '=')
...
+        import std.algorithm, std.stdio;
+
+        foreach (decoded; Base64.decoder(stdin.byLine()))
+        {
+            writeln(decoded);
+        }
+    }
+
+    ///
+    version (StdDdoc) unittest
```
thewilsonator, line 1694: ditto

thewilsonator (approved):

dukc:
I suspect buildKite is attempting to execute the unit tests that are compiled with `version=stdDdoc` and thus ends up waiting for input forever. Sigh - I think I have to either make them really tested or move them back to the documentation comment.

berni44:
> I suspect buildKite is attempting to execute the unit tests that are compiled with `version=stdDdoc` and thus ends up waiting for input forever. Sigh - I think I have to either make them really tested or move them back to the documentation comment.

Maybe they could be rewritten to not use `writeln` and `stdio` (I guess, that's what you mean with really tested)? As far as I know it is intentional, that the style tester complains here...

dukc:
Decided to move the first example back to the comment. Decided against rewriting, because 1: it'd be less practical example for the reader and 2: there's always the risk to lose the original intention of the example in the rewrite. The rewrite to a unit test should probably be done so that it still appears as `readln`/`writeln` in the docs, and I'm no test suite wizard. 

The second example should be no problem though, as it only prints, does not read from `stdio`.

berni44:
> The second example should be no problem though, as it only prints, does not read from `stdio`.

It garbles up the [output of the style checker](https://buildkite.com/dlang/phobos/builds/4909#b1cddd76-bda8-44ff-b9e0-2bc319f42235/274-987)... How about this:

```
    char[] result;
    auto encoded = Base64.encoder(cast(ubyte[])"0123456789");
    int witness = 0;
    foreach (n; map!q{a - '0'}(Base64.decoder(encoded)))
    {
        assert(n == witness++);
    }
```

Or, alternatively, because it wasn't the idea of this PR to improve the public unittests, I'd suggest to move it back into the docs and keep improving them for a different PR (if someone feels called) - let not the best be the enemy of the better...

berni44 (approved):

RazvanN7 (approved):

PR #7891, opened 2021-03-20
Rework documentation of std.format.spec : singleSpec

berni44:
I removed `fmt` from first sentence, because this sentence can be used in overview tables without context, so the reader has to guess what `fmt` is about.

I added a description of the template parameter, because I remember, that I was quite confused by template parameters, when I read the docs as a beginner. Such descriptions probably would have helped.

I added a bug, I found, because I think, listing bugs shows, that we care about bugs. And it motivates (at least me) to remove them. (Here, this won't be difficult, but I wanted to keep bug fixing and documentation separated.)

I changed the public unittest. IMHO, it should show, how `singleSpec` can be used, instead of examples how it throws, when wrongly used. I kept the old test though as a normal unittest and added a few more checks.

### DIFF
--- a/std/format/spec.d
+++ b/std/format/spec.d
@@ -797,17 +797,23 @@ if (is(Unqual!Char == Char))
 }
 
 /**
-Helper function that returns a `FormatSpec` for a single specifier given
-in `fmt`.
+Helper function that returns a `FormatSpec` for a single format specifier.
 
 Params:
-    fmt = A format specifier.
+    fmt = a $(MREF_ALTTEXT format string, std,format)
+          containing a single format specifier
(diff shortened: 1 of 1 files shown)


berni44:
In ddox the link to the description of the format string doesn't work. I'll have to find an other way to cope with this.

berni44:
I did not manage to create a link directly to the specs of format string, that works in ddoc and ddox. So I replaced it with a link to `std.format`. IMHO it would need a change in the dlang.org repositiory, but I'm not firm enough to do so, especially, because my local version of ddox does not create the links correctly, so I can't test the changes.

RazvanN7 (approved):

PR #7894, opened 2021-03-20
Fix Issue 21738 - std.format.spec: singleSpec should throw on "%%"

wolframw:
`singleSpec` doesn't work with "%%" since it doesn't really "do" anything. Therefore, `singleSpec` is changed to throw a `FormatException` if "%%" is passed as format specifier.

### DIFF
--- a/std/format/spec.d
+++ b/std/format/spec.d
@@ -842,9 +842,6 @@ Throws:
     A $(REF_ALTTEXT FormatException, FormatException, std,format) when the
     format string contains no format specifier or more than a single format
     specifier or when the format specifier is malformed.
-
-Known_bugs:
-    Does not throw on %%. ($(BUGZILLA 21738))
   */
 FormatSpec!Char singleSpec(Char)(Char[] fmt)
 {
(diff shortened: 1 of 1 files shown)


berni44 (approved):
You need to reword the commit message to start with "Fix issue 21738". So the bot will auto close the issue. All else is fine.

std/format/spec.d
```diff
@@ -806,8 +806,8 @@ Params:
 Returns:
     A `FormatSpec` with the specifier parsed.
 Throws:
-    A `FormatException` when more than one specifier is given or the specifier
-    is malformed.
+    A `FormatException` when more than one specifier is given, the specifier
+    is malformed, or when the specifier is "%%".
```
RazvanN7, line 810:
```suggestion
    is malformed or when the specifier is "%%".
```

berni44: With #7891 being merged, this isn't relevant anymore. There are more cases, where a `FormatException` can be thrown, so I used a more general term there and there is no need any more to write anything about `%%`.

But: When handling the merge conflicts, the "Known Bug" section should be removed.

wolframw: Thanks for the heads-up.

RazvanN7 (approved):

PR #7896, opened 2021-03-21
Rework documentation of std.format.package : sformat

berni44:
I removed `buf` from the first line, because this line might be used in other contexts, where `buf` is not known (for example in overview tables).

I switched the order of the two overloads, because in my opinion, the overload that resembles a function call in other languages (and without template constraints) is better understandable by new users and with the switch it's shown first in the docs. This is also the order of the examples below: First "normal" use, then check at compile time.

I removed the reference to `format()` in the comment about checks at compile time, because I couldn't find anything useful there, apart from a note, that it might produce faster code, which only applies to `format` but not to `sformat`.

I replaced the example, because I dislike unimaginative foo-bar examples and I wanted to put more emphasize on how the function has to be called, when checking at compile time is wished for.

### DIFF
--- a/std/format/package.d
+++ b/std/format/package.d
@@ -1853,29 +1853,39 @@ unittest
     assert(guessLength!char("%0*d") == 0);
 }
 
-/*****************************************************
- * Format arguments into buffer $(I buf) which must be large
- * enough to hold the result.
- *
- * Returns:
- *     The slice of `buf` containing the formatted string.
- *
- * Throws:
- *     A `RangeError` if `buf` isn't large enough to hold the
- *     formatted string.
- *
- *     A $(LREF FormatException) if the length of `args` is different
- *     than the number of format specifiers in `fmt`.
- */
-char[] sformat(alias fmt, Args...)(char[] buf, Args args)
-if (isSomeString!(typeof(fmt)))
-{
-    alias e = checkFormatException!(fmt, Args);
-    static assert(!e, e.msg);
-    return .sformat(buf, fmt, args);
-}
+/**
+Converts its arguments according to a format string into a buffer.
+The buffer has to be large enough to hold the formatted string.
 
-/// ditto
+The second version of `sformat` takes the format string as a template
+argument. In this case, it is checked for consistency at
+compile-time.
+
+Params:
+    buf = the buffer where the formatted string should go
+    fmt = a $(MREF_ALTTEXT format string, std,format)
+    args = a variadic list of arguments to be formatted
+    Char = character type of `fmt`
+    Args = a variadic list of types of the arguments
+
+Returns:
+    A slice of `buf` containing the formatted string.
+
+Throws:
+    A $(REF_ALTTEXT RangeError, RangeError, core, exception) if `buf`
+    isn't large enough to hold the formatted string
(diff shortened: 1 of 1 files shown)


std/format/package.d
```diff
@@ -1853,29 +1853,39 @@ unittest
...
-    alias e = checkFormatException!(fmt, Args);
-    static assert(!e, e.msg);
-    return .sformat(buf, fmt, args);
-}
+/**
+Converts its arguments according to a format string into a buffer.
+The buffer has to be large enough to hold the formatted string.
 
-/// ditto
+The second version of `sformat` takes the format string as template
```
RazvanN7, line 1860:
```suggestion
The second version of `sformat` takes the format string as a template
```

std/format/package.d
```diff
@@ -1853,29 +1853,39 @@ unittest
...
+argument. In this case, it is checked for consistency at
+compile-time.
+
+Params:
+    buf = the buffer, where the formatted result should go
```
RazvanN7, line 1865:
```suggestion
    buf = the buffer where the formatted string should go
```

std/format/package.d
```diff
@@ -1853,29 +1853,39 @@ unittest
...
+    A slice of `buf` containing the formatted string.
+
+Throws:
+    A $(REF_ALTTEXT RangeError, RangeError, core, exception) if `buf`
+    isn't large enough to hold the formatted string
+    and a $(LREF FormatException) if formatting did not succeed.
+
+Note:
+    In theory this function should be `@nogc`. But with the current
+    implementation there are some cases, where allocations occur:
```
RazvanN7, line 1881:
```suggestion
    implementation there are some cases where allocations occur:
```

std/format/package.d
```diff
@@ -1931,13 +1941,43 @@ char[] sformat(Char, Args...)(return scope char[] buf, scope const(Char)[] fmt,
...
+
+/// The format string can be checked at compile-time:
+@safe pure unittest
+{
+    char[20] buf;
+
+    assert(sformat!"Here are %d %s."(buf[], 3, "apples") == "Here are 3 apples.");
+
+    // This line doesn't compile, because 3.14 cannot be formatted with %d:
+    // writeln(sformat!"Here are %d %s."(buf[], 3.14, "apples"));
```
RazvanN7, line 1970: It looks kind of bad to have commented code in public examples. Will the unittest become `@system` if we replace this comment with: `static assert(!__traits(compiles, sformat!"Here are %d %s."(buf[], 3.14, "apples")));` ?

RazvanN7 (changes requested):

std/format/package.d
berni44: Using `__traits` is possible, but I dislike it. For a beginner, I think, it's much more comprehensible to have commented code, than a `__traits` statement. (Note, that I added this `__traits` statement in the next (nonpublic) unittest, to actually verify this statement.)

Whatever we decide upon here, we should do the same for the similar example of `format` (a PR that was already merged yesterday).

RazvanN7 (approved):

std/format/package.d
RazvanN7: Fair enough.

PR #7900, opened 2021-03-21
Improve some assertions in std.range

CyberShadow:


n8sh (approved):
Can merge as soon as style checker's objections about excessively long lines are addressed.

RazvanN7 (approved):

PR #7906, opened 2021-03-22
std.file docs: Tweak dirEntries examples

ntrel:
Use selective imports for `listdir` example, not fully qualified calls.
Don't use `args`, use current directory for working runnable example.

Move parallel dmd example into a documented unittest. Use `pattern` argument instead of filter & endsWith for simplicity. Note the `listdir` example already uses `map`.

Only run these 'tests' for `version(StdDdoc)`.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -4958,7 +4958,7 @@ version (StdDdoc)
 {
     string[] listdir(string pathname)
     {
-        import std.algorithm : filter, map;
+        import std.algorithm.iteration : filter, map;
         import std.array : array;
         import std.path : baseName;
 
--- a/std/file.d
+++ b/std/file.d
@@ -4940,19 +4940,6 @@ foreach (DirEntry e; dirEntries("dmd-testing", SpanMode.breadth))
     writeln(e.name, "\t", e.size);
 }
 
-// Iterate over all *.d files in current directory and all its subdirectories
-auto dFiles = dirEntries("", SpanMode.depth).filter!(f => f.name.endsWith(".d"));
-foreach (d; dFiles)
-    writeln(d.name);
-
-// Hook it up with std.parallelism to compile them all in parallel:
-foreach (d; parallel(dFiles, 1)) //passes by 1 file to each thread
-{
-    string cmd = "dmd -c "  ~ d.name;
-    writeln(cmd);
-    std.process.executeShell(cmd);
-}
-
 // Iterate over all D source files in current directory and all its
 // subdirectories
 auto dFiles = dirEntries("","*.{d,di}",SpanMode.depth);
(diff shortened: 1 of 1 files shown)


std/file.d
```diff
@@ -4966,28 +4953,49 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
...
-        import std.algorithm;
-        import std.array;
-        import std.file;
-        import std.path;
+        import std.algorithm.iteration : filter, map;
+        import std.array : array;
+        import std.path : baseName;
 
-        return std.file.dirEntries(pathname, SpanMode.shallow)
+        return dirEntries(pathname, SpanMode.shallow)
```
PetarKirov, line 4965:
```suggestion
        return pathname.dirEntries(SpanMode.shallow)
```

std/file.d
```diff
@@ -4966,28 +4953,49 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
...
             .filter!(a => a.isFile)
-            .map!(a => std.path.baseName(a.name))
+            .map!(a => baseName(a.name))
             .array;
     }
 
-    void main(string[] args)
-    {
+    void main() {
```
PetarKirov, line 4971: The style check fails because each curly brace must be on its own line.

std/file.d
```diff
@@ -4966,28 +4953,49 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
...
         import std.stdio;
 
-        string[] files = listdir(args[1]);
-        writefln("%s", files);
-     }
+        string[] files = listdir("");
```
PetarKirov, line 4974: 1. Could use UFCS
2. Using `"."` makes it more clear that we're listing the contents of the current directory

```suggestion
        string[] files = ".".listdir();
```

std/file.d
```diff
@@ -4966,28 +4953,49 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
...
+    import std.stdio;
+    import std.parallelism;
+    import std.process;
+
+    // Iterate over all *.d files in current directory and all its subdirectories
+    auto dFiles = dirEntries("", "*.d", SpanMode.depth);
+    foreach (d; dFiles)
+        writeln(d.name);
+
+    foreach (d; parallel(dFiles, 1)) // process each entry in a new thread
```
PetarKirov, line 4992:
```suggestion
    foreach (d; parallel(dFiles, 1)) // process each entry in a separate thread
```

`std.parallelism.parallel` uses a shared thread pool for the whole program, so there are at most `N+1` (IIRC) parallel invocations of `dmd`, where `N` is the number of CPU threads (e.g. 8, on a 4 core CPU with 2-way SMT). As was written before, the comment implied that there would be up to `dFiles.length` dmd processes in flight.

std/file.d
```diff
@@ -4966,28 +4953,49 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
...
+    {
+        // compile source file
+        string cmd = "dmd -c "  ~ d.name;
```
PetarKirov, line 4995:
```suggestion
        string cmd = "dmd -c " ~ d.name;
```

CyberShadow:
I would also like to suggest using UFCS in examples where appropriate, as it is closer to idiomatic D.

std/file.d
```diff
@@ -4970,24 +4957,45 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
...
+    import std.stdio;
+    import std.parallelism;
+    import std.process;
+
+    // Iterate over all *.d files in current directory and all its subdirectories
+    auto dFiles = dirEntries(".", "*.d", SpanMode.depth);
+    foreach (d; dFiles)
+        writeln(d.name);
+
+    foreach (d; parallel(dFiles, 1)) // process each entry in a separate thread
```
CyberShadow, line 4992:
```suggestion
    foreach (d; dFiles.parallel(1)) // process each entry in a separate thread
```

RazvanN7:
@ntrel I agree with the other reviewers, UFCS makes the code much nicer to read.

RazvanN7 (approved):
I'm gonna merge this as is and we can apply the required changes in a follow-up PR

std/file.d
```diff
@@ -4970,24 +4957,45 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
...
-     }
+        string[] files = listdir(".");
+        writeln(files);
+    }
+}
+
+/// Compile D files in parallel:
+version (StdDdoc)
+@system unittest
+{
```
berni44, line 4982: I think, this unittest introduced [issue 21796](https://issues.dlang.org/show_bug.cgi?id=21796).

See [logfile of circle-ci](https://app.circleci.com/pipelines/github/dlang/phobos/3976/workflows/90ccdf2d-ff37-4e70-b7da-ca80d626c604/jobs/13524).

std/file.d
```diff
@@ -4970,24 +4957,45 @@ auto dirEntries(string path, SpanMode mode, bool followSymlink = true)
...
+    auto dFiles = dirEntries(".", "*.d", SpanMode.depth);
+    foreach (d; dFiles)
+        writeln(d.name);
+
+    foreach (d; parallel(dFiles, 1)) // process each entry in a separate thread
+    {
+        // compile source file
+        string cmd = "dmd -c " ~ d.name;
+        writeln(cmd);
+        std.process.executeShell(cmd);
```
ibuclaw, line 4997: WHAT!

ibuclaw:
Hi @ntrel - This PR caused a [regression](https://issues.dlang.org/show_bug.cgi?id=21796), and has been [reverted](https://github.com/dlang/phobos/pull/8009).  Please feel free to resend a version of this PR without the regressing part.  Thanks!

PR #7907, opened 2021-03-23
std.format: Add %f for x87-reals

berni44:
The main problem here is, that the algorithms are designed for mantissas that are at max 60 bit long (maximum that fits in an ulong, because we need 4 bits for carry), but the mantiassa of x87 reals has 63 bit. So we need some extra code to get around these 3 extra bits.

The second added unittest contains a few corner cases where the new code is needed. It also covers the bit extraction at CTFE which could not be covered in #7862.

I ran more than 20 million random external tests comparing the output of `snprintf` with this implementation, without any difference reported. (That took several hours. Formatting reals is rather slow (with `snprintf` as well as with this implementation), because of the large output.)

### DIFF
--- a/std/format/internal/floats.d
+++ b/std/format/internal/floats.d
@@ -130,7 +130,10 @@ if (is(T == float) || is(T == double)
     }
 
     if (T.mant_dig == 64)
-        assert(false); // not yet implemented
+    {
+        if (f.spec != 'f' && f.spec != 'F')
+            assert(false); // not yet implemented
+    }
 
     final switch (f.spec)
     {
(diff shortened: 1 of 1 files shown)


thewilsonator:
Ping me when this is good (or you get stuck)

berni44:
@thewilsonator IMHO, this is good to go now. I had to fight with two bugs at the same time:

- Issue 21757: Logarithm not working correctly on some computers with reals in CTFE
- Issue 20972: Wrong real literals on some windows computers

I had to comment out some unittests to get around them. While the second one doesn't hurt much here, we have to keep the first one in mind when going life to avoid using CTFE with these computers.

RazvanN7 (approved):

PR #7908, opened 2021-03-24
std.experimental.checkedint opBinaryRight with integer left-hand side does not compile for any operators except + and -

n8sh:


thewilsonator (approved):
does this have an issue number?

PR #7909, opened 2021-03-24
checkedint.Abort and checkedint.Warn should be `@safe`

n8sh:


std/experimental/checkedint.d
```diff
@@ -1471,7 +1471,7 @@ default behavior.
 */
 struct Warn
 {
-    import std.stdio : stderr;
+    import std.stdio : writefln;
```
thewilsonator, line 1474: unrelated but shouldn't this be private imports?

thewilsonator (approved):

std/experimental/checkedint.d
n8sh: In current version of DMD writing `private import` is no longer necessary because imports are private by default.

PR #7910, opened 2021-03-24
std.experimental.checkedint.Checked is not compatible with "%d" and "%x" integer format specifiers

n8sh:
It might make sense to support something like `hookToString` but that can be done later.


thewilsonator (approved):

thewilsonator:
please rebase

PR #7911, opened 2021-03-24
Rework documentation of std.format.write : formatValue

berni44:
A lot of the former description and most of the unittests are of interest for more than just the "plumping" `formatValue`. So I moved this to the package description, to avoid loss, (but left all unchanged for better reviewability and because I plan to rework the package description anyway, when I'm through with all functions).

### DIFF
--- a/std/format/package.d
+++ b/std/format/package.d
@@ -360,6 +360,44 @@ My friends are "John", "Nancy".
 My friends are John, Nancy.
 )
 
+   Aggregates:
+   `struct`, `union`, `class`, and `interface` are formatted by calling `toString`.
+
+   `toString` should have one of the following signatures:
+
+   ---
+   void toString(Writer, Char)(ref Writer w, scope const ref FormatSpec!Char fmt)
+   void toString(Writer)(ref Writer w)
+   string toString();
+   ---
+
+   Where `Writer` is an $(REF_ALTTEXT output range, isOutputRange, std,range,primitives)
+   which accepts characters. The template type does not have to be called `Writer`.
+
+   The following overloads are also accepted for legacy reasons or for use in virtual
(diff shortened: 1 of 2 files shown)


std/format/package.d
```diff
@@ -360,6 +360,44 @@ My friends are "John", "Nancy".
...
+   void toString(Writer, Char)(ref Writer w, scope const ref FormatSpec!Char fmt)
+   void toString(Writer)(ref Writer w)
+   string toString();
+   ---
+
+   Where `Writer` is an $(REF_ALTTEXT output range, isOutputRange, std,range,primitives)
+   which accepts characters. The template type does not have to be called `Writer`.
+
+   The following overloads are also accepted for legacy reasons or for use in virtual
+   functions. It's recommended that any new code forgo these overloads if possible for
```
etienne02, line 378:
```suggestion
   functions. It's recommended that any new code forgot these overloads if possible for
```

std/format/package.d
```diff
@@ -383,6 +421,300 @@ import std.range.primitives : isInputRange;
...
+{
+    import std.array : appender;
+    import std.format.spec : singleSpec;
+
+    auto w = appender!string();
+    auto spec = singleSpec("%s");
+    char[2] two = ['a', 'b'];
+    formatValue(w, two, spec);
+
+    assert(w.data == "ab");
```
etienne02, line 520: Shouldn’t be `assert(w.data == "[a, b]");` here if it’s like dynamic arrays ?

berni44: These are both examples that I just copy pasted from one place to the other. I'd like to take care of them in a later PR and don't apply any changes here. IMHO, there is a lot more to do... This PR is mainly about the docs of `formatValue` - see my comment above.

RazvanN7 (approved):

PR #7913, opened 2021-03-24
std.format: Tweak format string docs

ntrel:
Add grammar links and headings.
Add example for positional format spec, move existing paragraph.
Document `%u` format char.
Make array examples runnable.

### DIFF
--- a/std/format/package.d
+++ b/std/format/package.d
@@ -45,10 +45,9 @@ $(TR $(TH Function Name) $(TH Description)
    $(REF_ALTTEXT $(D unformatValue), unformatValue, std, format, read) are
    used for the plumbing.
 
-   Format_String:
+   $(ADEF format-string)$(H2 Format strings)
 
-   <a name="format-string">$(I Format strings)</a>
-   consist of characters interspersed with $(I format
+   Format strings consist of characters interspersed with $(I format
    specifications). Characters are simply copied to the output (such
    as putc) after any necessary conversion to the corresponding UTF-8
    sequence.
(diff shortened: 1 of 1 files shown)


berni44:
While I'm normally quite happy, when people try to improve Phobos, for me, this PR is a little bit inconvenient. Since about a month I'm reworking `std.format` completely. I started with splitting the module in several sub modules, refactored most of it and since last weekend I'm reworking the docs. There are four functions left:  `formatValue` is in review state (#7911), `formattedWrite` in preparation, the other two (`unformatValue` and `formattedRead` I will probably postpone a little bit, because I need to better understand the read part which I treated up to now a stepmotherly, before doing so.

After that I planned to rework the module docs, like this PR does, but on a much more thorough base (there is much more undocumented stuff and there are more things, that need to be corrected). A lot of this rework, I have already done locally (weeks ago), but not added yet, because I can't do everything at once and always have to wait until the changes are reviewed.

So forgive me, asking you to revoke this PR. Would you mind me?

RazvanN7:
@ntrel What are your thoughts on @berni44 's comments?

RazvanN7:
ping @ntrel

ntrel:
If @berni44 has incompatible changes then it's ok with me to close this. I'd like to know when they're submitted. BTW this has ddoc problem anyway where `$(DL` is not expanded for some reason. I couldn't work out why.

berni44:
> I'd like to know when they're submitted.

Partly they are already submitted see #7919. The followup #7935 is currently in review process, and when this is merged, I think, there will about three or four more to follow until the docs are completed. My goal is to get this finished before next stable is branched off. Looks like we are well in our schedule.

RazvanN7:
It seems like @berni44 is on top of this. @ntrel thanks anyway for your work and sorry for not being able to merge this.

PR #7914, opened 2021-03-24
make std.experimental.checkedint.Checked!T.toHash callable when Checked!T is shared

n8sh:


std/experimental/checkedint.d
```diff
@@ -562,6 +562,44 @@ if (isIntegral!T || is(T == Checked!(U, H), U, H))
         }
     }
 
+    /// ditto
+    size_t toHash(this This)() const nothrow @safe if (is(This == shared))
```
n8sh, line 566: This signature is because when I used

```d
size_t toHash()() const shared nothrow @safe
```
I got this error:
>Error: `std.experimental.checkedint.Checked!(int, Hook1).Checked.toHash` called with argument types `() immutable` matches both:
>`std.experimental.checkedint.Checked!(int, Hook1).Checked.toHash() const`
>and:
>`std.experimental.checkedint.Checked!(int, Hook1).Checked.toHash!().toHash() shared const`

I don't think that should have been an error since non-template functions are supposed to always match before template functions.

RazvanN7: Wouldn't it be better if you merged both toHash functions into a single templated this one and avoid the code duplication?

RazvanN7 (changes requested):
Merge the `toHash` functions.

RazvanN7:
ping @n8sh

std/experimental/checkedint.d
n8sh: @RazvanN7 I need to retain a non-template version of `toHash` or else `TypeInfo.getHash` won't pick it up. Demonstration:

https://run.dlang.io/is/gyqq5g
```d
import std.stdio: writeln;

struct S1
{
    size_t toHash() const nothrow @safe { return 44; }
}

struct S2
{
    size_t toHash(this _)() const nothrow @safe { return 44; }
}

void main()
{
    S1 s1;
    S2 s2;
    writeln(s1.toHash()); // 44
    writeln(typeid(s1).getHash(&s1)); // 44
    writeln(s2.toHash()); // 44
    writeln(typeid(s2).getHash(&s2)); // 1364076727
}
```

n8sh:
>Merge the `toHash` functions.

@RazvanN7 keeping a top-level non-template version of `toHash` is necessary for `TypeInfo.getHash`. `TypeInfo.getHash` is used for built-in associative array keys so I believe it is critical. I could move common functionality into a mixin:

```d
    size_t toHash() const nothrow @safe
    {
        return mixin(hashExpr);
    }

    /// ditto
    size_t toHash(this _)() shared const nothrow @safe
    {
        import core.atomic : atomicLoad, MemoryOrder;
        static if (is(typeof(this.payload.atomicLoad!(MemoryOrder.acq)) P))
        {
            auto payload = __ctfe ? cast(P) this.payload
                                  : this.payload.atomicLoad!(MemoryOrder.acq);
        }
        else
        {
            alias payload = this.payload;
        }
        return mixin(hashExpr);
    }

    private
    {
        static if (hasMember!(Hook, "hookToHash"))
            enum hashExpr = q{hook.hookToHash(payload)};
        else static if (stateSize!Hook > 0)
        {
            static if (hasMember!(typeof(payload), "toHash"))
                enum hashExpr = q{payload.toHash() ^ hashOf(hook)};
            else
                enum hashExpr = q{hashOf(payload) ^ hashOf(hook)};
        }
        else static if (hasMember!(typeof(payload), "toHash"))
            enum hashExpr = q{payload.toHash()};
        else
            enum hashExpr = q{.hashOf(payload)};
    }
```

or would you prefer another approach?

std/experimental/checkedint.d
RazvanN7: Darn... OK, this is good then.

RazvanN7 (approved):

RazvanN7:
Thanks, @n8sh !

PR #7915, opened 2021-03-24
Rework documentation of std.format.write : formattedWrite

berni44:
I moved a public unittest, that's about grouping to package, because I think, that's something, that is relevant for other functions of `std.format` to. It even does not use `formattedWrite` directly. I will cope with that later, when reworking the package docs.

I renamed the template parameter `A` to `Args` for consistency with `format` and `sformat`. While at other places it is common to use single letters as template parameters, throughout `std.format` they are descriptive.

I changed the order of the two overloads, like I did with `format` and `sformat` because I think, for beginners the one without template parameters is better comprehensible. For the same reason I did not use idiomatic D for the first example.

The output range is named a writer here, I guess that is for historic reasons. In the docs I named it an output range. I think, users will immediately know, what an output range should be able to do, while a writer might not be so clear.

The description of the return value is currently not correct. ~~First of all, I can't see any good use for this. But as it's there, at least it should give something useful. The current behavior would probably be best described as: The largest positional argument used if positional arguments appear and the number of arguments formatted else. I suggest to use the description I provided, file a bug report and change this.~~

### DIFF
--- a/std/format/write.d
+++ b/std/format/write.d
@@ -20,76 +20,37 @@ import std.format.internal.write;
 import std.format.spec : FormatSpec;
 import std.traits : isSomeString;
 
-/**********************************************************************
-   Interprets variadic argument list `args`, formats them according
-   to `fmt`, and sends the resulting characters to `w`. The
-   encoding of the output is the same as `Char`. The type `Writer`
-   must satisfy $(D $(REF isOutputRange, std,range,primitives)!(Writer, Char)).
+/**
+Converts its arguments according to a format string and writes
+the result to an output range.
 
-   The variadic arguments are normally consumed in order. POSIX-style
-   $(HTTP opengroup.org/onlinepubs/009695399/functions/printf.html,
-   positional parameter syntax) is also supported. Each argument is
-   formatted into a sequence of chars according to the format
-   specification, and the characters are passed to `w`. As many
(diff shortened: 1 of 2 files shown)


berni44:
I noticed, that the return value is used by `format` for deciding, if an Exception is thrown. So I reworded it's description.

std/format/write.d
```diff
@@ -20,76 +20,37 @@ import std.format.internal.write;
...
-   via the grouping format specifiers $(B %&#40;) and $(B %&#41;). Each
-   matching pair of $(B %&#40;) and $(B %&#41;) corresponds with a single array
-   argument. The enclosed sub-format string is applied to individual array
-   elements.  The trailing portion of the sub-format string following the
-   conversion specifier for the array element is interpreted as the array
-   delimiter, and is therefore omitted following the last array element. The
-   $(B %|) specifier may be used to explicitly indicate the start of the
-   delimiter, so that the preceding portion of the string will be included
-   following the last array element.  (See below for explicit examples.)
-
```
RazvanN7, line 50: Isn't this information relevant? Why is it deleted?

berni44: I thought I moved it to module scope. Must have made a mistake here.

RazvanN7 (approved):

berni44:
removed trailing spaces

PR #7916, opened 2021-03-24
Fix Issue 17381 - Checked format string is permissive after floating point argument

berni44:
On my computer there is no failure, but in the [Online D Editor](https://run.dlang.io/is/YrEFFp) it reports failures since 2.074.1, but worked before.

I just added a unittest to find out on which computers this fails.

### DIFF
--- a/std/format/internal/floats.d
+++ b/std/format/internal/floats.d
@@ -16,8 +16,6 @@ module std.format.internal.floats;
 
 import std.format.spec : FormatSpec;
 
-package(std.format) enum ctfpMessage = "Cannot format reals at compile-time.";
-
 package(std.format) enum RoundingMode { up, down, toZero, toNearestTiesToEven, toNearestTiesAwayFromZero }
 
 package(std.format) auto printFloat(T, Char)(return char[] buf, T val, FormatSpec!Char f,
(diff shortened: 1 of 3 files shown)


berni44:
Strange. Seems to work on all computers here but fails in the online editor...

berni44:
Meanwhile I found out, that the error was triggered by formatting floats not being CTFEable. Which isn't the case any more (at least for `float` and `double`), and hence it is correct that the unittest passes on all testers.

berni44:
The problem was the special handling of the ctfp-exception in `checkFormatException`, which elided the check if all arguments were used.

For fixing this, I used a trick: Formatting NaNs doesn't use `snprintf`, so I allowed them being formatted instead of throwing on all `reals`. `checkFormatException` uses `Args.init` and hence now always works with floats in CTFE.

PetarKirov:
The online editor only provides already released  compiler versions, so any changes done in a pull request won't be visible in the editor, until a new version is released.

That said, as one of the maintainers of @dlang-tour, I am in favor of adding such feature, but it's not a small amount of work.

berni44:
> The online editor only provides already released compiler versions, so any changes done in a pull request won't be visible in the editor, until a new version is released.

You can choose "dmd-nightly", which sounds like it should be compiled once a day from the github sources. But it's 2.091 currently - see [bugreport](https://issues.dlang.org/show_bug.cgi?id=21773).

And when choosing "all dmd compilers", it stops with 2.094.

RazvanN7 (approved):

berni44:
Rebased to restart the tests.

PR #7918, opened 2021-03-25
Fix Issue 15888 - std.format should not produce deprecated hexstrings

berni44:


aG0aep6G (changes requested):
Printing "\uXYZW" for a wchar 0xXYZW is not correct. "\u...." refers to a code point. A wchar is a code *unit*.

Example:
```d
void main()
{
    import std.stdio;
    writeln(cast(wchar[]) [0xD83D, 0xDE08]); /* prints "😈" */
    writeln("\uD83D\uDE08"); /* Doesn't even compile. */
}
```

aG0aep6G:
@thewilsonator, please don't merge this yet. And in general, please, please, please give others a chance to look at a PR before merging it.

berni44:
> Printing "\uXYZW" for a wchar 0xXYZW is not correct. "\u...." refers to a code point. A wchar is a code _unit_.
> 
> Example:
> 
> ```d
> void main()
> {
>     import std.stdio;
>     writeln(cast(wchar[]) [0xD83D, 0xDE08]); /* prints "😈" */
>     writeln("\uD83D\uDE08"); /* Doesn't even compile. */
> }
> ```

So what is it supposed to write? IMHO the examples you gave in the issue aren't valid either. (One is still a hexstring, while the other one doesn't compile...)

aG0aep6G:
> So what is it supposed to write? IMHO the examples you gave in the issue aren't valid either. (One is still a hexstring, while the other one doesn't compile...)

How about `[cast(dchar) 0xFFFFFFF, cast(dchar) 0x42]` or `cast(dchar[]) [0xFFFFFFF, 0x42]`? I'm not sure if there's a meaningful difference between the two.

Just throwing an exception could also be an option.

berni44:
> > So what is it supposed to write? IMHO the examples you gave in the issue aren't valid either. (One is still a hexstring, while the other one doesn't compile...)
> 
> How about `[cast(dchar) 0xFFFFFFF, cast(dchar) 0x42]` or `cast(dchar[]) [0xFFFFFFF, 0x42]`? I'm not sure if there's a meaningful difference between the two.
> 
> Just throwing an exception could also be an option.

OK, let's try this one. Thanks for that idea and thanks for noting in time, that my first try was a bad idea...

aG0aep6G (approved):

RazvanN7 (approved):

PR #7920, opened 2021-03-25
Deprecate public access of enforceValidFormatSpec

berni44:
That is the third (and last) item in `std.format` that has been made public without any good reason. It's probably only used in `formatValueImpl`.

I did not manage to find a useful workaround here, but IMHO there is no real use case either.

### DIFF
--- a/changelog/deprecate_enforceValidFormatSpec.dd
+++ b/changelog/deprecate_enforceValidFormatSpec.dd
@@ -1,0 +1,4 @@
+Deprecate std.format : enforceValidFormatSpec
+
+`enforceValidFormatSpec` from `std.format` has been accidentally made
+public and will be removed from public view in 2.107.0.
--- a/std/format/spec.d
+++ b/std/format/spec.d
@@ -902,22 +902,13 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
     assertThrown!FormatException(singleSpec("%%"));
 }
 
+// @@@DEPRECATED_[2.107.0]@@@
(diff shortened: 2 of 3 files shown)


std/format/spec.d
```diff
@@ -902,22 +902,13 @@ FormatSpec!Char singleSpec(Char)(Char[] fmt)
     assertThrown!FormatException(singleSpec("%%"));
 }
 
+// @@@DEPRECATED_[2.107.0]@@@
+deprecated("enforceValidFormatSpec was accidentally made public and will be removed in 2.107.0")
 void enforceValidFormatSpec(T, Char)(scope const ref FormatSpec!Char f)
```
MoonlightSentinel, line 907: Maybe replace the entire function with a deprecated public alias?

berni44: The deprecation message would then talk about an alias instead of a function. Don't know if this makes a difference...

changelog/deprecate_enforceValidFormatSpec.dd
```diff
@@ -0,0 +1,4 @@
+Deprecate std.format : enforceValidFormatSpec
+
+`enforceValidFormatSpec` from `std.format` has been accidentally made
+public and will be removed from public view in 2.107.0.
```
RazvanN7, line 4: Is there an alternative to using `enforceValidFormatSpec`? Probably nobody is using it, but just in case.

berni44: No, I don't think so - as I wrote above: "I did not manage to find a useful workaround here, but IMHO there is no real use case either."

RazvanN7 (approved):

PR #7922, opened 2021-03-26
Revert "Fix Issue 21731 - SumType should provide convenient access to the typ…"

atilaneves:
Reverts dlang/phobos#7886

### DIFF
--- a/std/sumtype.d
+++ b/std/sumtype.d
@@ -790,18 +790,6 @@ public:
             return this.match!hashOf;
         }
     }
-
-    /**
-     * Returns the index of the type of the `SumType`'s current value in the
-     * `SumType`'s `Types`.
-     *
-     * If the `SumType` is qualified, returns the index of the type in `Types`
-     * whose qualified version matches the `SumType`'s current value.
-     */
-    size_t typeIndex() const
-    {
-        return tag;
-    }
 }
 
 // Construction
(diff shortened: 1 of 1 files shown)


n8sh (approved):
Approving and marking for immediate merging since the submitter is the one who has to approve all new public symbols and he does not appear to have approved the PR being reverted.

Geod24:
> Approving and marking for immediate merging since the submitter is the one who has to approve all new public symbols and he does not appear to have approved the PR being reverted.

It would have been nice to have an explanation as to why this is being reverted though. Any kind of context you might have right now will not be known to other people 6 months from now.

atilaneves:
> > Approving and marking for immediate merging since the submitter is the one who has to approve all new public symbols and he does not appear to have approved the PR being reverted.
> 
> It would have been nice to have an explanation as to why this is being reverted though. Any kind of context you might have right now will not be known to other people 6 months from now.

The rationale is that APIs should prevent misuse, and not going through `match` is IMHO not the way to use a sum type. *All* cases should be handled, every time.

pbackus:
> The rationale is that APIs should prevent misuse, and not going through match is IMHO not the way to use a sum type. All cases should be handled, every time.

I agree with this, but do not understand how it justifies excluding `typeIndex`, since`typeIndex` does not allow the user to bypass `match` or avoid handling all cases. Is some other kind of misuse you believe it would enable?

atilaneves:
> > The rationale is that APIs should prevent misuse, and not going through match is IMHO not the way to use a sum type. All cases should be handled, every time.
> 
> I agree with this, but do not understand how it justifies excluding `typeIndex`, since`typeIndex` does not allow the user to bypass `match` or avoid handling all cases. Is some other kind of misuse you believe it would enable?

Good point. I still don't see why this would be useful though, and after asking in the forum I didn't see any compelling answers.

pbackus:
> Good point. I still don't see why this would be useful though, and after asking in the forum I didn't see any compelling answers.

The main answer given on the forums was [serialization](https://forum.dlang.org/post/hmdsjejgtiyqwcxhwvae@forum.dlang.org). For example, having access to `typeIndex` makes it easy to serialize

```d
SumType!(int, string)("hello")
```

to

```json
{ "type": 1, "value": "hello" }
```

Of course, you could equally well choose a different schema for your serialized data; for example

1. `{ "type": "string", "value": "hello" }`; or even just
2. `"hello"`

...but these have their own disadvantages. (1) requires you to define a mapping between types and strings, which is non-trivial in the general case, and (2) makes it impossible to distinguish a plain `string` from a `SumType` in the serialized data. I wouldn't go so far as to say that the `typeIndex` approach is the *best* one, but there are legitimate reasons one might prefer it over the others.

PR #7923, opened 2021-03-28
std.typecons.RefCounted, std.container.array.Array, & similar structs that manage their own memory do not need to be scanned unless GC-allocated memory is reachable through them

n8sh:


PetarKirov:
I suggest using UDA as the API for type author to override the default policy:

```D
@(GC.ShouldScan.no)
struct MyType
{
    // ...
}
```

RazvanN7:
@n8sh Please add the word "Fix" at the beginning of your commit message so that the bot understands that you are actually fixing the issue.

PetarKirov:
BTW, we should check if hasIndirections can be optimized via `https://dlang.org/spec/traits.html#getPointerBitmap`, as IIRC @rainers added this trait, because the equivalent recursive template implementation was way too slow to compile. Also we may consider integratimg the handling of `GC.ShouldScan` UDA inside `getPointerBitmap`.

RazvanN7:
ping @n8sh

n8sh:
@RazvanN7 recently I was traveling. Updated commit message as requested.

Imperatorn:
Would be nice with a short description what it does and why it's needed. Like pseudo 

Why? Improve performance... etc
How? Annotate types with uda to enable... etc

n8sh:
> I suggest using UDA as the API for type author to override the default policy:
> 
> ```d
> import core.memory : GC;
> 
> @(GC.ShouldScan.no)
> struct MyType
> {
>     // ...
> }
> ```

@PetarKirov Is there a way to do this conditionally from inside the definition of the struct?

n8sh:
> Would be nice with a short description what it does and why it's needed.

@Imperatorn did you see the comment above `shouldGCScan`?

>Code that manually allocates memory without using the GC is responsible for calling GC.addRange if there is the possibility the memory might contain a pointer through which GC-allocated memory is reachable. Conservatively determining this via std.traits.hasIndirections can lead to false positives when nesting multiple types that manage their own memory (for instance a refcounted array of refcounted items). This template addresses that problem by acting like std.traits.hasIndirections but adding an escape hatch: any struct or union that defines `enum bool __GC_NO_SCAN = true` is regarded as having no pointers that can be used to reach GC-allocated memory. Names beginning with two underscores are reserved for use by the language so there is not a danger of conflicting with an enum that is identically-named by coincidence.

Bear in mind that this template is not public so the explanation is written for Phobos developers.

thewilsonator:
>  Is there a way to do this conditionally from inside the definition of the struct?

If `MyType` is a template you can do 

```d
template Mytype(T)
{
    alias whatever = logic goes here;
    @whatever
    struct MyType { ... }
}
```
not sure about for a non-template

RazvanN7:
@n8sh is it possible that this could fix this issue: https://github.com/dlang-community/libdparse/pull/387 ?

PR #7924, opened 2021-03-28
Add empty submodules to std.math.

berni44:
Since splitting out a whole has not worked out well here (see #7887), I'd like to try a different approach: This PR provides only a skeleton, that needs to be filled. This way it is possible to move the unproblematic functions and care about the problematic ones separately.

I tried to find good names for the subpackages, by going through math packages of other languages. The rough idea was to use the table at the top of `std.math`s docs as a guide. I add `integral` though, which is a new submodule. It should contain integer maths, like `isqrt` or `isSquare` or such things. IMHO this is something, that Phobos is missing.

### DIFF
--- a/std/math/floats.d
+++ b/std/math/floats.d
@@ -1,0 +1,16 @@
+// Written in the D programming language.
+
+/**
+This package is currently in a nascent state and may be subject to
+change. Please do not use it yet, but stick to $(MREF std, math).
+
+Copyright: Copyright The D Language Foundation 2000 - 2011.
+License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+Authors:   $(HTTP digitalmars.com, Walter Bright), Don Clugston,
+           Conversion of CEPHES math library to D by Iain Buclaw and David Nadlinger
+Source: $(PHOBOSSRC std/math/floats.d)
+ */
+
+module std.math.floats;
+
+// Will contain functions like feqrel, nextDown, isClose
--- a/std/math/hardware.d
+++ b/std/math/hardware.d
@@ -1,0 +1,16 @@
+// Written in the D programming language.
+
+/**
+This package is currently in a nascent state and may be subject to
+change. Please do not use it yet, but stick to $(MREF std, math).
+
+Copyright: Copyright The D Language Foundation 2000 - 2011.
+License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+Authors:   $(HTTP digitalmars.com, Walter Bright), Don Clugston,
+           Conversion of CEPHES math library to D by Iain Buclaw and David Nadlinger
+Source: $(PHOBOSSRC std/math/hardware.d)
+ */
+
+module std.math.hardware;
+
+// Will contain IeeeFlags and FloatingPointControl
--- a/posix.mak
+++ b/posix.mak
@@ -234,7 +234,8 @@ PACKAGE_std_experimental_allocator_building_blocks = \
   kernighan_ritchie null_allocator package quantizer \
   region scoped_allocator segregator stats_collector
 PACKAGE_std_format = package read spec write $(addprefix internal/, floats read write)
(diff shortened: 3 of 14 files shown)


std/math/hardware.d
```diff
@@ -0,0 +1,23 @@
+// Written in the D programming language.
+
+/**
+This package is currently in a nascent state and may be subject to
+change. Please do not use it yet, but stick to $(MREF std, math).
+
+Copyright: Copyright The D Language Foundation 2000 - 2011.
+           D implementations of tan, atan, atan2, exp, expm1, exp2, log, log10, log1p,
+           log2, floor, ceil and lrint functions are based on the CEPHES math library,
+           which is Copyright (C) 2001 Stephen L. Moshier $(LT)steve@moshier.net$(GT)
```
ibuclaw, line 10: We don't need a reference to the Cephes library in modules where these functions don't even exist.

i.e: Mention tan, atan, atan2 only in `std.math.trigonometry`; floor, ceil, lrint only in `std.math.rounding`, etc.

std/math/reminder.d
```diff
@@ -0,0 +1,23 @@
...
+           the following terms:
+License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+Authors:   $(HTTP digitalmars.com, Walter Bright), Don Clugston,
+           Conversion of CEPHES math library to D by Iain Buclaw and David Nadlinger
+Source: $(PHOBOSSRC std/math/reminder.d)
+ */
+
+module std.math.reminder;
+
+// Will contain functions like fmod and remainder
```
dkorpel, line 23: The file is named 'reminder' instead of 'remainder'

berni44: Thanks for noting this.

std/math/hardware.d
berni44: When you told me yesterday at beerconf, I was too tired to change this immediately. Now a new version is online.

ibuclaw: That's fine, I thought I'd just write a comment for posterity though.

std/math/floats.d
```diff
@@ -0,0 +1,16 @@
...
+change. Please do not use it yet, but stick to $(MREF std, math).
+
+Copyright: Copyright The D Language Foundation 2000 - 2011.
+License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+Authors:   $(HTTP digitalmars.com, Walter Bright), Don Clugston,
+           Conversion of CEPHES math library to D by Iain Buclaw and David Nadlinger
+Source: $(PHOBOSSRC std/math/floats.d)
+ */
+
+module std.math.floats;
```
ibuclaw, line 14: Don't do anything yet, but I wonder if this could be broken down into two separate modules as well i.e: `std.math.comparison` for `feqrel`, `isClose`, ... and `std.math.floatmanip` (for want of a better name) for `nextDown`, `ilogb`, ...

ibuclaw (commented):
Seems reasonable to me, but ping @atilaneves.

std/math/floats.d
berni44: Sounds like a good idea. I'll wait for other comments on this.

berni44: An other approach might be to put the whole floating point stuff (this,  the hardware stuff, and maybe even `CustomFloat` from `std.numeric`) into a separate package, for example `std.floats`, with own submodules. In the end, the focus of this stuff is not on math, but on floating points.

atilaneves (approved):

ibuclaw:
DMD will now check all functions in the `std.math` package for possible intrinsics(dlang/dmd#12314), so we can begin moving declarations without impediment.

std/math/floats.d
berni44: Wrote a [forum post](https://forum.dlang.org/post/atovzoedfzlvujfvpvkb@forum.dlang.org) asking this. Let's see what happens.

RazvanN7 (approved):

PR #7925, opened 2021-03-29
Rework documentation of std.format.read : formattedRead

berni44:
I renamed the template parameters `R` and `S` to `Range` and `Args` for consistency with other templates used in `std.format`.

I changed the order of the two overloads, like I did with `format` and `sformat` because I think, for beginners the one without template parameters is better comprehensible. For the same reason I did not use idiomatic D for the first example.

### DIFF
--- a/std/format/read.d
+++ b/std/format/read.d
@@ -20,35 +20,36 @@ import std.format.internal.read;
 import std.traits : isSomeString;
 
 /**
-Reads characters from $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
-`r`, converts them according to `fmt`, and writes them to `args`.
+Reads an input range according to a format string and stores the read
+values into its arguments.
+
+The second version of `formattedRead` takes the format string as
+template argument. In this case, it is checked for consistency at
+compile-time.
 
 Params:
-    r = The range to read from.
-    fmt = The format of the data to read.
-    args = The drain of the data read.
+    r = an $(REF_ALTTEXT input range, isInputRange, std, range, primitives),
+        where the formatted input is read from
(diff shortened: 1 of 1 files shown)


std/format/read.d
```diff
@@ -20,43 +20,44 @@ import std.format.internal.read;
 import std.traits : isSomeString;
 
 /**
-Reads characters from $(REF_ALTTEXT input range, isInputRange, std,range,primitives)
-`r`, converts them according to `fmt`, and writes them to `args`.
+Reads an input range according to a format string and stores the read
```
RazvanN7, line 23: I suggest keeping the link the `isInputRange` function.

std/format/read.d
```diff
@@ -20,43 +20,44 @@ import std.format.internal.read;
...
+values into its arguments.
+
+The second version of `formattedRead` takes the format string as
+template argument. In this case, it is checked for consistency at
+compile-time.
```
RazvanN7, line 28: Why not put this information as a comment directly above the second function?

berni44: Is this possible? I'll have a look...

berni44: I can't see, how this works out well - maybe I do not know something I should. With my attempts, the comment is either not shown at all or the two overloads are split and I need to duplicate the whole docs... (And we did it this way with `format`, `sformat` and so on too.)

std/format/read.d
RazvanN7: Ok.

std/format/read.d
RazvanN7: Hmm... I see. I guess this is fine too.

RazvanN7 (approved):

PR #7926, opened 2021-03-29
Rework documentation of std.format.read : unformatValue

berni44:
I moved the original public unittests to module scope, because I think they apply to `formattedRead` as well. I just copied them without change for better reviewability and will take care of the module docs in a separate PR once this PR (and #7925) are merged.

### DIFF
--- a/std/format/read.d
+++ b/std/format/read.d
@@ -15,6 +15,117 @@
  */
 module std.format.read;
 
+/// Booleans
+@safe pure unittest
+{
+    import std.format.spec : singleSpec;
+
+    auto str = "false";
+    auto spec = singleSpec("%s");
+    assert(unformatValue!bool(str, spec) == false);
+
+    str = "1";
+    spec = singleSpec("%d");
+    assert(unformatValue!bool(str, spec));
+}
+
+/// Null values
+@safe pure unittest
+{
+    import std.format.spec : singleSpec;
+
+    auto str = "null";
+    auto spec = singleSpec("%s");
+    assert(str.unformatValue!(typeof(null))(spec) == null);
+}
+
+/// Integrals
+@safe pure unittest
+{
+    import std.format.spec : singleSpec;
+
+    auto str = "123";
+    auto spec = singleSpec("%s");
+    assert(str.unformatValue!int(spec) == 123);
+
+    str = "ABC";
+    spec = singleSpec("%X");
+    assert(str.unformatValue!int(spec) == 2748);
+
+    str = "11610";
+    spec = singleSpec("%o");
+    assert(str.unformatValue!int(spec) == 5000);
+}
+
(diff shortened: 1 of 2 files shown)


std/format/read.d
```diff
@@ -415,133 +526,40 @@ uint formattedRead(R, Char, S...)(auto ref R r, const(Char)[] fmt, auto ref S ar
...
-    str = "1";
-    spec = singleSpec("%d");
-    assert(unformatValue!bool(str, spec));
-}
-
-/// Null values
-@safe pure unittest
-{
-    import std.format.spec : singleSpec;
+Reads a value from an _input range and converts it according to a
```
RazvanN7, line 529:
```suggestion
Reads a value from the given _input range and converts it according to a
```

std/format/read.d
```diff
@@ -415,133 +526,40 @@ uint formattedRead(R, Char, S...)(auto ref R r, const(Char)[] fmt, auto ref S ar
...
-    auto str = "aaa";
-    auto spec = singleSpec("%s");
-    assert(str.unformatValue!char(spec) == 'a');
-
-    // Using a numerical format spec reads a Unicode value from a string
-    str = "65";
-    spec = singleSpec("%d");
-    assert(str.unformatValue!char(spec) == 'A');
+Params:
+    input = an $(REF_ALTTEXT input range, isInputRange, std, range, primitives),
```
RazvanN7, line 533:
```suggestion
    input = the $(REF_ALTTEXT input range, isInputRange, std, range, primitives),
```

std/format/read.d
```diff
@@ -415,133 +526,40 @@ uint formattedRead(R, Char, S...)(auto ref R r, const(Char)[] fmt, auto ref S ar
...
+            where the formatted input is read from
```
RazvanN7, line 534:
```suggestion
            to read from
```

std/format/read.d
```diff
@@ -415,133 +526,40 @@ uint formattedRead(R, Char, S...)(auto ref R r, const(Char)[] fmt, auto ref S ar
...
+    spec = a $(MREF_ALTTEXT format string, std,format)
+    T = type to return
+    Range = the type of the input range `input`
+    Char = the character type used for `spec`
 
-    str = "41";
-    spec = singleSpec("%x");
-    assert(str.unformatValue!char(spec) == 'A');
+Returns:
+    The value read.
```
RazvanN7, line 541: I think that the old message (A value from `input` of type `T`) is more better and more descriptive

RazvanN7 (changes requested):

std/format/read.d
berni44: Changed that in the overview table in `package.d` too.

RazvanN7 (approved):

PR #7927, opened 2021-03-29
Fix Issue 15227 - std.format undocumented grammar

berni44:
This is a followup on #7919 but can be handled primarily independent. Its main purpose is to replace the description and the grammar of format strings with something, that describes the current implementation.

The order of the `Parameters` is currently not strictly enforced. I'm even unsure if  `Precision` and `Separator` should be reversed - what do you think?

Concerning issue 15227: There are some more missing things in the documentation of `std.format` mentioned. I will take care of them in followup PRs. Anyway I consider this a fix of the issue, because it's main theme is about the grammar and I don't want to fritter about this.

### DIFF
--- a/std/format/package.d
+++ b/std/format/package.d
@@ -60,60 +60,92 @@ Limitation: This package does not support localization, but
     adheres to the rounding mode of the floating point unit, if
     available.
 
-   Format_String:
+$(SECTION3 Format Strings)
 
-   <a name="format-string">$(I Format strings)</a>
-   consist of characters interspersed with $(I format
-   specifications). Characters are simply copied to the output (such
-   as putc) after any necessary conversion to the corresponding UTF-8
-   sequence.
+The functions contained in this package use $(I format strings). A
+format string describes the layout of another string for reading or
+writing purposes. A format string is composed of normal text
+interspersed with $(I format specifiers). A format specifier starts
+with a percentage sign $(B '%'), optionally followed by one or more
+$(I parameters) and ends with a $(I format indicator). A format
+indicator may be a simple $(I format character) or a $(I compound
+indicator).
 
-   The format string has the following grammar:
+$(I Format strings) are composed according to the following grammar:
 
 $(PRE
 $(I FormatString):
-    $(I FormatStringItem)*
+    $(I FormatStringItem) $(I FormatString)
 $(I FormatStringItem):
-    $(B '%%')
-    $(B '%') $(I Position) $(I Flags) $(I Width) $(I Separator) $(I Precision) $(I FormatChar)
-    $(B '%$(LPAREN)') $(I FormatString) $(B '%$(RPAREN)')
-    $(B '%-$(LPAREN)') $(I FormatString) $(B '%$(RPAREN)')
-    $(I OtherCharacterExceptPercent)
+    $(I Character)
+    $(I FormatSpecifier)
+$(I FormatSpecifier):
+    $(B '%') $(I Parameters) $(I FormatIndicator)
+
+$(I FormatIndicator):
+    $(I FormatCharacter)
+    $(I CompoundIndicator)
+$(I FormatCharacter):
+    $(I see remark below)
+$(I CompoundIndicator):
+    $(B '$(LPAREN)') $(I FormatString) $(B '%$(RPAREN)')
+    $(B '$(LPAREN)') $(I FormatString) $(B '%|') $(I Delimiter) $(B '%$(RPAREN)')
+$(I Delimiter)
+    $(I empty)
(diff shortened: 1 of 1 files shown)


std/format/package.d
```diff
@@ -45,61 +45,95 @@ $(TR $(TH Function Name) $(TH Description)
...
-   as putc) after any necessary conversion to the corresponding UTF-8
-   sequence.
+The functions of this package use $(I format strings), which describe
+the layout of an other string $(MDASH) in case of writing of the
+output string and in case of reading of the input string. A format
+string essentially consists of text interspersed with $(I format
+specifiers). A $(I format specifier) starts with a percentage sign,
+followed by some optional $(I parameters) and ends with a $(I format
+indicator) which can be either a simple $(I format character) or a
+$(I compound indicator).
```
RazvanN7, line 58: Suggested rewrite of section:

"The functions contained in this package use $(I format strings). A format string describes the layout of another string for reading or writing purposes. A format string is composed of normal text interspersed with  $(I format specifiers).
A format specifier starts with a percentage sign (`%`), followed by one or more $(I parameters) and ends with a
 $(I format indicator). A format indicator may be a simple $(I format character) or a $(I compound indicator)."

std/format/package.d
```diff
@@ -45,61 +45,95 @@ $(TR $(TH Function Name) $(TH Description)
...
 
-   The format string has the following grammar:
+More formally, $(I format strings) are composed according to the
```
RazvanN7, line 59:
```suggestion
$(I format strings) are composed according to the
```

std/format/package.d
```diff
@@ -45,61 +45,95 @@ $(TR $(TH Function Name) $(TH Description)
...
+    $(I Digit) $(I Digits)
+$(I NonZeroDigit):
+    $(B '1')|$(B '2')|$(B '3')|$(B '4')|$(B '5')|$(B '6')|$(B '7')|$(B '8')|$(B '9')
 $(I Digit):
     $(B '0')|$(B '1')|$(B '2')|$(B '3')|$(B '4')|$(B '5')|$(B '6')|$(B '7')|$(B '8')|$(B '9')
-$(I FormatChar):
-    $(B 's')|$(B 'c')|$(B 'b')|$(B 'd')|$(B 'o')|$(B 'x')|$(B 'X')|$(B 'e')|$(B 'E')|$(B 'f')|$(B 'F')|$(B 'g')|$(B 'G')|$(B 'a')|$(B 'A')|$(B '|')
 )
 
+Note: $(I FormatCharacter) is left unspecified. Theoretically it can
```
RazvanN7, line 126:
```suggestion
Note: $(I FormatCharacter) is unspecified. It can
```

std/format/package.d
```diff
@@ -45,61 +45,95 @@ $(TR $(TH Function Name) $(TH Description)
...
+be any character, that has no other purpose in this grammer, but it
```
RazvanN7, line 127:
```suggestion
be any character that has no other purpose in this grammar, but it
```

std/format/package.d
```diff
@@ -45,61 +45,95 @@ $(TR $(TH Function Name) $(TH Description)
...
+is strongly recommended to stick to (lower- and uppercase) letters
```
RazvanN7, line 128:
```suggestion
is recommended to assign (lower- and uppercase) letters
```

std/format/package.d
```diff
@@ -45,61 +45,95 @@ $(TR $(TH Function Name) $(TH Description)
...
+$(MDASH) this might even be enforced in the future.
```
RazvanN7, line 129: I don't think it's necessary to mention this.

RazvanN7 (changes requested):

std/format/package.d
berni44: "followed by one or more parameters": Can also be zero, like in `%s`. Is "followed by optional parameters" ok?

RazvanN7: I would say: "optionally followed by one or more parameters"

std/format/package.d
RazvanN7: Maybe I wasn't clear, but I was just referring the last line (129) - I don't think it's necessary to mention that "this might even be enforced in the future". I'm fine with the rest of the paragraph.

berni44: Oh, yeah, I got that wrong...

RazvanN7 (approved):
Thanks!

andralex:
Thanks for doing this. FWIW at the time I just mentioned that POSIX positional arguments are supported with a link and left it at that. It's nicer to have the whole thing in the grammar.

PR #7931, opened 2021-04-01
Rework documentation of std.format.read: Module description

berni44:
This one was more work than I expected. I first had to go through the whole source code to understand what really happens, because no useful documentation was available. I found several issues while doing so, but wrote the docs as if these issues were not there, using what I think was intended.

I wonder a little bit, what's the use case of reading `null`...

### DIFF
--- a/std/format/read.d
+++ b/std/format/read.d
@@ -1,17 +1,47 @@
 // Written in the D programming language.
 
 /**
-   This is a submodule of $(MREF std, format).
-   It provides some helpful tools.
+This is a submodule of $(MREF std, format).
 
-   Copyright: Copyright The D Language Foundation 2000-2013.
+It provides two functions for reading formatted input: $(LREF
+unformatValue) and $(LREF formattedRead). The former reads a single
+value. The latter reads several values at once and matches the
+characters found between format specifiers.
 
-   License: $(HTTP boost.org/LICENSE_1_0.txt, Boost License 1.0).
+Parameters are ignored, except for the ones consisting of a single
+$(B '*'). See $(LREF formattedRead) for more information.
 
-   Authors: $(HTTP walterbright.com, Walter Bright), $(HTTP erdani.com,
-   Andrei Alexandrescu), and Kenji Hara
+A space outside of a format specifier has a special meaning: it
+matches any sequence of whitespace characters, not just a single
+space.
 
-   Source: $(PHOBOSSRC std/format/read.d)
+The following combinations of format characters and types are
+available:
+
+$(BOOKTABLE ,
+$(TR $(TH) $(TH s) $(TH c) $(TH d, u, b, o, x, X) $(TH e, E, f, g, G) $(TH r) $(TH compound))
+$(TR $(TD `bool`) $(TD yes) $(TD $(MDASH)) $(TD yes) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)))
+$(TR $(TD `null`) $(TD yes) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)))
+$(TR $(TD $(I integer)) $(TD yes) $(TD $(MDASH)) $(TD yes) $(TD $(MDASH)) $(TD yes) $(TD $(MDASH)))
+$(TR $(TD $(I floating point)) $(TD yes) $(TD $(MDASH)) $(TD $(MDASH)) $(TD yes) $(TD yes) $(TD $(MDASH)))
+$(TR $(TD $(I character)) $(TD yes) $(TD yes) $(TD yes) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)))
+$(TR $(TD $(I string)) $(TD yes) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD yes))
+$(TR $(TD $(I array)) $(TD yes) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD yes))
+$(TR $(TD $(I associative array)) $(TD yes) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD yes))
+)
+
+Below are highlighted examples on how these combinations are used
+with $(LREF unformatValue), however, they apply for $(LREF
+formattedRead) also
+
+Copyright: Copyright The D Language Foundation 2000-2013.
+
+License: $(HTTP boost.org/LICENSE_1_0.txt, Boost License 1.0).
+
+Authors: $(HTTP walterbright.com, Walter Bright), $(HTTP erdani.com,
+Andrei Alexandrescu), and Kenji Hara
+
+Source: $(PHOBOSSRC std/format/read.d)
  */
 module std.format.read;
 
@@ -22,11 +52,11 @@ module std.format.read;
 
     auto str = "false";
     auto spec = singleSpec("%s");
-    assert(unformatValue!bool(str, spec) == false);
+    assert(str.unformatValue!bool(spec) == false);
 
     str = "1";
     spec = singleSpec("%d");
-    assert(unformatValue!bool(str, spec));
+    assert(str.unformatValue!bool(spec) == true);
 }
 
 /// Null values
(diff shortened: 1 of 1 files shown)


std/format/read.d
```diff
@@ -1,35 +1,69 @@
...
-   Andrei Alexandrescu), and Kenji Hara
-
-   Source: $(PHOBOSSRC std/format/read.d)
+This is a submodule of $(MREF std, format).
+
+It provides two functions for reading formatted $(LREF unformatValue)
+reads a single value, while $(LREF formattedRead) reads several
+values at once and matches the characters found between format
+specifiers.
+
```
RazvanN7, line 10: Suggested rewording:

"It provides two functions for reading formatted input: $(LREF unformatValue) and $(LREF formattedRead).
The former reads a single value. The latter reads several values at once and matches the characters found
between format specifiers."

std/format/read.d
```diff
@@ -1,35 +1,69 @@
...
+For reading purpose, only part of the possibilities of $(MREF_ALTTEXT
+format strings, std,format) is used: The parameters are almost
+completely ignored. The only exception is a single $(B '*'): It
+denotes that the read value is not assigned to a variable, but is
+just discarded. A $(B '*') can only be applied to a $(B 'c'),
+$(B 'd') or $(B 'u') format character.
```
RazvanN7, line 16: I don't understand this paragraph.

std/format/read.d
```diff
@@ -1,35 +1,69 @@
...
+
+A space outside of a format specifier has a special meaning: It
```
RazvanN7, line 18:
```suggestion
A space outside of a format specifier has a special meaning: it
```

std/format/read.d
```diff
@@ -1,35 +1,69 @@
...
+$(TR $(TD $(I character)) $(TD yes) $(TD yes) $(TD yes) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)))
+$(TR $(TD $(I string)) $(TD yes) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD yes))
+$(TR $(TD $(I array)) $(TD yes) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD yes))
+$(TR $(TD $(I associative array)) $(TD yes) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD $(MDASH)) $(TD yes))
+)
+
+The examples below show how to use these combinations with $(LREF
+unformatValue). Everything explained there can also be applied to
+$(LREF formattedRead).
+
```
RazvanN7, line 40: Suggested rewording:

"Below are highlighted examples on how these combinations are used with $(LREF
unformatValue), however, they apply for $(LREF formattedRead) also.

std/format/read.d
```diff
@@ -1,35 +1,69 @@
...
+Authors: $(HTTP walterbright.com, Walter Bright), $(HTTP erdani.com,
+Andrei Alexandrescu), and Kenji Hara
+
+Source: $(PHOBOSSRC std/format/read.d)
  */
 module std.format.read;
 
-/// Booleans
+/// Booleans can be read as words `false` and `true` as well as as integers `0` and `1`.
```
RazvanN7, line 52: I think that the unittest is fairly straightforward, no extra explanations are needed. I suggest undoing this change.

std/format/read.d
```diff
@@ -1,35 +1,69 @@
...
 @safe pure unittest
 {
     import std.format.spec : singleSpec;
 
     auto str = "false";
     auto spec = singleSpec("%s");
-    assert(unformatValue!bool(str, spec) == false);
+    assert(str.unformatValue!bool(spec) == false);
```
RazvanN7, line 59: Nice! UFCS for the win!

std/format/read.d
```diff
@@ -39,7 +73,11 @@ module std.format.read;
     assert(str.unformatValue!(typeof(null))(spec) == null);
```
RazvanN7, line 73: This is useful if you are, for example, reading code from a file and want to interpret it.

std/format/read.d
```diff
@@ -1,35 +1,69 @@
...
+    assert(str.unformatValue!bool(spec) == false);
 
     str = "1";
     spec = singleSpec("%d");
-    assert(unformatValue!bool(str, spec));
+    assert(str.unformatValue!bool(spec) == true);
 }
 
-/// Null values
+/// Null values can be read as the word `null`.
```
RazvanN7, line 66: The same goes here, the example is self-explanatory, no need for the extra comment.

std/format/read.d
```diff
@@ -39,7 +73,11 @@ module std.format.read;
     assert(str.unformatValue!(typeof(null))(spec) == null);
 }
 
-/// Integrals
+/**
+Integrals can be read as signed or unsigned decimal value, as dual,
+octal or hexadezimal number. It's also possible to read integers as raw values.
+Reading as raw values depends on the endianess of the processor.
+ */
```
RazvanN7, line 80: Maybe this comment can be broken down into several smaller comments that are put directly above the code paragraph. E.g:

```d
/// Integrals
unitest
{
       // signed decimal values
       auto str = "123";
       auto spec = singleSpec("%s");
       assert(str.unformatValue!int(spec) == 123);
 
       // hexadecimal values
       str = "ABC";
       spec = singleSpec("%X");
       assert(str.unformatValue!int(spec) == 2748);

     .....
}
```

std/format/read.d
```diff
@@ -55,9 +93,17 @@ module std.format.read;
...
+    spec = singleSpec("%r");
+    auto result = str.unformatValue!short(spec);
+    assert(result == 373 /* little endian */ || result == 29953 /* big endian */ );
 }
 
-/// Floating point numbers
+/**
+Floating point numbers can be read in natural or in scientific
+notation. When reading as raw value endianess matters.
+ */
```
RazvanN7, line 106: ditto

std/format/read.d
```diff
@@ -66,18 +112,28 @@ module std.format.read;
...
+    assert(str.unformatValue!double(spec).isClose(1e17));
+
+    str = "\x40\x00\x00\xBF";
+    spec = singleSpec("%r");
+    auto result = str.unformatValue!float(spec);
+    assert(isClose(result, -0.5) /* little endian */ || isClose(result, 2.0) /* big endian */ );
 }
 
-/// Character input ranges
+/// Characters can be read as single characters and as numbers.
```
RazvanN7, line 126: ditto

std/format/read.d
```diff
@@ -91,7 +147,7 @@ module std.format.read;
     assert(str.unformatValue!dchar(spec) == '✓');
 }
 
-/// Arrays and static arrays
+/// Strings and arrays can be read as a single value and as a compound type.
```
RazvanN7, line 150: ditto + Strings are actually arrays behind the scenes so I'm not sure this extra mention is necessary.

std/format/read.d
```diff
@@ -113,18 +169,29 @@ module std.format.read;
...
+    spec = singleSpec("%(%s,%)");
+    assert(str.unformatValue!(int[])(spec) == [1, 2, 3]);
+
+    str = "cool";
+    spec = singleSpec("%(%c%)");
+    assert(str.unformatValue!(char[])(spec) == ['c', 'o', 'o', 'l']);
 }
 
-/// Associative arrays
+/// Associative arrays can be read as single values and as a compound type.
```
RazvanN7, line 182: ditto

RazvanN7 (changes requested):

std/format/read.d
berni44: Maybe I should add an example, probably something like this (taken from the non public unittests):

```
    import std.typecons : Tuple;

    Tuple!(int, float) t;
    char[] line = "1 7643 2.125".dup;
    formattedRead(line, "%s %*u %s", t);
    assert(t[0] == 1 && t[1] == 2.125);
```

RazvanN7: Maybe this shouldn't the emphasized here, but rather explained at unittest site.

std/format/read.d
berni44: All other examples use UFCS, so I changed it here for consistency. ;-)

std/format/read.d
berni44: Instead of `"%s"` the format string could just be `"null"`. Would work equally well, won't it?

std/format/read.d
berni44: They have got different implementations and beginners might not be aware of strings being essentially the same as arrays.

std/format/read.d
RazvanN7: I guess, but then you would have to add `null` as a format specifier which really doesn't worth it.
Anyway, this is just a minor case. I don't have any strong opinions.

std/format/read.d
RazvanN7: OK, still I think it's better if we put the "String" comment right above the example.

std/format/read.d
berni44: As this only applies to `formattedWrite` anyway, I put the example there and added an explaining line in the docs there. I kept part of this paragraph, because I think it should be known, that only part of the grammar applies for reading.

std/format/read.d
```diff
@@ -1,17 +1,49 @@
...
 
-   Copyright: Copyright The D Language Foundation 2000-2013.
+It provides two functions for reading formatted input: $(LREF
+unformatValue) and $(LREF formattedRead). The former reads a single
+value. The latter reads several values at once and matches the
+characters found between format specifiers.
 
-   License: $(HTTP boost.org/LICENSE_1_0.txt, Boost License 1.0).
+For reading purpose, only part of the possibilities of $(MREF_ALTTEXT
+format strings, std,format) is used: The parameters are almost
```
RazvanN7, line 12: What do you mean by "the parameters are completely ignored" ?

berni44: It doesn't matter if you write `%-#030.10,7d` or `%d`. There is no difference. The only exception is `%*d`.

RazvanN7: I see. Maybe this rewording sounds better: "Format specifiers are ignored, except for the ones starting with a single $(B '*'). See $(LREF formattedRead) for more information." ?

berni44: Seems like we've got some babel confusion here - see the discussion in #7935. Maybe we need to rethink the wording of the grammar?

According to the (new) grammar, the "format specifier" includes the letter at the end, which is not ignored, only the parameters of that specifier. Ideally, we should link each of these words directly to the definition in the grammar, but as far as I can see, there is currently no way to do so correctly in ddoc and ddox at the same time; maybe not at all in ddox.

Suggestion: "Parameters are ignored, except for the ones consisting of a single $(B '*'). See $(LREF formattedRead) for more information."

RazvanN7: Yeah, I saw the discussion after I left the previous comment. I agree with the suggestion.

RazvanN7 (approved):

PR #7935, opened 2021-04-04
std.format: Add overview of format indicator

berni44:
Formatting decimals with `%b`, `%o` and `%x` is often referred to as writing this number in binary, octal and hexadecimal. But that is not, what (C-)users expect and what the current implementation produces: It is just an image of the underlying bits. That means `format!"%x"(-10);` results in `fffffff6` instead of `-a`. I tried to follow this with the descriptions.

A similar problem occurs with `%u`, which is often called as formatting as unsigned integer. This was needed with C, because C has no traits and cannot know if the argument was meant to be a signed or an unsigned integer, as far as I know. In D this does not make any sense and in the current implementation `%u` just produces the same results as `%d`. IMHO `%u` could be discarded completely. Yet, people think of it as a feature (e.g. #7913). To make the best out of it, I choose to describe it similar to `%b`, `%o` and `%x` as a (yet buggy implemented) representation of the bit pattern. This seems to me closest to the intention behind `%u`. An alternative would be to kick it into the long grass. Would be fine for me, but might be perceived as a breaking change.

### DIFF
--- a/std/format/package.d
+++ b/std/format/package.d
@@ -144,6 +144,60 @@ recommended to assign (lower- and uppercase) letters.
 Note: The $(I Parameters) of a $(I CompoundIndicator) are currently
 limited to a $(B '-') flag.
 
+$(SECTION4 Format Indicator)
+
+The $(I format indicator) can either be a single character or an
+expression surrounded by $(B %\() and $(B %\)). It specifies the
+basic manner in which a value will be formatted and is the minimum
+requirement to format a value.
+
+The following characters can be used as $(I format characters):
+
+$(BOOKTABLE ,
+   $(TR $(TH FormatCharacter) $(TH Semantics))
+   $(TR $(TD $(B 's'))
+        $(TD To be formatted in a human readable format.
+             Can be used with all types.))
+   $(TR $(TD $(B 'c'))
+        $(TD To be formatted as a character.))
+   $(TR $(TD $(B 'd'))
+        $(TD To be formatted as a signed decimal integer.))
+   $(TR $(TD $(B 'u'))
+        $(TD To be formatted as a decimal image of the underlying bit representation.))
+   $(TR $(TD $(B 'b'))
+        $(TD To be formatted as a binary image of the underlying bit representation.))
+   $(TR $(TD $(B 'o'))
+        $(TD To be formatted as an octal image of the underlying bit representation.))
+   $(TR $(TD $(B 'x') / $(B 'X'))
+        $(TD To be formatted as a hexadecimal image of the underlying bit representation.))
+   $(TR $(TD $(B 'e') / $(B 'E'))
+        $(TD To be formatted as a real number in decimal scientific notation.))
+   $(TR $(TD $(B 'f') / $(B 'F'))
+        $(TD To be formatted as a real number in decimal natural notation.))
+   $(TR $(TD $(B 'g') / $(B 'G'))
+        $(TD To be formatted as a real number in decimal short notation.
+             Depending on the number, a scientific notation or
+             a natural notation is used.))
+   $(TR $(TD $(B 'a') / $(B 'A'))
+        $(TD To be formatted as a real number in hexadezimal scientific notation.))
+   $(TR $(TD $(B 'r'))
+        $(TD To be formatted as raw bytes.
+             The output may not be printable and depends on endianess.))
+)
+
+The $(I compound indicator) can be used to describe compound types
+like arrays or structs in more detail. A compound type is enclosed
+within $(B '%\(') and $(B '%\)'). It either contains a $(I format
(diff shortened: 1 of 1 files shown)


std/format/package.d
```diff
@@ -144,6 +144,59 @@ recommended to assign (lower- and uppercase) letters.
 Note: The $(I Parameters) of a $(I CompoundIndicator) are currently
 limited to a $(B '-') flag.
 
+$(SECTION4 Format Indicator)
+
+The $(I format indicator) can either be a single character or an
+expression surrounded by $(B %\() and $(B %\)). It specifies the
+basic manner in which a value will be formatted.
```
Geod24, line 151:
```suggestion
expression surrounded by $(B %\() and $(B %\)). It is the minimum
requirement to format a value.
```
Perhaps ?

std/format/package.d
```diff
@@ -144,6 +144,59 @@ recommended to assign (lower- and uppercase) letters.
 Note: The $(I Parameters) of a $(I CompoundIndicator) are currently
 limited to a $(B '-') flag.
 
+$(SECTION4 Format Indicator)
```
Geod24, line 147: [The `printf` manual uses the term "Conversion specifier"](https://man7.org/linux/man-pages/man3/printf.3.html):
> Each _conversion specification_ is introduced by the character %, and ends with a _conversion specifier_.

Geod24 (changes requested):

std/format/package.d
berni44: Your sentence is a good idea, I'll add it. But I feel unwell with removing my suggestion - this sentence was meant to define the purpose of the format indicator, which IMHO would be missing if this sentence is removed.

std/format/package.d
berni44: When I started to rewrite the docs here, I encountered different namings for the same thing at various places. For example have a look at [printf in C++](https://www.cplusplus.com/reference/cstdio/printf/). It uses the word "specifier" for the whole thing as well as the letter at the end. In the former docs of `std.format` it wasn't used consistent either.

So I tried to unify the wording throughout the whole module, which resulted in a rewrite of the grammar (see PR #7927). IMHO this wording is consistent in itself and changing a single term would make other changes necessary. For example I'm using "format specifier" for the whole thing. If we use "conversion specifier" here, this needs to be changed too, to avoid confusion.

Another thing, that comes to my mind: Although `printf` was originally the model for this package, `std.format` has already diverged over the time (grouping is done with comma, not with a tick; there is no need to tell about `long` values in D, `%s` was added, and so on) and in my opinion we should not cling too strong to `printf` - it has some drawbacks (e.g. `%p` is unsafe and luckily was never added to Phobos). IMHO D can do better. :-)

RazvanN7 (approved):

RazvanN7:
@Geod24 OK to merge this?

berni44:
@RazvanN7 Why is this not merged now? Can you have a look please?

thewilsonator:
Because the request for changes from @Geod24 is still there

Geod24:
@berni44 : IMO using yet another wording for this is worse than picking one out of two existing wordings.
I would favor C has it is more established and widespread, but I don't have a personal preference.

I however don't think that this concern should block the good work you're doing documenting this module, so I'll dismiss my review and remove `auto-merge` to give you time to read this reply, and decide whether or not you want to change it. If you don't, feel free to re-add `auto-merge` yourself. I've invited your back in the org, please don't delete your account again ;)

Geod24 (approved):

berni44:
> @berni44 : IMO using yet another wording for this is worse than picking one out of two existing wordings.
> I would favor C has it is more established and widespread, but I don't have a personal preference.

I agree, that using what is there, is a good idea in general. However, I think, that the wording of C here is problematic, because some words, like "conversion specification" and "conversion specifier" can be easily confused (it actually took me a while to notice, that these two are two different things). And an other reason is, that I would need to reword all the places, where docs have changed already. I'm a little bit lazy here...

> I however don't think that this concern should block the good work you're doing documenting this module, so I'll dismiss my review and remove `auto-merge` to give you time to read this reply, and decide whether or not you want to change it. If you don't, feel free to re-add `auto-merge` yourself.

Unfortunately I've got no write access, so I can't. (Aside from not being able to add and remove certain buttons, I'm happy with having not write access, because that makes me feel safe that I cannot break anything accidentally.)

Geod24:
> Unfortunately I've got no write access, so I can't. (Aside from not being able to add and remove certain buttons, I'm happy with having not write access, because that makes me feel safe that I cannot break anything accidentally.)

Check your emails. We have protection in place for master and stable so you can't screw those up. But if you really don't like the idea, just reject the invite.

PR #7939, opened 2021-04-06
Add `std.math.integral : ilog2`

berni44:
The idea of `std.math.integral` is to add some functions on integers, that are often useful.

I start with a rather simple one: `ilog2`, which calculates logarithm on base 2 and rounds the result down to the next integer. This is useful for algorithms, that have to do estimates based on the size of a value. E.g. `1 << (ilog2(x)/2)` can be used as a seed for herons's method adapted for integral square root calculation.

The implementation is the fastest I'm aware of. Average computation time for `longs` is on my computer about 0.01ns, using ldc2 with optimization turned on.

### DIFF
--- a/std/math/integral.d
+++ b/std/math/integral.d
@@ -6,9 +6,70 @@ change. Please do not use it yet, but stick to $(MREF std, math).
 
 Copyright: Copyright The D Language Foundation 2021 - .
 License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+Authors:   Bernhard Seckinger
 Source: $(PHOBOSSRC std/math/integral.d)
  */
 
 module std.math.integral;
 
-// Will contain functions like isqrt, ilog2, isPrime and maybe also gcd from std.numeric
+import std.traits : Unqual;
+
+/**
+Calculates the base-2 logarithm of an integral number using binary search.
+
+The result can also be interpreted as the position of the most significant
+1 bit in the binary representation of that number.
+
+The algorithm uses a special processor instruction, if available and
+a very fast software implementation else.
+
+Params:
+    value = an integral number
+    Int = the type of the integral number
+
+Returns:
+    For positive values the base-2 logarithm rounded down to the next integer
+    value and 0 for zero and negative values.
+ */
+Unqual!Int ilog2(Int)(Int value)
+{
+    import core.bitop : bsr;
+
+    if (value <= 0) return 0;
+    return cast(typeof(return)) bsr(value);
+}
+
+///
+@safe pure unittest
+{
+    assert(ilog2(8) == 3);
+
+    // the result is rounded down
+    assert(ilog2(10) == 3);
+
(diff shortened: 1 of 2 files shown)


RazvanN7:
cc @atilaneves

RazvanN7 (commented):
Looks good to me.

dkorpel (changes requested):
The implementation can be simplified to:
```D
Unqual!Int ilog2(Int)(Int value)
{
    import core.bitop: bsr;
    if (value <= 0) return 0;
    return cast(typeof(return)) bsr(value);
}
```
Scanning for the most significant bit has a special instruction on most processors, and druntime has a function exposing it (with a fast software fall-back).

ibuclaw (changes requested):
Use `bsr` - though I wonder if we really need a new function for something that is just a trivial shell around another.

berni44:
> Scanning for the most significant bit has a special instruction on most processors, and druntime has a function exposing it (with a fast software fall-back).

Oh, thanks - I didn't know that.

berni44:
> Use `bsr` - though I wonder if we really need a new function for something that is just a trivial shell around another.

Well, just the fact, that I did not know about `bsr` reveals already, that there is some need. OK, I didn't look into `druntime`, admitted. But anyway, I would have overlooked `bsr`, because it's name doesn't tell one anything, when not being familiar with processor instructions. And well, in some sense it's a coincidence, that the position of the leftmost bit and `log2` is the same for integers. With `ilog10` this doesn't work anymore and people may wonder why `ilog2` was not implemented. For example look at [issue 9300](https://issues.dlang.org/show_bug.cgi?id=9300). If the author of that issue would have been aware of `bsr` he would probably have chosen a different example...

atilaneves:
> > Use `bsr` - though I wonder if we really need a new function for something that is just a trivial shell around another.
> 
> Well, just the fact, that I did not know about `bsr` reveals already, that there is some need. OK, I didn't look into `druntime`, admitted. But anyway, I would have overlooked `bsr`, because it's name doesn't tell one anything, when not being familiar with processor instructions. And well, in some sense it's a coincidence, that the position of the leftmost bit and `log2` is the same for integers. With `ilog10` this doesn't work anymore and people may wonder why `ilog2` was not implemented. For example look at [issue 9300](https://issues.dlang.org/show_bug.cgi?id=9300). If the author of that issue would have been aware of `bsr` he would probably have chosen a different example...

Is bsr literally the same thing? If so we definitely don't need two functions.

berni44:
> Is bsr literally the same thing? If so we definitely don't need two functions.

`std.math` contains several functions that are merely wrappers around a function of `druntime`, e.g. several of the trigonometric functions, or `rint`, `fabs` or `sqrt`. And funnily (I didn't know that before I just yet scanned through the package): An overload of [`ilogb`](https://dlang.org/phobos/std_math.html#.ilogb) for integers which is essentially, what I suggest here... *g*

PR #7941, opened 2021-04-06
std.math: Move classic algebraic functions into submodule.

berni44:


thewilsonator (approved):

berni44:
Added forgotten macros.

PR #7942, opened 2021-04-07
std.math: Move pow function into submodule.

berni44:
Moving pow to submodule seems to be a little bit more problematic, so I'm doing this in a single step. Let's see if it works...


berni44:
Some windows computers fail with a unittest in `std.internal.math.gammafunction.d`. Here the log of the x64 version, which is more verbose:

```
core.exception.AssertError@std\internal\math\gammafunction.d(1053): unittest failure
----------------
0x00007FF71340BB12 in d_unittestp
0x00007FF7131D508B in std.internal.math.gammafunction.__unittest_L1024_C7 at D:\a\1\phobos\std\internal\math\gammafunction.d(1054)
0x00007FF7131D8F27 in void std.internal.math.gammafunction.__modtest()
0x00007FF71342F139 in int core.runtime.runModuleUnitTests().__foreachbody2(object.ModuleInfo*)
0x00007FF71341B9A7 in int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)).__lambda2(immutable(object.ModuleInfo*))
0x00007FF713432794 in int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*))).__foreachbody2(ref rt.sections_win64.SectionGroup)
0x00007FF7134326FF in int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*)))
0x00007FF71341B967 in int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*))
0x00007FF71342EF87 in runModuleUnitTests
0x00007FF713418D25 in void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).runAll()
0x00007FF713418C7F in void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).tryExec(scope void delegate())
0x00007FF713418A86 in d_run_main2
0x00007FF713404FB3 in d_run_main
0x00007FF7110F9122 in unittest._d_cmain!().main at D:\a\1\phobos\../druntime\import\core\internal\entrypoint.d(30)
0x00007FF7134F1378 in __scrt_common_main_seh at d:\agent\_work\2\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl(288)
0x00007FFAD01E84D4 in BaseThreadInitThunk
0x00007FFAD2221821 in RtlUserThreadStart
1/130 modules FAILED unittests
std\internal\math\gammafunction.d(1053): [unittest] unittest failure
```

Anyone an idea, what's the reason here?

thewilsonator:
MS win has `double == real` and the gamma function tests in particular test more precision than they really should. 
I take it this PR is a cut and paste job for the most part?

evaluate the failing assertion
`feqrel(betaIncompleteInv(5, 10, 0.2000002972865658842L), 0.229121208190918L) >= real.mant_dig - 3`  and see how far off it is.

berni44:
> I take it this PR is a cut and paste job for the most part?

Aside from a few needed imports it's cut and paste...

berni44:
It's only one digit: .0...799 instead of .0...800.

Anyway I wonder, why the calculation changed. It shouldn't.

berni44:
`eBay/tsv-utils` fails too, due to the same problems as encountered with the gamma function.

IMHO, we should find out, why dmd produces different code here and fix that...

RazvanN7:
I wonder if might be possible that the compiler searches for pow in both druntime and std.math. Maybe these have different implementations and now since pow is moved from std.math, the compiler picks the one in druntime?

Also, there's the style checker:

```
std/internal/math/gammafunction.d(26:1)[warn]: A unittest should be annotated with at least @safe or @system
`

**Comment by berni44** [comment]
> Also, there's the style checker:

I'm aware of that, but I see no merit in starting the testers again, when the main issue is not fixed... ;-) (Ideally, the gammafunction should not be touched at all.)

**Comment by RazvanN7** [comment]
Ok, I was just pointing that out :-"

**Comment by ibuclaw** [comment]
The compiler lowers `^^` operator to
```
static import std.math;
std.math.pow(x, y);
```

Hazarding a guess, but you might need to force that the std version takes precedent with:

```
alias pow = std.math.exponential.pow;
```
In the package module.


Ideally pow should really be in druntime because it is a compiler interface, however the non-trivial template implementation makes it awkward to move without also privately copying a good chunk of meta and traits along with it.

**Comment by berni44** [comment]
> Hazarding a guess, but you might need to force that the std version takes precedent with:
> 
> ```
> alias pow = std.math.exponential.pow;
> ```
> 
> In the package module.

For `round` and `trunc` this didn't work. But maybe here it does. Let's try...

**Comment by berni44** [comment]
> For `round` and `trunc` this didn't work. But maybe here it does. Let's try...

No, unfortunately not...

**Comment by ibuclaw** [comment]
> > For `round` and `trunc` this didn't work. But maybe here it does. Let's try...
> 
> No, unfortunately not...

Another shot in the dark. Remove the public import of core.math from the std.math package.

In its place, use public selective imports in the appropriate submodules.

```
// e.g in std.math.exponential
public import core.math : pow;
```
An intentional side effect being new core.math built-ins aren't implicitly pulled into std.math.

**Comment by berni44** [comment]
> Another shot in the dark. Remove the public import of core.math from the std.math package.

`core.math` isn't publically imported, only static. I moved it into the functions anyway to see, if this has any effect...

**Comment by ibuclaw** [comment]
> > Another shot in the dark. Remove the public import of core.math from the std.math package.
> 
> `core.math` isn't publically imported, only static. I moved it into the functions anyway to see, if this has any effect...
> 

Oh, then I suppose it'll likely have no effect at all them, except if there're bugs in the import visibility system.

**Comment by ibuclaw** [comment]
> > `core.math` isn't publically imported, only static. I moved it into the functions anyway to see, if this has any effect...
> > 
> 
> Oh, then I suppose it'll likely have no effect at all them, except if there're bugs in the import visibility system.

Saying that though, it all seems to be green?

**Comment by berni44** [comment]
> Saying that though, it all seems to be green?

Unfortunately not. Still the same problem.

**Comment by ibuclaw** [comment]
@berni44 - I've managed to find some time to check this proper.

Compiling gammafunction.d on x86_64-linux using `-vcg-ast` and diff'd the generated dump before and after this patch.  This difference in codegen has no relation to a change in this PR.

```
@@ -5473,7 +5474,8 @@ pow!(real, real)
 				}
 				x = -x;
 			}
-			return sign * exp2(yl2x(x, y));
+			real w = exp2(y * log2(x));
+			return sign * w;
 		}
 		return impl(x, y);
 	}
```

Looks like the culprit is that you forgot to define `version = INLINE_YL2X;` for DigitalMars platforms.

File: std/math/exponential.d [hunk 1]
```diff
@@ -16,8 +16,827 @@ License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
...
+Unqual!F pow(F, G)(F x, G n) @nogc @trusted pure nothrow
+if (isFloatingPoint!(F) && isIntegral!(G))
+{
+    import std.math : abs, floor, isNaN, log2;
+    import std.traits : Unsigned;
+
+    // NaN ^^ 0 is an exception defined by IEEE (yields 1 instead of NaN)
+    if (isNaN(x)) return n ? x : 1.0;
+
+    import std.traits : Unsigned;
```
ibuclaw, line 48: You're importing this twice here.
```suggestion
    import std.traits : Unsigned;

    // NaN ^^ 0 is an exception defined by IEEE (yields 1 instead of NaN)
    if (isNaN(x)) return n ? x : 1.0;

```

std/math/exponential.d
```diff
@@ -16,8 +16,827 @@ License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
...
+    INFIN = &infin;
+    PLUSMNINF = &plusmn;&infin;
+    LT = &lt;
+    GT = &gt;
  */
 
 module std.math.exponential;
 
-// Will contain functions like pow, exp, log2
+import std.traits :  isFloatingPoint, isIntegral, Largest, Unqual;
```
ibuclaw, line 34:
```suggestion
import std.traits :  isFloatingPoint, isIntegral, Largest, Unqual;

static import core.math;

version (DigitalMars)
{
    version = INLINE_YL2X;        // x87 has opcodes for these
}
```

std/math/package.d
```diff
@@ -155,7 +156,6 @@ public import std.math.rounding;
 public import std.math.traits;
 public import std.math.trigonometry;
 
-static import core.math;
 static import core.stdc.math;
```
ibuclaw, line 159:
```suggestion
static import core.math;
static import core.stdc.math;
```

std/math/package.d
```diff
@@ -167,6 +167,8 @@ deprecated("std.meta.AliasSeq was unintentionally available from std.math "
            ~ "and will be removed after 2.102. Please import std.meta instead")
 public import std.meta : AliasSeq;
 
+alias pow = std.math.exponential.pow;
+
```
ibuclaw, line 171:
```suggestion
```

std/math/package.d
```diff
@@ -229,7 +231,7 @@ version (D_HardFloat)
     version (IeeeFlagsSupport) version = FloatingPointControlSupport;
 }
 
-version (StdUnittest) private
+package(std.math)
```
ibuclaw, line 234:
```suggestion
version (StdUnittest) package(std.math)
```

std/math/package.d
```diff
@@ -295,13 +297,25 @@ version (StdUnittest) private
  *      Results are undefined if |x| >= $(POWER 2,64).
  */
 pragma(inline, true)
-real cos(real x) @safe pure nothrow @nogc { return core.math.cos(x); }
+real cos(real x) @safe pure nothrow @nogc
+{
+    static import core.math;
+    return core.math.cos(x);
+}
```
ibuclaw, line 304:
```suggestion
real cos(real x) @safe pure nothrow @nogc { return core.math.cos(x); }
```

std/math/package.d
```diff
@@ -295,13 +297,25 @@ version (StdUnittest) private
...
 ///ditto
 pragma(inline, true)
-double cos(double x) @safe pure nothrow @nogc { return core.math.cos(x); }
+double cos(double x) @safe pure nothrow @nogc
+{
+    static import core.math;
+    return core.math.cos(x);
+}
```
ibuclaw, line 311:
```suggestion
double cos(double x) @safe pure nothrow @nogc { return core.math.cos(x); }
```

std/math/package.d
```diff
@@ -295,13 +297,25 @@ version (StdUnittest) private
...
 ///ditto
 pragma(inline, true)
-float cos(float x) @safe pure nothrow @nogc { return core.math.cos(x); }
+float cos(float x) @safe pure nothrow @nogc
+{
+    static import core.math;
+    return core.math.cos(x);
+}
```
ibuclaw, line 318:
```suggestion
float cos(float x) @safe pure nothrow @nogc { return core.math.cos(x); }
```

ibuclaw:
Just realised I can commit the suggestions myself.

ibuclaw:
New error:
```
D:\a\1\dmd\test\..\..\phobos\std\math\exponential.d(667): Error: `yl2x` cannot be interpreted at compile time, because it has no available source code
compilable\test5227.d(120): Error: cannot interpret `<error>` at compile time
compilable\test5227.d(120):        while evaluating `pragma(msg, powf)`
```
Which is correct, when LDC/GDC is building DMD, the resultant compiler can not interpret yl2x at compile-time.

ibuclaw:
So this is blocked by the compiler now.
https://github.com/dlang/dmd/blob/e658df1cc3a1622dea8e3614a5c3d94742ebbf52/src/dmd/builtin.d#L89-L93

It looks only for module names ending with `math` (i.e: core.math, std.math), so it returns `BUILTIN.unimpl` so tries to interpret the body.

berni44:
> So this is blocked by the compiler now.
> https://github.com/dlang/dmd/blob/e658df1cc3a1622dea8e3614a5c3d94742ebbf52/src/dmd/builtin.d#L89-L93
> 
> It looks only for module names ending with `math` (i.e: core.math, std.math), so it returns `BUILTIN.unimpl` so tries to interpret the body.

Wilsonator pointed me to the same line in PR #7943.

ibuclaw:
> > So this is blocked by the compiler now.
> > https://github.com/dlang/dmd/blob/e658df1cc3a1622dea8e3614a5c3d94742ebbf52/src/dmd/builtin.d#L89-L93
> > It looks only for module names ending with `math` (i.e: core.math, std.math), so it returns `BUILTIN.unimpl` so tries to interpret the body.
> 
> Wilsonator pointed me to the same line in PR #7943.

Just validating a fix just now with `compilable/test5227.d`, it looks like `fabs` stopped being matched in #7941, but this wasn't noticed because the body just forwards to `core.math`.

ibuclaw:
No longer blocked by compiler built-ins support. Looks like one remaining failure is due to missing symbols.

std/math/package.d
```diff
@@ -229,7 +230,7 @@ version (D_HardFloat)
     version (IeeeFlagsSupport) version = FloatingPointControlSupport;
 }
 
-version (StdUnittest) private
+version (StdUnittest) package(std.math)
```
berni44, line 233: The `version (StdUnittest)` here causes the failure of circleci`. This was the reason, why I removed it. I don't know now, if I should change anything or not...

ibuclaw: I haven't checked, but it sounds like the test is compiled with `-version=StdUnittest`, but linked to the release library?

The `-allinst` flag should force the symbol to be emitted, though perhaps the better thing to do would be to replace `equalsDigit` with `isClose` in a prequel PR.

berni44: This applies also to several other functions too, e.g. the trigonometric ones. I'd suggest to make this temporary `package(std.math)` (without version unittest) and when the splitting is done, go for `isClose`? (What troubles me, is the thought of having a large copy & paste here, which needs to be fixed after such a prequel PR. It's easy to do mistakes (like dropping stuff) which neither git notices nor the test suite.)

berni44: (And yes, maybe it would be better, if circleci would be fixed.)

ibuclaw: Ok

ibuclaw: I mean, I'm not particularly thrilled to introduce a new symbol into the library that is only used by the testsuite, but if we are quick to remove `equalsDigit`, then fine.

berni44: Well, it's still not public, only package. Shouldn't hurt too much.

std/math/package.d
```diff
@@ -1485,6 +1486,7 @@ float atanh(float x) @safe pure nothrow @nogc { return atanh(cast(real) x); }
  */
 pragma(inline, true)
 long rndtol(real x) @nogc @safe pure nothrow { return core.math.rndtol(x); }
+
```
ibuclaw, line 1489: Random space inserted here, probably from the many trials and rebasings.

ibuclaw (changes requested):
One last nit, and this is good to go.

berni44:
Many thanks for your help!

ibuclaw (approved):

PR #7943, opened 2021-04-07
std.math: Move most of the rounding functions into submodule.

berni44:
I left out `floor` and `ceil` because moving `floorImpl` to the submodule fails with a strange error message. I guess, there is some compiler magic involved here, making `union` work in CTFE.


berni44:
Here `compilable/test5227.d` fails. It's two function that cannot be evaluated at runtime: `round` and `trunc`. Looks, like the calls to `core.stdc.math.roundl` and `core.stdc.math.truncl` in CTFE are restricted to `std.math`, but cannot be called from `std.math.rounding`.

berni44:
It's not the calls I mentioned above. So I suspect dmd searching `round` and `trunc` in `std.math` and not in `std.math.rounding`; that is, the public imports are ignored... I need help here from someone who knows the inner workings of dmd.

thewilsonator:
https://github.com/dlang/dmd/blob/master/src/dmd/builtin.d#L91 is the function in question.

RazvanN7:
Have you tried making a public alias to the symbols that were moved to the submodules? That should work.

Otherwise, we can make the compiler acknowledge the new submodules, although I would rather not go down that route.

berni44:
> Have you tried making a public alias to the symbols that were moved to the submodules? That should work.

Wouldn't that clash with the public imports? I'll check it out.

berni44:
Didn't clash, so I probably do not understand something with visibility and scope... But that's my fault. I'll try to pick that up.

Anyway, the error has not vanished, but this time brought up a new detail:

```
D:\a\1\dmd\test\..\..\phobos\std\math\package.d(4851): Error: `asm` statements cannot be interpreted at compile time
D:\a\1\dmd\test\..\..\phobos\std\math\rounding.d(437):        called from here: `getControlState()`
compilable\test5227.d(70):        called from here: `round(5.5L)`
compilable\test5227.d(70):        while evaluating `pragma(msg, roundr)`
```

Line 4851 of `package.d` is inside `getControlState()` of `FloatingPointControl`.

std/math/rounding.d
```diff
@@ -19,4 +19,573 @@ Source: $(PHOBOSSRC std/math/rounding.d)
...
+ *
+ * This is generally the fastest method to convert a floating-point number
+ * to an integer. Note that the results from this function
+ * depend on the rounding mode, if the fractional part of x is exactly 0.5.
+ * If using the default rounding mode (ties round to even integers)
+ * lrint(4.5) == 4, lrint(5.5)==6.
+ */
+long lrint(real x) @trusted pure nothrow @nogc
+{
+    version (InlineAsm_X87)
```
ibuclaw, line 223: Missing version definition at top of file.

std/math/rounding.d
```diff
@@ -19,4 +19,573 @@ Source: $(PHOBOSSRC std/math/rounding.d)
...
+}
+
+/**
+ Returns the integer portion of x, dropping the fractional portion.
+ This is also known as "chop" rounding.
+ `pure` on all platforms.
+ */
+real trunc(real x) @trusted nothrow @nogc pure
+{
+    version (InlineAsm_X87_MSVC)
```
ibuclaw, line 504: Missing version definition

std/math/package.d
```diff
@@ -167,6 +174,9 @@ deprecated("std.meta.AliasSeq was unintentionally available from std.math "
            ~ "and will be removed after 2.102. Please import std.meta instead")
 public import std.meta : AliasSeq;
 
+public alias round = std.math.rounding.round;
+public alias trunc = std.math.rounding.trunc;
```
ibuclaw, line 178: Remove this as is unnecessary

std/math/rounding.d
```diff
@@ -19,4 +19,573 @@ Source: $(PHOBOSSRC std/math/rounding.d)
 
 module std.math.rounding;
 
-// Will contain functions like ceil, floor, round, quantize
+static import core.math;
+static import core.stdc.math;
+
+import std.traits : isFloatingPoint, isIntegral, Unqual;
+
```
ibuclaw, line 26:
```suggestion

version (D_InlineAsm_X86)    version = InlineAsm_X86_Any;
version (D_InlineAsm_X86_64) version = InlineAsm_X86_Any;

version (InlineAsm_X86_Any) version = InlineAsm_X87;
version (InlineAsm_X87)
{
    static assert(real.mant_dig == 64);
    version (CRuntime_Microsoft) version = InlineAsm_X87_MSVC;
}

```

ibuclaw (changes requested):
Squash and add missing versions.

ibuclaw (approved):

PR #7944, opened 2021-04-07
Make string/character auto-escaping more prominent

MetaLang:
This functionality is unexpected and buried in between examples down the page. This adds a note just below where compound specifiers are first introduced to make it more prominent for anyone reading the docs.

### DIFF
--- a/std/format/package.d
+++ b/std/format/package.d
@@ -343,7 +343,8 @@ limited to a $(B '-') flag.
     string will be included following the last array element.
 
     NOTE: Inside a grouping format specifier, strings and characters are
-    escaped automatically. To avoid this behavior, add '-' flag to "%(".
+    escaped automatically. To avoid this behavior, add the $(B '-') flag
+    to `"%$(LPAREN)"`.
 
     Example using array and nested array formatting:
     -------------------------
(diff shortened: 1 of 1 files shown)


pbackus:
Only change I'd make is to replace "add the `-` flag to `%(`" with something like "use `%-(` instead of `%(`"—both because it's more direct, and because the `-` in `%-(` is not actually a flag, grammatically.

MetaLang:
> Only change I'd make is to replace "add the `-` flag to `%(`" with something like "use `%-(` instead of `%(`"—both because it's more direct, and because the `-` in `%-(` is not actually a flag, grammatically.

I'm pretty much just copying what was already there and moving it up a bit. However, changing it to make it more clear is a good idea so I'll do that while I'm at it.

MetaLang:
For both the dlang.phobos failures, it shows the following error message:
```
@azure-pipelinesazure-pipelines
/ dlang.phobos (Windows_VisualD x86-mscoff)
Build log #L36
Git fetch failed with exit code: 128
```

The issue seems unrelated to this PR, which only changes the docs.

pbackus:
I restarted the tests and they're green now.

pbackus (approved):

RazvanN7 (approved):

MetaLang:
CircleCI seems to be stuck "waiting for status to be reported". Does anyone know how to give it a kick to get things moving again?

thewilsonator:
Force push.

PR #7945, opened 2021-04-08
Rework examples of std.format.write

berni44:
I intentionally skipped the `struct` examples. I want to cope with these halve a dozen `toString` variations in a separate PR.

I removed the `delegates`-example. Formatting delegates is creepy and buggy currently, and the example does not even use `formatValue` but uses `to` from `std.conv`. IMHO formatting delegates (and functions?) should be completely revised before giving an (almost useless) example.

### DIFF
--- a/std/format/write.d
+++ b/std/format/write.d
@@ -51,22 +51,28 @@ Source: $(PHOBOSSRC std/format/write.d)
 module std.format.write;
 
 /**
- * `bool`s are formatted as `"true"` or `"false"` with `%s` and as `1` or
- * `0` with integral-specific format specs.
+`bool`s are formatted as `"true"` or `"false"` with `%s` and like the
+`byte`s 1 and 0 with all other format characters.
  */
 @safe pure unittest
 {
     import std.array : appender;
     import std.format.spec : singleSpec;
 
-    auto w = appender!string();
-    auto spec = singleSpec("%s");
-    formatValue(w, true, spec);
+    auto w1 = appender!string();
+    auto spec1 = singleSpec("%s");
+    formatValue(w1, true, spec1);
 
-    assert(w.data == "true");
+    assert(w1.data == "true");
+
+    auto w2 = appender!string();
+    auto spec2 = singleSpec("%#x");
+    formatValue(w2, true, spec2);
+
+    assert(w2.data == "0x1");
 }
 
-/// `null` literal is formatted as `"null"`.
+/// The `null` literal is formatted as `"null"`.
 @safe pure unittest
 {
     import std.array : appender;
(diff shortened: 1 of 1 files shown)


berni44:
I wonder why the bot didn't catch this one...

std/format/write.d
```diff
@@ -79,129 +85,219 @@ module std.format.write;
...
+
+    assert(w2.data == "0xec");
 }
 
-/// Strings are formatted like $(REF printf, core, stdc, stdio)
+/**
+Strings are formatted as a sequence of characters with `%s`.
+Non-printable characters are not escaped. With a compound specifier
+the string is treated like a range of characters. With $(MREF_ALTTEXT
+compound specifiers, std,format) strings are treated special.
```
RazvanN7, line 176:
```suggestion
compound specifiers, std,format) strings are treated differently.
```

RazvanN7: Also, why is this last sentence necessary?

std/format/write.d
```diff
@@ -79,129 +85,219 @@ module std.format.write;
...
 }
 
 /**
- * Associative arrays are formatted by using `':'` and `", "` as
- * separators, and enclosed by `'['` and `']'`.
+Associative arrays are formatted by using `':'` and `", "` as
+separators, enclosed by `'['` and `']'` when used with `%s`. It's
+also possible to use a compound specifier for better control.
+
+Please note, that the order of the elements is not defined and hence
```
RazvanN7, line 246:
```suggestion
Please note, that the order of the elements is not defined, therefore
```

std/format/write.d
```diff
@@ -51,22 +51,28 @@ Source: $(PHOBOSSRC std/format/write.d)
 module std.format.write;
 
 /**
- * `bool`s are formatted as `"true"` or `"false"` with `%s` and as `1` or
- * `0` with integral-specific format specs.
+`bool`s are formatted as `"true"` or `"false"` with `%s` and like the
+`byte`s 1 and 0 with all other format characters.
```
RazvanN7, line 55: I think it sounds better if we drop the `byte` mentione. E.g: "[...] and as 1 or 0 with all other format characters."

std/format/write.d
```diff
@@ -79,129 +85,219 @@ module std.format.write;
...
+    auto spec2 = singleSpec("%x");
+    formatValue(w2, -1337, spec2);
+
+    assert(w2.data == "fffffac7");
 }
 
-/// Floating-point values are formatted like $(REF printf, core, stdc, stdio)
+/**
+Floating-point values are formatted in natural notation with `%f`, in
+scientific notation with `%e`, in short notation with `%g`, and in
```
RazvanN7, line 113:
```suggestion
scientific notation with `%e`, in short notation with `%g` and in
```

std/format/write.d
```diff
@@ -79,129 +85,219 @@ module std.format.write;
...
+hexadecimal scientific notation with `%a`. They are rounded according
+to the current rounding mode of the floating point unit, if available.
```
RazvanN7, line 115:
```suggestion
to the current rounding mode of the floating point unit.
```

RazvanN7: If you want to keep the `if available` part you also need to specify what happens in the contrary case. If you go down that road, I suggest rewording so that the sentence starts with that: "If a rounding mode is available/provided [...], otherwise [...]."

std/format/write.d
berni44: That's this annoying "feature", that strings and characters are treated like literals when they appear inside `%(...%)`.

std/format/write.d
berni44: I added that, because of `%r`. That's the only difference (see #7946).

std/format/write.d
berni44: In a recent PR I was asked to add such a "Harvard comma", because the docs are US English? I don't know much about it, just want to make sure...

RazvanN7: Ok, that's fine. Although, personally, I dislike the serial comma.

berni44: Me too. Look weird.

RazvanN7 (approved):

PR #7947, opened 2021-04-08
Deprecate std.typetuple.

berni44:
I suggest to deprecate this. It was removed in 2015 and is now been long around...


berni44:
Seems to be too early to do so. Still too many software, that relies on it...

PR #7951, opened 2021-04-11
std.format: Replace snprintf for x87-reals

berni44:
Finally going life with formatting x87-reals.

I consider this a fix for several issues, although the issues might still remain for some exotic reals: There is no official list, which reals D does support and also no way for me to test if the issues still exist. That means, the issues can never be closed if we do not decide at some time, that it's enough and anyway any fix of these issues will high likely be to support printing floats for that exotic reals, so there is no benefit in leaving the issues open.

### DIFF
--- a/changelog/formatting_floats_in_CTFE.dd
+++ b/changelog/formatting_floats_in_CTFE.dd
@@ -1,4 +1,4 @@
-`float` and `double` values can be formatted at compile time
+Floating point numbers can be formatted at compile time
 
 Example:
 ------
@@ -12,4 +12,5 @@ enum golden_ratio = format!"|%+-20.10E|"((1 + sqrt(5.0)) / 2);
 static assert(golden_ratio == "|+1.6180339887E+00   |");
 ------
 
-Note: compile time formatting `real`s remains unsupported.
+Note: compile time formatting `real`s is only supported for 64-bit
+reals and 80-bit reals.
--- a/std/format/internal/write.d
+++ b/std/format/internal/write.d
@@ -1041,6 +1041,15 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     static assert(format("%s", 0.6L) == "0.6");
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=9297
+@safe pure unittest
+{
+    static if (real.mant_dig == 64) // 80 bit reals
+    {
+        assert(format("%.25f", 1.6180339887_4989484820_4586834365L) == "1.6180339887498948482072100");
+    }
+}
+
 /*
     Formatting a `creal` is deprecated but still kept around for a while.
  */
--- a/std/format/internal/write.d
+++ b/std/format/internal/write.d
@@ -893,7 +893,7 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     real a = 0.16;
     real b = 0.016;
     assert(format("%.1f", a) == "0.2");
-//    assert(format("%.2f", b) == "0.02"); // Windows still fails here...
+    assert(format("%.2f", b) == "0.02");
 
     double a1 = 0.16;
     double b1 = 0.016;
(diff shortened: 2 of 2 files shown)


thewilsonator:
> There is no official list, which reals D does support and also no way for me to test if the issues still exist. 

From memory this is:
* IEEE 32 bit (float),
* IEEE 64 bit (double), 
* intel 80-bit x87 ,
* IBM double-double, 
* IEEE 128-bit

Take a look at `RealFormat` in phobos.

thewilsonator:
> so there is no benefit in leaving the issues open.

Sounds good to me.

thewilsonator (approved):

berni44:
> > There is no official list, which reals D does support and also no way for me to test if the issues still exist.
> 
> From memory this is:
> 
> * IEEE 32 bit (float),
> * IEEE 64 bit (double),
> * intel 80-bit x87 ,
> * IBM double-double,
> * IEEE 128-bit
> 
> Take a look at `RealFormat` in phobos.

`RealFormat` (which is not public) defines:

```
    ieeeHalf,
    ieeeSingle,
    ieeeDouble,
    ieeeExtended,   // x87 80-bit real
    ieeeExtended53, // x87 real rounded to precision of double.
    ibmExtended,    // IBM 128-bit extended
    ieeeQuadruple,
```

Docs of `std.math` state:

- 64 bit Big-endian  'double' (eg PowerPC))
- 128 bit Big-endian 'quadruple' (eg SPARC))
- 64 bit Little-endian 'double' (eg x86-SSE2))
- 80 bit Little-endian, with implied bit 'real80' (eg x87, Itanium))
- 128 bit Little-endian 'quadruple' (not implemented on any known processor!))
- Non-IEEE 128 bit Big-endian 'doubledouble' (eg PowerPC) has partial support)

Language reference doesn't state anything, as far as I know.

changelog/formatting_floats_in_CTFE.dd
```diff
@@ -1,4 +1,4 @@
-`float` and `double` values can be formatted at compile time
+`float`, `double` and certain `real` values can be formatted at compile time
```
PetarKirov, line 1: 64-bit and 80-bit `real`s accounts for 100% of the platforms supported by dmd, so I think the note below is sufficient.

```suggestion
`float`, `double` and `real` values can be formatted at compile time
```

std/format/internal/write.d
```diff
@@ -1031,6 +1033,30 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     assert(format("%#.6g",a) == "-1.00000e+06");
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=8424
+@safe pure unittest
+{
+    static assert(format("%s",0.6f) == "0.6");
+    static assert(format("%s",0.6) == "0.6");
+    static assert(format("%s",0.6L) == "0.6");
```
PetarKirov, line 1041:
```suggestion
    static assert(format("%s", 0.6f) == "0.6");
    static assert(format("%s", 0.6) == "0.6");
    static assert(format("%s", 0.6L) == "0.6");
```

std/format/internal/write.d
```diff
@@ -648,7 +649,7 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
             return;
         }
 
-        enforceFmt(!__ctfe, "Cannot format reals at compile-time.");
+        enforceFmt(!__ctfe, "The reals on this computer are not supported at compile-time.");
```
PetarKirov, line 652:
```suggestion
        enforceFmt(!__ctfe, mixin(`"Unsupported `real` type: real.sizeof = `, real.sizeof, ` | real.mant_dig = `, real.mant_dig, `"`));
```

PetarKirov (approved):

RazvanN7 (approved):
Awesome! Thanks!

changelog/formatting_floats_in_CTFE.dd
berni44: I suggest "Floating point numbers can be ..."? No need to list all of them...

PR #7957, opened 2021-04-12
std.format: Add description of flags, width, precision, separator and positional arguments.

berni44:
I kept the description of the flags rather brief. They will be given in more detail in the followup PR, which will feature a large table of types and format chars, which will give special meanings of the flags in more detail.

### DIFF
--- a/std/format/package.d
+++ b/std/format/package.d
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
 the elements of the compound type will be formatted. This $(I format
 specifier) can be a $(I compound indicator) itself.
 
+$(SECTION4 Flags)
+
+There are several flags that affect the outcome of the formatting.
+
+$(BOOKTABLE ,
+   $(TR $(TH Flag) $(TH Semantics))
+   $(TR $(TD $(B '-'))
+        $(TD When the formatted result is shorter then the value
+             given by the width parameter, the output is right
+             justified. With the $(B '-') flag this is changed
+             to left justification.
+
+             There are two exceptions where the $(B '-') flag has a
+             different meaning: (1) with $(B 'r') it denotes to use little
+             endian and (2) in case of a compound indicator it means that
+             no special handling of the members is applied.))
+   $(TR $(TD $(B '+')&nbsp;/&nbsp;$(B '&nbsp;'))
+        $(TD Applies to numerical values. By default, positive numbers are not
+             formatted to include the `+` sign. With one of these two flags present,
+             positive numbers are preceded by a plus sign or a space.
+             When both flags are present, a plus sign is used.
+
+             In case of $(B 'r'), a big endian format is used.))
+   $(TR $(TD $(B '0'))
+        $(TD Is applied to numerical values that are printed right justified.
+             If the zero flag is present, the space left to the number is
+             filled with zeros instead of spaces.))
+   $(TR $(TD $(B '#'))
+        $(TD Denotes that an alternative output must be used. This depends on the type
+             to be formatted and the $(I format character) used. See the
+             sections below for more information.))
+)
+
+$(SECTION4 Width\, Precision and Separator)
+
+The $(I width) parameter specifies the minimum width of the result.
+
+The meaning of $(I precision) depends on the format indicator. For
+integers it denotes the minimum number of digits printed, for
+real numbers it denotes the number of fractional digits and for
+strings and compound types it denotes the maximum number of elements
+that are included in the output.
+
+A $(I separator) is used for formatting numbers. If it is specified,
+the output is divided into chunks of three digits, separated by a $(B
+','). The number of digits in a chunk can be given explicitly by
+providing a number or a $(B '*') after the $(B ',').
+
+In all three cases the number of digits can be replaced by a $(B
+'*'). In this scenario, the next argument is used as the number of
+digits. If the argument is a negative number, the $(I precision) and
+$(I separator) parameters are considered unspecified. For ($I width),
+the absolute value is used and the $(B '-') flag is set.
+
+The $(I separator) can also be followed by a $(B '?'). In that case,
+an additional argument is used to specify the symbol that should be
+used to separate the chunks.
+
+$(SECTION4 Position)
+
+By default, the arguments are processed in the provided order. With
+the $(I position) parameter it is possible to address arguments
+directly. It is also possible to denote a series of arguments with
+two numbers separated by $(B ':'), that are all processed in the same
+way. The second number can be omitted. In that case the series ends
+with the last argument.
+
+It's also possible to use positional arguments for $(I width), $(I
+precision) and $(I separator) by adding a number and a $(B
+'$(DOLLAR)') after the $(B '*').
 
 
 
-    $(BOOKTABLE Flags affect formatting depending on the specifier as
-    follows., $(TR $(TH Flag) $(TH Types&nbsp;affected) $(TH Semantics))
-
-    $(TR $(TD $(B '-')) $(TD numeric, bool, null, char, string, enum, pointer) $(TD Left justify the result in
-        the field.  It overrides any $(B 0) flag.))
-
-    $(TR $(TD $(B '+')) $(TD numeric) $(TD Prefix positive numbers in
-    a signed conversion with a $(B +).  It overrides any $(I space)
-    flag.))
-
-    $(TR $(TD $(B '#')) $(TD integral ($(B 'o'))) $(TD Add to
-    precision as necessary so that the first digit of the octal
-    formatting is a '0', even if both the argument and the $(I
-    Precision) are zero.))
-
-    $(TR $(TD $(B '#')) $(TD integral ($(B 'x'), $(B 'X'))) $(TD If
-       non-zero, prefix result with $(B 0x) ($(B 0X)).))
-
-    $(TR $(TD $(B '#')) $(TD floating) $(TD Always insert the decimal
-       point and print trailing zeros.))
-
-    $(TR $(TD $(B '0')) $(TD numeric) $(TD Use leading
-    zeros to pad rather than spaces (except for the floating point
-    values `nan` and `infinity`).  Ignore if there's a $(I
-    Precision).))
-
-    $(TR $(TD $(B ' ')) $(TD numeric) $(TD Prefix positive
-    numbers in a signed conversion with a space.)))
 
     $(DL
-        $(DT $(I Width))
-        $(DD
-        Only used for numeric, bool, null, char, string, enum and pointer types.
-        Specifies the minimum field width.
-        If the width is a $(B *), an additional argument of type $(B int),
-        preceding the actual argument, is taken as the width.
-        If the width is negative, it is as if the $(B -) was given
-        as a $(I Flags) character.)
-
-        $(DT $(I Precision))
-        $(DD Gives the precision for numeric conversions.
-        If the precision is a $(B *), an additional argument of type $(B int),
-        preceding the actual argument, is taken as the precision.
-        If it is negative, it is as if there was no $(I Precision) specifier.)
-
-        $(DT $(I Separator))
-        $(DD Inserts the separator symbols ',' every $(I X) digits, from right
-        to left, into numeric values to increase readability.
-        The fractional part of floating point values inserts the separator
-        from left to right.
-        Entering an integer after the ',' allows to specify $(I X).
-        If a '*' is placed after the ',' then $(I X) is specified by an
-        additional parameter to the format function.
-        Adding a '?' after the ',' or $(I X) specifier allows to specify
-        the separator character as an additional parameter.
-        )
-
         $(DT $(I FormatChar))
         $(DD
         $(DL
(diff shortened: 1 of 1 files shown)


std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
+             right justified. With the $(B '-') flag this is changed
+             to left justification.
+
+             There are two exceptions where the $(B '-') flag has a
+             different meaning: With $(B 'r') it denotes to use little
+             endian and in case of a compound indicator it means that
+             no special handling of the members is applied.))
+   $(TR $(TD $(B '+')&nbsp;/&nbsp;$(B '&nbsp;'))
+        $(TD Applies only to numbers. Normally, positiv numbers are not
+             preceded by anything. With one of these two flags present,
```
RazvanN7, line 219:
```suggestion
             formatted to include the $(B +) sign. With one of these two flags present,
```

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
+             positiv numbers are preceded by a plus sign or a space.
```
RazvanN7, line 220:
```suggestion
             positive numbers are preceded by a plus sign or a space.
```

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
+             When both flags are present, a plus sign is used.
+
+             In case of $(B 'r') it denotes to use big endian.))
```
RazvanN7, line 223:
```suggestion
             In case of $(B 'r'), a big endian format is used.))
```

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
+   $(TR $(TD $(B '0'))
+        $(TD Applies only to numbers that are printed right justified.
```
RazvanN7, line 225:
```suggestion
        $(TD Is applied to numerical values that are printed right justified.
```

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
+             If the zero flag is present, the space left to the number is
+             filled with zeros instead of spaces.))
+   $(TR $(TD $(B '#'))
+        $(TD An alternativ output shall be used. This depends on the type
```
RazvanN7, line 229:
```suggestion
        $(TD Denotes that an alternative output must be used. This depends on the type
```

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
+             sections below for more information.))
+)
 
-    $(BOOKTABLE Flags affect formatting depending on the specifier as
-    follows., $(TR $(TH Flag) $(TH Types&nbsp;affected) $(TH Semantics))
+$(SECTION4 Width\, Precision and Separator)
 
-    $(TR $(TD $(B '-')) $(TD numeric, bool, null, char, string, enum, pointer) $(TD Left justify the result in
-        the field.  It overrides any $(B 0) flag.))
+The $(I width) specifies the minimum width of the result.
```
RazvanN7, line 236:
```suggestion
The $(I width) parameter specifies the minimum width of the result.
```

RazvanN7: This sentence is kind of tautological: "The width specifies the width of the result".

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
-    formatting is a '0', even if both the argument and the $(I
-    Precision) are zero.))
+A $(I separator) is only used for formatting numbers. If it is
+specified, the output is divided into chunks of three digits,
+separated by a $(B ','), for better readability. The number of digits
+in a chunk can also be given explicitly.
 
-    $(TR $(TD $(B '#')) $(TD integral ($(B 'x'), $(B 'X'))) $(TD If
-       non-zero, prefix result with $(B 0x) ($(B 0X)).))
+In all three cases the number can be replaced by a $(B '*'). In that
```
RazvanN7, line 249:
```suggestion
In all three cases the number of digits can be replaced by a $(B '*').
```

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
+case the next argument is used as the number. If the argument is a
```
RazvanN7, line 250:
```suggestion
In this scenario, the next argument is used as the number of digits. If the argument is a
```

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
+negative number, in case of $(I width), the $(B '-') flag is applied,
```
RazvanN7, line 251:
```suggestion
negative number,  the $(I width) parameter is set to $(B '-'),
```

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
+in case of $(I precision) and $(I separator) it is, as if no
+precision or separator is specified.
```
RazvanN7, line 253: Replace the above two lines with: " whereas the $(I precision) and $(I separator) parameters are considered unspecified". Overall, the paragraph will look something like this:

"In all three cases the number of digits can be replaced by a $(B '*'). In this scenario, the next argument is used as the number of digits. If the argument is a negative number,  the $(I width) parameter is set to $(B '-'), whereas the $(I precision) and $(I separator) parameters are considered unspecified."

I hope I did not change the semantical meaning of the paragraph. What do you think?

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
 
-    $(TR $(TD $(B '#')) $(TD floating) $(TD Always insert the decimal
-       point and print trailing zeros.))
+The $(I separator) can also be followed by a $(B '?'). In that case
```
RazvanN7, line 255:
```suggestion
The $(I separator) can also be followed by a $(B '?'). In that case,
```

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
 
-    $(TR $(TD $(B '0')) $(TD numeric) $(TD Use leading
-    zeros to pad rather than spaces (except for the floating point
-    values `nan` and `infinity`).  Ignore if there's a $(I
-    Precision).))
+$(SECTION4 Position)
 
-    $(TR $(TD $(B ' ')) $(TD numeric) $(TD Prefix positive
-    numbers in a signed conversion with a space.)))
+Normally, the arguments are processed in order. With the $(I
```
RazvanN7, line 261:
```suggestion
By default, the arguments are processed in order. With the $(I
```

RazvanN7: What kind of order do you refer to? Lexical order?

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
-        The fractional part of floating point values inserts the separator
-        from left to right.
-        Entering an integer after the ',' allows to specify $(I X).
-        If a '*' is placed after the ',' then $(I X) is specified by an
-        additional parameter to the format function.
-        Adding a '?' after the ',' or $(I X) specifier allows to specify
-        the separator character as an additional parameter.
-        )
+It's also possible to use positional arguments for $(I width), $(I
+precision) and $(I separator), by adding a number and a $(B
```
RazvanN7, line 269:
```suggestion
precision) and $(I separator) by adding a number and a $(B
```

RazvanN7 (changes requested):
A couple of suggestions for future PRs, if I may:

 - use "by default" instead of "normally"
 - try to not repeat the same expression in one paragraph, but rather use synonyms. For example: you use "in this case" a lot. There were paragraphs of 5 lines that had "in this case" 6-7 times. These can be  substituted by "in this scenario", "in this situation", "in these circumstances" etc. or could be ditched entirely by carefully rephrasing the sentence.
 - aim for small and concise sentences.

std/format/package.d
berni44: The important thing is _minimum_.

std/format/package.d
berni44: From left to right (the order given in the function call).

std/format/package.d
berni44: I think everything is ok, but the width stuff. The width parameter is set to `abs(value)` and the dash-flag is set. 

I rewrote it this way:

"If the argument is a negative number, the $(I precision) and
$(I separator) parameters are considered unspecified. For ($I width)
the absolute value is used and the $(B '-') flag is set."

OK?

berni44:
> A couple of suggestions for future PRs, if I may:

I'll try. :-)

> * try to not repeat the same expression in one paragraph, but rather use synonyms.

That's not easy for me. I'm often happy when I manage to find one way to express what I want to say. Sometimes I find synonyms in the dictionary, but I'm unsure if they really express, what I want to say. For example, I wanted to write here again "I'll try", like above. The dictionary suggest "I'll struggle", but I've got the feeling, that this is something else. Same with "I'll bother". Maybe "I'll make an effort" might be good?

Or turning it the other way round: I'm quite thankful, that you are helping me here... :-)

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
+that are included in the output.
 
-    $(TR $(TD $(B '#')) $(TD integral ($(B 'o'))) $(TD Add to
-    precision as necessary so that the first digit of the octal
-    formatting is a '0', even if both the argument and the $(I
-    Precision) are zero.))
+A $(I separator) is only used for formatting numbers. If it is
+specified, the output is divided into chunks of three digits,
+separated by a $(B ','), for better readability. The number of digits
+in a chunk can also be given explicitly.
```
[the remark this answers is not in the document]
RazvanN7: I suggest specifying how you can do that. E.g.: "The number of digits in a chunk can be given explicitly by [...]"

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
 
-    $(TR $(TD $(B '#')) $(TD integral ($(B 'x'), $(B 'X'))) $(TD If
-       non-zero, prefix result with $(B 0x) ($(B 0X)).))
+In all three cases the number of digits can be replaced by a $(B
+'*'). In this scenario, the next argument is used as the number of
+digits. If the argument is a negative number, the $(I precision) and
+$(I separator) parameters are considered unspecified. For ($I width)
```
RazvanN7, line 252:
```suggestion
$(I separator) parameters are considered unspecified. For ($I width),
```

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
+the absolute value is used and the $(B '-') flag is set.
```
RazvanN7, line 253: Nice, this last sentence sounds so much better.

std/format/package.d
```diff
@@ -198,66 +198,81 @@ contain exactly one $(I format specifier), which specifies the way,
...
 
-    $(TR $(TD $(B '0')) $(TD numeric) $(TD Use leading
-    zeros to pad rather than spaces (except for the floating point
-    values `nan` and `infinity`).  Ignore if there's a $(I
-    Precision).))
+$(SECTION4 Position)
 
-    $(TR $(TD $(B ' ')) $(TD numeric) $(TD Prefix positive
-    numbers in a signed conversion with a space.)))
+By default, the arguments are processed in order. With the $(I
```
RazvanN7, line 261:
```suggestion
By default, the arguments are processed in the provided order. With the $(I
```

std/format/package.d
RazvanN7: Yeah, that's great thanks!

RazvanN7:
> Or turning it the other way round: I'm quite thankful, that you are helping me here... :-)

Don't worry about it, it's my pleasure.

std/format/package.d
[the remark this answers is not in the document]
berni44: "by providing a number or a $(B '*') after the $(B ',')." ?

RazvanN7 (approved):

RazvanN7:
Looks good! Thanks! Might need a rebase though if buildkite keeps failing.

thewilsonator:
`_D3std6format8internal6floats__T11printFloatFVbi1TdTaZQwFNaNfNkAadSQCnQCm4spec__T10FormatSpecTaZQpEQDtQDsQDoQDi12RoundingModeAyaimbZQCr: error: undefined reference to '_D3std4math4ceilFNaNbNiNeeZe'`

RazvanN7:
Obviously not related to this PR (which simply adds documentation)

berni44:
> `_D3std6format8internal6floats__T11printFloatFVbi1TdTaZQwFNaNfNkAadSQCnQCm4spec__T10FormatSpecTaZQpEQDtQDsQDoQDi12RoundingModeAyaimbZQCr: error: undefined reference to '_D3std4math4ceilFNaNbNiNeeZe'`

IMHO this is a bug in the testers. They mix different versions of Phobos whenever the branch is not based on the HEAD. Sometimes this leads to such messages. We had several of them in the past, when we split `std.format`. So this is probably due to the merging of PR #7956 between creating this PR and now. I'll rebase and it will vanish.

PR #7959, opened 2021-04-12
Make isDigit branchless

UplinkCoder:
It results in smaller code which is always good

### DIFF
--- a/std/ascii.d
+++ b/std/ascii.d
@@ -314,7 +314,7 @@ bool isUpper(dchar c) @safe pure nothrow @nogc
   +/
 bool isDigit(dchar c) @safe pure nothrow @nogc
 {
-    return '0' <= c && c <= '9';
+    return (cast(ubyte)(c - '0')) < 10;
 }
 
 ///
--- a/std/ascii.d
+++ b/std/ascii.d
@@ -30,7 +30,7 @@ $(TR $(TD Validation) $(TD
         $(LREF isWhite)
 ))
 $(TR $(TD Conversions) $(TD
-        $(LREF toLower) 
+        $(LREF toLower)
         $(LREF toUpper)
 ))
 $(TR $(TD Constants) $(TD
--- a/std/ascii.d
+++ b/std/ascii.d
@@ -30,7 +30,7 @@ $(TR $(TD Validation) $(TD
         $(LREF isWhite)
 ))
 $(TR $(TD Conversions) $(TD
-        $(LREF toLower)
+        $(LREF toLower) 
         $(LREF toUpper)
 ))
 $(TR $(TD Constants) $(TD
@@ -314,7 +314,7 @@ bool isUpper(dchar c) @safe pure nothrow @nogc
   +/
 bool isDigit(dchar c) @safe pure nothrow @nogc
 {
-    return (cast(ubyte)(c - '0')) < 10;
+    return (cast(uint)(c - '0')) < 10;
 }
 
 ///


std/ascii.d
```diff
@@ -314,7 +314,7 @@ bool isUpper(dchar c) @safe pure nothrow @nogc
   +/
 bool isDigit(dchar c) @safe pure nothrow @nogc
 {
-    return '0' <= c && c <= '9';
+    return (cast(ubyte)(c - '0')) < 10;
```
thewilsonator, line 317: this would give incorrect results for `cast(dchar)('0'+256)`

UplinkCoder: uh did I forget to change it to uint?
let me fix that

UplinkCoder:
@thewilsonator can you put an auto-merge-squash?

std/ascii.d
```diff
@@ -30,7 +30,7 @@ $(TR $(TD Validation) $(TD
         $(LREF isWhite)
 ))
 $(TR $(TD Conversions) $(TD
-        $(LREF toLower)
+        $(LREF toLower)
```
thewilsonator, line 33: no

thewilsonator (approved):

std/ascii.d
UplinkCoder: damn that webinterface. fixing

maxhaton:
On which compiler? https://d.godbolt.org/z/fMfjjYcaM code is actually slightly longer on ldc. 

https://d.godbolt.org/z/TeGq5Y1x6 The way that they don't get inlined is probably a more pressing issue

UplinkCoder:
@maxhaton i fixed the code. try it again.

maxhaton:
> @maxhaton i fixed the code. try it again.

https://d.godbolt.org/z/noh54odGo for reference. Was smaller code in reference to the source or binary?

UplinkCoder:
@maxhaton it's smaller for unoptimized builds and dmd.

thewilsonator:
I haven't look at the generated code, but in general we shouldn't be pandering for DMD at the expense of LDC an GDC. I understand this is useful for unoptimised build.

UplinkCoder:
I could also put a micro-optimisation pass into dmd to transform (v <= C1 && v >= C2) into ((v - C1) > (C1 - C2))

UplinkCoder:
There is no expense.
ldc and gdc to the rewrite I made manually automatically.
they generate the same code with and without this patch.

maxhaton:
> I could also put a micro-optimisation pass into dmd to transform (v <= C1 && v >= C2) into ((v - C1) > (C1 - C2))

I guess you could, but it's probably just more spaghetti in the backend in exchange for a speedup that most people interested wouldn't even notice because optimized dmd builds are borderline considered harmful next to LLVM and GNU

RazvanN7 (commented):
This probably needs a rebase to fix the buildkite failure.

UplinkCoder:
I can't rebase it I think because I don't know where the branch lives

thewilsonator:
it lives https://github.com/dlang/phobos/tree/UplinkCoder-patch-1

berni44:
See #7800 for a similar PR that was refuted.

dukc:
This is going to be incompatible with using some third-party overflow checking tools. Is that a problem?

UplinkCoder:
Seeing the critique in the previous PR.
I have to agree that this is unreadable code.

ljmf00:
> Seeing the critique in the previous PR.
> I have to agree that this is unreadable code.

The readability can be fixed with a comment. Also, musl libc implements it that way.

ljmf00:
It is also interesting to see how a PR move forward depending on the people :eyes:

PR #7964, opened 2021-04-13
merge stable

MartinNowak:
- Fix issue #21702 - avoid quadratic template expansion in constraints of multiple search term versions of std.algorithm.searching.startsWith & endsWith
- Fix Issue 21725 - Specifying null as bitfields variable name now fails
- Fix Issue 21721 - casting std.BigInts to built-in floating point types not allowed in `@safe` code without -preview=dip1000
- Fix Issue 21724 - std.algorithm.mutation.copy fails on overlapping arrays if the source array's pointer is less than the destination array's pointer
- Fix issue 21716: std.regex performance regression.
- Fix Issue 21758 - std.experimental.checkedint opBinaryRight with integer left-hand side does not compile for any operators except + and -

### DIFF
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -374,14 +374,22 @@ if (isInputRange!SourceRange && isOutputRange!(TargetRange, ElementType!SourceRa
         assert(tlen >= slen,
                 "Cannot copy a source range into a smaller target range.");
 
-        immutable overlaps = __ctfe || () @trusted {
+        immutable overlaps = () @trusted {
             return source.ptr < target.ptr + tlen &&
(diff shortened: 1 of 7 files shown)


n8sh (approved):

RazvanN7 (approved):

RazvanN7:
```
_D3std6format8internal6floats__T11printFloatFVbi1TdTaZQwFNaNfNkAadSQCnQCm4spec__T10FormatSpecTaZQpEQDtQDsQDoQDi12RoundingModeAyaimbZQCr: error: undefined reference to '_D3std4math4ceilFNaNbNiNeeZe'
```

Maybe this needs a rebase?

PR #7965, opened 2021-04-13
std.format: New version for formatting integral values using writeAligned

berni44:
I first tried to refactor the old implementation, but soon noticed, that this is a long path leading to nothing. So removed most of it and wrote it anew. What I kept is the handling of "%r" and the calculation of `base`, as well as some other smaller ideas.

What I skipped is: 
- Using three functions instead of one. Seems to me, that this was a relict from ancient times, not necessary anymore. 
- Support for `cent` and `ucent`. The old implementation would not have worked anyway and I see not need in supporting something, that might never be added. And if needed, it should not be a problem to make this work for `cent` too.
- Extra handling for one digit numbers. I checked this. With modern optimization techniques, this had no effect at all.

What changed:
- `%u` has been handled like `%d` and was not documented. I recently added it to the docs and handled it like `%b`, `%o` and `%x`. That means, the value is now cast to `unsigend` and treated like an unsigned, that is, not sign is printed, even if plus or space flag are present. That is consistent with the other three format characters and thus `%u` got a meaning. (I had to change a unittest though.)
- Formatting 0 with 0 precision resulted in an empty string. Two unittests checked for that, which suggests, that someone found that natural. I can't go with that. There is a big difference if I write "There are bugs" or "There are 0 bugs". I even considered this a bug, see issue 21820. So I changed these two unittests.

Speed (in ns, with ldc2 and optimization):
type | old | new
-|-|-
byte | 129 | 134
ubyte | 126 | 131
short | 161 | 157
ushort | 155  | 146
int | 235 | 233
uint | 234 | 204
long | 396 | 378
ulong | 393 | 342

I used a random sample of parameters for the format string and random numbers. Aside from bytes the new version is faster.

I also did these tests with `"%s"` and one digit values. The new version was about 1ns faster for all types (103ns vs. 102ns).

### DIFF
--- a/std/format/internal/write.d
+++ b/std/format/internal/write.d
@@ -394,6 +394,28 @@ if (is(IntegralTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     formatTest(long.min, "-9223372036854775808");
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=21777
+@safe pure unittest
+{
+    assert(format!"%20.5,d"(cast(short) 120) == "              00,120");
+    assert(format!"%20.5,o"(cast(short) 120) == "              00,170");
+    assert(format!"%20.5,x"(cast(short) 120) == "              00,078");
+    assert(format!"%20.5,2d"(cast(short) 120) == "             0,01,20");
+    assert(format!"%20.5,2o"(cast(short) 120) == "             0,01,70");
+    assert(format!"%20.5,4d"(cast(short) 120) == "              0,0120");
+    assert(format!"%20.5,4o"(cast(short) 120) == "              0,0170");
+    assert(format!"%20.5,4x"(cast(short) 120) == "              0,0078");
+    assert(format!"%20.5,2x"(3000) == "             0,0b,b8");
+    assert(format!"%20.5,4d"(3000) == "              0,3000");
+    assert(format!"%20.5,4o"(3000) == "              0,5670");
+    assert(format!"%20.5,4x"(3000) == "              0,0bb8");
+    assert(format!"%20.5,d"(-400) == "             -00,400");
+    assert(format!"%20.30d"(-400) == "-000000000000000000000000000400");
+    assert(format!"%20.5,4d"(0) == "              0,0000");
+    assert(format!"%0#.8,2s"(12345) == "00,01,23,45");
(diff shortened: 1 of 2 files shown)


Imperatorn:
> I first tried to refactor the old implementation, but soon noticed, that this is a long path leading to nothing. So removed most of it and wrote it anew. What I kept is the handling of "%r" and the calculation of `base`, as well as some other smaller ideas.
> 
> What I skipped is:
> 
> * Using three functions instead of one. Seems to me, that this was a relict from ancient times, not necessary anymore.
> * Support for `cent` and `ucent`. The old implementation would not have worked anyway and I see not need in supporting something, that might never be added. And if needed, it should not be a problem to make this work for `cent` too.
> * Extra handling for one digit numbers. I checked this. With modern optimization techniques, this had no effect at all.
> 
> What changed:
> 
> * `%u` has been handled like `%d` and was not documented. I recently added it to the docs and handled it like `%b`, `%o` and `%x`. That means, the value is now cast to `unsigend` and treated like an unsigned, that is, not sign is printed, even if plus or space flag are present. That is consistent with the other three format characters and thus `%u` got a meaning. (I had to change a unittest though.)
> * Formatting 0 with 0 precision resulted in an empty string. Two unittests checked for that, which suggests, that someone found that natural. I can't go with that. There is a big difference if I write "There are bugs" or "There are 0 bugs". I even considered this a bug, see issue 21820. So I changed these two unittests.
> 
> Speed (in ns, with ldc2 and optimization):
> 
> type	old	new
> byte	129	134
> ubyte	126	131
> short	161	157
> ushort	155	146
> int	235	233
> uint	234	204
> long	396	378
> ulong	393	342
> I used a random sample of parameters for the format string and random numbers. Aside from bytes the new version is faster.
> 
> I also did these tests with `"%s"` and one digit values. The new version was about 1ns faster for all types (103ns vs. 102ns).

Nice work, do you have any hypothesis on why bytes are slower? 🤔

berni44:
> Nice work, do you have any hypothesis on why bytes are slower?

Not really. They were a little bit faster in the 1 digit test but only about 0.3ns which could also be noise. If I should guess, I'd say, the optimizer did different optimizations. But it's out of the blue.

RazvanN7 (approved):

PR #7967, opened 2021-04-15
Rework documentation of aggregates of std.format.write

berni44:
This one was not easy for me, because I needed to understand some concepts that where new to me. Thanks to @pbackus and @FeepingCreature who helped me here. I hope, I got everything right.

I mainly used the former description from `formatValue` which was temporarily moved to `package.d` and tried to improve it (it was a little bit outdated, but still in good shape) and added two more examples.

I did not include the new overloads of `toString` with `-preview=in` from @Geod24. As far as I understood the preview concept, this should not yet be made public in the docs. If I'm wrong here, please tell me.

### DIFF
--- a/std/format/write.d
+++ b/std/format/write.d
@@ -32,12 +32,53 @@ $(SECTION3 Structs$(COMMA) Unions$(COMMA) Classes$(COMMA) and Interfaces)
 
 Aggregate types can define various `toString` functions. If this
 function takes a $(REF_ALTTEXT FormatSpec, FormatSpec, std, format,
-spec) as argument, the function decides which format characters are
-accepted. If no `toString` is defined and the aggregate is an
-$(REF_ALTTEXT input range, isInputRange, std, range, primitives), it
-is treated like a range, that is $(B 's'), $(B 'r') and a compound
-specifier are accepted. In all other cases aggregate types only
-accept $(B 's').
+spec) or a $(I format string) as argument, the function decides
+which format characters are accepted. If no `toString` is defined and
+the aggregate is an $(REF_ALTTEXT input range, isInputRange, std,
+range, primitives), it is treated like a range, that is $(B 's'), $(B
+'r') and a compound specifier are accepted. In all other cases
+aggregate types only accept $(B 's').
+
+`toString` should have one of the following signatures:
+
+---
+void toString(Writer, Char)(ref Writer w, const ref FormatSpec!Char fmt)
+void toString(Writer)(ref Writer w)
+string toString();
+---
+
+Where `Writer` is an $(REF_ALTTEXT output range, isOutputRange,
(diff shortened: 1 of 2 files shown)


std/format/write.d
```diff
@@ -32,12 +32,52 @@ $(SECTION3 Structs$(COMMA) Unions$(COMMA) Classes$(COMMA) and Interfaces)
...
+
+---
+void toString(void delegate(const(char)[]) sink, const ref FormatSpec!char fmt);
+void toString(void delegate(const(char)[]) sink, string fmt);
+void toString(void delegate(const(char)[]) sink);
+---
+
+When several of the above `toString` versions are available, the
+versions with `Writer` take precedence over the versions with a
+`sink`. `string toString()` is the last considered.
```
RazvanN7, line 67:
```suggestion
`sink`. `string toString()` has the lowest priority.
```

std/format/write.d
```diff
@@ -32,12 +32,52 @@ $(SECTION3 Structs$(COMMA) Unions$(COMMA) Classes$(COMMA) and Interfaces)
...
+
+If non of these `toString` versions are available, the aggregates
```
RazvanN7, line 69:
```suggestion
If none of the above mentioned `toString` versions are available, the aggregates
```

std/format/write.d
```diff
@@ -32,12 +32,52 @@ $(SECTION3 Structs$(COMMA) Unions$(COMMA) Classes$(COMMA) and Interfaces)
...
+will be formatted with other means, in the following order:
```
RazvanN7, line 70:
```suggestion
will be formatted by other means, in the following order:
```

std/format/write.d
```diff
@@ -302,27 +342,25 @@ their base value else.
...
 
 /**
- * Formatting a struct by defining a method `toString`, which takes an output
- * range.
- *
- * It's recommended that any `toString` using $(REF_ALTTEXT output ranges, isOutputRange, std,range,primitives)
- * use $(REF put, std,range,primitives) rather than use the `put` method of the range
- * directly.
+`structs`, `unions`, `classes` and `interfaces` can be formatted in
+several different ways. The following example show this only for `structs`.
```
RazvanN7, line 346:
```suggestion
several different ways. The following example highlights `struct` formatting, however, it applies to other aggregates as well.
```

RazvanN7, line 346:
```suggestion
several different ways. The following example highlights `struct` formatting, however, it applies to other aggregates as well.
```

std/format/write.d
```diff
@@ -302,27 +342,25 @@ their base value else.
...
+For other aggregates it's the same.
```
RazvanN7, line 347:
```suggestion

```

RazvanN7 (changes requested):

RazvanN7 (approved):

PR #7968, opened 2021-04-15
Replace equalsDigit with isClose

berni44:
`isClose` does essentially the same as `equalsDigit`, is faster and doesn't use `libc`. This change was suggested by @ibuclaw in #7942 (second to last "show resolved" parts).

### DIFF
--- a/std/math/exponential.d
+++ b/std/math/exponential.d
@@ -218,9 +218,9 @@ if (isFloatingPoint!(F) && isIntegral!(G))
 
 @safe @nogc nothrow unittest
 {
-    import std.math : equalsDigit;
+    import std.math : isClose;
 
-    assert(equalsDigit(pow(2.0L, 10L), 1024, 19));
+    assert(isClose(pow(2.0L, 10L), 1024, 1e-18));
 }
 
 // https://issues.dlang.org/show_bug.cgi?id=21601
@@ -775,9 +775,9 @@ if (isFloatingPoint!(F) && isFloatingPoint!(G))
 
 @safe @nogc nothrow unittest
 {
-    import std.math : equalsDigit;
+    import std.math : isClose;
 
-    assert(equalsDigit(pow(2.0L, 10.0L), 1024, 19));
+    assert(isClose(pow(2.0L, 10.0L), 1024, 1e-18));
 }
 
 @safe pure nothrow @nogc unittest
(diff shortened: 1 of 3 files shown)


std/math/exponential.d
```diff
@@ -1191,7 +1190,7 @@ private T expImpl(T)(T x) @safe pure nothrow @nogc
     // High resolution test (verified against GNU MPFR/Mathematica).
     assert(exp(0.5L) == 0x1.A612_98E1_E069_BC97_2DFE_FAB6_DF34p+0L);
 
-    assert(equalsDigit(exp(3.0L), E * E * E, useDigits));
+    assert(isClose(exp(3.0L), E * E * E, 1e-15));
```
ibuclaw, line 1193: If possible, would be good to keep the tests at their current precision, to protect against accidental regressions.

RazvanN7 (approved):

std/math/exponential.d
berni44: I think it is actually about the same precision. When I replace `1e-15` with `real.sizeof > double.sizeof ? 1e-16 : 1e-15`, what you probably intended, some of the tests break (e.g. the `acos`-test in `trigonometry`). When I'm not completely wrong, this is, because `isClose` takes the difference and hence the next digit into account, while `equalsDigit` doesn't, so actually I should have reduced the numbers in the tests without `useDigits` by one. (But as they all work, I think there is no need to do so, isn't it?)

ibuclaw: Ok, but the CI doesn't run on `real == double ` targets, could it be possible that we might still have to do the opposite and conditionally relax them?

berni44: I sometimes wonder, if these tests are ever run on `real == double` targets at all. With the switch from `approxEqual` to `isClose` these objections where also raised (not sure, but maybe by you) and I felt quite uneasy about this but since I didn't hear a single complain.

Having said this, I'm not attached to this and we just could relax them.

ibuclaw: They are ran infrequently by ldc usually shortly after a new release by @kinke.

I'd like to be a bit more proactive, but that will have to wait until later this year when gdc will more closely keep in sync with master, making it a bit easier to do parallel testing on different target servers.

ibuclaw: FAOD, I certainly don't mind forcing a re-review of these tests an ARM and Aarch64, the tests should ideally be ensuring that we are accurate to the highest precision we can get away with isClose. Ultimately though, it probably won't be me doing that work in a follow-up PR.

PR #7969, opened 2021-04-16
std.format: Add overview table of all combinations of types and format characters.

berni44:
Final spurt: After this there will be only one more PR before the document format marathon ends. :-)

As this table has a normative touch, repetitions are somewhat intentional.

I also fixed two nits that I overlooked in the previous PR.

### DIFF
--- a/std/format/package.d
+++ b/std/format/package.d
@@ -231,7 +231,7 @@ $(BOOKTABLE ,
              sections below for more information.))
 )
 
-$(SECTION4 Width\, Precision and Separator)
+$(SECTION4 Width$(COMMA) Precision and Separator)
 
 The $(I width) parameter specifies the minimum width of the result.
 
@@ -249,7 +249,7 @@ providing a number or a $(B '*') after the $(B ',').
 In all three cases the number of digits can be replaced by a $(B
 '*'). In this scenario, the next argument is used as the number of
 digits. If the argument is a negative number, the $(I precision) and
-$(I separator) parameters are considered unspecified. For ($I width),
+$(I separator) parameters are considered unspecified. For $(I width),
 the absolute value is used and the $(B '-') flag is set.
 
 The $(I separator) can also be followed by a $(B '?'). In that case,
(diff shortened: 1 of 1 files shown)


std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
 precision) and $(I separator) by adding a number and a $(B
 '$(DOLLAR)') after the $(B '*').
 
+$(SECTION4 Types)
+
+The following table contains detailed descriptions for every
+available combination of type and format character. To avoid
+unnecessary repetition, here are some notes that apply to all or
+several entries:
```
RazvanN7, line 277: I would reorganize this paragraph the following way:

- start by explaining what this section does. E.g: "This section describes the result of combining types with format characters. It is organized in 2 subsections: a list of general information regarding the formatting of types in the presence of format characters and a table that contains details for every available combination of type and format character".
- continue with the general description paragraph: "When formatting types, the following rules apply:".
- finish with the table: "This table contains descriptions for every possible combination of type and format character"

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+
+$(UL
+  $(LI If the format character is upper case, the outcome is identical
+       to the lower case version, but uses upper case letters.)
```
RazvanN7, line 281: "If the format character is upper case, the resulting string will be formatted using upper case letters"

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+  $(LI The default precision for floating point numbers is 6 digits.)
+  $(LI Floating point numbers are rounded according to the
+       current rounding mode.)
```
RazvanN7, line 284: What is the current rounding mode?

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+  $(LI The floating point values `NaN` and `Infinity` are formatted as
+       `nan` and `inf`, possibly preceded by a sign.)
```
RazvanN7, line 286:
```suggestion
       `nan` and `inf`, possibly preceded by a  $(B +) or $(B -) sign.)
```

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+   $(TR $(MULTIROW_CELL 3, $(I Integral))
+        $(TD $(B 's'), $(B 'd'))
+            $(TD A signed decimal number. The $(B '#') flag is ignored.)
+   )
+   $(TR $(TD $(B 'b'), $(B 'o'), $(B 'u'), $(B 'x'), $(B 'X'))
+            $(TD An unsigned binary, decimal, octal or hexadecimal number.
+
+                 In case of $(B 'o') and $(B 'x'), the $(B '#') flag
+                 denotes that the number must be preceded by `0` and `0x`, with
+                 the exception of the value 0, where this not applies. For
```
RazvanN7, line 330:
```suggestion
                 the exception of the value 0, where this does not apply. For
```

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+                 fractional digits.
+
+                 When there are no fractional digits and the $(B '#') flag
+                 is $(I not) present, the dot is omitted.
+
+                 $(I Please note: the difference between $(B 'f') and $(B 'F')
+                 is only visible for `NaN` and `Infinity`.))
+   )
+   $(TR $(TD $(B 's'), $(B 'g'), $(B 'G'))
+            $(TD Short notation: If the absolute value is larger as `10 ^^ precision`
```
RazvanN7, line 358:
```suggestion
            $(TD Short notation: If the absolute value is larger than `10 ^^ precision`
```

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+                 Inside of a compound indicator the string is surrounded by double quotes
+                 and non printable characters are escaped. This can be avoided
+                 by preceding the compound indicator with a $(B '-') flag
+                 $(LPAREN)e.g. `"%-$(LPAREN)%s%$(RPAREN)"`$(RPAREN).)
+   )
+   $(TR $(TD $(B 'r'))
+            $(TD The sequence of characters, each formatted with $(B 'r').)
+   )
+   $(TR $(TD compound)
+            $(TD Like an array of characters.)
```
RazvanN7, line 409: Doesn't "As an array of characters" sound better than "Like an array of characters"? I, personally, don't like the latter. If you agree, please replace all such occurrences with "as".

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+   )
+   $(TR $(MULTIROW_CELL 3, $(I Array))
+        $(TD $(B 's'))
+            $(TD When the elements are characters, the array is formatted like
```
RazvanN7, line 413:
```suggestion
            $(TD When the elements are characters, the array is formatted as
```

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+   $(TR $(TD $(B 'r'))
+            $(TD The sequence of the elements, each formatted with $(B 'r').)
+   )
+   $(TR $(TD compound)
+            $(TD The sequence of the elements, each formatted according to the specifications
+                 given inside of the compound specifier.)
+   )
+   $(TR $(MULTIROW_CELL 2, $(I Associative Array))
+        $(TD $(B 's'))
+            $(TD As a sequence of the elements in unpredictable order. The output is
```
RazvanN7, line 428: I see that you used "as" here. Sounds so much better.

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+            $(TD The sequence of the elements, each formatted with $(B 'r').)
+   )
+   $(TR $(TD compound)
+            $(TD The sequence of the elements, each formatted according to the specifications
+                 given inside of the compound specifier.)
+   )
+   $(TR $(MULTIROW_CELL 1, $(I Struct))
+        $(TD $(B 's'))
+            $(TD When the struct has neither an applicable `toString`
+                 nor is an input range, it is formatted like `Type(field1, field2, ...)`.)
```
RazvanN7, line 464:
```suggestion
                 nor is an input range, it is formatted as follows: `StructType(field1, field2, ...)`.)
```

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+   $(TR $(MULTIROW_CELL 1, $(I Class))
+        $(TD $(B 's'))
+            $(TD When the class has neither an applicable `toString`
+                 nor is an input range, it is formatted as the
+                 fully qualified name of the class.)
+   )
+   $(TR $(MULTIROW_CELL 1, $(I Union))
+        $(TD $(B 's'))
+            $(TD When the union has neither an applicable `toString`
+                 nor is an input range, it is formatted as it's base name.)
```
RazvanN7, line 475:
```suggestion
                 nor is an input range, it is formatted as its base name.)
```

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+   )
+   $(TR $(MULTIROW_CELL 2, $(I Pointer))
+        $(TD $(B 's'))
+            $(TD A null pointer is formatted as 'null'. All other pointers are
+                 formatted as hexadecimal numbers with format character $(B 'X').)
```
RazvanN7, line 480:
```suggestion
                 formatted as hexadecimal numbers with the format character $(B 'X').)
```

std/format/package.d
```diff
@@ -269,119 +269,242 @@ It's also possible to use positional arguments for $(I width), $(I
...
+   $(TR $(TD compound)
+            $(TD The sequence of the elements, each formatted according to the specifications
+                 given inside of the compound specifier.)
+   )
+   $(TR $(MULTIROW_CELL 1, $(I Delegate))
+        $(TD $(B 's'), $(B 'r'), compound)
+            $(TD Like the `.stringof` of this delegate treated as a string.
+
+                 $(I Please note: The implementation is currently buggy
+                 and it's use is discouraged.))
```
RazvanN7, line 502:
```suggestion
                 and its use is discouraged.))
```

RazvanN7 (changes requested):
Mind the comments. Also, have you seen the actual rendering of this table? Maybe you can post it here as I'm afraid it may not look that well, given the amound of information present in it.

std/format/package.d
berni44: This is decided by the hardware and can be queried and changed (on some computers) by using `std.math.hardware : FloatingPointControl`. If not available, default is round-to-nearest.

std/format/package.d
berni44: I was quite unsure here. The German translation is the same for both. I'll change everything to "as".

std/format/package.d
berni44: Suggestion: "Rounding of floating point numbers adheres to the rounding mode of the floating point unit, if available."

RazvanN7: Yes, that's great.

RazvanN7 (approved):

PR #7970, opened 2021-04-16
std.math: Move floating point operations into operations submodule.

berni44:
In #7924 @ibuclaw suggested to split this module in two versions, therefore I postponed this PR until now. But I think, time is pressing meanwhile, because we should finish the split before `2.097.0-beta` is out, which means, we've got only two weeks left (after the split there should be a little bit cleaning up done too).

This PR would mean: All in one module, hardware stuff (`IEEEFlags` and `FloatingPointControl`) in a separate.

@ibuclaw suggested to split this module into one containing the 'comparing' stuff and one containing the 'float manip' stuff, while the hardware stuff goes still into a separate module.

A third possibility would be to create a new top level module `std.floats` and put everything there (still with the question if it's got one, two or three submodules).

What are your opinions on this?


ibuclaw:
I'd probably concede in calling this module `std.math.operations`.
1. As per the table category in the [documentation](https://dlang.org/phobos/std_math.html).
2. They are all sort of vaguely under the same "general operations" category under IEEE 754 (except for  fma which is "arithmetic").

berni44:
> I'd probably concede in having them lumped together if we were calling this module `std.math.operations`.

That's fine for me.

> except for `fma` which is "arithmetic"

Maybe, we should put it into the "algebraic" section? IMHO it might fit there too.

berni44:
Changed the name of the submodule to `operations`.

ibuclaw (approved):

PR #7971, opened 2021-04-16
std.format: Make formatting nan and inf 100% @nogc.

berni44:
This is a first step towards making formatting floats 100% `@nogc`, by using `writeAligned`. To do so, `formatValueImpl` in `internal.write` has to pass the writer to `printFloat`. 

To avoid having to do so in all unittests too, I wrote a small wrapper for the unittests. This wrapper is not `@nogc`, but the unittests in `std.format.internal.floats` don't need to be, only the one which guards against allocating.

In this last test, I used `NoOpSink` from `std.format.package.d` as a writer, because for that test the formatted result is not important. Therefore the tests which take nan or inf as argument now return an empty string. I added one more test there, showing, that indeed no allocations happen, even when the width is larger than the buffer size.

Finally I had to make `writeAligned` `package(std.format)` scope (default in `internal.write`), to be able to access it from `internal.floats`.

PS: Meanwhile I prepared some followup PRs: The roadmap looks like this: Next one will make formatting zeros `@nogc` and add a necessary enhancement to `writeAligned` in `std.format.interal.write`. After that, there are three more PRs, that can be done in parallel: One about `%a`, one about `%e` and one about `%f` (with the last two covering the `%g` case too). After that, there will be a (not yet prepared) clean up PR which (probably among other things) removes no longer needed buffers.

### DIFF
--- a/std/format/internal/floats.d
+++ b/std/format/internal/floats.d
@@ -18,8 +18,21 @@ import std.format.spec : FormatSpec;
 
 package(std.format) enum RoundingMode { up, down, toZero, toNearestTiesToEven, toNearestTiesAwayFromZero }
 
-package(std.format) auto printFloat(T, Char)(return char[] buf, T val, FormatSpec!Char f,
-                                             RoundingMode rm = RoundingMode.toNearestTiesToEven)
+// wrapper for unittests
+private auto printFloat(T, Char)(return char[] buf, T val, FormatSpec!Char f,
+                                 RoundingMode rm = RoundingMode.toNearestTiesToEven)
+if (is(T == float) || is(T == double)
+    || (is(T == real) && (T.mant_dig == double.mant_dig || T.mant_dig == 64)))
+{
+    import std.array : appender;
+    auto w = appender!string();
+
+    auto result = printFloat(buf, w, val, f, rm);
+    return result ~ w.data;
+}
+
+package(std.format) auto printFloat(Writer, T, Char)(return char[] buf, auto ref Writer w, T val,
+    FormatSpec!Char f, RoundingMode rm = RoundingMode.toNearestTiesToEven)
 if (is(T == float) || is(T == double)
     || (is(T == real) && (T.mant_dig == double.mant_dig || T.mant_dig == 64)))
 {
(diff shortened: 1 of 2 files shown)


berni44:
I found out, that this doesn't work with `sformat`. Have to see, how this can be fixed.

berni44:
Found the bug: The `writers` were not `auto ref` and thus copied...

std/format/internal/floats.d
```diff
@@ -18,8 +18,21 @@ import std.format.spec : FormatSpec;
...
+{
+    import std.array : appender;
+    auto w = appender!string();
+
+    auto result = printFloat(buf, w, val, f, rm);
+    return result ~ w.data;
+}
+
+package(std.format) auto printFloat(Writer, T, Char)(return char[] buf, auto ref Writer w, T val,
+    FormatSpec!Char f, RoundingMode rm = RoundingMode.toNearestTiesToEven)
```
RazvanN7, line 35: Shouldn't this be `@nogc` then?

std/format/internal/floats.d
```diff
@@ -3282,62 +3283,68 @@ private auto printFloat0(bool g, Char)(return char[] buf, FormatSpec!Char f, str
 // check no allocations
 @system unittest
 {
+    import std.format : NoOpSink;
+    auto w = NoOpSink();
+
     import core.memory;
     auto stats = GC.stats;
```
RazvanN7, line 3290: So the GC does get called, but no allocations are performed. I wonder how we can trnasition to making the unittest `@nogc`.

RazvanN7:
cc @atilaneves

std/format/internal/floats.d
berni44: Not yet. It's only `nan` and `inf`. All other values do still allocate. (The plan is to make 0 @nogc in the next PR (needs a change to `writeAligned`). And then to add three more PRs about `%a`, `%e` and `%f`, which can be done in parallel. When all these PRs are through, it's completely `@nogc`.)

(Aside from this, I thought, that templates should not be marked with `@nogc` because that is inferred automatically?)

RazvanN7: > Not yet. It's only nan and inf. All other values do still allocate. (The plan is to make 0 @nogc in the next PR (needs a change to writeAligned). And then to add three more PRs about %a, %e and %f, which can be done in parallel. When all these PRs are through, it's completely @nogc.)

This should be stated in the description of this PR and all subsequent PRs need to link to this one so that reviewers can easily grasp what is happening here.

> (Aside from this, I thought, that templates should not be marked with @nogc because that is inferred automatically?)

That is true, however, for templates that of which all instantiations are `@nogc`, the attribute can be slapped there. Of course, if the template calls a user defined function then it cannot be made `@nogc`, but in this case we can slap `@nogc` to unittests.

berni44: > This should be stated in the description of this PR and all subsequent PRs need to link to this one so that reviewers can easily grasp what is happening here.

Sorry. I didn't know that, when I wrote that PR at friday. At that time I just tried to find a start into making floating points `@nogc`. During the weekend I checked out how to continue and finally ended with the four followup PRs. It didn't occur to me, that I should have added this here. (I'll do so in a moment though.)

RazvanN7 (approved):

std/format/internal/floats.d
berni44: I could do

```
struct NoOpSink
{
    void put(E)(scope const E) pure @safe @nogc nothrow {}
}

auto w = NoOpSink();
```

instead. This should use the stack if I'm not completely wrong.

atilaneves (approved):

PR #7973, opened 2021-04-16
std.math.algebraic: Add overload for three-dimensional hypotenuse

ibuclaw:
Computes the square root of the sum of the squares of x, y, and z.

### DIFF
--- a/std/math/algebraic.d
+++ b/std/math/algebraic.d
@@ -369,6 +369,86 @@ real hypot(real x, real y) @safe pure nothrow @nogc
         }
 }
 
+/***********************************************************************
+ * Calculates the distance of the point (x, y, z) from the origin (0, 0, 0)
+ * in three-dimensional space.
+ * The distance is the value of the square root of the sums of the squares
+ * of x, y, and z:
+ *
+ *      sqrt($(POWER x, 2) + $(POWER y, 2) + $(POWER z, 2))
+ *
+ * Note that the distance between two points (x1, y1, z1) and (x2, y2, z2)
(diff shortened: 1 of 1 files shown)


std/math/algebraic.d
```diff
@@ -369,6 +369,86 @@ real hypot(real x, real y) @safe pure nothrow @nogc
...
+}
+
+///
+@safe unittest
+{
+    import std.math : feqrel;
+
+    assert(hypot(1.0, 2.0, 2.0) == 3.0);
+    assert(hypot(2.0, 3.0, 6.0) == 7.0);
+    assert(hypot(1.0, 4.0, 8.0) == 9.0);
```
berni44, line 416: IMHO, this should use `isClose` - and `feqrel` isn't used at all...

berni44 (approved):
I like the idea with `ratio` here. I compared the two-dimensional version of this with the current implementation (which uses several `if` to try to achieve the same) and the correct output from `bc`. This one is much better. For example: `hypot(4.04689e+2474, 1.87428e+2461) == 4.04689e+2474` with this algorithm (all bits correct), while it is `inf` with the current implementation.

So I suggest to implement this version for two dimensions too. (And that makes me wonder, if we should use variadic arguments to make it work for arbitrary dimensions? Would be "best is worst enemy of better" though.)

std/math/algebraic.d
ibuclaw: Was a copy-pasto from the other hypot unittest block.

ibuclaw:
@berni44 Yeah, the other implementation needs some love.  For a start, it forces the user to work with x87, which makes it a poor performer.

PR #7974, opened 2021-04-16
fix Issue 21834 - std.numeric.gcd can't handle negative values

ibuclaw:
This is a pretty lame restriction.  Parameters should always be transformed to an unsigned value.

### DIFF
--- a/std/numeric.d
+++ b/std/numeric.d
@@ -2939,20 +2939,28 @@ Returns:
 typeof(Unqual!(T).init % Unqual!(U).init) gcd(T, U)(T a, U b)
 if (isIntegral!T && isIntegral!U)
 {
-    if (a == 0)
-        return b;
-    if (b == 0)
-        return a;
+    // Operate on a common type between the two arguments.
+    alias UCT = Unsigned!(CommonType!(Unqual!T, Unqual!U));
+
+    // `std.math.abs` doesn't support unsigned integers, and `T.min` is undefined.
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

ibuclaw:
Eh... you can't signed and unsigned types as parameters for `gcd`, that seems like another oversight.

thewilsonator:
please merge this when you're satisfied with it.

ibuclaw:
> please merge this when you're satisfied with it.

If you're happy with the new signature...

On reviewing this morning, I'd feel more sure if there was a mixed types test added to verify it is handled correctly.

I'll have a small dig around the git history as well to add a few more comments, as why DigitalMars uses Eulers algorithm is beyond me.

std/numeric.d
```diff
@@ -2934,49 +2934,56 @@ Params:
...
             return a;
 
-        import core.bitop : bsf;
-        import std.algorithm.mutation : swap;
+        version (DigitalMars)
+        {
+            while (b)
+            {
+                immutable t = b;
+                b = a % b;
```
berni44, line 2964: `gcd(-15,cast(ubyte) 45)` breaks here with "Error: cannot implicitly convert expression `a % cast(int)b` of type `int` to `ubyte`"

ibuclaw:
@thewilsonator @berni44 - rebased on top of other PR refactoring/fixes and added more tests.

berni44 (approved):

PR #7975, opened 2021-04-17
std.numeric: Implement least common multiple

ibuclaw:
Seemed surprising that there was a `gcd` without an `lcm`.  I wonder if anyone might know why.

### DIFF
--- a/std/numeric.d
+++ b/std/numeric.d
@@ -2959,21 +2959,27 @@ if (isIntegral!T && isIntegral!U)
     if (bx == 0)
         return ax;
 
+    return gcdImpl(ax, bx);
+}
+
+private typeof(T.init % T.init) gcdImpl(T)(T a, T b)
+if (isIntegral!T)
+{
+    pragma(inline, true);
     import core.bitop : bsf;
     import std.algorithm.mutation : swap;
 
-    immutable uint shift = bsf(ax | bx);
-    ax >>= ax.bsf;
-
+    immutable uint shift = bsf(a | b);
+    a >>= a.bsf;
     do
     {
-        bx >>= bx.bsf;
-        if (ax > bx)
-            swap(ax, bx);
-        bx -= ax;
-    } while (bx);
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):
could you add this to the table of contents comment at the top of the file.

thewilsonator:
```
std/numeric.d(3115): Warning: Ddoc: parameter count mismatch, expected 2, got 0
std/numeric.d(3115):        Note that the format is `param = description`
```

ibuclaw:
> could you add this to the table of contents comment at the top of the file. 

OK, but given how #7954 is going, I'll need to add another overload for mixed signed/unsigned.

ibuclaw:
Also ping @atilaneves to both validate the new symbol and ensure that I'm not overlooking a function with a different name. :-)

berni44:
> ```
> std/numeric.d(3115): Warning: Ddoc: parameter count mismatch, expected 2, got 0
> std/numeric.d(3115):        Note that the format is `param = description`
> ```

This is strange. Seems like it wants `a` and `b` to be added in the parameter section. That's fine. But I wonder why it doesn't complain at `gcd` for the same reason...

ibuclaw:
> This is strange. Seems like it wants `a` and `b` to be added in the parameter section. That's fine. But I wonder why it doesn't complain at `gcd` for the same reason...

My first thought is that perhaps it eludes ddoc somehow by not quite being in the right format...

In any case, I'll park this for the moment, once again I'll have to break things up into a string of PRs, as this particular function is quite a big rabbit hole.

ibuclaw:
Moved the meat of the gcd implementation into a `private gcdImpl` function, which unconditionally assumes that all qualifiers have been stripped, types have been converted to unsigned.  This saves on doing a redundant abs() operation after lcm() has taken care of it.

RazvanN7 (approved):

atilaneves (approved):

PR #7977, opened 2021-04-17
std.numeric: Support mixing integral types in gcd()

ibuclaw:
This is neccesary for fixing gcd to be able to handle negative values (Issue 21834).

### DIFF
--- a/std/numeric.d
+++ b/std/numeric.d
@@ -2928,43 +2928,44 @@ an efficient algorithm such as $(HTTPS en.wikipedia.org/wiki/Euclidean_algorithm
 or $(HTTPS en.wikipedia.org/wiki/Binary_GCD_algorithm, Stein's) algorithm.
 
 Params:
-    T = Any numerical type that supports the modulo operator `%`. If
-        bit-shifting `<<` and `>>` are also supported, Stein's algorithm will
+    a = Integer value of any numerical type that supports the modulo operator `%`.
+        If bit-shifting `<<` and `>>` are also supported, Stein's algorithm will
         be used; otherwise, Euclid's algorithm is used as _a fallback.
+    b = Integer value of any equivalent numerical type.
 
 Returns:
     The greatest common divisor of the given arguments.
  */
-T gcd(T)(T a, T b)
-if (isIntegral!T)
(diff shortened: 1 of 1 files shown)


std/numeric.d
```diff
@@ -2975,6 +2983,40 @@ if (isIntegral!T)
...
+            assert(gcd(T(9),   U(6))   == 3);
+            assert(gcd(T(3),   U(4))   == 1);
+            assert(gcd(T(32),  U(24))  == 8);
+            assert(gcd(T(5),   U(6))   == 1);
+            assert(gcd(T(54),  U(36))  == 18);
+        }
+    }
+}
+
+@safe unittest
```
berni44, line 3016: A leftover?

berni44 (approved):

std/numeric.d
ibuclaw: Oops, yes. :-)

std/numeric.d
```diff
@@ -2928,19 +2928,20 @@ an efficient algorithm such as $(HTTPS en.wikipedia.org/wiki/Euclidean_algorithm
...
 Returns:
     The greatest common divisor of the given arguments.
  */
-T gcd(T)(T a, T b)
-if (isIntegral!T)
+typeof(Unqual!(T).init % Unqual!(U).init) gcd(T, U)(T a, U b)
+if (isIntegral!T && isIntegral!U)
 {
-    static if (is(T == const) || is(T == immutable))
+    static if (is(T == const) || is(T == immutable) || is(U == const) || is(T == immutable))
```
berni44, line 2942:
```suggestion
    static if (is(T == const) || is(T == immutable) || is(U == const) || is(U == immutable))
```

ibuclaw: Ah yes, I was meant to test mixing const/immutable as well.

ibuclaw: There's no need to actually have this, using `CommonType(Unqual!T, Unqual!U)` instead for holding the intermediaries.

berni44:
Is this good to go now? I think, with my new write rights, I'm allowed to add the merge label now...

ibuclaw:
@berni44 - sure, go for it.  I'll fix it up again in the next PR for negative integers.

PR #7979, opened 2021-04-17
std.math: Move hardware support into hardware submodule.

berni44:
The style checker complains, that some public unittests would need imports (that are private inside of `IeeeFlags`).  I don't know, why that worked, when it was in `package.d`. Maybe dscanner was tweaked somehow? What should we do about this?


berni44:
The confusion arose, because there are two more public symbols, which were not listed in the overview table. (And the reason was, that I forgot to add some `version` statements.) Now it should be good to go.

ibuclaw:
Just need to force a new commit has and push again to re-trigger the auto-tester.

RazvanN7 (approved):

PR #7983, opened 2021-04-19
Move unittests from std.format.internal.write at correct place.

berni44:
Several unittests in `std.format` are scattered around. I collected all of `std.format.internal.write` that belonged to a different place and put them where they should be (e.g. next to the function they test.)

There was one unittest that I moved to `std.typecons`. The original bug resided in `std.format`, but meanwhile the implementation of `std.typecons` changed and `formatElement` isn't called there anymore. IMHO it makes no sense to keep this in `std.format`.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -1615,6 +1615,22 @@ if (distinctFieldNames!(Specs))
     assert(!(t == t));
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=8015
+@safe unittest
+{
+    struct MyStruct
+    {
+        string str;
+        @property string toStr()
+        {
+            return str;
+        }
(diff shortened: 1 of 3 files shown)


std/typecons.d
```diff
@@ -1615,6 +1615,24 @@ if (distinctFieldNames!(Specs))
     assert(!(t == t));
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=8015
+@safe unittest
+{
+    import std.typecons : Tuple;
```
RazvanN7, line 1621: This import is no longer required, right?

berni44: You are right...

RazvanN7 (approved):

PR #7988, opened 2021-04-20
Fix 20886 - Retain tempCString buffer in browse

MoonlightSentinel:
The temporary buffer returned by `tempCString` was assigned to a `char*` and hence destroyed after the statement.
This worked by accident for small URL's because the space was allocated on the stack and not overwritten before forking.

~~IDK how to add a test for this.~~

### DIFF
--- a/std/process.d
+++ b/std/process.d
@@ -4194,7 +4194,8 @@ else version (Posix)
             //args[0] = "x-www-browser".ptr;  // doesn't work on some systems
             args[0] = "xdg-open".ptr;
 
-        args[1] = url.tempCString();
+        const buffer = url.tempCString(); // Retain buffer until end of scope
+        args[1] = buffer;
         args[2] = null;
 
         auto childpid = core.sys.posix.unistd.fork();
@@ -4206,7 +4207,29 @@ else version (Posix)
         }
         if (browser)
             free(cast(void*) browser);
+
+        version (StdUnittest)
+        {
+            // Verify that the test script actually suceeds
+            int status;
+            const check = waitpid(childpid, &status, 0);
+            assert(check != -1);
+            assert(status == 0);
+        }
     }
 }
 else
     static assert(0, "os not supported");
+
+version (Windows) { /* Doesn't use BROWSER */ }
+else
+@system unittest
+{
+    import std.conv : text;
+    import std.range : repeat;
+    immutable string url = text("http://", repeat('x', 249));
+
+    TestScript prog = `if [ "$1" != "` ~ url ~ `" ]; then exit 1; fi`;
+    environment["BROWSER"] = prog.path;
+    browse(url);
+}
--- a/std/process.d
+++ b/std/process.d
@@ -4139,42 +4139,6 @@ version (Windows)
         ShellExecuteW(null, "open", url.tempCStringW(), null, null, SW_SHOWNORMAL);
     }
 }
-else version (OSX)
-{
-    import core.stdc.stdio;
-    import core.stdc.string;
-    import core.sys.posix.unistd;
-
-    void browse(scope const(char)[] url) nothrow @nogc
-    {
-        const(char)*[5] args;
-
-        auto curl = url.tempCString();
-        const(char)* browser = core.stdc.stdlib.getenv("BROWSER");
-        if (browser)
-        {   browser = strdup(browser);
-            args[0] = browser;
-            args[1] = curl;
-            args[2] = null;
-        }
-        else
-        {
-            args[0] = "open".ptr;
-            args[1] = curl;
-            args[2] = null;
-        }
-
-        auto childpid = core.sys.posix.unistd.fork();
-        if (childpid == 0)
-        {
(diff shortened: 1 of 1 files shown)


CyberShadow:
> IDK how to add a test for this.

Probably overkill but you could create a shell script that writes its arguments to a file, and set `BROWSER` to the path to said script. Some `std.process` unittests do something like this.

std/process.d
```diff
@@ -4194,7 +4194,8 @@ else version (Posix)
             //args[0] = "x-www-browser".ptr;  // doesn't work on some systems
             args[0] = "xdg-open".ptr;
 
-        args[1] = url.tempCString();
+        const buffer = url.tempCString();
```
PetarKirov, line 4197: Plea

MoonlightSentinel: Sure.

std/process.d
```diff
@@ -4210,3 +4211,16 @@ else version (Posix)
...
+
+version (Windows) { /* Doesn't use BROWSER */ }
+else
+@system unittest
+{
+    import std.conv : text;
+    import std.range : repeat;
+    immutable string url = text("http://", repeat('x', 249));
+
+    TestScript prog = `if [ "$1" != "` ~ url ~ `"; then exit 1; fi`;
```
kinke, line 4223: Missing ` ]` before `;`; so the test apparently doesn't fail correctly at the moment... :)

std/process.d
```diff
@@ -4194,7 +4194,8 @@ else version (Posix)
             //args[0] = "x-www-browser".ptr;  // doesn't work on some systems
             args[0] = "xdg-open".ptr;
 
-        args[1] = url.tempCString();
+        const buffer = url.tempCString(); // Retain buffer until end of scope
+        args[1] = buffer;
```
kinke, line 4198: The `version(OSX)` above is almost identical, except that it already used a local for the buffer and uses `open` instead of `xdg-open`. If the 2 extra array elements for Mac are superfluous, the 2 blocks could be unified.

std/process.d
RazvanN7: I think that the usage of `tempCString` is wrong here. The variable is a hack and shoudn't be necessary. It really isn't obvious why the extra variable is needed and the documentation of `tempCString` is lacking, to say the least. 

Why not use `std.string.toStringz` [1] instead?

[1] https://dlang.org/phobos/std_string.html#.toStringz

std/process.d
MoonlightSentinel: Good catch. The problem is that `browse` doesn't wait for the forked process but returns immediatly...

I could add smth like this:

```d
version (StdUnittest)
{
    int status;
    const check = waitpid(childpid, &status, 0);
    assert(check != -1);
    assert(status == 0);
}
```

std/process.d
CyberShadow: It looks right to me. AIUI, `tempCString` returns an object which owns the string, but also is implicitly convertible to the string it owns. It is a dangerous pattern (which is why it's not public, I assume). So, if you want the string to live for longer than a D statement, then it has to be given a lifetime, such as saving it to a variable.

std/process.d
MoonlightSentinel: `execvp` expects a `null` pointer as terminator for the array of arguments, so the two superfluous elements shouldn't matter.

std/process.d
MoonlightSentinel: `toStringz` always allocates a copy on the heap while `tempCString` prefers an internal buffer for small strings (< 256).
The C-string doesn't need to outlive this function, so `toStringz` is unnecessary here.

RazvanN7: Ok, I just thought that using `toStringz` achieves the same result but in a safer manner.

RazvanN7 (approved):

std/process.d
MoonlightSentinel: Added a commit to merge both implementations (to not duplicate the additional check for `StdUnittest`)

kinke: Thanks!

nordlow: Would `-dip1000` have captured this?

nordlow:
Why didn't `-dip1000` disallow?

CyberShadow:
> Why didn't `-dip1000` disallow this?

[The code wasn't marked as `@trusted`](https://forum.dlang.org/post/s685qi$4r3$1@digitalmars.com), which bypasses the check.

> Would `-dip1000` have captured this?

[Yes](https://run.dlang.io/is/LE89ny) (if `@trusted` were to be removed).

nordlow:
> [The code wasn't marked as `@trusted`](https://forum.dlang.org/post/s685qi$4r3$1@digitalmars.com), which bypasses the check.

You mean:

The code _was_ marked as `@trusted`, right?

CyberShadow:
Err, yes :)

MoonlightSentinel:
> Why didn't `-dip1000` disallow this?

https://issues.dlang.org/show_bug.cgi?id=21868

PR #7990, opened 2021-04-21
std.stdio: Remove 'under-the-hood' C I/O functions that are now declared in druntime

ibuclaw:
Unfortunately, looks like I'm going to have to deprecate and rename everything else, as they are all (unintentionally?) public. :-(

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -154,29 +154,11 @@ version (Windows)
 
 version (Posix)
 {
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

ibuclaw:
Forgot to check in the explicit casts needed for the `size == 4` path.

RazvanN7 (approved):

PR #7999, opened 2021-04-22
std.stdio: Deprecate aliases that should be internal to implementation

ibuclaw:
These should have never been public.
```
import std.stdio;
FPUTC('a', cast(_iobuf*)core.stdc.stdio.stdout);
```

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -159,36 +159,121 @@ version (Posix)
 
 version (DIGITAL_MARS_STDIO)
 {
-    alias FPUTC = _fputc_nlock;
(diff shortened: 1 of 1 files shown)


RazvanN7 (approved):

RazvanN7:
@ibuclaw This is good to go, apart that it needs a changelog entry.

n8sh:
What is the motivation for this? Name conflicts?

ibuclaw:
>  What is the motivation for this? Name conflicts?

Implementation details should **never** be public. And they are meant to be private otherwise there would be ddoc for these functions and aliases.

PR #8000, opened 2021-04-23
std.format: Make formatting 0.0 100% @nogc.

berni44:
Followup on #7971; see there for the roadmap.

The calls to `printFloat0` are replaced by a calls to `writeAligned` which does the same more general and without allocating.

`writeAligned` needed another parameter: fractional digits are yet another stretchable element due to trailing zeros. The interpretation of `precision` for floating point values differs from the interpretation of it for integral values. Therefore I added an `enum PrecisionType` to distinguish the various meanings of `precision`. Eventually it will replace the boolean parameter `integer_precision` - for the time being I added an additional wrapper.

### DIFF
--- a/std/format/internal/floats.d
+++ b/std/format/internal/floats.d
@@ -136,18 +136,18 @@ if (is(T == float) || is(T == double)
     final switch (f.spec)
     {
         case 'a': case 'A':
-            return printFloatA(buf, val, f, rm, sgn, exp, mnt, is_upper);
+            return printFloatA(buf, w, val, f, rm, sgn, exp, mnt, is_upper);
         case 'e': case 'E':
-            return printFloatE!false(buf, val, f, rm, sgn, exp, mnt, is_upper);
+            return printFloatE!false(buf, w, val, f, rm, sgn, exp, mnt, is_upper);
         case 'f': case 'F':
-            return printFloatF!false(buf, val, f, rm, sgn, exp, mnt, is_upper);
+            return printFloatF!false(buf, w, val, f, rm, sgn, exp, mnt, is_upper);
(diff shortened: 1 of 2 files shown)


std/format/internal/write.d
```diff
@@ -925,6 +925,12 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     assert(format("%a", r) == "0x1p-20");
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=21836
+@safe pure unittest
```
thewilsonator, line 929: `@nogc`?

n8sh: Not `@nogc` since some code paths in the called function allocate, even though those paths are not taken when called with the argument in the unit test.

thewilsonator: do we have an `assertGCNotCalled` test helper?

berni44: There is a `@nogc`-test in `std.format.internal.floats`. I think, it can be made `@nogc` once the three followups on this are merged and formatting floats is completely `@nogc`, not only zeros, nans and infs.

std/format/internal/floats.d
```diff
@@ -3183,109 +3198,6 @@ if (is(T == float) || is(T == double)
...
-            result[$ - 3] = '+';
-            result[$ - 4] = is_upper ? 'E' : 'e';
-        }
-    }
-
-    return result;
-}
-
 // check no allocations
 @system unittest
```
berni44, line 3202: Here is this test.

thewilsonator (approved):

PR #8001, opened 2021-04-23
Fix Issue 21853 - std.format: formatting real.max in CTFE fails

berni44:
I introduced this bug in #7862: The reason is, that `log2` rounds up (due to floating points not precise enough) when the result is very close to an integer and thus the appended `floor` won't work anymore.

I fixed this by checking for `max_exp - 1` explicitly and for all others by detecting, that the first (implied) bit of the mantissa is 0 and not 1 as it should be. In this case the result is corrected accordingly.

PS: I would really appreciate if we could have a `cast(ulong[2])` for reals in CTFE, similar to the available casts for `float` and `double`. This would safe us from such trouble.

### DIFF
--- a/std/format/internal/floats.d
+++ b/std/format/internal/floats.d
@@ -55,6 +55,8 @@ if (is(T == float) || is(T == double)
                 exp = 32767;
             else if (abs(val2) < real.min_normal)
                 exp = 0;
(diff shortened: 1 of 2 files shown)


thewilsonator (approved):

berni44:
Forgot to check for x87 reals in the unittest.

PR #8002, opened 2021-04-23
Fix Issue 21853 - std.format: formatting real.max in CTFE fails

berni44:
A copy of #8001. The bot had not noticed my forced push, so I tried to restart the testers with closing and reopening, but I'm not allowed to reopen now, because there are forced pushes...

I had to remove windows from the unittest due to issed 20972, hence to forced push.


berni44:
Found a way around the windows bug. That's better, because now it fixes the issue also for those computers.

thewilsonator (approved):

berni44:
This is good to go now. *phew*

PR #8004, opened 2021-04-24
Move bitpattern extraction from std.format to std.math.

berni44:
Extracting bitpatterns from floating point values is something, that is not only useful in `std.format`. I therefore suggest to move this to `std.math` and eventually make it public.

The new function is only `package(std)` yet, because I'd like to improve it, before going public (if at all): I think, that the implied bit of the mantissa should be added, that the bias should be subtracted from the exponent and that the stupid bit from x87s should be provided. I didn't add this here, to make the review easier and also to avoid merge conflicts with #8000 and its followups.

I think #5825 tried to add something similar, but I don't understand the details of that PR and maybe it's also outdated.

### DIFF
--- a/std/math/operations.d
+++ b/std/math/operations.d
@@ -1693,3 +1693,306 @@ if (isFloatingPoint!T)
         }
     }}
 }
+
+package(std): // not yet public
+
+struct FloatingPointBitpattern(T)
+if (isFloatingPoint!T)
+{
+    static if (T.mant_dig <= 64)
+    {
+        ulong mantissa;
+    }
+    else
+    {
+        ulong mantissa_lsb;
+        ulong mantissa_msb;
+    }
+
+    int exponent;
+    bool negative;
+}
+
+FloatingPointBitpattern!T extractBitpattern(T)(T val)
+if (isFloatingPoint!T)
+{
+    import std.math : floatTraits, RealFormat;
+
+    FloatingPointBitpattern!T ret;
+
+    alias F = floatTraits!T;
+    static if (F.realFormat == RealFormat.ieeeExtended)
+    {
+        if (__ctfe)
+        {
+            import std.math.algebraic : abs, fabs;
+            import std.math.rounding : floor;
+            import std.math.traits : isInfinity, isNaN;
+            import std.math.exponential : log2;
+
+            if (isNaN(val) || isInfinity(val))
+                ret.exponent = 32767;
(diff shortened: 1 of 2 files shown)


thewilsonator:
See also `std.bitmanip.{FloatRep,DoubleRep}`

berni44:
> See also `std.bitmanip.{FloatRep,DoubleRep}`

Unfortunately this doesn't work for reals and cannot be used in CTFE due to unions... But would be nice to have a single place and item, that works for everything... Yet, I think, this is not possible with the current environment.

ibuclaw:
> I think #5825 tried to add something similar, but I don't understand the details of that PR and maybe it's also outdated.
> 

The hope that it lowers the boundary for supporting other kinds of float. Also the float/int casts currently used in std.math cause bus errors on strict alignment targets such as SPARC.

RazvanN7 (approved):

std/math/operations.d
```diff
@@ -1693,3 +1693,306 @@ if (isFloatingPoint!T)
...
+if (isFloatingPoint!T)
+{
+    static if (T.mant_dig <= 64)
+    {
+        ulong mantissa;
+    }
+    else
+    {
+        ulong mantissa_lsb;
+        ulong mantissa_msb;
```
ibuclaw, line 1709: So this PR would immediately break 128-bit real targets then?

berni44: What do you mean with "break"? 128-bit reals are not supported by this code. If you call `FloatingPointBitpatten` with a `real` on such a target you'll get an `static assert` message.

std/math/operations.d
```diff
@@ -1693,3 +1693,306 @@ if (isFloatingPoint!T)
...
+        {
+            ulong ival = () @trusted { return *cast(uint*) &val; }();
+        }
+        else static if (F.realFormat == RealFormat.ieeeDouble)
+        {
+            ulong ival = () @trusted { return *cast(ulong*) &val; }();
+        }
+        else
+        {
+            static assert(false, "Floating point type `" ~ F.realFormat ~ "` not supported.");
```
berni44, line 1799: Here is the `static assert`.

std/math/operations.d
ibuclaw: Ok, I had missed that this type only got instantiated inside a `mant_dig == 64` block. Just threw me off that you introduced these but do nothing in with them.

PR #8008, opened 2021-04-24
Clean up after making formating floating point numbers @nogc.

berni44:
Followup on #8005, #8006 and #8007.

- Add note for the changelog.
- Remove several useless buffers
- Remove two items from note in docs of `sformat` about GC usage.

I also tried to annotate a unittest with `@nogc`, but that failed. The reason is, that there are several calls to range primitives (e.g. `put` in `writeAligned` or indirectly in `graphemeStride`, where the compiler does not manage to deduce that they are `@nogc`.

Please note that the changes in `std.format.internal.write` look like much, but aside from the first few lines it's moving everything four spaces to the right, because the `static if` doesn't need this anymore.

### DIFF
--- a/changelog/formatting_floats_nogc.dd
+++ b/changelog/formatting_floats_nogc.dd
@@ -1,0 +1,8 @@
+Floating point numbers don't allocate with the GC anymore.
+
+The implementation of formatting floating point numbers has been
+reworked. We made sure that working examples never allocate with the
+GC, however, we are still using exceptions which are GC managed.
+Therefore, code that uses formatting correctly will never allocate,
+but in the case of exceptions, the GC will be used to allocate the
+exception. We are working on DIP 1008 to solve this issue.
--- a/std/format/internal/floats.d
+++ b/std/format/internal/floats.d
@@ -19,7 +19,7 @@ import std.format.spec : FormatSpec;
 package(std.format) enum RoundingMode { up, down, toZero, toNearestTiesToEven, toNearestTiesAwayFromZero }
 
 // wrapper for unittests
-private auto printFloat(T, Char)(return char[] buf, T val, FormatSpec!Char f,
+private auto printFloat(T, Char)(T val, FormatSpec!Char f,
                                  RoundingMode rm = RoundingMode.toNearestTiesToEven)
 if (is(T == float) || is(T == double)
     || (is(T == real) && (T.mant_dig == double.mant_dig || T.mant_dig == 64)))
@@ -27,11 +27,11 @@ if (is(T == float) || is(T == double)
     import std.array : appender;
     auto w = appender!string();
 
-    auto result = printFloat(buf, w, val, f, rm);
-    return result ~ w.data;
+    printFloat(w, val, f, rm);
+    return w.data;
 }
 
-package(std.format) auto printFloat(Writer, T, Char)(return char[] buf, auto ref Writer w, T val,
+package(std.format) void printFloat(Writer, T, Char)(auto ref Writer w, T val,
     FormatSpec!Char f, RoundingMode rm = RoundingMode.toNearestTiesToEven)
 if (is(T == float) || is(T == double)
     || (is(T == real) && (T.mant_dig == double.mant_dig || T.mant_dig == 64)))
@@ -142,23 +142,27 @@ if (is(T == float) || is(T == double)
 
         f.flZero = false;
         writeAligned(w, sgn, "", (mnt == 0) ? ( is_upper ? "INF" : "inf" ) : ( is_upper ? "NAN" : "nan" ), f);
-        return buf[0 .. 0];
+        return;
     }
 
     final switch (f.spec)
     {
         case 'a': case 'A':
-            return printFloatA(buf, w, val, f, rm, sgn, exp, mnt, is_upper);
+            printFloatA(w, val, f, rm, sgn, exp, mnt, is_upper);
+            break;
         case 'e': case 'E':
-            return printFloatE!false(buf, w, val, f, rm, sgn, exp, mnt, is_upper);
+            printFloatE!false(w, val, f, rm, sgn, exp, mnt, is_upper);
+            break;
         case 'f': case 'F':
-            return printFloatF!false(buf, w, val, f, rm, sgn, exp, mnt, is_upper);
+            printFloatF!false(w, val, f, rm, sgn, exp, mnt, is_upper);
+            break;
         case 'g': case 'G':
-            return printFloatG(buf, w, val, f, rm, sgn, exp, mnt, is_upper);
+            printFloatG(w, val, f, rm, sgn, exp, mnt, is_upper);
+            break;
     }
 }
 
-private auto printFloatA(Writer, T, Char)(return char[] buf, auto ref Writer w, T val,
+private void printFloatA(Writer, T, Char)(auto ref Writer w, T val,
     FormatSpec!Char f, RoundingMode rm, string sgn, int exp, ulong mnt, bool is_upper)
 if (is(T == float) || is(T == double)
     || (is(T == real) && (T.mant_dig == double.mant_dig || T.mant_dig == 64)))
@@ -207,7 +211,7 @@ if (is(T == float) || is(T == double)
     {
         writeAligned(w, prefix[1 - sgn.length .. $], "0", ".", is_upper ? "P+0" : "p+0",
                      f, PrecisionType.fractionalDigits);
-        return buf[0 .. 0];
+        return;
     }
 
     // save integer part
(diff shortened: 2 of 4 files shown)


thewilsonator (approved):

RazvanN7:
If formatting floating point numbers is @nogc shouldn't some methods be marked as `@nogc` ?

berni44:
> If formatting floating point numbers is @nogc shouldn't some methods be marked as `@nogc` ?

I havn't found a way to do so and I don't completely understand the reason why. The test introduced in #7393 from @rainers shows (at least as I understand it), that GC is not used, meanwhile not even for arbitrary large output. The former implementation was limited to about 500 characters.

To find out, why the compiler thinks, a certain function is not `@nogc`, is rather difficult. It only tells that it thinks that the function isn't. But not more. I can go deeper and try to annotate the mentioned function with `@nogc`. And then go deeper again. That led me to the range primitives, where I had to stop this process. 

If anyone could provide a function annotated with `@nogc` (or could explain me, why this doesn't work) I'd appreciate that.

RazvanN7:
If a function is marked `@nogc` the compiler checks for 2 things: it checks that all functions that are called are marked as `nogc` and that there are no operations that call the garbage collector (you can check this file for more specifics on that [1]). For templated functions, however, `@nogc` can be inferred depending on the instantiation. There is the hammer way where you can simply mark a template as `@nogc` and force all instantiations to be `@nogc` (this is done in some places in phobos [2]).

My expectation would have been that the format function, even though it is a template, could be marked as `@nogc` as a result of your PRs. That was what I was referring to in my previous comment.

As for the unittest that you mentioned, my intuition is that the garbage collector is invoked (therefore the unittest cannot be marked as `@nogc`), however, no memory is allocated. That is why the garbage collector specific functions are used to check that no memory was actually allocated. @rainers correct me if I am wrong.

I would suggest trying to mark the functions that do not allocate as `@nogc` and see what happens. We can start from there.

[1] https://github.com/dlang/dmd/blob/master/src/dmd/nogc.d#L31
[2] https://github.com/dlang/phobos/blob/master/std/exception.d#L1039

aG0aep6G:
> If anyone could provide a function annotated with `@nogc` (or could explain me, why this doesn't work) I'd appreciate that.

You often can't use `@nogc` because something deep down might throw an exception. For example, `front` on a `string` might throw a UTFException.

berni44:
With the help of [dmd PR 12383](https://github.com/dlang/dmd/pull/12383) I got the following:

```
std/format/internal/floats.d(2971): Error: `@nogc` function `std.format.internal.floats.__unittest_L2964_C7` cannot call non-@nogc function `std.format.internal.floats.printFloat!(NoOpSink, float, dchar).printFloat`
std/format/internal/floats.d(34):          could not infer `@nogc` for `std.format.internal.floats.printFloat!(NoOpSink, float, dchar).printFloat` because:
std/format/internal/floats.d(144):          - calling `std.format.internal.write.writeAligned!(NoOpSink, string, string, string, dchar).writeAligned` which is not `@nogc`
std/format/internal/write.d(3235):            could not infer `@nogc` for `std.format.internal.write.writeAligned!(NoOpSink, string, string, string, dchar).writeAligned` because:
std/format/internal/write.d(3240):            - calling `std.format.internal.write.writeAligned!(NoOpSink, string, string, string, string, dchar).writeAligned` which is not `@nogc`
std/format/internal/write.d(3244):              could not infer `@nogc` for `std.format.internal.write.writeAligned!(NoOpSink, string, string, string, string, dchar).writeAligned` because:
std/format/internal/write.d(3264):              - calling `std.format.internal.write.getWidth!string.getWidth` which is not `@nogc`
std/format/internal/write.d(3537):                could not infer `@nogc` for `std.format.internal.write.getWidth!string.getWidth` because:
std/format/internal/write.d(3265):              - calling `std.format.internal.write.getWidth!string.getWidth` which is not `@nogc`
```

Here, the output stopped, but that far I had come without this new feature (although it would have saved me a lot of time). So I annotated `getWidth` with `@nogc` and got:

```
std/format/internal/write.d(3543): Error: `@nogc` function `std.format.internal.write.getWidth!(char[]).getWidth` cannot call non-@nogc function `std.format.internal.write.getWidth!(char[]).getWidth.all!(char[]).all`
std/algorithm/searching.d(120):          could not infer `@nogc` for `std.format.internal.write.getWidth!(char[]).getWidth.all!(char[]).all` because:
std/format/internal/write.d(3547): Error: `@nogc` function `std.format.internal.write.getWidth!(char[]).getWidth` cannot call non-@nogc function `std.uni.graphemeStride!char.graphemeStride`
std/uni/package.d(7105):          could not infer `@nogc` for `std.uni.graphemeStride!char.graphemeStride` because:
std/uni/package.d(7110):          - calling `std.uni.genericDecodeGrapheme!false.genericDecodeGrapheme!(const(char)[]).genericDecodeGrapheme` which is not `@nogc`
std/uni/package.d(6975):            could not infer `@nogc` for `std.uni.genericDecodeGrapheme!false.genericDecodeGrapheme!(const(char)[]).genericDecodeGrapheme` because:
std/uni/package.d(6999):            - calling `std.range.primitives.front!char.front` which is not `@nogc`
std/range/primitives.d(2498):              could not infer `@nogc` for `std.range.primitives.front!char.front` because:
std/range/primitives.d(2504):              - calling `std.utf.decode!(Flag.no, const(char)[]).decode` which is not `@nogc`
std/utf.d(1121):                could not infer `@nogc` for `std.utf.decode!(Flag.no, const(char)[]).decode` because:
std/utf.d(1137):                - calling `std.utf.decodeImpl!(true, Flag.no, const(char)[]).decodeImpl` which is not `@nogc`
std/utf.d(1443):                  could not infer `@nogc` for `std.utf.decodeImpl!(true, Flag.no, const(char)[]).decodeImpl` because:
```

I think this is enough. Looks like there is a long way to go...

rainers:
> I havn't found a way to do so and I don't completely understand the reason why. The test introduced in #7393 from @rainers shows (at least as I understand it), that GC is not used, meanwhile not even for arbitrary large output. The former implementation was limited to about 500 characters.

IIRC, the test was there to verify that no GC allocations happen with reasonable sizes, but it could still allocate for huge numbers, e.g. here: https://github.com/dlang/phobos/pull/7393/files#diff-0d234c6b34c08dbe8b91f67337884a45095fc2f4e6b1640bebe20cced56a3606R7066

When looking at the current code, this statement looks strange: https://github.com/dlang/phobos/blob/master/std/format/internal/write.d#L599. Is that `.dup` intentional?

berni44:
> IIRC, the test was there to verify that no GC allocations happen with reasonable sizes, but it could still allocate for huge numbers, e.g. here: https://github.com/dlang/phobos/pull/7393/files#diff-0d234c6b34c08dbe8b91f67337884a45095fc2f4e6b1640bebe20cced56a3606R7066

That's what the precursor PRs of this changed: Meanwhile no GC allocations happen, independed of sizes.

> When looking at the current code, this statement looks strange: https://github.com/dlang/phobos/blob/master/std/format/internal/write.d#L599. Is that `.dup` intentional?

Uh, that code is (old, buggy and) on the verge of being removed. It's meanwhile only used for exotic reals (128 bit and such stuff).

RazvanN7:
@berni44 How do you suggest we proceed here? Should we pull this and investigate how to make the called functions `@nogc`? If throwing exceptions is the problem, there's DIP1008 [1] which allows exceptions to be thrown in `@nogc` code, however, I doubt that that DIP is in a usable state.

[1] https://github.com/dlang/DIPs/blob/master/DIPs/other/DIP1008.md

berni44:
> @berni44 How do you suggest we proceed here? Should we pull this and investigate how to make the called functions `@nogc`? If throwing exceptions is the problem, there's DIP1008 [1] which allows exceptions to be thrown in `@nogc` code, however, I doubt that that DIP is in a usable state.

I guess, as long, as we have the problem with exceptions, we will not manage to add `@nogc` to any function. But dealing with this is beyond the scope of this PR.

Before merging this, I think we should ponder about the changelog entry. What do we want to tell users there? I fear, they will hope for being able to annotate functions with `@nogc` which won't work. On the other hand, I think, that we should communicate, that another step towards `@nogc` has been done. What is your opinion on that?

Finally, I think this PR should be merged, before 2.079.0-beta is branched off, which I expect to happen in 2 or 3 days.

RazvanN7:
> On the other hand, I think, that we should communicate, that another step towards @nogc has been done. What is your opinion on that?

I think the best way is to be sincere up-front. Something on the lines of: we made sure that working examples never allocate with the gc, however, we are still using exceptions which are gc managed. Therefore, code that uses formatting correctly will never allocate, but in the case of exceptions, the gc will be used to allocate the exception. We are working on dip1008 to solve this issue.

berni44:
> I think the best way is to be sincere up-front. Something on the lines of: we made sure that working examples never allocate with the gc, however, we are still using exceptions which are gc managed. Therefore, code that uses formatting correctly will never allocate, but in the case of exceptions, the gc will be used to allocate the exception. We are working on dip1008 to solve this issue.

OK. Done. Thanks.

RazvanN7 (approved):

PR #8009, opened 2021-04-24
fix Issue 21796 - Revert "std.file docs: Tweak dirEntries examples"

ibuclaw:
Reverts dlang/phobos#7906

1. Unit-tests should _never_ invoke the compiler (think ARM, or PPC where dmd doesn't exist).
2. PR caused a regression https://issues.dlang.org/show_bug.cgi?id=21796

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -4940,6 +4940,19 @@ foreach (DirEntry e; dirEntries("dmd-testing", SpanMode.breadth))
     writeln(e.name, "\t", e.size);
 }
 
+// Iterate over all *.d files in current directory and all its subdirectories
+auto dFiles = dirEntries("", SpanMode.depth).filter!(f => f.name.endsWith(".d"));
+foreach (d; dFiles)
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

berni44 (approved):

CyberShadow:
What happened here?

Why are we running code in `version (StdDdoc)`? Aren't many `StdDdoc` declaration stubs and are not meant to be compiled?

CyberShadow:
@berni44 @ibuclaw  Are you sure this is what caused the regression?

MoonlightSentinel:
Maybe the `test_extractor` ignores the `version (StdDoc)`? (since it caused failures for the `publictests` target)

CyberShadow:
Ahh, that does sound plausible.

MoonlightSentinel:
I'll check and probably raise a PR for the `test_extractor`

PR #8010, opened 2021-04-24
Fix Issue 21858 - std.format: centering output

berni44:
With having `writeAligned` it's not much work, to add centering output too. What do you think, is this extension worthwhile?


RazvanN7 (approved):
I like it. cc @atilaneves

atilaneves (approved):

PR #8011, opened 2021-04-25
Fix 21863 - FieldNameTuple returns emptry string for interfaces

MoonlightSentinel:
Symbols without members should produce an empty tuple. `""` is also not a member and hence requires additional special casing.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -2809,7 +2809,7 @@ template FieldNameTuple(T)
     else static if (is(T == class))
         alias FieldNameTuple = staticMap!(NameOf, T.tupleof);
     else
-        alias FieldNameTuple = AliasSeq!"";
+        alias FieldNameTuple = AliasSeq!();
 }
 
 ///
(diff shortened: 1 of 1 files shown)


Geod24 (approved):
LGTM

std/traits.d
```diff
@@ -2809,7 +2809,7 @@ template FieldNameTuple(T)
     else static if (is(T == class))
         alias FieldNameTuple = staticMap!(NameOf, T.tupleof);
     else
-        alias FieldNameTuple = AliasSeq!"";
+        alias FieldNameTuple = AliasSeq!();
```
RazvanN7, line 2812: But this is silently modifying behavior of code. The docs specifically say that "If `T` isn't a struct, class, or union, an expression tuple with an empty string is returned.". If we admit that an interface is not a class, then the current behavior is according to the spec.

If we decide that interfaces could be considered classes, then the fix should be made so that types that are not classes, unions or structs continue to return a tuple with an empty string.

RazvanN7 (changes requested):

RazvanN7:
@Geod24 This is silently modifying code behavior. The least we could do is to add a changelog entry

RazvanN7:
Also, what is the motivation of this change?

MoonlightSentinel:
Created #8014 to restrict this change to interfaces

std/traits.d
MoonlightSentinel: Interfaces  should be treated like classes because instances may have fields, they just aren't know at compile time. The current behaviour makes it pretty akward to enumerate fields using `static foreach (field; FieldNameTuple!T) ...` when `T` might be inferred as some interface.

PR #8012, opened 2021-04-25
merge stable

MartinNowak:
- Fix issue #21702 - avoid quadratic template expansion in constraints of multiple search term versions of std.algorithm.searching.startsWith & endsWith
- Fix Issue 21725 - Specifying null as bitfields variable name now fails
- Fix Issue 21721 - casting std.BigInts to built-in floating point types not allowed in `@safe` code without -preview=dip1000
- Fix Issue 21724 - std.algorithm.mutation.copy fails on overlapping arrays if the source array's pointer is less than the destination array's pointer
- Fix issue 21716: std.regex performance regression.
- Fix Issue 21758 - std.experimental.checkedint opBinaryRight with integer left-hand side does not compile for any operators except + and -
- Fix DIP1000 violation in sdt.digest.hmac
- fix Issue 21801 - std.typecons.ReplaceType does not work for `in` parameters
- Fix 20886 - Retain tempCString buffer in browse
- Merge OSX and Posix implementations of browse

### DIFF
--- a/.dscanner.ini
+++ b/.dscanner.ini
@@ -167,7 +167,6 @@ assert_without_msg="-etc.c.SQL_,\
 -std.experimental.allocator.typed,\
 -std.experimental.checkedint,\
 -std.file,\
--std.format,\
 -std.functional,\
 -std.getopt,\
 -std.internal.cstring,\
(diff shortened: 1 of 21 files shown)


MoonlightSentinel:
Blocked by  a merge conflict.

thewilsonator (approved):

PR #8014, opened 2021-04-26
Revert change to FieldNameTuple for non-interface types

MoonlightSentinel:
Partially reverts the changes made in #8011 for non-interface types to keep `AliasSeq!""` as a special error code. 

CC @RazvanN7 @Geod24

### DIFF
--- a/changelog/fieldnametuple_interface.dd
+++ b/changelog/fieldnametuple_interface.dd
@@ -1,0 +1,8 @@
+`FieldnameTuple` now returns an empty tuple for interfaces
+
+Previously `FieldNameTuple` returned `AliasSeq!""` for interfaces as done for
+non-aggregate types like `int`, `char*`, ... . This behaviour was surprising
+because an instance of that interface *may* have members that just are not
+known at compile time.
+
+`FieldNameTuple` will now return an empty `AliasSeq!()` for interfaces.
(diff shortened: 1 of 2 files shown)


RazvanN7 (approved):
A changelog entry should be added also.

atilaneves:
Why `AliasSeq!""` though?

RazvanN7:
@atilaneves That was the original implementation. I can only guess that the author wanted to make a distinction between a struct/class/union type that has 0 members (`AliasSeq!()) and a type that is not a struct/class/union (`AliasSeq!""). I find that this distinction is superfluous, however, users may rely on this behavior, therefore we need to revert the PR that changed this: #8011

RazvanN7:
ping @MoonlightSentinel . Please add a changelog entry.

MoonlightSentinel:
The original PR #2561 claimed that `AliasSeq!""` was consistent with other parts of `std.traits`...

TBH, `FieldNameTuple!int` should probably be an error instead.

RazvanN7 (approved):

MoonlightSentinel:
@RazvanN7 Added the changelog entry

RazvanN7:
`auto-merge` added. @atilaneves I will rapidly merge this as this reverts a wrong commit. If you have any suggestions to improve the status quo we can do that with subsequent PRs.

PR #8016, opened 2021-04-27
Fix issue 20756: Handle interface argument in `AllImplicitConversionTargets`.

FeepingCreature:
Replaces https://github.com/dlang/phobos/pull/7453

ping @berni44 , @RazvanN7 

This should probably go in 2.097 so we don't release an incomplete version of ImplicitConversionTargets a second time. Does that need master or stable?


RazvanN7:
Since #7881 targeted master, this should also target master.

ImplicitConversionTargets was deprecated.

RazvanN7 (approved):

berni44:
Hey great! I didn't touch the non built-ins, because I know too little about this.

FeepingCreature:
As far as I can see, the buildkite failure is spurious: there's no actual error, just deprecations; the build tries to set it to succeed despite deprecations, but it doesn't work.

RazvanN7:
Indeed, for some reason, the build hangs at some point for no apparent reason.

PR #8018, opened 2021-04-27
std.format: handle "exotic" reals by casting to double

berni44:
I'd like to get rid of using `libc` in `std.format`. Meanwhile it is only used to format "exotic" reals, that is reals which are not 64-bit-reals nor x87-reals. I suggest to handle them (until someone volunteers to implement `printFloat` for them), by downcasting them to double and formatting that double.

Advantages:
- No dependency of `libc` anymore, that is, `snprintf` is finally removed.
- And that means: no dependency of the bugs in `snprintf`.
- CTFE for all reals
- Identical results on all platforms (the results from different `libc` implementations vary)
- No more known-to-be-buggy code in the floating point specialization of `formatValueImpl`.
- Support for more than 500 characters in the output.
- Use of the new =-flag also for "exotic" reals.
- No interference with `snprintf` and `@nogc` inference.
- It might even be the case, that some "exotic" reals where not supported yet, because a `libc` implementation was missing. In that case, these reals are now supported.

Disadvantages:
- Lower precision for "exotic" reals.
- Large values of "exotic" reals are formatted as "inf".
- Might be perceived as a breaking change (by users of "exotic" reals). (I thought about starting a deprecation cycle, but have not found a good idea, how to do so: Using these reals is not deprecated nor do I want to ask the users to downcast them themselves, because if they are eventually implemented, these downcast will become bugs.)

In my opinion the advantages outweigh the disadvantages. What do you think?

### DIFF
--- a/std/format/internal/write.d
+++ b/std/format/internal/write.d
@@ -404,27 +404,9 @@ if (is(IntegralTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
 void formatValueImpl(Writer, T, Char)(auto ref Writer w, T obj, scope const ref FormatSpec!Char f)
 if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
 {
-    import std.algorithm.comparison : min;
     import std.algorithm.searching : find;
-    import std.ascii : isUpper;
     import std.format : enforceFmt;
-    import std.math.traits : isInfinity, isNaN, signbit;
     import std.range.primitives : put;
-    import std.string : indexOf, indexOfAny;
-
-    string nanInfStr(scope const ref FormatSpec!Char f, const bool nan,
-        const bool inf, const int sb, const bool up) @safe pure nothrow
-    {
-        return nan
-            ? up
-                ? sb ? "-NAN" : f.flPlus ? "+NAN" : (f.flSpace ? " NAN" : "NAN")
-                : sb ? "-nan" : f.flPlus ? "+nan" : (f.flSpace ? " nan" : "nan")
-            : inf
-                ? up
-                    ? sb ? "-INF" : f.flPlus ? "+INF" : (f.flSpace ? " INF" : "INF")
-                    : sb ? "-inf" : f.flPlus ? "+inf" : (f.flSpace ? " inf" : "inf")
-                : "";
-    }
 
     FloatingPointTypeOf!T val = obj;
     const char spec = f.spec;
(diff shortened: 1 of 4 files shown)


jondegenhardt:
This sounds good to me. The qualification being that I'm not particularly cognizant of what people might be using these exotic reals for.

RazvanN7 (approved):

n8sh:
>Might be perceived as a breaking change (by users of "exotic" reals).

This seems like a regression for anyone currently using this feature (`format` with greater-than-80-bit reals) while having no effect for anyone not currently using this feature, no?

berni44:
> > Might be perceived as a breaking change (by users of "exotic" reals).
> 
> This seems like a regression for anyone currently using this feature (`format` with greater-than-80-bit reals) while having no effect for anyone not currently using this feature, no?

It has indeed no effect on users not using this features (aside from people looking in the source code and wondering about more or less obvious bugs in the code, like @rainers did in #8008).

For users of this feature, there are pros and cons - see my list above. I can't say, what they prefer - I don't know any such user.

jondegenhardt:
It might be useful to include a couple examples of exotic reals that would have the print representation modified in the description of the PR. That would be useful for documentation.

Also - I'm wondering if it might be better to map numbers greater than `double.max` to `double.max` rather than infinity. My thought is that code written to read numeric values doesn't always handle `inf` if it's not expecting it. In these cases, `inf` might be a larger problem for affected cases than `double.max`.

Similarly, this might be a better choice for values less than `double.min`. I'm assuming that case can occur as well. But there, the choice is to map to either zero or `double.min`, and I suspect `double.min` might be preferable to zero.

berni44:
> It might be useful to include a couple examples of exotic reals that would have the print representation modified in the description of the PR. That would be useful for documentation.

To me it's unclear, which reals D supports. In `std.math` you can find some lists and recently someone pointed me to that as a reference.

I think, what I can say is, that some Sparcs have 128-bit reals. These will be truncated. I don't know if any other affected reals exist (other than in theory).

> Also - I'm wondering if it might be better to map numbers greater than `double.max` to `double.max` rather than infinity. My thought is that code written to read numeric values doesn't always handle `inf` if it's not expecting it. In these cases, `inf` might be a larger problem for affected cases than `double.max`.

I thought about that too, but it didn't occur to me to use `double.max`. I thought about something like `"<big_number>"` and rejected that, because it introduces some new symbols. But `double.max` might be a better choice.

> Similarly, this might be a better choice for values less than `double.min`. I'm assuming that case can occur as well. But there, the choice is to map to either zero or `double.min`, and I suspect `double.min` might be preferable to zero.

With `double.min` you mean `nextUp(0.0)` (there is no property for that). Difficult choice. I can't say, what I would prefer...

jondegenhardt:
> > Similarly, this might be a better choice for values less than double.min. I'm assuming that case can occur as well. B
> 
> ut there, the choice is to map to either zero or double.min, and I suspect double.min might be preferable to zero.
> 
> With double.min you mean nextUp(0.0) (there is no property for that). Difficult choice. I can't say, what I would prefer...

Sorry, I mean `double.min_normal`. Which surprising to me, is not the same as `nextUp(0.0)`. So perhaps `nextUp(0.0)` is the better choice. Not sure.

jondegenhardt:
Looking at this a bit further, perhaps my suggestion about really large exotic reals to `double.max` was off target. What I was really thinking was to map these to the largest value the new printing code supports, but not `inf`. So, that might be `real.max` rather than `double.max`. The underlying goal being to avoid mapping to `inf`.

Similarly for the smallest representable non-zero values. This might be `nextUp(0.0L)` (real) rather than `nextUp(0.0)`.

My reason for suggesting avoiding mapping to zero is that zero is often treated specially by programs that read in numbers. Not the reader code, but the code that acts on the read in values.

berni44:
> Looking at this a bit further, perhaps my suggestion about really large exotic reals to `double.max` was off target. What I was really thinking was to map these to the largest value the new printing code supports, but not `inf`. So, that might be `real.max` rather than `double.max`. The underlying goal being to avoid mapping to `inf`.

Which `real` do you mean with `real.max`? I guess x87-reals. But they are inaccessible on a computer with other reals. So I think, `double.max` might be best, we can do.

> Similarly for the smallest representable non-zero values. This might be `nextUp(0.0L)` (real) rather than `nextUp(0.0)`.

Same here.

> My reason for suggesting avoiding mapping to zero is that zero is often treated specially by programs that read in numbers. Not the reader code, but the code that acts on the read in values.

Sounds reasonable. On the other hand, some might argue with rounding modes... Don't know, what's best.

I just checked, what the conversion from `double` to `float` is doing near 0.0: It uses the current rounding mode. So users might use the rounding mode to specify, what they want to have. (Your version would be setting it to `roundUp`.)

jondegenhardt:
I don't think we are in disagreement. Just want to clarify though. I am suggesting functionality, not implementation. A different way to summarize my perspective:
* For a given number, if the current code does not print `inf`, then the new code should not either.
* For a given number, if the current code does not print zero, then the new code should not either.

For "current code", since that is coming from clib, my intent is consider the most common clib implementations as the baseline.

I'm a little less concerned about whether fallback for really large or really small is based on double or real. As your comments point out, there are a number of implementation and run-time considerations involved that I don't have experience with.

berni44:
I added the idea of @jondegenhardt - be aware though, that I cannot test this code. But I think, it's still simple enough to ensure by careful reading, that no bug crept in.

std/format/internal/write.d
```diff
@@ -453,194 +435,65 @@ if (is(FloatingPointTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
...
-                if (j > 0 && (firstLen - j) % fs.separators == 0)
-                {
-                    put(w, fs.separatorChar);
-                }
-                put(w, buf[j + firstDigit]);
-            }
+        // Numbers between the smallest representable double subnormal and 0.0
+        // are converted to the smallest representable double subnormal:
+        if (val > 0 && val < nextUp(0.0))
+            tval = nextUp(0.0);
```
jondegenhardt, line 462: `nextUp(0.0)` and `nextDown(-0.0)` can be compile time constants. e.g.
```
enum doubleLowest = nextUp(0.0);
enum doubleNegativeLowest = nextDown(-0.0);
```

This ensures no run-time calls. It also makes sense from the perspective that these are essentially derived from the  IEEE standard, so they have fixed values. Could also specify the type for clarity. e.g. `enum double xyz`. Not sure about the Phobos guideline on this.

jondegenhardt:
LGTM

kinke:
> To me it's unclear, which reals D supports. [...] I think, what I can say is, that some Sparcs have 128-bit reals. These will be truncated. I don't know if any other affected reals exist (other than in theory).

In practice, D `real` corresponds to C `long double` (modulo DMD for Windows MSVC targets). 128-bit quadruple precision isn't that exotic; it's used for 64-bit ARM (except for Mac, where Apple crippled it to double precision), incl. 64-bit Android (x86_64 too), and AFAIK, it's also the format for more recent PowerPC CPUs. So purely based on chip numbers, x87 precision is probably more exotic than quadruple precision.

PR #3833, opened 2015-11-25
PR "Tuple's toString made useful" once again.

Bolpat:
For the conversation and reasons for the reopening see [here](github.com/D-Programming-Language/phobos/pull/3594). In a nutshell, I did a wrong rebase.

From [Forum: Make Tuple.toString useful](http://forum.dlang.org/thread/srofuqyrdzphgzylbwds@forum.dlang.org):
With this, you can easily format a `Tuple` from its content, similar how to do with an array.
The `toString(delegate, FormatSpec)` function takes the formats `%(nested%)` and `%(nested%|separator%)` a bit different: The second one treats the Tuple as it was an array. The first one applies the nested format to the expanded fields, like `format(nested, tup.expand)` did.
With this, the forum question [Formatted output of range of tuples](http://forum.dlang.org/post/ntqawkupddjcijwtwovp@forum.dlang.org) gets a trivial answer: Use `"%(%(%s%| %)\n%)"` instead of `"%(%(%s %)\n%)"`.
It can be expected not to break any existing code.
(This is more or less the old message.)

PR #8020, opened 2021-04-29
Fix Issue 21808 - std.format: It should be possible to change the order of key and value of AAs.

berni44:
In the description of the issue it was considered a bug, that positional arguments cannot be used to change the order of key and value in an AA. This is not really a bug, because positional arguments are meant to select the argument of the function call and not the order of elements of a compound type.

Anyway, I like the idea of being able to change the order of key and value somehow. With the current implementation this is not possible. Further, positional arguments inside of a compound specifier make only sense for querying values like width or precision (see [issue 18599](https://issues.dlang.org/show_bug.cgi?id=18599)), but not for the whole element, because we know it already.

Although I generally dislike the abuse of features, I think, here it makes sense to use positional arguments: it feels natural and without it, it would be necessary to introduce something new, which has no use for anything else.

### DIFF
--- a/std/format/internal/write.d
+++ b/std/format/internal/write.d
@@ -1755,14 +1755,26 @@ if (is(AssocArrayTypeOf!T) && !is(T == enum) && !hasToString!(T, Char))
     enum const(Char)[] defSpec = "%s" ~ f.keySeparator ~ "%s" ~ f.seqSeparator;
     auto fmtSpec = spec == '(' ? f.nested : defSpec;
 
+    auto key_first = true;
+
     // testing correct nested format spec
     import std.format : NoOpSink;
     auto noop = NoOpSink();
     auto test = FormatSpec!Char(fmtSpec);
     enforceFmt(test.writeUpToNextSpec(noop),
         "nested format string for associative array contains no format specifier");
+    enforceFmt(test.indexStart <= 2,
+        "positional parameter in nested format string for associative array may only be 1 or 2");
+    if (test.indexStart == 2)
+        key_first = false;
+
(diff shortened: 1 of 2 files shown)


PetarKirov:
That's useful feature, nice! I find it related to being able to change the order of tuple elements, as essentially formatting an associative array `aa` comes down to formatting `aa.byPair` (which returns a range key-value tuples). Associative arrays are most likely handled specially for historical reasons, so perhaps we could simplify their implementation by using `byPair` and allowing tuple elements to be re-arranged (with the same syntax as this PR proposes)?

berni44:
> I find it related to being able to change the order of tuple elements,

Meanwhile @Bolpat wrote in the issue, that this was the source of his idea. I'm unsure if he had that in mind, when he wrote #3833 which introduced this feature (I just now realized that it's the same person who wrote that PR and the issue). If I should guess, I'd say it was not so. At least it was never documented (something we should do, if this PR is accepted). But I might be wrong on that.

> Associative arrays are most likely handled specially for historical reasons, so perhaps we could simplify their implementation by using `byPair` and allowing tuple elements to be re-arranged (with the same syntax as this PR proposes)?

I dislike this idea. The main reason is, that it would introduce a dependency on a "foreign" `toString` function. If I'm not completely wrong, currently `std.format` does not use any `toString` from outside `std.format`, but is self contained. I fear a little bit, that someone in the future (who is unaware of this connection) might change the `toString` from `std.typecons` and accidentally break something in `std.format` which was not covered by unittests.

RazvanN7 (approved):

PR #8022, opened 2021-04-30
extractBitpattern: Add implied bit to mantissa and subtract bias from exponent.

berni44:
Followup on #8004.

`extractBitpattern` now returns the mantissa with implied bit and subtracts the bias from the exponent.

In `std.format` I removed this repeating code. For `%a` I had to move part of the code ahead of the digit extraction (and alter it slightly) to keep it working. All else is straight forward.

### DIFF
--- a/std/format/internal/floats.d
+++ b/std/format/internal/floats.d
@@ -44,8 +44,6 @@ if (is(T == float) || is(T == double)
     int exp = bp.exponent;
     string sgn = bp.negative ? "-" : "";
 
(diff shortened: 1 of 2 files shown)


RazvanN7:
Since you had to modify tests, isn't this going to break existing code?

berni44:
> Since you had to modify tests, isn't this going to break existing code?

No. I introduced the tests in #8004 a few days ago and just wanted to do the change in two steps. (And it's not public yet.)

RazvanN7 (approved):

PR #8025, opened 2021-05-01
Fix Issues 21728, 21729 and 21730 (check for open file, before using trustedFread)

berni44:
It's my first PR against stable, so I hope I made no beginners mistake here.

### DIFF
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -1078,6 +1078,18 @@ Throws: `Exception` if `buffer` is empty.
         assertThrown(f.rawRead(u));
     }
 
+    // https://issues.dlang.org/show_bug.cgi?id=21728
+    @system unittest
+    {
+        import std.process : pipe;
+        import std.exception : assertThrown;
+
+        auto p = pipe();
+        p.readEnd.close;
+        ubyte[1] u;
+        assertThrown(p.readEnd.rawRead(u));
+    }
+
 /**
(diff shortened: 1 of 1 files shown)


std/stdio.d
```diff
@@ -4932,6 +4958,20 @@ private struct ChunksImpl
     f.close();
 }
 
+// Issue 21730 - null ptr dereferenced in ChunksImpl.opApply (SIGSEGV)
+@system unittest
+{
+    import std.exception : assertThrown;
+    static import std.file;
+
+    auto deleteme = testFilename();
```
berni44, line 4967: Some windows computers fail here with:

```
std.file.FileException@std\file.d(1065): C:\Users\VSSADM~1\AppData\Local\Temp\deleteme.dmd.unittest.pid1124-детка.stdio.d.4967: The process cannot access the file because it is being used by another process.
```

I have little experience with `testFilename` and even less experience with windows. What mistake do I make here?

CyberShadow: My guess would be that the following happens:

1. The file is created and is held open.
2. The unit test attempts to delete the file. This fails, and throws an exception.
3. The `scope(exit)` block runs as part of exception unwinding. It then tries to delete the file again, which again throws the exception.

berni44: Hey, thanks. I exchanged removing and closing. The original segfault still happens without the fix. Now I hope this works.

RazvanN7 (approved):

PR #8026, opened 2021-05-01
Simplify isSame

andralex:
Types can also be compared with the isSame trait. This saves a couple of seconds for the entire phobos unittest.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -1801,18 +1801,12 @@ private:
  *        a == b          - at least one rvalue (literals, enums, function calls)
  * __traits(isSame, a, b) - other cases (types, variables, functions, templates, etc.)
  */
-private template isSame(ab...)
-if (ab.length == 2)
+private template isSame(alias a, alias b)
 {
-    static if (!is(typeof(&ab[0] && &ab[1])) // at least one is an rvalue
-            && __traits(compiles, { enum isSame = ab[0] == ab[1]; })) // c-t comparable
-    {
-        enum isSame = ab[0] == ab[1];
-    }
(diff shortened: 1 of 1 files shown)


PetarKirov:
This implementation also passes the tests:

```D
template isSame(alias a, alias b)
{
    static if (__traits(compiles, a == b) && !__traits(compiles, &a))
        enum isSame = a == b;
    else
        enum isSame = __traits(isSame, a, b);
}
```

I suggest using it as it looks cleaner. In case it's not an exact replacement, then we should amend the unit test in order to check the case(s) where it differs.

andralex:
> This implementation also passes the tests:
> 
> ```d
> template isSame(alias a, alias b)
> {
>     static if (__traits(compiles, a == b) && !__traits(compiles, &a))
>         enum isSame = a == b;
>     else
>         enum isSame = __traits(isSame, a, b);
> }
> ```
> 
> I suggest using it as it looks cleaner. In case it's not an exact replacement, then we should amend the unit test in order to check the case(s) where it differs.

Awesome. Did so. Thank you!

andralex:
@PetarKirov actually I added a couple of unitests that don't pass with your implementation. For now I reverted back to the previous implementation. Let me know if you work it out!

thewilsonator (approved):

PetarKirov:
@andralex Thanks for the unit tests - I will check them out.

PetarKirov:
Actually I see that you have use a hybrid approach that used the static if condition from the old code, but 2 alias parameters from my suggestion. This was my first iteration, but I wanted to see if I can simplify the condition, but for that the test coverage was not thorough enough, so thanks for improving it.

PR #8027, opened 2021-05-02
std.process: Add scope to pipeProcessImpl command parameter

ibuclaw:
Both public `pipeProcess` and `pipeShell` functions accept a `scope` program/command parameter.  It looks like a mistake that the internal Impl function doesn't, which triggers an error if the target ABI requires that the caller destroys all arguments ([issue 21880](https://issues.dlang.org/show_bug.cgi?id=21880)).


thewilsonator (approved):

thewilsonator:
is this ready to go?

ibuclaw:
@thewilsonator yes.  Was just pending a rebuild of gdc locally.

Geod24 (approved):

PR #5288, opened 2017-03-17
Issue 13568: Add CT-checked format string overloads to std.format

ntrel:
* Add overloads for formattedWrite, formattedRead, format, sformat.

Check compile-time format strings match the given argument types. I'm using CTFE to evaluate `format(fmtStr, Args.init)` and see if an exception was thrown; the actual runtime code is unchanged. This is not optimal but makes this PR easy to implement and avoids code duplication/refactoring of `formattedWrite`. Users can start to pass format strings as compile-time arguments.

* Throw FormatError when formatValue is called with CTFE and a floating point argument.

This allows `snprintf` to be avoided in CTFE when checking format strings so floating point arguments can be checked.

@quickfur

PR #8029, opened 2021-05-02
Fix Issue 21456 - std.format does not accept enum member with string base type as template parameter

berni44:
Don't know if it wouldn't be better to have to return type just being `auto`...

### DIFF
--- a/std/format/package.d
+++ b/std/format/package.d
@@ -1344,8 +1344,8 @@ if (isSomeChar!Char)
 }
 
 /// ditto
-typeof(fmt) format(alias fmt, Args...)(Args args)
-if (isSomeString!(typeof(fmt)))
+auto format(alias fmt, Args...)(Args args)
+if (is(StringTypeOf!(typeof(fmt))))
 {
     import std.array : appender;
     import std.range.primitives : ElementEncodingType;
@@ -1397,6 +1397,25 @@ if (isSomeString!(typeof(fmt)))
     static assert(!__traits(compiles, format!"%f"(1.5, 2)));
     static assert(!__traits(compiles, format!"%s"(1.5L, 2)));
     static assert(!__traits(compiles, format!"%f"(1.5L, 2)));
+}
+
+// https://issues.dlang.org/show_bug.cgi?id=21456
+@safe pure unittest
+{
+    enum FMTS1 = "%s is %s";
+    assert(format!FMTS1("Pi", 3.14) == "Pi is 3.14");
+
+    enum
+    {
+        FMTS2 = "%s is %s"
+    }
+    assert(format!FMTS2("Pi", 3.14) == "Pi is 3.14");
+
+    enum FMTS3 : string
+    {
+        ONE = "%s is %s"
+    }
+    assert(format!(FMTS3.ONE)("Pi", 3.14) == "Pi is 3.14");
 }
 
 // called during compilation to guess the length of the


thewilsonator (approved):

thewilsonator:
> Don't know if it wouldn't be better to have to return type just being auto

if that works then yes please

std/format/package.d
```diff
@@ -1341,12 +1341,10 @@ if (isSomeChar!Char)
...
 
 /// ditto
-typeof(fmt) format(alias fmt, Args...)(Args args)
-if (isSomeString!(typeof(fmt)))
+auto format(alias fmt, Args...)(Args args)
+if (is(StringTypeOf!(typeof(fmt))))
 {
     import std.array : appender;
-    import std.range.primitives : ElementEncodingType;
-    import std.traits : Unqual;
```
RazvanN7, line 1349: Why are these imports made public?

berni44: Sorry, I needed that for stating the return type explicitly and forgot to undo this, when using `auto`.

RazvanN7 (approved):

andralex:
This is distinctly the wrong kind of generality.

I am very sorry we missed the initial support for enum types derived from strings. No we don't need to support them. Whoever has those enums may as well call `representation` and call it a day.

This pull should be undone and all support for enum strings should be deprecated.

berni44:
> This is distinctly the wrong kind of generality.
> 
> I am very sorry we missed the initial support for enum types derived from strings. No we don't need to support them. Whoever has those enums may as well call `representation` and call it a day.
> 
> This pull should be undone and all support for enum strings should be deprecated.

For me it is OK to revert this PR - I just tried to care about a bug and marking this bug as invalid would achieve my goals too.

Anyway, I would like to understand what is wrong with enum types derived from strings? If I would have known this, I would have probably marked the issue immediately as invalid and gave a reason there. Maybe you can point me to a place, where this is explained? (Or explain it yourself in brief?)

nordlow:
> This is distinctly the wrong kind of generality.
> 
> I am very sorry we missed the initial support for enum types derived from strings. No we don't need to support them. Whoever has those enums may as well call `representation` and call it a day.
> 
> This pull should be undone and all support for enum strings should be deprecated.

One great motive for ditching enum strings is that it avoids duplicated template instances.

For instance

```d
void format(alias fmt)() { pragma(msg, fmt); }
```

```d
unittest {
    format!("%s")();
    enum Fmt : string { s = "%s" }
    format!(Fmt.s)();
}
```
prints

```
%s
%s
```
meaning format is instantiated twice, completely unnecessary. This creates template bloat that slows down the compiler.

Geod24:
I am wondering, why is `fmt` an `alias` and not a simple `string` ? The only reason I can think of is to support `wstring` and `dstring`, but I wouldn't find that compelling, as `fmt` has to be CT-known, and the main reasons to have non UTF8 string types is either interacting with other APIs or micro optimizing for memory usage.

Having `fmt` be a `string` would have avoided the need for special support here. From a quick look, it also seems that reverting this will prevent types with `alias this` from working. Not that I think it's a good idea to support it (or enum format string), but if we do, it shouldn't require special code to do so.

berni44:
> I am wondering, why is `fmt` an `alias` and not a simple `string`?

Don't know. It was introduced in #5288 and not changed since. At least no unittest complains, when I replace it with `string fmt` and remove the constraint.

Geod24:
CC @ntrel

ntrel:
> why is fmt an alias and not a simple string ?

No real reason. I think I just used alias whilst working on the change as it's more flexible, then never thought to tighten it to string before submitting.

schveiguy:
> Having `fmt` be a `string` would have avoided the need for special support here.

This is probably the right answer. It should work with anything that currently works, except `wstring` and `dstring` formats. For those, you can probably use a converter, or just generate `string` in the first place.

A good first step might be reverting this, and seeing if replacing `alias` with `string` breaks anything.

schveiguy:
Note that we still would need to deprecate if we did move to accepting a string directly.

nordlow:
Are there more format specificer being `alias` that should be `string` in the helper functions for format``?

PR #8030, opened 2021-05-03
Add transition note from approxEqual to isClose.

berni44:
Like suggested in #7775 by @jondegenhardt I added the transition note from the release note and additional information on how to cope with 0.0 in the out of date note of `approxEqual` to make the transition easier.

### DIFF
--- a/std/math/operations.d
+++ b/std/math/operations.d
@@ -1006,7 +1006,11 @@ if (isFloatingPoint!(X))
 
    Warning:
         This template is considered out-dated. It will be removed from
-        Phobos in 2.106.0. Please use $(LREF isClose) instead.
+        Phobos in 2.106.0. Please use $(LREF isClose) instead. To achieve
+        a similar behaviour to `approxEqual(a, b)` use
+        `isClose(a, b, 1e-2, 1e-5)`. In case of comparing to 0.0,
+        `isClose(a, b, 0.0, eps)` should be used, where `eps`
+        represents the accepted deviation from 0.0."
 
    Params:
         value = Value to compare.


std/math/operations.d
```diff
@@ -1006,7 +1006,13 @@ if (isFloatingPoint!(X))
 
    Warning:
         This template is considered out-dated. It will be removed from
-        Phobos in 2.106.0. Please use $(LREF isClose) instead.
+        Phobos in 2.106.0. Please use $(LREF isClose) instead. To (almost)
+        keep the current behaviour of `approxEqual(a, b)` use
```
RazvanN7, line 1010: Reword the phrase to: "To achieve a similar behavior to `aproxEqual(a, b)` use .."

std/math/operations.d
```diff
@@ -1006,7 +1006,13 @@ if (isFloatingPoint!(X))
...
+        `isClose(a, b, 1e-2, 1e-5)`, but we recommend to adjust the code
+        to make it work with `isClose(a, b)` if none of the values is 0.0.
+        In case of comparing to 0.0 you should use `isClose(a, b, 0.0, eps)`
+        with a certain value `eps`. Which value to choose depends on your
+        use case.
```
RazvanN7, line 1015: I would replace the text that starts at: "[...] but we recomment to adjust the code [...]" until the end of the paragraph with: "In case of comparing to 0.0, `isClose(a, b, 0.0, eps) should be used, where `eps` represents the accepted deviation from the result".

berni44: I wrote "... from 0.0", because "the result" is already known.

RazvanN7 (approved):

n8sh (approved):

PR #8031, opened 2021-05-03
Make formatting and printing parameters scope

nordlow:
Depends on https://github.com/dlang/dmd/pull/12496

Got fed up with `write` and `writeln` not working with scoped parameters. This might enable more packages to start using `-dip1000`.

So dmd build picks up changes on this phobos banch, specifically then change from `std.stdio.writeln(T...)(T args)` to  `std.stdio.writeln(T...)(scope T args)`. This is highly indesirable when dmd, druntime and phobos reside in separate repos. I've alleviated this problem by temporarily disabling the test in https://github.com/dlang/dmd/pull/12496.

```
==============================
Test 'fail_compilation/fail19103.d' failed: 
expected:
----
fail_compilation/fail19103.d(12): Error: no property `writeln` for type `fail19103.C`, perhaps `import std.stdio;` is needed?
fail_compilation/fail19103.d(14): Error: no property `writeln` for type `S1`, perhaps `import std.stdio;` is needed?
fail_compilation/fail19103.d(16): Error: no property `writeln` for type `S2`, did you mean `std.stdio.writeln(T...)(T args)`?
----
actual:
----
fail_compilation/fail19103.d(12): Error: no property `writeln` for type `fail19103.C`, perhaps `import std.stdio;` is needed?
fail_compilation/fail19103.d(14): Error: no property `writeln` for type `S1`, perhaps `import std.stdio;` is needed?
fail_compilation/fail19103.d(16): Error: no property `writeln` for type `S2`, did you mean `std.stdio.writeln(T...)(scope T args)`?
----
diff:
----
 fail_compilation/fail19103.d(12): Error: no property `writeln` for type `fail19103.C`, perhaps `import std.stdio;` is needed?
 fail_compilation/fail19103.d(14): Error: no property `writeln` for type `S1`, perhaps `import std.stdio;` is needed?
-fail_compilation/fail19103.d(16): Error: no property `writeln` for type `S2`, did you mean `std.stdio.writeln(T...)(T args)`?
+fail_compilation/fail19103.d(16): Error: no property `writeln` for type `S2`, did you mean `std.stdio.writeln(T...)(scope T args)`?
```


std/format/internal/write.d
```diff
@@ -2127,7 +2127,10 @@ if (hasToString!(T, Char))
     }
     else static if (overload == HasToStringResult.hasSomeToString)
     {
-        static if (!noop) put(w, val.toString());
+        static if (!noop)
+            () @trusted {       // TODO remove
+                put(w, val.toString());
```
nordlow, line 2132: I'm not sure why this fails with change to `scope`.

Geod24:
> To get this in, shall I disable the failing test in dmd first have this merged and then put the dmd test back?

Better: change the test to not rely on Phobos

nordlow:
`Adress.setNameLen` fails to compile in the call to

```d
format("%s", typeid(this))
```

. Could it be because `TypeInfo.toString` is not scope?

nordlow:
> > To get this in, shall I disable the failing test in dmd first have this merged and then put the dmd test back?
> 
> Better: change the test to not rely on Phobos

It makes no sense to make `fail19103.d` independent of Phobos because it specifically reminds the user to import `std.stdio` when using `writeln`. In Phobos.

Geod24:
If you look at the issue in bugzilla, you'll see that it's not about `writeln` itself but about an accept-invalid.
Hence, the suggestion is not even relevant to the test!

nordlow:
> If you look at [the issue in bugzilla](https://issues.dlang.org/show_bug.cgi?id=19103), you'll see that it's not about `writeln` itself but about an accept-invalid.
> Hence, the suggestion is not even relevant to the test!

I don't understand. Can you elaborate?

Geod24:
I expanded on my comment in the DMD PR (https://github.com/dlang/dmd/pull/12496#issuecomment-831151579).
Let's move the conversation there.

nordlow:
Btw, what about qualify most `args` parameters as `const` aswell to reduce template bloat?

nordlow:
So it seems the parameters of `format` and `writeln` cannot be made scope when any of its parameters is a class or struct with a non-scope `toString`. I currently see two ways out of this;

1. make scope parameters be inferred scope for templated functions
2. deprecate non-scope toString members of `class` and `struct`

RazvanN7:
What are your plans with this PR @nordlow ?

nordlow:
I'm unsure about how to proceed. If scope is/will be inferred for parameters of function templates we can close this. Until we have better diagnostics for mismatched parameter qualifiers it's difficult to debug these problems. An enabling feature for these kinds of problems would be having this awesome work merged https://github.com/dlang/dmd/pull/12383

nordlow:
I believe the compiler should enforce toString to be `scope` which will effect this PR.

RazvanN7:
@nordlow According to the DIP [1], scope is inferred. Should we close this?

[1] https://github.com/dlang/DIPs/blob/master/DIPs/other/DIP1000.md#inference

nordlow:
Yes.

PR #8033, opened 2021-05-03
Simplify and improve Erase and EraseAll

andralex:
This improves memory consumed while building Phobos as follows:

Before:

Building Phobos release: KB=1571056 s=13.13
Building std.meta unittests: KB=326300 s=2.86

After:

Building Phobos release: KB=1518000 s=12.93
Building std.meta unittests: KB=323552 s=2.87

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -302,18 +302,17 @@ if (args.length >= 1)
 }
 
 /**
- * Returns an `AliasSeq` created from TList with the first occurrence,
- * if any, of T removed.
+ * Returns an `AliasSeq` created from `args[1 .. $]` with the first occurrence,
+ * if any, of `args[0]` removed.
  */
-template Erase(T, TList...)
+template Erase(args...)
+if (args.length >= 1)
 {
-    alias Erase = GenericErase!(T, TList).result;
-}
-
-/// Ditto
-template Erase(alias T, TList...)
-{
-    alias Erase = GenericErase!(T, TList).result;
+    private enum pos = staticIndexOf!(args[0], args[1 .. $]);
+    static if (pos < 0)
+        alias Erase = args[1 .. $];
+    else
+        alias Erase = AliasSeq!(args[1 .. pos + 1], args[pos + 2 .. $]);
 }
 
 ///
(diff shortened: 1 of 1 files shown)


maxhaton:
```
std/meta.d(348:22)[error]: no identifier for declarator
--
  | std/meta.d(348:48)[warn]: Empty declaration
```
DScanner doesn't like AliasAssign by the looks of things?

RazvanN7:
@andralex Are these PRs meant for merging or just to test try the idea out? I was under the impression that AliasAssign is going to be tested with phobos and if it brings any benefits we'll integrate it (possible under a switch) otherwise we will revert the addition.

If that is the case, what threshold are we considering? It seems that this addition requires 3.3% less memory when building phobos.

RazvanN7:
@maxhaton @andralex  Dscanner uses libdparse [1] for parsing, which uses dsymbol [2] to do symbol lookup. I assume that the latter needs to updated. However, we don't know at this point if AliasAssign is going to go into language.

Before we have a clear decision if AliasAssign is going to be part of the language, I suggest we do not integrate it in our standard lib.

RazvanN7:
cc @Hackerpilot

andralex:
> @andralex Are these AliasAssign PRs meant for merging or just to test/try the idea out? I was under the impression that AliasAssign is going to be tested with phobos and if it brings any benefits we'll integrate it (possibly under a switch), otherwise we will revert the addition.
> 
> If that is the case, what threshold are we considering? It seems that this addition requires 3.3% less memory when building phobos (and 0.84% less memory when building std.meta unittests).

Right now we have a catch-22: AliasAssign is not tried so we don't know if it belongs in D, and we don't know if it belongs in D so we don't test it.

These PRs are breaking the cycle. They reveal obvious advantages in simplicity that make the addition worthwhile as long as it doesn't affect negatively some metric. Virtually all PRs are net negative lines of code and the new code is much simpler.

Related: Arguably the language feature removes a limitation ("why can't I rebind an alias if I didn't use it?") rather than being a net addition. Removal of limitations has an easier path to acceptance than brand new features.

RazvanN7:
The error from the style checker seems to come from libdparse [1].

[1] https://github.com/dlang-community/libdparse/blob/6f0893f6b07ec561d382f26c81e74f7788056828/src/dparse/parser.d#L2376

n8sh (approved):

std/meta.d
```diff
@@ -302,18 +302,17 @@ if (args.length >= 1)
...
-
-/// Ditto
-template Erase(alias T, TList...)
+template Erase(args...)
+if (args.length >= 1)
 {
-    alias Erase = GenericErase!(T, TList).result;
+    private enum pos = staticIndexOf!(args[0], args[1 .. $]);
+    static if (pos < 0)
+        alias Erase = args[1 .. $];
```
nordlow, line 313: If so, what about optimizing for when `args[0]` is found at the end of `args[1 .. $]` aswell?

RazvanN7:
I have made this PR to libdparse: https://github.com/dlang-community/libdparse/pull/441 . That should fix the issue once it gets in and a new tag is released for both libdparse and dscanner.

nordlow:
> I have made this PR to libdparse: [dlang-community/libdparse#441](https://github.com/dlang-community/libdparse/pull/441) . That should fix the style issue once it gets in and a new tag is released for both libdparse and dscanner.

Are there any plans to transition dscanner to a tool that makes use of dmd as a library?

PR #8034, opened 2021-05-03
Improve and simplify NoDuplicates

andralex:
Use iteration instead of recursion, also replace EraseAllN with a simpler template.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -422,47 +422,38 @@ if (args.length >= 1)
 }
 
 /*
- * Erase any occurrence of the first `TList[0 .. N]` elements from `TList[N .. $]`.
+ * Merges `items1` assumed to consist of distinct elements with `items2` also 
+   assumed to consist of distinct elements. The result has no duplicates.
  *
  * Params:
- *   N = number of elements to delete from the `TList`
- *   TList = sequence of aliases
+ *   items1 = merge these
+ *   items2 = with these
  *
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):
do you have any compilation time statistics for this?

andralex:
@thewilsonator could not measure any difference in compilation times

n8sh (approved):

adamdruppe:
I see staticIndexOf coming up in many of these, it might be making itself into a target in the near future!

ibuclaw:
FYI, this has a rather subtle semantic change that has caused a huge amount of cascading errors in downstream libraries.

In a nutshell:
```
NoDuplicates!(immutable(T), T); // 2.098 => (immutable(T))    | 2.099 => (immutable(T), T)
NoDuplicates!(T, immutable(T)); // 2.098 => (T)               | 2.099 => (T, immutable(T))
NoDuplicates!(immutable(T), S); // 2.098 => (immutable(T), S) | 2.099 => (immutable(T), S)
NoDuplicates!(S, immutable(T)); // 2.098 => (S, T)            | 2.099 => (S, immutable(T))
```
Clearly, the 2.098 behaviour is wrong, but there was no warning here that this would happen.

andralex:
@ibuclaw hmmm, interesting. Too bad it was discovered late, otherwise we could have documented it as a bug fix.

PR #8036, opened 2021-05-03
Improve ReplaceAll

andralex:
According to `heaptrack` this improves memory consumption for this module's unittest in release mode by a couple of megabytes.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -581,30 +581,19 @@ if (args.length >= 2)
 }
 
 /**
- * Returns an `AliasSeq` created from TList with all occurrences
- * of T, if found, replaced with U.
+ * Returns an `AliasSeq` created from `args[2 .. $]` with all occurrences
+ * of `args[0]`, if any, replaced with `args[1]`.
  */
-template ReplaceAll(T, U, TList...)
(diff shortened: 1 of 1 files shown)


n8sh:
Instead of `ReplaceAll(args...)` how about keeping it as `ReplaceAll(alias T, alias U, TList...)`? Since [DMD 2.087.0: Template alias parameters now match basic types as a conversion](https://dlang.org/changelog/2.087.0.html#template_alias_matches_basic_types) that should work without either needing four declarations of `ReplaceAll` or using the varargs template parameter trick for mixing types and aliases.

n8sh (approved):

andralex:
> Instead of `ReplaceAll(args...)` how about keeping it as `ReplaceAll(alias T, alias U, TList...)`? Since [DMD 2.087.0: Template alias parameters now match basic types as a conversion](https://dlang.org/changelog/2.087.0.html#template_alias_matches_basic_types) that should work without either needing four declarations of `ReplaceAll` or using the varargs template parameter trick for mixing types and aliases.

Would love to, but there are still bugs, see forum discussion titled "What's the deal with..."

PR #8037, opened 2021-05-03
Improve Reverse

andralex:
This has no visible effect on build times but reduces max memory occupied.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -661,51 +661,13 @@ if (args.length >= 2)
 }
 
 /**
- * Returns an `AliasSeq` created from TList with the order reversed.
+ * Returns an `AliasSeq` created from `args` with the order reversed.
  */
-template Reverse(TList...)
+template Reverse(args...)
 {
-    static if (TList.length <= 1)
-    {
-        alias Reverse = TList;
-    }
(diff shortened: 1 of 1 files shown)


std/meta.d
```diff
@@ -661,51 +661,13 @@ if (args.length >= 2)
...
-    else
-    {
-        alias Reverse =
-            AliasSeq!(
-                Reverse!(TList[$/2 ..  $ ]),
-                Reverse!(TList[ 0  .. $/2]));
-    }
+    alias Reverse = AliasSeq!();
+    static foreach_reverse (arg; args)
+        Reverse = AliasSeq!(Reverse, arg);
```
CyberShadow, line 670: I did not know about this feature.

`dreg` says this works as of 2.095.0, but I see nothing about this in https://dlang.org/changelog/2.095.0.html.

CyberShadow: https://github.com/dlang/dmd/pull/11846

Somehow the changelog never made it to dlang.org, I wonder why.

n8sh (approved):

adamdruppe:
Nice foreach_reverse!

andralex:
> Nice foreach_reverse!

Yep, let the compiler do the heavy lifting!

PR #8038, opened 2021-05-03
Use mergesort for DerivedToFront

andralex:
Previous algorithm was quadratic and quite difficult to understand. This implementation uses mergesort, and takes advantage of static foreach to reduce template instantiations.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -718,17 +718,15 @@ template Reverse(TList...)
 }
 
 /**
- * Returns the type from TList that is the most derived from type T.
- * If none are found, T is returned.
+ * Returns the type from `TList` that is the most derived from type `T`.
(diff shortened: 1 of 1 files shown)


andralex:
Forgot to mention this PR also includes an improvement for MostDerived.

n8sh (approved):
Verified that the code is correct.

The old implementation of `DerivedToFront` has a bug when there are duplicate types. The new implementation does not reproduce this bug.

```D
import std.meta;

class A {}
class B : A {}
class C : B {}

pragma(msg, DerivedToFront!(A, A, A, B, B, B, C, C, C));
```
Prints prior to this PR:
>(C, B, A, A, A, A, A, A, A)

andralex:
@n8sh thanks! I added that code to the unittests.

PR #8039, opened 2021-05-03
Use alias assignment in staticMap

andralex:
There is no speedup in unittesting, but memory is reduced.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -771,29 +771,14 @@ template DerivedToFront(TList...)
     static assert(is(TL == AliasSeq!(C, B, A)));
 }
 
-private enum staticMapExpandFactor = 150;
-private string generateCases()
-{
-    string[staticMapExpandFactor] chunks;
-    chunks[0] = q{};
-    static foreach (enum i; 0 .. staticMapExpandFactor - 1)
-        chunks[i + 1] = chunks[i] ~ `F!(Args[` ~ i.stringof ~ `]),`;
-    string ret = `AliasSeq!(`;
-    foreach (chunk; chunks)
-        ret ~= `q{alias staticMap = AliasSeq!(` ~ chunk ~ `);},`;
-    return ret ~ `)`;
-}
-private alias staticMapBasicCases = AliasSeq!(mixin(generateCases()));
-
 /**
-Evaluates to $(D AliasSeq!(F!(T[0]), F!(T[1]), ..., F!(T[$ - 1]))).
+Evaluates to `AliasSeq!(fun!(args[0]), fun!(args[1]), ..., fun!(args[$ - 1]))`.
  */
-template staticMap(alias F, Args ...)
+template staticMap(alias fun, args...)
 {
-    static if (Args.length < staticMapExpandFactor)
-        mixin(staticMapBasicCases[Args.length]);
-    else
-        alias staticMap = AliasSeq!(staticMap!(F, Args[0 .. $/2]), staticMap!(F, Args[$/2 .. $]));
+    alias staticMap = AliasSeq!();
+    static foreach (arg; args)
+        staticMap = AliasSeq!(staticMap, fun!arg);
 }
 
 ///


n8sh (approved):

adamdruppe:
I'm a little bit more skeptical here because that old staticMap represented a lot of rounds of profiling and optimization. Did you test this with long items (like staticMap over 1000 items?) or with a variety of items (like 1000 different lengths)?

The old code is ugly but it did a good job in all these cases.

maxhaton:
> There is no speedup in unittesting, but memory is reduced.

@andralex #7756 

See also the # of elements vs time graphs I link to in the thread.

adamdruppe:
Those are different implementations to this one though. Do you still have the test rig so we can run it again this implementation?

andralex:
@adamdruppe just tested on a large build (20 GB, 70 seconds) and saw no measurable change in time and 0.5% memory consumption reduction.

adamdruppe:
Very interesting. Well, 0.5% of 20 GB  isn't nothing... not bad.

maxhaton:
@andralex @atilaneves `AliasAssign` is in the spec now, merge?

atilaneves:
> @andralex @atilaneves `AliasAssign` is in the spec now, merge?

Red CI.

adamdruppe:
Add this unittest

```
template id(alias what) {
        enum id = __traits(identifier, what);
}

enum A { a }

static assert(staticMap!(id, A.a) == AliasSeq!("a"));
```

and it will catch it.

fix is to use index like this in the loop:

```
template staticMap(alias fun, args...)
{
    alias staticMap = AliasSeq!();
    static foreach (idx, arg; args)
        staticMap = AliasSeq!(staticMap, fun!(args[idx])); // notice the index use
}
```


It is because going over values is different than going over variables. enums are weird in that they're a little of both. (btw if i was dictator im not sure i'd support this case. like either make it one or the other, require user explictness somehow. but meh idk)

adamdruppe:
That's a significant performance pessimization... we should confirm that...

andralex:
The version currently proposed is reminiscent of `quickSort`, which uses a quadratic insertion algorithm for small arrays and otherwise recurses:

```d
    template staticMap(alias fun, args...)
    {
        static if (args.length <= 8)
        {
            alias staticMap = AliasSeq!();
            static foreach (i; 0 .. args.length)
                staticMap = AliasSeq!(staticMap, fun!(args[i]));
        }
        else
        {
            alias staticMap = AliasSeq!(staticMap!(fun, args[0 .. $ / 2]), staticMap!(fun, args[$ / 2 .. $]));
        }
    }
```

This version beats the current version on @schveiguy's toughest test as follows. Existing version:

```
2.16user 0.29system 0:03.04elapsed 80%CPU (0avgtext+0avgdata 406452maxresident)k
45120inputs+4760outputs (821major+104640minor)pagefaults 0swaps
```

Proposed version:

```
2.12user 0.35system 0:03.04elapsed 81%CPU (0avgtext+0avgdata 345428maxresident)k
45120inputs+4760outputs (821major+89439minor)pagefaults 0swaps
```

schveiguy:
Hate to say this, but isn't this the same as it was before the current version? That is, https://github.com/dlang/phobos/commit/215a49432f2522963aca4b0e0a1c6394b04b7dfe

We should find out why it was changed away from that in the first place. Maybe there's a case that we're not considering. Ping @John-Colvin

andralex:
@schveiguy it's quite different innit? Unrolling is done manually, there are multiple tests etc. I wouldn't call the implementations comparable.

schveiguy:
Well, what I mean is that the recursion threshold is the same as it was. If that has anything to do with why the current code was introduced in the first place, we should make sure we don't cause a regression. John changed the threshold for a reason.

andralex:
Spoke to @WalterBright just now, he acknowledged that the growth pattern `x = AliasSeq!(x, one_more_element)` is quadratic. He has an idea on how to detect that pattern and optimize it so as to avoid copying and use exponential growth. However, he's busy with the C compiler so @nordlow if this is something you could look at that'd be positively awesome.

I propose we use this in the interim as an intermediate solution that is simple and effective.

nordlow:
Now that https://github.com/dlang/dmd/pull/14332 has been merged it might be worth reevaluating the benchmarks and maybe switch to alternative implementation of `staticMap`. FYI, @schveiguy @John-Colvin, @andralex.

PR #8040, opened 2021-05-03
Use alias assignment in Filter

andralex:
Code is greatly simplified, memory consumed is reduced.


n8sh:
```console
---
  | std/meta.d(885:20)[error]: no identifier for declarator
  | std/meta.d(885:44)[warn]: Empty declaration
  | posix.mak:577: recipe for target 'dscanner' failed
```

The style checker seems not to like this idiom.

n8sh (approved):

dukc (changes requested):
Wait a minute. Isn't alias assigment currently an experimental feature? If it is rejected or changed later on, we will have problems.

dukc:
Oh I see that you explained using experimental features in #8033. I assume you have discussed with Walter and/or Atila about the issue?

andralex:
> Oh I see that you explained using experimental features in #8033. I assume you have discussed with Walter and/or Atila about the issue?

Yes, @WalterBright and @atilaneves are in the loop. Essentially we can't get an estimate on how well the feature could fare outside the stdlib because the stdlib would be its primary beneficiary (which would then improve the life of client code).

The PRs are intentionally bite-sized so we can easily dustmite and pinpoint performance problems back to them.

adamdruppe:
Worst case you could always revert the specific commits that use this if the feature were removed too. Probably best to keep a list of them just in case.

adamdruppe:
It took me a sec to realize what the recursive AliasSeq assignment was with the eponymous alias but I get it now and indeed this is really quite decent. Cool.

wilzbach:
Someone would need to upgrade/improve/fix dscanner before this can be used.

nordlow:
> Someone would need to upgrade/improve/fix dscanner before this can be used.

That has been fixed. :)

PR #8043, opened 2021-05-04
std.utf: Add isValidCodepoint.

berni44:
For improving formatting characters with `std.format` I need this function (e.g. fixing [issue 21815](https://issues.dlang.org/show_bug.cgi?id=21815), but can also be used for [issue 6125](https://issues.dlang.org/show_bug.cgi?id=6125)). I could add it as a private function to `std.format`, but I think, it fits better in here and might be useful for users too. Especially, because the `dchar` version already exists here as `isValidDchar`, while `isValidWchar` and `isValidChar` are missing (and it's not D like to use these functions instead of a single template).

I did not manage to find out, why `isValidDchar` was designed like this. It was added in 2007 to git, but probably existed prior to that somewhere else. Eventually, `isValidCharacter` might replace `isValidDchar`. (Should I do so immediately?)

cc @atilaneves

### DIFF
--- a/std/utf.d
+++ b/std/utf.d
@@ -41,6 +41,7 @@ $(TR $(TD Index) $(TD
 ))
 $(TR $(TD Validation) $(TD
     $(LREF isValidDchar)
+    $(LREF isValidCodepoint)
     $(LREF validate)
 ))
 $(TR $(TD Miscellaneous) $(TD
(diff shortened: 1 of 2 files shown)


dukc:
What's the problem with `isValidDChar`? `char` and `wchar` convert implicitly to `dchar` so the old function can be used with all character types.

I suggest a different name. This function is named as if it was testing the validity of a code unit. `isCodePoint` would be much better.

aG0aep6G:
> What's the problem with `isValidDChar`? `char` and `wchar` convert implicitly to `dchar` so the old function can be used with all character types.

`dchar(0x80)` represents the code point U+0080. `char(0x80)` does not represent a code point. Having `char` and `wchar` implicitly convert to `dchar` is a mistake at the language level. `isValidDchar` does not give sensible results for all character types.

dukc:
> `dchar(0x80)` represents the code point U+0080. `char(0x80)` does not represent a code point. Having `char` and `wchar` implicitly convert to `dchar` is a mistake at the language level.

Or at least they should convert to `dchar.init` when the source has only a partial code point. But anyway, this explains why we might want a new function.

std/utf.d
```diff
@@ -306,6 +307,54 @@ pure nothrow @safe @nogc unittest
     });
 }
 
+/**
+Checks if a single character forms a valid code point.
```
atilaneves, line 311: This here already suggests to me that the name of the function should be `isValidCodepoint`.

dukc:
@berni44  You got tired of waiting for review?

@atilaneves is there anything anymore blocking merging this right away?

thewilsonator (approved):

thewilsonator:
Sorry for the late review. Looks fine.

MoonlightSentinel:
The changelog was not updated to the new name!

https://issues.dlang.org/show_bug.cgi?id=22503

PR #8044, opened 2021-05-04
Use alias assignment in Repeat

andralex:
No difference in memory taken or compilation time. To make sure I added a really large Repeat to the unittest.

### DIFF
--- a/std/meta.d
+++ b/std/meta.d
@@ -1465,59 +1465,30 @@ private template SmartAlias(T...)
 }
 
 /**
- * Creates an `AliasSeq` which repeats `TList` exactly `n` times.
+ * Creates an `AliasSeq` which repeats `items` exactly `n` times.
  */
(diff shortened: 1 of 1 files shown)


std/meta.d
```diff
@@ -1543,6 +1514,8 @@ template Repeat(size_t n, TList...)
 
     alias ImInt10 = Repeat!(10, int);
     static assert(is(ImInt10 == AliasSeq!(int, int, int, int, int, int, int, int, int, int)));
+
+    alias Big = Repeat!(1,000,000, int);
```
TheGag96, line 1518: Am I dumb, or did you mean,

```d
alias Big = Repeat!(1_000_000, int);
```
?

andralex: Ooops. I am dumb! Will fix tomorrow and measure again.

PR #8046, opened 2021-05-05
Make staticIsSorted use static foreach instead of recursion

andralex:
No difference measured in unittest compilation time/memory consumed for unittests.


thewilsonator (approved):

RazvanN7:
@andralex I think a rebase will fix the failing druntime tester. Please do it, so we can merge this.

RazvanN7 (approved):

PR #8048, opened 2021-05-05
Add missing SharedConstInoutOf and refactor QualifierOf

andralex:
Not sure how useful those XxxOf are, but added this one for completeness.

Also refactored QualifierOf into something that probably the compiler would have an easier time going through.

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -291,10 +291,7 @@ private
  * Returns:
  *     `T` with the `inout` qualifier added.
  */
-template InoutOf(T)
-{
-    alias InoutOf = inout(T);
-}
+alias InoutOf(T) = inout(T);
 
 ///
 @safe unittest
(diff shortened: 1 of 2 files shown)


thewilsonator:
```
generated/linux/debug/64/publictests/std_traits.d(488): Error: undefined identifier `MutableOf`
generated/linux/debug/64/publictests/std_traits.d(488):        while evaluating: `static assert(__traits(isSame, QualifierOf!int, MutableOf))`
```

RazvanN7:
Can't we just deprecate them all together?

RazvanN7:
@andralex Should we really add these for completeness? Wouldn't it be better to deprecate the existing one?

RazvanN7 (approved):
I'll just merge this and deprecate them altogether later if we reach that conclusion.

andralex:
@RazvanN7 Thanks. I think we should have a better procedure for deciding on deprecation, e.g. search github.com and code.dlang.org to assess the level of use of facilities, etc.

PR #8052, opened 2021-05-07
Replace the use of format() during compilation in std.traits with con…

andralex:
…catenations

In many cases the code is shorter and simpler...

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -780,8 +780,6 @@ private template fqnSym(alias T)
 private template fqnType(T,
     bool alreadyConst, bool alreadyImmutable, bool alreadyShared, bool alreadyInout)
 {
-    import std.format : format;
-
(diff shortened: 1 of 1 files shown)


MoonlightSentinel (changes requested):
~~`format` works with callback / output range based implementations of `toString`. String concatentation does not.~~

~~But `text(...)` might work IIRC.~~

maxhaton:
> `format` works with callback / output range based implementations of `toString`. String concatentation does not.
> 
> But `text(...)` might work IIRC.

Could you give an example of how this effects this change specifically?

MoonlightSentinel:
Sorry, missread the first change to affect other types than string

maxhaton (approved):

RazvanN7 (approved):

PR #8057, opened 2021-05-09
Revert "Fix Issue 21456 - std.format does not accept enum member with string base type as template parameter"

berni44:
Reverts dlang/phobos#8029

See discussion there and [discussion in the forum](https://forum.dlang.org/thread/s72dav$2jd5$1@digitalmars.com).


RazvanN7:
cc @andralex

andralex (approved):

PR #8058, opened 2021-05-09
Remove duplicate assumePureFunction from std.regex

MoonlightSentinel:
It's redundant and has the same invalid constraint that was
fixed for the other implementation


std/regex/internal/ir.d
```diff
@@ -567,7 +567,7 @@ private auto defaultFactoryImpl(Char)(const ref Regex!Char re)
 
 // Used to generate a pure wrapper for defaultFactoryImpl. Based on the example in
 // the std.traits.SetFunctionAttributes documentation.
-private auto assumePureFunction(T)(T t)
+auto assumePureFunction(T)(T t)
```
MoonlightSentinel, line 570: This is still internal because of the `package(std.regex):` at the beginning of this file.

RazvanN7 (approved):

PR #8060, opened 2021-05-10
Make unittests in std.file @safe

atilaneves:


std/file.d
```diff
@@ -758,7 +758,9 @@ if ((isInputRange!R && !isInfinite!R && isSomeChar!(ElementEncodingType!R) || is
 
    int[] a = [ 0, 1, 1, 2, 3, 5, 8 ];
    write(deleteme, a); // deleteme is the name of a temporary file
-   assert(cast(int[]) read(deleteme) == a);
+   const bytes = read(deleteme);
+   const fileInts = () @trusted { return cast(int[]) bytes; }();
```
berni44, line 762: I wonder, what is the advantage of an `@safe` unittest with an `@trusted` block over an `@system` unittest?

RazvanN7: I guess it is going to help transition to @safe by default.

RazvanN7 (approved):

std/file.d
berni44: Hey, thanks. See #8061 how I used this new insight. ;-)

PR #8068, opened 2021-05-12
Rename the initial value of some enums from init to _init

BorisCarvajal:
Avoid conflicts with builtin `.init`.

Related [DMD pull](https://github.com/dlang/dmd/pull/12512).

### DIFF
--- a/std/socket.d
+++ b/std/socket.d
@@ -86,7 +86,7 @@ else version (Posix)
 
     import core.stdc.errno;
 
-    enum socket_t : int32_t { init = -1 }
+    enum socket_t : int32_t { _init = -1 }
     private const int _SOCKET_ERROR = -1;
 
     private enum : int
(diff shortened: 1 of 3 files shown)


std/socket.d
```diff
@@ -86,7 +86,7 @@ else version (Posix)
 
     import core.stdc.errno;
 
-    enum socket_t : int32_t { init = -1 }
```
thewilsonator, line 89: It would seem this is a change of behaviour and is intended to be so:
```
import std.socket;
import std.stdio;
void main()
{
    socket_t sock;
    writefln("%d", sock); // prints -1
}
```

BorisCarvajal: `socket_t.init` will be equal to the first member value `socket_t._init`(-1), just as before.

thewilsonator: Ah, ok.

thewilsonator (approved):

PR #8070, opened 2021-05-12
Fix Issue 21916 - Error message is obfuscated when using wrong format…

CyberShadow:
… specifier at compile-time

Before:

```
.../std/format/package.d(748): Error: expression `FormatException(['I', 'n', 'c', 'o', 'r', 'r', 'e', 'c', 't', ' ', 'f', 'o', 'r', 'm', 'a', 't', ' ', 's', 'p', 'e', 'c', 'i', 'f', 'i', 'e', 'r', ' ', 'f', 'o', 'r', ' ', 'r', 'a', 'n', 'g', 'e', ':', ' ', '%', 'f'], ".../std/format/internal/write.d", 1722LU, null, null, 0u)` is not constant
.../std/format/package.d(1370):        while evaluating: `static assert(!e)`
test.d(5): Error: template instance `std.format.format!("%f", int[])` error instantiating
```

After:

```
.../std/format/package.d(1370): Error: static assert:  "Incorrect format specifier for range: %f"
test.d(5):        instantiated from here: `format!("%f", int[])`
```

No test because (AFAIK) we can't test static assert error messages in Phobos.

### DIFF
--- a/std/format/package.d
+++ b/std/format/package.d
@@ -732,7 +732,7 @@ if (isInputRange!Range)
 }
 
 // Used to check format strings are compatible with argument types
-package(std) static const checkFormatException(alias fmt, Args...) =
+package(std) enum checkFormatException(alias fmt, Args...) =
 {
     import std.conv : text;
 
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

berni44:
This also fixed [issue 21774](https://issues.dlang.org/show_bug.cgi?id=21774). :-)

PR #8076, opened 2021-05-14
Annotate with `return` or `return scope` to please dlang/dmd#12520

nordlow:
Unblocks dlang/dmd#12520.

`make -f posix.mak unittest` passes locally for me now when using dlang/dmd#12520.

There are a couple of places where I had to qualify as `@trusted` to silence. I've commented those places with a TODO.

Furthermore, there are several places where I have no idea how to explain to the compiler that the code should compile when lambdas inside `map` and `file`. For instance,

```d
@safe unittest
{
    string[] listdir(string pathname)
    {
        import std.algorithm;
        import std.array;
        import std.file;
        import std.path;

        return std.file.dirEntries(pathname, SpanMode.shallow)
            .filter!(a => a.isFile)
            .map!(a => std.path.baseName(a.name))
            .array;
    }

    void main(string[] args)
    {
        import std.stdio;

        string[] files = listdir(args[1]);
        writefln("%s", files);
     }
}
```

erroring as

```
std/file.d(4987): Error: returning `baseName(a.name())` escapes a reference to parameter `a`
std/file.d(4987):        perhaps annotate the parameter with `return`
std/algorithm/iteration.d(524):        instantiated from here: `MapResult!(__lambda3, FilterResult!(__lambda2, DirIterator))`
std/file.d(4987):        instantiated from here: `map!(FilterResult!(__lambda2, DirIterator))`
```

Not even qualifying `listdir` as `@trusted` silences the diagnostics. Shouldn't it?

### DIFF
--- a/std/bigint.d
+++ b/std/bigint.d
@@ -246,7 +246,7 @@ public:
      * Implements assignment operators from built-in integers of the form
      * `BigInt op= integer`.
      */
-    BigInt opOpAssign(string op, T)(T y) pure nothrow @safe
+    BigInt opOpAssign(string op, T)(T y) pure nothrow @safe return
         if ((op=="+" || op=="-" || op=="*" || op=="/" || op=="%"
           || op==">>" || op=="<<" || op=="^^" || op=="|" || op=="&" || op=="^") && isIntegral!T)
     {
@@ -414,7 +414,7 @@ public:
     /**
      * Implements assignment operators of the form `BigInt op= BigInt`.
      */
-    BigInt opOpAssign(string op, T)(T y) pure nothrow @safe
+    BigInt opOpAssign(string op, T)(T y) pure nothrow @safe scope return
         if ((op=="+" || op== "-" || op=="*" || op=="|" || op=="&" || op=="^" || op=="/" || op=="%")
             && is (T: BigInt))
     {
@@ -472,13 +472,14 @@ public:
     /**
      * Implements binary operators between `BigInt`s.
      */
-    BigInt opBinary(string op, T)(T y) pure nothrow @safe const
+    BigInt opBinary(string op, T)(T y) pure nothrow @safe const return scope
         if ((op=="+" || op == "*" || op=="-" || op=="|" || op=="&" || op=="^" ||
             op=="/" || op=="%")
             && is (T: BigInt))
     {
         BigInt r = this;
-        return r.opOpAssign!(op)(y);
+        r.opOpAssign!(op)(y);
+        return r;
     }
 
     ///
@@ -1454,19 +1455,19 @@ public:
     }
 
 private:
-    void negate() @safe pure nothrow @nogc
+    void negate() @safe pure nothrow @nogc scope
     {
         if (!data.isZero())
             sign = !sign;
     }
-    bool isZero() pure const nothrow @nogc @safe
+    bool isZero() pure const nothrow @nogc @safe scope
     {
         return data.isZero();
     }
     alias isNegative = sign;
 
     // Generate a runtime error if division by zero occurs
-    void checkDivByZero() pure const nothrow @safe
+    void checkDivByZero() pure const nothrow @safe scope
     {
         assert(!isZero(), "BigInt division by zero");
     }
--- a/std/internal/math/biguintcore.d
+++ b/std/internal/math/biguintcore.d
@@ -245,12 +245,12 @@ private:
 
     immutable(BigDigit) [] data = ZERO;
 
-    this(immutable(BigDigit) [] x) pure nothrow @nogc @safe
+    this(immutable(BigDigit) [] x) pure nothrow @nogc @trusted scope // TODO @safe when -dip1000 understand scope of immutable arrays
     {
        data = x;
     }
   package(std)  // used from: std.bigint
-    this(T)(T x) pure nothrow @safe if (isIntegral!T)
+    this(T)(T x) pure nothrow @safe scope if (isIntegral!T)
     {
         opAssign(x);
     }
@@ -260,7 +260,7 @@ private:
     };
 public:
     // Length in uints
-    @property size_t uintLength() pure nothrow const @safe @nogc
+    @property size_t uintLength() pure nothrow const @safe @nogc scope
     {
         static if (BigDigit.sizeof == uint.sizeof)
         {
@@ -272,7 +272,7 @@ public:
             ((data[$-1] & 0xFFFF_FFFF_0000_0000L) ? 1 : 0);
         }
     }
-    @property size_t ulongLength() pure nothrow const @safe @nogc
+    @property size_t ulongLength() pure nothrow const @safe @nogc scope
     {
         static if (BigDigit.sizeof == uint.sizeof)
         {
@@ -285,7 +285,7 @@ public:
     }
 
     // The value at (cast(ulong[]) data)[n]
-    ulong peekUlong(size_t n) pure nothrow const @safe @nogc
+    ulong peekUlong(size_t n) pure nothrow const @safe @nogc scope
     {
         static if (BigDigit.sizeof == int.sizeof)
         {
@@ -298,7 +298,7 @@ public:
         }
     }
 
-    uint peekUint(size_t n) pure nothrow const @safe @nogc
+    uint peekUint(size_t n) pure nothrow const @safe @nogc scope
     {
         static if (BigDigit.sizeof == int.sizeof)
         {
@@ -312,7 +312,7 @@ public:
     }
 
     ///
-    void opAssign(Tulong)(Tulong u) pure nothrow @safe if (is (Tulong == ulong))
+    void opAssign(Tulong)(Tulong u) pure nothrow @safe scope if (is (Tulong == ulong))
     {
         if (u == 0) data = ZERO;
         else if (u == 1) data = ONE;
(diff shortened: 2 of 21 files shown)


nordlow:
Added debug prints in bigint tests. Now

```
make -f posix.mak BUILD=debug std/bigint.test
```

prints

```
true, x:1000000500, l:1000000, x%l:500
true, x:1000000500, ul:1000000, x%ul:0
core.exception.AssertError@std/bigint.d(588): unittest failure
```

Something is very wrong here.

std/bigint.d
```diff
@@ -565,12 +566,27 @@ public:
...
+        } ();
         assert(is(typeof(x % l)  == long)   && x % l  == 500L);
+
+        () @trusted {
+            writeln(is(typeof(x % ul) == BigInt),
+                    ", x:", x,
+                    ", ul:", ul,
+                    ", x%ul:", x % ul);
+        } ();
         assert(is(typeof(x % ul) == BigInt) && x % ul == BigInt(500));
```
nordlow, line 588: This line fails via

```
make -f posix.mak BUILD=debug std/bigint.test
```

as

```
true, x:1000000500, l:1000000, x%l:500
true, x:1000000500, ul:1000000, x%ul:0
core.exception.AssertError@std/bigint.d(588): unittest failure
```

Something is very wrong here.

dkorpel: You made a `BigUint` constructor `this(scope immutable(BigDigit) [] x) pure nothrow @nogc @trusted` but it assigns `x` to `this.data`, therefore escaping the constructor's scope. I suspect this introduces memory corruption.

nordlow: Ahh. I mistook x for having infinite lifetime. My mistake. I'll fix. Thanks!

nordlow: Removing `scope` for `x` makes

```d
    BigUint opBinary(string op, Tulong)(Tulong y) pure nothrow @safe const scope
        if (op == ">>" && is (Tulong == ulong))
    {
        assert(y > 0, "Can not right shift BigUint by 0");
        uint bits = cast(uint) y & BIGDIGITSHIFTMASK;
        if ((y >> LG2BIGDIGITBITS) >= data.length) return BigUint(ZERO);
        uint words = cast(uint)(y >> LG2BIGDIGITBITS);
        if (bits == 0)
        {
            return BigUint(data[words..$]);
        }
        else
        {
            uint [] result = new BigDigit[data.length - words]; // TODO this fails
            multibyteShr(result, data[words..$], bits);

            if (result.length > 1 && result[$-1] == 0)
                return BigUint(trustedAssumeUnique(result[0 .. $-1]));
            else
                return BigUint(trustedAssumeUnique(result));
        }
    }
```

fails as

```
std/internal/math/biguintcore.d(722): Error: scope variable `this` assigned to non-scope parameter `x` calling std.internal.math.biguintcore.BigUint.this
std/bigint.d(314): Error: template instance `std.internal.math.biguintcore.BigUint.opBinary!(">>", ulong)` error instantiating
std/bigint.d(2356):        instantiated from here: `opOpAssign!(">>", int)`
make: *** [posix.mak:342: generated/linux/debug/64/libphobos2.a] Fel 1
```

. Is there another way of fixing this than replacing `data[words..$]` with `data[words..$].idup`?

I don't understand why `data[words..$]` can't be passed to the `BigUint.this(immutable(BigDigit) [] x)`.

?

std/internal/math/biguintcore.d
```diff
@@ -719,7 +719,7 @@ public:
         uint words = cast(uint)(y >> LG2BIGDIGITBITS);
         if (bits == 0)
         {
-            return BigUint(data[words..$]);
+            return BigUint(data[words..$].idup); // TODO can .idup be avoided?
```
nordlow, line 722: Is there a way to avoid this error message with the `.idup` here? I don't understand why `data[words..$]` can't be passed to `BigUint.ctor()`as data is an array of immutable values. Is it because this in this `opBinary` is qualified as scope?

nordlow: Making opBinary non-`scope` did it.

dkorpel: `immutable` is irrelevant, you annotated the function `scope` so `this.data` is `scope` and can't be passed to a non-scope constructor.

std/bigint.d
nordlow: This was fixed when I removed `scope` from `opBinary`.

nordlow:
`make -f posix.mak unittest` passes locally for me now.

dkorpel:
A general comment about BigInt:

A lot of the functions look like this (pseudo code):
```D
BigInt opBinary(BigInt other) {
    if (we're lucky) {
        return [bigint with slice of the existing big digits]
    } else {
        return [newly allocated big digits with computation result]
    }
}
```
There's two ways to approach this:
- annotate the function `return scope` so returning a `BigInt` with a slice of `this.data` is allowed, but putting an extra burden on the caller (a breaking change)
- annotate the function with `scope` and defensively `.dup` `this.data` in the lucky cases, as the common case has to re-allocate anyway.

I'd say: use `return scope` when the function is already `@nogc` (e.g. `opUnary!"-"`) do `scope` + `.dup` otherwise.

dkorpel:
A general comment about this PR:

It gives a neat overview, but please split it up into smaller PR's (one per module / part) so it can be reviewed more easily and regressions can be bisected with more granularity. I happen to have familiarity with bigint, but can't help you with most other modules, and adding `return` and especially `@trusted` without a good understanding of each module can be risky.

nordlow:
> It gives a neat overview, but please split it up at some point into smaller PR's (one per module / part) so it can be reviewed more easily and regressions can be bisected with more granularity.

All sub-pull-requests have now been created, @dkorpel

dkorpel:
> All sub-pull-requests have now been created, @dkorpel

Thank you!

dkorpel:
By the way, don't you get deprecations from druntime with your pull request?
When I started looking at whether phobos compiles with the bug fix, one of the first culprits I found was that BigInt.toString is not `@safe` anymore because it calls functions from std.format which rely on `.dup` in object.d to remove `scope` from parameters, but `.dup` fails to have `scope` inferred currently and relies on `pure` to compile anyway. Have you encountered anything similar?

nordlow:
> By the way, don't you get deprecations from druntime with your pull request?
> When I started looking at whether phobos compiles with the bug fix, one of the first culprits I found was that BigInt.toString is not `@safe` anymore because it calls functions from std.format which rely on `.dup` in object.d to remove `scope` from parameters, but `.dup` fails to have `scope` inferred currently and relies on `pure` to compile anyway. Have you encountered anything similar?

On `stable` I do not. 

Btw, shouldn't my recent fix to druntime be applied to master aswell?

dkorpel:
> On `stable` I do not. The only deprecations I get when compiling as
> 
> `make -f posix.mak unittest`
> 
> are
 
Same for me when checking out your PR, but when checking out [aG0aep6G's PR](https://github.com/dlang/dmd/pull/12010) I have the problem that `core.lifetime: move` is no longer `@safe`.

> Btw, shouldn't my recent fix to druntime be applied to master aswell?

Yeah it's kind of annoying when changes in master depend on changes in stable, not sure what to do.

nordlow:
> Yeah it's kind of annoying when changes in master depend on changes in stable, not sure what to do.

How to deal with this, @RazvanN7 @thewilsonator ?

CyberShadow:
AFAIK stable can be merged into master at any time, and anyone can propose a PR to do it.

nordlow:
> AFAIK stable can be merged into master at any time, and anyone can propose a PR to do it.

Is this the correct way of doing it? https://github.com/dlang/druntime/pull/3475

dkorpel:
> On `stable` I do not. The only deprecations I get when compiling as

I suspect now (haven't checked) the deprecations only get raised when you directly pass a `scope` variable to a non-scope parameter of a pure function, but it can still wrongly infer `scope` on a template function without warning.

If that's true, then there might need to be a second round of fixes to Phobos, for `std.format` and `BigInt.toString` and more :(

nordlow:
> > On `stable` I do not. The only deprecations I get when compiling as
> 
> I suspect now (haven't checked) the deprecations only get raised when you directly pass a `scope` variable to a non-scope parameter of a pure function, but it can still wrongly infer `scope` on a template function without warning.
> 
> If that's true, then there might need to be a second round of fixes to Phobos, for `std.format` and `BigInt.toString` and more :(

I don't follow. When should these problems I arise? I've already verified that the initial phobos branch with all the adjustments builds and runs all the unittests with the dmd branch.

dkorpel:
> I don't follow. When should these problems arise? I've already verified locally that #8076 builds and runs all the unittests without errors with [dlang/dmd#12520](https://github.com/dlang/dmd/pull/12520).

But does it work when you apply https://github.com/dlang/dmd/pull/10924 or https://github.com/dlang/dmd/pull/12010 to make it an error?
I'm afraid that your deprecations variant https://github.com/dlang/dmd/pull/12520 is not thorough enough, since e.g. you didn't stumble on https://github.com/dlang/druntime/pull/3472 like I did when applying  aG0aep6G's fix.

nordlow:
> But does it work when you apply [dlang/dmd#10924](https://github.com/dlang/dmd/pull/10924) or [dlang/dmd#12010](https://github.com/dlang/dmd/pull/12010) to make it an error?

My initial phobos branch buillt without any deprecations when using https://github.com/dlang/dmd/pull/10924/files, yes. As will all the sub-PRs.

WalterBright:
@nordlow I pulled out a couple of files from this PR into separate PRs to make it smaller and more tractable.

nordlow:
> @nordlow I pulled out a couple of files from this PR into separate PRs to make it smaller and more tractable.

Ok, thanks. Note that I already had pulled out the bigint part from this PR at https://github.com/dlang/phobos/pull/8081, though.

RazvanN7:
Does this depend solely on #8085 now?

nordlow:
Closing this because of sub-PR has been created and all but one has been merged

PR #8077, opened 2021-05-14
Fix issue 21926 - Allow leadings zeros in std.conv.octal(string)

baryluk:
First of all, the DDoc currently says that:

"Leading zero is allowed, but not required."

The current implementation doesn't respect that.

D lexer allows leading zero(s) also.

And when considering octal from string, there is no possibility
of confusion, as the comments in the code claim. Disallowing
leading zeros for octal conversion from string, does not help
with anything actually.

So lets allow leading zeros.

Allowing leading zeros, does help when implementing various
APIs (i.e. glibc, Linux kernel), where a lot of octal constant,
are defined with multiple leading zeros, for readability and alignment,
in C headers. Having to manually or automatically special case
these when porting such API definitions, is counter-productive.

Example from a Linux kernel (`include/uapi/linux/stat.h`):

```c
#define S_IFMT  00170000
#define S_IFSOCK 0140000
#define S_IFLNK	 0120000
#define S_IFREG  0100000
#define S_IFBLK  0060000
#define S_IFDIR  0040000
#define S_IFCHR  0020000
#define S_IFIFO  0010000
#define S_ISUID  0004000
#define S_ISGID  0002000
#define S_ISVTX  0001000
```

With this patch, now it is trivial and easier to convert these to D:

```d
...
enum S_ISVTX = octal!"0001000";
```

while being close to original. That helps with readability,
and long term maintenance.

In fact the run-time version provided by `parse!(int)(string, 8)`
also supports leading zeros already. So this makes `octal`
more consistent.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -2884,6 +2884,7 @@ do
 
     assert(parse!int(s = "0011001101101", 2) == 0b0011001101101);
     assert(parse!int(s = "765", 8) == octal!765);
+    assert(parse!int(s = "000135", 8) == octal!"135");
     assert(parse!int(s = "fCDe", 16) == 0xfcde);
 
     // https://issues.dlang.org/show_bug.cgi?id=6609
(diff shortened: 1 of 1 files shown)


baryluk:
Fixed style errors

thewilsonator (approved):
Looks good. I'm not quite sure why the tests are segfaulting on azure. 
Also is there a bug report for this? If so, please retitle the commit _message_ "Fix issue #####: Allow leading zeros in std.conv.octal", if not please file.

baryluk:
> Looks good. I'm not quite sure why the tests are segfaulting on azure.
> Also is there a bug report for this? If so, please retitle the commit _message_ "Fix issue #####: Allow leading zeros in std.conv.octal", if not please file.

Add the issue and renamed the commit.

The azur failures are gone, it looks to have been some other unrelated issue. The buildkite issue now also looks unrelated.

RazvanN7:
I'm quite sure a rebase will fix the failing builkite test.

RazvanN7 (approved):

baryluk:
"Rebased". All passing now.

PR #8081, opened 2021-05-16
Annotate std/bigint.d and std/internal/math/biguintcore.d to please d…

nordlow:
…lang/dmd#12520

Part of #8076

### DIFF
--- a/std/bigint.d
+++ b/std/bigint.d
@@ -246,7 +246,7 @@ public:
      * Implements assignment operators from built-in integers of the form
      * `BigInt op= integer`.
      */
-    BigInt opOpAssign(string op, T)(T y) pure nothrow @safe
+    BigInt opOpAssign(string op, T)(T y) pure nothrow @safe return
         if ((op=="+" || op=="-" || op=="*" || op=="/" || op=="%"
           || op==">>" || op=="<<" || op=="^^" || op=="|" || op=="&" || op=="^") && isIntegral!T)
     {
@@ -414,7 +414,7 @@ public:
     /**
      * Implements assignment operators of the form `BigInt op= BigInt`.
      */
-    BigInt opOpAssign(string op, T)(T y) pure nothrow @safe
+    BigInt opOpAssign(string op, T)(T y) pure nothrow @safe scope return
         if ((op=="+" || op== "-" || op=="*" || op=="|" || op=="&" || op=="^" || op=="/" || op=="%")
             && is (T: BigInt))
     {
@@ -472,7 +472,7 @@ public:
     /**
      * Implements binary operators between `BigInt`s.
      */
-    BigInt opBinary(string op, T)(T y) pure nothrow @safe const
+    BigInt opBinary(string op, T)(T y) pure nothrow @safe const return scope
         if ((op=="+" || op == "*" || op=="-" || op=="|" || op=="&" || op=="^" ||
             op=="/" || op=="%")
             && is (T: BigInt))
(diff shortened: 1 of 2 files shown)


std/bigint.d
```diff
@@ -472,13 +472,14 @@ public:
...
      */
-    BigInt opBinary(string op, T)(T y) pure nothrow @safe const
+    BigInt opBinary(string op, T)(T y) pure nothrow @safe const return scope
         if ((op=="+" || op == "*" || op=="-" || op=="|" || op=="&" || op=="^" ||
             op=="/" || op=="%")
             && is (T: BigInt))
     {
         BigInt r = this;
-        return r.opOpAssign!(op)(y);
+        r.opOpAssign!(op)(y);
```
nordlow, line 481: I'm gonna see if I can get make this build with this change.

nordlow:
What is the correct parameter qualification of 

```d
    static void divMod(scope return BigUint x, scope return BigUint y,
                       scope out BigUint quotient, scope out BigUint remainder) pure nothrow @safe
    {
        if (y.data.length > x.data.length)
        {
            quotient = 0uL;
            remainder = x;
        }
        else if (y.data.length == 1)
        {
            quotient = divInt(x, y.data[0]);
            remainder = BigUint([modInt(x, y.data[0])]);
        }
        else
        {
            BigDigit[] quot = new BigDigit[x.data.length - y.data.length + 1];
            BigDigit[] rem = new BigDigit[y.data.length];
            divModInternal(quot, rem, x.data, y.data);
            quotient = BigUint(removeLeadingZeros(trustedAssumeUnique(quot)));
            remainder = BigUint(removeLeadingZeros(trustedAssumeUnique(rem)));
        }
    }
```

, @Geod24 ? I can't find any references on assigning return scope parameter to out scope parameters in the DIP-25 and DIP-1000 specs.

thewilsonator:
As far as I understand:
    `return` qualified parameters on a `void` returning function are nonsensical. 
    `out` parameters don't make sense as `scope` in a pure function since they can only be derived from either a pure allocator (e.g. the GC) or from other input parameters. scope on function parameters means that that reference shall not escape the scope of the function.

nordlow:
> As far as I understand:
> `return` qualified parameters on a `void` returning function are nonsensical.
> `out` parameters don't make sense as `scope` in a pure function since they can only be derived from either a pure allocator (e.g. the GC) or from other input parameters. scope on function parameters means that that reference shall not escape the scope of the function.

Ok, thanks. I'll adjust.

dkorpel:
Why does `divMod` need `scope` parameters, is there any existing code calling it with `scope` BigInts?
If you have to make it work, you can do:
```D
// biguintcore.d(977):
remainder = BigUint(x.data.idup);
// biguintcore.d(981):
quotient = divInt(BigUint(x.data.idup), y.data[0]);
```
If you do that, maybe add a comment that the redundant allocation is due to the limitation that `return scope` for out parameters is not supported.

nordlow:
> Why does `divMod` need `scope` parameters, is there any existing code calling it with `scope` BigInts?
> If you have to make it work, you can do:
> 
> ```d
> // biguintcore.d(977):
> remainder = BigUint(x.data.idup);
> // biguintcore.d(981):
> quotient = divInt(BigUint(x.data.idup), y.data[0]);
> ```
> 
> If you do that, maybe add a comment that the redundant allocation is due to the limitation that `return scope` for out parameters is not supported.

This seems like a severe limitation. Has `return scope` for out parameters been considered, @WalterBright?

nordlow:
I've reverted to a the @safe defensive variant 

```
static void divMod(BigUint x, scope BigUint y,
                       out BigUint quotient, out BigUint remainder) pure nothrow @safe
```

that passes locally with dmd on master my machine.

nordlow:
> Why does `divMod` need `scope` parameters, is there any existing code calling it with `scope` BigInts?
> If you have to make it work, you can do:

Thanks for highlighting this. I've reverted to non-x-scope variant 

```
static void divMod(BigUint x, scope BigUint y,
                       out BigUint quotient, out BigUint remainder) pure nothrow @safe
```

that doesn't require any @trusted and that passes locally with dmd master my machine.

dkorpel (approved):

WalterBright (approved):

PR #8087, opened 2021-05-16
Annotate sha_SSSE3.d to please dlang/dmd#12520

nordlow:
Part of #8076

### DIFF
--- a/std/internal/digest/sha_SSSE3.d
+++ b/std/internal/digest/sha_SSSE3.d
@@ -194,13 +194,16 @@ version (USE_SSSE3)
     /**
      * Chooses the instruction sequence based on the 32bit or 64bit model.
      */
-    private nothrow pure string[] swt3264(string[] insn32, string[] insn64)
+    version (_32Bit)
     {
-        version (_32Bit)
+        private nothrow pure string[] swt3264(return scope string[] insn32, scope string[])
         {
             return insn32;
         }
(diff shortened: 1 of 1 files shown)


nordlow:
It's very unfortunate that our CI pipeline depends on diagnostic tools that keep breaking in expected ways. This makes it very difficult to be productive. For instance:

```
--

std/internal/digest/sha_SSSE3.d(206:62)[warn]: Parameter insn32 is never used.
```

at

https://buildkite.com/dlang/phobos/builds/5474#b4491b19-2dc5-4462-a564-de190f4535c0

The only reasonable way forward is to only depend on diagnostics available in dmd. If checking for unused variables is really needed implement it in the compiler.

MoonlightSentinel:
Well, the warning is correct and would've been raised by dmd as well. Remove the identifier of the unused parameter

nordlow:
> Well, the warning is correct and would've been raised by dmd as well. Remove the identifier of the unused parameter.
> 
> EDIT: W.r.t. the unpredictable, DScanner was able to figure out that both parameters can be used depending on the `version` - dmd most likely won't be able to do that given how it treats `version` conditions

Thanks, that's good. I see now that I can run `dscanner` locally.

Is `make style` the preferred entrypoint?

dkorpel (approved):

thewilsonator (approved):

PR #8088, opened 2021-05-16
Annotate std/json.d to please dlang/dmd#12520

nordlow:
Part of #8076

### DIFF
--- a/std/json.d
+++ b/std/json.d
@@ -152,14 +152,14 @@ struct JSONValue
      * Throws: `JSONException` for read access if `type` is not
      * `JSONType.string`.
      */
-    @property string str() const pure @trusted
+    @property string str() const pure @trusted return scope
     {
         enforce!JSONException(type == JSONType.string,
                                 "JSONValue is not a string");
         return store.str;
     }
     /// ditto
-    @property string str(string v) pure nothrow @nogc @safe
+    @property string str(return string v) pure nothrow @nogc @trusted return // TODO make @safe
     {
         assign(v);
         return v;
@@ -275,14 +275,14 @@ struct JSONValue
        (*a)["hello"] = "world";  // segmentation fault
        ---
      */
-    @property ref inout(JSONValue[string]) object() inout pure @system
+    @property ref inout(JSONValue[string]) object() inout pure @system return
     {
         enforce!JSONException(type == JSONType.object,
                                 "JSONValue is not an object");
         return store.object;
     }
     /// ditto
-    @property JSONValue[string] object(JSONValue[string] v) pure nothrow @nogc @safe
+    @property JSONValue[string] object(return JSONValue[string] v) pure nothrow @nogc @trusted // TODO make @safe
     {
         assign(v);
         return v;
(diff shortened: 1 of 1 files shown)


std/json.d
```diff
@@ -461,7 +461,7 @@ struct JSONValue
         assertNotThrown(json["h"].get!float);
     }
 
-    private void assign(T)(T arg)
+    private void assign(T)(scope T arg)
```
Geod24, line 464: How can it be `scope` given it is stored in the object ?

nordlow: Thanks. I'll check.

dkorpel: It can't, but the assignment happens in a `@trusted` lambda so the compiler doesn't complain.

nordlow: Without scope, compilation errors as

```
std/json.d(164): Error: scope variable `v` assigned to non-scope parameter `arg` calling std.json.JSONValue.assign!string.assign
std/json.d(287): Error: scope variable `v` assigned to non-scope parameter `arg` calling std.json.JSONValue.assign!(JSONValue[string]).assign
std/json.d(333): Error: scope variable `v` assigned to non-scope parameter `arg` calling std.json.JSONValue.assign!(JSONValue[]).assign
std/json.d(1210): Error: CTFE failed because of previous errors in `object`
std/json.d(1206):        called from here: `parseValue(member)`
std/json.d(1379):        called from here: `parseValue(root)`
std/json.d(1391):        called from here: `parseJSON("{ \"key1\": { \"key2\": 1 }}", -1, JSONOptions.none)`
std/json.d(1391):        while evaluating: `static assert(cast(int)parseJSON("{ \"key1\": { \"key2\": 1 }}", -1, JSONOptions.none).type() == 6)`
std/json.d(1237): Error: CTFE failed because of previous errors in `array`
std/json.d(1231):        called from here: `parseValue(element)`
std/json.d(1379):        called from here: `parseValue(root)`
std/json.d(1394):        called from here: `parseJSON("[[1]]", -1, JSONOptions.none)`
std/json.d(1394):        while evaluating: `static assert(cast(int)parseJSON("[[1]]", -1, JSONOptions.none).type() == 5)`
std/json.d(1210): Error: CTFE failed because of previous errors in `object`
std/json.d(1379):        called from here: `parseValue(root)`
std/json.d(1427):        called from here: `parseJSON("{\"a\" : 2}", -1, JSONOptions.none)`
std/json.d(1427):        while evaluating: `static assert(((JSONValue __dop1293 = parseJSON("{\"a\" : 2}", -1, JSONOptions.none);) , __dop1293.opIndex("a")).integer() == 2L)`
```

I have no idea what's going on here. Something with CTFE.

nordlow: What if I qualify `assign()` itself as scope?

nordlow: I believe we have a catch 22 here.

Qualifying v as return 

    @property string str(return string v) pure nothrow @nogc @safe
    {
        assign(v);
        return v;
    }

fails to be build with -dip1000 as

```
std/json.d(164): Error: scope variable `v` assigned to non-scope parameter `arg` calling std.json.JSONValue.assign!string.assign
```

because incorrect inference of `scope` on v because str is qualified as pure.

This won't build correctly until the dmd is merged. What to do, @Geod24, @thewilsonator ?

nordlow: I just pushed a forced update that uses @trusted along with a TODO for now. std/json.d builds locally with both master and  dlang/dmd#12520.

RazvanN7 (approved):
I agree that this is a better alternative, however, we must not forget about it.

nordlow:
> I agree that this is a better alternative, however, we must not forget about it.

Thank you!

PR #8101, opened 2021-05-17
Make it possible to use RefCounted from @safe code

atilaneves:
Reviews definitely required here. 

I thought it through and I *think* this is memory safe. I tried to do nasty things with `-preview=dip1000` and the compiler prevented me, so that's nice.

This is currently preventing `std.file.dirEntries` from being `@safe`.

### DIFF
--- a/std/typecons.d
+++ b/std/typecons.d
@@ -6359,8 +6359,9 @@ still be valid during the destructor call.  This allows the `T` to
 deallocate or clean up any non-GC resources immediately after the
 reference count has reached zero.
 
-`RefCounted` is unsafe and should be used with care. No references
-to the payload should be escaped outside the `RefCounted` object.
+If not using -preview=dip1000, `RefCounted` is unsafe and should be
+used with care. No references to the payload should be escaped outside
+the `RefCounted` object.
 
 The `autoInit` option makes the object ensure the store is
 automatically initialized. Leaving $(D autoInit ==
@@ -6414,7 +6415,7 @@ if (!is(T == class) && !(is(T == interface)))
             import core.lifetime : emplace, forward;
 
             allocateStore();
-            version (D_Exceptions) scope(failure) deallocateStore();
+            version (D_Exceptions) scope(failure) () @trusted { deallocateStore(); }();
             emplace(&_store._payload, forward!args);
             _store._count = 1;
         }
@@ -6436,7 +6437,7 @@ if (!is(T == class) && !(is(T == interface)))
                 import std.internal.memory : enforceCalloc;
                 auto ptr = enforceCalloc(1, Impl.sizeof);
                 _store = () @trusted { return cast(Impl*) ptr; }();
-                pureGcAddRange(&_store._payload, T.sizeof);
+                () @trusted { pureGcAddRange(&_store._payload, T.sizeof); }();
             }
             else
             {
@@ -6452,7 +6453,7 @@ if (!is(T == class) && !(is(T == interface)))
             {
                 pureGcRemoveRange(&this._store._payload);
             }
-            () @trusted { pureFree(_store); }();
+            pureFree(_store);
             _store = null;
         }
 
@@ -6550,7 +6551,7 @@ to deallocate the corresponding resource.
             return;
         // Done, destroy and deallocate
         .destroy(_refCounted._store._payload);
-        _refCounted.deallocateStore();
+        () @trusted { _refCounted.deallocateStore(); }();
     }
 
 /**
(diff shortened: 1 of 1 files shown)


std/typecons.d
```diff
@@ -6424,7 +6424,7 @@ if (!is(T == class) && !(is(T == interface)))
             import std.algorithm.mutation : moveEmplace;
 
             allocateStore();
-            moveEmplace(source, _store._payload);
+            () @trusted { moveEmplace(source, _store._payload); }();
```
thewilsonator, line 6427: what about `inisialize` just above this? Is that covered by tests?

std/typecons.d
```diff
@@ -6450,7 +6452,7 @@ if (!is(T == class) && !(is(T == interface)))
             {
                 pureGcRemoveRange(&this._store._payload);
             }
-            pureFree(_store);
+            () @trusted { pureFree(_store); }();
```
thewilsonator, line 6455: I'd feel better if `deallocateStore` was in a trusted lambda where it is called rather than have an `@trusted` function that deallocates things

atilaneves: I started off that way, then thought "when is this not memory-safe?". The reason for `free` itself to be `@system` is twofold: one can pass a pointer that wasn't allocated with malloc to it (such as to the stack), and the other is use-after-free problems if there are aliases to that memory elsewhere. 

I don't think either can happen here.

atilaneves: Now that I think of it, it definitely can because there's no check here about the ref count.

thewilsonator: (unresolving conversation)

std/typecons.d
atilaneves: I added a test to ensure `initialize` is `@safe`.

Geod24:
```D
@safe:

import std.stdio;
import std.typecons;

struct Container
{
    int[] data;
}

void main ()
{
    auto ptr = getPtr();
    writeln(ptr);
    const save = ptr.dup;
    auto other = getPtr();
    assert(save == ptr);
}

int[] getPtr ()
{
    int[42] local;
    return getPtr2(Container(local));
}

int[] getPtr2 (scope Container local)
{
    RefCounted!Container rc = local;
    return rc.refCountedPayload().data;
}
```
Compiles, print garbage, fails the assert.

Compiled with `./generated/osx/release/64/dmd -preview=dip1000 -run atila.d` and this PR.

std/typecons.d
atilaneves: I resolved it because I addressed the comment in the latest version.

atilaneves:
> ```d
> @safe:
> 
> import std.stdio;
> import std.typecons;
> 
> struct Container
> {
>     int[] data;
> }
> 
> void main ()
> {
>     auto ptr = getPtr();
>     writeln(ptr);
>     const save = ptr.dup;
>     auto other = getPtr();
>     assert(save == ptr);
> }
> 
> int[] getPtr ()
> {
>     int[42] local;
>     return getPtr2(Container(local));
> }
> 
> int[] getPtr2 (scope Container local)
> {
>     RefCounted!Container rc = local;
>     return rc.refCountedPayload().data;
> }
> ```
> 
> Compiles, print garbage, fails the assert.
> 
> Compiled with `./generated/osx/release/64/dmd -preview=dip1000 -run atila.d` and this PR.

This is what reviews are for. Good work and thanks! Now to try and fix...

Geod24:
> Now to try and fix...

As I mentioned in our previous discussion, given the nature of DIP1000, I have little hope that we can end up with a (truly) `@safe` container that is usable. I would love nothing more than to be proven wrong though.

n8sh (changes requested):
Marking this as "changes requested" as an advisory against merging--even if the author finds a way to make this correct--until either DIP1000 is enabled by default or (better) the PR only applies `@trusted` when DIP1000 is enabled.

std/typecons.d
```diff
@@ -6548,7 +6551,7 @@ to deallocate the corresponding resource.
             return;
         // Done, destroy and deallocate
         .destroy(_refCounted._store._payload);
-        _refCounted.deallocateStore();
+        () @trusted { _refCounted.deallocateStore(); }();
```
dukc, line 6554: I think it's only this line that needs to be `@system` without -dip1000. I'm not sure if there is a `version`/`enum` for `-dip1000` anywhere already, but if not:

```d
import std.traits : isSafe;
enum dip1000Enabled = isSafe!((int x) => *&x);
```

std/typecons.d
```diff
@@ -6414,7 +6415,7 @@ if (!is(T == class) && !(is(T == interface)))
             import core.lifetime : emplace, forward;
 
             allocateStore();
-            version (D_Exceptions) scope(failure) deallocateStore();
+            version (D_Exceptions) scope(failure) () @trusted { deallocateStore(); }();
             emplace(&_store._payload, forward!args);
```
radcapricorn, line 6419: The deallocation can't be trusted if the `emplace` call isn't pure.

ljmf00: I don't understand the correlation, can you elaborate?

radcapricorn: Impure `emplace` (i.e. an impure constructor call within it) may store a pointer somewhere, making use after free possible.

PR #8093, opened 2021-05-17
Annotate std/range/package.d to please dlang/dmd#12520

nordlow:
Part of #8076

PR #8092, opened 2021-05-17
Revert "Fix Issue 9489 - writeln of struct with disabled copy ctor"

ibuclaw:
Reverts dlang/phobos#8055

PR #8103, opened 2021-05-17
Revert "Annotate std/path.d to please dlang/dmd#12520"

ibuclaw:
Reverts dlang/phobos#8090

Something broke master, and this is the only change that touches absolutePath.

### DIFF
--- a/std/path.d
+++ b/std/path.d
@@ -115,13 +115,13 @@ version (StdUnittest)
 private:
     struct TestAliasedString
     {
-        string get() @safe @nogc pure nothrow return scope { return _s; }
+        string get() @safe @nogc pure nothrow { return _s; }
         alias get this;
         @disable this(this);
         string _s;
     }
 
-    bool testAliasedString(alias func, Args...)(scope string s, scope Args args)
+    bool testAliasedString(alias func, Args...)(string s, Args args)
     {
         return func(TestAliasedString(s), args) == func(s, args);
     }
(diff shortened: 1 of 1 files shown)


dkorpel:
What does the test output say? I can't find the broken commit.

CyberShadow:
@dkorpel  It's here:
https://github.com/dlang/phobos/pull/8092#issuecomment-842528947

CyberShadow (approved):

ibuclaw:
Logs of #8092 are gone now, but also seen in:
- #8093 https://buildkite.com/dlang/phobos/builds/5499#cc42663b-941b-4eb7-90fe-276d34a6d1d3/472-3414
- #8077 https://buildkite.com/dlang/phobos/builds/5502#2536380c-2b57-4ba2-b58a-48dc21d29f98/471-3463

And a few others at the time.

dkorpel:
So `void runbench.runTests(runbench.Config)` can't find `pure @safe immutable(char)[] std.path.absolutePath(immutable(char)[], lazy immutable(char)[])` because there's a new `return` annotation changing the mangle. Is the benchmark linking with new Phobos but importing old Phobos?

nordlow:
How to move forward with this?

Shall I try adding the exact same PR again without the change to `absolutePath`?

And then add a second PR with only the change to `absolutePath`?

ibuclaw:
> How to move forward with this?
> 
> Here's a similar PR with the change to `absolutePath` excluded:
> #8106
> 
> And then add a second PR with only the change to `absolutePath`?

Seems reasonable to me.

PR #8107, opened 2021-05-19
Annotate absolutePath in std/path.d to please dlang/dmd#12520

nordlow:
Follow-up to https://github.com/dlang/phobos/pull/8106 to see if the same linker-error triggered by https://github.com/dlang/phobos/pull/8090 happens again. IRC, the linker error was caused by picking a symbol with [incorrect mangling](https://github.com/dlang/phobos/pull/8103#issuecomment-842942889) of `absolutePath`.

As

`echo _D3std4path12absolutePathFNaNfAyaLQeZQh | ddemangle `

prints

`pure @safe immutable(char)[] std.path.absolutePath(immutable(char)[], lazy immutable(char)[])`

it seems the error was caused by the CI, for some unknown reason, not linking with the recently built version of phobos.

Previous try was reverted via https://github.com/dlang/phobos/pull/8103.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -143,7 +143,7 @@ else
 
 version (StdUnittest) private struct TestAliasedString
 {
-    string get() @safe @nogc pure nothrow { return _s; }
+    string get() @safe @nogc pure nothrow return scope { return _s; }
     alias get this;
     @disable this(this);
     string _s;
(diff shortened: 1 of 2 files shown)


thewilsonator:
cc @ibuclaw

nordlow:
Green now. Shal we try again, @ibuclaw?

PR #8110, opened 2021-05-23
[stable] std.format.internal.write: Replace brittle runtime check by compile-time check

kinke:
When checking whether `T.toString` for some class `T` is inherited from base class `Object`.

The runtime check fails when using a druntime DLL on Windows, where `Object.toString` in other DLLs/executables is a trampoline (in the druntime import library) to the dllimported function, with its own identity.

### DIFF
--- a/std/format/internal/write.d
+++ b/std/format/internal/write.d
@@ -2313,11 +2313,7 @@ if (is(T == class) && !is(T == enum))
         }
         else
         {
-            // string delegate() dg = &val.toString;
-            Object o = val;     // workaround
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):
I don't suppose you have a test for this? Also is there a bugzilla report for this?

kinke:
It's covered by unittests in the same module. The problem only shows up with a separate druntime DLL on Windows, covered by LDC now; that's not an option for DMD.

RazvanN7 (approved):

PR #8111, opened 2021-05-26
toStringz() - try to use reserved memory via capacity/length

vitalfadeev:
Module std.string. Function toStringz
Try to use reserved memory via capacity/length.
Goal -  prevent memory allocation. 
PROS: It faster in case when string has reserved memory. When no reserved memory then to allocate new string ( using old code ).
CONS: +1 operation for testing capacity / length.
Forum: https://forum.dlang.org/post/drlqiudbvdbhxsgeyfbk@forum.dlang.org

### DIFF
--- a/std/string.d
+++ b/std/string.d
@@ -343,12 +343,38 @@ do
      return s;
      +/
 
-    // Need to make a copy
-    auto copy = new char[s.length + 1];
-    copy[0 .. s.length] = s[];
-    copy[s.length] = 0;
+    /+
+    Try use existent memory. Test reserved memory.
+    if reserved 
+      add trailing zero
+      return pointer
+    else
+      allocate new memory block
+      copy string
+      add zero
+      return pointer
+    +/
 
-    return &assumeUnique(copy)[0];
+    // reserve
+    if ( s.capacity > s.length )
+    {
+        char* cptr = cast( char* ) s.ptr; // C ptr
+        char* zptr = cptr + s.length;     // zero ptr
+        *zptr = '\0';        
+
+        return cast( immutable(char)* ) cptr;
+    }
+    else
+
+    // allocate
+    {
+        // Need to make a copy
+        auto copy = new char[s.length + 1];
+        copy[0 .. s.length] = s[];
+        copy[s.length] = 0;
+
+        return &assumeUnique(copy)[0];
+    }
 }
 
 ///


std/string.d
```diff
@@ -343,12 +343,38 @@ do
...
+    // reserve
+    if ( s.capacity > s.length )
+    {
+        char* cptr = cast( char* ) s.ptr; // C ptr
+        char* zptr = cptr + s.length;     // zero ptr
+        *zptr = '\0';        
 
-    return &assumeUnique(copy)[0];
+        return cast( immutable(char)* ) cptr;
+    }
```
aG0aep6G, line 366: You can't just write beyond `s.length`. You have to inform the GC that the array grows. Otherwise, appending to the original string will stomp over the terminator.

aG0aep6G (changes requested):

std/string.d
```diff
@@ -343,12 +343,38 @@ do
...
+      copy string
+      add zero
+      return pointer
+    +/
+
+    // reserve
+    if ( s.capacity > s.length )
+    {
+        char* cptr = cast( char* ) s.ptr; // C ptr
+        char* zptr = cptr + s.length;     // zero ptr
```
RazvanN7, line 362: `zptr` is not needed, you can always use `*(cptr+s.length) = "\0"`

std/string.d
```diff
@@ -343,12 +343,38 @@ do
...
+        *zptr = '\0';        
 
-    return &assumeUnique(copy)[0];
+        return cast( immutable(char)* ) cptr;
```
RazvanN7, line 365: The same applies for this cast; no need for the white spaces. Also, you cannot return `cptr` since it is initialized with a `scope` variable (parameter `s`). That is why the autotester is complaining.

std/string.d
vitalfadeev: Yes.  True.
Zero terminator not keeped after concatenate source string with other string.
Closing pull request.

auto dString = "D string" ~ 2.to!string;
auto cString = dString.toStringz();
dString = dString ~ "tail";
// cString[ $-1 ] != '\'0';

vitalfadeev:
Zero terminator not keeped after concatenate source string with other string.
Closing pull request.

auto dString = "D string" ~ 2.to!string;
auto cString = dString.toStringz();
dString = dString ~ "tail";
// cString[ $-1 ] != '\'0';

PR #8113, opened 2021-05-28
Annotate more of phobos with `return` and `scope`

dkorpel:
Continuation of https://github.com/dlang/phobos/pull/8076, but for https://github.com/dlang/dmd/pull/12010 instead of https://github.com/dlang/dmd/pull/10924

I'm still figuring out what to do with range functions like `array`, `text` and `toCase`. `scope` fails to be inferred for simple `char[]` cases, but putting `scope` on them forces every InputRange it's called on to be `scope` compatible, while a generic `InputRange` can escape its innards through global variables or Exceptions. Not that I *want* to support such ranges, but it would make transitioning to `-dip1000` more diffcult for users, so it may be better to leave them be and use work-arounds for current Phobos functions that need them to be `scope`, like I did in `std.getopt` by replacing `split` (not `scope` compatible, returns array of slices)  with `splitter` (lazy, `scope` compatible).

### DIFF
--- a/std/algorithm/mutation.d
+++ b/std/algorithm/mutation.d
@@ -1074,7 +1074,7 @@ Params:
 */
 void move(T)(ref T source, ref T target)
 {
-    moveImpl(source, target);
+    moveImpl(target, source);
 }
 
 /// For non-struct types, `move` just performs `target = source`:
@@ -1244,7 +1244,7 @@ pure nothrow @safe @nogc unittest
     static assert(is(typeof({ S s; move(s, s); }) == T));
 }
 
-private void moveImpl(T)(ref T source, ref T target)
+private void moveImpl(T)(ref scope T target, ref return scope T source)
 {
     import std.traits : hasElaborateDestructor;
 
@@ -1257,10 +1257,10 @@ private void moveImpl(T)(ref T source, ref T target)
         static if (hasElaborateDestructor!T) target.__xdtor();
     }
     // move and emplace source into target
-    moveEmplaceImpl(source, target);
+    moveEmplaceImpl(target, source);
 }
 
-private T moveImpl(T)(ref T source)
+private T moveImpl(T)(ref return scope T source)
 {
     // Properly infer safety from moveEmplaceImpl as the implementation below
     // might void-initialize pointers in result and hence needs to be @trusted
@@ -1269,10 +1269,10 @@ private T moveImpl(T)(ref T source)
     return trustedMoveImpl(source);
 }
 
-private T trustedMoveImpl(T)(ref T source) @trusted
+private T trustedMoveImpl(T)(ref return scope T source) @trusted
 {
     T result = void;
-    moveEmplaceImpl(source, result);
+    moveEmplaceImpl(result, source);
     return result;
 }
 
@@ -1415,7 +1415,7 @@ private T trustedMoveImpl(T)(ref T source) @trusted
     move(x, x);
 }
 
-private void moveEmplaceImpl(T)(ref T source, ref T target)
+private void moveEmplaceImpl(T)(ref scope T target, ref return scope T source)
 {
     import core.stdc.string : memcpy, memset;
     import std.traits : hasAliasing, hasElaborateAssign,
@@ -1486,7 +1486,7 @@ private void moveEmplaceImpl(T)(ref T source, ref T target)
  */
 void moveEmplace(T)(ref T source, ref T target) pure @system
 {
-    moveEmplaceImpl(source, target);
+    moveEmplaceImpl(target, source);
 }
 
 ///
@@ -2388,7 +2388,7 @@ Range remove(alias pred, SwapStrategy s = SwapStrategy.stable, Range)(Range rang
 @nogc @safe unittest
 {
     // @nogc test
-    int[10] arr = [0,1,2,3,4,5,6,7,8,9];
+    static int[] arr = [0,1,2,3,4,5,6,7,8,9];
     alias pred = e => e < 5;
 
     auto r = arr[].remove!(SwapStrategy.unstable)(0);
--- a/std/array.d
+++ b/std/array.d
@@ -1953,7 +1953,7 @@ private enum bool hasCheapIteration(R) = isArray!R;
    See_Also:
         For a lazy version, see $(REF joiner, std,algorithm,iteration)
   +/
-ElementEncodingType!(ElementType!RoR)[] join(RoR, R)(RoR ror, scope R sep)
+ElementEncodingType!(ElementType!RoR)[] join(RoR, R)(RoR ror, R sep)
 if (isInputRange!RoR &&
     isInputRange!(Unqual!(ElementType!RoR)) &&
     isInputRange!R &&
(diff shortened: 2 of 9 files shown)


dkorpel:
This should be good to go now: I can now locally build Phobos' unittests with https://github.com/dlang/dmd/pull/12010

I undid some of the scope annotations of [Per's PR](https://github.com/dlang/phobos/pull/8076) because they were only done to satisfy the requirement that `pure` functions always take `scope` arguments, [which I argued against](https://github.com/dlang/dmd/pull/10924#issuecomment-844170312).

Many range functions don't have correct `scope` inference yet, so I replaced some static arrays in `@safe` unittests with arrays in static data. They can be enhanced to work with scope ranges later, for now it's important to fix dip1000 first

std/algorithm/mutation.d
```diff
@@ -1415,7 +1415,7 @@ private T trustedMoveImpl(T)(ref T source) @trusted
     move(x, x);
 }
 
-private void moveEmplaceImpl(T)(ref T source, ref T target)
+private void moveEmplaceImpl(T)(ref scope T target, ref return scope T source)
```
thewilsonator, line 1418: Why the order change?

std/array.d
```diff
@@ -1953,7 +1953,7 @@ private enum bool hasCheapIteration(R) = isArray!R;
    See_Also:
         For a lazy version, see $(REF joiner, std,algorithm,iteration)
   +/
-ElementEncodingType!(ElementType!RoR)[] join(RoR, R)(RoR ror, scope R sep)
+ElementEncodingType!(ElementType!RoR)[] join(RoR, R)(RoR ror, R sep)
```
thewilsonator, line 1956: this removes `scope` I thought this PR was trying to add it, am I missing something?

std/path.d
```diff
@@ -2747,7 +2747,7 @@ else version (Posix)
     See_Also:
         $(LREF asAbsolutePath) which does not allocate
 */
-string absolutePath(return scope string path, lazy string base = getcwd())
```
thewilsonator, line 2750: ditto

std/path.d
```diff
@@ -2893,7 +2893,7 @@ if (isConvertibleToString!R)
     `Exception` if the specified _base directory is not absolute.
 */
 string relativePath(CaseSensitive cs = CaseSensitive.osDefault)
-    (scope return string path, lazy string base = getcwd())
```
thewilsonator, line 2896: ditto

std/random.d
```diff
@@ -3362,8 +3362,8 @@ if (isRandomAccessRange!Range)
     // Optionally @nogc std.random.randomCover
     // https://issues.dlang.org/show_bug.cgi?id=14001
     auto rng = Xorshift(123_456_789);
-    int[5] sa = [1, 2, 3, 4, 5];
-    auto r = randomCover(sa[], rng);
+    static immutable int[] sa = [1, 2, 3, 4, 5];
```
thewilsonator, line 3365: why is the change from a static array to a `static int[]` necessary?

std/algorithm/mutation.d
```diff
@@ -2388,7 +2388,7 @@ Range remove(alias pred, SwapStrategy s = SwapStrategy.stable, Range)(Range rang
 @nogc @safe unittest
 {
     // @nogc test
-    int[10] arr = [0,1,2,3,4,5,6,7,8,9];
+    static int[] arr = [0,1,2,3,4,5,6,7,8,9];
```
thewilsonator, line 2391: ditto

thewilsonator (commented):
Pardon my ignorance here

thewilsonator:
This otherwise looks OK

std/algorithm/mutation.d
dkorpel: Same as in druntime: https://github.com/dlang/druntime/pull/3480#issuecomment-845480887

std/random.d
dkorpel: A slice to a static array on the stack is `scope`, and `randomCover` doesn't support `scope` ranges yet. The test is `@nogc`, so an array literal can't be used.

std/algorithm/mutation.d
thewilsonator: OK makes sense

std/random.d
thewilsonator: OK

thewilsonator (approved):

std/path.d
dkorpel: `relativePath` calls `asRelativePath` which takes input ranges, and scope inference fails there

std/path.d
dkorpel: `absoultePath` calls `chainPath(baseVar, path).array`, and `scope` inference fails for `.array` currently

std/array.d
dkorpel: Range functions should ideally infer `scope` based on `empty()`, `popFront()`, `front()`, copy constructors etc. 
`scope` inference is not that good yet however ([issue 20674](https://issues.dlang.org/show_bug.cgi?id=20674)) so sometimes `scope` is manually applied to help the compiler. With issue 20150 fixed, this `scope R sep` gives an error:

> scope variable `sep` assigned to non-scope parameter `r` calling std.array.array!(Once).array

So the long term fix is to ensure `scope` inference works for both `join` and `array`, but that's an enhancement. For now it's important to get the dip1000 bug fix in.

thewilsonator:
Thanks for clearing that all up.

std/array.d
MoonlightSentinel: Won't this break existing code if the inference fails?

dkorpel: All the accepts-invalid fixes for dip1000 are breaking changes. Luckily dip1000 is still experimental.

MoonlightSentinel: But these changes are neither specific to DIP1000 nor about actually invalid code. It would be better to first improve the inference s.t. there is no unecessary breakage.

dkorpel: What code do you think would break without `-preview=dip1000`?

MoonlightSentinel: I don't have aspecific example ATM, will post one if I find a small example. 
But we should strive to not break valid code. `-dip1000` has existed for quite some time and is certainly used in the wild.

dkorpel: https://github.com/dlang/dmd/pull/12010 is going to break such dip1000 code in the wild undoubtedly, I just tested it on my own 37 KLOC dip1000 library and I had to add 3 `scope` annotations. Do you think we need a second switch `preview=dip1000fixed`?

MoonlightSentinel: Neither said nor implied that - some breakage will be unavoidable. I'm just concerned that these changes introduce a temporary regression that could be avoided by first improving the `scope` inference.

But the actual fallout might be neglegible given that your projects required minimal changes and buildkite passes as is.

thewilsonator:
please rebase or force push to start CircleCI

std/algorithm/mutation.d
```diff
@@ -1074,7 +1074,7 @@ Params:
 */
 void move(T)(ref T source, ref T target)
 {
-    moveImpl(source, target);
+    moveImpl(target, source);
```
nordlow, line 1077: Why the reversion of arguments?

dkorpel: https://github.com/dlang/druntime/pull/3480#issuecomment-845480887

nordlow: Thanks

dkorpel: I probably should have made a comment in the source code as well

PR #8115, opened 2021-05-29
Use splitter instead of split in getopt

dkorpel:
Split off from https://github.com/dlang/phobos/pull/8113
`split` eagerly returns a GC-allocated array of slices. Since `scope` only applies to the top level, this returned array escapes the input string as far as dip1000 is concerned.

### DIFF
--- a/std/getopt.d
+++ b/std/getopt.d
@@ -1106,7 +1106,7 @@ private struct configuration
 private bool optMatch(string arg, scope string optPattern, ref string value,
     configuration cfg) @safe
 {
-    import std.array : split;
+    import std.algorithm.iteration : splitter;
     import std.string : indexOf;
     import std.uni : toUpper;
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

std/getopt.d
```diff
@@ -1106,7 +1106,7 @@ private struct configuration
 private bool optMatch(string arg, scope string optPattern, ref string value,
     configuration cfg) @safe
 {
-    import std.array : split;
+    import std.algorithm : splitter;
```
thewilsonator, line 1109:
```
grep -nr 'import std.algorithm : ' $(find etc std -name '*.d') ; test $? -eq 1
--
&nbsp; | std/getopt.d:1109:    import std.algorithm : splitter;
```

I think you need to `import std.algorithm.mutation : splitter;` or whatever the submodule is.

PR #8116, opened 2021-05-29
Add `return scope` to realloc

dkorpel:
Split off from https://github.com/dlang/phobos/pull/8113
Since realloc could return the input if there is enough capacity, it should be `return scope`. While in practice realloc is all `@system` and `@trusted` code, it does seem to trip up issue 20150.

### DIFF
--- a/std/internal/cstring.d
+++ b/std/internal/cstring.d
@@ -145,7 +145,10 @@ if (isSomeChar!To && (isInputRange!From || isSomeString!From) &&
         {
             if (i + 1 == p.length)
             {
-                heapBuffer = trustedRealloc(p, strLength, heapBuffer is null);
+                if (heapBuffer is null)
+                    heapBuffer = trustedReallocStack(p, strLength);
+                else
+                    heapBuffer = trustedRealloc(heapBuffer);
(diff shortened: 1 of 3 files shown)


dkorpel:
Turns out that `std.internal.cstring: tempCString` is tricky, since it dynamically switches between stack and heap memory, which trips up `scope`.

dkorpel:
> Turns out that `std.internal.cstring: tempCString` is tricky, since it dynamically switches between stack and heap memory, which trips up `scope`.

I now split up the stack-logic and heap-logic into two functions, so that the former can be `scope` while the latter `return scope`.

RazvanN7 (approved):

RazvanN7:
Please rebase to restart circleCI

wilzbach:
@RazvanN7 maybe you could rebase for the author? The ping, rebase, wait for auto-merge label to be re-added cycle seems a bit inefficient.

dkorpel:
> @wilzbach Of course I can do the rebase myself, however, I think it's better to keep the PR author engaged.

Having to rebase my PRs over and over again because of CI issues is not very engaging though 😬

PR #8120, opened 2021-05-31
file: document dirEntries as no garantee of sorted output

ljmf00:
Some unittests should compare with a sorted array to be
correct.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -2917,6 +2917,8 @@ if (isConvertibleToString!R)
 @system unittest
 {
     import std.algorithm.comparison : equal;
+    import std.algorithm.sorting : sort;
+    import std.array : array;
     import std.path : buildPath;
 
     auto cwd = getcwd;
@@ -2927,9 +2929,9 @@ if (isConvertibleToString!R)
     dir.buildPath("a").write(".");
     dir.chdir; // step into dir
     "b".write(".");
-    dirEntries(".", SpanMode.shallow).equal(
-        [".".buildPath("b"), ".".buildPath("a")]
-    );
+    assert(dirEntries(".", SpanMode.shallow).array.sort.equal(
+        [".".buildPath("a"), ".".buildPath("b")]
+    ));
 }
 
 @safe unittest
@@ -4605,6 +4607,8 @@ enum SpanMode
 {
     import std.algorithm.comparison : equal;
     import std.algorithm.iteration : map;
+    import std.algorithm.sorting : sort;
+    import std.array : array;
     import std.path : buildPath, relativePath;
 
     auto root = deleteme ~ "root";
@@ -4613,19 +4617,19 @@ enum SpanMode
 
     root.buildPath("animals").mkdir;
     root.buildPath("animals", "cat").mkdir;
-    root.buildPath("animals", "dog").mkdir;
-    root.buildPath("plants").mkdir;
 
     alias removeRoot = (return scope e) => e.relativePath(root);
 
-    root.dirEntries(SpanMode.shallow).map!removeRoot.equal(
-        ["plants", "animals"]);
+    assert(root.dirEntries(SpanMode.depth).map!removeRoot.equal(
+        [buildPath("animals", "cat"), "animals"]));
 
-    root.dirEntries(SpanMode.depth).map!removeRoot.equal(
-        ["plants", "animals/dog", "animals/cat", "animals"]);
+    assert(root.dirEntries(SpanMode.breadth).map!removeRoot.equal(
+        ["animals", buildPath("animals", "cat")]));
 
-    root.dirEntries(SpanMode.breadth).map!removeRoot.equal(
-        ["plants", "animals", "animals/dog", "animals/cat"]);
+    root.buildPath("plants").mkdir;
+
+    assert(root.dirEntries(SpanMode.shallow).array.sort.map!removeRoot.equal(
+        ["animals", "plants"]));
 }
 
 private struct DirIteratorImpl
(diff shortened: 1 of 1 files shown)


std/file.d
```diff
@@ -2927,7 +2929,7 @@ if (isConvertibleToString!R)
     dir.buildPath("a").write(".");
     dir.chdir; // step into dir
     "b".write(".");
-    dirEntries(".", SpanMode.shallow).equal(
+    dirEntries(".", SpanMode.shallow).array.sort.equal(
         [".".buildPath("b"), ".".buildPath("a")]
```
CyberShadow, line 2933:
```suggestion
        [".".buildPath("a"), ".".buildPath("b")]
```
?

std/file.d
```diff
@@ -4618,13 +4622,13 @@ enum SpanMode
 
     alias removeRoot = (e) => e.relativePath(root);
 
-    root.dirEntries(SpanMode.shallow).map!removeRoot.equal(
+    root.dirEntries(SpanMode.shallow).array.sort.map!removeRoot.equal(
         ["plants", "animals"]);
```
CyberShadow, line 4626:
```suggestion
        ["animals", "plants"]);
```
?

std/file.d
```diff
@@ -4898,6 +4902,9 @@ public:
     directory is traversed. The name of each iterated directory entry
     contains the absolute or relative _path (depending on _pathname).
 
+    $(NOTE It's not guaranteed to return a sorted iterator as it depends on the
+    filesystem implementation)
```
CyberShadow, line 4906:
```suggestion
    $(NOTE The order of returned directory entries is as it is provided by the
    operating system / filesystem, and may not follow any particular sorting.)
```

std/file.d
CyberShadow: etc.

std/file.d
ljmf00: Ooof, this is weird. It makes sense the change actually, but it compiles in my machine and I didn't notice that. Maybe my brain is completely fried :upside_down_face:

CyberShadow: It will compile either way. Do you mean that it runs (and the assert passes)?

ljmf00: > It will compile either way. Do you mean that it runs (and the assert passes)?

I got the point now. I was not crazy. There's actually no testing there, no `assert()` around the equal. :facepalm:

CyberShadow: Ah, whoops!

std/file.d
```diff
@@ -4618,13 +4622,13 @@ enum SpanMode
...
+    root.dirEntries(SpanMode.shallow).array.sort.map!removeRoot.equal(
         ["plants", "animals"]);
 
-    root.dirEntries(SpanMode.depth).map!removeRoot.equal(
+    root.dirEntries(SpanMode.depth).array.sort.map!removeRoot.equal(
         ["plants", "animals/dog", "animals/cat", "animals"]);
 
-    root.dirEntries(SpanMode.breadth).map!removeRoot.equal(
+    root.dirEntries(SpanMode.breadth).array.sort.map!removeRoot.equal(
         ["plants", "animals", "animals/dog", "animals/cat"]);
```
ljmf00, line 4632: This gets problematic here as this is an example. Suggestions to make this unittest example clear?

CyberShadow: Why not just wrap the line in `assert(...)`?

When rendered on the website, it will be converted to a `writeln`, with the expected result in a comment.

ljmf00: Because `root.dirEntries(SpanMode.depth)` and `root.dirEntries(SpanMode.breadth)` sorted are the same but the example what to show the difference between them.

std/file.d
ljmf00:
```
/dev/shm/dtest/work/repo/dlang.org/web/phobos-prerelease/std_file.html:1544: [UNDEFINED MACRO] </div><div class="description"><!--UNDEFINED MACRO: "NOTE"-->
```

I thought NOTE macro was a thing, apparently, it doesn't.

std/file.d
wilzbach: Yep, just do `assert(dirEntries(".", SpanMode.shallow) == [".".buildPath("a"), ".".buildPath("b")])` and dlang.org should do the rest

RazvanN7:
@ljmf00 please rebase and apply the suggested feedback.

ljmf00:
> @ljmf00 please rebase and apply the suggested feedback.

Done. To reflect the actual behavior of `depth` and `breadth` search, I needed to change the tests.

CyberShadow:
Please have a look at the CI failures, I believe some of their cause was pointed out in a review that has not yet been addressed.

ljmf00:
> Please have a look at the CI failures, I believe some of their cause was pointed out in a review that has not yet been addressed.

Although the tests were not running, the tests seem correct, but the SpanMode.depth in Windows seems not behaving as it should be.

ljmf00:
@CyberShadow problem solved. The tests were wrong due to the malformed paths for Windows. `"foo/bar"` is different in windows `"foo\\bar"`. It's fixed with `buildPath("foo", "bar")`.

std/file.d
CyberShadow: @ljmf00 Would you mind fixing this too please? (Fix order and add `assert`)

Previously it was maybe-sometimes-working, now this looks like it never works.

ljmf00: Yes, I'm sorry, I forgot it. Should be good now

CyberShadow (approved):

CyberShadow:
LGTM, thanks!

RazvanN7 (approved):

PR #8123, opened 2021-06-01
Use `static foreach` for correct `scope` inference

dkorpel:
Part of https://github.com/dlang/phobos/pull/8113

Turns `foreach` over an AliasSeq into `static foreach` because of [issue 21209 - scope attribute inference with does not work well with foreach](https://issues.dlang.org/show_bug.cgi?id=21209). Without this, `text` and `chain` are not `scope` compatible.

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -454,6 +454,23 @@ template to(T)
     assert(text(null) == "null");
 }
 
+// Test `scope` inference of parameters of `text`
+@safe unittest
+{
+    static struct S
+    {
+        int* x; // make S a type with pointers
+        string toString() const scope
+        {
+            static int g = 0; // force toString to be impure for:
+            g++; // https://issues.dlang.org/show_bug.cgi?id=20150
+            return "S";
+        }
+    }
+    scope S s;
+    assert(text("a", s) == "aS");
+}
+
 // Tests for issue 11390
 @safe pure /+nothrow+/ unittest
 {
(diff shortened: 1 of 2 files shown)


thewilsonator:
can you add a test for this?

dkorpel:
I added a test for `text`. I tried to add one for chain, but I was unable with the current compiler behavior. Because of issue 20150 (-dip1000 defeated by pure), I have to force `chain` to be impure somehow, which I can only do by using an impure copy constructor. But then `scope` inference fails even after this patch, and I don't know why.

```D
// Test `scope` inference on `chain` parameters
version(unittest) private struct ImpureCopyCtor
{
    // not pure, for https://issues.dlang.org/show_bug.cgi?id=20150
    this(ref ImpureCopyCtor r) scope @nogc nothrow {}
    bool empty = true;
    void popFront() {}
    int* front;
}

@safe unittest
{
    auto chain1(ref ImpureCopyCtor r)
    {
        return chain(r, r);
    }
    scope ImpureCopyCtor r;
    chain1(r);
}
```

thewilsonator (approved):

thewilsonator:
CI seems stuck, can you please rebase or force push?

RazvanN7 (approved):

RazvanN7:
The autotester seems to be stuck. Rebase this to restart it.

thewilsonator:
Still stuck.

PR #8124, opened 2021-06-01
[stable] std.math.exponential: Revise recent pow() changes with integral exponent

kinke:
Revising #7823.

### DIFF
--- a/std/math/exponential.d
+++ b/std/math/exponential.d
@@ -65,6 +65,7 @@ Unqual!F pow(F, G)(F x, G n) @nogc @trusted pure nothrow
 if (isFloatingPoint!(F) && isIntegral!(G))
 {
     import core.math : fabs;
+    import std.math : floatTraits;
     import std.math.rounding : floor;
     import std.math.traits : isNaN;
     import std.traits : Unsigned;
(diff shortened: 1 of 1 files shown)


std/math/exponential.d
```diff
@@ -112,44 +113,22 @@ if (isFloatingPoint!(F) && isIntegral!(G))
 
     enum uint bias = F.max_exp - 1;
 
-    static if (is(F == float))
```
kinke, line 115: Issue 1: qualifiers ignored - slow generic branch taken for all non-mutable FP types.

std/math/exponential.d
```diff
@@ -112,44 +113,22 @@ if (isFloatingPoint!(F) && isIntegral!(G))
...
-    {
-        float f = cast(float) v;
-        uint ival = () @trusted { return *cast(uint*) &f; }();
```
kinke, line 118: `pow()` is `@trusted`, no need for these ugly lambdas.

std/math/exponential.d
```diff
@@ -112,44 +113,22 @@ if (isFloatingPoint!(F) && isIntegral!(G))
...
-        ulong ex = (ival >> 23) & 255;
-    }
-    else static if (is(F == double) || (is(T == real) && T.mant_dig == double.mant_dig))
```
kinke, line 121: No `T` here.

std/math/exponential.d
```diff
@@ -112,44 +113,22 @@ if (isFloatingPoint!(F) && isIntegral!(G))
...
-
-        // In the general case we have to fall back to log2, which is slower, but still
-        // a certain speed gain compared to not bailing out early.
-            // skipping subnormals by setting ex = bias
-        ulong ex = fabs(v) == F.infinity ? 2 * bias + 1 :
-            (fabs(v) < F.min_normal ? bias : cast(ulong) (floor(log2(fabs(v))) + bias));
+        alias RT = floatTraits!real;
+        ushort* vu = cast(ushort*) &v;
+        ex = ((vu[RT.EXPPOS_SHORT] & RT.EXPMASK) >> RT.EXPSHIFT)
+             - (cast(long) (RT.EXPBIAS >> RT.EXPSHIFT) + 1) + bias;
```
kinke, line 131: `v` is always a `real` - the previous code handled single/double/extended precision separately, converted v to variable-sized `F` and finally reinterpreted it as integer to get the exponent offsetted by `bias`. Here I simply extract the exponent directly from `v`, undo the `real` bias and apply the `F` bias again such as not having to touch the rest below.

std/math/exponential.d
```diff
@@ -112,44 +113,22 @@ if (isFloatingPoint!(F) && isIntegral!(G))
...
-            // in CTFE we cannot access the bit patterns and have therefore to
-            // fall back to the (slower) general case
-            // skipping subnormals by setting ex = bias
-            ex = fabs(v) == F.infinity ? 2 * bias + 1 :
-                (fabs(v) < F.min_normal ? bias : cast(ulong) (floor(log2(fabs(v))) + bias));
-        }
-        else
-        {
-            ulong[2] ival = () @trusted { return *cast(ulong[2]*) &v; }();
-            ex = ival[1] & 32767;
```
kinke, line 141: `real.sizeof == 12` on at least some x86 systems. Reinterpeting as 2 longs and reading the 2nd one exceeds the allocated stack size and is invalid.

PR #8125, opened 2021-06-01
Fix local variable lifetimes in systime parsing

dkorpel:
Blocking https://github.com/dlang/dmd/pull/12620

### DIFF
--- a/std/datetime/systime.d
+++ b/std/datetime/systime.d
@@ -8751,6 +8751,7 @@ public:
         auto found = (skipFirst ? str[1..$] : str).byCodeUnit.find('.', 'Z', '+', '-');
         auto dateTimeStr = str[0 .. $ - found[0].length];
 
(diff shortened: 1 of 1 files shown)


std/datetime/systime.d
```diff
@@ -8751,26 +8751,27 @@ public:
...
+        typeof(str.byCodeUnit) foundTZ; // needs to have longer lifetime than zoneStr
         typeof(str) fracSecStr;
         typeof(str) zoneStr;
 
         if (found[1] != 0)
         {
             if (found[1] == 1)
             {
-                auto foundTZ = found[0].find('Z', '+', '-');
+                foundTZ = found[0].find('Z', '+', '-')[0];
```
dkorpel, line 8762: I can't assign a Tuple here, since `Tuple.opAssign` is not `scope` compatible, so I keep the string and discard the index of the matched string. It's only used in a `!= 0` comparison to see if there was any match, so I replaced it by a `foundTZ.length` check.

thewilsonator (approved):

PR #8126, opened 2021-06-02
[stable] Revert #7823 - Faster pow implementation for integral exponents

kinke:
There are multiple issues with that commit, incl. out-of-bounds read on x86 systems and a significant performance decrease for qualified floating-point types; see #8124 for a full list of flaws.

I'm questioning the tackled goal in the first place, because I expect users to use that `pow` mainly with small integral exponents, *not* expecting underflows to 0 or overflows to ±infinity, so optimizing for these cases and sacrificing performance for the regular case doesn't make a lot of sense to me.

### DIFF
--- a/std/math/exponential.d
+++ b/std/math/exponential.d
@@ -64,13 +64,7 @@ version (D_HardFloat)
 Unqual!F pow(F, G)(F x, G n) @nogc @trusted pure nothrow
 if (isFloatingPoint!(F) && isIntegral!(G))
 {
-    import core.math : fabs;
-    import std.math.rounding : floor;
-    import std.math.traits : isNaN;
     import std.traits : Unsigned;
-
-    // NaN ^^ 0 is an exception defined by IEEE (yields 1 instead of NaN)
-    if (isNaN(x)) return n ? x : 1.0;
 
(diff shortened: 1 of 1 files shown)


std/math/exponential.d
```diff
@@ -69,9 +69,6 @@ if (isFloatingPoint!(F) && isIntegral!(G))
     import std.math.traits : isNaN;
     import std.traits : Unsigned;
 
-    // NaN ^^ 0 is an exception defined by IEEE (yields 1 instead of NaN)
-    if (isNaN(x)) return n ? x : 1.0;
-
```
kinke, line 74: `n == 0` already returns and used to return `1.0`. A `NaN` should be already propagated in the other cases too.

thewilsonator: Is there a test that covers this execution path?

kinke: I don't know; no test was added in #7823 in the first place though, and nobody complained back then.

Geod24 (approved):
Let's revert so that it doesn't affect the release.
CC @berni44 : Can you resubmit the PR with the issues pointed out by @kinke addressed ?

PR #8127, opened 2021-06-02
[stable] std.math.operations: Slightly revise extractBitpattern()

kinke:
Incl. fixing an out-of-bounds read on 32-bit x86 with `real.sizeof == 12` (detected via an LLVM assertion after merging to LDC).

### DIFF
--- a/std/math/operations.d
+++ b/std/math/operations.d
@@ -1729,9 +1729,9 @@ if (isFloatingPoint!T)
     {
         if (__ctfe)
         {
-            import core.math : fabs;
+            import core.math : fabs, ldexp;
             import std.math.rounding : floor;
-            import std.math.traits : isInfinity, isNaN;
+            import std.math.traits : isInfinity, isNaN, signbit;
             import std.math.exponential : log2;
 
             if (isNaN(val) || isInfinity(val))
@@ -1750,19 +1750,9 @@ if (isFloatingPoint!T)
             }
             else
             {
-                if (ret.exponent > 16382 + 64) // bias + bits of ulong
-                    val /= 2.0L ^^ (ret.exponent - (16382 + 64));
-                else
-                {
-                    auto delta = 16382 + 64 - (ret.exponent == 0 ? 1 : ret.exponent); // -1 in case of subnormals
-                    if (delta > 16383)
-                    {
-                        // need two steps to avoid overflow
-                        val *= 2.0L ^^ 16383;
-                        delta -= 16383;
-                    }
-                    val *= 2.0L ^^ delta;
-                }
+                auto delta = 16382 + 64 // bias + bits of ulong
+                             - (ret.exponent == 0 ? 1 : ret.exponent); // -1 in case of subnormals
+                val = ldexp(val, delta); // val *= 2^^delta
 
                 ulong tmp = cast(ulong) fabs(val);
                 if (ret.exponent != 32767 && ret.exponent > 0 && tmp <= ulong.max / 2)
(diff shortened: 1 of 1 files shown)


std/math/operations.d
```diff
@@ -1750,19 +1750,9 @@ if (isFloatingPoint!T)
             }
             else
             {
-                if (ret.exponent > 16382 + 64) // bias + bits of ulong
-                    val /= 2.0L ^^ (ret.exponent - (16382 + 64));
```
kinke, line 1754:
```
val / 2.0L ^^  (ret.exponent - (16382 + 64)) =
val * 2.0L ^^ -(ret.exponent - (16382 + 64)) =
val * 2.0L ^^ (16382 + 64 - ret.exponent)   => as in else branch (ret.exponent guaranteed to be != 0)
```

std/math/operations.d
```diff
@@ -1717,7 +1717,7 @@ if (isFloatingPoint!T)
     bool negative;
 }
 
-FloatingPointBitpattern!T extractBitpattern(T)(T val)
+FloatingPointBitpattern!T extractBitpattern(T)(T val) @trusted
```
thewilsonator, line 1720: please leave the trusted lambdas in place and remove this

berni44:
Sorry, due to personal reasons, I'll not find the time to review this. It was not doable by just a first glimse.

std/math/operations.d
RazvanN7: It looks like `ldexp` is not @safe. Anyway, having such functions marked as `@trusted` is bad practice.

kinke: `core.math.ldexp` is @safe. I get eye cancer from these trusted lambdas and prefer an @trusted for the whole function instead of 4 lambdas, especially for such a trivial function (except for the CTFE part) with the grand task of splitting the packed bits into 3 integral components, with the obvious need to reinterpret the FP value as integer.

RazvanN7: Maybe I am missing something, but it [seems](https://github.com/dlang/druntime/blob/master/src/core/math.d#L106) that `ldexp` is not **@safe** annotated.

I agree that the lambdas don't look very well, but they have the added benefit of reducing the search area when something goes wrong. Otherwise, someone unfamiliar with the code needs to scan entire function (~100 LOC). Even if you find the code trivial, that might not be the case for other developers. Also, the phobos convention is to reduce the trusted area to a minimum.

kinke: >Maybe I am missing something

https://github.com/dlang/druntime/blob/46bca630e501f01ce78124220d766e0f643547b0/src/core/math.d#L31`

> Otherwise, someone unfamiliar with the code needs to scan the entire function (~100 LOC).

Scan it for what?

---

The lambdas have been removed in their separate commit, so feel free to rework if you guys insist on them. I'm tired of debating such stuff.

RazvanN7: > Scan for what?

In case some safety issue appears.

kinke: I don't really follow - why would safety issues appear? They could have appeared previously with the function being @safe, and some called function (in the CTFE part, the runtime part doesn't call any function) changing to unsafe. Extracting integral sign, exponent and mantissa from an FP value should obviously be safe; @trusted just slightly simplifies the implementation wrt. the used unsafe reinterpret-casts.

Similar functions operating on the FP bits in [std.math.traits](https://github.com/dlang/phobos/blob/master/std/math/traits.d) (signbit, isIdentical, isInfinity, isSubnormal, ...) are all @trusted.

RazvanN7 (approved):

PR #8129, opened 2021-06-03
[stable] std.format.internal.floats: Restore compilability of unittests for targets with quadruple `real` precision

kinke:
Such as non-Apple AArch64; `printFloat()` doesn't support that format.


thewilsonator (approved):

thewilsonator:
@berni44 heads up

PR #8130, opened 2021-06-03
Fix std.algorithm findSplit methods module documentation

iK4tsu:
Improved the module documentation. Although the method documentation is correct, as it states the correct return type (a sub-type of `Tuple!()`), the same is not applied to the initial module documentation.

Made all tuples of Result have named types. This can improve readability and easiness when writing functional code.

### DIFF
--- a/std/algorithm/searching.d
+++ b/std/algorithm/searching.d
@@ -47,13 +47,13 @@ $(T2 findSkip,
         leaves `a` unchanged, whereas `findSkip(a, "c")` advances `a`
         to `"de"` and returns `true`.)
 $(T2 findSplit,
-        `findSplit("abcdefg", "de")` returns the three ranges `"abc"`,
+        `findSplit("abcdefg", "de")` returns a tuple of three ranges `"abc"`,
         `"de"`, and `"fg"`.)
 $(T2 findSplitAfter,
-        `findSplitAfter("abcdefg", "de")` returns the two ranges
-        `"abcde"` and `"fg"`.)
+`findSplitAfter("abcdefg", "de")` returns a tuple of two ranges `"abcde"`
+        and `"fg"`.)
 $(T2 findSplitBefore,
-        `findSplitBefore("abcdefg", "de")` returns the two ranges `"abc"`
+        `findSplitBefore("abcdefg", "de")` returns a tuple of two ranges `"abc"`
         and `"defg"`.)
 $(T2 minCount,
         `minCount([2, 1, 1, 4, 1])` returns `tuple(1, 3)`.)


iK4tsu:
It seems this is failing because of pointer casting at compile time. I think we should be able to use named tuples at compile, so is this fixable?

iK4tsu:
Reverted the last commit as it depends on modifications to `Tuple`. And I don't know if the changes needed to be done to allow named tuples at compile time can maintain the current behavior we have at the moment. So, this PR only improves documentation.

std/algorithm/searching.d
```diff
@@ -47,14 +47,14 @@ $(T2 findSkip,
         leaves `a` unchanged, whereas `findSkip(a, "c")` advances `a`
         to `"de"` and returns `true`.)
 $(T2 findSplit,
-        `findSplit("abcdefg", "de")` returns the three ranges `"abc"`,
-        `"de"`, and `"fg"`.)
+        `findSplit("abcdefg", "de")` returns a subtype of `Tuple!()` containing
+        three ranges `"abc"`, `"de"`, and `"fg"`.)
```
thewilsonator, line 51: I think just `a tuple of three ranges ...`

std/algorithm/searching.d
```diff
@@ -47,14 +47,14 @@ $(T2 findSkip,
...
 $(T2 findSplitAfter,
-        `findSplitAfter("abcdefg", "de")` returns the two ranges
-        `"abcde"` and `"fg"`.)
+        `findSplitAfter("abcdefg", "de")` returns a subtype of `Tuple!()`
+        containing two ranges `"abcde"` and `"fg"`.)
```
thewilsonator, line 54: ditto

std/algorithm/searching.d
```diff
@@ -47,14 +47,14 @@ $(T2 findSkip,
...
 $(T2 findSplitBefore,
-        `findSplitBefore("abcdefg", "de")` returns the two ranges `"abc"`
-        and `"defg"`.)
+        `findSplitBefore("abcdefg", "de")` returns a subtype of `Tuple!()`
+        containing two ranges `"abc"` and `"defg"`.)
```
thewilsonator, line 57: ditto

std/algorithm/searching.d
```diff
@@ -47,14 +47,14 @@ $(T2 findSkip,
...
 $(T2 minCount,
         `minCount([2, 1, 1, 4, 1])` returns `tuple(1, 3)`.)
```
thewilsonator, line 59: c.f. this

iK4tsu:
I was following the same description of the method documentation. But `tuple` also does the job.

> c.f. this

Also, what is `c.f.`?

thewilsonator:
https://en.wikipedia.org/wiki/Cf.

thewilsonator (approved):

RazvanN7:
I think the tester is stuck. Please rebase this to restart it.

PR #8133, opened 2021-06-04
std.conv: Avoid undefined bytes in toChars() results for radix 10

kinke:
The buffer is initialized from right-to-left in `initialize()`, and unused bytes are left alone. Previously, the whole result wasn't preinitialized with `T.init`, so the unused buffer bytes weren't well-defined. Initialize it now (with zeros), as the whole buffer is still used for equality/identity comparisons etc.


thewilsonator (approved):

thewilsonator:
Does this have a bug report?

kinke:
It now does.

thewilsonator (approved):

PR #8137, opened 2021-06-07
Issue 19842 - only class member functions must be used be used as slots

dukc:
From the bug report:

```d
import std.signals;
import std.stdio;

class Test
{
    mixin Signal!(int);

    void run ()
    {
        emit (1);
    }
}

class Tester
{
    void initialize ()
    {
        Test test = new Test ();
        test.connect ((int a) => trigger (a));
    }

    void trigger (int a)
    {
        writeln (a);
    }
}
```

Because `(int a) => trigger (a)` is not referring to any variable in `initialize`, it's easy to think it's the same as `&this.trigger`, but it's not. I'm attempting to be more clear on this.

### DIFF
--- a/std/signals.d
+++ b/std/signals.d
@@ -37,10 +37,10 @@
  *      $(LINK2 http://www.digitalmars.com/d/archives/16368.html, signals and slots)$(BR)
  *
  * Bugs:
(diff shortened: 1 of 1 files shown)


dukc:
Oops, forgot to target stable. Let's try again.

PR #8138, opened 2021-06-08
Fix std json static assert

atilaneves:


std/math/operations.d
```diff
@@ -1717,7 +1717,7 @@ if (isFloatingPoint!T)
     bool negative;
 }
 
-FloatingPointBitpattern!T extractBitpattern(T)(T val)
+FloatingPointBitpattern!T extractBitpattern(T)(T val) @trusted
```
thewilsonator, line 1720: I still think that this should use `@trused` lambdas but as this is a commit that is from stable, I'll let this go.

thewilsonator (approved):

RazvanN7:
You might want to rebase this as it contains some previously merged commits.

thewilsonator:
IIRC those were to stable

RazvanN7:
My bad. So do we want them in master also?

thewilsonator:
Well we should probably just merge stable instead.

ibuclaw:
2.097 is out, hasn't the merge happened yet?

atilaneves:
The failures seem completed unrelated.

thewilsonator:
I had those errors on a recent dmd PR, try rebasing.

atilaneves:
Errors again. I rebased on upstream/master since I don't really want to deal with this.

PR #8141, opened 2021-06-10
container.array: RangeT: (front|back|opIndex): pure nothrow.

DoctorNoobingstoneIPresume:
Currently, this Pull-Request is more of a Request-for-Comments (or Request-for-Guidance).

`std.container.array`: `RangeT`:

Since `front`, `back` and `opIndex` return a reference, I feel they can be `nothrow` (and, therefore, should be `nothrow`).

BTW, for `nothrow`, is the "can=>should" inference always true, please ?

Also, could they not also be `@pure` (just like `empty` and `length`) ?

BTW, for `@pure`, is the "can=>should" inference always true, please ?

Testing information:

    $ cd std/container/
    $ rdmd -unittest --main array.d
    1 modules passed unittests

Thank you for considering this.

### DIFF
--- a/std/container/array.d
+++ b/std/container/array.d
@@ -114,12 +114,12 @@ private struct RangeT(A)
     }
     alias opDollar = length;
 
-    @property ref inout(E) front() inout
+    @property ref inout(E) front() inout pure nothrow
     {
         assert(!empty, "Attempting to access the front of an empty Array");
         return _outer[_a];
     }
-    @property ref inout(E) back() inout
+    @property ref inout(E) back() inout pure nothrow
     {
         assert(!empty, "Attempting to access the back of an empty Array");
         return _outer[_b - 1];
@@ -167,7 +167,7 @@ private struct RangeT(A)
         }
     }
 
-    ref inout(E) opIndex(size_t i) inout
+    ref inout(E) opIndex(size_t i) inout pure nothrow
     {
         assert(_a + i < _b,
             "Attempting to fetch using an out of bounds index on an Array");


RazvanN7:
Generally, the more we annotate phobos (@safe, pure, @nogc, nothrow), the better.

DoctorNoobingstoneIPresume:
Partial answer:

I had not been aware of the attributes' being infered (for templated functions). Thank you.

RazvanN7:
You're welcome. OK to close this?

DoctorNoobingstoneIPresume:
Yes, this PR can be closed.

If there is time, I would like to ask something extra, for me.

Do you happen to know why inference-of-attributes is done for template functions, but not for regular functions ?

Or is it possible to point me to some documentation, please ? (I have not been able to find this in https://dlang.org/spec/template.html#function-templates or https://dlang.org/spec/function.html.)

Am I correct to think that inference-of-attributes is done individually for each instantiation of the function template ?

Then it would seem to me that it could also be done for regular functions -- but maybe it has been chosen not to do it. I wonder why.

Herringway:
If I understand it correctly, enabling attribute inference for non-templated functions complicates function prototypes.

`void func() { ... } ` and `void func();` would not be guaranteed to match. Worst case scenario is that you'd get a linker error, but that's not a great experience... Templates aren't typically used that way and so this limitation isn't a problem for them.

However, attribute inference IS done for functions with auto returns, since `auto func();` is an error.

RazvanN7:
> However, attribute inference IS done for functions with auto returns, since auto func(); is an error.

`auto` functions are treated as templated functions. The compiler rewrites `auto func() { ... }` to `T func(T)() {...}` behind the scenes.

dkorpel:
> `auto` functions are treated as templated functions. The compiler rewrites `auto func() { ... }` to `T func(T)() {...}` behind the scenes.

Really? From the users point of view, they are not treated as templates.
```D
auto autoFunc() {undefined();} // Error: undefined identifier `undefined`
T tempFunc(T)() {undefined();} // No error

auto autoFunc1() {}

pragma(msg, typeof(autoFunc1)); // pure nothrow @nogc @safe void()
pragma(msg, typeof(tempFunc)); // void (type of uninstantiated template)
```

DoctorNoobingstoneIPresume:
Thank you all for kind explanations !!

If I understand correctly:

(1) Inference needs the body of the function to be present.
(2) For some functions (such as those generated from templates and those with auto return type), the body needs to be present.

Is this correct, please ? (Sorry for abusing your time for my own selfish purposes -- but actually I am happy to do it.)

DoctorNoobingstoneIPresume:
> > `auto` functions are treated as templated functions. The compiler rewrites `auto func() { ... }` to `T func(T)() {...}` behind the scenes.
> 
> Really? From the users point of view, they are not treated as templates.
> 
> ```d
> auto autoFunc() {undefined();} // Error: undefined identifier `undefined`
> T tempFunc(T)() {undefined();} // No error
> 
> auto autoFunc1() {}
> 
> pragma(msg, typeof(autoFunc1)); // pure nothrow @nogc @safe void()
> pragma(msg, typeof(tempFunc)); // void (type of uninstantiated template)
> ```

However, it seems that `typeof (tempFunc ! bool)` gives `pure nothrow @nogc @safe void()` ?

dkorpel:
> Is this correct, please ? 

Yes.

RazvanN7:
@dkorpel Indeed, this was a mistake on my part. The compiler does not rewrite the auto function, however, it eagerly analyzes the body to deduce the return type and it does infer attributes. Since there are these similarities, I forgot that auto functions are not exactly templated functions, but in most cases they do behave like them.

@DoctorNoobingstoneIPresume (1) yes; (2) for templated functions and auto functions, yes, the compiler will error if the body is missing. Also, if you try to generated a .di (d header file), you will see that templates and auto functions gave their bodies outputted also.

DoctorNoobingstoneIPresume:
> Generally, the more we annotate phobos plain **functions** with @safe, pure, @nogc, nothrow the better. However, this PR is annotating templated functions. This is redundant because for templated functions the compiler automatically infers the attributes.
> Although there are situations in phobos where templated functions are manually annotated, I would say that this is an anti-pattern.

For existing situations (within phobos) where templated functions (and maybe functions with `auto` return type ?) are manually annotated:

I wonder what the cause/purpose has been:

(a) Mis-understanding that the attributes are automatically inferred ?

(b) Desire to clarify/document (some of) the attributes for the human readers of the code ?

(c) Desire to make sure that the generated code really obeys the manually specified attributes (and relying on the compiler to signal an error otherwise -- please correct me if I am wrong) ?

(d) The code started out as non-template and was later converted to template ?

Then we might want to think what to about existing and/or new code -- including maybe documenting best practices (if not already done)...

We might want to consider the cost/benefit ratio (e.g.: the following might be bad: git history changed with just some attributes no longer manually specified)...

RazvanN7:
I can't say for sure, but there are definitely cases where templated functions will be @safe/@nogc/nothrow/pure no matter what the instantiation types are. In those cases, the annotations serve as both documentation and safe guards in case someone modifies the code and breaks the contract. This is the only legitimate reason that I can think of.

DoctorNoobingstoneIPresume:
> I can't say for sure, but there are definitely cases where templated functions will be @safe/@nogc/nothrow/pure no matter what the instantiation types are. In those cases, the annotations serve as both documentation and safe guards in case someone modifies the code and breaks the contract. This is the only legitimate reason that I can think of.

Is that not also the (legitimate) case for `front`, `back`, `opIndex` (the `std.container.Array` functions manually annotated with `nothrow` and `pure` in this Pull Request) ? (-:

RazvanN7:
From my experience in phobos, manually annotating templated functions is not a common practice.

@atilaneves What is the policy on manually annotating templated functions?

atilaneves:
> From my experience in phobos, manually annotating templated functions is not a common practice.
> 
> @atilaneves What is the policy on manually annotating templated functions?

Don't, unless there's a very good reason for it. Even when one wants to make sure certain ones apply, it's better to add them to the unittest.

DoctorNoobingstoneIPresume:
> > From my experience in phobos, manually annotating templated functions is not a common practice.
> > @atilaneves What is the policy on manually annotating templated functions?
> 
> Don't, unless there's a very good reason for it. Even when one wants to make sure certain ones apply, it's better to add them to the unittest.

What is the rationale behind the policy (including the "better to add them as unittest"), please ?

DoctorNoobingstoneIPresume:
> @DoctorNoobingstoneIPresume (1) yes; (2) for templated functions and auto functions, yes, the compiler will error if the body is missing. Also, if you try to generated a .di (d header file), you will see that templates and auto functions have their bodies outputted also.

Thank you.

My understanding now is that being able to skip manual annotations is preferred. This ability has been avoided for non-template-non-auto functions because these functions allow declarations which are not definitions.

I am just wondering: Could inferrence be allowed for these functions as well and could the tool which generates .di files from .d files (`dmd -H`) be modified to add the "current" annotations to the function declarations ?

But maybe this is not done because:

(R1) .di files are sometimes generated by other means

and/or

(R2) .di files are sometimes not maintained up-to-date with the latest .d files

and/or

(3) some attributes are sometimes not considered by the authors as part of the specification of the function

(e.g.: a certain function could be marked as `nothrow` with the current implementation, but this is just an accident and the author wishes to reserve the possibility to change the implementation in a `throw`-ing way and it might actually be confusing for users of the function to see it marked as `nothrow`)... ?

atilaneves:
> > > From my experience in phobos, manually annotating templated functions is not a common practice.
> > > @atilaneves What is the policy on manually annotating templated functions?
> > 
> > 
> > Don't, unless there's a very good reason for it. Even when one wants to make sure certain ones apply, it's better to add them to the unittest.
> 
> What is the rationale behind the policy (including the "better to add them as unittest" part), please ?

Because template attributes are inferred and generally one wants that. Some code doesn't care about `@safe` for instance, and maybe depending on the actual types/aliases used they change. `map` is `@safe` if the lambda used is, and isn't if not.

And if a template function doesn't have a corresponding unittest then the codebase has other, bigger, problems.

DoctorNoobingstoneIPresume:
> > > > From my experience in phobos, manually annotating templated functions is not a common practice.
> > > > @atilaneves What is the policy on manually annotating templated functions?
> > > 
> > > 
> > > Don't, unless there's a very good reason for it. Even when one wants to make sure certain ones apply, it's better to add them to the unittest.
> > 
> > 
> > What is the rationale behind the policy (including the "better to add them as unittest" part), please ?
> 
> Because template attributes are inferred and generally one wants that.
> Some code doesn't care about `@safe` for instance, and maybe depending on the actual types/aliases used they change. `map` is `@safe` if the lambda used is, and isn't if not.

I believe that:

- this is not the case for these `front`, `back` and `opIndex` methods (they return references to elements);
- it will be very surprising if the implementation is changed in a way that violates the `nothrow` specification;
- we do want such violations to be detected as soon as possible (best: at compile-time);
- and we want this detection for as many cases as possible (e.g. as many instantiations as possible -- those in unittests and those in as many applications as possible).

> And if a template function doesn't have a corresponding unittest then the codebase has other, bigger, problems.

IMO, this is non-sequitur. I believe that:

- a unittest only tests a limited set of template instantiations (as opposed to the compiler automagically testing this for all the instantiations which are used in all applications -- not just in phobos, but third party, open-source, closed-source etc.);
- nothing prevents a unittest from being added in addition to the explicit specifications;
- testing these specifications in a unittest is more cumbersome than adding the respective keywords on the method;
- the explicit specifications, although redundant to the compiler in some cases (but IMO not in all cases, and not even in the majority of cases, as described in this comment), serve as documentation for the humans that yes, this function and this other functions really should be `nothrow` for all `T`'s out there.

The more I think about this, the more I believe that this PR is useful.

atilaneves:
Explicit annotations are a problem with D, though.

I understand your arguments, but I don't agree with them - a unit test also catches these problems at compile-time for instance.

DoctorNoobingstoneIPresume:
> Explicit annotations are a problem with D, though.

How are they a problem ?

> I understand your arguments, but I don't agree with them - a unit test also catches these problems at compile-time for instance.

As detailed above, IIUC, `unittest's detect such problems:

-for fewer instantiations (`T`s);
- with more effort for human writer and human readers.

Instead, we can (IMO should) simply manually annotate.

And change/document policy too -- based on documented rationale.

RazvanN7:
>How are they a problem, please ?

I am only speculating here, but I think that @atilaneves is referring to the cognitive load of seeing something like `void foo(T)(T) @safe pure @nogc nothrow`. Having them inferred allows for simpler declarations.

DoctorNoobingstoneIPresume:
> > How are they a problem, please ?
> 
> I am only speculating here, but I think that @atilaneves is referring to the cognitive load of seeing something like `void foo(T)(T) @safe pure @nogc nothrow`. Having them inferred allows for simpler declarations.

"Simpler" as in having one word eliminated for each attribute. Instead, a `unittest` is proposed, which is a lot of words for each attribute. => Not at all simpler (IMO).

Plus the `unittest` only tests a few subcases (a few `T`s), and with several words of source code text for *each* subcase. In comparison, the one word per attribute tests all subcases ever, in all code which ever uses this code (not just in Phobos).

Plus the `unittest` is much less connected to the actual function than having the signature modified. => Actually more effort for the human reader and for the human writer and worse self-documentation.

DoctorNoobingstoneIPresume:
IMO the purpose of automagical inferrence of attributes for template functions:

- is not related to programmers' getting away with writing-and-reading fewer words;

- is related to having some template functions being `nothrow` etc. if-and-only-if the corresponding operations (for the specific template arguments) are `nothrow` etc. -- indeed with less effort for the human writer (and the human reader) than, for example, https://en.cppreference.com/w/cpp/algorithm/swap !

@WalterBright Is it possible to explain the purpose of having attributes inferred for template functions (and provide guidelines -- at least for this PR, but more generally for Phobos and all D code), please ?

RazvanN7:
@DoctorNoobingstoneIPresume Look at this issue [1] and its fix [2]. This is why template functions should not be annotated with attributes.

[1] https://issues.dlang.org/show_bug.cgi?id=22101
[2] https://github.com/dlang/phobos/pull/8159

RazvanN7:
I am going to close this as it is not going to be accepted.

DoctorNoobingstoneIPresume:
> @DoctorNoobingstoneIPresume Look at this issue [1] and its fix [2]. This is why template functions should not be annotated with attributes.
> 
> [1] https://issues.dlang.org/show_bug.cgi?id=22101
> [2] #8159

How are those related, please ?

DoctorNoobingstoneIPresume:
> I am going to close this as it is not going to be accepted.

... Says the person not accepting it for obscure reasons.

CyberShadow:
I don't understand why there is this all of this discussion about templated functions, when the functions being modified here in this pull request are not templated (they are merely part of a templated struct, which by itself does not cause their attributes to be inferred).

CyberShadow:
@DoctorNoobingstoneIPresume Would you kindly add a test which demonstrates usage of `Array` in a situation which would be facilitated by the changes here.

aG0aep6G:
> I don't understand why there is this all of this discussion about templated functions, when the functions being modified here in this pull request are not templated (they are merely part of a templated struct, which by itself does not cause their attributes to be inferred).

But attributes are inferred in that scenario. Behold:

```d
struct S()
{
    void f() {}
}
void main() pure nothrow @safe @nogc
{
    S!() s;
    s.f(); /* no error; attributes are inferred */
}
```

CyberShadow:
Ah, thanks, my bad. I made a mistake when testing for this.

DoctorNoobingstoneIPresume:
> @DoctorNoobingstoneIPresume Would you kindly add a test which demonstrates usage of `Array` in a situation which would be facilitated by the changes here.
> 
> Edit: Also if you could please file an issue in the bug tracker describing the problem (e.g. `std.container.Array range methods are not usable in pure/nothrow`) and include the same sample program demonstrating the problem, and then update the commit message to reference this issue.

I believe that the attributes are inferred. (This might not exactly be a function template, but it is a templated entity, in particular -- it is a templated function -- which, despite the terminology, is not a function, but actually something more similar to a function template -- at least as per C++ terminology.)

That is not the issue.

The issue is (for `nothrow`, for example) checking (for all template arguments and in an idiomatic way and affecting the automatically generated documentation and specification) that the body of the templated function is not (accidentally, in the future) changed in such a way that an instance of the templated function can throw an exception.

Let me quote from @andralex' book (2010):

"5.11.2 The `nothrow` Function Attribute

The `nothrow` attribute specified with a function conveys the information that that function will never throw an exception. Just like `pure`, `nothrow` is compile-time-checked.

[...]"

In contrast with a function, for a templated function, with automagical inference of attributes (including the `nothrow` and `pure` attributes), I believe that the last part of this short intro paragraph (i.e. "Just like `pure`, `nothrow` is compile-time-checked [for all template arguments, in case of a templated function -- IIUC].") is missing.

This is natural (inference cannot co-exist with enforcement). But I believe that it is unfortunate.

And I believe we cannot efficiently bring back the checking by adding a `unittest`, because a `unittest` only tests an instance (or a limited set of instances) of the templated function.

It is true that for these three templated functions in particular, with their current two-line bodies which do not involve operations on the template arguments (methinks), a `unittest` does indeed cover all cases.

Still, this has the following problems (IMO):

(1) It still does not ensure that the checking is done for any modifications of the bodies. I am referring to modifications which introduce operations on the template arguments.

(2) A `unittest` is separate from the templated function (okay, still in the same module, D is great with that, but further from the templated function that the signature), it involves more writing than just a keyword per attribute, it involves more reading (that is, if the reader associates the `unittest` with the templated function, which is less probable than for a keyword in the signature) and it does not cause the attribute to appear in the automatically generated documentation (which is also a specification -- e.g. in C++, `std::vector <...>::begin` is specified as being `noexcept`).

(3) (More generally than these particular three two-line templated functions:) I believe (from the comments above and from my current reading of the documentation for functions and templates) that, for templated functions, currently it is not clear what the purpose of the inference of attributes is and whether we should use explicit attributes or not (and in which cases) -- what is the rationale ?

So the deeper purpose of this PR is to establish a common point of view so that we know what to do:

- for existing cases where attributes are explicitly specified (even though inferred);
- more importantly: for cases where we want attributes to be checked;
- and first and foremost: make us aware that we want that checking.

Do we want that checking ? (I believe: Yes, we want it.)

CyberShadow:
> The issue is (for `nothrow`, for example) checking (for all template arguments and in an idiomatic way and affecting the automatically generated documentation and specification) that the body of the templated function is not (accidentally, in the future) changed in such a way that an instance of the templated function can throw an exception.

Makes sense,

> And I believe we cannot efficiently bring back the checking by adding a `unittest`, because a `unittest` only tests an instance (or a limited set of instances) of the templated function.

Technically you can add a `unittest` to the aggregate, which would run if the user's app is built with `-unittest` (then it would test the instantiation with the user's type).

But in practice, it doesn't matter if we have one `unittest` which tests one instantiation or the attribute (or a `unittest` in the aggregate) which tests *every* instantiation. This is because if the goal is to catch regressions in (inferred) attributes, we can't test every instantiation in Phobos anyway. The compiler just doesn't have the capability to look at a template (not template instance) and say that, yep, for every *possible* instantiation of this template, this function inside it will indeed be `nothrow`.

> (2) A `unittest` is separate from the templated function (okay, still in the same module, D is great with that, but further from the templated function that the signature), it involves more writing than just a keyword per attribute, it involves more reading (that is, if the reader associates the `unittest` with the templated function, which is less probable than for a keyword in the signature) and it does not cause the attribute to appear in the automatically generated documentation (which is also a specification -- e.g. in C++, `std::vector <...>::begin` is specified as being `noexcept`).

Makes sense, but to me the difference is not that big to go against a BDFL's decision, personally.

> (3) (More generally than these particular three two-line templated functions:) I believe (from the comments above and from my current reading of the documentation for functions and templates) that, for templated functions, currently it is not clear what the purpose of the inference of attributes is and whether we should use explicit attributes or not (and in which cases) -- what is the rationale ?

Good question and I think the simple answer is simply to reduce the amount of syntactic load, having to write all these function attributes everywhere has been a common criticism of D and Phobos in particular.

DoctorNoobingstoneIPresume:
> > And I believe we cannot efficiently bring back the checking by adding a `unittest`, because a `unittest` only tests an instance (or a limited set of instances) of the templated function.
> 
> Technically you can add a `unittest` to the aggregate, which would run if the user's app is built with `-unittest` (then it would test the instantiation with the user's type).

Partial answer: I did not know that. Thank you. May I ask what that looks like, please ?

> But in practice, it doesn't matter if we have one `unittest` which tests one instantiation or the attribute (or a `unittest` in the aggregate) which tests _every_ instantiation. This is because if the goal is to catch regressions in (inferred) attributes, we can't test every instantiation in Phobos anyway. The compiler just doesn't have the capability to look at a template (not template instance) and say that, yep, for every _possible_ instantiation of this template, this function inside it will indeed be `nothrow`.

But I believe that with the explicit attributes, the checking is going to be done for all instantiations ever used, now and in the future, in Phobos and everywhere else, in release and non-release builds, whether unittests are enabled or not.

And for practical purposes, this is equivalent to the compiler testing the template itself (which I agree cannot be done).

And the price to pay is one keyword per attribute tested per function-or-templated-function.

> > (2) A `unittest` is separate from the templated function (okay, still in the same module, D is great with that, but further from the templated function that the signature), it involves more writing than just a keyword per attribute, it involves more reading (that is, if the reader associates the `unittest` with the templated function, which is less probable than for a keyword in the signature) and it does not cause the attribute to appear in the automatically generated documentation (which is also a specification -- e.g. in C++, `std::vector <...>::begin` is specified as being `noexcept`).
> 
> Makes sense, but to me the difference is not that big to go against a BDFL's decision, personally.


Partial answer: What/who is a BDFL, please ?

> > (3) (More generally than these particular three two-line templated functions:) I believe (from the comments above and from my current reading of the documentation for functions and templates) that, for templated functions, currently it is not clear what the purpose of the inference of attributes is and whether we should use explicit attributes or not (and in which cases) -- what is the rationale ?
> 
> Good question and I think the simple answer is simply to reduce the amount of syntactic load, having to write all these function attributes everywhere has been a common criticism of D and Phobos in particular.

Paetial answer: If we want checking (and documentation/specification), whether for normal functions or for templated functions, we have to write something. At least one word. Luckily: at most one word is possible. IMO.

CyberShadow:
> May I ask what that looks like, please ?

You literally just put the `unittest` block inside the template.

I don't know how well that will work today with how Phobos is linked etc. It's also probably an anti-feature because we want to do all of Phobos testing as part of our CI, and not when the D user happens to build their application with `-unittest`.

> But I believe that with the explicit attributes, the checking is going to be done for all instantiations ever used, now and in the future, in Phobos and everywhere else, in release and non-release builds, whether unittests are enabled or not.

Yes, but, that will happen too late. It will happen when D users try to instantiate the template. So, whether the attribute is there or not, the only difference is what kind of error message they will get. Without the attributes, inference will just move the error up to the point where the template instantiation is used in a context which requires this or that attribute. An attribute will also cause an error even in cases where otherwise the code would have otherwise compiled, which has benefits (causes regressions to be found quicker) and downsides (regressions break more code).

> Partial answer: What/who is a BDFL, please ?

[Benevolent Dictator for Life](https://en.wikipedia.org/wiki/Benevolent_dictator_for_life). For D, AFAIK this is currently Walter Bright and Atila Neves.

DoctorNoobingstoneIPresume:
> IMO the purpose of automagical inferrence of attributes for template functions:
> 
>     * is not related to programmers' getting away with writing-and-reading fewer words;
> 
>     * is related to having some template functions being `nothrow` etc. if-and-only-if the corresponding operations (for the specific template arguments) are `nothrow` etc. -- indeed with less effort for the human writer (and the human reader) than, for example, https://en.cppreference.com/w/cpp/algorithm/swap with its `noexcept (noexcept (conditions_on_the_template_arguments))`.
> 
> 
> IMO the functions modified by this PR should be `nothrow` in all cases (for all template arguments) -- and this should be part of their specification (and part of the automatically generated documentation) -- and this should be checked for all cases -- if possible, with minimum effort for humans (but IMO that effort cannot be zero -- so let us keep it at one-word-per-attribute-per-templated-function, right in the signature where those words can be seen, and let us cover all `T`s => not in a `unittest`).
> 
> @WalterBright, @CyberShadow, @andralex: Is it possible to explain the purpose of having attributes inferred for template functions (and provide guidelines -- at least for this PR, but more generally for Phobos and even all D code), please ?

I admit that another purpose is allowing functions generated from templated functions to be called in checked contexts (i.e. in functions with explicit `nothrow` attribute), even without explicit specification. I have only thought about this now.

But I believe this does not invalidate the other points.

DoctorNoobingstoneIPresume:
> > May I ask what that looks like, please ?
> 
> You literally just put the `unittest` block inside the template.

Thank you !

> I don't know how well that will work today with how Phobos is linked etc. It also probably would be an anti-feature because we want to do all of Phobos testing as part of our CI, and not when the D user happens to build their application with `-unittest`.

I also do not agree with `unittest` solution.

> > But I believe that with the explicit attributes, the checking is going to be done for all instantiations ever used, now and in the future, in Phobos and everywhere else, in release and non-release builds, whether unittests are enabled or not.
> 
> Yes, but, that will happen too late. It will happen when D users try to instantiate the template. So, whether the attribute is there or not, the only difference is what kind of error message they will get. Without the attributes, inference will just move the error up to the point where the template instantiation is used in a context which requires this or that attribute. An attribute will also cause an error even in cases where otherwise the code would have otherwise compiled, which has benefits (causes regressions to be found quicker) and downsides (regressions break more code).

(1) I now better understand and you are right -- there are indeed differences between the compiler checking the templated function itself and the compiler checking all the instantiations -- including at least a difference in the workflow for the dlang community:

IIUC: If the body of the templated function is changed in such a way that breaking `nothrow` only manifests in a third-party instantiation, the PR might be integrated (it passes Github checks), but later ("too late") a report comes from users and the PR has to be reverted and/or a new PR has to be added.

(2) Maybe we can use the following "stairway" of checking:

(level-0) No checks.
(level-1) Checks only when code is used in contexts which are checked.
(level-2) Checks for all instantiations (yes, with not ideal workflow -- bug reports bubbling up from end-users).
(level-3) Checks on the templated function itself (ideal workflow, but unfortunately not possible, I believe).

Next, we might want to:

(action-1) Evaluate the distance between each pair of levels.
(action-2) Determine whether checks for `nothrow` etc. are useful in practice.
(action-3) Determine whether there is a difference between functions and templated functions with regard to whether checks are useful in practice.

A partial attempt to answer (action-2) on my side is another quote from @andralex ' 2010 book. Let me add paragraphs to the one quoted in a previous comment:

"[...] For example:

```
import std.stdio;
nothrow void tryLog(string msg) {
   try {
      stderr.writeln(msg);
   } catch (Exception) {
      // Ignore exception
   }
}
```

The `tryLog` function does a best-effort attempt at logging a message. If an exception is thrown, it is silently ignored. This makes `tryLog` usable in critical sections of code. In some circumstances it would be silly for some important transaction to fail just because a log message couldn't be written. Code with transactional semantics relies critically on certain portions never throwing, and `nothrow` is the way you can ensure such facts statically."

> 
> > Partial answer: What/who is a BDFL, please ?
> 
> [Benevolent Dictator for Life](https://en.wikipedia.org/wiki/Benevolent_dictator_for_life). For D, AFAIK this is currently Walter Bright and Atila Neves.

Thank you.

It is obvious to me that I cannot force much more experienced community members to accept my PR's.

But I believe that for this PR, a clear rationale has not been given (@atilaneves: Please correct me -- e.g. I have replied to your last comment.).

And there has been repeated pressure to close the PR (@RazvanN7), again without clear rationale, when the deeper purpose is that we learn and understand and document best practices for everyone -- and an open PR attracts more opinions. And there should be some time allowed to allow responses (because we cannot also answer with counter-arguments on the same day).

Generally, I believe that rational arguments should be more important than even BDFL-versus-newbie -- that is how projects become better (IMO) and how more contributors are attracted.

CyberShadow:
> And there has been repeated pressure to close the PR

FWIW, there is pressure to close *all* PRs which do not carry an obvious improvement. And if the improvement is there but is not obvious, then it needs to be made obvious first (not just to current participants but to future ones too), using examples / tests / case studies / DIPs / forum discussions / other forms of rationalization.

> And if this PR is wrong, why not remove all explicit attributes which are inferred ?

If the attributes are tested in another way (such as a `unittest`) then that may be desirable.

DoctorNoobingstoneIPresume:
> > And there has been repeated pressure to close the PR
> 
> FWIW, there is pressure to close _all_ PRs which do not carry an obvious improvement.

I understand that this has not been personal.
May I ask why there is such pressure (for all PRs) ?

> And if the improvement is there but is not obvious, then it needs to be made obvious first (not just to current participants but to future ones too), using examples / tests / case studies / DIPs / forum discussions / other forms of rationalization.

I have tried to provide rationalization. Possibly incomplete, but currently I feel that the arguments against explicit attributes are incomplete too -- and there are un-answered questions, e.g. https://github.com/dlang/phobos/pull/8141#issuecomment-875276834. So a reviewer finds a problem with explicit attributes and closes this PR too, but the relationship between that problem and this PR is not clear.

I feel this is important especially since such discussion can serve others too, including in the future.

> > And if this PR is wrong, why not remove all explicit attributes which are inferred ?
> 
> If the attributes are tested in another way (such as a `unittest`) then that may be desirable.

I wish to ask why is `unittest` preferrable/desirable ? Currently, IIUC, the argument is related to criticism of too many annotations in the signature of a function. Is that the only argument, and it outweighs the arguments for checking ? Are the checks important for functions, but not for templated functions ? Or maybe they are not important for functions either ? Or not *that* important ?

Herringway:
The unittest confirms that the attributes were correctly inferred. It's not meant to guarantee that the attributes are always present on every instantiation, and it shouldn't be. There would be nothing gained from that except arbitrarily causing code to fail compilation.

For example...
```D
void f(alias func)() nothrow {
  func();
}

void throwingFunction() { throw new Exception("Oopsie woopsie!"); }
void nonThrowingFunction() nothrow {}

//alias t1 = f!throwingFunction; //Cannot instantiate, breaks nothrow promise
alias t2 = f!nonThrowingFunction;
```
Explicitly annotating templates only limits how they can be used. They can already be inferred nothrow, pure, etc, so there is nothing to be gained from explicit annotation.

All that needs to be checked is that the template itself does not introduce impure, throwing, etc behaviour, and a unittest or two satisfies that requirement just fine.

Simple functions don't have inferrence, but they also don't have multiple instantiations to worry about. So explicit annotations don't have the same downsides as they do on templates.

DoctorNoobingstoneIPresume:
> The unittest confirms that the attributes were correctly inferred. It's not meant to guarantee that the attributes are always present on every instantiation, and it shouldn't be. There would be nothing gained from that except arbitrarily causing code to fail compilation.

I believe that for these templated functions, it should be, i.e. we should make a documented-and-checked requirement that they never throw.

I would not say "arbitrarily cause compilation to fail", but rather "righteously cause compilation to fail, thankfully".

> For example...
> 
> ```d
> void f(alias func)() nothrow {
>   func();
> }
> 
> void throwingFunction() { throw new Exception("Oopsie woopsie!"); }
> void nonThrowingFunction() nothrow {}
> 
> //alias t1 = f!throwingFunction; //Cannot instantiate, breaks nothrow promise
> alias t2 = f!nonThrowingFunction;
> ```
> 
> Explicitly annotating templates only limits how they can be used. They can already be inferred nothrow, pure, etc, so there is nothing to be gained from explicit annotation.

They can be inferred with the current implementation, but they are not checked. The checking is what is gained. Plus it becomes part of the automatically generated documentation that they cannot throw.

> All that needs to be checked is that the template itself does not introduce impure, throwing, etc behaviour, and a unittest or two satisfies that requirement just fine.

IMO, adding the explicit attributes is what satisfies that requirement just fine. Advantages over unittest: (1) Single keyword with meaning understood by all who read. (2) Right there in the signature, not few/several/many lines above/below. (3) Automatically-generated documentation => Also, the attributes become part of the specification. (4) Checked by compiler even in applications which are not built with unittests enable. (5) Checked for all instantiations -- but, IIUC, as per @CyberShadow's reply above, this might also be achievable with an aggregate unittest (a unittest placed inside the template).

> Simple functions don't have inferrence, but they also don't have multiple instantiations to worry about. So explicit annotations don't have the same downsides as they do on templates.

I wish to ask: What are the downsides of explicit annotations (a) on normal functions; (b) on templated functions, please ?

DoctorNoobingstoneIPresume:
> @DoctorNoobingstoneIPresume Would you kindly add a test which demonstrates usage of `Array` in a situation which would be facilitated by the changes here.
> 
> Edit: Also if you could please file an issue in the bug tracker describing the problem (e.g. `std.container.Array range methods are not usable in pure/nothrow`) and include the same sample program demonstrating the problem, and then update the commit message to reference this issue.

Let us consider an application programmer who wishes to use an instantiation of one of these templated functions in a context where the no-fail guarantee is checked-and/or-important, e.g.:

* in a `nothrow` function;
* in a function which is called by a `nothrow` function;
* in any context where the "no-fail" error-safety guarantee is important (such as "code with transactional semantics" and/or code with other, weaker error-safety guarantees -- as those "[rely] critically on certain portions never throwing [...]") -- whether that context uses the `nothrow` annotation or not.

IMO, the problem is:

If we do not document/specify that the library function never fails,
the application programmer cannot reliably use it in his/her critical piece of code,
because he/she cannot rely on our never changing the library function in such a way that it actually can fail.

And if we do change the library function in a way in which it can fail,
but there is no early reporting of this change (e.g. compile error),
the problem is going to be detected (significantly) later and is going to be more difficult to debug.

PR #8143, opened 2021-06-14
std.container.Array: Simplify implementation of length (by calling reserve).

DoctorNoobingstoneIPresume:
Currently, the implementations of `length` and `reserve` have many lines in common.

For the case of `newLength > length`, many lines can be replaced with `reserve (newLength); _payload = _payload.ptr [0 .. newLength];`.

=> Easier to understand. Easier to re-use future improvements.

Testing information:

```
$ cd std/container/
$ rdmd -unittest --main array.d
1 modules passed unittests
```

### DIFF
--- a/std/container/array.d
+++ b/std/container/array.d
@@ -322,47 +322,9 @@ if (!is(immutable T == immutable bool))
                 return;
             }
             immutable startEmplace = length;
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):
Very nice!

PR #8144, opened 2021-06-16
Fix Issue 20393 - formattedRead accepts input, that should be rejected

taitaisama:
Function readUpToNextSpec in std/format/spec.d returned false when it encountered some specific types of errors and threw an exception on other types, however this wasn't taken into account in formattedRead in which it was assumed that readUpToNextSpec always threw exception. The change makes it so that readUpToNextSpec throws an exception instead of returning false.

### DIFF
--- a/std/format/read.d
+++ b/std/format/read.d
@@ -710,3 +710,11 @@ T unformatValue(T, Range, Char)(ref Range input, scope const ref FormatSpec!Char
     assert(result[0] == 'a');
 }
 
+@safe pure unittest
+{
+    import std.exception : assertThrown;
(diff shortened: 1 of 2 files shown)


std/format/read.d
```diff
@@ -710,3 +710,11 @@ T unformatValue(T, Range, Char)(ref Range input, scope const ref FormatSpec!Char
     assert(result[0] == 'a');
 }
 
+@safe pure unittest
```
thewilsonator, line 713: Please include a bugzilla link like the unit tests above.

thewilsonator (approved):

PR #8145, opened 2021-06-18
conv: add the ability to convert structs to associative arrays

ljmf00:

---

This will allow people to easily extract struct fields to an associative array. This is very common when you have a struct for configs and want to parse it to easily match some fields.

```d
struct Foo {
    int a = 5;
    float b = 0.5;
	string c = "foobar";
}

Foo a;
assert(["a": "5", "b": "0.5", "c": "foobar"] == a.to!(string[string]));
```

### DIFF
--- a/std/conv.d
+++ b/std/conv.d
@@ -1697,6 +1697,77 @@ if (!isImplicitlyConvertible!(S, T) &&
 }
 
 /**
+Struct to Associative array conversion converts each field and it's corresponding
+value.
+*/
+private T toImpl(T, S)(S value)
+if (is(S == struct) && isAssociativeArray!T &&
+    is(KeyType!T == string) && !is(typeof(value.opCast!T()) : T))
+{
+    cast(void) value; // mark as used, although used by the mixin(sym)
+
+    import std.typecons : Tuple, tuple;
+    import std.array : assocArray;
+
+    Tuple!(string, ValueType!T)[] tarr;
+    foreach (m; __traits(allMembers, S))
+    {
+        static immutable sym = "value." ~ m;
+        enum isFunctionOrTemplate = isFunction!(mixin(sym)) || __traits(isTemplate, mixin(sym));
+
+        static if ((isFunction!(mixin(sym)) &&
+                    functionAttributes!(mixin(sym)) & FunctionAttribute.property)
+                  || !isFunctionOrTemplate)
+        {
+            tarr ~= tuple(m, mixin(sym).to!(ValueType!T));
+        }
+    }
+
+    return tarr.assocArray;
+}
+
+///
+@safe pure nothrow unittest
+{
+    struct Foo {
+        int a = 5;
+        int b;
+    }
+
+    Foo a;
+    assert(["a": 5, "b": 0] == a.to!(int[string]));
+    assert(["a": "5", "b": "0"] == a.to!(string[string]));
+}
+
+// test with empty struct
+@safe pure nothrow unittest
+{
+    struct Foo {}
+
+    Foo a;
+    assert((int[string]).init == a.to!(int[string]));
+}
+
+// test with a struct with opCast to string[string]
+@safe pure nothrow unittest
+{
+    struct Foo {
+        string[string] opCast(T)() const
+            if (is(T == string[string]))
+        {
+            return (string[string]).init;
+        }
+
+        int a;
+        int b;
+    }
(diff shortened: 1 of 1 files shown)


thewilsonator:
cc @atilaneves

n8sh:
Rather than converting values to strings how about `sumtype`?

atilaneves:
> This is very common when you have a struct for configs and want to parse it to easily match some fields.

Wouldn't it be easier to use something like JSON there? I don't understand the use-case at all, or the motivation for including it in Phobos.

And even then, there's not nearly enough tests to capture what happens, say, here:

```d
struct Oops {
    int i;
    string s;
}

Oops().to!(int[string]);
```

ljmf00:
> > This is very common when you have a struct for configs and want to parse it to easily match some fields.
> 
> Wouldn't it be easier to use something like JSON there? I don't understand the use-case at all, or the motivation for including it in Phobos.

Well no. This is a different behaviour apart from what you can get with JSON. 
Also, there's currently no serialization support for JSON in the standard library.

> And even then, there's not nearly enough tests to capture what happens, say, here:
> 
> ```d
> struct Oops {
>     int i;
>     string s;
> }
> 
> Oops().to!(int[string]);
> ```

I missed that. This is some kind of a draft. I can make this code generation more efficient tho. I'll allocate more time to my PRs soon. Sorry for that.

Although that's not a Oops, that's a maybe oops. The point here would be to do the same conversions as implemented in `to!()` but with the struct fields. This is useful for fields with the same type and any type that has a matching `to!()` including non-primitive types.

For this struct, for example, its possible to do the following:
```d
struct Foo {
    string a;
    int b;
    float c;
}
```

```d
to!(string[string]).writeln;
to!(int[string]).writeln;
to!(float[string]).writeln;
```

It is possible to do the conversion if the `string` `a` is `"123"` or `"123.3"`, for example.

ljmf00:
I also need to adjust the priorities of this for structs implementing an `opCast`

ljmf00:
> Rather than converting values to strings how about `sumtype`?

I don't see `sumtype` particularly useful here. Can you elaborate?

RazvanN7:
@atilaneves Do you have a final decision on whether this should or shouldn't be included in phobos?

atilaneves:
> @atilaneves Do you have a final decision on whether this should or shouldn't be included in phobos?

Unless there's more motivation that I'm missing, I'd say no.

RazvanN7:
@ljmf00 I'm closing this until further motivation is provided. We can re-open it once @atilaneves gives his acceptance for this PR.

ljmf00:
> > @atilaneves Do you have a final decision on whether this should or shouldn't be included in phobos?
> 
> Unless there's more motivation that I'm missing, I'd say no.

Can you elaborate on your decision and the drawbacks that this can possibly have? I saw some cases where this is useful for high-level parsing and conversion. I stated them above and rewrote my PR description to explain my use case better.

> Rather than converting values to strings how about `sumtype`?

You upvoted here. Can you and @n8sh elaborate on how `sumtype` is useful here? Sumtype is used in type selection at compile-time, not at runtime.

atilaneves:
The drawback is that more code is a bad thing, always, and has to be justified with a reasonable trade-off. In other words: the default answer to "can I add this code?" should be "no".

I don't think this is a general enough addition to add to Phobos. I didn't find the explanations for why this is useful compelling.

ljmf00:
> The drawback is that more code is a bad thing, always, and has to be justified with a reasonable trade-off.

The trade-off here is making this logic completely manual. At least I would like to discuss a way of struct unpacking to a named tuple and convert a tuple to a dictionary. If there is a simple way to do this with the current tools of the standard library, I would accept it as a solution.

atilaneves:
> > The drawback is that more code is a bad thing, always, and has to be justified with a reasonable trade-off.
> 
> The trade-off here is making this logic completely manual.

Given that the user can write this code as part of their project or as a library that they put up on code.dlang.org, I don't understand how.

> At least I would like to discuss a way of struct unpacking to a named tuple and convert a tuple to a dictionary. If there is a simple way to do this with the current tools of the standard library, I would accept it as a solution.

I don't understand why it's necessary to be in the standard library. I don't understand why anyone would want to do this at all, but that's beside the point of this PR. Not everything belongs in the standard library, and that's fine.

PR #8147, opened 2021-06-21
[Reg 2.074] std.traits.isFloatingPoint, isNumeric, isUnsigned, and isSigned should never be true for SIMD vectors, imaginary/complex numbers, or enums with character base types

n8sh:
Should this PR target master instead of stable so if any code relies on this wrong behavior the breakage will happen in the next major release instead of the next minor release?

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -6333,7 +6333,7 @@ enum bool isIntegral(T) = __traits(isIntegral, T) && is(IntegralTypeOf!T);
 /**
  * Detect whether `T` is a built-in floating point type.
  */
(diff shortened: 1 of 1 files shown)


n8sh:
>Windows x64
>std\traits.d(6582): Error: integer constant expression expected instead of `cast(__vector(uint[4]))[0u, 0u, 0u, 0u]`

How odd. On my windows machine DMD 2.097 makes no complaint when compiling `enum EUInt4 : __vector(uint[4]) { a = __vector(uint[4]).init, }`.

thewilsonator (approved):

RazvanN7 (approved):

PR #8150, opened 2021-06-22
std.container.array: compiler allows Array as key of associative array but this does not work properly

n8sh:


n8sh:
I thought when a struct defines `opEquals` but not `toHash` the compiler raises an error/warning if that type is used as the key of an associative array.

RazvanN7:
@n8sh There's this filed bug report [1] which relates to alias this. That's a regression introduced in 2.068. Taking the code snippet in [2] and running it against all dmd compilers shows that before 2.068 the code failed to compile with "Error: AA key type Array should have 'size_t toHash() const nothrow @safe' if opEquals defined". I bet that the PR that introduced the regression in case of [1] is also the culprit for [2]. I think that [2] can be closed as a duplicate of [1].

I bet this PR is the culprit: https://github.com/dlang/dmd/pull/5001

[1] https://issues.dlang.org/show_bug.cgi?id=22075
[2] https://issues.dlang.org/show_bug.cgi?id=22065

RazvanN7:
Since https://issues.dlang.org/show_bug.cgi?id=22065 was closed, does this mean we can close this?

PR #8151, opened 2021-06-22
std.container.Array: insertBack: Extra checks (including addition overflow).

DoctorNoobingstoneIPresume:


thewilsonator:
Why is this necessary? 

Does this solve an issue? If so, please link to one.
Does this provide better diagnostics? If so, please add unittests that test for it (using `assertThrown!AssertError(testThatShouldFail())`

RazvanN7:

DoctorNoobingstoneIPresume:
> Why is this necessary?
> 
> Does this solve an issue? If so, please link to one.
> Does this provide better diagnostics? If so, please add unittests that test for it (using `assertThrown!AssertError(testThatShouldFail())`

First I wish to ask:

- Why are the pre-existing `assert`s necessary ? Do they provide better diagnostics ? Do they have `unittest`s ?

IMO the answer is negative.

But they are useful anyway for clarification of code: "If we have got to this point, 

> Why is this necessary?
> 
> Does this solve an issue? If so, please link to one.
> Does this provide better diagnostics? If so, please add unittests that test for it (using `assertThrown!AssertError(testThatShouldFail())`

First, let me ask back the same exact questions for the already-existing checks, e.g.:

- `assert`s modified by this PR;

- checks for `"Overflow"` in other methods.

RazvanN7:
Each PR should be accompanied by a rationale/justification. It is the burden of the contributor to demonstrate the utility of the PR.

> First, let me ask back the same exact questions for the already-existing checks, e.g.:

Already existing checks have their own circumstances that can be judged on a case by case basis, but that does not
represent a reason to accept new additions with no apparent benefit. It might be that some existing checks are well
justified whereas others have been accepted due to improper review. Such precedents do not represent a reason
to accept code without a rationale.

std/container/array.d
```diff
@@ -385,9 +385,17 @@ if (!is(immutable T == immutable bool))
...
             {
-                reserve(1 + capacity * 3 / 2);
+                import core.checkedint : addu;
+
+                bool overflow;
+                immutable size_t newCapacity = addu(capacity, capacity / 2 + 1, overflow);
+                if (overflow)
+                    assert(false, "Overflow");
+
+                reserve(newCapacity);
```
n8sh, line 398: Single-element `insertBack` prior to this PR:

https://github.com/dlang/phobos/blob/61c00065a8c7efdf21efb92ce69d7a9de4904ec9/std/container/array.d#L384-L397

If there is an overflow on line 390 `reserve` will be called with an argument smaller than the current `capacity`. After this the assertion on line 392 will fail with the message `""Failed to reserve memory"` because the call did not increase `capacity`. So overflow is already being detected.

DoctorNoobingstoneIPresume: In `array.d`, other checks for "Overflow" (which are less serious than this one -- checks after addition, compared to this one which is check after scaling) are enabled even in `-release` mode.

The "Failed to reserve memory" check is not.

DoctorNoobingstoneIPresume:
> Each PR should be accompanied by a rationale/justification. It is the burden of the contributor to demonstrate the utility of the PR.
> 
> > First, let me ask back the same exact questions for the already-existing checks, e.g.:
> 
> Already existing checks have their own circumstances that can be judged on a case by case basis, but that does not
> represent a reason to accept new additions with no apparent benefit. It might be that some existing checks are well
> justified whereas others have been accepted due to improper review. Such precedents do not represent a reason
> to accept code without a rationale.

Are we discussing (1) the particular already-existing checks which I have specifically indicated in my previous comment (i.e. checks modified by this PR and checks similar to this PR) ?

Or (2) generalities (which, IMO, are vague) ?

RazvanN7:
I do not know whether other checks are founded or not and that is not relevant for this discussion. My point is that this PR needs to prove its utility by providing a rationale or a case where the added checks are relevant, otherwise, the argument of consistency with other methods does not apply. Each method has been added by their author with a specific purpose in mind; if the author decided to put asserts, he/she may have had a good reason for that and it is possible that in this same scenario the same person may have deliberately decided to elide the checks. We do not know what the reasoning was, therefore, adding such checks needs a justification, one that you are not offering right now. Simply adding checks for the sake of consistency is not a good reason.

If you want to get this merged, please provide a rationale or a small code snippet (preferably link to a bugzilla issue) that proves the utility of this change.

DoctorNoobingstoneIPresume:
> I do not know whether other checks are founded or not and that is not relevant for this discussion.
My point is that this PR needs to prove its utility by providing a rationale or a case where the added checks are relevant, otherwise, the argument of consistency with other methods does not apply. Each method has been added by their author with a specific purpose in mind; if the author decided to put asserts, he/she may have had a good reason for that and it is possible that in this same scenario the same person may have deliberately decided to elide the checks. We do not know what the reasoning was, therefore, adding such checks needs a justification, one that you are not offering right now. Simply adding checks for the sake of consistency is not a good reason.
> 
> If you want to get this merged, please provide a rationale or a small code snippet (preferably link to a bugzilla issue) that proves the utility of this change.

But of course they are relevant -- consistency is important.

Also, how can you reason about a piece of code and not reason about a very similar piece of code ? Just because one has not yet been integrated, and another one has been ?

The authors might not be here to ask, or it might not be efficient to ask them, but we are programmers too and we can reason about their code and try to find out what has been in their minds (advantages and disadvantages of their approaches in the situations we analyze).

I believe that consistency is especially relevant for the Standard Library of a programming language, which should provide examples of best practices.

How can we check for overflow in a place and not in another place, especially one where a scaling operation is performed ?

How can I test that with a `unittest` ? I would need a 32-bit system (cause for 64-bit x86 systems, virtual addresses are currently limited to 48 bits, and anyway I cannot put that much RAM in a system) -- one in which an Array consumes 2.66 GiB -- so I need /LARGEADDRESSAWARE and maybe other things. The effort is disproportionate. But we can easily imagine the overflow even without such a test.

Why are we checking that the new `length` is equal to the old `length` plus the `length` of the inserted `Range` only when the inserted `Range` has a `length` method -- especially in code which iterates one-by-one through the inserted `Range` (and therefore can compute the length of the inserted `Range` without extra costs with regard to the asymptotic performance) ?

I find the requests imposed ("Please prove the utility, but do not ask us to judge similar code or to provide rationale.") just a way to favour reviewers over coders.

rich-de (approved):

RazvanN7 (approved):

DoctorNoobingstoneIPresume:
@rich-de, @RazvanN7: I appreciate the support.  >:D<

PR #8152, opened 2021-06-30
[stable] [Reg 2.097] std.typecons.RefCounted!T for struct T without an explicit toString or alias this previously could be converted to string even when uninitialized but now cannot

n8sh:


thewilsonator (approved):

std/typecons.d
```diff
@@ -6773,16 +6782,23 @@ pure @system unittest
...
     assert(to!string(cast(const) r) == to!string(cast(const) a));
     // Check that string conversion is still transparent for refcounted
     // structs that have alias this.
-    struct B { int b; alias b this; }
-    struct C { B b; alias b this; }
+    static struct B { int b; alias b this; }
+    static struct C { B b; alias b this; }
     assert(to!string(refCounted(C(B(123)))) == to!string(C(B(123))));
+    // https://issues.dlang.org/show_bug.cgi?id=22093
+    // Check that uninitialized refcounted structs that previously could be
```
Geod24, line 6796: Technically it should be "default-initialized", not uninitialized.

n8sh: I could have phrased that better. I meant "not initialized" to be understood as `RefCounted.refCountedStore.isInitialized == false`.

PR #8153, opened 2021-06-30
Fixed typos "iff" => "if"

sinisa-susnjar:


joakim-brannstrom:
Isn't that the term "if and only if" used in logic and math written as iff?

sinisa-susnjar:
Makes sense! Sorry for stirring things up!
Should I close it?

wilzbach:
Yup, I think iff makes a lot of sense.

That being said I actually decided to reopen it as it's not super common term and maybe shouldn't be used in the standard library documentation.

CyberShadow:
I think using something like [the `<abbr>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr) would be appropriate.

RazvanN7:
I know that @andralex usually uses this form.

andralex:
Folks, "iff" is a [word in the dictionary](https://www.merriam-webster.com/dictionary/iff). Wherever used, it adds precision to the documentation.

ljmf00:
> Folks, "iff" is a [word in the dictionary](https://www.merriam-webster.com/dictionary/iff). Wherever used, it adds precision to the documentation.

I don't agree with your argument. "Folks" is also a word present in the dictionary and it is not appropriate to be written in a standard specification or in scientific writing. The same way as `iff`. Some scientific journals consider it a jargon word and very subject-specific. See [1], [2], [3] and [4].

It is also a good thing to discuss what @wilzbach and @CyberShadow suggested rather than closing this PR right away :)

[1]: https://english.stackexchange.com/questions/428961/is-it-proper-to-use-iff-to-denote-if-and-only-if-in-academic-writing-papers
[2]: https://newbedev.com/is-using-iff-appropriate
[3]: https://english.stackexchange.com/questions/62067/is-iff-considered-a-real-word-or-just-an-abbreviation
[4]: https://www.wyzant.com/resources/answers/612993/is-it-proper-to-use-iff-to-denote-if-and-only-if-in-academic-writing-papers

PR #8154, opened 2021-07-01
Issue 21488 - Always compile phobos library with -fPIC on POSIX targets

ibuclaw:
Segmentation faults or linker errors occur on 32-bit PIE platforms if any object is not compiled with `-fPIC`.


ibuclaw:
Ping @MartinNowak don't know if you need to fix your release build scripts.

thewilsonator (approved):

PR #8155, opened 2021-07-03
std: array: refactor std.array unittest blocks and attribute annotation

ljmf00:

---

Read the patches description for context.

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -4123,8 +4123,8 @@ unittest
 // https://issues.dlang.org/show_bug.cgi?id=10690
 @safe pure nothrow unittest
 {
-    import std.algorithm;
-    import std.typecons;
+    import std.algorithm.iteration : filter;
+    import std.typecons : tuple;
     [tuple(1)].filter!(t => true).array; // No error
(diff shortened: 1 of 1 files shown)


Geod24 (approved):
LGTM, just two nits on the second commit:
- "seperate" => "separate"
- Moving one of the block makes the diff non-obvious. I thought at first the code was changed, but ignoring whitespace changes made it cleared;

Also, `std: array` => `std.array` ?

RazvanN7 (approved):

ljmf00:
> LGTM, just two nits on the second commit:
> 
> * "seperate" => "separate"
> * Moving one of the block makes the diff non-obvious. I thought at first the code was changed, but ignoring whitespace changes made it cleared;

Fixed. Yes, moving these blocks make diffs very hard to read, unfortunately.

> Also, `std: array` => `std.array` ?

Didn't know that, thanks for the clarification. I'll start committing that way. This could be added to the contribution guidelines, or maybe it is there somewhere and I missed it :upside_down_face:

PR #8156, opened 2021-07-04
cirrus-ci: bump OS CI versions to latest and keep last LTS supported

ljmf00:

---
Read the patch notes.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -52,14 +52,14 @@ task:
 
 # FreeBSD
 task:
-  name: FreeBSD 12.1 x64
+  name: FreeBSD 13.0 x64
   freebsd_instance:
-    image_family: freebsd-12-1
+    image_family: freebsd-13-0
     cpu: 4
     memory: 8G
   timeout_in: 60m
   environment:
     OS_NAME: freebsd
-    CI_DFLAGS: -version=TARGET_FREEBSD12
+    CI_DFLAGS: -version=TARGET_FREEBSD13
   install_bash_and_git_script: pkg install -y bash git
   << : *COMMON_STEPS_TEMPLATE
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -40,6 +40,17 @@ task:
 
 # Mac
 task:
+  name: macOS 11 x64
+  osx_instance:
+    image: big-sur-xcode
+  timeout_in: 60m
+  environment:
+    OS_NAME: darwin
+    # override Cirrus default OS (`darwin`)
+    OS: osx
+  << : *COMMON_STEPS_TEMPLATE
+
+task:
   name: macOS 10.15 x64
   osx_instance:
     image: catalina-xcode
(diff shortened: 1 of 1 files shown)


Geod24:
What's the rationale ? We usually tend to stick with the oldest supported release as it's more likely to catch issues.

RazvanN7:
> We usually tend to stick with the oldest supported release as it's more likely to catch issues.

What do you mean? Shouldn't we support the newest version?

kinke:
> Shouldn't we support the newest version?

No problem with adding a separate 13 job, but 12.x is still supported for a while (they support the 3 latest major versions). 12.1 can be bumped to 12.2 though.

ljmf00:
> > Shouldn't we support the newest version?
> 
> No problem with adding a separate 13 job, but 12.x is still supported for a while (they support the 3 latest major versions). 12.1 can be bumped to 12.2 though.

Yeah, sure, it makes sense to support both versions. I'll do that :)

ljmf00:
Unfortunately, there is interdependency with the DMD CI script. This should be avoided, IMHO. Ideally, this should either only build DMD or grab the prebuilt version. Maybe grabbing the prebuilt version could lead to some inconsistencies if the latest build fails to publish.

Is there any strong reason to run the DMD test suite in the standard library too?

kinke:
> Is there any strong reason to run the DMD test suite in the standard library too?

Probably not really for Phobos, but definitely for druntime. It doesn't run for long, so skipping it seems hardly worth it.

Wrt. the annoying FreeBSD 12.2 `diff` regression, see https://github.com/ldc-developers/dmd-testsuite/commit/3ffa31e655cf446b65fb23341e028e9743d171ce.

ljmf00:
> > Is there any strong reason to run the DMD test suite in the standard library too?
> 
> Probably not really for Phobos, but definitely for druntime. It doesn't run for long, so skipping it seems hardly worth it.

Yes, druntime makes a lot of sense, because a lot of tests there implicitly depends on `object.d` definitions and obviously compiler hooks, but DMD tests shouldn't depend on Phobos at all and if they do, I would say that is a bad practice.

> Wrt. the annoying FreeBSD 12.2 `diff` regression, see [ldc-developers/dmd-testsuite@3ffa31e](https://github.com/ldc-developers/dmd-testsuite/commit/3ffa31e655cf446b65fb23341e028e9743d171ce).

Oof. Thanks.

kinke:
> Yes, druntime makes a lot of sense, because a lot of tests there implicitly depends on object.d definitions and obviously compiler hooks

Exactly, and not only, e.g., there are also mangling tests which depend on `core.stdc.*` definitions.

> but DMD tests shouldn't depend on Phobos at all and if they do, I would say that is a bad practice.

Exactly, and that's why that's actually 'disallowed' - but it all depends on reviewers noticing, so I cannot rule out there are a few which import Phobos stuff (e.g., some `^^` usage, which lowers to `std.math.pow` - not sure if that import is implicit or not).

ljmf00:
> > Yes, druntime makes a lot of sense, because a lot of tests there implicitly depends on object.d definitions and obviously compiler hooks
> 
> Exactly, and not only, e.g., there are also mangling tests which depend on `core.stdc.*` definitions.
> 
> > but DMD tests shouldn't depend on Phobos at all and if they do, I would say that is a bad practice.
> 
> Exactly, and that's why that's actually 'disallowed' - but it all depends on reviewers noticing, so I cannot rule out there are a few which import Phobos stuff (e.g., some `^^` usage, which lowers to `std.math.pow` - not sure if that import is implicit or not).

Ok.

Either way, I can propose a change later to solve this interdependency. I'm now drafting a change in the DMD side https://github.com/dlang/dmd/pull/12822 .

RazvanN7:
@ljmf00 What is the course of action over here?

ibuclaw:
You're going to have to bump the bootstrap DMD version to 2.097.1 or later, so this will have to be parked until at least 2.098.0.  Even then, either until a time that we'll be forced off the ancient CI pipelines, or around the 2.108.0 release.  Whichever happens first.

PR #8162, opened 2021-07-09
std.container.Array: Payload "default constructor" simplifies Array.

DoctorNoobingstoneIPresume:
The purpose is reduce code duplication.
Thank you for considering this.


std/container/array.d
```diff
@@ -558,34 +550,8 @@ if (!is(immutable T == immutable bool))
...
-                GC.addRange(p, sz);
-            }
-            _data = Data(cast(T[]) p[0 .. 0]);
-            _data._capacity = elements;
-        }
-        else
-        {
-            _data.reserve(elements);
-        }
+            _data = Data(DefaultConstruct.value);
```
dkorpel, line 553:
```suggestion
            _data = Data.init;
```

std/container/array.d
```diff
@@ -428,32 +442,10 @@ if (!is(immutable T == immutable bool))
...
-        static if (hasIndirections!T)
-        {
-            if (p)
-                GC.addRange(p, T.sizeof * values.length);
-        }
-        _data = Data(p[0 .. values.length]);
+        _data = Data(DefaultConstruct.value);
+        _data.reserve(values.length);
+        foreach (ref value; values)
+            _data.insertBack(value);
```
dkorpel, line 448: `insertBack` has some overhead checking for capacity each time. Since you reserved the right length, you can just `emplace` each element directly like the original code does.

dkorpel (commented):
Love seeing a red diff! But maybe someone more familiar with `Array` than me should check whether the duplicated code didn't have subtle differences that were important e.g. for performance.

std/container/array.d
DoctorNoobingstoneIPresume: Thank you. I have updated accordingly.

std/container/array.d
```diff
@@ -428,32 +442,10 @@ if (!is(immutable T == immutable bool))
...
-        {
-            emplace(p + i, e);
-        }
-        static if (hasIndirections!T)
-        {
-            if (p)
-                GC.addRange(p, T.sizeof * values.length);
-        }
-        _data = Data(p[0 .. values.length]);
+        _data = Data(DefaultConstruct.value);
```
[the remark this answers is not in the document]
DoctorNoobingstoneIPresume: I would love this to work. We would have an idiom for a `struct` having some kind of a default constructor even in the presence of other constructors.

Unfortunately, this results in:

```
core.exception.AssertError@C:\Program Files (x86)\Dev\DigitalMars\dmd.2.096.1.windows\windows\bin\..\..\src\phobos\std\typecons.d(6631): Attempted to access an uninitialized payload.
----------------
0x00421CE5
0x00408BA0
0x00402323
0x00421ABD
0x0042A8BB
0x00425870
0x00423677
0x00421C43
0x7D4E7CC2 in FormatMessageW
1/1 modules FAILED unittests
C:\Program Files (x86)\Dev\DigitalMars\dmd.2.096.1.windows\windows\bin\..\..\src\phobos\std\typecons.d(6631): [unittest] Attempted to access an uninitialized payload.
```

std/container/array.d
```diff
@@ -24,6 +24,17 @@ import std.traits;
 
 public import std.container.util;
 
+pure @system unittest
+{
+    immutable(size_t) n = 100;
```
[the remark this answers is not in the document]
DoctorNoobingstoneIPresume: I have changed to `enum size_t n = 100;`. I feel that the `size_t` is helpful for me, as a hint that `n` holds a "number of elements". I hope that is okay.

std/container/array.d
```diff
@@ -24,6 +24,17 @@ import std.traits;
...
+    float[] a0; a0.length = n; a0[] = 1.41421f;
+    auto a1 = Array!float(a0);
+    assert(a1.length == n);
+    assert(a1.capacity == n); // Checks that a1's constructor has called reserve.
+    import std.algorithm.comparison;
```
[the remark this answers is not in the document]
DoctorNoobingstoneIPresume: Okay, this is easy to change and I have changed it.

I am just wondering whether this is somehow redundant: We see `equal` twice, once in the `import` declaration and once in the actual call to `equal`.

I am adding that this is a small scope. But even if it were a larger scope, we could use braces in order to reduce the effect of the `import` declaration to a smaller scope => IMO annihilate bad effect of un-selective `import` while also removing redundancy and improving convenience.

What do you think, please ?  (-:

std/container/array.d
```diff
@@ -24,6 +24,17 @@ import std.traits;
 
 public import std.container.util;
 
+pure @system unittest
+{
+    immutable(size_t) n = 100;
+    float[] a0; a0.length = n; a0[] = 1.41421f;
```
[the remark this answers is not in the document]
DoctorNoobingstoneIPresume: I felt that all statements really were part of the declaration-and-initialization of `a0`.

Anyway, I have changed to one statement per line. I am using braces as an IMO idiomatic hint (for the human readers and for IDE's which support collapsing-and-expanding brace-delimited scopes) that those statements are only there to initialize the thing defined immediately above the scope (`a0`) and do not affect the rest of the code => also make it easy to factor out the code in a separate function.

I hope that is okay.

DoctorNoobingstoneIPresume:
> Love seeing a red diff!

Thank you for constructive and encouraging review.

This PR is a continuation of https://github.com/dlang/phobos/pull/8143. Both PRs have been motivated by seeing repeated code, similar to Payload.reserve.

> But maybe someone more familiar with `Array` than me should check whether the duplicated code didn't have subtle differences that were important e.g. for performance.

Such checks (and all constructive reviews) are very welcome !

For now, I have tried to fix the performance problem spotted in your review (`insertBack` redundantly checking `length` vs `capacity`).

RazvanN7:
The buildkite failure might be fixed by a rebase.

std/container/array.d
```diff
@@ -24,6 +24,25 @@ import std.traits;
...
+        copy(iota(0, n), a0);
+        enum float factor = 1.1f;
+        a0[] *= factor;
+        assert(a0[$ - 1] == (n - 1) * factor);
+    }
+    auto a1 = Array!float(a0);
+    assert(a1.length == n);
+    assert(a1.capacity == n); // Checks that a1's constructor has called reserve.
+    import std.algorithm.comparison : equal;
+    assert(equal(a1[], a0[]));
```
dkorpel, line 43: Try to make unittests focussed and self-contained if you can. If it's only testing that "a1's constructor has called reserve" then the complex initialization of the float array is not needed:
```D
immutable int[4] a0 = [10, 20, 30, 40];
auto a1 = Array!int(a0[]);
assert(a1.length == a0.length);
assert(a1.capacity == a0.length); // Checks that a1's constructor has called reserve.
assert(a1[] == a0[]);
```

DoctorNoobingstoneIPresume:
> The buildkite failure might be fixed by a rebase.

Thank you. Let us see !

std/container/array.d
DoctorNoobingstoneIPresume: > Try to make unittests focussed and self-contained if you can. If it's only testing that "a1's constructor has called reserve" then the complex initialization of the float array is not needed:
> 
> ```d
> immutable int[4] a0 = [10, 20, 30, 40];
> auto a1 = Array!int(a0[]);
> assert(a1.length == a0.length);
> assert(a1.capacity == a0.length); // Checks that a1's constructor has called reserve.
> assert(a1[] == a0[]);
> ```

Unfortunately, for some lengths, the test passes even if reserve is not called (at least with the insertBack version, which is a correct version, even though it might be less efficient than the emplace version).

I have now added even more comments in the source code for the `unittest`, so that others do not fall in the same traps I have.

While the line count has increased, IMO it is quite readable by humans. And (I dare say) useful.

DoctorNoobingstoneIPresume:
I am very curious as to why `Data.init` cannot be used (https://github.com/dlang/phobos/pull/8162#discussion_r668377463).

This comment is just to inform that, currently, with my low skills and high laziness, it is going to take some time to investigate.

std/container/array.d
```diff
@@ -428,32 +481,28 @@ if (!is(immutable T == immutable bool))
...
+            {
+                // We do not simply write "_data.insertBack(value);"
+                // because that might perform, on each iteration, a now-redundant check of length vs capacity.
+                // Thanks to @dkorpel (https://github.com/dlang/phobos/pull/8162#discussion_r667479090).
+
+                import core.lifetime : emplace;
+                emplace(_data._payload.ptr + _data._payload.length, value);
+
+                // We increment the length after each iteration (as opposed to adjusting it just once, after the loop)
+                // in order to improve error-safety (in case one of the calls to emplace throws).
```
pbackus, line 499: So if one of the calls to `emplace` throws, the `Array` will be partially populated, with some unspecified `length` <= `values.length`? This does not seem like desirable behavior, and as far as I can tell is not how it worked before. Please preserve the original behavior here.

RazvanN7:
@DoctorNoobingstoneIPresume Any updates on this?

DoctorNoobingstoneIPresume:
> 
> 
> @DoctorNoobingstoneIPresume Any updates on this?

@RazvanN7 @pbackus I apologize for the delay. I have only today resumed investigations -- currently I am trying to better understand RefCounted. I hope to answer again soon.

std/container/array.d
DoctorNoobingstoneIPresume: @pbkacus: First, I apologize for long delay. Second:

For this case -- when one of the calls to `emplace` throws -- after *some* (but not *all*) calls to emplace have succeeded:

- the pre-existing behaviour is that the `length` remains unchanged -- just as if *no* elements have been added -- but IMO this is incorrect because *some* elements have been successfully added;

- the proposed behaviour is that `length` is increased with the number of elements which have been successfully added (indeed, a number less than `values.length`);

- but `length` should not be increased with `values.length`, because that would signify that *all* requested elements have been successfully added.

Am I mis-understanding anything, please ?

pbackus: Your understanding is correct. However, I disagree with your opinion that the pre-existing behavior is wrong.

If the user requests an array containing a specific list of values, we must either give them what they requested, or report failure via an exception. Giving them *part* of what they requested is not acceptable. Users must be able to rely on the standard library to do what they ask it to do.

It is also highly inappropriate to introduce a behavioral change like this in a PR whose stated purpose is to "reduce code duplication."

pbackus:
> i.e. what is it about `RefCounted ! (...)` which makes a constructor with an empty body differ in result from `.init`.

In my experience the most likely cause of this is a [nested struct][1].

Nested structs have a hidden context pointer that must be initialized at runtime. When you construct an instance using constructor syntax (e.g., `MyStruct()`), this initialization is performed. However, when you use `.init`, this initialization is skipped, and the context pointer is set to `null` instead. Here's an example:

```d
unittest
{
    int x = 1;
    
    struct Nested
    {
        int y = 2;
        int fun() { return x + y; } // access context
    }

    Nested a = Nested(); // use constructor syntax
    assert(a.fun() == 3); // ok
    
    Nested b = Nested.init; // use .init
    assert(b.fun() == 3); // Segmentation fault
}
```

The reason for this is that the `.init` value of a type is always evaluated at compile time, not runtime.

[1]: https://dlang.org/spec/struct.html#nested

std/container/array.d
DoctorNoobingstoneIPresume: > 
> 
> Your understanding is correct. However, I disagree with your opinion that the pre-existing behavior is wrong.
> 
> If the user requests an array containing a specific list of values, we must either give them what they requested, or report failure via an exception. Giving them _part_ of what they requested is not acceptable. Users must be able to rely on the standard library to do what they ask it to do.

I agree that this constructor should either successfully construct an array containing all requested elements or propagate the Exception.

I believe that this is done both by pre-existing code and by the proposed changeset.

However !!

I believe that the pre-existing code incorrectly tracks the number of (successfully) constructed elements during the loop in the constructor. The tracked number of elements is zero during the entire loop and is only updated after the loop -- that is: if the entire loop has been successfully completed.

But if the loop is not completed:

- yes, the `Exception` is propagated (the user is not misled into thinking that his/her request for construction has completed successfully) -- this is good (and is kept in the proposed changeset);

- no, the already-constructed elements are not destroyed (because they have not been tracked) -- this is bad (and is corrected by the proposed changeset).

I have just added a `unittest`. With the pre-existing code, the `unittest` fails.

(The pre-existing code is temporarily enabled with `static if (1)` at line 553 in the current patchset, i.e. in https://github.com/DoctorNoobingstoneIPresume/dlang_phobos/commit/cd46b4ef7e8a165bc6bf56cc0c0e3511fb89541e. This is just temporary so we can easily test together. If we change that to "static if (0)", the new code is enabled and the `unittest` passes.)

> It is also highly inappropriate to introduce a behavioral change like this in a PR whose stated purpose is to "reduce code duplication."

I agree that the commit message (and possibly the code) can be improved. I have just tried to modify the commit message now. I hope it is better.

DoctorNoobingstoneIPresume: > (The pre-existing code is temporarily enabled with `static if (1)` at line 553 in the current patchset, i.e. in [DoctorNoobingstoneIPresume@cd46b4e](https://github.com/DoctorNoobingstoneIPresume/dlang_phobos/commit/cd46b4ef7e8a165bc6bf56cc0c0e3511fb89541e). This is just temporary so we can easily test together: if we change that line to `static if (0)`, the new code is enabled and the `unittest` passes.)

I have just changed that line to `static if (0)` in order to enable the new code: https://github.com/DoctorNoobingstoneIPresume/dlang_phobos/commit/26ca49aeecb9d930adfe3849630348e3091bbc03.

BTW: In Github, is there a way to see all the commits that a proposed changeset has referred to during its evolution, please ?

DoctorNoobingstoneIPresume: > I have just added a `unittest`. With the pre-existing code, the `unittest` fails.

The `unittest` can be found by searching for `S.s_nDestroyed == S.s_nConstructed`.

pbackus: Thanks for investigating this. It looks like this is a bug in the pre-existing code, so the best way to handle it would be to file an issue on [issues.dlang.org][1], and submit the fix as a separate PR.

[1]: https://issues.dlang.org/

DoctorNoobingstoneIPresume: An issue has been added as https://issues.dlang.org/show_bug.cgi?id=22337.

Could you please let me know whether the commit message of the PR should be modified to specify the issue number ?

Regarding making a separate PR: In this case, I believe this is just extra work for little benefit. With the current PR, we get both a fixed bug and simpler-and-more-clear code (if we do not count the `unittest`'s and the comments). The simpler code (with fewer lines of code) helps us reason about the bug and the fix.

pbackus: The commit message for a bug fix should be formatted according to these instructions: <https://wiki.dlang.org/Contributing_to_Phobos#Issues>

Using separate PRs for bug fixes and refactors makes the changes easier to review and makes the git history of the code more granular (which makes it easier to `git bisect`, for example). In any case, it is D policy not to mix refactors and bug fixes; see here: <https://github.com/dlang/dmd/blob/master/CONTRIBUTING.md#refactoring>

DoctorNoobingstoneIPresume: Is this not already very easy to review ?

RazvanN7: How easy a PR is to review is a subjective matter. However, we do have objective guidelines and those are that refactorings and bug fixes should go to separate PRs. I suggest that you use this PR to do the refactoring and make a subsequent PR to fix the bug.

This matter has been extensively discussed and contributors occasionally have arguments against this policy, however, experience has shown that it is best to separate refactorings from bug fixes.

DoctorNoobingstoneIPresume: The "objective" guidelines referred above (https://github.com/dlang/dmd/blob/master/CONTRIBUTING.md#refactoring) state that:

"[...] Pull requests that do non-trivial refactorings should not include other changes, such as new feature or bug fix. [...]"

IMO, this "refactoring" is trivial.

PR #8163, opened 2021-07-11
Fix betterC unittest of SumType for issue 20150

dkorpel:
Fix for https://github.com/dlang/dmd/pull/12010#issuecomment-877849728

### DIFF
--- a/std/sumtype.d
+++ b/std/sumtype.d
@@ -993,6 +993,7 @@ version (D_BetterC) {} else
 }
 
 // const SumTypes
+version (D_BetterC) {} else // not @nogc, https://issues.dlang.org/show_bug.cgi?id=22117
 @safe unittest
 {
     auto _ = const(SumType!(int[]))([1, 2, 3]);


pbackus:
Currently, these attributes are inferred, which means that if `value`'s copy constructor escapes a reference to its argument, `value` can be inferred as non-`scope`. Adding an explicit `scope` annotation here would cause valid code to be rejected; for example:

```d
// Types with non-scope copy constructors
@safe unittest
{
    static int* globalPtr;
    static int globalInt;

    static struct S
    {
        @safe:
        int* p;
        this(int* p) { this.p = p; }
        this(ref S other)
        {
            globalPtr = other.p;
            this.p = other.p;
        }
    }

    auto s = S(&globalInt);
    SumType!S x = s;
}
```

This currently compiles with `-preview=dip1000`, but would not compile with these changes applied.

I think the correct approach here is to first remove the workaround for [issue 21229][1] (which has since been fixed), since that's what's currently preventing `value` from being inferred as `scope`. Hopefully, once that's done, no explicit annotation will be necessary.

[1]: https://issues.dlang.org/show_bug.cgi?id=21229

dkorpel:
> I think the correct approach here is to first remove the workaround for [issue 21229](https://issues.dlang.org/show_bug.cgi?id=21229) (which has since been fixed)

Where do I find this workaround?

Also, do you mind if I change the unittest to use a global (non-scope) array in the meantime? The test is called 'const SumTypes', having `scope` inferred on the constructors doesn't look integral to the test.

pbackus:
> Where do I find this workaround?

Oops, looks like I forgot to add the comment marking it in the Phobos version. It's the whole `mixin("Storage newStorage = { "` bit in the constructors. I can fix it myself in a separate PR if that would be easier.

> Also, do you mind if I change the unittest to use a global (non-scope) array in the meantime? The test is called 'const SumTypes', having scope inferred on the constructors doesn't look integral to the test.

 Sure, that's fine. Or leave it as a heap-allocated array and `version` it out in BetterC.

pbackus:
Turns out the constructor workaround is still required because the fix for issue 21229 was incomplete, leaving behind [issue 22118](https://issues.dlang.org/show_bug.cgi?id=22118). So, no `scope` for `SumType` just yet.

pbackus (approved):
LGTM

PR #8164, opened 2021-07-11
Fix issue 18632 - enable use of fromStringz with char[n]

dkorpel:


RazvanN7 (approved):

RazvanN7:
LGTM

RazvanN7:
One observation is that `fromStringz` was originally designed to convert a C string (null terminated char*) to a D string (char[]). Since a static array is already a D string, it can be argued that allowing the case that this PR enables is pointless. However, I think this enables the use of fromStringz in generic code where it is not obvious whether the string is a C string or a D string. As such, I don't have anything against this addition, however, I'll wait for others to weight in.

dkorpel:
> Since a static array is already a D string, it can be argued that allowing the case that this PR enables is pointless.

The use case I have in mind it C api's using structures with fixed size, e.g:
```D
struct WIN32_FIND_DATAW {
    DWORD           dwFileAttributes;
    FILETIME        ftCreationTime;
    // (...)
    WCHAR[MAX_PATH] cFileName = 0;
}
```
With the new overload, you can safely extract `cFileName.fromStringz` instead of `cFileName.ptr.fromStringz`.

Geod24:
Additionally, some C API accept a pointer and a length and will write to that. E.g. `getcwd`. In which case, one might want to use a static array to avoid GC allocations.

PR #8166, opened 2021-07-14
Fix Issue 22125 - std.process.Config was changed to a struct but miss…

CyberShadow:
… operator overloads


CyberShadow:
Oops, stable

PR #8167, opened 2021-07-14
Fix Issue 22125 - std.process.Config was changed to a struct but miss…

CyberShadow:
… operator overloads


MoonlightSentinel:
The bug report also mentioned the other binary operators which are missing from this PR?

CyberShadow:
Well, yes. Some binary operators make more sense than others. It's hard to think of a use case for `^` (XOR) or `^^` (exponentiation). But, we might as well use metaprogramming to support everything.

RazvanN7 (approved):

PR #8175, opened 2021-07-22
[DSSv3] Fix Issue 13684 - std.numeric.isRectangular

lucica28:
Added new function isRectangular

### DIFF
--- a/std/numeric.d
+++ b/std/numeric.d
@@ -1121,6 +1121,57 @@ T findRoot(T, DF)(scope DF f, const T a, const T b)
     return findRoot(f, a, b, (T a, T b) => false);
 }
 
+bool isRectangular(T)(T arr)
+if (is(typeof(arr[0])) && is(typeof(arr[0][0])))
+{
+    import std.array: join;
+
+    for (int i = 0; i < arr.length - 1; i++)
+        if(arr[i].length != arr[i + 1].length) return false;
+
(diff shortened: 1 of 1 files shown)


std/numeric.d
```diff
@@ -1121,6 +1121,57 @@ T findRoot(T, DF)(scope DF f, const T a, const T b)
     return findRoot(f, a, b, (T a, T b) => false);
 }
 
+bool isRectangular(T)(T arr)
```
thewilsonator, line 1124: Needs documentation

thewilsonator (commented):
Retitle the commit message to say "Fix issue 13684 - add std.numeric.isRectangular"

RazvanN7:
The commit message should contain the string "Fix Issue 13684 - ..." so that the bot picks it up. Upon merger, the bot  will then automatically close the issue.

kinke:
I don't think Phobos needs such a function; it may encourage people to think jagged arrays (arrays of arrays) are a reasonable way for such rectangular multidimensional arrays, while we have a much better suited library for that: http://mir-algorithm.libmir.org/mir_ndslice.html

Using `join()` for the recursion and thus requiring potentially huge temporary buffers for such a check is bad.

RazvanN7:
@lucica28 As per @kinke 's comment I am going to close this PR. Feel free to close the issue as WONTFIX.

PR #8176, opened 2021-07-23
Fix Issue #22140: FunctionTypeOf fails for template opCall without any templated argument

ljmf00:

### DIFF
--- a/std/traits.d
+++ b/std/traits.d
@@ -2279,15 +2279,18 @@ usually used for compile-time reflection purposes.
 template FunctionTypeOf(func...)
 if (func.length == 1 && isCallable!func)
 {
-    static if (is(typeof(& func[0]) Fsym : Fsym*) && is(Fsym == function) || is(typeof(& func[0]) Fsym == delegate))
+    static if ((is(typeof(& func[0]) Fsym : Fsym*) && is(Fsym == function)) || is(typeof(& func[0]) Fsym == delegate))
     {
         alias FunctionTypeOf = Fsym; // HIT: (nested) function symbol
     }
-    else static if (is(typeof(& func[0].opCall) Fobj == delegate))
+    else static if (is(typeof(& func[0].opCall) Fobj == delegate) || is(typeof(& func[0].opCall!()) Fobj == delegate))
     {
         alias FunctionTypeOf = Fobj; // HIT: callable object
     }
-    else static if (is(typeof(& func[0].opCall) Ftyp : Ftyp*) && is(Ftyp == function))
+    else static if (
+            (is(typeof(& func[0].opCall) Ftyp : Ftyp*) && is(Ftyp == function)) ||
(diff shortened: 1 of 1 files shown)


std/traits.d
```diff
@@ -2279,15 +2279,18 @@ usually used for compile-time reflection purposes.
 template FunctionTypeOf(func...)
 if (func.length == 1 && isCallable!func)
 {
-    static if (is(typeof(& func[0]) Fsym : Fsym*) && is(Fsym == function) || is(typeof(& func[0]) Fsym == delegate))
+    static if ((is(typeof(& func[0]) Fsym : Fsym*) && is(Fsym == function)) || is(typeof(& func[0]) Fsym == delegate))
     {
         alias FunctionTypeOf = Fsym; // HIT: (nested) function symbol
     }
-    else static if (is(typeof(& func[0].opCall) Fobj == delegate))
+    else static if (is(typeof(& func[0].opCall) Fobj == delegate) || is(typeof(& func[0].opCall!()) Fobj == delegate))
```
ljmf00, line 2286: Can someone provide me with an example of where this is applicable?

std/traits.d
```diff
@@ -2279,15 +2279,18 @@ usually used for compile-time reflection purposes.
 template FunctionTypeOf(func...)
 if (func.length == 1 && isCallable!func)
 {
-    static if (is(typeof(& func[0]) Fsym : Fsym*) && is(Fsym == function) || is(typeof(& func[0]) Fsym == delegate))
+    static if ((is(typeof(& func[0]) Fsym : Fsym*) && is(Fsym == function)) || is(typeof(& func[0]) Fsym == delegate))
```
RazvanN7, line 2282: This change is unnecessary

RazvanN7 (changes requested):
Needs tests.

std/traits.d
RazvanN7:
```d
void fun()
{
    int a;
    struct B { int opCall() { return a; }}
}
```
}

RazvanN7:
ping @ljmf00

std/traits.d
ljmf00: It is necessary for clarity. I based my change on DScanner suggestions. If you prefer I can revert it, but the rationale is that it clarifies the logic precedence of the expression.

https://github.com/dlang-community/D-Scanner/blob/2963358eb4a24064b0893493684d4075361297eb/src/dscanner/analysis/logic_precedence.d#L45

std/traits.d
ljmf00: Done

ljmf00:
> ping @ljmf00

Thanks for the ping :+1:

RazvanN7 (approved):

RazvanN7:
Thanks @ljmf00 !

PR #8179, opened 2021-07-25
[DSSv3] Fix Issue 14250 - add function for reading a password

teodutu:
A new module named `getpass` was added, which currently exposes a single
homonymous function. This function reads and returns user input from the
current terminal without echoing it.

As suggested by @MartinNowak, the `getpass` function was placed in a
separate module as its purpose is differs from `getopt`.


pbackus:
Adding a new module to Phobos is a big commitment. This seems like it would be much better off as a package on code.dlang.org, at least to start with.

ibuclaw:
Why?

PR #8181, opened 2021-07-26
[DSSv3] Fix issue - 16339

lucica28:
std.array.insertInPlace: added tests for class and union.

### DIFF
--- a/std/array.d
+++ b/std/array.d
@@ -1364,7 +1364,7 @@ if (isSomeString!(T[]) && allSatisfy!(isCharOrStringOrDcharRange, U))
 }
 
 ///
-@system unittest
+@safe pure unittest
 {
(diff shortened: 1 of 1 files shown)


std/array.d
```diff
@@ -1364,13 +1364,46 @@ if (isSomeString!(T[]) && allSatisfy!(isCharOrStringOrDcharRange, U))
 }
 
 ///
-@safe pure unittest
+@system unittest
```
RazvanN7, line 1367: Before this addition, the unittest tested that the code is both @safe and pure. With this addition the first paragraph no longer tests that. Ideally, the added tests should be pure and @safe; if that is not possible, then just create a new unittest block that is @system. (from looking at the code, I don't see why purity is lost; also, the class example should be @safe)

RazvanN7 (approved):

PR #8182, opened 2021-07-29
[DSSv3] Fix issue 20865 - add function isSubPath

lucica28:
Fix issue 20865 - added function isSubPath and tests

### DIFF
--- a/std/path.d
+++ b/std/path.d
@@ -3912,28 +3912,29 @@ if (isConvertibleToString!Range)
   *     true if `path` is a subpath of `path2`
   *     false otherwise
 */
-bool isSubPath(Range)(Range path1, Range path2)
+bool isSubPath(Range)(Range subPath, Range path)
 if ((isRandomAccessRange!Range && hasLength!Range && hasSlicing!Range &&
     isSomeChar!(ElementEncodingType!Range) || isNarrowString!Range) &&
     !isConvertibleToString!Range)
 {
-    import std.algorithm.searching : startsWith, endsWith;
-    import std.string : indexOf, chompPrefix, chop;
+    import std.array : array;
+    
+    if (!isValidPath(subPath) || !isValidPath(path)) return false;
 
-    if (!isValidPath(path1) || !isValidPath(path2)) return false;
+    auto subPathAbsolute = asAbsolutePath(subPath).array;
+    auto pathAbsolute = asAbsolutePath(path).array;
+    subPathAbsolute = buildNormalizedPath(subPathAbsolute);
+    pathAbsolute = buildNormalizedPath(pathAbsolute);
 
-    if (path1.endsWith("/")) path1 = chop(path1);
-    if (path1.endsWith("\\")) path1 = chop(path1);
-    if (path1.startsWith("~")) path1 = expandTilde(path1);
-    if (path2.startsWith("~")) path2 = expandTilde(path2);
-    if (!isAbsolute(path1)) path1 = absolutePath(path1);
-    if (!isAbsolute(path2)) path2 = absolutePath(path2);
-    path1 = buildNormalizedPath(path1);
-    path2 = buildNormalizedPath(path2);
+    auto subPathComponents = pathSplitter(subPathAbsolute).array;
+    auto pathComponents = pathSplitter(pathAbsolute).array;
 
-    return (path2.startsWith(path1) &&
-            indexOf(chompPrefix(path2, path1), '/') == 0 ||
-            indexOf(chompPrefix(path2, path1), '\\') == 0);
+    if (subPathComponents.length > pathComponents.length) return false;
+
+    for (uint i = 0; i < subPathComponents.length; i++)
+        if (!(subPathComponents[i] == pathComponents[i])) return false;   
+
+    return true;
 }
 
 @safe unittest
@@ -3942,14 +3943,12 @@ if ((isRandomAccessRange!Range && hasLength!Range && hasSlicing!Range &&
     {
         assert(isSubPath("/foo/bar","/foo/bar/some"));
         assert(!isSubPath("/foo/bar/some/file/..", "/foo/bar"));
-        assert(isSubPath("~/foo", "~/foo/bar/some/file"));
         assert(!isSubPath("/foo/bar", "/foo/bars"));
     }
 
     version (Windows)
     {
         assert(isSubPath("c:/foo/", "c:/foo/bar"));
-        assert(isSubPath("c:/foo", "c:/foo/bar"));
         assert(isSubPath("c:/foo/..", "c:/foo/bar/../."));
         assert(!isSubPath("c:/foo", "c:/foobar"));
     }
--- a/std/path.d
+++ b/std/path.d
@@ -3902,6 +3902,48 @@ if (isConvertibleToString!Range)
     assert(testAliasedString!isValidPath("/foo/bar"));
 }
 
+/** Checks whether `path1` is a subpath of `path2`
+  *
+  * Params:
+  *     path1 = string or Range of characters
+  *     path2 = string or Range of characters
+  *
+  * Returns:
+  *     true if `path` is a subpath of `path2`
+  *     false otherwise
+*/
+bool isSubPath(Range)(Range path1, Range path2)
+if ((isRandomAccessRange!Range && hasLength!Range && hasSlicing!Range &&
+    isSomeChar!(ElementEncodingType!Range) || isNarrowString!Range) &&
+    !isConvertibleToString!Range)
+{
+    import std.algorithm.searching : startsWith;
+    if (path1.startsWith("~")) path1 = expandTilde(path1);
+    if (path2.startsWith("~")) path2 = expandTilde(path2);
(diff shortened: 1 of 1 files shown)


std/path.d
```diff
@@ -3902,6 +3902,48 @@ if (isConvertibleToString!Range)
...
+  *     false otherwise
+*/
+bool isSubPath(Range)(Range path1, Range path2)
+if ((isRandomAccessRange!Range && hasLength!Range && hasSlicing!Range &&
+    isSomeChar!(ElementEncodingType!Range) || isNarrowString!Range) &&
+    !isConvertibleToString!Range)
+{
+    import std.algorithm.searching : startsWith;
+    if (path1.startsWith("~")) path1 = expandTilde(path1);
+    if (path2.startsWith("~")) path2 = expandTilde(path2);
```
CyberShadow, line 3922: None of the other functions in `std.path` do implicit tilde expansion, please remove

std/path.d
```diff
@@ -3902,6 +3902,48 @@ if (isConvertibleToString!Range)
...
+    if (!isAbsolute(path1)) path1 = absolutePath(path1);
+    if (!isAbsolute(path2)) path2 = absolutePath(path2);
```
CyberShadow, line 3924: I don't think this will work with non-string arguments.

std/path.d
```diff
@@ -3902,6 +3902,48 @@ if (isConvertibleToString!Range)
...
+  *
+  * Params:
+  *     path1 = string or Range of characters
+  *     path2 = string or Range of characters
+  *
+  * Returns:
+  *     true if `path` is a subpath of `path2`
+  *     false otherwise
+*/
+bool isSubPath(Range)(Range path1, Range path2)
```
CyberShadow, line 3915: Please use more descriptive parameter names than `path1` and `path2`.

Please also consider UFCS. In `path1.isSubPath(path2)`, is it immediately obvious which argument is the prefix and which is the longer path? Is a better function name and argument order possible?

CyberShadow (changes requested):
Bugzilla issues are not gospel, they may need adjustments before the enhancement they describe is suitable for addition.

CyberShadow:
For reference, here is my implementation of this function from ae:

https://github.com/CyberShadow/ae/blob/3ddbf59cf4e6bea3a59d132ad813773b1ea8dcdc/utils/path.d#L78-L96

std/path.d
```diff
@@ -3902,6 +3902,48 @@ if (isConvertibleToString!Range)
...
+{
+    import std.algorithm.searching : startsWith;
+    if (path1.startsWith("~")) path1 = expandTilde(path1);
+    if (path2.startsWith("~")) path2 = expandTilde(path2);
+    if (!isAbsolute(path1)) path1 = absolutePath(path1);
+    if (!isAbsolute(path2)) path2 = absolutePath(path2);
+    path1 = buildNormalizedPath(path1);
+    path2 = buildNormalizedPath(path2);
+
+    return (path2.startsWith(path1));
```
CyberShadow, line 3928: I think this is wrong, as `isSubPath("/ab", "/a")` will return true.

lucica28: @CyberShadow could you please take a look at the last commit? I tried to cover your recommendations.

std/path.d
```diff
@@ -3902,6 +3902,58 @@ if (isConvertibleToString!Range)
     assert(testAliasedString!isValidPath("/foo/bar"));
 }
 
+/** Checks whether `path1` is a subpath of `path2`
+  *
+  * Params:
+  *     path1 = string or Range of characters
+  *     path2 = string or Range of characters
```
CyberShadow, line 3909: These still refer to the old names.

std/path.d
```diff
@@ -3902,6 +3902,58 @@ if (isConvertibleToString!Range)
...
+  *
+  * Returns:
+  *     true if `path` is a subpath of `path2`
+  *     false otherwise
+*/
+bool isSubPath(Range)(Range subPath, Range path)
```
CyberShadow, line 3915: May I suggest the name `pathStartsWith`, for consistency with `std.algorithm.startsWith`.

The order of the arguments would then be inverted.

Or, is there a reason why this makes more sense that I'm not seeing?

std/path.d
```diff
@@ -3902,6 +3902,58 @@ if (isConvertibleToString!Range)
...
+    
+    if (!isValidPath(subPath) || !isValidPath(path)) return false;
+
+    auto subPathAbsolute = asAbsolutePath(subPath).array;
+    auto pathAbsolute = asAbsolutePath(path).array;
+    subPathAbsolute = buildNormalizedPath(subPathAbsolute);
+    pathAbsolute = buildNormalizedPath(pathAbsolute);
+
+    auto subPathComponents = pathSplitter(subPathAbsolute).array;
+    auto pathComponents = pathSplitter(pathAbsolute).array;
```
CyberShadow, line 3930: I'm sure that it's possible to implement this function with no allocations.

std/path.d
```diff
@@ -3902,6 +3902,58 @@ if (isConvertibleToString!Range)
...
+  *     false otherwise
+*/
+bool isSubPath(Range)(Range subPath, Range path)
+if ((isRandomAccessRange!Range && hasLength!Range && hasSlicing!Range &&
+    isSomeChar!(ElementEncodingType!Range) || isNarrowString!Range) &&
+    !isConvertibleToString!Range)
+{
+    import std.array : array;
+    
+    if (!isValidPath(subPath) || !isValidPath(path)) return false;
```
CyberShadow, line 3922: Why is this necessary?

std/path.d
```diff
@@ -3902,6 +3902,58 @@ if (isConvertibleToString!Range)
...
+
+    auto subPathAbsolute = asAbsolutePath(subPath).array;
+    auto pathAbsolute = asAbsolutePath(path).array;
+    subPathAbsolute = buildNormalizedPath(subPathAbsolute);
+    pathAbsolute = buildNormalizedPath(pathAbsolute);
```
CyberShadow, line 3927: Are all these transformations necessary?

`asAbsolutePath`, in particular, makes the result depend on the current directory (which also precludes this function from being `pure`). I think this (whether to expand or not relative paths based on the current directory) could go either way, but, we ought to rationalize the decision.

std/path.d
```diff
@@ -3902,6 +3902,58 @@ if (isConvertibleToString!Range)
...
+        assert(!isSubPath("/foo/bar", "/foo/bars"));
+    }
+
+    version (Windows)
+    {
+        assert(isSubPath("c:/foo/", "c:/foo/bar"));
+        assert(isSubPath("c:/foo/..", "c:/foo/bar/../."));
+        assert(!isSubPath("c:/foo", "c:/foobar"));
+    }
+}
```
CyberShadow, line 3955: - Please add tests where the paths are a range.
- Please add tests where both paths are relative.
- Please add tests where only one of the paths is relative, and the result depends on the current directory (assuming that we decide to keep `asAbsolutePath`).

CyberShadow (commented):
This one might be a bit tricky for a DSS task. There's also a bunch of subjective questions which may require more deliberation.

std/path.d
CyberShadow: My recommendation is to require that path expansion is explicitly done on the caller side.

Instead, do this:
- If both paths are absolute, just compare the prefix (plus dir separator).
- If both paths are relative, just compare the prefix (plus dir separator).
- If only one of the paths is relative, assert.

RazvanN7:
@atilaneves Is it worthwhile to add such a function in phobos?

atilaneves:
> @atilaneves Is it worthwhile to add such a function in phobos?

I don't think so, no. The bug report doesn't list a reason for why this is needed, and even if it were super useful I'd call is `isSubPathOf` and reverse the order of the parameters. Nobody is going to remember what should come first otherwise.

andre2007:
@atilaneves 
Sorry, yes the reason is missing in the bug report.

In the context of security I need this function to check for path traversal attacks.
It is not easy to get it right, and therefore I thought it would be quite useful to have it in Phobos.

https://owasp.org/www-community/attacks/Path_Traversal

atilaneves:
> @atilaneves
> Sorry, yes the reason is missing in the bug report.
> 
> In the context of security I need this function to check for path traversal attacks.
> It is not easy to get it right, and therefore I thought it would be quite useful to have it in Phobos.
> 
> https://owasp.org/www-community/attacks/Path_Traversal

And is it right in this PR? Otherwise: can you not write the function yourself in your project or put it up on dub?

andre2007:
Having a dub package with just one function `isSubPath` would maybe lead to the same issues NPM is facing with packages like isNumber, isOdd, isEven, ...

My gut feeling is, the only right place for such a function is std.path. But of course I can also implement it in my local packages too.

I do not care that much about the naming, as long as I can use it to check wheter a given path is a sub path of another path. The question regarding the naming I would leave to you, the Phobos experts...

RazvanN7:
@lucica28 Please reopen this if you are willing to take to the finish line and @atilaneves accepts the addition.

PR #8183, opened 2021-08-01
[DSSv3] Fix Issue 16210 - std.utf.byUTF can be made into a bidirectional range

vladchicos:
I also added unittests and an example.

### DIFF
--- a/std/utf.d
+++ b/std/utf.d
@@ -4383,24 +4383,24 @@ if (isSomeChar!C)
 
                 static if (isBidirectionalRange!R)
                 {
-                    @property dchar back() scope // 'scope'' required by call to decodeBack() below
+                    @property dchar back() scope // 'scope' required by call to decodeBack() below
                     {
-                        if (backBuff == Empty)
+                        if (backBuff != Empty)
+                            return cast(dchar) backBuff;
+
+                        auto c = r.back;
+                        static if (is(RC == wchar))
+                            enum firstMulti = 0xD800; // First high surrogate.
+                        else
+                            enum firstMulti = 0x80; // First non-ASCII.
+                        if (c < firstMulti)
                         {
-                            auto c = r.back;
-                            static if (is(RC == wchar))
-                                enum firstMulti = 0xD800; // First high surrogate.
-                            else
-                                enum firstMulti = 0x80; // First non-ASCII.
-                            if (c < firstMulti)
-                            {
-                                r.popBack;
-                                backBuff = cast(dchar) c;
-                            }
-                            else
-                            {
-                                backBuff = () @trusted { return decodeBack!useReplacementDchar(r); }();
-                            }
+                            r.popBack;
+                            backBuff = cast(dchar) c;
+                        }
+                        else
+                        {
+                            backBuff = () @trusted { return decodeBack!useReplacementDchar(r); }();
                         }
                         return cast(dchar) backBuff;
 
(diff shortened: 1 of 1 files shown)


std/utf.d
```diff
@@ -4369,10 +4381,45 @@ if (isSomeChar!C)
                     }
                 }
 
+                static if (isBidirectionalRange!R)
+                {
+                    @property dchar back() scope // 'scope'' required by call to decodeBack() below
+                    {
+                        if (backBuff == Empty)
```
thewilsonator, line 4388:
```
if (backBuff != Empty)
    return cast(dchar) backBuff;
...
return cast(dchar) backBuff;

File: std/utf.d [hunk 2]
```diff
@@ -4446,10 +4510,53 @@ if (isSomeChar!C)
                     }
                 }
 
+                static if (isBidirectionalRange!R)
+                {
+                    @property auto back() scope // 'scope' required by call to decodeBack() below
+                    {
+                        if (backPos == backFill)
```
thewilsonator, line 4517: ditto

thewilsonator (commented):
please use early returns

std/utf.d
vladchicos: I think I solved it.

RazvanN7 (approved):

std/utf.d
```diff
@@ -4278,9 +4278,9 @@ private int impureVariable;
  *      Does not use GC if `useReplacementDchar` is set to `UseReplacementDchar.no`
  *
  * Returns:
- *      A forward range if `R` is a range and not auto-decodable, as defined by
+ *      A bidirectional range if `R` is a range and not auto-decodable, as defined by
  *      $(REF isAutodecodableString, std, traits), and if the base range is
- *      also a forward range.
+ *      also a bidirectional range.
```
RazvanN7, line 4283: You are omitting the fact that if the underlying range is a forward range, a forward range is returned.

std/utf.d
```diff
@@ -4325,10 +4325,22 @@ if (isSomeChar!C)
                     this.buff = buff;
                 }
 
+                static if (isBidirectionalRange!R)
+                {
+                    this(return R r, uint frontBuff, uint backBuff)
+                    {
+                        this.r = r;
+                        this.buff = frontBuff;
+                        this.backBuff = backBuff;
```
RazvanN7, line 4334: Is this constructor needed? I don't see any calls to it. Also, it seems that it is not covered by tests.

std/utf.d
```diff
@@ -4394,9 +4441,26 @@ if (isSomeChar!C)
                     this.buf = buf;
                 }
 
+                static if (isBidirectionalRange!R)
+                {
+                    this(return R r, ushort frontPos, ushort frontFill,
+                         ushort backPos, ushort backFill, C[4 / C.sizeof] buf)
```
RazvanN7, line 4447: Again, I don't see any calls to this constructor. Is this needed?

RazvanN7 (changes requested):

RazvanN7:
ping @vladchicos

RazvanN7:
This looks like a rebase screw up.

std/utf.d
vladchicos: @RazvanN7  ,sorry for the delay and thanks for pointing it out. Right now the constructor is not used, but I discovered a problem in the meantime.
Consider this piece of code
```
wchar[] s = ['Ă', 'ț'];
auto r = s.byUTF!char;
r.popBack;
auto rCopy = r.save
//right now r == [0xc4, 0x82, 0xc8];
//and rCopy == [0xc4, 0x82];
```
It is fixable by having different save functions for forward and bidirectional ranges.
(And this way I will need the new constructor).
Is it ok if I do this, right?
I also pushed a commit to see what I mean, now I have to fix the rebasing.

std/utf.d
vladchicos: @RazvanN7 fixed it.

std/file.d
```diff
@@ -402,24 +402,24 @@ version (Posix) private void[] readImpl(scope const(char)[] name, scope const(FS
...
                               SECURITY_ATTRIBUTES *lpSecurityAttributes, DWORD dwCreationDisposition,
-                              DWORD dwFlagsAndAttributes, HANDLE hTemplateFile) @trusted
+                              DWORD dwFlagsAndAttributes, HANDLE hTemplateFile)
     {
         return CreateFileW(namez, dwDesiredAccess, dwShareMode,
                            lpSecurityAttributes, dwCreationDisposition,
                            dwFlagsAndAttributes, hTemplateFile);
 
     }
-    static trustedCloseHandle(HANDLE hObject) @trusted
```
RazvanN7, line 418: This is not right.

std/utf.d
RazvanN7: I see, so the saved back buff is not considered in the initial version of `save`.

std/utf.d
```diff
@@ -4361,18 +4374,60 @@ if (isSomeChar!C)
                     buff = Empty;
                 }
 
-                static if (isForwardRange!R)
+                static if (isBidirectionalRange!R)
```
RazvanN7, line 4377: you could move the `static if` inside the save function:

```d
static if (isForwardRange!R)
{
    auto save()
    {
         static if (isBidirectionalRange!R) { ... }
         else  { ... }
    }

vladchicos: @RazvanN7 I changed it, I fixed the rebase and I added an unittest for the save() function. I'm sorry for the rebasing mistake.

RazvanN7 (approved):

PR #8184, opened 2021-08-01
[DSSv3] Fix Issue 16218 - Windows std.file.readImpl should be marked @system

vladchicos:
I marked readImpl with @system because most of its code was already @trusted.

### DIFF
--- a/std/file.d
+++ b/std/file.d
@@ -402,24 +402,24 @@ version (Posix) private void[] readImpl(scope const(char)[] name, scope const(FS
 
 
 version (Windows) private void[] readImpl(scope const(char)[] name, scope const(FSChar)* namez,
-                                          size_t upTo = size_t.max) @safe
+                                          size_t upTo = size_t.max) @system
 {
     import core.memory : GC;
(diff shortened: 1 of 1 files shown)


RazvanN7:
I think  that rebasing might fix the current tester failures.

RazvanN7:
I'm kind of surprised that this function is not called from @safe code in phobos.

std/file.d
```diff
@@ -402,24 +402,24 @@ version (Posix) private void[] readImpl(scope const(char)[] name, scope const(FS
 
 
 version (Windows) private void[] readImpl(scope const(char)[] name, scope const(FSChar)* namez,
-                                          size_t upTo = size_t.max) @safe
+                                          size_t upTo = size_t.max) @system
```
RazvanN7, line 405: Since the posix version is @trusted, I would rather have this function be @trusted also.

vladchicos: @RazvanN7, I changed it. Now it seems to pass the tests.

RazvanN7 (approved):

